PR #39, opened 2016-04-21
Remove default service name and appId from default install, ref. DCOS-6411

mlunoe:
Before:

After:

### DIFF
--- a/src/js/components/modals/InstallPackageModal.js
+++ b/src/js/components/modals/InstallPackageModal.js
@@ -1,5 +1,5 @@
 import classNames from 'classnames';
-import {Form, Modal} from 'reactjs-components';
+import {Modal} from 'reactjs-components';
 import mixin from 'reactjs-mixin';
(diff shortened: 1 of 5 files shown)


src/js/components/modals/InstallPackageModal.js
```diff
@@ -462,10 +412,9 @@ class InstallPackageModal extends mixin(InternalStorageMixin, TabsMixin, StoreMi
       <div>
         <ReviewConfig
           packageIcon={cosmosPackage.getIcons()['icon-small']}
-          packageType={name}
-          packageName={appId}
+          packageName={name}
           packageVersion={version}
-          configuration={configuration} />
+          configuration={this.getPackageConfiguration()} />
```
kennyt: isn't `getPackageConfiguration` always returning an empty object at the moment?

kennyt:
i see nothing wrong with this PR 👍

PR #40, opened 2016-04-22
DCOS-5935 Added storemixin to vanilla objects to avoid duplicating constants

MatApple:


plugins/oauth/hooks.js
```diff
@@ -16,26 +13,9 @@ let SDK = require('./SDK').getSDK();
...
-  CONFIG_STORE_LISTENERS.forEach(event => {
-    ConfigStore.addChangeListener(event, handler);
-  });
-
-  ConfigStore.fetchConfig();
-}
+let configResponseCallback = null;
 
-module.exports = {
+module.exports = _.extend({
```
kennyt: Let's use `Object.assign` here and remove underscore

mesosphere-web:
Reviewer: @rcorral

plugins/oauth/hooks.js
```diff
@@ -55,6 +35,10 @@ module.exports = {
     this.actions.forEach(action => {
       SDK.Hooks.addAction(action, this[action].bind(this));
     });
+    this.store_initializeListeners([{
```
mlunoe: Shouldn't this be `store_listeners`?

mlunoe: Oh! Disregard that! Didn't see it was not a component

mlunoe:
Aside from @kennyt's comment I think this looks good! 🚢

plugins/oauth/hooks.js
rcorral: +1

rcorral: it would probably also be easier to read if you did `Object.assign({}, StoreMixin, {other,stuff})`
Rather than having `StoreMixin` at the bottom of the page.

MatApple: Great suggestion! Done.

rcorral:
Rebase and then 🚢

PR #41, opened 2016-04-22
Make the FilterButtons filterByKey prop optional

orlandohohmeier:
Refactor the `FilterButtons` component, to make the `filterByKey` prop optional, as the item list could also be a plain list of values.

### DIFF
--- a/src/js/components/__tests__/FilterButtons-test.js
+++ b/src/js/components/__tests__/FilterButtons-test.js
@@ -46,19 +46,29 @@ describe('FilterButtons', function () {
     });
   });
 
-  describe('#getCountByKey', function () {
+  describe('#getCount', function () {
+
+    beforeEach(function () {
+      this.itemList = [
+        'f0',
+        'f0',
+        'f1'
+      ];
+
+    });
+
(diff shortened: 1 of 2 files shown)


src/js/components/__tests__/FilterButtons-test.js
```diff
@@ -46,19 +46,29 @@ describe('FilterButtons', function () {
...
 
     it('returns a hash map with only key "all" if no items given', function () {
-      var counts = this.instance.getCountByKey([], this.key);
+      var counts = this.instance.getCount([], this.key);
       expect(counts).toEqual({all: 0});
     });
 
     it('creates a hash map of filter counts', function () {
-      var counts = this.instance.getCountByKey(this.itemList, this.key);
+      var counts = this.instance.getCount(this.itemList, this.key);
```
mlunoe: How is this and the one above supposed to work with key? I think you are overriding `this.itemList` in the `beforeEach` you introduced and `getCount` does not take a key anymore

orlandohohmeier: 💩  Those slipped. Sorry. I'll fix the leftovers and push a fix.

PR #6, opened 2016-04-19
Introduce filter-bar styles

daiyiwhale:
Creating specific styles for filter-bar situations.

PR #46, opened 2016-04-22
Add Next/Prev buttons for logs

kennyt:
Using styles created here #6 

px


mesosphere-web:
Reviewer: @rcorral

src/styles/components/buttons.less
```diff
@@ -35,6 +35,28 @@ components/buttons.less
...
+
+    &-down {
+      &:after {
+        content: '';
+        border-top: 6px solid color-lighten(@neutral, 10);
+        border-left: 5px solid transparent;
+        border-right: 5px solid transparent;
+        vertical-align: middle;
+        display: inline-block;
+      }
```
jfurrow: Let's combine this with the above so that we're not repeating the same styles. Also :abcd:
```css
&-up,
&-before {
  &:after {
    // Shared styles
  }
}
```

jfurrow:
Do we have to apply an inline style with this search plugin? It already has a class, it would be nice if we could assign the background color in our CSS file. Also I don't think `red` is the right color.

jfurrow:
I'm noticing a couple of issues:
* It's jarring how the text box moves location when you start typing into it. I think it should stay in the same location, though I realize the space is limited. Let's get a designer's input here.
* It's super laggy when searching large logs, I think we should debounce `handleFilterChange`.
* There seems to be a bug with the number of search results found:

kennyt:
@jfurrow sorry that style tag was in an outdated version of the code. fixed now.

i've fixed all bullet points except for the web worker

src/js/components/FilterInputText.js
```diff
@@ -93,8 +95,8 @@ var FilterInputText = React.createClass({
     let inputContainerClasses = classNames({
       'form-control form-control-group filter-input-text-group': true,
       'form-control-inverse': props.inverseStyle,
-      'focus': focus
-    });
+      'focus': focus,
```
mlunoe: Don't think we need that trailing comma

src/js/components/TaskDebugView.js
```diff
@@ -1,24 +1,70 @@
...
+    let filterInput = ReactDOM.findDOMNode(this.refs.filterInput);
 
+    if (filterInput) {
+      filterInput.removeEventListener('keydown', this.handleKeyDown);
+    }
+  }
+
+  handleKeyDown(event) {
+    let {keyCode} = event;
+    if (keyCode === 13) {
```
mlunoe: We have a keyboard util, let's add this code to the map

src/js/components/TaskDebugView.js
```diff
@@ -1,24 +1,70 @@
...
+  }
+
+  componentWillReceiveProps(nextProps, nextState) {
+    let {searchString, totalFound} = nextState;
+    let updatedState = {};
+
+    if (totalFound === 0) {
+      updatedState.watching = 0;
+    } else if (this.state.watching === 0 ||
+      this.state.searchString !== searchString && searchString != null) {
```
mlunoe: 1. Let's check if `searchString` is defined before we try to determine if it is not equal to `this.state.searchString`.
2. Maybe we could also define better variables? Usually in the `componentWillReceiveProps` we prefix with next for next state, so it becomes `searchString` and `nextSearchString`

src/js/components/MesosLogView.js
```diff
@@ -197,6 +204,22 @@ class MesosLogView extends mixin(StoreMixin) {
...
+    let node = logContainer.querySelector('.highlight.selected');
+    if (!node) {
+      return;
+    }
+
+    let containerHeight = logContainer.clientHeight;
+    let containerScrollTop = logContainer.scrollTop;
+
+    if ((node.offsetTop > containerHeight + containerScrollTop) ||
+      node.offsetTop < containerScrollTop) {
```
mlunoe: I can't remember, but are both `offsetTop` and `scrollTop` well-supported in all browsers?

src/styles/components/filter-bar.less
```diff
@@ -4,6 +4,10 @@
   flex-wrap: wrap-reverse;
   width: 100%;
 
+  &.flex-wrap {
```
mlunoe: We already have `flex-box-wrap`: https://github.com/mesosphere/canvas/blob/0.2.1/layout/foundation.less#L450

src/styles/components/side-panel-detail.less
```diff
@@ -226,6 +226,23 @@ Side Panel Detail
 
     width: @base-spacing-unit * 7;
 
+    &-wide {
+
+      width: @base-spacing-unit * 12;
```
mlunoe: When we create styles like these we need to make sure to create that spacing for all screens using all of the base spacing unit definitions: `@base-spacing-unit`, `@base-spacing-unit-screen-mini`, `@base-spacing-unit-screen-small`, etc.

src/styles/components/buttons.less
```diff
@@ -35,6 +35,29 @@ components/buttons.less
         }
       }
     }
+
+    &-up,
```
mlunoe: I'm not sure I like those `&-up`, `&-down`, `&-wide`, etc. It makes it hard to search and find your class definition

mlunoe:
This might be more related to the next task, but I'm experiencing a lack too, when I hold down enter to go to next match. The same thing happens if I just step through them clicking the arrows.

src/js/components/FilterInputText.js
kennyt: done

src/js/components/TaskDebugView.js
kennyt: done

src/js/components/TaskDebugView.js
kennyt: done

src/js/components/MesosLogView.js
kennyt: you're right, I think `offsetTop` might've had some problems. I'll create a DOMUtils method for that. I believe scrollTop was good (we use that in reactjs-components heavily for the tables)

kennyt: created a PR here #58

rcorral:
Can we make all of this fit in one line:
img

src/styles/components/side-panel-detail.less
kennyt: this is for the input in the side panels though, and the width is static until a certain point

mlunoe: I know, but the margin should still change along with everything else when screen becomes smaller or bigger.
The `base-spacing-unit` has different values for different screens

mlunoe:
We could make it so the dropdown is used whenever there is more than two logs

src/styles/components/side-panel-detail.less
kennyt: but why would we change the width when we don't want to change the width?

kennyt:
@rcorral @mlunoe alright i made it so that it becomes a dropdown when it's bigger than 2. there is still the possibility that it might break onto multiple lines, though if the name of it is bigger than usual

src/styles/components/buttons.less
mlunoe: @kennyt I didn't get a response on this

kennyt: changed it to `-up-arrow` `-down-arrow`

jfurrow:
@mlunoe @kennyt I am seeing a conversation in my email, but Github throws an error when I try to view it, so I'm adding my two cents here:

In regards to using the `&-up-arrow` class concatenation syntax (I made up this term), I think we should stick with it. I think the nesting keeps the classes nicely organized without adding unnecessary specificity. I do see your point about it being difficult to search for the full class name, but personally I think the benefits outweigh the negatives.

kennyt:
@rcorral alright cool, styles implemented where it won't break lines until mobile width

src/styles/components/buttons.less
jfurrow: @kennyt This still needs :abcd: :)

kennyt: done

src/styles/components/buttons.less
mlunoe: This doesn't address the problem. It is hard for me to search for the class definition if the class is not defined as a whole string

kennyt: wait so you're against using `&` for less?

PR #47, opened 2016-04-22
Introduce FilterBar component

daiyiwhale:
Introduces the `FilterBar` component to wrap component children in `filter-bar` styles. The `rightAlignLastChild` and `rightAlignLastNChildren` props are optional parameters to right-align items in the FilterBar. 

Depends on https://github.com/dcos/dcos-ui/pull/6

### DIFF
--- a/src/js/components/FilterBar.js
+++ b/src/js/components/FilterBar.js
@@ -3,7 +3,7 @@ import React from 'react';
 class FilterBar extends React.Component {
 
   getFilterBarLeft(filterItems, rightAlignCount) {
-    if (rightAlignCount) {
+    if (rightAlignCount > 0) {
       filterItems = filterItems.slice(0, filterItems.length - rightAlignCount);
     }
 
@@ -15,7 +15,7 @@ class FilterBar extends React.Component {
   }
 
   getFilterBarRight(filterItems, rightAlignCount) {
-    if (!rightAlignCount) {
+    if (rightAlignCount === 0) {
       return null;
     } else {
       filterItems = filterItems.slice(rightAlignCount * -1);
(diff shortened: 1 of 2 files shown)


mesosphere-web:
Reviewer: @jfurrow

src/js/components/FilterBar.js
```diff
@@ -0,0 +1,67 @@
...
+        {this.getFilterBarLeft(filterItems, rightAlignLastNChildren)}
+        {this.getFilterBarRight(filterItems, rightAlignLastNChildren)}
+      </div>
+    );
+  }
+}
+
+FilterBar.propTypes = {
+  rightAlignLastChild: React.PropTypes.bool,
+  rightAlignLastNChildren: React.PropTypes.number
```
kennyt: hmm this makes it kinda complicated to use indexes. why not just have two props: `leftChildren`, `rightChildren` and then that's it? It'll make it so we can't pass it as children, but it feels better than slicing react children

daiyiwhale: I felt like slicing the children is a better user experience because you simply have to wrap your components and let `FilterBar` take care of the rest. React is totally cool with it to: http://facebook.github.io/react/docs/top-level-api.html#react.children.toarray

kennyt: alright that sounds good then. what was the reason of not just using an index and having both the bool/index?

daiyiwhale: Again for a good user experience; most situations in our UI will only require one element to be right aligned, and `rightAlignLastChild` is super readable. `rightAlignLastNChildren` is to account for future situations designers might throw at us. It's almost no overhead for usability gained, but I'm pretty ambivalent overall if you feel strongly!

kennyt: yeah i would say let's just go with one prop. because it does add another level of complexity of which prop takes precedence. all people would have to do is just pass `1` instead of `true` imo

daiyiwhale: 👍

daiyiwhale: done!

kennyt:
LGTM

daiyiwhale:
Ready for review 👍

jfurrow:
Nice @daiyiwhale!

PR #52, opened 2016-04-25
[2] Use webpack for dev server and build

MatApple:
## Merging into feature branch
No longer need gulpfile other than running a server for build artifacts for verification. `gulp run 'connect:server'`

NOTE: Tests will fail till we merge into feature branch


mesosphere-web:
Reviewer: @orlandohohmeier

webpack/make-webpack-config.js
```diff
@@ -0,0 +1,191 @@
...
+  return {
+    entry: options.production ? './src/js/index.js' : [
+      'webpack-dev-server/client?http://localhost:8080',
+      'webpack/hot/only-dev-server',
+      './src/js/index.js',
+    ],
+
+    devServer: {
+      proxy: {
+        '*': 'http://mat-vytkb-elasticl-gaugjfo9dd8u-2145658064.us-west-2.elb.amazonaws.com/'
```
kennyt: should this be here?

webpack/make-webpack-config.js
```diff
@@ -0,0 +1,191 @@
...
+  var sassLoaders = scssLoaders + '?indentedSyntax=sass';
+  var lessLoaders = cssLoaders + '!less';
+
+  if (options.production) {
+    cssLoaders = extractForProduction(cssLoaders);
+    sassLoaders = extractForProduction(sassLoaders);
+    scssLoaders = extractForProduction(scssLoaders);
+    lessLoaders = extractForProduction(lessLoaders);
+
+    analyticsKey = '51ybGTeFEFU1xo6u10XMDrr6kATFyRyh';
```
kennyt: should these keys be here?

webpack/make-webpack-config.js
```diff
@@ -0,0 +1,191 @@
...
+          test: /\.css$/,
+          loader: cssLoaders,
+        },
+        {
+          test: /\.sass$/,
+          loader: sassLoaders,
+        },
+        {
+          test: /\.scss$/,
+          loader: scssLoaders,
```
jfurrow: What are we using the `Sass` and `SCSS` loaders for?

webpack/make-webpack-config.js
```diff
@@ -0,0 +1,191 @@
...
+
+function extractForProduction(loaders) {
+  return ExtractTextPlugin.extract('style', loaders.substr(loaders.indexOf('!')));
+}
+
+module.exports = function (options) {
+  var cssLoaders = 'style!css!postcss';
+  var scssLoaders = cssLoaders + '!sass';
+  var sassLoaders = scssLoaders + '?indentedSyntax=sass';
+  var lessLoaders = cssLoaders + '!less';
```
jfurrow: What are we using the `Sass` and `SCSS` loaders for?

webpack/make-webpack-config.js
MatApple: Moved them to package.json. Unfortunately they need to be kept somewhere with global context because they are used in multiple places. Dependents should be consolidated but lets worry about that later. package.json seems like a good candidate for this configuration for now.

webpack/make-webpack-config.js
MatApple: Plugin developers have more flexibility to use tools of their choice.

orlandohohmeier: If we want to provide this freedom of choice, we should consider different plugin mechanisms. They're so tightly coupled at the moment that they should use the very same tools to reduce the amount of conflicts due to different tooling.

webpack/webpack.config.js
```diff
@@ -0,0 +1,38 @@
+var ExtractTextPlugin = require('extract-text-webpack-plugin');
```
orlandohohmeier: I'd like us to use ES6 also for the configuration, and webpack is smart enough to load/parse the file with babel if you add `babel` to the file name ( `webpack.config.babel.js`).  What do you think, should we give it a try?

webpack/webpack.config.js
```diff
@@ -0,0 +1,38 @@
...
+
+try {
+  // If this fails - you should create a proxy.dev.js by copying proxy.template.js
+  // and customizing the proxy as needed.
+  proxy = require('./proxy.dev.js');
+} catch (err) {
+  proxy = require('./proxy.template.js');
+}
+
+module.exports = require('./make-webpack-config')({
```
orlandohohmeier: I'm not that sure about this `make-webpack-config` process. It seems a little over engineered and is hard to follow. Why not one very explicit and descriptive configuration file for the default configuration and a second one that extends/overwrites (`Object.assign`) specific configurations?

webpack/make-webpack-config.js
jfurrow: After thinking about it a bit more, I'm not sure this is the right solution. What if a plugin developer wants to use Stylus or PostCSS or something else? I think we should allow plugin developers to use whatever they like.

mlunoe: We can't account for all wishes to develop plugins in this setup. I say remove Sass/SCSS for now, and let's think about how we can become tool-agnostic for plugin development as a separate task

src/styles/components/icons.less
```diff
@@ -213,7 +213,7 @@ Icon (huge)
 
     .icon.icon-sprite.icon-sprite-mini {
 
-      background-image: url("@{path-images}components/icons/icons-mini.png?@{cache-buster}");
```
mlunoe: Did we decide to remove the `cache-buster`?

orlandohohmeier: It's handed by the webpack loader.

webpack/build.config.js
```diff
@@ -0,0 +1,31 @@
+var srcFolder = './src';
```
mlunoe: Could we please `git mv`, so it is easier to follow changes (and go back in history, if needed)?

webpack/make-webpack-config.js
MatApple: OK you guys have a valid point. I'll remove it for now.

webpack/make-webpack-config.js
MatApple: No, it has been moved to a separate dev file which is ignored by git.

webpack/webpack.config.js
MatApple: That's an interesting idea. Didn't know Webpack could do that. I'd like to use ES6 for build tools too. I'll try it.

MatApple:
Unfortunately made this PR against master. Closing and reopening (https://github.com/dcos/dcos-ui/pull/87)

PR #54, opened 2016-04-25
[4] Remove deprecated testutil

MatApple:
## Merging into feature branch

NOTE: Tests will fail till we merge into feature branch

PR #53, opened 2016-04-25
[3] Updating imports for css and images

MatApple:
## Merging into feature branch
- Import images and styles directly within JavaScript

NOTE: Tests will fail till we merge into feature branch

PR #51, opened 2016-04-25
[1] Upgrade build process for plugins

MatApple:
## Merging into feature branch
- Remove hard coded plugins reference. Now set external plugin location with env variable.
- Generate jest config dynamically.

NOTE: Tests will fail till we merge into feature branch

PR #55, opened 2016-04-25
Upgrading build process

MatApple:
## Feature branch
- No more virtual machine proxy for dcos ui. Proxy is part of dev server.
- Now have hot loading on dev server.
- No hard coded references to external plugins. External plugins location is now
completely configurable with npm config key. Now have full flexibility of where plugins are in relation to dcos-ui and the plugin directory name.
- No more copying over plugins into dcos-ui repo via gulp. Webpack will load and watch plugins specified in the directory set by 
```
npm config set dcos-ui:external_plugins ../relative/path/to/plugins/folder
```
- Use webpack's watch for all files. No more gulp watch.
- Full cache busting on all files. Every file (images, css imports e.g. icons, css, Js) has webpack's build hash prepended to it's file name. [hash]-name.[ext].
- Reduced built bundle to only 1.54MB (1.7MB with plugins) of JavaScript and 400kB of css. Reduced dev bundle size from 11MB of JavaScript to 7MB. Force React to build with NODE_ENV=production to reduce footprint of React. Pruned fixtures from production build.
- Dev build process no longer includes long running tasks like image optimization.
- Everything is now linted including unit and integration tests.
- Linting runs on plugins and dcos-ui on every file change and results show in terminal.
- Lint errors now force CI testing to fail early.
- Styles (css, less, sass) and images (png, jpeg, svg, etc) can now be bundled within plugins easily and imported as regular modules `import avatar from '../img/avatar';`

### PRs open (select once merged):
- [x] https://github.com/dcos/dcos-ui/pull/51
- [x] ~~https://github.com/dcos/dcos-ui/pull/52~~ [closed]
- [x] https://github.com/dcos/dcos-ui/pull/87
- [x] https://github.com/dcos/dcos-ui/pull/53
- [x] https://github.com/dcos/dcos-ui/pull/54

### NOTE:
None of the tests will pass on the above PRs till they all merge into this branch. I split them apart for easier code review. Don't bother testing locally till all code is in this PR.

### Outstanding:
- [x] Include Dygraphs in new build
- [x] Deal with Config.dev.js on production builds.
- [x] Remove all Gulpfile tasks except `connect:server` so we can still validate builds. [edit: Also use it for ensuring Config.dev.js is present on production builds.]
- [x] Remove gulp related dependencies from package.json
- [x] Shrinkwrap
- [x] Update README


mesosphere-web:
Reviewer: @philipnrmn

jfurrow:
Nice work!

README.md
```diff
@@ -5,6 +5,7 @@
 Node 4.x (and above) is **required**. We suggest using [nvm](https://github.com/creationix/nvm) or [n](https://github.com/tj/n) to keep multiple Node versions on your system.
 
 
+
```
kennyt: can you write unit tests, integration tests, docs, and create a wiki, and send out an email to the team for this change?

MatApple: 👍  It's a placeholder for bigger changes. The README will be updated with actual information.

kennyt: nah dude it was a joke cause it's just a newline

mlunoe: It's really hard to review this change. Could we break it into separate PRs?

jfurrow: Could you maybe add some comments that explains this change? Nothing huge, but two or three paragraphs about what's going on should be sufficient!

daiyiwhale: Can you test in all browsers? not sure if Edge is supported

webpack/webpack.dev.babel.js
```diff
@@ -0,0 +1,107 @@
...
+module.exports = Object.assign({}, webpackConfig, {
+  entry: [
+    'webpack-dev-server/client?http://localhost:' + PORT,
+    'webpack/hot/only-dev-server',
+    './src/js/index.js'
+  ],
+  devtool: '#eval-source-map',
+  output: {
+    path: './build',
+    filename: 'index.js'
```
rcorral: Why are these paths hardcoded? Shouldn't they come from the config?

jest/genConfig.js
```diff
@@ -0,0 +1,79 @@
+/*
+  Generates configuration for jest.
+ */
+var fs = require('fs');
+var path = require('path');
+var pluginsFolder = process.env.DCOS_UI_PLUGINS;
+var testPaths = ['src'];
+if (pluginsFolder) {
+  try {
+    require('../' + pluginsFolder);
```
rcorral: What if the plugins path is an absolute path? won't this fail?

rcorral: Do we even need this require?

package.json
```diff
@@ -47,121 +48,70 @@
...
+    "less": "2.6.1",
+    "less-loader": "2.2.3",
     "localStorage": "1.0.3",
     "marked": "0.3.5",
-    "mkdirp": "0.5.1",
     "node-libs-browser": "1.0.0",
-    "run-sequence": "1.1.5",
+    "react-hot-loader": "1.3.0",
+    "postcss-loader": "0.8.2",
+    "react-hot-loader": "1.3.0",
```
rcorral: this is here twice.
Also would've loved to see the added features on a different PR (post this massive one)

webpack/build.config.js
```diff
@@ -0,0 +1,31 @@
...
+};
+
+var files = {
+  srcJS: dirs.srcJS + '/index.js',
+  distJS: dirs.distJS + '/index.js',
+  srcCSS: dirs.srcCSS + '/index.less',
+  distCSS: dirs.distCSS + '/index.css',
+  srcHTML: dirs.src + '/index.html',
+  distHTML: dirs.dist + '/index.html'
+};
```
rcorral: It doesn't look like you use any of these variables.

package.json
```diff
@@ -47,121 +48,70 @@
...
+    "test": "npm run lint && npm run lint-plugins && node jest/genConfig.js && NODE_PATH=node_modules jest --config=jest/config.json --no-cache",
+    "jest": "node jest/genConfig.js && NODE_PATH=node_modules jest --config=jest/config.json --no-cache",
+    "start": "webpack-dev-server --config ./webpack/webpack.dev.babel.js --hot --progress --colors --content-base ./build",
+    "build": "./node_modules/.bin/gulp ensureConfig && NODE_ENV='production' webpack --config ./webpack/webpack.production.babel.js"
   },
-  "jest-webpack-alias": {
-    "configFile": ".webpack.config.js"
+  "analytics": {
+    "development": "39uhSEOoRHMw6cMR6st9tYXDbAL3JSaP",
+    "production": "51ybGTeFEFU1xo6u10XMDrr6kATFyRyh"
```
rcorral: Not sure if this is the best spot for these config keys.

webpack/webpack.config.babel.js
```diff
@@ -0,0 +1,64 @@
+import autoprefixer from 'autoprefixer';
+import colorLighten from 'less-color-lighten';
+import path from 'path';
+
+function absPath() {
+  let args = [].slice.apply(arguments);
+  args.unshift(__dirname, '..');
+
+  return path.resolve.apply(path.resolve, args);
+}
```
rcorral: Could you use this instead?
https://nodejs.org/api/path.html#path_path_normalize_path

webpack/webpack.dev.babel.js
```diff
@@ -0,0 +1,107 @@
+import ExtractTextPlugin from 'extract-text-webpack-plugin';
+import HtmlWebpackPlugin from 'html-webpack-plugin';
+import StringReplacePlugin from 'string-replace-webpack-plugin';
+import WebpackNotifierPlugin from 'webpack-notifier';
+
+import packageInfo from '../package';
+import webpackConfig from './webpack.config.babel';
+
+let PORT = process.env.PORT || 8080;
```
rcorral: Could we default to 4200 as that's what most people use today?

jest/genConfig.js
MatApple: We need to check the path is valid. If it isn't a valid path then we don't add the path to jest because jest won't be able to resolve it. The next line will never execute.

webpack/webpack.dev.babel.js
MatApple: What do you think about getting rid of the `build.config.js` file? Those variables aren't really needed because we only use them here. This webpack config is for this repo and doesn't need to be generalized.

package.json
MatApple: Good catch.

webpack/build.config.js
MatApple: You're right, I'd say we get rid of this file because it is now an unnecessary layer of abstraction.

package.json
MatApple: Ideally they are part of the telemetry plugin. I'll have to check if we can put them there now that we've changed the way we do telemetry. Previously it was difficult because we had multiple dependents.

webpack/webpack.config.babel.js
MatApple: Sure!

webpack/webpack.dev.babel.js
MatApple: Yes, makes sense!

webpack/webpack.dev.babel.js
MatApple: I've removed `build.config.js`.

package.json
MatApple: Fixed

webpack/build.config.js
MatApple: Gone!

webpack/webpack.dev.babel.js
MatApple: Done

package.json
MatApple: Moved to Config file. There was only a single instance where we were replacing `@@ANALYTICS_KEY`.

webpack/webpack.dev.babel.js
mlunoe: Did you push?

gulpfile.js
```diff
@@ -1,259 +1,18 @@
...
-      baseDir: config.dirs.dist
-    },
-    socket: {
-      domain: 'localhost:4200'
-    }
-  });
 });
 
-gulp.task('connect:server', function () {
+gulp.task('serve', function () {
```
mlunoe: Won't this still be rivalling the webpack-dev-server?

gulpfile.js
```diff
@@ -1,259 +1,18 @@
...
-}
 
-function ensurePluginDirectoryExists() {
-  // Make temp plugins directory if doesn't exist
-  if (!fs.existsSync(config.dirs.pluginsTmp)) {
-    mkdirp(config.dirs.pluginsTmp, function () {
-      console.log('Created ' + config.dirs.pluginsTmp);
-    });
+gulp.task('ensureConfig', function () {
+  // Make sure we have a Config.dev so we don't error on Config loading
```
mlunoe: Couldn't we do something similar to what we are doing with the proxy here?

package.json
```diff
@@ -47,121 +48,70 @@
...
     "source-map-loader": "0.1.5",
+    "string-replace-webpack-plugin": "0.0.3",
+    "style-loader": "0.13.1",
     "transform-loader": "0.2.3",
     "webpack": "1.13.0",
+    "webpack-dev-server": "1.14.1",
     "webpack-notifier": "1.3.0"
   },
   "scripts": {
     "clean": "rm -rf build",
```
mlunoe: This has been a mistake for a long time, but the folder being produced is called 'dist': https://github.com/dcos/dcos-ui/pull/55/files#diff-4c60f7b3842aaf35cce1238f92a3ea6fR2

mlunoe: Not something we need to fix here

gulpfile.js
```diff
@@ -1,261 +1,18 @@
...
-      domain: 'localhost:4200'
-    }
-  });
 });
 
-gulp.task('connect:server', function () {
+gulp.task('serve', function () {
   connect.server({
     port: 4200,
     root: config.dirs.dist
```
mlunoe: I think this file was just deleted

MatApple: Seems to still be alive!

package.json
MatApple: Yeah that clean can just get removed. It's not cleaning anything!

gulpfile.js
MatApple: This is a bit ugly but it works for now. Lets mark it as an improvement to make in the future. What do you think?

gulpfile.js
mlunoe: Not for me

gulpfile.js
MatApple: Yes, in terms of default port usage. Just don't run them at the same time, or change the port for dev server.

package.json
mlunoe: I think we should make sure to clean out the repository before we run the build, so we don't by accident rely on old files from an earlier build

gulpfile.js
mlunoe: Sounds good

gulpfile.js
MatApple: Sorry, yes the config file was removed. There was so much config in it that isn't being used anymore

MatApple: Fixed

gulpfile.js
mlunoe: Ok, maybe this could also be marked as an improvement, but I think the webpack server could take care of this too, right?

MatApple: Yeah it can. Its just a little more involved because we need to write a plugin.

package.json
MatApple: OK will update to clean out `dist`

MatApple: Done

gulpfile.js
mlunoe: Ok, let's mark it as an improvement then!

webpack/webpack.config.babel.js
```diff
@@ -0,0 +1,67 @@
...
+
+// Can override this with npm config set dcos-ui:external_plugins ../some/relative/path/to/repo
+let pluginsDir = absPath(process.env.npm_package_config_external_plugins || 'plugins');
```
rcorral: What's the purpose of defaulting to `'plugins'`?

gulpfile.js
```diff
@@ -1,261 +1,18 @@
...
-  }
-}
 
-function ensurePluginDirectoryExists() {
-  // Make temp plugins directory if doesn't exist
-  if (!fs.existsSync(config.dirs.pluginsTmp)) {
-    mkdirp(config.dirs.pluginsTmp, function () {
-      console.log('Created ' + config.dirs.pluginsTmp);
-    });
+gulp.task('ensureConfig', function () {
```
rcorral: I'm not sure this is needed. Why can't the build just fail until someone creates the config.dev file?

MatApple: It will break CI

README.md
rcorral: Can this be removed?

jest/genConfig.js
```diff
@@ -0,0 +1,74 @@
+/*
```
rcorral: Could we make this filename `gen-config.js` I don't think it needs to be camel cased.

webpack/webpack.config.babel.js
MatApple: Defaulting to `plugins` provides us with a resolvable directory so we don't get any errors. It won't be used other than to safely resolve `require.context` in `Loader.js`. If pluginsDir is invalid (can only happen if a developer does `npm config set dcos-ui:external_plugins some/dir`) where `some/dir` is invalid then we throw an error because it's not resolvable.

package.json
```diff
@@ -47,121 +52,65 @@
...
     "babel-runtime": "6.6.1",
-    "browser-sync": "2.12.3",
-    "del": "2.2.0",
-    "envify": "3.4.0",
+    "css-loader": "0.23.1",
     "eslint": "2.0.0",
     "eslint-loader": "1.3.0",
     "eslint-plugin-react": "2.6.4",
+    "extract-text-webpack-plugin": "1.0.1",
+    "file-loader": "0.8.5",
```
rcorral: Where is this used?

package.json
```diff
@@ -47,121 +52,65 @@
...
     "babel-preset-es2015": "6.6.0",
+    "babel-preset-jest": "11.0.2",
     "babel-preset-react": "6.5.0",
+    "babel-preset-react-hmre": "1.1.1",
+    "babel-register": "6.8.0",
     "babel-runtime": "6.6.1",
-    "browser-sync": "2.12.3",
-    "del": "2.2.0",
-    "envify": "3.4.0",
+    "css-loader": "0.23.1",
```
rcorral: Is this needed for `postcss-loader`?

README.md
MatApple: Yes, I still need to update the README to reflect the new dev and build process.

package.json
```diff
@@ -47,121 +52,65 @@
...
     "webpack": "1.13.0",
+    "webpack-dev-server": "1.14.1",
     "webpack-notifier": "1.3.0"
   },
   "scripts": {
-    "clean": "rm -rf build",
+    "clean": "rm -rf dist",
     "prebuild-ee": "./scripts/validate-tests",
-    "build-ee": "npm run clean && npm test && NODE_ENV='production' ./node_modules/.bin/gulp dist",
+    "build-ee": "./node_modules/.bin/gulp ensureConfig && npm run clean && npm run lint && npm run lint-plugins && npm test && npm run build",
```
rcorral: Can we rename this to `build-with-plugins`

jest/genConfig.js
MatApple: Done

package.json
```diff
@@ -47,121 +52,65 @@
...
     "dist": "npm run build-ee",
     "prebuild-ce": "./scripts/validate-tests",
-    "build-ce": "npm run clean && npm test && NODE_ENV='production' ./node_modules/.bin/gulp dist",
-    "serve": "NODE_ENV='development' ./node_modules/.bin/gulp serve",
+    "build-ce": "./node_modules/.bin/gulp ensureConfig && npm run clean && npm run lint && npm test && npm run build",
```
rcorral: Maybe this can be `build-without-plugins`

package.json
MatApple: To load images :)

package.json
MatApple: `postcss-loader` basically provides you with entry points to run `autoprefixer` etc. Then to load the result we still need `css-loader`.

webpack/webpack.config.babel.js
rcorral: Shouldn't this variable be called `externalPluginsDir`?

rcorral: I think it's ok for the build to fail. The goal is to build the application the way you have your config set rather than building the application in a different way that what you'd want.

webpack/webpack.dev.babel.js
```diff
@@ -0,0 +1,108 @@
...
+
+// Defaults to value in package.json.
+// Can override with npm config set dcos-ui:port 80
+let PORT = process.env.npm_package_config_port;
+let proxy;
+
+try {
+  // If this fails - you should create a proxy.dev.js by copying proxy.template.js
+  // and customizing the proxy as needed.
+  proxy = require('./proxy.dev.js');
```
rcorral: Again here, I think it's ok to fail if the file doesn't exist.
Maybe we can have an npm script that just copies this proxy file to `.dev` and same thing for the `config.template` that way it's easy for someone to get started.

MatApple: Good idea, but I wouldn't want someone to start the server for the first time and run into an error. I'll create script.

gulpfile.js
rcorral: We can add a line to CI to create the file before the tests run.

webpack/webpack.dev.babel.js
MatApple: Done

package.json
MatApple: It's now just `build` and `build-with-plugins`

gulpfile.js
MatApple: Thats fair, but I think since we have it, it doesn't make sense to remove it since its helping people get up and running quicker. It's nice for someone to clone the repo and `npm i && npm start`. In exchange for removing this, we'd have to write more docs (which we need at some point, just not immediately) :). What do you think?

webpack/webpack.config.babel.js
MatApple: Could you suggest a default configuration so it won't fail to build by default? Most people who get started won't have an external plugins directory. I updated the name of the variable to be more explicit!

package.json
```diff
@@ -47,121 +52,64 @@
...
-    "test": "./node_modules/.bin/gulp copy:external-plugins && npm run lint && jest --no-cache"
+    "lint-plugins": "PLUGINS=$npm_package_config_external_plugins; ./node_modules/.bin/eslint \"$PLUGINS/**/*.js\" --quiet",
+    "prebuild": "./scripts/validate-tests",
+    "serve": "NODE_ENV='development' npm run dist && ./node_modules/.bin/gulp serve",
+    "serve-notify": "export NOTIFY=true && npm run serve",
+    "scaffold": "./node_modules/.bin/gulp ensureConfig && ./node_modules/.bin/gulp ensureDevProxy",
+    "start": "PORT=$npm_package_config_port; webpack-dev-server --config ./webpack/webpack.dev.babel.js --hot --progress --colors --content-base ./build --port $PORT",
+    "test": "node jest/gen-config.js && NODE_PATH=node_modules jest --config=jest/config.json --no-cache",
+    "testing": "NODE_ENV='testing' npm run dist && ./node_modules/.bin/gulp serve",
+    "dist": "NODE_ENV='production' webpack --config ./webpack/webpack.production.babel.js"
```
rcorral: I like the idea of renaming `dist` to something else.
What if this became `build-application`?

README.md
```diff
@@ -36,8 +36,6 @@ Node 4.x (and above) is **required**. We suggest using [nvm](https://github.com/
 6. Navigate to http://localhost:4200
 
 
-After installing all development dependencies and configuring your local environment (steps 1-4 above), you can run test by running `npm test`.  You can build production assets by running `npm run dist`.  If you are actively developing, you may want to run `npm start` to avoid having to refresh your browser with every change.
-
```
MatApple: This info is not needed anymore as its explained in previous steps or further down the page.

PR #50, opened 2016-04-23
Use FilterBar component

daiyiwhale:
Wrapping all filter bars with `FilterBar` component for consistent handling of wrapping and margins.

Depends on https://github.com/dcos/dcos-ui/pull/47.

PR #56, opened 2016-04-25
Upgrade Node Side Panel to full page

daiyiwhale:
Node Side Panel is a full-fledged citizen.

Note that Task View remains a side panel.

Known issues:
- <s>The filter bar is not using a `FilterBar` component yet in this PR, it is done here: https://github.com/dcos/dcos-ui/pull/50</s>
- <s>loading page moving slightly when loading</s>


mesosphere-web:
Reviewer: @mlunoe

daiyiwhale:
Ah there's an issue with the loading page moving slightly when loading, I'll look into that now.

src/js/components/NodeSidePanelContents.js
```diff
@@ -168,7 +169,7 @@ class NodeSidePanelContents extends SidePanelContents {
             {this.tabs_getUnroutedTabs()}
           </ul>
         </div>
-        {this.tabs_getTabView()}
+        <RouteHandler/>
```
mlunoe: Nice!

src/js/components/FilterByTaskState.js
```diff
@@ -1,5 +1,6 @@
-import React from 'react';
+import classNames from 'classnames';
 import {Dropdown} from 'reactjs-components';
+import React from 'react';
```
MatApple: Need to wrap in no-lint :)

src/js/components/TaskView.js
```diff
@@ -1,3 +1,4 @@
+import classNames from 'classnames';
 import mixin from 'reactjs-mixin';
 import React from 'react';
```
MatApple: no-lint wrapper

kennyt:
is the caret supposed to look like this?

caret

src/js/components/NodeSidePanelContents.js
```diff
@@ -1,6 +1,7 @@
 /* eslint-disable no-unused-vars */
```
kennyt: do we even need this component anymore?

src/js/pages/nodes/NodeDetailPage.js
```diff
@@ -0,0 +1,373 @@
...
+    );
+  }
+
+  renderTasksTabView() {
+    let nodeID = this.props.params.nodeID;
+    let tasks = MesosStateStore.getTasksFromNodeID(this.props.params.nodeID);
+
+    let contents = this.getLoadingScreen();
+
+    let timeSinceMount = (Date.now() - this.mountedAt) / 1000;
```
kennyt: we can remove all of the `timeSinceMount` and `this.mountedAt` logic. that was there to not break animations. there are no more animations

daiyiwhale:
The design looks like this: 



It seems weird to me to have an icon for the caret when there is a character symbol caret, but yeah I could make it bigger.

src/js/pages/nodes/NodeDetailPage.js
```diff
@@ -0,0 +1,373 @@
...
+    );
+  }
+
+  getBreadcrumbs(nodeID) {
+    return (
+      <h5 className="inverse">
+        <Link className="headline emphasize" to="nodes">
+          Nodes
+        </Link>
+        <span style={{'margin': '0 15px'}}>
```
kennyt: look at how `TaskDirectoryView` does breadcrumbs. so we can avoid style tags 👍

daiyiwhale: Oh shoot that was supposed to be temporary, I'm embarrassed I left it here ahaha

src/js/pages/nodes/NodeDetailPage.js
```diff
@@ -0,0 +1,373 @@
...
+import SidePanels from '../../components/SidePanels';
+import StringUtil from '../../utils/StringUtil';
+import TabsMixin from '../../mixins/TabsMixin';
+import TaskView from '../../components/TaskView';
+import Units from '../../utils/Units';
+
+// number to fit design of width vs. height ratio
+const WIDTH_HEIGHT_RATIO = 4.5;
+
+const METHODS_TO_BIND = [
```
kennyt: we can remove this if there isn't anything to bind :)

src/js/pages/nodes/NodeDetailPage.js
```diff
@@ -0,0 +1,373 @@
...
+      tasks: 'Tasks',
+      health: 'Health',
+      details: 'Details'
+    };
+
+    this.state = {
+      currentTab: Object.keys(this.tabs_tabs).shift()
+    };
+
+    METHODS_TO_BIND.forEach((method) => {
```
kennyt: we can remove this if there isn't anything to bind :)

src/js/pages/nodes/NodeDetailPage.js
```diff
@@ -0,0 +1,373 @@
...
+      Active: StringUtil.capitalize(node.active.toString().toLowerCase()),
+      Registered: DateUtil.msToDateStr(
+        node.registered_time.toFixed(3) * 1000
+      ),
+      'Master Version': MesosStateStore.get('lastMesosState').version
+    };
+
+    return (
+      <div className="container container-fluid flush">
+        <DescriptionList
```
kennyt: here's what the contrast looks like

ctrst

I wonder if we'll need to introduce inverse styling to the descriptionlist

daiyiwhale: good idea, I will do it

kennyt:
closing by backdrop is off

gif

kennyt:
also, you should be able to go into a nodes detail by clicking on a node in the grid view. this is not happening anymore

kennyt:
when you open a task side panel from the nodes detail page, the back button says `Back` when it should say `Close`

src/js/components/FilterByTaskState.js
```diff
@@ -56,13 +57,22 @@ class FilterByTaskState extends React.Component {
   }
 
   render() {
+    let {inverseStyle} = this.props;
+
+    let dropdownMenuClassSet = classNames({
```
mlunoe: Could we add these as class options in stead of the `inverseStyle` options, since this will never change after definition:
```js
FilterByTaskState.defaultProps = {
  buttonClassName: 'button dropdown-toggle text-align-left',
  dropdownMenuClassName: 'dropdown-menu'
  ...
}
```

That way you have full control over it:
```js
<FilterByTaskState
    buttonClassName="button button-inverse dropdown-toggle text-align-left"
    dropdownMenuClassName="dropdown-menu inverse"
    ... />
```

src/js/components/NodeSidePanelContents.js
mlunoe: +1

src/js/components/TaskTable.js
```diff
@@ -177,12 +179,15 @@ class TaskTable extends React.Component {
   }
 
   render() {
+    let classSet = classNames({
```
mlunoe: Same thing here with default values

src/js/components/TaskView.js
```diff
@@ -159,6 +158,10 @@ class TaskView extends mixin(SaveStateMixin) {
   getContent() {
     let state = this.state;
     let tasks = this.props.tasks || [];
+    let headerClassSet = classNames({
```
mlunoe: Same thing here with default class values

src/js/components/UnitHealthDropdown.js
```diff
@@ -34,10 +35,22 @@ class UnitHealthDropdown extends mixin(InternalStorageMixin) {
   }
 
   render() {
+    let {inverseStyle} = this.props;
```
mlunoe: Same thing here with default class values

src/js/pages/nodes/NodeDetailPage.js
```diff
@@ -0,0 +1,373 @@
...
+          Error finding node
+        </h3>
+        <p className="flush">
+          {`Did not find a node by the id "${this.props.params.nodeID}"`}
+        </p>
+      </div>
+    );
+  }
+
+  getResourceChart(resource, totalResources) {
```
mlunoe: This seems copy-pasted from `SidePanelContents.js`, could we make it a component, since we will need this for both services and nodes?

daiyiwhale: [whispers] they're soon to be removed from services so I think it's ok to have them just like this

mlunoe: I actually still think this should be a component in it self

daiyiwhale: Agreed, but I feel it to be low priority as I could be refactoring for many days if I were to change all the things I felt dissatisfied with in all the code I have moved here ): It should probably be a separate task and addressed when we have time to work on the backlog

mlunoe: I think a separate PR is the right way to go, but I believe this should be part of this task. We are moving towards using the same information in different ways with this task - this code was used by both sidepanels before - and I think it is the right place to tackle that up front.

mlunoe: This does of course not mean a full refactoring of this file, but just making sure that we breaking code into reasonable components.

src/js/pages/nodes/NodeDetailPage.js
daiyiwhale: done

src/js/pages/nodes/NodeDetailPage.js
daiyiwhale: done

src/js/pages/nodes/NodeDetailPage.js
daiyiwhale: good catch, done

src/js/components/NodeSidePanelContents.js
daiyiwhale: all gone!

src/js/pages/nodes/NodeDetailPage.js
daiyiwhale: Just checked with Canvas and this just how dark the inverse styles for text is: http://mesosphere.github.io/canvas/#typography-inverse-styling

src/js/pages/nodes/NodeDetailPage.js
kennyt: i never doubted you 👍

src/js/components/FilterByTaskState.js
kennyt: if you do this, this would also include adding the non-inverse classes back to the `TaskViews` which are not inverse (of which there are many)

src/js/pages/nodes/NodeDetailPage.js
daiyiwhale: I will do this 👍

src/js/components/FilterByTaskState.js
daiyiwhale: yes, and also at times we need to pass the inverse styles down through many child components (`NodeDetailPage` is inverse and passes the inverse flag to its child `TaskView` which passes to its child `TaskTable`), so I think it's meaningful to have this specific inverse flag

src/js/pages/nodes/NodeDetailPage.js
```diff
@@ -0,0 +1,356 @@
...
+
+    this.state = {
+      currentTab: Object.keys(this.tabs_tabs).shift()
+    };
+  }
+
+  componentDidMount() {
+    super.componentDidMount(...arguments);
+
+    this.mountedAt = Date.now();
```
kennyt: you've still got one more

src/js/components/FilterByTaskState.js
kennyt: i do believe we should move towards having the classes replacable like lunoe is saying though. the amount of times we've run into trouble with this issue is significant. the question of whether it should happen in this PR i'm not really sure.

mlunoe: Agreed!

daiyiwhale: We'd still need a specific way to pass the inverse tags through 2 layers of children though

kennyt: yeah so i think `TaskView` should have the boolean and then everything else will have the defaultProps for className

daiyiwhale: makes sense, I'll make a new PR for that since it touches a lot more than just NodeSidePanel

daiyiwhale:
@kennyt, re: caret, we're going to introduce that mini sized caret when the hamburg team creates the breadcrumb component! This will be a placeholder for now. The `unified-ui` doesn't even have a placeholder yet so I could also remove it altogether until the real mini breadcrumbs arrive.

daiyiwhale:
re: go into a nodes detail by clicking on a node in the grid view: fixed

daiyiwhale:
re: back button says Back when it should say Close: this side panel is also being nuked soon so I don't see the need to fix it in this PR (:

src/js/pages/nodes/NodeDetailPage.js
daiyiwhale: done!

src/js/components/FilterByTaskState.js
daiyiwhale: https://github.com/dcos/dcos-ui/issues/80

kennyt:
tests are failing

MatApple:
Hey Chris, should there be spacing between input and dropdown button?

MatApple:
Looks like when we hit the back button from node detail and then select a different node, we have cached data in the component. Can make this transition a little friendlier? If the request to get new Node detail hangs for some reason, we will be stuck looking at an old Node detail

src/js/routes/nodes.js
```diff
@@ -5,67 +5,59 @@ let Redirect = Router.Redirect;
...
         }
       ]
     },
     {
       type: Route,
-      name: 'nodes-grid',
-      path: 'grid/?',
-      handler: NodesGridView,
+      name: 'node-detail',
+      path: 'node/:nodeID/?',
```
mlunoe: Could we make this just `/nodes/{nodeID}`?

MatApple:
As discussed, lets update the routing so we have `nodes/{nodeID}` instead of `nodes/node/{nodeID}` :)

src/js/components/DescriptionList.js
```diff
@@ -2,16 +2,16 @@ import React from 'react';
...
-    let {headline} = this.props;
+    let {headline, headlineClassName} = this.props;
     if (!headline) {
       return null;
     }
 
     // Wrap in headline element and classes
     return (
-      <h6 className="flush-top">
+      <h5 className={headlineClassName}>
```
mlunoe: Did we verify this change with the designers and check that it still looks nice everywhere we use it?

daiyiwhale: yes, the designers don't want h6 anywhere (they made this request a while ago during a design sync)

daiyiwhale:
Routes changed 👍

daiyiwhale:
:D :D :D :D

mlunoe:
So excited!!! Great work @daiyiwhale! This is really looking beautiful!

PR #60, opened 2016-04-26
Feature - add healthbar component

Poltergeist:
This adds a `HealthBar` component which wraps the `StatusBar` with a `Tooltip` and generates the text content for the tooltip.

image

The styling is part of #61

### DIFF
--- a/src/js/components/HealthBar.js
+++ b/src/js/components/HealthBar.js
@@ -1,0 +1,70 @@
+import classNames from 'classnames';
+import React from 'react';
+import {Tooltip} from 'reactjs-components';
+
+import HealthBarMap from '../constants/HealthBar';
+import StatusBar from './StatusBar';
+
+class HealthBar extends React.Component {
+  static get defaultProps() {
+    return {
+      instancesCount: null
+    };
+  }
+
+  renderToolTip() {
+    let {tasks, instancesCount} = this.props;
+
+    tasks = Object.keys(tasks).filter(function (task) {
+      return tasks[task] !== 0 && task !== 'tasksRunning';
+    }).map(function (task, index) {
+      let percentage = parseInt(tasks[task] / instancesCount * 100, 10);
+
+      let classSet = classNames(HealthBarMap[task].cssName, 'dot icon');
+
+      return (
+        <div key={index}>
+          <span className={classSet}></span>
+          {` ${tasks[task]} ${HealthBarMap[task].label} (${percentage} %)`}
+        </div>
+      );
+    }).filter(function (item) {
+      return item != null;
+    });
+
+    if (tasks.length === 0) {
+      return 'No Running Tasks';
+    }
+
+    return tasks;
+  }
+
+  render() {
+    let {tasks, instancesCount} = this.props;
+
+    if (tasks == null) {
+      return null;
+    }
+
+    tasks = Object.keys(tasks).filter(function (task) {
+      return task !== 'tasksRunning';
+    }).map(function (taskStatus) {
+      return {className: HealthBarMap[taskStatus].cssName, value: tasks[taskStatus]};
+    });
+
+    return (
+      <Tooltip content={this.renderToolTip()}>
+        <StatusBar
+          data={tasks}
+          scale={instancesCount}/>
+      </Tooltip>
+    );
+  }
+}
+
+HealthBar.propTypes = {
+  instancesCount: React.PropTypes.number,
+  tasks: React.PropTypes.object.isRequired
+}
+
+module.exports = HealthBar;
--- a/src/js/components/HealthBar.js
+++ b/src/js/components/HealthBar.js
@@ -6,50 +6,58 @@ import HealthBarMap from '../constants/HealthBar';
 import StatusBar from './StatusBar';
 
 class HealthBar extends React.Component {
-  renderToolTip() {
-    let {tasks, instancesCount} = this.props;
 
-    tasks = Object.keys(tasks).filter(function (task) {
-      return tasks[task] !== 0 && task !== 'tasksRunning';
+  getMappedTasksSummary(tasksSummary) {
+    return Object.keys(tasksSummary).filter(function (task) {
+      return task !== 'tasksRunning';
+    }).map(function (taskStatus) {
+      return {
+        className: HealthBarMap[taskStatus].className,
+        value: tasksSummary[taskStatus]
+      };
+    });
+  }
+
+  getTaskList(tasksSummary, instancesCount) {
+    return Object.keys(tasksSummary).filter(function (task) {
+      return tasksSummary[task] !== 0 && task !== 'tasksRunning';
     }).map(function (task, index) {
-      let percentage = parseInt(tasks[task] / instancesCount * 100, 10);
+      let percentage = parseInt(tasksSummary[task] / instancesCount * 100, 10);
 
-      let classSet = classNames(HealthBarMap[task].cssName, 'dot icon');
+      let classSet = classNames(HealthBarMap[task].className, 'dot icon');
 
       return (
         <div key={index}>
           <span className={classSet}></span>
-          {` ${tasks[task]} ${HealthBarMap[task].label} (${percentage} %)`}
+          {` ${tasksSummary[task]} ${HealthBarMap[task].label} (${percentage} %)`}
         </div>
       );
-    }).filter(function (item) {
-      return item != null;
     });
+  }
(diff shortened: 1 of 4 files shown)


mesosphere-web:
Reviewer: @kennyt

daiyiwhale:
Rad! ⚡

src/js/components/HealthBar.js
```diff
@@ -0,0 +1,78 @@
+import React from 'react';
```
daiyiwhale: (don't forget to alphabetize imports!)

kennyt: isn't this already alphabetized?

daiyiwhale: I think it was rebased after I commented, but yep it is now :'D

Poltergeist: :-) This is what I did :-)

src/js/components/HealthBar.js
```diff
@@ -0,0 +1,70 @@
+import classNames from 'classnames';
+import React from 'react';
+import {Tooltip} from 'reactjs-components';
+
+import HealthBarMap from '../constants/HealthBar';
+import StatusBar from './StatusBar';
+
+class HealthBar extends React.Component {
+  static get defaultProps() {
```
orlandohohmeier: Didn't we agreed to not use the static `defaultProps` getter for now?

src/js/components/HealthBar.js
kennyt: haha damn, sneaky sneaky

src/js/components/HealthBar.js
```diff
@@ -0,0 +1,68 @@
...
+  }
+}
+
+HealthBar.defaultProps = {
+    instancesCount: null
+};
+
+HealthBar.propTypes = {
+  instancesCount: React.PropTypes.number,
+  tasks: React.PropTypes.object.isRequired
```
kennyt: what is this object consist of? from what i can gather, it looks like the key is status -> value is the #
if that is the case, is this more like `taskSummaries`?

src/js/components/HealthBar.js
```diff
@@ -0,0 +1,68 @@
...
+        <StatusBar
+          data={tasks}
+          scale={instancesCount}/>
+      </Tooltip>
+    );
+  }
+}
+
+HealthBar.defaultProps = {
+    instancesCount: null
```
kennyt: indented too far

src/js/components/__tests__/HealthBar-test.js
```diff
@@ -0,0 +1,82 @@
...
+  describe('PropTypes', function () {
+    it('should throw an error if no tasks prop is provided', function () {
+      spyOn(console, 'error');
+      this.instance = ReactDOM.render(
+        <HealthBar />,
+        this.container
+      );
+      expect(console.error)
+        .toHaveBeenCalledWith(
+          'Warning: Failed propType: Required prop `tasks` was not' +
```
kennyt: let's just change this to make sure that it was called and was passed a string. maybe react might upgrade and decide to change the wording or something.

in fact, i think just checking it was called is fine too

src/js/components/HealthBar.js
```diff
@@ -0,0 +1,68 @@
...
+
+      let classSet = classNames(HealthBarMap[task].cssName, 'dot icon');
+
+      return (
+        <div key={index}>
+          <span className={classSet}></span>
+          {` ${tasks[task]} ${HealthBarMap[task].label} (${percentage} %)`}
+        </div>
+      );
+    }).filter(function (item) {
```
kennyt: why do we need this filter? won't the previous map ensure that it returns something?

src/js/components/HealthBar.js
```diff
@@ -0,0 +1,68 @@
...
+import {Tooltip} from 'reactjs-components';
+
+import HealthBarMap from '../constants/HealthBar';
+import StatusBar from './StatusBar';
+
+class HealthBar extends React.Component {
+  renderToolTip() {
+    let {tasks, instancesCount} = this.props;
+
+    tasks = Object.keys(tasks).filter(function (task) {
```
kennyt: can we break this logic out into a method and give it a good name for what it's doing?

src/js/components/HealthBar.js
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  render() {
+    let {tasks, instancesCount} = this.props;
+
+    if (tasks == null) {
+      return null;
+    }
+
+    tasks = Object.keys(tasks).filter(function (task) {
```
kennyt: let's break this logic out into a method and give it a good name

src/js/components/HealthBar.js
```diff
@@ -0,0 +1,68 @@
...
+    );
+  }
+}
+
+HealthBar.defaultProps = {
+    instancesCount: null
+};
+
+HealthBar.propTypes = {
+  instancesCount: React.PropTypes.number,
```
kennyt: is it possible for us to derive `instancesCount` from `tasks` instead of passing a prop?

src/js/components/HealthBar.js
MatApple: This confused me too. Perhaps we can rename to something like `taskCountByStatus` or similar.

src/js/components/HealthBar.js
```diff
@@ -0,0 +1,68 @@
...
+      let classSet = classNames(HealthBarMap[task].cssName, 'dot icon');
+
+      return (
+        <div key={index}>
+          <span className={classSet}></span>
+          {` ${tasks[task]} ${HealthBarMap[task].label} (${percentage} %)`}
+        </div>
+      );
+    }).filter(function (item) {
+      return item != null;
```
MatApple: This will never be `null` right? Lets remove this filter

src/js/components/HealthBar.js
```diff
@@ -0,0 +1,68 @@
...
+    });
+
+    if (tasks.length === 0) {
+      return 'No Running Tasks';
+    }
```
MatApple: We should put this above line 12. Base cases should return early :)

src/js/components/HealthBar.js
```diff
@@ -0,0 +1,68 @@
...
+
+    return tasks;
```
MatApple: Instead of reassigning the mapped tasks back to `tasks` again, we can just return at line 12.

src/js/components/__tests__/HealthBar-test.js
rcorral: +1

src/js/constants/HealthBar.js
```diff
@@ -0,0 +1,20 @@
...
+  tasksHealthy: {
+    cssName: 'healthy',
+    label: 'Healthy'
+  },
+  tasksUnhealthy: {
+    cssName: 'unhealthy',
+    label: 'Unhealthy'
+  },
+  tasksStaged: {
+    cssName: 'staged',
```
rcorral: Could you rename `cssName` to `className`. It's more consistent with the rest of the codebase.

src/js/components/HealthBar.js
rcorral: +1
One thing we realized about two months ago was to use unit tests to test methods in a component and leave the testing of the rendering to integration tests.
Putting it into a method allows us to test the unit only.
Then the integration tests don't have to care if the components are called Tooltip/StatusBar, it's just about what's rendered :)
...sorry for the verbose explanation.

src/js/components/HealthBar.js
Poltergeist: I'll add a shape for that.

src/js/components/HealthBar.js
Poltergeist: Actually it is not possible. If the cluster does not have enough offers you might have a couple of instances running or staged but the sum of both might not equal the defined instances.

src/js/components/HealthBar.js
Poltergeist: I am with you on this one, but in this case it is not possible tasks is not an array there and the decision whether we have something to show or not is decided by the filter in line 12.

src/js/components/HealthBar.js
kennyt: sounds good 👍

kennyt:
this looks great now :ship:

what do you think @MatApple ?

MatApple:
Looks good to me too. 🚢

PR #61, opened 2016-04-26
Feature add healthBar service table

Poltergeist:
Dependent on #60 

This adds the Healthbar Component to the service table. 

Sorting by status will sort by health.

image


mesosphere-web:
Reviewer: @MatApple

mesosphere-web:
Reviewer: @philipnrmn

src/js/components/ServicesTable.js
```diff
@@ -109,15 +111,27 @@ var ServicesTable = React.createClass({
     );
   },
 
-  renderHealth: function (prop, service) {
-    let appHealth = service.getHealth();
+  renderStatus: function (prop, service) {
+    let instances = service.getInstancesCount();
```
kennyt: let's call this `instancesCount`

src/js/components/ServicesTable.js
```diff
@@ -109,15 +111,27 @@ var ServicesTable = React.createClass({
...
+    let serviceStatus = service.getStatus();
+    let serviceStatusClassSet = StatusMapping[serviceStatus] || '';
+    let tasks = service.getTasksSummary();
```
kennyt: let's call this `tasksSummary` or `taskSummary` god damn how do i english 💯

src/styles/components/statusbar.less
```diff
@@ -0,0 +1,49 @@
...
+    }
+  }
+}
+
+.status-bar-column {
+  @media (min-width: @screen-medium-max) {
+    width: 330px;
+  }
+  width: 138px;
+
```
kennyt: got an extra newline here

MatApple:
Error - `Cannot find module './HealthBar' from 'ServicesTable.js'`

src/js/components/ServicesTable.js
MatApple: +1 `taskSummary` with one `s` :)

src/js/components/ServicesTable.js
MatApple: or just `instanceCount` since count implies there are multiple.

src/js/components/ServicesTable.js
```diff
@@ -18,6 +16,10 @@ import {Table} from 'reactjs-components';
 import TableUtil from '../utils/TableUtil';
 var Units = require('../utils/Units');
 
+const StatusMapping = {
+  'Running': 'color-1'
+};
```
MatApple: The picture shows red - do we need to map more colors here?

src/js/components/ServicesTable.js
```diff
@@ -18,6 +16,10 @@ import {Table} from 'reactjs-components';
 import TableUtil from '../utils/TableUtil';
 var Units = require('../utils/Units');
 
+const StatusMapping = {
+  'Running': 'color-1'
```
MatApple: What do you think about changing the class name to something more descriptive? I know it should be abstract but perhaps `progressColor-1`

src/js/components/ServicesTable.js
MatApple: +1 `taskSummary`

src/js/components/ServicesTable.js
MatApple: nvm - on second thought I think its fine.

src/styles/components/statusbar.less
```diff
@@ -0,0 +1,49 @@
...
+    background-color: currentColor;
+    border-left: 1px solid @black;
+    &:first-child {
+      border-left: none;
+    }
+  }
+}
+
+.status-bar-column {
+  @media (min-width: @screen-medium-max) {
```
mlunoe: Let's move this out to its own section

src/styles/components/statusbar.less
```diff
@@ -0,0 +1,49 @@
+.status-bar {
+  display: inline-flex;
+  background-color: @gray;
+  margin-right: @base-spacing-unit * 2/3;
```
mlunoe: When you use `@base-spacing-unit`, make sure to create styles for all screens, since we have different variables for different screens: `@base-spacing-unit`, `@base-spacing-unit-screen-mini`, `@base-spacing-unit-screen-small`, etc.

See here for an example: https://github.com/dcos/dcos-ui/blob/master/src/styles/components/filter-bar.less

src/styles/components/statusbar.less
```diff
@@ -0,0 +1,49 @@
...
+  display: inline-block;
+  .color-1 {
+    color: @white;
+  }
+  .color-2 {
+    color: @yellow;
+  }
+}
+
+.unknown {
```
mlunoe: 1. These are defined globally. Could we either move them to a more fitting location or wrap them?
It also feels spammy to define them multiple places: https://github.com/dcos/dcos-ui/pull/61/files#diff-297ecda992ddfcaf45f3d9d4a59aa535R37
I think it would make more sense to define them in one of these two files.

2. I'm actually curious to why we have both style files. Isn't this just different specific situations of the same general style?

src/js/components/ServicesTable.js
orlandohohmeier: Let's stick to the scope of this PR! The [`getTasksSummary`}(https://github.com/dcos/dcos-ui/blob/feature/unified-ui/src/js/structs/Service.js#L107) method is part of the service struct interface and is not part of this feature. Let's do this refactoring in a different branch. PRs welcome!

src/js/components/ServicesTable.js
Poltergeist: This is for the text on the right side of the healthBar.

src/styles/components/statusbar.less
Poltergeist: Decided to only use the spacing on `large` screens since the text which needs the margin is only shown on large screens.

src/styles/components/statusbar.less
Poltergeist: Added that media section at the bottom of the file for all cases.

src/styles/components/statusbar.less
Poltergeist: Those color definitions where for the dots since the bar has a to specific background colors so the background color was not overwritten I added a .dot block around those colors.

src/styles/components/statusbar.less
```diff
@@ -0,0 +1,59 @@
...
+  display: inline-block;
+  .running-state {
+    color: @white;
+  }
+  .deploying-state {
+    color: @yellow;
+  }
+}
+
+.dot {
```
mlunoe: If these are `dot`-styles, then let's move them to `badges.less` or even better it's own `dot.less` styles

src/styles/components/statusbar.less
```diff
@@ -0,0 +1,59 @@
...
+  }
+}
+
+@media (min-width: @screen-medium-max) {
+  .status-bar {
+    margin-right: @base-spacing-unit * 2/3;
+  }
+
+  .status-bar-column {
+    width: 330px;
```
mlunoe: Why are we hard-coding this width? This doesn't seem flexible to other use-cases

src/js/components/ServicesTable.js
```diff
@@ -1,12 +1,10 @@
 import {Link} from 'react-router';
 var React = require('react');
-import {Tooltip} from 'reactjs-components';
 
 import Cluster from '../utils/Cluster';
 var EventTypes = require('../constants/EventTypes');
 import Framework from '../structs/Framework';
-var HealthLabels = require('../constants/HealthLabels');
-var HealthTypesDescription = require('../constants/HealthTypesDescription');
+import HealthBar from './HealthBar';
```
mlunoe: I think this should be `StatusBar` now

mlunoe: Hmm.. wait a minute. Is there a component missing?

mlunoe: Arh, ok. This needs a rebase

mlunoe:
Ok, I'm going to work on this PR to get it mergable

pierlo-upitup:
Thanks @mlunoe ! I'm assigning this myself for another review!

src/js/components/ServicesTable.js
kennyt: i was simply talking about the variable

PR #64, opened 2016-04-26
Feature - Add service detail task list

orlandohohmeier:
task-list
Add a service detail task list which either displays a list of framework tasks including the scheduler tasks or a list of Marathon application tasks.

### DIFF
--- a/src/js/components/FilterButtons.js
+++ b/src/js/components/FilterButtons.js
@@ -25,7 +25,7 @@ class FilterButtons extends React.Component {
   }
 
   getFilterButtons() {
-    let {filterByKey, filters, itemList, selectedFilter} = this.props;
+    let {filterByKey, filters, inverseStyle, itemList, selectedFilter} = this.props;
 
     if (filterByKey) {
       itemList = itemList.map(function (item) {
@@ -36,8 +36,8 @@ class FilterButtons extends React.Component {
     let filterCount = this.getCount(itemList);
 
     return filters.map((filter) => {
-      let classSet = classNames({
-        'button button-stroke button-inverse': true,
+      let classSet = classNames('button button-stroke',{
+        'button-inverse': inverseStyle,
         'active': filter.toLowerCase() === selectedFilter.toLowerCase()
       });
 
@@ -62,19 +62,21 @@ class FilterButtons extends React.Component {
 }
 
 FilterButtons.defaultProps = {
-  renderButtonContent: function (title) {return title; },
-  onFilterChange: function () {}
+  inverseStyle: false,
+  onFilterChange: function () {},
+  renderButtonContent: function (title) {return title; }
 };
 
 FilterButtons.propTypes = {
-  // Optional function to generate button text. args: (filter, count)
-  renderButtonContent: React.PropTypes.func,
   filters: React.PropTypes.array,
   // The key in itemList that is being filtered
   filterByKey: React.PropTypes.string,
+  inverseStyle: React.PropTypes.bool,
+  itemList: React.PropTypes.array.isRequired,
   // A function that returns the onClick for a filter button given the filter.
   onFilterChange: React.PropTypes.func,
-  itemList: React.PropTypes.array.isRequired,
+  // Optional function to generate button text. args: (filter, count)
+  renderButtonContent: React.PropTypes.func,
   // The filter in props.filters that is currently selected.
   selectedFilter: React.PropTypes.string
 };
--- a/src/js/components/FilterButtons.js
+++ b/src/js/components/FilterButtons.js
@@ -64,7 +64,7 @@ class FilterButtons extends React.Component {
 FilterButtons.defaultProps = {
   inverseStyle: false,
   onFilterChange: function () {},
-  renderButtonContent: function (title) {return title; }
+  renderButtonContent: function (title) { return title; }
 };
 
 FilterButtons.propTypes = {
--- a/src/js/components/ServiceDetail.js
+++ b/src/js/components/ServiceDetail.js
@@ -5,7 +5,6 @@ import {StoreMixin} from 'mesosphere-shared-reactjs';
 import InternalStorageMixin from '../mixins/InternalStorageMixin';
 import MesosStateStore from '../stores/MesosStateStore';
 import Service from '../structs/Service';
-import Framework from '../structs/Framework';
 import ServiceInfo from './ServiceInfo';
 import TabsMixin from '../mixins/TabsMixin';
 import TaskView from './TaskView';
@@ -52,12 +51,7 @@ class ServiceDetail
 
   renderTasksTabView() {
     let {service} = this.props;
-    let tasks = [];
-
-    // TODO (orlandohohmeier): Get application tasks by application id
-    if (service instanceof Framework) {
-      tasks = MesosStateStore.getTasksFromServiceName(service.getName());
-    }
+    let tasks = MesosStateStore.getTasksByServiceId(service.getId());
 
     return (<TaskView tasks={tasks} inverseStyle={true} />);
   }
--- a/src/js/components/ServiceDetail.js
+++ b/src/js/components/ServiceDetail.js
@@ -3,9 +3,12 @@ import React from 'react';
 import {StoreMixin} from 'mesosphere-shared-reactjs';
 
 import InternalStorageMixin from '../mixins/InternalStorageMixin';
+import MesosStateStore from '../stores/MesosStateStore';
 import Service from '../structs/Service';
+import Framework from '../structs/Framework';
 import ServiceInfo from './ServiceInfo';
 import TabsMixin from '../mixins/TabsMixin';
+import TaskView from './TaskView';
 
 const METHODS_TO_BIND = [];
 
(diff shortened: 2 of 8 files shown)


mesosphere-web:
Reviewer: @aldipower

aldipower:
Yipii, starting reviewing.

src/js/components/FilterButtons.js
```diff
@@ -36,8 +36,8 @@ class FilterButtons extends React.Component {
     let filterCount = this.getCount(itemList);
 
     return filters.map((filter) => {
-      let classSet = classNames({
-        'button button-stroke button-inverse': true,
+      let classSet = classNames('button button-stroke',{
```
kennyt: can we add another `prop` to this component called `className` and have it default to `button button-stroke` and then use it here? that way, we can have the buttons not stroke if don't want it to :)

aldipower: This is a good idea.
I suggest to add this only when needed. 
Could we make usage of this generalizing thing already?

src/js/components/FilterButtons.js
```diff
@@ -62,19 +62,21 @@ class FilterButtons extends React.Component {
 }
 
 FilterButtons.defaultProps = {
-  renderButtonContent: function (title) {return title; },
-  onFilterChange: function () {}
+  inverseStyle: false,
+  onFilterChange: function () {},
+  renderButtonContent: function (title) {return title; }
```
aldipower: Please add an extra space before the `return` and make sure your editors linting is working properly. Thank you.

src/js/components/FilterButtons.js
```diff
@@ -62,19 +62,21 @@ class FilterButtons extends React.Component {
...
-  renderButtonContent: React.PropTypes.func,
   filters: React.PropTypes.array,
   // The key in itemList that is being filtered
   filterByKey: React.PropTypes.string,
+  inverseStyle: React.PropTypes.bool,
+  itemList: React.PropTypes.array.isRequired,
   // A function that returns the onClick for a filter button given the filter.
   onFilterChange: React.PropTypes.func,
-  itemList: React.PropTypes.array.isRequired,
+  // Optional function to generate button text. args: (filter, count)
```
aldipower: I know it's not you comment. 

I don't trust the `args: (filter, count)`. Could you just remove this portion of the comment?

src/js/components/ServiceDetail.js
```diff
@@ -48,7 +50,13 @@ class ServiceDetail
   }
 
   renderTasksTabView() {
-    return (<span>Tasks Placeholder</span>);
+    let {service} = this.props;
```
aldipower: `service` is only used once. No need to cache this.

src/js/components/ServiceDetail.js
```diff
@@ -67,13 +75,22 @@ class ServiceDetail
             {this.tabs_getUnroutedTabs()}
           </ul>
         </div>
-        {this.tabs_getTabView()}
+        <div className="
```
aldipower: Not sure if the line breaks are correct here?
Are they?

The dangling `"`  looks strange.

src/js/components/FilterButtons.js
orlandohohmeier: My linting is working like charm - eslint just don't complains about the missing space. :(

src/js/components/TaskView.js
```diff
@@ -133,55 +137,74 @@ class TaskView extends mixin(SaveStateMixin) {
...
+    let totalNumberOfTasks = tasks.length;
 
-    if (state.searchString !== '') {
-      tasks = StringUtil.filterByString(tasks, 'name', state.searchString);
+    // Get task states based on TaskStates types
+    let taskStates = tasks.map(function (task) {
+      let {stateTypes} = TaskStates[task.state];
+
+      return stateTypes.find(function (state) {
+        return state === 'active' || state === 'completed';
```
aldipower: Please use the constant instead of the strings to compare.

TaksState.ACTIVE

src/js/components/FilterButtons.js
kennyt: i actually don't think we have a linting for that rule. there are specific cases where we don't do it. i think this is good 👍

src/js/components/TaskView.js
orlandohohmeier: There is no such thing as `TaksState.ACTIVE` – I also wanted to use the constants directly, and I think we should refactor the TaksState structure, but that should be part of a different PR.

src/js/components/ServiceDetail.js
orlandohohmeier: There is in general no need to cache those objects, as the performance benefits are in the mini micro scale, but that's a different story.

src/js/components/FilterButtons.js
orlandohohmeier: I'm ok with removing it, but would keep it, as the documentation is valuable if you're implement your own `renderButtonContent` method. What do you think?

src/js/components/FilterButtons.js
orlandohohmeier: @kennyt You wouldn't add a `space`?

src/js/stores/MesosStateStore.js
```diff
@@ -164,6 +164,37 @@ var MesosStateStore = Store.createStore({
     return [];
   },
 
+  getTasksByServiceId: function (serviceId) {
+    // Convert serviceId to Mesos service name
+    let mesosServiceName = serviceId.split('/').slice(1).reverse().join('.');
```
aldipower: Yummy, spaghetti. Like this one here. :)

src/js/components/FilterButtons.js
aldipower: OK, fine with it. It just feels like the wrong place to document this.

src/js/components/TaskView.js
aldipower: OK sorry. I didn't had the structure of TaskStates in mind.
You're right, TaskStates should be refactored.
There should be constants created for this, but this is not part of this PR of course.

aldipower:
Works like expected in FF. Awesome job. Thank you.

aldipower:
Oh, I have linting errors while trying to run the tests. So I cannot run the tests.

orlandohohmeier:
@aldipower I've rebased the branch and you shouldn't see linting errors anymore.

aldipower:
Perfect! Thanx a lot!

aldipower:
LGTM @Poltergeist Merge if you like.

src/js/components/ServiceDetail.js
```diff
@@ -48,32 +50,43 @@ class ServiceDetail
...
-          <ServiceInfo service={service} />
+          <ServiceInfo service={this.props.service} />
           <ul className="tabs list-inline flush-bottom container-pod
             container-pod-short-top inverse">
             {this.tabs_getUnroutedTabs()}
           </ul>
         </div>
-        {this.tabs_getTabView()}
+        <div className="side-panel-tab-content side-panel-section container
+        container-pod container-pod-short container-fluid container-fluid-flush
```
kennyt: these should be indented one deeper

src/js/stores/MesosStateStore.js
```diff
@@ -164,6 +164,37 @@ var MesosStateStore = Store.createStore({
...
+    }
+
+    // Combine framework (if matching framework was found) and filtered
+    // Marathon tasks. This will give you a list of framework tasks including
+    // the scheduler tasks or a list of Marathon application tasks.
+    return frameworks.reduce(function (serviceTasks, framework) {
+      let {tasks=[], completed_tasks={}, name} = framework;
+
+      if (name === mesosServiceName) {
+        return  serviceTasks.concat(tasks,completed_tasks);
```
kennyt: i'm pretty sure there's an extra space after `return`

src/js/components/TaskView.js
```diff
@@ -133,55 +137,74 @@ class TaskView extends mixin(SaveStateMixin) {
...
-              handleFilterChange={this.handleStatusFilterChange}
-              totalTasksCount={tasks.length}
-              currentStatus={state.filterByStatus}/>
+        <FilterHeadline
+          inverseStyle={inverseStyle}
+          onReset={this.resetFilter}
+          name="Tasks"
+          currentLength={tasks.length}
+          totalLength={totalNumberOfTasks} />
+        <div className="filter-bar">
```
kennyt: there's a new `FilterBar` component that can do all of this for you :)

orlandohohmeier: But it will only replace the div, right? Anyway, let's refactor this in a different PR.

kennyt: it'll replace this div, the `filter-bar-left` div, and the `filter-bar-item` divs

orlandohohmeier: Cool!

src/js/components/FilterButtons.js
orlandohohmeier: I like the idea, but let's please stick to scope of this PR and extend the `FilterButtons` component in a different PR.

Poltergeist:
Hit the Button 🔘 👍 LGTM works as advertised. :shipit: 🐴

PR #65, opened 2016-04-26
Fix broken testing task

MatApple:
Plugin testing is broken since upgrading `babel-jest`. This should fix.


mesosphere-web:
Reviewer: @Poltergeist

PR #66, opened 2016-04-26
Introduce & Use ClickToSelect Component

daiyiwhale:
This PR introduces the `ClickToSelect` Component. 

Usage: wrap a component with `ClickToSelect` to cause browser highlights of the components contents when clicked.

### DIFF
--- a/src/js/components/ClickToSelect.js
+++ b/src/js/components/ClickToSelect.js
@@ -1,0 +1,39 @@
+import React from 'react';
+
+const METHODS_TO_BIND = [
+  'selectAll'
+];
+
+class ClickToSelect extends React.Component {
+  constructor() {
+    super();
+
+    METHODS_TO_BIND.forEach((method) => {
+      this[method] = this[method].bind(this);
+    });
+  }
+
+  selectAll(event) {
+    event.preventDefault;
+    let selection = document.getSelection();
+    let node = this.target;
+
+    selection.selectAllChildren(node);
+  }
+
+  render() {
+    return (
+      <span
+        onClick={this.selectAll}
+        ref={(node) => this.target = node}>
+        {this.props.children}
+      </span>
+    );
+  }
+}
+
+ClickToSelect.propTypes = {
+  children: React.PropTypes.any
+}
+
+module.exports = ClickToSelect;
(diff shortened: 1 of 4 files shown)


mesosphere-web:
Reviewer: @mlunoe

src/js/components/ClickToSelect.js
```diff
@@ -0,0 +1,39 @@
...
+      this[method] = this[method].bind(this);
+    });
+  }
+
+  selectAll(event) {
+    event.preventDefault;
+    let selection = document.getSelection();
+    let node = this.target;
+
+    selection.selectAllChildren(node);
```
jfurrow: We could just do `document.getSelection().selectAllChildren(this.target);` and avoid setting these variables

src/js/components/ClickToSelect.js
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  render() {
+    return (
+      <span
+        onClick={this.selectAll}
+        ref={(node) => this.target = node}>
```
jfurrow: We could just pass a string to `ref` and then use that in place of `this.target`. So you would reference the DOM node by using `this.refs.{someString}`

daiyiwhale: Ah that's much simpler, good idea!

src/js/components/ClickToSelect.js
daiyiwhale: I thought what I wrote would be easier to read, but actually now that I look at it the one-liner is pretty understandable as well haha. Changed!

jfurrow: That's fair! I don't feel strongly about it, so whatever you'd prefer is fine with me.

src/js/components/ClickToSelect.js
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+
+class ClickToSelect extends React.Component {
+  constructor() {
+    super();
+    this.selectAll = this.selectAll.bind(this);
+  }
+
+  selectAll() {
+    document.getSelection().selectAllChildren(this.refs.node);
```
kennyt: did you ensure that `selectAllChildren` works on all browsers?

https://developer.mozilla.org/en-US/docs/Web/API/Selection/selectAllChildren

i suppose if it doesn't, then it simply won't select and it will be ok

daiyiwhale: The support charts are full of mysterious `?` but I tested in latest Chrome, FF, Safari, Edge, and IE11 and it seemed to be functioning as advertised 👍

PR #67, opened 2016-04-27
Bug/dcos 6086 better repository error handling

mlunoe:
New errors comes first and foremost from Cosmos, but if no message is provided nice fallbacks are provided:

Here is a link to the cosmos error definitions:
https://github.com/dcos/cosmos/blob/master/cosmos-server/src/main/scala/com/mesosphere/cosmos/CosmosError.scala


mesosphere-web:
Reviewer: @rcorral

MatApple:
Not your code, but there is a typo in the 3rd image. `You can got to..` should be `You can go to..`

src/js/components/CosmosErrorMessage.js
```diff
@@ -0,0 +1,97 @@
...
+  }
+  getMessage() {
+    let {type, message} = this.props.error;
+
+    switch (type) {
+      // Specific uninstall errors
+      case 'IncompleteUninstall':
+      case 'UninstallNonExistentAppForPackage':
+        if (!message) {
+          message = 'Unable to uninstall package. Please check your connection and try again.';
```
MatApple: Is this error caused by a network issue?
If that were the case I'd expect the `type` to be defined client side. If these are client side errors then perhaps we should use constants.

src/js/components/CosmosErrorMessage.js
```diff
@@ -0,0 +1,97 @@
...
+        break;
+    }
+
+    return message;
+  }
+
+  appendRepositoryLink(message) {
+    return (
+      <span>
+        {`${message} You can got to the `}
```
MatApple: Woops - I thought the typo was coming from the server :)

src/js/components/CosmosErrorMessage.js
```diff
@@ -0,0 +1,97 @@
...
+          {this.getMessage()}
+        </div>
+      </div>
+    );
+  }
+};
+
+CosmosErrorMessage.defaultProps = {
+  className: 'text-overflow-break-word column-small-8 column-small-offset-2 column-medium-6 column-medium-offset-3',
+  error: {message: 'An error occured.'},
```
MatApple: Perhaps we should put a default `type` in here too.

src/js/components/CosmosErrorMessage.js
```diff
@@ -0,0 +1,97 @@
...
+  header: 'An Error Occured',
```
MatApple: The default case will render a modal with a heading and message:
### An Error Occured
An Error Occured
___
Maybe the message could provide some type of action like `Please try again` instead of a blanketing error message?

src/js/components/CosmosErrorMessage.js
```diff
@@ -0,0 +1,97 @@
...
+      case 'RepositoryUriConnection':
+      case 'RepositoryUriSyntax':
+        if (!message) {
+          message = 'The URL for a repository is not valid, or its host did not resolve. You might need to change the URL of this repository.';
+        }
+        message = this.appendRepositoryLink(message);
+        break;
+    }
+
+    return message;
```
kennyt: what do you think about prepending the `type` to the message? like how javascript does it

`TypeError: 'forEach' is not a function...'`

src/js/components/CosmosErrorMessage.js
```diff
@@ -0,0 +1,97 @@
...
+    switch (type) {
+      // Specific uninstall errors
+      case 'IncompleteUninstall':
+      case 'UninstallNonExistentAppForPackage':
+        if (!message) {
+          message = 'Unable to uninstall package. Please check your connection and try again.';
+        }
+        break;
+
+      // // Specific install errors
```
mlunoe: Double comment for everybody!

rcorral: huh?

mlunoe: It was a reminder for myself to remove the double comment

src/js/components/CosmosErrorMessage.js
mlunoe: Fixed

src/js/components/CosmosErrorMessage.js
mlunoe: I removed this fallback, since we expect the server to always provide a message

src/js/components/CosmosErrorMessage.js
mlunoe: Removed the fallback, since we expect the server to always to have a message

src/js/components/CosmosErrorMessage.js
mlunoe: Added "Please try again". I don't actually expect this case to ever happen, this is only in the case where we don't actually get a message from the server, which we always should

src/js/components/CosmosErrorMessage.js
mlunoe: No type needed, since we only need to add the page instructions if this is a repository related error

src/js/components/CosmosErrorMessage.js
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    return message;
+  }
+
+  appendRepositoryLink(message) {
+    return (
+      <span>
+        {`${message} You can go to the `}
+        <a href="/#/system/overview/repositories/">Repositories Settings</a>
```
kennyt: let's use a `Link` tag if possible

mlunoe: It's not since this will be used in the modal too. So I'd rather have this work everywhere we use it, rather than adding context to modal somehow.

PR #70, opened 2016-04-27
Add tab for deployments

philipnrmn:
This PR introduces a tabbed navigation to the services page, and splits the functionality previously in `ServicesPage` out to `ServicesTab`. The `DeploymentsTab` currently only shows an empty state. 

Note that in order to avoid conflicts (eg with a service named `deployments`) the service detail routes have been updated to use the `service` prefix, eg `/services/web-server` is now accessible at `/service/web-server`.

### DIFF
--- a/src/js/pages/services/DeploymentsTab.js
+++ b/src/js/pages/services/DeploymentsTab.js
@@ -1,0 +1,27 @@
+import React from 'react';
+
+import AlertPanel from '../../components/AlertPanel';
+
+var DeploymentsTab = React.createClass({
+
+  displayName: 'DeploymentsTab',
+
+  renderEmpty: function () {
+    return (
+      <AlertPanel
+        title="No Deployments"
+        iconClassName="icon icon-sprite icon-sprite-jumbo
+            icon-sprite-jumbo-white icon-services flush-top">
+        <p className="flush">Active deployments will be shown here.</p>
+      </AlertPanel>
+    );
+  },
+
+  render: function () {
+    return this.renderEmpty();
+  }
+
+});
+
+module.exports = DeploymentsTab;
+
--- a/src/js/pages/services/DeploymentsTab.js
+++ b/src/js/pages/services/DeploymentsTab.js
@@ -2,11 +2,13 @@ import React from 'react';
 
 import AlertPanel from '../../components/AlertPanel';
 
-var DeploymentsTab = React.createClass({
+class DeploymentsTab extends React.Component {
 
-  displayName: 'DeploymentsTab',
+  constructor() {
+    super(...arguments);
+  }
 
-  renderEmpty: function () {
+  renderEmpty() {
     return (
       <AlertPanel
         title="No Deployments"
@@ -15,13 +17,13 @@ var DeploymentsTab = React.createClass({
         <p className="flush">Active deployments will be shown here.</p>
       </AlertPanel>
     );
-  },
+  }
 
-  render: function () {
+  render() {
     return this.renderEmpty();
   }
 
-});
+}
 
 module.exports = DeploymentsTab;
 
--- a/src/js/pages/services/ServicesTab.js
+++ b/src/js/pages/services/ServicesTab.js
@@ -1,0 +1,158 @@
+import React from 'react';
+import {StoreMixin} from 'mesosphere-shared-reactjs';
+
+import AlertPanel from '../../components/AlertPanel';
+import Config from '../../config/Config';
+import DCOSStore from '../../stores/DCOSStore';
+import FilterHeadline from '../../components/FilterHeadline';
+import SaveStateMixin from '../../mixins/SaveStateMixin';
+import Service from '../../structs/Service';
+import ServiceDetail from '../../components/ServiceDetail';
+import ServiceFilterTypes from '../../constants/ServiceFilterTypes';
+import ServiceSidebarFilters from '../../components/ServiceSidebarFilters';
+import ServicesTable from '../../components/ServicesTable';
+import ServiceTree from '../../structs/ServiceTree';
+import SidebarActions from '../../events/SidebarActions';
+import SidePanels from '../../components/SidePanels';
+
+var DEFAULT_FILTER_OPTIONS = {
+  filterHealth: null,
+  searchString: ''
+};
+
+let saveState_properties = Object.keys(DEFAULT_FILTER_OPTIONS);
+
+var ServicesTab = React.createClass({
+
+  displayName: 'ServicesTab',
+
+  saveState_key: 'servicesPage',
+
+  saveState_properties,
+
+  mixins: [SaveStateMixin, StoreMixin],
+
+  statics: {
+    // Static life cycle method from react router, that will be called
+    // 'when a handler is about to render', i.e. on route change:
+    // https://github.com/rackt/react-router/
+    // blob/master/docs/api/components/RouteHandler.md
+    willTransitionTo: function () {
+      SidebarActions.close();
+    }
+  },
+
+  contextTypes: {
+    router: React.PropTypes.func
+  },
+
+  getInitialState: function () {
+    return Object.assign({}, DEFAULT_FILTER_OPTIONS);
+  },
+
+  componentWillMount: function () {
+    this.store_listeners = [{name: 'dcos', events: ['change']}];
+  },
+
+  handleFilterChange: function (filterValues, filterType) {
+    var stateChanges = Object.assign({}, this.state);
+    stateChanges[filterType] = filterValues;
+
+    this.setState(stateChanges);
+  },
+
+  resetFilterQueryParams: function () {
+    let router = this.context.router;
+    let queryParams = router.getCurrentQuery();
+
+    Object.values(ServiceFilterTypes).forEach(function (filterKey) {
+      delete queryParams[filterKey];
+    });
+
+    router.transitionTo(router.getCurrentPathname(), {}, queryParams);
+  },
+
+  resetFilter: function () {
+    var state = Object.assign({}, this.state, DEFAULT_FILTER_OPTIONS);
+    this.setState(state, this.resetFilterQueryParams);
+  },
+
+  getContents: function (id) {
+    // Render loading screen
+    if (!DCOSStore.dataProcessed) {
+      return (
+        <div className="container container-fluid container-pod text-align-center
+            vertical-center inverse">
+          <div className="row">
+            <div className="ball-scale">
+              <div />
+            </div>
+          </div>
+        </div>
+      );
+    }
+
+    // Find item in root tree and default to root tree if there is no match
+    let item = DCOSStore.serviceTree.findItemById(id) || DCOSStore.serviceTree;
+
+    // Render service table
+    if (item instanceof ServiceTree && item.getItems().length > 0) {
+      let {state} = this;
+      let services = item.getItems();
+      let filteredServices = item.filterItemsByFilter({
+        health: state.filterHealth,
+        name: state.searchString
+      }).getItems();
+
+      return (
+        <div className="flex-box flush flex-mobile-column">
+          <ServiceSidebarFilters
+            handleFilterChange={this.handleFilterChange}
+            services={services} />
+          <div className="flex-grow">
+            <FilterHeadline
+              inverseStyle={true}
+              onReset={this.resetFilter}
+              name="Services"
+              currentLength={filteredServices.length}
+              totalLength={services.length} />
+            <ServiceSearchFilter
+              handleFilterChange={this.handleFilterChange} />
+            <ServicesTable
+              services={filteredServices} />
+          </div>
+          <SidePanels
+            params={this.props.params}
+            openedPage="services"/>
+        </div>
+      );
+    }
+
+    // Render service detail
+    if (item instanceof Service) {
+      return (<ServiceDetail service={item} />);
+    }
+
+    // Render empty panel
+    return (
+      <AlertPanel
+        title="No Services Installed"
+        iconClassName="icon icon-sprite icon-sprite-jumbo
+          icon-sprite-jumbo-white icon-services flush-top">
+        <p className="flush-bottom">
+          Use the {Config.productName} command line tools to find and install
+          services.
+        </p>
+      </AlertPanel>
+    );
+  },
+
+  render: function () {
+    let {id} = this.props.params;
+
+    return this.getContents(decodeURIComponent(id));
+  }
+
+});
+
+module.exports = ServicesTab;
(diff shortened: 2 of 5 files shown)


mesosphere-web:
Reviewer: @MatApple

src/js/routes/services.js
```diff
@@ -1,37 +1,60 @@
...
-      path: 'task-detail/:taskID/?'
-    }
-  ]
+  services: {
+    type: Route,
+    name: 'services',
+    path: 'services/?',
+    handler: ServicesPage,
+    children: [
+      // Undocumented RR 0.13.5 behaviour: unnamed routes are treated as default
```
mlunoe: Could we make this explicit by adding a `Redirect`, like we do everywhere else?

src/js/routes/services.js
```diff
@@ -1,37 +1,60 @@
...
+      },
+      {
+        type: Route,
+        name: 'services-deployments',
+        path: 'deployments/',
+        handler: DeploymentsTab
+      }
+    ]
+  },
+  serviceDetails: {
```
mlunoe: Could we do something similar to what we do here to not add another top level route?
https://github.com/dcos/dcos-ui/pull/56/files#diff-cd50a5d0c3afa7ada47e360dc9000693R41

philipnrmn: This is way neater. Thanks.

src/js/routes/services.js
philipnrmn: Yep good shout.

MatApple:
We get two active tabs at the top! http://cl.ly/2d431P0i1Q00/Image%202016-04-28%20at%2012.03.05%20PM.png

MatApple:
Also, think we have to change the tab name. This is from design -

MatApple:
When I click on a service in the Services table, nothing happens! Am I missing something or will this change in another PR?

MatApple:
Deployments don't seem to be showing up. I can verify that they are there in Marathon but not showing here.

src/js/pages/services/DeploymentsTab.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+
+import AlertPanel from '../../components/AlertPanel';
+
+var DeploymentsTab = React.createClass({
```
MatApple: Is there a reason for not extending `React.Component`?

src/js/pages/services/ServicesTab.js
```diff
@@ -0,0 +1,223 @@
...
+import SidePanels from '../../components/SidePanels';
+
+var DEFAULT_FILTER_OPTIONS = {
+  filterHealth: null,
+  searchString: ''
+};
+
+let saveState_properties = Object.keys(DEFAULT_FILTER_OPTIONS);
+
+var ServicesTab = React.createClass({
```
MatApple: How about `class ServicesTab extends mixins(SaveStateMixin, StoreMixin)` :)

src/js/pages/services/ServicesTab.js
```diff
@@ -0,0 +1,223 @@
...
+
+  statics: {
+    // Static life cycle method from react router, that will be called
+    // 'when a handler is about to render', i.e. on route change:
+    // https://github.com/rackt/react-router/
+    // blob/master/docs/api/components/RouteHandler.md
+    willTransitionTo: function () {
+      SidebarActions.close();
+    }
+  },
```
MatApple: Once we use ES6, we can put `statics` and `contextTypes` at the bottom of the file.

src/js/pages/services/ServicesTab.js
rcorral: Why no an es6 class?

src/js/pages/services/ServicesTab.js
```diff
@@ -0,0 +1,223 @@
...
+    stateChanges[filterType] = filterValues;
+
+    this.setState(stateChanges);
+  },
+
+  resetFilterQueryParams: function () {
+    let router = this.context.router;
+    let queryParams = router.getCurrentQuery();
+
+    Object.values(ServiceFilterTypes).forEach(function (filterKey) {
```
MatApple: `Object.values` has really bad browser support and still considered experimental. Not sure its in `babel-polyfill` is it?

src/js/pages/services/ServicesTab.js
kennyt: let's create this component by `class ServicesTab extends React.Component` (i forget the technical name for this)

i always refer to the component `ServiceOverlay` if i need a reference

src/js/pages/services/DeploymentsTab.js
philipnrmn: No. I guess this is preferred? I'll open a PR with a documentation update if so.

src/js/pages/services/ServicesTab.js
```diff
@@ -0,0 +1,223 @@
...
+          icon-sprite-jumbo-white icon-services flush-top">
+        <p className="flush-bottom">
+          Use the {Config.productName} command line tools to find and install
+          services.
+        </p>
+      </AlertPanel>
+    );
+  },
+
+  getContents: function (id) {
```
MatApple: This function is defined twice

src/js/pages/services/ServicesTab.js
philipnrmn: In an earlier PR both the `ServicesPage` and the `ServicesTab` were converted to an ES6 class. That PR suffered from constant merge issues as the `ServicesPage` on the feature branch was updated. I'll resist making the same mistake again.

src/js/pages/services/ServicesTab.js
philipnrmn: This was adopted from `ServicesPage` and not introduced in this PR, so I'd prefer not to address it here if that's cool.

philipnrmn:
Hey Mat, thanks for the review. Couple of responses:


This is an issue with the tabs mixin which I'll address in a separate PR


The tab wording change was agreed with @leemunroe and is already reflected in the latest Services Deployments design.

src/js/pages/services/ServicesTab.js
philipnrmn: ref https://github.com/dcos/dcos-ui/pull/70#discussion_r61494313

src/js/pages/services/ServicesTab.js
```diff
@@ -0,0 +1,223 @@
...
+    willTransitionTo: function () {
+      SidebarActions.close();
+    }
+  },
+
+  contextTypes: {
+    router: React.PropTypes.func
+  },
+
+  getInitialState: function () {
```
MatApple: Lets use `getInitialState() {` style for our method declarations :)

src/js/pages/services/ServicesTab.js
philipnrmn: Thanks for the catch. Too many rebases!

src/js/routes/services.js
```diff
@@ -1,37 +1,69 @@
 import Router from 'react-router';
-let Route = Router.Route;
+let {Route, Redirect} = Router;
```
MatApple: Could just do `import {Route, Redirect} from 'react-router';`

src/js/pages/services/ServicesTab.js
philipnrmn: That would require this to be a class, wouldn't it?

src/js/routes/services.js
```diff
@@ -1,37 +1,69 @@
...
 
+import DeploymentsTab from '../pages/services/DeploymentsTab';
 import ServiceOverlay from '../components/ServiceOverlay';
 import ServicesPage from '../pages/ServicesPage';
+import ServicesTab from '../pages/services/ServicesTab';
 
-let servicesRoutes = {
+let services = {
```
MatApple: We try to be more explicit when naming variables. Have a look at how we name elsewhere (`/routes/universe.js`, `/routes/nodes.js` etc). Let's keep `servicesRoutes` 👍

src/js/pages/services/ServicesTab.js
kennyt: yeah, `Object.values` is in the polyfill so this is good

src/js/pages/services/DeploymentsTab.js
pierlo-upitup: I'd say for new components with should follow the new class paradigm. +1 for the updated docs .

src/js/routes/services.js
philipnrmn: That's an interesting point. Happy to obey the convention, but since we import it using `import services from 'services.js';` should we think about updating the module names too?

src/js/pages/services/DeploymentsTab.js
MatApple: We try to use ES6 whenever we have a choice :) What documentation are you referring too?

src/js/pages/services/ServicesTab.js
MatApple: I still think we should change this to be an ES6 class irrespective of merge issues. Modern, consistent code should always be chosen over less work when rebasing.

src/js/pages/services/ServicesTab.js
MatApple: I understand you may not have written this code, but we should still change it. We shouldn't be merging anything that has show stoppers in it, irrespective of who wrote it. :)

src/js/routes/services.js
MatApple: I see where you're coming from. Not sure we need to go as far as renaming the modules (since they are nested within the routes directory, thus providing context) but it would be nice if we at least imported like `import servicesRoutes from './routes/services';`. But I'd suggest that is out of scope here.

src/js/pages/services/ServicesTab.js
philipnrmn: I have to disagree. I don't think it's justifiable to make purely stylistic changes to code on which multiple members of the team are currently working.

src/js/pages/services/DeploymentsTab.js
philipnrmn: It would be really nice to know about those sorts of conventions before tripping over them at review-time, so I was thinking about some new documentation to cover it.

MatApple:
Looks like we loose our active sidebar tab because service detail routes were changed -

src/js/pages/services/ServicesTab.js
philipnrmn: thanks @kennyt 

@MatApple totally agreed, but this particular line is already merged.

philipnrmn:
Looks like a good argument for keeping the `/services/` route prefix. That's annoying. 

I'll update the routing again.

src/js/pages/services/DeploymentsTab.js
rcorral: Sorry you're going through this Phil. I think it was helpful for the rest of the team when we were in HH. Maybe it would make sense for us to do an overview with you since you weren't able to make it.

For now would you mind making this an es6 class since this is a new component?

src/js/routes/services.js
```diff
@@ -1,37 +1,69 @@
...
       type: Route,
-      name: 'services-detail',
-      path: ':id/?'
+      name: 'services-page',
+      path: '/services/?',
+      children: [
+        {
+          type: Route,
+          name: 'services-all',
+          path: 'all/',
```
kennyt: @rcorral i remember you had something to say on doing `/#/services/all`

philipnrmn: @kennyt as in we shouldn't? Would be interested to hear suggestions for a better routing scheme - not sure I'm happy with routes like `/services/all/%2Fsome-service`

philipnrmn:
OK, I have now addressed all feedback except for converting `ServicesPage` and `ServicesTab` to ES6 classes.

src/js/pages/services/ServicesTab.js
rcorral: So just so it's clear for myself, this is a brand new component, but some parts came from the `ServicesPage` component?
Phil, which other PRs depend on this change that would cause merge conflicts for people? If there's none, I don't understand the opposition for sticking with this style.

src/js/routes/services.js
rcorral: @philipnrmn can we make the routes work like this:
`/services` <-- All services list
`/services/:id` <-- Detail page of a service
`/services/:id/ui` <-- Opens the service overlay and the UI for the service

I have some comments about the rest of the URLs, but non blockers for this PR.

src/js/pages/services/ServicesTab.js
MatApple: It looks like this will disrupt at least 1 other PR so lets 🚢. Thanks @philipnrmn!

MatApple:
Thanks @philipnrmn!

rcorral:
@philipnrmn could you see this [last comment](https://github.com/dcos/dcos-ui/pull/70#discussion_r61514365) before we ship.

kennyt:
conflicts

src/js/routes/services.js
philipnrmn: Sure, we can do that. The reason we didn't already was that the route to the deployments tab is `/services/deployments`, do there's a potential conflict if a user adds a service named `deployments`. What's your preferred deployments route?

rcorral: Aren't the `:id`s always prefixed with at least one slash?
Therefore it wouldn't be possible for a user to create an app called `deployments`?

philipnrmn:
my hundredth rebase :(

philipnrmn:
OK @rcorral I fixed the routing up as you suggested. Note that this re-introduces the tab misbehaviour @MatApple pointed out here: https://github.com/dcos/dcos-ui/pull/70#issuecomment-215530051. I'll fix that in a separate PR as it's an issue with the `TabMixin`'s route detection.

src/js/routes/services.js
philipnrmn: ...good point.

src/js/pages/services/ServicesTab.js
philipnrmn: Hey @rcorral, sorry for the delayed response - didn't see this thread yesterday for some reason. 

The outer `ServicesPage` module remains, but all its content is split out into  `ServicesTab`.  That means that whenever a PR which changes to `ServicesPage` gets merged - and pretty much half the PRs into the `feature/unified-ui` branch touch it - the work of splitting `ServicesPage` has to be redone during the subsequent rebase.

It's worth noting that this is reflected in the git history `ServicesTab` is actually the result of a `git mv ServicesPage.js services/ServicesTab.js` - this means that `ServicesTab` is technically not the new component, rather `ServicesPage` is (the identical naming makes this confusing and git doesn't handle it perfectly). 

This caused a lot of overhead in a previous PR on a private repository, where I had updated the relevant classes to ES6 style, because it wasn't just a copy-paste job - things like initial component state had to be updated as well. This was painstaking and error-prone, as well as being rather annoying for everybody involved. 

`I don't understand the opposition for sticking with this style.`
Excuse the quote, but I wanted to address it directly because I seem to be having a lot of difficulty communicating one thing: I don't oppose the adoption of ES6 classes for components. I would like this component to be an ES6 class. I will convert it to an ES6 class in a future PR. 

I am opposing making such a change _in this PR_ because I want to avoid either me or any other developer getting stuck in a rebase loop as a result.

kennyt:
:ship:

PR #72, opened 2016-04-27
Feature - Service page text search

aldipower:
This reintroduces the service page text search, which enables a search of the complete Marathon Services (Groups, Apps).
The search is stored as a query param and AND-connected to the sidebar filters.

The search field could be placed anywhere in the application, so it's basically a global search placed locally.

Thx @kennyt for the help and preliminary work.

search1
 
search2

### DIFF
--- a/src/js/components/ServiceSearchFilter.js
+++ b/src/js/components/ServiceSearchFilter.js
@@ -1,0 +1,61 @@
+import mixin from 'reactjs-mixin';
+import React from 'react';
+
+import FilterInputText from './FilterInputText';
+import QueryParamsMixin from '../mixins/QueryParamsMixin';
+import ServiceFilterTypes from '../constants/ServiceFilterTypes';
+
+const METHODS_TO_BIND = ['setSearchString'];
+
+class ServiceSearchFilter extends mixin(QueryParamsMixin) {
+  static get defaultProps() {
+    return {
+    };
+  }
+
+  constructor() {
+    super();
+
+    this.state = {
+      searchString: ''
+    };
+
+    METHODS_TO_BIND.forEach((method) => {
+      this[method] = this[method].bind(this);
+    });
+  }
+
+  componentDidMount() {
+    this.updateFilterStatus();
+  }
+
+  componentWillReceiveProps() {
+    this.updateFilterStatus();
+  }
+
+  setSearchString(filterValue) {
+    this.setQueryParam(ServiceFilterTypes.TEXT, filterValue);
+    this.props.handleFilterChange(filterValue, ServiceFilterTypes.TEXT);
+  }
+
+  updateFilterStatus() {
+    let {state} = this;
+    let searchString =
+      this.getQueryParamObject()[ServiceFilterTypes.TEXT] || '';
+
+    if (searchString !== state.searchString) {
+      this.setState({searchString},
+        this.props.handleFilterChange.bind(null, searchString, ServiceFilterTypes.TEXT));
+    }
+  }
+
+  render() {
+    return (
+      <FilterInputText
+        searchString={this.state.searchString}
+        handleFilterChange={this.setSearchString} />
+    );
+  }
+};
+
+module.exports = ServiceSearchFilter;
--- a/src/js/components/ServiceSearchFilter.js
+++ b/src/js/components/ServiceSearchFilter.js
@@ -52,8 +52,10 @@ class ServiceSearchFilter extends mixin(QueryParamsMixin) {
   render() {
     return (
       <FilterInputText
-        searchString={this.state.searchString}
-        handleFilterChange={this.setSearchString} />
+        handleFilterChange={this.setSearchString}
+        inverseStyle={true}
+        placeholder="Search"
+        searchString={this.state.searchString} />
     );
   }
 };
--- a/src/js/components/ServiceSearchFilter.js
+++ b/src/js/components/ServiceSearchFilter.js
@@ -1,5 +1,7 @@
 import mixin from 'reactjs-mixin';
+/* eslint-disable no-unused-vars */
 import React from 'react';
+/* eslint-enable no-unused-vars */
 
 import FilterInputText from './FilterInputText';
 import QueryParamsMixin from '../mixins/QueryParamsMixin';
--- a/src/js/components/ServiceSearchFilter.js
+++ b/src/js/components/ServiceSearchFilter.js
@@ -10,11 +10,6 @@ import ServiceFilterTypes from '../constants/ServiceFilterTypes';
 const METHODS_TO_BIND = ['setSearchString'];
 
 class ServiceSearchFilter extends mixin(QueryParamsMixin) {
-  static get defaultProps() {
-    return {
-    };
-  }
-
   constructor() {
     super();
 
@@ -22,7 +17,7 @@ class ServiceSearchFilter extends mixin(QueryParamsMixin) {
       searchString: ''
     };
 
-    METHODS_TO_BIND.forEach((method) => {
+    METHODS_TO_BIND.forEach(method => {
       this[method] = this[method].bind(this);
     });
   }
--- a/src/js/components/ServiceSearchFilter.js
+++ b/src/js/components/ServiceSearchFilter.js
@@ -17,7 +17,7 @@ class ServiceSearchFilter extends mixin(QueryParamsMixin) {
       searchString: ''
     };
 
-    METHODS_TO_BIND.forEach(method => {
+    METHODS_TO_BIND.forEach((method) => {
       this[method] = this[method].bind(this);
     });
   }
--- a/src/js/stores/MarathonStore.js
+++ b/src/js/stores/MarathonStore.js
@@ -160,7 +160,11 @@ var MarathonStore = Store.createStore({
   },
 
   processMarathonGroups: function (data) {
-    let groups = new ServiceTree(data);
+    let groups = new ServiceTree({items: data.apps, filterProperties: {
+      name: function (item) {
+        return item.id;
+      }
+    }});
 
     let apps = groups.reduceItems(function (map, item) {
       if (item instanceof Service) {
--- a/src/js/stores/MarathonStore.js
+++ b/src/js/stores/MarathonStore.js
@@ -160,11 +160,13 @@ var MarathonStore = Store.createStore({
   },
 
   processMarathonGroups: function (data) {
-    let groups = new ServiceTree({items: data.apps, filterProperties: {
+    data.filterProperties = {
       name: function (item) {
         return item.id;
       }
-    }});
+    };
+
+    let groups = new ServiceTree(data);
 
     let apps = groups.reduceItems(function (map, item) {
       if (item instanceof Service) {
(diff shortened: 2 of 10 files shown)


mesosphere-web:
Reviewer: @orlandohohmeier

src/js/components/ServiceSearchFilter.js
```diff
@@ -0,0 +1,65 @@
+import mixin from 'reactjs-mixin';
+/* eslint-disable no-unused-vars */
+import React from 'react';
+/* eslint-enable no-unused-vars */
```
pierlo-upitup: Btw – we could also just
```
// eslint-disable no-unused-vars
import React from 'react';
```
rather than having to disable/re-enable each time. cc @rcorral

aldipower: Nice, thx.

src/js/components/ServiceSearchFilter.js
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  constructor() {
+    super();
+
+    this.state = {
+      searchString: ''
+    };
+
+    METHODS_TO_BIND.forEach((method) => {
```
pierlo-upitup: Is there a guideline consensus regarding no `()` for single-argument fat-arrows functions? If not, I'd suggest `forEach(method => {` .

src/js/components/ServiceSearchFilter.js
```diff
@@ -0,0 +1,65 @@
...
+import FilterInputText from './FilterInputText';
+import QueryParamsMixin from '../mixins/QueryParamsMixin';
+import ServiceFilterTypes from '../constants/ServiceFilterTypes';
+
+const METHODS_TO_BIND = ['setSearchString'];
+
+class ServiceSearchFilter extends mixin(QueryParamsMixin) {
+  static get defaultProps() {
+    return {
+    };
```
pierlo-upitup: What's the use for returning an empty `props` object?

src/js/components/ServiceSearchFilter.js
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  constructor() {
+    super();
+
+    this.state = {
+      searchString: ''
```
pierlo-upitup: shouldn't this be a 
```
getInitialState() {
    return { searchString: '' };
}
```
?

src/js/components/ServiceSearchFilter.js
aldipower: None. :)

src/js/components/ServiceSearchFilter.js
aldipower: Not with the new Class syntax. You set the initial state object (which is bound to the class) in the constructor.

src/js/components/ServiceSearchFilter.js
aldipower: I think there is no guideline. I also prefer to eliminate the parenthesis.

src/js/components/ServiceSearchFilter.js
```diff
@@ -0,0 +1,65 @@
...
+        this.props.handleFilterChange.bind(null, searchString, ServiceFilterTypes.TEXT));
+    }
+  }
+
+  render() {
+    return (
+      <FilterInputText
+        handleFilterChange={this.setSearchString}
+        inverseStyle={true}
+        placeholder="Search"
```
pierlo-upitup: How about `Search all services` ?

aldipower: That is not what the layout is saying.
Maybe a little bit long too.

src/js/components/ServiceSearchFilter.js
pierlo-upitup: So it makes sense to get rid of this, I guess?

aldipower: Absolutely.

src/js/components/ServiceSearchFilter.js
pierlo-upitup: cc @rcorral

src/js/components/ServiceSearchFilter.js
pierlo-upitup: Then again – this probably should be a consensus decision. Let's skip it for now.

aldipower: ok.

pierlo-upitup:
The feature works and the code LGTM after all the comments are resolved. Thanks!

mlunoe:
Closing until tests are up

aldipower:
Tests added

src/js/structs/__tests__/ServiceTree-test.js
```diff
@@ -147,6 +148,79 @@ describe('ServiceTree', function () {
...
+          {id: 'gamma', cmd: 'cmd', labels: {RANDOM_LABEL: 'random'}}
+        ],
+        groups: [
+          {
+            id: '/test', apps: [
+              {id: 'foo', cmd: 'cmd'},
+              {id: 'bar', cmd: 'cmd'}
+            ], groups: []
+          }
+        ],
```
pierlo-upitup: Would be nice to have the indentation agree on one format throughout!  I suggest something like
```
apps: [
	{
	  id: 'alpha',
	  cmd: 'cmd'
	},
	{
	  id: 'beta',
	  cmd: 'cmd',
	  labels: {DCOS_PACKAGE_FRAMEWORK_NAME: 'beta'},
	  tasksHealthy: 1,
	  tasksRunning: 1
	},
	{
	  id: 'gamma', 
	  cmd: 'cmd', 
	  labels: {
	  	RANDOM_LABEL: 'random'
  	  }
  	}
],
groups: [
	{
	  id: '/test', apps: [
	    {
	   	  id: 'foo', 
	   	  cmd: 'cmd'
	   	},
	    {
	      id: 'bar', 
	      cmd: 'cmd'
	    }
	 ], 
	 groups: []
	}
],
```
what do you think?

aldipower: Looks beautiful to me. Change it.

src/js/structs/__tests__/ServiceTree-test.js
```diff
@@ -147,6 +148,79 @@ describe('ServiceTree', function () {
...
+    it('should filter by health', function () {
+      let filteredServices = this.instance.filterItemsByFilter({
+        health: [HealthTypes.HEALTHY]
+      }).getItems();
+
+      expect(filteredServices.length).toEqual(1);
+      expect(filteredServices[0].id).toEqual('beta');
+    });
+
+    it('should filter AND-connected', function () {
```
pierlo-upitup: I suggest `it should perform a union with multiple filters`

pierlo-upitup: Or `it should perform a logical AND with multiple filters`

src/js/stores/MarathonStore.js
```diff
@@ -160,6 +160,12 @@ var MarathonStore = Store.createStore({
   },
 
   processMarathonGroups: function (data) {
+    data.filterProperties = {
+      name: function (item) {
+        return item.id;
+      }
+    };
+
```
pierlo-upitup: I think this mutation does not belong here. I need to think about this a bit more.

src/js/components/ServiceSearchFilter.js
```diff
@@ -0,0 +1,60 @@
...
+
+class ServiceSearchFilter extends mixin(QueryParamsMixin) {
+  constructor() {
+    super();
+
+    this.state = {
+      searchString: ''
+    };
+
+    METHODS_TO_BIND.forEach(method => {
```
orlandohohmeier: Please don't omit the parentheses, even if only one parameter is present to stay consistent with the rest of our code base.

**Examples**
arrow-functions

src/js/stores/MarathonStore.js
orlandohohmeier: I've just refactored the List and Tree struct (https://github.com/dcos/dcos-ui/pull/75), as it has always bothered me, that you have to pass the filter properties as constructor parameter.  With the change in place you're able to pass a definition of filterProperties whenever you filter items by text.

BTW: If you don't want to use the fancy `filterItemsByText` method, you could use the plain `filterItems` method. `filterItems` works like the `Array.prototype.filter` method but is applied to the whole tree. This would allow you, for exmaple, to use something like `fuzzaldrin` to implement a fuzzy filter/search in a few lines.

src/js/components/ServiceSearchFilter.js
pierlo-upitup: I suggested that presuming there was no guideline for it. My bad!

aldipower: OK, will readd it.

aldipower:
I'll add some integration tests too.

src/js/structs/__tests__/ServiceTree-test.js
aldipower: Pure lyric. Thx.

aldipower:
I've add integration tests for the search text filter.
Also I've fixed the sidebar filter integration tests, the behavior changed with the logical AND.

src/js/components/ServiceSearchFilter.js
kennyt: i actually didn't know that existed. i would 100% rather do that. A lot of our codebase is with the `/*` way unfortunately

src/js/structs/__tests__/ServiceTree-test.js
```diff
@@ -147,26 +186,135 @@ describe('ServiceTree', function () {
     });
   });
 
+  describe('#filterItemsByFilter', function () {
+
+    beforeEach(function () {
+      this.instance = new ServiceTree({
```
kennyt: this code is repeated in almost all `describe` blocks. let's just put this in a `beforeEach` in the top-most `describe` block

aldipower: You're right there is a lot of repetition. But almost every object differs slightly. So it looks easier then it is.
For the sake of progress I like to refactor this in another PR.

src/js/stores/MarathonStore.js
kennyt: #75 is merged. can we use any of the logic from there to replace modifying the data object here?

src/js/structs/__tests__/ServiceTree-test.js
kennyt: yes one option would be to create the `this.instance` and bring everything that is similar into the top-most describe.

then in the individual describes, you can modify `this.instance` in the way you need to for those specific tests, which will cut down on the code.

in another PR sounds good to me. just be sure to do it soon, since I believe that more servicetree tests will be coming

kennyt:
after this [comment](https://github.com/dcos/dcos-ui/pull/72#discussion_r61501933) i think this looks good

src/js/structs/__tests__/ServiceTree-test.js
aldipower: Cool, yeah, I had it it that way in mind too.

src/js/stores/MarathonStore.js
aldipower: I don't see any logic there that would help?

pierlo-upitup: This needs a rebase on `unified-ui` now

PR #74, opened 2016-04-28
fix links to the documentation

virtualroot:


mesosphere-web:
Reviewer: @pierlo-upitup

daiyiwhale:
Thanks for cleaning up our docs links, this is very important information to set straight right away! Your changes have brought to my attention that the https://dcos.io/docs site does not redirect URLs in a smart way in every situation so I have created this ticket that you can follow if interested: https://dcosjira.atlassian.net/browse/DCOS-65

src/js/config/Config.js
```diff
@@ -9,7 +9,7 @@ var Config = {
   networkingAPIPrefix: '/networking/api/v1',
   cosmosAPIPrefix: '/package',
   delayAfterErrorCount: 5,
-  documentationURI: 'https://dcos.io/docs',
+  documentationURI: 'https://dcos.io/docs/latest',
```
daiyiwhale: Instead of `/latest` we should be hitting `/<dcos-version>`. There is currently not a clean way to get the DC/OS version in the correct format for the docs. The DC/OS version is at the `/dcos-metadata/dcos-version.json` endpoint but it returns version in `x.y.z` format rather than `x.y` needed for the docs site. 

I filed this ticket: https://dcosjira.atlassian.net/browse/DCOS-66

jfurrow:
The consensus is that we'll be using the full version number in the URL (e.g. `/1.7.1` in place of `/latest`) and then handle the necessary redirects on the docs websites.

daiyiwhale:
And we will be getting the full version number from `/dcos-metadata/dcos-version.json`?

PR #76, opened 2016-04-28
Refactor the service detail task view

orlandohohmeier:
task-tab-mesos-polling
Refactor the service detail and move the tasks view into it's own component - this enables us to only poll the Mesos state, while a user is looking at the tasks list.

### DIFF
--- a/src/js/components/ServiceDetailTaskTab.js
+++ b/src/js/components/ServiceDetailTaskTab.js
@@ -1,0 +1,37 @@
+import mixin from 'reactjs-mixin';
+import React from 'react';
+import {StoreMixin} from 'mesosphere-shared-reactjs';
+
+import MesosStateStore from '../stores/MesosStateStore';
+import Service from '../structs/Service';
+import TaskView from './TaskView';
+
+class ServiceDetailTaskTab extends mixin(StoreMixin) {
+
+  constructor() {
+    super(...arguments);
+
+    this.store_listeners = [
+      {name: 'state', events: ['success']}
+    ];
+  }
+
+  render() {
+    let tasks = MesosStateStore.getTasksByServiceId(this.props.service.getId());
+
+    return (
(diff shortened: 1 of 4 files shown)


mesosphere-web:
Reviewer: @mlunoe

src/js/components/ServiceDetailTaskTab.js
```diff
@@ -0,0 +1,37 @@
...
+import Service from '../structs/Service';
+import TaskView from './TaskView';
+
+class ServiceDetailTaskTab extends mixin(StoreMixin) {
+
+  constructor() {
+    super(...arguments);
+
+    this.store_listeners = [
+      {name: 'state', events: ['success']}
```
pierlo-upitup: I have no clue what this does at all.

kennyt: this adds a property that the `StoreMixin` will then use. this essentially makes the component listen to the `'state'` store for the `'success'` event (in which case it will then re-render)

src/js/components/__tests__/ServiceDetailTaskTab-test.js
```diff
@@ -0,0 +1,55 @@
...
+  });
+
+  describe('#render', function () {
+
+    it('renders task view', function () {
+      var taskView =  TestUtils
+        .findRenderedComponentWithType(this.instance, TaskView);
+
+      expect(taskView).toBeDefined();
+
```
kennyt: got an extra newline here

kennyt:
this is awesome

src/js/components/ServiceDetailTaskTab.js
pierlo-upitup: what is the `state` store?

pierlo-upitup: OK I found it. src/js/constants/StoreConfig.js:227

PR #79, opened 2016-04-29
Remove filter propeties from marathon store

aldipower:
Bring the filter properties into a better location.
Separation of concerns.

### DIFF
--- a/src/js/structs/Tree.js
+++ b/src/js/structs/Tree.js
@@ -127,11 +127,17 @@ module.exports = class Tree extends List {
         }, {ids: filter.ids});
       }
 
-      if (filter.name) {
-        let tree =
-          new this.constructor(Object.assign({},
-            this, {items: services, filterProperties: this.getFilterProperties()}));
-        services = tree.filterItemsByText(filter.name).getItems();
+      if (filter.id) {
+        let filterProperties = Object.assign({}, this.getFilterProperties(), {
+          name: function (item) {
+            return item.id;
+          }
+        });
+
+        let tree = new this.constructor(Object.assign({},
+          this, {items: services, filterProperties: filterProperties}));
+
+        services = tree.filterItemsByText(filter.id).getItems();
       }
 
       if (filter.health != null && filter.health.length !== 0) {
--- a/src/js/structs/Tree.js
+++ b/src/js/structs/Tree.js
@@ -134,10 +134,7 @@ module.exports = class Tree extends List {
           }
         });
 
-        let tree = new this.constructor(Object.assign({},
-          this, {items: services, filterProperties: filterProperties}));
-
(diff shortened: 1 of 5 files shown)


mesosphere-web:
Reviewer: @daiyiwhale

src/js/structs/Tree.js
```diff
@@ -127,11 +127,17 @@ module.exports = class Tree extends List {
...
 
-      if (filter.name) {
-        let tree =
-          new this.constructor(Object.assign({},
-            this, {items: services, filterProperties: this.getFilterProperties()}));
-        services = tree.filterItemsByText(filter.name).getItems();
+      if (filter.id) {
+        let filterProperties = Object.assign({}, this.getFilterProperties(), {
+          name: function (item) {
+            return item.id;
```
orlandohohmeier: Why are you using the item `id` for the `name` filter?

src/js/structs/Tree.js
```diff
@@ -127,11 +127,17 @@ module.exports = class Tree extends List {
...
+          }
+        });
+
+        let tree = new this.constructor(Object.assign({},
```
orlandohohmeier: You don't need to create a new instance of tree to define the filter properties – just pass them to the `filterItemsByText` as a second argument.

orlandohohmeier:
Not part of the PR, but let's please remove the `filterByFilter` method (legacy code) and use the simple `filterItems` method instead.  If `filterItems` is not sufficient, to do the service filtering, let's please introduces the respective method to the `ServiceTree` instead of the general `Tree` struct.

pierlo-upitup:
@aldipower needs a good ol' rebase

src/js/structs/Tree.js
aldipower: Danke.

aldipower:
@pierlo-upitup Gerebased.

@orlandohohmeier Yes, do you know if there is a ticket existent?

src/js/structs/Tree.js
aldipower: Because a lot of marathon services doesn't have a name property.

orlandohohmeier: All service instance (frameworks as well as applications) do provide an id (`getId`) and name (`getName`) getter – let's use them.

orlandohohmeier:
@aldipower Probably not... Would you mind creating one?

src/js/structs/Tree.js
aldipower: OK, I think I've got what you've meant now.

pierlo-upitup:
Tested against latest unified-ui and works. LGTM.

philipnrmn:
Beat me to the merge :)

PR #80, opened 2016-04-29
Remove inverseStyle Prop

daiyiwhale:
Allows for more control over component classes, requested here: https://github.com/dcos/dcos-ui/pull/56#discussion_r61338091

Depends on https://github.com/dcos/dcos-ui/pull/56

### DIFF
--- a/src/js/components/HealthTab.js
+++ b/src/js/components/HealthTab.js
@@ -143,7 +143,8 @@ class HealthTab extends React.Component {
             inverseStyle={true} />
           <UnitHealthDropdown
             initialID="all"
-            inverseStyle={true}
+            className="button dropdown-toggle text-align-left button-inverse"
+            dropdownMenuClassName="dropdown-menu inverse"
             onHealthSelection={this.handleHealthSelection} />
         </FilterBar>
         <Table
--- a/src/js/components/TaskView.js
+++ b/src/js/components/TaskView.js
@@ -129,15 +129,18 @@ class TaskView extends mixin(SaveStateMixin) {
   }
 
   getTaskTable(tasks) {
+    let {inverseStyle, parentRouter} = this.props;
+
     let classSet = classNames({
       'table table-borderless-outer table-borderless-inner-columns': true,
       'flush-bottom': true,
-      'inverse': this.props.inverseStyle
+      'inverse': inverseStyle
     });
+
     return (
       <TaskTable
         className={classSet}
-        parentRouter={this.props.parentRouter}
+        parentRouter={parentRouter}
         tasks={tasks} />
     );
   }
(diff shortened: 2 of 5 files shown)


mesosphere-web:
Reviewer: @kennyt

src/js/components/TaskView.js
```diff
@@ -159,6 +163,18 @@ class TaskView extends mixin(SaveStateMixin) {
   getContent() {
     let state = this.state;
     let tasks = this.props.tasks || [];
```
mlunoe: I know this wasn't you, but could we use `[]` as the default prop for `tasks`

mlunoe: Then we can also do this:
```js
let {inverseStyle , tasks} = this.props;
```

src/js/components/UnitHealthDropdown.js
```diff
@@ -34,23 +34,32 @@ class UnitHealthDropdown extends mixin(InternalStorageMixin) {
   }
 
   render() {
+    let props = this.props;
```
mlunoe: Same here:
```js
let {className, dropdownMenuClassName, onHealthSelection} = this.props;
```

src/js/components/DescriptionList.js
```diff
@@ -2,16 +2,16 @@ import React from 'react';
...
-    let {headline} = this.props;
+    let {headline, headlineClassName} = this.props;
     if (!headline) {
       return null;
     }
 
     // Wrap in headline element and classes
     return (
-      <h6 className="flush-top">
+      <h5 className={headlineClassName}>
```
mlunoe: Have you verified this change with design and checked the usage everywhere, that it still looks how we want it?

src/js/pages/nodes/NodeDetailPage.js
```diff
@@ -0,0 +1,255 @@
+import {Link} from 'react-router';
```
mlunoe: Could we `git mv` this file, so we can see the changes here?

src/js/components/icons/IconChevron.js
```diff
@@ -12,25 +12,23 @@ class IconBack extends React.Component {
 
     return (
       <svg
-        height="16px"
```
mlunoe: Why remove the width and height?

daiyiwhale: the width and height were being controlled by the parent containers and not these attributes

src/js/components/DescriptionList.js
mlunoe: It has been verified

src/js/components/HealthTab.js
```diff
@@ -131,7 +131,7 @@ class HealthTab extends React.Component {
       <div>
         <FilterHeadline
           currentLength={visibleData.length}
-          inverseStyle={false}
+          inverseStyle={true}
```
jfurrow: Should this still be here?

src/js/components/TaskView.js
```diff
@@ -168,17 +184,19 @@ class TaskView extends mixin(SaveStateMixin) {
...
+        <span className={headerClassSet}>
           {this.getHeaderText(tasks)}
         </span>
         <FilterBar>
           <FilterInputText
             className="flush-bottom"
             searchString={state.searchString}
             handleFilterChange={this.handleSearchStringChange}
-            inverseStyle={false} />
+            inverseStyle={this.props.inverseStyle} />
```
kennyt: this should be removed right?

kennyt:
is this rebased? all the changes in the PR are from this PR?

src/js/components/HealthTab.js
daiyiwhale: I figured that the `FilterHeadline` is so specific that there is really no scenario we will be trying to apply other classes to it. Also the inverse styling is applied to it in a complicated, I think to expose all the styles is unnecessary (especially given that in this project we would never need to use them)

src/js/components/TaskView.js
daiyiwhale: At the moment TaskViews are inverse in nodes/task-view and not inverse in services/task-view, but will soon both be inverse, so this is temporarily accounting for both scenarios

daiyiwhale:
It is now!

src/js/components/HealthTab.js
jfurrow: I'm replying to this via email, since I can't see it on Github: That makes
sense, thanks for the explanation!

jfurrow: Haha, and suddenly I can see the whole thread now... fucking Github.

PR #82, opened 2016-04-29
[Feature] Introduce Header layout component

daiyiwhale:
Introducing a `Header` component to wrap a very common layout pattern: icon aligned with title and subtitle and divider bar.

### DIFF
--- a/src/js/components/Header.js
+++ b/src/js/components/Header.js
@@ -1,0 +1,80 @@
+import React from 'react';
+
+class Header extends React.Component {
+
+  getIcon(icon, className) {
+    if (icon == null) {
+      return null;
+    }
+
+    return (
+      <div className="media-object-item">
+        <div className={className}>
+          {icon}
+        </div>
+      </div>
+    );
+  }
+
+  getTitle(title, className) {
+    if (title == null) {
+      return null;
+    }
+
+    return (
+      <h1 className={className}>
+        {title}
+      </h1>
+    );
+  }
+
+  getSubTitle(subTitle) {
+    if (subTitle == null) {
+      return null;
+    }
+
+    return (<div>{subTitle}</div>);
+  }
+
+  render() {
+    let props = this.props;
+
+    return (
+      <div className={props.className}>
+        <div className={props.dividerClassName}>
+          <div className={props.mediaWrapperClassName}>
+            <div className="media-object flex-box flex-box-align-vertical-center">
+              {this.getIcon(props.icon, props.iconClassName)}
+              <div className="media-object-item">
+                {this.getTitle(props.title, props.titleClassName)}
+                {this.getSubTitle(props.subTitle)}
+              </div>
+            </div>
+          </div>
+        </div>
+      </div>
+    );
+  }
+}
+
+Header.defaultProps = {
+  className: 'container container-fluid container-pod container-pod-short flush flush-top',
+  mediaWrapperClassName: 'media-object-spacing-wrapper media-object-spacing-narrow media-object-offset',
+  dividerClassName: 'container-pod container-pod-short flush-top container-pod-divider-bottom container-pod-divider-bottom-align-right container-pod-divider-inverse',
+  titleClassName: 'flush inverse',
+  iconClassName: 'icon icon-large icon-image-container icon-app-container',
+};
+
+Header.propTypes = {
+  icon: React.PropTypes.node,
+  subTitle: React.PropTypes.node,
+  title: React.PropTypes.string,
+
+  className: React.PropTypes.string,
+  dividerClassName: React.PropTypes.string,
+  mediaWrapperClassName: React.PropTypes.string,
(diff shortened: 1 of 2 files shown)


mesosphere-web:
Reviewer: @MatApple

MatApple:
Love the size of this PR @daiyiwhale! So nice to review! :)

src/js/components/Header.js
```diff
@@ -0,0 +1,80 @@
...
+class Header extends React.Component {
+
+  getIcon(icon, className) {
+    if (icon == null) {
+      return null;
+    }
+
+    return (
+      <div className="media-object-item">
+        <div className={className}>
```
MatApple: Think we can just do `{className}` right?

src/js/components/Header.js
```diff
@@ -0,0 +1,80 @@
...
+
+  render() {
+    let props = this.props;
+
+    return (
+      <div className={props.className}>
+        <div className={props.dividerClassName}>
+          <div className={props.mediaWrapperClassName}>
+            <div className="media-object flex-box flex-box-align-vertical-center">
+              {this.getIcon(props.icon, props.iconClassName)}
```
MatApple: What do you think about making these `this.get..` methods variables by `let props`? Personal preference but it might make this render method a little easier on the eye :)

kennyt: they'd then lose the `this` context and we'd have to have a methods to bind

mlunoe:
Looks good!

kennyt:
looks good. let's get some tests in here and merge

src/js/components/Header.js
```diff
@@ -0,0 +1,80 @@
+import React from 'react';
+
+class Header extends React.Component {
```
kennyt: wait, i've got one comment. let's rename this component into something else other than `Header` because it's really undescriptive

daiyiwhale: PageHeader?

src/js/components/Header.js
```diff
@@ -0,0 +1,80 @@
...
+    );
+  }
+}
+
+Header.defaultProps = {
+  className: 'container container-fluid container-pod container-pod-short flush flush-top',
+  mediaWrapperClassName: 'media-object-spacing-wrapper media-object-spacing-narrow media-object-offset',
+  dividerClassName: 'container-pod container-pod-short flush-top container-pod-divider-bottom container-pod-divider-bottom-align-right container-pod-divider-inverse',
+  titleClassName: 'flush inverse',
+  iconClassName: 'icon icon-large icon-image-container icon-app-container',
```
philipnrmn: I'm unsure about all these explicit class-related props on this component - it feels a bit overcomplicated and makes me question the usefulness of this component. 

If we do need these props, can we make them extensible using `classNames` or similar? At present they're just overriden, which means that every instance of the component has to stay in synch with these defaults.

mlunoe: That's exactly the situation that we have been burnt by before. Having some classes that cannot be removed/changed by the user if not needed

philipnrmn: That's the beauty of `classNames` though:

```
let defaults = "blue square";
let additions = "circle";
let subtractions = {square: false};
// prints "blue circle"
console.log(classNames(defaults, subtractions, additions));
```

philipnrmn: Still, deciding whether to use `classNames` or not is bookkeeping really. I feel like a situation where the user needs to manage the internal class structure of a component is a result of a bigger problem - either that the component is too complex and should be broken into smaller pieces, or that the code structure is being dictated by the styles.

philipnrmn: (apologies to @daiyiwhale for hijacking this PR to discuss what appears to be a project-level convention!)

daiyiwhale: Let's put it on the agenda for our next HH <> SF sync-up!

mlunoe: Right, but in the case of the `inverse` style this is something that will not change for the instance of the components lifetime. Why not just make it a property?
For anything related to the state of the component I would agree that `classNames` is a good path to go down.
But yes, let's take it at the SF <> HH sync meeting

daiyiwhale: I just made a PR that was removing the `inverseStyle` prop from many components due to popular demand :P

kennyt:
write some simple tests for this component and let's :ship: it!

src/js/components/Header.js
kennyt: 👍

src/js/components/Header.js
jfurrow: I had never though to use `classnames` in the way that @philipnrmn suggests, but I really like that.

src/js/components/Header.js
daiyiwhale: done!

jfurrow: Hi Chris! I'm testing Github right now :)

PR #85, opened 2016-04-29
Feature: Add Dialog component

Poltergeist:
This adds a Dialog component which wraps the Modal component,

You may add buttons with css `classNames` and `onClick` handler and `text` Also a `title` can be set with a prop and the `severity` which is based on the Constants, `info`, `warning` and `danger`.

This PR does only contain the component and only a basic styling (removing grey footer and header). 

Sorry for the Size of this PR, but 🎆 260 lines are tests 🎉 

image

### DIFF
--- a/src/js/components/Dialog.js
+++ b/src/js/components/Dialog.js
@@ -1,0 +1,93 @@
+import classNames from 'classnames';
+import React from 'react';
+
+import DialogSeverity from '../constants/DialogSeverity';
+import {Modal} from 'reactjs-components';
+
+class Dialog extends React.Component {
+  getButtons() {
+    return this.props.buttons.map(function (button, index) {
+      let classSet = classNames('button', button.className);
+      return (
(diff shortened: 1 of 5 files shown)


mesosphere-web:
Reviewer: @pierlo-upitup

rcorral:
@Poltergeist could we see some screenshots of what these Dialogs look like?

Poltergeist:
I added a screenshot as mentioned in the description there has been no css added yet.

rcorral:
It looks very similar to the Confirm Modal we already have in reactjs components.
http://mesosphere.github.io/reactjs-components/#confirm

Any chance we can use that or add more options to it?

pierlo-upitup:
As @rcorral said, what I see in the PR doesn't justify the creation of a new Dialog component that replicates the already existing Confirm component. I'm closing this – please re-open if I missed something, @Poltergeist .

PR #88, opened 2016-04-30
Show post uninstall notes, ref. DCOS-6181

mlunoe:
Adding post uninstall notes and aligning post install notes with post uninstall notes.

Post uninstall notes:

Post install notes:


mesosphere-web:
Reviewer: @rcorral

src/js/components/PackagesTable.js
```diff
@@ -252,26 +166,16 @@ class PackagesTable extends mixin(StoreMixin) {
           columns={this.getColumns()}
           colGroup={this.getColGroup()}
           data={this.props.packages.getItems().slice()}
-          sortBy={{prop: 'name', order: 'desc'}} />
+          sortBy={{prop: 'appId', order: 'desc'}} />
```
mlunoe: Sorting was not working before

src/js/components/PackagesTable.js
```diff
@@ -175,30 +129,6 @@ class PackagesTable extends mixin(StoreMixin) {
     );
   }
 
-  getUpgradeCell(prop, cosmosPackage) {
```
mlunoe: Moved down under `getUninstallButton`

src/js/components/modals/UninstallPackageModal.js
```diff
@@ -0,0 +1,197 @@
...
+
+    if (!cosmosPackage) {
+      return this.getEmptyNode();
+    }
+
+    return (
+      <div className="container-pod container-pod-short text-align-center">
+        <h3 className="flush-top">Are you sure?</h3>
+        <p>
+          {`${cosmosPackage.getAppIdName()} will be uninstalled from ${Config.productName}. All tasks belonging to this package will be killed.`}
```
aldipower: This is over 80 chars.

src/js/structs/UniversePackage.js
```diff
@@ -76,6 +90,20 @@ class UniversePackage extends Item {
     );
   }
 
+  getPostInstallNotes() {
+    return Util.findNestedPropertyInObject(
+      this.get('package'),
+      'postInstallNotes'
+    )
```
aldipower: Semicolon missing.

aldipower:
Besides very little minor stylistics this looks good from my side.
Great first take.

src/js/components/modals/UninstallPackageModal.js
mlunoe: This is a string and not code

aldipower: To be exact, it's a string definition in a portion of markup, which is embedded in code.
Why do you make exceptions to strings?

mlunoe: We don't use the 80 character rule for strings and regex's, since it quickly becomes hard to read and verify if everything is correct (where it does not when it is one string). The situation is opposite for code; the short lines makes it easier to understand what is going on. That is why the 80 character rule is not a hard rule.

We've ended up in a lot of bad situations with class names for example.

aldipower: OK, fair enough. Just find it not consequent.
E.g. my editors window width is set up to the 80 chars and to be able to read your string I have to scroll to the right.
I cannot rely on the 80 chars in dcos ui. I don't like that.
But as I said, there are other things in the world that are much worse, so I am fine with it, if you don't want to change this.

src/js/components/icons/IconCircleCheckmark.js
```diff
@@ -0,0 +1,28 @@
...
+
+class IconCircleCheckmark extends React.Component {
+  render() {
+    return (
+      <svg
+        height="64px"
+        viewBox="0 0 64 64"
+        width="64px"
+        {...this.props}>
+        <g>
```
jfurrow: We don't need this group anymore.

mlunoe: Good eyes! Thanks! It's removed!

src/js/components/modals/UninstallPackageModal.js
rcorral: I'm with @mlunoe at least it's what we've done in the past. Maybe we can change in the future.

rcorral:
Works great!

PR #89, opened 2016-04-30
Open service ui with modifier + mouse click

mlunoe:


mesosphere-web:
Reviewer: @jfurrow

MatApple:
Looks good to me.

src/js/components/ServiceList.js
```diff
@@ -35,8 +36,19 @@ let ServiceList = React.createClass({
     return !_.isEqual(this.props, nextProps) || changedState;
   },
 
-  handleServiceClick: function (serviceName) {
-    this.context.router.transitionTo('dashboard-panel', {serviceName});
+  handleServiceClick: function (service, event) {
+    // Open service in new window/tab if service has a web URL
+    if (service.webui_url &&
+      (event.ctrlKey || event.shiftKey || event.metaKey)) {
```
rcorral: Modifier keys are cmd for mac and ctrl for windows/linux

src/js/components/ServiceList.js
```diff
@@ -35,8 +36,19 @@ let ServiceList = React.createClass({
     return !_.isEqual(this.props, nextProps) || changedState;
   },
 
-  handleServiceClick: function (serviceName) {
-    this.context.router.transitionTo('dashboard-panel', {serviceName});
+  handleServiceClick: function (service, event) {
+    // Open service in new window/tab if service has a web URL
+    if (service.webui_url &&
```
jfurrow: The `Service` struct has a `#getWebURL` method.

kennyt: unfortunately, this might not be using the service struct.

but i wonder if it should be

mlunoe: It doesn't, and yes maybe it should. I didn't see it as part of this PR, though

src/js/components/ServiceList.js
mlunoe: This handles if you open a new tab or a new window with a modifier key, i.e. shift, meta (command on mac) or ctrl (for Windows, will still be work as secondary click for Mac). Here is a reference: http://stackoverflow.com/a/20087506/1008519

src/js/components/ServiceList.js
jfurrow: That makes sense!

PR #90, opened 2016-05-02
Remove Underscore [2 of 2]

daiyiwhale:
As much as this will be terrible to review, I assure you it was even more terrible to put together. :'D


mesosphere-web:
Reviewer: @kennyt

src/js/components/NodesGridDials.js
```diff
@@ -82,12 +82,12 @@ var NodesGridDials = React.createClass({
     var props = this.props;
 
     if (props.showServices) {
-      config = this.getServiceSlicesConfig(node);
+      config = this.getServiceSlicesConfig(node) || [];
```
kennyt: looking at `getServicesSlicesConfig`, i think it will always return at least an empty array, so this is unecessary

src/js/components/NodesGridDials.js
```diff
@@ -82,12 +82,12 @@ var NodesGridDials = React.createClass({
...
     } else {
-      config = this.getUsedSliceConfig(node);
+      config = this.getUsedSliceConfig(node) || [];
```
kennyt: looking at `getUsedSliceConfig`, i think it will always return an array

src/js/components/NodesGridDials.js
```diff
@@ -144,7 +144,9 @@ var NodesGridDials = React.createClass({
   },
 
   getDials: function () {
-    return _.map(this.props.hosts, function (node) {
+    let hosts = this.props.hosts || [];
```
kennyt: `hosts` will always be an array because of the prop definition

src/js/components/NodesGridDials.js
```diff
@@ -144,7 +144,9 @@ var NodesGridDials = React.createClass({
...
+
+    return hosts.map(function (node) {
```
kennyt: `map` doesn't take a 2nd argument. this whole block of code should not be working right now (i haven't tested)

src/js/components/ServiceList.js
```diff
@@ -40,7 +40,7 @@ let ServiceList = React.createClass({
   },
 
   getServices: function (services, healthProcessed) {
-    return _.map(services, function (service) {
+    return services.map(function (service) {
```
kennyt: `map` doesn't take a 2nd argument. this whole block of code should not be working right now (i haven't tested)

use `=>`

src/js/components/Sidebar.js
```diff
@@ -71,7 +70,7 @@ var Sidebar = React.createClass({
       defaultMenuItems
     );
 
-    return _.map(menuItems, function (routeKey) {
+    return menuItems.map(function (routeKey) {
```
kennyt: `map` doesn't take a 2nd argument. this whole block of code should not be working right now (i haven't tested)

use `=>`

src/js/components/charts/HostTimeSeriesChart.js
```diff
@@ -27,10 +26,10 @@ var HostTimeSeriesChart = React.createClass({
...
     var props = this.props;
     var roundUpValue = props.roundUpValue;
-
-    var maxSlavesCount = _.max(props.data, function (slave) {
-      return slave.slavesCount;
-    }).slavesCount;
+    let slavesCounts = props.data.map(function (agent) {
+      return agent.slavesCount;
+    });
+    var maxSlavesCount = Math.max(...slavesCounts);
```
kennyt: `let`

src/js/components/charts/ResourceBarChart.js
```diff
@@ -64,10 +63,11 @@ let ResourceBarChart = React.createClass({
   getModeButtons: function () {
     let selectedResource = this.props.selectedResource;
 
-    return _.map(ResourceTypes, function (info, key) {
+    return Object.keys(ResourceTypes).map(function (key) {
```
kennyt: map doesn't take context as arg. let's use `=>`

src/js/components/charts/TasksChart.js
```diff
@@ -39,6 +39,7 @@ var TasksChart = React.createClass({
       tasks = getEmptyTaskData();
     }
 
+    // ?????? wtf
```
kennyt: ?

src/js/components/charts/TimeSeriesChart.js
```diff
@@ -79,7 +80,7 @@ var TimeSeriesChart = React.createClass({
     // happens after mount and ends up keeping the axis code outside of react
     // unfortunately.
     // If non `data` props change then we need to update the whole graph
-    if (!_.isEqual(_.omit(props, 'data'), _.omit(nextProps, 'data'))) {
+    if (!_.isEqual(Util.omit(props, 'data'), Util.omit(nextProps, 'data'))) {
```
kennyt: i see a `_.isEqual` here

src/js/components/charts/TimeSeriesChart.js
```diff
@@ -346,7 +347,7 @@ var TimeSeriesChart = React.createClass({
     var area = this.getArea(props.y, xTimeScale, yScale, firstSuccess);
     var valueLine = this.getValueLine(xTimeScale, yScale, firstSuccess);
 
-    return _.map(props.data, function (stateResource, i) {
+    return props.data.map(function (stateResource, i) {
```
kennyt: use `=>` and don't pass context

src/js/components/charts/TimeSeriesChart.js
```diff
@@ -367,9 +368,9 @@ var TimeSeriesChart = React.createClass({
   },
 
   getCircleList: function (props, yScale, width, height) {
-    return _.map(props.data, function (obj, i) {
+    return props.data.map(function (obj, i) {
```
kennyt: use `=>` and don't pass context

src/js/events/TaskDirectoryActions.js
```diff
@@ -8,13 +6,21 @@ import MesosStateStore from '../stores/MesosStateStore';
...
 function findWithID(stateObject, listProps, id) {
-  let idCondition = {id};
-  let searchItem;
+  let stateObjectFound;
   let length = listProps.length;
   for (let i = 0; i < length; i++) {
-    searchItem = _.findWhere(stateObject[listProps[i]], idCondition);
-    if (searchItem) {
-      return searchItem;
+    let tasks = stateObject[listProps[i]] || [];
```
kennyt: can't we use `.find` for this?

src/js/components/charts/TasksChart.js
daiyiwhale: ooops :'D

src/js/components/charts/TasksChart.js
```diff
@@ -70,14 +71,14 @@ var TasksChart = React.createClass({
...
-    return _.reduce(tasks, function (acc, task) {
+    return tasks.reduce(function (acc, task) {
       return acc + task.value;
     }, 0);
   },
 
   getTasks: function (_tasks) {
     var validTasks = Object.keys(taskInfo);
-    var tasks = _.pick(_tasks, validTasks);
+    var tasks = _.pick(_tasks, validTasks) || [];
```
mlunoe: `_.pick` is still here here

src/js/components/charts/TimeSeriesChart.js
daiyiwhale: _.isEqual is the most horrifying thing ever, I almost feel like we should keep _ in the repo just for _.isEqual

src/js/components/charts/TasksChart.js
```diff
@@ -39,6 +39,7 @@ var TasksChart = React.createClass({
       tasks = getEmptyTaskData();
     }
 
+    // ?????? wtf
     var numberOfTasks = _.size(taskInfo);
 
     return _.map(taskInfo, function (info, key) {
```
mlunoe: `_.map`

src/js/components/AnimatedLogo.js
```diff
@@ -41,6 +40,10 @@ var AnimatedLogo = React.createClass({
     this.stopAnimationLoop();
   },
 
+  randomInt: function (min, max) {
```
rcorral: Tests

src/js/mixins/GetSetMixin.js
```diff
@@ -24,7 +24,7 @@ var GetSetMixin = {
   },
 
   set: function (data) {
-    if (!_.isObject(data) || _.isArray(data)) {
+    if (!_.isObject(data) || Array.isArray(data)) {
```
mlunoe: `_.isObject`

src/js/components/AnimatedLogo.js
```diff
@@ -24,7 +23,7 @@ var AnimatedLogo = React.createClass({
 
   componentWillMount: function () {
     this.internalStorage_set({
-      logoGradientID: _.uniqueId('logoGradient'),
+      logoGradientID: `logoGradient-${Math.random()}`,
```
rcorral: This doesn't quite guarantee a unique ID. It would be nice if we created a util for this and not have it using floats.

daiyiwhale: True, but we aren't even using this file anymore!

src/js/components/AnimatedLogo.js
```diff
@@ -162,7 +165,7 @@ var AnimatedLogo = React.createClass({
   },
 
   buildTriangles: function (strokeID) {
-    return _.map(['center', 'right', 'left'], function (position) {
+    return ['center', 'right', 'left'].map(function (position) {
```
kennyt: use `=>` and remove the 2nd argument here for map

src/js/components/FilterByService.js
```diff
@@ -52,7 +51,7 @@ var FilterByService = React.createClass({
     });
     let items = [defaultItem].concat(this.props.services);
 
-    return _.map(items, function (service) {
+    return items.map(function (service) {
```
kennyt: use `=>` and remove the 2nd argument here for map

src/js/components/FilterHealth.js
```diff
@@ -48,7 +53,7 @@ var FilterHealth = React.createClass({
   getFilterButtons: function () {
     var mode = this.props.healthFilter;
 
-    return _.map(buttonMap, function (value, key) {
+    return Object.keys(buttonMap).map(function (key) {
```
kennyt: use `=>` and remove the 2nd argument here for map

src/js/components/AnimatedLogo.js
mlunoe: Nuke it! 💥

kennyt:
doing a search in the project for `_.` i see there are still a lot of underscores left

mlunoe:
Same for ` _ ` and `underscore` to catch `import`s and `requires`

src/js/components/charts/TasksChart.js
aldipower: lol

kennyt:
tests failing

kennyt:
conflicts

src/js/components/NodesGridView.js
```diff
@@ -65,7 +65,11 @@ var NodesGridView = React.createClass({
    * @param  {Object} props
    */
   componentWillReceiveProps: function (props) {
-    let ids = _.pluck(props.services, 'id');
+    let ids = props.services.reduce(function (acc, value) {
```
kennyt: why don't we just use a `map` here?

src/js/components/NodesGridView.js
```diff
@@ -65,7 +65,11 @@ var NodesGridView = React.createClass({
...
+      acc.push(value['id']);
```
kennyt: we can do `value.id`

src/js/components/NodesGridView.js
daiyiwhale: good idea, also I can probably use map instead of pluck for a number of times which I will investigate right now

src/js/components/charts/HostTimeSeriesChart.js
```diff
@@ -27,10 +26,10 @@ var HostTimeSeriesChart = React.createClass({
   getMaxY: function () {
     var props = this.props;
     var roundUpValue = props.roundUpValue;
-
-    var maxSlavesCount = _.max(props.data, function (slave) {
-      return slave.slavesCount;
-    }).slavesCount;
+    let slavesCounts = props.data.map(function (agent) {
```
mlunoe: This could be written as a reduce:
```js
let maxSlavesCount = props.data.reduce(function (max, {slavesCount}) {
  if (max < slavesCount) {
    return slavesCount;
  }

  return max;
}, 0);
```

But not sure which one is prettier `¯\_(ツ)_/¯`

src/js/components/charts/TasksChart.js
```diff
@@ -4,15 +4,16 @@ var React = require('react');
...
 var tasksPerRow = 3;
-var taskInfo = {
+const TASK_INFO = {
   'TASK_RUNNING': {label: 'Tasks running', colorIndex: 4},
   'TASK_STAGING': {label: 'Tasks staging', colorIndex: 1}
 };
 
 function getEmptyTaskData() {
-  return _.map(taskInfo, function (val, key) {
+  return _.map(TASK_INFO, function (val, key) {
```
mlunoe: `_.map`

src/js/components/charts/TasksChart.js
```diff
@@ -34,15 +35,15 @@ var TasksChart = React.createClass({
...
   },
 
-  getTaskInfo: function (tasks) {
+  getTASK_INFO: function (tasks) {
     if (tasks.length === 0) {
       tasks = getEmptyTaskData();
     }
 
-    var numberOfTasks = _.size(taskInfo);
+    var numberOfTasks = _.size(TASK_INFO);
```
mlunoe: `_.size`

src/js/components/charts/TasksChart.js
```diff
@@ -34,15 +35,15 @@ var TasksChart = React.createClass({
...
 
-    return _.map(taskInfo, function (info, key) {
-      var task = _.findWhere(tasks, {name: key});
+    return _.map(TASK_INFO, function (info, key) {
```
mlunoe: `_.map`

src/js/stores/MarathonStore.js
```diff
@@ -74,7 +74,7 @@ var MarathonStore = Store.createStore({
   },
 
   shouldPoll: function () {
-    return !_.isEmpty(this.listeners(MARATHON_APPS_CHANGE));
+    return !(this.listeners(MARATHON_APPS_CHANGE).length === 0);
```
mlunoe: This could just be `return !!this.listeners(MARATHON_APPS_CHANGE).length`

src/js/structs/LogBuffer.js
```diff
@@ -24,10 +22,7 @@ class LogBuffer extends List {
   constructor(options = {}) {
     super(...arguments);
 
-    this.configuration = _.defaults(
-      _.pick(options, ...Object.keys(DEFAULT_OPTIONS)),
-      DEFAULT_OPTIONS
-    );
+    this.configuration = Object.assign({}, DEFAULT_OPTIONS, options);
```
mlunoe: This is not doing exactly the same thing. The `pick` is limiting the the applied properties from `options`

src/js/utils/DOMUtils.js
```diff
@@ -28,21 +29,19 @@ var DOMUtils = {
 
     var computeInnerBound = function (acc, key) {
       var val = parseInt(compstyle[key], 10);
-      if (_.isNaN(val)) {
+      if (isNaN(val)) {
```
mlunoe: `_.isNaN` is not the same as `isNaN`, read here: http://underscorejs.org/#isNaN.

We need to make sure this is covering the cases that we want it to cover

daiyiwhale: oh man I just found out about this: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isNaN

src/js/components/charts/HostTimeSeriesChart.js
daiyiwhale: They're both lovely :D

src/js/structs/LogBuffer.js
daiyiwhale: You're right, I've updated the refactor \o/

src/js/components/NodesGridDials.js
```diff
@@ -175,14 +173,14 @@ var NodesGridDials = React.createClass({
...
+    });
   },
 
   // Zero-height spacer items force dial charts in the last line of the flex layout
   // not to spread themselves across the line.
   getSpacers: function () {
-    return _.times(30, function (n) {
-      return <div className="nodes-grid-dials-spacer" key={n}></div>;
+    return Array(30).map(function (v, index) {
+      <div className="nodes-grid-dials-spacer" key={index}></div>
```
kennyt: you probably need to return this `div` :)

src/js/utils/Util.js
```diff
@@ -1,5 +1,26 @@
 const Util = {
 
+  findWhere: function (findInObject, attributes) {
+    let keys = Object.keys(attributes);
+    let length = keys.length;
+
+    return Array.find.call(null, findInObject, function (object) {
```
kennyt: why don't we just do `findInObject.find(`?

can we call `findInObject` something else because it's an array not an object?

src/js/utils/MesosStateUtil.js
```diff
@@ -8,25 +8,25 @@ const MesosStateUtil = {
...
         }
 
         var frameworkKey = task.framework_id;
-        if (_.contains(filter, framework.id)) {
+        if (filter.includes(framework.id)) {
           frameworkKey = 'other';
         }
 
-        var resources = _.pick(task.resources, 'cpus', 'disk', 'mem');
+        let resources = task.resources;
```
kennyt: why don't we pick out the resources anymore?

src/js/components/charts/BarChart.js
```diff
@@ -71,7 +72,7 @@ var BarChart = React.createClass({
       stack: this.getStack(),
       xScale,
       yScale,
-      clipPathID: _.uniqueId('clip')
+      clipPathID: `clip-${Math.random().toString().slice(2)}`
```
kennyt: can we create a util for this? all it would have to do is start with an id that is 1 and then everytime it gets called, the just increments it so it will always be unique. rather than doing this

src/js/components/charts/DialChart.js
```diff
@@ -18,7 +17,6 @@ var DialChart = React.createClass({
   mixins: [InternalStorageMixin],
 
   propTypes: {
-    // [{colorIndex: 0, name: 'Some Name', value: 4}]
```
kennyt: this comment might've been there to show what the data looks like right?

src/js/components/charts/DialChart.js
```diff
@@ -80,14 +78,20 @@ var DialChart = React.createClass({
...
+    var namedSlices = slices.reduce(function (indexByName, slice) {
+      indexByName[slice.name] = slice;
+      return indexByName;
+    }, {});
+    var namedData = data.reduce(function (indexByName, datum) {
+      indexByName[datum.name] = datum;
+      return indexByName;
+    }, {});
+    var normalizedNamedData = Object.assign({}, namedSlices, namedData);
+    return Object.values(normalizedNamedData);
```
kennyt: can we have newlines before each of these returns

src/js/components/charts/DialChart.js
daiyiwhale: that's true, it'll stick it back

src/js/components/charts/DialChart.js
daiyiwhale: I find this way more confusing to read 

this isn't much better :'D 

I will concede to this though:

src/js/utils/MesosStateUtil.js
daiyiwhale: couple lines down, now using `RESOURCE_KEYS` instead of `_.each(resources`

src/js/utils/Util.js
daiyiwhale: findwhere is for objects, find is for arrays

daiyiwhale: ohhh I see what you're saying

src/js/components/charts/DialChart.js
mlunoe: Please add newlines before `return` statements. It is only if `return` is the only line in that block we don't add a newline before.

src/js/utils/DOMUtils.js
mlunoe: It doesn't seem like there is good support for it and Babel doesn't seem to have a polyfill for it: https://babeljs.io/repl/#?evaluate=true&lineWrap=false&presets=es2015%2Creact%2Cstage-2&experimental=false&loose=false&spec=true&playground=false&code=Number.isNaN('ajs')

src/js/utils/DOMUtils.js
```diff
@@ -28,21 +29,19 @@ var DOMUtils = {
 
     var computeInnerBound = function (acc, key) {
       var val = parseInt(compstyle[key], 10);
-      if (_.isNaN(val)) {
+      if (Number.isNaN(val)) {
```
mlunoe: It doesn't seem like there is good support for it and Babel doesn't seem to have a polyfill for it: https://babeljs.io/repl/#?evaluate=true&lineWrap=false&presets=es2015%2Creact%2Cstage-2&experimental=false&loose=false&spec=true&playground=false&code=Number.isNaN('ajs')

kennyt: although `Number#isNaN` is not heavily supported, `isNaN` by itself apparently is

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isNaN

src/js/structs/LogBuffer.js
```diff
@@ -24,10 +22,14 @@ class LogBuffer extends List {
...
 
-    this.configuration = _.defaults(
-      _.pick(options, ...Object.keys(DEFAULT_OPTIONS)),
-      DEFAULT_OPTIONS
-    );
+    this.configuration = Object.assign({}, DEFAULT_OPTIONS);
+
+    let optionsKeys = Object.keys(options);
+    Object.keys(DEFAULT_OPTIONS).forEach(function (key) {
+      if (optionsKeys.includes(key)) {
```
mlunoe: Let's use `options.hasOwnProperty(key)` instead here

src/js/utils/DOMUtils.js
mlunoe: Right, but the discussion started with `isNaN` and `_.isNaN` not being the same thing. So I wanted to make sure if we use `NaN` that it is covering the cases that we need.

kennyt: ah i see. i didn't see the previous discussion

mlunoe: It seems that the `parseInt(value, 10);` has got us covered:

kennyt: that's different from the `_.isNaN` though, isn't it?

kennyt: nvm, there's a `parseInt` before this line. i see what you're saying

kennyt:
tests failing

src/js/components/charts/DialChart.js
daiyiwhale: How's the readability of the second one? I think I'll go with that.

src/js/utils/DOMUtils.js
daiyiwhale: fear not all our fears can be laid to rest: 


(reference: https://kangax.github.io/compat-table/es6/#test-Number_properties)

src/js/components/__tests__/NodesGridDials-test.js
```diff
@@ -54,46 +55,52 @@ describe('NodesGridDials', function () {
...
 
     it('contains a usedPercentage property which is a number', function () {
       expect(typeof this.activeSlices.usedPercentage).toEqual('number');
     });
 
     it('contains a slice for the used resource', function () {
-      var slice = _.findWhere(this.activeSlices.data, {
-        name: this.resourceType.label
+      var slice = this.activeSlices.data.find(function (datum) {
+        return datum.name === this.resourceType.label
```
kennyt: this will need `=>` due to using `this` inside of it

src/js/components/charts/ChartStripes.js
```diff
@@ -15,7 +14,7 @@ var ChartStripes = React.createClass({
     var count = props.count;
     var width = props.width / (2 * count);
 
-    return _.map(_.range(0, count), function (i) {
+    return Object.keys(Array(count).fill()).map(function (i) {
```
kennyt: won't `i` be `undefined` every single time at the moment?

src/js/components/charts/DialChart.js
```diff
@@ -80,14 +79,23 @@ var DialChart = React.createClass({
...
-    var normalizedNamedData = _.extend(namedSlices, namedData);
-    return _.values(normalizedNamedData);
+    var namedSlices = slices.reduce(function (indexByName, slice) {
+      indexByName[slice.name] = slice;
+      return indexByName;
+    }, {});
+
+    var namedData = data.reduce(function (indexByName, datum) {
+      indexByName[datum.name] = datum;
+      return indexByName;
```
kennyt: don't forget to put a newline before the returns!

src/js/components/charts/TasksChart.js
```diff
@@ -31,26 +35,31 @@ var TasksChart = React.createClass({
     var newTasks = this.getTasks(nextProps.tasks);
 
     // If equal, do not update
-    return !_.isEqual(previousTasks, newTasks);
+    return !Object.keys(previousTasks).reduce(function (isEqual, key) {
```
kennyt: do we even want to do this? we're replacing this anyways right? we can leave it here since it's more work to remove it

src/js/components/charts/ChartStripes.js
daiyiwhale: We're mapping over the keys so I would be the indexes (0, 1, ...)

kennyt: oh i see, you're called `Object.keys` on an array. i think that's really odd. plus, the keys would be strings so it's not really what we want.

why not just do `Array(count).map` and then use the index? (2nd argument)

daiyiwhale: We'll need `Array(count).fill().map()` because without `.fill()` it would be `Array [ <3 empty slots> ]` but yes that would be better, thanks! :D

src/js/components/__tests__/NodesGridDials-test.js
daiyiwhale: oops I'd made those changes already but forgot to push them up :'D

src/js/components/charts/TimeSeriesChart.js
daiyiwhale: We are replacing `_.isEqual` with https://www.npmjs.com/package/deep-equal

src/js/components/AnimatedLogo.js
```diff
@@ -1,209 +0,0 @@
-var _ = require('underscore');
```
rcorral: Sooo sad to see this goo :(

src/js/utils/Util.js
```diff
@@ -1,6 +1,17 @@
+let id = 0;
+
 const Util = {
 
   /**
+   * Give a unique ID for this session.
+   * @return {Integer}               A unique id.
+   */
+  uniqueID: function () {
```
rcorral: This needs a test or two.

src/js/components/charts/TasksChart.js
```diff
@@ -1,19 +1,24 @@
...
 var DialChart = require('./DialChart');
 
-var tasksPerRow = 3;
-var taskInfo = {
+const TASKS_PER_ROW = 3;
+const TASK_INFO = {
   'TASK_RUNNING': {label: 'Tasks running', colorIndex: 4},
   'TASK_STAGING': {label: 'Tasks staging', colorIndex: 1}
 };
+const DISPLAYED_TASK_KEYS = ['TASK_RUNNING', 'TASK_STAGING'];
```
rcorral: `const DISPLAYED_TASK_KEYS = Object.keys(TASK_INFO);`

src/js/mixins/GetSetMixin.js
```diff
@@ -24,7 +22,7 @@ var GetSetMixin = {
   },
 
   set: function (data) {
-    if (!_.isObject(data) || _.isArray(data)) {
+    if (!(typeof data === 'object' && !Array.isArray(data))) {
```
rcorral: This is a bit confusing to read now, but it works :p

daiyiwhale: Haha I'll leave a comment!

src/js/utils/DOMUtils.js
```diff
@@ -1,4 +1,5 @@
-var _ = require('underscore');
+const WIDTH_ATTRIBUTES = ['paddingLeft', 'paddingRight', 'borderLeftWidth', 'borderRightWidth'];
+const HEIGHT_ATTRIBUTES = ['paddingTop', 'paddingBottom', 'borderTopWidth', 'borderBottomWidth'];
```
rcorral: 🔤

src/js/components/AnimatedLogo.js
daiyiwhale: (':

src/js/utils/Util.js
daiyiwhale: done! https://github.com/dcos/dcos-ui/pull/151

daiyiwhale:
Remaining comments addressed here: https://github.com/dcos/dcos-ui/pull/152

PR #91, opened 2016-05-02
Introduce a recurrent task util

orlandohohmeier:
Add a util to schedule and control recurrent tasks, like polling an actions (e.g. `MesosStateActions .fetchState` or `fetchVersion`). The util will execute the scheduled task in the given interval, but will "pause" the execution of all tasks if the application tab/window is inactive.

This util will enable us to unify the various implementations of polling mechanisms, it also provides the foundations to precisely poll service specific endpoints (e.g. service detail configurations).


mesosphere-web:
Reviewer: @daiyiwhale

src/js/utils/RecurrentTaskUtil.js
```diff
@@ -0,0 +1,53 @@
...
+    if (time >= interval + lastExecution) {
+      tasks.set(id, {interval, lastExecution: time, task});
+      task();
+    }
+  });
+
+  // Using `requestAnimationFrame` to execute the tasks has the great advantage
+  // that it will automatically "pause" the execution if the application window
+  // is inactive.
+  global.requestAnimationFrame(executeTasks);
```
pierlo-upitup: 👍

src/js/utils/RecurrentTaskUtil.js
```diff
@@ -0,0 +1,53 @@
...
+
+const RecurrentTaskUtil = {
+  /**
+   * Schedule recurrent task
+   * @param {string} id
+   * @param {function} task
+   * @param {number} interval
+   */
+  scheduleTask(id, task, interval) {
+    tasks.set(id, {lastExecution: -interval, interval, task});
```
pierlo-upitup: Nice use-case for a `Map` indeed!

pierlo-upitup: I'm just puzzled about the `-interval`. Why can't that be just a `0`? Additionally, I'd probably throw in some sanitisation for the `interval` argument (and possibly make it  an optional arg with a default value too).

src/js/utils/RecurrentTaskUtil.js
```diff
@@ -0,0 +1,53 @@
...
+}
+
+function stopTaskExecution() {
+  executing = false;
+}
+
+const RecurrentTaskUtil = {
+  /**
+   * Schedule recurrent task
+   * @param {string} id
```
pierlo-upitup: I'd suggest to name this `taskId` throughout.

src/js/utils/RecurrentTaskUtil.js
```diff
@@ -0,0 +1,53 @@
+const tasks = new Map();
+let executing = false;
+
+function executeTasks(time) {
```
pierlo-upitup: how about `timestamp`?

aldipower:
Nice idea!

src/js/utils/RecurrentTaskUtil.js
```diff
@@ -0,0 +1,53 @@
...
+  tasks.forEach(function ({task, lastExecution, interval}, id) {
+    if (time >= interval + lastExecution) {
+      tasks.set(id, {interval, lastExecution: time, task});
+      task();
```
pierlo-upitup: Might be a good idea to validate this is a function before calling it?

src/js/utils/RecurrentTaskUtil.js
```diff
@@ -0,0 +1,53 @@
+const tasks = new Map();
+let executing = false;
+
+function executeTasks(time) {
+  tasks.forEach(function ({task, lastExecution, interval}, id) {
+    if (time >= interval + lastExecution) {
+      tasks.set(id, {interval, lastExecution: time, task});
```
pierlo-upitup: `tasks.set(id, {task, lastExecution: time, interval}`

src/js/utils/RecurrentTaskUtil.js
```diff
@@ -0,0 +1,53 @@
...
+  scheduleTask(id, task, interval) {
+    tasks.set(id, {lastExecution: -interval, interval, task});
+    startTaskExecution();
+  },
+
+  /**
+   * Stop task
+   * @param {function} id
+   */
+  stopTask(id) {
```
pierlo-upitup: For all intents and purposes this should be called `deleteTask`. The API doesn't offer a way to `stop`(and `resume` for the matter) a Task at this point.

src/js/utils/RecurrentTaskUtil.js
pierlo-upitup: Also, how about enforcing an `isFunction` check before proceeding? This would be even better than what I suggested at https://github.com/dcos/dcos-ui/pull/91/files#r61786512

kennyt:
was this util written in order to replace the `startPolling` in all of the stores?

kennyt:
just wondering, is this performance-friendly? every ~17ms we're iterating through the tasks and checking if it is time to do the action again.

philipnrmn:
Nice implementation!

➕ for @kennyt's concerns- I'd have thought a 500ms period was enough. Am I missing something? Will this be used for animation as well as polling?

pierlo-upitup:
@philipnrmn & @kennyt – yeah I agree we could probably throttle the check within `requestAnimationFrame`.

philipnrmn:
I was just going to suggest a much longer period in a `setInterval`.

pierlo-upitup:
@philipnrmn but then you'd lose the advantage of suspending the polling operations on window blur.

kennyt:
we do currently have someone in order to suspend polling on window blur. it stops polling if you're gone for more than 60 seconds

philipnrmn:
`dcos-ui` has been merged. Please re-open this PR against `master`.

PR #93, opened 2016-05-02
Allow the default values for AdvancedConfig to be changed

kennyt:


mesosphere-web:
Reviewer: @mlunoe

src/js/components/SchemaForm.js
```diff
@@ -118,9 +118,15 @@ class SchemaForm extends React.Component {
   }
 
   getNewDefinition() {
-    return SchemaUtil.schemaToMultipleDefinition(
+    let definition = SchemaUtil.schemaToMultipleDefinition(
       this.props.schema, this.getSubHeader, this.getLabel
     );
+
+    if (this.props.model) {
```
aldipower: 'let {props} = this;` it's used twice now.

src/js/components/SchemaForm.js
```diff
@@ -118,9 +118,15 @@ class SchemaForm extends React.Component {
...
+      SchemaFormUtil.mergeModelIntoDefinition(this.props.model, definition);
```
mlunoe: 1. This function doesn't exist yet?
2. And is it supposed to modify `definition`? Not sure that's the best path to go down. maybe just return a new `definition`?

kennyt: this function exists already. this function is a merge, similar to `Object.assign` or `extend`

jfurrow:
Looks good, let's 📦 when @aldipower's comment is addressed.

kennyt:
@jfurrow @aldipower resolved comment 👍

PR #94, opened 2016-05-03
Introduce Services Breadcrumb component

pierlo-upitup:
This PR introduces the `ServicesBreadcrumb` component for both the Services list and the Services details page.

It's a minimalistic port of the Marathon UI `BreadcrumbComponent`, in that it leaves out the `Volume` and `Task` partials out (both routes are not yet present in DC/OS UI) as well as discarding the collapsible approach completely for the moment. 

Additionally, instead of introducing new CSS styles, it adopts the already present `.breadcrumbs` behaviour as seen in the Task details sidebar (See `TaskDirectoryView.js`). This means that the result does not match 1:1 the design mockups, which need to be revised anyway (cc @leemunroe).

breads


mesosphere-web:
Reviewer: @rcorral

src/js/components/ServicesBreadcrumb.js
```diff
@@ -0,0 +1,113 @@
+import React from 'react';
+import {Link} from 'react-router';
+
+import ServicesPathUtil from '../utils/ServicesPathUtil';
+
+const breadcrumbIcon = (
+  <i className="icon icon-sprite icon-sprite-small icon-back forward" />
+);
+
+var ServicesBreadcrumb = React.createClass({
```
orlandohohmeier: Please use the ES6 class syntax

src/js/components/ServicesBreadcrumb.js
```diff
@@ -0,0 +1,113 @@
...
+  displayName: 'ServicesBreadcrumb',
+
+  propTypes: {
+    appId: React.PropTypes.string,
+    groupId: React.PropTypes.string,
+    taskId: React.PropTypes.string,
```
orlandohohmeier: Please remove `taskId` and `volumeId` as they're unused.

src/js/components/ServicesBreadcrumb.js
```diff
@@ -0,0 +1,113 @@
...
+  },
+
+  getGroupLinks: function () {
+    var {appId, groupId} = this.props;
+
+    if (groupId == null) {
+      return null;
+    }
+
+    const pathParts = groupId.split('/').slice(0, -1);
```
orlandohohmeier: Please consider using a combination of ServiceTree `filterItems` and `reduceItems` in conjunction with Service `getId` to create a list of links instead of this string fiddling.

src/js/utils/ServicesPathUtil.js
```diff
@@ -0,0 +1,20 @@
+const ServicesPathUtil = {
```
orlandohohmeier: Let's please refactor the Breadcrumb component to use the feature rich Service (ServiceTree) interface instead of introducing an util to extract ids/paths from strings.

src/js/pages/services/ServicesTab.js
```diff
@@ -108,6 +110,15 @@ var ServicesTab = React.createClass({
 
     // Find item in root tree and default to root tree if there is no match
     let item = DCOSStore.serviceTree.findItemById(id) || DCOSStore.serviceTree;
+    let groupId = ServicesPathUtils.getGroupFromAppId(item.id);
```
orlandohohmeier: Please use the correct getter `item.getId()` instead of accessing the property directly. The `item.id` property is only avialable due to a scruffy item struct implementation.

src/js/pages/services/ServicesTab.js
```diff
@@ -108,6 +110,15 @@ var ServicesTab = React.createClass({
...
+
+    if (item instanceof ServiceTree) {
+      groupId = item.id;
+
+      if (groupId != null && !groupId.endsWith('/')) {
```
orlandohohmeier: This shouldn't be necessary and seems very error prone. Please refactor the breadcrump component to simplify it's usage.

src/js/utils/ServicesPathUtil.js
pierlo-upitup: I don't see any getter that would help me achieve what I need to do here...?

src/js/pages/services/ServicesTab.js
pierlo-upitup: I wonder why we didn't introduce a `get id()` for the `Item` instead of `getId()`.

orlandohohmeier: We did (at least for the new structs), but had to revert that changes as it collides with the current `Item` struct and any refactoring of the item struct was out of scope. :(

src/js/utils/ServicesPathUtil.js
```diff
@@ -0,0 +1,20 @@
...
+  getRelativePath(id, currentGroup) {
+    if (!currentGroup.endsWith('/')) {
+      currentGroup += '/';
+    }
+    if (id.startsWith(currentGroup)) {
+      return id.substring(currentGroup.length);
+    }
+    return id;
+  },
+  getGroupFromAppId(id) {
```
rcorral: Could you add a line break between the function definition and the `},`

src/js/components/ServicesBreadcrumb.js
```diff
@@ -0,0 +1,113 @@
+import React from 'react';
+import {Link} from 'react-router';
+
+import ServicesPathUtil from '../utils/ServicesPathUtil';
+
+const breadcrumbIcon = (
+  <i className="icon icon-sprite icon-sprite-small icon-back forward" />
```
rcorral: We have a `/components/icons` folder, it would be nice to put this there. There may even already be one there.

src/js/components/ServicesBreadcrumb.js
```diff
@@ -0,0 +1,113 @@
...
+  getGroupLinks: function () {
+    var {appId, groupId} = this.props;
+
+    if (groupId == null) {
+      return null;
+    }
+
+    const pathParts = groupId.split('/').slice(0, -1);
+
+    return pathParts.map((name, i) => {
```
rcorral: Double arrow isn't needed here as we don't need the ctx inside.

src/js/components/ServicesBreadcrumb.js
```diff
@@ -0,0 +1,113 @@
...
+      var id = pathParts.slice(0, i + 1).join('/');
```
rcorral: `let`

src/js/components/ServicesBreadcrumb.js
```diff
@@ -0,0 +1,113 @@
...
+        group = (<span>{name}</span>);
+      }
+
+      return (
+        <span className="crumb" key={id}>
+          {breadcrumbIcon}
+          {group}
+        </span>
+      );
+    }).slice(1);
```
rcorral: It would be nice to either see this `slice` in a new line like this:
```
})
.slice(1);
```
or assign the result to `map` to a variable and then do the slice in a separate line `return myVar.slice(1);`
Otherwise I find it so easy to have that slice be overlooked.

src/js/components/ServicesBreadcrumb.js
```diff
@@ -0,0 +1,113 @@
...
+  },
+
+  getAppLink: function () {
+    var appId = this.props.appId;
+    var groupId = this.props.groupId;
```
rcorral: `let` please!

src/js/components/ServicesBreadcrumb.js
```diff
@@ -0,0 +1,113 @@
...
+          {this.getRootLink()}
+          {this.getGroupLinks()}
+          {this.getAppLink()}
+        </h3>
+      </div>
+    );
+  }
+});
+
+export default ServicesBreadcrumb;
```
rcorral: Please use `module.exports = ServicesBreadcrumb;`
The latest version babel doesn't always handle this correctly and we have to then use the `.default`

src/js/utils/ServicesPathUtil.js
orlandohohmeier: I don't see any need for this util if you refactor the breadcrumb component as suggested in https://github.com/dcos/dcos-ui/pull/94#discussion_r61862251

src/js/components/ServicesBreadcrumb.js
orlandohohmeier: Let me illustrate what I have in mind:

**Usage** 
```javascript
<ServicesBreadcrumb currentId={groupId} />
```

**Generate Links** (*pseudo code*)
```javascript
  let links = DCOSStore.serviceTree
      .filterItems(function filterItemsById(item) {
           return currentId.startsWith(item.getId())
      })
      .mapItems(function generateLink(item) {
            return (<Link param={id:item.getId()}/>)
       })
      .flattenItems();
```

pierlo-upitup: Is that really a good idea in terms of performance though? Iterating 3 times over the whole tree just to get one item's path components?

aldipower: I may hook in?

If performance is a concern then @pierlo-upitup approach is the faster one I guess, because there is no need to filter the tree.
Each filter operation has a cost of 3 function calls (filterItemsById, startsWith, getId) for each step in the iteration.
I mean it's not too bad, but the string fiddeling approach is much more straight forward in this case.

On the other hand I find @orlandohohmeier approach more descriptive.

But I tend to the performance approach here, that is descriptive too..

Just my 2¢ and only a humble opinion.

Don't let us waist time on discussing on elegance.
"Beauty is in the eye of the beholder."

orlandohohmeier: You could and probably should reduce this to one iteration by using `reduceItems` to create one flat, filtered list of links. The above was only meant to illustrate the idea.

kennyt:
i think you might've rebased with master accidentally

src/js/components/ServicesBreadcrumb.js
pierlo-upitup: Although those make sense as they are configurable SVGs... What I have here is a simple assignment to avoid repeating the markup a bunch of times.

pierlo-upitup:
Addressed all comments and assigned to @aldipower

pierlo-upitup:
Also please note: not shown in the gif above, the breadcrumb now gets replaced by the filter headline only when a filter is applied, as per the designs.

src/js/components/ServicesBreadcrumb.js
```diff
@@ -0,0 +1,77 @@
...
+    if (groupId === '/') {
+      breadcrumbNodes[0] = (
+        <span className="crumb" key="/">Services</span>
+      );
+    }
+
+    breadcrumbNodes = breadcrumbNodes.concat(
+      DCOSStore.serviceTree.reduceItems(function (memo, item) {
+        if (groupId.startsWith(item.getId())) {
+          const id = item.getId();
```
aldipower: Swap lines, so this is assign before the `if`, it's already used there.

src/js/components/ServicesBreadcrumb.js
```diff
@@ -0,0 +1,77 @@
...
+          const name = item.getName();
+
+          let linkOrText = (
```
aldipower: Maybe `breadcrumb` instead of `linkOrText?

src/js/pages/services/ServicesTab.js
```diff
@@ -118,18 +119,33 @@ var ServicesTab = React.createClass({
         name: state.searchString
       }).getItems();
 
+      let breadcrumbsOrFilterHeadline = (
```
aldipower: This is a quantum variable. I don't know the content until I measure. :)
I am fine with it for now.
But it has some downsides like rafactorability/atomicity and separation of concerns.
I would like to see this in two separate variables next time. :)

aldipower:
Thank you for doing this!
It's a performant and straight-forward approach!

:ship: then :airplane: 

It needs some integration tests and could be generalized for reusage in other pages later on.
I've created tech-debt issues for that, so we can merge this.

aldipower:
Works perfectly in FF. Mergeable.

Although it definitely needs another design pass later as you already mentioned.
breadcrumbs

pierlo-upitup:
Please hold merge – working on the design atm.

pierlo-upitup:
crumbs

@leemunroe I did a first pass to get the styles closer to the designs as shown in the gif – still not there 100% but we can review this later together.

aldipower:
Perfect! Thank you.
I can navigate now, finally, missed that feature. :))))

leemunroe:
@pierlo-upitup Nice! @philipnrmn and @aldipower gave me a demo earlier. Some styling things to consider but functionality wise works well.

Styling to look at:
* Reducing the size of the arrow (so there isn't a jump)
* Changing colors of the breadcrumb items (apart from the current item)
* Reducing size of the breadcrumbs on the detail view

mlunoe:
Could we roll this out to the rest of the application?
Here is a really small example of a breadcrumb using existing classes and icon: https://github.com/dcos/dcos-ui/blob/master/src/js/pages/nodes/NodeDetailPage.js#L77

pierlo-upitup:
@mlunoe we have a techdebt issue to track that effort https://mesosphere.atlassian.net/browse/MARATHON-1033

PR #92, opened 2016-05-02
Update query params to reflect user saved state

pierlo-upitup:
This PR ensures that the query params reflect the user filter selection automatically persisted by the `SaveStateMixin`. 

filters

(ignore the `Breadcrumb` text in the gif)

PR #99, opened 2016-05-04
Fix services tab query param persistence

philipnrmn:
This PR fixes the bug highlighted in this comment: https://github.com/dcos/dcos-ui/pull/92#discussion_r61952582

Previously, query parameters were not shown in the URL even though they persisted over navigation.

### DIFF
--- a/src/js/pages/__tests__/ServicesTab-test.js
+++ b/src/js/pages/__tests__/ServicesTab-test.js
@@ -1,6 +1,8 @@
 jest.dontMock('../services/ServicesTab');
 jest.dontMock('../../mixins/InternalStorageMixin');
 jest.dontMock('../../mixins/SaveStateMixin');
+jest.mock('../../stores/UserSettingsStore');
+jest.mock('../../mixins/QueryParamsMixin');
 
 /* eslint-disable no-unused-vars */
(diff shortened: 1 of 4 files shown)


mesosphere-web:
Reviewer: @rcorral

src/js/pages/__tests__/ServicesTab-test.js
```diff
@@ -32,6 +39,22 @@ describe('ServicesTab', function () {
...
+      this.container
+    );
+    expect(QueryParamsMixin.setQueryParam).toHaveBeenCalledWith(
+      'filterHealth', ['0']
+    );
+    expect(QueryParamsMixin.setQueryParam).toHaveBeenCalledWith(
+      'searchString', 'foo'
+    );
+  });
+
```
pierlo-upitup: Thanks for these!

philipnrmn:
I wasn't happy with having to override methods inline in the `require` block, so I've introduced some proper mocks for the mixins.

philipnrmn:
NB tests are actually passing - the fail is just a bad config in Velocity.

aldipower:
Thx! Does the trick!
The mocks are also a good improvement!

PR #108, opened 2016-05-05
Make npm honor environment variables

MatApple:
`npm config set` doesn't work in CI where we have protected user directories `/Users/[username]/..`.

To honor setting npm config via environment variables, we cannot have underscores in our keys because they get converted to hyphens and therefore do not override the npm config.


mesosphere-web:
Reviewer: @kennyt

PR #118, opened 2016-05-05
Feature - Add task detail page

aldipower:
page1
page2

* The route looks like this now:
`#/services/[serviceID]/task-detail/[taskID]/`

* I've put in an additional rudimentary 'Marathon Task Configuration' that shows up below the Mesos task information. Extending this Marathon information is a huge task, that should be done outside of this PR. Also I think it needs a refinement together with @leemunroe at some point, how to merge Mesos and Marathon task information.

* I've adjusted the ServiceBreadcrumbs to be able to display and route the task id. But the ServiceBreadcrumbs are not generic right now, so I disabled them in the 'Node' task view, which is actually the same. @pierlo-upitup 

* The status (Running, Finished, etc) in the PageHeader isn't shown up emphazied, even though it is set as emphasize. This is a Canvas UI definition, that I am not meant to influence. If Canvas UI changes the emphazise setting, this status line will follow automatically, because the 'emphasize' is set.

* The code container in the logs view needs some adjustment, so the "Go to bottom" button isn't shown up accidentally. But this is a Reactjs-components fix. This PR is not responsible for it.

Without the help of @kennyt and @mlunoe I would have been screwed up.


mesosphere-web:
Reviewer: @MatApple

src/js/components/TaskDetail.js
```diff
@@ -287,37 +353,66 @@ class TaskSidePanelContents extends SidePanelContents {
...
-        </div>
+        {this.getBasicInfo(task, node)}
         {this.tabs_getTabView()}
       </div>
     );
   }
 }
 
-module.exports = TaskSidePanelContents;
+module.exports = TaskDetail;
```
pierlo-upitup: I don't see the `propTypes` – please add them!

pierlo-upitup:
Awesome work!!! Still reviewing but this is really great progress.

I noticed two minor UX issues 👓 :

1) By default the Details tab is selected (which is fine) however the tab is the 2nd in order. We should probably rearrange them.

defaulttab

2) The log viewer shows the "Go To Bottom" button even when there is no need for it (the log itself is shorter than the textarea). The button should not be shown unless we are scrolling away from the bottom. If this is a known bug in DC/OS and it wasn't introduced by this PR then I'm fine with solving it in a separate one.

gotobottom

src/js/components/TaskDetail.js
```diff
@@ -1,21 +1,26 @@
...
 import TaskStates from '../constants/TaskStates';
-import TaskUtil from '../utils/TaskUtil';
 import Units from '../utils/Units';
 
+import InternalStorageMixin from '../mixins/InternalStorageMixin';
+import TabsMixin from '../mixins/TabsMixin';
+
 const TABS = {
   files: 'Files',
   details: 'Details',
```
pierlo-upitup: Lets move `details` so it's the first one, as we now assume that to be the case. Then we don't need to explicitly assing the string value to the `this.state.currentTab` at line `37`.

src/js/components/TaskDetail.js
```diff
@@ -27,14 +32,14 @@ const METHODS_TO_BIND = [
...
-class TaskSidePanelContents extends SidePanelContents {
+class TaskDetail extends mixin(InternalStorageMixin, TabsMixin, StoreMixin) {
   constructor() {
     super(...arguments);
 
     this.tabs_tabs = Object.assign({}, TABS);
 
     this.state = {
-      currentTab: Object.keys(this.tabs_tabs).shift(),
+      currentTab: 'details',
```
pierlo-upitup: This should to be the first item in the `this.tabs_tabs` array so we don't rely on manual configuration. See comment https://github.com/dcos/dcos-ui/pull/118/files#r62645384

src/js/components/TaskDetail.js
```diff
@@ -151,56 +151,119 @@ class TaskSidePanelContents extends SidePanelContents {
...
-  getBasicInfo(task, node) {
-    // Hide when no task or when we are viewing debug tab
-    if (task == null || this.state.currentTab === 'debug') {
-      return null;
-    }
+  getBasicInfo(task) {
+    let taskIcon = (
+      <div
+        className="icon icon-large icon-image-container icon-app-container">
+        <img src={task.getImages()['icon-large']} />
```
pierlo-upitup: I'm not convinced that the `Task` struct should be responsible for holding an array of images. Is it the case that Structs generally are supposed to hold view-data? I personally would favour referencing the `ServiceImages` object from the component itself.

src/js/components/TaskDetail.js
```diff
@@ -151,56 +151,119 @@ class TaskSidePanelContents extends SidePanelContents {
...
+        (task.ipAddresses == null || task.ipAddresses.length === 0)) {
+      return (<dd>None</dd>);
+    }
+
+    if (service != null &&
+        service.ipAddress != null &&
+        service.ipAddress.discovery != null &&
+        service.ipAddress.discovery.ports != null &&
+        task.ipAddresses != null &&
+        task.ipAddresses.length > 0) {
```
pierlo-upitup: If I got that right these would need a `-2` indentation ?

src/js/components/TaskDetail.js
```diff
@@ -151,56 +151,119 @@ class TaskSidePanelContents extends SidePanelContents {
...
+
+      if (endpoints.length) {
+        return endpoints.map(function (endpoint) {
+          return (
+            <a href={`//${endpoint}`} target="_blank">{endpoint}</a>
+          );
+        });
+      }
+
+      return 'n/a';
```
pierlo-upitup: Should we be consistent and return this within a `dd` as we do at line 190?

src/js/components/TaskDetail.js
```diff
@@ -151,56 +151,119 @@ class TaskSidePanelContents extends SidePanelContents {
...
+    }
+
+    let headerValueMapping = {
+      Host: task.host,
+      Ports: task.ports,
+      Endpoints: this.getTaskEndpoints(task),
+      Status: this.getTaskStatus(task),
+      'Staged at': task.stagedAt,
+      'Started at': task.startedAt,
+      Version: task.version
```
pierlo-upitup: I'd prefer to have all keys here be within single quotes for consistency.

src/js/components/TaskDetail.js
```diff
@@ -151,56 +151,119 @@ class TaskSidePanelContents extends SidePanelContents {
...
       return null;
     }
 
     let node = MesosStateStore.getNodeFromID(task.slave_id);
+
+    if (node == null) {
+      return (
+        <p>Cannot find task information.</p>
+      );
+    }
```
pierlo-upitup: Why do we give up entirely in this case? The only info we're missing is the `node.hostname` and `node.id`.

src/js/components/TaskDetail.js
```diff
@@ -151,56 +151,119 @@ class TaskSidePanelContents extends SidePanelContents {
...
+    return task.ports.map(function (port) {
+      let endpoint = `${task.host}:${port}`;
+      return (
+        <a href={`//${endpoint}`} target="_blank">{endpoint}</a>
+      );
+    });
+  }
+
+  getMarathonTaskDetailsDescriptionList() {
+    let task = MarathonStore.getTaskFromTaskID(this.props.params.taskID);
```
pierlo-upitup: Why aren't we passing the `task` like we do for the other `get-` methods?

src/js/components/TaskDetail.js
```diff
@@ -226,22 +289,28 @@ class TaskSidePanelContents extends SidePanelContents {
...
-          className="container container-fluid container-pod container-pod-short flush-bottom"
+          className="container container-fluid flush container-pod container-pod-super-short flush-top"
           hash={headerValueMapping}
           headline="Configuration" />
         <DescriptionList
-          className="container container-fluid container-pod container-pod-short flush-bottom"
+          className="container container-fluid flush container-pod container-pod-super-short flush-top"
           hash={labelMapping}
           headline="Labels" />
+        {this.getMarathonTaskDetailsDescriptionList()}
```
pierlo-upitup: We should pass `task` to this function.

src/js/components/TaskDetail.js
```diff
@@ -287,37 +353,66 @@ class TaskSidePanelContents extends SidePanelContents {
     super.tabs_handleTabClick(...arguments);
   }
 
+  getNotFound(itemType) {
```
pierlo-upitup: Why the premature generalisation here? It seems to me we only ever call this with `task` as the `itemType`. If we have concrete use cases then I'd also like to see the `taskID` being an argument so that we can more easily extract this whole thing into an own component if appropriate.

src/js/components/TaskDetail.js
```diff
@@ -287,37 +353,66 @@ class TaskSidePanelContents extends SidePanelContents {
...
+    return (
+      <div className="container container-fluid container-pod text-align-center">
+        <h3 className="flush-top text-align-center">
+          {`Error finding ${itemType}`}
+        </h3>
+        <p className="flush">
+          {`Did not find a ${itemType} by the id "${this.props.params.taskID}"`}
```
pierlo-upitup: To expand on my comment above, relying on `this.props.params.taskID` makes this generic error node far less generic.

src/js/components/TaskDetail.js
```diff
@@ -287,37 +353,66 @@ class TaskSidePanelContents extends SidePanelContents {
...
+
+  getExpandButton() {
+    if (!this.state.showExpandButton) {
+      return null;
+    }
+
+    return (
+      <button
+        className="button button-stroke button-expand"
+        onClick={this.handleExpand}>
```
pierlo-upitup: This was part of the `SidePanelContents.js` component which we don't extend anymore, so the functionality is missing.

src/js/components/TaskDetail.js
```diff
@@ -151,56 +151,119 @@ class TaskSidePanelContents extends SidePanelContents {
...
+      Ports: task.ports,
+      Endpoints: this.getTaskEndpoints(task),
+      Status: this.getTaskStatus(task),
+      'Staged at': task.stagedAt,
+      'Started at': task.startedAt,
+      Version: task.version
+    };
+
+    return (
+      <DescriptionList
```
pierlo-upitup: Due to the `DescriptionList` using the `hash` keys as JSX `key` values we get a good old `tracker.js:10 Warning: Each child in an array or iterator should have a unique "key" prop. Check the render method of `DescriptionList`. It was passed a child from TaskDetail`. Outside of this PR we should fix this.

src/js/components/TaskDetail.js
```diff
@@ -151,56 +151,119 @@ class TaskSidePanelContents extends SidePanelContents {
...
 
-  getBasicInfo(task, node) {
-    // Hide when no task or when we are viewing debug tab
-    if (task == null || this.state.currentTab === 'debug') {
-      return null;
-    }
+  getBasicInfo(task) {
+    let taskIcon = (
+      <div
+        className="icon icon-large icon-image-container icon-app-container">
```
pierlo-upitup: Remove this `div` please, `PageHeader` adds it for us already.

src/js/components/TaskDetail.js
```diff
@@ -151,56 +151,119 @@ class TaskSidePanelContents extends SidePanelContents {
...
+      Endpoints: this.getTaskEndpoints(task),
+      Status: this.getTaskStatus(task),
+      'Staged at': task.stagedAt,
+      'Started at': task.startedAt,
+      Version: task.version
+    };
+
+    return (
+      <DescriptionList
+        className="container container-fluid flush container-pod container-pod-super-short flush-top"
```
pierlo-upitup: As `labels` can potentially be very long, encoded strings I suggest adding a `ddClassName="text-overflow-break-word"` to constraint the width of the page. 

Here's a gif for illustration purposes:

wordwrap

src/js/routes/services.js
```diff
@@ -24,7 +25,15 @@ let serviceRoutes = {
         {
           type: Route,
           name: 'services-detail',
-          path: ':id/?'
+          path: ':id/?',
+          children: [
+            {
+              type: Route,
+              name: 'services-task-panel',
```
pierlo-upitup: We should rename this route to a more appropriate `services-task-detail` now that the `panel` is gone.

src/js/routes/nodes.js
```diff
@@ -45,12 +46,8 @@ let nodesRoutes = {
         {
           type: Route,
           name: 'nodes-task-panel',
-          path: 'task-detail/:taskID/?'
-        },
-        {
-          type: Route,
-          name: 'node-detail-tasks',
-          path: 'task-detail/:taskID/?'
+          path: 'task-detail/:taskID/?',
```
pierlo-upitup: Why do we even need this route again? Can't we just use the one in `services.js` for the task detail route?

pierlo-upitup: Nvm – I just realised that we can reach this from the Nodes page.

src/js/components/TaskTable.js
```diff
@@ -196,7 +189,10 @@ class TaskTable extends React.Component {
 
 TaskTable.propTypes = {
   className: React.PropTypes.string,
-  parentRouter: React.PropTypes.func.isRequired,
+  parentRouter: React.PropTypes.oneOfType([
```
pierlo-upitup: Why do we need to pass down the `ServiceDetailTaskTab` router as a prop to all these components? Is it a workaround for tests?

src/js/routes/nodes.js
```diff
@@ -45,12 +46,8 @@ let nodesRoutes = {
...
+          handler: TaskDetail
```
pierlo-upitup: We should probably wrap this in a NodeDetailPage handler though, so that we can display the Nodes breadcrumbs. Without that, when you click on a Task from the Nodes page you end up in a navigation dead end.

src/js/stores/MarathonStore.js
```diff
@@ -159,6 +159,27 @@ var MarathonStore = Store.createStore({
     return this.get('apps')[name];
   },
 
+  getServiceFromTaskID: function (taskID) {
+    let serviceName = taskID.split('.')[0].split('_');
+    serviceName = serviceName[serviceName.length - 1];
```
pierlo-upitup: This doesn't seem very robust. I'd like to suggest extracting the `taskID` parsing to a Util so that we can test the hell out of it and generally to add a few somewhat more defensive checks.

src/js/structs/Task.js
```diff
@@ -0,0 +1,16 @@
+import Item from './Item';
+import ServiceImages from '../constants/ServiceImages';
+
+module.exports = class Task extends Item {
+  getId() {
+    return this.get('id') || '';
```
pierlo-upitup: What's the idea behind returning an empty string instead of `null`?

src/js/structs/Task.js
```diff
@@ -0,0 +1,16 @@
...
+  }
+
+  getImages() {
```
pierlo-upitup: Please share your thoughts on my comment https://github.com/dcos/dcos-ui/pull/118/files#r62646421

src/styles/layout/page.less
```diff
@@ -121,6 +121,17 @@
...
   }
+
+
+  // Make sure to grow Gemini view in page content
+  > .gm-scroll-view {
+
+    display: flex;
+    flex-direction: column;
+    flex: 1 0 auto;
+
```
pierlo-upitup: Remove extra newline?

pierlo-upitup:
Extremely good job! A few nitpicks/suggestion but it mostly works flawlessly. Thanks!

src/js/components/TaskDetail.js
aldipower: Could you explain @mlunoe ?

src/js/components/TaskDetail.js
aldipower: I will get this left ;)

aldipower:
@pierlo-upitup You did an amazing, precise and helpful review job. Thanx a lot!
I will take care of all your comments.

src/js/components/TaskDetail.js
aldipower: I would like to stay explicit here, because TABS is an object which is unsorted by definition.
It's just good luck to have it sorted.

src/js/components/TaskDetail.js
aldipower: Even this is an unsorted object..

src/js/components/TaskDetail.js
aldipower: Did it the other way around.

src/js/components/TaskDetail.js
aldipower: lol yes

src/js/components/TaskTable.js
aldipower: Yes, it's for the tests. Sorry.

src/js/routes/nodes.js
aldipower: It is wrapped in something like that.
But I've disabled the breadcrumbs for the node pages, because the need to be generalized first.

src/js/components/TaskDetail.js
aldipower: Nope.
One is the 'Mesos task', the other is a 'Marathon task'.
I've renamed the var to make it more clear.

src/js/structs/Task.js
aldipower: I wonder too.

src/js/components/MesosLogView.js
```diff
@@ -281,7 +281,7 @@ class MesosLogView extends mixin(StoreMixin) {
 
     return (
       <pre
-        className="flex-grow flush-bottom"
+        className="flex-grow flush-bottom prettyprint "
```
MatApple: Looks like we have an extra space at the end :)

src/js/components/TaskDetail.js
aldipower: That task comes from the Marathon store not from the Mesos store.
I've renamed the var to clarify this.

src/js/components/PageHeader.js
```diff
@@ -59,22 +60,25 @@ class PageHeader extends React.Component {
...
   title: React.PropTypes.string,
 
   className: React.PropTypes.string,
   dividerClassName: React.PropTypes.string,
+  iconClassName: React.PropTypes.string,
   mediaWrapperClassName: React.PropTypes.string,
-  titleClassName: React.PropTypes.string,
-  iconClassName: React.PropTypes.string
+  subTitleClassName: React.PropTypes.string,
+  titleClassName: React.PropTypes.string
```
MatApple: 🔤 ordering :)

src/js/components/TaskDetail.js
```diff
@@ -1,24 +1,29 @@
...
 
 import DescriptionList from './DescriptionList';
+import MarathonStore from '../stores/MarathonStore';
 import MesosStateStore from '../stores/MesosStateStore';
 import MesosSummaryStore from '../stores/MesosSummaryStore';
 import ResourceTypes from '../constants/ResourceTypes';
 import RequestErrorMsg from './RequestErrorMsg';
-import SidePanelContents from './SidePanelContents';
+import ServicesBreadcrumb from './ServicesBreadcrumb';
+import PageHeader from './PageHeader';
```
MatApple: 🔤

src/js/components/TaskDetail.js
```diff
@@ -1,24 +1,29 @@
...
 import TaskDebugView from './TaskDebugView';
 import TaskDirectoryView from './TaskDirectoryView';
 import TaskDirectoryStore from '../stores/TaskDirectoryStore';
 import TaskStates from '../constants/TaskStates';
-import TaskUtil from '../utils/TaskUtil';
 import Units from '../utils/Units';
 
+import InternalStorageMixin from '../mixins/InternalStorageMixin';
```
MatApple: 🔤

src/js/components/TaskDetail.js
MatApple: When you iterate over an Objects keys they will be returned in insertion order which is the same as the order declared when creating the Object. e.g. `{first:.., second:..}`. So we should be OK because we don't mutate the `tabs_tabs` Object in a way that would disrupt the order :)

src/js/components/TaskDetail.js
```diff
@@ -151,65 +151,126 @@ class TaskSidePanelContents extends SidePanelContents {
...
+      service.ipAddress.discovery.ports != null &&
+      task.ipAddresses != null &&
+      task.ipAddresses.length > 0) {
+
+      let ports = service.ipAddress.discovery.ports;
+      let endpoints = task.ipAddresses.reduce(function (memo, address) {
+        ports.forEach(port => {
+          memo.push(`${address.ipAddress}:${port.number}`);
+        });
+        return memo;
```
MatApple: #nit but we should have an extra line of whitespace above this `return memo;`

src/js/components/TaskDetail.js
aldipower: Interesting, that is new to me. Do you have a link for that? 
I thought that's why a Map was introduced. To ensure the order.
But we use an Object here, not a Map. So the order is not guaranteed. I am pretty sure.

src/js/components/TaskDetail.js
aldipower: Sorry, don't get it. :) Do you mean I should place the Mixins into the normal imports?

src/js/components/TaskDetail.js
```diff
@@ -151,65 +151,126 @@ class TaskSidePanelContents extends SidePanelContents {
...
+      MarathonStore.getTaskFromTaskID(this.props.params.taskID);
+
+    if (marathonTask == null) {
+      return null;
+    }
+
+    let headerValueMapping = {
+      'Host': marathonTask.host,
+      'Ports': marathonTask.ports,
+      'Endpoints': this.getTaskEndpoints(marathonTask),
```
MatApple: 🔤 's

src/js/components/PageHeader.js
aldipower: Sorry, don't get it. :)

src/js/components/TaskDetail.js
```diff
@@ -151,65 +151,126 @@ class TaskSidePanelContents extends SidePanelContents {
...
       .getServiceList();
     let service = services.filter({ids: [task.framework_id]}).last();
 
     let headerValueMapping = {
-      ID: task.id,
-      Service: `${service.name} (${service.id})`,
-      Node: `${node.hostname} (${node.id})`
+      'ID': task.id,
+      'Service': `${service.name} (${service.id})`,
+      'Node': `${node.hostname} (${node.id})`
```
MatApple: Curious why these have been added?

src/js/components/TaskDetail.js
aldipower: Actually this is sorted by importance/meaning.

src/js/components/TaskDetail.js
```diff
@@ -287,37 +351,56 @@ class TaskSidePanelContents extends SidePanelContents {
...
   }
 
+  getNotFound(item, itemID) {
+    return (
+      <div className="container container-fluid container-pod text-align-center">
+        <h3 className="flush-top text-align-center">
+          {`Error finding ${item}`}
+        </h3>
+        <p className="flush">
+          {`Did not find a ${item} by the id "${itemID}"`}
```
MatApple: How about `Did not find a ${item} with id "${itemID}"` ?

src/js/components/TaskTable.js
MatApple: Can't we just use `JestUtil.stubRouterContext`? It's based on [https://github.com/reactjs/react-router/blob/0.13.x/docs/guides/testing.md](https://github.com/reactjs/react-router/blob/0.13.x/docs/guides/testing.md)

src/js/components/__tests__/TaskTable-test.js
```diff
@@ -24,10 +25,17 @@ describe('TaskTable', function () {
     };
 
     this.container = document.createElement('div');
+
     this.instance = ReactDOM.render(
-      <TaskTable tasks={Tasks} parentRouter={this.parentRouter} />,
+      JestUtil.stubRouterContext(
+        TaskTable,{
```
MatApple: Think we need a space between `,{`

src/js/stores/MarathonStore.js
```diff
@@ -159,6 +159,30 @@ var MarathonStore = Store.createStore({
     return this.get('apps')[name];
   },
 
+  getServiceNameFromTaskID: function (taskID) {
```
MatApple: This method probably should be on a `Task` struct right?

src/js/stores/MarathonStore.js
```diff
@@ -159,6 +159,30 @@ var MarathonStore = Store.createStore({
...
+    let serviceName = taskID.split('.')[0].split('_');
+    return serviceName[serviceName.length - 1];
+  },
+
+  getServiceFromTaskID: function (taskID) {
+    let service = this.get('apps')[this.getServiceNameFromTaskID(taskID)];
+    if (service == null) {
+      return new Service();
```
MatApple: Should we still return a Service if no service is found? What would happen if we returned `null`?

src/js/stores/MarathonStore.js
```diff
@@ -159,6 +159,30 @@ var MarathonStore = Store.createStore({
...
+    }
+    return new Service(service.snapshot);
```
MatApple: We need an extra line above this `return` :)

src/js/stores/MarathonStore.js
```diff
@@ -159,6 +159,30 @@ var MarathonStore = Store.createStore({
...
+  },
+
+  getTaskFromTaskID: function (taskID) {
+    let service = this.getServiceFromTaskID(taskID);
+    if (service == null || service.tasks == null || !service.tasks.length) {
+      return null;
+    }
+
+    return service.tasks.find(function (task) {
```
MatApple: Is `find` in Babel Polyfill?

src/js/stores/MarathonStore.js
MatApple: I wonder if this should be on the `Task` struct?

src/js/stores/__tests__/MesosStateStore-test.js
```diff
@@ -142,12 +142,22 @@ describe('MesosStateStore', function () {
 
     it('should find a currently running task', function () {
```
MatApple: What do you think about adding a test to check that we get an `instanceof` `Task` struct?

src/js/stores/__tests__/MesosStateStore-test.js
```diff
@@ -142,12 +142,22 @@ describe('MesosStateStore', function () {
...
       var result = MesosStateStore.getTaskFromTaskID(1);
-      expect(result).toEqual({id: 1});
+      expect(result).toEqual({
```
MatApple: You can also do `expect(result.get()).toEqual({id: 1});` which is a little less work for yourself 😃

src/js/stores/MarathonStore.js
aldipower: yes :)

src/js/structs/Task.js
```diff
@@ -0,0 +1,16 @@
+import Item from './Item';
+import ServiceImages from '../constants/ServiceImages';
+
+module.exports = class Task extends Item {
+  getId() {
+    return this.get('id');
+  }
+
+  getImages() {
```
MatApple: What do you think about having specific methods for each image size instead? `getLargeImage()`, `getSmallImage()` etc

leemunroe:
@aldipower Nice work Felix. We discussed a lot of this in person but documenting it here. Mostly stylistic.

* Is it important to differentiate Mesos details from Marathon details with a heading or could these all be combined? I'm not quite clear on the differences and if users expect them to be called out as different groups of information
* The icon, heading and state positioning should be consistent throughout so when I go from Service detail to Task detail there isn't a jump in alignment
* For the list of key:value details it feels like the key column is very wide in most cases and could be narrower
* I think "Log Viewer" could just be "Logs"
* There is some odd styling left on the log container e.g. white border
* The margin between the tabs and search bar, then between the search bar and logs container, looks different, I would suggest keeping these consistent
* Add a tooltip to the download button so it is clear what it does
* Ensure long values are wrapped and don't introduce a horizontal scrollbar

src/js/components/MesosLogView.js
```diff
@@ -281,7 +281,7 @@ class MesosLogView extends mixin(StoreMixin) {
 
     return (
       <pre
-        className="flex-grow flush-bottom"
+        className="flex-grow flush-bottom prettyprint"
```
mlunoe: From the looks of the screenshots it still looks like the border is somehow bigger to the left and bottom than top and right

src/js/components/TaskDetail.js
pierlo-upitup: Still I think the `task` should be passed consistently to keep the code consistent 

`getMarathonTaskDetailsDescriptionList(marathonTask)`
...
`getResources(mesosTask)`

aldipower: OK

src/js/structs/Task.js
aldipower: Sounds like overkill to me. Sorry. :)

src/js/components/MesosLogView.js
aldipower: Yes, lee wrote this down in the notes above, so this shouldn't get lost.
Because this PR doesn't introduce the prettyprint itself, this should also be fixed outside of this PR for the sake of keeping the PRs small and consistent.

src/js/stores/MarathonStore.js
aldipower: You're right. This should return 'null', otherwise it's a too confusing behavior.

src/js/stores/MarathonStore.js
aldipower: This function returns a Service, so I think it doesn't belong to the Task.

src/js/components/MesosLogView.js
aldipower: OK, sorry, since this bug was introduced in this PR I fixed it here. :) Done.

src/js/components/TaskDetail.js
```diff
@@ -222,10 +222,7 @@ class TaskDetail extends mixin(InternalStorageMixin, TabsMixin, StoreMixin) {
     });
   }
 
-  getMarathonTaskDetailsDescriptionList() {
-    let marathonTask =
-      MarathonStore.getTaskFromTaskID(this.props.params.taskID);
-
+  getMarathonTaskDetailsDescriptionList(marathonTask) {
```
pierlo-upitup: Great!

src/js/components/TaskDetail.js
```diff
@@ -248,24 +245,22 @@ class TaskDetail extends mixin(InternalStorageMixin, TabsMixin, StoreMixin) {
     );
   }
 
-  renderDetailsTabView() {
-    let task = MesosStateStore.getTaskFromTaskID(this.props.params.taskID);
-
-    if (task == null || !MesosSummaryStore.get('statesProcessed')) {
+  getMesosTaskDetailsDescriptionList(mesosTask) {
```
pierlo-upitup: naming 🎳

src/js/components/TaskDetail.js
```diff
@@ -276,30 +271,51 @@ class TaskDetail extends mixin(InternalStorageMixin, TabsMixin, StoreMixin) {
...
+        className="container container-fluid flush container-pod container-pod-super-short flush-top"
+        hash={labelMapping}
+        headline="Labels" />
+    );
+  }
+
+  renderDetailsTabView() {
+    let mesosTask = MesosStateStore.getTaskFromTaskID(this.props.params.taskID);
+
+    let marathonTask =
```
pierlo-upitup: Why not `const` for these two guys?

pierlo-upitup:
@leemunroe I've captured the comments not addressed in the PR in a separate Jira for later.

src/js/components/TaskDetail.js
aldipower: Yeah, `const` is better. We should it use more regularly.

src/js/components/PageHeader.js
MatApple: Sorry for being vague! I was referring to the ABC ordering of the prop keys defined in this Object :) Just a nitpick - looks good!

src/js/components/TaskDetail.js
MatApple: Ok cool 👍

src/js/stores/MarathonStore.js
MatApple: Ok that's fair!

src/js/structs/Task.js
MatApple: Ok :)

src/js/stores/MarathonStore.js
orlandohohmeier: As this method is only parsing the task id I would rather move it to the `Task` struct or an util, as suggested by @MatApple.

orlandohohmeier:
Let's please make sure, that we (re)implement all the integration test (https://github.com/dcos/dcos-ui/blob/f39b03810d4ea57588d939ff2bf5b8750a9b9d08/tests/pages/services/TaskSidePanel-cy.js) to ensure the task detail page is properly working.

philipnrmn:
Obviously those can't exactly be re-implemented, but I can certainly implement some more tests if we want to keep the numbers up.

philipnrmn:
Last two rebase attempts on this branch were unsuccessful, but I have at least pushed some new tests.

src/js/components/TaskDetail.js
```diff
@@ -166,12 +166,17 @@ class TaskDetail extends mixin(InternalStorageMixin, TabsMixin, StoreMixin) {
...
+      'media-object-offset': false
+    };
+
     return (
       <PageHeader
         icon={taskIcon}
         subTitle={TaskStates[task.state].displayName}
         navigationTabs={tabs}
-        mediaWrapperClassName="media-object-spacing-wrapper"
+        mediaWrapperClassName={mediaWrapperClassNames}
```
philipnrmn: NB this can safely be merged, but won't actually work until a bug in `PageHeader` is fixed (#188)

pierlo-upitup:
🚢

src/js/components/TaskTable.js
orlandohohmeier: There is a `JestUtil` to wrap the component so that you can use the context to get the router.

src/js/structs/Task.js
```diff
@@ -0,0 +1,16 @@
+import Item from './Item';
+import ServiceImages from '../constants/ServiceImages';
+
+module.exports = class Task extends Item {
```
rcorral: Should there be tests written for this struct?

src/js/components/PageHeader.js
```diff
@@ -90,15 +90,25 @@ class PageHeader extends React.Component {
...
-          {children}
+          {navigationTabs}
         </div>
       </div>
     );
   }
 }
 
+PageHeader.defaultProps = {
+  className: 'container container-fluid container-pod container-pod-short flush flush-top',
```
daiyiwhale: This is probably an artifact of rebasing but we don't need these default classnames anymore since we are utilising the subtractive feature of `classnames` now

src/js/components/PageHeader.js
```diff
@@ -121,7 +131,8 @@ PageHeader.propTypes = {
   iconClassName: React.PropTypes.oneOfType([
     React.PropTypes.string,
     React.PropTypes.object,
-  ])
+  ]),
+  subTitleClassName: React.PropTypes.string,
```
daiyiwhale: Can we allow `subTitleClassName` to be a string or object in order to harness the full ability of `classnames`? (see above proptype declarations)

src/js/components/TaskDetail.js
```diff
@@ -0,0 +1,438 @@
...
+    if (service != null &&
+      service.ipAddress != null &&
+      service.ipAddress.discovery != null &&
+      service.ipAddress.discovery.ports != null &&
+      task.ipAddresses != null &&
+      task.ipAddresses.length > 0) {
+
+      let ports = service.ipAddress.discovery.ports;
+      let endpoints = task.ipAddresses.reduce(function (memo, address) {
+        ports.forEach(port => {
```
rcorral: No need to use phat arrow ;)

src/js/components/PageHeader.js
philipnrmn: Ah OK. Will remove them.

src/js/components/PageHeader.js
philipnrmn: I don't think it's in the master PageHeader so I'm not sure what it's still doing here...

daiyiwhale: Hm this PR doesn't seem to be using `subTitleClassName` ..... 👻

philipnrmn: ugh, it was introduced in this PR (https://github.com/dcos/dcos-ui/commit/95154b3523f3f7547def875a60846fdff58a3b20). Yes definitely agreed then.

src/js/components/TaskTable.js
orlandohohmeier: Added a Jira to track this. Shouldn't block this PR.

orlandohohmeier:
Seems like one of the refactorings broke the layout. 😖

orlandohohmeier:
*phew* Rebasing fixed the [layout issue](https://github.com/dcos/dcos-ui/pull/118#issuecomment-219882367)

src/js/structs/Task.js
orlandohohmeier: ✔️ Added tests

src/js/stores/MarathonStore.js
orlandohohmeier: I've create a Jira to keep track of this. Let's not block this PR.

philipnrmn:
FWIW the layout issue was addressed in my comment here: https://github.com/dcos/dcos-ui/pull/118/commits/ae12976ba489ecf2577627f14e56dce4d4cb04e8#r63559757

Looks like it was overly terse, will elaborate more next time.

src/js/components/PageHeader.js
```diff
@@ -108,20 +113,27 @@ class PageHeader extends React.Component {
...
 PageHeader.defaultProps = {
-  actionButtons: []
+  actionButtons: [],
+  className: 'container container-fluid container-pod container-pod-short flush flush-top',
+  dividerClassName: 'container-pod container-pod-short flush-top flush-bottom container-pod-divider-bottom container-pod-divider-bottom-align-right container-pod-divider-inverse',
+  iconClassName: 'icon icon-large icon-image-container icon-app-container',
+  mediaWrapperClassName: 'media-object-spacing-wrapper media-object-spacing-narrow media-object-offset',
+  subTitleClassName: 'emphasize',
+  titleClassName: 'flush inverse'
 };
```
philipnrmn: defaultProps are back again... let's kill 'em

src/js/components/PageHeader.js
```diff
@@ -108,20 +113,27 @@ class PageHeader extends React.Component {
...
 
 PageHeader.defaultProps = {
-  actionButtons: []
+  actionButtons: [],
+  className: 'container container-fluid container-pod container-pod-short flush flush-top',
+  dividerClassName: 'container-pod container-pod-short flush-top flush-bottom container-pod-divider-bottom container-pod-divider-bottom-align-right container-pod-divider-inverse',
+  iconClassName: 'icon icon-large icon-image-container icon-app-container',
+  mediaWrapperClassName: 'media-object-spacing-wrapper media-object-spacing-narrow media-object-offset',
+  subTitleClassName: 'emphasize',
+  titleClassName: 'flush inverse'
```
philipnrmn: defaultProps are back again! Let's get rid of them again

(c.f. https://github.com/dcos/dcos-ui/pull/118/files/d9a9be346cf5f776ece2e3180785d82d06deaed6#r63564663)

pierlo-upitup: Damnit github, broken link

philipnrmn: Fixed the link, sorry.

pierlo-upitup:
🚀

src/js/components/Page.js
```diff
@@ -105,7 +105,7 @@ var Page = React.createClass({
         {this.getPageHeader(title, navigation)}
         <GeminiScrollbar autoshow={true} className="page-content
           container-scrollable inverse" ref="pageRef">
-          <div className="flex-container-col container container-fluid container-pod container-pod-short-top">
+          <div className="flex-container-col flex-shrink flex-grow container container-fluid container-pod container-pod-short-top">
```
mlunoe: This did not fix the log view issues that we experienced and it had other implications on the application: https://github.com/dcos/dcos-ui/pull/228. Let's make sure to fix this properly.

PR #123, opened 2016-05-06
[1 of n] DCOS-6523: Add ServiceOptions to ServiceDetail

jfurrow:
I'm splitting this up into multiple PRs to keep them smaller, but I don't know how many there will be in total — probably around 5.

### DIFF
--- a/src/js/components/ServiceOptions.js
+++ b/src/js/components/ServiceOptions.js
@@ -88,7 +88,7 @@ class ServiceOptions extends mixin(InternalStorageMixin, StoreMixin) {
       </a>
     ];
 
-    if (service.getWebURL()) {
+    if (service.getWebURL && service.getWebURL()) {
       buttons.unshift(
         <a className="button button-primary" href={service.getWebURL()}
           target="_blank" key="open-service">
--- a/src/js/components/ServiceOptions.js
+++ b/src/js/components/ServiceOptions.js
@@ -62,9 +62,7 @@ class ServiceOptions extends mixin(StoreMixin) {
     this.setState({editConfigModalOpen: false});
   }
 
-  render() {
-    let {service} = this.props;
-
+  getServiceOptionButtons(service) {
     let editButtonClasses = classNames('button button-inverse button-stroke', {
       'disabled': !this.state.packageFetched
     });
@@ -85,9 +83,15 @@ class ServiceOptions extends mixin(StoreMixin) {
       );
     }
 
+    return buttons;
+  }
+
+  render() {
+    let {service} = this.props;
+
     return (
       <div className="button-collection flush-bottom">
-        {buttons}
+        {this.getServiceOptionButtons(service)}
       </div>
     );
   }
--- a/src/js/components/ServiceOptions.js
+++ b/src/js/components/ServiceOptions.js
@@ -1,0 +1,115 @@
+import classNames from 'classnames';
+import mixin from 'reactjs-mixin';
+import React from 'react';
+import {StoreMixin} from 'mesosphere-shared-reactjs';
+
+import CosmosPackagesStore from '../stores/CosmosPackagesStore';
+import Framework from '../structs/Framework';
+import InternalStorageMixin from '../mixins/InternalStorageMixin';
+import Service from '../structs/Service';
+
+const METHODS_TO_BIND = [
+  'handleEditConfigClick',
+  'handleConfigModalClose'
+];
+
+class ServiceOptions extends mixin(InternalStorageMixin, StoreMixin) {
+  constructor() {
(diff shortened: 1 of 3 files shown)


mesosphere-web:
Reviewer: @mlunoe

src/js/components/ServiceOptions.js
```diff
@@ -0,0 +1,115 @@
...
+    });
+
+    let buttons = [
+      <a className={editButtonClasses} key="edit-config"
+        onClick={this.handleEditConfigClick}>
+        Edit
+      </a>
+    ];
+
+    if (service.getWebURL()) {
```
kennyt: the `Service` struct doesn't have this right? so we need to check if this method exists before (you had this before)

jfurrow: You're right, for some reason I was thinking that the method was on the `Service` struct (and maybe it should be?) because `Framework` extends `Service`...

src/js/components/ServiceOptions.js
```diff
@@ -0,0 +1,115 @@
...
+class ServiceOptions extends mixin(InternalStorageMixin, StoreMixin) {
+  constructor() {
+    super(...arguments);
+
+    this.state = {
+      editConfigModalOpen: false
+    };
+
+    this.internalStorage_set({
+      servicePackage: null
```
kennyt: can we just blow this up?

in every part of the code where you need to access `servicePackage`, you can go directly to the source and do `CosmosPackagesStore.getPackageDetails()` and then use that variable

if you do this, you can remove the internalStorage mixin, + all of the methods where you manually listen to the cosmos store. (`StoreMixin` got your back because it'll forceUpdate and you'll have the package on the render)

src/js/components/ServiceOptions.js
```diff
@@ -0,0 +1,115 @@
...
+    }];
+
+    METHODS_TO_BIND.forEach((method) => {
+      this[method] = this[method].bind(this);
+    });
+  }
+
+  componentDidMount() {
+    super.componentDidMount(...arguments);
+
```
kennyt: let's store `this.props.service` as a variable since it's mentioned twice

src/js/components/ServiceOptions.js
```diff
@@ -0,0 +1,115 @@
...
+  handleEditConfigClick() {
+    let {servicePackage} = this.internalStorage_get();
+
+    if (servicePackage != null) {
+      this.setState({editConfigModalOpen: true});
+    }
+  }
+
+  handleConfigModalClose() {
+    this.setState({editConfigModalOpen: false});
```
kennyt: we never use `editConfigModalOpen` in this component. i'm assuming we'll have use for it later on 👍

kennyt:
this looks good. imo the only thing you'd need to unit test is the fact that it adds another button if there is a URL

src/js/components/ServiceOptions.js
jfurrow: That's a good point, but since `CosmosPackagesStore` only stores `packageDetails` for one package at a time, I don't know for sure if it's the package I want.

With your suggestion, it's possible that `CosmosPackagesStore#getPackageDetails` will return details for a package that I previously had selected.

With the way I have it implemented currently, I know for sure that the `packageDetails` will pertain to the service that I just opened, because I assume it's null until the fetch is successful.

What do you think? Perhaps we should change the way `CosmosPackagesStore` stores and retrieves package details?

kennyt:
when the button is disabled, it's hard to tell that the button is actually disabled. don't know how we should fix that though 

px

src/js/components/ServiceOptions.js
jfurrow: Done!

src/js/components/ServiceOptions.js
jfurrow: Yep that will be consumed by code in the next PR :)

src/js/components/ServiceOptions.js
kennyt: what you could do then is have a boolean called 'packageFetched' in `state` and default that to false, then set to true on the `onSuccess`. after that, you could pull directly from the Cosmos store.

what do you think?

jfurrow: That sounds good!

jfurrow: Done!

jfurrow:
@kennyt Tests added :)

jfurrow:
Regarding the "disabled" button, I'll bring it up with the designers!

jfurrow:
Closing to point to the feature branch `feature-branch/software-updates`

PR #129, opened 2016-05-06
[1 of n] DCOS-6523: Add ServiceOptions

jfurrow:


mesosphere-web:
Reviewer: @MatApple

kennyt:
Reviewed the previous PR already 👍

src/js/components/ServiceOptions.js
```diff
@@ -0,0 +1,107 @@
...
+    });
+
+    let buttons = [
+      <a className={editButtonClasses} key="edit-config"
+        onClick={this.handleEditConfigClick}>
+        Edit
+      </a>
+    ];
+
+    if (service.getWebURL && service.getWebURL()) {
```
rcorral: why wouldn't `#getWebURL` be available?

rcorral: If the answer is because we don't know if it's an instance of a Framework, then let's do that check instead of this one and then assume that the method is then available.

jfurrow: Done!

PR #130, opened 2016-05-06
[#5 Software Updates] Add Update Actions

MatApple:


mesosphere-web:
Reviewer: @orlandohohmeier

src/js/events/ServicePlanActions.js
```diff
@@ -0,0 +1,76 @@
...
+
+import AppDispatcher from './AppDispatcher';
+import Config from '../config/Config';
+import RequestUtil from '../utils/RequestUtil';
+
+const ServicePlanActions = {
+
+  fetchPlan: function (serviceID) {
+    RequestUtil.json({
+      url: `${Config.rootUrl}/service${serviceID}${Config.servicePlanAPIPath}`,
```
jfurrow: Will `serviceID` always be prefixed with a `/`?

MatApple: Good catch John. There will always be a prefix of `/` for services. The above works but only because Marathon is smart enough to notice we're missing and encoded `/` at the beginning of the serviceID so it adds it. The correct way of making this request is to encode the serviceID so it includes the `/` prefix as `%2F`, since all serviceID's essentially represent a path in a filesystem.

kennyt:
is this rebased? all the changes are from this PR?

src/js/events/__tests__/ServicePlanActions-test.js
```diff
@@ -0,0 +1,113 @@
+jest.dontMock('../ServicePlanActions');
+
+import ActionTypes from '../../constants/ActionTypes';
+import AppDispatcher from '../../events/AppDispatcher';
+import Config from '../../config/Config';
+import ServicePlanActions from '../ServicePlanActions';
+import RequestUtil from '../../utils/RequestUtil';
```
kennyt: :abc:

kennyt:
merging after tests pass and the one :abc:

src/js/events/__tests__/ServicePlanActions-test.js
MatApple: fixed

PR #134, opened 2016-05-09
Misc/services ui changes

pierlo-upitup:
Just some minor UI changes:

- Use `h4` instead of `h3` to fit the filter headline result text 
- Introduce placeholder buttons for Create Group / Deploy Service and make use of `FilterBar` to arrange them horizontally

bars

cc @leemunroe

### DIFF
--- a/src/js/components/ServiceSearchFilter.js
+++ b/src/js/components/ServiceSearchFilter.js
@@ -3,6 +3,7 @@ import mixin from 'reactjs-mixin';
 import React from 'react';
 /* eslint-enable no-unused-vars */
 
+import FilterBar from '../components/FilterBar';
 import FilterInputText from './FilterInputText';
 import QueryParamsMixin from '../mixins/QueryParamsMixin';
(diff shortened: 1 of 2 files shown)


mesosphere-web:
Reviewer: @rcorral

kennyt:
tried it out. looks great

philipnrmn:
Looks nice 👍

leemunroe:
@pierlo-upitup Nice! One thing, I'm noticing a slight jump in the UI when it switches to results mode. Not sure why. But it's a bit jarring.

pierlo-upitup:
@leemunroe yep the jump is there – I could fix this quick'n'dirty but I preferred to postpone this to later. 

Maybe @jfurrow can chime in on what could be a possible **good** solution that doesn't involve writing a bunch of unnecessary CSS fixes but that deals with Canvas UI `ul.h4` which is present in the FilterHeadline.

mlunoe:
It seems that there is a difference in margin between heading and the field, which makes the whole content 'jump' up. That we should fix

PR #135, opened 2016-05-09
Introduce struct for deployments

philipnrmn:
This PR introduces a 'struct' for the `Deployment` type, in preparation for a second PR which will introduce the `Deployment` to the `MarathonStore`

This PR must be merged before #136.

### DIFF
--- a/src/js/structs/__tests__/Deployment-test.js
+++ b/src/js/structs/__tests__/Deployment-test.js
@@ -1,0 +1,32 @@
+let Deployment = require('../Deployment');
+
+describe('Deployment', function () {
+
+  describe('#get...', function () {
+    it('calls through to `Item.get`', function () {
+      let deployment = new Deployment({
+        id: 'deployment-id',
+        version: '2001-01-01T01:01:01.001Z',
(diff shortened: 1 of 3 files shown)


mesosphere-web:
Reviewer: @rcorral

src/js/structs/__tests__/Deployment-test.js
```diff
@@ -0,0 +1,16 @@
+let Deployment = require('../Deployment');
```
kennyt: can you write a single test for all the simple `.get('string')`s?

philipnrmn: Sure. For my own interest, could you give me a quick run-down on why that would be a useful test?

philipnrmn: @kennyt added a test and squashed.

kennyt: if the API changed, we would then change the dummy data we're using and the tests would fail 👍

kennyt:
merging on test pass

PR #136, opened 2016-05-09
Fetch deployments from Marathon

philipnrmn:
~~This PR is dependent on functionality introduced in #135~~

This PR introduces functionality to the `MarathonStore`. It now fetches data from the Marathon deployments endpoint (ref [the docs](http://mesosphere.github.io/marathon/docs/rest-api.html#deployments)) and holds it on the store, inline with similar functionality which fetches services data from Marathon. 

This PR must be merged before #140


mesosphere-web:
Reviewer: @mlunoe

src/js/events/MarathonActions.js
```diff
@@ -47,6 +50,37 @@ module.exports = {
       };
     },
     {delayAfterCount: Config.delayAfterErrorCount}
+  ),
+
+  fetchDeployments: RequestUtil.debounceOnError(
```
mlunoe: Could we get an entry in `StoreConfig.js` for these actions?

src/js/stores/MarathonStore.js
```diff
@@ -217,10 +236,17 @@ var MarathonStore = Store.createStore({
...
+        MarathonStore.processMarathonDeployments(action.data);
+        break;
       case ActionTypes.REQUEST_MARATHON_GROUPS_ERROR:
         MarathonStore.processMarathonGroupsError();
         break;
+      case ActionTypes.REQUEST_MARATHON_DEPLOYMENTS_ERROR:
+        MarathonStore.processMarathonDeploymentsError();
+        break;
       case ActionTypes.REQUEST_MARATHON_GROUPS_ONGOING:
+      case ActionTypes.REQUEST_MARATHON_DEPLOYMENTS_ONGOING:
```
mlunoe: Let's not make these fall through. This needs its own handler. If you are not going to use the functionality I suggest you remove it

src/js/stores/__tests__/MarathonStore-test.js
```diff
@@ -176,4 +177,14 @@ describe('MarathonStore', function () {
 
   });
 
+  describe('#processMarathonDeployments', function () {
```
mlunoe: Usually we test all the actions that the store receives. Here is a good example to look at:
https://github.com/dcos/dcos-ui/blob/master/src/js/stores/__tests__/UserStore-test.js

src/js/events/MarathonActions.js
mlunoe: Let's also get some tests for this new action. Here is a good example to look at: https://github.com/dcos/dcos-ui/blob/master/src/js/events/UsersActions.js

src/js/stores/MarathonStore.js
philipnrmn: `MarathonStore.processOngoingRequest` is a no-op, so presumably both cases should be removed?

src/js/events/MarathonActions.js
```diff
@@ -1,7 +1,10 @@
 import {
   REQUEST_MARATHON_GROUPS_SUCCESS,
   REQUEST_MARATHON_GROUPS_ERROR,
-  REQUEST_MARATHON_GROUPS_ONGOING
+  REQUEST_MARATHON_GROUPS_ONGOING,
+	REQUEST_MARATHON_DEPLOYMENTS_SUCCESS,
```
kennyt: indentation is off

src/js/stores/MarathonStore.js
```diff
@@ -56,7 +63,8 @@ var MarathonStore = Store.createStore({
   mixins: [GetSetMixin],
 
   getSet_data: {
-    apps: {}
+    apps: {},
+    deployments: []
```
kennyt: can we default this to an empty instance of `DeploymentsList`?

src/js/events/MarathonActions.js
philipnrmn: It is off on Github, but not in the file.

src/js/stores/MarathonStore.js
philipnrmn: Yes! This will be in a separate PR shortly.

src/js/events/MarathonActions.js
philipnrmn: ah, tabs.

src/js/stores/MarathonStore.js
mlunoe: Yeah, let's remove them both then, including the `processOngoingRequest` or just not add the `REQUEST_MARATHON_DEPLOYMENTS_ONGOING`. Up to you

jfurrow: For what it's worth, this pattern exists all over the UI for handling ongoing requests. I would vote we decide what to do with them and then take care of them all at once.

src/js/stores/MarathonStore.js
philipnrmn: https://github.com/dcos/dcos-ui/pull/140/files#diff-8b6eb6158535aefd0b151ff7df2802a3R67

(in PR #140)

src/js/stores/MarathonStore.js
philipnrmn: I created DCOS-6761 to track this.

rcorral: @jfurrow the pattern that exists on the application is what @mlunoe said, we have a unique handler for the ONGOING requests, we don't have the same handler as other ONGOING requests. So let's not create this tech debt.

jfurrow: The pattern I'm referring to is putting noop functions everywhere. But in this case, the noop is called `processOngoingRequests` so it could, in theory, process the ongoing requests for both groups and deployments.

Regardless, I think we should not be creating these noops unless there is a good reason to do so.

philipnrmn:
I addressed all feedback and rebased downstream branches.

PR #138, opened 2016-05-09
Don't render with Gemini on initial render of SchemaForm

jfurrow:


mesosphere-web:
Reviewer: @mlunoe

src/js/components/SchemaForm.js
```diff
@@ -252,10 +257,18 @@ class SchemaForm extends React.Component {
       );
     });
 
+    if (!this.state.renderGemini) {
```
mlunoe: Could we get a comment about why this is necessary? Something like: 'Gemini measures the height/width which causes the animation to cancel'

mlunoe:
After [this](https://github.com/dcos/dcos-ui/pull/138#discussion_r62591805) comment I think it looks good: ⛵

src/js/components/SchemaForm.js
jfurrow: Done!

PR #139, opened 2016-05-09
Make integration tests work for actual cluster, ref. DCOS-6597

mlunoe:


mesosphere-web:
Reviewer: @kennyt

tests/pages/dashboard/ComponentsList-cy.js
```diff
@@ -2,7 +2,8 @@ describe('ComponentsList [10b]', function () {
 
   context('No Components Found [10f]', function () {
 
-    it('shows error when components not found [10d]', function () {
+    // TODO: Turn into unit test
+    xit('shows error when components not found [10d]', function () {
```
kennyt: why is this required to be a unit test? i can see how this could also be an integration test

tests/pages/nodes/node/HealthTab-cy.js
```diff
@@ -36,15 +42,15 @@ describe('Node Health Tab [0fa]', function () {
       cy.get('@filterHealth').click();
       cy.get('.dropdown').find('li').contains('Healthy').click();
       cy.get('.page-content td .text-success').should(function ($row) {
-        expect($row.length).to.equal(18);
+        expect($row.length).to.be.at.least(5);
```
kennyt: where'd you get 5 from?

tests/pages/nodes/node/HealthTab-cy.js
```diff
@@ -36,15 +42,15 @@ describe('Node Health Tab [0fa]', function () {
...
       });
     });
 
     it('filters by health check name [0ff]', function () {
       cy.get('@filterTextbox').type('logrotate');
       cy.get('.page-content').within(function () {
         cy.get('td a').should(function ($row) {
-          expect($row.length).to.equal(2);
+          expect($row.length).to.be.at.least(2);
```
kennyt: where'd you get 2 from?

tests/pages/dashboard/ComponentsList-cy.js
kennyt: ahh i see. you can't force it to be empty. would it be possible for you to mock the response for tests where you need to mock it?

mlunoe: The problem with that is that it is not an actual integration test then. We can easily test this as a unit test, but until we have something where we can actually change the cluster, this is going to be hard.

The whole point of having these tests is that they should work for actual clusters too

tests/pages/nodes/node/HealthTab-cy.js
mlunoe: It's a good number :smile:. Ok, seriously: There are certain health components we can rely on being there and having at least 5 tests that there are some without having to test the exact number. I can add a comment for this if this deserves an explanation

tests/pages/nodes/node/HealthTab-cy.js
mlunoe: Same as here: https://github.com/dcos/dcos-ui/pull/139#discussion_r62901262

tests/pages/nodes/node/HealthTab-cy.js
mlunoe: Actually, it's a good point. I re-wrote the tests to be more robust: calculating the exact numbers it should show.

tests/pages/nodes/node/HealthTab-cy.js
mlunoe: See reply here: https://github.com/dcos/dcos-ui/pull/139#discussion_r62904526

tests/pages/nodes/node/HealthTab-cy.js
```diff
@@ -33,20 +39,31 @@ describe('Node Health Tab [0fa]', function () {
     });
 
     it('filters by health [0fe]', function () {
```
kennyt: very nice

tests/pages/nodes/node/HealthTab-cy.js
```diff
@@ -33,20 +39,31 @@ describe('Node Health Tab [0fa]', function () {
...
+          expect($row.length).to.equal($healthyRows.length);
+        });
+        // Unhealthy rows should not show
+        cy.get('.page-content td .text-danger').should(function ($row) {
+          expect($row.length).to.equal(0);
+        });
       });
     });
 
     it('filters by health check name [0ff]', function () {
```
kennyt: good stuff

kennyt:
looks good, haven't tested it out though

PR #140, opened 2016-05-10
Deployment table (deployment data only)

philipnrmn:
This PR depends on functionality introduced in #136. 

It introduces a table which shows deployments data. 

The deployments table final design looks like this:

This PR brings it to the following state:

The eventual intent is to have a separate table row for each deployment and each app which is affected by that deployment. In the interests of minimal changes per PR, only the deployment rows are implemented at present. Status bars and `Rollback` links will be implemented separately.


mesosphere-web:
Reviewer: @rcorral

src/js/pages/services/DeploymentsTab.js
```diff
@@ -1,11 +1,73 @@
...
 import React from 'react';
+import {Table} from 'reactjs-components';
 
 import AlertPanel from '../../components/AlertPanel';
+import EventTypes from '../../constants/EventTypes';
+import MarathonStore from '../../stores/MarathonStore';
+import ResourceTableUtil from '../../utils/ResourceTableUtil';
+import StringUtil from '../../utils/StringUtil'
+
+const columnClassName = ResourceTableUtil.getClassName;
```
orlandohohmeier: Please rename the const, as the name kinda suggests that its simply holding class name (`string`) but it's actually a reference to the class name getter. 
BTW: Do we really need to cache this? I don't see any real performance improvements in doing so.

src/js/pages/services/DeploymentsTab.js
```diff
@@ -1,11 +1,73 @@
...
     super(...arguments);
+    this.onDeploymentsChange = this.onDeploymentsChange.bind(this);
+
+    this.state = {
+      deployments: MarathonStore.get('deployments')
+    };
+  }
+
+  componentDidMount() {
+    MarathonStore.addChangeListener(
```
orlandohohmeier: Not part of this PR, but it would be cool, if we could get the data from the DCOStore instead, to have components only deal with one unified store.

src/js/pages/services/DeploymentsTab.js
```diff
@@ -19,8 +81,34 @@ class DeploymentsTab extends React.Component {
     );
   }
 
+  renderPopulated() {
+    let state = this.state;
+    let deployments = state.deployments.getItems();
```
orlandohohmeier: Let's use a destructuring to get the deployments from the current state.

orlandohohmeier:
@philipnrmn Looks quite good, besides a few nits. Solid work!

src/js/pages/services/DeploymentsTab.js
philipnrmn: Yeah I wondered about this. It's less a cache and more a set of statics. 

I was using `ServicesTable` as a reference implementation and this is one of the deviations I made. It seemed rather odd to me to recreate the whole object, complete with methods, on every render. 

I agree that from the user's perspective there is zero difference either way.

src/js/pages/services/DeploymentsTab.js
philipnrmn: Yep, working towards that in the long run.

src/js/pages/services/DeploymentsTab.js
philipnrmn: I did this, but I really dislike the resulting code.

philipnrmn:
Addressed the nits and fixed up.

philipnrmn:
No idea why those tests were failing...

*walks away whistling innocently*

orlandohohmeier:
All test passed, let's 🚢 🇮🇹

src/js/pages/services/DeploymentsTab.js
```diff
@@ -1,11 +1,73 @@
...
+];
 
 class DeploymentsTab extends React.Component {
 
   constructor() {
     super(...arguments);
+    this.onDeploymentsChange = this.onDeploymentsChange.bind(this);
+
+    this.state = {
+      deployments: MarathonStore.get('deployments')
```
pierlo-upitup: I find it quite unsettling that we're assigning the Deployments List by reference to our state – we have no protection from external mutations. That's a generic issue with our Structs approach, not a problem in this PR per se.

PR #142, opened 2016-05-10
Fix the lint plugins script

orlandohohmeier:
Use correct npm config variable as the linting would otherwise only lint `/**/*.js` due to the undefined env.


mesosphere-web:
Reviewer: @rcorral

MatApple:
@orlandohohmeier If we change this here, we need to change it everywhere. The same variable is used within `/src`!

orlandohohmeier:
@MatApple Oh, yeah! You're right. I'll fix them.

orlandohohmeier:
Closing this PR as discussed with @MatApple as we've to consider the CI integration constraints (`npm set`) and need to "fix" it in a different way.

PR #145, opened 2016-05-10
Remove upgrade actions from PackagesTable

jfurrow:


mesosphere-web:
Reviewer: @mlunoe

kennyt:
can you write in the descriptions why this is happening? did people not want this on the table anymore?

jfurrow:
@kennyt Honestly I'm not exactly sure why it's happening...

kennyt:
@jfurrow what i mean was i didn't know if either it's being removed because it's never coming back, or if you're removing it because you're refactoring into another component. i'm assuming it's never coming back 👍

jfurrow:
@kennyt Ahh right on, yeah I think either it will be implemented elsewhere, or maybe in this table, but in a future release.

kennyt:
px

is the `Remove` button supposed to be there?

kennyt:
:ship: after that right align comment

src/js/components/PackagesTable.js
```diff
@@ -97,7 +82,7 @@ class PackagesTable extends React.Component {
         headerClassName: getClassName,
         heading: function () {},
         prop: 'uninstall',
-        render: this.getUpgradeCell,
```
mlunoe: I think we need to decrease the column width for this column as well

jfurrow: Done!

jfurrow:
@kennyt It's right-aligned now :)

kennyt:
tested this and it looks good 👍

PR #153, opened 2016-05-11
Only highlight tab on prefix match

philipnrmn:
This PR updates the tab utility so that instead of making a tab active when its name exists anywhere in the route, it only makes the tab active when its name exists at the start of the route. 

- The tab named `universe-packages` will be active with either of the routes `universe-packages` or `universe-packages-detail`.
- The tab named `services` will be active with `services` and `services-detail` but not with `-services-deployment`. 

This PR provides the basis for the fix in #166

### DIFF
--- a/src/js/utils/__tests__/TabsUtil-test.js
+++ b/src/js/utils/__tests__/TabsUtil-test.js
@@ -6,6 +6,7 @@ describe('TabsUtil', function () {
   describe('#getTabs', function () {
     beforeEach(function () {
       this.tabs = {foo: 'bar', baz: 'qux', corge: 'grault'};
+      this.hierarchicalTabs = {foo: 'bar', foobar: 'foobar', '-foo': 'baz'};
       this.getElement = function () {};
       spyOn(this, 'getElement');
     });
@@ -24,7 +25,7 @@ describe('TabsUtil', function () {
       expect(result.length).toEqual(3);
     });
 
-    it('should return li\'s', function () {
+    it('should return LIs', function () {
       var result = TabsUtil.getTabs(this.tabs, 'baz', this.getElement);
 
       expect(result[0].type).toEqual('li');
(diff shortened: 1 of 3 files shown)


mesosphere-web:
Reviewer: @mlunoe

kennyt:
i remember there was a reason it was implemented this hacky way.

for example, this breaks the highlighting of the package-detail for universe.

px

if you find a way to fix that, then this PR will be cool

daiyiwhale:
Hm ):

philipnrmn:
Thanks for the catch @kennyt.

philipnrmn:
💡 What about if we just use startsWith? That way we can continue to use prefixes in routes, and we can also keep the readability of the `service-deployment` route by introducing a prefix like a hyphen.

src/js/utils/__tests__/TabsUtil-test.js
```diff
@@ -24,22 +25,40 @@ describe('TabsUtil', function () {
...
+    it('should return LIs', function () {
       var result = TabsUtil.getTabs(this.tabs, 'baz', this.getElement);
 
       expect(result[0].type).toEqual('li');
       expect(result[1].type).toEqual('li');
       expect(result[2].type).toEqual('li');
     });
 
-    it('should return element\'s with one active class', function () {
+    it('should return elements with one active class', function () {
```
philipnrmn: I fixed up a couple of `Deppenapostrophe` while I was in here.

philipnrmn:
Tests and lint are passing locally, looks like the Velocity build failed during `npm install`.

pierlo-upitup:
Yeah, I see a bunch of untar errors like `npm ERR! tar.unpack untar error /root/.npm/core-js/1.2.6/package.tgz`. Weird!

pierlo-upitup:
So how do we get this PR merged @rcorral ? I tried adding `npm cache clean` to our build config but it didn't seem to work. I can't think of other reasons why the build failed here – any tips?

rcorral:
We run the tests again :) trying that now.

src/js/utils/__tests__/TabsUtil-test.js
kennyt: good stuff

src/js/utils/__tests__/TabsUtil-test.js
```diff
@@ -24,22 +25,42 @@ describe('TabsUtil', function () {
...
       var result = TabsUtil.getTabs(this.tabs, 'baz', this.getElement);
 
       expect(result[0].props.className).toEqual('tab-item');
       expect(result[1].props.className).toEqual('tab-item active');
       expect(result[2].props.className).toEqual('tab-item');
     });
 
+    it('should not highlight routes contained within the class name',
+        function () {
+          var tabs = TabsUtil.getTabs(this.hierarchicalTabs, '-foo',
```
mlunoe: See my note here: https://github.com/dcos/dcos-ui/pull/166/files#r63264866

I think we should update this test to not use the `-` prefix

PR #155, opened 2016-05-11
Add service detail configuration tab

orlandohohmeier:
---
⚠️  _WIP: Tests missing for many of the components and I'm unhappy with the sub list rendering_

---

Introduce a service detail configuration tab. 

**AC**
- [x] Service detail configuration tab
- [x] Handle service versions data in DCOSStore 
- [x] Tests for all components
- [x] Proper config (sub) list styling

/cc @leemunroe

### DIFF
--- a/src/js/components/ConfigurationView.js
+++ b/src/js/components/ConfigurationView.js
@@ -1,0 +1,242 @@
+import mixin from 'reactjs-mixin';
+import React from 'react';
+import {StoreMixin} from 'mesosphere-shared-reactjs';
+
+import DCOSStore from '../stores/DCOSStore'
+import DescriptionList from './DescriptionList';
+import Service from '../structs/Service';
+import StringUtil from '../utils/StringUtil';
+
+const sectionClassName = 'container-fluid container-pod container-pod-super-short flush flush-bottom';
+
+function fetchVersion(service, versionID) {
+  if (service.getVersions().get(versionID) == null) {
+    DCOSStore.fetchServiceVersion(service.getId(), versionID);
+  }
+}
+
+class ConfigurationView extends mixin(StoreMixin) {
+  constructor() {
+    super(...arguments);
+
+    this.store_listeners = [{name: 'dcos', events: ['change']}];
+  }
+
+  componentWillMount() {
+    super.componentWillMount();
+
+    let {service, versionID} = this.props;
+    fetchVersion(service, versionID);
+  }
+
+  componentWillReceiveProps({service, versionID}) {
+    super.componentWillReceiveProps(...arguments);
+
+    fetchVersion(service, versionID);
+  }
+
+  getGenralSection({id, cmd, cpus, disk, instances, mem}) {
+    let headerValueMapping = {
+      'ID': id,
+      'CPUs': cpus,
+      'Memory (MiB)': mem,
+      'Disk Space (Mib)': disk,
+      'Instances': instances,
+      'Command': cmd
+    };
+
+    return (
+      <DescriptionList className={sectionClassName}
+        hash={headerValueMapping}
+        headline="General" />
+    );
+  }
+
+  getDockerContainerSection({container}) {
+    if (container == null || container.docker == null) {
+      return null;
+    }
+
+    let {docker} = container;
+
+    let headerValueMapping = {
+      'Force Pull Image': docker.forcePullImage,
+      'Image': docker.image,
+      'Network': docker.network,
+      'Parameters': StringUtil.arrayToCommaSeparatedString(docker.parameters),
+      'Privileged': docker.privileged
+    };
+
+    return (
+      <DescriptionList className={sectionClassName}
+        hash={headerValueMapping}
+        headline="Docker Container" />
+    );
+  }
+
+  getPortDefinitionsSection({portDefinitions}) {
+    if (portDefinitions == null) {
+      return null;
+    }
+
+    let portConfigurations = portDefinitions.map(
+      function (portDefinition, index) {
+        let headline = `Port Definition ${index + 1}`;
+
+        if (portDefinition.name) {
+          headline += ` (${portDefinition.name})`;
+        }
+
+        return (
+          <DescriptionList className="nested-description-list"
+            hash={portDefinition}
+            headline={headline}
+            key={index} />
+        );
+      });
+
+    return (
+      <div className={sectionClassName}>
+        <h5>Port Definitions</h5>
+        {portConfigurations}
+      </div>
+    );
+  }
+
+  getHealthChecksSection({healthChecks}) {
+    if (healthChecks == null) {
+      return null;
+    }
+
+    let healthCheckConfigurations = healthChecks.map(
+      function (healthCheck, index) {
+
+        let headerValueMapping = {
+          'Protocol': healthCheck.protocol,
+          'Grace Period Seconds': healthCheck.gracePeriodSeconds,
+          'Interval Seconds': healthCheck.intervalSeconds,
+          'Timeout Seconds': healthCheck.intervalSeconds,
+          'Max Consecutive Failures': healthCheck.maxConsecutiveFailures
+        };
+
+        switch (healthCheck.protocol) {
+          case 'COMMAND':
+            headerValueMapping = Object.assign(headerValueMapping, {
+              'Command': healthCheck.command.value,
+            });
+            break;
+          case 'HTTP':
+            headerValueMapping = Object.assign(headerValueMapping, {
+              'Path': healthCheck.path,
+              'Port': healthCheck.port,
+              'Port Index': healthCheck.portIndex || 'PORT_INDEX',
+              'Port Type': healthCheck.portType,
+              'Ignore Http1xx': healthCheck.ignoreHttp1xx,
+            });
+            break;
+          case 'TCP':
+            headerValueMapping = Object.assign(headerValueMapping, {
+              'Port': healthCheck.port,
+              'Port Index': healthCheck.portIndex || 'PORT_INDEX',
+              'Port Type': healthCheck.portType
+            });
+            break;
+        }
+
+        return (
+          <DescriptionList className="nested-description-list"
+            hash={headerValueMapping}
+            headline={`Health Check ${index + 1} (${healthCheck.protocol})`}
+            key={index} />
+        );
+      });
+
+    return (
+      <div className={sectionClassName}>
+        <h5>Health Checks</h5>
+        {healthCheckConfigurations}
+      </div>
+    );
+  }
+
+  getLabelSection({labels}) {
+    return (
+      <DescriptionList className="nested-description-list"
+        hash={labels}
+        headline="Labels" />
+    );
+  }
+
+  getEnvironmentVariablesSection({env}) {
+    return (
+      <DescriptionList className="nested-description-list"
+        hash={env}
+        headline="Environment Variables" />
+    );
+  }
+
+  getOtherSection(config) {
+    let headerValueMapping = {
+      'Constraints': StringUtil.arrayToCommaSeparatedString(config.constraints),
+      'Dependencies': StringUtil.arrayToCommaSeparatedString(config.dependencies),
+      'Resource Roles': StringUtil.arrayToCommaSeparatedString(config.acceptedResourceRoles),
+      'Executor': config.executor,
+      'Backoff Factor': config.backoffFactor,
+      'Backoff Seconds': config.backoffSeconds,
+      'Max Launch Delay': config.maxLaunchDelaySeconds,
+      'URIs': StringUtil.arrayToCommaSeparatedString(config.uris),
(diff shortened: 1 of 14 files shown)


mesosphere-web:
Reviewer: @mlunoe

src/js/components/ServiceDetailConfigurationTab.js
```diff
@@ -0,0 +1,25 @@
...
+import Service from '../structs/Service';
+
+class ServiceConfigurationTab extends React.Component {
+
+  render() {
+    let {service} = this.props;
+    let {currentVersionID} = service.getVersionInfo();
+    return (
+      <ConfigurationView
+        headline={`Current Configuration (${currentVersionID})`}
```
pierlo-upitup: `Current Version` seems more appropriate.

src/js/components/ConfigurationView.js
```diff
@@ -0,0 +1,76 @@
...
+
+import DCOSStore from '../stores/DCOSStore'
+import DescriptionList from './DescriptionList';
+import Service from '../structs/Service';
+
+class ConfigurationView extends mixin(StoreMixin) {
+  constructor() {
+    super();
+
+    this.store_listeners = [{name: 'dcos', events: ['change']}];
```
pierlo-upitup: Isn't that a very generic event to listen to? Sounds like we need to check if the version actually changed in our `shouldComponentUpdate`?

src/js/components/ConfigurationView.js
```diff
@@ -0,0 +1,76 @@
...
+      'Backoff Factor': configuration.backoffFactor,
+      'Backoff': configuration.backoffSeconds,
+      'Max Launch Delay': configuration.maxLaunchDelaySeconds,
+      'URIs': configuration.uris,
+      'User': configuration.user,
+      'Args': configuration.args,
+      'Version': configuration.version
+    };
+
+    return (<DescriptionList
```
pierlo-upitup:
```
return (
  <DescriptionList
```
?

src/js/components/ConfigurationView.js
```diff
@@ -0,0 +1,76 @@
...
+    super();
+
+    this.store_listeners = [{name: 'dcos', events: ['change']}];
+  }
+
+  render() {
+    let {headline, service, versionID} = this.props;
+    let configuration = service.getVersions().get(versionID);
+
+    if (!configuration) {
```
pierlo-upitup: `configuration == null`

src/js/components/ConfigurationView.js
```diff
@@ -0,0 +1,76 @@
...
+      DCOSStore.fetchServiceVersion(service.getId(), versionID);
```
pierlo-upitup: We should perform this dance in `componentDidMount` and update the internal state with whatever Service Version the DCOSStore emits when ready by listening to the appropriate events on the Store. 
The render function should be "pure" and just concern itself to render whatever is in the state, without side-effects.

src/js/components/ConfigurationView.js
```diff
@@ -0,0 +1,76 @@
...
+      'URIs': configuration.uris,
+      'User': configuration.user,
+      'Args': configuration.args,
+      'Version': configuration.version
+    };
+
+    return (<DescriptionList
+      hash={headerValueMapping}
+      headline={headline} />);
+  }
```
pierlo-upitup: Please refactor the `render` function to just check if `this.state.version` is set and either return the loading screen or the DescriptionList.

src/js/components/ServiceDetailConfigurationTab.js
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+
+import ConfigurationView from './ConfigurationView';
+import Service from '../structs/Service';
+
+class ServiceConfigurationTab extends React.Component {
+
+  render() {
+    let {service} = this.props;
+    let {currentVersionID} = service.getVersionInfo();
```
pierlo-upitup: Do you envision a case where we might want to pass the ConfigurationView a version that is not the current one? If not, I'd suggest to just pass it the `service`.

src/js/components/ConfigurationView.js
```diff
@@ -0,0 +1,76 @@
+import React from 'react';
+import mixin from 'reactjs-mixin';
+import {StoreMixin} from 'mesosphere-shared-reactjs';
+
+import DCOSStore from '../stores/DCOSStore'
+import DescriptionList from './DescriptionList';
+import Service from '../structs/Service';
+
+class ConfigurationView extends mixin(StoreMixin) {
```
pierlo-upitup: I'd call this `ServiceConfigurationView`.

src/js/components/ConfigurationView.js
pierlo-upitup: nvm – the `DCOS_CHANGE` event is actually emitted for the right things. I only have a problem with its name, but that's out of this PR.

src/js/components/DescriptionList.js
```diff
@@ -23,11 +23,14 @@ class DescriptionList extends React.Component {
 
       // Check whether we are trying to render an object that is not a
       // React component
-      if (typeof value === 'object' && !Array.isArray(value) &&
-        value !== null && !React.isValidElement(value)) {
+      if (typeof value === 'object' && value !== null &&
+        !React.isValidElement(value)) {
 
+        // TODO (orlandohohmeier): indicate (e.g. indent) sub list for better ux
```
pierlo-upitup: I'd suggest to json.stringify objects into a `<pre>` like in Marathon UI. No need for sublists.

src/js/components/ConfigurationView.js
```diff
@@ -0,0 +1,76 @@
...
+      'Dependencies': configuration.dependencies,
+      'Labels': configuration.labels,
+      'Resource Roles': configuration.acceptedResourceRoles,
+      'Container': configuration.container,
+      'CPUs': configuration.cpus,
+      'Environment': configuration.env,
+      'Executor': 'todo',
+      'Health Checks': configuration.healthChecks,
+      'Instances': configuration.instances,
+      'IP Address': configuration.ipAddress,
```
pierlo-upitup: So all optional values will have a headline but no data?

src/js/components/ConfigurationView.js
pierlo-upitup: Or even `ServiceDetailConfigurationView`

src/js/components/ServiceDetailConfigurationTab.js
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+
+import ConfigurationView from './ConfigurationView';
+import Service from '../structs/Service';
+
+class ServiceConfigurationTab extends React.Component {
```
pierlo-upitup: does not match filename

src/js/components/ServiceDetailConfigurationTab.js
```diff
@@ -0,0 +1,25 @@
...
+
+  render() {
+    let {service} = this.props;
+    let {currentVersionID} = service.getVersionInfo();
+    return (
```
pierlo-upitup: Add a newline before return

leemunroe:
Nice one @orlandohohmeier. Just confirming you only want feedback on what is in the configuration tab for this PR (i.e. ignoring that the action buttons are missing and "Running" vs "Healthy")?

This is [this view](https://invis.io/3Q6PARCUD#/144153366_Services_V7_-Service_Detail-_Cassandra-_Configuration-).

* The "compare versions" feature is missing - we've talked about that being a nice to have and assuming it's a different effort, just noting it here for transparency
* If we're not going to compare versions, we'll still need a way to view historical versions and rollback to - can you confirm that is on the roadmap?
* Some of the formatting and visual hierarchy needs some style tweaking - maybe we can work on that together in SF e.g. it's not quite clear what is a heading at the minute
* What do we want to do about fields that have no value? I know this is something we've chatted about in the past, whether to leave empty, N/A, "Unspecified" or not show that row /cc @philipnrmn @pierlo-upitup 
* We could maybe cut down some spacing so the key:value pairs are closer together - chatted with @aldipower about this too, where we could use `column-2` instead of 3 (Sidenote: the styles and format we use here should be the same as what we use on the task detail view)
* For any JSON blobs, would love to see similar `<pre>` formatting carried through from what we have in Marathon today. Not in the mockup but I was definitely a fan of that.

image

src/js/components/ConfigurationView.js
mlunoe: Make that `componentWillMount`

src/js/components/DescriptionList.js
```diff
@@ -23,11 +23,14 @@ class DescriptionList extends React.Component {
 
       // Check whether we are trying to render an object that is not a
       // React component
-      if (typeof value === 'object' && !Array.isArray(value) &&
-        value !== null && !React.isValidElement(value)) {
+      if (typeof value === 'object' && value !== null &&
```
mlunoe: Why the change to accept arrays now? I'm sure there was a reason for it being there before.

mlunoe: This is used in different places, so let's check it still works as expected there

src/js/components/ConfigurationView.js
orlandohohmeier: I'd stick with the current naming as to be consistent with the `ServiceDetailTaskTab` and `TaskView` naming.

src/js/components/DescriptionList.js
orlandohohmeier: I opted for the sublists as it seemed easier than finding a way to properly render JSON blocks within a DescriptionList.

src/js/components/ConfigurationView.js
orlandohohmeier: Yeah, just as in the current Marathon UI version...

src/js/components/DescriptionList.js
orlandohohmeier: I wasn't able to so spot any issues and all the tests passed...

src/js/components/ConfigurationView.js
orlandohohmeier: Oh, yeah! This is a WIP implementation, as I wasn't sure how `componentWillMount` will work in combination with the `StoreMixin`. I will definitely refactor this!

src/js/components/ServiceDetailConfigurationTab.js
orlandohohmeier: Yeah, there we will an extend version of the configuration tab to display not only the current but also older versions.

src/js/components/ServiceDetailConfigurationTab.js
orlandohohmeier: "Current Configuration" is what the design says, but I'll update it.

src/js/components/ConfigurationView.js
```diff
@@ -0,0 +1,76 @@
+import React from 'react';
+import mixin from 'reactjs-mixin';
+import {StoreMixin} from 'mesosphere-shared-reactjs';
```
kennyt: :abc: these imports :)

src/js/components/DescriptionList.js
kennyt: well we should definitely keep the array check. this is because `value` cannot be an array.

orlandohohmeier: It's transformed into an object (`Object.assign({}, value)`) so that we will get sublists for arrays as well. But @leemunroe is currently exploring different ways to render the configuration details and I might drop this change...

kennyt: so then the keys would be the `index` and the value would be the values in the array? do we want that?

orlandohohmeier: We could join the index and the parent key/heading, but I will not iterate on this until we found a visually appealing way as the current styling neither works for "normal" object sublist nor array based sublist. So this change might get dropped.

src/js/components/DescriptionList.js
pierlo-upitup: And what about https://github.com/dcos/dcos-ui/pull/155#issuecomment-218921707 ?

pierlo-upitup:
Let me know once the tests are ready I'll be happy to give this another pass.

src/js/components/ConfigurationView.js
```diff
@@ -0,0 +1,90 @@
...
+
+function fetchVersion(service, versionID) {
+  if (service.getVersions().get(versionID) == null) {
+    DCOSStore.fetchServiceVersion(service.getId(), versionID);
+  }
+}
+
+class ConfigurationView extends mixin(StoreMixin) {
+  constructor() {
+    super();
```
jfurrow: We should pass `...arguments` to `super`

orlandohohmeier:
I'll split this into two separate PRs as it grew into a beast...

orlandohohmeier:
I've splitted the PR into smaller chunks. Please review #220, #221, and #222 instead.

PR #159, opened 2016-05-11
[EASY REVIEW] Edit checkbox table

kennyt:


src/js/components/CheckboxTable.js
```diff
@@ -48,12 +48,13 @@ class CheckboxTable extends React.Component {
 
   bulkCheck(isChecked) {
     let checkedIDs = [];
-    let {data, onCheckboxChange} = this.props;
+    let {data, onCheckboxChange, uniqueProperty} = this.props;
 
     if (isChecked) {
       data.forEach(function (datum) {
-        checkedIDs.push(datum.id);
+        checkedIDs.push(datum[uniqueProperty]);
```
kennyt: we don't always want to look at the `id` field. we should choose what the id is

src/js/components/CheckboxTable.js
```diff
@@ -104,7 +106,7 @@ class CheckboxTable extends React.Component {
   renderCheckbox(prop, row) {
     let {checkedItemsMap, uniqueProperty} = this.props;
     let rowID = row[uniqueProperty];
-    let checked = null;
+    let checked = false;
```
kennyt: form doesn't work if it's null. needs to be false

src/js/components/CheckboxTable.js
```diff
@@ -164,7 +166,7 @@ class CheckboxTable extends React.Component {
         buildRowOptions={this.getTableRowOptions}
         className={tableClassSet}
         columns={columns}
-        colGroup={this.getColGroup()}
+        colGroup={getColGroup()}
```
kennyt: allow external col group

mesosphere-web:
Reviewer: @jfurrow

src/js/components/CheckboxTable.js
jfurrow: Can we just add this to `propTypes`? And maybe add a noop as the `defaultProp` as I don't think we'll need it.

jfurrow:
After my one comment, it looks good!

PR #165, opened 2016-05-12
Simplify getTaskFromTaskID

aldipower:


pierlo-upitup:
Cleaner, nicer. LGTM

pierlo-upitup:
@orlandohohmeier for a review

src/js/stores/MesosStateStore.js
```diff
@@ -128,20 +128,15 @@ var MesosStateStore = Store.createStore({
...
+      });
 
-        if (equalTask) {
-          foundTask = task;
-          break;
-        }
-      }
-    }
+      return foundTask;
+    });
```
rcorral: Would be nice to make it more efficient instead rather than looping through all the services:
```

foo:
for (var i = 0; i < 100000; i++) {
  for (var j = 0; j < 9999999; j++) {
    if (j === 99) {
      break foo;
    }
  }
}
console.log(i, j);
```

orlandohohmeier: @rcorral It's not looping through all services – `Array.prototype.some()` executes the callback until it finds one where callback returns a true value, thus I don't see any performance issues here.

orlandohohmeier:
LGTM! Let's 🚢

src/js/stores/MesosStateStore.js
```diff
@@ -128,20 +128,15 @@ var MesosStateStore = Store.createStore({
     let services = this.get('lastMesosState').frameworks;
     let foundTask = null;
 
-    for (let serviceIndex in services) {
-      let service = services[serviceIndex];
+    services.some(function (service) {
```
kennyt: is it just me or before we were doing `for.. in` which is for objects, and here we are doing `.some` which is for arrays.

am i missing something?

orlandohohmeier: Yeah, we were doing `for (let serviceIndex in services)` which is less-than-ideal as services (lastMesosState. frameworks) is an array. So we should have fix that anyway.

kennyt: oh alright. so it has always been an array. got it.

orlandohohmeier: Forget about the `Array.prototype.find` part, as it wouldn't work... :(

PR #166, opened 2016-05-12
Fix services/deployments tab highlighting

philipnrmn:
NB: this PR depends on #153

Before:
screen shot 2016-05-11 at 21 29 37

After:
screen shot 2016-05-11 at 21 29 54

### DIFF
--- a/src/js/routes/services.js
+++ b/src/js/routes/services.js
@@ -7,7 +7,7 @@ import ServicesTab from '../pages/services/ServicesTab';
 
 let serviceRoutes = {
   type: Route,
-  name: 'services',
+  name: 'services-page',
   handler: ServicesPage,
   path: '/services/?',
   children: [
--- a/src/js/components/ServiceOverlay.js
+++ b/src/js/components/ServiceOverlay.js
@@ -108,7 +108,7 @@ class ServiceOverlay extends mixin(InternalStorageMixin) {
 
     // Did not find a service.
     if (!service) {
-      this.context.router.transitionTo('services');
+      this.context.router.transitionTo('services-page');
       this.context.router.transitionTo('services-panel', {serviceName});
       return;
     }
--- a/src/js/components/ServicesBreadcrumb.js
+++ b/src/js/components/ServicesBreadcrumb.js
@@ -22,7 +22,7 @@ class ServicesBreadcrumb extends React.Component {
 
     let breadcrumbNodes = [(
       <span className="crumb" key="/">
-        <Link to="services">Services</Link>
+        <Link to="services-page">Services</Link>
       </span>
     )];
 
--- a/src/js/components/Sidebar.js
+++ b/src/js/components/Sidebar.js
@@ -16,7 +16,7 @@ var MesosSummaryStore = require('../stores/MesosSummaryStore');
 var MetadataStore = require('../stores/MetadataStore');
 var SidebarActions = require('../events/SidebarActions');
 
-let defaultMenuItems = ['dashboard', 'services', 'nodes-list', 'universe', 'system'];
+let defaultMenuItems = ['dashboard', 'services-page', 'nodes-list', 'universe', 'system'];
 
 let {Hooks} = PluginSDK;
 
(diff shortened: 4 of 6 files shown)


mesosphere-web:
Reviewer: @mlunoe

src/js/routes/services.js
```diff
@@ -13,7 +13,7 @@ let serviceRoutes = {
   children: [
     {
       type: Route,
-      name: 'services-deployments',
+      name: '-services-deployments',
```
pierlo-upitup: why is the `-` necessary?

src/js/mixins/TabsMixin.js
```diff
@@ -6,6 +6,49 @@ import React from 'react';
...
+ *   render() {
+ *     return (
+ *       <Page
+ *         navigation={this.getNavigation()}>
+ *         <RouteHandler />
+ *       </Page>
+ *     );
+ *   }
+ * }
+ */
```
pierlo-upitup: Thanks for the 🏡  nice gardening

src/js/utils/TabsUtil.js
```diff
@@ -21,7 +21,7 @@ const TabsUtil = {
     return tabSet.map(function (tab, index) {
       let tabClass = classNames({
         'tab-item': true,
-        'active': currentTab.indexOf(tab) > -1
+        'active': currentTab.startsWith(tab)
```
pierlo-upitup: 👍

src/js/routes/services.js
pierlo-upitup: I get it. I wonder if we can find a better solution?

philipnrmn: `TabMixin` is expected to highlight tabs when their names appear within the route (for context, see https://github.com/dcos/dcos-ui/pull/153#issuecomment-218617058)

The upstream PR #153 tightens that requirement so that tabs are highlighted when their names appear at the start of the route. 

It's nice to reflect the route (`/services/deployments`) in the name, but doing that without some character prefixing the name causes the `TabMixin` to highlight both `services` and `services-deployments`. Adding a character prefix is the least hacky way I could come up with which allowed me to reflect the route in the name.

pierlo-upitup: Thanks for the nice explanation. Yeah, I can't think of anything better in-place, so I'm OK with it.

philipnrmn: One option is just to name it 'deployments', which I'm also OK with - it makes it seem like it might be a left-hand menu route though, which it isn't.

pierlo-upitup:
Added `hold_merge` as https://github.com/dcos/dcos-ui/pull/153 needs to be merged in first, correct @philipnrmn ?

src/js/routes/services.js
pierlo-upitup: Route names are deprecated since react-router 1.0, so I wouldn't worry too much about it.

kennyt:
can you rebase so i can see the changes more clearly?

orlandohohmeier:
LGTM. :shipit:

src/js/routes/services.js
mlunoe: I am late to the game here, my apologies.
I know that this will eventually be removed from react-router, but this hack to differentiate the top level route is not very nice. At least it got me confused and puzzled for some time until I started inspecting it in depth, which is not favourable. 
I believe we should stay consistent (and it shouldn't be too big of a change) with how we do it in the nodes page and name it `service-page`, even if it requires updating routes around the applicaiton: https://github.com/dcos/dcos-ui/blob/master/src/js/routes/nodes.js#L16

orlandohohmeier:
Hold merge until we've addressed https://github.com/dcos/dcos-ui/pull/166/files#r63264866

src/js/routes/services.js
orlandohohmeier: @mlunoe I've renamed the routes respectively.

orlandohohmeier:
I've addressed all the issues... :shipit:

kennyt:
tests failing. i just reran the build. will merge if tests pass

PR #168, opened 2016-05-12
Preserve serviceID format when dispatching actions

jfurrow:
@MatApple Let me know if you'd prefer to do this in some other way, but `ServicePlanStore` expects the dispatched `serviceID` to not be URI encoded.


MatApple:
Good idea @jfurrow! I'll add it to the extra actions I added in the other open PR

PR #172, opened 2016-05-13
Allow creating Service Groups

pierlo-upitup:
Allow users to create empty Service Groups.

groups

*NOTE*: it would be nice if the cursor would be placed at the end of the input field when focusing, but the correct way to introduce this functionality is to implement this in the reactjs-components `FieldInput.js` https://github.com/mesosphere/reactjs-components/blob/master/src/Form/FieldInput.js . Skipping for now. 

Also, validation in `dcos-ui` happens on submit, not on keyDown. This means the error is persisted until the user tries to submit again, which is not ideal. Discussion postponed. 

cc @leemunroe


mesosphere-web:
Reviewer: @mlunoe

src/js/pages/services/ServicesTab.js
```diff
@@ -148,14 +164,28 @@ var ServicesTab = React.createClass({
           <div className="flex-grow">
             {breadcrumbs}
             {filterHeadline}
-            <ServiceSearchFilter
-              handleFilterChange={this.handleFilterChange} />
+            <FilterBar
+              rightAlignLastNChildren={2}>
```
kennyt: `FilterBar` doesn't need to be on multiple lines (it can fit)

src/js/components/modals/ServiceGroupFormModal.js
```diff
@@ -0,0 +1,109 @@
+import mixin from 'reactjs-mixin';
+/* eslint-disable no-unused-vars */
+import React from 'react';
+/* eslint-enable no-unused-vars */
+
+import {StoreMixin} from 'mesosphere-shared-reactjs';
```
kennyt: `StoreMixin` should be in the externals right below `React`

src/js/components/modals/ServiceGroupFormModal.js
```diff
@@ -0,0 +1,109 @@
...
+import FormModal from '../FormModal';
+import MarathonStore from '../../stores/MarathonStore';
+
```
kennyt: there should be no newline between `MarathonStore` and `Validator`

kennyt:
after the 3 nits, I'd add my label

src/js/components/modals/ServiceGroupFormModal.js
```diff
@@ -0,0 +1,108 @@
...
+    ];
+  }
+
+  render() {
+    return (
+      <FormModal
+        ref="form"
+        disabled={this.state.disableNewGroup}
+        onClose={this.props.onClose}
+        onSubmit={this.handleNewGroupSubmit}
```
mlunoe: For your point about the error persisting you can just implement `onChange` with a reset state like so:
https://github.com/dcos/dcos-ui/blob/master/src/js/components/modals/AddRepositoryFormModal.js#L152

mlunoe:
@pierlo-upitup: It would be nice to see a PR for reactjs-components fixing the `InputField.j` problem you are mentioning here: https://github.com/dcos/dcos-ui/pull/172#issue-154685317

That can land separately to this PR and it seems necessary to the experience this feature is introducing

src/js/pages/services/ServicesTab.js
```diff
@@ -68,6 +72,18 @@ var ServicesTab = React.createClass({
     });
   },
 
+  handleCloseGroupFormModal: function () {
+    this.setState({
+      isServiceGroupFormModalShown: false
```
mlunoe: #nit I think this could be one line

src/js/pages/services/ServicesTab.js
```diff
@@ -68,6 +72,18 @@ var ServicesTab = React.createClass({
...
+    });
+  },
+
+  handleOpenGroupFormModal: function () {
+    this.setState({
+      isServiceGroupFormModalShown: true
```
mlunoe: #nit I think this could be one line

src/js/pages/services/ServicesTab.js
```diff
@@ -148,14 +164,27 @@ var ServicesTab = React.createClass({
           <div className="flex-grow">
             {breadcrumbs}
             {filterHeadline}
-            <ServiceSearchFilter
-              handleFilterChange={this.handleFilterChange} />
+            <FilterBar rightAlignLastNChildren={2}>
```
mlunoe: This function is becoming huge. Could we break this into smaller functions?
I know not all of this was introduced in this PR, but it is extending it, so I think it makes sense to handle it up front

mlunoe:
Merging since @kennyt already gave it his blessing

pierlo-upitup:
@mlunoe thanks for the feedback, will try to address your comments asap.

src/js/components/modals/ServiceGroupFormModal.js
pierlo-upitup: @Poltergeist  this only works for global-form errors, not for inline field validation. Additionally, the `onChange` seems to be triggered on field focus, which means that as soon as I click on the field I lose the error message. 

weird

Also, I'm not sure if the error `h4` is inline with the general styles. cc @leemunroe

pierlo-upitup:
@mlunoe @leemunroe this reactjs-components PR will enable the cursor position fix https://github.com/mesosphere/reactjs-components/pull/257

src/js/components/modals/ServiceGroupFormModal.js
Poltergeist: This is the same behaviour as we used in other places. For example the add repository dialog:

The inline field validation is triggered by the FormModal component.

image

image

pierlo-upitup: Thanks for the clarification. If this is a common pattern then 🚢

PR #173, opened 2016-05-13
Add CTA buttons when browsing an empty service group

pierlo-upitup:
This PR depends on https://github.com/dcos/dcos-ui/pull/172

Please merge #172 first!

edit from kenny: #172 is merged

image

### DIFF
--- a/src/js/pages/services/ServicesTab.js
+++ b/src/js/pages/services/ServicesTab.js
@@ -104,6 +104,20 @@ var ServicesTab = React.createClass({
     this.setState(state, this.resetFilterQueryParams);
   },
 
+  getAlertPanelFooter: function () {
+    return (
+      <div className="button-collection flush-bottom">
+        <button className="button button-stroke button-inverse"
+          onClick={this.handleOpenGroupFormModal}>
+          Create Group
+        </button>
+        <button className="button button-success">
+          Deploy Service
+        </button>
+      </div>
+    );
+  },
+
   getContents: function (id) {
     // Render loading screen
     if (!DCOSStore.dataProcessed) {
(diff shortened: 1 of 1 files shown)


mesosphere-web:
Reviewer: @MatApple

MatApple:
Hey @pierlo-upitup @leemunroe, is it in the plan to keep the breadcrumbs for this view? It would be nice to click back out to the parent group when you arrive at this instead of losing all context! :)

leemunroe:
Good call. Yeh the breadcrumbs here would be useful. 

And true, maybe change to `Create groups to organize...`

Poltergeist:
Breadcrumbs seem out of focus for that PR, I added the copy change.

src/js/pages/services/ServicesTab.js
```diff
@@ -103,6 +102,20 @@ var ServicesTab = React.createClass({
     this.setState(state, this.resetFilterQueryParams);
   },
 
+  getAlertPanelFooter: function () {
+    return (
+      <div className="button-collection flush-bottom">
+        <button className="button button-stroke button-inverse"
+          onClick={this.handleOpenGroupFormModal}>
```
kennyt: unfortunately, i believe the modal will never ever show, even if this is clicked. this is because we render the modal in `getServiceTreeView` (which we don't call if there are empty services).

this fix to this to render the modal always in `#render` right next to `#getContents`

Poltergeist: solved this issue, nice catch.

src/js/pages/services/ServicesTab.js
```diff
@@ -202,18 +216,25 @@ var ServicesTab = React.createClass({
...
   render: function () {
     let {id} = this.props.params;
+    id = decodeURIComponent(id);
+    let {state} = this;
+    let item = DCOSStore.serviceTree.findItemById(id) || DCOSStore.serviceTree;
 
-    return this.getContents(decodeURIComponent(id));
+    return (
+      <div>
+        {this.getContents(id)}
```
kennyt: can we pass in the item since we use it inside of `getContents` also?

kennyt:
adding my label. i think this looks good after my last comment

src/js/pages/services/ServicesTab.js
Poltergeist: Solved this issue by passing item instead of id.

pierlo-upitup:
@MatApple breadcrumb component added:

screen shot 2016-05-17 at 11 32 26

philipnrmn:
Happy to 🚢  once rebased

pierlo-upitup:
rebased it is!

MatApple:
Nice work @pierlo-upitup!

PR #179, opened 2016-05-13
DCOS-5902: Remove dummy data from UniversePackage struct

jfurrow:


mesosphere-web:
Reviewer: @kennyt

src/js/components/PackageUpgradeOverview.js
```diff
@@ -1,135 +0,0 @@
-import React from 'react';
```
kennyt: can you say something about why this is being removed?

jfurrow: This is now going to be implemented in the feature branch `feature-branch/software-updates`, so this component shouldn't be here.

PR #181, opened 2016-05-13
DCOS-6727: Collapse side panel with left bracket keyboard shortcut

jfurrow:
Leaves the sidebar behavior the same responsively, but allows users to toggle the sidebar with the left-bracket key `[`.


mesosphere-web:
Reviewer: @mlunoe

src/js/components/Page.js
```diff
@@ -2,8 +2,11 @@ var classNames = require('classnames');
 var GeminiScrollbar = require('react-gemini-scrollbar');
 var React = require('react');
 
+import EventTypes from '../constants/EventTypes';
```
kennyt: can we do the style we've been doing recently? like
```js
{PAGE_UPDATE, etc} from 'path/here'
```

src/js/components/Page.js
```diff
@@ -28,14 +31,28 @@ var Page = React.createClass({
       rendered: true
     });
     this.forceUpdate();
+    SidebarStore.addChangeListener(
```
kennyt: this seems like something we've already abstracted with the `StoreMixin`. let's just add the `SidebarStore` to the `StoreConfig` and use it from there

src/js/components/Sidebar.js
```diff
@@ -30,6 +30,10 @@ var Sidebar = React.createClass({
     router: React.PropTypes.func
   },
 
+  getInitialState: function () {
+    return {sidebarExpanded: true};
```
kennyt: let's use `SaveStateMixin` also :D

src/js/components/charts/Chart.js
```diff
@@ -38,6 +40,7 @@ var Chart = React.createClass({
       this.updateWidth();
     }
     window.addEventListener('resize', this.updateWidth);
+    SidebarStore.addChangeListener(EventTypes.PAGE_UPDATE, this.updateWidth);
```
kennyt: let's use StoreMixin :)

src/js/components/Sidebar.js
jfurrow: Ahh shit, okay I'll add a `TODO` here.

src/js/components/Page.js
jfurrow: Good call, I should have done this to begin with :)

src/js/components/Page.js
jfurrow: Removed because I implented the StoreMixin

src/js/components/Sidebar.js
jfurrow: https://mesosphere.atlassian.net/browse/DCOS-6909

src/js/components/charts/Chart.js
jfurrow: Done!

kennyt:
can you rebase this? i has to rebase this in order to test it

(it's because we upgraded to npm 3 now)

kennyt:
comment #1: you need to do `event.preventDefault` or something because `CMD + [` has some other use for me (it's going back in history.

comment #2: should we implement `CMD + ]` to enlarge the sidebar? instead of having `CMD + [` be the toggle

jfurrow:
@kennyt It's not `CMD + [` it's just `[`. And the button will toggle it back and forth, so no need for the opposite bracket to have the opposite behavior.

src/js/components/Page.js
```diff
@@ -23,19 +25,31 @@ var Page = React.createClass({
     ])
   },
 
+  componentWillMount: function () {
+    this.store_listeners = [
```
kennyt: actually since this is the old type of component, you can add it on directly to the object (similar to how you do `mixins: [etc, etc]`

src/js/components/charts/Chart.js
```diff
@@ -24,6 +25,12 @@ var Chart = React.createClass({
 
   componentWillMount: function () {
     this.internalStorage_set({width: null});
+    this.store_listeners = [
```
kennyt: actually since this is the old type of component, you can add it on directly to the object (similar to how you do `mixins: [etc, etc]`

src/js/components/Page.js
jfurrow: Done!

src/js/components/charts/Chart.js
jfurrow: Done!

jfurrow:
@kennyt The CSS is fixed, too (I had a CSS comment that wasn't commented, woops)

src/js/constants/StoreConfig.js
```diff
@@ -196,6 +199,13 @@ const ListenersDescription = {
     listenAlways: false
   },
 
+  sidebar: {
```
kennyt: every single store has `listenAlways` and `unmountWhen` set. can you just default them to `true` and a function that returns true, respectively?

jfurrow: Good call, missed those somehow. Done!

rcorral:
There's an exception that occurs when I run the applicaiton.

jfurrow:
@rcorral It's fixed. It had something to do with adding `store_listeners` as a key on the object that gets passed to React's `#createClass`, somehow it worked before the npm3 PR was merged.

PR #184, opened 2016-05-16
Add action buttons to page header

MatApple:


mesosphere-web:
Reviewer: @pierlo-upitup

daiyiwhale:
Just one feedback! Could we get `.page-header-container-right` to pad out the `actionButtons` so we don't have to pass in the buttons with padding each time?

MatApple:
Done @daiyiwhale! Great suggestion :)

pierlo-upitup:
I guess it'd be nice to get some context regarding where this would be useful and possibly an example usage (even if just as a screenshot). 
Code LGTM though :) 🚢

PR #187, opened 2016-05-17
Make dropdown menu fix only work for direct children

mlunoe:
When dropdowns are not direct children, but still children these introduced styles will still take effect, but they shouldn't.

Before:

After:

### DIFF
--- a/src/styles/layout/media-object.less
+++ b/src/styles/layout/media-object.less
@@ -138,7 +138,7 @@ components/media-objects.less
     padding-bottom: @base-spacing-unit * 1/2;
     padding-right: @base-spacing-unit * 1/2;
 
-    .dropdown-menu.down {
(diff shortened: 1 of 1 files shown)


mesosphere-web:
Reviewer: @aldipower

src/styles/layout/media-object.less
```diff
@@ -138,7 +138,7 @@ components/media-objects.less
     padding-bottom: @base-spacing-unit * 1/2;
     padding-right: @base-spacing-unit * 1/2;
 
-    .dropdown-menu.down {
+    > .dropdown-menu.down {
```
jfurrow: #nit Can we add an `&` before each of these? So it will read: `& > .dropdown-menu.down {`

mlunoe: Done

PR #188, opened 2016-05-17
De-dupe PageHeader classnames attributes

philipnrmn:
PageHeader was using the `classnames` API correctly, but classes were not actually being removed because we weren't using the `dedupe` version.

```
// Without dedupe:
classnames('class-1 class 2', {'class-1': false})
> 'class-1 class 2'
// With dedupe:
classnames('class-1 class 2', {'class-1': false})
> 'class-1'
```

### DIFF
--- a/src/js/components/PageHeader.js
+++ b/src/js/components/PageHeader.js
@@ -1,4 +1,4 @@
-import classNames from 'classnames';
+import classNames from 'classnames/dedupe';
 import React from 'react';
 
 class PageHeader extends React.Component {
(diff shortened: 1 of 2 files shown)


mesosphere-web:
Reviewer: @MatApple

rcorral:
Should we use `dedupe` everywhere we use `classnames`?

philipnrmn:
If we want this behaviour, then I should say yes.

daiyiwhale:
Oh wow, I hadn't noticed the dedupe version is 5x slower. I suppose we should be mindful

PR #189, opened 2016-05-17
Test dependencies for licenses that we don't want.

rcorral:
We are avoiding dependencies with a GPL license.
This PR adds a way to programmatically check and ensure we don't have any with a GPL license.


mesosphere-web:
Reviewer: @kennyt

PR #191, opened 2016-05-17
Bug: Node Table resource bar animation

daiyiwhale:
The resource bars on the Node Table were not always animating and displaying values properly (for example, when loading the UI in a different page and navigating to the Node page). This fix animates transform rather than width which will render properly on page changes.

### DIFF
--- a/src/styles/components/progressbar.less
+++ b/src/styles/components/progressbar.less
@@ -20,7 +20,10 @@
   .bar {
     box-shadow: 1px 0px 0px 0px @body-background-color;
     height: 100%;
-    transition: width 600ms;
+    width: 100%;
+    transform: scaleX(0);
+    transform-origin: 0;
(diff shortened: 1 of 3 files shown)


mesosphere-web:
Reviewer: @mlunoe

src/styles/components/progressbar.less
```diff
@@ -20,7 +20,10 @@
   .bar {
     box-shadow: 1px 0px 0px 0px @body-background-color;
     height: 100%;
-    transition: width 600ms;
+    width: 100%;
```
jfurrow: Let's just :abcd: the `width` property :)

jfurrow:
@daiyiwhale just the minor :abcd: comment then let's merge this!

daiyiwhale:
@jfurrow ✔️

PR #192, opened 2016-05-17
Add default prop for actionButtons

MatApple:


mesosphere-web:
Reviewer: @pierlo-upitup

PR #202, opened 2016-05-17
Fix DCOSStore data handling

orlandohohmeier:
Pull the data from the stores, as the` MesosSummaryStore` doesn't provide the state summary as payload.

### DIFF
--- a/src/js/stores/DCOSStore.js
+++ b/src/js/stores/DCOSStore.js
@@ -9,10 +9,11 @@ import Framework from '../structs/Framework';
 import MarathonStore from './MarathonStore';
 import MesosSummaryStore from './MesosSummaryStore';
 import ServiceTree from '../structs/ServiceTree';
+import SummaryList from '../structs/SummaryList';
 
 const METHODS_TO_BIND = [
-  'processMarathonData',
-  'processMesosData'
+  'onMarathonGroupsChange',
+  'onMesosSummaryChange'
 ];
 
 class DCOSStore extends EventEmitter {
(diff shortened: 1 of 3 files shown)


mesosphere-web:
Reviewer: @jfurrow

src/js/stores/DCOSStore.js
```diff
@@ -26,50 +27,42 @@ class DCOSStore extends EventEmitter {
...
-  processMarathonData(data) {
-    if (!(data instanceof ServiceTree)) {
+  onMarathonGroupsChange() {
+    let groups = MarathonStore.get('groups');
+    if (!(groups instanceof ServiceTree)) {
       return;
     }
 
-    this.data.marathon = data;
+    this.data.marathon = groups;
```
kennyt: i love :heart: love :heart: love :heart: love :heart: love :heart: love :heart: love :heart: love :heart: love :heart: the fact that this store doesn't use the getsetmixin

src/js/stores/DCOSStore.js
```diff
@@ -26,50 +27,42 @@ class DCOSStore extends EventEmitter {
...
     this.data = {
       marathon: new ServiceTree(),
-      mesos: {
-        frameworks: []
-      },
+      mesos: new SummaryList(),
       dataProcessed: false
     };
 
     this.proxyListeners = [
```
kennyt: we recently refactored `StoreMixin` so that it doesn't only work on components. it can work on straight up objects now. cc @MatApple 

we should create a jira for using it here

orlandohohmeier: But it wouldn't solve the "proxy" problem for me, would it?

kennyt: what's the proxy problem?

orlandohohmeier: Sorry, my comment wasn't very specific. I was wondering if you guys added enough magic, to enable it to automatically register listeners on a different store whenever someone adds a listener to this store 🔮⚛✨

kennyt: hmm we didn't add that level of magic, but we did create methods where you can actually do that yourself. @MatApple probably knows best here since he created it

src/js/stores/DCOSStore.js
```diff
@@ -26,50 +27,42 @@ class DCOSStore extends EventEmitter {
...
-   */
-  processMesosData(data) {
-    if (data == null || !Array.isArray(data.frameworks)) {
+  onMesosSummaryChange() {
+    let states = MesosSummaryStore.get('states');
+    if (!(states instanceof SummaryList)) {
       return;
     }
-    this.data.mesos = data;
-    this.data.dataProcessed = true;
```
mlunoe: Why is this not needed anymore? And not sure why it was set in both `processMarathonData` and `processMesosData` before either, but is it sufficient to have it in one place?

mlunoe:
Looks pretty good otherwise!

src/js/stores/DCOSStore.js
orlandohohmeier: We should always wait for the Marathon data, due to the way the data get's merged, thus the data processed flag will be updated once Marathon is ready.

jfurrow:
This looks great!

PR #204, opened 2016-05-18
Add ToggleButton

mlunoe:
This introduced a toggle button to use as a module in external plugins

### DIFF
--- a/src/js/components/ToggleButton.js
+++ b/src/js/components/ToggleButton.js
@@ -8,43 +8,32 @@ class ToggleButton extends React.Component {
       checked,
       children,
       className,
-      id,
       onChange
     } = this.props;
 
     return (
-      <span className={classNames(className)}>
+      <label className={classNames(className)}>
         <input
           className={classNames(checkboxClassName)}
-          id={id}
+          checked={checked}
           name="checkbox"
           onChange={onChange}
           type="checkbox" />
-        <label
-          checked={checked}
-          htmlFor={id}>
-          {children}
-        </label>
-      </span>
+        <span>{children}</span>
+      </label>
     );
   }
 }
 
 ToggleButton.defaultProps = {
   checked: false,
-  id: Date.now(),
   onChange: function () {},
-
   checkboxClassName: 'toggle-button'
 }
 
 ToggleButton.propTypes = {
   checked: React.PropTypes.bool,
   children: React.PropTypes.node,
-  id: React.PropTypes.oneOfType([
-    React.PropTypes.string,
-    React.PropTypes.number
-  ]),
   onChange: React.PropTypes.func,
 
   checkboxClassName: React.PropTypes.oneOfType([
--- a/src/js/components/ToggleButton.js
+++ b/src/js/components/ToggleButton.js
@@ -1,0 +1,62 @@
+import classNames from 'classnames';
+import React from 'react';
+
+class ToggleButton extends React.Component {
+  render() {
+    let {
+      checkboxClassName,
+      checked,
+      children,
+      className,
+      id,
+      onChange
+    } = this.props;
+
+    return (
+      <span className={classNames(className)}>
+        <input
+          className={classNames(checkboxClassName)}
+          id={id}
+          name="checkbox"
+          onChange={onChange}
+          type="checkbox" />
+        <label
+          checked={checked}
+          htmlFor={id}>
+          {children}
+        </label>
+      </span>
+    );
+  }
+}
+
+ToggleButton.defaultProps = {
+  checked: false,
+  id: Date.now(),
+  onChange: function () {},
+
+  checkboxClassName: 'toggle-button'
+}
+
+ToggleButton.propTypes = {
+  checked: React.PropTypes.bool,
+  children: React.PropTypes.node,
+  id: React.PropTypes.oneOfType([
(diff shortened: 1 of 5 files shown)


mesosphere-web:
Reviewer: @kennyt

Poltergeist:
This is cool. Thanks for going the extra mile. 👍

src/styles/components/toggle-button.less
```diff
@@ -0,0 +1,48 @@
+.toggle-button {
+  display: none;
+
+
+  + label {
```
jfurrow: Can we get an `&` here? `& + label {`

src/styles/components/toggle-button.less
```diff
@@ -0,0 +1,48 @@
...
+  &:checked + label {
+    &:before {
+      background-color: @toggle-checked-color;
+      border-color: @toggle-checked-color;
+    }
+    &:after {
+      transform: translateX(0);
+    }
+  }
+}
```
jfurrow: Let's :abcd: these properties

src/styles/components/toggle-button.less
```diff
@@ -0,0 +1,48 @@
+.toggle-button {
+  display: none;
+
+
```
jfurrow: Extra newline here :)

src/styles/components/toggle-button.less
```diff
@@ -0,0 +1,48 @@
...
+    &:before {
+      position: absolute;
+      top: 0;
+      right: 0;
+      display: block;
+      width: @toggle-width;
+      height: @toggle-height;
+      border-radius: @toggle-height;
+      border: rgba(0, 0, 0, 0.2) @toggle-border-width solid;
+      content: " ";
```
jfurrow: #nit elsewhere we use `content: '';`, can we use that here also? (single quotes without a space)

src/styles/components/toggle-button.less
```diff
@@ -0,0 +1,48 @@
...
+
+    &:before {
+      position: absolute;
+      top: 0;
+      right: 0;
+      display: block;
+      width: @toggle-width;
+      height: @toggle-height;
+      border-radius: @toggle-height;
+      border: rgba(0, 0, 0, 0.2) @toggle-border-width solid;
```
jfurrow: #nit Can we order these shorthand values as `{width} {style} {color}`? That's how we do it elsewhere, and I would like for it to be consistent, even though it doesn't matter.

src/js/components/ToggleButton.js
```diff
@@ -0,0 +1,62 @@
...
+
+  checkboxClassName: React.PropTypes.oneOfType([
+    React.PropTypes.string,
+    React.PropTypes.number,
+    React.PropTypes.array
+  ]),
+  className: React.PropTypes.oneOfType([
+    React.PropTypes.string,
+    React.PropTypes.number,
+    React.PropTypes.array
```
jfurrow: We should probably add `React.PropTypes.object` for each of these class props since we're passing them into the `classnames` module anyway.

src/js/components/ToggleButton.js
```diff
@@ -0,0 +1,62 @@
...
+          id={id}
+          name="checkbox"
+          onChange={onChange}
+          type="checkbox" />
+        <label
+          checked={checked}
+          htmlFor={id}>
+          {children}
+        </label>
+      </span>
```
jfurrow: You could wrap the entire component in a `label` element, then use a `span` in place of what is currently a `label`. This way you don't need to use the `id` and `htmlFor` attributes as the browser will handle that behavior without the explicit `for` attribute. Though, I'm not sure how that might affect accessibility...

jfurrow:
I added some #nitpick comments, but it looks great & works really nicely.

src/styles/components/toggle-button.less
```diff
@@ -0,0 +1,48 @@
...
+      transform: translateX(@toggle-height - @toggle-width);
+    }
+  }
+
+  &:checked + label {
+    &:before {
+      background-color: @toggle-checked-color;
+      border-color: @toggle-checked-color;
+    }
+    &:after {
```
jfurrow: Let's get some newlines above these pseudo-element selectors.

src/js/components/ToggleButton.js
mlunoe: That `number` was supposed to be `object`.
Done

src/styles/components/toggle-button.less
mlunoe: done

src/styles/components/toggle-button.less
mlunoe: done

src/styles/components/toggle-button.less
mlunoe: done

src/styles/components/toggle-button.less
mlunoe: done

src/styles/components/toggle-button.less
mlunoe: done

src/styles/components/toggle-button.less
mlunoe: done

jfurrow:
Lovely!

PR #206, opened 2016-05-18
Fix/use page header consistently

pierlo-upitup:
This makes sure we have the same markup and same behaviour across Task and Service detail pages.

From:
from

To:
to


mesosphere-web:
Reviewer: @MatApple

PR #207, opened 2016-05-18
Establish Jobs page and routes

pierlo-upitup:
All this PR accomplishes is to establish the Jobs overview page and routes. Consider it as the first step towards having the various Jobs views.

screen shot 2016-05-18 at 14 04 25

### DIFF
--- a/src/js/pages/JobsPage.js
+++ b/src/js/pages/JobsPage.js
@@ -1,0 +1,66 @@
+import React from 'react';
+import {RouteHandler} from 'react-router';
+
+import Page from '../components/Page';
+import TabsMixin from '../mixins/TabsMixin';
+
+var JobsPage = React.createClass({
+
+  contextTypes: {
+    router: React.PropTypes.func
+  },
+
+  mixins: [TabsMixin],
+
+  displayName: 'JobsPage',
+
+  statics: {
+    routeConfig: {
+      label: 'Jobs',
+      icon: 'jobs',
+      matches: /^\/jobs/
+    }
+  },
+
+  getInitialState: function () {
+    return {
+      currentTab: 'jobs-page'
+    };
+  },
+
+  componentWillMount: function () {
+    this.tabs_tabs = {
+      'jobs-page': 'Jobs'
+    };
+    this.updateCurrentTab();
+  },
+
+  updateCurrentTab: function () {
+    let routes = this.context.router.getCurrentRoutes();
+    let currentTab = routes[routes.length - 1].name;
+    if (currentTab != null) {
+      this.setState({currentTab});
+    }
+  },
+
+  getNavigation: function () {
+    return (
+      <ul className="tabs list-inline flush-bottom inverse">
+        {this.tabs_getRoutedTabs()}
+      </ul>
+    );
+  },
+
+  render: function () {
+    return (
+      <Page
+        navigation={this.getNavigation()}
+        title="Jobs">
+        <RouteHandler />
+      </Page>
+    );
+  }
+
+});
+
+module.exports = JobsPage;
--- a/src/js/pages/JobsPage.js
+++ b/src/js/pages/JobsPage.js
@@ -1,57 +1,44 @@
+import mixin from 'reactjs-mixin';
 import React from 'react';
 import {RouteHandler} from 'react-router';
 
 import Page from '../components/Page';
+import SidebarActions from '../events/SidebarActions';
 import TabsMixin from '../mixins/TabsMixin';
 
-var JobsPage = React.createClass({
+class JobsPage extends mixin(TabsMixin) {
+  constructor() {
+    super();
 
-  contextTypes: {
-    router: React.PropTypes.func
-  },
+    this.tabs_tabs = { 'jobs-page': 'Jobs'};
+    this.state = {currentTab: 'jobs-page'};
+  }
 
-  mixins: [TabsMixin],
-
-  displayName: 'JobsPage',
-
-  statics: {
-    routeConfig: {
-      label: 'Jobs',
-      icon: 'jobs',
-      matches: /^\/jobs/
-    }
-  },
-
-  getInitialState: function () {
-    return {
-      currentTab: 'jobs-page'
-    };
-  },
-
-  componentWillMount: function () {
-    this.tabs_tabs = {
-      'jobs-page': 'Jobs'
-    };
+  componentWillMount() {
     this.updateCurrentTab();
-  },
+  }
 
-  updateCurrentTab: function () {
+  componentWillReceiveProps() {
+    this.updateCurrentTab();
+  }
+
+  updateCurrentTab() {
     let routes = this.context.router.getCurrentRoutes();
     let currentTab = routes[routes.length - 1].name;
     if (currentTab != null) {
       this.setState({currentTab});
     }
-  },
+  }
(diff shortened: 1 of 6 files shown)


mesosphere-web:
Reviewer: @mlunoe

src/js/pages/JobsPage.js
```diff
@@ -0,0 +1,66 @@
+import React from 'react';
+import {RouteHandler} from 'react-router';
+
+import Page from '../components/Page';
+import TabsMixin from '../mixins/TabsMixin';
+
+var JobsPage = React.createClass({
```
philipnrmn: Sorry to do this, but we have a directive that new components should be classes extending `React.Component` (ref https://github.com/dcos/dcos-ui/pull/70#discussion-diff-61490798). Let's update this one.

src/js/pages/jobs/JobsTab.js
```diff
@@ -0,0 +1,76 @@
+import React from 'react';
+import {RouteHandler} from 'react-router';
+import {StoreMixin} from 'mesosphere-shared-reactjs';
+
+import AlertPanel from '../../components/AlertPanel';
+import DCOSStore from '../../stores/DCOSStore';
+import QueryParamsMixin from '../../mixins/QueryParamsMixin';
+import SidebarActions from '../../events/SidebarActions';
+
+var JobsTab = React.createClass({
```
philipnrmn: Let's update this to an ES6 class as along with `JobsPage`.

src/js/pages/JobsPage.js
pierlo-upitup: Damnit, I always pick the bad examples for my copypasta!

philipnrmn: It's so frustrating to run across these undocumented conventions >.<

philipnrmn:
It'd be good to have an integration test to check that the page loads - not necessarily in this PR though, up to you.

pierlo-upitup:
Added a couple of very basic integration tests for now. 

screen shot 2016-05-18 at 14 57 30

src/js/pages/jobs/JobsTab.js
```diff
@@ -0,0 +1,65 @@
...
+import {RouteHandler} from 'react-router';
+import {StoreMixin} from 'mesosphere-shared-reactjs';
+
+import AlertPanel from '../../components/AlertPanel';
+import DCOSStore from '../../stores/DCOSStore';
+
+class JobsTab extends mixin(StoreMixin) {
+
+  constructor() {
+    super();
```
philipnrmn: Not sure that we actually need to call `super()`. If it's defensive, let's call it with `...arguments`.

src/js/pages/jobs/JobsTab.js
```diff
@@ -0,0 +1,65 @@
...
+
+    this.store_listeners = [
+      {name: 'dcos', events: ['change']}
+    ];
+  }
+
+  /* eslint-disable no-unused-vars */
+  getContents(id) {
+  /* eslint-enable-no-unused-vars */
```
mlunoe: Let's just add the `id` when we need it and remove these for now

src/js/pages/jobs/JobsTab.js
```diff
@@ -0,0 +1,65 @@
...
+      return (
+        <RouteHandler />
+      );
+    }
+
+    // Render empty panel
+    return (
+      <AlertPanel
+        title="No Jobs Found"
+        iconClassName="icon icon-sprite icon-sprite-jumbo
```
mlunoe: Strings (including classNames) and regex' are ok to take up more than 80. That is why it is not a hard rule

pierlo-upitup: Mmm, OK. But I've seen `className`s newlining like very often – you OK with me leaving it this way?

kennyt: i've also do classnames on newlines and i haven't ever heard of an agreement on whether to do it or not.

pierlo, sorry for the nit hour but if you do it this way, can you have the 2nd line indented the same level as the child `<p>`?

src/js/pages/JobsPage.js
```diff
@@ -0,0 +1,66 @@
...
+
+import Page from '../components/Page';
+import SidebarActions from '../events/SidebarActions';
+import TabsMixin from '../mixins/TabsMixin';
+
+class JobsPage extends mixin(TabsMixin) {
+  constructor() {
+    super();
+
+    this.tabs_tabs = { 'jobs-page': 'Jobs'};
```
kennyt: can we remove the space between the `{` and `'`?

kennyt:
i'd merge after the nits

src/js/pages/jobs/JobsTab.js
mlunoe: We have this in other places, but I would like us to be more consistent. This is, however, not a show stopper for the PR.

src/js/pages/JobsPage.js
```diff
@@ -0,0 +1,66 @@
...
+import React from 'react';
+import {RouteHandler} from 'react-router';
+
+import Page from '../components/Page';
+import SidebarActions from '../events/SidebarActions';
+import TabsMixin from '../mixins/TabsMixin';
+
+class JobsPage extends mixin(TabsMixin) {
+  constructor() {
+    super();
```
rcorral: can you do `super(...arguments)` instead?

src/js/pages/JobsPage.js
```diff
@@ -0,0 +1,66 @@
...
+    this.tabs_tabs = { 'jobs-page': 'Jobs'};
+    this.state = {currentTab: 'jobs-page'};
+  }
+
+  componentWillMount() {
+    this.updateCurrentTab();
+  }
+
+  componentWillReceiveProps() {
+    this.updateCurrentTab();
```
rcorral: Can you add `super. componentWillReceiveProps(...arguments)`

src/js/pages/jobs/JobsTab.js
```diff
@@ -0,0 +1,63 @@
...
+import {RouteHandler} from 'react-router';
+import {StoreMixin} from 'mesosphere-shared-reactjs';
+
+import AlertPanel from '../../components/AlertPanel';
+import DCOSStore from '../../stores/DCOSStore';
+
+class JobsTab extends mixin(StoreMixin) {
+
+  constructor() {
+    super();
```
rcorral: Same here please!

src/js/pages/JobsPage.js
pierlo-upitup: Can you explain why? I've seen inconsistent usage across the codebase and in this case I'd argue that TabsMixin has no constructor per se so I don't see the need for `...arguments` here?

rcorral: Because it's really easy to miss in the future. This component could evolve to having other mixins and it's really nice to not have to think about which methods the mixin uses from react's lifecycle if we just call super with arguments.

orlandohohmeier: @rcorral Following that argument, we should always call the `super` whenever we override a method, as we otherwise miss something, if one of the classes in the chain evolves.
Let's please add this topic to our agenda and discuss it, as I see value in doing so, but on the other hand it kinda seems like overhead.

PR #211, opened 2016-05-18
Add deploy service methods to marathon store

Poltergeist:
Introduction of the createService action which posts an app definition to marathon. And notifies the store of a successful or unsuccessful POST request.

The store emits in both cases an event which will be picked up by a later component and then either closes a modal or displays the error.

This is a dependency for #212

PR #210, opened 2016-05-18
Add HeaderText to AdvancedConfig

Poltergeist:
This introduces the capability of adding a header to the `advancedConfig` component. This is needed to show a generic title instead of the package name and version.

This is a dependency for #212

PR #208, opened 2016-05-18
Extend service struct

Poltergeist:
Extend the service struct with a couple of get functions, those are needed for the deploy a service modal.

Also introducing a `ServiceUtil.js` this contains for now 3 functions:
- `convertFormModelToService` - This function is creating a `Service` object from the a `formModel` which is provided by the `AdvancedConfig` component.
- `createFormModelFromSchema` - This is creating a `formModel` with a given `Schema` and `Service`.
- `getAppDefinitionFromService` - This is creating a marathon `appDefinition` from a `Service` Object. 

This is a dependency for #212

PR #212, opened 2016-05-18
Add deploy service modal

Poltergeist:
---
This PR does include the following functionality:
 - Create Modal
 - JSON editor
 - Parsing data back and forth between those two.
 - The JSON editor is submitting the JSON you supply which means any app definition can be POSTed to the server
 - The form only does supply the given fields.
 - Switching to the form will cut the definition to the fields which are provided.
 - Some validation.
 - Refactoring of the SchemaForm
   - Remove the Header from the component and moving it to the modal
   - Clearing the timeout if component does unmount

This PR does **not** include:
 - A final design
 - A full fledged form
 - Complete Validation
 - Dependent fields
 - Multiple rows
 - Right error Handling
 - Update the `AlertPanel` inside empty groups to reveal the Deploy Modal 

---

IMAGES: 

---

image

image

image

This PR has a couple of dependencies:
#208 (~~not~~ **merged** ~~yet~~)
#209 (~~not~~ **merged** ~~yet~~)
#210 (~~not~~ **merged** ~~yet~~)
#211 (~~not~~ **merged** ~~yet~~)


mesosphere-web:
Reviewer: @jfurrow

pierlo-upitup:
I've force-pushed some changes to fix rebase issues against current master.

src/js/pages/services/ServicesTab.js
```diff
@@ -53,7 +54,8 @@ var ServicesTab = React.createClass({
 
   getInitialState: function () {
     return Object.assign({}, DEFAULT_FILTER_OPTIONS, {
-      isServiceGroupFormModalShown: false
+      isServiceGroupFormModalShown: false,
+      isDeployServiceModalShown: false
```
pierlo-upitup: I'd like more consistency in naming here. Given we already have `ServiceGroupFormModal.js`, I'd suggest naming our new modal `ServiceFormModal.js`. As we'll likely re-use this also for editing a Service, I would not have the `Deploy` in its name.

src/js/pages/services/ServicesTab.js
```diff
@@ -72,10 +74,18 @@ var ServicesTab = React.createClass({
...
+  },
+
   handleCloseGroupFormModal: function () {
     this.setState({isServiceGroupFormModalShown: false});
   },
 
+  handleOpenDeployServiceModal: function () {
+    this.setState({isDeployServiceModalShown: true});
+  },
+
```
pierlo-upitup: This starts to look like a 🍝 . How about we refactor these 4 methods into 1 that accepts a string and a boolean? In any case it makes sense to also toggle the "other" modal when showing this one, as we can't hold 2 modals at once.

Something like 
```
handleOpenModal: function(modalName, show = true) {
   // ...
   if (modalName === "Deploy Group Name Here") {
     this.setState({
       isServiceModalShown: !show, 
       isServiceGroupModalShown: show
    });
 }
```

src/js/pages/services/ServicesTab.js
```diff
@@ -5,6 +5,7 @@ import {StoreMixin} from 'mesosphere-shared-reactjs';
 import AlertPanel from '../../components/AlertPanel';
 import DCOSStore from '../../stores/DCOSStore';
 import FilterBar from '../../components/FilterBar';
+import DeployServiceModal from '../../components/modals/DeployServiceModal';
```
pierlo-upitup: see below for naming

src/js/pages/services/ServicesTab.js
```diff
@@ -213,7 +223,8 @@ var ServicesTab = React.createClass({
                     onClick={this.handleOpenGroupFormModal}>
               Create Group
             </button>
-            <button className="button button-success">
+            <button className="button button-success"
+              onClick={this.handleOpenDeployServiceModal}>
```
pierlo-upitup: Let's not forget to add similar behaviour to our AlertPanel

pierlo-upitup:
This is starting to look great!

src/js/components/modals/DeployServiceModal.js
```diff
@@ -0,0 +1,206 @@
...
+import Service from '../../structs/Service';
+import ServiceUtil from '../../utils/ServiceUtil';
+import ServiceSchema from '../../constants/ServiceSchema';
+
+const METHODS_TO_BIND = [
+  'getAdvancedSubmit',
+  'onMarathonStoreServiceCreateSuccess',
+  'onMarathonStoreServiceCreateError',
+  'handleChange',
+  'handleSubmit'
```
pierlo-upitup: 🔤 ?

src/js/components/modals/DeployServiceModal.js
```diff
@@ -0,0 +1,206 @@
...
+  constructor() {
+    super(...arguments);
+
+    this.internalStorage_set({
+      hasFormErrors: false,
+      pendingRequest: false
+    });
+
+    let model =
+      ServiceUtil.createFormModelFromSchema(new Service(), ServiceSchema);
```
pierlo-upitup: This should be able to also accept a `service` as a prop, which if not set is `new Service()` by default.

src/js/components/modals/DeployServiceModal.js
```diff
@@ -0,0 +1,206 @@
...
+    this.state = {
+      service: ServiceUtil.createServiceFromFormModel(model),
+      model: model,
+      errorMessage: null,
+      schema: ServiceSchema
```
pierlo-upitup: Seems like we don't need to keep state for `this.state.schema` ?

src/js/components/modals/DeployServiceModal.js
```diff
@@ -0,0 +1,206 @@
...
+
+  }
+
+  getModalContents() {
+    return (
+      <div>
+        <AdvancedConfig
+          headerText="Deploy New Service"
+          schema={this.state.schema}
+          mode={this.state.model}
```
kennyt: i believe this is supposed to be `model={...}` :)

src/js/components/modals/DeployServiceModal.js
```diff
@@ -0,0 +1,206 @@
...
+        events: ['serviceCreateError', 'serviceCreateSuccess']
+      }
+    ];
+
+    METHODS_TO_BIND.forEach((method) => {
+      this[method] = this[method].bind(this);
+    });
+  }
+
+  componentDidMount() {
```
kennyt: this isn't necessary unless you're going to put something here in the future

src/js/components/modals/DeployServiceModal.js
Poltergeist: I do like the idea of the default

src/js/pages/services/ServicesTab.js
```diff
@@ -242,9 +253,11 @@ var ServicesTab = React.createClass({
           open={state.isServiceGroupFormModalShown}
           parentGroupId={item.getId()}
           onClose={this.handleCloseGroupFormModal}/>
+        <DeployServiceModal open={state.isDeployServiceModalShown}
+          onClose={this.handleCloseDeployServiceModal}/>
       </div>
-    )
-  }
+    );
+  },
```
Poltergeist: This `,` is too much.

src/js/components/modals/ServiceFormModal.js
```diff
@@ -0,0 +1,244 @@
...
+  }
+
+  handleCancel() {
+    this.props.onClose();
+  }
+
+  handleSubmit() {
+    if (this.state.json) {
+      MarathonStore.createService(JSON.parse(this.state.app));
+      return ;
```
Poltergeist: remove the whitespace before the `;`

src/js/components/SchemaForm.js
```diff
@@ -78,9 +84,13 @@ class SchemaForm extends React.Component {
       return;
     }
 
-    setTimeout(() => {
+    if (this.timer) {
+      clearTimeout(this.timer);
```
kennyt: is this still needed?

src/js/components/modals/ServiceFormModal.js
```diff
@@ -0,0 +1,244 @@
...
+    let model = ServiceUtil.createFormModelFromSchema(ServiceSchema);
+    this.setState({
+      service: ServiceUtil.createServiceFromFormModel(model),
+      model: model,
+      app: '',
+      errorMessage: null
+    });
+  }
+
+  handleChange({isValidated, model}) {
```
kennyt: we can remove this method now (and also inside of methods_to_bind)

src/js/components/modals/ServiceFormModal.js
```diff
@@ -0,0 +1,244 @@
...
+class ServiceFormModal extends mixin(StoreMixin) {
+  constructor() {
+    super(...arguments);
+
+    let model =
+      ServiceUtil.createFormModelFromSchema(ServiceSchema);
+    this.state = {
+      service: ServiceUtil.createServiceFromFormModel(model),
+      model: model,
+      json: false,
```
kennyt: can we call this `jsonMode` or something like that?

src/js/components/modals/ServiceFormModal.js
```diff
@@ -0,0 +1,244 @@
...
+  getAdvancedSubmit(triggerSubmit) {
+    this.triggerAdvancedSubmit = triggerSubmit;
+  }
+
+  getErrorMessage() {
+    let {errorMessage} = this.state;
+    if (!errorMessage) {
+      return null;
+    }
+    return (
```
kennyt: can we put a newline before this?

src/js/components/modals/ServiceFormModal.js
```diff
@@ -0,0 +1,244 @@
...
+          showGutter={true}
+          showPrintMargin={false}
+          theme="monokai"
+          value={this.state.app}
+          width="100%"/>
+      );
+    }
+
+    return (
+      <AdvancedConfig
```
mlunoe: I think we can use `SchemaForm` here directly.
Actually looking at `AdvancedConfig` now, it is only a wrapper for `SchemaForm`. I think we should try to remove it all together, but not in this PR.

src/js/components/modals/ServiceFormModal.js
```diff
@@ -0,0 +1,244 @@
...
+      let model = this.triggerAdvancedSubmit().model;
+      let service = ServiceUtil.createServiceFromFormModel(model)
+      this.setState({
+        app: JSON.stringify(ServiceUtil
+          .getAppDefinitionFromService(service), null, 2),
+        model,
+        service
+      });
+    }
+    this.setState({json: !this.state.json});
```
kennyt: if `!this.state.json` is true, then we'd be setting state twice in a row. i think react might be smart enough to do it in one go, but could we just create an object called `nextState` and then have one `setState` at the end?

src/js/components/modals/ServiceFormModal.js
```diff
@@ -0,0 +1,244 @@
+import mixin from 'reactjs-mixin';
+import {Modal} from 'reactjs-components';
+import React from 'react';
+import {StoreMixin} from 'mesosphere-shared-reactjs';
+import Ace from 'react-ace';
```
kennyt: :abc:

src/js/components/modals/ServiceFormModal.js
```diff
@@ -0,0 +1,244 @@
...
+          width="100%"/>
+      );
+    }
+
+    return (
+      <AdvancedConfig
+        model={
+          ServiceUtil.createFormModelFromSchema(
+            ServiceSchema,
+            this.state.service
```
mlunoe: if you do `let {json, service} = this.state;` in the top you should be able to have `model={ServiceUtil.createFormModelFromSchema(ServiceSchema, service)` in one line here

src/js/components/modals/ServiceFormModal.js
mlunoe: Could we rename this to something more meaningful? e.g. `jsonMode`?

src/js/components/modals/ServiceFormModal.js
```diff
@@ -0,0 +1,244 @@
...
+
+class ServiceFormModal extends mixin(StoreMixin) {
+  constructor() {
+    super(...arguments);
+
+    let model =
+      ServiceUtil.createFormModelFromSchema(ServiceSchema);
+    this.state = {
+      service: ServiceUtil.createServiceFromFormModel(model),
+      model: model,
```
mlunoe: You can do just `model,` here

src/js/pages/services/ServicesTab.js
```diff
@@ -5,9 +5,14 @@ import {StoreMixin} from 'mesosphere-shared-reactjs';
 import AlertPanel from '../../components/AlertPanel';
 import DCOSStore from '../../stores/DCOSStore';
 import FilterBar from '../../components/FilterBar';
+import ServiceFormModal from '../../components/modals/ServiceFormModal';
```
kennyt: :abc:

src/js/components/modals/ServiceFormModal.js
```diff
@@ -0,0 +1,244 @@
...
+    ];
+
+    METHODS_TO_BIND.forEach((method) => {
+      this[method] = this[method].bind(this);
+    });
+  }
+
+  componentWillReceiveProps(nextProps) {
+    super.componentWillReceiveProps(...arguments);
+    let {props} = this;
```
mlunoe: No need to declare this variable. Just access `this.props.open` directly

src/js/components/modals/ServiceFormModal.js
```diff
@@ -0,0 +1,244 @@
...
+    if (!props.open && nextProps.open) {
+      this.resetState();
+    }
+  }
+
+  resetState() {
+    let model = ServiceUtil.createFormModelFromSchema(ServiceSchema);
+    this.setState({
+      service: ServiceUtil.createServiceFromFormModel(model),
+      model: model,
```
mlunoe: just `model,` will do

src/js/components/modals/ServiceFormModal.js
```diff
@@ -0,0 +1,244 @@
...
+      app: '',
+      errorMessage: null
+    });
+  }
+
+  handleChange({isValidated, model}) {
+    if (isValidated) {
+      this.setState({
+        service: ServiceUtil.createServiceFromFormModel(model),
+        model: model,
```
mlunoe: `model,`

src/js/components/modals/ServiceFormModal.js
```diff
@@ -0,0 +1,244 @@
...
+      service = new Service(JSON.parse(app));
+    } catch (e) {
+
+    }
+    this.setState({app, service})
+  }
+
+  handleJSONToggle() {
+    if (!this.state.json) {
+      let model = this.triggerAdvancedSubmit().model;
```
mlunoe: `let {model} = this.triggerAdvancedSubmit()`

src/js/components/modals/ServiceFormModal.js
```diff
@@ -0,0 +1,244 @@
...
+    }
+    return (
+      <div>
+        <h4 className="text-align-center text-danger flush-top">
+          {errorMessage.message}
+        </h4>
+        <pre className="text-danger">
+          {JSON.stringify(errorMessage.details, null, 2)}
+        </pre>
+        <button onClick={this.handleClearError}
```
mlunoe: :abc: and move `onClick` down in same level as `className`

src/js/pages/services/ServicesTab.js
```diff
@@ -210,10 +225,12 @@ var ServicesTab = React.createClass({
             <ServiceSearchFilter
               handleFilterChange={this.handleFilterChange} />
             <button className="button button-stroke button-inverse"
-                    onClick={this.handleOpenGroupFormModal}>
+                    onClick={() =>
```
kennyt: this should be indented just one level deeper than `button`

src/js/events/MarathonActions.js
```diff
@@ -54,7 +54,7 @@ module.exports = {
       error: function (xhr) {
         AppDispatcher.handleServerAction({
           type: REQUEST_MARATHON_SERVICE_CREATE_ERROR,
-          data: RequestUtil.getErrorFromXHR(xhr),
+          data: JSON.parse(xhr.response),
```
mlunoe: We have `RequestUtil. parseResponseBody` I believe you could use here: https://github.com/dcos/mesosphere-shared-reactjs/blob/master/utils/RequestUtil.js#L112

src/js/components/SchemaForm.js
mlunoe: We need comment here, at least

src/js/components/modals/ServiceFormModal.js
Poltergeist: Was changed due to another comment to 
```JS
nextState.model = this.triggerAdvancedSubmit().model;
```

src/js/components/modals/ServiceFormModal.js
```diff
@@ -0,0 +1,244 @@
...
+    return (
+      <AdvancedConfig
+        model={
+          ServiceUtil.createFormModelFromSchema(
+            ServiceSchema,
+            this.state.service
+          )
+        }
+        schema={ServiceSchema}
+        getTriggerSubmit={this.getAdvancedSubmit}/>
```
Poltergeist: 🔤

src/js/components/SchemaForm.js
Poltergeist: This is debouncing the validation and making sure that it is only run once.

src/js/components/modals/ServiceFormModal.js
Poltergeist: Removed it. 💥

src/js/components/modals/ServiceFormModal.js
Poltergeist: 👍 did that.

src/js/components/modals/ServiceFormModal.js
Poltergeist: 👍 done.

src/js/components/modals/ServiceFormModal.js
Poltergeist: Worked thanks for the comment.

src/js/components/modals/ServiceFormModal.js
Poltergeist: Done that. 👍

src/js/components/modals/ServiceFormModal.js
Poltergeist: ➡️

src/js/components/modals/ServiceFormModal.js
Poltergeist: 👍 done.

src/js/components/modals/ServiceFormModal.js
Poltergeist: Thanks did that all over this file. 👍

src/js/components/modals/ServiceFormModal.js
Poltergeist: done

src/js/components/modals/ServiceFormModal.js
Poltergeist: done

src/js/pages/services/ServicesTab.js
Poltergeist: done 👍

src/js/events/MarathonActions.js
Poltergeist: 👍 Tests pass

src/js/components/modals/ServiceFormModal.js
kennyt: don't think you pushed it up :)

Poltergeist: https://github.com/dcos/dcos-ui/pull/212/files#diff-076a6fbfc2ab2daac905f988b78234a2R88 yes I did.

kennyt:
wow, just tried this. this is pretty fucking awesome

src/js/components/SchemaForm.js
```diff
@@ -53,6 +53,13 @@ class SchemaForm extends React.Component {
     this.setState({renderGemini: true});
   }
 
+  componentWillUnmount() {
+    // Unschedule all validation if component unmounts.
+    if (this.timer) {
+      clearTimeout(this.timer);
```
kennyt: after all of the `clearTimeout`s, do we want to do `this.timer = null` or something?

src/js/components/modals/ServiceFormModal.js
```diff
@@ -0,0 +1,242 @@
...
+      errorMessage
+    });
+  }
+
+  handleCancel() {
+    this.props.onClose();
+  }
+
+  handleSubmit() {
+
```
kennyt: got an extra newline here

src/js/components/modals/ServiceFormModal.js
```diff
@@ -0,0 +1,242 @@
...
+          JSON mode
+        </ToggleButton>
+        <button
+          className="button button-large button-success flush-bottom"
+          onClick={this.handleSubmit}>
+          Deploy
+        </button>
+      </div>
+    );
+
```
kennyt: get an extra newline here

src/js/components/modals/ServiceFormModal.js
```diff
@@ -0,0 +1,242 @@
...
+          value={app}
+          width="100%"/>
+      );
+    }
+
+    return (
+      <SchemaForm
+        getTriggerSubmit={this.getAdvancedSubmit}
+        model={ServiceUtil.createFormModelFromSchema(ServiceSchema, service)}
+        onChange={Function.prototype}
```
kennyt: can you make an edit to `SchemaForm` to have a default function in this PR so you don't have to do this?

src/js/utils/ServiceUtil.js
```diff
@@ -8,26 +8,38 @@ const getFindPropertiesRecursive = function (service, item) {
...
+      definition.id = formModel.General.id;
+      definition.cmd = formModel.General.cmd;
+      definition.cpus = formModel.General.cpus;
+      definition.mem = formModel.General.mem;
+      definition.disk = formModel.General.disk;
+      definition.instances = formModel.General.instances;
+    }
+
+    if (formModel['Container Settings'] != null) {
+
```
kennyt: get an extra newline here

src/js/components/modals/ServiceFormModal.js
```diff
@@ -0,0 +1,242 @@
+import Ace from 'react-ace';
+import mixin from 'reactjs-mixin';
+import {Modal} from 'reactjs-components';
+import React from 'react';
+import {StoreMixin} from 'mesosphere-shared-reactjs';
+
+import 'brace/mode/json';
+import 'brace/theme/monokai';
+import 'brace/ext/language_tools';
```
kennyt: should we do this here or import them in `index.less`? what's the difference?

kennyt:
one thing that happens: if you're on the json mode and close it and then reopen it, the animation is broken and also you lose the model completely.

gf

kennyt:
after my last round of nits and the closing when in json mode issue, i think this looks good. sorry for the waterfall review

src/js/components/SchemaForm.js
Poltergeist: removed it.

src/js/components/modals/ServiceFormModal.js
Poltergeist: Those are actually javascript files. So we should keep them here.
```sh
beautify.js               linking.js                static_highlight.js
chromevox.js              modelist.js               statusbar.js
elastic_tabstops_lite.js  old_ie.js                 textarea.js
emmet.js                  searchbox.js              themelist.js
error_marker.js           settings_menu.js          whitespace.js
keybinding_menu.js        spellcheck.js
language_tools.js         split.js
```

kennyt: of course, sorry about that

Poltergeist:
I adressed those nits but I couldn't address the animation issue. Losing the date by canceling the dialog is intentional.

src/js/components/modals/ServiceFormModal.js
Poltergeist: No problem ;-)

kennyt:
but that's not the custom data. that's the default data that comes from the schema. should it be there like it is for the form mode?

src/js/components/SchemaForm.js
jfurrow: `this.timer` will always hold the ID of the `setTimeout`, is that intentional?

Poltergeist:
You are right @kennyt, the reset was causing this problem. I fixed the reset so that upon opening the modal the user will see the form pre-filled with the default data.

src/js/components/SchemaForm.js
Poltergeist: This is intentional since we will need the id for the `cancelTimeout` if the component unmounts before the timeout has been triggered.

jfurrow: That makes sense, thanks!

pierlo-upitup:
I took the liberty to force-push after rebasing on master

pierlo-upitup:
I don't see the json toggle ... ? 
image

pierlo-upitup:
I still see some open comments – can you please address them so we can merge? Thx!

src/js/components/modals/DeployServiceModal.js
```diff
@@ -0,0 +1,206 @@
+import mixin from 'reactjs-mixin';
+import {Modal} from 'reactjs-components';
+import React from 'react';
+import {StoreMixin} from 'mesosphere-shared-reactjs';
+
+import AdvancedConfig from '../AdvancedConfig';
```
pierlo-upitup: This guy is about to be suppressed! https://github.com/dcos/dcos-ui/pull/261

Poltergeist:
It seems like you force pushed a very old state. I'll fix it with my state.

pierlo-upitup:
Oh jeez! That's really weird, I'm absolutely sure I pulled the latest from the branch... ! 😱 Hope it's not too bad!!! Sorry! 😢

Poltergeist:
No problem I rebased and force pushed now everything should be there 👍

src/js/components/SchemaForm.js
Poltergeist: The util debounce only debounces a function timeout but it is not possible to just cancel a timeout if the function ain't needed anymore e.g. on `unmount`.

src/js/components/SchemaForm.js
```diff
@@ -181,9 +193,13 @@ class SchemaForm extends React.Component {
     );
   }
 
-  getServiceHeader() {
+  getFormHeader() {
```
pierlo-upitup: Although to be fair this ends up being the modal header, right>

Poltergeist: This is the header of the advanced package install modal. And that header ain't needed in the deploy service modal.

src/js/components/modals/ServiceFormModal.js
```diff
@@ -0,0 +1,240 @@
...
+];
+
+class ServiceFormModal extends mixin(StoreMixin) {
+  constructor() {
+    super(...arguments);
+
+    let model =
+      ServiceUtil.createFormModelFromSchema(ServiceSchema);
+    this.state = {
+      app: JSON.stringify({id:'', cmd:''}, null, 2),
```
pierlo-upitup: Now that we call everything `service`, `app` is just confusing. Let's name this one `jsonDefinition` or similar, as its sole purpose is to hold the JSON mode blob.

src/js/components/modals/ServiceFormModal.js
```diff
@@ -0,0 +1,240 @@
...
+    super.componentWillReceiveProps(...arguments);
+    if (!this.props.open && nextProps.open) {
+      this.resetState();
+    }
+  }
+
+  resetState() {
+    let model = ServiceUtil.createFormModelFromSchema(ServiceSchema);
+    this.setState({
+      app: '',
```
pierlo-upitup: why is this not the same default as in the constructor?

src/js/components/modals/ServiceFormModal.js
```diff
@@ -0,0 +1,240 @@
...
+  handleClearError() {
+    this.setState({
+      errorMessage: null
+    });
+  }
+
+  handleJSONChange(app) {
+    let {service} = this.state;
+    try {
+      service = new Service(JSON.parse(app));
```
pierlo-upitup: This is mutating state directly. Please `let service = Object.assign({}, this.state.service)`.

src/js/components/modals/ServiceFormModal.js
```diff
@@ -0,0 +1,240 @@
...
+    } catch (e) {
+
+    }
```
pierlo-upitup: Not a huge fan of silencers but feel free to ignore me.

src/js/components/modals/ServiceFormModal.js
```diff
@@ -0,0 +1,240 @@
...
+    this.setState({app, service})
+  }
+
+  handleJSONToggle() {
+    let nextState = {};
+    if (!this.state.jsonMode) {
+      let {model} = this.triggerAdvancedSubmit();
```
pierlo-upitup: `this.triggerAdvancedSubmit()` is assumed to be existing and defined. However it seems like this is only true if `getAdvancedSubmit` is called first. This seems very brittle and error-prone. 

Beyond hardening this call with some `isFunction` check, I wonder if we should even rely on this get/set mechanism at all.

It seems like we expect the child `SchemaForm` to pass back that function to us so we can store it for later usage ( see https://github.com/dcos/dcos-ui/pull/212/files/90f2cebeb466daabed171e6924ce92aec9cc5b52#diff-076a6fbfc2ab2daac905f988b78234a2R202 ). 
Furthermore we have a lot of references to `Advanced` properties/functions and that smells of tight coupling. 
Can we improve this situation somehow?

src/js/components/modals/ServiceFormModal.js
```diff
@@ -0,0 +1,240 @@
...
+      let service = ServiceUtil.createServiceFromFormModel(model);
+      nextState.model = model;
+      nextState.service = service;
+      nextState.app = JSON.stringify(ServiceUtil
+        .getAppDefinitionFromService(service), null, 2);
```
pierlo-upitup: We seem to be performing the same dance over and over (get the model, create the form structure, parse it back into JSON). How about extracting this into a reusable function that lets us specify the `direction` (ie: from model to form, or from form to model). Just an idea.

src/js/components/modals/ServiceFormModal.js
```diff
@@ -0,0 +1,240 @@
...
+  }
+
+  onMarathonStoreServiceCreateError(errorMessage) {
+    this.setState({
+      errorMessage
+    });
+  }
+
+  handleCancel() {
+    this.props.onClose();
```
pierlo-upitup: `props.onClose` is optional, let's either make it required or ensure we don't try to call `undefined`().

src/js/components/modals/ServiceFormModal.js
```diff
@@ -0,0 +1,240 @@
...
+  }
+
+  handleSubmit() {
+    if (this.state.jsonMode) {
+      MarathonStore.createService(JSON.parse(this.state.app));
+      return;
```
pierlo-upitup: Shouldn't we reset `errorMessage` also when submitting from JSON mode?

src/js/components/modals/ServiceFormModal.js
```diff
@@ -0,0 +1,240 @@
...
+    }
+    if (this.triggerAdvancedSubmit) {
+      let {model} = this.triggerAdvancedSubmit();
```
pierlo-upitup: The way this function works is pretty obscure to me, honestly.

src/js/components/modals/ServiceFormModal.js
```diff
@@ -0,0 +1,240 @@
...
+      let service = ServiceUtil.createServiceFromFormModel(model);
+      this.setState({
+        service: service,
+        model: model,
+        errorMessage: null
```
pierlo-upitup: we could also `this.setState({service, model, errorMessage: null});`

src/js/components/modals/ServiceFormModal.js
```diff
@@ -0,0 +1,240 @@
...
+      });
+      MarathonStore
+        .createService(ServiceUtil.getAppDefinitionFromService(service));
+    }
+  }
+
+  getAdvancedSubmit(triggerSubmit) {
+    this.triggerAdvancedSubmit = triggerSubmit;
+  }
```
pierlo-upitup: ❓ A getter with side-effects is not that awesome.

pierlo-upitup: Edit: consider this comment out of scope. This anti-pattern belongs to https://github.com/dcos/dcos-ui/blob/master/src/js/components/SchemaForm.js#L49

pierlo-upitup: Although we could at least rename our own `getAdvancedSubmit` to something that better represents what it does.

src/js/constants/ModalKeys.js
```diff
@@ -0,0 +1,6 @@
+let MODAL_KEYS = {
+  SERVICE_FORM_MODAL: 'SERVICE_FORM_MODAL',
+  SERVICE_GROUP_FORM_MODAL: 'SERVICE_GROUP_FORM_MODAL'
+};
+
+module.exports = MODAL_KEYS;
```
pierlo-upitup: Are these even used?

pierlo-upitup: How about calling these `SERVICE_MODAL_IDS`?

pierlo-upitup: Actually that's a bad idea, I can see how we can expand these with other modal ids in the future. nvm

src/js/constants/ServiceSchema.js
```diff
@@ -0,0 +1,91 @@
+/* eslint-disable no-unused-vars */
+import React from 'react';
+/* eslint-enable no-unused-vars */
+
+let SERVICE_SCHEMA = {
```
pierlo-upitup: Is this really a schema? It seems more like a SchemaForm configuration.
My point is: are we going to be able to use this same schema as a reference to perform JSON definition validation/transformations later on?

src/js/pages/services/ServicesTab.js
```diff
@@ -210,10 +226,11 @@ var ServicesTab = React.createClass({
             <ServiceSearchFilter
               handleFilterChange={this.handleFilterChange} />
             <button className="button button-stroke button-inverse"
-                    onClick={this.handleOpenGroupFormModal}>
+              onClick={() => this.handleOpenModal(SERVICE_GROUP_FORM_MODAL)}>
               Create Group
             </button>
-            <button className="button button-success">
+            <button className="button button-success"
+              onClick={() => this.handleOpenModal(SERVICE_FORM_MODAL)}>
```
pierlo-upitup: I hate to do this but I think I heard somewhere that we are supposed to use the METHODS_TO_BIND way of doing things consistently...

src/js/utils/ServiceUtil.js
```diff
@@ -8,26 +8,37 @@ const getFindPropertiesRecursive = function (service, item) {
...
   createServiceFromFormModel: function (formModel) {
-    let values = Object.keys(formModel).reduce(function (memo, section) {
-      Object.keys(formModel[section]).forEach(function (attribute) {
-        memo[attribute] = formModel[section][attribute];
-      });
+    let definition = {};
 
-      return memo;
-    }, {});
+    if (formModel.General != null) {
```
pierlo-upitup: maybe out of scope for this PR but I'd like us to enforce some checks on the received `formModel` so that we don't explode if it's not an object. 

We should really start considering `Flow` static typing...

src/js/utils/ServiceUtil.js
```diff
@@ -42,6 +53,14 @@ const ServiceUtil = {
     appDefinition.instances = service.getInstancesCount();
     appDefinition.cmd = service.getCommand();
 
+    let containerSettings = service.getContainerSettings();
+    if (
+      containerSettings && containerSettings.docker &&
+      containerSettings.docker.image
+    ) {
+      appDefinition.container = service.getContainerSettings();
```
pierlo-upitup: why call the function again?

src/js/utils/__tests__/ServiceUtil-test.js
```diff
@@ -5,7 +5,7 @@ var Service = require('../../structs/Service');
 var ServiceUtil = require('../ServiceUtil');
 
 describe('ServiceUtil', function () {
-  describe('#converModelToService', function () {
+  describe('#createServiceFromFormModel', function () {
```
pierlo-upitup: thx for that!

pierlo-upitup:
Sorry for crashing the party but I've noticed a few things I'd like to see discussed before hitting Merge! Thanks

src/js/components/modals/ServiceFormModal.js
Poltergeist: We only want to have a valid service in the state, also we want to validate the JSON. This is ensuring both if you have a better idea I am open for it.

pierlo-upitup: I agree, but silencing the whole of line 82 here can lead to situations where we don't know if the JSON parsing itself failed, or if the constructor pulled some unexpected behaviour.

I think I'd prefer to have 
```
try {
  let appDefinition = JSON.parse(app);
} catch(e) {}

this.setState({app, service: new Service(appDefinition)});
```
If the Service constructor doesn't like the input, it could return `null`?

src/js/components/modals/ServiceFormModal.js
Poltergeist: The triggerAdvancedSubmit returns the data from the form including validation, model and definition of the form. This is needed to get the data from outside of the component. However the naming I must confess isn't good. I removed the Advanced part and the getter function for the trigger is called `getTriggerSubmit` instead of `getAdvancedSubmit`.

src/js/components/modals/ServiceFormModal.js
Poltergeist: onClose has a default value of `function () {}` so undefined won't be called.

pierlo-upitup: Cool I missed that. Damn ES6-style statics.

src/js/constants/ServiceSchema.js
Poltergeist: Well it is a schema in terms of what the schemaForm calls it. From this schema the formDefenition is created which translates to we could add the validation here.

src/js/pages/services/ServicesTab.js
Poltergeist: Only if it is a ES6 class right?

pierlo-upitup: Well, true. But I'm also not sure about using the `() => ` inline is OK.

src/js/utils/ServiceUtil.js
Poltergeist: Added a test for `{}` and `null` as a parameter and made sure that they work.

src/js/utils/ServiceUtil.js
Poltergeist: good question!

src/js/utils/__tests__/ServiceUtil-test.js
Poltergeist: You are welcome and you were right.

src/js/pages/services/ServicesTab.js
Poltergeist: Well it is in other places e.g. [HealthTab.js:118](https://github.com/dcos/dcos-ui/blob/master/src/js/components/HealthTab.js#L118)

mlunoe:
Solid work @Poltergeist

PR #213, opened 2016-05-18
Breadcrumb Component

MatApple:
This is largely inspired by the awesome Marathon Breadcrumb component but abstracted to work dynamically on top of react-router.

### DIFF
--- a/src/js/components/Breadcrumb.js
+++ b/src/js/components/Breadcrumb.js
@@ -60,6 +60,7 @@ class Breadcrumb extends React.Component {
 
   handleResize() {
     let availableWidth = this.getAvailableWidth();
+
     this.setState({
       availableWidth,
       collapsed: this.shouldCollapse(availableWidth, this.state.expandedWidth)
@@ -95,15 +96,15 @@ class Breadcrumb extends React.Component {
   getExpandedWidth() {
     // array/splat casts NodeList to array
     let listItems = [...ReactDOM.findDOMNode(this).children]
-      .filter(function (_, i) {
+      .filter(function (_, index) {
         // Filter out even nodes containing '>'
-        return i % 2;
+        return index % 2;
       });
 
     return listItems
-      .map((item, n) => {
-        let isFirstItem = n === 0;
-        let isLastItem = n === listItems.length - 1;
+      .map((item, index) => {
+        let isFirstItem = index === 0;
+        let isLastItem = index === listItems.length - 1;
         if (isFirstItem || isLastItem) {
           return this.getWidthFromExpandedItem(item);
         }
@@ -145,6 +146,33 @@ class Breadcrumb extends React.Component {
     });
 
     return namedRoutes;
+  }
+
+  getCurrentRouteParams() {
+    // Isolated for easier testing
+    return this.context.router.getCurrentParams();
+  }
+
+  getCrumb(label, route, params, key) {
+    return (
+      <li key={key}>
+        <Link to={route}
+            params={params}
+            title={label}>
+          {label}
+        </Link>
+      </li>
+    );
+  }
+
+  getCrumbDivider(key) {
+    return (
+      <li key={key} >
+        <IconChevron
+          className="icon icon-small"
+          isForward={true} />
+      </li>
+    );
   }
 
   shouldCollapse(availableWidth, expandedWidth) {
@@ -192,15 +220,10 @@ class Breadcrumb extends React.Component {
     }
     // Can return multiple route objects to get rendered in crumbs.
     // Useful for splitting a route parameter like a file path.
-    return [{
-      label,
-      route,
-      params
-    }];
+    return [{label, route, params}];
   }
 
   renderCrumbsFromRoute() {
-    let {router} = this.context;
     let {buildCrumb, shift} = this.props;
 
     if (buildCrumb == null) {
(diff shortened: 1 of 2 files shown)


mesosphere-web:
Reviewer: @orlandohohmeier

src/styles/components/breadcrumbs.less
```diff
@@ -31,3 +31,111 @@ Breadcrumbs
...
+      overflow: hidden;
+
+      &:hover {
+        color: color-lighten(@neutral, 10);
+      }
+    }
+
+    .icon {
+      opacity: 0.5;
+      width: 14px;
```
mlunoe: :abc:

src/js/components/Breadcrumb.js
```diff
@@ -0,0 +1,295 @@
...
+      return crumbs;
+    }, []);
+  }
+
+  render() {
+    let classSet = classNames(
+      'list-unstyled breadcrumb',
+      {
+        collapsed: this.state.collapsed,
+        inverse: this.props.inverse
```
mlunoe: Could we just add the inverse class and use `classnames/dedupe` to let the user override it if needed, like we are doing here?
https://github.com/dcos/dcos-ui/pull/214/files#diff-6d467002be6b32378982c60715cf6919R110

src/js/components/Breadcrumb.js
```diff
@@ -0,0 +1,295 @@
...
+    } else {
+      // Split Label on space and - then capitalize each word and join.
+      // /foo-bar/ --> Foo Bar
+      label = label.split(splitLabelRegex).map(function (word) {
+        return StringUtil.capitalize(word);
+      }).join(joinLabelCharacter);
+    }
+    // Can return multiple route objects to get rendered in crumbs.
+    // Useful for splitting a route parameter like a file path.
+    return [{
```
mlunoe: #nit Let's put this in one line

src/js/components/Breadcrumb.js
```diff
@@ -0,0 +1,295 @@
...
+  getExpandedWidth() {
+    // array/splat casts NodeList to array
+    let listItems = [...ReactDOM.findDOMNode(this).children]
+      .filter(function (_, i) {
+        // Filter out even nodes containing '>'
+        return i % 2;
+      });
+
+    return listItems
+      .map((item, n) => {
```
mlunoe: Could we use `index` instead of `n`?

src/js/components/Breadcrumb.js
```diff
@@ -0,0 +1,295 @@
...
+    let link = item.children[0];
+    let textWidth = link.scrollWidth - link.offsetWidth;
+
+    return textWidth + PADDED_ICON_WIDTH;
+  }
+
+  getExpandedWidth() {
+    // array/splat casts NodeList to array
+    let listItems = [...ReactDOM.findDOMNode(this).children]
+      .filter(function (_, i) {
```
mlunoe: `index` in stead of `i`

src/js/components/Breadcrumb.js
```diff
@@ -0,0 +1,295 @@
...
+        name,
+        params,
+        labelIsParam
+      );
+
+      crumbObjects.forEach((crumbParams, crumbIndex) => {
+        let {label, route, params} = crumbParams;
+
+        crumbs.push(
+          <li key={crumbKey++}>
```
mlunoe: Could we move the view-parts to their own functions? e.g. `getLink` and `getArrow`, that should trim this function a little bit down in size and remove the `// Push divider ( > )` comment, since that should then be obvious

mlunoe:
Some small comments, but I think in general it looks good. NB: I did not test this out

src/js/components/Breadcrumb.js
```diff
@@ -0,0 +1,295 @@
...
+    // Can return multiple route objects to get rendered in crumbs.
+    // Useful for splitting a route parameter like a file path.
+    return [{
+      label,
+      route,
+      params
+    }];
+  }
+
+  renderCrumbsFromRoute() {
```
mlunoe: Could we test this function?

src/styles/components/breadcrumbs.less
MatApple: done

src/js/components/Breadcrumb.js
MatApple: This is actually really tricky to test and haven't found an appropriate way yet. I think it would be much easier as an integration test on the pages that implement it.

src/js/components/Breadcrumb.js
orlandohohmeier: Any chance we can combine the filter and map?

MatApple: Not conveniently because we use `listItems.length` in the `map` function. We could of course get round that with some math and put everything inside a `reduce` but I think it will make it less obvious. The `listItems.length` will rarely be > 10 I'd imagine so there isn't any performance concerns.

src/js/components/Breadcrumb.js
```diff
@@ -0,0 +1,305 @@
...
+    if (this.state.collapsed !== nextState.collapsed) {
+      return true;
+    }
+
+    return !DeepEqual(this.props, nextProps);
+  }
+
+  handleResize() {
+    let availableWidth = this.getAvailableWidth();
+    this.setState({
```
orlandohohmeier: 👾

orlandohohmeier: Please add newline between the `let` and `this.setState`...   #nit

orlandohohmeier:
Great Job!

src/js/components/Breadcrumb.js
MatApple: done!

pierlo-upitup:
Shouldn't the Breadcrumb detect changes in the router and re-render the crumbs based on that? I've tried to use it a new component and it doesn't seem to do so.

PR #215, opened 2016-05-18
Show notifications in tabs

kennyt:
px

you won't be able to reproduce this on your own, but that's what it looks like. i'd be open to comments on the implementation. sidebar badge still needs help with css

### DIFF
--- a/src/js/mixins/TabsMixin.js
+++ b/src/js/mixins/TabsMixin.js
@@ -5,6 +5,7 @@ import React from 'react';
 /* eslint-enable no-unused-vars */
 
 import TabsUtil from '../utils/TabsUtil';
+import NotificationStore from '../stores/NotificationStore';
 
 /**
  * Adds tabs-specific methods onto a class.
@@ -93,9 +94,9 @@ const TabsMixin = {
    * @return {Component} React component to render
    */
   tabs_getRoutedItem(customClasses, tab) {
+    let notificationCount = NotificationStore.getNotificationCount(tab);
     let tabLabelClass = classNames(customClasses, {'tab-item-label': true});
-
-    return (
+    let link = (
       <Link
         to={tab}
         className={tabLabelClass}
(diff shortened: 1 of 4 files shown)


mesosphere-web:
Reviewer: @mlunoe

daiyiwhale:
This looks rad, nice!

src/js/mixins/TabsMixin.js
```diff
@@ -93,16 +94,27 @@ const TabsMixin = {
    * @return {Component} React component to render
    */
   tabs_getRoutedItem(customClasses, tab) {
+    let notificationCount = NotificationStore.getNotificationCount(tab);
```
rcorral: Is `tab` too generic? Meaning, could we have conflicting tab names?

kennyt: hmm, good point. most of our tab names are named exactly the same as the route name, but there are a few tabs that aren't routed yet. what do you think we should do?

jfurrow:
@kennyt Let me know when you want to go over the styling of this.

src/js/mixins/TabsMixin.js
```diff
@@ -93,16 +94,27 @@ const TabsMixin = {
...
     let tabLabelClass = classNames(customClasses, {'tab-item-label': true});
-
-    return (
+    let link = (
```
mlunoe: We can just return this directly, if we put it below the `if`-statement

src/js/pages/SystemPage.js
```diff
@@ -84,13 +85,25 @@ class SystemPage extends mixin(TabsMixin) {
   }
 
   getRoutedItem(tab) {
-    return (
+    let notificationCount = NotificationStore.getNotificationCount(tab);
+    let link = (
```
mlunoe: Same thing here

philipnrmn:
Hey Kenny, nice work :)

One small CSS issue - in page header tabs, when a notification is added, characters with descenders get cut off:



I was able to fix this with

```css
.badge-container .tab-item-label { line-height: 1em; }
```

jfurrow:
Kenny and I chatted and are going to tackle the styling in another PR, so I'm merging this.

PR #217, opened 2016-05-18
Adds support for any resource type that mesos supports now and in the future

rcorral:
Mesos will be adding support for GPU resource, which breaks the UI.
This pull request introduces support for any resource that Mesos adds in the future. Currently it doesn't actually show the unknown resources anywhere in the UI. So this is the minimal amount of work needed so that the UI doesn't crash when new resources are added.

### DIFF
--- a/src/js/components/charts/ResourceBarChart.js
+++ b/src/js/components/charts/ResourceBarChart.js
@@ -3,7 +3,7 @@ const React = require('react');
 
 const BarChart = require('./BarChart');
 const Chart = require('./Chart');
-const ResourceTypes = require('../../constants/ResourceTypes');
+const ResourcesUtil = require('../../utils/ResourcesUtil');
 
 // number to fit design of width vs. height ratio
 const WIDTH_HEIGHT_RATIO = 4.5;
@@ -43,7 +43,7 @@ let ResourceBarChart = React.createClass({
     return [{
       id: 'used_resources',
       name: selectedResource + ' allocated',
-      colorIndex: ResourceTypes[selectedResource].colorIndex,
+      colorIndex: ResourcesUtil.getResourceColor(selectedResource),
       values: props.resources[selectedResource]
     }];
   },
@@ -63,18 +63,20 @@ let ResourceBarChart = React.createClass({
   getModeButtons: function () {
     let selectedResource = this.props.selectedResource;
 
-    return Object.keys(ResourceTypes).map((key) => {
+    let resourceColors = ResourcesUtil.getResourceColors();
+    let resourceLabels = ResourcesUtil.getResourceLabels();
+
+    return ResourcesUtil.getDefaultResources().map((resource) => {
       let classSet = classNames('button button-stroke button-inverse', {
-        'active': selectedResource === key
+        'active': selectedResource === resource
       });
-      let info = ResourceTypes[key];
 
       return (
         <button
-            key={key}
-            className={classSet + ' path-color-' + info.colorIndex}
-            onClick={this.handleSelectedResourceChange.bind(this, key)}>
-          {info.label}
+            key={resource}
+            className={`${classSet} path-color-${resourceColors[resource]}`}
+            onClick={this.handleSelectedResourceChange.bind(this, resource)}>
+          {resourceLabels[resource]}
         </button>
       );
     });
@@ -93,8 +95,9 @@ let ResourceBarChart = React.createClass({
     );
   },
 
-  getHeadline: function (info) {
-    let headline = info.label + ' Allocation Rate';
+  getHeadline: function (resource) {
+    let label = ResourcesUtil.getResourceLabel(resource);
+    let headline = `${label} Allocation Rate`;
 
     return (
       <div>
@@ -109,8 +112,6 @@ let ResourceBarChart = React.createClass({
   },
 
   render: function () {
-    let info = ResourceTypes[this.props.selectedResource];
-
     return (
       <div className="chart panel panel-inverse">
         <div className="panel-header panel-header-large no-border flush-bottom">
@@ -118,7 +119,7 @@ let ResourceBarChart = React.createClass({
             {this.getModeButtons()}
           </div>
           <div className="inverse">
-            {this.getHeadline(info)}
+            {this.getHeadline(this.props.selectedResource)}
           </div>
         </div>
         <div className="panel-content" ref="panelContent">
--- a/src/js/constants/ResourceTypes.js
+++ b/src/js/constants/ResourceTypes.js
@@ -1,7 +1,0 @@
-var ResourceTypes = {
-  cpus: {label: 'CPU', colorIndex: 0},
-  mem: {label: 'Memory', colorIndex: 6},
-  disk: {label: 'Disk', colorIndex: 3}
-};
-
-module.exports = ResourceTypes;
--- a/src/js/utils/ResourcesUtil.js
+++ b/src/js/utils/ResourcesUtil.js
@@ -29,6 +29,7 @@ const availableColors = Array(9).fill().map(function (value, index) {
 
 function getAvailableColors() {
   let colors = availableColors.slice(0);
+
   // Concat the array twice. In case there's many resources.
   // Unlikely that there will more than 4 anyways.
   return colors.concat(colors);
@@ -110,6 +111,7 @@ const ResourcesUtil = {
     }
 
     let index = opts.resourceList.indexOf(resource);
+
     return opts.availableColors[index];
   },
 
@@ -117,6 +119,7 @@ const ResourcesUtil = {
     // Map the default ones first
     let map = this.getDefaultResources().reduce((memo, resource) => {
       memo[resource] = this.getResourceColor(resource);
+
       return memo;
     }, {});
 
@@ -127,6 +130,7 @@ const ResourcesUtil = {
       memo[resource] = this.getResourceColor(resource, {
         resourceList: rest
       });
+
       return memo;
     }, map);
   }
--- a/src/js/utils/ResourcesUtil.js
+++ b/src/js/utils/ResourcesUtil.js
@@ -1,0 +1,127 @@
+import CompositeState from '../structs/CompositeState';
+import Node from '../structs/Node';
+import StringUtil from './StringUtil';
+
+const DefaultResourceTypes = {
+  cpus: {label: 'CPU', colorID: 0},
+  mem: {label: 'Memory', colorID: 6},
+  disk: {label: 'Disk', colorID: 3}
+};
+
+const usedColors = Object.keys(DefaultResourceTypes)
+.map(function (resource) {
+  return DefaultResourceTypes[resource].colorID;
+});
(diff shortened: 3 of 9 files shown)


mesosphere-web:
Reviewer: @philipnrmn

src/js/utils/ResourcesUtil.js
```diff
@@ -0,0 +1,127 @@
...
+      return StringUtil.capitalize(resource.toLowerCase());
+    }
+  },
+
+  getResourceLabels: function () {
+    let resources = this.getAvailableResources();
+
+    return resources.reduce((memo, resource) => {
+      memo[resource] = this.getResourceLabel(resource);
+      return memo;
```
kennyt: ah so you're jumping onto the `reduce` wagon. that's cool. can you put a newline before `return`?

src/js/utils/ResourcesUtil.js
```diff
@@ -0,0 +1,127 @@
...
+    }, {});
+
+    let rest = this.getAdditionalResources();
+
+    // Map the rest
+    return rest.reduce((memo, resource) => {
+      memo[resource] = this.getResourceColor(resource, {
+        resourceList: rest
+      });
+      return memo;
```
kennyt: can you put a newline here too? we all implicitly agreed that returns should have newlines before them unless it's by itself

src/js/constants/ResourceTypes.js
```diff
@@ -1,7 +0,0 @@
-var ResourceTypes = {
```
kennyt: this file is still being used in a test

src/js/utils/ResourcesUtil.js
```diff
@@ -0,0 +1,127 @@
...
+});
+const availableColors = Array(9).fill().map(function (value, index) {
+  return index;
+}).filter(function (value) {
+  return usedColors.indexOf(value) === -1;
+});
+
+function getAvailableColors() {
+  let colors = availableColors.slice(0);
+  // Concat the array twice. This is in case there's many resource.
```
kennyt: `many resources`

src/js/utils/__tests__/ResourcesUtil-test.js
```diff
@@ -0,0 +1,171 @@
...
+      let resources = ResourcesUtil.getAvailableResources();
+      // Should at least have cpu, mem, disk
+      expect(resources.length).toBeGreaterThan(2);
+    });
+
+    it('gets available resources from a node', function () {
+      CompositeState.getNodesList = createFnWithResources({foo: 0, bar: 0});
+
+      let resources = ResourcesUtil.getAvailableResources();
+      // Should at least have cpu, mem, disk
```
philipnrmn: I think these repeated comments can be removed.

src/js/utils/ResourcesUtil.js
```diff
@@ -0,0 +1,127 @@
...
+  return colors.concat(colors);
+}
+
+const ResourcesUtil = {
+  getDefaultResources: function () {
+    return Object.keys(DefaultResourceTypes);
+  },
+
+  getAvailableResources: function (excludeList = []) {
+    let item = CompositeState.getNodesList().getItems()[0];
```
philipnrmn: I feel like we should use a safe accessor here instead of relying on it being `length > 0`. 

I know a cluster should technically always have nodes, but I can imagine some bad state occurring which causes reference errors.

src/js/utils/ResourcesUtil.js
```diff
@@ -0,0 +1,127 @@
...
+
+    if (!item) {
+      item = CompositeState.getServiceList().getItems()[0];
+
+      if (!item) {
+        let fakeUsedResources = this.getDefaultResources()
+        .reduce(function (memo, resource) {
```
philipnrmn: I think this should be indented one deeper.

src/js/utils/ResourcesUtil.js
```diff
@@ -0,0 +1,127 @@
...
+          memo[resource] = 0;
+          return memo;
+        }, {});
+        item = new Node({used_resources: fakeUsedResources});
```
philipnrmn: Here we create a new Node with dummy `used_resources`, then call `get('used_resources')` and never reference the created Node again. I reckon it'd be a bit nicer to pass the dummy resources straight down.

src/js/utils/ResourcesUtil.js
```diff
@@ -0,0 +1,127 @@
...
+    rest.sort();
+    return rest;
+  },
+
+  getResourceLabel: function (resource) {
+    if (DefaultResourceTypes[resource]) {
+      return DefaultResourceTypes[resource].label;
+    }
+
+    if (resource.length <= 3) {
```
philipnrmn: This catches `CPU`, right? Can we either get an explanatory comment, or just do `/cpu/i.test(resource)`?

philipnrmn: Hmm, `RAM` as well I guess. OK, at least a comment then.

src/js/utils/ResourcesUtil.js
```diff
@@ -0,0 +1,127 @@
...
+          memo[resource] = 0;
+          return memo;
+        }, {});
+        item = new Node({used_resources: fakeUsedResources});
+      }
+    }
+
+    let resources = Object.keys(item.get('used_resources'));
+
+    if (excludeList.length) {
```
philipnrmn: I like the optimisation, but ISTR remember a convention about testing falsey values explicitly? Ignore me if that's wrong.

src/js/utils/ResourcesUtil.js
```diff
@@ -0,0 +1,127 @@
...
+
+    let rest = this.getAdditionalResources();
+
+    // Map the rest
+    return rest.reduce((memo, resource) => {
+      memo[resource] = this.getResourceColor(resource, {
+        resourceList: rest
+      });
+      return memo;
+    }, map);
```
philipnrmn: This is kind of subjective, but I think it'd be a bit more readable to `return Object.extend(defaultResourceColors, additionalResourceColors)`. No major issue with this implementation though if you prefer it.

src/js/utils/ResourcesUtil.js
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    return resources;
+  },
+
+  getAdditionalResources: function () {
+    let rest = this.getAvailableResources(this.getDefaultResources());
+    // We sort so we get the same color in future calls to this method
+    rest.sort();
+    return rest;
```
philipnrmn: Unimportant, but `return rest.sort();` works if you want to save a line.

src/js/components/charts/ResourceBarChart.js
```diff
@@ -63,18 +63,20 @@ let ResourceBarChart = React.createClass({
...
-      let info = ResourceTypes[key];
 
       return (
         <button
-            key={key}
-            className={classSet + ' path-color-' + info.colorIndex}
-            onClick={this.handleSelectedResourceChange.bind(this, key)}>
-          {info.label}
+            key={resource}
+            className={`${classSet} path-color-${resourceColors[resource]}`}
```
philipnrmn: Kind of feels like these two belong in the `classNames` above.

src/js/utils/ResourcesUtil.js
```diff
@@ -0,0 +1,127 @@
...
+import StringUtil from './StringUtil';
+
+const DefaultResourceTypes = {
+  cpus: {label: 'CPU', colorID: 0},
+  mem: {label: 'Memory', colorID: 6},
+  disk: {label: 'Disk', colorID: 3}
+};
+
+const usedColors = Object.keys(DefaultResourceTypes)
+.map(function (resource) {
```
philipnrmn: should this be indented?

src/js/utils/ResourcesUtil.js
philipnrmn: Also, dummy resources could be held as a static object at the top of the file rather than generated fresh each time.

src/js/utils/ResourcesUtil.js
rcorral: no

src/js/components/charts/ResourceBarChart.js
rcorral: Not changing the original implementation, just changing the variable used.

src/js/utils/ResourcesUtil.js
rcorral: Not the usual pattern right now.

src/js/utils/ResourcesUtil.js
rcorral: Done.

src/js/utils/ResourcesUtil.js
rcorral: I was going to do it but `Object.extend` isn't working with tests. So quickly gave up.

rcorral: Opted not to do this.

src/js/utils/ResourcesUtil.js
rcorral: Done.

src/js/constants/ResourceTypes.js
rcorral: Fixed.

src/js/utils/__tests__/ResourcesUtil-test.js
rcorral: Removed.

src/js/utils/ResourcesUtil.js
rcorral: Good idea, created a cached instance.

src/js/utils/ResourcesUtil.js
rcorral: Done

src/js/utils/ResourcesUtil.js
rcorral: Not sure if there's a standard way of doing this. I made the check more explicit.

src/js/utils/ResourcesUtil.js
rcorral: ¯\\_(ツ)_/¯

src/js/utils/ResourcesUtil.js
MatApple: I think you meant to use `Object.assign(` :)

src/js/utils/ResourcesUtil.js
MatApple: Should have a new line above return 👍

rcorral: Done. This needs to be an eslint rule.

src/js/components/charts/ResourceBarChart.js
philipnrmn: How meta!

src/js/utils/ResourcesUtil.js
kennyt: you haven't put newlines before this return and some others in this file

kennyt:
tested this out, works great. just one more comment about adding newlines before returns then i'll merge this thing

PR #220, opened 2016-05-19
Integrate Marathon version/config data

orlandohohmeier:
Integrate the Marathon version/config data. This will lay the foundation for the service detail configuration tab. These commits are cherry picked from #155 in order to reduce the PR size.

### DIFF
--- a/src/js/stores/DCOSStore.js
+++ b/src/js/stores/DCOSStore.js
@@ -26,7 +26,9 @@ class DCOSStore extends EventEmitter {
     });
 
     this.data = {
-      marathon: new ServiceTree(),
+      marathon: {
+        serviceTree: new ServiceTree()
+      },
       mesos: new SummaryList(),
       dataProcessed: false
     };
@@ -46,12 +48,12 @@ class DCOSStore extends EventEmitter {
   }
 
   onMarathonGroupsChange() {
-    let groups = MarathonStore.get('groups');
-    if (!(groups instanceof ServiceTree)) {
+    let serviceTree = MarathonStore.get('groups');
+    if (!(serviceTree instanceof ServiceTree)) {
       return;
     }
 
-    this.data.marathon = groups;
+    this.data.marathon.serviceTree = serviceTree;
     this.data.dataProcessed = true;
     this.emit(DCOS_CHANGE);
   }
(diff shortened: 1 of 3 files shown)


src/js/stores/DCOSStore.js
```diff
@@ -25,7 +25,9 @@ class DCOSStore extends EventEmitter {
     });
 
     this.data = {
-      marathon: new ServiceTree(),
+      marathon: {
+        tree: new ServiceTree()
```
pierlo-upitup: `services`? Why do we want to name this after the data structure it uses rather than what it represents?

src/js/stores/DCOSStore.js
```diff
@@ -35,36 +37,53 @@ class DCOSStore extends EventEmitter {
...
+  fetchServiceVersions(serviceID) {
+    MarathonStore.fetchServiceVersions(serviceID);
+  }
+
+  /**
+   * Process Marathon groups data
    * @param {ServiceTree} data
    */
-  processMarathonData(data) {
+  processMarathonGroups(data) {
```
pierlo-upitup: I'd argue `processMarathonServiceTree` would be more explicit, given that it definitely expects a `ServiceTree`?

src/js/stores/DCOSStore.js
```diff
@@ -130,20 +149,19 @@ class DCOSStore extends EventEmitter {
...
       // Create framework dict from Mesos data
-      let frameworks = mesos.frameworks
-        .reduce(function (map, framework) {
-          map[framework.name] = framework;
+      frameworks = frameworks.reduce(function (map, framework) {
+        map[framework.name] = framework;
 
-          return map;
-        }, {});
+        return map;
```
pierlo-upitup: This is not a `Map` – so naming it `map` is kinda confusing. How about the classic `memo`?

src/js/stores/DCOSStore.js
```diff
@@ -130,20 +149,19 @@ class DCOSStore extends EventEmitter {
...
+      }, {});
```
pierlo-upitup: ...or make the `{}` a `new Map()` ?

pierlo-upitup:
Only a few naming nits, rest seems solid.

src/js/stores/DCOSStore.js
orlandohohmeier: I simply named it tree as I wasn't happy with all other names that had come to my mind... I've changed it to `serviceTree` as `services` alone kinda looks like it's holding a flat list, hope that's ok for you.

src/js/stores/DCOSStore.js
orlandohohmeier: This method was already renamed in a different PR, which tackled issues with the general data handling.

PR #221, opened 2016-05-19
Add a configuration view

orlandohohmeier:
---

ℹ️   _This branch depends on changes that will be introduced with #236, thus it will fail until the respective branch is merged._

---
Introduce a config/version view. This component will be used in the service detail configuration tab. These commits are cherry picked from #155 in order to reduce the PR size.

_Integrated config view_

### DIFF
--- a/src/js/components/ConfigurationView.js
+++ b/src/js/components/ConfigurationView.js
@@ -1,0 +1,242 @@
+import mixin from 'reactjs-mixin';
+import React from 'react';
+import {StoreMixin} from 'mesosphere-shared-reactjs';
+
+import DCOSStore from '../stores/DCOSStore'
+import DescriptionList from './DescriptionList';
+import Service from '../structs/Service';
+import StringUtil from '../utils/StringUtil';
+
+const sectionClassName = 'container-fluid container-pod container-pod-super-short flush flush-bottom';
+
+function fetchVersion(service, versionID) {
+  if (service.getVersions().get(versionID) == null) {
+    DCOSStore.fetchServiceVersion(service.getId(), versionID);
+  }
+}
+
+class ConfigurationView extends mixin(StoreMixin) {
+  constructor() {
+    super(...arguments);
+
+    this.store_listeners = [{name: 'dcos', events: ['change']}];
+  }
+
+  componentWillMount() {
+    super.componentWillMount();
+
+    let {service, versionID} = this.props;
+    fetchVersion(service, versionID);
+  }
+
+  componentWillReceiveProps({service, versionID}) {
+    super.componentWillReceiveProps(...arguments);
+
+    fetchVersion(service, versionID);
+  }
+
+  getGenralSection({id, cmd, cpus, disk, instances, mem}) {
+    let headerValueMapping = {
+      'ID': id,
+      'CPUs': cpus,
+      'Memory (MiB)': mem,
+      'Disk Space (Mib)': disk,
+      'Instances': instances,
+      'Command': cmd
+    };
+
+    return (
+      <DescriptionList className={sectionClassName}
+        hash={headerValueMapping}
+        headline="General" />
+    );
+  }
+
+  getDockerContainerSection({container}) {
+    if (container == null || container.docker == null) {
+      return null;
+    }
+
+    let {docker} = container;
+
+    let headerValueMapping = {
+      'Force Pull Image': docker.forcePullImage,
+      'Image': docker.image,
+      'Network': docker.network,
+      'Parameters': StringUtil.arrayToJoinedString(docker.parameters),
+      'Privileged': docker.privileged
+    };
+
+    return (
+      <DescriptionList className={sectionClassName}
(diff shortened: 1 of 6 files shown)


src/js/components/ConfigurationView.js
```diff
@@ -0,0 +1,242 @@
...
+    fetchVersion(service, versionID);
+  }
+
+  componentWillReceiveProps({service, versionID}) {
+    super.componentWillReceiveProps(...arguments);
+
+    fetchVersion(service, versionID);
+  }
+
+  getGenralSection({id, cmd, cpus, disk, instances, mem}) {
```
pierlo-upitup: General?

src/js/components/ConfigurationView.js
```diff
@@ -0,0 +1,242 @@
...
+
+class ConfigurationView extends mixin(StoreMixin) {
+  constructor() {
+    super(...arguments);
+
+    this.store_listeners = [{name: 'dcos', events: ['change']}];
+  }
+
+  componentWillMount() {
+    super.componentWillMount();
```
pierlo-upitup: So to understand: when using `reactjs-mixin` we need to always remember to call `super.[lifeCycleMethod]`?

pierlo-upitup: Actually, I don't see why this one is needed. In any case, StoreMixin only defines `didMount` and `willUnmount` LF methods.

src/js/components/ConfigurationView.js
```diff
@@ -0,0 +1,242 @@
...
+
+    let {service, versionID} = this.props;
+    fetchVersion(service, versionID);
+  }
+
+  componentWillReceiveProps({service, versionID}) {
+    super.componentWillReceiveProps(...arguments);
```
pierlo-upitup: Same here. I feel lost 👻

src/js/components/ConfigurationView.js
```diff
@@ -0,0 +1,242 @@
...
+
+function fetchVersion(service, versionID) {
+  if (service.getVersions().get(versionID) == null) {
+    DCOSStore.fetchServiceVersion(service.getId(), versionID);
+  }
+}
+
+class ConfigurationView extends mixin(StoreMixin) {
+  constructor() {
+    super(...arguments);
```
pierlo-upitup: I'd like to know the case for passing `...arguments` here VS just `super()`.

src/js/components/ConfigurationView.js
```diff
@@ -0,0 +1,242 @@
...
+          case 'COMMAND':
+            headerValueMapping = Object.assign(headerValueMapping, {
+              'Command': healthCheck.command.value,
+            });
+            break;
+          case 'HTTP':
+            headerValueMapping = Object.assign(headerValueMapping, {
+              'Path': healthCheck.path,
+              'Port': healthCheck.port,
+              'Port Index': healthCheck.portIndex || 'PORT_INDEX',
```
pierlo-upitup: hat does `PORT_INDEX` mean?

src/js/utils/StringUtil.js
```diff
@@ -1,6 +1,14 @@
 import marked from 'marked';
 
 const StringUtil = {
+  arrayToCommaSeparatedString: function (array) {
```
pierlo-upitup: I don't see a huge need for this function, to be honest – but I guess we could make it a tad more generic for example by having the `separator` be an optional argument?
How about `arrayToJoinedString(array = [], separator = ',') { ` ?

pierlo-upitup:
Cosmetics/nits – rest LGTM

src/js/components/ConfigurationView.js
orlandohohmeier: The `...arguments` kinda makes sure, that this component is future proof and we're not missing something if we update one of the parent classes. 
I don't think that we have a rule for these yet, thus I suggest to discuss this in our next sync meeting. 

/cc @rcorral /link https://github.com/dcos/dcos-ui/pull/207#discussion_r63930608

src/js/components/ConfigurationView.js
orlandohohmeier: Upsi, that's is actually a bug. I somehow screwed that one. I blame the jetlag... The `PORT_INDEX` is actually the default `healthCheck.portType` value which defines the `portIndex` value type.

src/js/utils/StringUtil.js
orlandohohmeier: Good call. I only extracted that one in to a function as I was repeating myself over and over again. I'll generalize the method.

pierlo-upitup:
#220 has landed, please rebase etc.

src/js/components/ConfigurationView.js
rcorral: This is a pattern we've been doing for a very long time now. It makes it easy to add new mixins without having to think about which methods each of them implement.
The overhead for the reward is negligible.

PR #222, opened 2016-05-19
Add service detail configuration tab

orlandohohmeier:
---

ℹ️   _This branch depends on changes that will be introduced with #221, thus it will fail until the respective branch is merged._

---
Introduce a service detail configuration tab. These commits are cherry picked from #155 in order to reduce the PR size.

_Integrated config view_


pierlo-upitup:
LGTM. Waiting for the dependencies to be merged.

PR #224, opened 2016-05-19
Only cast list items if type is set

philipnrmn:
This PR addresses concerns over performance problems with untyped list initiation. I ran a couple of benchmarks - initiating a `NodesList` (untyped, inherits from `List`) of 1,000 items

 - Before: `1,653 ops/sec`
 - After: `3,395 ops/sec`


mesosphere-web:
Reviewer: @orlandohohmeier

PR #226, opened 2016-05-19
Bug - Give PageHeader correct padding when it has no tabs

daiyiwhale:
This class was added to `PageHeader`, and needs to be removed in these special cases when there are no tabs. 

For example, 

before:

after:

### DIFF
--- a/src/js/components/PageHeader.js
+++ b/src/js/components/PageHeader.js
@@ -84,9 +84,10 @@ class PageHeader extends React.Component {
     );
 
     let dividerClasses = classNames(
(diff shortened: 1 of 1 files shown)


mesosphere-web:
Reviewer: @mlunoe

kennyt:
px

this still happens. is this supposed to fix that?

daiyiwhale:
Dammit that's a new different bug haha. One sec

daiyiwhale:
Thanks for checking for me! I should have done that

kennyt:
merging on integration pass

PR #228, opened 2016-05-19
Bug - Disappearing padding from pages

daiyiwhale:
This bug happened when scrolling is set to `always` on Mac and the content flowed off the page (note the lack of margin under the secondary tabs):

picture of bug

To fix I removed `flex-shrink` so the page doesn't force all the children to shrink. 

I have checked the pages but I'd love a second glance! <3

Shoutout to @jfurrow for his champion css debug insight as always ✨

### DIFF
--- a/src/js/components/Page.js
+++ b/src/js/components/Page.js
@@ -105,7 +105,7 @@ var Page = React.createClass({
         {this.getPageHeader(title, navigation)}
         <GeminiScrollbar autoshow={true} className="page-content
           container-scrollable inverse" ref="pageRef">
-          <div className="flex-container-col flex-grow container container-fluid container-pod container-pod-short-top">
+          <div className="flex-container-col container container-fluid container-pod container-pod-short-top">
             {this.getChildren()}
           </div>
         </GeminiScrollbar>
--- a/src/js/components/Page.js
+++ b/src/js/components/Page.js
@@ -105,7 +105,7 @@ var Page = React.createClass({
         {this.getPageHeader(title, navigation)}
         <GeminiScrollbar autoshow={true} className="page-content
           container-scrollable inverse" ref="pageRef">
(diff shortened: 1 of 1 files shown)


mesosphere-web:
Reviewer: @philipnrmn

kennyt:
@mlunoe was there a reason we had `flex-shrink` here? for some reason i remember you had a reason

philipnrmn:
I've tested everything I can think of and it all looks good to me! Interested to see whether @mlunoe knows something I missed...

mlunoe:
This was introduced here: https://github.com/dcos/dcos-ui/commit/ab352353cb5401ff0f1253707fc53f280cd1085b#diff-2b67d6ccc87cd6654d79e88a5d127eeeL108

The purpose was to have the page with the log view properly grow and shrink to the page height, but I never verified that it didn't have implications anywhere else in the application. This was a suggested approach, but it is not ideal. I would much rather have the fix contained inside the log view.

src/js/components/Page.js
```diff
@@ -105,7 +105,7 @@ var Page = React.createClass({
         {this.getPageHeader(title, navigation)}
         <GeminiScrollbar autoshow={true} className="page-content
           container-scrollable inverse" ref="pageRef">
-          <div className="flex-container-col flex-shrink flex-grow container container-fluid container-pod container-pod-short-top">
```
mlunoe: Let's make it explicitly `flex-no-shrink`

mlunoe:
Let's create a ticket to fix the log view and properly, because this will have implications on log view. However, the log view needs to be fixed anyway because it is not working properly now even.
Reference: https://github.com/dcos/dcos-ui/pull/118/files#r64079221

daiyiwhale:
Ticket is there, comment addressed!

PR #231, opened 2016-05-20
Bug - StatusBar value not showing

daiyiwhale:
In ServicesPage, the `StatusBar` component was not showing value because of previously changed styles that animated `transform` rather than `width`. This PR updates `StatusBar` accordingly.


mesosphere-web:
Reviewer: @orlandohohmeier

PR #232, opened 2016-05-20
Use RequestUtil from mesosphere-shared-reactjs

jfurrow:
`RequestUtil` in https://github.com/dcos/mesosphere-shared-reactjs now contains all of the extra methods that previously existed only in DCOS UI.


mesosphere-web:
Reviewer: @MatApple

mlunoe:
Hmm... it feels sad to remove tests and not add them in the other repository. We should really add tests there. This shouldn't hold up this PR though

kennyt:
@mlunoe there are tests in mesosphere-shared-reactjs for these no?

jfurrow:
@mlunoe I didn't remove any tests, in fact I added 3 more :)

PR #238, opened 2016-05-23
Extract parent group id from input field

pierlo-upitup:
Fix the Create Group modal form to reproduce existing behaviour in Marathon UI.

Extract the parent group ID from the input field itself and show indication of target group in the modal header itself.

Example:

screen shot 2016-05-23 at 14 05 10
screen shot 2016-05-23 at 14 04 36

cc @MatApple and @leemunroe

### DIFF
--- a/src/js/components/modals/ServiceGroupFormModal.js
+++ b/src/js/components/modals/ServiceGroupFormModal.js
@@ -56,8 +56,12 @@ class ServiceGroupFormModal extends mixin(StoreMixin) {
   }
 
   handleNewGroupSubmit(model) {
+    let {parentGroupId} = this.props;
+
     this.setState({disableNewGroup: true});
-    MarathonStore.createGroup(model);
+    MarathonStore.createGroup(Object.assign({}, model,
+      {id: `${parentGroupId}/${model.id}`})
+    );
   }
 
   getErrorMessage() {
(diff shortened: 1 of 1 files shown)


mesosphere-web:
Reviewer: @kennyt

src/js/components/modals/ServiceGroupFormModal.js
```diff
@@ -92,25 +91,30 @@ class ServiceGroupFormModal extends mixin(StoreMixin) {
...
         onChange={this.resetState}
-        open={this.props.open}
+        open={props.open}
         definition={this.getNewGroupFormDefinition()}>
         <h2 className="modal-header-title text-align-center flush-top">
-          Create New Group
+          Create Group
         </h2>
+        <p className="text-align-center flush-top">
+          Enter a path for the new group under&nbsp;
```
kennyt: i'd just do `{'Enter a path for the new group under '}`

Poltergeist:
Works like advertised, thanks for the change and putting effort into this PR.

leemunroe:
@pierlo-upitup Do we need a trailing slash at the end of `/dadadaa`? Your call. I ask as it is what we currently have in Marathon.

Also consider using `Create Group` as the button label. I'd like to see us use more description button text as a design guideline. A good way to think about it is buttons (and links) should still make sense in isolation.

Other than those pet peeves, l👀ks g👀d

pierlo-upitup:
@leemunroe I decided to leave the path without trailing slash for now as I think it looks cleaner – if you disagree let me know. 

The button now spells `Create Group` as asked. 🚢

pierlo-upitup:
Running the cypress tests locally works, not sure why Velocity reports the failure.

MatApple:
This is awesome @pierlo-upitup! Thanks for going the extra mile - well done :)

PR #242, opened 2016-05-24
Add React ace component

Poltergeist:
This adds `react-ace` as an dependency which will be used for the `JSON` editor.

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,6 +39,7 @@
     "moment": "2.13.0",
     "query-string": "4.1.0",
     "react": "0.14.7",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -39,6 +39,7 @@
     "moment": "2.13.0",
     "query-string": "4.1.0",
     "react": "0.14.7",
+    "react-ace": "^3.4.1",
```
rcorral: Please remove the `^`! we don't use it :)

mesosphere-web:
Reviewer: @daiyiwhale

PR #245, opened 2016-05-24
Introduce job structs

orlandohohmeier:
---

⚠️   _This branch is still work in progress. Tests missing._ 

---

Introduce basic job and job tree struct. These structs are used to type and abstract the Chronos API responses.


orlandohohmeier:
Split this into a job and job tree branch.

PR #246, opened 2016-05-24
[2 of 2 Chronos] Introduce chronos store

orlandohohmeier:
---
~~⚠️   _This branch is still work in progress. It also depends on changes that will be introduced with #250 and #251, thus it will fail until the respective branch is merged._~~

---
Introduce and integrate a Chronos store to fetch, parse and store jobs data.


src/js/stores/ChronosStore.js
```diff
@@ -0,0 +1,121 @@
...
+
+    // Handle app actions
+    this.dispatcherIndex = AppDispatcher.register(({source, action}) => {
+      if (source !== SERVER_ACTION) {
+        return false;
+      }
+      switch (action.type) {
+        case REQUEST_CHRONOS_JOBS_SUCCESS:
+          if (Array.isArray(action.data)) {
+            action.data.forEach((job) => {
```
mlunoe: No need to parse data, just pass to `JobTree`

src/js/constants/EventTypes.js
```diff
@@ -1,11 +1,13 @@
...
+  'APP_STORE_CHANGE',
+  'AUTH_LOGIN_REDIRECT',
   'AUTH_USER_LOGIN_ERROR',
-  'AUTH_USER_LOGOUT_SUCCESS',
+  'AUTH_USER_LOGIN_SUCCESS',
   'AUTH_USER_LOGOUT_ERROR',
-  'AUTH_LOGIN_REDIRECT',
-  'APP_STORE_CHANGE',
+  'AUTH_USER_LOGOUT_SUCCESS',
+  'CHRONOS_JOBS_CHANGE',
```
mlunoe: Just introducing ` 'CHRONOS_JOBS_CHANGE'` and `'CHRONOS_JOBS_ERROR'` here. The reset is :abc:

src/js/stores/DCOSStore.js
```diff
@@ -34,12 +38,20 @@ class DCOSStore extends EventEmitter {
         serviceTree: new ServiceTree(),
         versions: new Map()
       },
+      chronos: {
+        jobTree: new JobTree(),
```
mlunoe: Not quite sure if this is the right way of doing this, as we try to refrain from storing structs in the stores (and potentially the omni store), and the wrap the data in the structs when the are returned from the store.

rcorral:
retest this please

src/js/stores/ChronosStore.js
```diff
@@ -0,0 +1,143 @@
...
+  }
+
+  /**
+   * @type {JobTree}
+   */
+  get jobTree() {
+    return new JobTree(this.data.jobTree);
+  }
+
+  static get storeID() {
```
rcorral: Why does this need to be a function?
Can we put it at the top to keep consistent with the rest of our stores?

src/js/stores/ChronosStore.js
```diff
@@ -0,0 +1,143 @@
...
+  }
+
+  /**
+   * Adds the listener for the specified event
+   * @param {string} eventName
+   * @param {Function} callback
+   * @return {ChronosStore} ChronosStore instance
+   * @override
+   */
+  on(eventName, callback) {
```
rcorral: Why are we creating this inconsistency in the APIs? Why can't we have `addChangeListener` and `removeChangeListener`

src/js/stores/ChronosStore.js
```diff
@@ -0,0 +1,143 @@
...
+    super.on(eventName, callback);
+
+    // Start polling if there is at least one listener
+    if (this.shouldPoll()) {
```
rcorral: I'm confused why this is being done here instead of doing it inside `addChangeListener`

src/js/stores/ChronosStore.js
```diff
@@ -0,0 +1,143 @@
...
+  }
+}
+
+VisibilityStore.addChangeListener(VISIBILITY_CHANGE, handleInactiveChange);
+
+class ChronosStore extends EventEmitter {
+  constructor() {
+    super(...arguments);
+
+    this.data = {
```
rcorral: This is also another inconsistency in how we store data. Why are we doing this?
Is it because otherwise we're storing structs?

src/js/stores/ChronosStore.js
mlunoe: Yeah, I'll move them

mlunoe: done

orlandohohmeier: The `addChangeListener` was (if I got it right) only introduced as there was no way to overwrite the native methods (pre es6 times). Starting/stopping the polling here, makes sure that we cover all cases, even those which aren't using the `addChangeListener`.

rcorral: I mean...we never use `#on` directly. If someone does that means they're trying to go around things on purpose.

src/js/stores/ChronosStore.js
orlandohohmeier: There is no need for the `GetSetMixin`, thus we're simple storing the data in a plain object. Is this a problem?

src/js/stores/ChronosStore.js
```diff
@@ -0,0 +1,117 @@
...
+      jobTree: {id: '/', items: []}
+    };
+
+    // Handle app actions
+    this.dispatcherIndex = AppDispatcher.register(({source, action}) => {
```
rcorral: Could we define this function as we do for other stores so that it's consistent. At the end of the file?

src/js/stores/ChronosStore.js
rcorral: I think the bigger question is, why are we not storing the raw data?

src/js/stores/ChronosStore.js
orlandohohmeier: In my humble opinion, we should either provide a magic way to start polling if someone subscribes to specific event no matter if she is using `addChangeListener`, `addEventListener` or `on`, or we should provide a very explicit method to start the polling (e.g. `addChangeListenerAndStartPolling`).

src/js/stores/ChronosStore.js
mlunoe: Yes, we are not storing structs. Why is this inconsistent? We are storing data like this in other `class`es.

src/js/stores/ChronosStore.js
mlunoe: This is a `class` and we decided to use getters for stores, so this is how to access internal data. This is not inconsistent with what we have done elsewhere: https://github.com/dcos/dcos-ui/blob/master/src/js/stores/DCOSStore.js#L219

src/js/stores/ChronosStore.js
mlunoe: Again, this is how we assign local properties in other classes.

orlandohohmeier: We need to do this in the constructor, as this is an ES6 store.

rcorral: Does it need to be an ES6 store?

rcorral: It's so easy to look for this function at the end of the file as I always know it's going to be there. And I actually do find myself looking through this function quite often.

orlandohohmeier: Why shouldn't it? Whats wrong with adapting the new standards? BTW: I thought we decided to do so.

mlunoe: Alternatively we should pull it out and add it separately after class declaration:
```js
class ChronosStore extends EventEmitter {
  ...
}

ChronosStore.dispatcherIndex =  AppDispatcher.register(function ({source, action}) {
  ...
};
```

But we don't do it like that with other classes. Only for proptypes and defaultprops in components

src/js/stores/ChronosStore.js
rcorral: You're right since it's an ES6 class. But does it need to be?
It's really nice to find the name of this class right away.
Right now there's over 20 stores for which the styles are in one consistent way. You've linked to one store that has done things inconsistently and that's too bad. But let's not keep breaking windows ;)

src/js/stores/ChronosStore.js
rcorral: I'm cool with having ES6 classes for stores when they're all consistent. It should be pretty low effort to go through all stores and make them consistent. So let's do that and then this totally makes sense.

src/js/stores/ChronosStore.js
mlunoe: I don't understand you comment (https://github.com/dcos/dcos-ui/pull/246#discussion_r64953391) @rafa. We are storing raw data. This is a good default as we can assume the root group, but no jobs

rcorral: We should be storing the raw data IMO because that gets sent to the omni store too when we use the getset mixin. We should not have some stores that don't push their data there.

rcorral: @mlunoe you're right on the storing data, I missed that. Again, probably because I have to be looking for other things rather than the same style of code that the other stores use. It's really important we keep things consistent.

orlandohohmeier: We're storing raw data in this store. I don't see any issue here.

src/js/stores/ChronosStore.js
orlandohohmeier: I'm ok with updating the rest of the stores, but thought that the prefered rollout strategy for components and stores is to implement all new ones in ES6. And to be honest, I think we do have other issues with inconsistency then complete stores that apply to all ES6 compatible style rules. Take this method definition for example, which is inconsistent with how stuff is defined in that store https://github.com/dcos/dcos-ui/blob/master/src/js/stores/AuthStore.js#L59

rcorral: I'm not sure what you're pointing out with your example. Maybe if you could explain a little more.

Can we agree that there's code which has made it through review that later we discovered to be inconsistent. This is quite confusing to developers, specially if I'm going to create a new Store, which style do I use? We should manage our technical debt and not create code which is inconsistent.
So please, let us do one of two things:
* Keep this store as it is with these new styles and refactor the rest of the stores to match this style now.
* Let's change the styles of this store to match the rest of our stores and at a later point refactor all stores at once to this new style.

orlandohohmeier: I added the example as I wanted to illustrate that we do have more urgent problems with inconsistency than a new store applying the new ES6 style guide. As I wrote, I thought that the prefered rollout strategy for the ES6 classes and such is to implement new components and stores(!) following the ES6 style guide. Thus I didn't thought that there is anything wrong with this store.

I'll estimate the effort to update all the store and use that estimate to find a good solution for this "inconsistency". 

No matter how we're proceeding with this store, I'd like you to add an contribution file or something similar to document the latest decisions on how to rollout he latest ES6 features, and your plan on how to take care of all the inconsistency that is already present, as we probably wasted time doing this store in the new ES6 style and I like us to prevent this in the future.

orlandohohmeier: The effort needed to update all the stores seems low, thus I'm going to update them as discussed.

pierlo-upitup: I'm just chiming in to say that as @orlandohohmeier wrote, we seem to have a different strategy regarding components, @rcorral. It seems like all new React components are supposed to be ES6 classes (which is a PR-blocking requirement) even without a full-on refactoring process being required for preexisting non-ES6 components. 

Without having clear "consistency" guidelines written in a document these comments start feeling very arbitrary and ultimately become the cause of confusion, frustration and worst of all they waste our time. Let's please fix this once and for all.

orlandohohmeier:
After agreeing on the new store pattern, is there anything else that stops us from merging this?

src/js/stores/ChronosStore.js
```diff
@@ -0,0 +1,117 @@
...
+  }
+}
+
+function handleInactiveChange() {
+  let isInactive = VisibilityStore.get('isInactive');
+  if (isInactive) {
+    stopPolling();
+  }
+
+  if (!isInactive && ChronosStore.shouldPoll()) {
```
philipnrmn: Apologies for the comment on a merged PR, but I don't think there is an instantiated ChronosStore in scope here, so I think this will cause reference errors.

orlandohohmeier: Fixed in https://github.com/dcos/dcos-ui/pull/295

PR #250, opened 2016-05-24
Introduce job struct

orlandohohmeier:
Introduce basic job struct – this struct will be used to type and abstract the Chronos API response.

### DIFF
--- a/src/js/structs/Job.js
+++ b/src/js/structs/Job.js
@@ -5,7 +5,7 @@ module.exports = class Job extends Item {
     super(...arguments);
 
     const id = this.getId();
-    if (id !== '/' && !id.startsWith('/') || id.endsWith('/')) {
+    if (id !== '/' && (!id.startsWith('/') || id.endsWith('/'))) {
       throw new Error(`Id (${id}) must start with a leading slash ("/") and should not end with a slash, except for root id wich is only a slash`);
     }
   }
--- a/src/js/structs/Job.js
+++ b/src/js/structs/Job.js
@@ -13,7 +13,7 @@ module.exports = class Job extends Item {
   }
 
   getId() {
-    return this.get('id') || '';
+    return this.get('id') || '/';
   }
 
   getName() {
--- a/src/js/structs/Job.js
+++ b/src/js/structs/Job.js
@@ -1,6 +1,16 @@
 import Item from './Item';
 
 module.exports = class Job extends Item {
+  constructor(options = {}) {
+    super(...arguments);
+
+    const id = this.getId();
+    if (!id.startsWidth('/') || id.endsWidth('/')) {
+      throw new Error(
+        'Id must start with a leading slash ("/") and should not end with a slash.'
+      );
+    }
+  }
 
   getId() {
     return this.get('id') || '';
--- a/src/js/structs/Job.js
+++ b/src/js/structs/Job.js
@@ -1,7 +1,7 @@
 import Item from './Item';
 
 module.exports = class Job extends Item {
-  constructor(options = {}) {
+  constructor() {
     super(...arguments);
 
     const id = this.getId();
--- a/src/js/structs/Job.js
+++ b/src/js/structs/Job.js
@@ -5,10 +5,8 @@ module.exports = class Job extends Item {
     super(...arguments);
 
     const id = this.getId();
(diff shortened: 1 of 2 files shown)


mesosphere-web:
Reviewer: @Poltergeist

src/js/structs/Job.js
```diff
@@ -0,0 +1,13 @@
+import Item from './Item';
+
+module.exports = class Job extends Item {
+
+  getId() {
+    return this.get('id') || '';
+  }
+
+  getName() {
+    return this.getId().split('/').pop();
```
aldipower: This could probably lead to unwanted behavior if the id has a dangling `/`.
```
-> '/a/b/c/'.split('/')
<- Array [ "", "a", "b", "c", "" ]
```

src/js/structs/__tests__/Job-test.js
```diff
@@ -0,0 +1,29 @@
+let Job = require('../Job');
+
+describe('Job', function () {
```
aldipower: Maybe adding a test with a dangling slash makes sense?

pierlo-upitup:
After @aldipower 's comments are addressed, this can be automerged.

src/js/structs/Job.js
```diff
@@ -0,0 +1,21 @@
+import Item from './Item';
+
+module.exports = class Job extends Item {
+  constructor(options = {}) {
+    super(...arguments);
+
+    const id = this.getId();
+    if (id !== '/' && (!id.startsWith('/') || id.endsWith('/'))) {
+      throw new Error(`Id (${id}) must start with a leading slash ("/") and should not end with a slash, except for root id wich is only a slash`);
```
orlandohohmeier: `which` not `which`  #nit

src/js/structs/Job.js
```diff
@@ -0,0 +1,21 @@
+import Item from './Item';
+
+module.exports = class Job extends Item {
+  constructor() {
+    super(...arguments);
+
+    const id = this.getId();
+    if (id !== '/' && (!id.startsWith('/') || id.endsWith('/'))) {
+      throw new Error(`Id (${id}) must start with a leading slash ("/") and should not end with a slash, except for root id which is only a slash`);
```
aldipower: @mlunoe 
I think it's not a good idea to force this and throw an error.
There is this small possibility that the API maybe returns an ID with a slash at the end.
In this case our whole application will fail now. 
Also I think the `id` shouldn't validated in the constructor of `Job`. I don't expect it at this point.

src/js/structs/Job.js
```diff
@@ -0,0 +1,21 @@
...
+    }
+  }
+
+  getId() {
+    return this.get('id') || '/';
+  }
+
+  getName() {
+    return this.getId().split('/').pop();
```
aldipower: @mlunoe 
My earlier comments pointed more in the direction, that `getName` should be able to handle `id`s with an ending slash, so that the correct name will be returned anyway.
This makes the application more fail safe. Now with the thrown exception, the opposite is the case.

src/js/structs/Job.js
orlandohohmeier: The backend is using the same [PathId](https://github.com/mesosphere/marathon/blob/master/src/main/scala/mesosphere/marathon/state/PathId.scala) as Marathon and as far as I can tell already strips dangling slashs for us. 
In the very rare case we actually receive malformed data we should handle it in the same way we handle malformed JSON. Let's catch these errors in the Chronos store and report them to the user instead of introducing a ton of complex regex patterns in an attempt to fix a broken backend response. 

/cc @gkleiman

mlunoe: Your comment actually spawned a larger conversation about what we should and should not do in the structs. We decided to move the whole building of the tree to a 'reducer'-like function that is being used in the actions before the struct is being created. That way we can handle any malformed data gracefully before handing it off to the struct:
https://github.com/dcos/dcos-ui/pull/246/files#diff-976ccf0fd4e369643bb74528d0c47785R20

aldipower: @orlandohohmeier @mlunoe 
Yeah, cool, doing it the ChronosStore is a good idea imho! Is that also the way it could work with the Services?

BTW, never trust the backend! ;)

PR #251, opened 2016-05-24
Introduce job tree struct

orlandohohmeier:
---
ℹ️   _This branch depends on changes that will be introduced with #250, thus it will fail until the respective branch is merged._

---

Introduce basic job tree struct – this struct will be used to type and abstract the Chronos API response. It provide the needed functionality to parse and create sub trees (if needed), to insert jobs in the correct location based on id/path matching.

### DIFF
--- a/src/js/structs/JobTree.js
+++ b/src/js/structs/JobTree.js
@@ -19,6 +19,12 @@ module.exports = class JobTree extends Tree {
     this.id = '/';
 
     if (typeof id == 'string') {
+      if (id !== '/' && (!id.startsWith('/') || id.endsWith('/'))) {
+        throw new Error(`Id (${id}) must start with a leading slash ("/") ` +
+          'and should not end with a slash, except for root id which is only ' +
+          'a slash.');
+      }
+
       this.id = id;
     }
 
--- a/src/js/structs/JobTree.js
+++ b/src/js/structs/JobTree.js
@@ -30,7 +30,7 @@ module.exports = class JobTree extends Tree {
 
     // Parse and add items to the current tree
     if (Array.isArray(items)) {
-      items.forEach((item)=> {
+      items.forEach((item) => {
         // Create instances of JobTree for tree like items and add them
         if ((item.items != null && Array.isArray(item.items))
           && !(item instanceof JobTree)) {
@@ -63,11 +63,12 @@ module.exports = class JobTree extends Tree {
       throw new TypeError('item is neither an instance of Job nor JobTree');
     }
 
+    const id = this.getId();
     const itemId = item.getId();
 
-    if (!itemId.startsWith(this.getId())) {
+    if (!itemId.startsWith(id)) {
       throw new Error(
-        `item id (${itemId}) doesn't match tree id (${this.getId()})`
+        `item id (${itemId}) doesn't match tree id (${id})`
       );
     }
 
@@ -82,7 +83,7 @@ module.exports = class JobTree extends Tree {
     const [parentId] = itemId.match(/\/.*?(?=\/?[^/]+\/?$)/);
 
     // Add item to the current tree if it's the actual parent tree
-    if (this.getId() === parentId) {
+    if (id === parentId) {
       super.add(item);
       return;
     }
--- a/src/js/structs/JobTree.js
+++ b/src/js/structs/JobTree.js
@@ -1,0 +1,112 @@
+import Job from './Job';
+import Tree from './Tree';
+
+module.exports = class JobTree extends Tree {
+  /**
+   * (Chronos) JobTree
+   * @param {{
+   *          id:string,
+   *          items:array<({id:string, items:array}|*)>,
+   *          filterProperties:{propertyName:(null|string|function)}
+   *        }} options
+   * TODO: DCOS-7369 - Add remove item method
+   * @constructor
+   * @struct
+   */
+  constructor({id, items, filterProperties} = {}) {
+    super({filterProperties});
+
+    this.id = '/';
+
+    if (typeof id == 'string') {
+      this.id = id;
+    }
+
+    // Parse and add items to the current tree
+    if (Array.isArray(items)) {
+      items.forEach((item)=> {
+        // Create instances of JobTree for tree like items and add them
+        if ((item.items != null && Array.isArray(item.items))
+          && !(item instanceof JobTree)) {
+          this.add(new this.constructor(
+            Object.assign({filterProperties: this.getFilterProperties()}, item)
+          ));
+
+          return;
+        }
+
+        // Convert items into instance of Job and add them to the current tree
+        if (!(item instanceof Job) && !(item instanceof JobTree)) {
+          this.add(new Job(item));
+
+          return;
+        }
+
+        this.add(item);
+      })
+    }
+  }
+
+  /**
+   * Add item - This method will create sub trees if needed to insert the item
+   * at the correct location (based on id/path matching).
+   * @param {JobTree|Job} item
+   */
+  add(item) {
+    if (!(item instanceof Job || item instanceof JobTree)) {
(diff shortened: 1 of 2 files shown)


mesosphere-web:
Reviewer: @mlunoe

src/js/structs/JobTree.js
```diff
@@ -0,0 +1,111 @@
...
+
+module.exports = class JobTree extends Tree {
+  /**
+   * (Chronos) JobTree
+   * @param {{
+   *          id:string,
+   *          items:array<({id:string, items:array}|*)>,
+   *          filterProperties:{propertyName:(null|string|function)}
+   *        }} options
+   * TODO (orlandohohmeier): Add remove item method
```
mlunoe: Let's add the JIRA task number here

src/js/structs/JobTree.js
```diff
@@ -0,0 +1,111 @@
...
+    super({filterProperties});
+
+    this.id = '/';
+
+    if (typeof id == 'string') {
+      this.id = id;
+    }
+
+    // Parse and add items to the current tree
+    if (Util.isArray(items)) {
```
mlunoe: We should remove this Util. We are using `Array.isArray` now

src/js/structs/JobTree.js
```diff
@@ -0,0 +1,111 @@
...
+   * @param {JobTree|Job} item
+   */
+  add(item) {
+    if (!(item instanceof Job || item instanceof JobTree)) {
+      throw new TypeError('item is neither an instance of Job nor JobTree');
+    }
+
+    if (!item.getId().startsWith(this.getId())) {
+      throw new Error(
+        `item id (${item.getId()})  doesn't match tree id (${this.getId()})`
```
mlunoe: I think you have an extra space here in between `)` and `doesn't`

src/js/structs/JobTree.js
```diff
@@ -0,0 +1,110 @@
...
+   * Add item - This method will create sub trees if needed to insert the item
+   * at the correct location (based on id/path matching).
+   * @param {JobTree|Job} item
+   */
+  add(item) {
+    if (!(item instanceof Job || item instanceof JobTree)) {
+      throw new TypeError('item is neither an instance of Job nor JobTree');
+    }
+
+    if (!item.getId().startsWith(this.getId())) {
```
pierlo-upitup: How about caching `item.getId()` so we can reuse it in the next 3 occurences?

src/js/structs/JobTree.js
```diff
@@ -0,0 +1,110 @@
...
+
+    // Check if the item is already present in the tree (including sub trees)
+    if (this.findItemById(item.getId())) {
+      // TODO: DCOS-7370 - Update item (type Job) data
+      return;
+    }
+
+    // Get the parent id (e.g. /group) by matching every thing but the item
+    // name including the preceding slash "/" (e.g. /id).
+    let [parentId] = item.getId().match(/\/.*?(?=\/?[^/]+\/?$)/);
```
pierlo-upitup: are we protected against `.getId()` returning a string with no leading `/`? for example `group/sub/test` here would become `/sub`

src/js/structs/JobTree.js
```diff
@@ -0,0 +1,110 @@
...
+
+    // Add item to the current tree if it's the actual parent tree
+    if (this.getId() === parentId) {
+      super.add(item);
```
pierlo-upitup: ❓

src/js/structs/__tests__/JobTree-test.js
```diff
@@ -0,0 +1,115 @@
...
+          new JobTree({id: '/group/foo', items: []}),
+          {id: '/group/bar', items: []},
+          {id: '/group/alpha'},
+          new Job({id: '/group/beta'}),
+          {id: '/group/wibble/wobble'}
+        ],
+      });
+    });
+
+    it('defaults id to root tree (groups) id', function () {
```
pierlo-upitup: Let's try to have human readable descriptions 😄 . How about `it(creates the root group if no items are supplied)`?

src/js/structs/__tests__/JobTree-test.js
```diff
@@ -0,0 +1,115 @@
...
+      let tree = new JobTree({items: []});
+      expect(tree.getId()).toEqual('/');
+    });
+
+    it('sets correct tree (groups) id', function () {
```
pierlo-upitup: how about `it(creates the group ids correctly)` and since we're at it how about adding 
```
expect(this.instance.getId()).toEqual('/group');
expect(this.instance.getItems()[0].getId()).toEqual('/group/foo');
```
?

pierlo-upitup: NVM – I see you perform similar checks later.

pierlo-upitup:
A few comments and failing tests needs fixing. 
```
 Runtime Error
Error: Cannot find module '../Job' from 'JobTree-test.js'
    at Runtime._resolveNodeModule (/Users/pierlo/Development/dcos-ui/node_modules/jest-cli/src/Runtime/Runtime.js:451:11)
    at Object.<anonymous> (/Users/pierlo/Development/dcos-ui/src/js/structs/__tests__/JobTree-test.js:1:169)
    at jasmine2 (/Users/pierlo/Development/dcos-ui/node_modules/jest-jasmine2/src/index.js:231:16)
    at process._tickCallback (node.js:406:9)
```

orlandohohmeier:
@pierlo-upitup The test fail at the moment as this depends on stuff that will be introduced with #250.

src/js/structs/__tests__/JobTree-test.js
orlandohohmeier: This test verifies that `JobTree` is using the default (root tree) group id, if no id is provided.

src/js/structs/JobTree.js
orlandohohmeier: Good catch! We will probably never ever get such data for the API, but I'll harden our implementation to make sure we're not creating wrong structures based on corrupted data.

src/js/structs/JobTree.js
orlandohohmeier: Yeah, should make it a little easier to read the lines.

src/js/structs/JobTree.js
```diff
@@ -0,0 +1,110 @@
...
+    if (this.findItemById(item.getId())) {
+      // TODO: DCOS-7370 - Update item (type Job) data
+      return;
+    }
+
+    // Get the parent id (e.g. /group) by matching every thing but the item
+    // name including the preceding slash "/" (e.g. /id).
+    let [parentId] = item.getId().match(/\/.*?(?=\/?[^/]+\/?$)/);
+
+    // Add item to the current tree if it's the actual parent tree
```
orlandohohmeier: We're trying to find the parent of an item (`Job` or `JobTree`), by extracting the`parentId` from the `itemId`. We're adding it to the current tree if the `parentId` matches the current `JobTree` id, otherwise we're trying to find the parent (`findItemById`) or create a new parent (`JobTree`) and add the item to that one.
**TLDR;** This method recursively creates all the parents needed to insert an item at the correct spot (based on it's id).

src/js/structs/JobTree.js
orlandohohmeier: This call adds an item to the current list of items.

src/js/structs/JobTree.js
```diff
@@ -0,0 +1,118 @@
...
+          'and should not end with a slash, except for root id which is only ' +
+          'a slash.');
+      }
+
+      this.id = id;
+    }
+
+    // Parse and add items to the current tree
+    if (Array.isArray(items)) {
+      items.forEach((item)=> {
```
mlunoe: #nit missing space after `)` and before `=>`

PR #254, opened 2016-05-24
Updates to Breadcrumb

daiyiwhale:
A number of updates to `Breadcrumb` based on my attempt to use `Breadcrumb`

- trimmed down `buildCrumb` default function to keep parsing minutia external to that function so overriding this function with a prop requires less repeated code
- explicitly set the `buildCrumb` default function in `defaultProps`
- added `buildCrumbCustomParams` prop to supply custom parameters to `buildCrumb` function for custom labels. This prop also helps trigger `shouldComponentUpdate` when custom params are updated

### DIFF
--- a/src/js/components/Breadcrumb.js
+++ b/src/js/components/Breadcrumb.js
@@ -1,7 +1,7 @@
 import classNames from 'classnames';
 import DeepEqual from 'deep-equal';
 import {Link} from 'react-router';
-import React from 'react';
+import React, {PropTypes} from 'react';
 import ReactDOM from 'react-dom';
 
 import IconChevron from './icons/IconChevron';
@@ -11,7 +11,33 @@ const COLLAPSE_BUFFER = 12;
 const LAST_ITEM_OFFSET = 150; // Difference between scrollWidth and outerWidth
 const PADDED_ICON_WIDTH = 38; // Width of icon + padding
 
-const {PropTypes} = React;
+/**
+ * Builds a crumb or multiple crumbs from the label and matching route
+ * @param {String} label Url path part
+ * @param {String} route Matching route name
+ * @param {Object} routeParams Router routeParams Object
+ * @param {Bool}   labelIsParam Boolean indicating if the label represents a
+ * route parameter.
+ * @return {Array} Array of crumbs containing the label and route.
+ * The returned label is the text shown for the crumb. The route is
+ * the route name which will be set on the Link-to attribute.
+ * Each returned crumb will be spliced into the resulting breadcrumb.
+ */
+const buildCrumb = function (label, route, routeParams, labelIsParam) {
+  // Extract param value as label e.g. (:userID -> Foo)
+  if (labelIsParam) {
+    // Decode value
+    label = decodeURIComponent(routeParams[label]);
+  } else {
+    // Split Label on space and - then capitalize each word and join.
+    // /foo-bar/ --> Foo Bar
+    label = StringUtil.idToTitle([label], ['-']);
+  }
+
+  // Can return multiple route objects to get rendered in crumbs.
+  // Useful for splitting a route parameter like a file path.
+  return [{label, route}];
+}
 
 class Breadcrumb extends React.Component {
   constructor() {
@@ -46,6 +72,12 @@ class Breadcrumb extends React.Component {
   }
 
   shouldComponentUpdate(nextProps, nextState) {
+    let {buildCrumbCustomParams} = this.props;
+
+    if (!DeepEqual(nextProps.buildCrumbCustomParams, buildCrumbCustomParams)) {
+      return true;
+    }
+
     if (nextState.availableWidth == null || nextState.expandedWidth == null) {
       this.updateDimensions();
 
@@ -153,11 +185,11 @@ class Breadcrumb extends React.Component {
     return this.context.router.getCurrentParams();
   }
 
-  getCrumb(label, route, params, key) {
+  getCrumb(label, route, key) {
     return (
       <li key={key}>
         <Link to={route}
-            params={params}
+            params={this.getCurrentRouteParams()}
             title={label}>
           {label}
         </Link>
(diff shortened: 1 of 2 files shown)


mesosphere-web:
Reviewer: @aldipower

src/js/components/Breadcrumb.js
```diff
@@ -183,56 +215,14 @@ class Breadcrumb extends React.Component {
...
-    return [{label, route, params}];
-  }
 
   renderCrumbsFromRoute() {
-    let {buildCrumb, shift} = this.props;
-
-    if (buildCrumb == null) {
-      buildCrumb = this.buildCrumb;
-    }
+    let {buildCrumb, buildCrumbCustomParams, shift} = this.props;
```
kennyt: won't the `buildCrumb` at the top of the file never be used because of this right here?

daiyiwhale: yep!

src/js/components/Breadcrumb.js
```diff
@@ -46,6 +72,12 @@ class Breadcrumb extends React.Component {
   }
 
   shouldComponentUpdate(nextProps, nextState) {
+    let {buildCrumbCustomParams} = this.props;
+
+    if (!DeepEqual(nextProps.buildCrumbCustomParams, buildCrumbCustomParams)) {
```
MatApple: Don't think we need this right? The last part of this method already checks if props have changed

src/js/components/Breadcrumb.js
```diff
@@ -242,18 +232,23 @@ class Breadcrumb extends React.Component {
...
+        label = label.slice(1);
+      }
 
       let crumbObjects = buildCrumb(
         label,
         name,
         params,
-        labelIsParam
+        labelIsParam,
+        buildCrumbCustomParams
```
MatApple: I think we should put this argument in the `buildCrumbs` function at the top and we should use it in that default `buildCrumbs` method.

src/js/components/Breadcrumb.js
daiyiwhale: we need it because `nextState.availableWidth == null || nextState.expandedWidth == null` will return `false` before the last part of this method has a chance to return `true`

src/js/components/Breadcrumb.js
daiyiwhale: Oops, totally forgot about this. This is what happens when I try to build a theoretical component instead of implementing it in parallel (':

src/js/components/Breadcrumb.js
MatApple: Ahh Ok, so maybe we should just move the entire DeepEqual up then?

src/js/components/Breadcrumb.js
daiyiwhale: done

src/js/components/Breadcrumb.js
daiyiwhale: I thought there might have been a reason you had `nextState.availableWidth == null || nextState.expandedWidth == null ` first, to circuit some process?

MatApple: No it just defaulted to this order :) If it works in the other order, lets do it!

src/js/components/Breadcrumb.js
```diff
@@ -124,43 +120,37 @@ class Breadcrumb extends React.Component {
...
     return this.context.router.getCurrentParams();
   }
 
-  getCrumb(label, route, params, key) {
-    return (
-      <li key={key}>
-        <Link to={route}
-            params={params}
+  getCrumb(crumb, key) {
+    let content = crumb;
```
aldipower: This assignment looks a little bit weird because, you still using `crumb` directly below.
I think you can get rid of `content`.

src/js/components/Breadcrumb.js
```diff
@@ -124,43 +120,37 @@ class Breadcrumb extends React.Component {
...
+    if (crumb.hasOwnProperty('route')) {
+      route = crumb.route;
+    }
+
+    if (crumb.hasOwnProperty('label')) {
+      label = crumb.label;
+    }
+
+    if (route) {
+      content = (
```
aldipower: You could override `crumb` here, so you can get rid of the `content`-var.

aldipower: Or isn't that possible because `crumb` is passed in by reference?

src/js/components/Breadcrumb.js
```diff
@@ -183,86 +173,43 @@ class Breadcrumb extends React.Component {
...
     }
-    // Can return multiple route objects to get rendered in crumbs.
-    // Useful for splitting a route parameter like a file path.
-    return [{label, route, params}];
-  }
 
-  renderCrumbsFromRoute() {
-    let {buildCrumb, shift} = this.props;
+    let crumbConfiguration = route.buildBreadCrumb();
+    let crumbs = crumbConfiguration.getCrumbs(this.context.router);
```
aldipower: `this.context.router` is used twice in this function and could be cached.

aldipower: I have no other solution to offer, but I find these lines slightly confusing.

On line 185 I "build a breadcrumb", then on the next line I am using this breadcrumb to "get the crumbs".
This is a little bit weird. Naming problem I think.

src/js/components/Breadcrumb.js
```diff
@@ -183,86 +173,43 @@ class Breadcrumb extends React.Component {
...
-        if (pathIndex !== paths.length - 1
-          || crumbIndex !== crumbObjects.length - 1) {
+    return crumbs.reduce((memo, crumb, crumbIndex) => {
+      memo.push(this.getCrumb(crumb, ++crumbKey));
 
-          crumbs.push(this.getCrumbDivider(crumbKey++));
-        }
-      });
+      if (crumbs.length - 1 !== crumbIndex) {
+        memo.push(this.getCrumbDivider(++crumbKey));
```
aldipower: By passing a something like `"divider"+crumbIndex` as a key, there would be no need for a `crumbKey`-var anymore.

aldipower:
Thx! This makes the things easier!
You removed more lines then you added, this is in the most cases good. :)

src/js/components/Breadcrumb.js
rcorral: Opted not to do this, as it's just a different way of doing things.

src/js/components/Breadcrumb.js
rcorral: Done.

src/js/components/Breadcrumb.js
rcorral: Done.

src/js/components/Breadcrumb.js
rcorral: Done.

rcorral:
retest this please

src/js/components/Breadcrumb.js
aldipower: It saves a variable, which safes memory and lines of code.
Of course it is different. :)

rcorral:
retest this please

colin-msphere:
retest this please

rcorral:
Having @daiyiwhale review this PR because I've basically refactored a bunch of it.

PR #255, opened 2016-05-25
[2/8] Deployments Table - Don't use ServicesList to list services

philipnrmn:
`ServicesList` is actually a list used for instances of `Framework`, not `Service`. In order to avoid casting our non-framework services to frameworks, let's use an array to hold services on deployments instead of a `List`.

### DIFF
--- a/src/js/stores/DCOSStore.js
+++ b/src/js/stores/DCOSStore.js
@@ -13,7 +13,6 @@ import Framework from '../structs/Framework';
 import MarathonStore from './MarathonStore';
 import MesosSummaryStore from './MesosSummaryStore';
 import NotificationStore from './NotificationStore';
-import ServicesList from '../structs/ServicesList';
 import ServiceTree from '../structs/ServiceTree';
 import SummaryList from '../structs/SummaryList';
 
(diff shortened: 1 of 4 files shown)


mesosphere-web:
Reviewer: @jfurrow

kennyt:
do we want to create a struct for this?

orlandohohmeier:
I think, we should repurpose the `ServiceList` to handle all kinds of services and not only `Frameworks, but that's out of scope of this PR. I'm fine with using a plain array here, especially as there is no parsing and stuff needed. Let's create a Jira and schedule some time to properly refactor the service list.

philipnrmn:
@orlandohohmeier 👍  

`ServicesList` is a specialised list  which casts all its members to `Framework` - this was reasonable in previous versions of DCOS since the only displayed services were all frameworks, and I think it's still useful in things like the dashboard services health panel. 

I'd therefore like to keep it more or less as-is, but rename it so that its role is clear. Then we can introduce a `ServicesList` which actually is a list of services. 

I also think this is a good case for a Jira ticket to be fixed later on, so I'll go ahead and create one now.

MatApple:
+1 for @philipnrmn doing this - thanks Phil!

+1 for what @orlandohohmeier has proposed - that we make the ServicesList handle Frameworks and Apps.

orlandohohmeier:
Let's :shipit:

PR #259, opened 2016-05-25
Add date to relative time

kennyt:


mesosphere-web:
Reviewer: @philipnrmn

src/js/utils/DateUtil.js
```diff
@@ -21,6 +21,10 @@ const DateUtil = {
     return moment.unix(ms).fromNow();
   },
 
+  dateToRelativeTime: function (date) {
```
rcorral: Would be nice to have a test and to also have some JDoc saying that the `date` param needs to be an instance of `Date`

kennyt: alright done

philipnrmn:
I'm OK with the implementation, but what's the intention of this PR? Why not just call the `moment` method inline?

philipnrmn:
I'm merging this, but I'm still curious about what it's for.

kennyt:
@philipnrmn to test it and also if we ever wanted to stop using moment and use something else, we'd have to change it in one place

PR #260, opened 2016-05-25
Allows adding additional configurations to routes.

rcorral:
We want to be able to attach additional configuration options to routes so that different components in the system can make use of them. This will allow for that.


mesosphere-web:
Reviewer: @Poltergeist

rcorral:
Merging, as I need this to continue working.

PR #263, opened 2016-05-26
Adjust task detail styles

aldipower:
Before:
td-before

After:
td-after


rcorral:
retest this please

aldipower:
done

pierlo-upitup:
cc @leemunroe what do you think?

leemunroe:
Wow, that's a lot of endpoints. What if we come up with a show more/show less solution to help tidy this up?

PR #265, opened 2016-05-26
Use timefield for date

aldipower:
Display the local time in the task detail list

### DIFF
--- a/src/js/components/MarathonTaskDetailsList.js
+++ b/src/js/components/MarathonTaskDetailsList.js
@@ -58,6 +58,14 @@ class MarathonTaskDetailsList extends React.Component {
     return task.status;
   }
 
+  getTimeField(time) {
+    return (
+      <time dateTime={time} title={time}>
+        {new Date(time).toLocaleString()}
+      </time>
+    );
+  }
+
   getMarathonTaskDetailsDescriptionList(task) {
     if (task == null) {
       return null;
@@ -68,8 +76,8 @@ class MarathonTaskDetailsList extends React.Component {
       'Ports': task.ports,
       'Endpoints': this.getTaskEndpoints(task),
       'Status': this.getTaskStatus(task),
-      'Staged at': task.stagedAt,
-      'Started at': task.startedAt,
+      'Staged at': this.getTimeField(task.stagedAt),
+      'Started at': this.getTimeField(task.startedAt),
       'Version': task.version
     };
 


rcorral:
retest this please

aldipower:
done

src/js/components/MarathonTaskDetailsList.js
```diff
@@ -58,6 +58,14 @@ class MarathonTaskDetailsList extends React.Component {
     return task.status;
   }
 
+  getTimeField(time) {
+    return (
+      <time dateTime={time} title={time}>
+        {new Date(time).toLocaleString()}
```
philipnrmn: Should we be using locale strings here, or does it make more sense to use an ISO string? Or do we want something from `moment`?

aldipower: This is directly taken from Marathon UI and I remember it was a requirement from a user to do it with locale strings.

philipnrmn: Ça plane pour moi. We can always adjust it later.

@leemunroe does that work for you?

leemunroe: @philipnrmn Hmm can you give me more context about what's going on here? Screenshot?

philipnrmn: @leemunroe sure thing. 

task detail panel

This is about how the `Staged At`/`Started At` values should be formatted. The data from the server is an [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) combined date and time (like the `Version` value), which isn't so readable. 

The method which Marathon uses and which Felix is using here
* Formats the date according to the user's configuration
* Converts the date to the user's local timezone

I wondered whether this is still our preferred way of doing things. 

NB this discussion shouldn't block merge on this PR.

leemunroe: Looks good, happy to keep consistent with Marathon for now. This might be something that needs a deeper discussion outside of this PR. `How do we want to handle dates and times?`

We have a few options bouncing around. User time zone, server time zone, UTC. Absolute or relative. I'm not sure how consistent we are. 

One idea is showing the alternate on hover (either tooltip or title attribute). Another is clicking a date/time to switch. cc @air

air: I love the idea of clicking a timestamp to view alternative formats, since it's easy then to copy/paste.
We should be able to draw a short set of rules for how we'll be consistent with displaying timestamps.

PR #267, opened 2016-05-27
add create modal optional section

Poltergeist:
---
It is depended on #212 (**merged**)

---

This PR adds the logic for the optional section and also introduces the fields for that section to the modal. 

image

### DIFF
--- a/src/js/constants/ServiceSchema.js
+++ b/src/js/constants/ServiceSchema.js
@@ -43,7 +43,7 @@ let SERVICE_SCHEMA = {
           title: 'Instances',
           type: 'number',
           getter: function (service) {
-            return (service.getInstancesCount() || 0);
+            return (service.getInstancesCount() || 0) + '';
           }
         },
         cmd: {
--- a/src/js/constants/ServiceSchema.js
+++ b/src/js/constants/ServiceSchema.js
@@ -140,6 +140,13 @@ let SERVICE_SCHEMA = {
             return service.getAcceptedResourceRoles() &&
               service.getAcceptedResourceRoles().join(', ');
           }
+        },
+        user: {
+          title: 'user',
+          type: 'string',
+          getter: function (service) {
+            return service.getUser();
+          }
         }
       }
     }
--- a/src/js/constants/ServiceSchema.js
+++ b/src/js/constants/ServiceSchema.js
@@ -104,6 +104,19 @@ let SERVICE_SCHEMA = {
           getter: function (service) {
             return service.getExecutor();
           }
+        },
+        uris: {
+          title: 'URIs',
+          type: 'string',
+          description: 'Comma-separated list of valid URIs.',
+          getter: function (service) {
+            if (!service.getFetch()) {
+              return null;
+            }
+            return service.getFetch().map(function (item) {
+              return item.uri;
+            }).join(', ');
+          }
         }
       }
     }
--- a/src/js/constants/ServiceSchema.js
+++ b/src/js/constants/ServiceSchema.js
@@ -129,6 +129,17 @@ let SERVICE_SCHEMA = {
                 return item.join(':');
               }).join(',')
           }
+        },
+        acceptedResourceRoles: {
+          title: 'Accepted Resource Roles',
+          type: 'string',
+          description: 'Comma-separated list of resource roles. Marathon ' +
+          'considers only resource offers with roles in this list for ' +
+          'launching tasks of this app.',
+          getter: function (service) {
+            return service.getAcceptedResourceRoles() &&
+              service.getAcceptedResourceRoles().join(', ');
+          }
         }
       }
     }
--- a/src/js/constants/ServiceSchema.js
+++ b/src/js/constants/ServiceSchema.js
@@ -92,6 +92,20 @@ let SERVICE_SCHEMA = {
         }
       },
       required: []
+    },
+    'Optional': {
+      type: 'object',
+      properties: {
+        executor: {
+          title: 'Executor',
+          type: 'string',
+          description: 'Executor must be the string \'//cmd\', a string ' +
+          'containing only single slashes (\'/\'), or blank.',
+          getter: function (service) {
+            return service.getExecutor();
+          }
+        }
+      }
     }
   },
   required: [
--- a/src/js/constants/ServiceSchema.js
+++ b/src/js/constants/ServiceSchema.js
@@ -2,10 +2,10 @@
 import React from 'react';
 /* eslint-enable no-unused-vars */
 
-import General from './service-schema/general';
-import Optional from './service-schema/optional';
+import General from './service-schema/General';
+import Optional from './service-schema/Optional';
 
-let SERVICE_SCHEMA = {
+let ServiceSchema = {
   type: 'object',
   properties: {
     General: General,
@@ -49,4 +49,4 @@ let SERVICE_SCHEMA = {
   ]
 };
 
-module.exports = SERVICE_SCHEMA;
+module.exports = ServiceSchema;
--- a/src/js/constants/ServiceSchema.js
+++ b/src/js/constants/ServiceSchema.js
@@ -22,28 +22,28 @@ let SERVICE_SCHEMA = {
           description: 'The amount of CPUs which are used for the service',
           type:'number',
           getter: function (service) {
-            return (service.getCpus() || '');
+            return `${service.getCpus() || ''}`;
           }
         },
         mem: {
           title: 'Mem (MiB)',
           type: 'number',
           getter: function (service) {
-            return (service.getMem() || '');
+            return `${service.getMem() || ''}`;
           }
         },
         disk: {
           title: 'Disk (MiB)',
           type: 'number',
           getter: function (service) {
-            return (service.getDisk() || '');
+            return `${service.getDisk() || ''}`;
           }
         },
         instances: {
           title: 'Instances',
           type: 'number',
           getter: function (service) {
-            return (service.getInstancesCount() || 0) + '';
+            return `${service.getInstancesCount() || 0}`;
           }
         },
         cmd: {
@@ -142,7 +142,7 @@ let SERVICE_SCHEMA = {
           }
         },
         user: {
-          title: 'user',
+          title: 'User',
           type: 'string',
           getter: function (service) {
             return service.getUser();
--- a/src/js/constants/ServiceSchema.js
+++ b/src/js/constants/ServiceSchema.js
@@ -2,64 +2,13 @@
 import React from 'react';
 /* eslint-enable no-unused-vars */
 
+import General from './service-schema/general';
+import Optional from './service-schema/optional';
+
 let SERVICE_SCHEMA = {
   type: 'object',
   properties: {
-    General: {
-      description: 'Configure your container',
-      type: 'object',
-      properties: {
-        id: {
-          title: 'ID',
-          description: 'The id for the service',
-          type: 'string',
-          getter: function (service) {
-            return service.getId();
-          }
-        },
-        cpus: {
-          title: 'CPUs',
-          description: 'The amount of CPUs which are used for the service',
-          type:'number',
-          getter: function (service) {
-            return `${service.getCpus() || ''}`;
-          }
-        },
-        mem: {
-          title: 'Mem (MiB)',
-          type: 'number',
-          getter: function (service) {
-            return `${service.getMem() || ''}`;
-          }
-        },
-        disk: {
-          title: 'Disk (MiB)',
-          type: 'number',
-          getter: function (service) {
-            return `${service.getDisk() || ''}`;
-          }
-        },
-        instances: {
-          title: 'Instances',
-          type: 'number',
-          getter: function (service) {
-            return `${service.getInstancesCount() || 0}`;
-          }
-        },
-        cmd: {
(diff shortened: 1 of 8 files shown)


mesosphere-web:
Reviewer: @jfurrow

src/js/components/modals/ServiceFormModal.js
```diff
@@ -0,0 +1,240 @@
...
+];
+
+class ServiceFormModal extends mixin(StoreMixin) {
+  constructor() {
+    super(...arguments);
+
+    let model =
+      ServiceUtil.createFormModelFromSchema(ServiceSchema);
+    this.state = {
+      app: JSON.stringify({id:'', cmd:''}, null, 2),
```
aldipower: It seems weird to me that the Modal handles the app as a string. ?
Further I think this should come from a store.

src/js/components/modals/ServiceFormModal.js
```diff
@@ -0,0 +1,240 @@
...
+  handleClearError() {
+    this.setState({
+      errorMessage: null
+    });
+  }
+
+  handleJSONChange(app) {
+    let {service} = this.state;
+    try {
+      service = new Service(JSON.parse(app));
```
aldipower: The JSON parsing seems to me more the job of a request util?
Feels wrong to have this here.

aldipower: OK, sorry! It is the JSON mode, now I understand it better.
I would like to see this logic traversed out of this component.
But I am not sure if this is out of scope of this PR. Ignore this comment, if so. :)

src/js/components/modals/ServiceFormModal.js
```diff
@@ -0,0 +1,240 @@
...
+          onClick={this.handleSubmit}>
+          Deploy
+        </button>
+      </div>
+    );
+  }
+
+  getModalContents() {
+    let {app, jsonMode, service} = this.state;
+    if (jsonMode) {
```
aldipower: Still have the feeling the this should be part of another component. But that's another PR.

src/js/components/SchemaForm.js
```diff
@@ -78,7 +85,12 @@ class SchemaForm extends React.Component {
       return;
     }
 
-    setTimeout(() => {
+    // The cleartimeout is there to debounce the validation. And to make
+    // sure it is only run once.
+    if (this.timer) {
+      clearTimeout(this.timer);
+    }
+    this.timer = setTimeout(() => {
```
pierlo-upitup: Can't we use `Util.debounce`?

Poltergeist:
This commit does contain all the commits of #212 please wait with reviewing until #212 is merged.

src/js/constants/ServiceSchema.js
```diff
@@ -81,6 +81,63 @@ let SERVICE_SCHEMA = {
         }
       },
       required: []
+    },
+    'Optional': {
```
aldipower: Why does this `key` start with a capital letter? Looks very weird.

pierlo-upitup: it is used as a title of the modal section

src/js/constants/ServiceSchema.js
```diff
@@ -45,7 +45,7 @@ let SERVICE_SCHEMA = {
           type: 'number',
           default: 1,
           getter: function (service) {
-            return (service.getInstancesCount() || 0);
+            return (service.getInstancesCount() || 0) + '';
```
aldipower: Not sure why this has to be an string, but fair enough, there will be a reason for that.

pierlo-upitup: Actually if that's the case I'd rather see an ES6 template string here. ````
```
`${service.getInstancesCount() || 0}`
```

aldipower: Oh, just so that the "type" is set to 'number' above.
So this shouldn't be converted to string here I guess.

src/js/utils/ServiceUtil.js
```diff
@@ -28,6 +28,27 @@ const ServiceUtil = {
         definition.instances = formModel.General.instances;
       }
 
+      if (formModel.Optional != null) {
+        definition.executor = formModel.Optional.executor;
+        definition.fetch = formModel.Optional.uris &&
+          formModel.Optional.uris.split(',')
```
aldipower: The capital letter in `Optional` looks wrong.

src/js/constants/ServiceSchema.js
pierlo-upitup: Well the `type` there is the `input type="number"` i think. Maybe the string conversion is needed because reading an input's value (`input.value`) would return a string anyway?

pierlo-upitup: In fact I just saw this on master when opening the dialog: `tracker.js:10 Warning: Failed propType: Invalid prop `startValue` of type `number` supplied to `FieldInput`, expected `string`. Check the render method of `FormControl`.`

aldipower: Ah ok! So this should be explicitly of type "string"?

pierlo-upitup: No! It only affects the style of the input, it will show the stepper controls I believe.

src/js/utils/ServiceUtil.js
```diff
@@ -54,6 +75,12 @@ const ServiceUtil = {
...
     appDefinition.cmd = service.getCommand();
 
+    appDefinition.executor = service.getExecutor();
+    appDefinition.fetch = service.getFetch();
+    appDefinition.constraints = service.getConstraints();
+    appDefinition.acceptedResourceRoles = service.getAcceptedResourceRoles();
+    appDefinition.user = service.getUser();
+
     let containerSettings = service.getContainerSettings();
     if (
```
aldipower: ??

src/js/utils/ServiceUtil.js
```diff
@@ -54,6 +75,12 @@ const ServiceUtil = {
     appDefinition.instances = service.getInstancesCount();
     appDefinition.cmd = service.getCommand();
```
aldipower: why this extra line at this place?

aldipower:
Could you also remove the extra lines at those places? Thx.
https://github.com/dcos/dcos-ui/pull/267/files#diff-a44f44c46bf6bdbc44737179ffb8970dL2
https://github.com/dcos/dcos-ui/pull/267/files#diff-a44f44c46bf6bdbc44737179ffb8970dR64

src/js/constants/ServiceSchema.js
aldipower: Oh, confused now.
I don't think this "types" here are used to set the input field type.
There's is also a type of `object` and so on here in this scheme.
They are meant to be real javascript types. (Virtually)

src/js/utils/ServiceUtil.js
pierlo-upitup: The `SchemaForm` leverages the key as the title of the section. 
Either way, may I suggest to `let {Optional} = formModel;`? So that we don't have to repeat ourselves so much.

src/js/constants/ServiceSchema.js
pierlo-upitup: I've updated my comments: https://github.com/dcos/dcos-ui/pull/267/files/5b436ccc925cdcf6271d75b71155964504b0d810#r65355596

aldipower: Oh, so, no convertion to "string". :)

src/js/constants/ServiceSchema.js
aldipower: Sorry, using the key as a value to display is a very bad idea.

pierlo-upitup: Oh, I totally agree. But that's how the SchemaForm works right now and I think refactoring it to add a `sectionId` or similar is out of scope.

aldipower: Confirmative.

src/js/utils/ServiceUtil.js
aldipower: I see, confirmative.

src/js/constants/ServiceSchema.js
Poltergeist: the type is as PG mentioned is used for a very simple auto validation but sadly it is not used for the input field. Which means we have to provide a string which is then validated if it is a number. There is definitively some refactoring possible.

src/js/utils/ServiceUtil.js
Poltergeist: Accepted will change, and also do that for the General part above. Should I do that in this PR or do you prefer if we do that in another PR?

src/js/utils/ServiceUtil.js
Poltergeist: Maybe I should add a comment here I thought it would be handy to have a section per section lets say

```JS
// General
definition.id = service.getID();
...
...

// Optional
definition.executor = service.getExecutor();
...
...

File: src/js/constants/ServiceSchema.js [hunk 10]
```diff
@@ -81,6 +81,63 @@ let SERVICE_SCHEMA = {
...
+          description: 'Comma-separated list of resource roles. Marathon ' +
+          'considers only resource offers with roles in this list for ' +
+          'launching tasks of this app.',
+          getter: function (service) {
+            return service.getAcceptedResourceRoles() &&
+              service.getAcceptedResourceRoles().join(', ');
+          }
+        },
+        user: {
+          title: 'user',
```
pierlo-upitup: `User`

src/js/utils/ServiceUtil.js
Poltergeist: It is a long condition

```JS
    if (
      containerSettings && containerSettings.docker &&
      containerSettings.docker.image
    ) {
```

What we could do here is
```JS

    if (containerSettings && containerSettings.docker &&
      containerSettings.docker.image
    ) {
```
My assumption was that it is more readable as it is.

src/js/utils/ServiceUtil.js
pierlo-upitup: Do it here pls

src/js/utils/ServiceUtil.js
pierlo-upitup: Let's please
```
if (containerSettings && 
  containerSettings.docker &&
  containerSettings.docker.image) {
```

src/js/utils/ServiceUtil.js
aldipower: Ok that makes sense. Comments would be cool.

src/js/utils/ServiceUtil.js
aldipower: I would be fine with everything but the first.

Poltergeist:
What do you think if we pull each section from the schema to its own file and import it separately? This would give us the advantage of not having one huge file.

src/js/utils/ServiceUtil.js
kennyt: we usually do what PG said 👍

src/js/utils/ServiceUtil.js
Poltergeist: Added below this line 😄

src/js/constants/service-schema/general.js
```diff
@@ -0,0 +1,61 @@
+/* eslint-disable no-unused-vars */
+import React from 'react';
+/* eslint-enable no-unused-vars */
+
+let GENERAL = {
```
aldipower: Only "constants" should be written in capital letters.

src/js/constants/ServiceSchema.js
```diff
@@ -2,64 +2,13 @@
 import React from 'react';
 /* eslint-enable no-unused-vars */
 
+import General from './service-schema/general';
```
aldipower: In general (no I don't mean the var name itself ;) ) the letter casing should follow the file casing.
So I suggest to rename the file `general` to `General`.

src/js/constants/ServiceSchema.js
```diff
@@ -77,7 +26,8 @@ let SERVICE_SCHEMA = {
         }
       },
       required: []
-    }
+    },
+    'Optional': Optional
```
aldipower: Cool, this decouples it a little bit, I like that.

Poltergeist:
Renamed the files

aldipower:
Nice with ice!

src/js/utils/ServiceUtil.js
```diff
@@ -8,7 +8,7 @@ const getFindPropertiesRecursive = function (service, item) {
 
       return memo;
     }
-    memo[subItem] = item[subItem].getter(service) || item[subItem].default;
+    memo[subItem] = item[subItem].getter && item[subItem].getter(service) || item[subItem].default;
```
orlandohohmeier: Please use a propper condition instead of boolean operators for this. #nit

orlandohohmeier:
As discussed I'd like to move the schema into a different location (e.g. config) as it's definitely not a constant (any more). Besides that I suggest to make use of the newly [introduced `title` attribute](https://github.com/dcos/dcos-ui/pull/321). But none of this is in the scope of this PR. Let's just keep it in mind while advancing the modal.

orlandohohmeier:
Looks good to me. Great job! Let's 🚢  🇮🇹

PR #269, opened 2016-05-27
Move chronos jobs data to a reducer-like function

mlunoe:
This moves chronos jobs data to a reducer-like function that will restructure before passing it to the application

This comment spawned a larger conversation about what we should and should not do in the structs. We decided to move the whole building of the tree to a 'reducer'-like function that is being used in the actions before the struct is being created. That way we can handle any malformed data gracefully before handing it off to the struct:
https://github.com/dcos/dcos-ui/pull/250#discussion_r64840695

Ping team: @MatApple @daiyiwhale @kennyt @rcorral @jfurrow @orlandohohmeier @philipnrmn @pierlo-upitup @Poltergeist @aldipower

### DIFF
--- a/src/js/events/ChronosActions.js
+++ b/src/js/events/ChronosActions.js
@@ -21,6 +21,11 @@ const ChronosActions = {
             let data;
             try {
               data = ChronosUtil.parseJobs(response);
+              AppDispatcher.handleServerAction({
+                type: REQUEST_CHRONOS_JOBS_SUCCESS,
+                data
+              });
+              resolve();
             } catch (error) {
               AppDispatcher.handleServerAction({
                 type: REQUEST_CHRONOS_JOBS_ERROR,
@@ -28,12 +33,6 @@ const ChronosActions = {
               });
               reject();
             }
-
-            AppDispatcher.handleServerAction({
-              type: REQUEST_CHRONOS_JOBS_SUCCESS,
-              data
-            });
-            resolve();
           },
           error: function (e) {
             AppDispatcher.handleServerAction({
--- a/src/js/events/ChronosActions.js
+++ b/src/js/events/ChronosActions.js
@@ -31,7 +31,6 @@ const ChronosActions = {
             }
           },
           error: function (e) {
-            console.log(e.message);
             AppDispatcher.handleServerAction({
               type: REQUEST_CHRONOS_JOBS_ERROR,
               data: e.message
--- a/src/js/events/ChronosActions.js
+++ b/src/js/events/ChronosActions.js
@@ -17,9 +17,11 @@ const ChronosActions = {
           url: `${Config.rootUrl}/chronos/jobs`,
           data: [{name: 'embed', value: 'activeJobs'}],
           success: function (response) {
+            let data = ChronosUtil.parseJobs(response);
+
             AppDispatcher.handleServerAction({
               type: REQUEST_CHRONOS_JOBS_SUCCESS,
-              data: response
+              data
             });
             resolve();
           },
--- a/src/js/events/ChronosActions.js
+++ b/src/js/events/ChronosActions.js
@@ -27,14 +27,11 @@ const ChronosActions = {
               });
               resolve();
             } catch (error) {
-              AppDispatcher.handleServerAction({
-                type: REQUEST_CHRONOS_JOBS_ERROR,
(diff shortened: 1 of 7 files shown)


mesosphere-web:
Reviewer: @kennyt

src/js/events/ChronosActions.js
```diff
@@ -19,9 +20,20 @@ const ChronosActions = {
...
+              data = ChronosUtil.parseJobs(response);
+            } catch (error) {
+              AppDispatcher.handleServerAction({
+                type: REQUEST_CHRONOS_JOBS_ERROR,
+                data: error
+              });
+              reject();
+            }
+
             AppDispatcher.handleServerAction({
```
orlandohohmeier: Let's move the whole `handleServerAction` and `resolve` part into the `try` block or add a return to the `catch` block. Otherwise we're always dispatching the success event.

mlunoe: Oh yeah! Good catch! Done.

orlandohohmeier:
At this introduces a new pattern, let's make sure that we have a plan to rollout this new pattern to the rest of our codebase where needed.

rcorral:
Sounds good! I like it!

src/js/events/ChronosActions.js
```diff
@@ -19,13 +20,20 @@ const ChronosActions = {
...
+                type: REQUEST_CHRONOS_JOBS_SUCCESS,
+                data
+              });
+              resolve();
+            } catch (error) {
+              this.error(error);
+            }
           },
           error: function (e) {
+            console.log(e.message);
```
orlandohohmeier: Please remove this log.

src/js/utils/ChronosUtil.js
```diff
@@ -0,0 +1,95 @@
...
+   * (based on id/path matching).
+   * @param {{
+   *          id:string,
+   *          items:array<({id:string, items:array}|*)>,
+   * }} parent tree to add item to
+   * @param {{id:string}} item job to add to parent
+   * @param {[type]} jobsAlreadyAdded hash of id and jobs of jobs that
+   * have already been added to the parent
+   */
+  addJob(parent, item, jobsAlreadyAdded) {
```
orlandohohmeier: Let's make this method "private"

src/js/events/ChronosActions.js
mlunoe: Woops! I'll remove

src/js/utils/ChronosUtil.js
mlunoe: I added it here to be able to test the function

mlunoe: I'll look into testing both functions from `parseJobs`

mlunoe: done!

mlunoe:
All comments have been addressed

src/js/events/ChronosActions.js
```diff
@@ -17,11 +18,17 @@ const ChronosActions = {
...
+            try {
+              data = ChronosUtil.parseJobs(response);
+              AppDispatcher.handleServerAction({
+                type: REQUEST_CHRONOS_JOBS_SUCCESS,
+                data
+              });
+              resolve();
+            } catch (error) {
+              this.error(error);
+            }
```
jfurrow: Why do we need this `try...catch` block?

mlunoe: Becuase `ChronosUtil.parseJobs` can throw an error if the data being parsed is not of correct format. That is we in the `cacth` block call `this.error` with the message provided

kennyt: should this happen inside of a store? almost all of our Actions never do anything significant to the data

mlunoe: So in the description, I tried to explain that this using a more reducer-like approach to the data-parsing, meaning that it parses up in front, instead of when we build out the struct. The latter approach could become CPU heavy. The same change will be done for the ServiceTree.

kennyt:
integration failing so can't merge

mlunoe:
I'm trying to get the build of this to pass. It passes locally, but times out on the server:

orlandohohmeier:
Let's please hold merge until we figured out why those test fail every once in a while...

PR #270, opened 2016-05-27
Fix sidebar toggle button display problems

jfurrow:
Previously I had naively toggled the user `dropup` menus for all breakpoints. With this PR, the `sidebar-footer-user-dropdown` will be hidden from a viewport width of `0` up to `screen-small`, displayed from `screen-small` to `screen-medium`, and from there the toggle class `is-expanded` will determine its visibility.

### DIFF
--- a/src/styles/components/user-dropdown-menu.less
+++ b/src/styles/components/user-dropdown-menu.less
@@ -37,7 +37,6 @@
 .user-dropdown {
   display: none;
 
-  .sidebar.is-expanded &,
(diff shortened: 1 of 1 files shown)


mesosphere-web:
Reviewer: @daiyiwhale

pierlo-upitup:
To be honest I can't see a difference and I am not sure I even saw the problem this PR is solving to begin with, but the code LGTM 👍

PR #271, opened 2016-05-27
[1/2 Breadcrumbs] Refactor Breadcrumb components to allow for usage with route configuration

rcorral:
With this change we will be able to create breadcrumbs that are driven from route configurations like this: 
```js
          {
            type: Route,
            name: 'system-overview-units-unit-nodes-detail',
            path: 'components/:unitID/?',
            handler: UnitsHealthDetail,
            buildBreadCrumb: function () {
              return {
                parentCrumb: 'system-overview-units',
                getCrumbs: function (router) {
                  return [
                    <UnitsHealthDetailBreadcrumb
                      parentRouter={router}
                      routeName="system-overview-units-unit-nodes-detail"
                      />
                  ]
                }
              }
            }
          }
```

These breadcrumb components will come later.

**See [this](https://github.com/dcos/dcos-ui/pull/272/files#diff-dca06449f7494b90f4a887405c69765bR27) for an implementation example**

### DIFF
--- a/src/js/components/Breadcrumb.js
+++ b/src/js/components/Breadcrumb.js
@@ -1,251 +1,52 @@
-import classNames from 'classnames';
-import DeepEqual from 'deep-equal';
-import {Link} from 'react-router';
+import mixin from 'reactjs-mixin';
+/* eslint-disable no-unused-vars */
 import React, {PropTypes} from 'react';
-import ReactDOM from 'react-dom';
+/* eslint-enable no-unused-vars */
+import {StoreMixin} from 'mesosphere-shared-reactjs';
 
-import IconChevron from './icons/IconChevron';
+import Crumb from './Crumb';
 
-const COLLAPSE_BUFFER = 12;
-const LAST_ITEM_OFFSET = 150; // Difference between scrollWidth and outerWidth
-const PADDED_ICON_WIDTH = 38; // Width of icon + padding
-
-class Breadcrumb extends React.Component {
+class Breadcrumb extends mixin(StoreMixin) {
   constructor() {
     super(...arguments);
 
     this.state = {
-      availableWidth: null,
-      collapsed: false,
-      expandedWidth: null
+      isLoadingCrumb: true
     };
-
-    this.handleResize = this.handleResize.bind(this);
   }
 
-  componentDidMount() {
-    if (global.window != null) {
-      window.addEventListener('resize', this.handleResize);
-      window.addEventListener('focus', this.handleResize);
-    }
-  }
+  getBackupCrumbLabel() {
+    let {parentRouter, routeName} = this.props;
+    let route = parentRouter.namedRoutes[routeName];
+    let lastParam = route.paramNames[route.paramNames.length - 1];
+    let currentParamValue = parentRouter.getCurrentParams()[lastParam];
 
-  componentDidUpdate() {
-    this.updateDimensions();
-  }
-
-  componentWillUnmount() {
-    if (global.window != null) {
-      window.removeEventListener('resize', this.handleResize);
-      window.removeEventListener('focus', this.handleResize);
-    }
-  }
-
-  shouldComponentUpdate(nextProps, nextState) {
-    if (nextState.availableWidth == null || nextState.expandedWidth == null) {
-      this.updateDimensions();
-
-      return false;
-    }
-    if (this.state.collapsed !== nextState.collapsed) {
-      return true;
-    }
-
-    return !DeepEqual(this.props, nextProps);
-  }
-
-  handleResize() {
-    let availableWidth = this.getAvailableWidth();
-
-    this.setState({
-      availableWidth,
-      collapsed: this.shouldCollapse(availableWidth, this.state.expandedWidth)
-    });
-  }
-
-  updateDimensions() {
-    let availableWidth = this.getAvailableWidth();
-    let expandedWidth = this.getExpandedWidth();
-
-    this.setState({
-      availableWidth,
-      expandedWidth,
-      collapsed: this.shouldCollapse(availableWidth, expandedWidth)
-    });
-  }
-
-  getAvailableWidth() {
-    return ReactDOM.findDOMNode(this).offsetWidth;
-  }
-
-  getWidthFromExpandedItem(item) {
-    return item.offsetWidth + PADDED_ICON_WIDTH;
-  }
-
-  getWidthFromCollapsedItem(item) {
-    let link = item.children[0];
-    let textWidth = link.scrollWidth - link.offsetWidth;
-
-    return textWidth + PADDED_ICON_WIDTH;
-  }
-
-  getExpandedWidth() {
-    // array/splat casts NodeList to array
-    let listItems = [...ReactDOM.findDOMNode(this).children]
-      .filter(function (_, index) {
-        // Filter out even nodes containing '>'
-        return index % 2;
-      });
-
-    return listItems
-      .map((item, index) => {
-        let isFirstItem = index === 0;
-        let isLastItem = index === listItems.length - 1;
-        if (isFirstItem || isLastItem) {
-          return this.getWidthFromExpandedItem(item);
-        }
-
-        return this.getWidthFromCollapsedItem(item);
-      })
-      .reduce(
-        function (totalWidth, itemWidth) { return totalWidth + itemWidth },
-        this.getLastItemWidth()
-      );
-  }
-
-  getLastItemWidth() {
-    let lastItem = ReactDOM.findDOMNode(this).lastChild;
-    let lastItemLink = lastItem.firstChild;
-
-    return lastItemLink.scrollWidth + LAST_ITEM_OFFSET;
-  }
-
-  getCurrentRouteParams() {
-    // Isolated for easier testing
-    return this.context.router.getCurrentParams();
-  }
-
-  getCrumb(crumb, key) {
-    let label = null;
-    let route = null;
-
-    if (crumb.hasOwnProperty('route')) {
-      route = crumb.route;
-    }
-
-    if (crumb.hasOwnProperty('label')) {
-      label = crumb.label;
-    }
-
-    if (route) {
-      crumb = (
-        <Link to={route.to}
-            params={route.params}
-            title={label}>
-          {label}
-        </Link>
-      );
-    }
-
-    return (
-      <li key={key}>
-        {crumb}
-      </li>
-    );
-  }
-
-  getCrumbDivider(key) {
-    return (
-      <li key={key} >
-        <IconChevron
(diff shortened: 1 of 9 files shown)


mesosphere-web:
Reviewer: @MatApple

src/js/components/Breadcrumbs.js
```diff
@@ -0,0 +1,242 @@
+import classNames from 'classnames';
```
rcorral: I renamed the `Breadcrumb` file to `Breadcrumbs` and now it looks like a brand new file :( sorry.

src/js/components/Breadcrumbs.js
```diff
@@ -0,0 +1,242 @@
...
+    return crumbs.reduce((memo, crumb, crumbIndex) => {
+      memo.push(this.getCrumb(crumb, ++crumbKey));
+
+      if (crumbs.length - 1 !== crumbIndex) {
+        memo.push(this.getCrumbDivider(++crumbKey));
+      }
+
+      return memo;
+    }, []);
+  }
```
rcorral: This method and `buildCrumbs` are the ones that have changed from previous implementation, everything else should be the same.

src/js/components/Crumb.js
```diff
@@ -0,0 +1,28 @@
+import {Link} from 'react-router';
+import React, {PropTypes} from 'react';
+
+class Crumb extends React.Component {
```
rcorral: This component will allow us to create a single Crumb, that is then later wrapped in an `li` at least for now.

src/js/components/Breadcrumb.js
```diff
@@ -1,251 +1,52 @@
...
 
-import IconChevron from './icons/IconChevron';
+import Crumb from './Crumb';
 
-const COLLAPSE_BUFFER = 12;
-const LAST_ITEM_OFFSET = 150; // Difference between scrollWidth and outerWidth
-const PADDED_ICON_WIDTH = 38; // Width of icon + padding
-
-class Breadcrumb extends React.Component {
+class Breadcrumb extends mixin(StoreMixin) {
```
rcorral: This component will now be used by other components as a parent class.
This will allow us to easily build breadcrumb items with this class, as it abstracts a lot of the functionality that each breadcrumb will require.

pierlo-upitup:
> With this change we will be able to create breadcrumbs that are driven from route configurations like this:

@rcorral I really wish the Breadcrumbs would be able to build themselves automatically based on the current Route (and maybe add a prop to override the path if needed) rather than us having to write such imperative code. 
This is to illustrate what I mean: https://github.com/svenanders/react-breadcrumbs

src/js/components/Breadcrumb.js
pierlo-upitup: Breadcrumb shouldn't need to extend the `StoreMixin` imho. What is the `StoreMixin` solving here for this and any of the inheriting classes that can't be dealt with using `props`?

pierlo-upitup:
Shouldn't we have tests?

rcorral:
Hey @pierlo-upitup I talked this over with @MatApple for some time last week and came up with two solutions, for which we picked this current one.
1. **Dynamic breadcrumbs**: This sounds ideal at first, but will cause us to do two things. 1) Nest all of our routes. Which means we need to have an `if` statement in the `render` of each view to determine if we show the contents of that view or if it's a child view. 2) The entire route tree needs to load before the target view loads. This quite painful because we will have to do consecutive xhr calls to get detailed information about each crumb along the way as in most cases we want a custom crumb. Because we have to make xhr calls, any of them can fail at any point, which will cause us to show the error page.
2. **Breadcrumb components**: This looks like a lot more work, and it is. But it gives us a lot more flexibility and non of the "cons" as the other method. It's just more work at first. This is why we've picked this as the best solution out of the two.

src/js/components/Breadcrumb.js
```diff
@@ -1,251 +1,52 @@
...
   constructor() {
```
kennyt: if we're going to extend the `StoreMixin`, can we do `this.store_listeners = [];` so it won't throw an error if the things that extend this don't want to use it?

src/js/components/Breadcrumbs.js
```diff
@@ -0,0 +1,242 @@
+import classNames from 'classnames';
+import DeepEqual from 'deep-equal';
```
kennyt: question: do we usually camelcase the things that we import that are functions? like `classNames`

src/js/components/Crumb.js
```diff
@@ -0,0 +1,28 @@
...
+  render() {
+    let {props} = this;
+    let content = props.label;
+
+    if (props.route) {
+      content = (
+        <Link to={props.route.to}
+            params={props.route.params}
```
kennyt: this line and the one below it are indented one level too deep

src/js/components/Breadcrumbs.js
rcorral: I think people just grab the example from the documentation for these modules

kennyt: should we formally make a rule for it to be 🐫 🐫 🐫

src/js/components/Breadcrumb.js
rcorral: Good call, done.

src/js/components/Crumb.js
rcorral: Done.

rcorral:
@pierlo-upitup added tests for the components that I added.
The Breadcrumbs.js file makes heavy use of react-router...which is a bitch to test.

src/js/components/BreadcrumbLink.js
```diff
@@ -0,0 +1,32 @@
+import {Link} from 'react-router';
+import React, {PropTypes} from 'react';
+
+class Breadcrumb extends React.Component {
```
mlunoe: Isn't this supposed to be `BreadcrumbLink`?

src/js/components/BreadcrumbSegment.js
```diff
@@ -0,0 +1,58 @@
+import mixin from 'reactjs-mixin';
+/* eslint-disable no-unused-vars */
+import React, {PropTypes} from 'react';
+/* eslint-enable no-unused-vars */
+import {StoreMixin} from 'mesosphere-shared-reactjs';
+
+import BreadcrumbLink from './BreadcrumbLink';
+
+class BreadcrumbSegment extends mixin(StoreMixin) {
```
pierlo-upitup: Could you please explain why this breadcrumb segment component needs to extend the StoreMixin? I would prefer keeping this component store-agnostic and solving the problem using props if possible.

pierlo-upitup:
Hey @rcorral I'm a little confused by the naming and the actual hierarchy. 

In which case would we use `Breadcrumbs` vs `BreadcrumbSegment`? And what is the relation between the two?

rcorral:
Clearly I've failed at naming, so I'm opened to suggestions. But this is the thought right now:
`Breadcrumbs`: Is a drop-in component to render the breadcrumbs for that page
`BreadcrumbSegment`: Is a component that we can extend that will render a breadcrumb segment.
`BreadcrumbLink`: Creates the link to page from breadcrumb.

src/js/components/BreadcrumbLink.js
rcorral: Fixed.

rcorral:
Renamed BreadcrumbLink to BreadcrumbSegmentLink.
Let's not have this PR be blocked on naming, I'm fine doing the renaming if someone comes up with a better name afterwards.

src/js/components/BreadcrumbSegment.js
rcorral: Removed!

src/js/components/BreadcrumbSegment.js
```diff
@@ -0,0 +1,54 @@
...
+
+class BreadcrumbSegment extends React.Component {
+  constructor() {
+    super(...arguments);
+
+    this.state = {
+      isLoadingCrumb: true
+    };
+
+    this.store_listeners = [];
```
mlunoe: Do we need this when it is not using the `StoreMixin`?

src/js/components/__tests__/BreadcrumbSegment-test.js
```diff
@@ -0,0 +1,47 @@
...
+
+  it('renders the label', function () {
+    let instance = TestUtils.renderIntoDocument(
+      <BreadcrumbSegment routeName="foo" parentRouter={this.parentRouter} />
+    );
+
+    expect(instance.getBackupCrumbLabel()).toEqual('baz');
+  });
+
+  it('renders the label', function () {
```
mlunoe: Renders the link?

src/js/components/BreadcrumbSegmentLink.js
```diff
@@ -0,0 +1,32 @@
+import {Link} from 'react-router';
+import React, {PropTypes} from 'react';
+
+class BreadcrumbSegmentLink extends React.Component {
+  render() {
+    let {props} = this;
+    let content = props.label;
+
+    if (props.route) {
+      content = (
```
mlunoe: #nit we could just `return` here and on line 18
(should not hold op the PR)

src/js/components/BreadcrumbSegment.js
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  getCrumbLabel() {
+    return '';
+  }
+
+  getBackupCrumbLabel() {
```
mlunoe: I have a hard time reasoning about what is going on here in this function. Could you add an explanatory comment in the code?

src/js/components/BreadcrumbSegment.js
rcorral: Done.

src/js/components/__tests__/BreadcrumbSegment-test.js
rcorral: Done.

src/js/components/BreadcrumbSegmentLink.js
rcorral: Done.

src/js/components/BreadcrumbSegment.js
rcorral: Done.

mlunoe:
Awesome! Looks good! 🚢

src/js/components/BreadcrumbSegment.js
rcorral: Nvm, I brought it back, because it's not possible to do `extends mixin(BreadcrumbSegment, StoreMixin) {` at the moment as the first mixin is a react component. Plus I foresee all components that extend `BreadcrumbSegment` to use the storemixin.

PR #272, opened 2016-05-27
[2/2 Breadcrumbs] Implemented breadcrumb on Node Unit page.

rcorral:
First implementation of new Breadcrumbs.
I also had to fix a bunch of things that were missing from the HealthUnit pages, like handling loading and errors. This is something that we should've done in the beginning. There were also some typos.


mesosphere-web:
Reviewer: @philipnrmn

src/js/pages/system/UnitsHealthNodeDetail.js
```diff
@@ -33,12 +43,22 @@ class UnitsHealthNodeDetail extends mixin(StoreMixin) {
...
   }
 
-  getErrorNotice() {
-    return (
-      <div className="container container-pod">
-        <RequestErrorMsg />
-      </div>
-    );
+  onUnitHealthStoreUnitSuccess() {
+    let isLoading = this.state.isLoading & ~FLAGS.UNIT_LOADED;
```
jfurrow: Not gonna lie, the bitwise stuff in here is confusing to me. But I should understand it better anyway so I'm fine with keeping it.

rcorral: Yeah, they'll take some getting used to. But it's really nice for components that depend on more than one request before rendering. We could even tie in the error as a flag too. That would be nice that way we don't have another variable tracking that.

rcorral:
Closing this and opening a new one against master.

PR #273, opened 2016-05-27
[2/2 Breadcrumbs] Implemented breadcrumb on Node Unit page.

rcorral:


mesosphere-web:
Reviewer: @Poltergeist

src/js/pages/system/UnitsHealthDetail.js
```diff
@@ -47,6 +55,24 @@ class UnitsHealthDetail extends mixin(StoreMixin) {
     UnitHealthStore.fetchUnitNodes(this.props.params.unitID);
   }
 
+  onUnitHealthStoreUnitSuccess() {
+    let isLoading = this.state.isLoading & ~FLAGS.UNIT_LOADED;
```
pierlo-upitup: Reasoning about these flags is really hard. Also, it's a pattern I haven't seen anywhere else in the codebase. Additionally I am not even sure I understand the need for these in the first place, being both flags constant values.

I also find `isLoading` to not be of type `bool` very strange, but that could be fixed with a `!!`. Still, I'd like to suggest reconsidering this approach to something less contrived, or to at least have comments to explain what's going on.

src/js/pages/system/UnitsHealthDetail.js
```diff
@@ -19,6 +20,11 @@ const METHODS_TO_BIND = [
   'resetFilter'
 ];
 
+const FLAGS = {
+  UNIT_LOADED: parseInt('1', 2),
```
pierlo-upitup: ... or use ES6 : `0b1` and `0b10`

src/js/components/Breadcrumbs.js
```diff
@@ -0,0 +1,242 @@
+import classNames from 'classnames';
```
orlandohohmeier: What is the difference between `Breadcrumbs` and `Breadcrumb` and how do they relate to the `Crumb` component?

src/js/routes/factories/system.js
```diff
@@ -19,19 +24,56 @@ let RouteFactory = {
...
+            buildBreadCrumb: function () {
+              return {
+                parentCrumb: 'system-overview-units-unit-nodes-detail',
+                getCrumbs: function (router) {
+                  return [
+                    <UnitsHealthNodeDetailBreadcrumb
+                      parentRouter={router}
+                      routeName="system-overview-units-unit-nodes-node-detail"
+                      />
+                  ]
```
kennyt: this needs a semicolon

src/js/components/Crumb.js
```diff
@@ -0,0 +1,28 @@
...
+
+class Crumb extends React.Component {
+  render() {
+    let {props} = this;
+    let content = props.label;
+
+    if (props.route) {
+      content = (
+        <Link to={props.route.to}
+            params={props.route.params}
```
kennyt: this is indented too far

src/js/pages/system/UnitsHealthDetail.js
rcorral: I'm up for a different solution, but what I don't want is to have `n` number of variables tracking if different things have loaded. I think this is the right tool for the job, and I'm hoping you guys see it once you get used to looking at it. 
Maybe this example helps:
```
let myVar = 0b111;
let MY_FLAG = 0b11;

myVar &= ~MY_FLAG;
myVar.toString(2); // 100
myVar &= ~MY_FLAG;
myVar.toString(2); // 100 (it doesn't keep flipping bits)
```

It's also really nice to do `if (this.state.isLoading)` in the `render` method. Rather than `if (fooIsLoading && barIsLoading && ...)`.

kennyt:
i still don't understand how to use it. your example is a case where it is special. how do i just use it for the biggest use case which is what is in the params?

`Secrets > secret/path/here`

src/js/pages/system/UnitsHealthDetail.js
mlunoe: Could we create some wrapper methods to make it more descriptive? Something like: `getFlagValue` that does return a `bool`?

rcorral: A mixin?

src/js/routes/factories/system.js
rcorral: Done.

src/js/components/Breadcrumbs.js
rcorral: @orlandohohmeier I've renamed the components so that they make more sense.
`Breadcrumbs`: Is a drop-in component to render the breadcrumbs for that page
`BreadcrumbSegment`: Is a component that we can extend that will render a breadcrumb segment.
`BreadcrumbLink`: Creates the link to page from breadcrumb.

src/js/pages/system/UnitsHealthDetail.js
mlunoe: I would prefer a util or just functions on the component

rcorral: So a Util that you would just pass a variable `isLoading` and the bits that you want to turn off?

mlunoe: Yeah, and it returns the new variable. Something like that

orlandohohmeier: I don't see any need for bitwise operators here... You're dealing only with two flags, and I don't see any issue with a slightly bloated but easy digestible condition instead of a fancy obscure bitwise operation. Let's please not introduce them, as they are hard to understand (at least for most of the people) and inconsistent with the rest of our codebase.
How about a solution using a map or something similar that is, as well able to handle lots of flags without bloating the condition? 

```javascript
/**
* Check if all resources are loaded
* @param {Map} resourceMap
* @return boolean
*/
function areAllResourcesLoaded(resourceMap){
  return Array.from(resourceMap.values()).every(function ({loaded}}) {
  
    return loaded;
  });
}

// Init resource map
let resourceMap = new Map();
resourceMap.set('units', {loaded:false});
resourceMap.set('nodes', {loaded:false});
resourceMap.set('foo', {loaded:false})
resourceMap.set('bar', {loaded:false})

// Update map in load handler
resourceMap.set('units', {loaded:true});
resourceMap.set('nodes', {loaded:true});
resourceMap.set('foo', {loaded:true})
resourceMap.set('bar', {loaded:true})

// Check if everything is loaded
if(!areAllResourcesLoaded(resourceMap)){
    console.log('🕑  still loading')
}else {
   console.log('🎉  everything is loaded')
}
```
_The above is untested and should only illustrate a solution using a map or array_

pierlo-upitup: +1 Can't agree more. Additionally for the specific case of this PR we could just hide the logical checks away inside of a simple function and be done with it.

rcorral: Changed to just use more variables to track this stuff.

mlunoe:
This can be rebased now

kennyt:
tests failing

src/js/pages/system/breadcrumbs/UnitsHealthDetailBreadcrumb.js
```diff
@@ -0,0 +1,41 @@
...
+    ];
+  }
+
+  componentDidMount() {
+    super.componentDidMount();
+
+    let {unitID} = this.props.parentRouter.getCurrentParams();
+    let unit = UnitHealthStore.getUnit(unitID);
+
+    if (unit.get('id')) {
```
kennyt: don't have a getter for this?

src/js/pages/system/breadcrumbs/UnitsHealthNodeDetailBreadcrumb.js
```diff
@@ -0,0 +1,41 @@
...
+    ];
+  }
+
+  componentDidMount() {
+    super.componentDidMount();
+
+    let {unitID, unitNodeID} = this.props.parentRouter.getCurrentParams();
+    let node = UnitHealthStore.getNode(unitNodeID);
+
+    if (node.get('host_ip')) {
```
kennyt: don't have a getter for this?

src/js/pages/system/breadcrumbs/UnitsHealthNodeDetailBreadcrumb.js
```diff
@@ -0,0 +1,41 @@
...
+
+  onUnitHealthStoreNodeSuccess() {
+    this.setState({isLoadingCrumb: false});
+  }
+
+  getCrumbLabel() {
+    let {unitNodeID} = this.props.parentRouter.getCurrentParams();
+    let node = UnitHealthStore.getNode(unitNodeID);
+
+    return node.get('host_ip');
```
kennyt: don't have a getter for this?

kennyt:
px

lots of height

kennyt:
flicts

src/js/pages/system/breadcrumbs/UnitsHealthDetailBreadcrumb.js
rcorral: Can you point me to it? I don't know which one you're talking about.

src/js/pages/system/breadcrumbs/UnitsHealthNodeDetailBreadcrumb.js
rcorral: Same here. I don't know which one you're talking about.

src/js/pages/system/breadcrumbs/UnitsHealthNodeDetailBreadcrumb.js
rcorral: And same here.

rcorral:
@kennyt fixed styling
img

PR #275, opened 2016-05-28
[ES6 class] Introduce base store

orlandohohmeier:
This base store will be extended by all other stores and it's intended to replace the `Store.create` util as well as the GetSet mixin.

### DIFF
--- a/src/js/stores/BaseStore.js
+++ b/src/js/stores/BaseStore.js
@@ -10,10 +10,6 @@ if (global.__DEV__) {
 import {APP_STORE_CHANGE} from '../constants/EventTypes';
 
 class BaseStore extends EventEmitter {
-  constructor() {
-    super(...arguments);
-  }
-
   get(key) {
     if (typeof this.getSet_data === 'undefined') {
       return null;
--- a/src/js/stores/BaseStore.js
+++ b/src/js/stores/BaseStore.js
@@ -1,0 +1,52 @@
+import {EventEmitter} from 'events';
+
+let PluginSDK;
+
+// Hack until we fix circular dependency - DCOS-5040
+if (global.__DEV__) {
+  PluginSDK = require('PluginSDK');
+}
+
+import {APP_STORE_CHANGE} from '../constants/EventTypes';
+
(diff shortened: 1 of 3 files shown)


mesosphere-web:
Reviewer: @mlunoe

pierlo-upitup:
I'm not sold on the need for https://github.com/dcos/dcos-ui/pull/275/files#diff-f7d4948c1e227d032d593d0eebcc7eb3R14 but LGTM.

src/js/stores/BaseStore.js
```diff
@@ -0,0 +1,52 @@
...
+if (global.__DEV__) {
+  PluginSDK = require('PluginSDK');
+}
+
+import {APP_STORE_CHANGE} from '../constants/EventTypes';
+
+class BaseStore extends EventEmitter {
+  constructor() {
+    super(...arguments);
+  }
```
pierlo-upitup: This is not needed at all.

src/js/stores/__tests__/BaseStore-test.js
```diff
@@ -0,0 +1,65 @@
...
+    it('should return undefined if no key is given', function () {
+      expect(this.instance.get()).toEqual(null);
+    });
+
+    it('should return undefined if given an object', function () {
+      expect(this.instance.get({})).toEqual(null);
+    });
+
+    it('returns null if property hasn\'t been defined', function () {
+      expect(this.instance.get('foo')).toEqual(null);
```
pierlo-upitup: Shouldn't it blow up in this case? In a sense the store is lying to us. If I'm accessing a property that was never defined, I'd like to know that rather than having the store tell me that property is defined and set to `null`.
What do you think?

orlandohohmeier: I agree on that and I would like to completely remove the `get` and `set` methods but that's a different story. The intention of this store was to replace the `Store.create` util as well as the `GetSet` mixin, thus it replicates their interface and we should not alter it for now.

pierlo-upitup: I see. Merging.

PR #284, opened 2016-05-28
[ES6 class] Update MesosSummaryStore to es6 class

orlandohohmeier:
---
ℹ️   _This branch depends on features that will be introduced with #275, thus it will fail until the respective branch is merged._

---
Update store to es6 class – this is part of series of PRs to update all the stores as discussed in https://github.com/dcos/dcos-ui/pull/246/files/b4b52fcdb5f899fd74f9bd68cfa3d216177fcecf#r64962494

### DIFF
--- a/src/js/stores/MesosSummaryStore.js
+++ b/src/js/stores/MesosSummaryStore.js
@@ -1,4 +1,4 @@
-import {Store} from 'mesosphere-shared-reactjs';
+import BaseStore from './BaseStore';
 
 var AppDispatcher = require('../events/AppDispatcher');
 import ActionTypes from '../constants/ActionTypes';
@@ -8,7 +8,6 @@ import {
   MESOS_SUMMARY_CHANGE,
   MESOS_SUMMARY_REQUEST_ERROR
 } from '../constants/EventTypes';
-var GetSetMixin = require('../mixins/GetSetMixin');
 var MesosSummaryUtil = require('../utils/MesosSummaryUtil');
 var MesosSummaryActions = require('../events/MesosSummaryActions');
 import SummaryList from '../structs/SummaryList';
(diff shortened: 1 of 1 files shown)


mesosphere-web:
Reviewer: @mlunoe

src/js/events/MesosSummaryActions.js
```diff
@@ -84,12 +85,25 @@ var MesosSummaryActions = {
     Config.getRefreshRate(),
     function (resolve, reject) {
 
-      return function (timeScale) {
+      return function (timeScale, placeholder = false) {
+        if (placeholder) {
+          AppDispatcher.handleServerAction({
+            type: REQUEST_SUMMARY_PLACEHOLDER,
```
rcorral: I think we need a better solution for this.

src/js/stores/MesosSummaryStore.js
```diff
@@ -38,12 +37,7 @@ function startPolling() {
       } else {
         // If not active, push null placeholder. This will ensure we maintain
         // history when navigating back, for case where history server is down.
-
-        // Use {silent: true} Because we only want to push a summary on the stack without side
-        // effects (like re-rendering etc). The tab is out of focus so we
-        // don't want it to do any work. It only matters that there is
-        // appropriate history when we return focus to the tab.
-        MesosSummaryStore.processSummaryError({silent: true});
+        MesosSummaryActions.fetchSummary(undefined, true);
```
rcorral: Can we instead just do this at the bottom of the file: `MesosSummaryStoreInstance = new MesosSummaryStore();`
and call `MesosSummaryStoreInstance.processSummaryError({silent: true});` just like we were doing before?

I don't like the idea of calling a method on an Actions file that we know will return faked data as if it were making a call to an API. I don't believe that's the best layer to fake the data.

orlandohohmeier: I'll change it back to the `processSummaryError({silent: true});` method call as the refactoring is actually not in the scope of this PR. But I'd like to discuss this in our next meeting, as I think that actions, following the flux pattern, are the right place to do that kind of stuff.

/cc @mlunoe

mlunoe: Sounds good!

PR #286, opened 2016-05-28
[ES6 class] Update MesosLogStore to es6 class

orlandohohmeier:
---
ℹ️   _This branch depends on features that will be introduced with #275, thus it will fail until the respective branch is merged._

---
Update store to es6 class – this is part of series of PRs to update all the stores as discussed in https://github.com/dcos/dcos-ui/pull/246/files/b4b52fcdb5f899fd74f9bd68cfa3d216177fcecf#r64962494

### DIFF
--- a/src/js/stores/MesosLogStore.js
+++ b/src/js/stores/MesosLogStore.js
@@ -1,30 +1,62 @@
-import {Store} from 'mesosphere-shared-reactjs';
+import BaseStore from './BaseStore';
 
 import ActionTypes from '../constants/ActionTypes';
 import AppDispatcher from '../events/AppDispatcher';
 import Config from '../config/Config';
-import EventTypes from '../constants/EventTypes';
-import GetSetMixin from '../mixins/GetSetMixin';
+import EventTypes from '../constants/EventTypes';;
 import Item from '../structs/Item';
 import LogBuffer from '../structs/LogBuffer';
 import MesosLogActions from '../events/MesosLogActions';
 
 const MAX_FILE_SIZE = 50000;
 
-const MesosLogStore = Store.createStore({
-  storeID: 'mesosLog',
+class MesosLogStore extends BaseStore {
 
-  mixins: [GetSetMixin],
+  constructor() {
+    super(...arguments);
 
-  addChangeListener: function (eventName, callback) {
+    this.dispatcherIndex = AppDispatcher.register((payload) => {
+      let source = payload.source;
+      if (source !== ActionTypes.SERVER_ACTION) {
+        return false;
+      }
+
+      let action = payload.action;
+
+      switch (action.type) {
+        case ActionTypes.REQUEST_MESOS_LOG_SUCCESS:
+          this.processLogEntry(action.slaveID, action.path, action.data);
+          break;
+        case ActionTypes.REQUEST_MESOS_LOG_ERROR:
+          this.processLogError(action.slaveID, action.path);
+          break;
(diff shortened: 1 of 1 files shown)


mesosphere-web:
Reviewer: @kennyt

src/js/stores/MesosLogStore.js
```diff
@@ -1,30 +1,62 @@
...
-
-  mixins: [GetSetMixin],
-
-  addChangeListener: function (eventName, callback) {
+class MesosLogStore extends BaseStore {
+
+  constructor() {
+    super(...arguments);
+
+    this.dispatcherIndex = AppDispatcher.register((payload) => {
```
aldipower: Maybe I miss how it really works. :)

I wonder how this is unregistered properly? The store has no deconstructor. This looks like a memory leak here.

A function gets registered at a 'global' object `AppDispatcher` in a constructor that could be instantiated n-times. 
I think nobody noticed yet, because MesosLogStore is only instantiated once?

orlandohohmeier: @aldipower If we wouldn't export an [instance](https://github.com/dcos/dcos-ui/pull/286/files/39bd1df7e5b21ad2d676aac226deed7ae360be09#diff-d2871a01da8a1ac9c58e9633e5e84505R203) but the class itself we would be in real danger of creating memory leaks as there is no unregister. 
We could introduce a new singleton pattern which makes sure that we will always have only one instance in memory even if we export the class. But thats a different story and not part of this PR.

aldipower: I didn't notice that the instance is exported, so that makes it kind of safe, true.
But to be honest, the only technically save solution I can think about to register the dispatcher-function is:
Outside of the class, using an instance created by a singleton.
Not in the constructor.

But yeah, not in this PR maybe.

orlandohohmeier: If we can make sure that one can only create a single instance of this class, then there is no need to move the registering part. The current "singleton" pattern is the cheapest of all available patterns but we were using it already (e.g. [CompositeState](https://github.com/dcos/dcos-ui/blob/master/src/js/structs/CompositeState.js#L103)).

Let's add this topic to our agenda and decide on a true and safe singleton pattern and refactor the store accordingly. I'd favor a pattern that allows us to extend singleton classes and one that is not only relying on "proper" file handling (e.g. case sensitivity).

aldipower: Yes, logically there is no need to move it then, but it still remains very dirty. It's not fail-safe in the long term. E.g. if somebody other has to refactor this in some years? You cannot ensure the person is aware of this specialty. So it's good to eliminate this soon.

Agreeing with everything else.

aldipower: @orlandohohmeier I like to here your opinion on my last comment.

orlandohohmeier: I'm not sure where especially you see the risk, if we can make sure that there is only a single instance of a class. I'd love to hear what risks you have in mind and why you think it's dirty. 
I've added this topic to our agenda and would like to define a set of requirements and features and then have look at the current state of the art javascript singleton patterns and pick one.

aldipower: It's simple, I see the risk in "making something sure". It adds an unnecessary level of complexity that needs to be understood by the person that is working on that file.

orlandohohmeier: Gotcha... So one of our requirements should be that the pattern is easy to understand and maintain.

BTW: The plain object singleton (old store pattern) is anything than memory save (see: file handling)

PR #294, opened 2016-05-31
Adds module type examples

rcorral:
This is so that people aren't confused when writing new code which modules they should use as examples.

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,6 +109,21 @@ If you want to add a new npm package to 'node_modules':
     }
     ```
 
+## Module examples
+
+Check out these module types that you can use as a blueprint.
+
+* [Table](https://github.com/dcos/dcos-ui/blob/master/src/js/components/ServicesTable.js)
+* [Modal](https://github.com/dcos/dcos-ui/blob/master/src/js/components/modals/AddRepositoryFormModal.js)
+* [Actions](https://github.com/dcos/dcos-ui/blob/master/src/js/events/CosmosPackagesActions.js)
+* [Main navigation page](https://github.com/dcos/dcos-ui/blob/master/src/js/pages/SystemPage.js)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -109,6 +109,21 @@ If you want to add a new npm package to 'node_modules':
...
+
+Check out these module types that you can use as a blueprint.
+
+* [Table](https://github.com/dcos/dcos-ui/blob/master/src/js/components/ServicesTable.js)
+* [Modal](https://github.com/dcos/dcos-ui/blob/master/src/js/components/modals/AddRepositoryFormModal.js)
+* [Actions](https://github.com/dcos/dcos-ui/blob/master/src/js/events/CosmosPackagesActions.js)
+* [Main navigation page](https://github.com/dcos/dcos-ui/blob/master/src/js/pages/SystemPage.js)
+* [Tab page](https://github.com/dcos/dcos-ui/blob/master/src/js/pages/system/RepositoriesTab.js)
+* [Adding Routes](https://github.com/dcos/dcos-ui/blob/master/src/js/routes/nodes.js)
+* [Store](https://github.com/dcos/dcos-ui/blob/master/src/js/stores/ChronosStore.js)
```
kennyt: hmm one comment i have about the chronos store is that it doesn't use the `StoreMixin` to listen to the `VisibilityStore`. 

For example something like:

```js
 this.store_initializeListeners([{
    name: 'visibility',
    events: ['change']
}]);
```

jfurrow:
This is a great idea! I think we could add `Breadcrumbs

pierlo-upitup:
I propose to keep this updated as we merge new components that should be reused.

README.md
kennyt: dangit this comment was ignored. stores are gonna be messed up

kennyt: omg a store has already been created that followed this pattern. we should rectify this ASAP

PR #295, opened 2016-06-01
Fix ChronosStore visibility change handling

orlandohohmeier:
Adjust the ChronosStore visibility change handling to fix reference errors.


mesosphere-web:
Reviewer: @jfurrow

pierlo-upitup:
I also 🍒 -picked this onto current master and it fixes the following
image


Nice catch!

PS Outside of this PR, could we turn off the ChronosStore fetching until the API is there?

PR #296, opened 2016-06-01
Add getter for storeID on MarathonStore

Poltergeist:
Add getter for storeID on MarathonStore


mesosphere-web:
Reviewer: @kennyt

PR #297, opened 2016-06-01
Fix ChronosStore storeID getter

orlandohohmeier:
Move the `storeID` getter to the instance, as the StoreMixin accesses storeIDs as a property of the stores not the constructors.


mesosphere-web:
Reviewer: @aldipower

philipnrmn:
Nice catch 👍

aldipower:
Nice! With test! :)

PR #298, opened 2016-06-01
Create FormUtil to create multiple rows

kennyt:


mesosphere-web:
Reviewer: @orlandohohmeier

src/js/utils/FormUtil.js
```diff
@@ -0,0 +1,89 @@
...
+        propValue[valueIndex] = instanceValue;
+      }
+    });
+
+    return Object.assign({}, model, {[prop]: propValue});
+  }
+};
+
+global.test = FormUtil.getMultipleFields;
+global.xform = FormUtil.modelToCombinedProps;
```
pierlo-upitup: Leftovers?

src/js/utils/FormUtil.js
```diff
@@ -0,0 +1,89 @@
...
+  });
+}
+
+function getProp(key) {
+  return key.split('[')[0];
+}
+
+function getPropIndex(key) {
+  let value = key.split('[');
+  return parseInt(value[1].split(']')[0])
```
Poltergeist: The two lines above could better be solved with a regular expression to be clearer in what is happening.
```JS
parseInt(key.match(\[(\d*)\])[1], 10)
```
And to be a bit more error safe there is also this option
```JS
parseInt(key.match(\[\d*\]).join().replace(/\[|\]/, ''), 10)
```

src/js/utils/FormUtil.js
```diff
@@ -0,0 +1,89 @@
...
+}
+
+function getPropKey(key) {
+  let value = key.split('[');
+  return value[1].split('.')[1];
+}
```
Poltergeist: In this case you are trying to find the last string after a `.` or the the first one after a `[` ?

src/js/utils/FormUtil.js
```diff
@@ -0,0 +1,89 @@
...
+
+function isNotMultipleProp(key) {
+  return !key.includes('[') || !key.includes(']');
```
Poltergeist: This also exludes keys with the name `uid[]` which is properly never going to happen.

src/js/utils/FormUtil.js
```diff
@@ -0,0 +1,89 @@
...
+}
+
+const FormUtil = {
+  getMultipleFields: function (prop, count, definition) {
```
orlandohohmeier: This is more a create multiple fields definition than a get fields functions, isn't it?

src/js/utils/FormUtil.js
```diff
@@ -0,0 +1,89 @@
...
+    let definitions = [];
+    let multipleDefinition = Array.isArray(definition);
```
orlandohohmeier: Please use `isArray` directly or rename this to something like (is/has)MultipleDefinitions.

src/js/utils/FormUtil.js
```diff
@@ -0,0 +1,89 @@
...
+
+    for (let i = 0; i < count; i++) {
+      let definitionInstance;
+      if (multipleDefinition) {
+        definitionInstance = cloneDefinition(definition).map(function (definitionClone) {
```
orlandohohmeier: It's not an instance of a definition, is it? Please consider a different name.

orlandohohmeier: Why are you cloning the definitions array? You could simple use map (which will create a new array) and then create   new definitions instead of altering them.

src/js/utils/FormUtil.js
```diff
@@ -0,0 +1,89 @@
...
+          definitionClone.name = `${prop}[${i}].${definitionClone.name}`;
+
+          return definitionClone;
+        });
+      } else {
+        definitionInstance = Object.assign({}, definition);
+        definitionInstance.name = `${prop}[${i}]`;
```
orlandohohmeier: You're overwriting the definition name here. That could lead to interesting bugs.

src/js/utils/FormUtil.js
orlandohohmeier: Why do we need the prop? Shouldn't be the index enough to group fields?

src/js/utils/FormUtil.js
pierlo-upitup: As long as the regex is tested and with a clear, human readable description (or maybe even a comment) then I agree.

src/js/utils/__tests__/FormUtil-test.js
```diff
@@ -0,0 +1,79 @@
...
+          name: 'uid'
+        });
+      });
+
+      it('returns the correct amount of fields', function () {
+        expect(this.result.length).toEqual(2);
+      });
+
+      it('returns the correct names for fields', function () {
+        expect(this.result[0].name).toEqual('uid[0]');
```
orlandohohmeier: I think the result is wrong.  I'd expoerc the follwing result `uid[0].uid`. See https://github.com/dcos/dcos-ui/pull/298/files#r65526886 for reference.

src/js/utils/FormUtil.js
kennyt: yeah :)

src/js/utils/FormUtil.js
kennyt: after the `.` so `uid[1].name` will return `name`

src/js/utils/FormUtil.js
kennyt: cloning the definition so i can modify the name which i do below

src/js/utils/FormUtil.js
kennyt: cloning the definition so i can modify the name which i do below

src/js/utils/__tests__/FormUtil-test.js
kennyt: this is for one property. if you wanted to have multiple properties, it would then lead to that. (there are tests below this for multiple properties)

orlandohohmeier: It should produce the same result no matter how many fields you have. There is no reason to have different results.

kennyt: well it would be confusing because if there are multiple props, it turns into an object.

it'll be:
`ports: [{name: xyz, password: secret}]`


rather than if it's just one prop it'll straight up be the value:
`ports: [100, string, something]`

orlandohohmeier: I'd say it should always return the same type and the data should always follow the same structure. Otherwise it's pretty hard to use the util one has to add multiple type checks as the results are kinda unpredictable.

orlandohohmeier: I was also wondering if we really need the the prop to group things. Wouldn't the index be enough?

pierlo-upitup:
I think it's important to keep in mind what is the ultimate use-case we are trying to shoot for, here. Ideally, I would like to have a way to have support for multi-field duplicable rows out of the box. 

Imagine the `Environmental Variables` section of the create service modal. Every entry there is composed of 2 fields, `key` and `value`. The 2 things together compose an entry. They only make sense as a tuple. Now, these can be duplicated, added in-place (insert in position) and deleted individually. 

With all that in mind, imagine if we could just write something like this as part of our `ServiceSchema.js`:

```
// ...
'Environmental Variables': {
  description: 'Environmental variables',
  type: 'object',
  properties: {
    env: {
      type: 'tuple',//  <--  or maybe "shape" or whatever
      duplicable: true, // <-- magic!
      properties: {
         key: { // ... },
         value: {  // ... }
      }
    }
  }
}
```
The duplicable row mechanism should be transparent to us. What do you think?

src/js/utils/FormUtil.js
```diff
@@ -0,0 +1,89 @@
...
+      return object.slice();
+    }
+
+    if (typeof object === 'object') {
+      return Object.assign({}, object);
+    }
+
+    return object;
+  });
+}
```
pierlo-upitup: We should consider moving this to a more generic file (`Util.js`? ). We may need a good solution for deep clones also in other sections. See https://github.com/mesosphere/marathon-ui/blob/master/src/js/helpers/Util.js#L132

src/js/utils/FormUtil.js
```diff
@@ -0,0 +1,89 @@
...
+function getPropKey(key) {
+  let value = key.split('[');
+  return value[1].split('.')[1];
+}
+
+function isNotMultipleProp(key) {
+  return !key.includes('[') || !key.includes(']');
+}
+
+const FormUtil = {
```
pierlo-upitup: Let's be mindful of out-of-control Util proliferation. Can't we stick to one of the already existing `SchemaFormUtil.js` or `SchemaUtil`?

src/js/utils/__tests__/FormUtil-test.js
kennyt: i removed this method from the util

src/js/utils/FormUtil.js
kennyt: this no longer exists. i ported over `deepCopy` from marathon-ui (good stuff guys)

src/js/utils/FormUtil.js
kennyt: unfortunately, i can't use either of those because they are used on schema-related stuff. this is form-related stuff

src/js/utils/FormUtil.js
```diff
@@ -0,0 +1,209 @@
...
+function getPropIndex(key) {
+  let value = key.split('[');
+  return parseInt(value[1].split(']')[0])
+}
+
+function getPropKey(key) {
+  return key.split('.')[1];
+}
+
+function isNotMultipleProp(key) {
```
rcorral: what is a multiple prop? Could you add a comment to some of these functions?

src/js/utils/FormUtil.js
```diff
@@ -0,0 +1,209 @@
...
+  }
+
+  return newDefinition;
+}
+
+const FormUtil = {
+  /**
+   * @return {Array} parentDefinition the modified definition.
+   * @param {Object} options Object containing properties for
+   * transformingDefinition. The following properties are:
```
rcorral: Interesting organization of JDOC, could we keep [it consistent](https://github.com/dcos/dcos-ui/blob/master/src/js/utils/TableUtil.js#L59) with others?

rcorral: It would also be nice to have a blurb about what this method should be used for

src/js/utils/FormUtil.js
```diff
@@ -0,0 +1,209 @@
...
+      throw 'Need prop, parent definition, and item definition to transform.'
+    }
+
+    if (instancesIndex == null) {
+      instancesIndex = 0;
+    }
+
+    if (propID == null) {
+      propID = Util.uniqueID(prop);
+    }
```
rcorral: You could move this `if` inside the `else` block in line 97, as you only use it there.

kennyt: done

src/js/utils/FormUtil.js
kennyt: done

src/js/utils/FormUtil.js
kennyt: done

PR #299, opened 2016-06-01
[WIP] [Feedback Requested] Introduce ExpandingTable component

jfurrow:
I'm introducing a component that allows table rows to be expanded & collapsed. I have a couple of concerns about the best approach for some things and wanted to get everyone's feedback before fully committing. Here's where it will be implemented: 

Here's the component with a very simple toggle button: 

Here's a component you can copy/paste to test it: https://gist.github.com/jfurrow/30934b2d553f94d5009c61a58c56bdc8

Limitations:
* It doesn't generate new `<tr>` elements as you might expect. It simply renders additional elements inside each `<td>`. This is a decision @rcorral and I made to keep it simple. The downside to this is that when a row is expanded, all child nodes must be the exact same height, or the horizontal alignment will be off. It also means that text in the child nodes cannot wrap.
* The toggle button for expanding the row is implemented by the component that implements the table (via the `#render` method that is passed in with the `columns` — example in the above gist).
* It requires each row in the `data` to have a unique `id` attribute. In the `Table` component itself, we construct table cell IDs, so we could do something like that here too, but I thought explicit IDs would be safer.

Let me know what you guys think of this. If it looks good, I'll write tests and create a proper PR.

### DIFF
--- a/src/js/components/ExpandingTable.js
+++ b/src/js/components/ExpandingTable.js
@@ -1,0 +1,94 @@
+import React from 'react';
+import {Table} from 'reactjs-components';
+
+class ExpandingTable extends React.Component {
+  constructor() {
+    super(...arguments);
+
+    this.state = {
+      expandedRows: []
+    };
+  }
+
+  defaultRenderer(prop, row) {
+    return row[prop];
+  }
+
+  expandRow(row) {
+    let {expandedRows} = this.state;
+    let rowID = this.getRowID(row);
+    let selectedRowIndex = expandedRows.indexOf(rowID);
+
+    // If the selected row is already expanded, then we want to collapse it.
+    if (selectedRowIndex > -1) {
+      expandedRows.splice(selectedRowIndex, 1);
+    } else {
+      expandedRows.push(rowID);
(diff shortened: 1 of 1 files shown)


mesosphere-web:
Reviewer: @daiyiwhale

orlandohohmeier:
I think @philipnrmn also set out to implement an expanding table for the deployments. Maybe you guys could join forces.

philipnrmn:
Hey @jfurrow, nice work!

Your summary of the advantages/disadvantages of the rows versus child of `<td>` is also my experience from the `DeploymentsTable`. I think my ideal implementation would use separate rows and scoping, but  I found that this was too big a job to retrofit to the `Table` component so I left it as-is. 

What would be super useful though is to be able to pass the `Table` component a list of more than one type and have appropriate rows rendered, so we didn't have to fit the data to the component but could just pass the result of a store get directly.

mlunoe:
Nice work @jfurrow!

jfurrow:
Closing this and re-opening a new, official PR.

PR #312, opened 2016-06-02
Adjust validate-tests script

orlandohohmeier:
Change the validation pattern to check for the new Jasmin 2 `fit` instead of the old `it.only`.

Run the following command to test the script. It should echo "valid" if there is no `f(it|describe)` nor `(context|it).only` and "invalid" if you add "f" or "only" to one of the tests.
```
eval scripts/validate-tests && echo "valid" || echo "invalid"
```

### DIFF
--- a/scripts/validate-tests
+++ b/scripts/validate-tests
@@ -1,3 +1,3 @@
 #!/usr/bin/env bash
-! egrep --quiet -R "describe\.only\(|it\.only\(" ./src && \
+! egrep --quiet -R "fdescribe\\(|fit\(" ./src && \
 ! egrep --quiet -R "context\.only\(|it\.only\(" ./tests


mesosphere-web:
Reviewer: @philipnrmn

scripts/validate-tests
```diff
@@ -1,3 +1,3 @@
 #!/usr/bin/env bash
-! egrep --quiet -R "describe\.only\(|it\.only\(" ./src && \
-! egrep --quiet -R "context\.only\(|it\.only\(" ./tests
```
kennyt: i believe this line is for cypress. did cypress also change?

orlandohohmeier: Whoops! You're right, Cypress is using the Mocha which hasn't changed.

kennyt:
i don't know why integration tests are failing. i'll merge once they pass (just rebuilt it)

pierlo-upitup:
retest this please

PR #313, opened 2016-06-02
Move jasmine.clock installation to specific tests

jfurrow:
Previously we were running `jasmine.clock().install()` in `setupTestFramework`, which worked for some tests, but one of those tests altered the behavior of timing functions for all subsequent tests.

This PR will "uninstall" all application timers (which are created with `global.setTimeout` and `global.setInterval`), then "install" the mock Jasmine timers. It looks funny to see the `uninstall` and `install` methods called consecutively, but it's necessary as far as I can tell due to our use of `global.set{Interval|Timeout}`. Read about it here: https://github.com/jasmine/jasmine/issues/895#issuecomment-127370465

### DIFF
--- a/src/js/stores/__tests__/ChronosStore-test.js
+++ b/src/js/stores/__tests__/ChronosStore-test.js
@@ -17,16 +17,21 @@ const JobTree = require('../../structs/JobTree');
 describe('ChronosStore', function () {
 
   beforeEach(function () {
+    // Clean up application timers
+    jasmine.clock().uninstall();
+    // Install our custom jasmine timers
+    jasmine.clock().install();
     ChronosActions.fetchJobs = jasmine.createSpy('fetchJobs');
   });
 
   afterEach(function () {
+    jasmine.clock().uninstall();
     ChronosStore.removeAllListeners();
   });
 
   describe('constructor', function () {
 
-    it('should call the fetchJobs 2 times', function () {
+    it('should call the fetchJobs 3 times', function () {
       ChronosStore.addChangeListener(
         EventTypes.CHRONOS_JOBS_CHANGE,
         function () {}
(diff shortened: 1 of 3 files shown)


mesosphere-web:
Reviewer: @Poltergeist

src/js/stores/__tests__/ChronosStore-test.js
```diff
@@ -35,7 +40,7 @@ describe('ChronosStore', function () {
       jasmine.clock().tick(2 * Config.getRefreshRate());
       // Finish up outstanding timers
       jest.runOnlyPendingTimers();
-      expect(ChronosActions.fetchJobs.calls.count()).toEqual(2);
+      expect(ChronosActions.fetchJobs.calls.count()).toEqual(3);
```
jfurrow: This test needs to be changed to 3 calls of `#fetchJobs`. `#fetchJobs` is called immediately when `setInterval` is initiated, then twice more due to the mocked passage of time (`2 * Config.getRefreshRate()`).

src/js/stores/__tests__/ChronosStore-test.js
```diff
@@ -17,16 +17,21 @@ const JobTree = require('../../structs/JobTree');
 describe('ChronosStore', function () {
 
   beforeEach(function () {
+    // Clean up application timers
+    jasmine.clock().uninstall();
```
Poltergeist: If the uninstall is called in the `afterEach` do we really need it again in the `beforeEach`? If that is not the case I would like to have it rather in the `afterEach` then in the `beforeEach` since it is a tear down task.

Poltergeist:
One nit™ but else it looks fine.

src/js/stores/__tests__/ChronosStore-test.js
jfurrow: Thanks for trying this out! I removed it from `afterEach`

PR #314, opened 2016-06-02
Add Nested Service Links Component

MatApple:
Splits an AppId into a bunch of links to the service and parent groups. Optionally takes a taskId, creating clickable links for the task and parent service/groups.

Currently restricted to use within a table cell using existing classes. Could be altered without much trouble to work outside of a table but I don't currently see where we'd do that.

### DIFF
--- a/src/js/components/NestedServiceLinks.js
+++ b/src/js/components/NestedServiceLinks.js
@@ -1,0 +1,178 @@
+import classNames from 'classnames/dedupe';
+import {Link} from 'react-router';
+import React, {PropTypes} from 'react';
+
+class NestedServiceLinks extends React.Component {
+
+  getMinorLink(label, params, key, classes) {
+    return (
+      <div key={key} className="table-cell-value">
+        <div className={classes}>
+          <Link to="services-detail"
+              params={params}
+              title={label}>
+            {label}
+          </Link>
+        </div>
+      </div>
+    );
+  }
+
+  getCrumbDivider(key) {
+    return (
+      <div key={key}>
+        <i className="icon icon-sprite icon-sprite-mini icon-chevron flush" />
+      </div>
+    );
+  }
+
+  getMinorLinks() {
+    let componentKey = 0;
+    let {minorLinkClassName, taskID} = this.props;
+    let minorLinkClasses = classNames(
+      'text-overflow service-link',
+      minorLinkClassName
+    );
+    let nestedGroups = this.getServicePathParts();
+    let popCount = 1;
+
+    if (taskID != null) {
+      popCount = 0;
+    }
+    nestedGroups = nestedGroups.slice(0, nestedGroups.length - popCount);
+
+    let links = nestedGroups.reduce((components, part, index) => {
+      let id = encodeURIComponent(nestedGroups.slice(0, index + 1).join('/'));
+
+      components.push(
+        this.getMinorLink(part, {id}, componentKey++, minorLinkClasses)
+      );
+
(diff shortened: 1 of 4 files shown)


mesosphere-web:
Reviewer: @jfurrow

src/js/components/NestedServiceLinks.js
```diff
@@ -0,0 +1,178 @@
...
+  taskID: null,
+  className: 'inverse',
+  minorLinkClassName: 'inverse'
+};
+
+NestedServiceLinks.propTypes = {
+  serviceID: PropTypes.string.isRequired,
+  taskID: PropTypes.string,
+  // Classes
+  className: PropTypes.oneOfType([
```
mlunoe: We're missing array here as an option for classes

src/js/components/NestedServiceLinks.js
```diff
@@ -0,0 +1,178 @@
...
+        <div className={minorLinkWrapperClasses}>
+          {this.getMinorLinks()}
+        </div>
+      </div>
+    );
+  }
+}
+
+NestedServiceLinks.defaultProps = {
+  taskID: null,
```
mlunoe: Is there a reason for this being explicitly set to `null`? I believe we can leave it out

mlunoe:
Really nice work @MatApple!

src/js/components/NestedServiceLinks.js
mlunoe: Maybe we can use this construct: https://github.com/mesosphere/reactjs-components/pull/279/files#diff-312372494fa36bc00aa112862d70f194R37

src/styles/components/nested-service-links.less
```diff
@@ -0,0 +1,26 @@
...
+--------------------------------------------------------------------------------
+
+NestedServiceLinks
+
+--------------------------------------------------------------------------------
+----------------------------------------------------------------------------- */
+
+.nested-service-link-container {
+  padding-top: @base-spacing-unit * 0.5 !important;
+  padding-bottom: @base-spacing-unit * 0.5 !important;
```
jfurrow: Why the `!important` rule?

MatApple: We are! It's the padding on the table cell itself. Open to better ideas :)

src/js/components/NestedServiceLinks.js
MatApple: I just like that its explicit because we rely on it being falsy or truthy. Not declaring it works too but this tells me the prop is used if provided or not.

src/js/components/NestedServiceLinks.js
mlunoe: @MatApple Any update on this?

MatApple: @mlunoe done! Great suggestion thanks 👍

src/styles/components/nested-service-links.less
jfurrow: I don't have a great solution. But, I do think we should try to override the Canvas specificity by adding more selectors rather than using `!important`. Although it's just another hack to increase the specificity, it's not as bad (IMO) as `!important`

pierlo-upitup:
This is really cool. Let's use it in the search results page!

src/styles/components/nested-service-links.less
MatApple: Good call @jfurrow - done!

jfurrow:
Nice! Feel free to merge when tests pass! :ship:

PR #317, opened 2016-06-02
Using StoreMixin for ChronosStore

mlunoe:


mesosphere-web:
Reviewer: @kennyt

src/js/stores/ChronosStore.js
```diff
@@ -16,8 +16,6 @@ import {
   SERVER_ACTION
 } from '../constants/ActionTypes';
 
-import VisibilityStore from './VisibilityStore';
```
jfurrow: We still use `VisibilityStore` in here to check visibility

mlunoe: I was too fast 🏇

ssk2:
Retest this please.

ssk2:
retest this please

ssk2:
retest this please

ssk2:
retest this please

src/js/stores/ChronosStore.js
```diff
@@ -66,10 +66,7 @@ class ChronosStore extends EventEmitter {
       return true;
     });
 
-    VisibilityStore.addChangeListener(
-      VISIBILITY_CHANGE,
-      this.onVisibilityStoreChange.bind(this)
-    );
+    this.store_initializeListeners([{name: 'visibility', events: ['change']}]);
```
orlandohohmeier: Do we really wan't to introduce the store mixin only to add a listener to the visibility store? I thought our overall plan is to "remove" mixins...  I'd rather introduce something like the recurrent task util (https://github.com/dcos/dcos-ui/pull/91) to handle this in a clean consistent way. What do you think?

orlandohohmeier:
Do we really want to introduce the store mixin only to add a listener to the visibility store? I thought our overall plan is to "remove" mixins...  I'd rather introduce something like the recurrent task util (https://github.com/dcos/dcos-ui/pull/91) to handle this in a clean consistent way. What do you think?

kennyt:
@orlandohohmeier i thought we agreed to have some mixins. like the `StoreMixin`. we have many stores already using the `StoreMixin` to listen to other stores atm (like this PR)

orlandohohmeier:
@kennyt Yeah, sorry my comment was misleading. I meant, that we agreed (if I recall correctly) to be cautions with the usage and introduction of mixins and that we consider the alternatives before introducing new ones. 
If we want to move towards this store pattern, let's make sure that we have a rollout plan in place and discuss this before introduce it, as there aren't any stores that use the store mixin, at least as far as I can tell. 

/cc @dcos/frontend

kennyt:
@orlandohohmeier apologies, all of the stores that use `StoreMixin` are inside of the private plugins. no store in dcos-ui uses it because there's been no need to until now i suppose

rcorral:
Hey guys, you're both right! Let's have a rollout plan for this as there's many others stores that can benefit from this!

mlunoe:
This change is global. The ones in the plugins have already landed

mlunoe:
Guys, since change is already in the plugins, could we get this in, and then make a decision whether we want to change it? @rcorral @orlandohohmeier

orlandohohmeier:
@mlunoe Do we need this now? Otherwise I'd suggest to wait with this adjustment and the needed discussion (which might cost use more time than we want) until we've implemented all the features for 1.8. What do you think?

rcorral:
I agree with @orlandohohmeier we should wait on this rather than adding more inconsistencies. It's not something that we need, but an improvement. Let's rolls this out globally later and let's plan for it so it doesn't add additional unplanned work.

MatApple:
@orlandohohmeier I actually really like the Util you proposed.

MatApple:
Lets close this for another day.

PR #330, opened 2016-06-06
Introduce STATUS sidebar filter

pierlo-upitup:
Adds support for filtering services by STATUS.

filters

*NOTES*

1. I really dislike the current coupling we have going on inside `Tree#filterItemsByFilter` where we have an `if` per filter. The Tree itself should provide a generic mechanism to filter items, not the specific implementations an instance of the Tree (in our case a ServiceTree) may or may not want to support. This is a no-no and we should refactor this later on. (**see below**)

2. I re-enabled the Cypress tests for `SidebarFilter` by simply using `cy.viewport` to set a minimum working width until we have the responsive designs ready. I consider this a preferable approach instead of just shutting down the tests.

screen shot 2016-06-06 at 12 07 31

UPDATE
----

As agreed with @orlandohohmeier I will create another PR to move the `filterItemsByFilter` method to the `ServiceTree`. This is an initial effort to decouple the service-specific filtering logic from the generic Tree class.

### DIFF
--- a/src/js/constants/ServiceStatusLabels.js
+++ b/src/js/constants/ServiceStatusLabels.js
@@ -1,7 +1,7 @@
-var StatusLabels = {
+var ServiceStatusLabels = {
   RUNNING: 'Running',
   DEPLOYING: 'Deploying',
   SUSPENDED: 'Suspended'
 };
 
-module.exports = StatusLabels;
+module.exports = ServiceStatusLabels;
--- a/src/js/structs/ServicesList.js
+++ b/src/js/structs/ServicesList.js
@@ -25,14 +25,6 @@ class ServicesList extends List {
           });
         });
       }
-
-      if (filters.status != null && filters.status.length !== 0) {
-        services = services.filter(function (service) {
-          return filters.status.some(function (statusValue) {
-            return service.getServiceStatus().key === parseInt(statusValue, 10);
-          });
-        });
-      }
     }
 
     return new ServicesList({items: services});
--- a/src/js/structs/ServicesList.js
+++ b/src/js/structs/ServicesList.js
@@ -25,6 +25,14 @@ class ServicesList extends List {
           });
         });
       }
+
+      if (filters.status != null && filters.status.length !== 0) {
+        services = services.filter(function (service) {
+          return filters.status.some(function (statusValue) {
+            return service.getServiceStatus().key === parseInt(statusValue, 10);
+          });
+        });
+      }
     }
 
     return new ServicesList({items: services});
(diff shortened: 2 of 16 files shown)


mesosphere-web:
Reviewer: @philipnrmn

src/js/structs/ServicesList.js
```diff
@@ -25,6 +25,14 @@ class ServicesList extends List {
...
         });
       }
+
+      if (filters.status != null && filters.status.length !== 0) {
+        services = services.filter(function (service) {
+          return filters.status.some(function (statusValue) {
+            return service.getServiceStatus().key === parseInt(statusValue, 10);
+          });
+        });
+      }
```
pierlo-upitup: @orlandohohmeier is this what you mean?

orlandohohmeier: @pierlo-upitup No. Sorry, I wasn't clear enough. I was referring to [`Tree.prototype.filterItems`](https://github.com/dcos/dcos-ui/blob/f156ca9b6449f39dbaf6687a951815ad53a21639/src/js/structs/Tree.js#L53) which is the method that is used to implement the [`ServiceTree.prototype.findItemById`](https://github.com/dcos/dcos-ui/blob/f156ca9b6449f39dbaf6687a951815ad53a21639/src/js/structs/ServiceTree.js#L98) as well as `filterItemsByText` . But I wouldn't introduce any new methods on the tree and rather use `filterItems` directly as the filter behavior is very specific to the service table and not necessarily part of the data structure.

orlandohohmeier:
I get lot's of type error, due to missing `getServiceStatus
TypeError

pierlo-upitup:
Got it. I missed the `ServiceTree#getServiceStatus` implementation.

pierlo-upitup:
Integration tests are working locally.

philipnrmn:
LGTM. Only very slight concern would be the naming of `StatusTypes` and `StatusLabels` - do we want to make it clear that they are specific to Services?

pierlo-upitup:
@philipnrmn that's a really good point – in fact it crossed my mind as I was adding those. Will rename them to `ServiceStatus*`.

src/js/constants/ServiceStatusLabels.js
```diff
@@ -0,0 +1,7 @@
+var StatusLabels = {
```
philipnrmn: Shall we do the variable names just for completeness?

pierlo-upitup:
Velocity is reporting:

```
 73 passing (7m)
  19 pending
  2 failing

  1) Service Search Filters Filters services table sets the correct search string filter query params:
     Error: Cypress command timeout of '4270ms' exceeded.
      at http://localhost:2020/__cypress/static/js/cypress.js:3254:31

  2) Service Search Filters Filters services table will clear filters by clear all link click:
     Error: Cypress command timeout of '4270ms' exceeded.
      at http://localhost:2020/__cypress/static/js/cypress.js:3254:31
```

But I have no such issues when running Cypress tests locally.

pierlo-upitup:
retest this please

philipnrmn:
No I also ran them locally before labelling. How frustrating.

PR #331, opened 2016-06-06
Fix SchemaForm SideTabs

Poltergeist:
With the introduction of the new title attribute for schema sections there was a bug introduced which made it impossible to navigate to a section with a title attribute.

PR #332, opened 2016-06-06
Move service schema to config directory

Poltergeist:
This PR moves the service schema files to the ~~`config`~~ `schemas` directory and also uses the new title attribute. Finally the `Container Settings` do also get their own file.

Depends on the fix in #331 to work.


mesosphere-web:
Reviewer: @pierlo-upitup

Poltergeist:

aldipower:
In my opinion the config-directory is reserved for application-wise configurations like the config.dev.js.
Putting a scheme into the config-dir I think we mix the things up.
Calling it `schemes` would be much moar obvious to me. :)

aldipower:
The title-attr is great btw! :)

pierlo-upitup:
I agree with @aldipower – `config/` does not strike me as being the right place for this.

Poltergeist:
I addressed the concerns about the config directory and moved the files to a new directory called `schemas`. My idea is to have a sub directory for each schema and a main file in the `schemas` directory itself
```
schemas
  |__ ServiceSchema.js
  |__ service-schema/
      |__ general.js

pierlo-upitup:
retest this please

aldipower:
:+1:

PR #333, opened 2016-06-06
Feature - 1066 Service actions - Buttons

aldipower:
This introduces the service actions buttons/dropdowns -> Edit, Scale, Suspend, Destroy .
There is no underlaying functionality yet for the sake of atomic PRs. This will follow next.

service-dropdown

As you can see the dropdown needs some polishing in the reactjs-components itself, it's displayed to widely and is hits the right boundary.

@MatApple Thx for preparing the `actionButtons` property! :)


mesosphere-web:
Reviewer: @jfurrow

Poltergeist:
Looks good to me! Only one question why is `getDropdownItems` a function I mean the array could also be a instance attribute or a constant. Or a local variable. Why did you solved this with a function?

aldipower:
Good point! Doesn't make that much sense. :smile:

src/js/components/ServiceInfo.js
```diff
@@ -7,7 +8,73 @@ import PageHeader from './PageHeader';
...
+        Edit
+      </button>
+    ), (
+      <Dropdown
+        key="actions-dropdown"
+        buttonClassName="button button-stroke button-inverse dropdown-toggle"
+        dropdownMenuClassName="dropdown-menu"
+        dropdownMenuListClassName="dropdown-menu-list"
+        dropdownMenuListItemClassName="clickable"
+        wrapperClassName="dropdown"
```
jfurrow: You can add `anchor-right` to the `wrapperClassName` prop which will prevent it from running off the screen

src/js/components/ServiceInfo.js
```diff
@@ -7,7 +8,73 @@ import PageHeader from './PageHeader';
...
+  onActionsItemSelection(item) {
+    if (item.id === ServiceActionItem.EDIT) {
+      this.setState({
+        isServiceFormModalShown: true
+      });
+    }
+  }
+
+  getActionButtons() {
+    const dropdownItems = [{
```
kennyt: can we move this to the top of the file since it's a `const`?

src/js/components/ServiceInfo.js
pierlo-upitup: Let's do this!

pierlo-upitup:
Existing nitpicks asides, this LGTM. Change & Merge!

I'm OK with skipping tests for now, but we should have integration tests for each individual action once they're implemented.

src/js/components/ServiceInfo.js
aldipower: Whoop whoop! This works nicely. Thanx for the hint @jfurrow .

rightaligned

src/js/components/ServiceInfo.js
aldipower: Nope, it will change it's contents in a later PR.

PR #336, opened 2016-06-06
Introduce service task failure data

orlandohohmeier:
Fetch service task (last) failure data alongside groups and provide getter - this is needed to implement the service debug tab.

/cc @mlunoe


aldipower:
I like it.

PR #339, opened 2016-06-06
Introduce n/a status as a fallback

pierlo-upitup:
This prevents exceptions when none of the existing statuses match.

### DIFF
--- a/src/js/constants/ServiceStatusTypes.js
+++ b/src/js/constants/ServiceStatusTypes.js
@@ -1,7 +1,8 @@
 var ServiceStatusTypes = {
   RUNNING: 0,
   DEPLOYING: 1,
-  SUSPENDED: 2
+  SUSPENDED: 2,
+  NA: 4
 };
 
 module.exports = ServiceStatusTypes;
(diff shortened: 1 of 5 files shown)


mesosphere-web:
Reviewer: @daiyiwhale

src/js/constants/ServiceStatusTypes.js
```diff
@@ -1,7 +1,8 @@
 var ServiceStatusTypes = {
   RUNNING: 0,
   DEPLOYING: 1,
-  SUSPENDED: 2
+  SUSPENDED: 2,
+  NA: 4
```
aldipower: In marathon-ui the `4` is `waiting`. Maybe we should have this in mind, not sure. Feel free to ignore.

pierlo-upitup:
Yeah, but it seems to me that we don't use the key for any sorting here so I guess we're good.

src/js/constants/ServiceStatusTypes.js
pierlo-upitup: Yeah, but it seems to me that we don't use the key for any sorting here so I guess we're good.

orlandohohmeier:
Could you please add a test verify taht `getServiceStatus` returns n/a if nothing matching. Rest looks good to me!

PR #346, opened 2016-06-06
ServiceForm & Duplicate Rows

kennyt:
px

OK, there can probably be some styling improvements

In order to test this PR you need to edit a vendor file. In `node_modules/reactjs-components/lib/Form/FormControl.js` on line `95` add this:
``` JS     
      if (_react2.default.isValidElement(definition)) {
        return definition;
      }
```


mesosphere-web:
Reviewer: @aldipower

src/js/components/ServiceForm.js
```diff
@@ -0,0 +1,230 @@
...
+    let {model, schema} = this.props;
+    let definition = SchemaUtil.schemaToMultipleDefinition(
+      schema, this.getSubHeader, this.getLabel
+    );
+
+    if (model) {
+      SchemaFormUtil.mergeModelIntoDefinition(model, definition);
+    }
+
+    this.transformPorts(definition, model.environmentVariables.ports);
```
Poltergeist: Could this be automatic? I mean for all array typed fields, this would reduce redundancy when we add new array typed fields. That we have to add it in multiple places.

src/js/components/ServiceForm.js
```diff
@@ -0,0 +1,230 @@
...
+        onClick={this.handleRemoveRow.bind(this, generalDefinition, prop, id)}>
+        X
+      </button>
+    );
+  }
+
+  getModel() {
+    let model = this.triggerTabFormSubmit();
+    model = SchemaFormUtil.processFormModel(model, this.multipleDefinition);
+    model.environmentVariables = FormUtil.modelToCombinedProps(
```
Poltergeist: Can we make this automatic so that all sections which have combined Props do this. Rather then the need to add those fields here manually?

src/js/components/ServiceForm.js
```diff
@@ -0,0 +1,230 @@
...
+      newDefinition
+    );
+    newDefinition.push(
+      this.getRemoveVariableButton(definition, prop, propID)
+    );
+    definition.splice(definition.length - 1, 0, newDefinition);
+    this.forceUpdate();
+  }
+
+  transformPorts(definition, startValues) {
```
Poltergeist: This is tight coupled with one definition, is it possible to generalize this, I would like to have this for all array fields.

pierlo-upitup:
This PR depends on https://github.com/mesosphere/reactjs-components/pull/284

src/js/components/ServiceForm.js
```diff
@@ -0,0 +1,207 @@
...
+import GeminiUtil from '../utils/GeminiUtil';
+import SchemaFormUtil from '../utils/SchemaFormUtil';
+import SchemaUtil from '../utils/SchemaUtil';
+import TabForm from './TabForm';
+import Util from '../utils/Util';
+
+const METHODS_TO_BIND = [
+  'validateForm', 'getTriggerTabFormSubmit', 'handleFormChange',
+  'handleExternalSubmit', 'getRemoveVariableButton', 'getAddNewRowButton'
+];
```
pierlo-upitup: shouldnt these be each on their own line?

src/js/components/ServiceForm.js
```diff
@@ -0,0 +1,207 @@
...
+
+  getModel() {
+    let model = this.triggerTabFormSubmit();
+    return SchemaFormUtil.processFormModel(model, this.multipleDefinition);
+  }
+
+  getNewDefinition() {
+    let {model, schema} = this.props;
+    let definition = SchemaUtil.schemaToMultipleDefinition(
+      schema, this.getSubHeader, this.getLabel, this.getRemoveVariableButton, this.getAddNewRowButton
```
pierlo-upitup: this line breaks the 80 chars

src/js/components/ServiceForm.js
```diff
@@ -0,0 +1,207 @@
...
+    );
+
+    if (model) {
+      SchemaFormUtil.mergeModelIntoDefinition(model, definition, this.getRemoveVariableButton);
```
pierlo-upitup: This one too

src/js/components/ServiceForm.js
```diff
@@ -0,0 +1,207 @@
...
+        className="button"
+        onClick={
+          this.handleAddRow.bind(this, prop, generalDefinition, definition)
+        }>
+        Add New Variable
+      </button>
+    );
+  }
+
+  getRemoveVariableButton(generalDefinition, prop, id) {
```
pierlo-upitup: Is it about Envvars? then it doesnt belong here.

Poltergeist: This is about the duplicatable rows. It is generalised and handles all of the rows.

pierlo-upitup: So what does `variable` mean here?

Poltergeist: You are right I looked at the wrong line. Changed this to default to `Add New Line` but made it configurable via the schema.

orlandohohmeier:
What was the reasoning behind the introduction of a new `ServiceForm` component, which replicate lots of the `SchemaForm`, instead of extending the existing component? I'd prefer the latter, as we will need the very same functionality for the job form and could use general solution.

src/js/components/ServiceForm.js
```diff
@@ -0,0 +1,225 @@
...
+  componentWillUnmount() {
+    // Unschedule all validation if component unmounts.
+    if (this.timer) {
+      clearTimeout(this.timer);
+    }
+  }
+
+  componentDidUpdate() {
+    // Timeout necessary due to modal content height updates on did mount
+    setTimeout(() => {
```
orlandohohmeier: Let's use `requstAnimationFrame` as we want to wait until stuff is render. and the modal height is updated.

src/js/components/ServiceForm.js
```diff
@@ -134,9 +92,11 @@ class ServiceForm extends React.Component {
     );
   }
 
-  getModel() {
+  getDataTriple() {
```
orlandohohmeier: Can we please rename this method? I'd liked `getModel` better...

orlandohohmeier:
@Poltergeist I think extending the `SchemaForm` is a step in the right direction, but we still need a general solution for schema based forms that allow duplicable fields. But thats maybe not in the scope of this PR. What do you think?

src/js/utils/SchemaFormUtil.js
```diff
@@ -1,15 +1,47 @@
...
+    definitionToSet.definition.find(function (field, i) {
+      if (FormUtil.isFieldInstanceOfProp(prop, field)) {
+        firstIndex = i;
+        return true;
+      }
+
+      return false;
+    });
+
+    FormUtil.removePropID(definitionToSet.definition, prop);
```
orlandohohmeier: Why do we need to remove props here? Doe we mutate the respective definitions? Can't we simply create a proper set of definitions with out that prop?

orlandohohmeier:
I'm not sure I understand all the nifty details of this, as the `SchemaFrom` is highly complex and in parts a little confusing. I'd love a team session to learn all the details of it.  Anyway... Besides a few nits and comments and my wish for a general form that allows duplicable fields, I'd say this looks pretty good.  
Let's get this merge and then refactor this into a general `DuplicableFieldsSchemaForm` _(we definitely need a better name)_ component, which we can use for the jobs as well.

src/js/utils/SchemaFormUtil.js
Poltergeist: This is used if a row is present in the model, (think edit modal). What it does is removing an empty field if there are already fields provided. 

With this function it looks like:
image

If the function is not called it would look like:

image

orlandohohmeier: Ok, than let's adjust the name to something "meaningful"... What about 
`FormUtil.removeEmptyFields` or something similar?

kennyt: actually `removePropID` will completely remove all fields if they match regardless if they have values or not. but we do this in the beginning so we always remove empty rows.

(you're probably going to use this method when you're creating health checks and someone switches the dropdown value. you blow away all fields for that certain instance `healthCheck[0]` and then create new fields for `healthCheck`s

kennyt:
removed the environment variables logic, which will be in a new PR

orlandohohmeier:

orlandohohmeier:
@kennyt Thanks for refactoring this! Looks pretty good. Let's :ship: 🇮🇹

Poltergeist:
I really like the final implementation, I do think that this approach gives us a good base for further development. LGTM 🚢 🆔

PR #347, opened 2016-06-07
Display svg icons correctly

daiyiwhale:
SVG icons being clipped without `display: block property`


mesosphere-web:
Reviewer: @kennyt

PR #349, opened 2016-06-07
Introduce missing service statuses

pierlo-upitup:
This PR introduces support for the `DELAYED` and `WAITING` status. From the [Marathon docs](https://mesosphere.github.io/marathon/docs/marathon-ui.html#application-status-reference):

> Additionally, by inspecting the v2/queue endpoint the following states can also be determined:

**Waiting**

Marathon is waiting for offers from Mesos. True whenever an app has `queueEntry.delay.overdue === true`.

**Delayed**

An app is considered delayed whenever too many tasks of the application failed in a short amount of time. Marathon will pause this deployment and retry later. True if the queue endpoint returns the following JSON condition: `queueEntry.delay.overdue === false`

image

TODO (out of scope)
- [ ] I'd like some guidance on how to add fixtures for the integration tests covering these new statuses in the `SidebarFilter-cy.js` test case
- [ ] verify progress bar styles match the new status values


mesosphere-web:
Reviewer: @daiyiwhale

src/js/constants/ServiceStatusTypes.js
```diff
@@ -1,8 +1,10 @@
...
-  RUNNING: 0,
-  DEPLOYING: 1,
-  SUSPENDED: 2,
-  NA: 4
+  NA: 0,
+  RUNNING: 1,
+  DEPLOYING: 2,
+  SUSPENDED: 3,
+  DELAYED: 4,
+  WAITING: 5
```
orlandohohmeier: I'm not so sure about the order... Isn't "delayed" more important than "waiting" and "running"/"deploying" more important than "suspended"?

pierlo-upitup: So you suggest:

```
NA
SUSPEND
RUNNING
DEPLOYING
WAITING
DELAYED
```
?

src/js/stores/DCOSStore.js
```diff
@@ -98,6 +113,21 @@ class DCOSStore extends EventEmitter {
...
+  onMarathonQueueChange(nextQueue) {
+    let {marathon:{queue}} = this.data;
+
+    nextQueue.forEach((entry) => {
+      if (entry.app == null) {
+        return;
+      }
+
+      let serviceId = entry.app.id;
+      queue.set(serviceId, entry);
```
orlandohohmeier: You don't need the `serviceId` variable here...

pierlo-upitup: no but I thought it made the code easier to read

src/js/structs/Service.js
```diff
@@ -126,6 +126,15 @@ module.exports = class Service extends Item {
   getServiceStatus() {
     let {tasksRunning} = this.getTasksSummary();
     let deployments = this.getDeployments();
+    let queue = this.getQueue();
+
+    if (queue != null && queue.delay && queue.delay.overdue) {
```
orlandohohmeier: I'd like some documentation here, as it's really hard to distinguish the both conditions.

src/js/structs/ServiceTree.js
```diff
@@ -168,21 +169,40 @@ module.exports = class ServiceTree extends Tree {
   }
 
   getServiceStatus() {
-    let {tasksRunning} = this.getTasksSummary();
-    let deployments = this.getDeployments();
+    // Return the service status with the highest index
+    let treeStatus = this.getStatusSummary().reduce((memo, count, statusKey) => {
+      if (count > 0) {
+        Object.keys(ServiceStatus).forEach(statusId => {
```
orlandohohmeier: Do we really need to iterate through the `ServiceStatus` object over and over again? How about creating a temporary map, or adding a different getter?

src/js/constants/ServiceStatusTypes.js
orlandohohmeier: Yep.

src/js/structs/ServiceTree.js
```diff
@@ -168,21 +169,40 @@ module.exports = class ServiceTree extends Tree {
   }
 
   getServiceStatus() {
-    let {tasksRunning} = this.getTasksSummary();
-    let deployments = this.getDeployments();
+    // Return the service status with the highest index
+    let treeStatus = this.getStatusSummary().reduce((memo, count, statusKey) => {
```
orlandohohmeier: Wy are you reducing the `getStatusSummary` result?  And what's the difference?

src/js/stores/DCOSStore.js
```diff
@@ -71,6 +79,13 @@ class DCOSStore extends EventEmitter {
   }
 
   /**
+   * Fetch launch queue from Marathon
+   */
+  fetchQueue() {
```
aldipower: I would prefer a more descriptive name here, so we could get rid of the distracting comment.
Maybe `fetchMarathonLaunchQueue()`.
As always just a suggestion.

pierlo-upitup: how about `fetchServicesQueue` ?

src/js/structs/Service.js
aldipower: No documentation please for such things (that nobody reads).

To make it more descriptive we could refactor the code, which makes it even faster:
```JavaScript
if (queue != null && queue.delay) {
  if (queue.delay.overdue) {
     return ServiceStatus.WAITING;
  }
  return ServiceStatus.DELAYED;
}
```

src/js/stores/DCOSStore.js
aldipower: Sounds good.

src/js/structs/Service.js
pierlo-upitup: This would be such a perfect place for a ternary, 😢

src/js/structs/ServiceTree.js
pierlo-upitup: I realise the code is not very terse. I will think of a more self-explanatory way of doing the same thing.

src/js/stores/DCOSStore.js
pierlo-upitup: LOL i just realised we dont even need this!

src/js/structs/ServiceTree.js
```diff
@@ -168,21 +168,19 @@ module.exports = class ServiceTree extends Tree {
...
-    let deployments = this.getDeployments();
-
-    if (deployments.length > 0) {
-      return ServiceStatus.DEPLOYING;
-    }
-
-    if (tasksRunning > 0) {
-      return ServiceStatus.RUNNING;
-    }
+    return this.reduceItems(function (statusSummary, item) {
```
orlandohohmeier: I'd rename the parameter to something as simple as `serviceStatus`, as we not  dealing with summary anymore.  #nit

pierlo-upitup: Good point. I'll go for `serviceTreeStatus` though, as it represents the final tree status.

src/js/stores/DCOSStore.js
orlandohohmeier: I'm ok with that and would even encourage us to do this more often, but I think we agreed that we not store properties (for what ever reason) if we only access it once. Your call.

orlandohohmeier:
Great job! ✨ 💯

pierlo-upitup:
All comments addressed.

src/js/stores/DCOSStore.js
```diff
@@ -218,7 +240,10 @@ class DCOSStore extends EventEmitter {
         return item;
       }
 
-      let options = {versions: versions.get(item.getId())};
+      let options = {
+        versions: versions.get(item.getId()),
+        queue: queue.get(item.getId())
```
aldipower: Second call of the getId-function, please cache it.

aldipower:
Cool! :) Works and looks good!

PR #352, opened 2016-06-07
Feature - 1066 Service Actions dialogs

aldipower:
This implements the dialogs/confirm-boxes for the service actions (scale, suspend, destroy).
There is no real action behind them, this will follow next.

Destroy:
confirm-destroy

Scale:
confirm-scale

Suspend:
confirm-suspend

### DIFF
--- a/src/js/components/ServiceDetail.js
+++ b/src/js/components/ServiceDetail.js
@@ -15,7 +15,9 @@ import TabsMixin from '../mixins/TabsMixin';
 const METHODS_TO_BIND = [
   'onActionsItemSelection',
   'onAcceptDestroyConfirmDialog',
+  'onAcceptSuspendConfirmDialog',
   'onCancelDestroyConfirmDialog',
+  'onCancelSuspendConfirmDialog',
   'onCloseServiceFormModal'
 ];
 
@@ -34,7 +36,8 @@ class ServiceDetail extends mixin(InternalStorageMixin, TabsMixin) {
     this.state = {
       currentTab: Object.keys(this.tabs_tabs).shift(),
       isServiceFormModalShown: false,
-      isServiceDestroyConfirmShown: false
+      isServiceDestroyConfirmShown: false,
+      isServiceSuspendConfirmShown: false
     };
 
     METHODS_TO_BIND.forEach((method) => {
@@ -50,6 +53,9 @@ class ServiceDetail extends mixin(InternalStorageMixin, TabsMixin) {
       case ServiceActionItem.DESTROY:
         this.setState({isServiceDestroyConfirmShown: true});
         break;
+      case ServiceActionItem.SUSPEND:
+        this.setState({isServiceSuspendConfirmShown: true});
+        break;
     }
   }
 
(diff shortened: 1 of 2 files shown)


mesosphere-web:
Reviewer: @orlandohohmeier

pierlo-upitup:
Shouldn't the heading color always match the severity level? @leemunroe

pierlo-upitup:
retest this please

aldipower:
@pierlo-upitup @leemunroe Just FYI - Unfortunately the color of `text-warning` (ugly yellow) compared to `button-warning` differs a lot in Canvas UI. So we don't have matching colors in the moment.

src/js/components/ServiceDetail.js
```diff
@@ -8,11 +9,17 @@ import ServiceDetailConfigurationTab from './ServiceDetailConfigurationTab';
...
 import ServicesBreadcrumb from './ServicesBreadcrumb';
 import TabsMixin from '../mixins/TabsMixin';
 
 const METHODS_TO_BIND = [
   'onActionsItemSelection',
+  'onAcceptDestroyConfirmDialog',
+  'onAcceptSuspendConfirmDialog',
+  'onCancelDestroyConfirmDialog',
+  'onCancelSuspendConfirmDialog',
+  'onCloseScaleFormModal',
```
pierlo-upitup: shouldnt this be `onCancelScaleFormModal`?

src/js/components/ServiceDetail.js
```diff
@@ -30,7 +37,10 @@ class ServiceDetail extends mixin(InternalStorageMixin, TabsMixin) {
 
     this.state = {
       currentTab: Object.keys(this.tabs_tabs).shift(),
-      isServiceFormModalShown: false
+      isServiceFormModalShown: false,
+      isServiceDestroyConfirmShown: false,
+      isServiceScaleFormModalShown: false,
+      isServiceSuspendConfirmShown: false
```
pierlo-upitup: 😱

pierlo-upitup:
Doesn't this PR only implement `ServiceScaleFormModal` ?

leemunroe:
Very nice.

* I'm a bit concerned about the color of the scale and suspend buttons, I would expect the scale button the be standard modal call to action color (green)
* Can we change the width of the instance input? Likely never going to be more than 3 characters so I would expect the width to better reflect that

aldipower:
@pierlo-upitup No, the other dialogs are `Confirm`-elements from reactjs-components as you can see in the code.

@leemunroe Yeah, I will make the buttons to have the standard color and change the width of the input.

src/js/components/ServiceDetail.js
aldipower: Yes, really bad.

src/js/components/ServiceDetail.js
aldipower: I think the wording for closing a _Modal_ is `close`, we use it like that also at other places.
On the other hand the wording in the _Confirm_ is `cancel`.
So if I change this, the things will be mixed up at another place.
I would like to stick with this.

src/js/components/ServiceDetail.js
orlandohohmeier: There is always only one modal open, right? How about one variable holding a modal id or nothing to toggle all these modal states?

aldipower: Good idea, at least this looks much better then. Underlaying problem will remain the same..
Will do it.

aldipower:
@orlandohohmeier I've refactored the show states. Feels much better now.

@leemunroe I've changed the button colors to the `primary`-color as used application wide.
Reducing the width of the input field looks silly though. If you like you could show me what you have exactly in mind?

confirm-scale

Notice the funny text.. ;)
confirm-suspend

orlandohohmeier:
L:eyes:ks good to me! Let's :shipit:

Poltergeist:
Merge

kennyt:
did this PR get merged in with failing integration tests?

Poltergeist:
@kennyt This was my bad, there was one test failing in the integration tests. In the last couple of days the integration tests were failing a lot, not because they were actually failing but because they were flaky that was why I probably didn't paid enough attention to them. I am sorry for this and will pay more attention next time.

PR #354, opened 2016-06-07
Change proptype for tabform definition into object

kennyt:


mesosphere-web:
Reviewer: @pierlo-upitup

PR #355, opened 2016-06-07
Adjust PageHeader component styling

jfurrow:
It looks like we forgot one of the icon classes to give it rounded corners. I also added `align-items: center` to help align the icon with the header.

Before:

After:

### DIFF
--- a/src/js/components/PageHeader.js
+++ b/src/js/components/PageHeader.js
@@ -11,7 +11,7 @@ class PageHeader extends React.Component {
     }
 
     let iconClasses = classNames(
-      'icon icon-large icon-image-container',
+      'icon icon-large icon-app-container icon-image-container',
       iconClassName
     );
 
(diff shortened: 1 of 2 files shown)


mesosphere-web:
Reviewer: @aldipower

src/js/components/PageHeader.js
```diff
@@ -11,7 +11,7 @@ class PageHeader extends React.Component {
     }
 
     let iconClasses = classNames(
-      'icon icon-large icon-image-container',
+      'icon icon-large icon-app-container icon-image-container',
```
daiyiwhale: I think `icon-app-container` wasn't in there before so this component could be more generic, letting the pages using the pageheaders chose to add either `icon-app-container` or `icon-user-container` etc depending on what kind of icon it needs. Not sure if that makes more sense of if non-`icon-app-container` pages should use `classnames.dedupe` to remove `icon-app-container` and add `icon-{thing}-container`?

jfurrow:
Closing this, PageHeader refactor coming up shortly!

PR #356, opened 2016-06-07
Upgrade jest and babel dependencies.

rcorral:
Fixes really long error messages when a test would fail.


mesosphere-web:
Reviewer: @jfurrow

PR #363, opened 2016-06-08
Adds defaults to composite sate.

rcorral:
So that if the Marathon response comes before Mesos there's no error.


mesosphere-web:
Reviewer: @Poltergeist

rcorral:
retest this please

PR #364, opened 2016-06-08
Refactor PageHeader component

jfurrow:
The `PageHeader` component was a little difficult to work with previously, due to its inheritance of multiple unrelated style components. This PR introduces styles specifically for the `PageHeader` that more closely match the design. Its behavior at smaller viewport widths is more consistent as well.

I also decided to change the classname of our page's tabbed navigation, which was previously called `page-header`. I think this will help clarify the roles of the elements.

### DIFF
--- a/src/js/components/Page.js
+++ b/src/js/components/Page.js
@@ -59,7 +59,7 @@ var Page = React.createClass({
     }
 
     return (
-      <div className="page-header-navigation">
+      <div className="page-navigation-list">
         <div className="container container-fluid container-pod container-pod-short flush-bottom">
           {navigation}
         </div>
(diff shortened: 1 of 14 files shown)


mesosphere-web:
Reviewer: @pierlo-upitup

pierlo-upitup:
LGTM although it's quite a lot of new CSS!

jfurrow:
@pierlo-upitup That's true, but I do think this is better — please let me know if you're opposed!

The classes that we applied to this previously got the `PageHeader` to match the designs _most_ of the way, but not fully, and there were a ton of classes that were added to get some basic properties (notably the 5+ `container-pod` variations just to get the grey divider). I also think the `media-object` classes are being abused in general and I was really delighted to remove them, but I'm happy to rethink this if you're opposed to this solution!

MatApple:
🚢

PR #371, opened 2016-06-09
Allow the UI to load even if requests to endpoints are failing.

rcorral:
This will allow users to access the UI in case that they made a mistake while configuring different parts of the UI.
This will also allow access to the Components Health to debug why components in the cluster could not be working.


mesosphere-web:
Reviewer: @philipnrmn

src/js/index.js
```diff
@@ -44,37 +44,58 @@ RequestUtil.json = function (options = {}) {
...
-    setTimeout(function () {
-      let routes = RouterUtil.buildRoutes(appRoutes.getRoutes());
-      let router = Router.run(routes, function (Handler, state) {
-        Config.setOverrides(state.query);
-        ReactDOM.render(
-          (<Provider store={PluginSDK.Store}>
-            <Handler state={state} />
-          </Provider>),
-          domElement);
+(function () {
```
MatApple: Do we need this closure?

rcorral: It just keeps the code that is related to each other together, if you feel strongly about it I can remove it. I just figured it's easy to come back to this in the future and see what is related in this file, since it's just one long thing of stuff.

MatApple: Thats Ok, I don't feel strongly.

src/js/components/RequestErrorMsg.js
```diff
@@ -40,11 +48,13 @@ class RequestErrorMsg extends React.Component {
 }
 
 RequestErrorMsg.defaultProps = {
+  columnClasses: {},
   header: 'Cannot Connect With The Server',
   message: getDefaultMessage()
 };
 
 RequestErrorMsg.propTypes = {
+  columnClasses: React.PropTypes.object,
```
MatApple: A better declaration here would be:
```
columnClasses: React.PropTypes.oneOfType([
    React.PropTypes.array,
    React.PropTypes.object,
    React.PropTypes.string
  ]),
```

src/js/index.js
```diff
@@ -90,7 +91,20 @@ RequestUtil.json = function (options = {}) {
   }
 
   function onConfigurationError() {
-    // Render configuration error
+    // Try to find appropriate DOM element or fallback
+    let element = document.querySelector('#canvas div') || domElement;
+    let columnClasses = {
```
MatApple: This could be a single String or an array of Strings.

src/js/components/RequestErrorMsg.js
rcorral: Done.

PR #376, opened 2016-06-09
1066 - Enable service scale action with error handling

aldipower:
This enables the service scale dialog and triggers the actual scale action. Engage!

### DIFF
--- a/src/js/components/modals/ServiceScaleFormModal.js
+++ b/src/js/components/modals/ServiceScaleFormModal.js
@@ -5,6 +5,7 @@ import React from 'react';
 import {StoreMixin} from 'mesosphere-shared-reactjs';
 
 import FormModal from '../FormModal';
+import MarathonStore from '../../stores/MarathonStore';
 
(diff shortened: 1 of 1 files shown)


mesosphere-web:
Reviewer: @jfurrow

src/js/components/modals/ServiceScaleFormModal.js
```diff
@@ -58,7 +97,7 @@ class ServiceGroupFormModal extends mixin(StoreMixin) {
         min: 0,
         name: 'instances',
         placeholder: instancesCount,
-        value: instancesCount,
+        value: instancesCount.toString(),
```
orlandohohmeier: Does this really needs to be a string?

pierlo-upitup: As it ends up inside an `input` field it will eventually become one, so I guess yes. But ideally the Form should take care of that for us.

orlandohohmeier:
LGTM 💯 ✨

PR #378, opened 2016-06-09
[6] Deployments Table - improve data interpolation

philipnrmn:
Previously, we assumed that groups data was available from the Marathon store when we populated the DCOS Store with deployments data. This was not only bad in theory but also in practice, as the deployments endpoint is significantly quicker than the groups endpoint and the race condition thus created was almost always tripped.

This PR:
- [x] ensures that the deployments are not processed until the groups data has been received
- [x] invokes the deployments processing code immediately after the groups data has been received
- [x] ensures that if the deployments service data is stale (ie, the deployment contains references to deleted services), those services are trimmed

Outside the scope of this feature we should either update the name of the `dataProcessed` attribute of `DCOSStore` to reflect that it only relates to groups data, or extend it so that it is only true when all endpoints have returned some data, or (my preference) remove it entirely, since IMO components should not have to perform any sort of introspection against stores.

### DIFF
--- a/src/js/stores/DCOSStore.js
+++ b/src/js/stores/DCOSStore.js
@@ -139,6 +139,10 @@ class DCOSStore extends EventEmitter {
 
     this.data.marathon.serviceTree = serviceTree;
     this.data.dataProcessed = true;
+
+    // Populate deployments with services data immediately
+    this.onMarathonDeploymentsChange();
+
     this.emit(DCOS_CHANGE);
   }
 
--- a/src/js/stores/DCOSStore.js
+++ b/src/js/stores/DCOSStore.js
@@ -107,6 +107,9 @@ class DCOSStore extends EventEmitter {
   }
 
   onMarathonDeploymentsChange() {
+    if (!this.data.dataProcessed) {
+      return;
+    }
     let deploymentsList = MarathonStore.get('deployments');
     let serviceTree = MarathonStore.get('groups');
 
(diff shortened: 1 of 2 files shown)


mesosphere-web:
Reviewer: @daiyiwhale

philipnrmn:
retest this please

philipnrmn:
retest this please

philipnrmn:
retest this please

philipnrmn:
Third time's the charm.

src/js/stores/DCOSStore.js
```diff
@@ -120,7 +123,13 @@ class DCOSStore extends EventEmitter {
     this.data.marathon.deploymentsList = deploymentsList
       .mapItems(function (deployment) {
         let ids = deployment.getAffectedServiceIds();
-        let services = ids.map(serviceTree.findItemById.bind(serviceTree));
+        let services = ids.reduce(function (memo, id) {
+          let service = serviceTree.findItemById(id);
```
orlandohohmeier: How about moving the data merging to the getter - similar the services? With that change you wouldn't need the `dataProcessed` conditions as you could return the deployment list whenever the needed data is present. You're maybe even able to provide a subset until all needed data is present. What do you think?

Poltergeist:
LGTM Lets 🚢  `:this:` 

CI passes: 
---
image
---

Tests pass:
---
image
---

orlandohohmeier:

philipnrmn:
¯\_(ツ)_/¯

src/js/stores/DCOSStore.js
philipnrmn: Hey @orlandohohmeier, sorry I didn't see this until just now. I agree that's a much nicer approach, let's make that change separately.

PR #382, opened 2016-06-09
1066 - Delete service action

aldipower:
Getting an error with the StoreMixin that I really don't understand, after the services got deleted with the new 'Destroy'-action in the dropdown.
It's seems the according Component is already unmounted, so the Mixin tries to work on a null-object?

storemixin

@jfurrow Would be great if you could take a quick look? But don't waste too much time. :)

----
This implements the actual delete service action request. Engage!


mesosphere-web:
Reviewer: @Poltergeist

jfurrow:
@aldipower I just pushed a commit (https://github.com/dcos/dcos-ui/pull/382/commits/a3b3b2ff1f521d51137b946efc24f4ca9bf5d4fb) that fixes the issue. It was due to `ServiceScaleFormModal` using the `StoreMixin` when it didn't need to (specifically because `store_listeners` was not defined on the `ServiceScaleFormModal` instance).

Shoutout to @rcorral who was a massive help in debugging this!

aldipower:
Oh ok! Now I get it.
Thank you very much for your help. Very kind of you.

Hah, it's that the store_listeners are set and used in the base branch `feature/1066-service-actions`.
So trying to keep the PRs more atomic screwed me up a little bit here. I either took the wrong approach of splitting the things or I wasn't careful enough.

Thanx again for your help and time. @jfurrow @rcorral

Poltergeist:
```
src/js/components/modals/ServiceScaleFormModal.js
  42:22  error  'instances' is defined but never used  no-unused-vars
```
Is this fixed in the target base branch? `if (true)` I'll add my label.

aldipower:
@Poltergeist More then fixed! It's actually used there!!!

PR #381, opened 2016-06-09
[1 of 2 Debug Tab] Add service debug tab

mlunoe:
Desktop:

Mobile:

Task with no statistics (the scenario will happen for each section, when information is not available):

Note: The table in the images will be added in an upcoming PR.

PR #384, opened 2016-06-09
[2 of 2 Debug Tab] Add service debug tab stats table

mlunoe:
---
~~⚠️ Dependent on this PR: https://github.com/dcos/dcos-ui/pull/381
ℹ️ Reviewer look here for the diff between the two branches: https://github.com/dcos/dcos-ui/compare/mlunoe/add-service-debug-tab...mlunoe/add-service-debug-tab-stats-table?expand=1~~

---

Task with no statistics (the scenario will happen for each section, when information is not available):

### DIFF
--- a/src/js/components/ServiceDetailDebugTab.js
+++ b/src/js/components/ServiceDetailDebugTab.js
@@ -3,6 +3,7 @@ import React from 'react';
 import DateUtil from '../utils/DateUtil';
 import DescriptionList from './DescriptionList';
 import Service from '../structs/Service';
+import TaskStatsTable from './TaskStatsTable';
 
 class ServiceDetailDebugTab extends React.Component {
   getValueText(value) {
@@ -72,6 +73,18 @@ class ServiceDetailDebugTab extends React.Component {
     return <DescriptionList hash={LastVersionChangeValueMapping} />;
   }
 
+  getTaskStats() {
+    let {taskStats} = this.props.service;
+
+    if (taskStats == null || Object.keys(taskStats).length === 0) {
+      return (
+        <p>This app does not have task statistics</p>
+      );
+    }
+
+    return <TaskStatsTable taskStats={taskStats} />
+  }
+
   render() {
     return (
       <div>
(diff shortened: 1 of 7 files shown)


mesosphere-web:
Reviewer: @orlandohohmeier

src/js/components/ServiceDetailDebugTab.js
```diff
@@ -0,0 +1,114 @@
...
+          {lastConfigChangeAt} ({new moment(lastConfigChangeAt).fromNow()})
+        </span>
+      )
+    };
+
+    return (<DescriptionList hash={LastVersionChangeValueMapping} />);
+  }
+
+  getTaskStats() {
+    let {taskStats} = this.props.service;
```
orlandohohmeier: Why aren't you using the service getter? Using [`Service.prototype. getTaskStats`](https://github.com/dcos/dcos-ui/blob/master/src/js/structs/Service.js#L168) would give you `TaskStats` struct which would save you a lot of unnecessary fiddling with the raw data.

src/js/components/TaskStatsTable.js
```diff
@@ -0,0 +1,161 @@
...
+const keyCaptionMap = {
+  startedAfterLastScaling: 'Started After Last Scaling',
+  withLatestConfig: 'With Latest Config',
+  withOutdatedConfig: 'With Outdated Config',
+  totalSummary: 'Total Summary'
+};
+
+class TaskStatsTable extends React.Component {
+  getTaskStats() {
+    let {taskStats} = this.props;
```
orlandohohmeier: I'd like to know why you decided to work with the raw data instead of using the respective structs, as proposed in https://github.com/dcos/dcos-ui/pull/384/files#r66639580

orlandohohmeier:
LGTM 💯 ✨

MatApple:
@mlunoe what is the difference between `Average Seconds` and `Median Seconds` in this table?
1) shouldn't we name it `mean` and `median`?
2) it looks like we have the exact same value which makes me wonder if we're using the same algo for both?

orlandohohmeier:
@MatApple These values are calculated by Marathon - they describe the task life time. You will see a difference if your task fail.   
```
          "lifeTime" : { 
            // Measured from `"startedAt"` (timestamp of the Mesos TASK_RUNNING status update) of each running task 
            // until now.
            "averageSeconds" : 20.0,
            "medianSeconds" : 10.0
          }
```

mlunoe:
@MatApple It is not the same. When there has only been one task they will be the same however. Having multiple different tasks you will start to see differences. I'm using the terms that have been decided on from Marathon, and have been ported by design

PR #391, opened 2016-06-10
Add labels section to the service form modal

Poltergeist:
This adds the functionality to add Labels in the create modal.

image
image

### DIFF
--- a/src/js/schemas/service-schema/Labels.js
+++ b/src/js/schemas/service-schema/Labels.js
@@ -1,4 +1,4 @@
-var Labels = {
+const Labels = {
   type: 'object',
   title: 'Labels',
   properties: {
@@ -9,14 +9,12 @@ var Labels = {
       getter: function (service) {
         let labels = service.getLabels() || {};
 
-        return Object.keys(labels).reduce(function (memo, key) {
-          memo.push({
+        return Object.keys(labels).map(function (key) {
+          return {
             key,
             value: labels[key]
-          });
-
-          return memo;
-        }, []);
+          };
+        });
       },
       itemShape: {
         properties: {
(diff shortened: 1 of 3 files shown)


mesosphere-web:
Reviewer: @philipnrmn

aldipower:
Daumen hoch! :+1:

src/js/schemas/service-schema/Labels.js
```diff
@@ -0,0 +1,37 @@
+var Labels = {
```
kennyt: `let`

src/js/schemas/service-schema/Labels.js
```diff
@@ -0,0 +1,37 @@
...
+  title: 'Labels',
+  properties: {
+    labels: {
+      type: 'array',
+      duplicable: true,
+      addLabel: 'Add Label',
+      getter: function (service) {
+        let labels = service.getLabels() || {};
+
+        return Object.keys(labels).reduce(function (memo, key) {
```
kennyt: i feel like this could be better implemented using `map`. what do you think?

kennyt:
after my two super nits, :ship:

src/js/schemas/service-schema/Labels.js
Poltergeist: You are right I was missing that.

src/js/schemas/service-schema/Labels.js
Poltergeist: changed it to const

src/js/schemas/service-schema/Labels.js
```diff
@@ -0,0 +1,35 @@
...
+      addLabel: 'Add Label',
+      getter: function (service) {
+        let labels = service.getLabels() || {};
+
+        return Object.keys(labels).map(function (key) {
+          return {
+            key,
+            value: labels[key]
+          };
+        }, []);
```
kennyt: you don't need the 2nd argument `[]` :D

PR #394, opened 2016-06-10
Use ms in DateUtil#msToRelativeTime instead of seconds

jfurrow:
Previously the method `msToRelativeTime` expected seconds, now it will take milliseconds as its name suggests


mesosphere-web:
Reviewer: @orlandohohmeier

PR #400, opened 2016-06-13
Fix labels error if no label is provided

Poltergeist:
💯 💯 💯 💯

### DIFF
--- a/src/js/utils/ServiceUtil.js
+++ b/src/js/utils/ServiceUtil.js
@@ -84,10 +84,12 @@ const ServiceUtil = {
       }
 
       if (labels != null) {
-        definition.labels = labels.labels.reduce(function (memo, item) {
-          memo[item.key] = item.value;
-          return memo;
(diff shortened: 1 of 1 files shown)


mesosphere-web:
Reviewer: @mlunoe

src/js/utils/ServiceUtil.js
```diff
@@ -84,10 +84,12 @@ const ServiceUtil = {
       }
 
       if (labels != null) {
-        definition.labels = labels.labels.reduce(function (memo, item) {
-          memo[item.key] = item.value;
-          return memo;
-        }, {});
+        if (labels.labels != null) {
```
philipnrmn: Could combine these into one check...

Poltergeist: You are right, I thought about keeping it this way in case something gets added later to that section, which is probably not the case. But since you mentioned it I changed it to be only one condition.

PR #388, opened 2016-06-10
[7] Deployments Table - show rollback link

philipnrmn:
This PR:

- [x] adds a deployment link to the deployments table
- [x] adds an action to the marathon store to delete the deployment
- [x] introduces a confirmation modal
- [x] removes the deployment from the store if the delete request is succesful
- [x] handles errors on the delete request

This PR depends on ~~#372~~

PR #401, opened 2016-06-13
[8] Deployments Table - styling

philipnrmn:
This PR brings the styling of the deployments table into line with the designs. 

Before and after:

- [x] Column widths are correctly set
- [x] Long group and service names do not overflow the line
- [x] Location, start time and action columns are removed on narrow screens

- [x] Would be great to have @leemunroe's opinion on this. 

NB: this PR depends on #388

### DIFF
--- a/src/js/pages/services/DeploymentsTab.js
+++ b/src/js/pages/services/DeploymentsTab.js
@@ -81,10 +81,10 @@ class DeploymentsTab extends mixin(StoreMixin) {
 
       return (
         <dd key={index}>
-          <span className="icon icon-small icon-image-container icon-app-container deployment-service-icon">
-            <img src={image} />
-          </span>
           <Link to="services-detail" params={{id}} className="deployment-service-name">
+            <span className="icon icon-small icon-image-container icon-app-container deployment-service-icon">
+              <img src={image} />
+            </span>
             {StringUtil.capitalize(service.getName())}
           </Link>
         </dd>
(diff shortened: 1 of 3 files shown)


mesosphere-web:
Reviewer: @Poltergeist

philipnrmn:
Made some further tweaks after a review session with @leemunroe

philipnrmn:
retest this please

leemunroe:
@philipnrmn Is it just me or is the started column vertically aligned to the top and the status isn't? Would expect these to be the same (probably both top) when things like wrapping occurs on other cells.

image

rcorral:
Looks good, just needs a rebase!

src/js/pages/services/DeploymentsTab.js
```diff
@@ -124,6 +141,30 @@ class DeploymentsTab extends mixin(StoreMixin) {
     );
   }
 
+  renderAction(prop, deployment) {
```
Poltergeist: As mentioned in #411 this should be right before `renderEdit` 🔤  feel free to change it in either PR

Poltergeist:
Looks good to me except one last nit™️ 

image

❤️  it❗🚢  it❗

philipnrmn:
@leemunroe the deployment ID should never wrap, because it will throw all the row alignments out. Would you please retest making sure you've pulled the latest changes?

Separately, I also vertically aligned everything top.

philipnrmn:
@Poltergeist re alphabetical order, I'm not sure how to approach this. At the moment ordering within this component follows the convention of column order, and methods like `getModalContent` directly below `renderModal` to keep them in context.

I'd like either to stay with this convention or convert the entire file to alphabetical order; I don't mind which but if it's the latter I'd prefer to address it in a separate PR rather than piggybacking off this styling one.

leemunroe:
👍 Sounds good. Was going off the gif screenshot you left above.

PR #402, opened 2016-06-13
Introduce job form component

orlandohohmeier:
Introduce a job form component – it's mainly needed to work around an issue (DCOS-7669) with the schema form, but might also serve other needs in the future.

/cc @kennyt @Poltergeist

### DIFF
--- a/src/js/components/JobForm.js
+++ b/src/js/components/JobForm.js
@@ -1,0 +1,52 @@
+import React from 'react';
+
+import SchemaForm from './SchemaForm';
+import SchemaFormUtil from '../utils/SchemaFormUtil';
+
+const METHODS_TO_BIND = [
+  'handleFormChange',
+  'validateForm'
+];
+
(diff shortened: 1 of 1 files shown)


mesosphere-web:
Reviewer: @philipnrmn

orlandohohmeier:
Let me add that I'm very unhappy with this solution, and would love if we could fix `SchemaForm` instead, but I don't have the time nor do I know enough about all the form components to fix it myself.

aldipower:
@orlandohohmeier @kennyt 
I am a little bit unhappy with this, because it disguise an underlaying problem, which makes it even more difficult to detect then.
But I also see the need for this right now.
I hope the underlaying architecture will be corrected in the future, so we can get rid of this.
Otherwise we will run in much more trouble later.

aldipower:
@orlandohohmeier 
I see! Yes! :)

kennyt:
going to be fixing it this week

PR #411, opened 2016-06-13
[9] Deployments Table - rollback and remove for 'starting' deployments

philipnrmn:
This functionality was added after talking with @leemunroe 

This PR introduces a distinction between a starting and a non-starting deployment. A deployment is considered to be starting if it deploys a service for the first time (or in more exact terms, if one of its actions is of type `StartApplication`)

When a deployment starts an application, rolling it back also deletes the linked service. We felt that this was not sufficiently well telegraphed to the user, and that both the rollback link and the accompanying modal should reflect this. 

This PR therefore

- [x] Adds an `isStarting` method to the `Deployment` struct
- [x] Shows a 'Rollback and Remove' link on starting deployments
- [x] Shows updated rollback modal text on starting deployments

NB this PR depends on #401

### DIFF
--- a/src/js/pages/services/DeploymentsTab.js
+++ b/src/js/pages/services/DeploymentsTab.js
@@ -143,10 +143,14 @@ class DeploymentsTab extends mixin(StoreMixin) {
 
   renderAction(prop, deployment) {
     if (deployment != null) {
+      let linkText = 'Rollback';
+      if (deployment.isStarting()) {
+        linkText = linkText + ' & Remove';
+      }
       return (
         <a className="deployment-rollback button button-link button-danger table-display-on-row-hover"
             onClick={this.handleRollbackClick.bind(null, deployment)}>
-          Rollback
+          {linkText}
         </a>
       );
     }
--- a/src/js/pages/services/DeploymentsTab.js
+++ b/src/js/pages/services/DeploymentsTab.js
@@ -211,7 +211,7 @@ class DeploymentsTab extends mixin(StoreMixin) {
         <col className="hidden-mini" />
         <col className="hidden-mini" style={{width: '120px'}} />
         <col style={{width: '240px'}} />
-        <col className="hidden-mini" style={{width: '120px'}} />
+        <col className="hidden-mini" style={{width: '160px'}} />
       </colgroup>
     );
   }
@@ -251,17 +251,6 @@ class DeploymentsTab extends mixin(StoreMixin) {
   renderRollbackModal() {
     let {deploymentToRollback} = this.state;
     if (deploymentToRollback != null) {
-      let serviceNames = deploymentToRollback.getAffectedServices()
-        .map(function (service) {
-          return StringUtil.capitalize(service.getName());
-        });
-      let listOfServiceNames = StringUtil.humanizeArray(serviceNames);
-      let serviceCount = serviceNames.length;
-
-      let service = StringUtil.pluralize('service', serviceCount);
-      let its = (serviceCount === 1) ? 'its' : 'their';
-      let version = StringUtil.pluralize('version', serviceCount);
-
       return (
         <Confirm
           closeByBackdropClick={true}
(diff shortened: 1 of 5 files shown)


mesosphere-web:
Reviewer: @Poltergeist

leemunroe:
I think this definitely better communicates what will happen.

Curious what @aquamatthias thinks as this gets rid of the "Stop" action as he previously requested.

@philipnrmn Note: In the gif you provided the modal jumps ever so slightly when you hover over the button. Does that happen IRL? If so we should fix it.

src/js/stores/MarathonStore.js
```diff
@@ -337,6 +340,18 @@ class MarathonStore extends GetSetBaseStore {
     this.emit(MARATHON_DEPLOYMENTS_ERROR);
   }
 
+  processMarathonDeploymentRollback(data) {
+    let id = data.originalDeploymentID;
+    if (id != null) {
+      let deployments = this.get('deployments')
+        .filterItems(function (deployment) {
+          return deployment.id !== id;
```
kennyt: can we do `getId()`?

aldipower:
Technically works as expected and looks nice! @leemunroe I don't see any flickering of the dialog IRL, must be the .gif.

src/js/pages/services/DeploymentsTab.js
```diff
@@ -180,11 +245,58 @@ class DeploymentsTab extends mixin(StoreMixin) {
...
+          <div className="container-pod container-pod-short text-align-center">
+            <h3 className="flush-top">You're About To Rollback The Deployment</h3>
+            <p>{this.getRollbackModalText(deploymentToRollback)}</p>
+          </div>
+        </Confirm>
+      );
+    }
+  }
+
+  getRollbackModalText(deploymentToRollback) {
```
Poltergeist: This should be after getColGroup following 🔤 .

src/js/pages/services/DeploymentsTab.js
```diff
@@ -124,6 +141,34 @@ class DeploymentsTab extends mixin(StoreMixin) {
     );
   }
 
+  renderAction(prop, deployment) {
```
Poltergeist: This should be before `renderEmpty` following the :abc:

Poltergeist:
First I thought this is a lot of code for a dialog and then I realised it is:
image

👍  good job.

It looks good except some small nits on my end.

aldipower:
Besides @kennyt and @Poltergeist nits this code looks good! :)

Although I have one note.
I cannot detect any error handling of the request fails?

src/js/pages/services/DeploymentsTab.js
```diff
@@ -271,17 +264,35 @@ class DeploymentsTab extends mixin(StoreMixin) {
...
+
+    if (deploymentToRollback.isStarting()) {
+      return `This will stop the current deployment of ${listOfServiceNames}
+              and start a new deployment to remove the affected ${service}.`
+    }
+
+    return `This will stop the current deployment of ${listOfServiceNames} and
+            start a new deployment to revert the affected ${service} to ${its}
+            previous ${version}.`
+  }
```
pierlo-upitup: :rage4:

philipnrmn:
Addressed @Poltergeist's alphabetisation concern in #401 https://github.com/dcos/dcos-ui/pull/401#issuecomment-225873634

pierlo-upitup:
Nice one! As we see green let's 🚢

philipnrmn:
@leemunroe what's do you think re @aldipower's query here https://github.com/dcos/dcos-ui/pull/411#issuecomment-225839685 - what should the deployments tab do when a rollback attempt results in a server error?

philipnrmn:
Re lots of code, this had two upstream unmerged PRs - with those merged and this branch rebased, we're down to a more manageable `+107 −19`

philipnrmn:
Re @aldipower's query here https://github.com/dcos/dcos-ui/pull/411#issuecomment-225839685:
#438

PR #372, opened 2016-06-09
[5] Deployments Table - Introduce service rows

philipnrmn:
This PR introduces services to the deployments tables and adds styles for them. The `tree-list` will hopefully eventually be contributed upstream to Canvas-UI, but for the moment has been added as a dcos-ui specific component. 

__NB: I have had to create a new feature branch after the last one got badly out of synch. See #370 for more detail.__

PR #249, opened 2016-05-24
[4/8] Deployments Table - Use DCOSStore instead of MarathonStore

philipnrmn:
Note that this work is based off a feature branch (`deployments-table`).

This PR relies on ~~#241~~ and ~~#248~~ (it may be merged before #248 without issue)

The intention of the PR is to keep the behaviour of the deployments table exactly the same, but making use of the `DCOSStore` instead of the `MarathonStore`. 

Testing:
```bash
git checkout feature/deployment-dcos-store
curl "https://patch-diff.githubusercontent.com/raw/dcos/dcos-ui/pull/248.patch" | git apply
```

Note: the above is no longer necessary since the merge of #248

PR #247, opened 2016-05-24
[3/8] Deployments Table - Add badge to deployments tab

philipnrmn:
Note that this work is based off a feature branch (`deployments-table`).

This PR relies on ~~#241~~ and ~~#215~~ (but may be merged before ~~#215~~ without issue). 

This PR adds a notification badge to the Deployment tab. 

moving pictures

Testing:
```bash
git checkout feature/deployment-tab-badge
curl "https://patch-diff.githubusercontent.com/raw/dcos/dcos-ui/pull/215.patch" | git apply
```

(Note - no longer necessary following rebase)

PR #412, opened 2016-06-13
[All] Deployments Table

philipnrmn:
This PR merges the deployments table feature branch into master. Each component PR has already been reviewed. Please do not merge until the following list is complete:

Component PRs:
- [x] #241 [1/8] Deployments Table - Deployment struct updates
- [x] #255 [2/8] Deployments Table - Don't use ServicesList to list services
- [x] #247 [3/8] Deployments Table - Add badge to deployments tab
- [x] #249 [4/8] Deployments Table - Use DCOSStore instead of MarathonStore
- [x] #372 [5] Deployments Table - Introduce service rows
- [x] #378 [6] Deployments Table - improve data interpolation
- [x] #388 [7] Deployments Table - show rollback link
- [x] #401 [8] Deployments Table - styling
- [x] #411 [9] Deployments Table - rollback and remove for 'starting' deployments
- [x] #425 [10] Deployments Table - Replace breadcrumbs with nested services links
- [x] #430 [11] Use store mixin forced updates instead of react lifecycle
- [ ] #??? [12] Handle error response on rollback deployments


mesosphere-web:
Reviewer: @aldipower

src/js/pages/services/DeploymentsTab.js
```diff
@@ -1,73 +1,158 @@
...
+      {name: 'dcos', events: ['change'], suppressUpdate: true}
+    ];
+
+    METHODS_TO_BIND.forEach((method) => {
+      this[method] = this[method].bind(this);
+    }, this);
+  }
+
+  onDcosStoreChange() {
+    this.setState({deployments: DCOSStore.deploymentsList});
```
rcorral: We typically avoid this patter of storing data that we already have in components. You could instead remove the `supressUpdate: true` and just get the `deploymentsList` where you need it in this component.
This avoids storing the data in the store and in `n` components that are being rendered.
The component will automatically rerender if you remove the `suppressUpdate`

src/js/stores/DCOSStore.js
```diff
@@ -97,6 +106,37 @@ class DCOSStore extends EventEmitter {
...
+          if (service != null) {
+            memo.push(service);
+          }
+          return memo;
+        }, []);
+
+        return Object.assign({affectedServices: services}, deployment);
+      });
+
+    this.emit(DCOS_CHANGE);
```
rcorral: Would it be better to have a different event name? Otherwise anything that listens for `DCOS_CHANGE`.
We typically have a new event name per endpoint and it looks like the deployments data comes from a different endpoint than the list of running apps.

src/js/pages/services/DeploymentsTab.js
```diff
@@ -1,73 +1,158 @@
...
+  }
+
+  renderAffectedServices(prop, deployment) {
+    return (
+      <dl className="deployment-services-list flush-top flush-bottom tree-list">
+        <dt className="deployment-id text-uppercase">{deployment.id}</dt>
```
rcorral: Can you use the `#getId` method on the Deployment struct?

src/js/pages/services/DeploymentsTab.js
```diff
@@ -1,73 +1,158 @@
...
+      );
+    });
+  }
+
+  renderLocation(prop, deployment) {
+    const services = deployment.getAffectedServices();
+    const items = services.map(function (service, index) {
+      return (
+        <li key={index}>
+          <ServicesBreadcrumb
```
rcorral: We should be using the new `Breadcrumbs` component, all of the new components will be using it and we will be removing the old `ServicesBreadcrumb`. See this #416

src/js/stores/DCOSStore.js
pierlo-upitup: We are doing this as part of the `DCOS` omnistore and this is consistent with what's in the rest of the file:

https://github.com/dcos/dcos-ui/blob/master/src/js/stores/DCOSStore.js#L108
https://github.com/dcos/dcos-ui/blob/master/src/js/stores/DCOSStore.js#L122
https://github.com/dcos/dcos-ui/blob/master/src/js/stores/DCOSStore.js#L163
etc...

src/js/pages/services/DeploymentsTab.js
pierlo-upitup: Let's do that in a separate PR.

philipnrmn: #425

philipnrmn:
NB tests will fail until #425 is merged.

src/js/stores/DCOSStore.js
philipnrmn: Yeah I was aiming for consistency with the Marathon Groups data which is used in the Services tab. I would like to see more fine-grained events eventually though.

src/js/pages/services/DeploymentsTab.js
philipnrmn: Addressed directly on this branch in fbc6a0e

src/js/pages/services/DeploymentsTab.js
philipnrmn: Happy to respect the convention if it's genuinely preferred, but I'd like to understand what the benefit is of breaking with React idiom. Forcing updates via the store mixin makes it hard for me to understand the update/render cycle and circumvents `shouldComponentUpdate`.

rcorral: Yes, please let's keep things consistent. At the moment it is the preferred way of doing things.
A few thoughts:
* Storing the same datum in multiple components isn't a good practice
* If we truly don't want to rerender when an event gets fired through Storemixin then let's use `suppressUpdate` but it looks like we do want this behavior right now.
* ForceUpdate isn't pretty, I agree. Let's consider other solutions for the future.

src/js/stores/DCOSStore.js
rcorral: Fair, it just sucks that so much stuff renders all the time :(

src/js/pages/services/DeploymentsTab.js
philipnrmn: #430

pierlo-upitup:
Seems to me all comments have been addressed minus `Handle error response on rollback deployments`. Can we merge this and address that outside of this PR please?

aldipower:
Very nice work! :ok_hand:

aldipower:
retest this please

PR #415, opened 2016-06-13
Fix missing table borders in minified CSS

MatApple:
Fixes missing table borders in minified CSS. Bug in cssnano http://cssnano.co/optimisations/#mergelonghand


mesosphere-web:
Reviewer: @kennyt

PR #416, opened 2016-06-13
Using new breadcrumbs in services page.

rcorral:


mesosphere-web:
Reviewer: @pierlo-upitup

rcorral:
retest this please

daiyiwhale:
Code looks good! I'm not sure about this breadcrumb trail naming though, it's a bit confusing to have the service name twice when looking at the detail page. But I also don't know what would be better (naming the second breadcrumb label `tasks`? `details`? hm)

daiyiwhale:
Oh wait that's just how the tasks happened to be named, ignore me!

pierlo-upitup:
I think this PR removed the `Services` entry from the default (ie unfiltered) page. 

missing crumb

@rcorral what do you think?

rcorral:
I can definitely add it back, I did this on purpose because I felt it was redundant for it to just say `Service`
If you think we should have it, I'll put up a PR. Let me know @pierlo-upitup

pierlo-upitup:
Absolutely. As you can see in the gif the layout feels broken with the Services heading. Also, this is what the mockups indicate.

rcorral:
PR #445

PR #410, opened 2016-06-13
Change service modal styling

Poltergeist:
image

This changes the header of the service form modal to look a bit more welcoming and also now the header includes the JSON toggle button.

Fix for the react warning is PR'ed at mesosphere/reactjs-components#286

Contains #409 [~~unmerged~~merged]

PR #419, opened 2016-06-14
Add form groups

Poltergeist:
~~This is based on #410~~ [~~un~~merged™️]

>This adds the functionallity to have multiple fields in one line.
It converts the schema to the right definition. In the model however
those groups are omitted because they are onlye needed for the view
layer

image
image

### DIFF
--- a/src/js/utils/SchemaFormUtil.js
+++ b/src/js/utils/SchemaFormUtil.js
@@ -117,7 +117,7 @@ let SchemaFormUtil = {
         return;
       }
 
-      let nextDefinition = definition.definition.find(
+      let nextDefinition = [].concat.apply([], definition.definition).find(
         function (definitionField) {
           return definitionField.name === path
             || definitionField.title === path;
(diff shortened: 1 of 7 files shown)


mesosphere-web:
Reviewer: @mlunoe

pierlo-upitup:
YAY!

src/js/utils/SchemaFormUtil.js
```diff
@@ -117,7 +117,7 @@ let SchemaFormUtil = {
         return;
       }
 
-      let nextDefinition = definition.definition.find(
+      let nextDefinition = [].concat.apply([], definition.definition).find(
```
orlandohohmeier: Let's be a little more verbose here and write `Array.prototype.concat.apply` this is not only easier to read but also don't create an unneeded instance of array. 

BTW:  Be aware that `concat.apply([],[[[1]]])` will return `[[1]]` not  `[1]`. So please make sure we only have on level of nested arrays?

pierlo-upitup:
Tested and working.

Poltergeist:
image
`cypress run` passes

orlandohohmeier:
LGTM! 🚢

PR #420, opened 2016-06-14
Add Job docker getter

orlandohohmeier:
/cc @jfurrow

PR #396, opened 2016-06-10
[Jobs Table: 1 of 2] Add additional structs for Jobs table

jfurrow:
These structs will be later used by the Jobs table.

This PR introduces two methods on the `Jobs` struct:
* `#getLastRunStatus` retrieves the status and time of the most recent run
* `#getScheduleStatus` retrieves the status of the schedule

PR #422, opened 2016-06-14
Introduce job util

orlandohohmeier:
---
ℹ️   _This branch depends on features that will be introduced with #396 and #420, thus it will fail until the respective branches get merged._

---

This util is used to create job specs from job structs as well as to create from models from jobs and respective form schemas.

### DIFF
--- a/src/js/utils/JobUtil.js
+++ b/src/js/utils/JobUtil.js
@@ -1,0 +1,102 @@
+import Job from '../structs/Job';
+
+// TODO: DCOS-7747 Move this method as well as `createFormModelFromSchema` into
+// the SchemaUtil and refactor it accordingly.
+const getFindPropertiesRecursive = function (job, item) {
+
+  return Object.keys(item).reduce(function (memo, subItem) {
+    if (item[subItem].type === 'object') {
+      memo[subItem] = getFindPropertiesRecursive(job, item[subItem].properties);
+
+      return memo;
+    }
+
+    memo[subItem] = item[subItem].default;
+
+    if (item[subItem].getter) {
+      memo[subItem] = item[subItem].getter(job);
+    }
+
+    return memo;
+  }, {});
+};
+
+const JobUtil = {
+  createJobFromFormModel: function (formModel) {
+    if (formModel == null) {
+      return new Job();
+    }
+
+    let spec = {};
+    let {
+      general = {
+        id: null,
+        description: null
+      },
+      labels = {},
+      docker,
+      schedule
+    } = formModel;
+
+    spec.id = general.id;
+    spec.description = general.description;
+
+    if (Array.isArray(labels.items)) {
+      spec.labels = labels.items.reduce(function (memo, {key, value}) {
+        memo[key] = value || null;
+
+        return memo;
+      }, {});
+    }
+
+    spec.run = {
+      cmd: general.cmd,
+      cpus: general.cpus,
+      mem: general.mem,
+      disk: general.disk,
+      docker
+    };
+
+    spec.schedules = [schedule];
+
+    return new Job(spec);
+  },
+
+  createFormModelFromSchema: function (schema, service = new Job()) {
+
+    return getFindPropertiesRecursive(service, schema.properties);
+  },
+
+  createJobSpecFromJob: function (job) {
+    let spec = {};
+
+    spec.id = job.getId() || null;
+    spec.description = job.getDescription() || null;
+
+    spec.run = {
+      cmd: job.getCommand(),
+      cpus: job.getCpus(),
+      mem: job.getMem(),
+      disk: job.getDisk(),
+    };
+
+    let labels = job.getLabels();
+    if (Object.keys(labels).length > 0) {
+      spec.labels = labels;
+    }
+
+    let docker = job.getDocker();
+    if (docker.image) {
+      Object.assign(spec.run, {docker});
+    }
+
+    let [schedule = {}] = job.getSchedules();
+    if (schedule.id != null && schedule.cron != null) {
+      spec.schedules = [schedule];
+    }
+
+    return spec;
+  }
+};
+
+module.exports = JobUtil;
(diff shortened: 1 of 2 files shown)


mesosphere-web:
Reviewer: @philipnrmn

src/js/utils/JobUtil.js
```diff
@@ -0,0 +1,91 @@
+import Job from '../structs/Job';
+
+const getFindPropertiesRecursive = function (job, item) {
```
MatApple: Not sure about the naming of this :) maybe we could avoid the `getFind`?

orlandohohmeier: I've to admit that I bluntly stole this one from the `ServiceUtil` and didn't thought about the naming. But yeah, let's find a better name for it.

src/js/utils/JobUtil.js
```diff
@@ -0,0 +1,91 @@
...
+    let spec = {};
+    let {general = {}, labels = {}, docker, schedule} = formModel;
+
+    spec.id = general.id;
+    spec.description = general.description;
+
+    if (Array.isArray(labels.items)) {
+      spec.labels = labels.items.reduce(function (memo, {key, value}) {
+        memo[key] = value || null;
+        return memo;
```
MatApple: #nit lets have a newline above `return`

src/js/utils/JobUtil.js
```diff
@@ -0,0 +1,91 @@
...
+  createJobFromFormModel: function (formModel) {
+    if (formModel == null) {
+      return new Job();
+    }
+
+    let spec = {};
+    let {general = {}, labels = {}, docker, schedule} = formModel;
+
+    spec.id = general.id;
+    spec.description = general.description;
```
MatApple: Should we have defaults for `general.id` and `general.description`?

orlandohohmeier: Yeah, good idea. But let's do that in a separate PR.

src/js/utils/JobUtil.js
```diff
@@ -0,0 +1,91 @@
...
+      docker: docker
+    };
+
+    spec.schedules = [schedule];
+
+    return new Job(spec);
+  },
+
+  createFormModelFromSchema: function (schema, job = new Job()) {
+
```
aldipower: remove extra line please

src/js/utils/JobUtil.js
```diff
@@ -0,0 +1,91 @@
+import Job from '../structs/Job';
+
+const getMatchingProperties = function (job, item) {
+
```
aldipower: no extra line

orlandohohmeier:

src/js/utils/JobUtil.js
```diff
@@ -0,0 +1,90 @@
+import Job from '../structs/Job';
+
+const getMatchingProperties = function (job, item) {
```
mlunoe: Could we get a description of this function or some comments? Not sure what exactly is going on here.
Picking out properties from a schema, but what is the `getter`?

src/js/utils/__tests__/JobUtil-test.js
```diff
@@ -0,0 +1,102 @@
...
+        id: 'test',
+        cmd: 'sleep 1000;'
+      });
+
+      expect(JobUtil.createJobSpecFromJob(job)).toEqual(
+        {
+          id: 'test',
+          description: null,
+          run: {
+            cpus: 0.01,
```
mlunoe: Not sure I understand where these values come from?

src/js/utils/JobUtil.js
```diff
@@ -0,0 +1,90 @@
...
+        return memo;
+      }, {});
+    }
+
+    spec.run = {
+      cmd: general.cmd,
+      cpus: general.cpus,
+      mem: general.mem,
+      disk: general.disk,
+      docker: docker
```
mlunoe: #nit this could be just `docker`

src/js/utils/__tests__/JobUtil-test.js
orlandohohmeier: It's default as defined in https://github.com/dcos/dcos-ui/blob/master/src/js/structs/Job.js

src/js/utils/JobUtil.js
Poltergeist: This is getting the model from the schema. `getter` is a attribute on the schema which is an anonymous function which returns the value for that field from the struct. This is used because a) it gives in the schema the possibility to transform the data from the struct to the form. Think key value to array, or array to comma separated list. b) it is a fairly simple approach which you can see directly on the schema. So if a field is filled out wrong there is just one point to look at.

This function is actually walking recursively  trough the schema and creates the model for the form from a struct. The idea is to fill out dynamically the model by giving a schema and a struct in this case the `jobs` struct.

src/js/utils/JobUtil.js
MatApple: Couldn't we just put them in L28? `{general = {id: null, description: null}, `?

src/js/utils/JobUtil.js
orlandohohmeier: We should introduced this method in the `SchemaUtil` instead and is should be capable to simply  create modal from a schema and additionally augmenting it with data object like the structs in this example.  
As we don't have time at the moment, I'd like to focus on the actual feature and do this refactoring in the next dev cycle. What do you think?

src/js/utils/JobUtil.js
orlandohohmeier: Yeah, why not.

src/js/utils/JobUtil.js
mlunoe: Yes, let's create a TODO for it! An move on!

mlunoe:
Feel free to merge once tests pass

orlandohohmeier:

PR #426, opened 2016-06-14
Fix Jobs icon

jfurrow:
I used the wrong icon before. Here's the new one:


mesosphere-web:
Reviewer: @pierlo-upitup

src/js/pages/JobsPage.js
```diff
@@ -2,7 +2,7 @@ import mixin from 'reactjs-mixin';
 import React from 'react';
 import {RouteHandler} from 'react-router';
 
-import IconJobs from 'babel!svg-react!../../img/components/icons/pages.svg?name=IconJobs';
+import IconJobs from 'babel!svg-react!../../img/components/icons/pages-code.svg?name=IconJobs';
```
MatApple: Are we going to load all SVGs like this? If so, maybe we should define the loaders in our webpack config so we don't have to prefix with `babel!svg-react!` here. What do you think?

jfurrow: Oh good call! I didn't realize we could do that, but will get it in this PR.

kennyt:
merge once mat's comment is done

src/js/pages/JobsPage.js
jfurrow: This is done now!

PR #431, opened 2016-06-15
Fix multiple row adding

Poltergeist:
This fixes multiple row adding. So if there is no field present the next field is added right above the button and not at the beginning of the field.
For clarification here the bug in action: 
screen recording 2016-06-15 at 02 34 pm

### DIFF
--- a/src/js/schemas/service-schema/Labels.js
+++ b/src/js/schemas/service-schema/Labels.js
@@ -9,10 +9,6 @@ const Labels = {
       getter: function (service) {
         let labels = service.getLabels() || {};
 
-        if (Object.keys(labels).length === 0) {
-          return null;
-        }
-
         return Object.keys(labels).map(function (key) {
           return {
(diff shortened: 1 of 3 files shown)


mesosphere-web:
Reviewer: @orlandohohmeier

src/js/schemas/service-schema/Labels.js
```diff
@@ -9,6 +9,10 @@ const Labels = {
       getter: function (service) {
         let labels = service.getLabels() || {};
 
+        if (Object.keys(labels).length === 0) {
+          return null;
```
orlandohohmeier: Do we really need to return `null` instead of an empty array?

Poltergeist: Changed this problem at the root, so if now a array is provided with a length === 0 the default row is created otherwise the provided array is used.

PR #437, opened 2016-06-15
Use classnames/dedupe in TabForm

jfurrow:
Allow these static classes to be overridden.


mesosphere-web:
Reviewer: @daiyiwhale

PR #439, opened 2016-06-16
Improve application loading

rcorral:
We will now wait for Mesos request to come back before we render the application.


mesosphere-web:
Reviewer: @mlunoe

src/js/index.js
```diff
@@ -56,38 +59,90 @@ RequestUtil.json = function (options = {}) {
...
-        let routes = RouterUtil.buildRoutes(appRoutes.getRoutes());
-        let router = Router.run(routes, function (Handler, state) {
-          Config.setOverrides(state.query);
-          ReactDOM.render(
-            (<Provider store={PluginSDK.Store}>
-              <Handler state={state} />
-            </Provider>),
-            domElement);
+      let loadedResources = {
+        mesos: false
```
MatApple: Lets put this on one line!

src/js/index.js
```diff
@@ -56,38 +59,90 @@ RequestUtil.json = function (options = {}) {
...
+      };
+      let mesosEvents = [MESOS_SUMMARY_CHANGE, MESOS_SUMMARY_REQUEST_ERROR];
+
+      // Responsible for checking if application is ready to render
+      function delayOrLoadApplication() {
+        // Check if we've loaded all resources first
+        let resources = Object.keys(loadedResources);
+        let allLoaded = resources.reduce(function (memo, resource) {
```
MatApple: This could be `let allLoaded = Object.values(loadedResources).some(function (loaded) { return !loaded; });`. We have `Object.values` via babel polyfill.

src/js/index.js
rcorral: Potentially Marathon. But not right now.

src/js/index.js
```diff
@@ -56,38 +59,90 @@ RequestUtil.json = function (options = {}) {
...
+          if (loadedResources[resource] === false) {
+            memo = false;
+          }
+
+          return memo;
+        }, true);
+
+        if (allLoaded === false) {
+          return;
```
MatApple: Don't we want to make sure we call this function again to check we're all loaded?
```
setTimeout(delayOrLoadApplication, msLeftOfDelay);
return;
```

src/js/index.js
```diff
@@ -56,38 +59,90 @@ RequestUtil.json = function (options = {}) {
...
+      function onMesosSummaryChange() {
+        // Keep polling until the system attaches another listener to summary
+        function keepPollingAlive() {
+          if (MesosSummaryStore.listeners(MESOS_SUMMARY_CHANGE).length > 1) {
+            MesosSummaryStore.removeChangeListener(
+              MESOS_SUMMARY_CHANGE, keepPollingAlive
+            );
+          }
+        }
+        MesosSummaryStore.addChangeListener(
```
MatApple: Don't think you want this nested inside `onMesosSummaryChange` do we? We're adding a new `keepPollingAlive` callback every time the `onMesosSummaryChange` is called.

src/js/index.js
```diff
@@ -56,38 +59,90 @@ RequestUtil.json = function (options = {}) {
...
-            </Provider>),
-            domElement);
+      let loadedResources = {
+        mesos: false
+      };
+      let mesosEvents = [MESOS_SUMMARY_CHANGE, MESOS_SUMMARY_REQUEST_ERROR];
+
+      // Responsible for checking if application is ready to render
+      function delayOrLoadApplication() {
+        // Check if we've loaded all resources first
```
MatApple: I'd suggest we have a `alreadyLoaded` flag so we don't run into accidentally calling `renderApplicationToDOM` more than once. If we have more than 1 item in `loadedResources` then this could happen because we're using `setTimeouts`.

src/js/index.js
```diff
@@ -56,38 +59,90 @@ RequestUtil.json = function (options = {}) {
...
+      }
+
+      // Mesos
+      MesosSummaryStore.init();
+      mesosEvents.forEach(function (event) {
+        MesosSummaryStore.addChangeListener(event, onMesosSummaryChange);
       });
+
+      function renderApplicationToDOM() {
+        setTimeout(function () {
```
MatApple: #nit what do you think about wrapping the call to `renderApplicationToDOM` in L88 with this `setTimeout` instead? Keeps this function clean.

src/js/index.js
rcorral: Good idea, done.

src/js/stores/MesosSummaryStore.js
```diff
@@ -169,7 +180,9 @@ class MesosSummaryStore extends GetSetBaseStore {
     let states = this.get('states');
 
     if (typeof data.date !== 'number') {
-      data.date = Date.now();
+      let lastRequestTime = this.getNextRequestTime();
+      this.set({lastRequestTime: lastRequestTime});
```
MatApple: `{lastRequestTime}`

src/js/index.js
rcorral: I can't think of a scenario where `renderApplicationToDOM` would get called twice. Because it doesn't go to that portion of the logic until everything has been loaded and at that point we're just waiting for the delay to pass.

src/js/stores/MesosSummaryStore.js
```diff
@@ -169,7 +180,9 @@ class MesosSummaryStore extends GetSetBaseStore {
     let states = this.get('states');
 
     if (typeof data.date !== 'number') {
-      data.date = Date.now();
+      let lastRequestTime = this.getNextRequestTime();
```
MatApple: Isn't this going to return the next request time or the current time if it's the first time this gets called?

src/js/index.js
```diff
@@ -56,38 +59,85 @@ RequestUtil.json = function (options = {}) {
...
-            domElement);
+      let loadedResources = {
+        mesos: false
+      };
+      let mesosEvents = [MESOS_SUMMARY_CHANGE, MESOS_SUMMARY_REQUEST_ERROR];
+
+      // Responsible for checking if application is ready to render
+      function delayOrLoadApplication() {
+        // Check if we've loaded all resources first
+        let allLoaded = Object.values(loadedResources).some(function (loaded) {
```
kennyt: isn't this variable really `someNotLoaded`?

src/js/index.js
rcorral: Good idea, done.

src/js/index.js
rcorral: Yup, renamed.

src/js/stores/MesosSummaryStore.js
rcorral: Yes. That's because the first request to the history server may fail.

src/js/stores/MesosSummaryStore.js
```diff
@@ -186,17 +199,31 @@ class MesosSummaryStore extends GetSetBaseStore {
...
+        data.unshift(Util.deepCopy(data[0]));
+      }
+    }
+
     // Multiply Config.stateRefresh in order to use larger time slices
     data = MesosSummaryUtil.addTimestampsToData(data, Config.getRefreshRate());
     data.forEach((datum) => {
       this.processSummary(datum, {silent: true});
     });
+    this.set({lastRequestTime: Date.now()});
```
MatApple: I think we could run into animation issues here because our interval isn't accurate - we effectively have `Config.getRefreshRate() + requestResponseTime`

src/js/stores/MesosSummaryStore.js
```diff
@@ -186,17 +199,31 @@ class MesosSummaryStore extends GetSetBaseStore {
...
+    this.emit(MESOS_SUMMARY_CHANGE);
   }
 
   processSummaryError(options = {}) {
     let unsuccessfulSummary = new StateSummary({successful: false});
     let states = this.get('states');
 
+    this.set({lastRequestTime: this.getNextRequestTime()});
```
MatApple: Should this be the same as L217?

rcorral:
retest this please

rcorral:
retest this please

src/js/utils/ApplicationUtil.js
```diff
@@ -0,0 +1,60 @@
...
+import Config from '../config/Config';
+import MesosSummaryStore from '../stores/MesosSummaryStore';
+
+module.exports = {
+  /**
+   * Checks that enough time has elapsed before rendering app
+   *
+   * @param  {Func} onDelayEndCallback Gets called after delay has ellapsed
+   */
+  renderOnDelayEnd(onDelayEndCallback) {
```
MatApple: Just an idea but seeing this isn't actually doing the rendering, maybe it could be `invokeAfterPageLoad(fn, delay)` and you could pass in `Config.applicationRenderDelay` as the `delay` argument to keep this more functional.

src/js/utils/ApplicationUtil.js
```diff
@@ -0,0 +1,60 @@
...
+  MESOS_SUMMARY_REQUEST_ERROR
+} from '../constants/EventTypes';
+import Config from '../config/Config';
+import MesosSummaryStore from '../stores/MesosSummaryStore';
+
+module.exports = {
+  /**
+   * Checks that enough time has elapsed before rendering app
+   *
+   * @param  {Func} onDelayEndCallback Gets called after delay has ellapsed
```
kennyt: `ellapsed` is spelled `elapsed`

src/js/utils/ApplicationUtil.js
rcorral: Did the first part, the second part I don't think we need just now, so I didn't do it.

src/js/utils/ApplicationUtil.js
rcorral: Done.

kennyt:
px there's a quick flash of no services, did that come from this PR?

rcorral:
Yeah that's fine for now. It didn't come from this PR though. It was already like that.

rcorral:
Tests on my branch: img
Tests on master: img

PR #443, opened 2016-06-16
Feature - Service version dropdown and displaying

aldipower:
This adds the version dropdown to the service configuration tab to load and select previous configurations.

* [x] Fetch versions
* [x] Implement dropdown
* [x] Show selected configuration
* [x] Tests
* [x] Integration tests with freshly added versions endpoint

dropdown-1

dropdown-2


mesosphere-web:
Reviewer: @orlandohohmeier

src/js/components/ServiceDetailConfigurationTab.js
```diff
@@ -1,19 +1,100 @@
...
+  getVersionsDropdown() {
+    let {service} = this.props;
+    let versions = service.getVersions();
+
+    if (versions.size < 2) {
+      return null;
+    }
+
+    let versionItems = [];
+    for (let version of versions.keys()) {
```
aldipower: I actually would prefer the functional way with a .map()-function but couldn't find a straight forward way to do this with a `Map`.

orlandohohmeier: This would work, but is may a little to much... `Array.from(versions.keys()).map(function (key) { return key} );`

orlandohohmeier:

pierlo-upitup:
We also want to be able to `Edit` and `Rollback` to the selected version (the latter only if the selected version is not the current). Will these land in a separate PR?

src/js/components/ServiceDetailConfigurationTab.js
```diff
@@ -1,19 +1,100 @@
...
+      selectedVersionID: null
+    };
+
+    METHODS_TO_BIND.forEach(method => {
+      this[method] = this[method].bind(this);
+    });
+  }
+
+  componentWillMount() {
+    this.setState({
```
rcorral: can we do `super.componentWillMount(...arguments);`

src/js/components/ServiceDetailConfigurationTab.js
```diff
@@ -1,19 +1,100 @@
...
+import DCOSStore from '../stores/DCOSStore';
 import Service from '../structs/Service';
 
+const METHODS_TO_BIND = [
+  'handleVersionSelection'
+];
+
 class ServiceDetailConfigurationTab extends React.Component {
+  constructor() {
+    super();
```
rcorral: can you add `...arguments`

src/js/components/ServiceDetailConfigurationTab.js
rcorral: Why can't we just use a `map`?
```
let versionItems = versions.map(function (version) {
...
});
```

orlandohohmeier: `versions` is an instance of [`Map`](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Map) and doesn't provide an array like `map` function.  Therefor we need to get the `keys()` before we can iterate them, which is similar to what we would have done if we would have used a plain `Object` to build the versions map.

aldipower:
@pierlo-upitup For the sake of smaller PRs the buttons are not part of this PR!

src/js/components/ServiceDetailConfigurationTab.js
aldipower: After I've added this I removed it again, because it makes the tests fail on the server for clear reasons.
componentWillMount is part of ReactDOM or React Native, which is not existent in this test and _not_ part of React.Component itself.
It's also a little bit doubtfully to me to call this explicitly without any actual need. That is against the principle of using things only if you need them.

philipnrmn:

philipnrmn:
Nice implementation. Is this considered styled? The dropdown changes width on selection.

philipnrmn:
Works On My Machine™
screen shot 2016-06-17 at 14 53 26

After a quick consult with @aldipower I'd like to merge this now - the width jump can be addressed in a separate PR.

PR #444, opened 2016-06-16
Add performance boost

Poltergeist:
This adds a performance hotfix which has been created by @orlandohohmeier.

 
image

image

### DIFF
--- a/src/js/components/modals/ServiceFormModal.js
+++ b/src/js/components/modals/ServiceFormModal.js
@@ -119,7 +119,8 @@ class ServiceFormModal extends mixin(StoreMixin) {
           'serviceCreateSuccess',
           'serviceEditError',
           'serviceEditSuccess'
-        ]
+        ],
+        suppressUpdate: true
       }
     ];
 
@@ -133,6 +134,13 @@ class ServiceFormModal extends mixin(StoreMixin) {
     if (!this.props.open && nextProps.open) {
       this.resetState();
     }
+  }
+
+  shouldComponentUpdate(nextProps, nextState) {
+    let {state, props} = this;
+    return props.open !== nextProps.open ||
+      state.jsonMode !== nextState.jsonMode ||
+      state.errorMessages !== nextState.errorMessage;
   }
 
   resetState() {
@@ -166,7 +174,7 @@ class ServiceFormModal extends mixin(StoreMixin) {
     } catch (e) {
 
     }
-    this.setState({jsonDefinition, service})
+    this.setState({jsonDefinition, service});
   }
 
   handleJSONToggle() {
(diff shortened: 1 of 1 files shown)


mesosphere-web:
Reviewer: @kennyt

src/js/components/modals/ServiceFormModal.js
```diff
@@ -135,6 +136,13 @@ class ServiceFormModal extends mixin(StoreMixin) {
     }
   }
 
+  shouldComponentUpdate(nextProps, nextState) {
+    let {state, props} = this;
+    return props.open !== nextProps.open ||
+      state.jsonMode !== nextState.jsonMode ||
+      state.errorMessages !== nextState.errorMessage;
+  }
```
pierlo-upitup: could there be a case where the user didn't fix the error message and we need to re-render it?

orlandohohmeier: We're re-rendering if the `errorMessage` is updated (e.g. cleared) or (!) the `jsonMode` is toggled or the modal open property is toggled. Every other case should be handled by the child components.

Poltergeist: the errorMessage gets set to null on change (reads: focus/blur/change) thus we have to re-render.

src/js/components/modals/ServiceFormModal.js
```diff
@@ -135,6 +136,13 @@ class ServiceFormModal extends mixin(StoreMixin) {
     }
   }
 
+  shouldComponentUpdate(nextProps, nextState) {
+    let {state, props} = this;
+    return props.open !== nextProps.open ||
+      state.jsonMode !== nextState.jsonMode ||
+      state.errorMessages !== nextState.errorMessage;
```
kennyt: here it's comparing `errorMessages` with `errorMessage`. the second one is not plural. is that what we want?

kennyt:
i'd merge after my comment is addressed

rcorral:
retest this please

src/js/components/modals/ServiceFormModal.js
Poltergeist: Good catch

Poltergeist:
image

Poltergeist:

Poltergeist:

src/js/components/modals/ServiceFormModal.js
```diff
@@ -166,7 +174,7 @@ class ServiceFormModal extends mixin(StoreMixin) {
     } catch (e) {
 
     }
-    this.setState({jsonDefinition, service})
+    this.setState({jsonDefinition, service});
```
aldipower: Strike!

PR #455, opened 2016-06-17
Add error handling to tab form

kennyt:


mesosphere-web:
Reviewer: @MatApple

src/js/components/TabForm.js
```diff
@@ -21,6 +24,7 @@ class TabForm extends React.Component {
     });
 
     this.triggerSubmit = function () {};
+    this.isValidated = false;
```
rcorral: Let's not set properties on the instance of the class, this is a bad practice. We can either opt to use internal storage or state if it makes sense to re-render.

rcorral: Also consider renaming to `isFormValid`

kennyt: both done

pierlo-upitup:
Aren't we supposed to move on from these mixins sooner rather than later? Why would we want to introduce the internal storage mixin instead of using good old state?

rcorral:
Sometimes doing a setState when you don't actually need to render is a real bugger because of child components having to re-render all the way down. Specially when we don't implement `shouldComponentUpdate` everywhere.

src/js/components/TabForm.js
```diff
@@ -49,18 +54,30 @@ class TabForm extends React.Component {
     this.setState({currentTab});
   }
 
+  handleFormError() {
+    this.internalStorage_update({isFormValidated: false});
+  }
```
pierlo-upitup: Why can't we just `this.props.onError` here instead of setting this flag and moving the error-handling logic to `this.handleExternalSubmit` ? Imho `onError` should be the only handler invoked when an error occurred, and the parent component should be responsible to decide wether or not to proceed with submit. This feels unintuitive to me.

kennyt: the reason for this is kind of complicated, and it's the best solution i can think of so far.

`TabForm` is kind of a special form because it actually renders many `Form`s. When you call a submit on `TabForm`, it then iterates through the `Forms` and calls submit on each of them. if one of those `Forms` error out, we need to set some flag that there was an error, so that when we're done iterating through the `Form`s, we can either decide to call `onError` or `onSubmit`

we don't call `setState` because we don't need to re-render and `setState` is async in some mysterious ways. doing this guarantees it to be synchronous

i'm not sure i got my point 100% across but i don't know of a better solution

rcorral:
@pierlo-upitup could you take a look if you get a chance please?
If not maybe we can merge this in and then revisit later.

orlandohohmeier:
@rcorral I'd say let's merge this now and revisit this later as I fear that any other solution might involve huge refactoring.  Would you my creating an issue to keep track of this?

rcorral:
Created a follow up task for this.

PR #457, opened 2016-06-19
Introduce update job action

orlandohohmeier:
Add an action to update jobs.


mesosphere-web:
Reviewer: @kennyt

src/js/events/ChronosActions.js
```diff
@@ -133,6 +134,26 @@ const ChronosActions = {
     });
   },
 
+  updateJob: function (jobID, data) {
```
aldipower: I am not familiar with the API yet, so forgive me my question.
Is _update_ the right wording here?
At the MarathonActions we call this part `editService` for example instead of update.
Just a question?!

src/js/events/ChronosActions.js
```diff
@@ -133,6 +134,26 @@ const ChronosActions = {
...
+    RequestUtil.json({
+      url: `${Config.rootUrl}/chronos/v0/scheduled-jobs/${jobID}`,
+      method: 'PUT',
+      data,
+      success: function () {
+        AppDispatcher.handleServerAction({
+          type: REQUEST_CHRONOS_JOB_CREATE_SUCCESS
```
aldipower: `REQUEST_CHRONOS_JOB_CREATE_SUCCESS` while doing an _update_ is kind of confusing.

src/js/events/ChronosActions.js
Poltergeist: I have the same disadvantage like you `@aldipower` my interpretation of the jobs would be that update is right because you update a timer but you edit the time? Jobs are running in a ir-/regular timeframe Service do run constantly. That is my interpretation but it doesn't mean I am right.

src/js/events/ChronosActions.js
orlandohohmeier: Whoops, I forgot to rename those...

src/js/events/ChronosActions.js
orlandohohmeier: The endpoint is documented as follows, but I'm fine to use a different name if you think "editJob" works better.

> Update a job. This endpoint will overwrite the complete job, it will not support PATCH-like updates.

What do you think?

aldipower: OK, difficult to say. Let's stay with _update_ imo.

rcorral:
Looks good! Could you put a screenshot of the integration tests passing?

orlandohohmeier:
Integration test results: 
>  `128 passing (4m)`
>  `19 pending`

PR #458, opened 2016-06-20
Add current group to the the app id field in the create modal

Poltergeist:
This adds the current group to the create modal so that the application id is prefilled with the current group id.

image

The regular expression `/^(\/.+)$/` is looking for the beginning of the string and matches if the string starts with a `/` and does contain more characters after the slash. This is combined into a group and then replaced with the first group which is the complete string and a `/` is appended. This is needed because in most case a path like `/group/another-group` will be given by `getId` except on root then the return value of `getId` would be `/` so in most cases we want to append a `/` so that the user can begin typing the id of their application.

another approach could be this
```JS
let applicationId = item.getId();
if (applicationId !== '/') {
  applicationId = applicationId + '/';
}
let service = new Service({id: applicationId});
```
I prefer the regex way because it is shorter, but I am also fine with changing to the solution proposed above.

### DIFF
--- a/src/js/pages/services/ServicesTab.js
+++ b/src/js/pages/services/ServicesTab.js
@@ -256,6 +256,8 @@ var ServicesTab = React.createClass({
     // Find item in root tree and default to root tree if there is no match
     let item = DCOSStore.serviceTree.findItemById(id) || DCOSStore.serviceTree;
 
+    let service = new Service({id: item.getId().replace(/^(\/.+)$/, '$1/')});
+
     // Make sure to grow when logs are displayed
     let routes = this.context.router.getCurrentRoutes();
     let classes = classNames({
@@ -270,6 +272,7 @@ var ServicesTab = React.createClass({
           parentGroupId={item.getId()}
           onClose={this.handleCloseGroupFormModal}/>
         <ServiceFormModal open={state.isServiceFormModalShown}
+          service={service}
           onClose={this.handleCloseServiceFormModal}/>
       </div>
     );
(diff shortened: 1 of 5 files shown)


mesosphere-web:
Reviewer: @mlunoe

src/js/pages/services/ServicesTab.js
```diff
@@ -256,6 +256,8 @@ var ServicesTab = React.createClass({
     // Find item in root tree and default to root tree if there is no match
     let item = DCOSStore.serviceTree.findItemById(id) || DCOSStore.serviceTree;
 
+    let service = new Service({id: item.getId().replace(/^(\/.+)$/, '$1/')});
```
kennyt: add a comment for this and i think it's good

rcorral:
An integration test for this would be awesome!

src/js/pages/services/ServicesTab.js
orlandohohmeier: I don't feel comfortable with instantiating a new service on every render call, just to pass the group/context. Let's introduce a prop to pass the context into the modal.

orlandohohmeier: /ping @Poltergeist

Poltergeist:

Poltergeist:
image

src/js/pages/services/ServicesTab.js
Poltergeist: @orlandohohmeier Changed the way to a prop so that the id is passed as a prop.

PR #459, opened 2016-06-20
Add Force update to edit service if service is in deployment

Poltergeist:


mesosphere-web:
Reviewer: @jfurrow

src/js/components/modals/ServiceFormModal.js
```diff
@@ -209,6 +210,7 @@ class ServiceFormModal extends mixin(StoreMixin) {
 
   onMarathonStoreServiceEditError(errorMessage) {
     this.setState({
+      force: errorMessage.message && /force=true/.test(errorMessage.message),
```
jfurrow: What do you think about putting this `/force=true/.test...` check in a separate method on the `ServiceFormModal`? Something like `#shouldForceEdit` that accepts the `message` and returns `/force=true/.test(message)`.

jfurrow:
One minor comment, but this looks good!

src/js/components/modals/ServiceFormModal.js
Poltergeist: Comment has been addressed.

src/js/events/MarathonActions.js
```diff
@@ -89,9 +89,15 @@ module.exports = {
     });
   },
 
-  editService: function (data) {
+  editService: function (data, force) {
+    let url = `${Config.rootUrl}${Config.marathonAPIPrefix}/apps/${data.id}`;
+
+    if (force === true) {
+      url += '?force=true';
```
rcorral: It would be good to add a simple unit test to make sure that the `force` argument changes the requested URL.

MatApple:
Hey @Poltergeist can you post a screenshot of integration tests passing? After that, and Rafas comment I think this looks good!

src/js/components/modals/ServiceFormModal.js
```diff
@@ -205,13 +206,18 @@ class ServiceFormModal extends mixin(StoreMixin) {
     });
   }
 
+  shouldForceUpdate(message = this.errorMessage) {
```
pierlo-upitup: Probably unrelated, but why aren't we storing `errorMessage` in this.state?

Poltergeist:
image

image

Added a unit test and fixed the integration tests.

src/js/components/modals/ServiceFormModal.js
Poltergeist: shouldForceUpdate is called in the setStates. This is why this function takes a parameter. But in fact the default is wrong. Changed it.

PR #463, opened 2016-06-21
Transform fillRule attribute to fill-rule

jfurrow:
Forces the `fillRule` prop, generated by `svg-react-loader`, to be rendered as `fill-rule`. This is necessary for SVG icons in React 0.14, but can be removed in React 15.x.

In all its glory...


mesosphere-web:
Reviewer: @aldipower

PR #465, opened 2016-06-21
Add default values to the service deploy modal schema

Poltergeist:
This adds default values to the Service Schema according to the marathon ui values.

```JSON
{
  "id": null,
  "cmd": null,
  "cpus": 1,
  "mem": 128,
  "disk": 0,
  "instances": 1
}
```
image

### DIFF
--- a/src/js/schemas/service-schema/General.js
+++ b/src/js/schemas/service-schema/General.js
@@ -22,6 +22,7 @@ let General = {
           title: 'CPUs',
           description: 'The amount of CPUs which are used for the service',
           type: 'number',
+          default: 1,
           getter: function (service) {
             return `${service.getCpus() || ''}`;
           }
@@ -29,6 +30,7 @@ let General = {
         mem: {
           title: 'Mem (MiB)',
           type: 'number',
+          default: 128,
           getter: function (service) {
             return `${service.getMem() || ''}`;
           }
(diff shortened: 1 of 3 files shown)


mesosphere-web:
Reviewer: @MatApple

rcorral:
@Poltergeist looks great, could you post a screenshot of the integration tests passing?

kennyt:
well, the `0` isn't showing up because inside of the `getter` of the schema, you do `disk || ''` and `0` will not be applied

Poltergeist:
image

I added also some integration tests for the defaults.

src/js/schemas/service-schema/General.js
```diff
@@ -22,29 +22,33 @@ let General = {
...
+            return `${service.getDisk() || 0}`;
           }
         },
         instances: {
           title: 'Instances',
           type: 'number',
+          default: 1,
           getter: function (service) {
-            return `${service.getInstancesCount() || 0}`;
+            return `${service.getInstancesCount() || ''}`;
```
pierlo-upitup: why empty string here and not 1 ?

Poltergeist: It actually needs to be 0. This is the default which is set if no value is available. so if getInstancesCount would return false for example the input field should show a 0 and not an empty string. I changed this behaviour for all number fields.

kennyt: but right here, you're still return `''` not `0`, no?

if you change this feel free to merge

Poltergeist: I changed that behaviour in a commit after the comment ;-)

PR #470, opened 2016-06-21
[1 of n: Icons] Implement page icons

jfurrow:

### DIFF
--- a/src/js/components/HostTable.js
+++ b/src/js/components/HostTable.js
@@ -5,6 +5,7 @@ import {StoreMixin} from 'mesosphere-shared-reactjs';
 import {Tooltip} from 'reactjs-components';
 
 var HostTableHeaderLabels = require('../constants/HostTableHeaderLabels');
+import IconServers from '../../img/components/icons/icon-medium/servers.svg?name=IconServers';
 var InternalStorageMixin = require('../mixins/InternalStorageMixin');
 var ResourceTableUtil = require('../utils/ResourceTableUtil');
 var ProgressBar = require('./charts/ProgressBar');
@@ -22,7 +23,7 @@ var HostTable = React.createClass({
   statics: {
     routeConfig: {
       label: 'Nodes',
-      icon: 'datacenter',
+      icon: <IconServers />,
       matches: /^\/nodes/
     }
   },
--- a/src/js/components/AccessDeniedPage.js
+++ b/src/js/components/AccessDeniedPage.js
@@ -39,8 +39,6 @@ module.exports = class AccessDeniedPage extends React.Component {
             flex-container-col">
             <AlertPanel
               footer={this.getFooter()}
-              iconClassName="icon icon-sprite icon-sprite-jumbo
-                icon-sprite-jumbo-white icon-lost-planet flush-top"
               title="Access Denied">
               <p>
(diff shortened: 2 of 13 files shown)


mesosphere-web:
Reviewer: @Poltergeist

jfurrow:
This broke tests, fixing now.

jfurrow:
retest this please

kennyt:
there's a legit failure in unit tests

jfurrow:
@kennyt derp, I thought it was integration tests for some reason.

The unit tests are fixed by #475

pierlo-upitup:
From a design standpoint, I gotta say I find the contrast between the sharp icon outlines and the antialiased fonts a tad incoherent. But I'm probably out of my element – feel free to ignore this, @jfurrow  @leemunroe @ashenden ?

jfurrow:
@pierlo-upitup In the screenshot above, the `fill` attribute was set inline for each icon to `#000` which might be to blame for the incoherent contrast. That's fixed in #471. Here's a recent screenshot (although @ashenden has suggested we bump up the opacity):

src/js/components/HostTable.js
```diff
@@ -5,6 +5,7 @@ import {StoreMixin} from 'mesosphere-shared-reactjs';
 import {Tooltip} from 'reactjs-components';
 
 var HostTableHeaderLabels = require('../constants/HostTableHeaderLabels');
+import IconServers from '../../img/components/icons/icon-medium/servers.svg?name=IconServers';
```
Poltergeist: How does this work, I mean how does it create a component from a plain svg file. I do like it, is this a React functionality?

jfurrow: It's actually a Webpack loader that takes the SVG and returns a React component. It's this one here: https://github.com/jhamlet/svg-react-loader

Poltergeist: Thanks I was very impressed and also a bit scared that this worked :-)

jfurrow: Hahaha, understandably! We opted to go this route (as opposed to creating simple React components that return SVG markup) to allow the design team to easily maintain the icons going forward.

Poltergeist:
There is one thing which I do not like about the icons, and that is that there are so many of them why aren't we using SVG sprites with svg symbols?

jfurrow:
@Poltergeist To be honest I did not see the benefit either way. I'm not opposed to using SVG sprites, but why do you prefer them?

Poltergeist:
Well with the use of the SVG to react component loader we are moving all the SVG code to the Javascript base. Which means that we are moving about 2MB of svg code into our already about 5 MB Javascript code base. Also we are adding a bit of react boilerplate onto every single svg. That is my first concern. 

I do like symbols personally a lot because I only must write a very small svg element and I have my logo with the name I probably know. The styling could still be done with css. So size and color for example. This is my opinion which is not that important. 

But I am very concerned about the file size of the javascript code. But anyways this is not connected to this PR it looks good to me. 🚢

jfurrow:
@Poltergeist That's a very good point. After discussing with @rcorral, I am going to pursue that route. Good catch!

jfurrow:
Closing to implement SVG sprite!

PR #475, opened 2016-06-21
[0 of n: Icons] Replace icons in tests with empty React elements

jfurrow:
Icons will break tests because they need to be run through a Webpack loader first. This replaces all icons with empty React elements.

Huge thanks to @rcorral for the help here!

### DIFF
--- a/jest/preprocessor.js
+++ b/jest/preprocessor.js
@@ -16,7 +16,7 @@ module.exports = {
       if (babel.util.canCompile(filename)) {
         // This will match:
         // import IconFoo from '../../icons/icon-medium/pages-code.svg?name=foo';
-        var matches = src.match(/(\w+) (.*?) (\w+)(.*?)(icons)(.*?)\.svg(\?.*?)$/img);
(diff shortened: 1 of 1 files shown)


mesosphere-web:
Reviewer: @MatApple

jest/preprocessor.js
```diff
@@ -14,6 +14,18 @@ module.exports = {
       }
       // Run our modules through Babel before running tests
       if (babel.util.canCompile(filename)) {
+        // This will match:
+        // import IconFoo from '../../icons/icon-medium/pages-code.svg?name=foo';
+        var matches = src.match(/(\w+) (.*?) (\w+)(.*?)(icons)(.*?)\.svg(\?.*?)$/img);
```
rcorral: @jfurrow could you check if this works too:
```
var matches = src.match(/^(\w+) (.*?) (\w+|=) '(.*?)(icons)(.*?)\.svg(\?.*?)';$/img);
```
I want to make sure we don't match other things on accident.

jfurrow: Yep, this works great as well. I'll use this instead.

PR #483, opened 2016-06-22
Add concurrency policy to schedule

orlandohohmeier:
Add a concurrency policy to the job schedule if it's undefined, as a schedule without a policy is invalid.

This is a hotfix, and we will replace it with a schedule struct that has proper defaults. We will also provide this as an option in the create/edit modal once we support other policies.

### DIFF
--- a/src/js/utils/JobUtil.js
+++ b/src/js/utils/JobUtil.js
@@ -90,6 +90,9 @@ const JobUtil = {
 
     let [schedule = {}] = job.getSchedules();
     if (schedule.id != null && schedule.cron != null) {
+      if (schedule.concurrencyPolicy == null) {
+        schedule.concurrencyPolicy = 'ALLOW';
+      }
       spec.schedules = [schedule];
     }
 
--- a/src/js/utils/__tests__/JobUtil-test.js
+++ b/src/js/utils/__tests__/JobUtil-test.js
@@ -101,5 +101,25 @@ describe('JobUtil', function () {
           }
         });
     });
+
+    it('should add concurrencyPolicy if schedule is defined', function () {
+      const job = new Job({
+        id: 'test',
+        run: {
+          cmd: 'sleep 1000;'
+        },
+        schedules: [{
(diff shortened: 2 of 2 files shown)


mesosphere-web:
Reviewer: @pierlo-upitup

src/js/utils/JobUtil.js
```diff
@@ -90,6 +90,9 @@ const JobUtil = {
 
     let [schedule = {}] = job.getSchedules();
     if (schedule.id != null && schedule.cron != null) {
+      if (schedule.concurrencyPolicy == null) {
+        schedule.concurrencyPolicy = 'ALLOW';
```
Poltergeist: Does it make sense to have a constant for `ALLOW`?

Poltergeist:
One question, everything else LGTM.

src/js/utils/JobUtil.js
orlandohohmeier: Does it provide any value to have a constant?

pierlo-upitup: nah...

pierlo-upitup:
Makes sense. LGTM.

src/js/utils/JobUtil.js
pierlo-upitup: But one question: shouldn't serverside validation itself fail if the user has not provided one value? Or shouldn't it add a default for us?

orlandohohmeier: Yeah, it would be nice if the backend could add the default for us. At the moment it just fails.

pierlo-upitup: That's bad. Have we also double-checked that `ALLOW` is the most sensible default?

orlandohohmeier: There is no other "valid" value than `ALLOW` at the moment.

orlandohohmeier:
Tests passing locally:
>  `135 passing (4m)`
>  `19 pending`

PR #486, opened 2016-06-22
Removing fill prop and adding icon-tiny family

ashenden:
* Removed `fill` attribute from SVGs.
* Added new Clipboard icon `icon-clipboard`
* Added new icon family `icon-tiny` for use in some specific UI components like breadcrumbs.


ashenden:
@jfurrow how do I figure out what failed above?

jfurrow:
@ashenden Our integration tests are failing currently. Don't worry about this right now.

PR #487, opened 2016-06-22
[3 of n: Icons] Introduce Icon component

jfurrow:
This is a simple component to make using icons as simple as possible (hopefully). The only thing that may be confusing is the `family` vs. `size` props. From @ashenden's "ICONS ICONS ICONS" email:
>Though our names imply size (e.g. "icon-small”) we think of them as a collection/family.  We could very well call these “icon-banana” or “icon-little-outline-style".  The name is only useful for letting us quickly relate the icons in this collection to each other.  It just so happens that those same names map nicely to the sizes we have in product.

For now we have only four families. If you need an icon larger than `meidum`, then use the `medium` family and whatever size you need.

What size/family do you think should be default? I've defaulted to `medium` but I'm not sure what's best.

### DIFF
--- a/src/js/components/Icon.js
+++ b/src/js/components/Icon.js
@@ -1,0 +1,60 @@
+import classNames from 'classnames';
+import React from 'react';
+
+import Util from '../utils/Util';
+
+class Icon extends React.Component {
+  render() {
+    let {props} = this;
+
+    let additionalProps = Util.omit(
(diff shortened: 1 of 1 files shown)


mesosphere-web:
Reviewer: @kennyt

jfurrow:
Integration tests passing locally:

PR #488, opened 2016-06-22
Add Health Checks to ServiceFormModal

MatApple:
### Note
- Has issues with preserving order of healthchecks. Altering the last healthcheck will re-render that healthcheck at the top. @kennyt is working on a fix for this.
- Haven't added required fields/client-side validation. cc/ @orlandohohmeier @pierlo-upitup. If it needs it, lets create a follow up PR for that.

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -4985,9 +4985,9 @@
       "resolved": "https://registry.npmjs.org/react-transform-hmr/-/react-transform-hmr-1.0.4.tgz"
     },
     "reactjs-components": {
-      "version": "0.14.0-beta.29",
-      "from": "reactjs-components@0.14.0-beta.29",
-      "resolved": "https://registry.npmjs.org/reactjs-components/-/reactjs-components-0.14.0-beta.29.tgz"
+      "version": "0.14.0-beta.30",
+      "from": "reactjs-components@0.14.0-beta.30",
+      "resolved": "https://registry.npmjs.org/reactjs-components/-/reactjs-components-0.14.0-beta.30.tgz"
     },
     "reactjs-mixin": {
       "version": "0.0.2",
--- a/src/js/components/SchemaForm.js
+++ b/src/js/components/SchemaForm.js
@@ -107,7 +107,17 @@ class SchemaForm extends mixin(StoreMixin) {
       propID,
       newDefinition
     );
-    newDefinition.push(
+
+    let deleteButtonTop = Object.values(definition.itemShapes || {})
+      .some(function (itemShape) {
+        return itemShape.deleteButtonTop;
+      });
+    let arrayAction = 'push';
+
+    if (deleteButtonTop) {
+      arrayAction = 'unshift';
+    }
+    newDefinition[arrayAction](
       this.getRemoveRowButton(definition, prop, propID)
     );
 
--- a/src/js/components/ServiceForm.js
+++ b/src/js/components/ServiceForm.js
@@ -1,6 +1,7 @@
 import {Hooks} from 'PluginSDK';
 import React from 'react';
 
+import FormUtil from '../utils/FormUtil';
 import SchemaForm from './SchemaForm';
 import SchemaFormUtil from '../utils/SchemaFormUtil';
 import SchemaUtil from '../utils/SchemaUtil';
@@ -9,6 +10,10 @@ const METHODS_TO_BIND = [
   'handleFormChange',
   'validateForm'
 ];
+
+const FIELDS_TO_WATCH = {
+  healthChecks: ['protocol', 'portType']
+};
 
 class ServiceForm extends SchemaForm {
   constructor() {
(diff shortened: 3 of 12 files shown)


mesosphere-web:
Reviewer: @daiyiwhale

MatApple:
retest this please

src/js/utils/SchemaUtil.js
```diff
@@ -54,6 +54,10 @@ function schemaToFieldDefinition(fieldName, fieldProps, formParent, isRequired,
     definition.checked = value;
   }
 
+  if (fieldProps.fieldType === 'number') {
```
kennyt: nice. i was meaning to do this

src/js/schemas/service-schema/HealthChecks.js
```diff
@@ -0,0 +1,132 @@
...
+          .properties;
+
+        instanceDefinition.forEach(function (definition) {
+          let prop = definition.name;
+          if (FormUtil.isFieldInstanceOfProp('healthChecks', definition)) {
+            prop = FormUtil.getPropKey(definition.name);
+          }
+          if (properties[prop].shouldShow) {
+            definition.formElementClass = {
+              'hidden-form-element': !properties[prop].shouldShow(service)
```
kennyt: nice, this is an interesting way of doing it. i like it. one problem with this is that if you choose protocol HTTP, type in a path, then choose protocol COMMAND, the path value will still be there if you switch to JSON mode.

a solution to this could be to do cleanup of the unnecessary props when switching

npm-shrinkwrap.json
```diff
@@ -2521,6 +2521,633 @@
...
+          "from": "wrappy@1",
+          "resolved": "https://registry.npmjs.org/wrappy/-/wrappy-1.0.1.tgz"
+        },
+        "xtend": {
+          "version": "4.0.1",
+          "from": "xtend@^4.0.0",
+          "resolved": "https://registry.npmjs.org/xtend/-/xtend-4.0.1.tgz"
+        }
+      }
+    },
```
kennyt: should all of this be in here even though we're only bumping RJSC? maybe you can just manually change .29 to .30 and leave all of these out

src/js/schemas/service-schema/HealthChecks.js
MatApple: Yeah I chose this path because it had the least resistance. Since the compiled definition is passed around everywhere, it's nice to just mutate the props on it instead of slicing and splicing. I agree that we should have a cleanup in one of the Utils. I'll put up another PR for that.

kennyt: i agree, i like this implementation a lot better. just leaving a note for the extra props 👍

npm-shrinkwrap.json
MatApple: Yeah I need to clean this up.

src/js/components/SchemaForm.js
```diff
@@ -107,7 +107,17 @@ class SchemaForm extends mixin(StoreMixin) {
       propID,
       newDefinition
     );
-    newDefinition.push(
+
+    let deleteButtonTop = Object.values(definition.itemShapes || {})
```
kennyt: what if we put `deleteButtonTop` on the definition itself, rather than on one of it's itemShapes

src/js/schemas/service-schema/HealthChecks.js
MatApple: Thanks mate! 😃

src/js/components/ServiceForm.js
```diff
@@ -33,16 +38,39 @@ class ServiceForm extends SchemaForm {
     Hooks.doAction('serviceFormMount', this);
   }
 
-  handleFormChange() {
+  handleFormChange(model, eventObj) {
+    // Since the keys we're watching are of the
+    // form `prop[id].key`, extract the key
+    let propKey = FormUtil.getPropKey(eventObj.fieldName);
+
+    let shouldUpdateDefinition = Object.keys(model).some(function (changeKey) {
```
kennyt: before we do all of this, can we make sure that it's a change event? `handleFormChange` will handle all events atm like blur, focus, etc. we can do a check to make sure it's a change event to save us some performance

npm-shrinkwrap.json
MatApple: Done.

src/js/components/ServiceForm.js
MatApple: Good idea. Done.

src/js/components/SchemaForm.js
MatApple: I tried that but it seems to get stripped out or put on a different definition. Do you think this could have issues?

kennyt: nah, it's cool

MatApple:
@kennyt I decided to add the attribute pruning in this PR, so there is another commit after you gave your label.

src/js/utils/ServiceUtil.js
```diff
@@ -194,16 +213,35 @@ const ServiceUtil = {
...
       if (healthChecks != null && healthChecks.healthChecks != null) {
         definition.healthChecks = healthChecks.healthChecks
-          .map(function (healthCheck) {
-            if (healthCheck.portType == null) {
-              healthCheck.portType = 'PORT_INDEX';
+          .reduce(function (memo, healthCheck) {
+            // Only set defaults if user has changed a value in the form.
+            // I.e. user has intent to create a healthCheck.
+            let hasSetValue = Object.values(healthCheck).some(function (value) {
+              return value != null && value !== false;
```
MatApple: This covers default `null` values and any checkboxes set to `false`.

pierlo-upitup:
Do we need to doublecheck integration tests are passing after the last commit?

kennyt:
conflicts

src/js/utils/ServiceUtil.js
```diff
@@ -194,16 +213,35 @@ const ServiceUtil = {
...
+              }
+              if (healthCheck.protocol == null) {
+                healthCheck.protocol = 'HTTP';
+              }
             }
-            if (healthCheck.protocol == null) {
-              healthCheck.protocol = 'HTTP';
+            // Only include healthCheck if it contains values other
+            // than default values
+            if (!(Object.keys(healthCheck).length < 2
```
kennyt: can we use `hasSetValue` here instead? isn't that the same thing?

kennyt:
alright this looks good to me after my last comment

src/js/utils/ServiceUtil.js
MatApple: You're right, missed that. Done

kennyt:
tested it out and it works well!

rcorral:
@MatApple can you post an image of the integration tests?

rcorral:
@MatApple also...any idea why there's so many changes to the shrinkwrap file?

kennyt:
@rcorral there's only 3 line changes in shrinkwrap. it just hides the diff cause it's a huge file

rcorral:
Oh you're right. My brain is not with it today.

MatApple:
All but one test pass - same test fails on master though too.

Poltergeist:
The test was not failing on master. 

This is the ServiceVersions test on the branch `ph/fix/docker-volumes`
image

The problem seems to be that the `cmd` and `id` field are getting cleared on the general section in the version edit modal.

MatApple:
Thats super weird because I tested it against master and it failed. Maybe its the testing gods messing with us

PR #490, opened 2016-06-22
Fix schema form warnings

kennyt:
px

### DIFF
--- a/src/js/components/modals/__tests__/InstallPackageModal-test.js
+++ b/src/js/components/modals/__tests__/InstallPackageModal-test.js
@@ -1,9 +1,16 @@
 jest.dontMock('../InstallPackageModal');
 jest.dontMock('../../CosmosErrorMessage');
 jest.dontMock('../../ReviewConfig');
+jest.dontMock('../../SchemaForm');
 jest.dontMock('../../../stores/CosmosPackagesStore');
 jest.dontMock('../../../mixins/InternalStorageMixin');
 jest.dontMock('../../../mixins/TabsMixin');
+jest.dontMock('../../../utils/FormUtil');
+jest.dontMock('../../../utils/GeminiUtil');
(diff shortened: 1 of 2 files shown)


mesosphere-web:
Reviewer: @Poltergeist

src/js/components/modals/__tests__/InstallPackageModal-test.js
```diff
@@ -63,7 +70,7 @@ describe('InstallPackageModal', function () {
         this.instance.getModalContents(),
         this.container
       ));
-      var result = node.querySelectorAll('p')[1];
+      var result = node.querySelectorAll('p')[4];
```
jfurrow: Why this change? Just curious.

kennyt: this is because we're actually rendering the schema form now, which will itself take up some nodes. therefore we need to look deeper

src/js/utils/SchemaFormUtil.js
```diff
@@ -194,7 +194,8 @@ let SchemaFormUtil = {
 
   validateModelWithSchema(model, schema) {
     let result = tv4.validateMultiple(model, schema);
-    if (result.valid) {
+
+    if (result == null || result.valid) {
```
kennyt: we need to do the `== null` check for tests. it has no significance for the app because tv4 will always return something

src/js/components/modals/__tests__/InstallPackageModal-test.js
jfurrow: Ahh cool. Thanks

MatApple:
I think there are more tests that need fixing too right? I'm seeing `DeploymentsTab-test`, `ServiceDetailTaskTab-test` as well.

Poltergeist:
Looks good lets :ship: this.

PR #494, opened 2016-06-23
Refactor the task detail tab configuration

orlandohohmeier:
Use the route definition (`Route.prototype.children`) instead of configuring a set of tabs that must match the corresponding route configurations. This prevents configuration errors and enables us to
reuse the task detail in more places (e.g. jobs)

### DIFF
--- a/src/js/pages/task-details/TaskDetail.js
+++ b/src/js/pages/task-details/TaskDetail.js
@@ -15,26 +15,6 @@ import TabsMixin from '../../mixins/TabsMixin';
 import TaskDirectoryStore from '../../stores/TaskDirectoryStore';
 import TaskStates from '../../constants/TaskStates';
 
-const SERVICES_TABS = {
-  'services-task-details-tab': 'Details',
-  'services-task-details-files': 'Files',
-  'services-task-details-logs': 'Logs',
-  'services-task-details-volumes': 'Volumes'
-};
-
-const NODES_TABS = {
-  'nodes-task-details-tab': 'Details',
-  'nodes-task-details-files': 'Files',
-  'nodes-task-details-logs': 'Logs',
-  'nodes-task-details-volumes': 'Volumes'
-};
-
-const VIRTUAL_NETWORKS_TABS = {
-  'virtual-networks-tab-detail-tasks-details-tab': 'Details',
-  'virtual-networks-tab-detail-tasks-details-files': 'Files',
-  'virtual-networks-tab-detail-tasks-details-logs': 'Logs'
-};
-
 const METHODS_TO_BIND = [
   'onTaskDirectoryStoreError',
   'onTaskDirectoryStoreSuccess'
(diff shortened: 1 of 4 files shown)


mesosphere-web:
Reviewer: @pierlo-upitup

src/js/pages/task-details/TaskDetail.js
```diff
@@ -67,14 +47,26 @@ class TaskDetail extends mixin(InternalStorageMixin, TabsMixin, StoreMixin) {
...
   componentWillMount() {
     super.componentWillMount(...arguments);
-    this.tabs_tabs = Object.assign({}, SERVICES_TABS);
-    if (this.props.params.nodeID != null) {
-      this.tabs_tabs = Object.assign({}, NODES_TABS);
-    }
-    if (this.props.params.overlayName != null) {
-      this.tabs_tabs = Object.assign({}, VIRTUAL_NETWORKS_TABS);
+
+    let routes = this.context.router.getCurrentRoutes();
```
aldipower: Could we use the functional approach? Better to read and understand imho.

```JavaScript
let taskDetailRoute = routes.find(route => route.handler === TaskDetail);

if (taskDetailRoute != null) {
  this.tabs_tabs = taskDetailRoute.children.reduce(
    ...
  );
  ...
}
```

orlandohohmeier: Yeah, good idea!

src/js/routes/factories/network.js
```diff
@@ -111,7 +111,8 @@ let RouteFactory = {
                 parentCrumb: 'virtual-networks-tab-detail-tasks-detail',
                 getCrumbs: function () { return []; }
               }
-            }
+            },
+            title: 'Details'
```
pierlo-upitup: I'm not opposing this, however it really feels like we're overloading our route configuration objects.

aldipower: Thought the same..

aldipower: The title is actually 'content'. That doesn't belong into a definition imo. But just my 2€¢.

pierlo-upitup: Yeah, maybe we should derive the title from a constant object that maps the route name to the tab's title. Just a thought.

orlandohohmeier: Problem with that is, that we end with various place, which one has to update only to add a new tab. Thats very error prone.

pierlo-upitup: That is also true.

orlandohohmeier:
Let me note that I'm thinking about moving the "dynamic" tab configuration into the tabs mixing. What do you think? Is it worth it, or should we just move forward with this one for now?

aldipower:
Ah ok, imo it's worth to do it. :)

pierlo-upitup:
@orlandohohmeier in principle it's a good idea, but I vote for postponing the mixin change for now.

orlandohohmeier:
Ok, I'll create an issue and add a todo so that we don't lose track.

src/js/pages/task-details/TaskDetail.js
orlandohohmeier: ✔️  Done

pierlo-upitup:
All tests passing locally, minus some unrelated test already fixed in another PR.

PR #500, opened 2016-06-23
[1 of 4: Icons] Remove high-specificity icon styles

jfurrow:
Previously we gave ALL `.icon` elements inside `table`s a margin, and the CSS was very specific.

I've introduced new classes, `.icon-margin-right` and `.icon-margin-left` and applied them where we relied on the general class for margins.


mesosphere-web:
Reviewer: @Poltergeist

kennyt:
works great

PR #508, opened 2016-06-23
Adds Models for Marathon Info and the clusters cryptographic id.

rcorral:


mesosphere-web:
Reviewer: @kennyt

src/js/events/ConfigActions.js
```diff
@@ -22,6 +21,24 @@ const ConfigActions = {
...
+      success: function (response) {
+        AppDispatcher.handleServerAction({
+          type: ActionTypes.REQUEST_CLUSTER_CCID_SUCCESS,
+          data: response
+        });
+      },
+      error: function (e) {
+        AppDispatcher.handleServerAction({
+          type: ActionTypes.REQUEST_CLUSTER_CCID_ERROR,
+          data: e.message
```
kennyt: do we need to do the whole `RequestUtil.getErrorFromXHR(xhr)` dance here? or is this cool?

rcorral: I removed this entirely because I don't think there will be an error response.

PR #519, opened 2016-06-24
Fix docker volumes

Poltergeist:
This also fixes the problem of adding a new row at the top of the form if no row was present.
 
image

image

### DIFF
--- a/src/js/components/SchemaForm.js
+++ b/src/js/components/SchemaForm.js
@@ -136,10 +136,10 @@ class SchemaForm extends mixin(StoreMixin) {
     }
 
     return (
-      <div className="form-row-element">
+      <div className="form-row-element"
+        prop={prop}>
         <a
           key={`${prop}-add-new-row`}
-          prop={prop}
           className="clickable row"
           onClick={
             this.handleAddRow.bind(this, prop, generalDefinition, definition)
--- a/src/js/utils/ServiceUtil.js
+++ b/src/js/utils/ServiceUtil.js
@@ -123,16 +123,17 @@ const ServiceUtil = {
           if (volumes.dockerVolumes) {
             volumesList = volumesList.concat(
               volumes.dockerVolumes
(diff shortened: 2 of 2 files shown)


mesosphere-web:
Reviewer: @philipnrmn

src/js/utils/ServiceUtil.js
```diff
@@ -123,6 +123,9 @@ const ServiceUtil = {
           if (volumes.dockerVolumes) {
             volumesList = volumesList.concat(
               volumes.dockerVolumes
+                .filter(function ({containerPath, hostPath}) {
```
kennyt: should this be indented at the same level as `map` below?

kennyt:
not 100% how this works, but if i fill all the fields out, switch to JSON then switch back container volumes is empty

px

src/js/components/SchemaForm.js
```diff
@@ -136,10 +136,10 @@ class SchemaForm extends mixin(StoreMixin) {
     }
 
     return (
-      <div className="form-row-element">
+      <div className="form-row-element"
+        prop={prop}>
```
kennyt: can you say something about what this is used for? not 100% clear to me atm

rcorral:
Guys, I'm going to merge this in so that we can provide a build. But @Poltergeist could you please take a look at the comments Kenny left?

kennyt:
sounds good

rcorral:
integration tests pass.

src/js/components/SchemaForm.js
Poltergeist: This is needed to find the exact place to add a new row. If this is not present the fields get added at the beginning of the form rather then above the button. This is only needed in the case if there isn't any row left.

Poltergeist:
The docker volume gets cleared if no docker image is present. Ideal would be to only show those fields if a docker volume is set.

PR #523, opened 2016-06-24
Use x icon for ServiceTable label filter instead of 'x' character

jfurrow:
Before:

After:


mesosphere-web:
Reviewer: @pierlo-upitup

PR #525, opened 2016-06-24
Allow duplicable rows to have titles

kennyt:
px

This also makes it so that the `Add New X` button isn't completely flush against the bottom


mesosphere-web:
Reviewer: @MatApple

PR #526, opened 2016-06-24
Clicking on the magnifying glass links to logs again

kennyt:
px

### DIFF
--- a/src/js/pages/task-details/TaskDetail.js
+++ b/src/js/pages/task-details/TaskDetail.js
@@ -139,10 +139,18 @@ class TaskDetail extends mixin(InternalStorageMixin, TabsMixin, StoreMixin) {
   }
 
   handleOpenLogClick(selectedLogFile) {
-    this.setState({selectedLogFile, currentTab: 'debug'});
+    let params = Object.assign(
+      {},
+      this.props.params,
+      {filePath: encodeURIComponent(selectedLogFile.get('path'))}
+    );
+    let currentRoutes = this.context.router.getCurrentRoutes();
+    let {logRouteName} = currentRoutes[currentRoutes.length - 1];
+    this.context.router.transitionTo(logRouteName, params);
   }
 
   getBasicInfo() {
+    let {selectedLogFile} = this.state;
     let task = MesosStateStore.getTaskFromTaskID(this.props.params.taskID);
 
     if (task == null) {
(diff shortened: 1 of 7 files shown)


mesosphere-web:
Reviewer: @jfurrow

rcorral:
@kennyt found one small issue, the dropdown doesn't populate as it should, before it would show you in the list all of the *.log files that were in the directory that you came from. Should be an easy fix.
![gif](http://cl.ly/0A3n2N17102k/Screen%20Recording%202016-06-25%20at%2012.37%20PM.gif0

src/js/pages/task-details/TaskLogsTab.js
```diff
@@ -98,12 +99,17 @@ class TaskLogsTab extends React.Component {
...
   handleViewChange(currentFile) {
-    this.setState({
-      currentFile,
-      watching: 0,
-      searchString: '',
-      totalFound: 0
-    });
+    let currentRoutes = this.context.router.getCurrentRoutes();
+    let lastRoute = currentRoutes[currentRoutes.length - 1];
+    console.log('in here', lastRoute);
```
MatApple: 🍷

src/js/pages/task-details/TaskDetail.js
```diff
@@ -139,10 +139,18 @@ class TaskDetail extends mixin(InternalStorageMixin, TabsMixin, StoreMixin) {
...
 
   handleOpenLogClick(selectedLogFile) {
-    this.setState({selectedLogFile, currentTab: 'debug'});
+    let params = Object.assign(
+      {},
+      this.props.params,
+      {filePath: encodeURIComponent(selectedLogFile.get('path'))}
+    );
+    let currentRoutes = this.context.router.getCurrentRoutes();
+    let {logRouteName} = currentRoutes[currentRoutes.length - 1];
```
orlandohohmeier: How about `let {logRouteName} = currentRoutes.pop()` which is a little less error prone.

kennyt:
@rcorral i've been trying to look into that problem and i just realized that it's also on master. it's a really weird problem where `TaskDetail` re-mounts when you change tabs so then it fetches the directory over again and blows away the old one.

this problem has probably been here since we made each of the tabs routed

src/js/pages/task-details/TaskDetail.js
rcorral: Not sure if this will cause a mutation.

rcorral:
@kennyt alright, let's merge this in. Could you take a look at a fix for the issue though?

src/js/pages/task-details/TaskLogsTab.js
rcorral: Let's address this.

src/js/pages/task-details/TaskLogsTab.js
```diff
@@ -188,6 +194,7 @@ class TaskLogsTab extends React.Component {
   }
 
   onItemSelection(obj) {
+    console.log(obj);
```
rcorral: And this.

src/js/pages/task-details/TaskDetail.js
orlandohohmeier: Good, point. Let's keep it as it is.

PR #527, opened 2016-06-24
Improve ServiceTable responsiveness, v2

jfurrow:
Here's another stab. I spoke with ashenden and we decided to display the total number of tasks in the tooltip at smaller viewport widths.

This also handles much larger numbers of tasks.


mesosphere-web:
Reviewer: @kennyt

PR #533, opened 2016-06-26
DCOS-7920: Use Webpack import instead of relying on referencing location

jfurrow:
This will prevent a bug where the path to the image might be incorrect.


mesosphere-web:
Reviewer: @pierlo-upitup

orlandohohmeier:
Looks good to me! 👍  I wonder if we should do the same for the rest of the icons/images and update all the other components as well. Nothing we need to do now, but should consider doing in the next few weeks. Please create an issue to track this.

Poltergeist:
This is awesome and works.

PR #534, opened 2016-06-27
Fix user dropup menu width inconsistency

jfurrow:
The width of the user dropup menu was not matching the sidebar with. This fixes that.

As an aside, the way we are setting the width of the sidebar for mobile devices is really strange to me (the negative margin value which matches the positive padding feels very hacky) and I'd like to simplify it. I created a tech-debt task for that.


mesosphere-web:
Reviewer: @mlunoe

PR #535, opened 2016-06-27
DCOS-7919: Sort TaskTable descending by default, newest tasks at top

jfurrow:
Unfortunately there is also a bug in the `Table` component, so `reactjs-components` needs a new release before this property will actually take effect. https://github.com/mesosphere/reactjs-components/pull/294


mesosphere-web:
Reviewer: @MatApple

Poltergeist:
Code looks good to me. Let's get the PR on reactjs components to a mergeble state. Before merging this.

kennyt:
i don't believe the hold merge applies anymore

PR #541, opened 2016-06-27
Fix parsing of service to form in networking section

Poltergeist:
Tests are passing.

### DIFF
--- a/src/js/schemas/service-schema/Networking.js
+++ b/src/js/schemas/service-schema/Networking.js
@@ -39,9 +39,11 @@ const Networking = {
 
         return portMappings.map(function (portMapping) {
           return {
-            lbPort: portMapping.hostPort || portMapping.containerPort,
+            lbPort: portMapping.hostPort || portMapping.containerPort ||
+              portMapping.port,
             name: portMapping.name,
-            protocol: portMapping.protocol
+            protocol: portMapping.protocol,
+            discovery: portMapping.port > 0
           };
         });
       },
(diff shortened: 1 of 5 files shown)


philipnrmn:
The discovery transform looks good. 

- [ ] We should also persist `networkType` to get the dropdown working. 
- [ ] When a docker image is set in bridge or user mode, we should interpret `docker.container.portMapings` rather than `portDefinitions'.

Poltergeist:
- [x] Removed the Network drop down from the container section.
- [x] Moved the network logic to networking section.

src/js/schemas/service-schema/Networking.js
```diff
@@ -9,11 +9,15 @@ const Networking = {
       title: 'Network Type',
       options: [
         {html: 'Host (Default)', id: 'host'},
-        'Bridge',
-        {html: 'Virtual Network: Dev', id: 'dev'},
+        {html: 'Bridge', id: 'bridge'},
+        {html: 'Virtual Network: Dev', id: 'user'},
```
philipnrmn: What's the logic behind this rename?

Poltergeist: Right I'll change it back. I introduced this by going down another route. My bad.

src/js/schemas/service-schema/Networking.js
```diff
@@ -39,9 +43,12 @@ const Networking = {
 
         return portMappings.map(function (portMapping) {
           return {
-            lbPort: portMapping.hostPort || portMapping.containerPort,
+            lbPort: portMapping.hostPort || portMapping.containerPort ||
```
aldipower: Not relevant for this PR, because it is working in this case, but this should have proper `null`-evaluation in the future, because a port `0` is treated as `false` in the logic. Could possibly lead to errors if a refactoring will be done at another place in the code.

PR #546, opened 2016-06-27
Allow checkbox labels to be rendered

jfurrow:
This PR enables `SchemaForm` checkboxes to be rendered with a label inside the `<label>` element. 

The properties that are changed may be a little confusing, but `SchemaUtil` previously would override the `showLabel` property on all `SchemaForm`s, which is what gets displayed on top of a checkbox, whereas the `label` property renders within the `<label>` element.

Before:

After:

### DIFF
--- a/src/js/schemas/service-schema/ContainerSettings.js
+++ b/src/js/schemas/service-schema/ContainerSettings.js
@@ -28,7 +28,6 @@ const ContainerSettings = {
       type: 'group',
       properties: {
         privileged: {
-          title: 'Extend runtime privileges',
           label: 'Extend runtime privileges',
           showLabel: false,
           type: 'boolean',
@@ -43,7 +42,6 @@ const ContainerSettings = {
           }
         },
         forcePullImage: {
-          title: 'Force pull image on launch',
           label: 'Force pull image on launch',
           showLabel: false,
           type: 'boolean',
--- a/src/js/schemas/service-schema/ContainerSettings.js
+++ b/src/js/schemas/service-schema/ContainerSettings.js
@@ -29,6 +29,8 @@ const ContainerSettings = {
       properties: {
         privileged: {
           title: 'Extend runtime privileges',
+          label: 'Extend runtime privileges',
+          showLabel: false,
           type: 'boolean',
           getter: function (service) {
             let container = service.getContainerSettings();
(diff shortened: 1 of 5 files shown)


mesosphere-web:
Reviewer: @kennyt

src/js/utils/SchemaUtil.js
```diff
@@ -18,23 +18,29 @@ function getValueFromSchemaProperty(fieldProps) {
   return value;
 }
 
-function getLabelFromSchemaProperty(fieldName, fieldProps, isRequired, renderLabel) {
+function getLabelFromSchemaProperty(fieldName, fieldProps, isRequired, renderLabel, definition) {
```
kennyt: can we change the name of this? it's no longer getting label, but setting label

src/styles/components/multiple-form.less
```diff
@@ -92,6 +92,21 @@ base/forms.less
 
   .multiple-form-modal {
 
+    .form-row-element {
+
+      &.checkbox {
+        &:extend(.align-self-flex-end);
+
+        &.column-1 {
```
kennyt: why do we need this?

src/js/schemas/service-schema/ContainerSettings.js
```diff
@@ -29,6 +29,8 @@ const ContainerSettings = {
       properties: {
         privileged: {
           title: 'Extend runtime privileges',
```
kennyt: do we still need the titles?

src/js/utils/SchemaUtil.js
jfurrow: Good call, done!

src/styles/components/multiple-form.less
jfurrow: We need this because if a checkbox is in a row by itself, we don't want it to be aligned to the end of the flex container.

jfurrow: Alright so I actually changed this so that it doesn't rely on the `column-1` class being present.

src/js/schemas/service-schema/ContainerSettings.js
jfurrow: Nope, removed! Thanks

jfurrow:
retest this please

PR #551, opened 2016-06-27
Update notification count on change

kennyt:
px

### DIFF
--- a/src/js/stores/DCOSStore.js
+++ b/src/js/stores/DCOSStore.js
@@ -129,11 +129,18 @@ class DCOSStore extends EventEmitter {
     let deploymentsList = MarathonStore.get('deployments');
     let serviceTree = MarathonStore.get('groups');
 
-    NotificationStore.addNotification(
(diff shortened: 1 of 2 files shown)


mesosphere-web:
Reviewer: @philipnrmn

src/js/stores/DCOSStore.js
```diff
@@ -129,12 +129,19 @@ class DCOSStore extends EventEmitter {
...
-    NotificationStore.addNotification(
-      'services-deployments',
-      'deployment-count',
-      deploymentsList.getItems().length
+    let deploymentListlength = deploymentsList.getItems().length;
+    let currentDeploymentCount = NotificationStore.getNotificationCount(
+      'services-deployments'
     );
 
+    if (deploymentListlength !== currentDeploymentCount) {
```
kennyt: we do this in order to not be constantly be emitting change events. this will prevent the complicated services logic to constantly re-render

PR #553, opened 2016-06-27
Update reactjs-components & affected components

jfurrow:
This updates `reactjs-components` and changes all implementations of the `Modal` as needed.

I removed a bunch of custom CSS for the login form, which was but there to remove padding from the outer container and add it to the inner container. The new `Modal` component does this out of the box.


mesosphere-web:
Reviewer: @orlandohohmeier

PR #555, opened 2016-06-27
Prevent page-navigation element from shrinking

jfurrow:
Explicitly set flex properties of `page-navigation`.


mesosphere-web:
Reviewer: @MatApple

kennyt:
accidentally committed to this branch, but it's also a small change. it stops capitalizing the location in the deployments tab

px

PR #557, opened 2016-06-27
Fix incorrect prop of micro vs. tiny for icons

jfurrow:
Removing any concept of `micro` in favor of `tiny` :)


mesosphere-web:
Reviewer: @kennyt

src/styles/components/icons.less
```diff
@@ -181,11 +181,6 @@ Icon (huge)
 
   }
 
-  .icon.icon-micro {
-    max-height: @icon-micro-height;
-    max-width: @icon-micro-width;
```
kennyt: should we also remove this variables or nah?

jfurrow: Oh yeah, good call!

PR #558, opened 2016-06-27
Listens to mesos state store success event to re-render UI after failed requests.

rcorral:


mesosphere-web:
Reviewer: @daiyiwhale

src/js/components/TaskView.js
```diff
@@ -42,23 +52,13 @@ class TaskView extends mixin(SaveStateMixin) {
...
 
-  componentWillUnmount() {
-    super.componentWillUnmount();
-
-    MesosStateStore.removeChangeListener(
-      EventTypes.MESOS_STATE_REQUEST_ERROR,
-      this.onMesosStateRequestError
-    );
+  onStateStoreSuccess() {
+    this.setState({mesosStateErrorCount: 0});
```
kennyt: let's do this only if `mesosStateErrorCount` is currently not `0`

rcorral: That's a good catch. Done.

rcorral: It would rerender. Supressupdate, just means don't call `forceUpdate` in the storemixin

src/js/components/TaskView.js
```diff
@@ -1,8 +1,8 @@
 import classNames from 'classnames';
 import mixin from 'reactjs-mixin';
+import {StoreMixin} from 'mesosphere-shared-reactjs';
```
jfurrow: :abcd:

jfurrow:
Other than the #nit it LGTM

PR #561, opened 2016-06-28
Fix hiding of LB Port on required conditions

MatApple:
Hide LB Port when not in host mode and when discovery is not checked.


mesosphere-web:
Reviewer: @Poltergeist

rcorral:
@philipnrmn please check this out when you get a chance!

src/js/components/ServiceForm.js
```diff
@@ -39,24 +53,47 @@ class ServiceForm extends SchemaForm {
...
+              nestedModel[FIELD_TO_TABS[key]][key] = combinedModel[key];
+            }
+
+            return;
+          }
+          if (combinedModel[key] != null) {
+            nestedModel[key] = combinedModel[key];
+          }
+        });
+        console.log(nestedModel.networking)
```
philipnrmn: Accidental `console.log`

MatApple: Removed!

philipnrmn:
I'm sorry, I've been too terse and not communicated the requirement properly. I posted this:
> hiding the LB Port field when in host mode with the discover field off

What I meant was that the LB Port field should only be hidden in host mode with the discover field off. So:

 - in User and Bridge mode the LB Port field should always be visible
 - in Host mode the LB Port field should be visible iff the discovery box is checked

MatApple:
Hey @philipnrmn, thanks for the feedback. I've updated to work as you describe.

philipnrmn:
Almost there! the LB Port field is currently not showing for me in User mode. It should always be visible in User mode.

MatApple:
@philipnrmn I don't even see an option in the `networkType` select field for User mode. Where can I find this?

The schema currently only has these options: 
```
options: [
        {html: 'Host (Default)', id: 'host'},
        {html: 'Bridge', id: 'bridge'},
        {html: 'Virtual Network: Dev', id: 'dev'},
        {html: 'Virtual Network: Prod', id: 'prod'}
      ],
```

kennyt:
conflicts

src/js/components/SchemaForm.js
```diff
@@ -174,36 +177,28 @@ class SchemaForm extends mixin(StoreMixin, InternalStorageMixin) {
   }
 
   getRemoveRowButton(generalDefinition, prop, id, title = null) {
+    let deleteButton = (
+      <div key={`${prop}${id}-remove`} className="align-self-flex-end">
```
kennyt: the reason i didn't do it this way was because this `div` needs the class `form-row-element` if it's by itself and doesn't need it if it's within the title

pierlo-upitup:
@MatApple Virtual Networks are expressed as `networkType: "USER"` – that is what @philipnrmn is referring to in his previous comment.

src/js/components/ServiceForm.js
```diff
@@ -293,7 +293,7 @@ class ServiceForm extends SchemaForm {
       if (!['host', 'bridge'].includes(model.networking.networkType)) {
         // Set expose to true as default
         model.networking.ports.map(function (port) {
-          port.expose = true;
```
MatApple: @Poltergeist currently setting this to false will filter this port mapping from the app definition that we send.

kennyt:
conflicts

orlandohohmeier:
networking duplicable rows
The port fields get rendered in the wrong place...

MatApple:
@orlandohohmeier pulling the latest changes I can't reproduce this

orlandohohmeier:
@MatApple I just pulled the very last changes and tested it... still broken.

MatApple:
@orlandohohmeier Fixed it - didn't pay close enough attention to your GIF first time round 😉

PR #564, opened 2016-06-28
Fetch service network options

orlandohohmeier:
network-type

**AC**
* [x] Fetch the virtual network overlays and provide them as options in the deploy service modal 
* [x] Adjust the network type getter to properly set the default

### DIFF
--- a/src/js/components/ServiceForm.js
+++ b/src/js/components/ServiceForm.js
@@ -5,9 +5,11 @@ import FormUtil from '../utils/FormUtil';
 import SchemaForm from './SchemaForm';
 import SchemaFormUtil from '../utils/SchemaFormUtil';
 import SchemaUtil from '../utils/SchemaUtil';
+import VirtualNetworksStore from '../stores/VirtualNetworksStore';
 
 const METHODS_TO_BIND = [
   'handleFormChange',
+  'onVirtualNetworksStoreSuccess',
   'validateForm'
 ];
 
@@ -23,9 +25,9 @@ class ServiceForm extends SchemaForm {
       this[method] = this[method].bind(this);
     });
 
-    this.store_listeners = Hooks.applyFilter(
-      'serviceFormStoreListeners', []
-    );
+    this.store_listeners = Hooks.applyFilter('serviceFormStoreListeners', [
+      {name: 'virtualNetworks', events: ['success']}
+    ]);
   }
 
   componentWillMount() {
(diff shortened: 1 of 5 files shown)


mesosphere-web:
Reviewer: @MatApple

src/js/components/ServiceForm.js
```diff
@@ -68,6 +70,26 @@ class ServiceForm extends SchemaForm {
...
+    let {networking} = this.multipleDefinition;
+    let virtualNetworks = VirtualNetworksStore.getOverlays()
+      .mapItems(function (overlay) {
+        let name = overlay.getName();
+
+        return {html: `Virtual Network: ${name}`, id: name}
+      }).getItems();
+
+    // networking.definition[0].options = [].concat(networkType.options,
+    // virtualNetworks);
```
pierlo-upitup: shouldnt this be removed?

src/js/components/ServiceForm.js
```diff
@@ -68,6 +70,26 @@ class ServiceForm extends SchemaForm {
...
+    networking.definition.forEach(function (definition) {
+      if (definition.name === 'networkType') {
+        definition.options = [].concat(networkType.options, virtualNetworks);
+      }
+      return definition;
```
pierlo-upitup: who's receiving these?

src/js/utils/ServiceUtil.js
```diff
@@ -324,7 +324,7 @@ const ServiceUtil = {
             definition.container.docker.network = 'BRIDGE';
           } else {
             definition.container.docker.network = 'USER';
-            definition.ipAddress = {networkName: 'd-overlay-1'};
+            definition.ipAddress = {networkName: networkType};
```
pierlo-upitup: can't we stick to calling this `networkName`?

src/js/utils/__tests__/ServiceUtil-test.js
```diff
@@ -322,9 +322,9 @@ describe('ServiceUtil', function () {
         it('adds the networkName field to the service', function () {
           let service = ServiceUtil.createServiceFromFormModel({
             containerSettings: { image: 'redis' },
-            networking: { networkType: 'user', ports: [{}] },
+            networking: { networkType: 'prod', ports: [{}] },
```
pierlo-upitup: wait – isnt `network type` still supposed to be USER?

rcorral:
Code looks good to me 👍

src/js/utils/__tests__/ServiceUtil-test.js
orlandohohmeier: I've to admit it's a little confusing and the prop name is anything but good (that's a different story and not part of this PR)... The network type is used to define whether service container network should be `HOST`, `BRIDGE`, or `USER` but it is also used to define the service `ipAddress.networkName`. Thus we assign the selected option to the `networkType` property and use that prop to determine the respective values. That's why we expect `service.ipAddress.networkName` to equal `networkType` and the reason why I adjusted the values.

src/js/utils/ServiceUtil.js
orlandohohmeier: See https://github.com/dcos/dcos-ui/pull/564#discussion_r68847355 for details.

PR #567, opened 2016-06-28
Don't use abbreviations for docker volumes

kennyt:
It was abbreviated before because there wasn't enough space. there is enough space now due to RJSC changes not counting JSX elements into width calculation

# Before
px

# After
px


mesosphere-web:
Reviewer: @jfurrow

Poltergeist:
LGTM 🎆 🚢 💯

PR #570, opened 2016-06-28
Add option for title between duplicable rows

MatApple:

### DIFF
--- a/src/js/components/SchemaForm.js
+++ b/src/js/components/SchemaForm.js
@@ -172,28 +172,36 @@ class SchemaForm extends mixin(StoreMixin, InternalStorageMixin) {
   }
 
   getRemoveRowButton(generalDefinition, prop, id, title = null) {
-    let deleteButton = (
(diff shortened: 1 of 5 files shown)


mesosphere-web:
Reviewer: @mlunoe

src/js/components/SchemaForm.js
```diff
@@ -161,16 +171,31 @@ class SchemaForm extends mixin(StoreMixin) {
...
         </button>
       </div>
     );
+
+    if (!title) {
+      return deleteButton;
+    }
+
+    return (
+      <div className="form-row-element duplicable-row-title-wrapper">
```
kennyt: this thing needs a key. like what we do on line 176

kennyt:
after my last comment, :ship:

src/js/components/SchemaForm.js
kennyt: done

PR #587, opened 2016-06-29
Listen for Job events in JobsTab

jfurrow:
We weren't listening for these events previously, so the table didn't update when a new job was added, nor when the job list was fetched.


mesosphere-web:
Reviewer: @aldipower

PR #590, opened 2016-06-29
Ensure ports are only added if either name or lbPort is set

Poltergeist:
image
Before when the user opened the deploy service modal and toggled to the JSON mode one port was set by default. This fix makes sure that only ports with a set name or lbPort are added.

### DIFF
--- a/src/js/utils/ServiceUtil.js
+++ b/src/js/utils/ServiceUtil.js
@@ -266,7 +266,7 @@ const ServiceUtil = {
         let networkType = networking.networkType;
         if (networking.ports != null) {
           networking.ports = networking.ports.filter(function (port) {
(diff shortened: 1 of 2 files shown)


mesosphere-web:
Reviewer: @mlunoe

rcorral:
Good catch!

orlandohohmeier:
The code looks to me. Please adjust the respective test cases as well, as the checks are failing.

pierlo-upitup:
Not too sure we actually want this. Let's keep it open and revisit this later.

pierlo-upitup:
@rcorral suggests we need this.

MatApple:
@Poltergeist @aldipower Are you guys able to fix the tests here so we can merge? Thanks guys! Nice work

PR #591, opened 2016-06-29
DCOS-8138 & DCOS-8139: volumes error messages while editing an application

Poltergeist:
This adds the nice error message from marathon which is provided if the user adds a network or local volume to an already existing application.

image

This is currently not possible in marathon. It is only possible to add volumes to a new service. In this PR the automatic addition of updateStrategy and residency is prevented while editing a configuration.


mesosphere-web:
Reviewer: @philipnrmn

rcorral:
@Poltergeist thanks for the hard work on this!
Code looks good!

PR #593, opened 2016-06-29
Fix the agent IP path for tasks

kennyt:
they didn't want it on the task detail, but really wanted the information shown on the virtual networks tasks table. the agent ip column was always `N/A` even though the data was there. this fixes that

px


mesosphere-web:
Reviewer: @pierlo-upitup

PR #596, opened 2016-06-30
Fix task finished text color

jfurrow:
Verified with @cleverroffey, this is the intended text color.

Before:

After:


mesosphere-web:
Reviewer: @daiyiwhale

kennyt:
should we also remove the class application to JSX?

jfurrow:
The class is applied generically (`task-status-${taskStatus}`) so I don't think it's worth removing only for this status.

daiyiwhale:
nice :D

PR #604, opened 2016-06-30
Revert "Revert "Ensure ports are only added if either name or lbPort is set""

rcorral:
Reverts dcos/dcos-ui#603


mesosphere-web:
Reviewer: @pierlo-upitup

PR #614, opened 2016-07-01
[2 of 2] DCOS-8142: Change LB Port fieldtype to number when enabled

jfurrow:
I verified this works in Chrome, Safari, Firefox, IE11 and Edge.

### DIFF
--- a/src/js/utils/SchemaUtil.js
+++ b/src/js/utils/SchemaUtil.js
@@ -61,7 +61,7 @@ function schemaToFieldDefinition(fieldName, fieldProps, formParent, isRequired, 
     definition.checked = value;
   }
 
-  if (fieldProps.fieldType === 'number') {
+  if (fieldProps.type === 'number') {
     definition.fieldType = 'number';
   }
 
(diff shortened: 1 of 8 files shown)


mesosphere-web:
Reviewer: @orlandohohmeier

src/js/utils/SchemaUtil.js
```diff
@@ -61,7 +61,7 @@ function schemaToFieldDefinition(fieldName, fieldProps, formParent, isRequired,
     definition.checked = value;
   }
 
-  if (fieldProps.fieldType === 'number') {
+  if (fieldProps.type === 'number') {
```
MatApple: I feel like this could have consequences elsewhere. We should check all of our schemas that are rendered into a SchemaForm to make sure they have the correct type set.

MatApple: @jfurrow ^^

jfurrow: I'm not sure what else needs to be changed, can you clarify? I believe @kennyt was saying that the schemas provided to the Install Modal already use `type` instead of `fieldType`.

MatApple: Just want to make sure all the schemas under `/service-schemas/**` adhere to this :)

jfurrow: Thanks guys! I searched the codebase and it looks like all `fieldType` keys are meant to be there (ie. they belong to an actual definition, not schema). Good call!

PR #615, opened 2016-07-01
Hide service breadcrumb on root view

daiyiwhale:
Remove redundant breadcrumb title on root `/services` view


mesosphere-web:
Reviewer: @pierlo-upitup

daiyiwhale:
Actually we want to show that breadcrumb on root view, but only hide if there are no services.

PR #617, opened 2016-07-01
[1 of 2] DCOS-8059: Change type key to fieldType

jfurrow:


mesosphere-web:
Reviewer: @MatApple

kennyt:
i know this is kinda messed up, but should we change the functionality to make it use `type` instead? this way, the Advanced Install modal will be able to take advantage of it with the universe package schemas

jfurrow:
So you're suggesting we make this change to reactjs-components? I'm fine with that, just want to make sure I understand what you're suggesting.

kennyt:
you'd make the change [here](https://github.com/dcos/dcos-ui/blob/fc02902d54660fdb80414c4ded71584dce833a8a/src/js/utils/SchemaUtil.js#L64)

similar to how the boolean does it a couple lines down. instead of looking in `fieldProps.fieldType`, it would look at `fieldProps.type`. this will allow us to build forms from legit schemas and use number fields.

what you would have to do is go through all the `service-schema`s and `job-schema`s and change all of the `fieldType: 'number'` to `type: 'number'`

jfurrow:
@kennyt Makes sense! I'll do that.

src/js/components/modals/AddRepositoryFormModal.js
```diff
@@ -91,7 +91,7 @@ class AddRepositoryFormModal extends mixin(StoreMixin) {
         value: ''
       },
       {
-        fieldType: 'number',
```
kennyt: you don't need this change :) this is a form definition

jfurrow: Good catch! Fixed!

MatApple:
Haha left the same comment as Kenny on Part 2 of this. Once we've verified all the schemas adhere to this change then lets 🚢 . Do we need the hold merge label till then?

MatApple:
@jfurrow ^^

jfurrow:
@MatApple That discussion is continued here: https://github.com/dcos/dcos-ui/pull/614/files/a2404207b20927c8fad068302ccff13c3926aa21#r69782580 I believe this is ready to ship.

MatApple:
@jfurrow can you post a pic of Integration tests passing locally? :)

jfurrow:
@MatApple for sure!

PR #629, opened 2016-07-05
DCOS-8260 Hide Disable Schedule button if one-off job

pierlo-upitup:
Remove the `Disable Schedule` button from the More dropdown if the Job has no schedule.


mesosphere-web:
Reviewer: @Poltergeist

src/js/pages/jobs/JobDetailPage.js
```diff
@@ -143,22 +143,27 @@ class JobDetailPage extends mixin(StoreMixin, TabsMixin) {
...
-        id: JobActionItem.SUSPEND
-      },
-      {
         html: 'Destroy',
         id: JobActionItem.DESTROY
       }
     ];
 
+    if (job.getSchedules().length > 0) {
+      dropdownItems.splice(1, 0, {
```
orlandohohmeier: I'd rather push the items in the right order, then using splice to and one in the middle...

pierlo-upitup: OK what is the advantage though?

orlandohohmeier: Whenever we want/need to change the oder we need to also adjust the `splice` call which could result in errors. If we just push the buttons in the correct order it's pretty easy as one only need to move the calls.

pierlo-upitup:
@orlandohohmeier I amended my commit to incorporate your feedback.

src/js/pages/jobs/JobDetailPage.js
```diff
@@ -143,21 +143,26 @@ class JobDetailPage extends mixin(StoreMixin, TabsMixin) {
...
+    let job = MetronomeStore.getJob(this.props.params.id);
+    let dropdownItems = [];
+
+    dropdownItems.push({
+      className: 'hidden',
+      html: 'More',
+      id: JobActionItem.MORE
+    });
+
+    if (job.getSchedules().length > 0) {
```
orlandohohmeier: How about only adding the button it there is a schedule and it's actually activated?

pierlo-upitup: That's a good point. I will do that and then (Outside of this PR) we should actually add a way to Enable Schedules back.

src/js/pages/jobs/JobDetailPage.js
```diff
@@ -143,21 +143,26 @@ class JobDetailPage extends mixin(StoreMixin, TabsMixin) {
...
   getActionButtons() {
-    let dropdownItems = [
-      {
-        className: 'hidden',
-        html: 'More',
-        id: JobActionItem.MORE
-      },
-      {
+    let job = MetronomeStore.getJob(this.props.params.id);
+    let dropdownItems = [];
```
kennyt: what do you think about just having the hidden/destroy objects in the array right when it's invoked so we don't have to do pushing, then inside of the check do a `splice` with the correct index?

orlandohohmeier: @kennyt I actually asked asked @pierlo-upitup to change this to array `push`  in https://github.com/dcos/dcos-ui/pull/629#discussion_r69585121, as using `splice` would mean that whenever we want to change the order or need to add another button we would have to also adjust the splice call which could result in errors.

pierlo-upitup:
Added some more commits to also include a `Enable Schedule` for disabled ones. Here's how it works:

enabledisable


cc @leemunroe

aldipower:
Code looks good to me!!

I am wondering if the 'enabled'-state is verbose enough to the user?
If it's disabled maybe it's not noticeable?

leemunroe:
@pierlo-upitup In the designs we had the schedule displayed up beside the title, looks like that is missing (so that may be a separate issue). If it is there, I would expect to also see it change from `Scheduled for every Thursday at 4:00pm` to `Schedule disabled`.

image

pierlo-upitup:
@leemunroe OK will create an issue for that other part.

orlandohohmeier:
@leemunroe @pierlo-upitup Let's revisit the designs as they still assume that there will be always only one schedule per job. We've changed the API a lot and it is already design to support multiple schedules per job and we probably need to support that in the UI as well.

src/js/events/MetronomeActions.js
```diff
@@ -176,20 +178,42 @@ const MetronomeActions = {
...
+          data: RequestUtil.parseResponseBody(xhr),
+          jobID,
+          xhr
+        });
+      }
+    });
+  },
+
+  enableSchedule: function (jobID, data) {
+    RequestUtil.json({
```
orlandohohmeier: Let's not introduce an `enableSchedule` method which is similar to the `disableSchedule` basically a general purpose update schedule method. Please see https://github.com/dcos/dcos-ui/pull/639 for details.

pierlo-upitup:
Closing this PR in favor of @orlandohohmeier 's  simplified approach : https://github.com/dcos/dcos-ui/pull/639. I'll create a new PR that only adds the view actions.

PR #633, opened 2016-07-05
DCOS-7823: Introduce collapsing string component

jfurrow:
The user specifies a number of characters that should not be truncated at the end of the string, or the default of 15 characters is used.

I opted not to use `classnames/dedupe` here because I thought the performance hit would not be worth it, as I don't expect we'll need to override these classnames. But let me know if you feel otherwise!

### DIFF
--- a/src/js/components/CollapsingString.js
+++ b/src/js/components/CollapsingString.js
@@ -46,8 +46,8 @@ class CollapsingString extends React.Component {
       return false;
     }
 
-    return !DeepEqual(this.props, nextProps)
-      || !DeepEqual(this.state, nextState);
+    return !DeepEqual(this.state, nextState)
+      || !DeepEqual(this.props, nextProps);
   }
 
   getParentWidth() {
--- a/src/js/components/CollapsingString.js
+++ b/src/js/components/CollapsingString.js
@@ -1,0 +1,166 @@
+import DeepEqual from 'deep-equal';
+import React from 'react';
+import ReactDOM from 'react-dom';
+
+import DOMUtils from '../utils/DOMUtils';
+
+const METHODS_TO_BIND = ['updateDimensions'];
+
+class CollapsingString extends React.Component {
+  constructor() {
+    super(...arguments);
+
+    this.state = {
+      collapsed: false,
+      parentWidth: null,
+      stringWidth: null
+    };
+
+    METHODS_TO_BIND.forEach((method) => {
+      this[method] = this[method].bind(this);
+    });
+  }
+
+  componentDidMount() {
+    if (global.window != null) {
+      window.addEventListener('resize', this.updateDimensions);
+      window.addEventListener('focus', this.updateDimensions);
+    }
+
+    this.updateDimensions();
+  }
+
+  componentDidUpdate() {
+    this.updateDimensions();
+  }
+
+  componentWillUnmount() {
+    if (global.window != null) {
+      window.removeEventListener('resize', this.updateDimensions);
+      window.removeEventListener('focus', this.updateDimensions);
+    }
+  }
+
+  shouldComponentUpdate(nextProps, nextState) {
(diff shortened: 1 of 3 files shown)


mesosphere-web:
Reviewer: @daiyiwhale

src/js/components/CollapsingString.js
```diff
@@ -0,0 +1,166 @@
...
+      window.removeEventListener('focus', this.updateDimensions);
+    }
+  }
+
+  shouldComponentUpdate(nextProps, nextState) {
+    if (nextState.parentWidth === null) {
+      return false;
+    }
+
+    if (this.state.collapsed !== nextState.collapsed
```
kennyt: can we just do `!DeepEqual(this.state, nextState)`? then put it on the same line as the deep equals on props below?

src/js/components/CollapsingString.js
```diff
@@ -0,0 +1,166 @@
...
+}
+
+CollapsingString.defaultProps = {
+  endLength: 15,
+  fullStringClassName: 'collapsing-string-full-string',
+  truncatedStringEndClassName: 'collapsing-string-truncated-end',
+  truncatedStringStartClassName: 'collapsing-string-truncated-start',
+  truncatedWrapperClassName: 'collapsing-string-truncated-wrapper',
+  wrapperClassName: 'collapsing-string'
+}
```
kennyt: `;`

src/js/components/CollapsingString.js
jfurrow: Haha yes, good call.

src/js/components/CollapsingString.js
jfurrow: Done!

src/js/components/CollapsingString.js
```diff
@@ -0,0 +1,161 @@
...
+    }
+  }
+
+  shouldComponentUpdate(nextProps, nextState) {
+    if (nextState.parentWidth === null) {
+      return false;
+    }
+
+    return !DeepEqual(this.props, nextProps)
+      || !DeepEqual(this.state, nextState);
```
kennyt: can we do the check for `state` first? i feel like that will be changing at a much higher pace since it changes on `resize`. this will save us comparing props constantly

jfurrow: Good call! Done

src/js/components/CollapsingString.js
```diff
@@ -0,0 +1,161 @@
...
+    let parent = null;
+    let node = ReactDOM.findDOMNode(this);
+
+    if (this.props.parentSelector != null) {
+      parent = DOMUtils.closest(node, this.props.parentSelector);
+    } else {
+      parent = node.parentNode;
+    }
+
+    return DOMUtils.getComputedWidth(parent);
```
MatApple: If `parent === null` will we get an issue here?

jfurrow: Good call, but I think we should default to `document.body` if no parent node is found

src/js/components/CollapsingString.js
```diff
@@ -0,0 +1,161 @@
...
+      return false;
+    }
+
+    return !DeepEqual(this.state, nextState)
+      || !DeepEqual(this.props, nextProps);
+  }
+
+  getParentWidth() {
+    let parent = null;
+    let node = ReactDOM.findDOMNode(this);
```
MatApple: Maybe we should have a check here to make sure we have a node and return early with `0` if we don't. Just in case we have a call to `updateDimensions` on the stack before we unmount.

jfurrow: Done!

src/js/components/CollapsingString.js
jfurrow: Done!

PR #639, opened 2016-07-06
DCOS-8260: Adjust update schedule methods

orlandohohmeier:
**AC**
* Rename the action method to reflect it's capabilities
* Adjust update schedule store events
* Rename suspend/disable schedule method
* Fix disable schedule method


mesosphere-web:
Reviewer: @Poltergeist

pierlo-upitup:
@orlandohohmeier I've pushed some more commits to allow the enabling/disabling of the current schedule from the job detail page.

src/js/pages/jobs/JobDetailPage.js
```diff
@@ -143,21 +149,34 @@ class JobDetailPage extends mixin(StoreMixin, TabsMixin) {
...
+    let dropdownItems = [];
+    let [schedule] = job.getSchedules();
+
+    dropdownItems.push({
+      className: 'hidden',
+      html: 'More',
+      id: JobActionItem.MORE
+    });
+
+    if (schedule != null && schedule.enabled) {
```
aldipower: This could be shortened to:
```JavaScript
if (schedule != null) {
  if (schedule.enabled) {
    ...
  } else {
    ...
  }
}
```

src/js/stores/MetronomeStore.js
```diff
@@ -166,10 +166,40 @@ class MetronomeStore extends EventEmitter {
     MetronomeActions.runJob(jobID);
   }
 
-  suspendSchedule(jobID) {
-    let schedule = this.getJob(jobID).schedules[0];
-    schedule.enabled = false;
-    MetronomeActions.suspendSchedule(jobID, schedule);
+  disableSchedule(jobID) {
+    let job = this.getJob(jobID);
+    if (job == null) {
```
aldipower: Add an extra line above
Remove the extra line below

src/js/pages/jobs/JobDetailPage.js
orlandohohmeier: What's the benefit?

src/js/stores/MetronomeStore.js
```diff
@@ -166,10 +166,40 @@ class MetronomeStore extends EventEmitter {
...
+
+      return null;
+    }
+
+    let [schedule] = job.getSchedules();
+    if (schedule == null) {
+
```
aldipower: Same here

src/js/stores/MetronomeStore.js
```diff
@@ -166,10 +166,40 @@ class MetronomeStore extends EventEmitter {
...
+      return null;
+    }
+
+    this.updateSchedule(jobID, Object.assign({}, schedule, {enabled:false}));
+  }
+
+  enableSchedule(jobID) {
+    let job = this.getJob(jobID);
+    if (job == null) {
+
```
aldipower: Remove extra line

src/js/stores/MetronomeStore.js
```diff
@@ -166,10 +166,40 @@ class MetronomeStore extends EventEmitter {
...
+      return null;
+    }
+
+    let [schedule] = job.getSchedules();
+    if (schedule == null) {
+
```
aldipower: Same here

src/js/stores/MetronomeStore.js
```diff
@@ -166,10 +166,40 @@ class MetronomeStore extends EventEmitter {
...
+      return null;
+    }
+
+    this.updateSchedule(jobID, Object.assign({}, schedule, {enabled:false}));
+  }
+
+  enableSchedule(jobID) {
```
orlandohohmeier: @pierlo-upitup What about adding a `toggleSchedule` method, instead of having to very similar methods?

src/js/stores/MetronomeStore.js
```diff
@@ -166,10 +166,40 @@ class MetronomeStore extends EventEmitter {
...
+      return null;
+    }
+
+    let [schedule] = job.getSchedules();
+    if (schedule == null) {
+
+      return null;
+    }
+
+    this.updateSchedule(jobID, Object.assign({}, schedule, {enabled:false}));
```
aldipower: Space behind the `:`

src/js/stores/MetronomeStore.js
```diff
@@ -166,10 +166,40 @@ class MetronomeStore extends EventEmitter {
...
+    let [schedule] = job.getSchedules();
+    if (schedule == null) {
+
+      return null;
+    }
+
+    this.updateSchedule(jobID, Object.assign({}, schedule, {enabled:true}));
+  }
+
+  updateSchedule(jobID, schedule) {
```
aldipower: If you add the null-check for the schedule here, you can get rid off all of them in the functions above.

orlandohohmeier: You mean this https://github.com/dcos/dcos-ui/pull/639/files/8d027183f2820e3ae1afbbfaf31043feff55938c#diff-cad4282743a1728a5757cf28370730acR193 check?

src/js/stores/MetronomeStore.js
orlandohohmeier: I thought there should always be an empty line before return...

src/js/pages/jobs/JobDetailPage.js
aldipower: You arrange and shorten logically (I mean the maths logic).
If you don't do it, you unnecessarily evaluate `schedule != null` twice!

src/js/stores/MetronomeStore.js
aldipower: That is completely new to me.

src/js/stores/MetronomeStore.js
aldipower: Yes and the others.

src/js/pages/jobs/JobDetailPage.js
orlandohohmeier: That's true, but you're increasing the cyclomatic complexity with your approach and I'd say it's harder to follow...

aldipower: I find it easier to follow!
And I doubt that this is increasing cyclomatic complexity, I guess it remains the same.
On a binary level yours have 2 x 1 levels = 2.
Mine has 1 x 2 levels = 2.

src/js/stores/MetronomeStore.js
pierlo-upitup: I'd say we did enough refactoring on this thing, honestly... :)

src/js/stores/MetronomeStore.js
orlandohohmeier: We need eslint rules for all of this! It's really tedious to give and receive this kind of feedback. Would you mind checking our best practice doc and also have a look, if there is a eslint rule that we could apply?

src/js/stores/MetronomeStore.js
orlandohohmeier: That wouldn't work as schedule would be defined as `{enabled:true}` if we don't do the respective checks. But we should also check for `null` values here.

src/js/stores/MetronomeStore.js
pierlo-upitup: I totally agree. To the point that I'd probably even feel like ignoring this.

src/js/stores/MetronomeStore.js
pierlo-upitup: joking aside, yeah that makes sense. i'll do it

src/js/pages/jobs/JobDetailPage.js
orlandohohmeier: Meaning it's just a matter of taste. I'm not religious about, so I'm fine with changing it. @pierlo-upitup would you mind adjusting it to make @aldipower happy?

pierlo-upitup: If @aldipower thinks the performance advantage is important I can do.

orlandohohmeier: I don't think there is any measurable performance difference. It's more a style questions...

src/js/stores/MetronomeStore.js
kennyt: I actually remember something different like: `always have newlines before returns unless the returns are by itself (like in an if statement)`

src/js/pages/jobs/JobDetailPage.js
aldipower: In this specific case I guess there won't be a significant performance difference. This is not the point.
It's about approaching solutions and logical simplicity. Your approach is more complex on the logical level.

The Apple I by Steve Wozniak for example was a masterpiece of logical connections, thus it needed less chips (expensive).

Another example which points to a similar direction:
If I see 4/6, I will shorten it to 2/3, for the sake of simplicity.

So, I really don't care if you change it or not. Just wanted to clarify my point

src/js/stores/MetronomeStore.js
aldipower: The eslint `newline-before-return`-rule treats both as correct, so I am not sure how this should forced with eslint.
http://eslint.org/docs/rules/newline-before-return

Additionally I refer to Kenny's sentence.

orlandohohmeier:
Integration tests pass locally... 
> `146 passing (6m)`
> `19 pending`

Poltergeist:
As agreed the Ci tests will be fixed in another PR.

PR #640, opened 2016-07-06
DCOS-8071: Fix status dots in task detail

Poltergeist:
This adds the right coloring for task states:

screen shot 2016-07-05 at 17 22 33
screen shot 2016-07-05 at 17 22 07

### DIFF
--- a/src/js/components/TaskTable.js
+++ b/src/js/components/TaskTable.js
@@ -159,7 +159,7 @@ class TaskTable extends React.Component {
 
     let running = ['TASK_RUNNING', 'TASK_STARTING'].includes(state) && unknown;
     let success = healthy && state === 'TASK_RUNNING';
-    let danger = (dangerState && !activeState && dangerState &&
+    let danger = (dangerState && !activeState &&
       ['TASK_ERROR', 'TASK_FAILED'].includes(state)) || healthy === false;
 
     let statusClass = classNames({
--- a/src/js/components/TaskTable.js
+++ b/src/js/components/TaskTable.js
@@ -143,9 +143,40 @@ class TaskTable extends React.Component {
     let statusClassName = TaskUtil.getTaskStatusClassName(task);
     let statusLabelClasses = `${statusClassName} table-cell-value`;
 
+    let {state} = task;
+
+    let dangerState = TaskStates[state].stateTypes.includes('failure');
+
+    let healthy = task.statuses.some(function (status) {
+      return status.healthy;
+    });
+
+    let unknown = task.statuses.some(function (status) {
+      return status.healthy == null;
+    });
+
+    let activeState = TaskStates[state].stateTypes.includes('active');
+
+    let running = ['TASK_RUNNING', 'TASK_STARTING'].includes(state) && unknown;
+    let success = healthy && state === 'TASK_RUNNING';
+    let danger = (dangerState && !activeState && dangerState &&
+      ['TASK_ERROR', 'TASK_FAILED'].includes(state)) || healthy === false;
+
+    let statusClass = classNames({
+      'dot': true,
+      inactive: !activeState,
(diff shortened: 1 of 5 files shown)


mesosphere-web:
Reviewer: @orlandohohmeier

jfurrow:
This looks great! Nice work!

daiyiwhale:
It's beautiful (': ✨

Poltergeist:
By the way are all states right?

\cc @leemunroe

src/js/components/TaskTable.js
```diff
@@ -157,9 +143,40 @@ class TaskTable extends React.Component {
...
+
+    let unknown = task.statuses.some(function (status) {
+      return status.healthy == null;
+    });
+
+    let activeState = TaskStates[state].stateTypes.includes('active');
+
+    let running = ['TASK_RUNNING', 'TASK_STARTING'].includes(state) && unknown;
+    let success = healthy && state === 'TASK_RUNNING';
+    let danger = (dangerState && !activeState && dangerState &&
```
orlandohohmeier: Let's check `dangerState` only once... 😀

orlandohohmeier: Any chance we can refactor this into separate sequential conditions? Might be easier to read.

orlandohohmeier:
Let's please double check it this state is valid:

leemunroe:
Looking great. Couple of nitpicks, open to discussion if you disagree on any.

* Looks like there are 2 shades of grey text going on e.g. `Finished` is different than `Killed`. Is there a good reason? I'd expect them to be the same shade.
* I'm not quite sure what the sequence is of all these statuses occur, can you confirm? `Created` > `Starting` > `Started` > `Running`
* Now that I see `Staging` as yellow I'm wondering why we decided that. I think we copied that from Marathon. Does it still make sense to call that status out as yellow?

Poltergeist:
The 2 shades of grey came from a document in which light grey was mentioned for those states. I can change that. That is also were the yellow color came from for `Staging`. 

I am currently trying to confirm that the status sequence is right.

leemunroe:
* There are two shades of grey for the dots, but the text should be the one shade of grey so it is legible
* And you're right about the yellow for staging, I just wanted to check with you guys if you had an opinion, lgtm

Poltergeist:
image
Updated the styles and removed the 2 shades of grey in the fonts.

The yellow is maybe drawing to much attention, staging isn't a bad state right? As a user I might be confused? I always consider red to be the color of an error and yellow being a warning / notice of unusual behaviour. This is what I do think about it.

kennyt:
hmm i'm currently testing it and staging looks like this:

px

leemunroe:
Yeh seems odd to me. Lets make it white. We can change it in the future if we find there's a good reason to call it out.

Re. staging having a red dot, can a task be unhealthy before it is `Running`? Doesn't seem right to me.

Poltergeist:
All states documented in Mesos in their order:

```
  TASK_STAGING
  TASK_STARTING
  TASK_RUNNING 

  TASK_KILLING

  TASK_FINISHED
  TASK_FAILED
  TASK_KILLED
  TASK_LOST
  TASK_ERROR 
```
I'll check for the staging part.

orlandohohmeier:
Is there a reason for not coloring tasks of suspended/idle services?

screen shot 2016-07-08 at 15 47 42
screen shot 2016-07-08 at 15 47 53

Poltergeist:
This doesn't seem right...

Poltergeist:

pierlo-upitup:
@orlandohohmeier i cant reproduce your 2nd screenshot. Can you ?

orlandohohmeier:
LGTM 💯 ✨

PR #641, opened 2016-07-06
Introduce method to concat list items

orlandohohmeier:
Add a method to concat list items.

### DIFF
--- a/src/js/structs/List.js
+++ b/src/js/structs/List.js
@@ -53,6 +53,30 @@ module.exports = class List {
     this.list.push(cast.call(this, item));
   }
 
+  /**
+   * @param {*} items
+   * @returns {List} List (or child class) containing concatenated items
+   */
+  concatItems(...items) {
+
+    // Prepend current list
+    items.unshift(this);
+
+    // Get items from lists
+    items = items.map(function (item) {
+      if (item instanceof List) {
+        return item.getItems();
+      }
+
+      return item;
+    });
+
+    // Flatten items
+    items = Array.prototype.concat.apply([], items);
+
+    return new this.constructor({items});
+  }
+
   getItems() {
     return this.list;
   }
--- a/src/js/structs/__tests__/List-test.js
+++ b/src/js/structs/__tests__/List-test.js
@@ -87,6 +87,31 @@ describe('List', function () {
 
   });
 
+  describe('#concatItems', function () {
+
+    beforeEach(function () {
+      this.list = new List({items: [1, 2]});
+    });
+
+    it('returns an instance of List', function () {
+      expect(this.list.concatItems()).toEqual(jasmine.any(List));
+    });
+
+    it('concatenate list items', function () {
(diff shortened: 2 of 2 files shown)


mesosphere-web:
Reviewer: @jfurrow

src/js/structs/List.js
```diff
@@ -53,6 +53,30 @@ module.exports = class List {
...
+    // Prepend current list
+    items.unshift(this);
+
+    // Get items from lists
+    items = items.map(function (item) {
+      if ((item instanceof List)) {
+        return item.getItems();
+      }
+
+      return item
```
aldipower: :(

src/js/structs/List.js
```diff
@@ -53,6 +53,30 @@ module.exports = class List {
...
+   * @returns {List} List (or child class) containing concatenated items
+   */
+  concatItems(...items) {
+
+    // Prepend current list
+    items.unshift(this);
+
+    // Get items from lists
+    items = items.map(function (item) {
+      if ((item instanceof List)) {
```
aldipower: Double parenthesis.
Looks pretty nice actually. ;)

src/js/structs/List.js
orlandohohmeier: I don't do this to offend you, but there is no eslint rule to enforce the semicolon and my IDE is not adding them as they aren't required by the ECMAScript (Automatic Semicolon Insertion).

aldipower:
I think this is the way to do it, so I am fine with it.

---
Not your concern, but I _don't_ like to introduce something probably very non-performant for a simple problem that is already solved very good.
Just want to shout this into the room.

src/js/structs/List.js
aldipower: I know. But to be honest, you are the only one, where I notice this regularly, so it is a little bit annoying. Nevermind.

rcorral:
@orlandohohmeier what's the specific use case?

src/js/structs/List.js
```diff
@@ -53,6 +53,30 @@ module.exports = class List {
...
+   */
+  concatItems(...items) {
+
+    // Prepend current list
+    items.unshift(this);
+
+    // Get items from lists
+    items = items.map(function (item) {
+      if (item instanceof List) {
+        return item.getItems();
```
pierlo-upitup: Don't we have a way to get a flattened array of `items` already from an arbitrarily nested list?

orlandohohmeier: Not, that I'm aware of...

orlandohohmeier:
@rcorral I needed something to concat three different lists (to build the job history) and was looking for something nicer than the following:

```
new List({items:[].concat(FirstList.getItems(),SecondList.getItems(),ThirdList.getItems()})
```

But I'll probably merge the items in a different place (before creating lists from them) and don't need this method any more.

kennyt:
so what's the deal with this PR now? is it unnecessary?

orlandohohmeier:
Yep, it's unnecessary. Closing it.

PR #643, opened 2016-07-06
Rename job run structs

orlandohohmeier:
Adjust the job run and job run list struct naming as they will serve as the base class for all job runs, including old once (history).

### DIFF
--- a/src/js/structs/JobRun.js
+++ b/src/js/structs/JobRun.js
@@ -2,7 +2,7 @@ import DateUtil from '../utils/DateUtil';
 import Item from './Item';
 import JobTaskList from './JobTaskList';
 
-module.exports = class JobActiveRun extends Item {
+class JobRun extends Item {
   getDateCreated() {
     return DateUtil.strToMs(this.get('createdAt'));
   }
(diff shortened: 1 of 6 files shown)


mesosphere-web:
Reviewer: @daiyiwhale

jfurrow:
retest this please

src/js/structs/JobRun.js
```diff
@@ -2,7 +2,7 @@ import DateUtil from '../utils/DateUtil';
 import Item from './Item';
 import JobTaskList from './JobTaskList';
 
-module.exports = class JobActiveRun extends Item {
+class JobRun extends Item {
   getDateCreated() {
```
pierlo-upitup: At present this only works for runs that belong to the `activeRuns`. The runs returned as part of the history have slightly dissimilar fields (`completedAt` vs `finishedAt` and maybe others...)

orlandohohmeier: Thats, true! But luckily they all use `createdAt` and that's the only date we're using at the moment. I'll add the rest (e.g. `status`) in the struct or `MetronomeUtil`, which should enable us to use the very same struct for all runs.

orlandohohmeier:
> `146 passing (6m)`
> `19 pending`

orlandohohmeier:

orlandohohmeier:

PR #648, opened 2016-07-06
Apply overflow: hidden to body

jfurrow:
All content overflow is handled within `div#canvas`, which will always fill the viewport, so this shouldn't be a problem. It prevents the ugly scrollbar from appearing when the user dropup menu is animating closed. I tested this in all browsers, but please try to break it.

### DIFF
--- a/src/styles/layout/foundation.less
+++ b/src/styles/layout/foundation.less
@@ -29,6 +29,10 @@ body,
 
 }
 
+body {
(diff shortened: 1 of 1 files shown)


mesosphere-web:
Reviewer: @pierlo-upitup

kennyt:
i can't find anything wrong

PR #653, opened 2016-07-07
Add secret value component

mlunoe:
Creating SecretValue component to be used internally and externally.

Reviewer: See external plugin branch of same name


mesosphere-web:
Reviewer: @daiyiwhale

kennyt:
i tested it out and it works great

PR #654, opened 2016-07-07
[1 of 3: CollapsingString] Use greater than, not greater than or equal to

jfurrow:
Previously the string would collapse if the string width matched exactly the parent width. This is not desirable.


mesosphere-web:
Reviewer: @MatApple

PR #659, opened 2016-07-07
Use 1/3rd of the end of the string by default

jfurrow:
@kennyt had the great suggestion of using 1/3rd of the length of the string as the default number of characters to display, instead of the arbitrary 15 I had selected previously.

### DIFF
--- a/src/js/components/CollapsingString.js
+++ b/src/js/components/CollapsingString.js
@@ -118,12 +118,13 @@ class CollapsingString extends React.Component {
 
   render() {
     let fullString = this.props.string;
+    let endLength = this.props.endLength || Math.floor(fullString.length * 1/3);
     let stringEnding = null;
 
     if (this.state.collapsed) {
       stringEnding = (
         <span className={this.props.truncatedStringEndClassName}>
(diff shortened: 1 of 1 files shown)


mesosphere-web:
Reviewer: @kennyt

src/js/components/CollapsingString.js
```diff
@@ -118,12 +118,13 @@ class CollapsingString extends React.Component {
 
   render() {
     let fullString = this.props.string;
+    let endLength = this.props.endLength || Math.floor(fullString.length * 1/3);
```
kennyt: not sure this might ever happen, but what if `props.endLength` is 0? i can imagine us sometime in the far future programmatically changing `endLength` to 0 for some reason.

sorry for the nit

jfurrow: Good call, fixed!

pierlo-upitup:
But that means the length is now unpredictable. Is this actually desirable?

kennyt:
@pierlo-upitup the length is only unpredictable if you do not pass in a `endLength`. the 1/3 is simply the default

PR #662, opened 2016-07-08
DCOS-8301: Integrate job history data

orlandohohmeier:


mesosphere-web:
Reviewer: @jfurrow

orlandohohmeier:
Pushed two more commits to fix the "finished" column as well.

src/js/utils/MetronomeUtil.js
```diff
@@ -91,5 +91,36 @@ module.exports = {
     });
 
     return rootTree;
+  },
+
+  parseJob(job) {
+    let {history}= job;
+
+    if (history == null) {
+
```
aldipower: Same here in this PR, see other PR.

src/js/utils/MetronomeUtil.js
```diff
@@ -91,5 +91,36 @@ module.exports = {
...
+      return job;
+    }
+
+    let {failedFinishedRuns = [], successfulFinishedRuns = []} = history;
+
+    failedFinishedRuns = failedFinishedRuns.map(function (jobRun) {
+
```
aldipower: This extra line looks so strange. Wobbling in my eyes.

aldipower:
Besides the extra line minor this looks very good!
Thank you! :)

src/js/utils/MetronomeUtil.js
orlandohohmeier: I'm ok with adjusting this, but as said in that other PR would really like to have an `eslint` rule for this. We're wasting way to much time with manually enforcing style! Either there is a rule or it shouldn't be considered an error – let's focus on the important parts and shipping features.

src/js/utils/MetronomeUtil.js
orlandohohmeier: Can we please don't comment like this on any ones work!

src/js/utils/MetronomeUtil.js
aldipower: As I wrote in the other PR, there the according eslint-rule treats both cases as valid.
So I cannot fulfil your request.

src/js/utils/MetronomeUtil.js
aldipower: Sorry, don't wanted to be harsh.
But reviewing is also work. 
And I kindly felt ignored here, which is also not so nice.
This 'extra line' thingy was already discussed in the other PR.

src/js/utils/MetronomeUtil.js
orlandohohmeier: I only added those blank lines as I was told, that we should always add blank lines in those cases.  Now it turns out that this blank lines should only be there under some circumstances. This is a waste of time! Time we could've better use to fix bugs or ship features.
Let's please add rules where possible, and agree to not block any PR by stylistic comments which are not covered in .eslintrc rules. Either there is a rule, or the style is important enough to write our own eslint plugin.  In any other case, we should just stop complaining!

aldipower: So if commenting on non-eslint stylistics is considered as a waste of time and we don't do it, I wonder how we should define eslint-rules. In my opinion eslint-rules grow over time, coming out from discussions.
Defining them all at once is first incomplete and very static. :(

orlandohohmeier: Defining a holistic set rules doesn't mean that we can't change them. It also doesn't mean that we should stop discussing these!  Defining them as eslint rules just ensures that we all, including third party developers, are aware of such rules. I'd suggest to start with a set of very well documented rules like those form [airbnb](https://github.com/airbnb/javascript), and adjust that set where wee see need. 

/cc @dcos/frontend

src/js/utils/MetronomeUtil.js
orlandohohmeier: I know, it's ton of work, and I really appreciate your work! I didn't added these lines to offend you. I thought we're supposed to, which is due to the lack of a defined set of rules.

rcorral:
retest this please

rcorral:
retest this please

PR #665, opened 2016-07-08
DCOS-8375: Reset dropdown filters on filter clear

daiyiwhale:
Dropdowns were not being reset when search filters were cleared. This PR fixes that.

before:
(Dropdown still shows 'Unhealthy' after filter is cleared)
filters are cleared but dropdown is still displaying old filter param noo

after:
(Dropdown resets to default text 'Filter By Service' after filter is cleared)
dropdown resetting upon clearing filters. aw yus.

### DIFF
--- a/src/js/components/FilterByService.js
+++ b/src/js/components/FilterByService.js
@@ -93,8 +93,9 @@ var FilterByService = React.createClass({
         dropdownMenuListItemClassName="clickable"
         wrapperClassName="dropdown"
         items={this.getDropdownItems()}
-        onItemSelection={this.handleItemSelection}
         initialID={this.getSelectedId(this.props.byServiceFilter)}
+        onItemSelection={this.handleItemSelection}
+        ref={(ref) => this.dropdown = ref}
         transition={true}
         transitionName="dropdown-menu" />
     );
--- a/src/js/components/FilterByService.js
+++ b/src/js/components/FilterByService.js
@@ -84,6 +84,12 @@ var FilterByService = React.createClass({
     return id;
   },
 
+  setDropdownValue: function (id) {
+    this.dropdown.setState({
+      selectedID: id
+    });
+  },
+
   render: function () {
     return (
       <Dropdown
--- a/src/js/components/HealthTab.js
+++ b/src/js/components/HealthTab.js
@@ -86,7 +86,7 @@ class HealthTab extends React.Component {
 
   resetFilter() {
     if (this.healthFilter !== null && this.healthFilter.dropdown !== null) {
-      this.healthFilter.dropdown.setState({selectedID: 'all'});
+      this.healthFilter.setDropdownValue('all');
     }
 
     this.setState({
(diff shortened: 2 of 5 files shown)


mesosphere-web:
Reviewer: @mlunoe

src/js/components/FilterByService.js
```diff
@@ -93,8 +93,9 @@ var FilterByService = React.createClass({
         dropdownMenuListItemClassName="clickable"
         wrapperClassName="dropdown"
         items={this.getDropdownItems()}
-        onItemSelection={this.handleItemSelection}
         initialID={this.getSelectedId(this.props.byServiceFilter)}
+        onItemSelection={this.handleItemSelection}
+        ref={(ref) => this.dropdown = ref}
```
kennyt: why don't you just make this `ref` a string and you can just access it by doing `this.refs.dropdown`

daiyiwhale: docs say that method is on the way to being deprecated: https://facebook.github.io/react/docs/more-about-refs.html#the-ref-string-attribute ):

daiyiwhale: lookin out for future me :P

kennyt: damn nice. that's awesome

jfurrow: Good find! I didn't know that... I've exclusively used string `ref`s up until now

daiyiwhale: the more you know 🌈

kennyt: what do you think about this: currently all the components are accessing the dropdown inside of `FilterByService`.

what if we had a method on `FilterByService` called `setDropdownValue` so that the components wouldn't have to access deep properties like `this.healthFilter.dropdown`

instead it would be like `this.healthFilter.setDropdownValue('all')`

this is nice because we can change the `dropdown` name and we'd only need to change it in one place

kennyt:
after my last comment, this looks good 👍

src/js/components/FilterByService.js
mlunoe: I like your suggestion @kennyt!
It also feels like we're doing something wrong if we have to access something nested within a subcomponent like that

daiyiwhale: I love that idea!

daiyiwhale:
done!

rcorral:
retest this please

src/js/components/FilterByService.js
mlunoe: Is there an update on this?

daiyiwhale: https://github.com/dcos/dcos-ui/pull/665/commits/2944772702a83b5db0d2ad1708ebbf05809257e6?

daiyiwhale:
retest this please

rcorral:
retest this please

kennyt:
tested this out and it looks good

daiyiwhale:
retest this please

PR #666, opened 2016-07-08
Service form validation foundation

kennyt:
This depends on:
https://github.com/mesosphere/reactjs-components/pull/303

This adds the functionality of validation for the schema forms. You'd create a field called `externalValidator` on the schema like this:

px

### DIFF
--- a/src/js/components/ServiceForm.js
+++ b/src/js/components/ServiceForm.js
@@ -104,6 +104,42 @@ class ServiceForm extends SchemaForm {
     this.props.onChange(...arguments);
   }
 
+  validateForm() {
+    let model = this.triggerTabFormSubmit();
+
+    // Clears all previous errors.
+    FormUtil.applyToDefinitions(this.multipleDefinition, function (definition) {
+      definition.showError = false;
+    });
+
+    // Apply all validations.
+    let validated = false;
+    FormUtil.applyToDefinitions(this.multipleDefinition, function (definition) {
+      if (typeof definition.externalValidator !== 'function') {
+        return;
+      }
+
+      let fieldValidated = definition.externalValidator(model, definition);
+      if (!fieldValidated && validated) {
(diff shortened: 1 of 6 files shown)


mesosphere-web:
Reviewer: @philipnrmn

src/js/components/ServiceForm.js
```diff
@@ -104,6 +104,40 @@ class ServiceForm extends SchemaForm {
     this.props.onChange(...arguments);
   }
 
+  validateForm() {
+    let model = this.triggerTabFormSubmit();
+
+    // Clears all previous errors.
+    FormUtil.forEachDefinition(this.multipleDefinition, function (definition) {
```
mlunoe: I am probably way too late to the game here, but why are we storing `multipleDefinition` on `this`?
Couldn't we use internalStorage for this?
This is probably not something for this PR, but I thought I'd mention it, as I do not like that pattern.

orlandohohmeier: @mlunoe I can't agree more with you! We should never store any data on components itself.  Let's get rid of this one during the schema form "rewrite", which is scheduled for the next release if I remember it correctly...

src/js/components/ServiceForm.js
```diff
@@ -104,6 +104,40 @@ class ServiceForm extends SchemaForm {
...
+      definition.showError = false;
+    });
+
+    // Apply all validations.
+    let validated = true;
+    FormUtil.forEachDefinition(this.multipleDefinition, function (definition) {
```
orlandohohmeier: Any chance we can combine the two `forEachDefinition` calls?

kennyt: perfect. don't know how i missed that

jfurrow:
LGTM and unit tests are passing.

PR #669, opened 2016-07-09
Allow components to be rendered in DescriptionList dt

daiyiwhale:
Previously, only strings could be rendered in DescriptionList `dt` because it was required to be a valid key value in an object. This PR extends the capabilities of DescriptionList so that `dt`s can contain components.

### DIFF
--- a/src/js/components/DescriptionList.js
+++ b/src/js/components/DescriptionList.js
@@ -21,16 +21,6 @@ class DescriptionList extends React.Component {
     return Object.keys(hash).map((key, index) => {
       let value = hash[key];
 
-      // Check if we need to render a component in the dt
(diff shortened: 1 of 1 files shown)


mesosphere-web:
Reviewer: @Poltergeist

kennyt:
hmm alright this is pretty cool. what do you think about instead of `renderKeys` being an array, it's a map to a render function. so it'd look like

```js
renderKeys = {
  clientSecret: function (key, value) {
    return (
      <h1>hello there</h1>
    )
  }
}
```

daiyiwhale:
Oh good idea! :D

daiyiwhale:
done!

PR #672, opened 2016-07-11
DCOS-8463: handle docker params as an array of objects

pierlo-upitup:
For some reason the previous implementation converted docker parameters to a record-like object, which produced an error when creating an app:
image

Docker Parameters should instead be an array of `{key: k, value: v}` objects as per the Marathon Docs: 

https://mesosphere.github.io/marathon/docs/native-docker.html

This PR does the following:

- intercept and use correct top-level dockerParameters mapping for server-side error
- treat docker parameters as an array
- add some basic unit tests

Full workflow GIF:

params

Tests are passing locally.


Poltergeist:
LGTM 🚢  Thanks for taking care of this.

PR #673, opened 2016-07-11
DCOS-8454: Fix error frame in service modal to not scroll if shown in small screens

Poltergeist:
This fixes a problem when an error is shown in smaller screens and the error message could be scrolled out of the visible area.

### DIFF
--- a/src/js/components/modals/ServiceFormModal.js
+++ b/src/js/components/modals/ServiceFormModal.js
@@ -331,15 +331,13 @@ class ServiceFormModal extends mixin(StoreMixin) {
     }
 
     return (
-      <div>
-        <div className="error-field text-danger">
-          <h4 className="text-align-center text-danger flush-top">
-            {errorMessage.message}
-          </h4>
-          <ul>
-            {errorList}
-          </ul>
-        </div>
+      <div className="error-field text-danger">
+        <h4 className="text-align-center text-danger flush-top">
+          {errorMessage.message}
+        </h4>
+        <ul>
+          {errorList}
+        </ul>
       </div>
     );
   }
@@ -372,6 +370,11 @@ class ServiceFormModal extends mixin(StoreMixin) {
     let {jsonDefinition, jsonMode, service} = this.state;
 
     if (jsonMode) {
+      let height = '370px';
+
+      if (this.state.errorMessage) {
+        height = '462px';
+      }
       return (
         <Ace editorProps={{$blockScrolling: true}}
           mode="json"
@@ -379,7 +382,7 @@ class ServiceFormModal extends mixin(StoreMixin) {
           showGutter={true}
           showPrintMargin={false}
           theme="monokai"
-          height="462px"
+          height={height}
           value={jsonDefinition}
           width="100%"/>
       );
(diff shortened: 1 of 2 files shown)


mesosphere-web:
Reviewer: @aldipower

pierlo-upitup:
@leemunroe we should also consider slimming down the padding in smaller screens. If you get a list of errors, the json window will become unusably short.

src/js/components/modals/ServiceFormModal.js
```diff
@@ -372,14 +370,19 @@ class ServiceFormModal extends mixin(StoreMixin) {
     let {jsonDefinition, jsonMode, service} = this.state;
 
     if (jsonMode) {
+      let height = '370px';
+
+      if (this.state.errorMessage) {
+        height = '462px';
+      }
```
pierlo-upitup: Hmmm... I'm a bit worried about these hardcoded numbers. Can't we think of a more robust solution? The problem begins with having to set the Ace editor height explicitly, but doing this doesn't make things much better.

Idea: calculate the actual height of the error block and subtract that from the desired Ace editor height? 
Also, what if the error block is taller than `92px` (which seems to be the hardcoded value set in `multiple-form.less` below)? 
If I'm not wrong the error block grows in height to accomodate the amount of errors shown...

src/styles/components/multiple-form.less
```diff
@@ -235,6 +235,15 @@ base/forms.less
...
       border-bottom: @red 1px solid;
       padding: (@base-spacing-unit / 2) @base-spacing-unit;
+      position: absolute;
+      width: 100%;
+      top: 88px;
+      left: 0;
+      z-index: 5;
+
+      & + div#brace-editor {
+        margin-top: 92px;
```
pierlo-upitup: See my comment above... This feels rather workaround-y and I'd love to suggest exploring other solutions if possible...

pierlo-upitup:
Point in case: the hardcoded values don't work with varying error block heights. 
case

Poltergeist:
The static numbers are not solving the problem, But we have to use static numbers since values like `auto` and `100%` are breaking the json mode and the ace editor isn't showing anymore

image

Possible solutions could be:
 - we could use assumption about the height of the error frame.
 - place the error Frame above the JSON editor and remove it onClick

jfurrow:
I think the best solution is @Poltergeist where the errors are rendered "on top of" the ACE editor and the user must dismiss them before continuing to edit the JSON. The only other reasonable solution to this is to set a static height on both the ACE editor and the error container, and then scroll the error container when it is taller than the static height...

Perhaps @leemunroe can offer some advice as to where this should be displayed, but it might be nice if the floating box which contains the errors is offset from the edges of the Ace container slightly, such that it's obvious to the user that the editor still exists behind the error messages. Somewhat similar to this screenshot:

leemunroe:
Are we able to make use of this functionality built in to Ace Editor?
image

If so I'd imagine
* We keep the global error banner to 1 line "There's an issue with your JSON"
* Highlight each issue/line with this Ace editor tooltip

rcorral:
retest this please

orlandohohmeier:

Poltergeist:
Closed since this solution does not work.

leemunroe:
@Poltergeist Is this being tracked elsewhere? Here are our proposed designs for the modal validation. https://invis.io/R381VNZXA

PR #674, opened 2016-07-11
Fix job detail run history integration tests

orlandohohmeier:
Adjust the default sort behaviour and fix the job run history integration tests.

### DIFF
--- a/tests/_fixtures/metronome/job.json
+++ b/tests/_fixtures/metronome/job.json
@@ -111,7 +111,7 @@
         }
       ]
     }
-  ] ,
+  ],
   "history": {
     "failedFinishedRuns": [
       {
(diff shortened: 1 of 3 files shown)


mesosphere-web:
Reviewer: @jfurrow

Poltergeist:
image

🚢  LGTM!

tests/_fixtures/metronome/job.json
```diff
@@ -111,7 +111,7 @@
         }
       ]
     }
-  ] ,
```
pierlo-upitup: i wonder if we should ignore this change – not a biggie but cypress keeps changing these fixtures every time it runs and it's pretty annoying...

pierlo-upitup:
If the fixture change does not bother you, 🚢

tests/_fixtures/metronome/job.json
orlandohohmeier: What do you recommend? Should I drop that change?

pierlo-upitup: As you please.

orlandohohmeier: Ok.. I'll not change it, as the cypress versions (0.16.4) I'm using, seem to be is ok with it.

PR #663, opened 2016-07-08
DCOS-8101: Add service state to the subline rather then the health status

Poltergeist:
image

image

image

This changes the sub headline on detail pages to states rather then health status.

PR #675, opened 2016-07-11
DCOS-8440: Add suspended state if service has 0 instances configured

Poltergeist:
image

If a service has 0 `instances` configured the service state will be set to suspended.

### DIFF
--- a/src/js/structs/Service.js
+++ b/src/js/structs/Service.js
@@ -146,8 +146,9 @@ module.exports = class Service extends Item {
     let deployments = this.getDeployments();
     let queue = this.getQueue();
 
-    if (this.getInstancesCount() === 0 &&
-      this.getTasksSummary().tasksRunning === 0
+    let instances = this.getInstancesCount();
+    if (instances === 0 &&
+      tasksRunning === 0
     ) {
       return ServiceStatus.SUSPENDED;
     }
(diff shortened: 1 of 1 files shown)


mesosphere-web:
Reviewer: @daiyiwhale

kennyt:
can we add a unit test for this?

src/js/structs/Service.js
```diff
@@ -146,6 +146,10 @@ module.exports = class Service extends Item {
     let deployments = this.getDeployments();
     let queue = this.getQueue();
 
+    if (this.getInstancesCount() === 0 && deployments.length === 0) {
```
pierlo-upitup: We need to also check against the amount of running tasks. From the docs: 

> An application with a target instances of 0 and whose running tasks count is 0. True when the JSON returned looks like: app.instances === 0 && app.tasksRunning === 0

https://mesosphere.github.io/marathon/docs/marathon-ui.html

Poltergeist: changed deployments for tasksRunning

orlandohohmeier:
Not necessarily in the scope of this PR, but let's make sure we're using/displaying the same data in the service detail.

suspended

pierlo-upitup:
@orlandohohmeier i believe that comes with https://github.com/dcos/dcos-ui/pull/663/files

src/js/structs/Service.js
```diff
@@ -146,6 +146,12 @@ module.exports = class Service extends Item {
     let deployments = this.getDeployments();
     let queue = this.getQueue();
 
+    if (this.getInstancesCount() === 0 &&
```
orlandohohmeier: Isn't this "duplicating" https://github.com/dcos/dcos-ui/pull/675/files#diff-6e602e3f57a52157c9dba6acf1858558R171?

Poltergeist: Good catch will fix it by moving the condition to the top.

orlandohohmeier:
LGTM! 🚢

PR #677, opened 2016-07-11
DCOS-8447: Add deployment action to deployment table

Poltergeist:
image

This adds the deployment action to the deployment table.

### DIFF
--- a/src/js/pages/services/DeploymentsTab.js
+++ b/src/js/pages/services/DeploymentsTab.js
@@ -152,13 +152,26 @@ class DeploymentsTab extends mixin(StoreMixin) {
   renderStatus(prop, deployment) {
     const title = `Step ${deployment.getCurrentStep()} of ${deployment.getTotalSteps()}`;
     const services = deployment.getAffectedServices();
-    const action = deployment.currentActions.map(function (action) {
-      return action.action;
-    }).join(', ');
+
+    let currentActions = {};
+    if (deployment.currentActions && deployment.currentActions.length > 0) {
+      currentActions =
+        deployment.currentActions.reduce(function (memo, action) {
+          memo[action.app] = action.action;
+
+          return memo;
+        }, {});
+    }
+
     const items = services.map(function (service, index) {
+      let statusText = service.getStatus();
+      if (currentActions[service.id] != null) {
+        statusText += ` (${currentActions[service.id]})`;
+      }
+
       return (
         <li key={index}>
-          {service.getStatus()} ({action})
+          {statusText}
         </li>
       );
     });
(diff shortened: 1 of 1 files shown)


mesosphere-web:
Reviewer: @mlunoe

src/js/pages/services/DeploymentsTab.js
```diff
@@ -152,10 +152,13 @@ class DeploymentsTab extends mixin(StoreMixin) {
...
     const title = `Step ${deployment.getCurrentStep()} of ${deployment.getTotalSteps()}`;
     const services = deployment.getAffectedServices();
+    const action = deployment.currentActions.map(function (action) {
+      return action.action;
+    }).join(', ');
     const items = services.map(function (service, index) {
       return (
         <li key={index}>
-          {service.getStatus()}
+          {service.getStatus()} ({action})
```
pierlo-upitup: We should probably make sure the `action != null` or we may end up showing `Deploying (undefined)` or worse.

src/js/pages/services/DeploymentsTab.js
```diff
@@ -152,10 +152,13 @@ class DeploymentsTab extends mixin(StoreMixin) {
   renderStatus(prop, deployment) {
     const title = `Step ${deployment.getCurrentStep()} of ${deployment.getTotalSteps()}`;
     const services = deployment.getAffectedServices();
+    const action = deployment.currentActions.map(function (action) {
+      return action.action;
```
pierlo-upitup: Shouldn't we ensure the action is affecting the current service? Keep in mind this will be on a per-app row. 

I wonder if this should be inside of the `map` starting at L157.

src/js/pages/services/DeploymentsTab.js
kennyt: instead of doing `action != null` we can do a `filter` after the `map` and before the `.join(', ')`

pierlo-upitup: @kennyt why iterate over the list twice anyway?

Poltergeist: action would never be null in that case. But it is not error safe if currentActions isn't present it would fail I changed this way to a reduce. And an if in the map below.

src/js/pages/services/DeploymentsTab.js
Poltergeist: I ensured that the action is only shown if there is an action for the current service.

pierlo-upitup:
LGTM! 
adding an example with multiple apps for reference:

image

pierlo-upitup:
🚢

PR #680, opened 2016-07-11
Add a more unique key onto mask to fix weird mask animation

kennyt:
# Before
px

# After
px


mesosphere-web:
Reviewer: @Poltergeist

PR #681, opened 2016-07-11
[#1 Action dropdown service row] Add edit and delete actions for groups

kennyt:
Adding these actions so in the future we can implement buttons on groups like from marathon-ui:

px


mesosphere-web:
Reviewer: @orlandohohmeier

pierlo-upitup:
Thanks @kennyt – actually I'm not sure if we need/want the `Edit Group` part because even a simple action such as changing a group name will result in a new deployment involving all of its applications, which in some cases may lead to unwanted results.

In case we do want to offer a way to edit an entire group though, I strongly suggest to warn the user about the outcome of the action very clearly, @leemunroe.

kennyt:
@pierlo-upitup the scaleBy also uses the `Edit Group` endpoint doesn't it? as far as i can tell from designs, there's no way in the UI that will allow you to change group name

rcorral:
retest this please

src/js/stores/MarathonStore.js
```diff
@@ -92,6 +96,22 @@ class MarathonStore extends GetSetBaseStore {
         case ActionTypes.REQUEST_MARATHON_GROUP_CREATE_SUCCESS:
           this.emit(MARATHON_GROUP_CREATE_SUCCESS);
           break;
+        case ActionTypes.REQUEST_MARATHON_GROUP_DELETE_ERROR:
+          let groupErrorMessage = action.data;
+          if (!Object.keys(groupErrorMessage).length) {
```
MatApple: `groupErrorMessage` is going to be emitted as either an Object or a String. Is that handled appropriately in views or should we assemble the string here when we get an Object?

kennyt: there are no views in this PR. i'll handle this issue on https://github.com/dcos/dcos-ui/pull/691

kennyt: @MatApple i now encode the group IDs for actions

MatApple:
retest this please

MatApple:
retest this please

PR #682, opened 2016-07-11
Implement TaskEndpointsList

jfurrow:
Here it is in action:
https://cl.ly/2z3L133h0A28

@leemunroe How do you want to handle this responsively? I'm not super confident about the way I implemented it. Also, what should the toggle text say after the user has clicked `more...`? Right now it says `less` which seems a little silly.


mesosphere-web:
Reviewer: @aldipower

leemunroe:
@jfurrow Nice. Anything particular you were concerned about? Responsiveness looks fine to me. It would be nice to use that extra space in the name column before the endpoints wrap, and similarly eat up the space between the state and CPU column before we hide the task name. But this is likely a larger issue beyond the goal of this.

More/Less sounds good to me. Would have liked to see the count e.g. `13 more...`

jfurrow:
@leemunroe Great! I'll add the count in there. Unfortunately we are a little restricted with the column widths. Right now we can either assign a fixed width or a percentage to the column width. I have it set to `20%` currently.

kennyt:
conflicts

jfurrow:
@mlunoe And I discussed this offline; I'm going to make use of his methods in #683 once they are moved into TaskUtil. So this can't be merged until that happens :)

src/js/components/__tests__/MarathonTaskDetailsList-test.js
```diff
@@ -1,106 +0,0 @@
-jest.dontMock('../MarathonTaskDetailsList');
```
jfurrow: This test file only tested a method that has been abstracted into multiple parts. I tested the new component in `TaskEndpointsList-test`

rcorral:
retest this please

src/js/components/TaskEndpointsList.js
```diff
@@ -0,0 +1,158 @@
...
+  getNullList() {
+    return <span>N/A</span>;
+  }
+
+  getEndpointsList(hosts, ports = []) {
+    let totalPortCount = ports.length;
+
+    return hosts.map((host, hostIndex) => {
+      let isExpanded = this.state.expandedHosts.includes(host);
+      let {portLimit} = this.props;
```
kennyt: we can bring `portLimit` instantiation outside of the `#map`

jfurrow: We actually can't because it gets changed if `isExpanded` is true

src/js/components/TaskTable.js
```diff
@@ -50,6 +52,15 @@ class TaskTable extends React.Component {
         sortFunction
       },
       {
+        className: this.getHostColumnClassname,
```
kennyt: just curious, is this cacheable? this will matter with large task tables which will be often

jfurrow: Yeah it can be cached if the list of hosts and ports are the exact same. How should I define that?

src/js/components/TaskEndpointsList.js
```diff
@@ -0,0 +1,158 @@
...
+import TaskUtil from '../utils/TaskUtil';
+
+const METHODS_TO_BIND = [
+  'getEndpointsList',
+  'getPortsList',
+  'getShowHideLink',
+  'handlePortsToggle'
+];
+
+class TaskEndpointsList extends React.Component {
```
kennyt: is there any way we can get a `shouldComponentUpdate` on this component? just the fact that it's going to be used on every row of the `TaskTable` makes me feel like this will be extremely helpful if it's possible

src/js/components/TaskTable.js
```diff
@@ -165,6 +184,12 @@ class TaskTable extends React.Component {
     );
   }
 
+  renderHost(prop, task) {
+    let marathonTask = MarathonStore.getTaskFromTaskID(task.id);
+
+    return <TaskEndpointsList portLimit={3} task={marathonTask} />;
```
kennyt: can we add a `key` here with the task id or something? i'm not 100% clear on this, but if we add a unique key, then react will know not to re-create the component or something. but let's just do it because we have nothing to lose

src/js/components/TaskTable.js
```diff
@@ -165,6 +184,12 @@ class TaskTable extends React.Component {
     );
   }
 
+  renderHost(prop, task) {
+    let marathonTask = MarathonStore.getTaskFromTaskID(task.id);
```
kennyt: actually, the fact that we pick out the task here is extremely expensive. if we have 500 tasks (which is quite common) then we're going to be iterating through a shit load of things. 

here's the implementation of `getTaskFromTaskID` https://github.com/dcos/dcos-ui/blob/master/src/js/stores/MarathonStore.js#L343

i'd say we should definitely find a way around this somehow before this merges. no idea how

src/js/components/TaskEndpointsList.js
kennyt: ahh i see what you're saying. sounds good

src/js/components/TaskTable.js
kennyt: it's just `cacheCell: true`

src/js/components/TaskTable.js
jfurrow: Good point, but what if the ports or hosts change for a given task? Then this component won't render the updated data because the task ID won't change.

jfurrow: Maybe we should implement some sort of caching in `MarathonStore#getTaskFromTaskID`? Is that crazy talk?

src/js/components/TaskTable.js
jfurrow: When using `cacheCell: true`, the cell doesn't update properly.

Using caching:


Not using caching:

src/js/components/TaskEndpointsList.js
jfurrow: Done!

src/js/components/TaskTable.js
jfurrow: I did add the key but I don't think that's going to help us out: https://facebook.github.io/react/docs/reconciliation.html#keys

jfurrow:
I broke tests. Fixing now :)

src/js/components/TaskTable.js
kennyt: sounds good. caching is a no go

kennyt:
i forgot what the consensus was. is this PR good to merge in, then create a mapping of the tasks later? or should this PR go in after that?

jfurrow:
I believe we are going to hold off on this until the caching of the tasks is completed.

mlunoe:
Beautiful! Code looks good and works great 😄 👍 Great work @jfurrow ! ⛵ 🚢 🚀

mlunoe:
The only test that is failing is also failing on master atm.:


I'm putting up a PR for the fix for that, which is unrelated to this PR.

MatApple:
LGTM, nice work @jfurrow!

PR #701, opened 2016-07-13
DCOS-8430: Rename State to Status in TaskTable

daiyiwhale:
Rename State to Status in TaskTable, for consistency across application.

### DIFF
--- a/src/js/components/TaskTable.js
+++ b/src/js/components/TaskTable.js
@@ -30,8 +30,8 @@ class TaskTable extends React.Component {
     return task.resources[prop];
   }
 
-  getStatusValue(task) {
-    return TaskStates[task['state']].displayName;
+  getStatusValue(state) {
+    return TaskStates[state].displayName;
   }
 
   getColumns() {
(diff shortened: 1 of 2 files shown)


mesosphere-web:
Reviewer: @orlandohohmeier

src/js/components/TaskTable.js
```diff
@@ -30,8 +30,8 @@ class TaskTable extends React.Component {
     return task.resources[prop];
   }
 
-  getStateValue(task, prop) {
-    return TaskStates[task[prop]].displayName;
+  getStatusValue(task) {
+    return TaskStates[task['state']].displayName;
```
jfurrow: We can just use `task.state` here

pierlo-upitup: or we could even `getStatusValue({state}) { return TaskStates[state].displayName; }` if we wanted to be fancy

pierlo-upitup:
Relevant tests are passing
image

src/js/components/TaskTable.js
daiyiwhale: done!

PR #704, opened 2016-07-13
Fix layout on service search results view

philipnrmn:
Bug report: 

> If I search or apply a filter to services, the table/layout shifts, which is jarring and looks like a mistake.
> 
> I think there are 2 issues here:
> 1. The title/breadcrumb is changing style - let's keep it the same i.e. where it changes from "Services" to "Showing 9 of 13 Services", keep this the same font size
> 2. The row height changes due to adding the service location - ideally we can keep the row the same height, similar to how Marathon works today

With this PR:

### DIFF
--- a/src/js/components/ServicesTable.js
+++ b/src/js/components/ServicesTable.js
@@ -187,10 +187,16 @@ var ServicesTable = React.createClass({
   getColumns: function () {
     let className = ResourceTableUtil.getClassName;
     let heading = ResourceTableUtil.renderHeading(ServiceTableHeaderLabels);
+    let {isFiltered} = this.props;
 
     return [
       {
-        className,
+        className: function () {
+          return classNames([
+            className.apply(this, arguments),
+            {'table-cell-short': isFiltered}
+          ]);
+        },
         headerClassName: className,
         prop: 'name',
         render: this.renderHeadline,
--- a/src/styles/components/breadcrumbs.less
+++ b/src/styles/components/breadcrumbs.less
@@ -142,3 +142,11 @@ Breadcrumbs
     }
   }
 }
+
+.breadcrumb-style-headline {
+  margin-bottom: @base-spacing-unit * 2/3;
+  .h4 {
+    font-size: @base-spacing-unit * .65;
+    line-height: @base-spacing-unit;
+  }
+}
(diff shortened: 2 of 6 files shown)


mesosphere-web:
Reviewer: @MatApple

philipnrmn:
On reflection I really dislike the negative margin, I'm going to remove it and drop the padding on the table cell instead.

rcorral:
Nice!

rcorral:
retest this please

daiyiwhale:
Aaah, you've scratched the collective itch :D

pierlo-upitup:
retest this please

src/js/components/ServicesTable.js
```diff
@@ -187,10 +187,16 @@ var ServicesTable = React.createClass({
...
 
     return [
       {
-        className,
+        className: function () {
+          return classNames([
+            className.apply(this, arguments),
+            {'table-cell-short': isFiltered}
+          ]);
+        },
```
pierlo-upitup: Smart!

src/styles/components/breadcrumbs.less
```diff
@@ -142,3 +142,11 @@ Breadcrumbs
     }
   }
 }
+
+.breadcrumb-style-headline {
+  margin-bottom: @base-spacing-unit * 2/3;
+  .h4 {
+    font-size: @base-spacing-unit * .65;
```
pierlo-upitup: I wonder if we should instead declare a font-size rather than using the base-spacing-unit for this...?

pierlo-upitup:
Besides my nit, thanks for this! Looks great.

orlandohohmeier:
Feel free to merge after addressing PGs nit. Great job!

kennyt:
retest this please

src/styles/components/breadcrumbs.less
kennyt: yeah same with `line-height` below this too. i believe `base-spacing-unit` is used for margin/padding only

kennyt:
LGTM after nits

src/styles/components/breadcrumbs.less
philipnrmn: Good shout! I'm going to resist the `line-height` suggestion because (possibly wrongly) the breadcrumb uses `base-spacing-unit` and I'd rather keep the two in line than deviate. Also, 24/13 is not a nice number...

rcorral:
retest this please

PR #708, opened 2016-07-13
Remove suffix suppression logic for JobDetailPage

jfurrow:
Before:

After:

### DIFF
--- a/src/js/pages/jobs/JobDetailPage.js
+++ b/src/js/pages/jobs/JobDetailPage.js
@@ -298,20 +298,14 @@ class JobDetailPage extends mixin(StoreMixin, TabsMixin) {
     });
 
     let timePrefix = null;
-    let shouldSuppressRelativeTime = dateCompleted == null;
-
-    if (shouldSuppressRelativeTime) {
-      timePrefix = 'for ';
-    }
 
     return [
       <span className={statusClasses} key="status-text">
         {status.displayName}
       </span>,
       <span key="time-running">
-        ({timePrefix}{DateUtil.msToRelativeTime(
-          longestRunningActiveRun.getDateCreated(),
-          shouldSuppressRelativeTime
+        (as of {timePrefix}{DateUtil.msToRelativeTime(
+          longestRunningActiveRun.getDateCreated()
         )})
       </span>
     ];


mesosphere-web:
Reviewer: @pierlo-upitup

src/js/pages/jobs/JobDetailPage.js
```diff
@@ -298,20 +298,14 @@ class JobDetailPage extends mixin(StoreMixin, TabsMixin) {
     });
 
     let timePrefix = null;
-    let shouldSuppressRelativeTime = dateCompleted == null;
```
pierlo-upitup:
```
ERROR in ./src/js/pages/jobs/JobDetailPage.js

  291:9  error  'dateCompleted' is defined but never used  no-unused-vars

✖ 1 problem (1 error, 0 warnings)
```

pierlo-upitup:
Nice! Just one question: why do we get `as of secs ago` in the heading, but just `secs ago` in the table? (see 2nd screenshot). I wonder if we should keep these readings consistent? @leemunroe

src/js/pages/jobs/JobDetailPage.js
```diff
@@ -298,20 +298,14 @@ class JobDetailPage extends mixin(StoreMixin, TabsMixin) {
...
 
     return [
       <span className={statusClasses} key="status-text">
         {status.displayName}
       </span>,
       <span key="time-running">
-        ({timePrefix}{DateUtil.msToRelativeTime(
-          longestRunningActiveRun.getDateCreated(),
-          shouldSuppressRelativeTime
+        (as of {timePrefix}{DateUtil.msToRelativeTime(
```
pierlo-upitup: I wonder if we really need the `as of` – if we do, let's consider having this as a consistent prefix?

leemunroe:
I'm not quite sure what the original goal if this fix was, so apologies if this is off topic.

If we're talking seconds we should use the word `seconds` not `secs` e.g. `33 seconds ago`

If we're not able to show seconds, or don't think it is valuable, we could use `Less than a minute ago` similar to how http://timeago.yarp.com/ works

In the screenshots I can't tell if a `title` attribute is present but I would expect the absolute time to be there when hovering the relative time.

And lastly, looks like we're missing the `Run Time` column. Without it I have to do Math to figure out how long the job took to run.
https://invis.io/G27YLX7JW#/162138848_Job_Detail_-_Run_History

leemunroe:
FYI Marathon times https://cl.ly/1y081k231E2x

jfurrow:
@leemunroe 
1. I am changing it so it doesn't display `Running (for secs)` and instead will display `Running (as of secs ago)`. If you'd prefer `(for secs)` then we can leave this as-is.
2. I agree `seconds` makes more sense.
3. We're using [Moment.js](http://momentjs.com/docs/) so we can easily display anything that they support.
4. The `title` attribute is not present currently, I will add it.
5. The `Run Time` column should be done as part of a separate task, I created a JIRA for it.

leemunroe:
Gotcha. Consistency is good. But `Running (as of secs ago)` sounds weird. I would sway towards `Running (less than a minute)`, and because the unit of time here isn't following a number, we don't need to abbreviate.

Thanks for creating Jira and adding title attr 🕐

jfurrow:
@leemunroe We're going to change the way we handle relative time throughout the entire UI, so the changes made in this PR are unnecessary. You can track the progress [in this JIRA](https://mesosphere.atlassian.net/browse/DCOS-8436).

PR #709, opened 2016-07-13
Making sure services always show an icon in service table

mlunoe:
Before:

After:

### DIFF
--- a/src/js/components/ServicesTable.js
+++ b/src/js/components/ServicesTable.js
@@ -102,13 +102,10 @@ var ServicesTable = React.createClass({
     );
   },
 
-  renderHeadline: function (prop, service) {
-    const id = encodeURIComponent(service.getId());
-    let itemImage = null;
-
+  getImage: function (service) {
     if (service instanceof ServiceTree) {
       // Get serviceTree image/icon
-      itemImage = (
+      return (
         <Icon
           className="icon-margin-right inverse"
           color="grey"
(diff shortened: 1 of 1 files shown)


mesosphere-web:
Reviewer: @jfurrow

src/js/components/ServicesTable.js
```diff
@@ -118,23 +115,25 @@ var ServicesTable = React.createClass({
...
-      );
-    }
+    // Get service image/icon
+    return (
+      <span
+        className="icon icon-small icon-image-container icon-app-container icon-margin-right">
+        <img src={service.getImages()['icon-small']}/>
+      </span>
+    );
+  },
```
pierlo-upitup: I like the simplification a lot – but I have one question: shouldn't the `ServiceTree` struct also implement a `getImages()` method so that we can get rid of the `instanceof` check completely here?

mlunoe: We test for everything `ServiceTree` here: https://github.com/dcos/dcos-ui/pull/709/files#diff-a4468ae3d6f9c66e921365d98b241156R106
The only thing that can be in this list is `ServiceTree` or `Service`

kennyt: i think what PG is saying is that if we can implement a `getImages()` onto the `ServiceTree` so that we don't even have to do the check and it'll just work for all instances.

but PG i don't think that's possible at the moment because `ServiceTree` uses the `Icon` component and everything else uses the `<img src=` element

rcorral: I don't think the structs should ever return any JSX. It should be strictly data.

pierlo-upitup: I agree – but https://github.com/dcos/dcos-ui/blob/master/src/js/structs/Service.js#L80

PR #713, opened 2016-07-14
DCOS-8432 Add version column to Tasks list

pierlo-upitup:
So,

Mesos tasks don't carry the Marathon version, so there is no right way to have this information for a completed task. The right solution to this would be that either Mesos starts storing the version info (maybe as a label) or Marathon starts keeping track of completed tasks.

The only possible solution without any major change is to take the Mesos Task ID and use it to grab the equivalent Marathon Task which contains the `version` we need to render the value. This means, however, that we can't display the version of completed tasks.

So when looking at the full, unfiltered task list, this means:

screen shot 2016-07-14 at 16 22 23

@leemunroe @pyronicide @air can you please chime in and let us know if this is good enough for 1.8 ?

### DIFF
--- a/src/js/components/TaskTable.js
+++ b/src/js/components/TaskTable.js
@@ -9,6 +9,7 @@ import ResourceTableUtil from '../utils/ResourceTableUtil';
 import TaskStates from '../constants/TaskStates';
 import TaskTableHeaderLabels from '../constants/TaskTableHeaderLabels';
 import TaskUtil from '../utils/TaskUtil';
+import TableUtil from '../utils/TableUtil';
 import Units from '../utils/Units';
 
 const METHODS_TO_BIND = [
@@ -34,6 +35,15 @@ class TaskTable extends React.Component {
 
   getStatusValue(task) {
     return TaskStates[task.state].displayName;
+  }
+
+  getVersionValue(task) {
+    let marathonTask = MarathonStore.getTaskFromTaskID(task.id);
+    if (marathonTask == null) {
+      return null;
+    }
+
+    return marathonTask.version;
   }
 
   getColumns() {
@@ -106,12 +116,20 @@ class TaskTable extends React.Component {
       {
         cacheCell: true,
         className,
+        getValue: this.getVersionValue,
         headerClassName: className,
         heading,
         prop: 'version',
         render: this.renderVersion,
         sortable: true,
-        sortFunction
+        sortFunction: TableUtil.getSortFunction('id', (task) => {
+          let version = this.getVersionValue(task);
+          if (version == null) {
+            return null;
+          }
+
+          return new Date(version);
+        })
       }
     ];
   }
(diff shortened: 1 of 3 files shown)


mesosphere-web:
Reviewer: @kennyt

leemunroe:
I don't know enough about it to make a hard call but to me it seems like it's more useful to have the versions that we do know.

pyronicide:
This is something that'll be super useful for debugging. Could we schedule this work in 1.9? I love the idea around including the version in the label, that's something we're going to want anyways in 1.9 as part of metrics and logging.

pierlo-upitup:
@leemunroe from a UX standpoint, what do you think about the empty version? Is empty preferable over `N/A`or some other indication that the version is not available ?

pierlo-upitup:
@pyronicide can you also comment on the proposed solution?

pyronicide:
I'm okay with keeping it empty in 1.8.

jfurrow:
Between this and the `Hosts` column, we are running out of space in this table. I think we'll need to find another way to display the `Hosts` and `Version` columns.

air:
Having an empty value for no-longer-with-us Tasks is totally fine for now.

pierlo-upitup:
OK, thanks. @leemunroe , @jfurrow has a point there – what do you suggest?

pierlo-upitup:
Relevant integration tests still passing, unit tests should pass too. 

We should probably add some more meaningful tests for the Service Task table page, but aint nobody got time for that!

pierlo-upitup:
PS here's the final result:

screen shot 2016-07-18 at 15 17 33
screen shot 2016-07-18 at 15 17 06


c @leemunroe

pierlo-upitup:
retest this please

src/js/components/TaskTable.js
```diff
@@ -100,6 +102,15 @@ class TaskTable extends React.Component {
         render: ResourceTableUtil.renderUpdated,
         sortable: true,
         sortFunction
+      },
+      {
+        className,
```
kennyt: can we cache this cell? (look on line 87 for `cacheCell` usage)

times when we can't cache are when the inputs are the same, but some external thing may change the render. for example, if you're rendering the task name in certain colors based on the status, then you can't cache.

but i believe the version cell is cacheable so we can safely cache it :)

orlandohohmeier:
LGTM! 🚢

Poltergeist:
If one sorts on the version column the version text gets removed from the table. 

version-gets-removed-on-sort mov

pierlo-upitup:
@kennyt seems like the issue is due to the cache property. With that removed, the values are back but the sorting is still acting up. Can you take a look? 

image

cc @Poltergeist

mlunoe:
There are some conflicts here

kennyt:
@pierlo-upitup i fixed the merge conflicts, fixed the sorting, and fixed the caching. should now be good

rcorral:
flicts!

pierlo-upitup:
retest this please

Poltergeist:
💅 🇮🇹

PR #716, opened 2016-07-14
Remove Average column in debug table, update header name and number rounding

MatApple:
Updates columns in Service Debug Table. Removed `Average Seconds` column, renamed `Median Seconds` -> `Median Lifetime`. Introduced new number rounding to report minutes when time is > 3600 secs.

Examples of new number rounding:


mesosphere-web:
Reviewer: @daiyiwhale

PR #717, opened 2016-07-14
Fix undefined environment variable

MatApple:
Prevents undefined key/values being set when we build the model and update the definition.

Example of what would happen if you set a container image and command.

### DIFF
--- a/src/js/schemas/service-schema/EnvironmentVariables.js
+++ b/src/js/schemas/service-schema/EnvironmentVariables.js
@@ -21,12 +21,20 @@ let EnvironmentVariables = {
         if (variableMap == null) {
           return [];
         }
-        return Object.keys(variableMap).map(function (key) {
-          return Hooks.applyFilter('variablesGetter', {
+
+        return Object.keys(variableMap).reduce(function (memo, key) {
+          if ((key == null || key === 'undefined')
+            && (variableMap[key] === 'undefined' || variableMap[key] == null)) {
+            return;
+          }
+
+          memo.push(Hooks.applyFilter('variablesGetter', {
             key,
             value: variableMap[key]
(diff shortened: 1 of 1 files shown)


mesosphere-web:
Reviewer: @mlunoe

src/js/schemas/service-schema/EnvironmentVariables.js
```diff
@@ -21,12 +21,20 @@ let EnvironmentVariables = {
         if (variableMap == null) {
           return [];
         }
-        return Object.keys(variableMap).map(function (key) {
-          return Hooks.applyFilter('variablesGetter', {
+
+        return Object.keys(variableMap).reduce(function (memo, key) {
+          if ((key == null || key === 'undefined')
```
kennyt: what if the key is legit 'undefined'?

kennyt:
LGTM after my one comment. although i'd be willing to merge even if my comment doesn't get addressed

src/js/schemas/service-schema/EnvironmentVariables.js
rcorral: just do `typeof key === 'undefined'`

kennyt: right here it's checking if the key is a string `undefined`. i forget the reason for this, but i remember sometime before this point, `undefined` got converted to `'undefined'`

MatApple: We only run the risk of filtering out a variable where both the key AND value is `undefined`, `"undefined"` or `null`. So I think its safe.

Kenny's right, `typeof key === 'undefined'` won't work reliably here because it gets converted to a String.

PR #693, opened 2016-07-12
Adjust multiple-form headings

jfurrow:

PR #718, opened 2016-07-14
Have schema checkbox labels on the right

kennyt:
depends on https://github.com/dcos/dcos-ui/pull/693

# Before
px

# After
px

### DIFF
--- a/src/js/utils/SchemaUtil.js
+++ b/src/js/utils/SchemaUtil.js
@@ -26,13 +26,12 @@ function setLabelFromSchemaProperty(fieldName, fieldProps, isRequired, renderLab
   }
 
   if (renderLabel && fieldProps.description) {
-    label = renderLabel(fieldProps.description, label);
+    label = renderLabel(fieldProps.description, label, fieldProps.type);
   }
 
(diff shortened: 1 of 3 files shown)


mesosphere-web:
Reviewer: @philipnrmn

src/js/utils/SchemaUtil.js
```diff
@@ -26,13 +26,12 @@ function setLabelFromSchemaProperty(fieldName, fieldProps, isRequired, renderLab
   }
 
   if (renderLabel && fieldProps.description) {
-    label = renderLabel(fieldProps.description, label);
+    label = renderLabel(fieldProps.description, label, fieldProps.type);
   }
 
   // Set the label property of checkboxes if a label is defined.
-  if (fieldProps.type === 'boolean' && fieldProps.label != null
-    && fieldProps.label !== '') {
```
rcorral: Why did you remove the last two checks?

kennyt: because even if there are no `label` property on the schema, we can still have a label from the `title` property, which we set above

PR #719, opened 2016-07-14
Allow schema utils to take options

kennyt:
A much much needed change for schema utils. The functionality is the same. The only thing that changes is instead of taking an insane amount of arguments in a specific order, it now takes an options object which it will derive the needed arguments

# Before
px

# After
px


kennyt:
retest this please

rcorral:
Much better for sure 👍

PR #721, opened 2016-07-14
Ensure we call application rendered at the right time.

rcorral:


mesosphere-web:
Reviewer: @jfurrow

src/js/index.js
```diff
@@ -52,6 +52,7 @@ RequestUtil.json = function (options = {}) {
     let contents = PluginSDK.Hooks.applyFilter('applicationContents', null);
     if (contents) {
       ReactDOM.render(contents, domElement);
+      PluginSDK.Hooks.doAction('applicationRendered');
```
kennyt: i don't know if this is what you want, but `ReactDOM.render` takes a 3rd argument that is the callback for when the `render` is done. is that what we want?

rcorral: Sounds good, why not!

PR #724, opened 2016-07-15
Revert "DCOS-8430: Rename State to Status in TaskTable"

rcorral:
Reverts dcos/dcos-ui#701
This broke the Nodes UI.

PR #725, opened 2016-07-15
Revert "Revert "DCOS-8430: Rename State to Status in TaskTable""

rcorral:
Reverts dcos/dcos-ui#724

@daiyiwhale I merged and reverted your PR after I noticed that it broke the nodes UI. This is a revert of my revert. So your original code. Please fix it up!

### DIFF
--- a/src/js/components/TaskTable.js
+++ b/src/js/components/TaskTable.js
@@ -30,8 +30,8 @@ class TaskTable extends React.Component {
     return task.resources[prop];
   }
 
-  getStatusValue(state) {
-    return TaskStates[state].displayName;
+  getStatusValue(task) {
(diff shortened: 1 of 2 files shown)


src/js/components/TaskTable.js
```diff
@@ -213,7 +213,7 @@ class TaskTable extends React.Component {
           <span className={statusClass}></span>
         </div>
         <span className={statusLabelClasses}>
-          {this.getStateValue(task, prop)}
+          {this.getStatusValue(task, prop)}
```
rcorral: Do we still need to pass `prop`?

daiyiwhale: fixed!

PR #728, opened 2016-07-15
Fallback image component

mlunoe:
Fallback kicking in:

If both url fails:

The component will also handle if it receives new props, and will retry loading the image


mesosphere-web:
Reviewer: @philipnrmn

mlunoe:

src/js/components/Image.js
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    event.target.src = fallbackSrc;
+  }
+
+  render() {
+    let {props} = this;
+
+    return (
+      <img src={props.src} onError={this.onImageError} {...props} />
```
jfurrow: You don't need to specify `src={props.src}` if you're going to use the spread operator with `props`

src/js/components/Image.js
```diff
@@ -0,0 +1,37 @@
...
+
+    METHODS_TO_BIND.forEach((method) => {
+      this[method] = this[method].bind(this);
+    });
+  }
+
+  onImageError(event) {
+    let {fallbackSrc} = this.props;
+    if (!fallbackSrc || event.target.src === fallbackSrc) {
+      return;
```
jfurrow: What do you think of actually hiding the image in this case? That way we'll never display a broken image.

kennyt:
LGTM after furrow's two comments

src/js/components/Image.js
mlunoe: True!

src/js/components/Image.js
mlunoe: Yeah, I can do that!

jfurrow:
This is great @mlunoe! Glad we have this! Are you going to implement this everywhere in a separate PR?

src/js/components/Image.js
mlunoe: done

src/js/components/Image.js
mlunoe: done. Also, had an offline conversation where we tried hiding the container, but that didn't feel right, so we decided not to go with that. See the result here:

src/js/components/Image.js
```diff
@@ -0,0 +1,69 @@
...
+      imageErrorCount: 0
+    };
+
+    METHODS_TO_BIND.forEach((method) => {
+      this[method] = this[method].bind(this);
+    });
+  }
+
+  componentWillReceiveProps(nextProps) {
+    // Try again if new urls were provided
```
kennyt: should we check if it's different from the current `src`?

src/js/components/Image.js
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    // Both src and fallback failed
+    this.setState({imageErrorCount: imageErrorCount + 1});
+  }
+
+  render() {
+    let {props, state: {imageErrorCount}} = this;
+    let classes = classNames(
+      {'hidden': imageErrorCount > 1},
```
kennyt: i thought we didn't want to hide it. did you push it up?

src/js/components/Image.js
```diff
@@ -0,0 +1,69 @@
...
+
+  onImageError(event) {
+    let {props: {fallbackSrc}, state: {imageErrorCount}} = this;
+
+    if (imageErrorCount === 0) {
+      event.target.src = fallbackSrc;
+    }
+
+    // Both src and fallback failed
+    this.setState({imageErrorCount: imageErrorCount + 1});
```
jfurrow: What is the purpose of counting the number of errors?

jfurrow: Also could we add a callback, incase the parent wants to do something on error?

src/js/components/Image.js
mlunoe: I figured it didn't matter too much. The only reason why this would be called is if you change the probs, and I believe it easier to not worry about what got changed and start over, which is less error prone. There are no performance gains to be had from this.

kennyt: `componentWillReceiveProps` gets called every time it gets re-rendered. it gets called even if the props are the same

src/js/components/Image.js
mlunoe: We did want to hide the image, not the container, where ever we use that

src/js/components/Image.js
mlunoe: 1. I need to know whether both src and fallback have failed in order to tell whether to add the class 'hidden' or not.
2. We can add that if we need it. We don't need it now.

src/js/components/Image.js
mlunoe: Arh, right! I always forget that that is the case. Good catch. I updated the PR to account for that

src/js/components/Image.js
mlunoe: See the "If both url fails" in the description for an example

src/js/components/Image.js
jfurrow: UGH FINE

kennyt:
my label is already on this, but i just wanna say LGTM again

src/js/components/Image.js
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    if (src === nextProps.src && fallbackSrc !== nextProps.fallbackSrc) {
+      newSrc = nextProps.fallbackSrc;
+    }
+
+    // Try again if a new url was provided
+    if (newSrc) {
+      let image = ReactDOM.findDOMNode(this);
+      image.src = newSrc;
```
jfurrow: What's the purpose of setting `src` directly on the node? Why not a more "React" approach by setting state?

kennyt: oh that's a good point actually

mlunoe:
Passing locally:

PR #730, opened 2016-07-15
Fix failing integration test

jfurrow:
The expandable row wasn't necessarily the last row in the table. This ensures that it only selects elements which are expandable.


mesosphere-web:
Reviewer: @pierlo-upitup

PR #733, opened 2016-07-15
Extend airbnb eslint

kennyt:
Extend airbnb eslint, then overwrite all the rules that are failing. We do this so we can remove each rule one by one to make incremental changes.


mesosphere-web:
Reviewer: @aldipower

.eslintrc
```diff
@@ -56,8 +58,82 @@
       "requireParamDescription": false
     }],
 
+    // Airbnb overwrites
```
kennyt: overwriting all the rules that are currently failing from airbnb's eslint. i'll go over each rule and fix them individually if we want that particular rule

package.json
```diff
@@ -69,9 +69,12 @@
     "babel-register": "6.9.0",
     "babel-runtime": "6.9.2",
     "css-loader": "0.23.1",
-    "eslint": "2.0.0",
+    "eslint": "^2.13.1",
```
rcorral: Let's remove the `^`

PR #736, opened 2016-07-15
[CNVS Upgrade] Migrating components: Breadcrumbs

jfurrow:


mesosphere-web:
Reviewer: @mlunoe

src/styles/components/breadcrumbs/styles.less
```diff
@@ -1,144 +1,312 @@
...
-  list-style: none !important;
-  padding-top: 0; // this + container-pod-short-top = 26px;
-  padding-bottom: @base-spacing-unit * 2/3;
-  margin: 0;
-
-  li {
+  .breadcrumb {
     display: flex;
+    flex: 0 0;
+    font-size: @body-font-size;
```
ashenden: We should do a `null` check on these either manually or using the `property-variant()`.  Looks like we do it further down in the media queries, just not here.

ashenden:
Looks great.  One comment.  :ship:

src/styles/components/breadcrumbs/styles.less
jfurrow: Done!

jfurrow:
@ashenden Done!

PR #738, opened 2016-07-15
Enable eslint no useless escapes

kennyt:
We can safely remove these escapes. It doesn't have any effect on functionality. http://eslint.org/docs/rules/no-useless-escape


mesosphere-web:
Reviewer: @pierlo-upitup

PR #741, opened 2016-07-15
[CNVS Upgrade] Migrating components: Buttons

jfurrow:


mesosphere-web:
Reviewer: @MatApple

src/styles/components/buttons/styles.less
```diff
@@ -1,19 +1,12 @@
...
-
-
-
-
-
-& when (@button-enabled) {
+& when (@buttons-enabled) {
 
   .go-to-bottom-button {
     bottom: 32px;
```
ashenden: This should be variable driven and change with screen-size.  Let's either resolve here, or note somehow as a TODO.

src/styles/components/buttons/styles.less
```diff
@@ -24,18 +17,33 @@ components/buttons.less
...
+          opacity: 0.01;
+          transform: translate(-50%, -100%);
+          transition: transform 350ms ease-in, opacity 350ms ease-in;
+        }
+
       }
+
     }
 
     &.button-short {
```
ashenden: Not quite sure what the intent of `button-short` is and where the the padding values are being derived from.  Seems very arbitrary.

src/styles/components/buttons/styles.less
```diff
@@ -49,48 +57,45 @@ components/buttons.less
       &.button-fixed-width-small {
         width: @base-spacing-unit * 3;
```
ashenden: Seems like this fixed width could create some issues on certain screen-sizes vs. something more fluid.

src/styles/components/buttons/styles.less
```diff
@@ -49,48 +57,45 @@ components/buttons.less
...
-      opacity: 0.01;
-      transform: translate(-50%, -100%);
-      transition: transform 350ms ease-in, opacity 350ms ease-in;
     }
+
   }
 
   .button-collection {
 
     & > .dropdown {
```
ashenden: This another one we should circle back around on.  Seems like there is a style-hack happening here as result of either the component being implemented incorrectly in DOM or we need to modify the source `.button-collection` object to fix whatever issue this is trying to solve.  Need more understanding.

src/styles/components/buttons/styles.less
```diff
@@ -102,149 +107,132 @@ components/buttons.less
...
-  Button Group
+    }
 
-  --------------------------------------------------------------------------------
-  ----------------------------------------------------------------------------- */
+  }
 
   .button-group {
-
     margin-bottom: @base-spacing-unit * 0.5;
```
ashenden: We should not need this style if we are using `.button-collection` to wrap a `.button-group`.

src/styles/components/buttons/styles.less
```diff
@@ -102,149 +107,132 @@ components/buttons.less
...
 
     .dot {
```
ashenden: The `.dot` and `.badge` styles can be removed now, as the new CNVS applies consistent spacing between child elements of a button.  NOTE: the text in the button needs to be wrapped in `<span>` tag for this to work.

src/styles/components/buttons/styles.less
```diff
@@ -102,149 +107,132 @@ components/buttons.less
...
 
   .button-align-content {
-
     align-items: center;
     display: flex;
     flex-flow: row nowrap;
-
   }
 
   .button-split-content {
```
ashenden: This is pretty cool.  We should bring this into CNVS.

src/styles/components/buttons/styles.less
```diff
@@ -102,149 +107,132 @@ components/buttons.less
...
-
     }
 
     .badge {
-
       margin-left: @base-spacing-unit * 0.25;
-
     }
 
     &-directions {
```
ashenden: Another one I'm concerned is a quick hack vs. fixing the problem in the CNVS button object directly.

src/styles/components/buttons/variables.less
```diff
@@ -0,0 +1,11 @@
+/* -----------------------------------------------------------------------------
+--------------------------------------------------------------------------------
+--------------------------------------------------------------------------------
+banner-plugin/variables.less
```
ashenden: Incorrect path.

src/styles/components/buttons/variables.less
```diff
@@ -0,0 +1,11 @@
...
+--------------------------------------------------------------------------------
+--------------------------------------------------------------------------------
+----------------------------------------------------------------------------- */
+
+
+
+@buttons-enabled:                                                               true;
```
ashenden: We already have this set to true in `styles/vendors/cnvs/components/button`.  Thoughts on whether we repeat here for components of the same name?

ashenden:
Looks great.  Added a handeful of inline comments.  Some need to resolved, others are more suggestions/thought-starters and things to consider as a possible second pass.

src/styles/components/buttons/variables.less
jfurrow: I would suggest overriding variables for each component, because it's possible that the author of CNVS may change the variable name, or decide to ship with buttons disabled.

src/styles/components/buttons/styles.less
jfurrow: Added a `TODO`!

src/styles/components/buttons/styles.less
jfurrow: Added a `TODO`!

src/styles/components/buttons/styles.less
jfurrow: Added a `TODO`!

src/styles/components/buttons/styles.less
jfurrow: Added a `TODO`!

src/styles/components/buttons/styles.less
jfurrow: Added a `TODO`!

src/styles/components/buttons/styles.less
jfurrow: Added a `TODO`!

src/styles/components/buttons/styles.less
jfurrow: Added a `TODO`!

src/styles/components/buttons/styles.less
jfurrow: Added a `TODO`!

src/styles/components/buttons/variables.less
jfurrow: Fixed!

jfurrow:
@ashenden I added `TODO`s. I agree with all of your assessments, just need to set aside time to address the issues.

PR #742, opened 2016-07-15
[CNVS Upgrade] Migrating components: Caret

jfurrow:


ashenden:
Looks good to go.  Do we even need this component though given it's used almost exclusively with `dropdown` elements which already draw these carets?

jfurrow:
I think this is only used in Table headers.

PR #748, opened 2016-07-18
[CNVS Upgrade] Migrating components: Charts 

jfurrow:


src/styles/components/charts/styles.less
```diff
@@ -1,342 +1,580 @@
...
-      fill: @white;
-      line-height: 1;
     }
 
-    .unit-label {
-      fill: @purple;
-      &.path-color-0 {
-        color: @purple;
-      }
+    & when (@typography-enabled) {
```
ashenden: What is the point of this check?  Looking at the applied styles, none seem to be dependent on Typography from CNVS.

src/styles/components/charts/styles.less
```diff
@@ -1,342 +1,580 @@
...
+  /**********************************/
+  /*  Specific to Nodes Grid View   */
+  /**********************************/
+  .nodes-grid-dials {
     display: flex;
-    justify-content: center;
-    margin-top: 10px;
+    flex-flow: row wrap;
+    justify-content: space-around;
+    margin-left: -(ceil(@base-spacing-unit * 0.3333));
```
ashenden: All these `* 0.3333` should probably be replaced with something like `* (1/3)` or `/ 3` to avoid the 4 point decimal cutoff and let the compiler do the exact math.

src/styles/components/charts/styles.less
```diff
@@ -1,342 +1,580 @@
...
+
+& when (@charts-enabled) and (@layout-screen-small-enabled) {
+
+  @media (min-width: @layout-screen-small-min-width) {
+
+    .chart {
+
+      .unit-bordered-horizontal-mini {
+        border-right: 1px solid @chart-axis-color;
+        margin-top: @base-spacing-unit-screen-small * 1/4;
```
ashenden: My pref would be to have this be `* 0.25` instead of the visible fraction.

ashenden:
Looks good overall.  Not quite sure everything going on here.  Assume it's stable.  We should do another pass on "charts" in general during 1.9 and see how we can make it better universally.

src/styles/components/charts/styles.less
jfurrow: Yeah, it's strange. It should be removed — I'll add a TODO.

src/styles/components/charts/styles.less
jfurrow: I opted for `1/3` here.

src/styles/components/charts/styles.less
jfurrow: I'll bring this up to the frontend team in our forthcoming code styling discussion. I'd vote for fractions everywhere :)

jfurrow:
@ashenden I totally agree, this is very confusing and there are a lot of oddly specific SVG styles. It needs a refactor.

PR #749, opened 2016-07-18
[CNVS Upgrade] Migrating components: Code Blocks

jfurrow:


src/styles/components/code/variables.less
```diff
@@ -0,0 +1,28 @@
...
+
+
+
+/* -----------------------------------------------------------------------------
+--------------------------------------------------------------------------------
+Code Block Colors
+--------------------------------------------------------------------------------
+----------------------------------------------------------------------------- */
+
+@code-block-pretty-print-border-color:                                          color-lighten(@neutral, 10);
```
ashenden: Is the hyphen between `pretty` and `print` intentional here?

ashenden:
Generally speaking, we can probably get rid of this component, as the new code element in CNVS is heavily improved.  Worth considering if we can strip this one of the roster.

Overall looks good though.

src/styles/components/code/variables.less
jfurrow: Nope, good catch! Fixed.

jfurrow:
Awesome!

PR #750, opened 2016-07-18
Fix Stack overflow for case where statusbar is scaled

MatApple:
Fixes Stack overflow bug when stealing portions in the case where there is no portion to steal from.

### DIFF
--- a/src/js/components/StatusBar.js
+++ b/src/js/components/StatusBar.js
@@ -8,43 +8,44 @@ const GROW_RATIO = 0.07;
 const MIN_SIZE = 7; // (% of statusBar width)
 
 function stealPortion(barSizes, indexesLessThanThreshold, unassignedPortion) {
-  if (indexesLessThanThreshold.length) {
-    // Try to steal some width from the largest portion
-    // and give it to the first portion in indexesLessThanThreshold
-    let maxSize = 0;
-    let maxIndex = 0;
-    // Find new Max (could be different after we've stolen from a portion already)
-    barSizes.forEach(function (relativeSize, index) {
-      if (relativeSize > maxSize) {
-        maxSize = relativeSize;
-        maxIndex = index;
-      }
-    });
-    // Possible we have unassigned portion that we should steal from first
-    if (unassignedPortion > maxSize) {
-      maxSize = unassignedPortion;
-      maxIndex = -1;
-    }
-
-    let currentGrowthIndex = indexesLessThanThreshold[0];
-    let sizeToSteal = maxSize * GROW_RATIO;
-    // Reassign the portion
-    barSizes[currentGrowthIndex] += sizeToSteal;
-    if (maxIndex > -1) {
-      barSizes[maxIndex] -= sizeToSteal;
-    } else {
-      unassignedPortion -= sizeToSteal;
-    }
-
-    if (barSizes[currentGrowthIndex] > MIN_SIZE) {
-      // Done with growing this portion. Otherwise it stays on the stack
-      // and will get another % added to it from the largest portion
-      // calculated in the next iteration.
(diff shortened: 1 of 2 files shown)


mesosphere-web:
Reviewer: @pierlo-upitup

MatApple:
retest this please

src/js/components/StatusBar.js
```diff
@@ -34,7 +43,7 @@ function stealPortion(barSizes, indexesLessThanThreshold) {
       indexesLessThanThreshold = indexesLessThanThreshold.slice(1);
     }
     // Recurse till we have grown everything above threshold
-    stealPortion(barSizes, indexesLessThanThreshold);
+    stealPortion(barSizes, indexesLessThanThreshold, unassignedPortion);
```
mlunoe: Can we get some tests to make sure this problem doesn't re-emerge?

src/js/components/StatusBar.js
```diff
@@ -7,7 +7,7 @@ const GROW_RATIO = 0.07;
 // Grow anything below this threshold by stealing from the biggest portion
 const MIN_SIZE = 7; // (% of statusBar width)
 
-function stealPortion(barSizes, indexesLessThanThreshold) {
+function stealPortion(barSizes, indexesLessThanThreshold, unassignedPortion) {
```
mlunoe: Maybe we could also move this to a Util function?

MatApple: I think it would be premature to move into a Util because it's fairly specific to this Component.

src/js/components/StatusBar.js
MatApple: I'll write some tests

src/js/components/StatusBar.js
mlunoe: Up to you! It was just a suggestion to make it easier for you to test and also remove clutter from the component ☺️

MatApple: Thanks for the suggestion 👍

src/js/components/StatusBar.js
MatApple: done!

src/js/components/StatusBar.js
```diff
@@ -7,7 +7,7 @@ const GROW_RATIO = 0.07;
...
   if (indexesLessThanThreshold.length) {
```
mlunoe: If we do like so, like we do in other places, we can unindent the whole function:
```js
if (!indexesLessThanThreshold.length) {
  return barSizes;
}
```

src/js/components/StatusBar.js
```diff
@@ -20,12 +20,21 @@ function stealPortion(barSizes, indexesLessThanThreshold) {
...
+      maxIndex = -1;
+    }
 
     let currentGrowthIndex = indexesLessThanThreshold[0];
     let sizeToSteal = maxSize * GROW_RATIO;
     // Reassign the portion
     barSizes[currentGrowthIndex] += sizeToSteal;
-    barSizes[maxIndex] -= sizeToSteal;
+    if (maxIndex > -1) {
+      barSizes[maxIndex] -= sizeToSteal;
```
mlunoe: This was not introduced by here, so this is unrelated to this PR, however, this function modifies barSizes. I think it would be much nicer if this function could return new segments based on what is given.

mlunoe:
The two last comments shouldn't hold up the PR, so I gave it my label

src/js/components/StatusBar.js
MatApple: Thanks for the suggestion but I don't think it's worth it to recreate a new Array on every iteration :)

src/js/components/StatusBar.js
MatApple: done!

PR #755, opened 2016-07-19
Pad out tooltip content on Services StatusBar

MatApple:
Before:

After:

### DIFF
--- a/src/js/components/HealthBar.js
+++ b/src/js/components/HealthBar.js
@@ -27,7 +27,7 @@ class HealthBar extends React.Component {
       let classSet = classNames(HealthBarStates[task].className, 'dot icon');
 
       return (
-        <div key={index}>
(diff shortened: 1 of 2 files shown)


mesosphere-web:
Reviewer: @aldipower

jfurrow:
Sorry I'm too late but I think this should be handled differently. I think we should use the `@base-spacing-unit` varaible here. Also, instead of adding padding to the top and bottom of each `.tooltip-line-item` element, we should apply margin to the bottom of those elements, like so:
```less
.tooltip-line-item {
  margin-bottom: @base-spacing-unit / 6; // this might be the wrong denomenator

  &:last-child{
    margin-bottom: 0;
  }

}
```

Just for future reference. I'll tackle this in the CNVS upgrade.

PR #757, opened 2016-07-19
Add Help tooltip to JSON editor in Service Modal

MatApple:
[DCOS-8445](https://mesosphere.atlassian.net/browse/DCOS-8445). Until we figure out how to do contextual help globally, this is what we've got. Currently configure the docs endpoint in Config.js. Perhaps these docs will end up living in dcos.io at some point.

### DIFF
--- a/src/js/components/modals/ServiceFormModal.js
+++ b/src/js/components/modals/ServiceFormModal.js
@@ -1,6 +1,6 @@
 import Ace from 'react-ace';
 import mixin from 'reactjs-mixin';
-import {Modal} from 'reactjs-components';
+import {Modal, Tooltip} from 'reactjs-components';
 import React from 'react';
 import {StoreMixin} from 'mesosphere-shared-reactjs';
 
@@ -8,6 +8,8 @@ import 'brace/mode/json';
 import 'brace/theme/monokai';
 import 'brace/ext/language_tools';
 
+import Config from '../../config/Config';
+import Icon from '../Icon';
 import MarathonStore from '../../stores/MarathonStore';
 import ServiceForm from '../ServiceForm';
 import Service from '../../structs/Service';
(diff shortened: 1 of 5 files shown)


mesosphere-web:
Reviewer: @philipnrmn

pierlo-upitup:
Thanks for this, however I'm seeing this issue on my device:

tooltip

MatApple:
Thanks @pierlo-upitup, I couldn't reproduce that but I've made some changes so that should never happen.

src/js/config/Config.js
```diff
@@ -11,6 +11,7 @@ var Config = {
   secretsAPIPrefix: '/secrets/v1',
   delayAfterErrorCount: 5,
   documentationURI: 'https://dcos.io/docs',
+  marathonJSONDocsURI: 'https://mesosphere.github.io/marathon/docs/generated/api.html#v2_apps_post',
```
rcorral: The URI isn't actually to the home of the docs, but to a modal in the docs. Maybe we should have a "generic" URI here. Also, maybe we shouldn't put it in the config because it's completely related to the services UI.

src/js/components/modals/ServiceFormModal.js
```diff
@@ -373,16 +375,38 @@ class ServiceFormModal extends mixin(StoreMixin) {
     let {jsonDefinition, jsonMode, service} = this.state;
 
     if (jsonMode) {
+      let toolTipContent = (
+        <div>
+          Use the JSON editor to enter Marathon Application definitions manually.
+          {' '}
+          <a href={Config.marathonJSONDocsURI} target="_blank">
+            Read more here.
```
rcorral: Maybe this should say `See API docs` or `See API documentation`.
What do you think?

MatApple:
@pierlo-upitup that problem is fixed now

src/js/components/modals/ServiceFormModal.js
```diff
@@ -373,16 +375,38 @@ class ServiceFormModal extends mixin(StoreMixin) {
     let {jsonDefinition, jsonMode, service} = this.state;
 
     if (jsonMode) {
+      let toolTipContent = (
+        <div>
+          Use the JSON editor to enter Marathon Application definitions manually.
+          {' '}
+          <a href={`${Config.marathonDocsURI}generated/api.html#v2_apps_post`} target="_blank">
+            Read more here.
```
rcorral: Maybe this should say See API docs or See API documentation.
What do you think?

PR #758, opened 2016-07-19
DCOS-8371: Add detailed error message on scale

Poltergeist:
screen shot 2016-07-19 at 11 08 55

This shows a more detailed error on scale if `object is not valid` is the error message which is provided by marathon.

### DIFF
--- a/src/js/components/modals/ServiceScaleFormModal.js
+++ b/src/js/components/modals/ServiceScaleFormModal.js
@@ -67,10 +67,9 @@ class ServiceScaleFormModal extends mixin(StoreMixin) {
 
   onMarathonStoreServiceEditError({message:errorMsg = '', details: details}) {
     this.resetState();
-    let isErrorMsgToShort = errorMsg.toLowerCase() === 'object is not valid' &&
-      details.length !== 0;
+    let hasDetails = details && details.length !== 0;
 
-    if (isErrorMsgToShort) {
+    if (hasDetails) {
       this.setState({
         errorMsg: details.reduce(function (memo, error) {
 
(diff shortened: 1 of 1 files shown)


mesosphere-web:
Reviewer: @MatApple

src/js/components/modals/ServiceScaleFormModal.js
```diff
@@ -65,8 +65,22 @@ class ServiceScaleFormModal extends mixin(StoreMixin) {
     this.props.onClose();
   }
 
-  onMarathonStoreServiceEditError({message:errorMsg}) {
+  onMarathonStoreServiceEditError({message:errorMsg = '', details: details}) {
     this.resetState();
+    let isErrorMsgToShort = errorMsg.toLowerCase() === 'object is not valid' &&
```
orlandohohmeier: Could you rename `isErrorMsgToShort` to something meaningful? #nit

orlandohohmeier:
Minor nit. Rest looks good to me.

src/js/components/modals/ServiceScaleFormModal.js
MatApple: I feel like this could be brittle checking for this exact string. What if we checked for `details` and tried to build the error from that first. If that doesn't exist, then we fall back to the generic `message` key? You guys might have more insight as to why we shouldn't do that though..?

Poltergeist: Changed this to `details&&details.length !== 0` and renamed it to `hasDetails`.

orlandohohmeier:
Feel free to merge once test past.

PR #761, opened 2016-07-19
[CNVS Upgrade] Moving Source Sans to hosted instead of Google CDN served

ashenden:


jfurrow:
We should also add a WOFF format for the future, I'll add this in just a few minutes.

PR #762, opened 2016-07-19
DCOS-8067 - Sort sidebar filter by label order

aldipower:
This sorts the sidebar filters by the order of the labels instead of the filter values.

filterorder

### DIFF
--- a/src/js/components/SidebarFilter.js
+++ b/src/js/components/SidebarFilter.js
@@ -167,7 +167,7 @@ class SidebarFilter extends mixin(QueryParamsMixin) {
             this.getFormLabel(filterLabels[filterLabel], filterLabel),
           labelClass: labelClassSet
         };
-    });
+      });
   }
 
   getHealthNodes() {
--- a/src/js/components/SidebarFilter.js
+++ b/src/js/components/SidebarFilter.js
@@ -139,30 +139,34 @@ class SidebarFilter extends mixin(QueryParamsMixin) {
     let {filterLabels, filterValues} = this.props;
     let {selectedNodes} = this.state;
 
-    return Object.keys(filterValues).map(filterValue => {
-      let value = filterValues[filterValue].toString();
-      let checked = selectedNodes.indexOf(value) > -1;
+    return Object.keys(filterLabels)
+      .filter(filterLabel => {
+        return filterValues[filterLabel] != null;
+      })
+      .map(filterLabel => {
+        let value = filterValues[filterLabel];
+        let checked = selectedNodes.indexOf(value.toString()) > -1;
 
-      let labelClassSet = classNames(
-        'side-list-item form-row-element form-element-checkbox inverse',
-        'row row-flex flex-align-items-center vertical-center flush clickable',
-        {
-          'filter-active': this.getCountByValue(filterValue) > 0,
-          'filter-checked': checked
-        }
-      );
+        let labelClassSet = classNames(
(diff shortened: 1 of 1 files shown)


mesosphere-web:
Reviewer: @mlunoe

src/js/components/SidebarFilter.js
```diff
@@ -139,30 +139,34 @@ class SidebarFilter extends mixin(QueryParamsMixin) {
...
-        value: checked,
-        fieldType: 'checkbox',
-        formElementClass: 'form-row-element checkbox flush',
-        name: filterValue,
-        label:
-          this.getFormLabel(filterLabels[filterValue], filterValue),
-        labelClass: labelClassSet
-      };
+    return Object.keys(filterLabels)
+      .filter(filterLabel => {
```
mlunoe: You are not using context here, so let's use a regular function

src/js/components/SidebarFilter.js
```diff
@@ -139,30 +139,34 @@ class SidebarFilter extends mixin(QueryParamsMixin) {
...
+        return filterValues[filterLabel] != null;
+      })
+      .map(filterLabel => {
```
mlunoe: #nit We also usually have parenthesis around the arguments

mlunoe:
Not sure I understand how this is sorting by label order? Could you explain?

aldipower:
@mlunoe It iterates over `filterLabels` instead of the `filterValues` now. That is the whole trick.

src/js/components/SidebarFilter.js
aldipower: Thanx, I've forgot about that.

src/js/components/SidebarFilter.js
```diff
@@ -136,34 +138,43 @@ class SidebarFilter extends mixin(QueryParamsMixin) {
...
+      })
+      .map((filterLabel) => {
+        let value = filterValues[filterLabel];
+        let checked = selectedNodes.indexOf(value.toString()) > -1;
+
+        let labelClassSet = classNames(
+          'side-list-item form-row-element form-element-checkbox inverse',
+          'row row-flex flex-align-items-center vertical-center flush clickable',
+          {
+            'filter-active': this.getCountByValue(filterLabel) > 0,
```
rcorral: Shouldn't you be passing `filterValue` here and not the `filterLabel`?

rcorral:
@aldipower looks good after my comment is answered.

Also, looking at the commits, is that a rebase gone wrong? I"m not sure why there's a merge commit in there.

src/js/components/SidebarFilter.js
aldipower: Yes the function naming is a little bit confusing, but it is working correctly here.
`filterLabel` is resolving to the actual value inside of the function.

```JavaScript
  getCountByValue(filterValue) {
    let {props} = this;
    let count = props.countByValue[props.filterValues[filterValue]];

    if (count == null) {
      return 0;
    }

    return count;
  }
```

`props.fitlerValues` contains this e.g.
```JavaScript
Object {UNHEALTHY: 0, HEALTHY: 1, IDLE: 2, NA: 3}
```
Where the `filterLabel` is this:
`"IDLE"`

aldipower:
@rcorral Thanx for looking over it. I've answered your comment, hope this helps.

There is no wrong rebase here, just another PR got merged into this, because they were dependent on each other.

Poltergeist:

PR #768, opened 2016-07-19
Fix docs version formatting

MatApple:
Make sure we only route to major doc versions. No patch versions or prerelease (anything after `-`)


mesosphere-web:
Reviewer: @aldipower

PR #774, opened 2016-07-19
Remove padding from non-scrolling portion of TabForm modal

jfurrow:
Before:

After:

Integration tests:

### DIFF
--- a/src/js/components/TabForm.js
+++ b/src/js/components/TabForm.js
@@ -161,7 +161,7 @@ class TabForm extends mixin(InternalStorageMixin) {
     let formRowClass = this.props.formRowClass;
 
     let panels = Object.keys(multipleDefinition).map((formKey, i) => {
-      let panelClassSet = classNames('form', {
+      let formPanelClassSet = classNames('form-panel', {
         'hidden': currentTab !== formKey
       });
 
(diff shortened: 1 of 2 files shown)


mesosphere-web:
Reviewer: @kennyt

src/js/components/TabForm.js
```diff
@@ -161,7 +161,7 @@ class TabForm extends mixin(InternalStorageMixin) {
     let formRowClass = this.props.formRowClass;
 
     let panels = Object.keys(multipleDefinition).map((formKey, i) => {
-      let panelClassSet = classNames('form', {
+      let formPanelClassSet = classNames('form-panel', {
         'hidden': currentTab !== formKey
```
jfurrow: With this change, we're hding the whole `form-panel` element instead of its child.

mlunoe:

src/js/components/TabForm.js
```diff
@@ -173,9 +173,9 @@ class TabForm extends mixin(InternalStorageMixin) {
       let formRowClassSet = classNames('row', formRowClass, formKey);
 
       return (
-        <div key={i} className="form-panel">
+        <div key={i} className={formPanelClassSet}>
           <Form
-            className={panelClassSet}
+            className="form"
```
mlunoe: Do we need this 'form' to be set? Isn't that default?

mlunoe:
Looks and works great! My one minimal comment should not hold up this PR :shipit:

src/js/components/TabForm.js
jfurrow: There's no default set that I'm aware of; am I missing something?

mlunoe: Had a conversation offline with @jfurrow. It seem that this is the only place we are using a `form` class and it is only being defined in `multiple-form.less`, so @jfurrow is now exploring a solution where we use `form`-tag instead

jfurrow: Done!

mlunoe:
Awesome! Looks good!

PR #776, opened 2016-07-20
Update time component

mlunoe:

### DIFF
--- a/src/js/components/ServiceDetailDebugTab.js
+++ b/src/js/components/ServiceDetailDebugTab.js
@@ -1,9 +1,9 @@
 import React from 'react';
 
-import TimeAgo from './TimeAgo';
 import DescriptionList from './DescriptionList';
 import Service from '../structs/Service';
 import TaskStatsTable from './TaskStatsTable';
+import TimeAgo from './TimeAgo';
 
 class ServiceDetailDebugTab extends React.Component {
   getValueText(value) {
@@ -32,9 +32,11 @@ class ServiceDetailDebugTab extends React.Component {
       'State': this.getValueText(state),
       'Message': this.getValueText(message),
       'Host': this.getValueText(host),
-      'Timestamp': <span>
-        {timestamp} (<TimeAgo time={new Date(timestamp)} />)
-      </span>,
+      'Timestamp': (
+        <span>
+          {timestamp} (<TimeAgo time={new Date(timestamp)} />)
+        </span>
+      ),
       'Version': <span>{version} (<TimeAgo time={new Date(version)} />)</span>
     };
 
--- a/src/js/components/ServiceDetailDebugTab.js
+++ b/src/js/components/ServiceDetailDebugTab.js
@@ -1,6 +1,6 @@
 import React from 'react';
 
-import DateUtil from '../utils/DateUtil';
+import TimeAgo from './TimeAgo';
 import DescriptionList from './DescriptionList';
 import Service from '../structs/Service';
 import TaskStatsTable from './TaskStatsTable';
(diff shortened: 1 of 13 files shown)


mesosphere-web:
Reviewer: @Poltergeist

src/js/components/TimeAgo.js
```diff
@@ -0,0 +1,65 @@
...
+class TimeAgo extends React.Component {
+  constructor() {
+    super(...arguments);
+
+    METHODS_TO_BIND.forEach((method) => {
+      this[method] = this[method].bind(this);
+    });
+
+    this.state = {
+      interval: global.setInterval(this.updateTime, this.props.updateInterval)
```
rcorral: I think this needs to change dynamically based on the time diff. It won't make sense to change every 30 seconds when something happened hours or days ago. Or even minutes. It would be nice if the component was smart enough to change this dynamically!

mlunoe: done

MatApple:
This is a cool component :) Nice work @mlunoe!

src/js/components/ServiceDetailDebugTab.js
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 
-import DateUtil from '../utils/DateUtil';
+import TimeAgo from './TimeAgo';
```
jfurrow: :abcd:

src/js/components/ServiceDetailDebugTab.js
```diff
@@ -27,17 +27,15 @@ class ServiceDetailDebugTab extends React.Component {
...
     let taskFailureValueMapping = {
       'Task ID': this.getValueText(taskId),
       'State': this.getValueText(state),
       'Message': this.getValueText(message),
       'Host': this.getValueText(host),
-      'Timestamp': <span>{timestamp} ({timeStampText})</span>,
-      'Version': <span>{version} ({DateUtil.msToRelativeTime(version)})</span>
+      'Timestamp': <span>
+        {timestamp} (<TimeAgo time={new Date(timestamp)} />)
+      </span>,
```
jfurrow: Can we add parentheses around this to be consistent with other areas of the app? So like this:
```javascript
'Timestamp': (
  ...jsx...
),
```

src/js/components/TimeAgo.js
```diff
@@ -0,0 +1,100 @@
+import React from 'react';
+
+import DateUtil from '../utils/DateUtil';
+import Util from '../utils/Util';
+
+const SECOND = 1000;
+const MINUTE = 60 * 1000;
+const HOUR = 60 * 60 * 1000;
+const DAY = 24 * 60 * 60 * 1000;
```
rcorral: could do this:
```
const SECOND = 1000;
const MINUTE = 60 * SECOND;
const HOUR = 60 * MINUTE;
const DAY = 24 * HOUR;
```

src/js/components/TimeAgo.js
```diff
@@ -0,0 +1,100 @@
...
+  }
+}
+
+TimeAgo.defaultProps = {
+  dontUpdate: false,
+  suppressSuffix: false
+};
+
+TimeAgo.propTypes = {
+  dontUpdate: React.PropTypes.bool,
```
rcorral: What do you think about calling this param `autoUpdate` and have it be truthy from the start. It feels a little more natural IMO.

src/js/components/TimeAgo.js
```diff
@@ -0,0 +1,100 @@
...
+      interval: global.setTimeout(this.updateTime, this.getUpdateInterval())
+    });
+  }
+
+  render() {
+    let {prefix, suppressSuffix, time} = this.props;
+    let relativeTime = DateUtil.msToRelativeTime(time, suppressSuffix);
+    if (time == null) {
+      relativeTime = 'N/A';
+    }
```
rcorral: If we do `let relativeTime = 'N/A'` first and then do `if (time) { DateUtil... }` then we don't have to fail the call to DateUtil because there's no time.

rcorral:
Looks pretty good! Just some small comments and I think it's good!

src/js/components/TimeAgo.js
mlunoe: Yes! Positive check, I like it! Done!

src/js/components/TimeAgo.js
mlunoe: Good point! Done!

src/js/components/TimeAgo.js
mlunoe: Yes, good point! Done

PR #777, opened 2016-07-20
DCOS-8087: hide docker volumes if no container image is provided

Poltergeist:
This is mostly @MatApple work, thanks again for jumping on this one. 
What it does is removing the docker volumes from the volumes section if no docker image is provided. 
docker-volumes mov

### DIFF
--- a/src/js/components/ServiceForm.js
+++ b/src/js/components/ServiceForm.js
@@ -1,4 +1,5 @@
 import {Hooks} from 'PluginSDK';
+import Classnames from 'classnames';
 import React from 'react';
 
 import FormUtil from '../utils/FormUtil';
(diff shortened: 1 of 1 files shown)


mesosphere-web:
Reviewer: @orlandohohmeier

src/js/components/ServiceForm.js
```diff
@@ -332,7 +361,8 @@ class ServiceForm extends SchemaForm {
           return port;
         });
       }
-      this.internalStorage_set({model});
+      let {dockerVolumesDefinition = null} = this.internalStorage_get();
```
philipnrmn: 💯 destructuring defaults

philipnrmn:
Nice!

src/js/components/ServiceForm.js
```diff
@@ -233,8 +233,37 @@ class ServiceForm extends SchemaForm {
...
   updateDefinitions() {
     let {model} = this.internalStorage_get();
-    let {networking} = this.multipleDefinition;
+    let {networking, volumes} = this.multipleDefinition;
+    let {containerSettings} = model;
 
+    let showDockerVolumes = containerSettings &&
+      containerSettings.image != null && containerSettings.image !== '';
+
+    let endIndex = -1;
```
orlandohohmeier: I really don't like that we rely on these indices, but I can't think of a better solution without refactoring huge parts of form. Let's make sure we get rid of these once we refactor the schema form.

orlandohohmeier:
LGTM! 🚢

PR #779, opened 2016-07-20
DCOS-8165: Introduce way to stop an active job run

pierlo-upitup:
This introduces a way to stop an active Job Run from the list.

**PLEASE NOTE**: the current Metronome API does not allow us to POST multiple Job Run IDs at once. To workaround this limitation, me and @orlandohohmeier decided to maintain the checkbox approach, but to limit the selection to one item (see screenshot). Once the API will add support for that (follow **DCOS-8763**), it will be trivial to introduce the multiple-items behaviour.
For this solution to work, a new boolean property `allowMultipleSelect` has been added to the `CheckboxTable`. 

The following is known but out of scope to keep the PR lean(er):

1) In case of error when submitting the request, the modal simply gets closed. What is the pattern here? Close the current dialog and show another one with the error message? @leemunroe 
2) `CheckboxTable` currently does not provide a simple way to exclude certain items from selection. In this case, we should not let users select a non-active job run
3) the FilterBar header jumps when the `Stop` button is shown. @jfurrow @leemunroe any ideas how to fix that?

screenflick movie 72

Relevant tests are passing:

image


mesosphere-web:
Reviewer: @aldipower

pierlo-upitup:
retest this please

src/js/components/ExpandingTable.js
```diff
@@ -17,7 +17,7 @@ class ExpandingTable extends React.Component {
   }
 
   expandRow(row) {
-    let {expandedRows} = this.state;
+    let {expandedRows} = Object.assign({}, this.state);
```
kennyt: why is this necessary?

src/js/components/ExpandingWithCheckboxTable.js
```diff
@@ -0,0 +1,28 @@
...
+
+class ExpandingWithCheckboxTable extends React.Component {
+  getColGroup() {
+    return (
+      <colgroup>
+        <col style={{width: '40px'}} />
+        <col />
+        <col style={{width: '120px'}} className="hidden-mini" />
+        <col style={{width: '160px'}} className="hidden-mini" />
+        <col style={{width: '160px'}} className="hidden-mini" />
```
kennyt: these seem very specific for a component like this, which is pretty general

src/js/pages/jobs/JobRunHistoryTable.js
```diff
@@ -233,24 +292,36 @@ class JobRunHistoryTable extends React.Component {
...
+            name="Run"
+            onReset={function () {}}
+            totalLength={totalRunCount} />
+          {this.getStopButton(hasCheckedTasks)}
+        </FilterBar>
+        <ExpandingWithCheckboxTable
+          allowMultipleSelect={false}
           className="job-run-history-table table table-hover inverse table-borderless-outer table-borderless-inner-columns flush-bottom"
           childRowClassName="job-run-history-table-child"
           columns={this.getColumns()}
```
kennyt: right here, you're passing columns into `ExpandingWithCheckboxTable` but then inside of that component, it has it's own `getColumns`. i think we can safely remove one right?

leemunroe:
`In case of error when submitting the request, the modal simply gets closed. What is the pattern here? Close the current dialog and show another one with the error message?` I'm not sure actually. How do we deal with this on other views? e.g. I delete a user and it fails

rcorral:
We usually don't close modals until we get a response from the server. In some places we'll show the error inside of the current modal, in other cases (rare) we show another modal on top with the error.

src/js/components/ExpandingTable.js
pierlo-upitup: https://github.com/dcos/dcos-ui/blob/master/src/js/components/ExpandingTable.js#L24-L30

> NEVER mutate state directly

https://facebook.github.io/react/docs/component-api.html

src/js/components/ExpandingWithCheckboxTable.js
pierlo-upitup: Agreed – I will refactor these out.

src/js/pages/jobs/JobRunHistoryTable.js
pierlo-upitup: Yes, I will do it this way so that we don't pollute the abstracted component with ad-hoc solutions. Thanks.

pierlo-upitup:
@rcorral @leemunroe I will create an issue with the remaining bits so we can come back to those later.

src/js/components/ExpandingTable.js
kennyt: but let's say we do something like `expandedRows.hello = 10` i think we're still mutating state. `expandedRows` in both cases would be the same object. the way around this would be to do `Object.assign({}, this.state.expandedRows)`, unless i'm missing something

src/js/components/ExpandingWithCheckboxTable.js
kennyt: honestly, after you take this method out, will this component even be necessary? it'll literally only have `render`

leemunroe:
@pierlo-upitup For the `Stop` button jumping, were we planning to add this Failed/Success filter as per the designs? If so I'd expect this button to live in this row and it's likely not an issue then.

image

src/js/components/ExpandingWithCheckboxTable.js
pierlo-upitup: Yeah. My initial plans were more ambitious in terms of what I wanted to achieve with this. But without any further refactorings on the CheckboxTable and ExpandingTable, this "composite" component doesn't make much sense anymore. I will move this to the JobRunHistoryTable itself to keep things simple.

src/js/components/ExpandingTable.js
pierlo-upitup: You're totally right! Fixed

pierlo-upitup:
Addressed @kennyt's comments – ready for another pass

aldipower:
My personal test in Chrome and FF says: :+1: 

The code also LGTM!

Good work man!

aldipower:
Unfortunately the integration tests for the jobs are failing.

pierlo-upitup:
True, will fix.

pierlo-upitup:
retest this please

src/js/pages/jobs/JobRunHistoryTable.js
```diff
@@ -3,37 +3,68 @@ import {Link} from 'react-router';
...
   getColGroup() {
     return (
       <colgroup>
-        <col style={{width: '40%'}} />
-        <col style={{width: '20%'}} />
-        <col style={{width: '20%'}} />
-        <col style={{width: '20%'}} />
+        <col style={{width: '40px'}} />
+        <col />
+        <col style={{width: '120px'}} className="hidden-mini" />
```
orlandohohmeier: Remove `hidden-mini` class

pierlo-upitup: Done!

PR #789, opened 2016-07-20
Fix service table filtering

kennyt:
the filtering now matches marathon-ui exactly. please try it out

before, the sidebar filtering would include groups. we no longer do that (unless you're doing a global search with the search string)

we also search through all the children now

### DIFF
--- a/src/js/components/ServiceSidebarFilters.js
+++ b/src/js/components/ServiceSidebarFilters.js
@@ -20,6 +20,10 @@ function getCountByFilters(services) {
   const volumesKey = ServiceOther.VOLUMES.key;
 
   return services.reduce(function (memo, service) {
+    if (service instanceof ServiceTree) {
+      return memo;
+    }
+
     let serviceStatus = service.getServiceStatus();
     let serviceHealth = service.getHealth();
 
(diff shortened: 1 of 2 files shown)


mesosphere-web:
Reviewer: @philipnrmn

src/js/structs/ServiceTree.js
```diff
@@ -115,16 +115,34 @@ module.exports = class ServiceTree extends Tree {
       }
 
       if (filter.health != null && filter.health.length !== 0) {
-        services = services.filter(function (service) {
```
rcorral: @kennyt could you explain a bit why each one of the changes on this file were made?

src/js/components/ServiceSidebarFilters.js
```diff
@@ -20,6 +20,10 @@ function getCountByFilters(services) {
   const volumesKey = ServiceOther.VOLUMES.key;
 
   return services.reduce(function (memo, service) {
+    if (service instanceof ServiceTree) {
+      return memo;
+    }
```
kennyt: we do not want to count `groups` into the filter count at all. this is because when we use sidebar filtering, we do not show groups

src/js/structs/ServiceTree.js
```diff
@@ -115,16 +115,34 @@ module.exports = class ServiceTree extends Tree {
...
+            }
+
+            if (service.getHealth().value === parseInt(healthValue, 10)) {
+              memo.push(service);
+            }
           });
-        });
+
+          return memo;
+        }, []);
```
kennyt: we need to do a recursive `reduce` because we want to traverse all of the children in order to build up all the services that match the filter.

previously, it would only get the top-most level in the filtering, and then for groups that contain children that match, it showed.

the current implementation does not show groups at all and shows only the children that match.

PR #799, opened 2016-07-20
DCOS-8414: Recalibrate checkboxes when filtered tasks changes

daiyiwhale:
Bug fix for TaskTable checkboxes not updating when task table gets new tasks/loses tasks:

checkbox confusion

### DIFF
--- a/src/js/components/TaskView.js
+++ b/src/js/components/TaskView.js
@@ -126,7 +126,10 @@ class TaskView extends mixin(SaveStateMixin, StoreMixin) {
     return this.state.mesosStateErrorCount >= 5;
   }
 
-  getFilteredTasks(filterByStatus, tasks, searchString) {
+  getFilteredTasks() {
+    let {tasks} = this.props;
+    let {filterByStatus, searchString} = this.state;
+
     if (searchString !== '') {
       tasks = StringUtil.filterByString(tasks, 'id', searchString);
     }
@@ -241,7 +244,7 @@ class TaskView extends mixin(SaveStateMixin, StoreMixin) {
       });
     });
 
-    tasks = this.getFilteredTasks(filterByStatus, tasks, searchString);
+    tasks = this.getFilteredTasks();
 
     let rightAlignLastNChildren = 0;
     let hasCheckedTasks = Object.keys(checkedItems).length !== 0;
--- a/src/js/components/TaskView.js
+++ b/src/js/components/TaskView.js
@@ -1,4 +1,5 @@
 import classNames from 'classnames';
+import deepEqual from 'deep-equal';
 import mixin from 'reactjs-mixin';
 import React from 'react';
 import {StoreMixin} from 'mesosphere-shared-reactjs';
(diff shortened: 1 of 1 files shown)


mesosphere-web:
Reviewer: @pierlo-upitup

src/js/components/TaskView.js
```diff
@@ -60,6 +61,24 @@ class TaskView extends mixin(SaveStateMixin, StoreMixin) {
     super.componentWillMount();
   }
 
+  componentWillReceiveProps(nextProps) {
+    let {filterByStatus, searchString} = this.state;
+    let tasks = this.getFilteredTasks(filterByStatus, nextProps.tasks, searchString);
+    let prevTasks = this.getFilteredTasks(filterByStatus, this.props.tasks, searchString);
+
+    if (!deepEqual(tasks, prevTasks)) {
```
rcorral: I feel that the two calls to `getFilteredTasks` and the `deepEqual` is going to get pretty expensive with thousands of tasks.
@daiyiwhale would you mind doing some performance tests for this method, with 1k, 10k and 100k tasks?

daiyiwhale: There isn't a super convenient way to use fixtures with TaskView so I made some observations on the OrganizationTab, and qualitatively we will be running into table rendering performance snags in scrolling before sorting becomes an issue

daiyiwhale: Also, for reasons I don't understand, performance on TaskView page is already unbearable when number of tasks is in the order of 100s :/

daiyiwhale: I created DCOS-8883, DCOS-8884, and DCOS-8885 to address the underlaying issues in the TaskTable. This PR addresses the bug reported in DCOS-8414 and shouldn't be blocked by performance concerns as they are now to be tracked in those three tickets (:

jfurrow: Sorry if I'm missing something, but instead of calling `getFilteredTasks` at the top of this method, can't we just call `deepEqual` with `nextProps.tasks` to `this.props.tasks` in this `if` statement? Then, if that's truthy, we can call `getFilteredTasks` on each set of tasks.

MatApple: Maybe I'm missing something here but can't we just do:
```
let checkedItems = {};

nextProps.tasks.forEach((task) => {
  if (this.state.checkedItems[task.id]) {
    checkedItems[task.id] = true;
  }
});

this.setState({checkedItems});
```

daiyiwhale: awesome, done!

jfurrow:
LGTM!

src/js/components/TaskView.js
```diff
@@ -219,11 +241,7 @@ class TaskView extends mixin(SaveStateMixin, StoreMixin) {
       });
     });
 
-    if (searchString !== '') {
-      tasks = StringUtil.filterByString(tasks, 'id', searchString);
-    }
-
-    tasks = this.filterByCurrentStatus(tasks);
+    tasks = this.getFilteredTasks(filterByStatus, tasks, searchString);
```
MatApple: We shouldn't pass around props and state as arguments :) Lets take out all the arguments in these methods that are accessible by state or props!

daiyiwhale: Ohh I had those in there when I thought we needed `getFilteredTasks` in `componentWillReceiveProps` and wanted to filter with nextProps, fixed!

src/js/components/TaskView.js
```diff
@@ -99,21 +112,33 @@ class TaskView extends mixin(SaveStateMixin, StoreMixin) {
     this.setState({checkedItems: {}, killAction: ''});
   }
 
-  filterByCurrentStatus(tasks) {
-    let status = this.state.filterByStatus;
-    if (status === 'all') {
+  filterByCurrentStatus(tasks, filterByStatus) {
```
MatApple: One more here :)

src/js/components/TaskView.js
```diff
@@ -99,21 +112,36 @@ class TaskView extends mixin(SaveStateMixin, StoreMixin) {
     this.setState({checkedItems: {}, killAction: ''});
   }
 
-  filterByCurrentStatus(tasks) {
-    let status = this.state.filterByStatus;
-    if (status === 'all') {
+  filterByCurrentStatus() {
```
MatApple: We still need to pass in the tasks here! Maybe we should call them `filteredTasks` because we're passing in an already filtered set from L140

PR #802, opened 2016-07-21
Typeahead component

daiyiwhale:
Introducing the typeahead component:

typeahead component workin

responsive typeahead


mesosphere-web:
Reviewer: @orlandohohmeier

src/js/components/FilterInputText.js
```diff
@@ -1,93 +1,112 @@
...
-  componentDidUpdate: function () {
-    if (this.state.focus) {
-      this.refs.filterInput.focus();
+    METHODS_TO_BIND.forEach(method => {
+      this[method] = this[method].bind(this);
+    });
+  }
+
+  componentDidUpdate(prevProps, prevState) {
+    if (prevState.state !== this.state.focus && this.state.focus && this.inputField) {
```
MatApple: Shouldn't `prevState.state` be `prevState.focus`? :)

src/js/components/FilterInputText.js
```diff
@@ -1,93 +1,112 @@
...
+    if (this.props.searchString) {
+      return this.getClearIcon();
+    }
+  }
 
-  render: function () {
-    var props = this.props;
-    var focus = this.state.focus;
+  render() {
+    let props = this.props;
```
MatApple: Seeing we're making adjustments, lets just pull out the keys we need -> `let {inverseStyle} = this.props;`

src/js/components/FilterInputText.js
```diff
@@ -1,93 +1,112 @@
...
+          size="mini"
+          className="clickable"
+          color={color} />
+        </a>
+      </span>
+    );
+  }
+
+  renderClearIcon() {
+    if (this.props.searchString) {
```
MatApple: This should just be in `getClearIcon`. We don't need `renderClearIcon` as well as `getClearIcon`.

src/js/components/FilterInputText.js
```diff
@@ -127,18 +141,27 @@ var FilterInputText = React.createClass({
...
+
+FilterInputText.defaultProps = {
+  handleFilterChange: function () {},
+  inverseStyle: false,
+  placeholder: 'Filter',
+  searchString: ''
+};
+
+FilterInputText.propTypes = {
+  handleFilterChange: PropTypes.func.isRequired,
```
MatApple: If this is required then we don't need it as a default prop too

src/js/components/FilterInputText.js
```diff
@@ -127,18 +141,27 @@ var FilterInputText = React.createClass({
...
+  inverseStyle: PropTypes.bool,
+  placeholder: PropTypes.string,
+  searchString: PropTypes.string.isRequired,
```
MatApple: Same here

src/js/components/FilterInputText.js
```diff
@@ -127,18 +141,27 @@ var FilterInputText = React.createClass({
...
+  sideText: PropTypes.node
```
MatApple: We should have a default prop set to `null` for this

src/js/components/Typeahead.js
```diff
@@ -0,0 +1,73 @@
...
+    if (this.inputField && this.inputField.value) {
+      return this.getClearIcon();
+    }
+  }
+
+  getInputField() {
+    let {
+      emptyLabel,
+      labelKey,
+      onChange,
```
MatApple: 🔤

src/js/components/Typeahead.js
```diff
@@ -0,0 +1,73 @@
+import InnerTypeahead from 'mesosphere-react-typeahead';
+/* eslint-disable no-unused-vars */
+import React, {PropTypes} from 'react';
+/* eslint-enable no-unused-vars */
+
+import FilterInputText from './FilterInputText';
+
+class Typeahead extends FilterInputText {
+  handleInputChange(input) {
```
MatApple: This method should go below the component lifecycle methods right?

src/js/components/FilterInputText.js
MatApple: Edit: I now see you have this so you can override this method when you use inheritance. What do you think about changing this so we have a method called something like `shouldRenderClearIcon` and that is the method we override. Then in the `render` method we do
```
let clearIcon = null;
if (this.shouldRenderClearIcon()) {
  clearIcon = this.getClearIcon();
}
```

I think that makes this component more readable without knowing that another component inherits from it and needs this control :)

src/js/components/Typeahead.js
```diff
@@ -0,0 +1,73 @@
...
+    this.props.handleFilterChange(input);
+  }
+
+  componentWillUpdate() {
+    if (this.inputField == null && this.typeahead) {
+      if (this.typeahead.refs.inner) {
+        if (this.typeahead.refs.inner.inputComponent) {
```
MatApple: Instead of the nested `if`s, lets just make them all a single `if` chained by `&&`

src/js/components/Typeahead.js
```diff
@@ -0,0 +1,73 @@
...
+          this.inputField = this.typeahead.refs.inner.inputComponent.input;
+        }
+      }
+    }
+  }
+
+  renderClearIcon() {
+    if (this.inputField && this.inputField.value) {
```
MatApple: if `value == 0` we wouldn't be able to clear it right? I think we should make this `this.inputField.value != null`.

src/js/components/Typeahead.js
```diff
@@ -0,0 +1,73 @@
...
+      placeholder,
+      selected
+    } = this.props;
+
+    if (selected === '') {
+      selected = [];
+    }
+
+    return (
+      <div className="typeahead inverse">
```
MatApple: We shouldn't hard code `inverse` style.

src/js/components/Typeahead.js
```diff
@@ -0,0 +1,73 @@
...
+      </div>
+    );
+  }
+}
+
+Typeahead.defaultProps = {
+  labelKey: 'name'
+};
+
+Typeahead.propTypes = {
```
MatApple: If none of these are required then we should probably declare default values in default props.

src/styles/components/typeahead.less
```diff
@@ -0,0 +1,27 @@
+.typeahead.inverse {
```
MatApple: I don't think we should couple the inverse styling to this component so tightly. Inverse should be a modifier to the base styling.

src/js/components/Typeahead.js
```diff
@@ -0,0 +1,73 @@
...
+  emptyLabel: PropTypes.string,
+  labelKey: PropTypes.string,
+  onChange: PropTypes.func,
```
kennyt: perhaps we can change the name of this prop from `onChange` to `onDropdownItemSelection` to make it clear that `onChange` does not apply to typing in the input field

philipnrmn:
I'm seeing this error during the build:
```
ERROR in ../react-typeahead/lib/index.js
Module build failed: Error: Cannot find module 'babel-eslint'
```

`npm install babel-eslint` fixed it, but I get a lot of linting errors from `react-typeahead` now.

src/js/components/Typeahead.js
```diff
@@ -0,0 +1,73 @@
...
+  items: PropTypes.array,
```
philipnrmn: Can we shape this to require objects with `id` and `name`? I like to use `propTypes` as a reference for what I need to provide.

philipnrmn:
Eventually, not necessarily this PR, I'd love to be able to use cursor keys to pick out a typeahead - switching to the mouse from typing feels weird.

philipnrmn:
Couple of minor things aside, it looks great and works nicely!

daiyiwhale:
I fixed the highlighting so selecting items with arrow keys works now!

src/styles/components/typeahead.less
daiyiwhale: hmm it seemed convenient to be able to to make the entire typeahead inverse or non-inverse styled with just one css class

daiyiwhale:
retest this please

jfurrow:
Do the designs call for the offset menu? It seems to me like it should be left-aligned with the textbox.

src/js/components/FilterInputText.js
```diff
@@ -1,117 +1,129 @@
...
+  }
 
-  handleClearInput: function () {
+  handleClearInput() {
     this.props.handleFilterChange('', ServiceFilterTypes.TEXT);
-  },
+  }
 
-  handleBlur: function () {
+  handleInputChange() { }
```
jfurrow: We can remove this

src/js/components/Typeahead.js
```diff
@@ -0,0 +1,83 @@
...
+/* eslint-enable no-unused-vars */
+
+import FilterInputText from './FilterInputText';
+
+class Typeahead extends FilterInputText {
+  handleInputChange(input) {
+    this.props.handleFilterChange(input);
+  }
+
+  handleClearInput() {
```
jfurrow: #nit Could we name these two handlers consistently? So you could just rename this to `handleInputClear`

src/styles/components/typeahead.less
jfurrow: @MatApple's saying that all of these extended styles only apply for the `.inverse` class. But, I see that the `Typeahead` component uses the same classes as Canvas, so there's no need to override most of those styles.

But I do think we should generically extend the `.clickable` class for `li`s, as well as the styles for the `li.active` declaration, regardless of the `.inverse` modifier.

jfurrow:
Great work @daiyiwhale! This is really cool.

src/js/components/FilterInputText.js
mlunoe: Done

src/js/components/FilterInputText.js
mlunoe: Done

src/js/components/Typeahead.js
mlunoe: Done

src/js/components/Typeahead.js
```diff
@@ -0,0 +1,83 @@
...
+    return (
+      <div className={classSet}>
+        <InnerTypeahead
+          emptyLabel={emptyLabel}
+          labelKey={labelKey}
+          onChange={onDropdownItemSelection}
+          onInputChange={this.handleInputChange}
+          options={items}
+          placeholder={placeholder}
+          ref={ref => { if (ref) { this.typeahead = ref.getInstance(); } }}
```
mlunoe: Could we get a comment about why we need to use the `this.getInstance` function, and maybe a reference to here? https://github.com/mesosphere/react-typeahead#public-methods

mlunoe:
Good comments @jfurrow. After his comments, I think this looks good.
Could you post a screenshot of the integration tests passing?

philipnrmn:
I'm getting an error where the `Typeahead` component is being used:
`Typeahead.js?:48 Uncaught TypeError: ref.getInstance is not a function`

jfurrow:
This actually stopped working for me. I'm receiving this error in the console:


Also the dropdown isn't appearing:

daiyiwhale:
Fixed the styles!! whew

src/styles/components/typeahead.less
```diff
@@ -0,0 +1,103 @@
+.form-control.form-control-group {
+
+  .typeahead {
+
+     .form-control {
+
+       &.bootstrap-typeahead-input-hint {
+
```
jfurrow: There's an extra newline here

src/styles/components/typeahead.less
```diff
@@ -0,0 +1,103 @@
...
+        margin-bottom: 0 !important;
+        margin-top: 0 !important;
```
jfurrow: Do we need these `!important` tags? How specific is the margin that we're overriding?

src/styles/components/typeahead.less
daiyiwhale: done!

src/styles/components/typeahead.less
daiyiwhale: Made a comment! (we're overriding Canvas `!important`s)

src/js/components/Typeahead.js
daiyiwhale: done!

src/styles/components/typeahead.less
```diff
@@ -0,0 +1,108 @@
...
+
+  .form-control-group.filter-input-text-group {
+
+    .typeahead {
+
+      .dropdown-menu {
+        margin-left: -32px;
+        // We are matching the width of .filter-input-text-group
+        // as defined in reactjs-components
+        min-width: 100px;
```
jfurrow: Let's use the variables that this pixel value is derived from.

src/styles/components/typeahead.less
```diff
@@ -0,0 +1,108 @@
...
+
+  .form-control-group.filter-input-text-group {
+
+    .typeahead {
+
+      .dropdown-menu {
+        margin-left: -33px;
+        // We are matching the width of .filter-input-text-group
+        // as defined in reactjs-components
+        min-width: 200px;
```
jfurrow: Same as above, let's use the variables that these pixel values are derived from.

src/styles/components/typeahead.less
daiyiwhale: Unfortunately they were hard-coded as `100px` and `200px` :/

jfurrow: Alright let's create the negative margins based on the variables and we should be good!

src/styles/components/typeahead.less
```diff
@@ -0,0 +1,108 @@
...
+          &:extend(.clickable);
+          &:extend(.dropdown-menu-list li.is-selectable all);
+
+          &.active {
+            &:extend(.dropdown-menu-list li.is-selectable:hover);
+          }
+
+          .highlight {
+            font-size: inherit;
+            line-height: inherit;
```
jfurrow: 👍 for the use of `inherit`!

daiyiwhale:
Variables applied to styles! :D

src/styles/components/typeahead.less
daiyiwhale: <3

src/styles/components/typeahead.less
daiyiwhale: done!

src/styles/components/typeahead.less
daiyiwhale: I did the thing!!

src/js/components/FilterInputText.js
daiyiwhale: addressed!

src/js/components/FilterInputText.js
daiyiwhale: addressed!

src/js/components/FilterInputText.js
daiyiwhale: `renderClearIcon()` has been removed entirely!

src/js/components/FilterInputText.js
daiyiwhale: addressed!

src/js/components/Typeahead.js
daiyiwhale: addressed!

src/js/components/Typeahead.js
daiyiwhale: method has been removed!

src/js/components/Typeahead.js
daiyiwhale: logic removed!

src/js/components/Typeahead.js
daiyiwhale: addressed!

src/js/components/Typeahead.js
daiyiwhale: addressed!

src/styles/components/typeahead.less
daiyiwhale: redid all these WIPs with much prettier .less! :D :D

src/js/components/Typeahead.js
mlunoe: This is done

src/js/components/FilterInputText.js
mlunoe: This is done

src/js/components/Typeahead.js
mlunoe: This is done

src/js/components/FilterInputText.js
```diff
@@ -1,121 +1,130 @@
...
-  componentDidUpdate: function () {
-    if (this.state.focus) {
-      this.refs.filterInput.focus();
+    METHODS_TO_BIND.forEach(method => {
+      this[method] = this[method].bind(this);
+    });
+  }
+
+  componentDidUpdate(prevProps, prevState) {
+    if (prevState.focus !== this.state.focus && this.state.focus && this.inputField) {
```
mlunoe: This is longer than 80 chars. You could do:
```js
  componentDidUpdate(prevProps, prevState) {
    let {focus} = this.state;
    if (prevState.focus !== focus && focus && this.inputField) {
      this.inputField.focus();
    }
  }
```

src/js/components/FilterInputText.js
```diff
@@ -1,121 +1,130 @@
...
+      this.inputField.focus();
     }
-  },
+  }
+
+  handleChange(e) {
```
mlunoe: #nit this could be in one line:
```js
  handleChange(event) {
    this.props.handleFilterChange(event.target.value, ServiceFilterTypes.TEXT);
  }
```

mlunoe: Plus, let's write out `event` instead of `e`

src/js/components/FilterInputText.js
daiyiwhale: done!!

src/js/components/FilterInputText.js
daiyiwhale: all fixed 👌

jfurrow:
@mlunoe (or anyone else), do you have any ideas on what we should do about this warning?


It appears to stem from a dependency of the typeahead component.

mlunoe:
Good find @jfurrow!

daiyiwhale:
forked repo & package are updated, this pr is ready to go!

jfurrow:
:ship:!

philipnrmn:
Is nobody else getting those linting errors in master now?

daiyiwhale:
I didn't get them ): Are you still using your npm linked `mesosphere-react-typeahead`?

philipnrmn:
yep

philipnrmn:
Unlinking the react-typeahead fixed the problem :)

daiyiwhale:
:party-parot:

PR #808, opened 2016-07-21
Always have children for confirm modal

kennyt:
If this doesn't get merged, this warning will show:

px

### DIFF
--- a/src/js/components/ServicesTable.js
+++ b/src/js/components/ServicesTable.js
@@ -219,25 +219,27 @@ var ServicesTable = React.createClass({
   },
 
   getDestroyConfirmDialog: function () {
-    let message = null;
     const {state} = this;
     let service = state.serviceToChange;
+    let serviceName = '';
     let itemText = 'Service';
     if (service instanceof ServiceTree) {
       itemText = 'Group';
     }
 
     if (service) {
-      message = (
-        <div className="container-pod flush-top container-pod-short-bottom">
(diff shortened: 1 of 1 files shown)


mesosphere-web:
Reviewer: @philipnrmn

src/js/components/ServicesTable.js
```diff
@@ -219,26 +219,28 @@ var ServicesTable = React.createClass({
...
-          <p>
-            Are you sure you want to destroy <span className="emphasize">{service.getId()}</span>? This action is irreversible.
-          </p>
-          {this.getErrorMessage()}
-        </div>
-      );
+      serviceName = service.getId();
     }
 
+    let message = (
```
mlunoe: If we are always going to have this as children, why not just put it as 'children'?
```js
    return (
      <Confirm
        ...>
        <div className="container-pod flush-top container-pod-short-bottom">
          <h2 className="text-danger text-align-center flush-top">Destroy {itemText}</h2>
          <p>
            Are you sure you want to destroy <span className="emphasize">{serviceName}</span>? This action is irreversible.
          </p>
          {this.getErrorMessage()}
        </div>
      </Confirm>
    );
```

mlunoe:
Thanks for fixing this btw!

src/js/components/ServicesTable.js
kennyt: honestly you're right. i was just going the way it was being done at that point. i'll make the change right now

kennyt: done

PR #814, opened 2016-07-21
Add hooks for plugins in service form errors

kennyt:
In the case the plugins add fields to the service form and it errors in the backend, we allow the plugins to catch that


mesosphere-web:
Reviewer: @jfurrow

PR #820, opened 2016-07-22
Add restart service views.

rcorral:
img

img

### DIFF
--- a/src/js/components/ServiceDetail.js
+++ b/src/js/components/ServiceDetail.js
@@ -144,18 +144,9 @@ class ServiceDetail extends mixin(InternalStorageMixin, StoreMixin, TabsMixin) {
   getDestroyConfirmDialog() {
     const {service} = this.props;
     const {state} = this;
-    const serviceID = (<span className="emphasize">{service.getId()}</span>);
-
-    let message = (
-      <div className="container-pod flush-top container-pod-short-bottom">
-        <h2 className="text-danger text-align-center flush-top">Destroy Service</h2>
-        <p>Destroying {serviceID} is irreversible. Are you sure you want to continue?</p>
-        {this.getErrorMessage()}
-      </div>
-    );
 
     return (
-      <Confirm children={message}
+      <Confirm
         disabled={state.disabledDialog === ServiceActionItem.DESTROY}
         open={state.serviceActionDialog === ServiceActionItem.DESTROY}
         onClose={this.closeDialog}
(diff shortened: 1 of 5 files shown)


mesosphere-web:
Reviewer: @daiyiwhale

rcorral:
img

src/js/components/ServiceDetail.js
```diff
@@ -145,6 +167,33 @@ class ServiceDetail extends mixin(InternalStorageMixin, StoreMixin, TabsMixin) {
...
+        <h2 className="text-danger text-align-center flush-top">Restart Service</h2>
+        <p>
+          Are you sure you want to restart <span className="emphasize">{service.getId()}</span>?
+        </p>
+        {this.getErrorMessage()}
+      </div>
+    );
+
+    return (
+      <Confirm children={message}
```
mlunoe: Could we not pass in the children as a prop, but just add it as markup?
```js
<Confirm
  ...>
  <div className="container-pod flush-top container-pod-short-bottom">
    <h2 className="text-danger text-align-center flush-top">Restart Service</h2>
    <p>
      Are you sure you want to restart <span className="emphasize">{service.getId()}</span>?
     </p>
     {this.getErrorMessage()}
  </div>
</Confirm>
```

src/js/components/ServicesTable.js
```diff
@@ -252,6 +273,41 @@ var ServicesTable = React.createClass({
...
+          <p>
+            Are you sure you want to restart <span className="emphasize">{service.getId()}</span>?
+          </p>
+          {this.getErrorMessage()}
+        </div>
+      );
+    }
+
+    return (
+      <Confirm children={message}
```
mlunoe: `Confirm` always need children, they cannot be `undefined`. See here: https://github.com/dcos/dcos-ui/pull/808

Also, same thing here with children as props

mlunoe:
This works well! Awesome feature 👍

src/js/components/ServicesTable.js
```diff
@@ -252,6 +273,41 @@ var ServicesTable = React.createClass({
     );
   },
 
+  getRestartConfirmDialog: function () {
+    let message = null;
```
kennyt: we don't wanna do this because it'll throw a prop error. `Confirm` requires children. 

see here https://github.com/dcos/dcos-ui/pull/808

kennyt:
works well for me. on an unrelated note, i wonder why we can't restart groups. isn't restarting a group simply restarting all the apps within that group?

marathon-ui also does not have restart for groups though so this is good

src/js/components/ServiceDetail.js
rcorral: Got a JIRA to clean up the dupe code of these modals for next sprint. Let's tackle this there, otherwise this change will cause an inconsistency in how these modals are implemented

src/js/components/ServicesTable.js
rcorral: Same answer as above!

src/js/components/ServicesTable.js
rcorral: Fixed up.

src/js/components/ServiceDetail.js
mlunoe: But this will throw warnings, first of all. Second of all, this PR already merged: https://github.com/dcos/dcos-ui/pull/808

mlunoe: This is why there is conflicts on this PR

src/js/components/ServicesTable.js
```diff
@@ -110,6 +112,17 @@ var ServicesTable = React.createClass({
     });
   },
 
+  onMarathonStoreServiceRestartSuccess: function () {
+    this.closeDialog();
+  },
+
+  onMarathonStoreServiceRestartError: function ({message:errorMsg}) {
```
kennyt: needs a space after `:`

PR #822, opened 2016-07-22
Fix hidden page content due to shrinking page content

jfurrow:
This PR fixes a bug where the page component content (`.page-content`) wouldn't grow to fill its parent when the user's OS has native overlay scrollbars.

I tested it in Chrome, Firefox, Safari, and IE11/Edge, but this seemingly innocuous change has the potential to break scrolling so it's worth being tested by others.

On OS X, you can toggle the overlay scrollbars vs. "normal" scrollbars in System Preferences > General:

If you're using a mouse, set it to "when scrolling" and you'll get overlay scrollbars. If you're using a trackpad, set it to "Automatically..." or "when scrolling."

### DIFF
--- a/src/styles/components/scrollbar.less
+++ b/src/styles/components/scrollbar.less
@@ -35,6 +35,7 @@
 }
 
 .gm-prevented {
+  flex: 1 1 auto;
   overflow: auto;
 }
 


mesosphere-web:
Reviewer: @kennyt

MatApple:
Every time we touch these gemini classes I get really nervous!

MatApple:
tested around a bit and it LGTM but I think we need to test more thoroughly on all browsers and OS scrollbar settings.

mlunoe:
Rebased this for you

mlunoe:
I'm still seeing the problem:

mlunoe:
Arh! Looking at wrong element! `.page-content` grows as it should 😊 :shipit:

mlunoe:
Tested in:
  
✅ ✅ ✅

jfurrow:
Thanks @mlunoe! I tested in Internet Explorer and Edge, so we should be good to go.

mlunoe:
Let's merge this sucker

PR #824, opened 2016-07-22
Fix split-content-button for Firefox

jfurrow:
Firefox doesn't support `display: flex;` for `button` elements ([read about it](https://bugzilla.mozilla.org/show_bug.cgi?id=984869#c2)), so this circumvents the issue by adding additional DOM nodes which receive the appropriate `flex` properties.

Tested in IE10/Edge, Safari, Firefox, and the ever-so-lovely Chrome.


mesosphere-web:
Reviewer: @aldipower

PR #837, opened 2016-07-25
Memoize task IDs

philipnrmn:
This is the simplest possible implementation of a task ID cache - we simply memoize in order to make repeated calls to `getTaskByTaskID(taskID)` inexpensive. 

Potential improvements: smarter cache-busting (we bust on every poll at present); denormalization on data load.

### DIFF
--- a/src/js/stores/MesosStateStore.js
+++ b/src/js/stores/MesosStateStore.js
@@ -97,6 +97,19 @@ class MesosStateStore extends GetSetBaseStore {
     return !!this.listeners(MESOS_STATE_CHANGE).length;
   }
 
+  indexTasksByID(lastMesosState) {
+    let taskIndex = {};
+
+    lastMesosState.frameworks.forEach(function (service) {
+      let tasks = service.tasks.concat(service.completed_tasks);
+      tasks.forEach(function (task) {
+        taskIndex[task.id] = task;
+      });
+    });
+
+    return taskIndex;
+  }
+
   getHostResourcesByFramework(filter) {
     return MesosStateUtil.getHostResourcesByFramework(
       this.get('lastMesosState'), filter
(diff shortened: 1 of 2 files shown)


mesosphere-web:
Reviewer: @MatApple

aldipower:
Thank you @philipnrmn !

src/js/stores/MesosStateStore.js
```diff
@@ -147,8 +148,12 @@ class MesosStateStore extends GetSetBaseStore {
   }
 
   getTaskFromTaskID(taskID) {
+    let taskCache = this.get('taskCache');
+    let foundTask = taskCache[taskID];
+    if (foundTask != null) {
+      return foundTask;
```
MatApple: Think we need to wrap this in the Task struct :)

src/js/stores/MesosStateStore.js
```diff
@@ -164,6 +169,7 @@ class MesosStateStore extends GetSetBaseStore {
       return null;
     }
 
+    taskCache[taskID] = foundTask;
     return new Task(foundTask);
```
MatApple: #nit newline before return statement

MatApple:
@philipnrmn just wondering if this is the correct approach. Memoization will reduce the 2 - n lookups to O(1) but the first lookup of every task is still expensive. I'm wondering how often we will be looking these tasks up more than once before we get a new set of data and the cache is cleared.

What if we looped through all tasks of all services and maintained a map of {taskId: task}, whenever we get new data so we don't have the expensive find operation?

philipnrmn:
Yes, that's an option. It's a relatively expensive initial operation which would have to be performed on every poll though. Let's instrument before going further.

MatApple:
Isn't it as expensive as a worst case single lookup of `getTaskFromTaskID`?

philipnrmn:
Yes (without side-effects obviously), but it's also a guaranteed cost.

src/js/stores/MesosStateStore.js
```diff
@@ -164,7 +169,10 @@ class MesosStateStore extends GetSetBaseStore {
       return null;
     }
 
-    return new Task(foundTask);
+    let result = new Task(foundTask);
+    taskCache[taskID] = result;
```
jfurrow: We should also call `this.set({taskCache})` here, right?

philipnrmn: `taskCache` is a reference here.

MatApple: You're right but I think it's good practice for us to set again.

philipnrmn: I would love to understand this better! I will catch up offline.

philipnrmn: I caught up offline and understand that it has to do with future-proofing. I'll do it and leave a clarifying comment.

philipnrmn:
Made the cache more aggressive, as suggested by @MatApple

src/js/stores/MesosStateStore.js
philipnrmn: In the event, I modified the flow so that `taskCache` is set once per flow.

PR #843, opened 2016-07-25
Call #getSchedules only if we are rendering a job

jfurrow:
Fixes failing integration tests.


mesosphere-web:
Reviewer: @philipnrmn

PR #845, opened 2016-07-26
DCOS-8352 Show pretty cron schedule in job detail page

pierlo-upitup:
Worst-case scenario:

image

Sweet spot:
screen shot 2016-07-26 at 15 33 26

cc @leemunroe as we discussed, we're moving the schedule text below the heading. Here you go!

### DIFF
--- a/src/js/pages/jobs/JobDetailPage.js
+++ b/src/js/pages/jobs/JobDetailPage.js
@@ -17,6 +17,7 @@ import JobRunHistoryTable from './JobRunHistoryTable';
 import MetronomeStore from '../../stores/MetronomeStore';
 import PageHeader from '../../components/PageHeader';
 import RequestErrorMsg from '../../components/RequestErrorMsg';
+import StringUtil from '../../utils/StringUtil';
 import TabsMixin from '../../mixins/TabsMixin';
 import TaskStates from '../../constants/TaskStates';
 
@@ -313,8 +314,7 @@ class JobDetailPage extends mixin(StoreMixin, TabsMixin) {
             id="repeat"
             size="mini"
             family="small" />
-          <span className="job-details-header-status">Scheduled</span>
-          <span>{scheduleText}</span>
+          <span>Scheduled {StringUtil.lowercase(scheduleText)}</span>
         </p>
       );
     }
--- a/src/js/pages/jobs/JobDetailPage.js
+++ b/src/js/pages/jobs/JobDetailPage.js
@@ -1,6 +1,7 @@
 import classNames from 'classnames';
 import {Confirm, Dropdown} from 'reactjs-components';
 import mixin from 'reactjs-mixin';
+import prettycron from 'prettycron';
 /* eslint-disable no-unused-vars */
 import React from 'react';
 /* eslint-enable no-unused-vars */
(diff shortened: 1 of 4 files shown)


mesosphere-web:
Reviewer: @daiyiwhale

leemunroe:
Nice one! Two things:

1. Can we start the line with `Scheduled to run every 0, 2, 4...` so it's clear what it is telling us?
2. Can we move it under the status?

Originally we had designed the status directly below the heading to the right of the icon. Was there a reason we didn't do that or did it get missed?

image

jfurrow:
@leemunroe The status rendering below the icon is an application-wide change to the `PageHeader` component, which was requested by @ashenden.

pierlo-upitup:
I will add the `Scheduled to run ...` copy.

pierlo-upitup:
@leemunroe actually things can get tricky pretty quickly here. The schedule may start with "Every" but also with specific times and other variations. So, prepending `Scheduled to run ` is not going to work with schedules such as `01:00 on the 6th in Sep`. 

I'd suggest to address this outside of this PR and possibly with a simpler approach. One idea:

[ ] **Scheduled** Every other monday
[ ] **Scheduled** 01:00 on the 6th in Sep

leemunroe:
@pierlo-upitup I'm not sure why `Scheduled to run` doesn't work here. `Scheduled to run 01:00 on the 6th in Sep` sounds good to me. But if you think there are situations it doesn't fit happy to fall back to your suggestion of `Scheduled`.

pierlo-upitup:
Shouldn't it be `Scheduled to run AT (time)`, `Scheduled to run ON (week day)` etc.? I will go ahead with the **Scheduled** suggestion as it seems more future-proof.

pierlo-upitup:
All feedback addressed. 

image

leemunroe:
Sorry my fault for not being clear. In your screenshot it looks like `Scheduled` is a separate thing, it's bold and has a space to the right of it and then there's a capital `E` for Every. I'm suggesting it all reads as one sentence `Scheduled to run every 5th minute past every hour on the 1st in February`.

Poltergeist:
👍 It is looking great to me, except the copy change. Please fix this.

MatApple:
LGTM - lets get the copy changed and merge :)

jfurrow:
Just a heads up, we can use the CSS selector `:first-letter` to change the casing of `Every` to `every`, but we can't change `Feb` to `February` unless we fork the [prettycron](https://github.com/azza-bazoo/prettycron) library.

pierlo-upitup:
@leemunroe OK, that's exactly what I suggested here https://github.com/dcos/dcos-ui/pull/845#issuecomment-235575723 and I thought you confirmed it. Anyway, I will remove the bolding and make it look like a single sentence. :shipit:

pierlo-upitup:
Following @jfurrow's suggestion the last commit adds a `.lowercase-first` class that does the trick.

image

pierlo-upitup:
**mesosphere_ci** retest this please

orlandohohmeier:
Seems like we lost the ellipsis from your first screenshot...

pierlo-upitup:
@orlandohohmeier you're right. That's because `.lowercase-first` needs a block-level element for the `:first-letter`, but that doesnt play nice with the ellipsis... 

@jfurrow can you think of a fix here?

pierlo-upitup:
In the end I decided to revert the CSS solution and to go for a `StringUtil.lowercase` function instead, as we already had a `capitalize` function. 

image

pierlo-upitup:
Relevant tests are passing

image

mlunoe:
@pierlo-upitup Let's make sure to run _all_ of the tests, this is what broke tests here:
https://github.com/dcos/dcos-ui/pull/865#issuecomment-236301669

I know this is overhead right now, but we're going to have to suck it in until we get it fixed

PR #851, opened 2016-07-27
Adding node and npm version check

mlunoe:
This check will run on `npm preinstall` and `npm start`

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,8 +17,8 @@
   },
   "license": "Apache-2.0",
   "engines": {
-    "node": "4.x.x",
-    "npm": "3.x.x"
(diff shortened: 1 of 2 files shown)


mesosphere-web:
Reviewer: @aldipower

mlunoe:
Hmmm... didn't actually work, since this is a hard restriction:


Come to think of it, it might be too strict. Will look into that

MatApple:
@mlunoe I still maintain that we shouldn't be focussing on this for now :)

MatApple:
Going to close this PR till we have some bandwidth to re-evaluate.

PR #854, opened 2016-07-27
Using the h4 class to not have nested header elements

mlunoe:
This is a quick fix to not have nested header element. We need to look into making the modal header able to do what we need it to do, and this is a symptom of it.


mesosphere-web:
Reviewer: @pierlo-upitup

PR #855, opened 2016-07-27
DCOS-8811: Use just 'DC/OS' instead of product name in CLI install modal

daiyiwhale:


mesosphere-web:
Reviewer: @jfurrow

mlunoe:
Do we know if design actually wants it this way and not a global change?

daiyiwhale:
I have alerted the designers and john.A. has responded that it is okay for some places to say just 'DC/OS' whereas other say 'Enterprise'. Also he will followup with a design audit in the future.

PR #856, opened 2016-07-27
Remove header navigation on Jobs page

MatApple:
We only have a single tab so we don't need this navigation. [DCOS-8578](https://mesosphere.atlassian.net/browse/DCOS-8578)


mesosphere-web:
Reviewer: @mlunoe

mlunoe:
Awesome! Let's remove some code! ⛵

PR #857, opened 2016-07-28
Using correct height measurement

mlunoe:
Before (I am not controlling being kept in the bottom - only trying to scroll back up):

After (I am controlling the scroll, every scroll direction is controlled by my interaction):


mesosphere-web:
Reviewer: @MatApple

MatApple:
Works great!

PR #866, opened 2016-07-28
DCOS-8588: Allow dropdown menus to appear outside of scrolling containers

jfurrow:
Here's the scroll event handler working with both native overlay scrollbars and Gemini overlay scrollbars:

Native:

Gemini:


mesosphere-web:
Reviewer: @Poltergeist

PR #868, opened 2016-07-28
Worst-case-scenario fix for deploy button issue

philipnrmn:
When creating a new service, invalid field values are not flagged to the user until the deploy button is hit twice. 

This fix is terrible, let's not use it.


mesosphere-web:
Reviewer: @mlunoe

MatApple:
We will be revisiting this.

PR #880, opened 2016-07-29
DCOS-8567: Use button-collection variables instead of custom spacing

jfurrow:
This PR introduces variables for the padding & margins on `filter-bar-item`. Vertical spacing should be the exact same, but the horizontal spacing now uses the `@button-collection-spacing` variable family.

After:


mesosphere-web:
Reviewer: @Poltergeist

Poltergeist:
LGTM !!! 💯  `%`

PR #886, opened 2016-07-29
Adding crude test of history server data in MesosSummaryActions

mlunoe:
This will make sure to fall back to `mesos/summary` endpoint if the data from history server doesn't validate according to this crude test.

More will be done to validate the data that is received from the server later.

### DIFF
--- a/src/js/events/MesosSummaryActions.js
+++ b/src/js/events/MesosSummaryActions.js
@@ -18,9 +18,9 @@ function testHistoryServerResponse(response) {
   // If the response is an empty object, that means something is whack
   // Fall back to making requests to Mesos
   // TODO (DCOS-7764): This should be improved to validate against a schema
-  if (response === {} ||
-      Array.isArray(response.frameworks) ||
-      Array.isArray(response.slaves)) {
+  if (!Object.keys(response).length ||
+      !Array.isArray(response.frameworks) ||
+      !Array.isArray(response.slaves)) {
     _historyServiceOnline = false;
   }
 }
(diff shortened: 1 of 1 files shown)


mesosphere-web:
Reviewer: @MatApple

jfurrow:
Can we add a `TODO` and a link to the JIRA ticket for whatever needs to be done in the future?

src/js/events/MesosSummaryActions.js
```diff
@@ -17,7 +17,10 @@ var _historyServiceOnline = true;
 function testHistoryServerResponse(response) {
   // If the response is an empty object, that means something is whack
   // Fall back to making requests to Mesos
-  if (response === {}) {
+  // TODO (DCOS-7764): This should be improved to validate against a schema
+  if (response === {} ||
+      Array.isArray(response.frameworks) ||
```
orlandohohmeier: Shouldn't this be `!Array.isArray(response.frameworks)` and `Array.isArray(response.slaves)`?

mlunoe: No, we need the `slaves` and `frameworks` to be arrays. Both of them

philipnrmn: `response === {}` will always be `false`. We could test `Object.keys(response).length`, but there's no real need to have the test at all.

philipnrmn: Also, with this code, we flag the history service as offline if 
- response is an object AND
  - response.frameworks is an array OR
  - response.slaves is an array.

I think a valid response from the history server is
 - response is an object AND
    - response.arrays is an array AND
    - response.frameworks is an array OR
 - response is an array of such objects

If that's right we should adjust the test to reflect it.

philipnrmn:
LGTM, just waiting on tests

src/js/events/MesosSummaryActions.js
mlunoe: Updated the test

PR #888, opened 2016-07-30
Align deployment rollback link right

philipnrmn:
Before:

After:

### DIFF
--- a/src/js/pages/services/DeploymentsTab.js
+++ b/src/js/pages/services/DeploymentsTab.js
@@ -228,7 +228,7 @@ class DeploymentsTab extends mixin(StoreMixin) {
         linkText = linkText + ' & Remove';
       }
       return (
-        <a className="deployment-rollback button button-link button-danger table-display-on-row-hover"
+        <a className="deployment-rollback button button-link button-danger table-display-on-row-hover pull-right"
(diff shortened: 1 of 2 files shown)


mesosphere-web:
Reviewer: @orlandohohmeier

src/js/pages/services/DeploymentsTab.js
```diff
@@ -228,7 +228,7 @@ class DeploymentsTab extends mixin(StoreMixin) {
         linkText = linkText + ' & Remove';
       }
       return (
-        <a className="deployment-rollback button button-link button-danger table-display-on-row-hover"
+        <a className="deployment-rollback button button-link button-danger table-display-on-row-hover pull-right"
```
mlunoe: I wasn't aware of this class, but I believe it is legacy. I think we should stay consistent with what we do other tables and use `text-aling-right` on the `td`, like here: https://github.com/dcos/dcos-ui/blob/master/src/js/components/PackagesTable.js#L51

jfurrow: Agreed, `pull-right` does exist in Canvas, but it makes use of `float: right;` which I am not a fan of. Though you could argue this is personal preference...

orlandohohmeier:
LGTM 💯 ✨

PR #891, opened 2016-08-01
Fix wide dropdown width

jfurrow:
This is an intermediary fix for the "wide dropdown" so that it doesn't render off the viewport. A real fix needs to be introduced in `reactjs-components`, which I have documented here: https://mesosphere.atlassian.net/browse/DCOS-8991

Before:

After:

### DIFF
--- a/src/styles/components/dropdown-menus.less
+++ b/src/styles/components/dropdown-menus.less
@@ -40,12 +40,6 @@ components/dropdown-menus.less
       display: block;
       width: 100%;
 
-      .dropdown-menu {
-
-        width: 100%;
-
-      }
-
     }
 
   }


mesosphere-web:
Reviewer: @mlunoe

philipnrmn:
Had an opportunity to review this whilst looking at separate functionality, LGTM.

mlunoe:
Shouldn't this dropdown be right aligned to match the arrow position?

jfurrow:
@mlunoe That would be weird, IMO.

mlunoe:
Ok, couldn't remember how we had the rest of the dropdowns

jfurrow:
The dropdown menu is meant to be as wide as the button, but that's the fix that needs to be made in reactjs-components.

mlunoe:
Arh, makes more sense now

PR #899, opened 2016-08-02
DCOS-8982: Showing verbose errors as they arrive from /package/install

wavesoft:
Changing `CosmosErrorMessage` to show all the errors received from the /package/install response. In addition, exploiting this opportunity I separated some common error paths from `ServiceFormModal` to a separate `ErrorPaths` file.

image


mesosphere-web:
Reviewer: @pierlo-upitup

pierlo-upitup:
**mesosphere_ci** retest this please

src/js/components/CosmosErrorMessage.js
```diff
@@ -71,7 +113,10 @@ CosmosErrorMessage.propTypes = {
   className: React.PropTypes.string,
   error: React.PropTypes.shape({
     message: React.PropTypes.string,
-    type: React.PropTypes.string
+    type: React.PropTypes.string,
+    data: React.PropTypes.shape({
+      errors: React.PropTypes.array
```
cruhland: @wavesoft note that the contents of the `data` field are dependent on the value of the `type` field in the error response. For `MarathonBadResponse`, this is [correct](https://github.com/dcos/cosmos/blob/525c58c94e3f5f61240418a104cdbb8f0a1d96c4/cosmos-server/src/main/scala/com/mesosphere/cosmos/CosmosError.scala#L65), but other error types have different fields. For example, `UnsupportedContentType` will have [`supported` and `actual` fields](https://github.com/dcos/cosmos/blob/525c58c94e3f5f61240418a104cdbb8f0a1d96c4/cosmos-server/src/main/scala/com/mesosphere/cosmos/CosmosError.scala#L83-L86) in its `data` object.

wavesoft: You are right! Changing `data` from `shape` to `object` in order to be more generic.

orlandohohmeier:
@leemunroe  @wavesoft I wonder if we should use the `CollapsibleErrorMessage` (https://github.com/dcos/dcos-ui/pull/911) component for these errors as well...

wavesoft:
Ah, that would be a nice idea. Let me try and I send you a screenshot.

wavesoft:
Nope, I stand corrected. I forgot that we have a floating part that will not play well in this particular case:

image

image

However I could make it a bit more configurable. I think by adding a `className` property to the `CollapsibleErrorMessage` component we could change it's look and feel solely through CSS classes. Shall I?

wavesoft:
Just for the record, I also tried the more generic collapsible error, that does not float the message on top, but then the modal does not resize when the message is expanded.

orlandohohmeier:
@daiyiwhale can you help with the modal resizing?

wavesoft:
@orlandohohmeier found it! Since I was using local state management on the `CollapsibleErrorMessage`, the state updates were not propagating all the way to the modal in order to re-draw. 

I solved it by forwarding the toggle event all the way up to the `InstallPackageModel` and then forcing an update to the model. 

However, in order for this to work we need #911 merged first!

toggler-works mov

wavesoft:
Everything should be addressed.

orlandohohmeier:
LGTM! 🚢

src/js/components/modals/InstallPackageModal.js
```diff
@@ -573,6 +581,7 @@ class InstallPackageModal extends mixin(InternalStorageMixin, TabsMixin, StoreMi
 
     return (
       <Modal
+        ref="modal"
```
daiyiwhale: Perhaps we should use callback refs instead, to avoid possible future tech debt? https://facebook.github.io/react/docs/more-about-refs.html#the-ref-string-attribute

src/js/components/CosmosErrorMessage.js
```diff
@@ -34,9 +36,38 @@ class CosmosErrorMessage extends React.Component {
...
+    if (data && data.errors) {
+
+      // Get an array of array of errors for every individual path
+      let errorsDetails = data.errors.map(function ({path, errors}) {
+        return errors.map(function (error) {
+          return (ErrorPaths[path] || path)+'.'+error;
+        });
+      });
+
+      // Flatten elemens array and return
```
daiyiwhale: I have fixed it for ya (:

daiyiwhale:
This is great! ✨

PR #889, opened 2016-08-01
DCOS-8978: Fixing bug that drops empty environment variables

wavesoft:
The `JSON.stringify` function is dropping keys with `undefined` values. This pull request makes sure that no environment variable value is in this state.

PR #903, opened 2016-08-03
DCOS-8808 & DCOS-8904: rethink the struct approach in the service form modal

Poltergeist:
This PR introduces a `getJSON` method to the Item struct which will return a JSON of the `_itemData` this enables us to

- Make unknown fields which have been added in JSON mode persistent after switching back to the form.
- Keep fields in edit mode which are not present in the form, for example `requirePorts`, `updateStrategy`, `residency` to name a few. Those are also exposed to the JSON mode. 
- Remove fields which have been set in an earlier version for example `user` right now it is only possible to rename the user but not to remove it from the form. because if the value is `null` the field was removed.

In general the JSON is now a lot more verbose but it is necessary.

image

image


mesosphere-web:
Reviewer: @mlunoe

src/js/constants/blacklistServiceConfig.js
```diff
@@ -0,0 +1,16 @@
+let BlackListServiceConfig = [
```
orlandohohmeier: Let's rename this file and provide the black list as a variable so that it is consistent with the other constant files.  How about `ServiceConfig.BLACK_LIST`?

src/js/structs/ServiceTree.js
```diff
@@ -48,6 +48,9 @@ module.exports = class ServiceTree extends Tree {
 
       // Check the DCOS_PACKAGE_FRAMEWORK_NAME label to determine if the item
       // should be converted to an Application or Framework instance.
+      if (item instanceof Framework || item instanceof Application) {
```
orlandohohmeier: Let's add this checks to the condition in line 37 https://github.com/dcos/dcos-ui/pull/903/files#diff-0eb72bc636980e431e5fe14b2a54616dR37

src/js/utils/ServiceUtil.js
```diff
@@ -2,6 +2,25 @@ import {Hooks} from 'PluginSDK';
 import Service from '../structs/Service';
 import VolumeConstants from '../constants/VolumeConstants';
 
+function isEmpty(data) {
```
orlandohohmeier: Can we move this to the validator util?

src/js/structs/Item.js
```diff
@@ -7,6 +7,9 @@ module.exports = class Item {
     this._itemData = item;
   }
 
+  toJSON() {
+    return JSON.stringify(this._itemData);
```
mlunoe: Let's do `this.get()` to not access the internal property directly and use the API in stead.

src/js/structs/ServiceTree.js
mlunoe: Let's check for `Service` here

src/js/utils/ServiceUtil.js
```diff
@@ -2,6 +2,25 @@ import {Hooks} from 'PluginSDK';
...
+  if (typeof(data) == 'number' || typeof(data) == 'boolean') {
```
mlunoe: We use `typeof data` elsewhere, let's stick to that, or change this everywhere

src/js/utils/ServiceUtil.js
```diff
@@ -2,6 +2,25 @@ import {Hooks} from 'PluginSDK';
...
+    return false;
+  }
+  if (typeof(data) == 'undefined' || data === null) {
+    return true;
+  }
+  if (typeof(data.length) != 'undefined') {
+    return data.length === 0;
+  }
+  var count = 0;
```
mlunoe: `let`

src/js/utils/ServiceUtil.js
```diff
@@ -2,6 +2,25 @@ import {Hooks} from 'PluginSDK';
...
+  for (var i in data) {
```
mlunoe: let's use a `reduce` for this

src/js/utils/ServiceUtil.js
mlunoe: Let's get some tests for this function

mlunoe: Also, we should add this to `Util`

mlunoe: Haha, just saw your note now @orlandohohmeier 😆

src/js/utils/ServiceUtil.js
```diff
@@ -174,7 +192,7 @@ const ServiceUtil = {
           if (externalVolumes.length > 0) {
             volumesList = volumesList.concat(externalVolumes);
           }
-          if (!isEdit) {
+          if (!isEdit && volumesList.length > 0) {
```
mlunoe: `&& volumesList.length` should be enough

src/js/constants/BlackListServiceConfig.js
```diff
@@ -0,0 +1,16 @@
+let BlackListServiceConfig = [
```
orlandohohmeier: Let's rename this file and provide the black list as a variable so that it is consistent with the other constant files. How about ServiceConfig.BLACK_LIST?

src/js/utils/ServiceUtil.js
```diff
@@ -364,6 +382,15 @@ const ServiceUtil = {
       }
     }
 
+    definition = Object.keys(definition).reduce(function (memo, key) {
+      if (!isEmpty(definition[key])) {
+        memo[key] = definition[key];
+      } else {
+        memo[key] = null;
+      }
+      return memo;
```
mlunoe: #nit newline before `return`

src/js/components/modals/ServiceFormModal.js
```diff
@@ -103,6 +104,15 @@ const responseAttributePathToFieldIdMap = {
   '/': 'general'
 };
 
+var cleanJSONdefinition = function (jsonDefinition) {
+  return Object.keys(jsonDefinition).filter(function (key) {
+    return !BlackListServiceConfig.includes(key);
+  }).reduce(function (memo, key) {
+    memo[key] = jsonDefinition[key];
+    return memo;
```
mlunoe: #nit newline before `return`

src/js/constants/BlackListServiceConfig.js
```diff
@@ -0,0 +1,16 @@
...
+  'version',
+  'versionInfo',
+  'deployments',
+  'tasks',
+  'taskStats',
+  'tasksHealthy',
+  'tasksRunning',
+  'tasksStaged',
+  'tasksUnhealthy',
+  '_itemData'
```
mlunoe: It this necessary? Isn't this an effect of the bug you discovered here: https://github.com/dcos/dcos-ui/pull/903/files#diff-0eb72bc636980e431e5fe14b2a54616dR51

src/js/structs/Item.js
Poltergeist: Do you mean `this.get('_itemData')` The whole point is to get the whole object which means also unknown values.

Poltergeist: Forget my comment, I just realized that `this.get()` will give you `_itemData`.

src/js/structs/ServiceTree.js
jfurrow: Done!

src/js/structs/Item.js
rcorral: A simple test would be great!

src/js/utils/ServiceUtil.js
jfurrow: Done!

src/js/utils/ServiceUtil.js
jfurrow: Done!

src/js/utils/ServiceUtil.js
jfurrow: Removed entirely! :)

src/js/utils/ServiceUtil.js
jfurrow: Done!

src/js/structs/Item.js
jfurrow: All of these are done :)

src/js/utils/ServiceUtil.js
jfurrow: Done!

src/js/constants/BlackListServiceConfig.js
jfurrow: Done!

src/js/utils/ServiceUtil.js
jfurrow: Done!

src/js/components/modals/ServiceFormModal.js
jfurrow: Done!

src/js/constants/blacklistServiceConfig.js
jfurrow: Done!

src/js/constants/BlackListServiceConfig.js
jfurrow: Yes, it's necessary. Without it we'll be passing the `_itemData` to the JSON editor:

jfurrow:
@Poltergeist @orlandohohmeier @mlunoe I believe I've addressed all of the comments, but let me know if I've missed anything.

src/js/structs/ServiceTree.js
```diff
@@ -34,7 +34,9 @@ module.exports = class ServiceTree extends Tree {
     // Converts items into instances of ServiceTree, Application or Framework
     // based on their properties.
     this.list = this.list.map((item) => {
-      if (item instanceof ServiceTree) {
+      if (item instanceof ServiceTree
+        || item instanceof Framework
```
mlunoe: Check for `Service`

src/js/components/modals/ServiceFormModal.js
```diff
@@ -103,6 +104,15 @@ const responseAttributePathToFieldIdMap = {
   '/': 'general'
 };
 
+var cleanJSONdefinition = function (jsonDefinition) {
+  return Object.keys(jsonDefinition).filter(function (key) {
+    return !ServiceConfig.BLACK_LIST.includes(key);
+  }).reduce(function (memo, key) {
+    memo[key] = jsonDefinition[key];
+    return memo;
```
mlunoe: Newline before `return`

src/js/constants/ServiceConfig.js
```diff
@@ -0,0 +1,18 @@
...
+    'version',
+    'versionInfo',
+    'deployments',
+    'tasks',
+    'taskStats',
+    'tasksHealthy',
+    'tasksRunning',
+    'tasksStaged',
+    'tasksUnhealthy',
+    '_itemData'
```
mlunoe: This should still not be the case. This means that we are instantiating a struct with struct data, and not raw data

src/js/utils/ServiceUtil.js
```diff
@@ -194,7 +195,7 @@ const ServiceUtil = {
 
           if (localVolumes.length > 0) {
             volumesList = volumesList.concat(localVolumes);
-            if (!isEdit) {
+            if (!isEdit && volumesList.length > 0) {
```
mlunoe: `&& volumesList.length` is enough here

mlunoe: Done

src/js/components/modals/ServiceFormModal.js
mlunoe: done

src/js/structs/ServiceTree.js
mlunoe: done

mlunoe:
I didn't get time to finish this up completely today. This: https://github.com/dcos/dcos-ui/pull/903#discussion_r73409725 is still a big issue, and should be fixed before this PR lands. This https://github.com/dcos/dcos-ui/pull/903/commits/1c092c18c3156cbe1412253be55213d77ab7f4a2 might have fixed it, but I did not verify. Cheers

Poltergeist:
Yeah Looks good can we merge this?

src/js/constants/ServiceConfig.js
```diff
@@ -0,0 +1,17 @@
+let ServiceConfig = {
+  BLACK_LIST: [
```
pierlo-upitup: Take it or leave it, but it's `blacklist` in one word ;)

pierlo-upitup:
When I create a new service, and switch to JSON immediately, I see 

image

src/js/stores/DCOSStore.js
```diff
@@ -350,6 +354,12 @@ class DCOSStore extends EventEmitter {
         options = Object.assign(options, frameworks[item.getName()]);
       }
 
+      if (item instanceof Item) {
+        return new item.constructor(
```
pierlo-upitup: Why isn't `new Item(...` enough here?

pierlo-upitup: oh I see.

Poltergeist:
This is a bug on master. @pierlo-upitup

Poltergeist:
The bug was possibly introduced with https://github.com/dcos/dcos-ui/pull/889 Maybe @wavesoft has a quick solution to this bug. I do not think that this should block the merge since it is already on master.

pierlo-upitup:
Adding `hold_merge` so @Poltergeist can push a quick fix for https://github.com/dcos/dcos-ui/pull/903#issuecomment-237580553

Besides that, LGTM

mlunoe:
I get this error, if I delete id:

Also this, if I try to switch back to the form:

mlunoe:
This bug persists:

src/js/components/modals/ServiceFormModal.js
```diff
@@ -258,13 +270,15 @@ class ServiceFormModal extends mixin(StoreMixin) {
     }
 
     if (this.state.jsonMode) {
-      let jsonDefinition = this.state.jsonDefinition;
-      marathonAction(JSON.parse(jsonDefinition), this.state.force);
+      let jsonDefinition = JSON.parse(this.state.jsonDefinition);
+      jsonDefinition = cleanJSONdefinition(jsonDefinition);
+      marathonAction(jsonDefinition, this.state.force);
+      jsonDefinition = JSON.stringify(jsonDefinition, null, 2);
```
pierlo-upitup: Isn't this a bit convoluted?
1) parse the json string
2) remove fields from it and convert to object
3) (call the action)
4) re-convert parsed json back to string
...

src/js/structs/Item.js
```diff
@@ -14,4 +14,8 @@ module.exports = class Item {
 
     return this._itemData[key];
   }
+
+  toJSON() {
```
orlandohohmeier: Let's move this to `Service`, as we're introducing it exclusively for the services.

pierlo-upitup: Won't this be helpful for Jobs too?

orlandohohmeier: Yeah, but I'd rather replicate it or introduce a new shared parent class than adding the method to all(!) of our struct.

src/js/components/modals/ServiceFormModal.js
orlandohohmeier: We're probably able to drastically simplify the whole `handleSubmit` process. I wonder if we should do this in a separate PR. I'd love if we could simplify it in the same way we've did it for the jobs (https://github.com/dcos/dcos-ui/blob/master/src/js/components/modals/JobFormModal.js#L157).

orlandohohmeier:
Let's merge this beast once the tests pass

PR #904, opened 2016-08-03
Introduce general validator utils

orlandohohmeier:
Add a few general validator utils to check if a value is a number, an integer or in in a specific range.

### DIFF
--- a/src/js/utils/ValidatorUtil.js
+++ b/src/js/utils/ValidatorUtil.js
@@ -27,6 +27,13 @@ var ValidatorUtil = {
 
       return memo;
     }, 0) === 0;
+  },
+
+  isNumberInRange: function (value, range = {}) {
+    const {min = 0, max = Number.POSITIVE_INFINITY} = range;
+    const number = parseFloat(value);
+
+    return !Number.isNaN(number) && number >= min && number <= max;
   }
 };
 
--- a/src/js/utils/ValidatorUtil.js
+++ b/src/js/utils/ValidatorUtil.js
@@ -1,4 +1,8 @@
 var ValidatorUtil = {
+  isDefined: function (value) {
+    return value != null && value !== '' || typeof value === 'number';
+  },
+
   isEmail: function (email) {
     return email != null &&
       email.length > 0 &&
--- a/src/js/utils/ValidatorUtil.js
+++ b/src/js/utils/ValidatorUtil.js
@@ -33,6 +33,12 @@ var ValidatorUtil = {
     }, 0) === 0;
   },
 
+  isInteger: function (value) {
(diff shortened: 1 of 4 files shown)


mesosphere-web:
Reviewer: @daiyiwhale

src/js/utils/ValidatorUtil.js
```diff
@@ -1,10 +1,33 @@
...
+    return Number.isInteger(number);
+  },
+
+  isNumber: function (value) {
+    const number = parseFloat(value);
+
+    return !Number.isNaN(number) && Number.isFinite(number);
+  },
+
+  isNumberInRange: function (value, range = {}) {
```
wavesoft: Minor comment, but I would also add an `exclusive (boolean)` parameter. When `true` do only `>` or `<` tests.
For example: `isNumberInRange( numberOfCPUs, 0, 65536, true );`, since a user can enter a 0.01 as value.

orlandohohmeier: Couldn't you just specify a different range in those cases? I'd rather keep the interface simple.

wavesoft: Yeah, you are right. That's over-engineering a bit. I think even `isNumber(x) && (x > 0)` could do the trick.

src/js/utils/__tests__/ValidatorUtil-test.js
```diff
@@ -3,6 +3,30 @@ jest.dontMock('../ValidatorUtil');
...
+    });
+
+    it('should verify that the value is defined', function () {
+      expect(ValidatorUtil.isDefined('a')).toBe(true);
+      expect(ValidatorUtil.isDefined('#')).toBe(true);
+      expect(ValidatorUtil.isDefined(0)).toBe(true);
+      expect(ValidatorUtil.isDefined(1)).toBe(true);
+      expect(ValidatorUtil.isDefined({})).toBe(true);
+      expect(ValidatorUtil.isDefined([])).toBe(true);
+    });
```
wavesoft: Perhaps some negative test cases too?

wavesoft:
Should `Number.POSITIVE_INFINITY` be considered a legal value? I agree that it makes sense in the general validation scope, but I don't think an `Infinity` would be nicely handled by the server, right?

orlandohohmeier:
@wavesoft Yeah, it might cause errors, but I don't think the util should make any assumptions about the server capabilities. It's up to the user of the util to configure it properly.

wavesoft:
Allright then.

Poltergeist:
Could you please rebase. Otherewise Looks good to me.

PR #907, opened 2016-08-03
DCOS-8258: Client-side validation on the JobForm

wavesoft:
I am using the `SchemaForm` validator that already exists. I just made the `tv4` errors a bit more user-friendly and introduced a `validator` function to the schema. This has a different syntax than the `externalValidator` used in the `ServicesForm` .. what's better is open for discussion.

dcos-8258


wavesoft:
Making the error message a bit prettier:

image

wavesoft:
I accidentally committed a few more files. Fixing now.

wavesoft:
A more complicate `JSON` validator is on it's way, but it might be better idea to add this on a separate pull. So, please have a look and if it's good let's merge asap.

src/js/utils/ValidatorUtil.js
```diff
@@ -1,11 +1,116 @@
...
+  /**
+   * Check if the given string is a valid cron component
+   *
+   * @param {string} component - The cron component to test
+   * @param {number} numMin - The minimum value accepted in the numerical components
+   * @param {number} numMax - The maximum value accepted in the numerical components
+   * @param {array} discreteValues - Other valid discreet values
+   * @returns {string} - Returns null if test passes otherwise return the error string
+   */
+  testCronComponent: function (component, numMin, numMax, discreteValues = []) {
```
orlandohohmeier: Can we move this to a dedicated `CRONValidatorUtil`or `ScheduleValidatorUtil` providing an `isCRONComponent` as well as a `isCRON` util?

src/js/schemas/job-schema/Schedule.js
```diff
@@ -41,6 +42,50 @@ const General = {
         let [schedule = {}] = job.getSchedules();
 
         return schedule.cron;
+      },
+      validator: function (value) {
+        let test;
```
orlandohohmeier: I'd love if we could move all these tests to an `isCRON` util.

src/js/utils/SchemaFormUtil.js
```diff
@@ -166,6 +166,127 @@ function processValue(value, valueType) {
...
+      Object.keys(prop.properties).forEach(function (propName) {
+        defClone.properties[propName] = prop.properties[propName];
+      });
+
+      // Adopt requirements
+      if (prop.required != null) {
+        defClone.required = defClone.required.concat(prop.required);
+      }
+
+      // Delete this property
```
orlandohohmeier: Let's try to document the "why" not the "what"...

src/js/utils/SchemaFormUtil.js
```diff
@@ -166,6 +166,127 @@ function processValue(value, valueType) {
...
+
+  });
+
+  return defClone;
+}
+
+/**
+ * Introduce custom validation function to tv4
+ */
+tv4.defineKeyword('validator', function (data, validationFunction) {
```
orlandohohmeier: Shouldn't we do this configuration in different/general place? Maybe adding a tv4 configuration file?

wavesoft: Yes, I am planning to move tv4 configuration + validation utilities in a separate file because I will also be re-used by the JSON validation. For this particular PR I guess this can stay like this.

orlandohohmeier:
I really like this approach and think way it's better than the current `externalValidator` interface, but I still have a few questions. Could you please elaborate how you plan to implement validations where we need the complete model for example validating the service `cmd` and `arg` configuration where only one is allowed to be configured. I'd also like if you could layout a plan how to migrate the services to this new approach.

src/js/utils/ValidatorUtil.js
wavesoft: I chose `testCRONComponent` instead of `isCRONComponent` because instead of returning a boolean value it provides the error description.

Naming aside, I separated the functions to a new util file.

src/js/schemas/job-schema/Schedule.js
wavesoft: Done

src/js/utils/SchemaFormUtil.js
wavesoft: Done

wavesoft:
@orlandohohmeier I was planning on creating a generic validation class that can be used to validate a data structure against anything. This way we can add a TV4 schema validator, or a custom validator. Something like:

```javascript
// 'data' should be the blob we eventually send to the server
// in order to be compatible with JSON or Form data
let validator = new GenericValidator( data );

// The second argument could be a function that will bring the data
// into a structured format that TV4 validator requires (in our case
// that's the `model`)
validator.add( new TV4Validator( schema ), dataToSchemaCompatibleFormatConverterFunction );

// A custom validator does not need any data conversion and has raw
// access to the data set.
validator.add( new CustomValidator() );

// Calling `validate` will collect errors from all the validators and return them
// as an array of {path: '', message: ''} objects. The path is referred to the
// property in the `data` object.
let errors = validator.validate();
```

which can also be nicely extended with plug-ins.

src/js/components/modals/JobFormModal.js
```diff
@@ -145,27 +152,87 @@ class JobFormModal extends mixin(StoreMixin) {
...
     let {isEdit, job} = this.props;
-    let jobSpec = JobUtil.createJobSpecFromJob(this.state.job);
 
-    if (!isEdit) {
-      MetronomeStore.createJob(jobSpec);
+    if (jsonMode) {
+      try {
+
+        // Parse JSON and create job object
+        let job = new Job(JSON.parse(this.state.jsonDefinition));
```
rcorral: Is this the only thing that we need to wrap in the try block or any of it can throw an error?

wavesoft: You are correct. However this whole branch will be re-written soon, since I am working on a client-side JSON validator.

src/js/utils/CRONValidatorUtil.js
```diff
@@ -0,0 +1,214 @@
...
+ */
+class CRONValidatorUtil {
+
+  /**
+   * Test if the given string is a valid CRON syntax
+   *
+   * @param {string} value - The cron string to test
+   * @returns {string} - Returns null if test passes otherwise return the error string
+   */
+  static testCronString(value) {
```
orlandohohmeier: We've decided to use plain objects instead of `static` class methods for utils. Let's use a simple object for now and revisit the decision later. I'll create an issue to revisit that decision as would prefer static methods.

src/js/utils/CRONValidatorUtil.js
```diff
@@ -0,0 +1,214 @@
...
+  /**
+   * Check if the given string is a valid cron component
+   *
+   * @param {string} component - The cron component to test
+   * @param {number} numMin - The minimum value accepted in the numerical components
+   * @param {number} numMax - The maximum value accepted in the numerical components
+   * @param {array} discreteValues - Other valid discreet values
+   * @returns {string} - Returns null if test passes otherwise return the error string
+   */
+  static testCronComponent(component, numMin, numMax, discreteValues = []) {
```
orlandohohmeier: Any chance that can refactor this, so that he util only returns a boolean? Or at least doesn't need to know about the error messages, maybe by introducing "exit" codes? Nonetheless we should strive to make this util consistent with the other validator utils.

src/js/utils/SchemaFormUtil.js
```diff
@@ -166,6 +166,129 @@ function processValue(value, valueType) {
...
+ * @returns {object} The flattened definition
+ */
+function unnestGroupsInDefinition(definition) {
+  let defClone = Object.assign({}, definition);
+  if (defClone.properties == null) {
+    return defClone;
+  }
+
+  // Make sure we are operating on a cloned properties
+  defClone.properties = Object.assign({}, defClone.properties);
```
orlandohohmeier: Let's move the cloning into it's own util - thinking about a general "deep" clone util.

src/js/utils/CRONValidatorUtil.js
wavesoft: Done

src/js/utils/CRONValidatorUtil.js
wavesoft: I can introduce an `isCronComponent` function to be consistent with the rest, but I won't use it. That's the reasoning behind the `testXXX` naming and not `isXXX` naming.

src/js/utils/SchemaFormUtil.js
wavesoft: This is also part of my general validation refactoring, so I would say wait for an upcoming PR 😄

src/js/utils/CRONValidatorUtil.js
```diff
@@ -0,0 +1,214 @@
+
+/**
+ * Validator utilities for CRON syntax
+ */
+var CRONValidatorUtil = {
```
orlandohohmeier: Let's split this into smaller validation methods (e.g. `isValidMinuteValue`, `isValidMonthValue`) to improve the testability as discussed.

MatApple:
conflicts!

wavesoft:
Conflicts should be resolved and Orlando's comment should be addressed.

wavesoft:
A bit of heads-up:
* Enabling the TV4 validator seemed like the fastest option when this PR was started, however things got complicated after the introduction of the toggling `runOnSchedule` field.
* I eventually used the same validation mechanism as on `ServiceForm` even though it's yat another tech debt.
* There was a mix-up on the PR and some commits were mangled with `master` commits. I managed to clean-up everything, it looks OK to me, but double check if this PR removes some of your commits.

Poltergeist:
One minor question why are you introducing the `jsonDefinition` to the `jobFormModal`?

wavesoft:
@Poltergeist : in order to perform the validation in a single place, during submit

src/js/components/modals/JobFormModal.js
```diff
@@ -62,7 +63,8 @@ class JobFormModal extends mixin(StoreMixin) {
       defaultTab: '',
       errorMessage: null,
       job: new Job(),
-      jsonMode: false
+      jsonMode: false,
+      jsonDefinition: ''
```
orlandohohmeier: Let's please not have more than one state property to store the job definition, as this makes it really hard to reason about state changes and might result in a wrong state.

wavesoft:
Added tests for the `JobValidatorUtil` and realised that the CRON pattern that I am using has some major issues.

src/js/schemas/job-schema/General.js
```diff
@@ -52,6 +91,22 @@ const General = {
           type: 'number',
           getter(job) {
             return `${job.getDisk() || ''}`;
+          },
+          externalValidator({general}, definition) {
```
orlandohohmeier: Let's reduce this to one condition as illustrated in the following:

```javascript
 externalValidator({general}, definition) {
    if (ValidatorUtil.isDefined(general.disk) && !ValidatorUtil.isNumberInRange(general.disk)) {
      definition.showError = 'Disk must be a positive integer';

      return false;
    }
    return true;
  }
```

src/js/schemas/job-schema/General.js
```diff
@@ -52,6 +91,22 @@ const General = {
...
+            if (ValidatorUtil.isEmpty(general.disk)) {
```
orlandohohmeier: Please don't use `ValidatorUtil.isEmpty` to check if a property is undefined - we need to refactor that method as the name is highly mislead but is meant to check if an object or array is empty. See DCOS-9453 for details.

src/js/components/JobForm.js
```diff
@@ -26,10 +27,7 @@ class JobForm extends SchemaForm {
   }
 
   handleExternalSubmit() {
-    this.validateForm();
```
orlandohohmeier: Why do we need to change this method?

src/js/components/JobForm.js
```diff
@@ -58,6 +56,33 @@ class JobForm extends SchemaForm {
...
+
+      let fieldValidated = definition.externalValidator(model, definition);
+      if (!fieldValidated) {
+        validated = false;
+      }
+    });
+
+    this.forceUpdate();
+
+    return {
```
orlandohohmeier: I know that this mirrors the `ServiceForm`, but do we actually need the return? Is there no way to simply validate the form and the return the data using `getDataTriple`?

src/js/components/modals/JobFormModal.js
```diff
@@ -150,29 +155,92 @@ class JobFormModal extends mixin(StoreMixin) {
...
-    try {
-      let job = new Job(JSON.parse(jsonDefinition));
-      this.setState({errorMessage: null, job});
-    } catch (error) {
-      // TODO: DCOS-7734 Handle error
-    }
+
+    // Keep the new JSON string in the JSON definition
+    // Validation will occur upon submission of the form
+    this.setState({
```
orlandohohmeier: Why do we need to change this method? What's wrong with converting the JSON definition to an `Job` struct?

src/js/components/modals/JobFormModal.js
```diff
@@ -150,29 +155,92 @@ class JobFormModal extends mixin(StoreMixin) {
...
+      errorMessage: null,
+      jsonDefinition
+    });
+
   }
 
   handleSubmit() {
-    let {isEdit, job} = this.props;
-    let jobSpec = JobUtil.createJobSpecFromJob(this.state.job);
+    let {isEdit} = this.props;
```
orlandohohmeier: This method looks rather complex now, could you please explain why we need all the logic in this method?

src/js/schemas/job-schema/General.js
wavesoft: Done

orlandohohmeier:
data-handling-issue mov
There is also an issue with the data handling. The from seems to restore the old values with every submit.

src/js/schemas/job-schema/General.js
wavesoft: Got it. Done!

src/js/components/modals/JobFormModal.js
wavesoft: Mainly it was just copy-past black magic. However I decided to have a thorough look and properly implement the entire logic. Check the updates.

wavesoft:
Ok, heads-up once again:
* CRON Validation goes away completely. Only a simpler 'valid-characters-only' implementation is in place until the [DCOS-9454](https://mesosphere.atlassian.net/browse/DCOS-9454) is sorted out
* Separated the change and submit logic:
** A `change` event in a form field or the JSON spec only updates their individual state. No validation of any sort takes place.
** A `submit` event triggers validation and returns the composed job using the `createJobFromEditorContents` function.
** The same thing occurs when the user switches between json/form mode. However in that case validation errors are not shown and in case of errors the default job from the properties is used.

src/js/components/JobForm.js
```diff
@@ -43,21 +37,60 @@ class JobForm extends SchemaForm {
...
-    this.validateForm();
-    // Provide updated model to get new multipleDefinition
-    let model = this.triggerTabFormSubmit();
-    this.multipleDefinition = this.getNewDefinition(model);
-    this.props.onChange(this.getDataTriple(model));
-    // Only update if schedule changed
-    if (scheduleEnabledChange) {
+    // Fetch the updated model from the <Form /> component
+    // nested deeper in the component tree.
+    this.model = this.triggerTabFormSubmit();
```
mlunoe: I am sorry if I am late to the game here, I have been working pretty hard to not set `model` on this and just work with local variables. Is there a reason why you decided to set `model` on this? If not I would really like to see that not happen

mlunoe:
Sorry, I am removing the review label, because of the many changes since.

src/js/components/JobForm.js
wavesoft: Yeah, I also try to avoid it as much as possible, but it comes fro the superclass `SchemaForm` that uses it. This is really nasty and the solution here is what I thought can be the minimal effort required to make this work without major refactoring on the `SchemaForm` (since there is already a plan for this).

wavesoft: Besides, even if I don't set the `.model` I will need to set the `.multipleDefinition`, because I want the validation to happen outside the `handleFormChange` event. Therefore since we are already violating the principle, I thought of keeping it more compatible with the whatever is the weird logic of `SchemaForm`

wavesoft:

src/js/components/JobForm.js
```diff
@@ -43,21 +37,60 @@ class JobForm extends SchemaForm {
...
+    // Forward the onChange event to the parent
+    // The parent is responsible for updating the form with the new model
+    // in order to reflect the changes
+    this.props.onChange(this.getDataTriple());
+
+    // Force an update if we have clicked on the checkbox
+    if (eventObj.fieldName === 'runOnSchedule') {
       this.forceUpdate();
     }
+
```
Poltergeist: #nit no new line at the end of a function.

src/js/components/modals/JobFormModal.js
```diff
@@ -123,60 +131,179 @@ class JobFormModal extends mixin(StoreMixin) {
...
+          }
+        });
+        if (!keepValidationErrors) {
+          return null;
+        }
+      }
+
+      return job;
+
+    } else {
```
Poltergeist: The else is not needed since we are returning at the end of the if block in line 206.

src/js/components/JobForm.js
wavesoft: ✔️

src/js/components/modals/JobFormModal.js
wavesoft: ✔️

src/js/schemas/job-schema/General.js
Poltergeist: This line did not change? Is that right?

wavesoft: Oops. Must be slipped during rebase. Fixed it.

Poltergeist:
Soime minor nits the rest looks good.

Poltergeist:
LGTM 👍 💯

src/js/components/JobForm.js
mlunoe: Alright, that is fair

src/js/schemas/job-schema/General.js
```diff
@@ -36,6 +49,19 @@ const General = {
           type:'number',
           getter(job) {
             return `${job.getCpus() || ''}`;
+          },
+          externalValidator({general}, definition) {
+            if (!ValidatorUtil.isNumber(general.cpus)) {
```
orlandohohmeier: Let's combine these condition using `!ValidatorUtil.isNumberInRange(general.cpus, {min:MesosConstants.MIN_CPUS})`

src/js/schemas/job-schema/General.js
```diff
@@ -44,6 +70,19 @@ const General = {
           type: 'number',
           getter(job) {
             return `${job.getMem() || ''}`;
+          },
+          externalValidator({general}, definition) {
+            if (!ValidatorUtil.isNumber(general.mem)) {
```
orlandohohmeier: Let's combine these conditions as well.

src/js/schemas/job-schema/General.js
orlandohohmeier: Please reduce this to one condition.

src/js/schemas/job-schema/Schedule.js
```diff
@@ -53,6 +69,26 @@ const Schedule = {
...
+      externalValidator({schedule}, definition) {
+        if (!schedule.runOnSchedule) {
+          return true;
+        }
+
+        if (!ValidatorUtil.isDefined(schedule.startingDeadlineSeconds)) {
+          return true;
+        }
+
+        if (!ValidatorUtil.isNumber(schedule.startingDeadlineSeconds)) {
```
orlandohohmeier: Let's reduce this to one condition as well.

src/js/utils/__tests__/JobValidatorUtil-test.js
```diff
@@ -0,0 +1,85 @@
...
+      expect(JobValidatorUtil.isValidJobID('Uppercase')).toBe(false);
+      expect(JobValidatorUtil.isValidJobID('job#1')).toBe(false);
+      expect(JobValidatorUtil.isValidJobID('job_1')).toBe(false);
+      expect(JobValidatorUtil.isValidJobID('job%1')).toBe(false);
+      expect(JobValidatorUtil.isValidJobID('job+1')).toBe(false);
+      expect(JobValidatorUtil.isValidJobID('job*1')).toBe(false);
+      expect(JobValidatorUtil.isValidJobID('job(1)')).toBe(false);
+    });
+
+    it('should properly handle multiple dots slashes', function () {
```
orlandohohmeier: should properly handle multiple dots ~~slashes~~

orlandohohmeier:
💥  🎆: 🚢 🇮🇹 ✨

PR #910, opened 2016-08-04
Fixing endpoint displays, ref. DCOS-8975

mlunoe:
BRIDGE network:

HOST network:

Virtual Networks is still missing

### DIFF
--- a/src/js/utils/TaskUtil.js
+++ b/src/js/utils/TaskUtil.js
@@ -2,42 +2,53 @@
 import React from 'react';
 /* eslint-enable no-unused-vars */
 
+import Node from '../structs/Node';
 import Util from './Util';
-import Service from '../structs/Service';
 
 const TaskUtil = {
   /**
    * Returns a list of ips or hosts from task
    * @param  {object} task to return ip or host list from
+   * @param  {object} node to get host name from, if necessary
    * @return {Array.<string>} an array of ip addresses or hosts
    */
-  getHostList(task = {}) {
-    let {ipAddresses} = task;
-    if (ipAddresses && ipAddresses.length) {
-      return ipAddresses.map(function (address) {
-        return address.ipAddress;
-      });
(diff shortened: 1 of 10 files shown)


mesosphere-web:
Reviewer: @jfurrow

mlunoe:
@Poltergeist Would be nice if I could get your eyes on this before tomorrow morning

mlunoe:
Tests for the `TaskEndpointsList` are failing, will fix tomorrow.

Poltergeist:

Poltergeist:
It looks good to me and works very fine! Sorry for the delay.

src/js/utils/TaskUtil.js
```diff
@@ -74,6 +85,29 @@ const TaskUtil = {
     }
 
     return portMappings;
+  },
+
+  getNetwork(task) {
```
jfurrow: What exactly does this return? Perhaps it could be called `#getNetworkType`?

src/js/utils/TaskUtil.js
```diff
@@ -2,42 +2,53 @@
...
+    let ports = this.getPorts(task);
+    let hostName = node.getHostName();
+
+    // networkType is 'HOST', but no host name
+    if (networkType !== 'BRIDGE' && networkType !== 'USER' && !hostName) {
+      return {ports: [], hosts: []};
+    }
+
+    // networkType is 'HOST'
+    if (networkType !== 'BRIDGE' && networkType !== 'USER') {
```
jfurrow: Do we need these negative checks? Can we not just use `networkType === 'HOST'`? Just curious, I don't know the details here...

mlunoe: This is future proofing, and defaulting it to handle it as 'HOST' network, so if they ever add another network type, we will handle it as 'HOST' by default

PR #911, opened 2016-08-04
DCOS-8454: Introducing a re-usable CollapsibleErrorMessage component

wavesoft:
This component is now used by the `ServiceFormModal` and `JobFormModal` for rendering the error messages.

collapsible-error mov

### DIFF
--- a/src/js/components/CollapsibleErrorMessage.js
+++ b/src/js/components/CollapsibleErrorMessage.js
@@ -1,4 +1,3 @@
-
 import Icon from './Icon';
 
 import React from 'react';
@@ -70,11 +69,11 @@ class CollapsibleErrorMessage extends React.Component {
   }
 
   /**
-   * [Custom component] The (show more/less) link
+   * [Fragment] The show more/less link
    *
    * @returns {React.Component} - The rendered content
    */
-  getDetailLink() {
+  getShowDetailsLink() {
     let {message, details} = this.props;
     let isDetailed = message && details && details.length;
 
@@ -101,11 +100,11 @@ class CollapsibleErrorMessage extends React.Component {
   }
 
   /**
-   * [Custom component] The list of errors in the details
+   * [Fragment] The list of errors in the details
    *
    * @returns {React.Component} - The rendered content
    */
-  getDetailMessages() {
+  getDetailsListItems() {
     let {details} = this.props;
 
     return details.map(function (message) {
@@ -118,11 +117,11 @@ class CollapsibleErrorMessage extends React.Component {
   }
 
   /**
-   * [Custom component] The fixed message part
+   * [Fragment] The fixed message part
    *
    * @returns {React.Component} - The rendered content
    */
-  getFixedPart() {
+  getFixedMessagePart() {
     let {message} = this.props;
     let isVisible = !!message;
 
@@ -141,7 +140,7 @@ class CollapsibleErrorMessage extends React.Component {
             className="icon-alert icon-margin-right"
             color="red" />
           {message}
-          {this.getDetailLink()}
+          {this.getShowDetailsLink()}
         </div>
       );
   }
@@ -151,7 +150,7 @@ class CollapsibleErrorMessage extends React.Component {
    *
    * @returns {React.Component} - The rendered content
    */
-  getToggledPart() {
+  getCollapsibleMessagePart() {
     let {message, details} = this.props;
     let isDetailed = !!message && details && details.length;
 
@@ -166,7 +165,7 @@ class CollapsibleErrorMessage extends React.Component {
         <div className="collapsible-toggled">
           <div>
             <ul>
-            {this.getDetailMessages()}
+            {this.getDetailsListItems()}
             </ul>
           </div>
         </div>
@@ -200,8 +199,8 @@ class CollapsibleErrorMessage extends React.Component {
     // Render message component
     return (
         <div className={className}>
-          {this.getFixedPart()}
-          {this.getToggledPart()}
+          {this.getFixedMessagePart()}
+          {this.getCollapsibleMessagePart()}
         </div>
       );
 
--- a/src/js/components/CollapsibleErrorMessage.js
+++ b/src/js/components/CollapsibleErrorMessage.js
@@ -8,10 +8,6 @@ import classNames from 'classnames';
  * Methods to bind in 'this' context
  */
 const METHODS_TO_BIND = [
-  'getDetailLink',
-  'getDetailMessages',
-  'getFixedPart',
-  'getToggledPart',
   'toggleExpanded'
 ];
 
@@ -112,8 +108,10 @@ class CollapsibleErrorMessage extends React.Component {
    */
   getDetailMessages() {
     let {details} = this.props;
+
     return details.map(function (message) {
       let msg = message.toString();
+
       return (
           <li key={msg}>{msg}</li>
         );
--- a/src/js/components/CollapsibleErrorMessage.js
+++ b/src/js/components/CollapsibleErrorMessage.js
@@ -49,7 +49,7 @@ class CollapsibleErrorMessage extends React.Component {
   }
 
   /**
-   * Set the default expandedg from the properties
+   * Set the default expanded state from the properties
    */
   componentWillMount() {
     this.setState({
--- a/src/js/components/CollapsibleErrorMessage.js
+++ b/src/js/components/CollapsibleErrorMessage.js
@@ -18,6 +18,20 @@ const METHODS_TO_BIND = [
 /**
  * An error bar component commonly used in the ServiceForm and the JobsForm
  * that contains collapsible error messages.
+ *
+ * @example <caption>How to use the CollapsibleErrorMessage</caption>
+ *   <CollapsibleErrorMessage
+ *
+ *        className="error-for-modal"
+ *
+ *        details={[ 'an', 'array', 'of', 'detailed errors' ]}
+ *        message="The message to display"
+ *
+ *        onToggle={function(isToggled) {
+ *          // Do something
+ *        }}
+ *
+ *      />
  */
 class CollapsibleErrorMessage extends React.Component {
 
@@ -40,12 +54,24 @@ class CollapsibleErrorMessage extends React.Component {
   }
 
   /**
+   * Set the default expandedg from the properties
+   */
+  componentWillMount() {
+    this.setState({
+      expanded: this.props.expanded
+    });
+  }
+
+  /**
    * Toggle the expanded state of the collapsible error bar
    */
   toggleExpanded() {
     this.setState({
       expanded: !this.state.expanded
     });
+
+    // Call onToggle callback
+    this.props.onToggle(this.state.expanded);
   }
 
   /**
@@ -71,7 +97,7 @@ class CollapsibleErrorMessage extends React.Component {
     // Render
     return (
         <span
-          className="collapsible-toggle-label text-danger clickable"
+          className="collapsible-toggle-label  clickable"
           onClick={this.toggleExpanded} >
           (Show {moreLess})
         </span>
@@ -110,7 +136,7 @@ class CollapsibleErrorMessage extends React.Component {
 
     // Render the fixed part of the message
     return (
-        <div className="collapsible-fixed text-align-center text-danger flush-top">
+        <div className="collapsible-fixed">
           <Icon
             family="mini"
             id="yield"
@@ -138,9 +164,10 @@ class CollapsibleErrorMessage extends React.Component {
     }
 
     // Render the toggled part of the message
+    // (Note: The nested div is used to float the contents when needed)
     return (
         <div className="collapsible-toggled">
-          <div className="text-danger">
+          <div>
             <ul>
             {this.getDetailMessages()}
             </ul>
@@ -187,14 +214,18 @@ class CollapsibleErrorMessage extends React.Component {
 
 CollapsibleErrorMessage.defaultProps = {
   className: '',
+  details: null,
+  expanded: false,
   message: '',
-  details: null
+  onToggle: function () { }
 };
 
 CollapsibleErrorMessage.propTypes = {
   className: React.PropTypes.string,
+  details: React.PropTypes.array,
+  expanded: React.PropTypes.bool,
   message: React.PropTypes.string,
-  details: React.PropTypes.array
+  onToggle: React.PropTypes.func
 };
 
 module.exports = CollapsibleErrorMessage;
--- a/src/js/components/CollapsibleErrorMessage.js
+++ b/src/js/components/CollapsibleErrorMessage.js
@@ -69,16 +69,15 @@ class CollapsibleErrorMessage extends React.Component {
   }
 
   /**
-   * [Fragment] The show more/less link
+   * The show more/less link
    *
-   * @returns {React.Component} - The rendered content
+   * @returns {React.Component|null} - The rendered content
    */
   getShowDetailsLink() {
     let {message, details} = this.props;
-    let isDetailed = message && details && details.length;
 
     // Check if we must not show the detail link at all
-    if (!isDetailed) {
+    if (!message || !details || !details.length) {
       return null;
     }
 
@@ -100,33 +99,32 @@ class CollapsibleErrorMessage extends React.Component {
   }
 
   /**
-   * [Fragment] The list of errors in the details
+   * The list of errors in the details
    *
    * @returns {React.Component} - The rendered content
    */
   getDetailsListItems() {
     let {details} = this.props;
 
-    return details.map(function (message) {
+    return details.map(function (message, i) {
       let msg = message.toString();
 
       return (
-          <li key={msg}>{msg}</li>
+          <li key={i}>{msg}</li>
         );
     });
   }
 
   /**
-   * [Fragment] The fixed message part
+   * The fixed message part
    *
-   * @returns {React.Component} - The rendered content
+   * @returns {React.Component|null} - The rendered content
    */
   getFixedMessagePart() {
     let {message} = this.props;
-    let isVisible = !!message;
 
     // If not visible, just exit
-    if (!isVisible) {
+    if (!message) {
       return null;
     }
 
@@ -148,7 +146,7 @@ class CollapsibleErrorMessage extends React.Component {
   /**
    * [Custom component] The toggled message part
    *
-   * @returns {React.Component} - The rendered content
+   * @returns {React.Component|null} - The rendered content
    */
   getCollapsibleMessagePart() {
     let {message, details} = this.props;
@@ -176,14 +174,13 @@ class CollapsibleErrorMessage extends React.Component {
   /**
    * React.js Render Function
    *
-   * @returns {React.Component} - The rendered content
+   * @returns {React.Component|null} - The rendered content
    */
   render() {
     let {message} = this.props;
-    let isVisible = !!message;
 
     // If not visible, just exit
-    if (!isVisible) {
+    if (!message) {
       return null;
     }
 
--- a/src/js/components/CollapsibleErrorMessage.js
+++ b/src/js/components/CollapsibleErrorMessage.js
@@ -62,12 +62,11 @@ class CollapsibleErrorMessage extends React.Component {
    * Toggle the expanded state of the collapsible error bar
    */
   toggleExpanded() {
-    this.setState({
-      expanded: !this.state.expanded
-    });
+    let expanded = !this.state.expanded;
+    this.setState({expanded});
 
     // Call onToggle callback
-    this.props.onToggle(this.state.expanded);
+    this.props.onToggle(expanded);
   }
 
   /**
--- a/src/js/components/CollapsibleErrorMessage.js
+++ b/src/js/components/CollapsibleErrorMessage.js
@@ -11,7 +11,7 @@ const METHODS_TO_BIND = [
   'getDetailLink',
   'getDetailMessages',
   'getFixedPart',
-  'getFloatingPart',
+  'getToggledPart',
   'toggleExpanded'
 ];
 
@@ -71,7 +71,7 @@ class CollapsibleErrorMessage extends React.Component {
     // Render
     return (
         <span
-          className="text-danger clickable"
+          className="collapsible-toggle-label text-danger clickable"
           onClick={this.toggleExpanded} >
           (Show {moreLess})
         </span>
@@ -108,13 +108,9 @@ class CollapsibleErrorMessage extends React.Component {
       return null;
     }
 
-    // Make sure at least one whitespace character
-    // exists after the message string
-    message += ' ';
-
     // Render the fixed part of the message
     return (
-        <div key="fixed" className="collapsible-fixed text-align-center text-danger flush-top">
+        <div className="collapsible-fixed text-align-center text-danger flush-top">
           <Icon
             family="mini"
             id="yield"
@@ -128,11 +124,11 @@ class CollapsibleErrorMessage extends React.Component {
   }
 
   /**
-   * [Custom component] The floating message part
+   * [Custom component] The toggled message part
    *
    * @returns {React.Component} - The rendered content
    */
-  getFloatingPart() {
+  getToggledPart() {
     let {message, details} = this.props;
     let isDetailed = !!message && details && details.length;
 
@@ -141,9 +137,9 @@ class CollapsibleErrorMessage extends React.Component {
       return null;
     }
 
-    // Render the floating part of the message
+    // Render the toggled part of the message
     return (
-        <div key="floating" className="collapsible-floating">
+        <div className="collapsible-toggled">
           <div className="text-danger">
             <ul>
             {this.getDetailMessages()}
@@ -169,20 +165,36 @@ class CollapsibleErrorMessage extends React.Component {
     }
 
     // Compile classes
-    let className = classNames('collapsible-error-message', {
-      expanded: this.state.expanded
-    });
+    let className = classNames(
+      this.props.className,
+      'collapsible-error-message',
+      {
+        'expanded': this.state.expanded
+      }
+    );
 
     // Render message component
     return (
         <div className={className}>
           {this.getFixedPart()}
-          {this.getFloatingPart()}
+          {this.getToggledPart()}
         </div>
       );
 
   };
 
+};
+
+CollapsibleErrorMessage.defaultProps = {
+  className: '',
+  message: '',
+  details: null
+};
+
+CollapsibleErrorMessage.propTypes = {
+  className: React.PropTypes.string,
+  message: React.PropTypes.string,
+  details: React.PropTypes.array
 };
 
 module.exports = CollapsibleErrorMessage;
--- a/src/js/components/CollapsibleErrorMessage.js
+++ b/src/js/components/CollapsibleErrorMessage.js
@@ -1,0 +1,110 @@
+
+import Icon from './Icon';
+
+import React from 'react';
+import classNames from 'classnames';
+
+/**
+ * An error bar component commonly used in the ServiceForm and the JobsForm
+ * that contains collapsible error messages.
+ */
+class CollapsibleErrorMessage extends React.Component {
+
+  /**
+   * Initialize superclass
+   */
+  constructor() {
+    super(...arguments);
+
+    // Initial state
+    this.state = {
+      expanded: false
+    };
+
+  }
+
+  /**
+   * Toggle the expanded state of the collapsible error bar
+   */
+  toggleExpanded() {
+    this.setState({
+      expanded: !this.state.expanded
+    });
+  }
+
+  /**
+   * React.js Render Function
+   *
+   * @returns {React.Component} - The rendered content
+   */
+  render() {
+    let {message, details} = this.props;
+    let isVisible = !!message;
+    let isDetailed = isVisible && details && details.length;
+    let contents = [];
+
+    // For sure, create a message body
+    contents.push(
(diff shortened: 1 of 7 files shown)


mesosphere-web:
Reviewer: @pierlo-upitup

src/js/components/CollapsibleErrorMessage.js
```diff
@@ -0,0 +1,110 @@
...
+    contents.push(
+        <div key="fixed" className="collapsible-fixed text-align-center text-danger flush-top">
+          <Icon
+            family="mini"
+            id="yield"
+            size="mini"
+            className="icon-alert icon-margin-right"
+            color="red" />
+          {message}
+          {(() => {
```
mlunoe: Let's keep functions out of the view declarations

src/js/components/CollapsibleErrorMessage.js
```diff
@@ -0,0 +1,110 @@
...
+   */
+  render() {
+    let {message, details} = this.props;
+    let isVisible = !!message;
+    let isDetailed = isVisible && details && details.length;
+    let contents = [];
+
+    // For sure, create a message body
+    contents.push(
+        <div key="fixed" className="collapsible-fixed text-align-center text-danger flush-top">
```
mlunoe: Could we create `get`-functions for these view parts, as we do elsewhere?

src/js/components/CollapsibleErrorMessage.js
```diff
@@ -0,0 +1,110 @@
...
+                );
+            }
+          })()}
+        </div>
+      );
+
+    // If we have details, create a floating element below
+    // that will show them
+    if (isDetailed && this.state.expanded) {
+      contents.push(
```
mlunoe: Same with this, let's create a separate `get`-function for this view part

src/js/components/CollapsibleErrorMessage.js
mlunoe: See our function hierarchy here: https://github.com/dcos/dcos-ui/blob/master/BEST_PRACTICES.md#component-method-order
(Custom component methods is where the `get`-functions should go)

src/js/components/CollapsibleErrorMessage.js
mlunoe: Here for an example for how we separate logic from view definition: https://github.com/dcos/dcos-ui/blob/master/src/js/components/ImageViewer.js#L45

```js
const METHODS_TO_BIND = [
  'toggleExpanded'
];

...
  constructor() {
    super(...arguments);

    this.state = {
      expanded: false
    };

    // We use our bind snippet, when context needs to be changed
    });
  }

  ...

  getShowDetailLink() {
    // Base cases return early
    if (!isDetailed) {
      return null;
    }
    let text = 'more';
    // We don't use ternary operators
    if (this.state.expanded) {
      text = 'less';
    }

    return (
       <span>
         &nbsp;
         <span
           className="text-danger clickable"
           onClick={this.toggleExpanded}>
           <small>{`(Show ${text})`}</small>
         </span>
       </span>
    );
  }
```

src/js/components/CollapsibleErrorMessage.js
```diff
@@ -0,0 +1,110 @@
...
+          {message}
+          {(() => {
+            if (isDetailed) {
+              return (
+                  <span>
+                    &nbsp;
+                    <span
+                      className="text-danger clickable"
+                      onClick={this.toggleExpanded.bind(this)}>
+                      <small>(Show {this.state.expanded ? 'less' : 'more'})</small>
```
mlunoe: Let's remove logic from view definition, see above

src/js/components/CollapsibleErrorMessage.js
```diff
@@ -0,0 +1,110 @@
...
+            color="red" />
+          {message}
+          {(() => {
+            if (isDetailed) {
+              return (
+                  <span>
+                    &nbsp;
+                    <span
+                      className="text-danger clickable"
+                      onClick={this.toggleExpanded.bind(this)}>
```
mlunoe: Let's use our bind snippet, see above

src/js/components/CollapsibleErrorMessage.js
```diff
@@ -0,0 +1,110 @@
...
+      );
+
+    // If we have details, create a floating element below
+    // that will show them
+    if (isDetailed && this.state.expanded) {
+      contents.push(
+          <div key="floating" className="collapsible-floating">
+            <div className="text-danger">
+              <ul>
+              {details.map(function (message) {
```
mlunoe: Same here, see above

src/js/components/modals/ServiceFormModal.js
```diff
@@ -342,37 +344,26 @@ class ServiceFormModal extends mixin(StoreMixin) {
...
     if (this.shouldForceUpdate(errorMessage)) {
       return (
-        <div className="error-field text-danger">
-          <h4 className="text-align-center text-danger flush-top">
-            App is currently locked by one or more deployments. Press the button
-            again to forcefully change and deploy the new configuration.
-          </h4>
-        </div>
+        <CollapsibleErrorMessage
+          message={'App is currently locked by one or more deployments. Press the button again ' +
```
mlunoe: We don't need to break strings, they don't abide by the 80 char rule. This should be fine:
`message="App is currently locked by one or more deployments. Press the button again to forcefully change and deploy the new configuration." />`
Or maybe even declare the message above if you will

src/styles/components/collapsible-error-message.less
```diff
@@ -0,0 +1,54 @@
...
+    position: relative;
+
+    > div {
+
+      // Fill the fixed part dimentions
+      position: absolute;
+      left: 0; width: 100%;
+
+      // Bring on fore-most z-index
+      z-index: 100000;
```
mlunoe: Let's base this off of canvas' modal z-index (or new or other variable you see fit to solve your issue) here: https://github.com/dcos/dcos-ui/blob/master/src/styles/variables/variables-z-index.less

src/styles/components/collapsible-error-message.less
```diff
@@ -0,0 +1,54 @@
...
+  // The floating part of the error message
+  //
+  .collapsible-floating {
+    position: relative;
+
+    > div {
+
+      // Fill the fixed part dimentions
+      position: absolute;
+      left: 0; width: 100%;
```
mlunoe: alphabetize, please and one statement per line

src/styles/components/collapsible-error-message.less
```diff
@@ -0,0 +1,54 @@
...
+// message and a floating drop-down with the error details
+//
+.collapsible-error-message {
+
+  //
+  // Shared styling between the two items
+  //
+  .collapsible-fixed, .collapsible-floating > div {
+    background: color-lighten(@red, 90);
+    padding: (@base-spacing-unit / 2) @base-spacing-unit;
```
mlunoe: When we use `@base-spacing-unit`, we need to declare it for all of the screen sizes, like here: https://github.com/dcos/dcos-ui/blob/master/src/styles/components/dashboard-health-list.less#L49
(this file is not perfect, unfortunately)

src/styles/components/collapsible-error-message.less
mlunoe: If this is only a problem when it is used in the modal, let's wrap it in a modal class.

src/styles/components/collapsible-error-message.less
mlunoe: This is maybe a better example: https://github.com/dcos/dcos-ui/blob/master/src/styles/components/page-header.less#L96

src/styles/components/collapsible-error-message.less
```diff
@@ -0,0 +1,54 @@
...
+      // We don't hide the border because this will
+      // introduce 1 px shift in the bottom content
+      border-bottom: color-lighten(@red, 90) 1px solid;
+    }
+    .collapsible-floating > div {
+      border-bottom: @red 1px solid;
+    }
+  }
+
+}
```
mlunoe: unix newline, please. You can add it to your editor to do it automatically

mlunoe:
Hope this gives you some pointers

wavesoft:
Wow, excellent pointers @mlunoe . I think you clarified most of my questions! Will work on the comments tomorrow morning.

src/js/components/CollapsibleErrorMessage.js
```diff
@@ -0,0 +1,110 @@
...
+            id="yield"
+            size="mini"
+            className="icon-alert icon-margin-right"
+            color="red" />
+          {message}
+          {(() => {
+            if (isDetailed) {
+              return (
+                  <span>
+                    &nbsp;
```
rcorral: Let's not use &nbsp;
There's gotta be a better way.

mlunoe: I was thinking the same thing, but couldn't come up with a nice solution on the spot

src/js/components/CollapsibleErrorMessage.js
wavesoft: Done

src/js/components/CollapsibleErrorMessage.js
wavesoft: Done, thanks!

src/js/components/CollapsibleErrorMessage.js
wavesoft: Done

src/js/components/CollapsibleErrorMessage.js
wavesoft: Done

src/js/components/CollapsibleErrorMessage.js
wavesoft: Done

src/js/components/CollapsibleErrorMessage.js
wavesoft: Done

src/js/components/modals/ServiceFormModal.js
wavesoft: It was mainly for readability purposes, but I guess even if the string spans in more than 1 line HTML don't care. 😄

src/styles/components/collapsible-error-message.less
wavesoft: Ok, this is up for discussion. Since this is supposed to be re-usable it can be used either in a modal or on the main page body. But if it's used in the page body *and* a modal is shown on top, the floater would be visible on top of the modal.

I guess a property `modal=true` can be useful in order to select a `z-index` above the modal or below? Is there any better solution?

src/styles/components/collapsible-error-message.less
wavesoft: Done

src/styles/components/collapsible-error-message.less
wavesoft: Done!

src/js/components/CollapsibleErrorMessage.js
wavesoft: I think it's better to add a whitespace after the message. More than one whitespace won't appear in HTML either ways ;)

src/styles/components/collapsible-error-message.less
wavesoft: In this case, this is not really used for per-screen-size spacing unit. It's more like a reference value for an small spacing around the label.

When I try the `proper` spacing implementation, it looks kind of weird on large screen sizes:
image

wavesoft:
I think I addressed all the comments. Let me know if there is any other question.

wavesoft:
I introduced the `className` property through which you can now style the error component. By default it only shows the main message and collapses the details.

src/js/components/CollapsibleErrorMessage.js
```diff
@@ -108,13 +108,9 @@ class CollapsibleErrorMessage extends React.Component {
...
     }
 
-    // Make sure at least one whitespace character
-    // exists after the message string
-    message += ' ';
-
     // Render the fixed part of the message
     return (
-        <div key="fixed" className="collapsible-fixed text-align-center text-danger flush-top">
+        <div className="collapsible-fixed text-align-center text-danger flush-top">
```
orlandohohmeier: It's not `fixed` anymore, is it?

wavesoft: It is, but the key is not needed any more, since I am not using the array approach.
The `floating` is now called `toggled`

src/js/components/CollapsibleErrorMessage.js
wavesoft: Even better one, I introduced the `collapsible-toggle-label` class that adds a left-margin, plus fixes the label width so it doesn't resize the content when it's toggled.

src/js/components/CollapsibleErrorMessage.js
orlandohohmeier: Shouldn't we then rename the class `collapsible-fixed` as well?

wavesoft: Well, that part doesn't hide, so it's still fixed, right?

src/styles/components/collapsible-error-message.less
```diff
@@ -0,0 +1,89 @@
+
+//
+// The CollapsibleErrorMessage is a component used on top
+// of forms or on other places to display a main error
+// message and a toggled drop-down with the error details
+//
```
jfurrow: Since all of our LESS is being refactored in an upcoming release, these comments don't need to change. But just a forewarning, in the future we are planning to adhere to three tiers of comment blocks. Stay tuned for a CSS style guide :)

src/styles/components/collapsible-error-message.less
```diff
@@ -0,0 +1,89 @@
...
+    // Prepare for floating the toggled part by
+    // simply adding position: absolute
+    > div {
+
+      // Some default padding
+      padding: (@base-spacing-unit / 2) @base-spacing-unit;
+
+      // Occupy maximum space
+      left: 0;
+      width: 100%;
```
jfurrow: Let's alphabetize these properties.

src/styles/variables/variables-z-index.less
```diff
@@ -35,6 +35,8 @@ Z-Index Depths
 @z-index-side-panel-container:                                                  @z-index-side-panel + 4;
 @z-index-side-panel-expand-button:                                              @z-index-side-panel + 10;
 @z-index-sidebar:                                                               @z-index-header;
+// The error message comes in 'body' and 'modal' varieties
+@z-index-errormsg-floater-modal:												@z-index-modal-container + 1;
```
jfurrow: Let's add more spaces to get the value of the variable to be at the 80th column.

src/styles/components/collapsible-error-message.less
wavesoft: Done

src/styles/variables/variables-z-index.less
wavesoft: Done

src/styles/components/collapsible-error-message.less
wavesoft: Good to know! 😄

src/js/components/CollapsibleErrorMessage.js
```diff
@@ -0,0 +1,231 @@
...
+import Icon from './Icon';
+
+import React from 'react';
+import classNames from 'classnames';
+
+/**
+ * Methods to bind in 'this' context
+ */
+const METHODS_TO_BIND = [
+  'getDetailLink',
```
mlunoe: I think you only need `toggleExpanded` here

src/js/components/CollapsibleErrorMessage.js
```diff
@@ -0,0 +1,231 @@
...
+  /**
+   * Toggle the expanded state of the collapsible error bar
+   */
+  toggleExpanded() {
+    this.setState({
+      expanded: !this.state.expanded
+    });
+
+    // Call onToggle callback
+    this.props.onToggle(this.state.expanded);
```
mlunoe: don't you want this to happen on the `setState` callback?

src/js/components/CollapsibleErrorMessage.js
```diff
@@ -0,0 +1,231 @@
...
+  }
+
+  /**
+   * [Custom component] The list of errors in the details
+   *
+   * @returns {React.Component} - The rendered content
+   */
+  getDetailMessages() {
+    let {details} = this.props;
+    return details.map(function (message) {
```
mlunoe: #nit newline before `return`

src/js/components/CollapsibleErrorMessage.js
```diff
@@ -0,0 +1,231 @@
...
+      let msg = message.toString();
+      return (
```
mlunoe: #nit newline before `return`

src/js/components/CollapsibleErrorMessage.js
```diff
@@ -0,0 +1,231 @@
...
+          {this.getToggledPart()}
+        </div>
+      );
+
+  };
+
+};
+
+CollapsibleErrorMessage.defaultProps = {
+  className: '',
```
mlunoe: This is strictly not necessary as it is just passed to `classNames(`, which will ignore `undefined`. Just a suggestion, and not a requirement! :)

src/js/components/CollapsibleErrorMessage.js
wavesoft: You are right. These were leftovers of a previous version. Removed.

src/js/components/CollapsibleErrorMessage.js
wavesoft: Done

src/js/components/CollapsibleErrorMessage.js
wavesoft: Done.

src/js/components/CollapsibleErrorMessage.js
wavesoft: I am not following you here... where do you refer to? Do you mean the `componentDidUpdate` callback that will be triggered after I updated the state?

The `onToggle` is used so the hosting component can hook a possible UI update callback according to the expanded status.

src/js/components/CollapsibleErrorMessage.js
wavesoft: You are right, but since it's a property I wanted to have a default value. `className : undefined` might also be an acceptable option I guess, but it looks a bit uglier.

Out of curiosity, is `undefined` or `''` causing bigger (the tiniest, neglectable, yet there) performance impact?

mlunoe: Not at all! AND I do like not having changing types in our variables, hence only a suggestion.
And thinking about it I think the argument of having variables that don't change type is better than 'faster to code'!
Sail on captain! ⛵

src/js/components/CollapsibleErrorMessage.js
mlunoe: Yes, but you are passing an old state value through to the function. `setState` is asynchronous, so you probably want to do something like:
```js
toggleExpanded() {
  this.setState({expanded: !this.state.expanded}, () => {
    // Call onToggle callback
    this.props.onToggle(this.state.expanded);
  });
}
```

Or even better, just store the variable:
```js
toggleExpanded() {
  let expanded = !this.state.expanded;
  this.setState({expanded});

  // Call onToggle callback
  this.props.onToggle(expanded);
}
```

wavesoft: Damn, you are right! To my defence I initially didn't plan to add an argument to the callback, but I thought that having it can only be beneficial 😄 
Fixed!

src/js/components/CollapsibleErrorMessage.js
```diff
@@ -0,0 +1,228 @@
...
+      return null;
+    }
+
+    // Get text
+    let moreLess = 'more';
+    if (this.state.expanded) {
+      moreLess = 'less';
+    }
+
+    // Render
```
rcorral: @wavesoft love a lot of the comments. A good rule of thumb that we use is: If the code is self explanatory, there's no need for a comment.

src/js/components/CollapsibleErrorMessage.js
```diff
@@ -0,0 +1,228 @@
+
```
rcorral: #nit let's not have these beginning lines.

src/js/components/CollapsibleErrorMessage.js
```diff
@@ -0,0 +1,228 @@
...
+          className="collapsible-toggle-label  clickable"
+          onClick={this.toggleExpanded} >
+          (Show {moreLess})
+        </span>
+      );
+
+  }
+
+  /**
+   * [Custom component] The list of errors in the details
```
rcorral: What does [Custom component] mean?

src/js/components/CollapsibleErrorMessage.js
```diff
@@ -0,0 +1,228 @@
...
+        );
+    });
+  }
+
+  /**
+   * [Custom component] The fixed message part
+   *
+   * @returns {React.Component} - The rendered content
+   */
+  getFixedPart() {
```
rcorral: I'm not sure what Part means.
Would `getFixedMessage` be a better name for this method?

src/js/components/CollapsibleErrorMessage.js
```diff
@@ -0,0 +1,228 @@
...
+        </div>
+      );
+  }
+
+  /**
+   * [Custom component] The toggled message part
+   *
+   * @returns {React.Component} - The rendered content
+   */
+  getToggledPart() {
```
rcorral: Could this one be `getCollapsible` message?

src/js/components/CollapsibleErrorMessage.js
```diff
@@ -0,0 +1,228 @@
...
+        </div>
+      );
+
+  }
+
+  /**
+   * React.js Render Function
+   *
+   * @returns {React.Component} - The rendered content
+   */
```
rcorral: Maybe this comment could also be removed.

rcorral: Sorry for all the nits!

src/js/components/modals/ServiceFormModal.js
```diff
@@ -403,37 +405,28 @@ class ServiceFormModal extends mixin(StoreMixin) {
...
-        <div className="error-field text-danger">
-          <h4 className="text-align-center text-danger flush-top">
-            App is currently locked by one or more deployments. Press the button
-            again to forcefully change and deploy the new configuration.
-          </h4>
-        </div>
+        <CollapsibleErrorMessage
+          className="error-for-modal"
+          message="App is currently locked by one or more deployments.
+            Press the button again to forcefully change and deploy the new configuration." />
```
rcorral: For these parameters that are strings and where you want a line break, it is preferred that we use this syntax, so that it doesn't cause the diff to look like this with the red background.
```
message={`App is currently...`}
```

src/styles/index.less
```diff
@@ -165,3 +165,4 @@ Components
 @import "components/tree-list.less";
 @import "components/typography.less";
 @import "components/user-dropdown-menu.less";
+@import "components/collapsible-error-message.less";
```
rcorral: Would be good to alphabetize this import, unless it needs to be applied at the end.

src/js/components/CollapsibleErrorMessage.js
wavesoft: Done

src/js/components/CollapsibleErrorMessage.js
wavesoft: @rcorral I am also for it. Some times I tend to comment just for visually balancing the code/comment ratio and I might be overdoing it. I will try to keep it only for the important parts and only for the *why*s

src/js/components/CollapsibleErrorMessage.js
wavesoft: The naming comes from the fact that the actual message is composed of two parts : the `fixed` part and the `toggled` part.

src/js/components/CollapsibleErrorMessage.js
wavesoft: This was partially the copy-paste daemon. The initial comment was `[Fragment]` but then I copy-pasted the wrong description. I wanted to denote that this function renders a fragment of the total.

src/js/components/CollapsibleErrorMessage.js
wavesoft: ok, I went with `getFixedMessagePart`

src/js/components/CollapsibleErrorMessage.js
wavesoft: I eventually went with `getCollapsibleMessagePart`

src/js/components/CollapsibleErrorMessage.js
wavesoft: Personally, I would suggest to go with adding the JSDoc comment to all of the function headers, even though the comments are be pointless. For visual coherence and documentation reasons. But of course if this does not follow the design guidelines please let me know.

src/js/components/modals/ServiceFormModal.js
wavesoft: Thanks for the tip! Done.

src/styles/index.less
wavesoft: Yup, you are right. Done.

src/js/components/CollapsibleErrorMessage.js
rcorral: Let's talk about as a team next Thursday and figure out what the majority prefers!

orlandohohmeier:
error-rendering

@wavesoft any chance that we can fix the error message position? Otherwise one might miss it if there are lots of options/properties.

wavesoft:
@orlandohohmeier Done. I fixed it on the header, so it's outside the scrollable region. 

fixed-error mov

src/js/components/CollapsibleErrorMessage.js
```diff
@@ -0,0 +1,227 @@
...
+  getShowDetailsLink() {
+    let {message, details} = this.props;
+    let isDetailed = message && details && details.length;
+
+    // Check if we must not show the detail link at all
+    if (!isDetailed) {
+      return null;
+    }
```
pierlo-upitup: I'd argue there is no need for the `isDetailed` variable, we could just 
```
let {message, details} = this.props;
if (message != null && details != null && details.length) {
  return null;
}
```
but that's really minor knittin' 😛

wavesoft: What about:

```javascript
if (!message || !details || !details.length) {
}
```

Because `message` could be also an empty string

src/js/components/CollapsibleErrorMessage.js
pierlo-upitup: Yeah I agree with @rcorral – and I'd actually say comments should be added sparsely and only when they actually add value.
There is a cost in maintaining comments relevant and up to date, and there is nothing worse than an outdated comment that is actually misleading the reader!

For example, the comment `// Render` (while not being a potential liability), adds absolutely zero value, because the `render` function is self-explanatory.

The other comment below, at line 103 instead:
```
/**
* [Fragment] The fixed message part
*
* @returns {React.Component} - The rendered content
*/
getFixedMessagePart() { // ...
```
makes me wonder about two things:
1) it seems to be introducing a very specific "syntax" (`[Fragment]` etc). In this case I'd argue it's introducing a style that seems inconsistent with other components, but maybe this is what the team agrees for the future? In any case I'd suggest tracking the effort to achieve consistent commenting style in the codebase
2) this function may actually return `null` in certain circumstances. The comment does not cater for that case, though.

All in all I'm not saying this should block the PR – but I wanted to share my thoughts on what I think are useful comments and what is actually unnecessary boilerplate. 

On the other hand, this is a good example of a useful comment:

https://github.com/dcos/dcos-ui/blob/master/src/js/utils/ContainerValidatorUtil.js#L7-L9
In this case, the comment is invaluable because it will tell the "future you" where that regex came from and what it means, so you don't have to spend an hour to figure it out. 

Just my 2 cents!

src/js/components/CollapsibleErrorMessage.js
```diff
@@ -0,0 +1,227 @@
...
+   * @returns {React.Component} - The rendered content
+   */
+  getDetailsListItems() {
+    let {details} = this.props;
+
+    return details.map(function (message) {
+      let msg = message.toString();
+
+      return (
+          <li key={msg}>{msg}</li>
```
pierlo-upitup: Let's not use `msg` as our key. If your errors are showing an identical string, this will mess up with React's rendering efforts.

In this case, the easiest will be:
```
return details.map(function (message, i) {
// ...
return  <li key={i}>{msg}</li>
```

src/js/components/CollapsibleErrorMessage.js
```diff
@@ -0,0 +1,227 @@
...
+   * @returns {React.Component} - The rendered content
+   */
+  getFixedMessagePart() {
+    let {message} = this.props;
+    let isVisible = !!message;
+
+    // If not visible, just exit
+    if (!isVisible) {
+      return null;
+    }
```
pierlo-upitup: Same as above, I don't think we need a dedicate variable for this.

src/js/components/CollapsibleErrorMessage.js
wavesoft: Good point. Done.

src/js/components/CollapsibleErrorMessage.js
wavesoft: Done!

pierlo-upitup:
I've added a few comments but this is great work and LGTM! :shipit:

PR #916, opened 2016-08-05
Making sure we are not loosing focus in service form network tab

mlunoe:
~~Related PR: https://github.com/dcos/dcos-ui/pull/561~~

~~This PR doesn't solve the whole problem, but makes it a little better, i.e. only happen for one field~~

This PR fixes a small issue, user would loose focus in the create service form when entering new ports.

### DIFF
--- a/src/js/components/SchemaForm.js
+++ b/src/js/components/SchemaForm.js
@@ -99,12 +99,14 @@ class SchemaForm extends mixin(StoreMixin, InternalStorageMixin) {
     // Default method.
   }
 
-  handleAddRow(prop, definition, newDefinition) {
+  handleAddRow(prop, definition, newDefinition, index) {
     let propID = Util.uniqueID(prop);
     newDefinition = FormUtil.getMultipleFieldDefinition(
       prop,
       propID,
-      newDefinition
+      newDefinition,
+      undefined,
+      index
     );
 
     let deleteButtonTop = Object.values(definition.itemShapes || {})
@@ -147,6 +149,7 @@ class SchemaForm extends mixin(StoreMixin, InternalStorageMixin) {
   }
 
   getAddNewRowButton(prop, generalDefinition, definition, labelText = '') {
+    let index = this.getIndexFromDefinition(generalDefinition);
     let label = 'Add New Line';
 
     if (labelText !== '') {
@@ -158,14 +161,28 @@ class SchemaForm extends mixin(StoreMixin, InternalStorageMixin) {
         <div className="row form-row-element">
           <a
             className="clickable row"
-            onClick={
-              this.handleAddRow.bind(this, prop, generalDefinition, definition)
-            }>
+            onClick={this.handleAddRow.bind(
+              this,
+              prop,
+              generalDefinition,
+              definition,
+              index
+            )}>
             + {label}
           </a>
         </div>
       </div>
     );
+  }
+
+  getIndexFromDefinition(definition) {
+    return definition.reduce(function (total, item) {
+      if (Array.isArray(item)) {
+        return total + 1;
+      }
+
+      return total;
+    }, 0);
   }
 
   getRemoveRowButton(generalDefinition, prop, id, title = null) {
--- a/src/js/components/SchemaForm.js
+++ b/src/js/components/SchemaForm.js
@@ -176,6 +176,8 @@ class SchemaForm extends mixin(StoreMixin, InternalStorageMixin) {
   }
 
   getIndexFromDefinition(definition) {
+    // This counts the number of arrays in the definition to determine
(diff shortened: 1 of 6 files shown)


mesosphere-web:
Reviewer: @jfurrow

src/js/components/ServiceForm.js
```diff
@@ -23,7 +23,7 @@ const DUPLICABLE_FIELDS_TO_WATCH = {
   },
   ports: {
     fields: ['discovery', 'protocol'],
-    blurOnly: ['name', 'lbPort'],
+    blurOnly: ['name'],
```
jfurrow: Could you add a short explanation as to why we don't need `lbPort` in this array? To be honest I don't really know what it's doing anyway...

mlunoe:
As it turns out, we need to watch lbPort after all

mlunoe:
ping @daiyiwhale!

mlunoe:
Removing Review label after changes

daiyiwhale:
Ahh you did it! amazing

daiyiwhale:
I am receiving this warning: `The specified value "Not Enabled" is not a valid number. The value must match to the following regular expression: -?(\d+|\d+\.\d+|\.\d+)([eE][-+]?\d+)?` when `Discovery` is not checked.

jfurrow:
I'm receiving the same warning:

orlandohohmeier:
I'm getting the same warning. It seems like we're not toggling the input type properly. Aside from that it works great - awesome job!

mlunoe:
Yes, I am looking into that. Also, there is some mysterious going on with the removal of 😞
Removing the first row, will remove all.

src/js/schemas/service-schema/Networking.js
```diff
@@ -83,11 +83,9 @@ const Networking = {
         });
       },
       filterProperties(service = {}, instanceDefinition, model) {
-        let properties = Networking
-          .properties
-          .ports
-          .itemShape
-          .properties;
+        let {properties} = Networking.properties.ports.itemShape;
```
mlunoe: All in this file is basically just making it easier to understand what is going on

mlunoe:
The warning problem is now fixed, ref: https://github.com/dcos/dcos-ui/pull/916#issuecomment-239280350

mlunoe:
Tests fixed

mlunoe:
ping @jfurrow

src/js/components/SchemaForm.js
```diff
@@ -99,12 +99,14 @@ class SchemaForm extends mixin(StoreMixin, InternalStorageMixin) {
...
-  handleAddRow(prop, definition, newDefinition) {
+  handleAddRow(prop, definition, newDefinition, index) {
     let propID = Util.uniqueID(prop);
     newDefinition = FormUtil.getMultipleFieldDefinition(
       prop,
       propID,
-      newDefinition
+      newDefinition,
+      undefined,
+      index
```
jfurrow: In a separate PR, I would love for us to pass an object with these properties, rather than carefully construct this list of arguments each time we call the `#getMultipleFieldDefinition` method. But again, that's not in scope of this PR!

src/js/utils/__tests__/FormUtil-test.js
```diff
@@ -21,21 +21,25 @@ describe('FormUtil', function () {
...
+      expect(result[1].name).toEqual('variable[2].value');
     });
 
     it('does not modify the definition', function () {
       let result = FormUtil.getMultipleFieldDefinition(
         'variable',
         1,
-        this.definition
+        this.definition,
+        null,
```
jfurrow: In `SchemaForm` we pass `undefined` to this method, whereas here we pass `null`. It looks like they both produce the same result, so I would prefer to be consistent (and `null` makes more sense than `undefined` to me)

jfurrow:
@mlunoe This seems to work. I'm struggling a bit to understand the logic but the confusion wasn't introduced in this PR. I left a minor comment, but otherwise it's good to go!

src/js/utils/__tests__/FormUtil-test.js
mlunoe: Done!

src/js/components/SchemaForm.js
mlunoe: Yeah, let's wait with that to the refactor. I don't want to mess around more with this than I actually have to

mlunoe:
@jfurrow Both @orlandohohmeier and I have tested this fairly throughly. The logic _is_ really hard to follow, and it took me a while to get there. I have tried to add descriptions where I felt they were needed.

jfurrow:
Thanks @mlunoe! Looks good.

PR #923, opened 2016-08-08
Fix spelling of 'occurred'

philipnrmn:
Spelling error in the Cosmos error modal, thanks to @ssk2 for the report. 

I also fixed up the same error in some of the tests while I was at it.


philipnrmn:
Re-rolling assignees since Mat's out.

mlunoe:
😂 Thanks for fixing this!

PR #926, opened 2016-08-08
DCOS-8110 Add framework version to service details page

pierlo-upitup:
image

I'd like some thoughts on the following:

1) the `ServiceInfo` uses the `PageHeader` component to render the framework name (`title` prop)
2) the `PageHeader` component passes the `title` down to the `CollapsibleString`
3) the `title` MUST be a string for the `CollapsibleString` magic to work
4) I can’t see a way to make this work by having the framework’s name and version be part of the `title` as seen in the designs

The only option I see here without refactoring anything is to show the version below the framework name, perhaps before the Status bar.

@leemunroe @jfurrow any thoughts?

### DIFF
--- a/src/js/components/PageHeader.js
+++ b/src/js/components/PageHeader.js
@@ -140,7 +140,7 @@ PageHeader.propTypes = {
   icon: React.PropTypes.node,
   navigationTabs: React.PropTypes.node,
   subTitle: React.PropTypes.node,
-  title: React.PropTypes.string,
+  title: React.PropTypes.node,
 
   className: classPropType,
   pageHeaderClassNames: classPropType,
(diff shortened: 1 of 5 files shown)


jfurrow:
I agree with your assessment; without refactoring, the version needs to be displayed below the title.

One solution (which involves a bit of refactoring) would be to accept both strings and React components in the `PageHeader` on the `title` prop. If it's a string, then it could use `CollapsingString` as it does now; otherwise it would just render the component. This would give more control to the parent of `PageHeader`.

pierlo-upitup:
@jfurrow yeah – otoh I think PageHeader is probably trying to take too much on its shoulders. I'd say it should just accept children and then it's up to the parent component to decide if it sould wrap the title in a CollapsibleString or not. This would give us more composability, and for example we could collapse the framework name but not its version...

@leemunroe any input from you would be great.

leemunroe:
Just to confirm here, what we're seeing in the screenshot above is **not** possible?

If we need to move it down for now, I would opt to show it right of the status bar. The intent being I'd like to see the status and status bar placement consistency across each service detail view.

PR #927, opened 2016-08-08
DCOS-8945 Allow excluding certain items from being selected in checkbox table

pierlo-upitup:
Exclude the non-running `jobRuns` from being selected and stopped, which would produce a 404 error.

disable

### DIFF
--- a/src/js/components/CheckboxTable.js
+++ b/src/js/components/CheckboxTable.js
@@ -113,8 +113,13 @@ class CheckboxTable extends React.Component {
   }
 
   renderCheckbox(prop, row) {
-    let {checkedItemsMap, uniqueProperty} = this.props;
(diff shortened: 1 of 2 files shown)


orlandohohmeier:
reset-selection-issue mov

We need to reset/update the `checkedItemsMap` when an item becomes disabled. I wonder if it's easier to keep track of these things if we combine the properties...

wavesoft:
I suggested a quick solution. I am working on a more proper one, following what @orlandohohmeier said before, however this touches many components and it will take some time. So if we are in a hurry, merge, otherwise wait 😄

wavesoft:
Keeping it like this for now.

orlandohohmeier:
LGTM! 🚢

PR #931, opened 2016-08-08
Fix breadcrumb route to universe package.

rcorral:
This PR fixes a few things:
* Fix breadcrumb route to universe package.
* Allows passing any additional parameter to BreadcrumbSegmentLink through `this.props.route`
* Fixes request to fetch package description, the version was being accessed incorrectly.


wavesoft:
Looks good, doesn't seem to affect other places where `Breadcrumb` is used.
However I see that the test is not checking anything about the breadcrumb (but I guess this is more for the integration test that the unit test?)

wavesoft:
Also, I am not sure if this is this PR's scope, but if you request an invalid version it simply breaks:

image

PR #933, opened 2016-08-08
Fixes color for disk.

rcorral:
Philip(p)s! I need your help...will this be a problem with the progressbars in the services page?
I'm fixing this because it was causing the disk progressbars in the Nodes page to look like the overcapacity ones.
See:
img


philipnrmn:
Works nicely and doesn't interfere with the services page.

Poltergeist:
This works nicely. Thanks for the fix.

PR #936, opened 2016-08-08
Adding node and npm version check

mlunoe:
This check will run on `npm preinstall` and `npm start`

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,8 +17,8 @@
   },
   "license": "Apache-2.0",
   "engines": {
-    "node": "4.x.x",
-    "npm": "3.x.x"
+    "node": "4.4.4",
+    "npm": "3.9.1"
   },
   "config": {
     "port": 4200,
@@ -109,8 +109,10 @@
     "webpack-notifier": "1.3.0"
   },
   "scripts": {
+    "check-env": "sh ./scripts/validate-engine-versions",
     "clean": "rm -rf dist",
     "shrinkwrap": "npm shrinkwrap --dev && ./node_modules/.bin/gulp fixShrinkwrap",
+    "preinstall": "npm run check-env",
     "prebuild-assets": "./node_modules/.bin/gulp checkDependencies && npm install compression-webpack-plugin@latest",
     "build-assets": "npm run clean && NODE_ENV=production webpack --config ./webpack/webpack.production.babel.js",
     "postbuild-assets": "rm -rf node_modules/compression-webpack-plugin",
@@ -118,11 +120,11 @@
     "build-with-external-plugins": "npm run lint && npm run lint-plugins && npm test && npm run build-assets",
     "lint": "bash -c 'opts=$@ ; if [[ $# -lt 1 ]]; then opts=\"src/\\*\\*/\\*.js\"; fi ; ./node_modules/.bin/eslint --quiet ${opts}' 0",
     "lint-plugins": "PLUGINS=$npm_config_externalplugins; npm run lint -- \"$PLUGINS/**/*.js\"",
-    "prebuild": "./scripts/validate-tests",
+    "prebuild": "sh ./scripts/validate-tests",
     "serve": "npm run build-assets && ./node_modules/.bin/gulp serve",
     "serve-notify": "export NOTIFY=true && npm run serve",
     "scaffold": "./node_modules/.bin/gulp ensureConfig && ./node_modules/.bin/gulp ensureDevProxy",
-    "start": "NODE_ENV=development webpack-dev-server --config ./webpack/webpack.dev.babel.js --hot --progress --colors --content-base ./build --host 0.0.0.0 --port $npm_package_config_port",
+    "start": "npm run check-env && NODE_ENV=development webpack-dev-server --config ./webpack/webpack.dev.babel.js --hot --progress --colors --content-base ./build --host 0.0.0.0 --port $npm_package_config_port",
     "test": "node jest/gen-config.js && NODE_PATH=node_modules jest --config=jest/config.json --no-cache",
     "testing": "NODE_ENV=testing webpack-dev-server --config ./webpack/webpack.dev.babel.js --progress --colors --content-base ./build --port $npm_package_config_port"
   },
--- a/package.json
+++ b/package.json
@@ -17,8 +17,8 @@
   },
   "license": "Apache-2.0",
   "engines": {
-    "node": "4.4.4",
-    "npm": "3.9.1"
+    "node": "4.4.x",
+    "npm": "3.9.x"
   },
   "config": {
     "port": 4200,
--- a/scripts/validate-engine-versions
+++ b/scripts/validate-engine-versions
@@ -1,4 +1,16 @@
 #!/usr/bin/env bash
+#
+# This script will check the following versions provided in package.json,
+# based on current versions of node and npm
+# {
+#   ...
+#   "engines": {
+#     "node": "4.4.x",
+#     "npm": "3.9.x"
+#   }
+#   ...
+# }
+#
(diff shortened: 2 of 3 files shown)


philipnrmn:
Are we OK with this not functioning on Windows?

mlunoe:
yes :grin:

package.json
```diff
@@ -109,20 +109,22 @@
...
+    "preinstall": "npm run check-env",
     "prebuild-assets": "./node_modules/.bin/gulp checkDependencies && npm install compression-webpack-plugin@latest",
     "build-assets": "npm run clean && NODE_ENV=production webpack --config ./webpack/webpack.production.babel.js",
     "postbuild-assets": "rm -rf node_modules/compression-webpack-plugin",
     "build": "npm run lint && npm test && npm run build-assets",
     "build-with-external-plugins": "npm run lint && npm run lint-plugins && npm test && npm run build-assets",
     "lint": "bash -c 'opts=$@ ; if [[ $# -lt 1 ]]; then opts=\"src/\\*\\*/\\*.js\"; fi ; ./node_modules/.bin/eslint --quiet ${opts}' 0",
     "lint-plugins": "PLUGINS=$npm_config_externalplugins; npm run lint -- \"$PLUGINS/**/*.js\"",
-    "prebuild": "./scripts/validate-tests",
+    "prebuild": "sh ./scripts/validate-tests",
```
rcorral: Why is this prefixed with `sh`?

scripts/validate-engine-versions
```diff
@@ -0,0 +1,30 @@
...
+#   }
+#   ...
+# }
+#
+PACKAGE_NODE_VERSION=$(node -p -e "require('./package.json').engines.node")
+NODE_VERSION=$(node -v)
+if [[ "v${PACKAGE_NODE_VERSION%.*}" == "${NODE_VERSION%.*}" ]]; then
+  echo "Node version is valid ($NODE_VERSION)"
+else
+  echo "ERROR: Node version v${PACKAGE_NODE_VERSION%.*}.x does not match required version ${NODE_VERSION%.*}.x"
```
jfurrow: These two version variables need to be flipped: `echo "ERROR: Node version v${NODE_VERSION%.*}.x does not match required version ${PACKAGE_NODE_VERSION%.*}.x"`

scripts/validate-engine-versions
```diff
@@ -0,0 +1,30 @@
...
+  exit 1;
+fi
+
+PACKAGE_NPM_VERSION=$(node -p -e "require('./package.json').engines.npm")
+NPM_VERSION=$(npm -v)
+if [[ "${PACKAGE_NPM_VERSION%.*}" == "${NPM_VERSION%.*}" ]]; then
+  echo "NPM version is valid (v$NPM_VERSION)"
+else
+  echo "ERROR: Node version v${NPM_VERSION%.*}.x does not match required version ${NPM_VERSION%.*}.x"
```
jfurrow: These two variables are the same, the second one should be `PACKAGE_NPM_VERSION`

package.json
mlunoe: To make sure it runs as shell script, i.e. more reliable

rcorral: In which cases would it not do that?

scripts/validate-engine-versions
mlunoe: LOL, fixed

scripts/validate-engine-versions
mlunoe: LOL, fixed

mlunoe:
Here is an example of the npm error printout:

package.json
mlunoe: I've been burnt by this in the passed, where it would run bash on mac, but shell in linux environments and fail. Just wanted to make sure that it was working for the environments we were using.
Do you have a suggestion to make it better?

mlunoe: Maybe we should change it to use `#!/bin/sh`:
http://stackoverflow.com/a/5725402/1008519

rcorral: Yeah let's do the shebang :)

mlunoe: ok, done

mlunoe:
ping @philipnrmn @daiyiwhale

package.json
philipnrmn: The `sh` is necessary unless the script has execute permissions.

package.json
```diff
@@ -17,8 +17,8 @@
   },
   "license": "Apache-2.0",
   "engines": {
-    "node": "4.x.x",
-    "npm": "3.x.x"
+    "node": "4.4.x",
+    "npm": "3.9.x"
```
philipnrmn: Any reason to use 3.9 and not 3.10?

mlunoe: Not particularly, other than that it is what we have been using for a while now, and not updated lately.
Could we update that separately, since that requires us to update our builds on dcos and docs-enterprise repos

package.json
rcorral: it does have execute permissions! :(

mlunoe: Follow up PR here: https://github.com/dcos/dcos-ui/pull/963

rainbowland302:
Hi, I have some problem when build the dcos-ui using windows.
1. the scripts file cannot be run for task prebuild.
_commit a511032d0f7969678932feaaa2b64f8ecad9d09a_
This commit try to fix it but it is reverted.
2.  $npm_package_config_port not works in windows.
_https://github.com/ngParty/Angular1-scaffold/issues/34_
3. There's also some dependency fails in windows.
I'd like to know do you have the plan to support the build on windows.
Sorry for the trouble, hope for your replies. Thank you.

mlunoe:
1. The prebuild is running a gulp task `checkDependencies` and installs `compression-webpack-plugin@latest` npm module. You can skip this step and just install manually like so: `npm i compression-webpack-plugin@latest`

  The commit you are pointing to is just checking the node and npm versions.

  What you can do here is to run the script by prefixing it with `sh` like so: `sh ./scripts/validate-engine-versions` in powershell. Otherwise, you can just skip this step yourself and manually check that you are running the node and npm versions listed in under `engines` in package.json.

2. As it states in the link you posted you can replace these config variables with manual values:
i.e. set `$npm_package_config_port` to 4200

3. Could you list which dependencies that fail, then I can try to help you.

My suspicion is that some of the environment variables are also not being set correctly. You could use this module and prefix all the environment variable assignments with `cross-env`: https://www.npmjs.com/package/cross-env

It would be awesome if you could take a stab at making the build work for Windows!

mlunoe:
The environment checks we could omit when running Windows (with `IF`-statements), that way our CI will still run the checks for you (NB: this will make the feedback loop larger for developers on Windows, but that is better than a non-working build :blush:)

rainbowland302:
Thank you for your kind reply. As mentioned before I've got following dependency fails:
1. _Node version is valid (v4.4.6) NPM version is valid (v3.9.5)
npm WARN optional Skipping failed optional dependency /chokidar/fsevents:
npm WARN notsup Not compatible with your operating system or architecture: fsevents@1.0.14_

2. _npm start and I got 'Uncaught TypeError: ReactHotAPI is not a function' in chrome console._
This is related to the react-hot-loader. It seems not a must, so I just run without --hot.

3. Then I got 'Uncaught TypeError: (0 , _typeof3.default) is not a function' in console.
This seems to be related to BABEL-6.

mlunoe:
Hey @rainbowland302 
1. This should be fine. That is just warnings and doesn't fail the build, so you can ignore that part. 

2. Yes, you can run without `--hot`!

3. Hmmm... it seems to be a bug in Babel 6, and might require an update of Babel, which is a little out of scope for us right now. Would you be willing to give it a try?

rainbowland302:
Sure, I'll have a try. But I'm afraid there will be other more failed dependencies. Perhaps, I may use Linux instead.

PR #937, opened 2016-08-08
Fix max listeners limit.

rcorral:
When rendering many dial chars, the Chart component listens to sidebar width change. If there's many charts, it'll throw a warning.

### DIFF
--- a/src/js/stores/SidebarStore.js
+++ b/src/js/stores/SidebarStore.js
@@ -99,4 +99,7 @@ class SidebarStore extends GetSetBaseStore {
 
 }
 
-module.exports = new SidebarStore();
+const store = new SidebarStore();
+store.setMaxListeners(1000);
+
+module.exports = store;


src/js/stores/SidebarStore.js
```diff
@@ -99,4 +99,7 @@ class SidebarStore extends GetSetBaseStore {
 
 }
 
-module.exports = new SidebarStore();
+const store = new SidebarStore();
+store.setMaxListeners(1000);
```
jfurrow: Maybe we should use `Infinity` here?

mlunoe:
Hmm... I wonder if we should not have this listener in the chart then?

MatApple:
Since this is only to prevent a warning which will now only be shown in the 99th percentile I'm going to merge.

mlunoe:
I don't think this a good solution. Could we get an actual solution for this?
I actually disagree with this entering the code base

MatApple:
I agree it's not the *best* solution, but it is the simplest. What do you propose?

mlunoe:
To not have the charts fire the sidebar action, but on the parent instead.

PR #938, opened 2016-08-08
[CNVS Upgrade] Add flex properties for .page-body's scroll container

jfurrow:
This fixes the Gemini Scrollbar content when native overlay scrollbars are enabled. The rest of `page/styles.less` still needs to be audited, but I'd like to wait until more of the components have been audited.

### DIFF
--- a/src/styles/components/scrollbar/styles.less
+++ b/src/styles/components/scrollbar/styles.less
@@ -55,24 +55,6 @@
 
   .gm-scrollbar-container {
 
-    &.gm-scrollbar-container-flex {
-
-      position: relative;
-
-      .gm-scroll-view {
-
-        position: absolute;
-
-        & > * {
-
-          position: relative;
-
-        }
-
-      }
-
-    }
-
     &.inverse {
 
       .gm-scrollbar {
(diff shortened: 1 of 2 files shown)


ashenden:
Do we need this style?  I believe I added a style for the gemini component that would do this on a case-by-case basis called `gm-scrollbar-container-flex`.  Maybe I'm missing something extra here.

jfurrow:
Yes, this is necessary. Gemini adds DOM nodes, one of which is `.gm-scroll-view`. When the device does not have native overlay scrollbars, Gemini sets inline `height` and `width` styles on `.gm-scroll-view`. In this case, the styles I added are not relevant: the height and width are determined by Gemini.


When the device has native overlay scrollbars, it does not set the `height` and `width` properties, so the DOM node doesn't grow to fill its parent (in this case the parent is `.page-body`).

jfurrow:
What specifically are you fixing with this? https://github.com/dcos/dcos-ui/blob/cnvs-upgrade/master/src/styles/components/scrollbar/styles.less#L58

Also, here: https://github.com/dcos/dcos-ui/blob/cnvs-upgrade/master/src/styles/components/scrollbar/styles.less#L66 I'm worried about the `> *` selector.

PR #940, opened 2016-08-09
DCOS-9118: Refactor service form modal to only keep service in state

Poltergeist:
This removes the jsonDefinition from the serviceFormModal.


mlunoe:
What did jsonDefinition do before and why can that be replaced?
Just trying to understand this PR all together

Poltergeist:
The JSON definition kept the service definition as a JSON string which was redundant because we were keeping it as well as a service struct which let to a lot of JSON parsing which wasn't necessary. Sorry for the missing explanation.

orlandohohmeier:
LGTM 💯 ✨

PR #942, opened 2016-08-09
Fixes dial charts not having the correct color mappings. refs: DCOS-8917

rcorral:
Also added the ability to show up to 30 different frameworks with the dialcharts, this should last for a while :)


src/styles/components/nodes-grid-view.less
```diff
@@ -78,6 +78,102 @@
...
+          background-color: rgba(121, 174, 163, 1);
+        }
+
+        &.service-color-30 {
+          background-color: rgba(112, 238, 156, 1);
+        }
+
+        &.service-color-31 {
+          background-color: rgba(181, 244, 74, 1);
+        }
```
jfurrow: Just curious, where are these colors coming from?

Also, we don't need to use `rgba` because the alpha channel is being set to 1. But I don't think it really makes a difference, so feel free to leave as-is.

jfurrow: It might be nice to put these colors in variables so we can easily change them in the future.

philipnrmn:
+1 on @jfurrow's comments, otherwise :shipit:

orlandohohmeier:
Looks good to me, but I suggest to change this to one class or inline style to set the `color` and then applying via `currentColor` instead of defining the same colors over and over again.

rcorral:
All set @jfurrow @philipnrmn @orlandohohmeier

jfurrow:
This doesn't implement @orlandohohmeier's request :(

He's saying apply the color like this, once for each unique color: `color: @variable;`
And then apply it to the various properties like this: `fill: currentColor; stroke: currentColor;`

mlunoe:
Arh, yes. That is a good point!

mlunoe:
Re-adding my label. Let's get this in and create a ticket to track to the style changes

rcorral:
To follow up on @jfurrow and @orlandohohmeier's comment https://mesosphere.atlassian.net/browse/DCOS-9680

PR #943, opened 2016-08-09
Fixes visibility store so that it doesn't set isInactive as false when it is active.

rcorral:
The problem was that if you switched between dcos tab and another tab in your browser and back to dcos, we wouldn't clear the timeout and after 30 seconds `isInactive` would be set to false, halting all ajax requests.

ps. this was fun to debug.

cc/ @MatApple


jfurrow:
lol @ "this was fun to debug"

Thanks for the explanation!

MatApple:
Nice find @rcorral 👍

PR #949, opened 2016-08-10
Chris/bug/service icons not displaying

daiyiwhale:
Some service icons were missing. It turns out that some Applications carry their icons in their `DCOS_PACKAGE_METADATA` `label` even though they are not a Framework. And so we will generalise the Service struct (from which both Framework and Application inherit) to check for a metadata label when rendering an icon.

Before: 
(artifactory is an Application)
where is your icon yo

After:
ah yes here


mlunoe:
A bunch of tests are failing on this PR

mlunoe:
Code looks good other than that

PR #950, opened 2016-08-10
DCOS-9170: Add support for unknown values in jobs

Poltergeist:
This adds the support which has been added to services recently to support unknown attributes in the jobs ui.

### DIFF
--- a/src/js/components/modals/JobFormModal.js
+++ b/src/js/components/modals/JobFormModal.js
@@ -10,6 +10,7 @@ import 'brace/ext/language_tools';
 
 import MetronomeStore from '../../stores/MetronomeStore';
 import Job from '../../structs/Job';
+import JobConfig from '../../constants/JobConfig';
 import JobForm from '../JobForm';
 import JobUtil from '../../utils/JobUtil';
 import JobSchema from '../../schemas/JobSchema';
@@ -52,6 +53,16 @@ const responseAttributePathToFieldIdMap = {
   '/schedules/{INDEX}/concurrencyPolicy': 'schedules/{INDEX}/concurrencyPolicy',
   '/schedules/{INDEX}/enabled': 'schedules/{INDEX}/enabled',
   '/labels': 'labels'
+};
+
+let cleanJSONspec = function (jsonDefinition) {
+  return Object.keys(jsonDefinition).filter(function (key) {
+    return !JobConfig.BLACKLIST.includes(key);
+  }).reduce(function (memo, key) {
+    memo[key] = jsonDefinition[key];
+
+    return memo;
+  }, {});
 };
 
 class JobFormModal extends mixin(StoreMixin) {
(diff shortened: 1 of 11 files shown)


src/js/components/modals/JobFormModal.js
```diff
@@ -54,6 +55,16 @@ const responseAttributePathToFieldIdMap = {
   '/labels': 'labels'
 };
 
+let cleanJSONspec = function (jsonDefinition) {
```
wavesoft: If this is similar to [the one in ServiceFormModal.js](https://github.com/dcos/dcos-ui/blob/master/src/js/components/modals/ServiceFormModal.js#L109) shouldn't this be moved to the utility functions?

Poltergeist: I actually thought about something different. I thought we could utilize https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify specifically the toJSON function.

wavesoft:
Besides the minor comment, everything looks good to me.

src/js/components/modals/JobFormModal.js
rcorral: I'm confused by your comment @Poltergeist 
Seems like we could reuse code?

Poltergeist: Yes, and no. Both functions are using a different Config which we could pass as a parameter. I would prefer something like this if we want to have a more generic code:

```JS
function cleanJSONSpec(jsonSpec) {
  return cleanJSON(jsonSpec, JobConfig);
} 

function cleanJSONDefinition(jsonDefinition) {
  return cleanJSON(jsonDefinition, ServiceConfig);
}

function cleanJSON(json, config) {
  return Object.keys(json).filter(function (key) {
    return !config.BLACKLIST.includes(key);
  }).reduce(function (memo, key) {
    memo[key] = json[key];

    return memo;
  }, {});
}
```
This would remove code duplication but also remove the overhead of adding the config as a general parameter which would mean more code duplication.

mlunoe: That works for me :relieved:

src/js/utils/CleanJSONUtil.js
```diff
@@ -0,0 +1,20 @@
+import JobConfig from '../constants/JobConfig';
+import ServiceConfig from '../constants/ServiceConfig';
+
+export function cleanJSONspec(jsonSpec) {
```
orlandohohmeier: Can we rename these utils so that it's obvious that this one is about jobs and the other one about services?

orlandohohmeier:
Aside from the naming issues, solid work. Thanks! Feel free to merge once you've adjusted the naming.

src/js/utils/CleanJSONUtil.js
Poltergeist: I'll go for `cleanJobJSON` and `cleanServiceJSON`

Poltergeist:
Renamed will merge when tests passed.

PR #953, opened 2016-08-10
[CNVS Upgrade] Use new filenames for reactjs-components

jfurrow:


ashenden:
Looks good to me, though a check is failing.

jfurrow:
All of the tests are going to fail and will need to be audited separately. The test rely on classes that have changed, elements are invisible, etc etc.

PR #946, opened 2016-08-10
DCOS-8237: Refactor MarathonStore

daiyiwhale:
Moving appropriate functionality from `MarathonStore` to `ServiceTree`.

PR #955, opened 2016-08-10
Show task health state in detail page

philipnrmn:
Previously, task health status was being inferred from the presence of the Mesos state.json endpoint in `frameworks[].tasks[].statuses[].healthy`. This is not guaranteed to be present. Separately, task health and running status were being conflated such that where a task was in an unhealthy state (eg Killing) the Running state was being shown. 

This PR:
 - introduces a fallback on the Marathon health checks where the Mesos task healthy property is not present
 - adds a 'Healthy' and 'Unhealthy' label to the status column
 - cleans up the state badges classes so that running and unhealthy do not conflict

This PR resolves https://dcosjira.atlassian.net/browse/DCOS-290

Before:
screen shot 2016-08-10 at 4 44 25 pm

After:
screen shot 2016-08-10 at 4 45 09 pm

_NB a separate bug is present in these screenshots where over-capacity tasks are counted twice. This is outside the scope of this PR_

### DIFF
--- a/src/js/components/TaskTable.js
+++ b/src/js/components/TaskTable.js
@@ -215,7 +215,7 @@ class TaskTable extends React.Component {
         <col style={{width: '40px'}} />
         <col />
         <col style={{width: '15%'}} className="hidden-mini" />
-        <col style={{width: '105px'}} />
+        <col style={{width: '115px'}} />
         <col style={{width: '40px'}} className="hidden-medium hidden-small hidden-mini" />
         <col style={{width: '85px'}} className="hidden-mini" />
         <col style={{width: '85px'}} className="hidden-mini" />
--- a/src/js/components/TaskTable.js
+++ b/src/js/components/TaskTable.js
@@ -79,12 +79,18 @@ class TaskTable extends React.Component {
     if (task.statuses == null) {
       return null;
     }
-    const healths = task.statuses.map((status) => status.healthy);
+    const healths = task.statuses.map(function (status) {
+      return status.healthy;
+    });
     const healthDataExists = healths.length > 0 && healths.every(
-      (health) => typeof health !== 'undefined'
+      function (health) {
+        return typeof health !== 'undefined';
+      }
     );
     if (healthDataExists) {
-      return healths.some((health) => health);
+      return healths.some(function (health) {
+        return health;
+      });
     }
 
     return null;
--- a/src/js/components/TaskTable.js
+++ b/src/js/components/TaskTable.js
@@ -39,6 +39,14 @@ class TaskTable extends React.Component {
   }
 
   getStatusValue(task) {
+    let taskHealth = this.getTaskHealth(task);
+    if (taskHealth === true) {
+      return 'Healthy';
+    }
+    if (taskHealth === false) {
+      return 'Unhealthy';
+    }
+
     return TaskStates[task.state].displayName;
   }
 
@@ -70,6 +78,9 @@ class TaskTable extends React.Component {
   }
 
   getTaskHealthFromMesos(task) {
+    if (task.statuses == null) {
+      return null;
+    }
     const healths = task.statuses.map((status) => status.healthy);
     const healthDataExists = healths.length > 0 && healths.every(
       (health) => typeof health !== 'undefined'
@@ -128,7 +139,7 @@ class TaskTable extends React.Component {
       {
         cacheCell: true,
         className,
-        getValue: this.getStatusValue,
+        getValue: this.getStatusValue.bind(this),
         headerClassName: className,
         heading,
         prop: 'status',
--- a/src/js/components/TaskTable.js
+++ b/src/js/components/TaskTable.js
@@ -55,12 +55,8 @@ class TaskTable extends React.Component {
     if (mesosTaskHealth !== null) {
       return mesosTaskHealth;
     }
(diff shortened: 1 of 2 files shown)


src/js/components/TaskTable.js
```diff
@@ -39,9 +39,59 @@ class TaskTable extends React.Component {
...
+
     return TaskStates[task.state].displayName;
   }
 
+  getTaskHealth(task) {
+    let mesosTaskHealth = this.getTaskHealthFromMesos(task);
+    if (mesosTaskHealth !== null) {
+      return mesosTaskHealth;
+    }
+    let marathonTaskHealth = this.getTaskHealthFromMarathon(task);
```
mlunoe: Couldn't we just return `this.getTaskHealthFromMarathon(task)` here, since the fallback is returning `null`

src/js/components/TaskTable.js
```diff
@@ -39,9 +39,59 @@ class TaskTable extends React.Component {
...
+    if (marathonTaskHealth !== null) {
+      return marathonTaskHealth;
+    }
+
+    return null;
+  }
+
+  getTaskHealthFromMarathon(task) {
```
mlunoe: These newly introduced methods seems like a good fit for util functions, could we move them there instead?

src/js/components/TaskTable.js
```diff
@@ -89,7 +139,7 @@ class TaskTable extends React.Component {
       {
         cacheCell: true,
         className,
-        getValue: this.getStatusValue,
+        getValue: this.getStatusValue.bind(this),
```
mlunoe: Let's add this to `METHODS_TO_BIND`

src/js/components/TaskTable.js
philipnrmn: We could, but my fear is we'll just end up hiding the problem and never fixing it. IMO this shouldn't be done in the component during render at all...

philipnrmn:
In an offline conversation we raised concerns about duplicate processing being done here and in CompositeStore, I'm happy to confirm that this is not the case, and this PR makes nothing worse 👍

philipnrmn:
NB this PR and #946 are in conflict with one another. Whichever one lands first will cause significant rework on the other.

philipnrmn:
Needs significant rework thanks to #946

philipnrmn:
These tests should really not be passing 😱

src/js/components/TaskTable.js
mlunoe: I can appreciate that! Then let's get a TODO in there and create a ticket to solve the problem!

mlunoe:
Needs a rebase

philipnrmn:
rebased.

mlunoe:
Label remains. @jfurrow: We need your :eyes: too

src/js/components/TaskTable.js
```diff
@@ -39,9 +40,55 @@ class TaskTable extends React.Component {
   }
 
   getStatusValue(task) {
+    let taskHealth = this.getTaskHealth(task);
+    if (taskHealth === true) {
```
jfurrow: What is the purpose of checking specifically for the boolean `true` and `false` values? Would coercion not be sufficient here? (e.g. `if (taskHealth)` and `if (!taskHealth)`). Just curious :)

src/js/components/TaskTable.js
```diff
@@ -39,9 +40,55 @@ class TaskTable extends React.Component {
...
+  }
+
+  getTaskHealthFromMesos(task) {
+    if (task.statuses == null) {
+      return null;
+    }
+    const healths = task.statuses.map((status) => status.healthy);
+    const healthDataExists = healths.length > 0 && healths.every(
+      (health) => typeof health !== 'undefined'
+    );
```
jfurrow: We have been enforcing a rule where the fat arrow function syntax is only used when we need to retain scope in the anonymous function. I'm up for nixing this rule FWIW, but as far as I know it's still enforced.

jfurrow:
A few nits, but it seems to work perfectly!

src/js/components/TaskTable.js
mlunoe: I guess this makes it more precise to differentiate between e.g. `null` as opposed to boolean values.
I also don't like this function to read:
```js
if (taskHealth) {
  return 'Healthy';
}
if (taskHealth === false) {
  return 'Unhealthy';
}
...
```

Which to me would be more confusing than what is already there

philipnrmn: Yeah, it's a confusing API - we only show task health as a status if `taskHealth` is defined. If it's undefined, we use the `state` property to infer a status (like `Killing`). This is [covered in the tests](https://github.com/dcos/dcos-ui/pull/955/files#diff-f1869143cacab7fd5bff4b5dbe2094f9R145) but maybe we should add a comment here to ease readability.

src/js/components/TaskTable.js
philipnrmn: Thanks for the catch. +1 for killing the rule, I felt mildly resentful as I made this change.

philipnrmn:
Thanks for the review, fixed the nits.

src/js/components/TaskTable.js
jfurrow: Thanks for the explanation, guys!

PR #956, opened 2016-08-11
Make all files viewable in log viewer

mlunoe:
Select stdout as default:

Display all files as viewable (not directories, of course):


philipnrmn:
This is working and the code LGTM! I'm just a bit concerned that there could be an awful lot of files in the sandbox and they're not all really suited to being tailed in a webview, for example it's very common to add a potentially very large tarball to an app configuration:

mlunoe:
That is a good point, but we decided if they wanted to go and inspect that, that should be fine. It is hard to reason about which files are viewable and which are not, so we figured instead of trying to restrain the user, then let them do what they want.

philipnrmn:
SGTM

PR #958, opened 2016-08-11
Make sure to only emit directory events when response matches request

mlunoe:
Before:

After:


jfurrow:
Unrelated to this PR, but we should not remove the directory breadcrumbs when changing directories. https://mesosphere.atlassian.net/browse/DCOS-9280

daiyiwhale:
Nice!

PR #959, opened 2016-08-11
DCOS-8568: Making inactive filter items fade 50%

wavesoft:
This makes a bit more obvious the fact that a filter without items is disabled.

image


philipnrmn:
👍

PR #961, opened 2016-08-11
Improve general validation rules

orlandohohmeier:
Improve is number checks and use util to test is integer and is number in range input.

### DIFF
--- a/src/js/utils/ValidatorUtil.js
+++ b/src/js/utils/ValidatorUtil.js
@@ -34,9 +34,8 @@ var ValidatorUtil = {
   },
 
   isInteger: function (value) {
-    const number = parseFloat(value);
-
-    return Number.isInteger(number);
+    return ValidatorUtil.isNumber(value) &&
+      Number.isInteger(parseFloat(value));
   },
 
   isNumber: function (value) {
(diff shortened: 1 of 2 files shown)


mlunoe:
Great! 👍 💯

src/js/utils/ValidatorUtil.js
```diff
@@ -34,22 +34,22 @@ var ValidatorUtil = {
   },
 
   isInteger: function (value) {
-    const number = parseFloat(value);
-
-    return Number.isInteger(number);
+    return ValidatorUtil.isNumber(value) &&
+      Number.isInteger(parseFloat(value));
```
rcorral: Is doing `parseFloat` a good idea?
Number.isInteger(parseFloat('1.1')); // true
Number.isInteger('1.1'); // false
If the calling function doesn't end up doing `parseFloat` it may have unexpected results.
I would prefer to put the responsibility of calling `parseFloat` in the calling function and not changing the type at all here.

orlandohohmeier: We probably should rename these methods as they intentional use `parseFloat` to also handle number like  strings. How do you feel about `isNumericValue`, `isIntegerValue` , and `isNumericValueInRange`?

rcorral: +1

orlandohohmeier: Are you ok doing this in a different PR as I'd like to update the other validator utils names as well?

rcorral: Works for me, let's just try and take care of it soon, if possible!

orlandohohmeier: Sure, I'll create Jira so that we don't loose track.

PR #969, opened 2016-08-12
DCOS-6965: Replace instances of ProgressBar with StatusBar

daiyiwhale:
Before:

After:

This PR unifies the two bars (progress and status) into one. It is now called StatusBar, but contains logic from the progress bar.


mlunoe:
Conflicts here, please rebase

src/styles/components/statusbar.less
```diff
@@ -50,3 +50,11 @@
   }
 
 }
+
+.host-table {
+  .status-bar {
+    margin-right: 20px;
```
mlunoe: This makes the bars smaller as the number gets larger. Do we want that?
See the right alignment of the bars being different here:

src/js/components/StatusBar.js
```diff
@@ -92,9 +92,10 @@ class StatusBar extends React.Component {
         return null;
       }
 
-      let style = {
-        width: `${scale}%`
-      };
+      if (!style) {
+        style = {};
```
mlunoe: You can do `let {className, style = {}} = status;` to default assign in the declaration (unless style is actually being assigned `null`, but from what I can tell it is not).

MatApple:
ping @daiyiwhale

src/styles/components/statusbar.less
mlunoe: Fixed

src/js/components/StatusBar.js
mlunoe: Fixed

src/styles/index.less
```diff
@@ -153,14 +153,13 @@ Components
...
 @import "components/page-header.less";
-@import "components/progressbar.less";
 @import "components/scrollbar.less";
 @import "components/secret-value.less";
 @import "components/service-sidebar.less";
 @import "components/service-table.less";
 @import "components/side-list.less";
 @import "components/side-tabs.less";
-@import "components/statusbar.less";
+@import "components/status-bar.less";
```
jfurrow: Nice!

src/js/components/PackageUpgradeOverview.js
```diff
@@ -1,135 +0,0 @@
-import React from 'react';
```
MatApple: @jfurrow  - are we sure we want to kill this file? I know there is some upgrades related code that sneaked into master instead of the feature branch. I see that you wrote this and don't want to remove it if it's needed for upgrades

src/js/plugin-bridge/PluginModules.js
```diff
@@ -71,9 +71,8 @@ module.exports = {
     NestedServiceLinks: 'NestedServiceLinks',
     Page: 'Page',
     PageHeader: 'PageHeader',
-    ProgressBar: 'charts/ProgressBar',
     RequestErrorMsg: 'RequestErrorMsg',
-    RouterUtil: 'RouterUtil',
```
MatApple: How come we're removing `RouterUtil`?

mlunoe: This was listed under components (where it is not), but also under utils (as it should)

src/js/components/PackageUpgradeOverview.js
mlunoe: I checked with @jfurrow before removing these, but let's be completely sure!

jfurrow: Yeah I recommended removing it because the files are not being used in master and I don't think it makes sense to maintain them. If we do merge the software update feature branch in the future, we can pick up these loose ends as needed.

src/js/plugin-bridge/PluginModules.js
MatApple: Oh 👍

PR #970, opened 2016-08-15
Fix npm lint script

orlandohohmeier:
Adjust the npm lint script to disable globbing (`set -f`) as this should be handled by eslint. I also adjusted it to only lint src/js by default.


mlunoe:
⛵ 💯 ✨

mlunoe:

PR #972, opened 2016-08-15
Use keywords instead of numbers to configure eslint

orlandohohmeier:
Use keywords instead of numbers to configure eslint


.eslintrc
```diff
@@ -33,34 +33,34 @@
   "extends": "airbnb",
 
   "rules": {
-    "eqeqeq": [2, "smart"],
-    "quotes": [2, "single"],
-    "indent": [2, 2, {"SwitchCase": 1}],
+    "eqeqeq": ["error", "smart"],
+    "quotes": ["error", "single"],
+    "indent": ["error", 2, {"SwitchCase": 1}],
     "camelcase": [0, {"properties": "never"}],
```
mlunoe: Shouldn't we change the `0`s too?

orlandohohmeier: ✔️ Adjusted

mlunoe:
I merged this, since this actually didn't make a change of the rule set we have

PR #989, opened 2016-08-22
Improve task table performance

philipnrmn:
Avoids costly re-render of services tab when marathon store updates.

Before:
screen shot 2016-08-22 at 12 36 39 am

After:
screen shot 2016-08-22 at 12 37 18 am


wavesoft:
Cool!

PR #990, opened 2016-08-22
DCOS-9034: Renaming VIPs when changing service

wavesoft:
Adding support for renaming service VIPs when committing service changes.

_NOTE: That's a tech debt. It might be a better idea to add `diff` support on service edit_

### DIFF
--- a/src/js/components/modals/ServiceFormModal.js
+++ b/src/js/components/modals/ServiceFormModal.js
@@ -146,14 +146,6 @@ class ServiceFormModal extends mixin(StoreMixin) {
       this[method] = this[method].bind(this);
     });
 
-    this.sourceService = null;
-
-  }
-
-  componentDidMount() {
-    if (this.props.service) {
-      this.sourceService = JSON.parse(this.props.service.toJSON());
-    }
   }
 
   componentWillReceiveProps(nextProps) {
@@ -340,10 +332,9 @@ class ServiceFormModal extends mixin(StoreMixin) {
       );
 
       // Handle service rename if this action renamed a field
-      if (this.props.isEdit && this.sourceService) {
-        let sourceName = this.sourceService['id'].split('/').pop();
-        if (service.getName() !== sourceName) {
-          service.handleRename( sourceName, service.getName() );
+      if (this.props.isEdit && this.props.service && this.props.service.getName()) {
+        if (service.getName() !== this.props.service.getName()) {
+          service.handleRename( this.props.service.getName(), service.getName() );
         }
       }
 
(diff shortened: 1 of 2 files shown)


src/js/components/modals/ServiceFormModal.js
```diff
@@ -145,6 +145,15 @@ class ServiceFormModal extends mixin(StoreMixin) {
     METHODS_TO_BIND.forEach((method) => {
       this[method] = this[method].bind(this);
     });
+
+    this.sourceService = null;
```
orlandohohmeier: Let's not store any data on the component instance.

src/js/components/modals/ServiceFormModal.js
```diff
@@ -329,6 +338,15 @@ class ServiceFormModal extends mixin(StoreMixin) {
         this.props.isEdit,
         JSON.parse(this.state.service.toJSON())
       );
+
+      // Handle service rename if this action renamed a field
+      if (this.props.isEdit && this.sourceService) {
+        let sourceName = this.sourceService['id'].split('/').pop();
```
orlandohohmeier: Can't we simplify this to use `this.props.service.getName()`? Let's not replicate the `getName` functionality.

wavesoft: Blast, you are right! Fixed.

wavesoft:
As @Poltergeist pointed out, it looks that `marathon` is already doing this. Does it have to be done on the UI too?

wavesoft:
This still seems to be useful when using JSON-editing

src/js/components/modals/ServiceFormModal.js
```diff
@@ -307,6 +308,14 @@ class ServiceFormModal extends mixin(StoreMixin) {
     }
 
     if (this.state.jsonMode) {
+
+      // Handle service rename if this action renamed a field
+      if (this.props.isEdit && this.props.service && this.props.service.getId()) {
+        if (this.state.service.getId() !== this.props.service.getId()) {
+          this.state.service.handleRename( this.props.service.getId(), this.state.service.getId() );
```
mlunoe: 1. We are more than 80 characters here.
2. There are also unnecessary spaces in the function call.
3. Finally, this seems to build on a bigger problem of having multiple sources of truth. Not sure we can do much about it, though :/

src/js/structs/Service.js
```diff
@@ -249,6 +264,20 @@ module.exports = class Service extends Item {
 
   }
 
+  handleRename(fromName, toName) {
```
mlunoe: Do we actually want to be able to modify a service in place?
This is usually a one-way creation from a json of some sort

wavesoft:
Ok, apparently this seems not to be a particular issue, since the renaming of the VIP is already handled by the UI (when in form mode). Therefore we can assume that if somebody is using the JSON view he is an expert and it should be his business to re-name the VIP.

Under this rationale I think we could close and discard this commit.

PR #993, opened 2016-08-22
Increasing the durability of CosmosErrorMessage on invalid responses

wavesoft:
Make sure error responses are handled gracefully, even if they don't obey the standard.

### DIFF
--- a/src/js/components/CosmosErrorMessage.js
+++ b/src/js/components/CosmosErrorMessage.js
@@ -51,52 +51,48 @@ class CosmosErrorMessage extends React.Component {
       return null;
     }
 
-    // Check if this is a plain string
+    // Return early if we have some well-known or an unknown type
     if (typeof error === 'string') {
       return [error];
+    } else if (typeof error !== 'object') {
+      return null;
+    }
 
-    // Check if we should additionally append
-    // the error details as an unordered list
-    } else if (error.data && error.data.errors) {
+    // Return early if important fields are missing, or they are not
+    // in the expected format
+    if (!error.data || !error.data.errors) {
+      return null;
+    }
+    if (!Array.isArray(error.data.errors)) {
+      return [String(error.data.errors)];
+    }
 
-      // Check for invalid errors format
-      if (!Array.isArray(error.data.errors)) {
-        return [String(error.data.errors)];
+    // Get an array of array of errors for every individual path
+    let errorsDetails = error.data.errors.map(function (errorDetail) {
+
+      // Return early on unexpected error object format
+      if (!errorDetail) {
+        return [];
+      }
+      if (typeof errorDetail !== 'object') {
+        return [String(errorDetail)];
       }
 
-      // Get an array of array of errors for every individual path
-      let errorsDetails = error.data.errors.map(function (errorDetail) {
-
-        // Bail early on unexpected error object format
-        if (!errorDetail) {
-          return [];
-        }
-        if (typeof errorDetail !== 'object') {
-          return [String(errorDetail)];
-        }
-
-        // Extract details
-        let {path = '/', errors = []} = errorDetail;
-        if (!errors || !Array.isArray(errors)) {
-          return [];
-        }
-        return errors.map(function (error) {
-          return (ErrorPaths[path] || path)+'.'+error;
-        });
-
+      // Extract details
+      let {path = '/', errors = []} = errorDetail;
+      if (!errors || !Array.isArray(errors)) {
+        return [];
+      }
+      return errors.map(function (error) {
+        return (ErrorPaths[path] || path)+'.'+error;
       });
 
-      // Flatten elements in array and return
-      return errorsDetails.reduce(function (a, b) {
-        return a.concat(b);
-      });
+    });
 
-    } else {
-
-      // No errors
-      return null;
-
-    }
+    // Flatten elements in array and return
+    return errorsDetails.reduce(function (a, b) {
+      return a.concat(b);
+    });
 
   }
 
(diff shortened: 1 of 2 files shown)


src/js/components/CosmosErrorMessage.js
```diff
@@ -28,10 +28,14 @@ class CosmosErrorMessage extends React.Component {
   }
 
   getMessage() {
-    let {type, message} = this.props.error;
+    let {error} = this.props;
+    if (!error) {
+      return 'An error occurred (Description missing)';
```
jfurrow: Could we change this message to something that is more sentence-like, perhaps `An unknown error occurred.` or something similar?

jfurrow:
One minor suggestion, otherwise this looks great!

Poltergeist:
@wavesoft could you please rebase this branch?

src/js/components/CosmosErrorMessage.js
```diff
@@ -41,25 +46,58 @@ class CosmosErrorMessage extends React.Component {
...
-    });
 
-    // Flatten elements in array and return
-    return errorsDetails.reduce(function (a, b) {
-      return a.concat(b);
-    });
+    } else {
+
+      // No errors
+      return null;
```
mlunoe: Could we make edge cases return early? Makes it easier to read.

src/js/components/CosmosErrorMessage.js
```diff
@@ -41,25 +46,58 @@ class CosmosErrorMessage extends React.Component {
...
       return null;
     }
+
+    // Check if this is a plain string
+    if (typeof error === 'string') {
+      return [error];
+
     // Check if we should additionally append
     // the error details as an unordered list
+    } else if (error.data && error.data.errors) {
```
mlunoe: Not a fan of `else-if`s, they make it hard to reason about and it is unnecessary in this case, because you return in the block above. Could we do this instead:

```js
  getDetails() {
    let {error} = this.props;
    // No errors
    if (!error || !error.data || !error.data.errors) {
      return null;
    }

    // Check if this is a plain string
    if (typeof error === 'string') {
      return [error];

    }

    // Check for invalid errors format
    if (!Array.isArray(error.data.errors)) {
      return [String(error.data.errors)];
    }

    // Get an array of array of errors for every individual path
    let errorsDetails = error.data.errors.map(function (errorDetail) {

      // Bail early on unexpected error object format
      if (!errorDetail) {
        return [];
      }
      if (typeof errorDetail !== 'object') {
        return [String(errorDetail)];
      }

      // Extract details
      let {path = '/', errors = []} = errorDetail;
      if (!errors || !Array.isArray(errors)) {
        return [];
      }
      return errors.map(function (error) {
        return (ErrorPaths[path] || path)+'.'+error;
      });
    });

    // Flatten elements in array and return
    return errorsDetails.reduce(function (a, b) {
      return a.concat(b);
    });
  }
```

mlunoe:
I had a few stylistic suggestions. Works great!

src/js/components/CosmosErrorMessage.js
```diff
@@ -41,25 +46,54 @@ class CosmosErrorMessage extends React.Component {
...
-      // No errors
+    let {error} = this.props;
+    if (!error) {
+      return null;
+    }
+
+    // Return early if we have some well-known or an unknown type
+    if (typeof error === 'string') {
+      return [error];
+    } else if (typeof error !== 'object') {
```
mlunoe: This `else` is unnecessary, because of the `return` statement in block above

wavesoft: Could be a number, or boolean :)

mlunoe: Not sure what you are referring to, but this:
```js
if (typeof error === 'string') {
  return [error];
}
if (typeof error !== 'object') {
  return null;
}
```
is the same as this:
```js
if (typeof error === 'string') {
  return [error];
} else if (typeof error !== 'object') {
  return null;
}
```

If it is a string it will return in the first block, if it is not a string and not an object it will return in the second block. If it is none of these it will continue in both cases

wavesoft: Ah, I understood something completely different! - Fixed

src/js/components/CosmosErrorMessage.js
```diff
@@ -41,25 +46,55 @@ class CosmosErrorMessage extends React.Component {
...
         return (ErrorPaths[path] || path)+'.'+error;
       });
+
     });
 
     // Flatten elements in array and return
     return errorsDetails.reduce(function (a, b) {
       return a.concat(b);
     });
+
```
Poltergeist: This is too much :-) Please remove it.

src/js/components/modals/InstallPackageModal.js
```diff
@@ -66,6 +67,9 @@ class InstallPackageModal extends mixin(InternalStorageMixin, TabsMixin, StoreMi
     METHODS_TO_BIND.forEach((method) => {
       this[method] = this[method].bind(this);
     });
+
+    this.modalRef = undefined;
+
```
Poltergeist: Let's not add a empty new line add the end of a function.

Poltergeist:
Apart from two minor stylistic things. This look great. Thanks.

Poltergeist:
Reviewed 1 of 2 files at r1, 1 of 1 files at r4.
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dcos/dcos-ui/993)*

src/js/components/CosmosErrorMessage.js
mlunoe: Did this for you @wavesoft

src/js/components/modals/InstallPackageModal.js
mlunoe: Did this for you @wavesoft

PR #994, opened 2016-08-23
Remove network discovery text when discovery not checked

MatApple:
[DCOS-9119](https://mesosphere.atlassian.net/browse/DCOS-9119)

The expected behavior is that when a user checks the "Discovery" box that the blue icon and the information that accompanies it would display.

Currently, it is always being displayed which is not expected. This PR ensures the info is only displayed when a discovery checkbox is checked.


mlunoe:
Looks good!

PR #995, opened 2016-08-23
DCOS-9292: Prevent vip being overwritten by accident

Poltergeist:
This disables the form if a vip config is provided which does not contain the current service id.

image

### DIFF
--- a/src/js/components/modals/ServiceFormModal.js
+++ b/src/js/components/modals/ServiceFormModal.js
@@ -269,6 +269,35 @@ class ServiceFormModal extends mixin(StoreMixin) {
   shouldDisableForm(service) {
     let containerSettings = service.getContainerSettings();
 
+    let portDefinitions = service.getPortDefinitions();
+    let name = new RegExp(`${service.getId()}`, 'g');
+
+    if (portDefinitions) {
+      if (!portDefinitions.some(function (port) {
+        if (port.labels == null || Object.keys(port.labels).length === 0) {
+          return true;
+        }
+        return Object.keys(port.labels).some(function (key) {
+          return name.test(port.labels[key]);
+        });
+      })) {
+        return true;
+      }
+    }
+
+    if (containerSettings && containerSettings.docker && containerSettings.docker.portMappings) {
+      if (!containerSettings.docker.portMappings.some(function (port) {
+        if (port.labels == null || Object.keys(port.labels).length === 0) {
+          return true;
+        }
+        return Object.keys(port.labels).some(function (key) {
+          return name.test(port.labels[key]);
+        });
+      })) {
+        return true;
+      }
+    }
+
     return containerSettings != null && containerSettings.type === 'MESOS' && (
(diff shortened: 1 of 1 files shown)


orlandohohmeier:
Let's not remove the banner on change but when the attribute is compatible - any chance we can check this in the fly? We also need to properly need to reset the button styles.

src/js/components/modals/ServiceFormModal.js
```diff
@@ -269,6 +269,35 @@ class ServiceFormModal extends mixin(StoreMixin) {
...
+      }
+    }
+
+    if (containerSettings && containerSettings.docker && containerSettings.docker.portMappings) {
+      if (!containerSettings.docker.portMappings.some(function (port) {
+        if (port.labels == null || Object.keys(port.labels).length === 0) {
+          return true;
+        }
+        return Object.keys(port.labels).some(function (key) {
+          return name.test(port.labels[key]);
```
orlandohohmeier: Let's test for the complete string, as this would otherwise allow partial matches which is not correct.

Poltergeist:
@orlandohohmeier regarding the banner this is not part of this PR could we do this in another PR?

src/js/components/modals/ServiceFormModal.js
```diff
@@ -269,6 +269,35 @@ class ServiceFormModal extends mixin(StoreMixin) {
   shouldDisableForm(service) {
     let containerSettings = service.getContainerSettings();
 
+    let portDefinitions = service.getPortDefinitions();
+    let name = new RegExp(`${service.getId()}`, 'g');
+
+    if (portDefinitions) {
+      if (!portDefinitions.some(function (port) {
```
MatApple: Maybe we could pull this `some` statement out of the `if` and assign a variable to it. It's a little harder to follow the logic like this.

src/js/components/modals/ServiceFormModal.js
```diff
@@ -269,6 +269,35 @@ class ServiceFormModal extends mixin(StoreMixin) {
...
+        return Object.keys(port.labels).some(function (key) {
+          return name.test(port.labels[key]);
+        });
+      })) {
+        return true;
+      }
+    }
+
+    if (containerSettings && containerSettings.docker && containerSettings.docker.portMappings) {
+      if (!containerSettings.docker.portMappings.some(function (port) {
```
MatApple: Again here 👍

src/js/components/modals/ServiceFormModal.js
Poltergeist: Changed the regexp to `let name = new RegExp(`^${service.getId()}:[0-9]*$`, 'g');` 
But we have to check against `${service.getId()}:${port.port || port.containerPort || port.hostPort}` since the port can be in a lot of fields :-(.

MatApple:
LGTM - thanks @Poltergeist

orlandohohmeier:
LGTM 💯 ✨

PR #996, opened 2016-08-23
Updating mesosphere-shared-reactjs v0.0.20

mlunoe:
Only change in this version is this: https://github.com/dcos/mesosphere-shared-reactjs/pull/15, which makes sure that the hanging requests actually works.

This also fixes a problem, where the history server would always be set to offline in the beginning.

Before (with network delay):

Before (without network delay):

After:


src/js/events/MesosSummaryActions.js
```diff
@@ -64,7 +69,7 @@ function requestFromHistoryServer(resolve, reject, timeScale = 'last') {
       requestFromMesos(resolve, reject);
     },
     hangingRequestCallback() {
-      AppDispatcher.handleServerAction({type: REQUEST_MESOS_HISTORY_ONGOING});
+      AppDispatcher.handleServerAction({type: REQUEST_SUMMARY_HISTORY_ONGOING});
```
mlunoe: This action type was never defined and therefore never handled (or called before for that matter)

mlunoe:

wavesoft:
Tested in different network conditions and it LGTM. I can't reproduce the original problem though.

mlunoe:
If you open up `./node_modules/mesosphere-shared-reactjs/utils/RequestUtil.js` and replace the `createCallbackWrapper` with this:
```js
function createCallbackWrapper(callback, requestID) {
  return function () {
    setTimeout(() => {
      setRequestState(requestID, false);

      if (Util.isFunction(callback)) {
        callback.apply(null, arguments);
      }
    }, 2100);
  };
}
```

That will create a delay for some of the requests to trigger the hanging function for one and not for the other.

mlunoe:
Assigning you @rcorral since you have already looked at the code from mesosphere-shared-reactjs.

PR #998, opened 2016-08-23
Fix text overflow in UserDropup

jfurrow:
Before:

After:


mlunoe:
Looks good! Thanks for fixing!

PR #1002, opened 2016-08-24
Defaulting to string, where it is set to not change value type

mlunoe:
Was hitting a warning with this property being required, where it should not be

### DIFF
--- a/src/js/components/FilterInputText.js
+++ b/src/js/components/FilterInputText.js
@@ -35,7 +35,9 @@ class FilterInputText extends React.Component {
   handleChange(event) {
     let {target} = event;
 
-    this.props.handleFilterChange(target.value, ServiceFilterTypes.TEXT);
+    // Make sure to never emit falsy values
+    let value = target.value || '';
+    this.props.handleFilterChange(value, ServiceFilterTypes.TEXT);
   }
 
   handleInputClear() {
--- a/src/js/components/HealthTab.js
+++ b/src/js/components/HealthTab.js
@@ -36,7 +36,7 @@ class HealthTab extends React.Component {
     this.setState({healthFilter: selectedHealth.id});
   }
 
-  handleSearchStringChange(searchString) {
+  handleSearchStringChange(searchString = '') {
     this.setState({searchString});
   }
 
--- a/src/js/components/TaskView.js
+++ b/src/js/components/TaskView.js
@@ -100,7 +100,7 @@ class TaskView extends mixin(SaveStateMixin, StoreMixin) {
     this.setState({killAction: ''});
   }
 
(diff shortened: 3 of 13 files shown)


jfurrow:
We assign a default value of an empty string to `searchString`, so I am wondering how you were able to get the warning?

Furthermore, since we explicitly pass this property to controlled `input` element, so we should never allow the value to ever be `null` or `undefined`.

mlunoe:
The problem is exactly `null`, which will override the default assignment of empty string. What is the problem of having `null` or `undefined` as the value for an input field? We do that a lot of places

jfurrow:
Because then we will be switching between controlled inputs and uncontrolled inputs, and it will throw warnings in React 15.  https://facebook.github.io/react/docs/forms.html#controlled-components

But if we are doing it elsewhere then I guess we'll just fix them all at once when we upgrade.

mlunoe:
ok, that's fair. We should follow the recommendations for React.
I've changed users of the component to default to an empty string

mlunoe:
One thing though is that we probably should is to introduce a linting rule about not using mixed types for the same variable. Then we wouldn't hit these warnings again

philipnrmn:
Looks good!

Can we also avoid getting this warning in future implementations by checking `target.value` in the `handleChange` event of [FilterInputText](https://github.com/dcos/dcos-ui/blob/master/src/js/components/FilterInputText.js#L38)?

mlunoe:
@philipnrmn: Fixed!

philipnrmn:
Separately, this PR highlights a pre-existing performance issue - each of the affected components has many children and some form of tabular data. The data gets re-filtered, and the component and all its children get completely re-rendered, on each search string change. 

In the future, we should trap and handle circumstances where the user has cleared the input and handle them by not applying the filter, avoiding the costly filter.

Separately, the filter should not be called in `render`, which circumvents `shouldComponentUpdate`.

mlunoe:
I would actually argue to do the filtering in render, unless you have large amounts of data. I can't remember where I read it, but it comes from the "do more in render()" pattern. They even use it as an example here: https://facebook.github.io/react/docs/thinking-in-react.html#step-2-build-a-static-version-in-react

philipnrmn:
Kind of academic at this point, but it's a Friday afternoon so here's my $0.02:

Maybe the example isn't a perfect illustration of your point, but if the props/state data in that component changes, the view always needs re-rendering. In the `HealthTab` and similar, every keystroke in the search input triggers a re-render, and we don't know until we've performed the filter whether it's necessary or not. 

So I reckon the idiomatic way of doing it is to filter before setting state and then use checks in `shouldComponentUpdate` to prevent unnecessary renders.

PR #119, opened 2016-05-06
Optionally enforce types on List.

philipnrmn:
This PR adds an optional `type` property to `List`, in order to avoid having to enforce typing in each subclass of `List` (eg [here](https://github.com/dcos/dcos-ui/blob/master/src/js/structs/NodesList.js#L11-L18) in `NodeList`). 

If the `type` property is present, `List` will enforce typing on instantiation and when an item is added.

PR #1003, opened 2016-08-24
Add hooks for additional loading criteria to network page

MatApple:
Allows plugins to delay the rendering of the Network page by returning a Promise in a filter. Page will render when the Promise has resolved.

By default the page will not be delayed.

### DIFF
--- a/src/js/pages/NetworkPage.js
+++ b/src/js/pages/NetworkPage.js
@@ -43,6 +43,20 @@ class NetworkPage extends mixin(TabsMixin) {
 
   componentWillMount() {
     super.componentWillMount(...arguments);
+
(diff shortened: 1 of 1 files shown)


philipnrmn:
I like this and would cheerfully ship it!

I am a little apprehensive though. This PR introduces a small change to the way one specific page renders. I am reminded uncomfortably of #119, which I'm sure you'll remember introduced a degree of tension and discord quite disproportionate to its effect. Let's make sure that we consult as widely as possible so that we don't end up in that situation again.

MatApple:
Thanks @philipnrmn. I understand your concern and appreciate your understanding that we really only need this functionality on this page. If the need arises to have this elsewhere, let's reengineer and rollout globally.

PR #1004, opened 2016-08-24
Displaying hostname under host, and not endpoints

mlunoe:


jfurrow:
Looks good! I'm glad to see they chose not to display the endpoints list here!

PR #1006, opened 2016-08-25
DCOS-8081: Add Link to status documentation to the status column header

Poltergeist:
This adds a link to the marathon docs which explains the status column in more detail.

image


wavesoft:
LGTM

src/js/constants/Links.js
```diff
@@ -0,0 +1,4 @@
+export default {
+  statusHelpLink:'https://mesosphere.github.io/marathon/docs/marathon-ui' +
```
MatApple: Let's make use of the `marathonDocsURI` in `Config.js` to form the base URI.

Poltergeist: Done!

PR #1008, opened 2016-08-25
Affirm Universe Package config `title`s are utilised with tests and fixtures

daiyiwhale:
We are already using the `title` attribute in Universe Package configs, this PR is to affirm this to fixtures and tests.

### DIFF
--- a/src/js/utils/__tests__/SchemaUtil-test.js
+++ b/src/js/utils/__tests__/SchemaUtil-test.js
@@ -14,7 +14,8 @@ describe('SchemaUtil', function () {
                 id: {
                   type: 'string'
                 }
-              }
+              },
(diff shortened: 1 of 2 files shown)


src/js/utils/__tests__/SchemaUtil-test.js
```diff
@@ -14,7 +14,8 @@ describe('SchemaUtil', function () {
                 id: {
                   type: 'string'
                 }
-              }
+              },
+              title: "Application"
```
mlunoe: Needs to be single quote ;)

mlunoe:
Apart from my one small comment

daiyiwhale:
done!

PR #1009, opened 2016-08-25
Migrating Service UI bits to Internal Plugin :boom:

MatApple:
This PR moves all service-ui bits into an internal plugin called services-ui.

2 major things happened. 1 - moved files. 2 - updated import paths. All code is still tightly coupled at this stage through relative imports. **Reviewing this should be fast.**

Also introduced a folder called foundation-ui at the root. This is the starting point for us to place foundation files. A webpack alias in the `webpack.config.js` file allows us to pretend this folder is an npm module while we actively develop it.

Service upgrades files have not been moved yet (`ServicePlan.js`, etc). They will be moved later since they currently aren't functional.

Removed some dead files in the process which were not used anywhere (hence the absolute reduction).

Favicons and an `index.html` have been copied to `plugins/services` as a placeholder till they are updated to support running services as a standalone application.


mlunoe:
Looks good, works great

mlunoe:
👍

PR #1010, opened 2016-08-26
DCOS-9274: Fix volumes filter

Poltergeist:
This fixes the volume filter on the service table. It wasn't showing services with defined volumes.

volume-filter mov


orlandohohmeier:
LGTM! 🚢

wavesoft:
LGTM too

PR #1013, opened 2016-08-29
Decrease frequency of calls to forceUpdate

philipnrmn:
Please test using dcos/mesosphere-shared-reactjs#17

This is the first major chunk of work on improving the performance of dcos-ui by decreasing the number of times a component is rendered. Using the changes to the `StoreMixin` in the above PR, we specify whether or not a component should unmount after its first render, and whether a render should be forced by the mixin, or whether the component should handle its own updates via `state`. 

In [some cases](https://github.com/dcos/dcos-ui/blob/d2f674a4f40299ab6e189c5c918eab662f15457c/src/js/components/BreadcrumbSegment.js) it was possible to remove the store mixin altogether, and in [others](https://github.com/dcos/dcos-ui/commit/d2f674a4f40299ab6e189c5c918eab662f15457c) to remove specific store events which were unused. 

Where `suppressUpdate` has been added and is true, a component is able to render itself instead of being force-updated. Where `suppressUpdate` has been added and is false, the component still requires force-updating and should be refactored to avoid this (see especially the JobsPage and NetworkTab for instances of this).

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -4417,9 +4417,9 @@
       }
     },
     "mesosphere-shared-reactjs": {
-      "version": "0.0.20",
-      "from": "mesosphere-shared-reactjs@0.0.20",
-      "resolved": "https://registry.npmjs.org/mesosphere-shared-reactjs/-/mesosphere-shared-reactjs-0.0.20.tgz"
+      "version": "0.0.21",
+      "from": "mesosphere-shared-reactjs@0.0.21",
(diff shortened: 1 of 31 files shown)


jfurrow:
@philipnrmn I'm not sure if this is related to this PR, but I'm receiving the following error when trying to load any task's detail page (specifically accessing it from within the Services UI, for what it's worth):

mlunoe:
Rebased this for you @philipnrmn

mlunoe:
I get the same error

orlandohohmeier:
@philipnrmn Could you please rebase your branch?

mlunoe:
Looks good! Waiting for this to go in: dcos/mesosphere-shared-reactjs#17

PR #1014, opened 2016-08-29
[Pods] Moving Application logic to Service

wavesoft:
- This PR moves all the logic from the `Service` struct to `Application`, making it easier to introduce the `Pod` struct that will subclass from `Service`.
- `Service` struct has now a skeleton of the bare minimum functions required for a service to appear in the services list (and for the filters to operate correctly). Pods will implement the same fields accordingly.

_This is the first of a series of PRs regarding the Pods implementation and they are aimed for discussion on the `pods` feature branch._


src/js/components/__tests__/ServicesTable-test.js
```diff
@@ -8,11 +8,11 @@ const React = require('react');
 /* eslint-enable no-unused-vars */
 const ReactDOM = require('react-dom');
 const ServicesTable = require('../ServicesTable');
-const Service = require('../../structs/Service');
+const Application = require('../../structs/Application');
```
orlandohohmeier: This one should be a Service as it's dealing with pods as well

src/js/pages/__tests__/DeploymentsTab-test.js
```diff
@@ -16,7 +16,7 @@ const DCOSStore = require('../../stores/DCOSStore');
 const Deployment = require('../../structs/Deployment');
 const DeploymentsTab = require('../services/DeploymentsTab');
 const DeploymentsList = require('../../structs/DeploymentsList');
-const Service = require('../../structs/Service');
+const Application = require('../../structs/Application');
```
orlandohohmeier: Same here.

src/js/structs/__tests__/Application-test.js
```diff
@@ -0,0 +1,861 @@
+const Application = require('../Application');
+const HealthStatus = require('../../constants/HealthStatus');
+const ServiceImages = require('../../constants/ServiceImages');
+const ServiceStatus = require('../../constants/ServiceStatus');
+const TaskStats = require('../TaskStats');
+const VolumeList = require('../VolumeList');
+
+describe('Service', function () {
```
orlandohohmeier: Please rename the the test suite as well

src/js/components/__tests__/ServicesTable-test.js
wavesoft: Not true, because if I instantiate a new `Service`, all the health checks will always return dummy data, since `Service` is an abstract class:

```js
  getTasksSummary() {
    return {
      tasksHealthy: 0,
      tasksStaged: 0,
      tasksUnhealthy: 0,
      tasksUnknown: 0,
      tasksOverCapacity: 0,
      tasksRunning: 0
    };
  }
```

I add a test for a `Pod` instance when the struct is ready.

src/js/pages/__tests__/DeploymentsTab-test.js
wavesoft: The same should apply here. I will add it as a pending task to use also Pod instances here.

src/js/structs/__tests__/Application-test.js
wavesoft: Done 😄

src/js/components/__tests__/ServicesTable-test.js
orlandohohmeier: True, did't thought about that

orlandohohmeier:
LGTM 💯 ✨

orlandohohmeier:
Thanks for adjusting the method name! 👍

PR #1007, opened 2016-08-25
[Loader 1 of 2] Adding fetch previous log loader to mesos log

mlunoe:

PR #1016, opened 2016-08-29
[Loader 2 of 2] Add loader component

mlunoe:
---
⚠️ Followup PR to this: https://github.com/dcos/dcos-ui/pull/1007

---

This PR introduces a Loader component to be used around the application, centralizing code in one place. It is also exposing the component as a module to be used by plugins!

### DIFF
--- a/src/js/components/Loader.js
+++ b/src/js/components/Loader.js
@@ -9,35 +9,37 @@ import React from 'react';
 // Available types:
 let typeMap = {
   ballBeat: {className: 'ball-beat', divCount: 3},
-  ballClipRotate: {className: 'ball-clip-rotate', divCount: 1},
-  ballClipRotateMultiple:
-    {className: 'ball-clip-rotate-multiple', divCount: 2},
-  ballClipRotatePulse: {className: 'ball-clip-rotate-pulse', divCount: 2},
-  ballGridBeat: {className: 'ball-grid-beat', divCount: 9},
-  ballGridPulse: {className: 'ball-grid-pulse', divCount: 9},
-  ballPulseRise: {className: 'ball-pulse-rise', divCount: 5},
-  ballPulseSync: {className: 'ball-pulse-sync', divCount: 3},
-  ballRotate: {className: 'ball-rotate', divCount: 1},
   ballScale: {className: 'ball-scale', divCount: 1},
-  ballScaleMultiple: {className: 'ball-scale-multiple', divCount: 3},
-  ballScaleRandom: {className: 'ball-scale-random', divCount: 3},
-  ballScaleRipple: {className: 'ball-scale-ripple', divCount: 1},
-  ballScaleRippleMultiple:
-    {className: 'ball-scale-ripple-multiple', divCount: 3},
-  ballSpinFadeLoader: {className: 'ball-spin-fade-loader', divCount: 8},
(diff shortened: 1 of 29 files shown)


src/js/components/Loader.js
```diff
@@ -0,0 +1,86 @@
...
+  cubeTransition: {className: 'cube-transition', divCount: 2},
+  lineScale: {className: 'line-scale', divCount: 5},
+  lineScaleParty: {className: 'line-scale-party', divCount: 4},
+  lineScalePulseOut: {className: 'line-scale-pulse-out', divCount: 5},
+  lineScalePulseOutRapid:
+    {className: 'line-scale-pulse-out-rapid', divCount: 5},
+  lineSpinFadeLoader: {className: 'line-spin-fade-loader', divCount: 8},
+  pacman: {className: 'pacman', divCount: 5},
+  semiCircleSpin: {className: 'semi-circle-spin', divCount: 1},
+  triangleSkewSpin: {className: 'triangle-skew-spin', divCount: 1}
```
jfurrow: Do we need all of these? It would be nice to reduce these to just the few that we actually use.

mlunoe: I just thought it would be nice to have the possibilities as a reference here. What bothers you about it?

jfurrow: It seems superfluous because we will likely never use most of these; it's extra code with no benefit.

mlunoe: Right, however a good reference. We could link to the repository instead? Or comment out the ones we don't use?
It was just annoying to me to have to look this up and understand it again. That will not be the case next time we want to add a loader.

jfurrow: That's understandable. I guess I'd prefer to remove them and link to the repository of available options, but I don't care that much — if you'd prefer to comment them out, let's do that!

mlunoe: Done!

src/js/pages/jobs/JobDetailPage.js
```diff
@@ -260,13 +261,8 @@ class JobDetailPage extends mixin(StoreMixin, TabsMixin) {
 
   getLoadingScreen() {
     return (
-      <div className="container container-fluid container-pod text-align-center
```
wavesoft: #nit but perhaps use the ```{`string`}``` approach since it spans on multiple lines?

wavesoft:
Health is never populated on the nodes grid. It always stays on loading state:

image

mlunoe:
Rebased, and [this](https://github.com/dcos/dcos-ui/pull/1016#issuecomment-243709207) problem is fixed.

src/js/pages/jobs/JobDetailPage.js
mlunoe: You are commenting on lines I removed

wavesoft:
LGTM

mlunoe:
ping @daiyiwhale

src/js/components/Loader.js
```diff
@@ -0,0 +1,88 @@
...
+
+class Loader extends React.Component {
+  getDivs(length) {
+    return Array.from({length}).map(function (_, index) {
+      return <div key={index} />;
+    });
+  }
+
+  render() {
+    let {className, innerClassName, type} = this.props;
```
rcorral: You don't define `type` in the proptypes.

PR #1017, opened 2016-08-29
DCOS-9440: Fix component health download snapshot link

daiyiwhale:
Fixes link! Clicking the download button actually initiates a download now.

note: DCOS-9440 is the internal tracking number.
dcos ref: https://dcosjira.atlassian.net/browse/DCOS-335


mlunoe:
Tested this out, and it works well!

PR #1018, opened 2016-08-29
Add warning for auth0 config mismatch

MatApple:
Adding security warning so developer knows to update authHost config when using different auth host domain. Previously silent, early abort from login flow led to confusion.

### DIFF
--- a/plugins/oauth/components/LoginPage.js
+++ b/plugins/oauth/components/LoginPage.js
@@ -48,8 +48,7 @@ class LoginPage extends mixin(StoreMixin) {
 
   onMessageReceived(event) {
     if (event.origin !== SDK.config.authHost) {
-      console.warn(`Event Origin "${event.origin}" does not match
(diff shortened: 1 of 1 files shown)


plugins/oauth/components/LoginPage.js
```diff
@@ -48,6 +48,9 @@ class LoginPage extends mixin(StoreMixin) {
 
   onMessageReceived(event) {
     if (event.origin !== SDK.config.authHost) {
+      console.warn(`Event Origin "${event.origin}" does not match
+        allowed origin "${SDK.config.authHost}"`);
```
jfurrow: The extra whitespace in this string will be visible in the console, so let's just put it all on a single line.

MatApple: Yeah I don't know why I did that. Thanks @jfurrow!

mlunoe:
LGTM

jfurrow:
:ship:!

PR #1022, opened 2016-08-30
[Pods] Making sure toJSON is properly used throughout the structs

wavesoft:
Making sure `toJSON` on structs returns an object and moving the JSON
clean-up functions to the them.

### DIFF
--- a/src/js/structs/Item.js
+++ b/src/js/structs/Item.js
@@ -16,6 +16,6 @@ module.exports = class Item {
   }
 
   toJSON() {
-    return this._itemData;
+    return this.get();
   }
 };
--- a/src/js/structs/Item.js
+++ b/src/js/structs/Item.js
@@ -14,4 +14,8 @@ module.exports = class Item {
 
     return this._itemData[key];
   }
+
+  toJSON() {
+    return this._itemData;
+  }
 };
--- a/src/js/structs/Job.js
+++ b/src/js/structs/Job.js
@@ -1,3 +1,4 @@
+import {cleanJobJSON} from '../utils/CleanJSONUtil';
 import DateUtil from '../utils/DateUtil';
 import Item from './Item';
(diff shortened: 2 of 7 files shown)


src/js/structs/Item.js
```diff
@@ -14,4 +14,8 @@ module.exports = class Item {
 
     return this._itemData[key];
   }
+
+  toJSON() {
+    return this._itemData;
```
Poltergeist: Please use `this.get()` instead of using `_itemData` directly.

wavesoft: Well, we are on the `Item` class, and we are aware of the private properties of the class. In this particular case, I think `this.get()` will only add an extra useless call?

Poltergeist: I agree, but there is another understanding: https://github.com/dcos/dcos-ui/pull/903#discussion_r73374936 The idea is to use the api instead of accessing the internal values.

src/js/structs/Job.js
```diff
@@ -127,6 +128,6 @@ module.exports = class Job extends Item {
   }
 
   toJSON() {
-    return this.get();
+    return cleanJobJSON(this.get());
```
orlandohohmeier: Side note: We're going to remove the clean methods once we've  separated the status from definition/spec data.

Poltergeist: ???

Poltergeist:
Please rebase this branch and then feel free to merge this branch.

src/js/structs/Job.js
orlandohohmeier: We're going to refactor the service structs so that we have a `ServiceSpec` which will only include the "definition" and eventually `ServiceStatus` including all the status informations (the data removed by the clean methods). Thus we don't need those method any more.

PR #1028, opened 2016-08-30
DCOS-8360: deeper nesting styles of schema forms

daiyiwhale:
We are styling the nested schema form configs in the following way:

```
Tier 1 (which is currently the nav and the page header) should be an H2
Tier 2 an H3
Tier 3 an H4
Tier 4 an H5
```
Before:

After:


mlunoe:
Looks good to me! Since you changed some of the integration test fixtures, could you run them and post the result here?

mlunoe:
Tests still failing:

PR #1032, opened 2016-08-31
DCOS-9248: Remove overcapacity from bar

Poltergeist:
This PR does remove over capacity from the statusbar because the number was not clear to the user. There is now a text added next to the total task count if a service is over capacity.

Also the over capacity line is still in the tooltip.

over capacity:
image

within capacity:
image

/cc @leemunroe @silverem2

### DIFF
--- a/src/js/components/ServiceInfo.js
+++ b/src/js/components/ServiceInfo.js
@@ -93,6 +93,13 @@ class ServiceInfo extends React.Component {
     let runningTasksCount = tasksSummary.tasksRunning;
     let instancesCount = service.getInstancesCount();
     let runningTasksSubHeader = StringUtil.pluralize('Task', runningTasksCount);
+    let overCapacity = '';
+
+    if (tasksSummary.tasksOverCapacity !== 0) {
+      overCapacity =
+          ` (over capacity by ${tasksSummary.tasksOverCapacity} tasks)`;
+    }
+
     let subHeaderItems = [
       {
         classes: `media-object-item ${serviceStatusClassSet}`,
@@ -101,7 +108,7 @@ class ServiceInfo extends React.Component {
       },
       {
         classes: 'media-object-item',
-        label: `${runningTasksCount} ${runningTasksSubHeader}`,
+        label: `${runningTasksCount} ${runningTasksSubHeader}` + overCapacity,
         shouldShow: runningTasksCount != null && runningTasksSubHeader != null
       },
       {
--- a/src/js/components/ServiceInfo.js
+++ b/src/js/components/ServiceInfo.js
@@ -94,10 +94,18 @@ class ServiceInfo extends React.Component {
     let instancesCount = service.getInstancesCount();
     let runningTasksSubHeader = StringUtil.pluralize('Task', runningTasksCount);
     let overCapacity = '';
+    let disableToolTip = serviceStatus === 'Deploying';
 
     if (tasksSummary.tasksOverCapacity !== 0) {
       overCapacity =
           ` (over capacity by ${tasksSummary.tasksOverCapacity} tasks)`;
+    }
+
+    if (disableToolTip) {
+      tasksSummary = {
+        tasksStaged: instancesCount,
+        tasksRunning: 0
+      };
     }
 
     let subHeaderItems = [
@@ -114,8 +122,9 @@ class ServiceInfo extends React.Component {
       {
         label: (
           <HealthBar
+            disableToolTip={disableToolTip}
             tasksSummary={tasksSummary}
-            instancesCount={instancesCount} />
+            instancesCount={instancesCount}/>
         ),
         shouldShow: true
       }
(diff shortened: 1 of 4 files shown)


orlandohohmeier:
Test failed:
```
��� ServiceInfo ��� #render ��� it renders number of running tasks
  - Expected '2 Tasks (over capacity by NaN tasks)' to equal '2 Tasks'.
```

Poltergeist:
o.O Fixed the test.

leemunroe:
@Poltergeist Making sure I understand what is going on here. 

* In the over capacity screenshot, nginx is `Deploying`
* The ideal state of nginx is 3 tasks
* There are currently `6 Tasks` because there are 3 running healthy tasks and 3 tasks are deploying (therefore 3 over capacity)

Questions

* Does it make sense to have `(100%)` beside each in the tooltip? This seems odd to me. I think I would expect `3 Healthy of 6 (50%)` `3 Unknown of 6 (50%)` `3 Over Capacity (100%)`
* The bar doesn't look like it is deploying (i.e. doesn't have the striped barbershop style) which is concerning. Is that because we don't know which tasks are deploying?

Poltergeist:
The labels are divided into multiple parts,
- the percent was calculated using the `instances` setting not the `tasksRunning`. I use now what ever is higher either `tasksRunning` or `instances`. I used this because `tasksStaging` do not get added to the `tasksRunning`, alternative here would be to sum `tasksStaging` and `tasksRunning`.
- However I do have a problem with is the `3 Over Capacity (100%)` This would mean that this percentage is based on the instances defined and not the number of running tasks. I would rather think that 3 tasks are over capacity and those are 50% of all the tasks running then 100% over capacity.

About the bar, in this case I defined an application with a `nginx` container wit a healthcheck with an interval of 60 seconds, so the first 60 seconds the tasks is not healthy and the service is still deploying. So If I change the configuration to another `nginx` container like `nginx:latest`, a new task will be staged, which will be eventually be running, but first it will be `unkown` in health. After 60 seconds the health check will pass. At that moment the old running instances will be shutting down and after that the complete deployment will be completed and the service will be in the running state. 

In the screenshot you do see a service which still is deploying but the new tasks are already running. We are adding the barbershop stripes only to staging instances, should we add the stripes to all states if a service is deploying? So also to the `healthy`, `unhealthy`, `unknown`  etc.

leemunroe:
I think I would expect the deploying status to trump the health status i.e. if it is in a deploying state, always show the stripes, until the deployment has finished, then show the health. I think visually it is important to call this out to the user that it is deploying. Does that make sense?

Also sounds like something that needs a deeper look at in a future release, and whether we should separate out health/status and/or replace the bar with something else.

Poltergeist:
Okay I'll replace all the health informations with a deploying animation if a service is deploying. But this will mean that the user has no health information during a deployment even if this deployment will take longer. Do we also want to show something special for waiting services like a yellow barber stripe bar?

Poltergeist:
image
image

Poltergeist:

orlandohohmeier:
LGTM! 🚢

src/js/components/ServiceInfo.js
```diff
@@ -93,6 +93,21 @@ class ServiceInfo extends React.Component {
     let runningTasksCount = tasksSummary.tasksRunning;
     let instancesCount = service.getInstancesCount();
     let runningTasksSubHeader = StringUtil.pluralize('Task', runningTasksCount);
+    let overCapacity = '';
+    let disableToolTip = serviceStatus === 'Deploying';
```
orlandohohmeier: Let's use the const #nit

orlandohohmeier: Let's rename this var

orlandohohmeier:
Please reenable tooltip during deployment.

leemunroe:
@orlandohohmeier @Poltergeist agreed, we should have the tooltip in there, then at least we have parity (I think) with how Marathon UI works today i.e. if a deployment is happening, the whole bar is striped, and the tool tip makes it clear what is going on.

What I would like to see, and should have clarified above, is if there are 3 running healthy tasks, and 3 deploying (therefore 3 over capacity), then half the bar is striped. It sounds like we can't do this right now because we don't know which tasks are deploying vs deployed, correct? I would suggest we chat to backend to see if there's a way to make this distinction /cc @gkleiman @aquamatthias 


image

wavesoft:
LGTM!

src/js/components/ServiceInfo.js
```diff
@@ -93,6 +93,14 @@ class ServiceInfo extends React.Component {
     let runningTasksCount = tasksSummary.tasksRunning;
     let instancesCount = service.getInstancesCount();
     let runningTasksSubHeader = StringUtil.pluralize('Task', runningTasksCount);
+    let overCapacity = '';
+    let isDeploying = serviceStatus === 'Deploying';
+
+    if (tasksSummary.tasksOverCapacity !== 0) {
```
orlandohohmeier: Let's change this condition to `tasksOverCapacity > 0`, to not accidentally render the message if  `tasksOverCapacity` is undefined or null.

Poltergeist: Changed the check to `tasksOverCapacity > 0`

orlandohohmeier:
@leemunroe Could you please create a JIRA to track the adjustments needed to properly display overcapacity tasks?

orlandohohmeier:
LGTM 💯 ✨

PR #1033, opened 2016-08-31
[Pods] Handling new pod structures in the marathon groups response

wavesoft:
---
⚠️  ~~Depends on #1014~~

---

Also adding some fixtures for testing until we have real API.

### DIFF
--- a/src/js/utils/MarathonUtil.js
+++ b/src/js/utils/MarathonUtil.js
@@ -72,7 +72,7 @@ function parseApp(app) {
 
 function parsePod(pod) {
   let {id} = pod;
-  Object.defineProperty(app, '_type', {
+  Object.defineProperty(pod, '_type', {
     enumerable: false,
     configurable: false,
     writable: false,
--- a/src/js/utils/MarathonUtil.js
+++ b/src/js/utils/MarathonUtil.js
@@ -3,12 +3,6 @@ import VolumeStatus from '../constants/VolumeStatus';
 
 function parseApp(app) {
   let {id, tasks = [], container = {}} = app;
-  Object.defineProperty(app, '_type', {
-    enumerable: false,
-    configurable: false,
-    writable: false,
-    value: 'app'
-  });
 
   if (id == null || !id.startsWith('/') || id.endsWith('/')) {
     throw new Error(`Id (${id}) must start with a leading slash ("/") ` +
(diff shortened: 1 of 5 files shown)


wavesoft:
Dependant PR merged. Rebased and looking for review!

src/js/utils/MarathonUtil.js
```diff
@@ -3,6 +3,12 @@ import VolumeStatus from '../constants/VolumeStatus';
 
 function parseApp(app) {
   let {id, tasks = [], container = {}} = app;
+  Object.defineProperty(app, '_type', {
+    enumerable: false,
+    configurable: false,
+    writable: false,
+    value: 'app'
```
mlunoe: If we do the check here, we might as well do the Framework and Application check as well here, moving these lines to here: https://github.com/dcos/dcos-ui/pull/1033/files#diff-a44f44c46bf6bdbc44737179ffb8970dR77

This way we don't differentiate in different places about different types

wavesoft:
Decided to completely remove the tests from the location I had them and move the entire logic in a single place, in the `ServiceUtil` (because this function is also used by the `ServiceFormModal` to create an appropriate instance from JSON.

mlunoe:
Alright, lgtm

src/js/utils/ServiceUtil.js
```diff
@@ -63,8 +64,18 @@ const pruneHealthCheckAttributes = function (healthCheckSchema, healthCheck) {
   }, {});
 };
 
+function looksLikePod(data) {
+  // A pod definition contains a `containers` field, not found on the
+  // other definitions.
+  return Array.isArray(data.containers);
```
orlandohohmeier: Let's move this check to `createServiceFromDefinition` method, as it's only used once.

wavesoft:
`createServiceFromDefinition` is used both when receiving data from the server **AND** on the `ServiceForm`. However, the form is working on a different part of the data (only on the specs), while the server response contains everything.

Solution? Shall we separate these two functions, or is this `Service` --> `ServiceSpec` / `ServiceRuntime` is going to introduce more trouble?

wavesoft:
Will work on a solution on the #1037 PR were we have the notion of Specifications

Poltergeist:
Looks awesome good imo. 💯 🎆 1/2 🚢 🇮🇹  🍏

PR #1039, opened 2016-09-01
Properly checking if error/warning messages have changed

wavesoft:
This makes sure `ServiceFormModal` properly checks if the component should update.

This fixes a bug encountered when the form is in JSON edit mode. Previously, when a warning message is visible (ex. the JSON string contains unhanded attributes), the components updates on every character type, making Ace editor replace it's contents on every key press - effectively prohibiting further edit of it's contents.

### DIFF
--- a/src/js/components/modals/ServiceFormModal.js
+++ b/src/js/components/modals/ServiceFormModal.js
@@ -1,5 +1,6 @@
 import Ace from 'react-ace';
 import classNames from 'classnames';
+import deepEqual from 'deep-equal';
 import {Hooks} from 'PluginSDK';
 import mixin from 'reactjs-mixin';
 import {Modal, Tooltip} from 'reactjs-components';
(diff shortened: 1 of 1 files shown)


src/js/components/modals/ServiceFormModal.js
```diff
@@ -146,10 +147,25 @@ class ServiceFormModal extends mixin(StoreMixin) {
 
   shouldComponentUpdate(nextProps, nextState) {
     let {state, props} = this;
+
+    if ((typeof state.errorMessage !== typeof nextState.errorMessage) ||
```
mlunoe: We could abstract these into a one function

mlunoe:
One #nit, but looks good 😊

src/js/components/modals/ServiceFormModal.js
```diff
@@ -146,10 +147,25 @@ class ServiceFormModal extends mixin(StoreMixin) {
...
+       ((typeof state.errorMessage === 'object') && !deepEqual(
+          state.errorMessage,
+          nextState.errorMessage)
+       )) {
+      return true;
+    }
+
+    if ((typeof state.warningMessage !== typeof nextState.warningMessage) ||
```
orlandohohmeier: This condition looks like a duplicate...

src/js/components/modals/ServiceFormModal.js
wavesoft: Done 👍

orlandohohmeier:
LGTM 💯 ✨

PR #1041, opened 2016-09-02
DCOS-9104: Add Tasks started by a framework to the taskList

Poltergeist:
image
image

With this PR we are adding TASKS_RUNNING for frameworks from the mesosSummary to a running Framework.

### DIFF
--- a/src/js/structs/Framework.js
+++ b/src/js/structs/Framework.js
@@ -31,6 +31,11 @@ module.exports = class Framework extends Service {
     return tasksSummary;
   }
 
+  getInstancesCount() {
+    let tasksRunning = this.get('TASK_RUNNING') || 0;
+    return super.getInstancesCount() + tasksRunning;
+  }
+
   getName() {
     let labels = this.getLabels();
     if (labels && labels.DCOS_PACKAGE_FRAMEWORK_NAME) {
--- a/src/js/structs/Framework.js
+++ b/src/js/structs/Framework.js
@@ -21,6 +21,16 @@ module.exports = class Framework extends Service {
     return {value};
   }
 
+  getTasksSummary() {
+    let tasksSummary = super.getTasksSummary();
+
+    let tasksRunning = this.get('TASK_RUNNING') || 0;
+    tasksSummary.tasksRunning += tasksRunning;
+    tasksSummary.tasksUnknown += tasksRunning;
+
+    return tasksSummary;
+  }
+
   getName() {
     let labels = this.getLabels();
     if (labels && labels.DCOS_PACKAGE_FRAMEWORK_NAME) {
--- a/src/js/structs/Framework.js
+++ b/src/js/structs/Framework.js
@@ -22,7 +22,7 @@ module.exports = class Framework extends Service {
   }
 
   getTasksSummary() {
-    let tasksSummary = super.getTasksSummary();
+    let tasksSummary = Object.assign({}, super.getTasksSummary());
 
     let tasksRunning = this.get('TASK_RUNNING') || 0;
     tasksSummary.tasksRunning += tasksRunning;
--- a/src/js/schemas/service-schema/General.js
+++ b/src/js/schemas/service-schema/General.js
@@ -96,7 +96,13 @@ let General = {
           type: 'number',
           default: 1,
           getter(service) {
-            return `${service.getInstancesCount() || this.default}`;
+            let taskRunning = service.get('TASK_RUNNING') || 0;
+            let instances = service.getInstancesCount();
+            instances -= taskRunning;
+            if ((instances !== 0 && !instances) || instances < 0) {
+              instances = this.default;
+            }
+            return `${instances}`;
           }
         }
       }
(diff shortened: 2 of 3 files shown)


src/js/stores/MesosSummaryStore.js
```diff
@@ -178,6 +178,42 @@ class MesosSummaryStore extends GetSetBaseStore {
     });
   }
 
+  getFrameworkTasksByName(serviceName = '') {
+    let frameworks = this.get('frameworks');
+
+    if (frameworks == null) {
+      return false;
```
mlunoe: I don't think we want to return different types. Let's stick with `null`

src/js/stores/MesosSummaryStore.js
```diff
@@ -178,6 +178,42 @@ class MesosSummaryStore extends GetSetBaseStore {
     });
   }
 
+  getFrameworkTasksByName(serviceName = '') {
```
mlunoe: I don't think we need to default assign this to empty string

src/js/stores/MesosSummaryStore.js
```diff
@@ -178,6 +178,42 @@ class MesosSummaryStore extends GetSetBaseStore {
...
+    let frameworks = this.get('frameworks');
+
+    if (frameworks == null) {
+      return false;
+    }
+
+    return frameworks.filter((framework) => {
+      return framework.name === serviceName;
+    }).reduce((memo, framework) =>{
```
mlunoe: This seems like a weird process to me. Aren't we ever going to find at most one?
And if we find multiple we will override it with the one found last? Also, if there are multiple why not start with `{TASK_ERROR: 0,...}` as the initial value?

Assuming we will only find one, why not do something like this:
```js
getFrameworkTasksByName(serviceName) {
  let frameworks = this.get('frameworks') || [];
  let framework = frameworks.find(function (framework) {
    return framework.name === serviceName;
  });
  if (!framework) {
    return null;
  }

  return Object.assign({TASK_ERROR: 0, ...}, framework);
}
```

src/js/stores/MesosSummaryStore.js
mlunoe: Do we maybe want to move this to a Util? We have something _very_ similar in `MesosStateStore`

mlunoe: And could we call it something with `count` in it to make it obvious what is to be expected returned?

src/js/components/ServicesTable.js
```diff
@@ -294,6 +295,11 @@ var ServicesTable = React.createClass({
     let serviceStatus = service.getStatus();
     let serviceStatusClassSet = StatusMapping[serviceStatus] || '';
     let taskSummary = service.getTasksSummary();
+    let {TASK_RUNNING = 0} =
+        MesosSummaryStore.getFrameworkTasksByName(service.getName());
```
mlunoe: Having this added to the count here seems to be the wrong approach.

Could we get ServiceList from CompositeState (which merges information from Marathon and Mesos) and get the count through there?

orlandohohmeier: Let's not use the `CompositeState`, as we've decided to deprecate it with the introduction of the `DCOSStore`.  How about merging  the data within the `DCOSStrore`?

src/js/stores/MesosSummaryStore.js
orlandohohmeier: How abut `getFrameworkTaskSummaryByName` to match the service naming scheme?

Poltergeist:
Okay I found out that we are already merging the data in the DCOSStore and we had the relevant information all a long. This is why I decided that I will drop the commit and create new ones which add the relevant information (`TASK_RUNNING`) directly to the `tasksSummary` and `InstancesCount`. Regarding the comments, this is a complete rewrite so the code does not exist anymore sorry, but thanks for the comments, they where insightful. Regarding the `CompositeState` didn't we agree to not use it anymore?

src/js/structs/Framework.js
```diff
@@ -21,6 +21,21 @@ module.exports = class Framework extends Service {
     return {value};
   }
 
+  getTasksSummary() {
+    let tasksSummary = super.getTasksSummary();
+
+    let tasksRunning = this.get('TASK_RUNNING') || 0;
```
orlandohohmeier: What about the other tasks states?

src/js/structs/Framework.js
```diff
@@ -21,6 +21,21 @@ module.exports = class Framework extends Service {
     return {value};
   }
 
+  getTasksSummary() {
+    let tasksSummary = super.getTasksSummary();
```
orlandohohmeier: Let's copy (`Object.assign({}, super.getTasksSummary());`) the properties as we otherwise mutate the original tasks summary.

Poltergeist: I added this. in the commit => 378458e

mlunoe:
Code looks good, will test out as soon as I am done with my current task

orlandohohmeier:
LGTM 💯 ✨

PR #1043, opened 2016-09-02
Add organization routes dynamically

mlunoe:
This PR makes organization routes being added dynamically when sub-routes are defined.

### DIFF
--- a/plugins/oauth/hooks.js
+++ b/plugins/oauth/hooks.js
@@ -1,7 +1,7 @@
 /* eslint-disable no-unused-vars */
 import React from 'react';
 /* eslint-enable no-unused-vars */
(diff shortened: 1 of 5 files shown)


mlunoe:
ping @philipnrmn and @orlandohohmeier

orlandohohmeier:
@mlunoe Could you please some background? Why do we need those routs to be dynamic?

mlunoe:
Yes, sorry if that wasn't clear. The scenario where this need emerges if you turn on/off different plugins, e.g. oauth, auth-providers and organization.
In the case of all of the plugins turned off, we would like to see no organization tab all together (since there are no subtabs).

Does that answer your question?

PR #1045, opened 2016-09-05
DCOS-9604: Add Framework fields to service config blacklist

Poltergeist:
This adds framework fields from the state summary to the service config blacklist. To remove them from the service json in the service form.


wavesoft:
It feels so wrong having to do this 😨 
But yeah, with the current situation, this looks like a viable solution.

orlandohohmeier:
LGTM! 🚢

PR #1046, opened 2016-09-06
DCOS-9162: Add force setting to kill and scale if application is currently locked

Poltergeist:
The behaviour seems a bit off here. But the kill and scale does work now with the force option.

killandscale mov

### DIFF
--- a/src/js/components/KillTaskModal.js
+++ b/src/js/components/KillTaskModal.js
@@ -44,16 +44,16 @@ class KillTaskModal extends mixin(StoreMixin) {
     }
   }
 
-  shouldForceUpdate(message = this.state.errorMsg) {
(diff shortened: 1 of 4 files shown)


orlandohohmeier:
@Poltergeist The behavior inconsistent and feels very strange. Could you please align the `force` behavior with the other actions?

Poltergeist:
I'll do my best.

Poltergeist:
taskkill mov
Updated! I tried to do this in the most non destructive way.

mlunoe:
Also maybe update the text to say something that explains the action a little bit more, when you force it

mlunoe:
Added `Force ` in front of the button text to show the action better

PR #1047, opened 2016-09-06
DCOS-9498: Remove ipAddress on network switch to HOST

Poltergeist:
This PR will make sure that the ipAddress field in the config is removed if the user selects a different option from the dropdown.


orlandohohmeier:
LGTM 💯 ✨

PR #1048, opened 2016-09-06
[Pods] The pod UI-specific views

wavesoft:
image

PR #1050, opened 2016-09-07
[Pods] WIP: Introducing debug tab

wavesoft:
---
⚠️ Depends on #1048  

---

This PR adds a `Debug` tab to the pods-specific view.


wavesoft:
Replaced with #1082

PR #1051, opened 2016-09-07
DCOS-8898: Change history embed with historySummary embed on jobs request

Poltergeist:
This changes the emebed parameter for the jobs request and we only request the history
summary from metronome now. This will remove a lot of overhead in request size for
clusters which are using a lot of jobs.


orlandohohmeier:
LGTM 🎆 🚢 🇮🇹

PR #1052, opened 2016-09-07
Removing id property from ServiceFormModal

wavesoft:
_Not addressing pod-only problem, but is needed for a feature on the pods UI._

The `id=` property was causing some troubles on some new implementation that I am working on and it didn't look good eitherwise. Suggested a modification.


Poltergeist:
👍  LGTM 🎆 👻

mlunoe:
Rebased `feature/pods` and then this PR to be up to date

PR #1053, opened 2016-09-07
Always use itemID instead of description. refs: DCOS-9617

rcorral:


mlunoe:
Looks good, works well :shipit:

PR #1054, opened 2016-09-07
Fixes network tab

rcorral:
Renames discovery to loadBalanced that was missed and ensures there's no port definitions when setting up a VIP.


orlandohohmeier:
Thanks a bunch for fixing the missed `discovery` 💯 🚢 🇮🇹

PR #1055, opened 2016-09-07
Cleanup nodes routes

mlunoe:
This PR introduces routes for tabs on the node detail page

Before:

After:


rcorral:
Code looks good! I found two issues while testing:

Loading is off center:
img


This might've already been an issue before, if so, feel free not to fix. But we're losing the /health/ segment
gif

mlunoe:
Nice finds! Fixed both!

PR #1056, opened 2016-09-07
Properly set the expose network checkbox prop ID

orlandohohmeier:
Adjust `getExposeNetworkingCheckboxes` method to define a  checkbox prop
ID. Without this ID the`FormUtil.removePropID` method (used by the remove row button)  fails to remove the respective fields.
This change fixes an issue where it was impossible to remove endpoints
in the networking tab.


rcorral:
Works great!

PR #1057, opened 2016-09-07
Add port definitions for VIPs and DNS address for load balanced services. refs: DCOS-9191

rcorral:
VIP and Bridge port definition
img

Host port definition
img

### DIFF
--- a/src/js/components/ConfigurationView.js
+++ b/src/js/components/ConfigurationView.js
@@ -122,9 +122,9 @@ class ConfigurationView extends mixin(StoreMixin) {
         );
 
         // Check if this port is load balanced
-        if (portMapping.servicePort) {
+        if (portMapping.containerPort) {
           headerValueMapping[serviceAddressKey] =
-            this.buildHostName(id, portMapping.servicePort);
+            this.buildHostName(id, portMapping.containerPort);
         } else {
           delete headerValueMapping[serviceAddressKey];
         }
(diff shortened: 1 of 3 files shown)


src/js/components/ServiceForm.js
```diff
@@ -66,7 +67,7 @@ class ServiceForm extends SchemaForm {
     Hooks.doAction('serviceFormMount', this);
     // Update definition and build model
     this.updateDefinitions();
-    this.bananas(true, 0);
+    this.recalculateFormModelAndRender(true, 0);
```
rcorral: Sneaking this in here so that we no longer have a method called bananas.

src/js/components/ServiceForm.js
```diff
@@ -186,7 +187,7 @@ class ServiceForm extends SchemaForm {
         // Build out the hostname mappings
         let serviceAddress = HostUtil.stringToHostname(model.general.id);
         if (serviceAddress) {
-          serviceAddress += '.marathon.l4lb.thisdcos.directory';
+          serviceAddress += Networking.L4LB_ADDRESS;
```
rcorral: This is the real change.

mlunoe:
Code looks good, but I am seeing duplicate Port Definitions:

rcorral:
@mlunoe fixed thee issue!

src/js/components/ConfigurationView.js
```diff
@@ -219,12 +305,17 @@ class ConfigurationView extends mixin(StoreMixin) {
       );
     }
 
+    let portsSection = this.getVIPPortDefinitionsSection(config);
+    if (!portsSection) {
+      portsSection = this.getPortDefinitionsSection(config);
+    }
```
rcorral: @orlandohohmeier could you take a peek at this? Not sure how to handle only showing one of them otherwise. I think to the user we should just show a "Ports Definition" because in the modal it shows as one grouping instead of two.

mlunoe: Why don't we just move this call (`this.getPortDefinitionsSection(config);`) to be returned in stead of `null` here: https://github.com/dcos/dcos-ui/pull/1057/files#diff-b56f4ccf31412430b3161e86081b2e8eR106

orlandohohmeier: Yeah, let's try to combine these two

rcorral: Moved!

mlunoe:
Looks good! :shipit: Feel free to merge once tests pass

PR #1060, opened 2016-09-08
Rename method.

rcorral:


mlunoe:
:shipit: when build passes

PR #1062, opened 2016-09-09
[Pods] Adjust Marathon groups (pod) fixture

orlandohohmeier:
Fix inconsistencies in the Marathon groups (pod) fixture


wavesoft:
Nice, thanks 👍

PR #1067, opened 2016-09-12
Adding expandAll property on ExpandingTable

wavesoft:
This flag is useful when searching on the child elements of the `ExpandingTable` and you want to expand the entire table to make the items accessible. This should not discard the current selected item state.


mlunoe:
The code looks good, but let's get some tests

mlunoe:
Going to merge this, and let's handle the tests in a separate PR

PR #1069, opened 2016-09-12
[Pods] Adding PodViewFilter

wavesoft:
This introduces a re-usable filter bar component that contains plain text search and buttons.

Example usage:

```jsx
  // Required UI State properties
  this.setState({
    filter: {
      text: '',
      status: ''
    }
  });

  ...

  // Status mapper is a function that receives an
  // item from the `items` array and returns
  // a string that represents the status of this item
  var statusMapper = function(item) {
    if (!item.active) {
      return 'inactive';
    }
    return 'active';
  }

  ...

  <PodViewFilter
    filter={this.state.filter}
    inverseStyle={inverseStyle}
    items={items}
    onFilterChange={this.handleFilterChange}
    statusChoices={['all', 'active', 'inactive']}
    statusMapper={statusMapper}
    />
```

### DIFF
--- a/src/js/components/PodViewFilter.js
+++ b/src/js/components/PodViewFilter.js
@@ -41,18 +41,6 @@ class PodViewFilter extends React.Component {
     );
   }
 
-  getChildrenArray() {
-    let {children} = this.props;
-    if (!children) {
-      return [];
-    }
-    if (!Array.isArray(children)) {
-      return [children];
-    }
-
-    return children;
-  }
-
   getFilterButtons() {
     let {filter, items, inverseStyle, statusMapper, statusChoices} = this.props;
     if (!statusChoices.length) {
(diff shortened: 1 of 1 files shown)


src/js/components/PodViewFilter.js
```diff
@@ -0,0 +1,115 @@
...
+  getFilterButtonContent(filterName, count) {
+    return (
+      <span className="button-align-content">
+        <span className="label">{StringUtil.capitalize(filterName)}</span>
+        <span className="badge">{count || 0}</span>
+      </span>
+    );
+  }
+
+  getChildrenArray() {
```
mlunoe: This shouldn't be necessary, I think you can just add `{this.props.children}` in the render method

src/js/components/PodViewFilter.js
```diff
@@ -0,0 +1,115 @@
...
+  }
+
+  getFilterButtons() {
+    let {filter, items, inverseStyle, statusMapper, statusChoices} = this.props;
+    if (!statusChoices.length) {
+      return null;
+    }
+
+    return (
+        <FilterButtons
```
mlunoe: Indentation

src/js/components/PodViewFilter.js
mlunoe: Because of this: https://github.com/dcos/dcos-ui/blob/master/src/js/components/FilterBar.js#L47

mlunoe: I removed this here: https://github.com/dcos/dcos-ui/pull/1069/commits/6254e5cad83be310776401bfb8beeff606b512a2

mlunoe:
Thinking a bit about this afterwards I think we may be able to use this in multiple places, not only for pods. Could we create a ticket to roll this out for the rest of the application once `feature/pods` branch merges?

wavesoft:
I thought so to, but I had some naming issues. For example the `FilterBar` is already taken. Perhaps we need to spend some additional thinking on this, but sure 👍

PR #1071, opened 2016-09-12
[Pods] Skeleton pods view with placeholders

wavesoft:
This is only a placeholder view for the pods UI. (Breakdown of #1048 )

### DIFF
--- a/src/js/components/PodDetail.js
+++ b/src/js/components/PodDetail.js
@@ -1,0 +1,94 @@
+import classNames from 'classnames';
+import mixin from 'reactjs-mixin';
+import React from 'react';
+
+import Breadcrumbs from './Breadcrumbs';
+import Pod from '../structs/Pod';
+import PodHeader from './PodHeader';
+import TabsMixin from '../mixins/TabsMixin';
+
+const METHODS_TO_BIND = [
+];
+
+class PodDetail extends mixin(TabsMixin) {
+  constructor() {
+    super(...arguments);
+
+    this.tabs_tabs = {
+      instances: 'Instances',
+      configuration: 'Configuration',
+      debug: 'Debug'
+    };
+
+    this.state = {
+      currentTab: Object.keys(this.tabs_tabs).shift()
+    };
+
+    METHODS_TO_BIND.forEach((method) => {
+      this[method] = this[method].bind(this);
+    });
+  }
+
+  renderConfigurationTabView() {
+    let {pod} = this.props;
+    return (
+      <pre>
+      &lt;PodConfigurationView pod={pod.getId()} /&gt;
+      </pre>
+    );
+  }
+
+  renderDebugTabView() {
+    let {pod} = this.props;
+    return (
+      <pre>
+      &lt;PodDebugTabView pod={pod.getId()} /&gt;
+      </pre>
+    );
+  }
+
+  renderInstancesTabView() {
+    let {pod} = this.props;
+    return (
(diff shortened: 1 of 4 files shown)


src/js/components/PodDetail.js
```diff
@@ -0,0 +1,94 @@
...
+      </pre>
+    );
+  }
+
+  render() {
+    const {pod} = this.props;
+
+    return (
+      <div className="flex-container-col">
+        <div className={classNames(
```
jfurrow: Let's just pass a string with all of these classnames since we're not conditionally applying any of them.

src/js/components/PodHeader.js
```diff
@@ -0,0 +1,108 @@
...
+    let podIcon = null;
+    let podImages = pod.getImages();
+    if (podImages && podImages['icon-large']) {
+      podIcon = (
+        <img src={podImages['icon-large']} />
+      );
+    }
+
+    const tabs = (
+      <ul className={classNames([
```
jfurrow: Let's just pass a string with all of these classnames.

mlunoe:
Failed to rebase this guy, could you take a look?

src/js/components/PodDetail.js
```diff
@@ -0,0 +1,94 @@
...
+    this.state = {
+      currentTab: Object.keys(this.tabs_tabs).shift()
+    };
+
+    METHODS_TO_BIND.forEach((method) => {
+      this[method] = this[method].bind(this);
+    });
+  }
+
+  renderConfigurationTabView() {
```
mlunoe: Could you create routes for these, so we can use the `tabs_getRoutedTabs` and `RouteHandler`?

src/js/components/PodDetail.js
wavesoft: Done

src/js/components/PodHeader.js
wavesoft: Done

src/js/components/PodDetail.js
wavesoft: I will give it a try ;)

orlandohohmeier: Let's not create routes for these tabs, as the respective service app tabs aren't routed at the moment and we should do this in one go.

orlandohohmeier:
@wavesoft Aside from the lack of tests, solid work. Great job! ✨ 💯

wavesoft:
The only relevant test that I could think of is to test if `PodDetail` is rendered when a pod instance is passed to the tab.

orlandohohmeier:
@wavesoft please have a look at the `ServiceDetail` tests. We should have the same level of test for pods.

wavesoft:
This will be possible when we have merged all the individual pages. I added it on my ToDo list when the relevant PRs get merged.

src/js/components/PodDetail.js
mlunoe: It takes a lot of overhead to make tabs routed. If you have to re-do the code. I would really like to see the routes implemented here. We have a task to track making the corresponding service tabs routed

orlandohohmeier: I understand your concerns, but we shouldn't extend the scope due the named reasons and  the time constrains. Let's create an issue and add the routes once we've merged the pods stuff into master.

mlunoe: Ok, sounds good

PR #1074, opened 2016-09-12
Adding priority to userLogoutSuccess

mlunoe:
This PR adds priority to userLogoutSuccess action on the tracking plugin


mlunoe:
Since priority defaults to 10 this PR is unnecessary

PR #1075, opened 2016-09-13
Add Task Name column to task table.

rcorral:


mlunoe:
Code looks good and works well!

We could consider hiding the name column on medium screen size and downwards:


I'll leave that up to you. It shouldn't hold back the PR

rcorral:
Fixed!

PR #1068, opened 2016-09-12
[Pods] Exposing additional information to the pods struct

wavesoft:
Exposing some additional information on the pods struct.

PR #1077, opened 2016-09-13
[Pods] Introducing PodUtil

wavesoft:
This introduces the `PodUtil` helper that for now contains free-text filter functions.

### DIFF
--- a/src/js/utils/PodUtil.js
+++ b/src/js/utils/PodUtil.js
@@ -20,7 +20,7 @@ var PodUtil = {
    * Returns true if the container instance matches
    * the free-text given.
    *
-   * @param {PodInstance} instance - The Container to test
+   * @param {PodInstance} instance - The instance to test
    * @param {string} text - The text to test against
    * @returns {boolean} - Returns true if the test passes
    */
--- a/src/js/utils/PodUtil.js
+++ b/src/js/utils/PodUtil.js
@@ -1,0 +1,42 @@
+var PodUtil = {
+
+  /**
+   * Returns true if the container given matches
+   * the free-text given.
+   *
+   * @param {PodContainer} container - The Container to test
+   * @param {string} text - The text to test against
+   * @returns {boolean} - Returns true if the test passes
+   */
+  isContainerMatchingText(container, text) {
+    if (!text) {
+      return true;
+    }
+
+    return container.getName().indexOf(text) >= 0;
+  },
+
+  /**
+   * Returns true if the container instance matches
+   * the free-text given.
+   *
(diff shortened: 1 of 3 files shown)


src/js/utils/PodUtil.js
```diff
@@ -0,0 +1,42 @@
...
+   * @param {PodContainer} container - The Container to test
+   * @param {string} text - The text to test against
+   * @returns {boolean} - Returns true if the test passes
+   */
+  isContainerMatchingText(container, text) {
+    if (!text) {
+      return true;
+    }
+
+    return container.getName().indexOf(text) >= 0;
```
mlunoe: If you need the `getName` for here then, I would suggest that we use `getId` or call `getId` in the `getName` function where we need it to be the same

orlandohohmeier: @mlunoe I'm not sure I understand your comment. Whats the benefit of using `getId`?

wavesoft: @orlandohohmeier he refers to hist modification here : https://github.com/dcos/dcos-ui/pull/1068/commits/ce12e9b10635e4c92415ed90dcf750505110b104 (that breaks this PR).

I will re-introduce `getName()` and it will call `getId()` for now. However I wanted this separation because `getId()` should be used on server communication, while `getName()` on the UI.

wavesoft:
Added tests

src/js/utils/PodUtil.js
```diff
@@ -0,0 +1,42 @@
...
+  },
+
+  /**
+   * Returns true if the container instance matches
+   * the free-text given.
+   *
+   * @param {PodInstance} instance - The Container to test
```
orlandohohmeier: Looks like you confused pod instances and containers here

orlandohohmeier:
Minor nits aside, great job! 👍 💯

src/js/utils/PodUtil.js
wavesoft: Oops. Fixed

src/js/utils/PodUtil.js
```diff
@@ -0,0 +1,42 @@
...
+   * @param {string} text - The text to test against
+   * @returns {boolean} - Returns true if the test passes
+   */
+  isInstanceOrChildrenMatchingText(instance, text) {
```
wavesoft: I am adding this on the todo because it's going to add some merging overhead on the linked PRs.

src/js/utils/PodUtil.js
mlunoe: Ok, however I am not sure I understand the last bit about server communication?

PR #1070, opened 2016-09-12
[Pods] Adjust service actions to support pods

orlandohohmeier:
Adjust the service actions to support pods. I added mocked versions to enable proper testing.

_Please excuse the rather large number of changes. I figured it might be easier to land this with only one PR instead of a number smaller ones as I'm on vacation and wouldn't be able to do the rebase dance._

PR #1080, opened 2016-09-13
[Pods] Adding pod actions to pod-specific views

wavesoft:
---
~~⚠️  Depends on #1071~~
~~⚠️  Depends on #1070~~

---

This PR introduces the action buttons and links them to the appropriate pod-specific store actions.

_WIP: Using this PR as a base in order to integrate the `ServiceFormModal` in a separate PR_

### DIFF
--- a/src/js/components/PodDetail.js
+++ b/src/js/components/PodDetail.js
@@ -4,10 +4,12 @@ import React from 'react';
 import Breadcrumbs from './Breadcrumbs';
 import Pod from '../structs/Pod';
 import PodInstancesView from './PodInstancesView';
+import PodActionItem from '../constants/PodActionItem';
 import PodHeader from './PodHeader';
 import TabsMixin from '../mixins/TabsMixin';
 
 const METHODS_TO_BIND = [
+  'handleAction'
 ];
 
 class PodDetail extends mixin(TabsMixin) {
@@ -27,6 +29,22 @@ class PodDetail extends mixin(TabsMixin) {
     METHODS_TO_BIND.forEach((method) => {
       this[method] = this[method].bind(this);
     });
+  }
+
+  handleAction(action) {
+    switch (action.id) {
+      case PodActionItem.SCALE:
+        console.debug('Scale');
+        break;
+
+      case PodActionItem.EDIT:
+        console.debug('Edit');
+        break;
+
+      case PodActionItem.DESTROY:
+        console.debug('Destroy');
+        break;
+    }
   }
 
   renderConfigurationTabView() {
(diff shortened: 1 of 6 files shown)


philipnrmn:
I'm happy with the direction of this PR. I haven't done a line-by-line review as it's obviously still WIP and there's nothing more annoying than a close interrogation of unfinished work. Is it in a hold-merge state because of its (recently resolved) PR dependencies or will there be more commits on this branch?

mlunoe:
I tried rebasing this for you, but ran into conflicts. I'll let you take care of this guy

src/js/constants/PodActionItem.js
```diff
@@ -0,0 +1,8 @@
+const PodActionItem = {
+  EDIT: 'edit',
+  DESTROY: 'destroy',
+  SCALE: 'scale',
+  SUSPED: 'suspend'
```
philipnrmn: s/SUSPED/SUSPEND/

src/js/components/PodDetail.js
```diff
@@ -21,14 +28,41 @@ class PodDetail extends mixin(TabsMixin) {
...
+      activeActionDialog: null
     };
 
     METHODS_TO_BIND.forEach((method) => {
       this[method] = this[method].bind(this);
     });
   }
 
+  handleAction(action) {
+    switch (action.id) {
```
philipnrmn: Not a big fan of using an object arg with a single parameter here, it feels like more complexity than we need. 

If we made the action type the first name, we could bind in the JSX like
```onClick={props.handleAction.bind(null, PodActionItem.DESTROY)}```

But I'd actually prefer individual explicit handleDestroy/handleEdit/handleScale/handleSuspend  methods.

File: src/js/components/PodDetail.js [hunk 3]
```diff
@@ -56,17 +90,36 @@ class PodDetail extends mixin(TabsMixin) {
...
     const {pod} = this.props;
+    const {activeActionDialog} = this.state;
 
     return (
       <div className="flex-container-col">
         <div className="container-pod container-pod-divider-bottom-align-right container-pod-short-top flush-bottom flush-top media-object-spacing-wrapper media-object-spacing-narrow">
           <Breadcrumbs />
           <PodHeader
-            onDeploy={this.onActionDeploy}
+            onAction={this.handleAction}
```
philipnrmn: I think I'd prefer to have individual methods than one overloaded method.

src/js/components/PodDetail.js
MatApple: +1

src/js/components/PodDetail.js
wavesoft: Yeah, I also aggree on this. I will have a look on the ticket.

PR #1082, opened 2016-09-13
[Pods] Fleshing out PodDebugTabView

wavesoft:
---
~~⚠️ Requires #1071 to be merged~~ Merged!

---

This PR implements the DebugTab view. It shows the time of the last change, and it also extracts last termination information according to this spec: https://github.com/mesosphere/marathon/blob/feature/pods/docs/docs/rest-api/public/api/v2/types/podStatus.raml#L217

image

### DIFF
--- a/src/js/components/PodContainerTerminationTable.js
+++ b/src/js/components/PodContainerTerminationTable.js
@@ -1,0 +1,136 @@
+import classNames from 'classnames';
+import React from 'react';
+import {Table} from 'reactjs-components';
+
+import ResourceTableUtil from '../utils/ResourceTableUtil';
+import TableUtil from '../utils/TableUtil';
+
+const METHODS_TO_BIND = [
+  'getColumnHeading',
+  'renderColumnID',
+  'renderColumnState',
+  'renderColumnTerminationCode',
+  'renderColumnTerminationMessage'
+];
+
+class PodContainerTerminationTable extends React.Component {
+  constructor() {
+    super(...arguments);
+
+    METHODS_TO_BIND.forEach((method) => {
+      this[method] = this[method].bind(this);
+    });
+  }
+
+  getColumns() {
+    let {getClassName} = ResourceTableUtil;
+
+    return [
+      {
+        className: getClassName,
+        headerClassName: getClassName,
+        prop: 'ID',
+        heading: this.getColumnHeading,
+        render: this.renderColumnID
+      },
+      {
+        className: getClassName,
+        headerClassName: getClassName,
+        prop: 'Last State',
+        heading: this.getColumnHeading,
+        render: this.renderColumnState
+      },
+      {
+        className: getClassName,
+        headerClassName: getClassName,
+        prop: 'Code',
+        heading: this.getColumnHeading,
+        render: this.renderColumnTerminationCode
+      },
+      {
+        className: getClassName,
+        headerClassName: getClassName,
+        prop: 'Message',
+        heading: this.getColumnHeading,
+        render: this.renderColumnTerminationMessage
+      }
+    ];
+  }
+
+  getColGroup() {
+    return (
+      <colgroup>
+        <col />
+        <col style={{width: '160px'}} />
+        <col style={{width: '80px'}} />
+        <col style={{width: '160px'}} />
+      </colgroup>
+    );
+  }
+
+  getColumnHeading(prop, order, sortBy) {
+    let caretClassNames = classNames(
+      'caret',
+      {
+        [`caret--${order}`]: order != null,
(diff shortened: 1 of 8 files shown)


jfurrow:
I'd suggest moving the `Containers` row to its own section, so that the the table can fill the width of the page, rather than the column it's in currently. Here's my markdown representation of the idea:

#### Last Changes
```
Configuration   Wed...
```

#### Last Terminations
```
Terminated at...
Instance ID     pod...
Message         Pod...
Started At      Wed...
```

#### Containers
```
Containers table goes here
```

src/js/components/PodDetail.js
```diff
@@ -0,0 +1,93 @@
...
+      </pre>
+    );
+  }
+
+  render() {
+    const {pod} = this.props;
+
+    return (
+      <div className="flex-container-col">
+        <div className={classNames(
```
jfurrow: Let's just pass a string with all of these classnames since we're not conditionally applying any of them.

wavesoft:
Note that containers is an array on the last termination (that on it's an array on it's own).

src/js/components/PodDetail.js
wavesoft: This was a solution to break the long lines. Should I completely discard this?

wavesoft:
Regarding https://github.com/dcos/dcos-ui/pull/1082#issuecomment-246751759, I am in a discussion with @leemunroe to find out a nice solution to this. 

For now I moved the table at the end of the list:

image

src/js/components/PodContainerTerminationTable.js
```diff
@@ -0,0 +1,136 @@
...
+  render() {
+    return (
+        <Table
+          className={this.props.className}
+          columns={this.getColumns()}
+          colGroup={this.getColGroup()}
+          data={this.props.containers}
+          itemHeight={TableUtil.getRowHeight()}
+          sortBy={{prop: 'ID', order: 'desc'}}
+          />
```
mlunoe: #nit Let's move this up to be `sortBy={{prop: 'ID', order: 'desc'}} />`

src/js/components/PodDebugTabView.js
```diff
@@ -0,0 +1,84 @@
...
+import Pod from '../structs/Pod';
+import PodContainerTerminationTable from './PodContainerTerminationTable';
+import TimeAgo from './TimeAgo';
+
+class PodDebugTabView extends React.Component {
+  getTerminationHistory() {
+    let history = this.props.pod.getTerminationHistoryList().getItems();
+    if (!history.length) {
+      return (
+          <div>(No data)</div>
```
mlunoe: #nit indentation

src/js/components/PodDetail.js
```diff
@@ -40,10 +41,8 @@ class PodDetail extends mixin(TabsMixin) {
   renderDebugTabView() {
     let {pod} = this.props;
     return (
-      <pre>
-      &lt;PodDebugTabView pod={pod.getId()} /&gt;
-      </pre>
-    );
+        <PodDebugTabView pod={pod} />
```
mlunoe: #nit indentation

src/js/components/PodContainerTerminationTable.js
mlunoe: done

src/js/components/PodDebugTabView.js
mlunoe: done

src/js/components/PodDetail.js
mlunoe: done

mlunoe:
Let's wait for @leemunroe's approval

wavesoft:
After a discussion with lee yesterday we ended up with something like this:

image

* _No results_: Title `Last Terminations`, text `(No data)`
* _One result_: Title `Last Termination (?? ago)`
* _More than one result_: Append title `Terminated at ??/??/?? (?? ago)` for every other item

mlunoe:
I get this error when I enter the debug tab of a pod that has no tasks:

mlunoe:
Should we handle the error I pointed out, or is this expected?
Could you elaborate a little on the process to come to fix this?
@MatApple Do you know something I don't?

MatApple:
@mlunoe I tested it and didn't get the error.

mlunoe:
I get the error when I use fixtures and enter the debug tab. @wavesoft could you check it out?

wavesoft:
@mlunoe , @MatApple this seems to be unrelated. It looks like the fixtures have not reflected the latest API advancements. If you test it against a live cluster it works fine.

mlunoe:
Right, so we should update the fixtures!

PR #1083, opened 2016-09-13
[Pods] DCOS-9613: 1/2 Separating Application to ApplicationSpec

wavesoft:
This PR separates `Application` struct to `Application` + `ApplicationSpec`.

The first is used when we need to inspect the *status of a running instance*, while the latter is used when we *create or edit* an application.

### DIFF
--- a/src/js/components/ServiceDetailConfigurationTab.js
+++ b/src/js/components/ServiceDetailConfigurationTab.js
@@ -5,6 +5,7 @@ import ConfigurationView from './ConfigurationView';
 import DCOSStore from '../stores/DCOSStore';
 import MarathonStore from '../stores/MarathonStore';
 import Service from '../structs/Service';
+import ApplicationSpec from '../structs/ApplicationSpec';
 import ServiceFormModal from './modals/ServiceFormModal';
 import ServiceUtil from '../utils/ServiceUtil';
 
@@ -56,8 +57,8 @@ class ServiceDetailConfigurationTab extends React.Component {
       this.props.service.getVersions().get(this.state.selectedVersionID);
 
     MarathonStore.editService(
-      ServiceUtil.getAppDefinitionFromService(
-        new Service(serviceConfiguration)
+      ServiceUtil.getDefinitionFromSpec(
+        new ApplicationSpec(serviceConfiguration)
       )
     );
   }
--- a/src/js/structs/Application.js
+++ b/src/js/structs/Application.js
@@ -1,5 +1,4 @@
 import ApplicationSpec from './ApplicationSpec';
-import {cleanServiceJSON} from '../utils/CleanJSONUtil';
 import Config from '../config/Config';
 import FrameworkUtil from '../utils/FrameworkUtil';
 import HealthStatus from '../constants/HealthStatus';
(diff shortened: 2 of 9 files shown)


mlunoe:
Rebased this for you. Didn't have time to go through it, though :(

src/js/structs/Application.js
```diff
@@ -7,57 +9,28 @@ import TaskStats from './TaskStats';
...
-    return this.get('executor');
-  }
-
+  /**
+   * @override
+   */
   getSpec() {
-    // DCOS-9613: This should be properly implemented
-    return this;
+    return new ApplicationSpec(cleanServiceJSON(this.get()));
```
orlandohohmeier: Let's move the clean json part to the spec, so that we can creates specs directly from the server response.

src/js/components/ServiceDetailConfigurationTab.js
```diff
@@ -56,8 +57,8 @@ class ServiceDetailConfigurationTab extends React.Component {
       this.props.service.getVersions().get(this.state.selectedVersionID);
 
     MarathonStore.editService(
-      ServiceUtil.getAppDefinitionFromService(
-        new Service(serviceConfiguration)
+      ServiceUtil.getDefinitionFromSpec(
```
orlandohohmeier: We don't need the util once #1070 is merged

src/js/structs/Application.js
```diff
@@ -7,57 +9,28 @@ import TaskStats from './TaskStats';
...
-  }
-
-  getContainer() {
-    return this.get('container');
-  }
-
-  getConstraints() {
-    return this.get('constraints');
-  }
+  // getContainer() {
```
mlunoe: Could we just add this when we need it?

src/js/utils/ServiceUtil.js
```diff
@@ -451,55 +451,23 @@ const ServiceUtil = {
...
+    let definition = spec.toJSON();
 
-    appDefinition.updateStrategy = service.getUpdateStrategy();
-    appDefinition.residency = service.getResidency();
-    appDefinition.ipAddress = service.getIpAddress();
+    if (spec instanceof ApplicationSpec) {
 
-    // Environment Variables
-    appDefinition.env = service.getEnvironmentVariables();
+      // TODO: It looks that this hook operates over Application (Service)
```
mlunoe: Then let's make the change!

src/js/structs/Application.js
wavesoft: Done

src/js/components/ServiceDetailConfigurationTab.js
wavesoft: Yup, I kept it for making the PR functional on it's own until we have that PR merged.

src/js/utils/ServiceUtil.js
wavesoft: ok, I will pass down the `spec`. Where can I find the plug-ins that might be affected? -- If any

src/js/structs/Application.js
wavesoft: Blast. That should not be there. I commented that part out until I figure out if it was used or not. Apparently we are not using this function anywhere. Instead we are using the `getContainerSettings` (in spec).

mlunoe:
I worked with @wavesoft to get this into a mergable state.
I am merging this in to get the ball rolling, and unblock @wavesoft.
If there are any follow up comments, let's address them in a separate PR.

PR #1090, opened 2016-09-14
[CNVS Upgrade] Fix broken unit tests

jfurrow:
Velocity won't properly report the status of the unit tests because this still relies on the `cnvs-upgrade/master` branch of both `reactjs-components` and the plugins repository, but you can see them passing locally:

### DIFF
--- a/src/js/components/Panel.js
+++ b/src/js/components/Panel.js
@@ -1,6 +1,8 @@
 import classNames from 'classnames/dedupe';
 import React from 'react';
 
+import Util from '../utils/Util';
+
 const defaultClasses = {
   panel: 'panel',
   content: 'panel-content panel-cell panel-cell-narrow panel-cell-short panel-cell-borderless',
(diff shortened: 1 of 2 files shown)


jfurrow:
ping @MatApple :)

src/js/components/Panel.js
```diff
@@ -48,7 +50,8 @@ var Panel = React.createClass({
     let panelClasses = classNames(defaultClasses.panel, props.className);
 
     return (
-      <div className={panelClasses}>
+      <div className={panelClasses}
+        {...Util.omit(props, ['children', 'className', 'heading', 'footer'])}>
```
MatApple: Do we need to add extra "unknown" attributes?

jfurrow: I'm not sure what you mean, can you clarify?

jfurrow: We talked offline, but this is done!

PR #1096, opened 2016-09-15
Use parsedVersion instead of version.

rcorral:


MatApple:
@rcorral do we still want this in master? Looks like it has merged to 1.8 already?

mlunoe:
MARGED!

tamarrow:
thanks!

PR #1021, opened 2016-08-30
DCOS-9285: Disable checkboxes for non-marathon tasks

nLight:
this PR addresses the issue with a possibility to check and kill tasks that haven't been started by marathon. The issue is resolved by adding additional field `startedBy` with the name of a framework, i.e. `spark` that actually started the task, so we can use that information to filter out non-marathon tasks.

Additionally this PR fixes the way how CheckboxTable component renders the heading checkbox when there are disabled items.

screen shot 2016-08-30 at 14 45 43

PR #1097, opened 2016-09-16
DCOS-9285: Move scheduler name assignment to MesosStateActions

nLight:
In this PR I implemented the after merge comments from @mlunoe https://github.com/dcos/dcos-ui/pull/1021 regarding the place of the scheduler name assigning logic. So the logic has been moved to the `MesosStateActions` as a central place, where the data comes in into the app, therefore the data will be consistent across the whole app.

Works on the Services page and on the Nodes page as pointed out by @rcorral in https://github.com/dcos/dcos-ui/pull/1066

I also moved the CheckboxTable fix as well.

screen shot 2016-09-16 at 13 16 38

screen shot 2016-09-16 at 13 15 46

### DIFF
--- a/src/js/components/CheckboxTable.js
+++ b/src/js/components/CheckboxTable.js
@@ -52,7 +52,9 @@ class CheckboxTable extends React.Component {
 
   bulkCheck(isChecked) {
     let checkedIDs = [];
-    let {data, onCheckboxChange, uniqueProperty} = this.props;
+    let {data, onCheckboxChange, uniqueProperty, disabledItemsMap} = this.props;
+
+    data = data.filter((datum) => !disabledItemsMap[datum[uniqueProperty]]);
 
     if (isChecked) {
       data.forEach(function (datum) {
@@ -75,13 +77,14 @@ class CheckboxTable extends React.Component {
   renderHeadingCheckbox() {
     let checked = false;
     let indeterminate = false;
-    let {allowMultipleSelect, checkedItemsMap, data} = this.props;
+    let {allowMultipleSelect, checkedItemsMap, disabledItemsMap, data} = this.props;
 
     if (!allowMultipleSelect) {
       return null;
     }
 
     let checkedItemsCount = Object.keys(checkedItemsMap).length;
+    let disabledItemsCount = Object.keys(disabledItemsMap).length;
 
     if (checkedItemsCount > 0) {
       indeterminate = true;
@@ -89,7 +92,7 @@ class CheckboxTable extends React.Component {
       checked = false;
     }
 
-    if (checkedItemsCount === data.length && checkedItemsCount !== 0) {
+    if (checkedItemsCount + disabledItemsCount >= data.length && checkedItemsCount !== 0) {
       checked = true;
       indeterminate = false;
     }
--- a/src/js/components/CheckboxTable.js
+++ b/src/js/components/CheckboxTable.js
@@ -92,7 +92,7 @@ class CheckboxTable extends React.Component {
       checked = false;
     }
 
-    if (checkedItemsCount + disabledItemsCount >= data.length && checkedItemsCount !== 0) {
+    if (checkedItemsCount + disabledItemsCount === data.length && checkedItemsCount !== 0) {
       checked = true;
       indeterminate = false;
     }
--- a/src/js/components/CheckboxTable.js
+++ b/src/js/components/CheckboxTable.js
@@ -54,7 +54,9 @@ class CheckboxTable extends React.Component {
     let checkedIDs = [];
     let {data, onCheckboxChange, uniqueProperty, disabledItemsMap} = this.props;
 
-    data = data.filter((datum) => !disabledItemsMap[datum[uniqueProperty]]);
+    data = data.filter(function (datum) {
+      return !disabledItemsMap[datum[uniqueProperty]];
+    });
 
     if (isChecked) {
       data.forEach(function (datum) {
--- a/src/js/utils/MesosStateUtil.js
+++ b/src/js/utils/MesosStateUtil.js
@@ -2,29 +2,24 @@ import Util from './Util';
 
 const RESOURCE_KEYS = ['cpus', 'disk', 'mem'];
 
-function setSchedulerName(name, tasks) {
+function setIsStartedByMarathonFlag(name, tasks) {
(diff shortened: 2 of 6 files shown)


wavesoft:
Looks good to me 💯  ✨ 
poking @mlunoe to confirm that this fixes the issue he mentioned here: https://github.com/dcos/dcos-ui/pull/1021#discussion-diff-78253915

src/js/utils/MesosStateUtil.js
```diff
@@ -2,8 +2,34 @@ import Util from './Util';
...
+
+    newTask.scheduler = name;
+
+    return newTask;
+  });
+}
+
 const MesosStateUtil = {
 
+  assignSchedulerNameToTasks(state) {
```
mlunoe: I mentioned in [my comment](https://github.com/dcos/dcos-ui/pull/1021#r78254227), we do have the scheduler (framework) information on the tasks. If you look at tasks and look for `framework_id`, that should be enough information for you to tell whether this scheduled from a marathon instance or not (by a look up of the framework in Mesos data):


We do a similar thing here: https://github.com/dcos/dcos-ui/blob/master/src/js/pages/task-details/TaskDetailsTab.js#L41

However, thinking a bit more about this, it will probably cause some issues when we do this for all tasks (which could be _many_).

So what I suggest is that we add another property, exactly as you are doing here, but in stead tells us what we need to know:
e.g. `isStartedByMarathon = framework.name === 'marathon'` for tasks that have been started by the root Marathon framework.

Since the framework name `marathon` have been reserved for the root Marathon, this should work.

This solution is, however still far from ideal (because of the vague name matching). What really needs to happen is we need some type of label on the Framework level (or task level even) that will tell us which Marathon framework it was started from, if any.
That way we will be more robust towards the future, when we need to support different Marathons in the UI.

cc. @pyronicide How can we get the ball rolling on getting such a label?

nLight: Hey @mlunoe, the thing is that there's an upcomming feature when we would like to i.e. draw a little icon next to the task identifying which scheduler has launched it. We actually started exactly from `isStartedByMarathon` flag approach. It was, again, made by name matching. I agree that idally we should get this info from the backend. If it's critical we can go back to `isStartedByMarathon` and matching by framework_id to solve that specific issue with checkboxes.

mlunoe: Where is that feature needed? In the task table? And how are you then going to get the icon from that framework?

Anyways, the only reliable way of telling which framework it belongs to, is through the `framework_id` property and find the framework like this: https://github.com/dcos/dcos-ui/blob/master/src/js/pages/task-details/TaskDetailsTab.js#L41

If that is going to go in the task table I would strongly encourage that we revisit this solution and instead make the icon that we need available on the task when we get the (root) `marathon` groups information, where the icon information is coming from.

orlandohohmeier: @mlunoe I agree that we should use the schedule ID to accurately identify tasks lunched by a framework but I don't see any issue with adding the framework name to the tasks. 
I would use the framework name for now and switch to the scheduler ID once we need to support multiple Marathon instances as it required extra API calls to fetch the respective IDs. Are you ok with that?

src/js/utils/MesosStateUtil.js
```diff
@@ -2,8 +2,34 @@ import Util from './Util';
 
 const RESOURCE_KEYS = ['cpus', 'disk', 'mem'];
 
+function setSchedulerName(name, tasks) {
+  return tasks.map(function (task) {
```
orlandohohmeier: We could change this to the following to make sure that we don't accidentally overwrite a property.

```
return tasks.map(function (task) {
   return  Object.assign({scheduler:name}, task);
}
```
What do you think?

_P.S. We might need to rename the prop to something different e.g. `framework_name` to avoid collisions._

src/js/utils/MesosStateUtil.js
mlunoe: I would prefer that we introduce the suggested label to tell whether it has been started by the root Marathon.
Let's tackle the other scenario when that PR comes up. It is hard for me to review that need until I understand what that task is about.

The reason I don't like the name, is that it feels like duplicating information from the framework (id and name) instead of having that one identifier + helpful properties that works for the use cases.

I don't see what else you need the name for that you can't use the `framework_id` for.

src/js/utils/MesosStateUtil.js
nLight: Implemented 👍  for suggestion!

src/js/utils/MesosStateUtil.js
nLight: @mlunoe switched back to `isStartedByMarathon` approach. Let's discuss the PR today quickly in Slack, so we get it through.

src/js/components/CheckboxTable.js
```diff
@@ -75,21 +79,22 @@ class CheckboxTable extends React.Component {
   renderHeadingCheckbox() {
     let checked = false;
     let indeterminate = false;
-    let {allowMultipleSelect, checkedItemsMap, data} = this.props;
+    let {allowMultipleSelect, checkedItemsMap, disabledItemsMap, data} = this.props;
```
mlunoe: More than 80 characters here. We have a rule to keep code (not strings and regexes) within an 80 character limit.

src/js/components/CheckboxTable.js
```diff
@@ -75,21 +79,22 @@ class CheckboxTable extends React.Component {
...
+    let disabledItemsCount = Object.keys(disabledItemsMap).length;
 
     if (checkedItemsCount > 0) {
       indeterminate = true;
     } else {
       checked = false;
     }
 
-    if (checkedItemsCount === data.length && checkedItemsCount !== 0) {
+    if (checkedItemsCount + disabledItemsCount === data.length && checkedItemsCount !== 0) {
```
mlunoe: More than 80 chars

mlunoe: Also, I feel some of this logic of setting `indeterminate` and `checked` can be simplified a bit.

mlunoe:
Merged this, and created a follow up PR for the nit fixes: https://github.com/dcos/dcos-ui/pull/1106

PR #1092, opened 2016-09-14
[CNVS Upgrade: Stylelint 1 of 3] Add styles linter

jfurrow:
This PR introduces [stylelint](https://github.com/stylelint/stylelint) and a bunch of rules! A definition and example of every rule can be found [here](https://github.com/stylelint/stylelint/blob/master/docs/user-guide/rules.md).

Let me know if you have any questions or concerns. I'd like to get the team's approval on these rules before fixing all of the warnings. I quickly tested [stylefmt](https://github.com/morishitter/stylefmt) to automatically format our LESS files to adhere to these rules and it seems to work well!

The webpack configuration modifications will display any lint warning each time the LESS is changed.

PR #1102, opened 2016-09-16
[CNVS Upgrade: Stylelint 3 of 3] Add stylelint npm script

jfurrow:
~~Depends on #1092.~~ Merged!

This PR separates the `lint` script into two separate scripts; an `eslint` script and `stylelint` script, so that both JS and LESS lint warnings will fail the `build` script.


package.json
```diff
@@ -119,8 +121,10 @@
     "postbuild-assets": "rm -rf node_modules/compression-webpack-plugin",
     "build": "npm run lint && npm test && npm run build-assets",
     "build-with-external-plugins": "npm run lint && npm run lint-plugins && npm test && npm run build-assets",
-    "lint": "bash -c 'opts=$@ ; set -f; if [[ $# -lt 1 ]]; then opts=\"src/js/**/*.js\"; fi ; ./node_modules/.bin/eslint --quiet ${opts}' 0",
-    "lint-plugins": "PLUGINS=$npm_config_externalplugins; npm run lint -- \"$PLUGINS/**/*.js\"",
+    "lint": "npm run stylelint && npm run eslint",
+    "eslint": "bash -c 'opts=$@ ; set -f; if [[ $# -lt 1 ]]; then opts=\"src/js/**/*.js\"; fi ; ./node_modules/.bin/eslint --quiet ${opts}' 0",
+    "stylelint": "bash -c 'opts=$@ ; set -f; if [[ $# -lt 1 ]]; then opts=\"src/styles/**/*.less\"; fi ; ./node_modules/.bin/stylelint ${opts}' 0",
```
MatApple: You'll need to update the `opts` to reflect what's in master.

MatApple:
For only adding 2 new npm dependencies our Shrinkwrap has inflated ~6k. I think something is awry

mlunoe:
I noticed the many dependencies. Unfortunately there is no way of excluding them for the style linter. It is all or nothing. Meaning that it includes e.g. both postcss support and the lodash library :(

Not sure what we can do about it

package.json
jfurrow: Done!

webpack/webpack.dev.babel.js
```diff
@@ -87,8 +87,7 @@ module.exports = Object.assign({}, webpackConfig, {
     new StyleLintPlugin({
       configFile: '.stylelintrc',
       files: '**/*.less',
-      syntax: 'less',
-      failOnError: false
```
mlunoe: Do we not want to keep this configuration, when running in dev mode?

jfurrow: I don't think so, because it's annoying for the build to fail just because of a lint error. This is how `eslint` behaves too; you're alerted about the warning, but the build still compiles.

mlunoe: I think we are saying the same thing. But, it is the default and that is why it is being removed.

jfurrow: Yep!

PR #1104, opened 2016-09-19
Lint plugins by default

orlandohohmeier:
I've adjusted the eslint file pattern to lint plugins by default. Fixed a linting error along the way.


mlunoe:
Arh! Good catch!

PR #1106, opened 2016-09-19
80 character rule nit fixes

mlunoe:
---
Follow up PR to this one: https://github.com/dcos/dcos-ui/pull/1097

---

This just fixes some 80 character rules

### DIFF
--- a/src/js/components/CheckboxTable.js
+++ b/src/js/components/CheckboxTable.js
@@ -79,22 +79,27 @@ class CheckboxTable extends React.Component {
   renderHeadingCheckbox() {
     let checked = false;
     let indeterminate = false;
-    let {allowMultipleSelect, checkedItemsMap, disabledItemsMap, data} = this.props;
+    let {
+      allowMultipleSelect,
+      checkedItemsMap,
+      disabledItemsMap,
+      data
+    } = this.props;
 
     if (!allowMultipleSelect) {
       return null;
     }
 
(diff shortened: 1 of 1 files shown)


nLight:
Thanks @mlunoe! 👍  Shall we enable warnings for that rule?

src/js/components/CheckboxTable.js
```diff
@@ -79,22 +79,27 @@ class CheckboxTable extends React.Component {
...
+      data
+    } = this.props;
 
     if (!allowMultipleSelect) {
       return null;
     }
 
-    let checkedItemsCount = Object.keys(checkedItemsMap).length;
-    let disabledItemsCount = Object.keys(disabledItemsMap).length;
+    let checkedCount = Object.keys(checkedItemsMap).length;
```
orlandohohmeier: I'm ok with this change but don't think that shortening variables names is generally a good fix to obey to the 80 characters rule.

orlandohohmeier:
LGTM! 🚢

src/js/components/CheckboxTable.js
mlunoe: I initially moved code around, until I realized I thought the `Items` part was redundant. And so I landed on this. But yes, I do agree. That should not be a fix

mlunoe:
@orlandohohmeier and I have been working on a solution for that. There was some trouble about how we would tackle our exceptions, or how, if we make it a hard rule, we would handle breaking up our long canvas class names (and other situations)

PR #1110, opened 2016-09-20
[Pods][Testing] Fixing the scaling

wavesoft:
Fixing the format of the `scaling` property of the pod spec, and updating `ServiceScaleFormModal` and `ServiceSuspendModal` to properly use it.

In the previous mocks we were using:
```js
spec: {
  scaling: {
    fixed: {
      instances: 1
    }
  }
}
```

while it should be:
```js
spec: {
  scaling: {
    kind: 'fixed',
    instances: 1
  }
}
```


src/js/components/modals/ServiceScaleFormModal.js
```diff
@@ -106,6 +106,17 @@ class ServiceScaleFormModal extends mixin(StoreMixin) {
           id: service.id,
           scaleBy: parseInt(instances, 10)
         });
+      } else if (service instanceof Pod) {
+        let spec = Object.assign({}, service.getSpec().get(),
```
orlandohohmeier: Let's move this logic to the `ServiceUtil`, so that we have one method to create scaled service specs. This has the benefit that we have an easy testable and reusable method and less logic within the view components.

wavesoft: I introduced `MarathonSpecUtil` for all the spec-related mutations. Let me know if that follows your thinking.

orlandohohmeier: Let's have one util that deals with all kinds of services. I also think that we could add this to the `ServiceUtil` instead of introducing yet another util collection.

wavesoft: Renamed to `ServiceSpecUtil` and removed the `Pods` namespace inside it, as we discussed.

src/js/utils/ServiceSpecUtil.js
```diff
@@ -0,0 +1,33 @@
...
+  /**
+   * Creates a new PodSpec with the scaling specifications
+   * changed to fixed, and with the number of instances defined
+   * from the second argument.
+   *
+   * @param {PodSpec} spec - The spec to change
+   * @param {number} instancesCount - The new number of instances
+   * @returns {PodSpec} - Returns a new PodSpec instance
+   */
+  setFixedScaling(spec, instancesCount) {
```
orlandohohmeier: Would love if we could make this a general scale service method that know how to deal with pods and apps? That would allow us to remove all the type checks form the views.
What do you think?

wavesoft: I aggree. Let's do this.

orlandohohmeier: Why is that? I'd only create both specs (pod and app) everything else stays the same.

wavesoft: I thought you meant introducing a new `scaleXXX` action on `MarathonStore`. Having a single utility for both specs sounds better.

wavesoft: Done 👍

src/js/utils/ServiceSpecUtil.js
```diff
@@ -0,0 +1,81 @@
...
+  /**
+   * Creates a new PodSpec with the scaling specifications
+   * changed to fixed, and with the number of instances defined
+   * from the second argument.
+   *
+   * @param {PodSpec} spec - The spec to change
+   * @param {number} instancesCount - The new number of instances
+   * @returns {PodSpec} - Returns a new PodSpec instance
+   */
+  setPodFixedScaling(spec, instancesCount) {
```
orlandohohmeier: Can we rename this to `setPodInstances`, so that all those methods follow the same naming scheme?  #nit

src/js/utils/ServiceSpecUtil.js
```diff
@@ -0,0 +1,81 @@
...
+
+  /**
+   * High-level action for picking the correct scaling function accoring
+   * to spec type.
+   *
+   * @param {PodSpec|ApplicationSpec|FrameworkSpec} spec - The spec to change
+   * @param {number} instancesCount - The new number of instances
+   * @returns {PodSpec|ApplicationSpec|FrameworkSpec} - Returns a new instance
+   */
+  scaleServiceSpec(spec, instancesCount) {
```
orlandohohmeier: Can we rename this to `setServiceInstances`?  #nit

src/js/utils/ServiceSpecUtil.js
```diff
@@ -0,0 +1,81 @@
...
+
+  /**
+   * Create a new FrameworkSpec with the instances property
+   * set to the instancesCount specified.
+   *
+   * @param {FrameworkSpec} spec - The spec to change
+   * @param {number} instancesCount - The new number of instances
+   * @returns {FrameworkSpec} - Returns a new FrameworkSpec instance
+   */
+  setFrameworkInstances(spec, instancesCount) {
```
orlandohohmeier: Wonder if we really need a `setFrameworkInstances` util as `Framework` only extends `Application`.

src/js/utils/ServiceSpecUtil.js
wavesoft: I would vote against, because `fixed` scaling is only one type of possible scaling options. In the future we might see something else.

src/js/utils/ServiceSpecUtil.js
wavesoft: Sure, but since they are semantically different I wanted to separate them, just in case the way a framework scales is different. (plus we already have 2 different specs).

src/js/utils/ServiceSpecUtil.js
wavesoft: I was thinking of using different naming for higher-level functions. But `setServiceInstances` seems a better idea.

wavesoft: Fixed

src/js/utils/ServiceSpecUtil.js
orlandohohmeier: Hm... I'd assume for now that no matter what scaling strategy we're using we will always have a somehow defined number of instances. I'd name this one `setPodInstances` and introduce one to change the scaling type (e.g. `setPodScalingType`) when needed.  What do you think?

wavesoft: Ok, fixed 👍

src/js/utils/__tests__/ServiceSpecUtil-test.js
```diff
@@ -0,0 +1,144 @@
...
+
+describe('ServiceSpecUtil', function () {
+
+  describe('Pods', function () {
+
+    describe('#setPodFixedScaling', function () {
+
+      it('should properly create missing sections', function () {
+        var spec = new PodSpec({ });
+        var newSpec = ServiceSpecUtil.setPodFixedScaling(spec, 10);
```
orlandohohmeier: Seems like you missed to rename the methods here as well.

orlandohohmeier:
I'm sorry that I'm so rigid about the method names and util structure, but I think it's important to get this right.

wavesoft:
No prob, makes sense. Fixed 👍

Poltergeist:
@wavesoft please rebase this branch.

wavesoft:
Rebased ⚓

orlandohohmeier:
LGTM. Great work! 🎆 🚢 🇮🇹

PR #1112, opened 2016-09-20
[Pods][Testing] Normalising capitalisation of 'status' property

wavesoft:
In the spec, status string was supposed to be capitalised (https://github.com/mesosphere/marathon/blob/feature/pods/docs/docs/rest-api/public/api/v2/types/podStatus.raml#L100), but on the API it looks like it's not:

image

So, as a precaution I am normalising the status everywhere.

### DIFF
--- a/src/js/structs/Pod.js
+++ b/src/js/structs/Pod.js
@@ -47,7 +47,7 @@ module.exports = class Pod extends Service {
    * @override
    */
   getHealth() {
-    switch (this.get('status')) {
+    switch (this.getStatusString()) {
       // DEGRADED - The number of STABLE pod instances is less than the number
       // of desired instances.
       case 'DEGRADED':
@@ -136,6 +136,10 @@ module.exports = class Pod extends Service {
    */
   getSpec() {
     return this._spec;
+  }
+
+  getStatusString() {
+    return (this.get('status') || '').toUpperCase();
   }
 
   /**
--- a/src/js/structs/PodContainer.js
+++ b/src/js/structs/PodContainer.js
@@ -4,7 +4,7 @@ import StringUtil from '../utils/StringUtil';
 
 module.exports = class PodContainer extends Item {
   getContainerStatus() {
-    switch (this.get('status')) {
+    switch (this.getStatusString()) {
       case 'RUNNING':
         if (this.hasHealthChecks()) {
           if (this.isHealthy()) {
@@ -30,7 +30,7 @@ module.exports = class PodContainer extends Item {
 
       default:
         return Object.assign(Object.create(PodContainerStatus.NA), {
-          displayName: StringUtil.capitalize(this.get('status').toLowerCase())
+          displayName: StringUtil.capitalize(this.getStatusString().toLowerCase())
         });
     }
   }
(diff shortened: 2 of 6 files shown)


src/js/structs/Pod.js
```diff
@@ -138,6 +138,10 @@ module.exports = class Pod extends Service {
     return this._spec;
   }
 
+  getStatusString() {
+    return (this.get('status') || '').toUpperCase();
```
nLight: Is there a way to DRY this duplication?

src/js/structs/PodContainer.js
```diff
@@ -30,7 +30,7 @@ module.exports = class PodContainer extends Item {
 
       default:
         return Object.assign(Object.create(PodContainerStatus.NA), {
-          displayName: StringUtil.capitalize(this.get('status').toLowerCase())
+          displayName: StringUtil.capitalize(this.getStatusString().toLowerCase())
```
nLight: Here we make it uppercase and then lowercase again. Maybe it would be better to omit `toUpperCase` in `getStatusString`

wavesoft: I aggree, however the `getStatusString` also returns a string even if the `status` is missing. If we use `get('status')` and `status` is not defined, this is going to raise an exception.

src/js/structs/Pod.js
wavesoft: I don't really think it makes sense to create a utility function for this. The logic is pretty simple (just upper casing and checking for default value).

/cc @orlandohohmeier - what do you think?

orlandohohmeier: I agree that we don't need a utility function for this one, but I wonder why we upper case the status string?

wavesoft: We could lower-case if you prefer. I just think that upper-case strings seem more like a constant in the code and can be easily spotted.

src/js/structs/PodContainer.js
nLight: I mean we could keep `getStatusString` just move `toUpperCase` from it to the place where we actually use it

src/js/structs/Pod.js
orlandohohmeier: 1. Do we really expect to get different status strings form the backend?
2. Let's add a proper constant (constant file) if we need one

wavesoft: There are cases we know, such as this:
https://github.com/mesosphere/marathon/blob/feature/pods/docs/docs/rest-api/public/api/v2/types/podStatus.raml#L100

But also cases were we don't know what it could be, like this:
https://github.com/mesosphere/marathon/blob/feature/pods/docs/docs/rest-api/public/api/v2/types/podStatus.raml#L36

orlandohohmeier: Don't we need to know the possible values for that one as well?

wavesoft:
From the discussion I see that we should not do this feature.
Let's wait until @jdef has defined the last unknown part (https://github.com/mesosphere/marathon/blob/feature/pods/docs/docs/rest-api/public/api/v2/types/podStatus.raml#L36) and we should assume that the specs are followed correctly. 

Closing this PR, have a look on #1113

PR #1113, opened 2016-09-20
[Pods] Moving raw strings to constants files

wavesoft:
Moved all the state strings from the `Pod`, `PodInstance` and `PodContainer`  to a constants file.

### DIFF
--- a/src/js/constants/PodContainerState.js
+++ b/src/js/constants/PodContainerState.js
@@ -6,9 +6,14 @@
  *       the current values are assumed to be correct (DCOS-9852).
  */
 module.exports = {
+  STAGING: 'TASK_STAGING',
+  STARTING: 'TASK_STARTING',
+  STARTED: 'TASK_STARTED',
   RUNNING: 'TASK_RUNNING',
-  ERROR: 'TASK_ERROR',
-  FAILED: 'TASK_FAILED',
+  KILLING: 'TASK_KILLING',
   FINISHED: 'TASK_FINISHED',
-  KILLED: 'TASK_KILLED'
+  KILLED: 'TASK_KILLED',
+  FAILED: 'TASK_FAILED',
+  LOST: 'TASK_LOST',
+  ERROR: 'TASK_ERROR'
 };
--- a/src/js/constants/PodContainerState.js
+++ b/src/js/constants/PodContainerState.js
@@ -1,0 +1,14 @@
+//
+// The possible values of ContainerState enum for Pods, according to:
+// https://github.com/mesosphere/marathon/blob/feature/pods/docs/docs/rest-api/public/api/v2/types/podStatus.raml#L36
+//
+// TODO: Note that the SPEC currently states that this is not yet well-defiend,
+//       the current values are assumed to be correct:
+//
+module.exports = {
+  RUNNING: 'RUNNING',
+  ERROR: 'ERROR',
+  FAILED: 'FAILED',
+  FINISHED: 'FINISHED',
+  KILLED: 'KILLED'
+};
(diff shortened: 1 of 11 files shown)


wavesoft:
We checked the code with @jdef and we realized that all the responses are normalized to lower-case. Changed the constants to reflect this.

wavesoft:
@jdef pointed out that the containers are using the upper-case `TASK_XXX` constants from here: https://github.com/apache/mesos/blob/master/include/mesos/mesos.proto#L1497 - Updated the constants to reflect this change

src/js/constants/PodContainerState.js
```diff
@@ -0,0 +1,14 @@
+//
+// The possible values of ContainerState enum for Pods, according to:
+// https://github.com/mesosphere/marathon/blob/feature/pods/docs/docs/rest-api/public/api/v2/types/podStatus.raml#L36
+//
+// TODO: Note that the SPEC currently states that this is not yet well-defiend,
+//       the current values are assumed to be correct (DCOS-9852).
+//
+module.exports = {
+  RUNNING: 'TASK_RUNNING',
```
nLight: So task statuses are uppercase while everything else is lowercase?

src/js/constants/PodContainerState.js
```diff
@@ -0,0 +1,14 @@
+//
```
orlandohohmeier: Please use  `/** ... */` for multi-line comments. #nit

src/js/structs/PodContainer.js
```diff
@@ -16,21 +17,27 @@ module.exports = class PodContainer extends Item {
...
-      case 'KILLED':
+      case PodContainerState.KILLED:
         return PodContainerStatus.KILLED;
 
       default:
+        let statusString = this.get('status') || '';
+        if (statusString.startsWith('TASK_')) {
+          statusString = statusString.substr(5);
+        }
+        statusString = statusString.replace(/_/g, ' ');
```
orlandohohmeier: I'm not to sure about this "default"/fallback behavior – seems a little to complex. Let's use something like "NA" in case we receive an unknown status.

src/js/constants/PodContainerState.js
wavesoft: yes. Not only uppercase, plus prefixed with `TASK_` :(

src/js/structs/PodContainer.js
wavesoft: There are other statuses currently not handled (check this: https://github.com/apache/mesos/blob/master/include/mesos/mesos.proto#L1497 ) - I don't know which one of them can appear on the pods context.

src/js/constants/PodContainerState.js
wavesoft: Done. (I usually use `/** .. */` for jsdoc comments, this particular case was just notes)

nLight:
🚢 !

src/js/structs/PodContainer.js
orlandohohmeier: Let's only support the known set and add stuff when needed instead of converting these strings, as that could result in unwanted behavior.

src/js/constants/PodInstanceState.js
```diff
@@ -0,0 +1,11 @@
+//
```
orlandohohmeier: Let's use `/** ... */` here as well

src/js/constants/PodState.js
```diff
@@ -0,0 +1,9 @@
+//
```
orlandohohmeier: Same here

orlandohohmeier:
🚢 🚢 🚢

PR #1116, opened 2016-09-20
[CNVS Upgrade] Fix Nodes routes

jfurrow:
This PR fixes the node routes.

I moved the component previously called `NodesPage` to `NodesOverview`, and created a new component called `NodesPage` which is a simple page wrapper for all of the nodes routes.

Because all of the node children routes render within the new `NodesPage`, I removed all of the `Page` component wrappers that were previously used by the child routes.


philipnrmn:
Working great :) Thanks for fixing this so quickly John.

src/js/pages/NodesOverview.js
```diff
@@ -0,0 +1,320 @@
...
+        <Link className={gridClassSet} onClick={resetFilter} to="nodes-grid">Grid</Link>
+      </div>
+    );
+  },
+
+  getHostsPageContent() {
+    let {byServiceFilter, healthFilter, searchString, selectedResource} = this.state;
+    var data = this.internalStorage_get();
+    let nodes = data.nodes || [];
+    let nodesList = nodes.slice(0, NODES_DISPLAY_LIMIT);
```
MatApple: Curious about why we have this?

jfurrow: This was here before, I just renamed the file actually.

MatApple:
LGTM! As discussed with @philipnrmn - there will likely be plenty of alterations here with the refactor. 🚢  it

src/js/pages/NodesOverview.js
MatApple: Cool - maybe it was implemented prior to our smart table.

PR #1125, opened 2016-09-21
[Nodes Page] Purify node detail view

philipnrmn:
This PR is simply adds the pure render mixin to the NodeDetailTab and its only child, DescriptionList.

### DIFF
--- a/src/js/components/DescriptionList.js
+++ b/src/js/components/DescriptionList.js
@@ -1,6 +1,12 @@
+import PureRenderMixin from 'react-addons-pure-render-mixin';
 import React from 'react';
 
 class DescriptionList extends React.Component {
+  constructor() {
+    super(...arguments);
+    this.shouldComponentUpdate = PureRenderMixin.shouldComponentUpdate;
+  }
+
   getHeadline() {
     let {headline, headlineClassName} = this.props;
     if (!headline) {
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -5762,6 +5762,11 @@
       "from": "react-addons-css-transition-group@0.14.7",
       "resolved": "https://registry.npmjs.org/react-addons-css-transition-group/-/react-addons-css-transition-group-0.14.7.tgz"
     },
+    "react-addons-pure-render-mixin": {
+      "version": "0.14.7",
+      "from": "react-addons-pure-render-mixin@0.14.7",
+      "resolved": "https://registry.npmjs.org/react-addons-pure-render-mixin/-/react-addons-pure-render-mixin-0.14.7.tgz"
+    },
     "react-addons-test-utils": {
       "version": "0.14.7",
       "from": "react-addons-test-utils@0.14.7",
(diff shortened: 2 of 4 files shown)


src/js/components/DescriptionList.js
```diff
@@ -1,6 +1,12 @@
+import PureRenderMixin from 'react-addons-pure-render-mixin';
 import React from 'react';
 
 class DescriptionList extends React.Component {
+  constructor() {
+    super(...arguments);
+    this.shouldComponentUpdate = PureRenderMixin.shouldComponentUpdate;
```
mlunoe: Shouldn't the `PureRenderMixin.shouldComponentUpdate` be bound?
this.shouldComponentUpdate = PureRenderMixin.shouldComponentUpdate.bind(this);

philipnrmn: I don't think so. `DescriptionList.shouldComponentUpdate`as defined there is properly scoped.

mlunoe: I was just looking here: https://facebook.github.io/react/docs/pure-render-mixin.html
Isn't it supposed to have access to `this.props` and `this.state`?

philipnrmn: Yep, absolutely - and it does have access to `this.props` and `this.state` when it's called. I'm just trying to figure out why that example uses binding - do you know of any circumstances where it's invoked in the wrong scope?

mlunoe: Nope, just looked at that resource and figured we should follow it.
Looking into it a bit more, you are right. It shouldn't matter: https://github.com/facebook/react/blob/master/src/addons/ReactComponentWithPureRenderMixin.js

[See Babel example here](https://babeljs.io/repl/#?babili=false&evaluate=true&lineWrap=false&presets=es2015%2Ces2016%2Ces2017%2Creact%2Cstage-2&experimental=false&loose=false&spec=true&playground=false&code=let%20mixin%20%3D%20function%20()%20%7B%0A%20%20console.log(this.constructor)%3B%0A%7D%0A%0Aclass%20Hello%20%7B%0A%20%20constructor(prop)%20%7B%0A%20%20%20%20this.mixin%20%3D%20mixin%3B%0A%20%20%7D%0A%7D%0A%0Aclass%20There%20%7B%0A%20%20constructor(prop)%20%7B%0A%20%20%20%20this.mixin%20%3D%20mixin%3B%0A%20%20%7D%0A%7D%0A%0Alet%20hello%20%3D%20new%20Hello()%3B%0Ahello.mixin()%3B%0Alet%20there%20%3D%20new%20There()%3B%0Athere.mixin()%3B)

mlunoe:
Looks like you have some linting errors:

philipnrmn:
Thanks for the review! Yep the lint errors are from the base branch, not related to this PR :)

jfurrow:
My apologies for some of the lint warnings, I was moving too quickly and some of my PRs introduced two that I'm aware of, which have been fixed here: https://github.com/dcos/dcos-ui/pull/1123. Looks like more have crept in; I'll address these tomorrow morning!

PR #1107, opened 2016-09-19
Making Hooks a factory and adding isolated instance

mlunoe:
This PR isolates Hooks, so it now has to be used as a factory, and there can be multiple instances of hooks with isolated scope.

This PR will not introduce new hooks, but rather just make the existing one isolated and therefore have no change on the application.

This is a part of the preparation for building the Mount component and Mount service.

PR #1130, opened 2016-09-22
[Mount Component/Service 1 of 2] Feature: Add Mount component and service

mlunoe:
---

~~⚠️ Dependent on this PR: https://github.com/dcos/dcos-ui/pull/1107~~ Merged!

---

This PR introduces the Mount component and service.

Proposed file structure:

```
foundation-ui/
- mount/
- - index.js
- - MountPoint.js
- - MountService.js
- auth/
- - index.js
- - AuthPoint.js
- - AuthService.js
- index.js
```

To try it out you can add this to e.g. the DashboardPage (or see the included tests):

TODO for @orlandohohmeier:
- Tests need to be changed to follow the new structure
- We should probably throw a warning in dev mode if you register a component under same priority?
- There is a bug with the cloning of children: 
- It seems to re-render a lot. Maybe try to look into that and maybe implement `shouldComponentUpdate`
- fix conflict


mlunoe:
Btw. If you have a better idea for the naming and structuring of the files, I would love to hear from you!

foundation-ui/factories/ServiceFactory.js
```diff
@@ -0,0 +1,23 @@
+/*
+  Service Factory
+ */
+function ServiceFactory(hooks) {
+  return {
+    on(mountId /* , callback */) {
```
MatApple: If this is going to be used as the core for all Services, it might make more sense to rename the first argument to something other than `mountId`. Also I think we should explicitly define the arguments it can take so it's easier for people to trace down the api. Let's add in `, callback, priority)` explicitly.

foundation-ui/factories/ServiceFactory.js
```diff
@@ -0,0 +1,23 @@
...
+
+    removeListener(mountId /* , callback */) {
+      hooks.removeFilter(...arguments);
+      // Have Mount's run filter again so removed filter will be excluded
+      hooks.doAction(mountId);
+    }
+  };
+}
+
+module.exports = ServiceFactory;
```
MatApple: We should probably write some basic tests for these 2 methods.

foundation-ui/services/MountBundle.js
```diff
@@ -0,0 +1,147 @@
...
+
+  componentWillUnmount() {
+    hooks.removeAction(this.props.mountId, this.updateState);
+  }
+
+  updateState(props = this.props) {
+    let {children, mountId} = props;
+    let blackList = ['children', ...Object.keys(Mount.propTypes)];
+    // Filter props consumed by Mount to create childProps
+    let childProps = Object.keys(props).filter(function (key) {
```
MatApple: We don't need the extra filter here. We can just put the filter in the reduce function so we only iterate once.

foundation-ui/services/MountBundle.js
```diff
@@ -0,0 +1,147 @@
...
+
+Mount.propTypes = {
+  mountId: PropTypes.string.isRequired,
+  wrapperComponent: PropTypes.node,
+  alwaysWrap: PropTypes.bool
+};
+
+module.exports = {
+  Mount,
+  MountService: new ServiceFactory(hooks)
```
MatApple: I really like this 👍

foundation-ui/services/MountBundle.js
```diff
@@ -0,0 +1,147 @@
...
+  getChildContext() {
+    // Get context passed down from parent Mount if one exists
+    // and add to it so we can create a render tree chain of nested mountIds.
+    // This should just be an Array of string mountIds. We can use the mountIds
+    // to look up each Mount Component on the page and see debug info about it
+    // like which package inserted the component, what the priorities were for
+    // each component if multiple components were registered for a mountId etc.
+    const {mountChain = []} = this.context || {};
+
+    return {mountChain: [...mountChain, this.props.mountId]};
```
MatApple: We'll probably want to use `Symbols` I think here because we can guarantee uniqueness for each instance of mountId that might be on a page. But I think this is fine for now until we implement the debugging part. This is a great example of how we can build some powerful debugging in :)

foundation-ui/services/MountBundle.js
mlunoe: Let's not worry about iterations here.
This is more readable imo. That's why I did it like this.

foundation-ui/services/MountBundle.js
mlunoe: I'm sorry, I am not sure I understand what you mean by `Symbols`? Could you please elaborate?

foundation-ui/factories/ServiceFactory.js
mlunoe: We are already testing the functions that these call. I don't see what value it would provide to test the call of `addFilter` followed by the call to `doAction`.
Maybe you have scenario that is not covered by the other tests that I am missing?

mlunoe: Will add a simple test to make sure that the two methods are being called with the right arguments

mlunoe: Done

foundation-ui/factories/ServiceFactory.js
mlunoe: Done

foundation-ui/services/MountBundle.js
mlunoe: We talked offline about this. What was meant was a suggestion to use JavaScript Symbols, so we can differentiate between siblings with the same key.

However, we decided that this will work for now and we can give it another look once we implement the debugging tools

mlunoe:
*mesosphere-ci** Retest this please!

mlunoe:
@MatApple I covered all your comments

MatApple:
Nice work @mlunoe 💯

foundation-ui/services/MountBundle.js
orlandohohmeier: Wonder why we need to provide the mount chain in the first place? What's a benefit of a child node knowing the IDs of all it's parents? Having the current ID should be enough to debug if one of the mounts isn't working.

foundation-ui/services/MountBundle.js
```diff
@@ -0,0 +1,147 @@
...
+};
+
+Mount.defaultProps = {
+  wrapperComponent: 'div',
+  alwaysWrap: false
+};
+
+Mount.propTypes = {
+  mountId: PropTypes.string.isRequired,
+  wrapperComponent: PropTypes.node,
```
mlunoe: If you have multiple children. You cannot return an array in a React component

foundation-ui/services/MountBundle.js
```diff
@@ -0,0 +1,147 @@
...
+    // Filter props consumed by Mount to create childProps
+    let childProps = Object.keys(props).filter(function (key) {
+      return !blackList.includes(key);
+    }).reduce(function (memo, key) {
+      memo[key] = props[key];
+
+      return memo;
+    }, {});
+
+    let filteredChildren = hooks.applyFilter(
```
mlunoe: There's a whole example usage section at the top of the file. Is that not enough?

mlunoe: To you mountChain comment question @MatApple maybe you want to elaborate?

foundation-ui/services/MountBundle.js
```diff
@@ -0,0 +1,147 @@
+import React, {PropTypes} from 'react';
```
mlunoe: The problem with this structure is that we need to instantiate the `Hooks` to be used in both the service and component. To make it completely isolated we chose do to it in one file.
What I did was to then have a separate file for the Hooks instantiation (which we will need for every service/component pair) that the two files could include.

It will then look something like this:
* foundation
	* mount
		* hooks.js
		* MountService.js
		* MountComponent.js
	* auth
		* hooks.js
		* AuthService.js
		* AuthComponent.js
	* ...
I removed the `main.js` or `index.js` to not cause confusion of what you will get, but we could include it to expose the service and component.

One last thing: We don't postfix with `Component` or `Service` anywhere else, which I think would be good here to differentiate, but this looks kind of weird to me:
```js
<MountComponent id="foo">
  <h1>bar</h1>
</MountComponent>
```

However, naming it `Mount` we could make that be exposed in the `index.js`, so you could do this:
`import Mount from 'foundation-ui/Mount';`
or this:
`import Mount from 'foundation-ui/Mount/Mount';`
`import MountService from 'foundation-ui/Mount/MountService';`

foundation-ui/factories/ServiceFactory.js
```diff
@@ -0,0 +1,23 @@
+/*
+  Service Factory
+ */
+function ServiceFactory(hooks) {
```
mlunoe: It doesn't need to have 'factory' methods. This is a factory that creates services. See here for more examples:
https://medium.com/javascript-scene/javascript-factory-functions-vs-constructor-functions-vs-classes-2f22ceddf33e#.49vxo3zg0

foundation-ui/services/MountBundle.js
mlunoe: ✅ Check out the new proposed structure

foundation-ui/factories/ServiceFactory.js
mlunoe: Renamed to Service under the new structure

foundation-ui/services/MountBundle.js
```diff
@@ -0,0 +1,147 @@
...
+ *     return (
+ *       <Mount mountId="serviceDetail">
+ *         // What to show when serviceDetail doesn't mount
+ *         <EmptyPage />
+ *       </Mount>
+ *     );
+ *   }
+ * };
+ */
+class Mount extends React.Component {
```
mlunoe: ✅

foundation-ui/services/MountBundle.js
```diff
@@ -0,0 +1,147 @@
...
+    );
+  }
+}
+
+// Consume mountChain from parent Mount(s)
+Mount.contextTypes = {
+  mountChain: PropTypes.array
+};
+
+// Pass down mointChain context to children Mount(s)
```
mlunoe: ✅

foundation-ui/services/MountBundle.js
```diff
@@ -0,0 +1,147 @@
...
+  mountChain: PropTypes.array
+};
+
+Mount.defaultProps = {
+  wrapperComponent: 'div',
+  alwaysWrap: false
+};
+
+Mount.propTypes = {
+  mountId: PropTypes.string.isRequired,
```
mlunoe: ✅

foundation-ui/factories/ServiceFactory.js
```diff
@@ -0,0 +1,23 @@
+/*
```
mlunoe: ✅ Added jsDoc

mlunoe:

foundation-ui/services/MountBundle.js
orlandohohmeier: Yep, I only wondered why we need to make this configureable. Thinking about it, it actually is quite use full as this enables us to also use non-block elements as wrappers.

foundation-ui/services/MountBundle.js
MatApple: @orlandohohmeier @mlunoe a quick couple of reasons why the mountChain will be a good thing.
1) We can build in a simple dev tool that outlines Mount points on the page which would be useful for package developers. It would also be helpful to know the hierarchy of Mounts so you can ultimately trace back the ownership of code.
2) Since Mounts are the primary method of cross-package component injection, we may need to pass down additional information about the existing relay query higher up the tree so we can make smarter decisions about whether to fetch a new Relay query or rely on the existing one if it covers the fragment of the injected component.

mlunoe:

foundation-ui/services/MountBundle.js
orlandohohmeier: Looks good to me, thanks for updating! 💜

foundation-ui/services/MountBundle.js
orlandohohmeier: @MatApple We should be able to highlight mount points without the mount chain, I don't see a real need here; but I see benefits in having it, to allow us to pass down additional information. Wonder if we could achieve the same using props only, in that case I would vote to not use the context as the context API is still an [experimental feature](https://facebook.github.io/react/docs/context.html) and the data flow isn't as transparent as it is using props.

orlandohohmeier:
I finally found the time to complete the review. This is huge! Great work @mlunoe  ✨  💯 

I'm ok with merging the mount point PR as it is, but I would love to discuss the general use of hooks to implement the services. I like to have them to filter stuff but think we should implement very specific interfaces, as this enables us to properly validate input, they are easier to use as they provide a distinct interface and they allow us to have concrete sets of tests. 

/cc @dcos/frontend

mlunoe:
Had a conversation offline with @orlandohohmeier to make some changes as follows:
- Move the use of the hooks instance in the `MountPoint` to only being in the Service and instead have an API the MountPoint can use for this
- Not having a ServiceFactory, but just fletch out the `MointService` (and `AuthService`) as is with possible duplication to make it easier to understand
- Validate and test input in the `MointService` to catch possible misusages.

I am going to implement those changes now

mlunoe:
Ok, @MatApple and @orlandohohmeier: This has been updated to conform to the description above

orlandohohmeier:
@mlunoe @rcorral @wavesoft  I've addressed all of the open todos and adjusted the mount service and component as discussed. 

Changes/Updates:

* Simplified the mount service by removing the array/map optimization for now as it made it harder to refactor the service without providing a huge performance gain. I'll reintroduce the optimization once we've landed the service.
* Adjusted the mount service as well as the mount point to use flux like events and methods to properly pass data and handle changes.  
* Fixed issues with the component sorting as components should maintain order of addition if they aren't prioritized.
* Changed the service and mount point to use `React.Component` classes as well as stateless functional components instead of dealing with react elements. 
* Updated the "documentation" and examples to reflect the latest changes. 
* Added test that should properly cover all of the features.

foundation-ui/mount/MountPoint.js
```diff
@@ -0,0 +1,109 @@
...
+
+      return props;
+    }, {});
+
+    // Limit the number of components as configured and create elements
+    const elements = components.slice(0, limit).map((Component, index) => {
+      return (<Component {...filteredProps} key={index} />);
+    });
+
+    // Don't render prop.children if elements is defined
```
wavesoft: >// Don't render prop.children if elements is **not** defined?

foundation-ui/mount/MountService.js
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    // Add component descriptor and sort components list by priority and index
+    components.push({component, priority, type, index: components.length});
+    components.sort((a, b) => {
+      if (a.priority !== b.priority) {
+        return b.priority - a.priority;
+      }
+
+      return a.index - b.index;
```
wavesoft: I see a problem here:
* Register  a couple of components with varying priorities, call `unregisterComponent` -> Chance is you did not remove the item with the highest index. Calling `registerComponent` again will have unexpected behaviour since more than 1 items have the same `index`.

Suggestions:

1. You should also re-assign new `index` properties in the `unregisterComponent` function.
2. Use an ever-incrementing value for index, instead of `components.length`

foundation-ui/mount/MountPoint.js
```diff
@@ -0,0 +1,109 @@
...
+    this.setState({components: MountService.findComponentsWithType(type)});
+  }
+
+  render() {
+    const {alwaysWrap, children, limit, wrapper} = this.props;
+    const {components} = this.state;
+
+    // Filter consumed props to only pass on a "clean" set of props
+    const filteredProps = Object.keys(this.props).reduce((props, key) => {
+      if (!Object.prototype.hasOwnProperty.call(MountPoint.propTypes, key)) {
```
wavesoft: Cool! Only a quick question. Is the `MountPoint.propTypes` maintained at a release build? Or are they optimised-out?

foundation-ui/mount/MountService.js
wavesoft: Perhaps add a test for this case too?

foundation-ui/mount/MountPoint.js
orlandohohmeier: I actually don't know... I'll check our build and adjust this based on my findings.

foundation-ui/utils/__tests__/ReactUtil-test.js
```diff
@@ -0,0 +1,43 @@
...
+        [<span key="1">test</span>, <span key="2">test</span>]
+    );
+
+    expect(TestUtils.isElementOfType(result, 'div')).toEqual(true);
+  });
+
+  it('should not wrap if not necessary', function () {
+    const elements = ReactUtil.wrapElements(<span>test</span>);
+
+    expect(TestUtils.isElementOfType(elements, 'span')).toEqual(true);
```
orlandohohmeier: @wavesoft and I discussed this one offline and agreed that we don't need an extra test case, but might need to adjust the tag names to avoid confusions.

foundation-ui/mount/MountService.js
orlandohohmeier: 🙀  good catch! I'll properly go with the ever-incrementing value...

foundation-ui/mount/MountPoint.js
```diff
@@ -0,0 +1,109 @@
...
+      components: MountService.findComponentsWithType(this.props.type)
+    };
+
+    METHODS_TO_BIND.forEach((method) => {
+      this[method] = this[method].bind(this);
+    });
+  }
+
+  componentWillMount() {
+
```
mlunoe: #nit let's remove this newline

foundation-ui/mount/MountPoint.js
orlandohohmeier: They're included but I'm going to introduce an additional list so that we don't fail once we add an additional optimization step which removes them.

orlandohohmeier: Discussed this with @wavesoft and we're keeping the current solution for now as any extra list is highly error prone and like to be outdated.

orlandohohmeier:
@mlunoe @wavesoft addressed your comments

foundation-ui/utils/ReactUtil.js
```diff
@@ -0,0 +1,27 @@
...
+  /**
+   * Wrap React elements
+   *
+   * @param {Array.<ReactElement>|ReactElement} elements
+   * @param  {function|String} wrapper component
+   * @param {boolean} [alwaysWrap]
+   *
+   * @returns {ReactElement} wrapped react elements
+   */
+  wrapElements(elements, wrapper = 'div', alwaysWrap = false) {
```
rcorral: I think we'll need to be able to set attributes on the element itself, for example, for a <tr>, we could want to add classes to it.

foundation-ui/mount/MountService.js
```diff
@@ -0,0 +1,94 @@
+import {EventEmitter} from 'events';
+import {CHANGE} from './MountEvent';
+
+/**
+ * List of component descriptors
+ *
+ * @type {Array.<{component:function, priority:Number, type:String}>}
+ */
+const components = [];
```
orlandohohmeier: I removed that data structure as discussed in https://github.com/dcos/dcos-ui/pull/1130#issuecomment-259173317 to simplify the current implementation as it made the refactoring harder without providing a huge performance gain. I'll reintroduce the optimization once we've landed the packages and see need.

foundation-ui/mount/MountService.js
```diff
@@ -0,0 +1,94 @@
...
+
+    // Add component descriptor and sort components list by priority and rank
+    components.push({component, priority, type, rank: rank++});
+    components.sort((a, b) => {
+      if (a.priority !== b.priority) {
+        return b.priority - a.priority;
+      }
+
+      return a.rank - b.rank;
+    });
```
orlandohohmeier: The components are sorted in a way that the one with the highest priority comes first and the rest maintains it's position. This enables us to easily limit the components to the ones with the highest priority (e.g. only using the first two components). The latter ensures that we don't have any funny effects (element shuffle) caused by the sorting algorithm.  

### Example

In case one would add the following set of components in the respective order, the result would return a nicely ordered list.

#### Register Component
```
1. B {priority: 0}
2. C {priority: 0}
3. A {priority: 2}
4. D {priority: 0}
```
#### Find Components

```
[A,B,C,D]
```

*The following test verifies that the service returns the components in proper order:*
https://github.com/dcos/dcos-ui/pull/1130/files#diff-ab5a90aae6458cd23f55dae4c16b75b0R135

I'm happy to adjust the sorting if you think that the sorting needs to work differently.

foundation-ui/mount/Mount.js
```diff
@@ -0,0 +1,108 @@
...
+      components: MountService.findComponentsWithType(this.props.type)
+    };
+
+    METHODS_TO_BIND.forEach((method) => {
+      this[method] = this[method].bind(this);
+    });
+  }
+
+  componentWillMount() {
+    MountService.addListener(CHANGE, this.handleMountServiceChange);
```
orlandohohmeier: Dang, I actually had a look to make sure that I'm using the correct naming scheme but I must have the one file with a bad example... Going to adjust the naming.

foundation-ui/utils/ReactUtil.js
orlandohohmeier: Good point! There are different ways to solve this; either we change the interface to something like the following or we keep it as it is and create custom wrapper components as illustrated below.

### Wapper Props 

Add the ability to pass on props to the wrapper

```
wrapElements(elements, wrapper = 'div', props = null, alwaysWrap = false)
```

### Custom Wapper 

Possible way using a custom wrapper to define a custom class name. 

```
wrapElements(elements, ({children}) => {<div className="custom-wrapper">{children}</div>});
```

What would you prefer?

foundation-ui/mount/MountService.js
rcorral: Agreed, performance is probably marginal, it's more that it made more sense in my head to do it this other way. Cool with deferring.

foundation-ui/mount/MountService.js
rcorral: This makes sense. I was getting thrown off with what the most important priority was, as I was thinking the lowest would be most important, but that's because I'm comparing it to the priorities in our Hooks.
I'm cool with leaving it as it is, but it's worth considering that it's different between the two services and it could be nice to keep things consistent. Same with the Routing.

foundation-ui/utils/ReactUtil.js
rcorral: I'm indifferent, though I'm leaning more towards the wrapper props.

mlunoe: I would agree, that goes along the lines of only allowing what we need

orlandohohmeier: Ok, I'm not going to introduce the wrapper props for now as the latter works today, and we can extend the util later if we see need. Hope you guys are ok with that.

foundation-ui/mount/Mount.js
```diff
@@ -0,0 +1,108 @@
...
+
+    this.setState({components: MountService.findComponentsWithType(type)});
+  }
+
+  render() {
+    const {alwaysWrap, children, limit, wrapper} = this.props;
+    const {components} = this.state;
+
+    // Filter consumed props to only pass on a "clean" set of props
+    const filteredProps = Object.keys(this.props).reduce((props, key) => {
```
MatApple: Won't this prevent a Mount from passing on additional context via props?

I'd imagine we may want to do something like (bad example for dcos-ui but hopefully it illustrates my point):
```
<Mount type="service-table-row" serviceId="/foo">
 <tr>{ services[serviceId] }</tr>
</Mount>
```
So another component that gets injected can consume `serviceId`.

foundation-ui/mount/MountService.js
```diff
@@ -0,0 +1,94 @@
...
+   * @param  {function} component
+   * @param  {String} type
+   */
+  unregisterComponent(component, type) {
+    let i = components.length;
+    while (--i >= 0) {
+      if (components[i].component === component &&
+          components[i].type === type) {
+        components.splice(i, 1);
+        break;
```
MatApple: Since we can register the same component/type multiple times, we should either splice out all components with component/type (so we shouldn't break here), or, when registering we should make sure you can't register the same component/type combination more than once.

foundation-ui/mount/Mount.js
orlandohohmeier: It's only filtering the props that are used by the `Mount` component to pass on a "clean" set of props.

https://github.com/dcos/dcos-ui/pull/1130/files#diff-50755952fa7e5098afe99ed3fe6e5dcfR142

foundation-ui/mount/MountService.js
orlandohohmeier: I'm going to ensure that every component/type combination is only registered once.

orlandohohmeier:
ℹ️  I'm going to align service with the singleton pattern used in the navigation and routing service as it's a good first step towards a better pattern.

orlandohohmeier:
@MatApple @wavesoft can haz some labels pleaz?  🏷 🙇

foundation-ui/mount/Mount.js
```diff
@@ -0,0 +1,108 @@
+import React, {PropTypes} from 'react';
+
+import {MountService} from './index';
+import {CHANGE} from './MountEvent';
```
nLight: I named those `EventTypes` in both routing and navigation services, just saying :)

foundation-ui/mount/MountService.js
```diff
@@ -0,0 +1,123 @@
...
+   * @param {String} type - String which is used to match components with the
+   * respective mounts
+   * @param {Number} [priority] - Number which is used to sort the components
+   */
+  registerComponent(component, type, priority = 0) {
+    let {components, instance, rank} = this;
+
+    if (typeof component !== 'function') {
+      if (global.__DEV__) {
+        throw new TypeError('Provided component must be a ' +
```
nLight: ES6 template strings are multiline

foundation-ui/mount/MountService.js
```diff
@@ -0,0 +1,123 @@
...
+      }
+
+      return;
+    }
+
+    if (components.find((descriptor) =>
+        descriptor.component === component &&
+        descriptor.type === type)) {
+      if (global.__DEV__) {
+        throw new Error('Provided component/type combination ' +
```
nLight: Let's use template strings

foundation-ui/mount/MountService.js
```diff
@@ -0,0 +1,123 @@
...
+   * Register component with type and optional priority
+   *
+   * @this MountService~Context
+   * @param {function} component - React component which will be
+   * mounted based on the provided type
+   * @param {String} type - String which is used to match components with the
+   * respective mounts
+   * @param {Number} [priority] - Number which is used to sort the components
+   */
+  registerComponent(component, type, priority = 0) {
```
nLight: I like that we're going to have homogeneous services, same code style, same public API style 👍

foundation-ui/mount/__tests__/MountService-test.js
```diff
@@ -0,0 +1,159 @@
...
+    };
+
+    beforeEach(function () {
+      this.instance.registerComponent(SecondComponent, 'type', 0);
+      this.instance.registerComponent(ThirdComponent, 'type', 0);
+      this.instance.registerComponent(FirstComponent, 'type', 2);
+      this.instance.registerComponent(FourthComponent, 'type', 0);
+    });
+
+    it('should return list of matching components in proper order',
```
nLight: We may want to be more specific and name an array `array`

foundation-ui/mount/__tests__/MountService-test.js
```diff
@@ -0,0 +1,159 @@
...
+
+      expect(this.instance.findComponentsWithType('type'))
+          .toEqual([ReactComponent, FunctionalComponent]);
+    });
+
+    it('should throw if an object instead of a component was provided',
+        function () {
+          expect(() => {
+            this.instance.registerComponent({}, 'type', 0);
+          }).toThrow();
```
nLight: I'd provide an error message to all `.toThrow` expectations

foundation-ui/mount/MountService.js
orlandohohmeier: Why would I want that? It's a simple error message nothing more.

foundation-ui/mount/__tests__/MountService-test.js
orlandohohmeier: ✅  Adjusted the test to verify that the correct errors are thrown

foundation-ui/mount/__tests__/MountService-test.js
orlandohohmeier: ✅  Changed the test description to explicitly callout the type

foundation-ui/mount/Mount.js
orlandohohmeier: Do you want me to change the name?

orlandohohmeier:

orlandohohmeier:

orlandohohmeier:

orlandohohmeier:
@wavesoft @nLight Any chance we can merge this one, once the test pass?

nLight:
Right... since I need this to do secrets I'll be the one to review/merge it :)

PR #1139, opened 2016-09-26
DCOS-9955: Sum resources in the services table

Poltergeist:
Sum up the resources in the service table to represent the actual resources needed and not the once configured for an application.

image


wavesoft:
Nice, LGTM 💯

PR #1141, opened 2016-09-26
[Pods] DCOS-9893: Align pod instances table cells at the top

nLight:
This PR adds `vertical-align: top` to the pod-instances-table cells. I decided to go with having `td` in `.less` over adding a default class like `columns-select` to `ResourceTableUtil` because `ResourceTableUtil` is being used like everywhere.

https://mesosphere.atlassian.net/browse/DCOS-9893

*Before* checkbox and version are vertically aligned in the middle:
screen shot 2016-09-26 at 14 41 49

*After* checkbox and version are vertically aligned at the top:
screen shot 2016-09-26 at 14 41 34

### DIFF
--- a/src/styles/components/pod-instances-table.less
+++ b/src/styles/components/pod-instances-table.less
@@ -1,0 +1,3 @@
+.pod-instances-table td {
+  vertical-align: top;
+}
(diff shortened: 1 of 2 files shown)


wavesoft:
Nice! 💯 ✨

orlandohohmeier:
Wonder if we should `vertical-align: top` to all expandable tables... /cc @jfurrow

leemunroe:
@nLight ✨

jfurrow:
@orlandohohmeier that makes sense!

PR #1143, opened 2016-09-26
[Pods] DCOS-9892: Addressing visual issues on PodInstancesTable

wavesoft:
Addressing Lee's comments on the pod instance table.


wavesoft:

src/js/stores/MesosStateStore.js
```diff
@@ -157,6 +157,19 @@ class MesosStateStore extends GetSetBaseStore {
     return null;
   }
 
+  getNodeFromHostname(hostname) {
+    let nodes = this.get('lastMesosState').slaves;
+    console.log('Looking node with hostname:', hostname, 'in:', nodes);
```
mlunoe: I think this slipped in here

src/js/components/PodInstancesTable.js
```diff
@@ -37,6 +40,24 @@ class PodInstancesTable extends React.Component {
     });
   }
 
+  componentWillMount() {
```
mlunoe: Normally we use the StoreMixin instead of having these three methods, like so: https://github.com/dcos/dcos-ui/blob/master/src/js/components/ServiceDetailTaskTab.js#L15

But, seeing as we need to get this in, and it is bound change soon anyways, let's not worry about it for now

src/js/stores/MesosStateStore.js
mlunoe: Removed

PR #1149, opened 2016-09-27
[Mount Component/Service 2 of 2] Feature: Add Authorization component and service

mlunoe:
---

⚠️ Depending on this PR: https://github.com/dcos/dcos-ui/pull/1130

---

You can try it out in the Dashboard like so:

``` js
import {Authorization, Authorize} from 'foundation-ui/services/AuthorizeBundle';

  componentWillMount() {
    ...
    this.permissions.forEach((permission) => {
      Authorization.on(permission, this.checkPermission.bind(this, permission));
    });
  },

  permissions: [
    'dcos:ui:sidebar-tabs',
    'dcos:services:marathon:full'
  ],

  checkPermission(permission) {
    // some logic like looking at a cache of user permissions and determining
    // if they should see this component.
    let result = this.permissions.includes(permission);
    // hack to remove permissions after first call
    this.permissions = [];
    return result;
  },

  render() {
    ...
    return (
      <Page title="Dashboard">
        <Authorize authId="dcos:ui:sidebar-tabs" replacementComponent={<h1 className="inverse">No, no, no! Unauthorized!</h1>}>
          <h1 className="inverse">Authorized!</h1>
        </Authorize>
        ...
      </Page>
    );
}
```


mlunoe:

foundation-ui/services/AuthorizeBundle.js
```diff
@@ -0,0 +1,111 @@
+import React, {PropTypes} from 'react';
```
orlandohohmeier: Let's reorganize the file structure as described in https://github.com/dcos/dcos-ui/pull/1130#discussion_r80878513

foundation-ui/services/AuthorizeBundle.js
```diff
@@ -0,0 +1,111 @@
...
+import Hooks from '../services/Hooks';
+import ServiceFactory from '../factories/ServiceFactory';
+
+const hooks = new Hooks();
+const METHODS_TO_BIND = [
+  'updateState'
+];
+
+/*
+ * Example usage:
```
orlandohohmeier: Let's make this an JSDoc comment, with proper example notation.

foundation-ui/services/AuthorizeBundle.js
```diff
@@ -0,0 +1,111 @@
...
+      this[method] = this[method].bind(this);
+    });
+  }
+
+  componentWillMount() {
+    this.updateState();
+    // Register for changes in Authorize specific to this authId.
+    // Whenever Authorize has a change it will end up invoking updateState
+    // so we can get the newly filtered children.
+    hooks.addAction(this.props.authId, this.updateState);
```
orlandohohmeier: Why are we using hooks to register for an event?

foundation-ui/services/AuthorizeBundle.js
```diff
@@ -0,0 +1,111 @@
+import React, {PropTypes} from 'react';
+
+import Hooks from '../services/Hooks';
+import ServiceFactory from '../factories/ServiceFactory';
+
+const hooks = new Hooks();
```
orlandohohmeier: Why do we need to have an "private" instance of hooks?

foundation-ui/services/AuthorizeBundle.js
```diff
@@ -0,0 +1,111 @@
...
+};
+
+Authorize.propTypes = {
+  authId: PropTypes.string.isRequired,
+  defaultValue: PropTypes.bool,
+  replacementComponent: PropTypes.node
+};
+
+module.exports = {
+  Authorization: new ServiceFactory(hooks),
```
orlandohohmeier: Why don't we have a dedicated class/object dealing with authorization? All the foundation services should have very well defined, convenient, and testable APIs that are custom to the use case, to make if very easy to use those services.

orlandohohmeier: We need to find a better singleton pattern, but that shouldn't block this PR

foundation-ui/services/AuthorizeBundle.js
orlandohohmeier: Please provide an example as documented in https://github.com/dcos/dcos-ui/pull/1130#discussion_r80880867

foundation-ui/services/AuthorizeBundle.js
mlunoe: ✅

foundation-ui/services/AuthorizeBundle.js
mlunoe: This was a decision made in the design doc. Take a look at the "Formalizing Hook entry points" section in the design doc

foundation-ui/services/AuthorizeBundle.js
mlunoe: ✅

foundation-ui/services/AuthorizeBundle.js
mlunoe: See the design doc

mlunoe: Again, that is the purpose of the example in the top. Not sure exactly what you are looking for here

foundation-ui/services/AuthorizeBundle.js
mlunoe: ✅ This is done under the new structure

mlunoe: This is not a singleton, it is a factory. Changed the structure so it should be more clear.
Also, the `new` keyword is probably confusing as well and should not be there. I have removed it now.

mlunoe:

mlunoe:
@orlandohohmeier and @jfurrow this now conforms to the same structure as MountPoint and MountService here: https://github.com/dcos/dcos-ui/pull/1130

mlunoe:
@orlandohohmeier Could you review the AuthService and AuthPoint parts and ignore the MountPoint and MountService structure?

orlandohohmeier:
@mlunoe I'll look into it tomorrow morning

foundation-ui/auth/AuthService.js
```diff
@@ -0,0 +1,105 @@
...
+  /**
+   * Function to retrieve whether user is authorized for a given permission
+   * @param  {String} permission
+   * @param  {Boolean} defaultValue if no filter modifies value,
+   * this will be returned
+   * undefined or null
+   * @param  {Object} props to pass to filter function
+   * @return  {Boolean} wether the component is authorized or not
+   */
+  isAuthorized(permission, defaultValue, props) {
```
orlandohohmeier: Wonder why we need to pass all of these parameters, shouldn't it be enough to if a user has the required permissions?

foundation-ui/auth/AuthService.js
```diff
@@ -0,0 +1,105 @@
...
+  // AuthPoint API
+  /**
+   * Register for changes in AuthService specific to the given permission.
+   * Whenever AuthService has a change it invoke callback
+   * so we can get the newly filtered authorized value.
+   * @param {String} permission to listen for events
+   * @param {Function} callback to call when event fires
+   * @param {Number} priority of callback [-Infinity; Infinity]
+   */
+  addListener(permission, callback, priority) {
```
orlandohohmeier: Let's either use a different method name or emit a general "change" event to avoid confusion with common event emitter/dispatcher implementations.

foundation-ui/auth/AuthService.js
```diff
@@ -0,0 +1,105 @@
...
+const AuthService = {
+  // Package API
+  /**
+   * Registers package for updates on this AuthPoint permission
+   * @param  {String} permission of MointPoint to register package for
+   * @param  {Function} callback to call when content is being fetched
+   * @param  {Number} priority to register package callback with
+   * [-Infinity; Infinity]
+   */
+  register(permission, callback, priority) {
```
orlandohohmeier: I like us to provide a stricter api by defining and testing against an `AuthenticationAdapter` type. The adapter is then used to verify that a user has the required permission.

I picture the authentication service to work like illustrated (!) in the following "example":

```
class AuthenticationService {

    constructor(){
        this.authenticationAdapter = [];
    }

    /**
     * Register an adapter to perform authentication and authorization 
     * checks
     * @param {AuthenticationAdapter} adapter
     */
    registerAuthenticationAdapter(adapter) {
        if(!(adapter instanceof AuthenticationAdapter)){
            throw new TypeError('Adapter should be an instance of AuthenticationAdapter');
        }

        this.authenticationAdapter.push(adapter);
    }

    /**
    * Check  authentication adapters to verify that the user has 
    * the required permission
    */
    isAuthorized(permission){
        return  this.authenticationAdapter.every(function(adapter){
            return adapater.isAuthorized(permission) 
        });
    }
}
```

This API will give enough flexibility to deal with different authentication backends (e.g. oAuth) while making sure that the provided API is simple and easy to use.

foundation-ui/auth/AuthService.js
```diff
@@ -0,0 +1,105 @@
...
+ *     'myID',
+ *     React.Children.toArray(children),
+ *     props
+ *   );
+ *
+ *   // Do something with newContent
+ * }
+ * AuthService.addListener('myID', updateContent);
+ */
+const AuthService = {
```
orlandohohmeier: We should agree an a pattern for creating "singleton" services, as we at the moment either create objects directly or return instances by using new within the module scope. But that should not block this PR.

foundation-ui/auth/AuthPoint.js
```diff
@@ -0,0 +1,114 @@
...
+  }
+}
+
+AuthPoint.defaultProps = {
+  replacementComponent: null,
+  // If nothing registers we want to have everything authorized
+  defaultValue: true
+};
+
+AuthPoint.propTypes = {
```
orlandohohmeier: Shouldn't it be enough to have a prop to define the required permission?

foundation-ui/auth/AuthPoint.js
```diff
@@ -0,0 +1,114 @@
...
+    this.updateState(nextProps);
+  }
+
+  componentWillUnmount() {
+    // Hooks will need ability to remove previously added actions.
+    AuthService.removeListener(this.props.id, this.updateState);
+  }
+
+  updateState(props = this.props) {
+    let {defaultValue, id} = props;
```
orlandohohmeier: I don't understand why we need to pass all these props...

foundation-ui/auth/AuthPoint.js
```diff
@@ -0,0 +1,114 @@
...
+ *     return (
+ *       <AuthPoint id="dcos:ui:sidebar-tabs">
+ *         // Will not be shown if checkPermission (above) returns false.
+ *         <SystemTab />
+ *       </AuthPoint>
+ *     );
+ *   }
+ * };
+ */
+class AuthPoint extends React.Component {
```
orlandohohmeier: Can we rename this to "Auth", similar to the "Mount" component?

orlandohohmeier:
Wonder if there is a nice way to define fallback content in case a user doesn't have required permissions. Maybe something like the following:

```
<Auth permission="dcos:ui:sidebar-tabs" valid={true}>
    <SystemTab />
</Auth>
<Auth permission="dcos:ui:sidebar-tabs" valid={false}>
       Apparently you don't required permissions :(
</Auth>
```

```
<Auth permission="dcos:ui:sidebar-tabs">
    <SystemTab />
    <Fallback>
        Apparently you don't required permissions :(
    </Fallback>
</Auth>
```

orlandohohmeier:
Closing this PR in favor of #1362

PR #1150, opened 2016-09-27
[Pods] DCOS-9986: Using Table instead of CheckboxTable on PodInstancesTable

wavesoft:
Removing checkboxes from `PodInstancesTable`. However I am keeping the logic of the `CheckboxTable` since this is something we will roll-back into when we have the per-instance kill functionality.

image

### DIFF
--- a/src/js/components/PodInstancesTable.js
+++ b/src/js/components/PodInstancesTable.js
@@ -70,7 +70,6 @@ class PodInstancesTable extends React.Component {
   getColGroup() {
     return (
       <colgroup>
-        <col style={{width: '40px'}} />
         <col />
         <col style={{width: '120px'}} />
         <col style={{width: '120px'}} />
--- a/src/js/components/PodInstancesTable.js
+++ b/src/js/components/PodInstancesTable.js
@@ -12,6 +12,7 @@ import PodInstanceList from '../structs/PodInstanceList';
 import PodInstanceStatus from '../constants/PodInstanceStatus';
 import PodTableHeaderLabels from '../constants/PodTableHeaderLabels';
 import PodUtil from '../utils/PodUtil';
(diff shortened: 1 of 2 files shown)


orlandohohmeier:
@wavesoft the tests have failed... could you please have a look?

wavesoft:
@orlandohohmeier I am trying to figure out why...

wavesoft:
Ok, finally fixed the unit tests, using the following magic words ✨ 
```jsx
<Table
        ...
        itemHeight={TableUtil.getRowHeight()}
        ...
/>
```

😮

src/js/components/PodInstancesTable.js
```diff
@@ -357,9 +357,10 @@ class PodInstancesTable extends React.Component {
         data={this.getTableDataFor(instances, filterText)}
         expandAll={!!filterText}
         getColGroup={this.getColGroup}
+        itemHeight={TableUtil.getRowHeight()}
         onCheckboxChange={this.handleItemCheck}
-        sortBy={{prop: 'startedAt', order: 'desc'}}
```
mlunoe: Why does this change? Doesn't this follow how we order tables in general?

wavesoft: I don't know if there is a "common" way of sorting tables. I can change this back.
I was just surprised to see the items suddenly appear in reverse order in the tests (for the same `startedAt` value) and I thought of sorting them by something more unique, like the `name`, just in case two `startedAt` timestamps are identical.

mlunoe: Arh! Sorry, I thought this was the Pod version of tasks. I get it now. Yes, this makes sense.

Poltergeist:
LGTM 🎆

PR #1151, opened 2016-09-27
DCOS-9995: Adding support for force=true on destroy

wavesoft:
It looks like when destroying a service, we never properly checked for errors in order to prompt for force-destroy. This PR:

* Introduces the `force` option to `destroyGroup` and `destroyService`.
* Makes the error message a bit more generic

### DIFF
--- a/src/js/components/ServerErrorModal.js
+++ b/src/js/components/ServerErrorModal.js
@@ -56,7 +56,7 @@ module.exports = class ServerErrorModal extends mixin(StoreMixin) {
     });
   }
 
-  handleServerError({message:errorMessage}) {
+  handleServerError(errorMessage) {
     if (!errorMessage) {
       throw 'No error message defined!';
     }
--- a/src/js/components/ServerErrorModal.js
+++ b/src/js/components/ServerErrorModal.js
@@ -56,7 +56,7 @@ module.exports = class ServerErrorModal extends mixin(StoreMixin) {
     });
   }
 
-  handleServerError(errorMessage) {
+  handleServerError({message:errorMessage}) {
     if (!errorMessage) {
       throw 'No error message defined!';
     }
--- a/src/js/stores/MarathonStore.js
+++ b/src/js/stores/MarathonStore.js
@@ -41,6 +41,7 @@ import DeploymentsList from '../structs/DeploymentsList';
 import GetSetBaseStore from './GetSetBaseStore';
 import HealthStatus from '../constants/HealthStatus';
 import MarathonActions from '../events/MarathonActions';
+import Util from '../utils/Util';
 import {
   MARATHON_APPS_CHANGE,
   MARATHON_APPS_ERROR,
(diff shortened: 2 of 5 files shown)


src/js/stores/MarathonStore.js
```diff
@@ -174,11 +177,15 @@ class MarathonStore extends GetSetBaseStore {
           this.emit(MARATHON_GROUP_CREATE_SUCCESS);
           break;
         case REQUEST_MARATHON_GROUP_DELETE_ERROR:
-          let groupErrorMessage = action.data;
-          if (!Object.keys(groupErrorMessage).length) {
-            groupErrorMessage = 'Error destroying group';
+          data = action.data;
+          if (!data || (Util.isObject(data) && !data.message)) {
```
wavesoft: Here I tried to be compatible with error messages I am not aware of. Do we have any spec of the possible error messages that can appear on the non-pods version?

nLight:
tests 🔥  👀

wavesoft:
Sorry, spent too much time on the tests of https://github.com/dcos/dcos-ui/pull/1150 and didn't saw this was failing 😞

src/js/components/ServerErrorModal.js
```diff
@@ -56,7 +56,7 @@ module.exports = class ServerErrorModal extends mixin(StoreMixin) {
     });
   }
 
-  handleServerError(errorMessage) {
+  handleServerError({message:errorMessage}) {
```
MatApple: This isn't a good idea and will break our existing Server Errors. We should be using `RequestUtil.getErrorFromXHR(xhr)` in all the actions when the request errors. I just noticed we're using `RequestUtil.parseResponseBody(xhr)` in several places. https://github.com/dcos/dcos-ui/blob/master/src/js/events/MarathonActions.js#L112

src/js/stores/MarathonStore.js
MatApple: @wavesoft we should be using this https://github.com/dcos/dcos-ui/blob/master/src/js/events/MarathonActions.js#L64 which will extract the error from the XHR object.

src/js/components/ServerErrorModal.js
wavesoft: Hello @MatApple , this function does not forward the `details` array that might be needed in order to compile a more verbose response. For now I will use this function to the actions I introduced, but we might need to think about this.

Poltergeist: As @wavesoft mentioned we are using the parseResponseBody because the message does not give us enough information about the error. The message for some errors in marathon are for example `Invalid Format` if we surface that error the user does not know which field is triggering this error. In the details is a array containing a detailed information. The following JSON should show this dilemma:
```JSON
{
   "message":"Object is not valid",
   "details":[
      {
         "path":"/",
         "errors":[
            "AppDefinition must either contain one of 'cmd' or 'args', and/or a non-Mesos 'container'."
         ]
      }
   ]
}
```

But I agree that we need a single way to parse errors, but it should be more verbose.

wavesoft:

PR #1154, opened 2016-09-27
[Pods] Replace defaults with an explicit check

nLight:
This PR fixes the situation when marathon provides null instead of not providing anything for an optional property.


wavesoft:
LGTM! 💯

PR #1155, opened 2016-09-27
[Pods] Display container endpoints

nLight:
This PR addresses the differences between app port definitions and pod container endpoints. It seems like they used to be pretty similar that’s why we went ahead and extracted `ServiceConfigUtil`. But they have recently changed.

Keeping in mind those differences we might want to roll back the `ServiceConfigUtil .getPortDefinitionGroups` extraction.

The other reason why I decided not to implement links to load balanced endpoints is because there seems to be a bug in the way how we handle groups of apps/pods I discovered and will file in Jira.

### DIFF
--- a/src/js/components/PodContainerSpecView.js
+++ b/src/js/components/PodContainerSpecView.js
@@ -88,29 +88,31 @@ class PodContainerSpecView extends React.Component {
   }
 
   getEndpointsSection() {
-    let {container: {id, endpoints}} = this.props;
+    let {container: {endpoints}} = this.props;
 
     if (!endpoints || !endpoints.length) {
       return null;
     }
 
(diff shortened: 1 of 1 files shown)


Poltergeist:
flicts :-)

nLight:
@Poltergeist resolved! :)

Poltergeist:

MatApple:
Hey @nLight can we also track in JIRA what we might have to change (in relation to this PR) after that bug gets fixed that you're filing a JIRA for?

nLight:
Hey @MatApple! Thanks for the suggestion, I created a ticket that is blocked by the original one https://mesosphere.atlassian.net/browse/DCOS-10050

wavesoft:
Looks good to me! 👍

MatApple:
Thanks @nLight!

PR #1159, opened 2016-09-28
[Pods] Add quick link to logs to task rows

nLight:
This PR adds a link to a task's logs.

ℹ️  when testing it please note that Logs view seems not to be working in general. So you don't see any logs under the Log tab.

https://mesosphere.atlassian.net/browse/DCOS-9894

screen shot 2016-09-28 at 12 18 03

/cc @leemunroe

### DIFF
--- a/src/js/components/PodInstancesTable.js
+++ b/src/js/components/PodInstancesTable.js
@@ -6,6 +6,7 @@ import {Table} from 'reactjs-components';
 import CollapsingString from './CollapsingString';
 import EventTypes from '../constants/EventTypes';
 import ExpandingTable from './ExpandingTable';
+import Icon from './Icon';
 import MesosStateStore from '../stores/MesosStateStore';
 import Pod from '../structs/Pod';
 import PodInstanceList from '../structs/PodInstanceList';
@@ -22,6 +23,7 @@ const METHODS_TO_BIND = [
   'handleMesosStateChange',
   'renderColumnAddress',
   'renderColumnID',
+  'renderColumnLogs',
   'renderColumnResource',
   'renderColumnStatus',
   'renderColumnUpdated',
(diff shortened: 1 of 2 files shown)


src/js/components/PodInstancesTable.js
```diff
@@ -281,6 +290,31 @@ class PodInstancesTable extends React.Component {
     );
   }
 
+  renderColumnLogs(prop, row, rowOptions = {}) {
+    if (rowOptions.isParent) {
+      return <Icon id="" size="mini" />;
```
nLight: Wonder if there's a better way to have a space here.

leemunroe:
👍

src/js/components/PodInstancesTable.js
wavesoft: `return null` doesn't work?

wavesoft: You could try something like `<div>&nbsp;</div>` since the height is defined by the line height. However I would like a confirmation from the other guys, since we tend to avoid non-breaking space.

wavesoft:
Looks good to me ✨  💯 
Only one comment regarding the possible use of &nbsp; or other breaking character for the empty line.

src/js/components/PodInstancesTable.js
nLight: Awesome, thank you! I picked `span` though. Let's see what other guys say.

wavesoft: As discussed with orlando, just make sure to add a comment for clearly stating why we are using nbsp there.

wavesoft:
Conflicts 😄

nLight:
Fixed! @wavesoft 👋

orlandohohmeier:
💯 ✨  Code looks good to me, going to hit the button once the test pass...

nLight:
@orlandohohmeier it's 💚  🙇

PR #1161, opened 2016-09-28
[Pods] DCOS-10076: Pod health bar inconsistency between services table and pod detail

wavesoft:
Showing the deploying animation on the health bar in the pod detail view, as already shown like so on the services table list.

### DIFF
--- a/src/js/structs/Pod.js
+++ b/src/js/structs/Pod.js
@@ -177,10 +177,6 @@ module.exports = class Pod extends Service {
       taskSummary.tasksOverCapacity = totalInstances - definedInstances;
     }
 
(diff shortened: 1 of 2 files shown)


src/js/structs/Pod.js
```diff
@@ -176,6 +176,9 @@ module.exports = class Pod extends Service {
     if (totalInstances > definedInstances) {
       taskSummary.tasksOverCapacity = totalInstances - definedInstances;
     }
```
orlandohohmeier: #nit let's add an empty line here

wavesoft:
Ok, making things a bit more clear, and not introducing fake data.

PR #1162, opened 2016-09-28
[Pods] Using lastUpdated instead of lastChanged

wavesoft:
The debug tab renders incorrectly the `lastChanged` instead of `lastUpdated` in the Last Changes -> Configuration label. Fixed with this PR.

### DIFF
--- a/src/js/components/PodDebugTabView.js
+++ b/src/js/components/PodDebugTabView.js
@@ -62,12 +62,17 @@ class PodDebugTabView extends React.Component {
 
   getLastVersionChange() {
     let {pod} = this.props;
-    let lastChanged = pod.getLastChanged();
+    let lastUpdated = pod.getLastUpdated();
+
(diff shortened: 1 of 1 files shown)


nLight:
So it says `Last Changes` why would we use lastUpdated? It's a bit confusing.

wavesoft:
Ok, just to clarify things. I know the title says `Last Changed` however, the `lastChanged` property refers to the last time something changed to the pod state, while on the UI we look at the last time the user changed something.

So, in this case, we have to look at the `lastUpdated` property, which refers to the last time the configuration were updated. (Hence the label `Configuration` next to the value)

nLight:
@wavesoft let's have this comment in the code! Then it'd be ✨

wavesoft:
Done 🙇

MatApple:
LGTM! 🚢

PR #1163, opened 2016-09-28
Check for Mesos API capability before making the AJAX request.

rcorral:


src/js/events/MesosSummaryActions.js
```diff
@@ -104,12 +104,17 @@ var MesosSummaryActions = {
...
           'hasCapability', false, 'historyServiceAPI'
         );
 
-        if (!_historyServiceOnline || !canAccessHistoryServer) {
-          requestFromMesos(resolve, reject);
+        if (!_historyServiceOnline || !canAccessHistoryAPI) {
+          let canAccessMesosAPI = Hooks.applyFilter(
+            'hasCapability', false, 'mesosAPI'
+          );
+          if (canAccessMesosAPI) {
```
mlunoe: Wouldn't this mean that we will never make the request in some cases, which would in turn never load the UI?

rcorral: Yup, realized this after I put up the PR and that's why now we emit an event with an empty state.
Which turns out to be a very hacky fix. Got any better ideas on how to do this?
The main problem is that a lot of the pages rely on the MesosSummaryStore to say "hey I've received data" and then they render. The real fix is to go through each page and have them rely on their own store saying "hey I've got data".

mlunoe: Yup, the latter would be preferable. Hmm... not sure what a good temporary solution would look like.
Probably emit another event that is then handled in the MesosSummaryStore separately.
But that would probably be the same as what you are doing here.

How urgent is this? Could we make sure to make this change as we migrate to Relay? Or would it be more beneficial to have the actual fix in place before we start the migration?

rcorral: I think we can fix this once we move to relay.
It's a shitty fix, but good enough IMO. Doesn't really add tech-debt, it just doesn't improve the current way of doing things.
We need this so that the UI can load when Alice only has access to Jobs or Universe but not to Mesos.

mlunoe: works for me

MatApple: I agree - lets wait for the shift to relay.

PR #1169, opened 2016-09-29
[Pods] DCOS-9984:  Adjust task directory path handling to support pods

orlandohohmeier:
Introduce and integrate a new util to get the task directory from Mesos state which is capable of handling nested task groups (pods).

Closes DCOS-9984

### DIFF
--- a/src/js/utils/MesosStateUtil.js
+++ b/src/js/utils/MesosStateUtil.js
@@ -33,9 +33,9 @@ const MesosStateUtil = {
   getFramework(state, frameworkID) {
     const {frameworks, completed_frameworks} = state;
     return [].concat(frameworks, completed_frameworks).find(
-        function (framework) {
-          return framework != null && framework.id === frameworkID;
-        });
+      function (framework) {
+        return framework != null && framework.id === frameworkID;
+      });
   },
 
   /**
@@ -118,17 +118,17 @@ const MesosStateUtil = {
 
           // Find pod task and executor
           return [].concat(executor.tasks, executor.completed_tasks)
-              .every(function (task) {
-                if (task != null && task.id === taskID) {
-                  // For a detail documentation on how to construct the path
-                  // please see: https://reviews.apache.org/r/52376/
-                  taskPath =
-                      `${executor.directory}/tasks/${task.id}/${path}`;
-                  return false;
-                }
+            .every(function (task) {
+              if (task != null && task.id === taskID) {
+                // For a detail documentation on how to construct the path
+                // please see: https://reviews.apache.org/r/52376/
+                taskPath =
+                    `${executor.directory}/tasks/${task.id}/${path}`;
+                return false;
+              }
 
-                return true;
-              });
+              return true;
+            });
         });
 
     return taskPath;
--- a/src/js/utils/MesosStateUtil.js
+++ b/src/js/utils/MesosStateUtil.js
@@ -34,7 +34,7 @@ const MesosStateUtil = {
     const {frameworks, completed_frameworks} = state;
     return [].concat(frameworks, completed_frameworks).find(
         function (framework) {
-          return framework.id === frameworkID;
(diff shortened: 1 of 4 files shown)


src/js/utils/MesosStateUtil.js
```diff
@@ -26,6 +26,19 @@ const MesosStateUtil = {
...
   /**
+   * @param {{frameworks:array,completed_frameworks:array}} state
+   * @param {string} frameworkID
+   * @returns {{executors:Array, completed_executors:Array}|null} framework
+   */
+  getFramework(state, frameworkID) {
+    const {frameworks, completed_frameworks} = state;
+    return [].concat(frameworks, completed_frameworks).find(
+        function (framework) {
+          return framework.id === frameworkID;
```
wavesoft: Make sure `framework` is defined, because `[].concat()` will introduce a `null` or `undefined` value if either of frameworks or completed_frameworks is missing.

wavesoft:
When tests pass I am 👍

src/js/utils/MesosStateUtil.js
```diff
@@ -26,6 +26,19 @@ const MesosStateUtil = {
...
 
   /**
+   * @param {{frameworks:array,completed_frameworks:array}} state
+   * @param {string} frameworkID
+   * @returns {{executors:Array, completed_executors:Array}|null} framework
+   */
+  getFramework(state, frameworkID) {
+    const {frameworks, completed_frameworks} = state;
+    return [].concat(frameworks, completed_frameworks).find(
+        function (framework) {
```
nLight: #nit indentation 4 spaces instead of 2

src/js/utils/MesosStateUtil.js
```diff
@@ -70,8 +83,56 @@ const MesosStateUtil = {
...
+    const framework =
+        MesosStateUtil.getFramework(state, frameworkID);
+
+    if (framework == null) {
+      return taskPath;
+    }
 
+    // Find matching executor or task to construct the task path
+    [].concat(framework.executors, framework.completed_executors)
+        .every(function (executor) {
```
nLight: I'd re-indent that one as well <- 2 spaces #nit

src/js/utils/MesosStateUtil.js
```diff
@@ -26,6 +26,19 @@ const MesosStateUtil = {
   },
 
   /**
+   * @param {{frameworks:array,completed_frameworks:array}} state
+   * @param {string} frameworkID
+   * @returns {{executors:Array, completed_executors:Array}|null} framework
+   */
+  getFramework(state, frameworkID) {
+    const {frameworks, completed_frameworks} = state;
```
nLight: we could add defaults like  `{frameworks=[], completed_frameworks=[]}`

src/js/utils/MesosStateUtil.js
```diff
@@ -26,6 +26,19 @@ const MesosStateUtil = {
...
+    return [].concat(frameworks, completed_frameworks).find(
```
nLight: Then `frameworks.concat(completed_frameworks)`

src/js/utils/MesosStateUtil.js
```diff
@@ -26,6 +26,19 @@ const MesosStateUtil = {
...
+      function (framework) {
+        return framework != null && framework.id === frameworkID;
```
nLight: and it looks like we don't need that check anymore :) Unless `state.frameworks` or `state.completed_frameworks` give null, in case of `null` defaults won't work 😢

nLight:
Awesome work! Just a few #nits :shipit:

src/js/utils/MesosStateUtil.js
orlandohohmeier: Apparently `completed_frameworks` could be `null` but we could define the variables separately but I would vote to keep the null check anyway. What do you think? Should we change this?

PR #1170, opened 2016-09-29
[Pods] DCOS-10099: Getting total resources from instance information

wavesoft:
Making sure we always show the total count of resources, as received from marathon (instead of summing-up the current values of the view, ex. when a filter is applied).

### DIFF
--- a/src/js/components/PodInstancesTable.js
+++ b/src/js/components/PodInstancesTable.js
@@ -229,14 +229,14 @@ class PodInstancesTable extends React.Component {
       let children = this.getContainersWithResources(
         podSpec, containers, instance.getAgentAddress()
       );
-      let resourcesSum = instance.getResources();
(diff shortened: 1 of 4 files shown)


src/js/components/PodInstancesTable.js
```diff
@@ -206,9 +198,10 @@ class PodInstancesTable extends React.Component {
       let containers = instance.getContainers().filter(function (container) {
         return PodUtil.isContainerMatchingText(container, filterText);
       });
-      let {children, resourcesSum} = this.getContainersWithResources(
+      let children = this.getContainersWithResources(
         podSpec, containers, instance.getAgentAddress()
       );
+      let resourcesSum = instance.getResources();
```
orlandohohmeier: How about `let {cpus, mem} = instance.getResources();`?

rcorral:
🚢  after conflicts are resolved

PR #1171, opened 2016-09-29
[Pods] DCOS-10100: Getting resource details from spec when available

wavesoft:
This PR is a temporary solution until https://mesosphere.atlassian.net/browse/DCOS-10098 is addressed. It will automatically switch on using the new data when they become available.

In detail: We encountered a case were a pod contains a container not part of the `spec` (ex. you edit the spec and remove a container, but it is still present in the instances for some time). Therefore, we were not able to get the resource information from the spec and the UI was failing. This PR uses by default the `resources` information from the container itself and for the transient period until the issue is addressed it will also use the information from the spec if available.

### DIFF
--- a/src/js/components/PodInstancesTable.js
+++ b/src/js/components/PodInstancesTable.js
@@ -168,6 +168,8 @@ class PodInstancesTable extends React.Component {
 
     let children = containers.map(function (container) {
       let containerResources = container.getResources();
+
+      // TODO: Remove the following 4 lines when DCOS-10098 is addressed
       let containerSpec = podSpec.getContainerSpec(container.name);
       if (containerSpec) {
         containerResources = containerSpec.resources;
--- a/src/js/components/PodInstancesTable.js
+++ b/src/js/components/PodInstancesTable.js
@@ -167,9 +167,14 @@ class PodInstancesTable extends React.Component {
     };
 
     let children = containers.map(function (container) {
+      let containerResources = container.getResources();
       let containerSpec = podSpec.getContainerSpec(container.name);
-      Object.keys(containerSpec.resources).forEach(function (key) {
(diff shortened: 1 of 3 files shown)


src/js/components/PodInstancesTable.js
```diff
@@ -167,9 +167,14 @@ class PodInstancesTable extends React.Component {
     };
 
     let children = containers.map(function (container) {
+      let containerResources = container.getResources();
       let containerSpec = podSpec.getContainerSpec(container.name);
-      Object.keys(containerSpec.resources).forEach(function (key) {
-        resourcesSum[key] += containerSpec.resources[key];
+      if (containerSpec) {
+        containerResources = containerSpec.resources;
```
orlandohohmeier: Please add a `TODO` as we only needed this until DCOS-10098 is fixed

wavesoft: 👍

src/js/structs/PodContainer.js
```diff
@@ -58,6 +58,16 @@ module.exports = class PodContainer extends Item {
...
   }
 
+  getResources() {
+    let resources = this.get('resources') || {};
+    return Object.assign({
+      cpus: 0,
+      mem: 0,
+      gpus: 0,
+      disk: 0
+    }, resources);
```
nLight: #nit
we could just `Object.assign({...}, this.get('resources'));`
`Object.assign` takes care of `undefined` and `null`

nLight:
Awesome work! Just a small #nit, not sure if we even do it this way.

src/js/structs/PodContainer.js
wavesoft: You are right, cool!

PR #1172, opened 2016-09-29
[Pods] DCOS-9895: Have Pod and task tables use up available space before truncating

nLight:
With this PR less important columns in the pod's instances table will be hidden on the smaller screens as we do for the regular app tasks table. In addition to that I introduced min-/max-width to make use of the empty space as stated in the ticket.

screen shot 2016-09-29 at 13 54 33
sep-29-2016 13-51-35

### DIFF
--- a/src/js/components/PodInstancesTable.js
+++ b/src/js/components/PodInstancesTable.js
@@ -74,7 +74,7 @@ class PodInstancesTable extends React.Component {
     return (
       <colgroup>
         <col />
-        <col />
+        <col className="hidden-mini" />
         <col className="hidden-mini" />
         <col className="hidden-mini" />
         <col className="hidden-mini" />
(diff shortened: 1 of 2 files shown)


wavesoft:
Can you make sure there is always enough space for reasonable value ranges for all the columns? Unfortunately the current implementation of expanding table does not properly handle full row heights and if a value wraps to the next line we have problems. For example:

image

nLight:
@wavesoft interesting, for me 128 doesn't break the line

wavesoft:
Ok, couldn't reproduce the bug I reported. Could be a stale CSS.
Looks good! Excellent work! 💯  😄

leemunroe:
@nLight NICE!

Separate issue: Do we know why the vertical alignment of Host/Port is off? In your screen shots it looks to be a couple pixels higher than the other columns.

PR #1173, opened 2016-09-29
[Pods] DCOS-10089: Including pods on getTasksFromVirtualNetworkName

wavesoft:
The marathon task response is different between a `pod` and an `application`. This PR makes `getTasksFromVirtualNetworkName` aware of the pod structure in order for them to appear on the Virtual Networks table:

image

For reference, the differences on the task responses are:

Pods:

image

App:

image

### DIFF
--- a/src/js/utils/MesosStateUtil.js
+++ b/src/js/utils/MesosStateUtil.js
@@ -64,10 +64,11 @@ const MesosStateUtil = {
       let tasks = framework.tasks || [];
 
       return memo.concat(tasks.filter(function (task) {
-        return Util.findNestedPropertyInObject(
-          task,
-          'container.network_infos.0.name'
-        ) === overlayName;
+        let appPath = 'container.network_infos.0.name';
+        let podPath = 'statuses.0.container_status.network_infos.0.name';
+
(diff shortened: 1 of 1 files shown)


src/js/utils/MesosStateUtil.js
```diff
@@ -64,10 +64,16 @@ const MesosStateUtil = {
       let tasks = framework.tasks || [];
 
       return memo.concat(tasks.filter(function (task) {
-        return Util.findNestedPropertyInObject(
+        let appMatch = Util.findNestedPropertyInObject(
```
nLight: How about
```
let appPath = 'container.network_infos.0.name';
let podPath = 'statuses.0.container_status.network_infos.0.name';

return Util.findNestedPropertyInObject(task, appPath) === overlayName
        || Util.findNestedPropertyInObject(task, podPath) === overlayName;
```

Not sure if it's better

wavesoft: Definitely looks cleaner! Thanks! 👍

wavesoft: Done

PR #1175, opened 2016-09-29
[Pods] DCOS-10108 [2/3]: Adding killPodInstances marathon action

wavesoft:
This PR introduces the `killPodInstances` marathon action, that can be used to kill individual pod instances.


src/js/events/MarathonActions.js
```diff
@@ -495,6 +497,48 @@ var MarathonActions = {
         });
       }
     });
+  },
+
+  killPodInstances(pod, instanceIDs, force) {
+    let podID = encodeURIComponent(pod.getId());
+    let params = '';
+
+    if (!(pod instanceof Pod)) {
```
MatApple: Hey @wavesoft I like that you're thinking about these checks but do you really think we need them here? I think we should bring this up in a team sync to chat about it but for now we should remove them for consistency.

PR #1185, opened 2016-09-30
[Pods] DCOS-10183: Fix network labels destructuring

nLight:
This PR addresses the issue of using `network` as `labels` because of a bug in destructuring.


wavesoft:
LGTM! 🚢

PR #1188, opened 2016-09-30
DCOS-9891: Renaming Tasks to Instances in pods

wavesoft:
This PR addresses task naming issues:

* On the collection view `Tasks` are now called `Instances` on the pods (I don't have a large enough screen to get a screenshot of this)
* On the pod instance detail view, the filter header shows `Showing X of X Instances` instead of tasks.

image

### DIFF
--- a/src/js/components/ServiceDetailTaskTab.js
+++ b/src/js/components/ServiceDetailTaskTab.js
@@ -20,7 +20,10 @@ class ServiceDetailTaskTab extends mixin(StoreMixin) {
     let tasks = MesosStateStore.getTasksByService(this.props.service);
 
     return (
-      <TaskView tasks={tasks} inverseStyle={true} label="Instance"
+      <TaskView
+        tasks={tasks}
+        inverseStyle={true}
+        label="Instance"
         parentRouter={this.context.router} />
     );
   }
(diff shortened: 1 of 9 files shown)


wavesoft:
Ok, so apparently all 'Tasks' should be named 'Instances', if that's correct and we are in hurry we can merge this, however this means that all variable names must be changed to `instanceXXX` from `taskXXX`

MatApple:
Is this correct?

orlandohohmeier:
@wavesoft @MatApple Nope. We should only change "Task" to "Instance" in the overview and service detail views, but shouldn't touch the task detail and node page. I'll update the branch respectively.

orlandohohmeier:
The following labels/names have changed:






/cc @leemunroe @pyronicide @marcomonaco

src/js/components/ServiceDetailTaskTab.js
```diff
@@ -20,7 +20,7 @@ class ServiceDetailTaskTab extends mixin(StoreMixin) {
     let tasks = MesosStateStore.getTasksByService(this.props.service);
 
     return (
-      <TaskView tasks={tasks} inverseStyle={true}
+      <TaskView tasks={tasks} inverseStyle={true} label="Instance"
```
mlunoe: #nit let's break this into mulitple lines. Shouldn't hold the PR back, though

wavesoft:
LGTM 🚢

marcomonaco:
good 🚢

leemunroe:
👍

PR #1196, opened 2016-10-03
[CNVS Upgrade] Enable media-object styles

jfurrow:


src/styles/index.less
```diff
@@ -48,10 +48,14 @@
...
+// Iframes
 
-@import 'layout/sidebar/variables.less';
-@import 'layout/sidebar/styles.less';
+@import 'layout/iframes/styles.less';
+
+// Media Object
+
+@import 'layout/media-object/variables.less';
+@import 'layout/media-object/styles.less';
```
mlunoe: This file is still under `layout/media-object.less`

src/styles/index.less
```diff
@@ -61,9 +65,15 @@
 
 @import 'layout/page-body/styles.less';
 
-// @import 'layout/iframes.less';
-// @import 'layout/media-object.less';
-// @import 'layout/modal.less';
+// Responsive Utilities
+
+@import 'layout/responsive-utilities/styles.less';
```
mlunoe: This file needs to be moved from `utilities/responsive-utilities.less`

src/styles/index.less
```diff
@@ -48,10 +48,14 @@
 
 @import 'layout/header/styles.less';
 
-// Sidebar
+// Iframes
 
-@import 'layout/sidebar/variables.less';
-@import 'layout/sidebar/styles.less';
+@import 'layout/iframes/styles.less';
```
mlunoe: This file needs to be moved from `layout/iframes.less`

jfurrow: I lost a commit somewhere... this is fixed!

src/styles/index.less
jfurrow: I lost a commit somewhere... but this actually should have been removed. The responsive classes that we use (`.hidden-*`, `.visible-*` etc) are coming directly from CNVS right now, though they are mis-named and need to be fixed there. I spoke with John A about this already, it just needs to be done.

PR #1202, opened 2016-10-04
[CNVS Upgrade] Replace instances of container-pod with pod

jfurrow:
This PR replaces all instances of `container-pod` classes with `pod` where necessary. I removed some class assignments that are no longer necessary.

### DIFF
--- a/src/js/components/PodDetail.js
+++ b/src/js/components/PodDetail.js
@@ -90,16 +90,13 @@ class PodDetail extends mixin(TabsMixin) {
     const {activeActionDialog} = this.state;
 
     return (
-      <div className="flex-container-col">
-        <div className="container-pod container-pod-divider-bottom-align-right container-pod-short-top flush-bottom flush-top media-object-spacing-wrapper media-object-spacing-narrow">
-          <Breadcrumbs />
-          <PodHeader
-            onAction={this.handleAction}
-            pod={pod}
-            tabs={this.tabs_getUnroutedTabs()} />
-          {this.tabs_getTabView()}
-        </div>
-
+      <div>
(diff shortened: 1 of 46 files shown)


src/js/components/PodDetail.js
```diff
@@ -90,16 +90,13 @@ class PodDetail extends mixin(TabsMixin) {
     const {activeActionDialog} = this.state;
 
     return (
-      <div className="flex-container-col">
```
mlunoe: I think we might actually need this for the log view to grow to its parent

mlunoe: I think we might actually need this for the log view to grow to its parent

src/js/components/ServiceDetail.js
```diff
@@ -117,17 +117,11 @@ class ServiceDetail extends mixin(InternalStorageMixin, TabsMixin) {
...
-          container-pod-divider-bottom-align-right
-          container-pod-short-top flush-bottom flush-top
-          service-detail-header media-object-spacing-wrapper
-          media-object-spacing-narrow">
-          <Breadcrumbs />
-          <ServiceInfo onActionsItemSelection={this.onActionsItemSelection}
-            service={service} tabs={this.tabs_getUnroutedTabs()} />
-          {this.tabs_getTabView()}
-        </div>
+      <div>
```
mlunoe: Same thing here

src/js/components/PodDetail.js
jfurrow: The page flex properties need to be addressed in a separate PR; the `flex-container-col` class definition doesn't exist anymore.

PR #1207, opened 2016-10-04
[CNVS Upgrade] Fix table-cell-flexbox container

jfurrow:
This PR removes the dependency on flex-specific classes for this element.


mlunoe:
Let's make sure that we handle the other cases where `table-cell-value` is used.

jfurrow:

PR #1211, opened 2016-10-06
[Pods] DCOS-10321: Including completed pod instances on summary view

wavesoft:
This PR uses the mesos state endpoint to extend the list of instances in the `PodInstancesView` to include completed instances.

image

### DIFF
--- a/src/js/utils/MesosStateUtil.js
+++ b/src/js/utils/MesosStateUtil.js
@@ -3,9 +3,12 @@ import Util from './Util';
 
 const RESOURCE_KEYS = ['cpus', 'disk', 'mem'];
 
-// Based on the regex marathon uses here:
+// Based on the regex that marathon uses to validate task IDs,
+// but keeping only the 'instance-' prefix, that refers to pods.
+// The 'marathon-' prefix is used for task launched because of an AppDefinition.
+//
 // https://github.com/mesosphere/marathon/blob/feature/pods/src/main/scala/mesosphere/marathon/core/task/Task.scala#L134
-const POD_TASK_REGEX = /^(.+)\.(?:instance-|marathon-)([^_\.]+)[\._]([^_\.]+)$/;
+const POD_TASK_REGEX = /^(.+)\.instance-([^_\.]+)[\._]([^_\.]+)$/;
 
 function setIsStartedByMarathonFlag(name, tasks) {
   return tasks.map(function (task) {
@@ -14,6 +17,21 @@ function setIsStartedByMarathonFlag(name, tasks) {
 }
 
 const MesosStateUtil = {
+
+  /**
+   * De-compose the given task id into it's primitive components
+   *
+   * @param {String} taskID - The task ID to decompose
+   * @returns {{podID, instanceID}} Returns the ID components
+   */
+  decomposePodTaskId(taskID) {
+    let [, podID, instanceID, taskName] = POD_TASK_REGEX.exec(taskID);
+    return {
+      podID,
+      instanceID,
+      taskName
+    };
+  },
 
   flagMarathonTasks(state) {
     let newState = Object.assign({}, state);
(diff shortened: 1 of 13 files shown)


src/js/utils/MesosStateUtil.js
```diff
@@ -1,7 +1,12 @@
+import PodInstanceState from '../constants/PodInstanceState';
 import Util from './Util';
 
 const RESOURCE_KEYS = ['cpus', 'disk', 'mem'];
 
+// Based on the regex marathon uses here:
+// https://github.com/mesosphere/marathon/blob/feature/pods/src/main/scala/mesosphere/marathon/core/task/Task.scala#L134
+const POD_TASK_REGEX = /^(.+)\.(?:instance-|marathon-)([^_\.]+)[\._]([^_\.]+)$/;
```
nLight: I'd love the front-end to have no idea about these regexps :(

src/js/utils/MesosStateUtil.js
```diff
@@ -87,6 +92,98 @@ const MesosStateUtil = {
...
+
+    // Try to compose actual PodInstance structures from the information we
+    // have so far. Obviously we don't have any details, but we can populate
+    // most of the UI-interesting fields by summarising container details
+    return Object.keys(instancesMap).map(function (instanceId) {
+      let containers = instancesMap[instanceId];
+      let summaryProperties = containers.reduce(function (memo, instance) {
+        let {resources={}, lastChanged=0} = instance;
+
+        memo.resources.cpus += resources.cpus || 0;
```
nLight: Is it possible to get `null` for these resources? If not we could destructure `resources` object and provide defaults for them.

src/js/utils/PodUtil.js
```diff
@@ -36,7 +39,53 @@ var PodUtil = {
...
+    }
+
+    // De-compose PodInstances into plain objects, so we always operate
+    // with plain objects
+    let podInstancesMap = podInstances.reduceItems(function (memo, instance) {
+      memo[instance.getId()] = instance.get();
+      return memo;
+    }, {});
+
+    historicalInstances.forEach(function (instance) {
```
nLight: Every time I see `forEach` I think how it could be replaced with `map` or `reduce`

src/js/utils/MesosStateUtil.js
```diff
@@ -87,6 +92,98 @@ const MesosStateUtil = {
...
+      return framework.name === 'marathon';
+    });
+
+    if (!marathonFramework) {
+      return [];
+    }
+
+    let instancesMap = marathonFramework.completed_tasks.reduce(
+      function (memo, task) {
+        if (POD_TASK_REGEX.test(task.id)) {
```
wavesoft: Yup, that's actually a good idea. 👍

src/js/utils/MesosStateUtil.js
```diff
@@ -87,6 +92,98 @@ const MesosStateUtil = {
...
+        return memo;
+      }, {}
+    );
+
+    // Try to compose actual PodInstance structures from the information we
+    // have so far. Obviously we don't have any details, but we can populate
+    // most of the UI-interesting fields by summarising container details
+    return Object.keys(instancesMap).map(function (instanceId) {
+      let containers = instancesMap[instanceId];
+      let summaryProperties = containers.reduce(function (memo, instance) {
```
wavesoft: Well the task is just to summarise information, and it's tightly related to the rest of the function context. I think it's fairly ok to keep it like this.

src/js/utils/PodUtil.js
wavesoft: Yup, this was a left-over from a previous implementation were `reduce` was making things uglier. This version can actually be a bit cleaner with `reduce`. Fixed 👍

src/js/utils/MesosStateUtil.js
wavesoft: I think this would be cleaner if we are assigning values, but in this case we are adding. From what I see this would make things more complicating for 4 lines of code. Can you write down your suggestion?

nLight:
Awesome work!

PR #1212, opened 2016-10-06
DCOS-10394 (1.8): Adjust util to use the provided service port

orlandohohmeier:
---
ℹ️ This fix is for 1.8 only please see #1213 for respective master fix.

---

Adjust the service util to always use the provided service port data from the JSON definition.

Fixes DCOS-10394

### DIFF
--- a/src/js/utils/ServiceUtil.js
+++ b/src/js/utils/ServiceUtil.js
@@ -350,16 +350,17 @@ const ServiceUtil = {
               let lbPort = parseInt(port.lbPort || 0, 10);
               portMapping.containerPort = lbPort;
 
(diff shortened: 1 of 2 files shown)


src/js/utils/ServiceUtil.js
```diff
@@ -350,16 +350,17 @@ const ServiceUtil = {
               let lbPort = parseInt(port.lbPort || 0, 10);
               portMapping.containerPort = lbPort;
 
-              if (networkType === 'bridge') {
+              if (ValidatorUtil.isDefined(port.hostPort)) {
```
nLight: No need to check for `networkType `?

PR #1216, opened 2016-10-07
[WIP] Upgrade react-router

nLight:
⚠️  **This is just a WIP PR to give you a feeling about the upcoming change**

------

This PR upgrades react-router from 0.13 to 2.8.1
API has changed, some concepts have changed so please be patient. 
All suggestions are welcome!


src/js/pages/ClusterPage.js
```diff
@@ -8,7 +7,7 @@ class ClusterPage extends React.Component {
   render() {
     return (
       <Page title="Cluster">
-        <RouteHandler />
+        {this.props.children}
```
mlunoe: Are these an intermediate step to upgrading?
I really think these should remain routes

nLight: https://github.com/ReactTraining/react-router/blob/master/upgrade-guides/v1.0.0.md#routehandler

PR #1217, opened 2016-10-07
DCOS-10443: Deactivate task log cell caching

orlandohohmeier:
Don't cache the task log cells as otherwise the links will all point to the same task detail.

Closes DCOS-10443


wavesoft:
LGTM 🚢

PR #1218, opened 2016-10-07
[Nodes Page] Modularize the nodes page

philipnrmn:
This PR moves all the components which are specific to the Nodes Page to `plugins/nodes`. 

~~Testing this PR will require the `cnvs-upgrade` branch in `reactjs-components` checked out and linked locally, and use of the `REACT_JS_COMPONENTS_LOCAL=true` EV.~~


philipnrmn:
Rebased and fixed tests

PR #1220, opened 2016-10-07
[Nodes Page] Nodes detail page

philipnrmn:
⚠️ This PR depends on #1218

This PR separates the tabs in the nodes detail page into container and component. 

~~Testing this PR will require the `cnvs-upgrade` branch in `reactjs-components` checked out and linked locally, and use of the `REACT_JS_COMPONENTS_LOCAL=true` EV.~~

### DIFF
--- a/plugins/nodes/src/js/components/HealthTab.js
+++ b/plugins/nodes/src/js/components/HealthTab.js
@@ -1,3 +1,4 @@
+import PureRenderMixin from 'react-addons-pure-render-mixin';
 import React from 'react';
 import {ResourceTableUtil} from 'foundation-ui';
 import {Table} from 'reactjs-components';
@@ -22,6 +23,7 @@ class HealthTab extends React.Component {
   constructor() {
     super(...arguments);
 
+    this.shouldComponentUpdate = PureRenderMixin.shouldComponentUpdate;
     this.state = {
       healthFilter: 'all',
       searchString: ''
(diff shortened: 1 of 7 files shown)


plugins/nodes/src/js/components/HealthTab.js
```diff
@@ -22,6 +23,7 @@ class HealthTab extends React.Component {
   constructor() {
     super(...arguments);
 
+    this.shouldComponentUpdate = PureRenderMixin.shouldComponentUpdate;
```
MatApple: Are you sure we don't have to bind this? Seems like it might always return true because it might read `this.state` and `this.props` as `undefined`

src/js/pages/system/units-health-node-detail/NodeInfoPanel.js
```diff
@@ -0,0 +1,47 @@
...
+    super(...arguments);
+    this.shouldComponentUpdate = PureRenderMixin.shouldComponentUpdate;
+  }
+
+  render() {
+    const {summary, docsURL, output} = this.props;
+
+    return (
+      <div className="flex-container-col flex-grow">
+        <span className="h4 inverse flush-top">Summary</span>
```
MatApple: We should remove all `inverse` classes :) They no longer exist

MatApple:
I'm going to merge this and open a PR against `mod/1.10` to remove all `inverse` classes and bind the PureRenderMixins.

src/js/pages/system/units-health-node-detail/NodeInfoPanel.js
philipnrmn: Oh, nice catch. Thanks!

PR #1221, opened 2016-10-07
[CNVS Upgrade] Fix hidden-* classes

jfurrow:
These responsive classes are named incorrectly in the published version of CNVS. I submitted two PRs to CNVS to fix them, one of which has already merged. I don't currently have access to publish the NPM module. So if you'd like to test this, you'll need to `npm link` this branch of CNVS: https://github.com/mesosphere/canvas/pull/29

This PR updates all of the `hidden-*` classes. It's not exactly a one-to-one update, due to the following:
1. All screen sizes were renamed in CNVS, so `mini` became `small`, `small` became `medium`, `medium` became `large`, and `large` became `jumbo`.
2. Previously `hidden-small` and `hidden-medium` would hide elements only within the range of that breakpoint.
3. Previously `hidden-mini` hid elements at `mini` and below, and `hidden-large` hid elements at `large` and above.

These responsive classes are now more obvious:
1. `hidden-{breakpoint}` will hide elements only within that breakpoint's range.
2. `hidden-{breakpoint}-up` will hide elements at the breakpoint's min width and up.
3. `hidden-{breakpoint}-down` will hide elements at the breakpoint's max width and down.

Thus, I condensed classes like `hidden-medium hidden-small hidden-mini` into `hidden-large-down`.

### DIFF
--- a/src/js/components/SidebarFilter.js
+++ b/src/js/components/SidebarFilter.js
@@ -210,7 +210,7 @@ class SidebarFilter extends mixin(QueryParamsMixin) {
     let {props} = this;
 
     return (
-      <div className="side-list sidebar-filters hidden-medium hidden-small pod flush-top flush-left">
+      <div className="side-list sidebar-filters pod flush-top flush-left">
         <div className="sidebar-filters-header label">
           {this.getTitle()}
           {this.getClearLinkForFilter(props.filterType)}
--- a/src/js/components/SidebarLabelsFilter.js
+++ b/src/js/components/SidebarLabelsFilter.js
@@ -211,7 +211,7 @@ class SidebarLabelsFilters extends mixin(QueryParamsMixin) {
     }
 
     return (
-      <div className="side-list sidebar-filters hidden-medium hidden-small pod flush-top flush-left">
+      <div className="side-list sidebar-filters pod flush-top flush-left">
         {this.getLabelsDropdown()}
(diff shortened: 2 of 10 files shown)


src/js/components/SidebarFilter.js
```diff
@@ -210,7 +210,7 @@ class SidebarFilter extends mixin(QueryParamsMixin) {
     let {props} = this;
 
     return (
-      <div className="side-list sidebar-filters hidden-medium hidden-small pod flush-top flush-left">
+      <div className="side-list sidebar-filters pod flush-top flush-left">
```
jfurrow: I was able to remove these entirely because their parent, `ServiceSidebarFilters`, uses the appropriate class.

src/js/components/SidebarLabelsFilter.js
```diff
@@ -211,7 +211,7 @@ class SidebarLabelsFilters extends mixin(QueryParamsMixin) {
     }
 
     return (
-      <div className="side-list sidebar-filters hidden-medium hidden-small pod flush-top flush-left">
+      <div className="side-list sidebar-filters pod flush-top flush-left">
```
jfurrow: I was able to remove these entirely because their parent, `ServiceSidebarFilters`, uses the appropriate class.

philipnrmn:
Working nicely 👍

PR #1225, opened 2016-10-10
[CNVS Upgrade] Fix task log viewer

jfurrow:
This PR applies the correct classes for the log viewer.

### DIFF
--- a/src/js/components/FilterInputText.js
+++ b/src/js/components/FilterInputText.js
@@ -92,15 +92,15 @@ class FilterInputText extends React.Component {
     });
 
     return (
-      <span className="form-control-group-add-on form-control-group-add-on-append">
+      <span className="form-control-group-add-on">
         {sideText}
         <a onClick={this.handleInputClear}>
-        <Icon
-          family="mini"
-          id="ring-close"
-          size="mini"
-          className={iconClassNames}
-          color={color} />
+          <Icon
+            family="mini"
+            id="ring-close"
+            size="mini"
+            className={iconClassNames}
+            color={color} />
         </a>
       </span>
     );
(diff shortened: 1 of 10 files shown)


src/js/components/FilterInputText.js
```diff
@@ -92,15 +92,15 @@ class FilterInputText extends React.Component {
     });
 
     return (
-      <span className="form-control-group-add-on form-control-group-add-on-append">
+      <span className="form-control-group-add-on">
```
mlunoe: I still see the `form-control-group-add-on-append` and `form-control-group-add-on-prepend` in the Form styles for reactjs-components.
Should these styles live here or there? If here, then we should probably remove them from reactjs-components or the other way around.

jfurrow: Good call, I didn't realize that. I just took a quick glance at what's in `reactjs-components` and I believe that it should be removed from `reactjs-components`. The classname I'm using here comes from CNVS.

mlunoe: I still see them those classes in `FormsTabContent`, though.

jfurrow: The `FormsTabContent` component is part of @ashenden style guide; I suspect it's just leftover from the previous Canvas documentation. The `form-control-group-add-on*` family of classes should be styled via CNVS, not via some custom styling we added to reactjs-components, in my opinion.

In any case, the `form-control-group-add-on-append` class in reactjs-components has no properties that are being applied here, so it should be removed even if we decide to keep the custom styles in reactjs-components.

mlunoe: Ok, thanks for checking!
I missed that that was part of CNVS.

PR #1231, opened 2016-10-11
[CNVS Upgrade] Style table cell links

jfurrow:
This implements two `table-cell-link-*` classes throughout the application.

The `...-primary` class is bold link and the `...-secondary` class is the normal weight.


jfurrow:

PR #1234, opened 2016-10-12
[Pods] DCOS-10594: Fixing killPodInstance endpoint

wavesoft:
Fixing a bug with the `Kill` pod instance button. Apparently:

* ~~The pod spec is wrong, the correct endpoint is `/{id}::instance` (without tailing `s`)~~ fixed in marathon 1.4.0-snap16
* The pod `id` is actually a full path, without the heading slash. It should be injected as a URL component, without escaping

__Hold merge:__ Until marathon fixes the endpoint to `::instances` again.

### DIFF
--- a/src/js/events/MarathonActions.js
+++ b/src/js/events/MarathonActions.js
@@ -492,8 +492,12 @@ var MarathonActions = {
   },
 
   killPodInstances(pod, instanceIDs, force) {
-    let podID = pod.getId().substr(1);
+    let podID = pod.getId();
     let params = '';
+
+    if (podID[0] === '/') {
+      podID = podID.substr(1);
+    }
 
     if (force) {
       params = '?force=true';
--- a/src/js/events/MarathonActions.js
+++ b/src/js/events/MarathonActions.js
@@ -492,7 +492,7 @@ var MarathonActions = {
   },
 
   killPodInstances(pod, instanceIDs, force) {
-    let podID = encodeURIComponent(pod.getId());
+    let podID = pod.getId().substr(1);
     let params = '';
 
     if (force) {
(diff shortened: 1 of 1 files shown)


wavesoft:
So apparently this is not spec error, but implementation error. Will roll-back some parts of the PR.

src/js/events/MarathonActions.js
```diff
@@ -492,7 +492,7 @@ var MarathonActions = {
   },
 
   killPodInstances(pod, instanceIDs, force) {
-    let podID = encodeURIComponent(pod.getId());
+    let podID = pod.getId().substr(1);
```
orlandohohmeier: Let's use a regexp instead of `substr` to only remove leading slashes "/"

wavesoft: Done :nerd_face:

orlandohohmeier:

src/js/events/MarathonActions.js
```diff
@@ -492,9 +492,13 @@ var MarathonActions = {
   },
 
   killPodInstances(pod, instanceIDs, force) {
-    let podID = encodeURIComponent(pod.getId());
+    let podID = pod.getId();
     let params = '';
 
+    if (podID[0] === '/') {
+      podID = podID.substr(1);
```
Poltergeist: I would rather choose a regular expression to replace the `/` from the beginning of a string
```JS
podID = podID.replace(/^\//,'');
```

Or do the replace in line 495 
```JS
let podID = pod.getId().replace(/^\//,'');

Comment (wavesoft) [inline src/js/events/MarathonActions.js]: Ok, Regex might be cleaner, but the performance optimiser inside me kicked-in 😄 

(I really had to crank the iterations up to 1,000,000 in order to see some numbers, so this optimisation really doesn't make any sense, but:)
```
substr: 21ms
regex: 45ms
```

jfurrow:
:ship:

wavesoft:
_Rebased_

wavesoft:
@orlandohohmeier I didn't manage to reproduce the error you were referring to. Could it be a rebase issue? Can you have a look again?

orlandohohmeier:
LGTM 💯 ✨

PR #1238, opened 2016-10-13
Adds pre-commit hook that will get installed on npm install. This will run linters

rcorral:


.pre-commit-hook
```diff
@@ -0,0 +1,11 @@
...
+
+set -e
+
+echo "Pre-commit checks..."
+
+npm run lint;
+
+if [ "undefined" != `npm config get externalplugins` ]; then
+  npm run lint-plugins;
+fi
```
jfurrow: This needs more emoji...

jfurrow:
This is pretty sweet! It might get annoying when creating several commits from a bunch of changes, but it's so much better than the alternative.

.pre-commit-hook
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
+
+set -e
+
+echo "Pre-commit checks..."
```
jfurrow: This needs more emoji...

rcorral: Done!

PR #1243, opened 2016-10-13
[CNVS Upgrade] Create Tabs component

jfurrow:
This has been updated with all of the suggestions below (up to Oct. 18 at 2pm) and should be ready for review!

Here's how you'd use it:
```jsx
<Tabs vertical={true}>
  <TabButtons>
    <TabButton id="services" label="Services">
      <TabButton id="container-1" label="Container 1" />
      <TabButton id="container-2" label="Container 2">
        <TabButton id="container-3" label="Container 3" />
      </TabButton>
    </TabButton>
  </TabButtons>
  <TabViews>
    <TabView id="services">
      Services content
    </TabView>
    <TabView id="container-1">
      Container 1
    </TabView>
    <TabView id="container-2">
      Container 2
    </TabView>
    <TabView id="container-3">
      Something Else
    </TabView>
  </TabViews>
</Tabs>
```

Use of the wrapping `Tabs` component is optional; in that scenario, the owner of `TabViews` and `TabButtons` would be responsible for passing the `activeTab` to `TabViews` and `TabButtons`.

### DIFF
--- a/src/js/components/TabButton.js
+++ b/src/js/components/TabButton.js
@@ -1,0 +1,68 @@
+import classNames from 'classnames/dedupe';
+import React from 'react';
+
+class TabButton extends React.Component {
+  constructor() {
+    super();
+
+    this.handleClick = this.handleClick.bind(this);
+  }
+
+  getChildren() {
+    let {activeTab, children, onClick} = this.props;
+
+    return React.Children.map(children, (tabChild) => {
+      if (tabChild.type === TabButton) {
+        return React.cloneElement(tabChild, {activeTab, onClick});
+      }
+
+      return TabChild;
+    });
+  }
+
+  handleClick(event) {
+    event.stopPropagation();
+
+    if (this.props.onClick) {
+      this.props.onClick(this.props.id);
+    }
+  }
+
+  render() {
+    let {active, activeTab, className, label, labelClassName, id} = this.props;
+    let classes = classNames(
+      'menu-tabbed-item',
+      {
+        'active': active || activeTab === id
+      },
+      className
+    );
+    let labelClasses = classNames('menu-tabbed-item-label', labelClassName);
+
+    return (
+      <div className={classes}>
+        <span className={labelClasses} onClick={this.handleClick}>
+          {label}
+        </span>
+        {this.getChildren()}
+      </div>
+    );
+  }
+}
+
+const classProps = React.PropTypes.oneOfType([
+  React.PropTypes.array,
+  React.PropTypes.object,
+  React.PropTypes.string
+]);
+
+TabButton.propTypes = {
+  active: React.PropTypes.bool,
+  children: React.PropTypes.node,
+  className: classProps,
+  id: React.PropTypes.string.isRequired,
+  label: React.PropTypes.node,
+  labelClassName: classProps
+};
+
+module.exports = TabButton;
--- a/src/js/components/TabButton.js
+++ b/src/js/components/TabButton.js
@@ -1,11 +1,15 @@
 import classNames from 'classnames/dedupe';
 import React from 'react';
 
+const METHODS_TO_BIND = ['handleClick'];
+
 class TabButton extends React.Component {
   constructor() {
-    super();
+    super(...arguments);
 
-    this.handleClick = this.handleClick.bind(this);
+    METHODS_TO_BIND.forEach((method) => {
+      this[method] = this[method].bind(this);
+    });
   }
 
   getChildren() {
--- a/src/js/components/TabButtons.js
+++ b/src/js/components/TabButtons.js
@@ -1,0 +1,49 @@
+import classNames from 'classnames/dedupe';
+import React from 'react';
+
+class TabButtons extends React.Component {
+  getChildren() {
+    let {activeTab, children, onChange} = this.props;
+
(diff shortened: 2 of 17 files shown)


wavesoft:
Hey @jfurrow this is almost exactly what I did on my tab prototype 👍 

However, after some brainstorming with @orlandohohmeier , we figured out that this automatic button generation magic will quickly become black magic. So we ended up with a surprisingly simple, yet much cleaner solution for the tab buttons:

1.  Separate buttons to a different component, fully customizable, that share the same `activeTab` property with the `<Tabs />` component. Since buttons are something very simple and something rarely changes they can be easily written by hand. Something like:

  ```js
  <TabButtons activeTab={this.state.activeTab} activeLinkClass="selected" onTabClick={}>
    <Link path="/services" tabId="services">Services</Link>
    <Link path="/services" tabId="services">Services</Link>
  </TabButtons>
  ```
  
  (The `tabId` and `activeLinkClass` is my syntactic sugar suggestion that I came up with right now. It could be easily replaced with `className={classNames('active': this.state.activeTab === 'path')}` on every link, or we could even rely on the latest `<Link>` automatic selected detection)

2. The `onTabClick` event should be fired only by the buttons component. The `<Tabs>` should be a passive component. Perhaps replicating a lightweight logic of the `<Router />` to make it possible to migrate to a real router if the situation changes in the future (?)

3. Dynamic buttons should be generated in the same way dynamic tabs will be generated. (ex. as we would do `getMoMServices()` we could to a `getMoMServiceTabButtons` that use the same data source).

jfurrow:
Thanks for the feedback @wavesoft!

I think I'm following your suggested, and the logic is definitely more simple. I have some questions:
1. I'm not sure what you mean about the `<Tabs>` component having lightweight logic of a `<Router />`; what logic are you thinking for this?
2. Is this how you imagine it all working together?
```jsx
<TabButtons activeTab={this.state.activeTab} activeLinkClass="selected" onTabClick={(tabID) => { this.setState({activeTab: tabID}) }}>
  <Link path="/services" tabId="services">Services</Link>
  <Link path="/other" tabId="other">Other</Link>
</TabButtons>
<Tabs activeTab={this.state.activeTab}>
  <Tab tabId="services">
    Services Tab Content
  </Tab>
</Tabs>
```

rcorral:
I'm curious about how many times we'll have tabs that are away from their contents, for which this solution would be useful.

I fear that with your suggestion @wavesoft, we'll eventually end up with this type of pattern:

```
<SuperHighLevelContainer>
  <SomeWrappingComponent activeTab={this.state.activeTab} onChange={}>
    <TabButtons activeTab={this.state.activeTab} onChange={}>
      ...
    </TabButtons>
  </SomeWrappingComponent>

  <SomeOtherContainer activeTab={this.state.activeTab}>
    <Tabs activeTab={this.state.activeTab}>
      ...
    </Tabs>
  </SomeOtherContainer>
</SuperHighLevelContainer>
```

You could see how it could quickly get out of control to be passing around the `activeTab` state.

orlandohohmeier:
I suggest to have a rather inflexible tab component for the 80% use case which is implemented using a `Content` and `Menu` component. This would give as a set of components that is easy to use, serves all the described cases and separates the "logic" into small reusable chunks.  

```jsx
<Tabs>
  <Tab label="Parent C" id="parent-c">
    Parent C Content
  </Tab>
</Tabs>

// ------- 

<Menu selected={string}  onChange ={function}>
   <MenuItem|Link id={string}>
</Menu>

<Content section={string}>
  <Section id={string}>
    Section Content
  </Section>
</Content>
```

What do you think?

rcorral:
Yup, I get it and I like the proposal. I just want to avoid passing the `activeTab` down many components as that'll be error prone and confusing.
Would love to see if there's something we can do to avoid that pattern while still having this flexible approach.

jfurrow:
@wavesoft @orlandohohmeier @rcorral All comments addressed, let me know what you think

src/js/components/TabButton.js
```diff
@@ -0,0 +1,68 @@
+import classNames from 'classnames/dedupe';
+import React from 'react';
+
+class TabButton extends React.Component {
+  constructor() {
+    super();
```
mlunoe: #nit `...arguments`

src/js/components/TabButton.js
```diff
@@ -0,0 +1,68 @@
...
+
+    this.handleClick = this.handleClick.bind(this);
```
mlunoe: `METHODS_TO_BIND`

src/js/components/TabButtons.js
```diff
@@ -0,0 +1,49 @@
+import classNames from 'classnames/dedupe';
+import React from 'react';
+
+class TabButtons extends React.Component {
```
mlunoe: Let's figure out a way to differentiate component names so it is not only differentiated by `s`

src/js/components/TabButton.js
jfurrow: Done!

src/js/components/TabButton.js
jfurrow: Done!

mlunoe:
Awesome work! Looks good works great!

MatApple:
This looks great @jfurrow! Nice work!

PR #1244, opened 2016-10-14
DCOS-10624: Adjust get service by task id getter

orlandohohmeier:
Return the matching service instance instead of creating a new one using the abstract service struct.

Fixes DCOS-10624


nLight:
Awesome work! Thank you for the fix! 🚢

PR #1245, opened 2016-10-14
Debounce DCOSStore events

orlandohohmeier:
The DCOS change event is fired whenever one of the depended stores (e.g. Marathon) fires a change event; this results in an unnecessarily high amount of change events due to the unregular nature of those updates.

This PR introduces methods to debounce change event calls (max 4x per second), which significantly improves the overall performance.

_Please note that we besides reducing the number of change event calls also need to address the high number of network request as well as the large number of stacked calls._

**Before**
current

**After**
debounced mov

### DIFF
--- a/src/js/components/PodInstancesTable.js
+++ b/src/js/components/PodInstancesTable.js
@@ -246,6 +246,16 @@ class PodInstancesTable extends React.Component {
     return children;
   }
 
+  getDisabledItemsMap(instanceList) {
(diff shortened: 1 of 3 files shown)


philipnrmn:
No arguments with the implementation, but I'm not sure I like the approach of naive debounces. Can't we make the events more granular?

jfurrow:
This is awesome! Performance feels much better.

PR #1247, opened 2016-10-14
[Form Refactor]: The base structs and utils for the new form workflow

Poltergeist:
This does contain the structs and also the utils for the form refactoring. This is a basic implementation I hope it does explain how everything is working together.

```JS
let batch = new Batch();

// Create Reducer here with a object. These could also be deep nested but each 
// level does need to be generated.
let reducer = ReducerUtil.combineReducer({
  id: function (state = '', action) {
    if (action.type === 'SET' && action.path.join === 'id') {
      return action.value;
    }
    return state;
  },
  vip: function (state = '', action) {
    if (action.type === 'SET' && action.path.join === 'id') {
      this.id = id;
    }
    if (action.type === 'SET' && action.path.join === 'port') {
      this.port = port;
    }
    if (this.port != null && this.id != null) {
      return `${this.id}:${this.port}`;
    }
    return state;
  }
});

// Two events are getting added to the Batch.
batch.add(new SetAction(['id'], 'abcd');
batch.add(new SetAction(['port'], '8080');

let state = batch.reduce(reducer, {});

console.log(state); // output: {id: 'abcd', vip: 'abcd:8080'}
```

@mlunoe could you please look over this and let me know what you think so far.

Also Thanks to @orlandohohmeier @nLight @wavesoft for the input and support while pairing. It was a pleasure.


src/js/structs/SetAction.js
```diff
@@ -0,0 +1,9 @@
+import Action from './Action';
+
+class ActionSet extends Action {
```
orlandohohmeier: Mind the filename... Please either rename the class or the file.

src/js/structs/Batch.js
```diff
@@ -0,0 +1,88 @@
...
+ * }
+ *
+ *
+ */
+
+class Batch {
+
+  /**
+   * Initialize a new, empty immutable batch
+   *
```
orlandohohmeier: Let's remove this meaningless comment

src/js/structs/SetAction.js
orlandohohmeier: Not to sure if we actually need to create a set action...

src/js/structs/Batch.js
Poltergeist: removed.

src/js/structs/SetAction.js
Poltergeist: Sure, SetAction sounded better to me :-)

src/js/utils/ReducerUtil.js
```diff
@@ -0,0 +1,55 @@
...
+      function (reducer) {
+        return typeof reducers[reducer] === 'function';
+      }
+    );
+
+    // This is creating the context for this combined reducer.
+    let context = new WeakMap();
+
+    return function (state = {}, action, index = 0) {
+      let i = reducerKeys.length;
```
mlunoe: #nit let's use `index`

src/js/utils/ReducerUtil.js
```diff
@@ -0,0 +1,55 @@
...
+        let reducer = reducers[key];
+
+        if (index === 0 || !context.has(reducer)) {
+          context.set(reducer, {});
+        }
+
+        state[key] =
+          reducer.call(context.get(reducer), state[key], action, state);
+      }
+      return state;
```
mlunoe: #nit newline before `return`

src/js/utils/ReducerUtil.js
```diff
@@ -0,0 +1,55 @@
...
+   *
+   * @param {Object} reducers An object whose values correspond to different
+   * reducer functions that need to be combined into one.
+   *
+   * @returns {Function} A reducer function that invokes every reducer inside
+   *   the passed object, and builds a state object with the same shape.
+   */
+  combineReducers(reducers = {}) {
+    let reducerKeys = Object.keys(reducers).filter(
+      function (reducer) {
```
mlunoe: Could we use `reducerName` or `reducerKey` here to be more descriptive?

src/js/utils/ReducerUtil.js
```diff
@@ -0,0 +1,55 @@
...
+        let key = reducerKeys[i];
+
+        let reducer = reducers[key];
+
+        if (index === 0 || !context.has(reducer)) {
+          context.set(reducer, {});
+        }
+
+        state[key] =
+          reducer.call(context.get(reducer), state[key], action, state);
```
mlunoe: Could we get an explanatory comment about what is going on here?
The `state[key]` and `state` is confusing me

src/js/structs/Batch.js
```diff
@@ -0,0 +1,83 @@
...
+ *
+ *      return <Visualize data={data} />
+ *   }
+ * }
+ *
+ *
+ */
+
+class Batch {
+
```
mlunoe: #nit remove newline

src/js/structs/Batch.js
```diff
@@ -0,0 +1,83 @@
...
+   * NOTE: This is currently mutating the underlying batch array, but
+   *       this will change in the future, so use this function assuming
+   *       it will return the new batch you should operate upon!
+   *
+   * @param {Action} item
+   * @returns {Batch} Returns the instance for the batch for chaining.
+   */
+  add(item) {
+    this.batch.push(item);
+    return this.instance;
```
mlunoe: #nit newline before `return`

src/js/utils/__tests__/ReducerUtil-test.js
```diff
@@ -0,0 +1,163 @@
+const ReducerUtil = require('../ReducerUtil');
+
+const idReducer = function (state = '', action) {
+  if (action.path.join('') === 'id') {
```
mlunoe: Not sure understand the idea behind this part, having an array so you need to do `action.path.join('')`. Could you explain a use case where this is necessary/helpful?

Poltergeist: Well for the test this is probably to specific. 

My take on the idea for the path being an array was that you also could deep traverse into an object with logic if you have a path. This would also work if you path string ist something like: `container.docker.id` which you would split to an array. But then fields are not allowed to have a . in the name.

I will change this here to be a `String`.

src/js/utils/ReducerUtil.js
Poltergeist: I will remove the last state for now. And will add a comment.

src/js/utils/ReducerUtil.js
mlunoe: Then we should probably rename this to `reducerIndex`?

src/js/utils/__tests__/ReducerUtil-test.js
mlunoe: Alright, that makes sense. No, let's keep the array. I was just curious to the format.
I think it makes sense, so you can use functions like `action.path.includes('container')`.

src/js/structs/Action.js
```diff
@@ -0,0 +1,21 @@
+class Action {
```
Poltergeist: FormAction seems to be more fitting.

src/js/structs/Batch.js
```diff
@@ -0,0 +1,83 @@
...
+ *   }
+ * }
+ *
+ *
+ */
+
+class Batch {
+
+  constructor() {
+    const context = {instance:this, batch:[]};
```
Poltergeist: I'll add a comment

``
// This creates a context for the functions which have access to the batch variable and also 
// access to the instance. So `this` can be returned and functions could be chained.
``

This is also necessary, if one extends the batch and overwrites the add method to do some filtering.

src/js/structs/Batch.js
Poltergeist: done.

src/js/utils/ReducerUtil.js
Poltergeist: done.

src/js/utils/ReducerUtil.js
Poltergeist: done

src/js/utils/__tests__/ReducerUtil-test.js
Poltergeist: Okay.

src/js/structs/Batch.js
Poltergeist: done.

src/js/structs/FormAction.js
```diff
@@ -0,0 +1,21 @@
+class FormAction {
+  constructor(path, value, type = 'SET') {
+    Object.defineProperties(this, {
```
Poltergeist: I still plan to add a constant file for all the types. With those we could essentially do a simple validation. I just wanted to do that in a later step, as this PR is already introducing a couple of new things. 

So do you think I should introduce this also in this PR or rather in a later one?

jfurrow:
@Poltergeist This is awesome! Great test coverage too. Nice work!!!

Poltergeist:
@mlunoe @orlandohohmeier @MatApple 

I am trying to put up a to do list for this. If I missed something please add it here.

- [ ] Decide if the formAction should live in the structs or somewhere else.


Also in another PR I will be adding: 

- [ ] Type validation for formActions
- [ ] FormActionTypes constants

src/js/structs/Batch.js
Poltergeist: done.

src/js/structs/__tests__/Actions-test.js
```diff
@@ -0,0 +1,43 @@
+const FormAction = require('../FormAction');
+
+describe('FormAction', function () {
```
orlandohohmeier: Let's keep class, file and test name in sync.

src/js/utils/__tests__/ReducerUtil-test.js
```diff
@@ -0,0 +1,163 @@
+const ReducerUtil = require('../ReducerUtil');
+
+const idReducer = function (state = '', action) {
```
orlandohohmeier: Let's move this into the respective `describe` block

mlunoe: Or before each

mlunoe:
I think this looks good after you have addressed the outstanding comments

orlandohohmeier:
💯 ✨ This is great! Thanks for all your hard work.

PR #1252, opened 2016-10-15
Bind PureRenderMixin

MatApple:


philipnrmn:
I talked with @mlunoe about this in https://github.com/dcos/dcos-ui/pull/1125#pullrequestreview-1184322

I don't think it's actually necessary; do you know of a case where `shouldComponentUpdate` is invoked in the wrong scope?

PR #1260, opened 2016-10-18
Adjust the Mesos task path util

orlandohohmeier:
Align the util with the latest Mesos API advancements as the "pod" task data changed which resulted in wrong paths. It also no longer necessary to iterate over the list of tasks as we already have all the bits to construct the paths.

Closes DCOS-10807

### DIFF
--- a/src/js/utils/MesosStateUtil.js
+++ b/src/js/utils/MesosStateUtil.js
@@ -1,3 +1,4 @@
+import ExecutorTypes from '../constants/ExecutorTypes';
 import PodInstanceState from '../constants/PodInstanceState';
 import Util from './Util';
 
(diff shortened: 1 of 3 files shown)


greggomann:
I ran a one-off DC/OS build with this PR and it works great, thanks @orlandohohmeier! 🚢

src/js/utils/MesosStateUtil.js
```diff
@@ -225,26 +225,25 @@ const MesosStateUtil = {
...
         .every(function (executor) {
-          // Find app/framework executor
-          if (executor != null &&
-              (executor.id === executorID || executor.id === taskID)) {
+          if (executor.id === executorID || executor.id === taskID) {
+
+            // Use the executor task path construct if it's a "pod" / TaskGroup
+            // executor (type: DEFAULT), otherwise fallback to the default
+            // app/framework behavior.
+            if (executor.type === 'DEFAULT') {
```
rcorral: Not a blocker for PR, but would be nice if this `DEFAULT` was extracted to a constant.

wavesoft:
Looks good to me! I also share the same feeling with @rcorral , but I don't really think there is going to another type any time soon.

Thanks for addressing this so quickly! 💯 ✨

orlandohohmeier:
@rcorral @wavesoft I've introduced an executor type constant, hope that's what you guys had in mind.

PR #1266, opened 2016-10-19
[CNVS Upgrade] Remove spaces before variable values

jfurrow:
This PR removes the extra space characters in all of the variables files.


ashenden:
FWIW I think this is a bad call.  If it's the majority opinion, I'll ride with it.  But, generally speaking, the move to remove spaces results in a reader having to scan a list of variables (potentially 100s) both vertically and horizontally.  By "column-izing" our variables, it makes scanning almost exclusively vertical.

If it gets us closer to using variables more, I'll concede -- reluctantly.

ashenden:
Excerpt from a convo with @jfurrow about this `give me an IDE plugin that artificially does the column visualization for me and I'll STFU`

PR #1269, opened 2016-10-19
[Services 1] Purifying ServiceDetail pages and direct decendents

MatApple:
Purifies ServiceDetail and all immediate tabs within ServiceDetail (tasks, configuration, debug).

Updates TaskView and TaskTable to be wrapped in a container, which assumes responsibility for the merging of data attributes from Marathon and Mesos. This merging was previously handled in the TaskTable itself and will eventually be handled on the GraphQL server. This new container is then used in NodeDetail also.


plugins/services/src/js/tasks/ServiceTasksContainer.js
```diff
@@ -0,0 +1,75 @@
...
+    if (hasError) {
+      return <RequestErrorMsg />;
+    }
+
+    let tasks = MesosStateStore.getTasksByService(this.props.service);
+
+    return (
+      <TasksContainer
+        tasks={tasks}
+        label="Instance" />
```
MatApple: ✔️

PR #1271, opened 2016-10-19
DCOS-10421: Adjust the Spec health check getter

orlandohohmeier:
Make sure that the `ApplicationSpec` health check getter returns "cloned" objects so that no one is accidentally mutating them. This change resolves an issue where the service form cancel button isn't working due to unwanted mutations.

Closes DCOS-10421

### DIFF
--- a/src/js/structs/ApplicationSpec.js
+++ b/src/js/structs/ApplicationSpec.js
@@ -48,7 +48,13 @@ module.exports = class ApplicationSpec extends ServiceSpec {
   }
 
   getHealthChecks() {
-    return this.get('healthChecks');
+    let healthChecks = this.get('healthChecks') || [];
+
+    // "Clone" health checks to ensure that no one is accidentally altering the
+    // properties.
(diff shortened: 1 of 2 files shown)


rcorral:
This seems like it's treating the symptom and not the problem, is that right?
Any chance we can do this cloning before the mutation occurs?

orlandohohmeier:
It depends on the perspective, either we have the struct making sure that no one is accidentally mutating it's data (immutable data structures) or we make sure that we're not mutating the struct when ever we're using it. I opted for the first but could easily fix this in a different place if you want me to. What do you recommend?

rcorral:
Talked with @orlandohohmeier offline. He's convinced me this is a good approach 👍

wavesoft:
I would add the same concern with @rcorral , because the data from the structs shouldn't be treated as mutable (because even if we switch to an immutable solution the mutation will cause an exception), but if you guys have discussed this through, I trust your judgement 👍 

Thanks @orlandohohmeier , LGTM 💯 ✨

PR #1272, opened 2016-10-19
[Services 2] Prepare Pods for Relay

MatApple:
Prepares Pods for relay integration.


orlandohohmeier:
@MatApple I've rebased the branch for you and changed the base branch to master

plugins/services/src/js/pod-detail/PodDetail.js
```diff
@@ -44,94 +37,70 @@ class PodDetail extends mixin(TabsMixin) {
   }
 
   handleActionDestroy() {
-    this.setState({
-      activeActionDialog: PodActionItem.DESTROY
-    });
-
+    const {pod} = this.props.pod;
+    this.context.modalHandlers.deleteService({pod});
```
orlandohohmeier: Why are we relying on the context to handle the actions?

plugins/services/src/js/pod-instances/PodInstancesContainer.js
```diff
@@ -0,0 +1,227 @@
...
+    });
+  }
+  /**
+   * Sets the pending actionType to false in state which will in turn be
+   * pushed down to children via props. Can optional set an error for the
+   * actionType
+   * @param  {String} actionType
+   * @param  {Any} error
+   */
+  unsetPendingAction(actionType, error = null) {
```
orlandohohmeier: Let's move all these methods to handle modals into some kind of a modal service instead of passing these actions down using context and parameter. The `PodInstancesContainer` shouldn't need to handle all of this.

jfurrow:
This seems to work well! I pushed a small style change but otherwise I think it's good to go. :ship:

MatApple:
@orlandohohmeier thanks for rebasing this for me! Lets get this in and I'll address your comments in the coming week. I'll be changing the way modals are handled and I'd like to chat about that some more.

PR #1276, opened 2016-10-20
Adding timezone-mock package

wavesoft:
This PR adds the `timezone-mock` package in the development dependencies in order for the tests to be able to mock timezone when `Date.toLocaleString()` is used.


wavesoft:
I just realised that this module is quite bad and we shouldn't use it. I will provide a utility function in the `JestUtil`.

PR #1278, opened 2016-10-20
JestUtil: Adding mockTimezone

wavesoft:
This PR introduces the `JestUtil. mockTimezone` function that can be used to provide the default timezone to be used by the `Date` object.

(Since unfortunately `jest.clock` does not provide this functionality)

### DIFF
--- a/src/js/utils/JestUtil.js
+++ b/src/js/utils/JestUtil.js
@@ -31,6 +31,10 @@ class RouterStub {
   static setRouteComponentAtDepth() {}
 }
 
+// Default prototype functions when mocking timezone
+const defaultGetTimezoneOffset = Date.prototype.getTimezoneOffset;
+const defaultToLocaleString = Date.prototype.toLocaleString;
+
 const JestUtil = {
   renderAndFindTag(instance, tag) {
     var result = TestUtils.renderIntoDocument(instance);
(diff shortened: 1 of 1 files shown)


src/js/utils/JestUtil.js
```diff
@@ -105,6 +109,32 @@ const JestUtil = {
...
+    let timezoneOffset = (
+          new Date(date.toLocaleString(undefined, {timeZone: 'UTC'}))
+        - new Date(date.toLocaleString(undefined, {timeZone: timezone}))
+      ) / 1000 / 60;
+
+    /* eslint-disable no-extend-native */
+    Date.prototype.getTimezoneOffset = function () {
+      return timezoneOffset;
+    };
+    Date.prototype.toLocaleString = function (locale=undefined, options={}) {
```
jfurrow: #nit typically we put spaces in between the assignment operator (e.g. `locale = undefined, options = {}`), I'm surprised the linter isn't configured for this. So, I guess you're free to do as you please here!

jfurrow:
I had a minor #nit, but as the linter isn't configured for this, it's not too important to me. Approved!

src/js/utils/JestUtil.js
mlunoe: Yeah, I thought we had a rule for this 🤔

mlunoe:
Feel free to ⛵️ as soon as that one comment is done

PR #1279, opened 2016-10-20
[CNVS Upgrade] Fix up the nodes view

jfurrow:
This PR fixes up the Nodes view:
* I fixed the double padding between the panel header and the panel content in the `ResourceBarChart`
* I fixed the dark graph colors in the nodes grid view
* I moved the width and margin properties of the nodes grid sidebar to the sidebar itself rather than random children (it was confusing)
* I fixed some erroneously defined transition properties, so toggling the task list will transition as intended
* I replaced the old, yucky, native `input` checkbox with the new, yummy, fancy checkbox from reactjs-components.

The `FilterHeadline` margins seem a little tight, but that's everywhere in the app, so I'll address that in a future PR.

### DIFF
--- a/plugins/nodes/src/js/components/NodesGridView.js
+++ b/plugins/nodes/src/js/components/NodesGridView.js
@@ -1,4 +1,5 @@
 import classNames from 'classnames';
+import {Form} from 'reactjs-components';
 import PureRender from 'react-addons-pure-render-mixin';
 import React from 'react';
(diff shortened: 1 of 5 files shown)


philipnrmn:
thank you for this tasty 🍰

rcorral:
Thanks for all the clean up @jfurrow!

PR #1283, opened 2016-10-24
DCOS-9773: Stop assuming RequestUtil has a default timeout

jfurrow:
Depends on https://github.com/dcos/mesosphere-shared-reactjs/pull/24

We no longer need to set "long enough" timeouts on endpoints where 2 seconds is not enough, because RequestUtil no longer imposes default timeouts. The only request that we want to time out after 2 seconds is the request made in `MesosStateSummary`.


rcorral:
@jfurrow merged the downstream PR :)

rcorral:
ping @jfurrow

nLight:
So dcos/mesosphere-shared-reactjs#24 got merged, thus I removed `hold merge` label

jfurrow:
@nLight @rcorral Thanks! Sorry, I should have taken care of this sooner.

PR #1290, opened 2016-10-25
Validation of batch types

Poltergeist:
This is adding the support of type validation in the `Transaction`s 

I choose to use `Symbols` as values over `Strings` since they are unique and can not produce bugs because we just pass in Strings. 

If a invalid type is added a TypeError will be thrown, which should give enough information.

About the `DEL` type, this type is currently just in there for testing purposes it might be that this will be removed/replaced at a later time.

### DIFF
--- a/src/js/structs/Transaction.js
+++ b/src/js/structs/Transaction.js
@@ -1,9 +1,9 @@
-import FormActionTypes from '../constants/TransactionTypes';
+import TransactionTypes from '../constants/TransactionTypes';
 
-const validTypes = Object.values(FormActionTypes);
+const validTypes = Object.values(TransactionTypes);
 
 class Transaction {
-  constructor(path, value, type = FormActionTypes.SET) {
+  constructor(path, value, type = TransactionTypes.SET) {
     if (!validTypes.includes(type)) {
       throw new TypeError(`Only the following types are allowed: ${validTypes}`);
     }
--- a/src/js/structs/Transaction.js
+++ b/src/js/structs/Transaction.js
@@ -1,5 +1,12 @@
+import FormActionTypes from '../constants/TransactionTypes';
+
+const validTypes = Object.values(FormActionTypes);
+
 class Transaction {
-  constructor(path, value, type = 'SET') {
+  constructor(path, value, type = FormActionTypes.SET) {
+    if (!validTypes.includes(type)) {
+      throw new TypeError(`Only the following types are allowed: ${validTypes}`);
+    }
     Object.defineProperties(this, {
       path: {
         value: path,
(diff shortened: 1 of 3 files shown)


src/js/structs/Transaction.js
```diff
@@ -1,5 +1,12 @@
+import FormActionTypes from '../constants/TransactionTypes';
```
MatApple: Maybe we should make this variable `TransactionTypes` just to keep things consistent.

src/js/structs/Transaction.js
```diff
@@ -1,5 +1,12 @@
...
+
+const validTypes = Object.values(FormActionTypes);
+
 class Transaction {
-  constructor(path, value, type = 'SET') {
+  constructor(path, value, type = FormActionTypes.SET) {
+    if (!validTypes.includes(type)) {
+      throw new TypeError(`Only the following types are allowed: ${validTypes}`);
```
MatApple: The `validTypes` in the message will be Symbols. Perhaps we should output the `Object.keys(FormActionTypes)`.

src/js/structs/__tests__/Transaction-test.js
```diff
@@ -1,15 +1,20 @@
...
 describe('Transaction', function () {
   describe('#constructor', function () {
     it('should have the type SET', function () {
       const transaction = new Transaction(0, 0);
-      expect(transaction.type).toEqual('SET');
+      expect(transaction.type).toEqual(TransactionTypes.SET);
+    });
+
+    it('should throw a Error for random type', function () {
+      expect(()=>new Transaction(0, 0, 'DEL')).toThrowError(TypeError);
```
MatApple: Lets space out the fat arrow notation `() => new ..`

src/js/structs/__tests__/Transaction-test.js
```diff
@@ -1,15 +1,20 @@
...
     });
 
-    it('should have the type DEL', function () {
-      const transaction = new Transaction(0, 0, 'DEL');
-      expect(transaction.type).toEqual('DEL');
+    it('should accept SET constant', function () {
+      expect(()=>new Transaction(0, 0, TransactionTypes.SET)).not.toThrowError();
```
MatApple: and here

src/js/structs/Transaction.js
Poltergeist: The error message right now is:
```
TypeError: Only the following types are allowed: Symbol(SET),Symbol(DEL)
```
I am okay with changing it to `SET, DEL, ...` , if that is still clear enough that only the predefined values are valid types and not strings.

src/js/structs/__tests__/Transaction-test.js
Poltergeist: done

src/js/structs/__tests__/Transaction-test.js
Poltergeist: done

PR #1294, opened 2016-10-27
Add Parser Util

Poltergeist:
This adds a ParserUtil which contains a function named combineParsers with this function it is possible to create a composable set of parsers which will return a flat list of items. In our use case this will be a set of transactions.

The util can be used like this:

```JS
const parser = ParserUtil.combineParsers([
  function (state) {
    return {
      type: SET,
      path: 'id',
      value: state.id
  },
  function (state) {
    return {
      type: SET,
      path: 'cmd',
      value: state.cmd
    };
  }
]);

const appDefinition = {
  id: 'test',
  cmd: 'sleep 100;'
};

console.log(parser(appDefinition));

// Will resolve into: 
// [
//   {
//     type: SET,
//     path: 'id',
//     value: 'test'
//   },
//   {
//     type: SET,
//     path: 'cmd',
//     value: 'sleep 100;'
//   }
// ]);

```

A more complex example with some nesting is also possible:

```JS
const containerParser = function (state) {
  if (state.container != null && state.container.docker != null) {
    return ParserUtil.combineParsers([
      function (state) {
        return {
          type: SET,
          path: 'container.docker.image',
          value: state.container.docker.image
        };
      }
    ])(state);
  }
  return [];
};

let parser = ParserUtil.combineParsers([
  function (state) {
    return {
      type: SET,
      path: 'id',
      value: state.id
    } 
  },
  containerParser
]);

let state = {
  id: 'docker',
  container: {
    docker: {
      image: 'nginx'
    }
  }
};

console.log(parser(state));

// Will result into:
// [
//   {
//     type: SET,
//     path: 'id',
//     value: 'docker'
//   },
//   {
//     type: SET,
//     path: 'container.docker.image',
//     value: 'nginx'
//   }
// ]
```

This will enable us to have a tree structure of parsers based on conditions. But the result will always be a flat list.

### DIFF
--- a/src/js/utils/ParserUtil.js
+++ b/src/js/utils/ParserUtil.js
@@ -1,0 +1,29 @@
+module.exports = {
+  combineParsers(parsers = []) {
+    parsers = parsers.filter(
+      function (parser) {
+        return typeof parser === 'function';
+      }
+    ).reverse();
+
+    return function (state = {}) {
+      let index = parsers.length;
+
+      const transactionLog = [];
+
+      while (--index >= 0) {
+        let parser = parsers[index];
+
+        let transaction = parser(state);
+
+        if (transaction instanceof Array) {
+          transactionLog.push(...transaction);
+        } else {
+          transactionLog.push(transaction);
+        }
+      }
+
+      return transactionLog;
+    };
+  }
+};
--- a/src/js/structs/Transaction.js
+++ b/src/js/structs/Transaction.js
@@ -11,15 +11,18 @@ class Transaction {
     Object.defineProperties(this, {
       path: {
         value: path,
-        writable: false
+        writable: false,
+        enumerable: true
       },
       value: {
         value,
-        writable: false
+        writable: false,
+        enumerable: true
       },
       type: {
         value: type,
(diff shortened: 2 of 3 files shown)


src/js/utils/ParserUtil.js
```diff
@@ -0,0 +1,29 @@
+module.exports = {
+  combineParsers(parsers = []) {
+    parsers = parsers.filter(
+      function (parser) {
+        return typeof parser === 'function';
```
wavesoft: Why this test here? Isn't the user of this function always supposing to pass functions as arguments?

Poltergeist: This is just to make sure that we are not executing something which is not a function. We will get an array which should contain functions.

Poltergeist:
The combineParses have their use-case right now for transactions but, they are fairly free defined so you also could create something totally different.

wavesoft:
I am not yet fully convinced that this is indeed so generic, but we can always rename it in the future if we really have that problem.

Good work sir, LGTM 🚢 💯 ✨

PR #1295, opened 2016-10-27
Fixed undefined OS and default to linux. refs:open-DCOS-477

rcorral:


src/js/components/modals/CliInstallModal.js
```diff
@@ -18,9 +18,12 @@ class CliInstallModal extends React.Component {
   constructor() {
     super(...arguments);
 
-    this.state = {
-      selectedOS: browserInfo().os
-    };
+    let selectedOS = browserInfo().os;
+    if (Object.values(osTypes).includes(selectedOS) === false) {
```
MatApple: should we just do `!` instead of `=== false`?

MatApple:
`CliInstallModal-test.js` is failing.

src/js/components/modals/CliInstallModal.js
MatApple: This is fine but syntactically we usually don't do explicit false equality checks - usually just use `!` prefix. Just a #nit.

src/js/components/modals/CliInstallModal.js
```diff
@@ -18,9 +18,12 @@ class CliInstallModal extends React.Component {
   constructor() {
     super(...arguments);
 
-    this.state = {
-      selectedOS: browserInfo().os
-    };
+    let selectedOS = browserInfo().os;
```
rcorral: I know it's confusing, but this works. If you do `linux` lowercase it won't work.

MatApple: `browserInfo().os` returns a lowercase string right? Because we check `Object.values(osTypes).includes(selectedOS)` here which would only match if it were lowercase. So something is wrong here - either L21 should be mapped to uppercase or we need to lowercase `selectedOS` before we do `Object.values(osTypes).includes(selectedOS)`.

orlandohohmeier:
Any update? Would love to land this fix.

src/js/components/modals/CliInstallModal.js
rcorral: @MatApple you're right. Fixed the issue. Not sure how I thought it worked before when I tested it.

src/js/components/modals/CliInstallModal.js
rcorral: potato potatoe! I removed the explicit check for boolean.

MatApple:
LGTM 🚢

PR #1297, opened 2016-10-27
Introduce a new pre-install script

orlandohohmeier:
Add a `pre-install` script that is install/linking git hooks in a non-destructive way.


wavesoft:
LGTM! ✨ 💯

scripts/utils/git
```diff
@@ -0,0 +1,50 @@
...
+
+# Install/link GIT hooks
+# Example: install_hooks "pre-commit pre-push"
+install_hooks() {
+  # Check if .git/hooks directory exists
+  if [ ! -d ".git/hooks" ]
+  then
+    warning "Couldn't install hooks as the directory ${NC}.git/hooks/${RED}"\
+              "doesn't exists"
+    return
```
nLight: We could also create`hooks` directory if `.git` exists

nLight:
Please consider my comment and merge it, good job!

scripts/utils/git
orlandohohmeier: As far as I know there is by default always a `.git/hooks` directory, thus I decided not to add the needed extra command to create that directory if it doesn't exists. Do you think we should add this case?

nLight: I'm fine wth that assumption

nLight:
Looks like I can't delete the branch @orlandohohmeier

nLight:
Ok, it became possible and I did it. Interesting...

PR #1298, opened 2016-10-27
Update the pre-commit hook

orlandohohmeier:
Update the `pre-commit` to only lint staged files.

PR #1300, opened 2016-10-27
Add a pre-push hook

orlandohohmeier:
---

⚠️ This PR depends on changes that will get introduced with #1298 

---

Add a pre-push hook to run the unit test before pushing stuff to origin.

### DIFF
--- a/scripts/pre-push
+++ b/scripts/pre-push
@@ -1,0 +1,11 @@
+#!/usr/bin/env bash
+
+SCRIPT_PATH="$(dirname "$0")/$(dirname "$(readlink "$0")")"
+
+# Import utils
+source ${SCRIPT_PATH}/utils/message
+
+title "Run pre-push hook..."
+
+header "Run test..."
+npm run test -- -b
--- a/scripts/pre-push
+++ b/scripts/pre-push
@@ -7,5 +7,11 @@ source ${SCRIPT_PATH}/utils/message
 
 title "Run pre-push hook..."
 
+header "Run lint..."
+npm run lint &&
+if [ "undefined" != `npm config get externalplugins` ]; then
+  npm run lint-plugins || exit $?
+fi
+
 header "Run test..."
 npm run test -- -b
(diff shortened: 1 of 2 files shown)


scripts/pre-push
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+
+source ${0%/*}/utils/message
+
+title "Run pre-push hook..."
+
+# Run the complete suite of tests
+npm run test -- -b
```
orlandohohmeier: We should actually run the linting as well to ensure that we introduce errors with a rebase...

orlandohohmeier:

wavesoft:
Looks great, I am OK with this, but it takes an awfully long time on every push. 
My concern is that sooner or later, people will start using `skip validation`...

MatApple:
@orlandohohmeier can we close this PR now?

orlandohohmeier:
@MatApple I'm fine with closing it, but would vote to merge it and risk that people will start skipping the validation (`--no-verify`) as running it every now and then is already better than never. 
We also have the commitment to improve the test performance which will improve the `pre-push` performance as well...

MatApple:
I'm with you on that @orlandohohmeier - thought since https://github.com/dcos/dcos-ui/issues/1340 merged we are good. But I think it's good to have this as too. We can at least trial it to see if it isn't too disruptive.

orlandohohmeier:
@MatApple @wavesoft There is a rather hacky way to get the base branch which we then could us to only test "changed" files..

```
base_branch=$(git show-branch | sed "s/].*//" | grep "\*" |
  grep -v "$(git rev-parse --abbrev-ref HEAD)" | head -n1 | sed "s/^.*\[//")
```

MatApple:
Works for me! Hacky is Ok here :)

nLight:
@orlandohohmeier let's rebase and merge it!

mlunoe:
Rebased, tested and labelled!

PR #1301, opened 2016-10-28
List.js: Adding combine function to merge and remove duplicates

wavesoft:
This PR introduces the `combine` function to the `List` that creates a new list using the items from itself and the passed one, discarding all possible duplicates.

### DIFF
--- a/src/js/structs/List.js
+++ b/src/js/structs/List.js
@@ -1,4 +1,5 @@
 import Item from './Item';
+import ObjectUtil from '../utils/ObjectUtil';
 import StringUtil from '../utils/StringUtil';
 
 /**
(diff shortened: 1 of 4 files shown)


src/js/structs/List.js
```diff
@@ -66,6 +66,26 @@ module.exports = class List {
...
+   * Return a new list with the items currently present in this list and
+   * the list on the first argument, without duplicates.
+   *
+   * @param {List} list - The list to combine with this list
+   * @returns {List} Returns a new list with the combined items
+   */
+  combine(list) {
+    let myItems = this.getItems();
+    let newItems = list.reduceItems(function (items, item) {
+      if (myItems.indexOf(item) === -1) {
```
nLight: When I see multiple array lookups I always want to use a hash (Object) to get unique values.

wavesoft: Yeah, I thought about ways to optimise this, but Object won't work, since `item`s are objects and object keys can only be strings. A `Map` could be used for what you suggest, however I am not sure it's widely supported yet.

wavesoft: Well, as a matter of fact `O(N²)` for such a core function, especially when it's used by the filters it's a bad idea. I will actually suggest a much faster yet a bit more complicate to read solution with `O(2N)` time.

wavesoft:
Ok, new version is here. Ugly, but faster. Comments please :nerd_face:

nLight:
@wavesoft I'm proud of you! 👍

wavesoft:
Allrighty, now it should be cleaner to read 😃

src/js/structs/List.js
```diff
@@ -74,15 +75,35 @@ module.exports = class List {
...
-      return items;
-    }, []);
+    // Instead of using the O(n²) approach of iterating over `newItems` and
+    // checking if each item exist on `myItems`, we use a trick that brings the
+    // complexity down to O(n) (well, O(2n) to be precise).
+
+    // We first generate a unique object for the current combine operation
+    // and we mark all the objects of `myItems` with it, declaring them as used.
+    let comparisonTag = {};
+    for (let i=0, l=myItems.length; i<l; ++i) {
```
orlandohohmeier: Let's use more descriptive names for `i`, `j` and `l` as you use them not only within the loop and it probably helps people to better understand this method.

orlandohohmeier:
Nit aside this is great - really like how you optimize the combine method 💯 ✨ Great job!

MatApple:
Nice work @wavesoft - clever solution. `O(n + m)` space and `O(n + m)` time is probably as good as this gets 👍

MatApple:
Lets ship it after Orlando's comment is addressed.

orlandohohmeier:
Let's merge this once the test pass. Great job! 💯 🚢 🚀

PR #1305, opened 2016-10-28
Fix loading indicators! 🕸 

jfurrow:
This PR removes the wrapping `div` that we used in conjunction with the `Loader` component almost everywhere. We don't need it because it adds extra margin where it shouldn't in 99% of the cases.

In the future, if we do decide to add margin or padding, we can do so in the `Loader` component itself and it'll just work everywhere :)

### DIFF
--- a/plugins/nodes/src/js/components/NodesGridView.js
+++ b/plugins/nodes/src/js/components/NodesGridView.js
@@ -49,7 +49,7 @@ var NodesGridView = React.createClass({
     }
 
     var loadingClassSet = classNames({
-      'hidden': hasLoadingError
+      hidden: hasLoadingError
     });
 
     return (
--- a/plugins/nodes/src/js/components/NodesTable.js
+++ b/plugins/nodes/src/js/components/NodesTable.js
@@ -64,11 +64,7 @@ var NodesTable = React.createClass({
     let requestReceived = this.props.receivedNodeHealthResponse;
 
     if (!requestReceived) {
-      return (
-        <Loader
-          innerClassName="loader-small"
-          type="ballBeat" />
-      );
+      return <Loader innerClassName="loader-small" type="ballBeat" />;
     }
 
     let health = node.getHealth();
--- a/plugins/nodes/src/js/components/NodesTable.js
+++ b/plugins/nodes/src/js/components/NodesTable.js
@@ -64,7 +64,7 @@ var NodesTable = React.createClass({
     let requestReceived = this.props.receivedNodeHealthResponse;
 
     if (!requestReceived) {
-      return <Loader innerClassName="loader-small" type="ballBeat" />;
+      return <Loader size="small" type="ballBeat" />;
     }
 
     let health = node.getHealth();
--- a/plugins/services/src/js/components/MesosLogView.js
+++ b/plugins/services/src/js/components/MesosLogView.js
@@ -338,7 +338,7 @@ class MesosLogView extends mixin(StoreMixin) {
     return (
       <Loader
         className="pod pod-tall-bottom flush-right flush-left"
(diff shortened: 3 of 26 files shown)


plugins/nodes/src/js/components/NodesGridView.js
```diff
@@ -49,7 +49,7 @@ var NodesGridView = React.createClass({
     }
 
     var loadingClassSet = classNames({
-      'hidden': hasLoadingError
+      hidden: hasLoadingError
```
jfurrow: The boyscout in me did this while I was editing this for testing purposes... I'll leave this unless somebody objects!

plugins/nodes/src/js/components/NodesTable.js
```diff
@@ -64,11 +64,7 @@ var NodesTable = React.createClass({
     let requestReceived = this.props.receivedNodeHealthResponse;
 
     if (!requestReceived) {
-      return (
-        <Loader
-          innerClassName="loader-small"
-          type="ballBeat" />
-      );
+      return <Loader innerClassName="loader-small" type="ballBeat" />;
```
jfurrow: More scout activity...

plugins/services/src/js/components/MesosLogView.js
```diff
@@ -336,11 +336,10 @@ class MesosLogView extends mixin(StoreMixin) {
 
     // Show loader since we will start a request for more logs
     return (
-      <div className="pod flush-top">
-        <Loader
-          innerClassName="loader-small"
-          type="ballSpinFadeLoader" />
-      </div>
+      <Loader
+        className="pod pod-tall-bottom flush-right flush-left"
```
jfurrow: This loader was actually offset to the right due to some extra margin on the left. I also added `pod-tall-bottom` because the loader didn't appear centered vertically without it, due to some extra padding on the container of this element.

src/js/components/Loader.js
```diff
@@ -45,15 +45,15 @@ let typeMap = {
 class Loader extends React.Component {
   getDivs(length) {
     return Array.from({length}).map(function (_, index) {
-      return <div key={index} />;
+      return <div className="loader-element" key={index} />;
```
jfurrow: I added a class to these elements to target them via CSS, rather than using the `> div` selector.

src/js/components/Loader.js
```diff
@@ -45,15 +45,15 @@ let typeMap = {
...
     });
   }
 
   render() {
     let {className, innerClassName, type} = this.props;
     let config = typeMap[type] || typeMap.ballScale;
     let classes = classNames(
-      'loader horizontal-center vertical-center',
+      'loader horizontal-center',
```
jfurrow: I removed vertical-center because it wasn't actually doing anything for us.

plugins/nodes/src/js/components/NodesTable.js
orlandohohmeier: Do we actually need the `innerClassName`? Would love to get rid of these props in favor of more descriptive props like `size="small"`...

orlandohohmeier:
Great job! 💯 🎆

plugins/nodes/src/js/components/NodesTable.js
jfurrow: Sorry for the delay. This is done now!

jfurrow:
_mesosphere-ci_ retest this please

orlandohohmeier:
Perfect! 💯  🚢  🇮🇹

plugins/nodes/src/js/components/NodesTable.js
orlandohohmeier: np

PR #1316, opened 2016-11-01
[Service create form 1 of 2] Introduce fields: input and textarea

mlunoe:
This PR introduces two fields; FieldInput and FieldTextarea.

They are both fairly simple; they do not hold state and are only concerned with rendering

Example usage:
```js
class ServiceFormSection extends Component {
  getIDHelpBlock() {
    return (
      <span>
        {"Include the path to your service, if applicable. E.g. /dev/tools/my-service. "}
        <a href="https://mesosphere.github.io/marathon/docs/application-groups.html" target="_blank">
          More information
        </a>.
      </span>
    );
  }

  render() {
    let {data, errors} = this.props;

    return (
      <form className="flex row" onChange={this.handleChange}>
        <FormGroup className="column-8">
          <FieldLabel>
            SERVICE NAME <span className="text-danger">*</span>
          </FieldLabel>
          <FieldInput
            name="id"
            type="text"
            value={data.id} />
          <FieldHelp>{this.getIDHelpBlock()}</FieldHelp>
          <FieldError>{errors.id}</FieldError>
        </FormGroup>

        <FormGroup className="column-4">
          <FieldLabel>
            INSTANCES <span className="text-danger">*</span>
          </FieldLabel>
          <FieldInput
            name="instances"
            min={0}
            type="number"
            value={data.instances} />
          <FieldError>{errors.instances}</FieldError>
        </FormGroup>
      </form>
    );
  }
}
```

### DIFF
--- a/src/js/components/form/FieldInput.js
+++ b/src/js/components/form/FieldInput.js
@@ -55,6 +55,12 @@ class FieldInput extends React.Component {
     );
   }
 
+  getElement(attributes) {
+    return (
+      <input {...attributes} />
+    );
+  }
+
   render() {
     let {
       error,
@@ -70,7 +76,7 @@ class FieldInput extends React.Component {
     return (
       <div className={classes}>
         {this.getLabel()}
-        <input {...attributes} />
+        {this.getElement(attributes)}
         {this.getHelpBlock()}
         {this.getErrorMsg()}
       </div>
--- a/src/js/components/form/FieldInput.js
+++ b/src/js/components/form/FieldInput.js
@@ -1,0 +1,117 @@
+import classNames from 'classnames/dedupe';
+import deepEqual from 'deep-equal';
+import React from 'react';
+
+import Util from '../../utils/Util';
+
+class FieldInput extends React.Component {
+  shouldComponentUpdate(nextProps) {
+    return !deepEqual(this.props, nextProps);
+  }
+
+  getErrorMsg() {
+    let {helpBlockClass, error} = this.props;
+
+    if (!error) {
+      return null;
+    }
+
+    return (
+      <span className={classNames(helpBlockClass)}>
+        {error}
+      </span>
+    );
+  }
+
+  getHelpBlock() {
+    let {helpBlock, helpBlockClass} = this.props;
+
+    if (!helpBlock) {
+      return null;
+    }
+
+    return (
+      <p className={classNames(helpBlockClass)}>
+        {helpBlock}
+      </p>
+    );
+  }
+
+  getLabel() {
+    let {label, labelClass} = this.props;
+
+    if (!label) {
+      return null;
+    }
+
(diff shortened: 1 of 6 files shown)


src/js/components/form/FieldTextarea.js
```diff
@@ -0,0 +1,36 @@
...
+
+    let classes = classNames(
+      {[formGroupErrorClass]: !!error},
+      formGroupClass
+    );
+
+    return (
+      <div className={classes}>
+        {this.getLabel()}
+        <textarea {...attributes} />
```
Poltergeist: Let's do a this.getElement(attributes) helper function. Which we do overwrite in the FieldTextarea.

so 
```JS
getElement(attributes) {
  return (
    <textarea {...attributes} />
  );
}
```

would be used in render as `{this.getElement(attributes)}` 

and we only need to overwrite the getElement function rather then the `render`, as far as I see this only the element is the only difference in the render function compared with the FieldInput. we are even omitting the props based on `FieldInput.propTypes`.

Another solution, which is more composable, is that we have a wrapping component which does contain all the help block, error block and labels and does also await children which then could be a input or a textarea

e.g.:
```HTML
<FormElementWrapper>
  <input type="text" />
</FormElementWrapper>

Comment (mlunoe) [inline src/js/components/form/FieldTextarea.js]: Done!

**Comment by mlunoe** [comment]
@Poltergeist All comments addressed

File: src/js/components/form/FieldInput.js [hunk 2]
```diff
@@ -0,0 +1,123 @@
...
+
+  getErrorMsg() {
+    let {helpBlockClass, error} = this.props;
+
+    if (!error) {
+      return null;
+    }
+
+    return (
+      <span className={classNames(helpBlockClass)}>
```
orlandohohmeier: Why are we using the `helpBlockClass` to style the error message?

orlandohohmeier:
I'm worrying that by modeling the form element by inheritance and providing a large chunk for configuration isn't really an improvement over the component that we have today. I'd love to see us composing those form elements from a wrapping `FormElement` and a set components like an error, help, and the actual form element (e.g. input) similar to the tab components. 

```
<FormElement>
   <FormElementLabel>Email</FormElementLabel>
   <FormElementError>Oh shit something went wrong</FormElementError>
   <FormElementHelp>Just enter your email..</FormElementHelp>
   <input type="email">
</FormElement>
```

Anyway, I'm ok to merge the PR as it, to unblock you as the code looks good to me.

mlunoe:
@orlandohohmeier and I talked offline and decided to get this in to unblock and then I will work with @jfurrow to create a follow-up PR to make the form fields more composable!

jfurrow:
Nice work @mlunoe!

orlandohohmeier:
Great job! 💯 ✨

PR #1319, opened 2016-11-02
Upgrade Jest to v16

philipnrmn:
We're currently using Jest v12. 

Jest v15 introduced a [raft of features](https://facebook.github.io/jest/blog/2016/09/01/jest-15.html) including significant performance improvements. These were mostly to do with Jest's configuration, for example automocking is now off by default. 

Since our tests were written with old Jest in mind, I've re-enabled these defaults and this PR does not represent a performance improvement. However, it does allow us to incrementally fix automocks and timers and hopefully improve test run times in the long term. In the meantime there's more transparent test run states and nicer `--watch` functionality to enjoy.


philipnrmn:
Tests are passing, the failure is `gulp checkDependencies`.

package.json
```diff
@@ -85,8 +85,8 @@
     "html-webpack-plugin": "2.16.0",
     "image-webpack-loader": "1.7.0",
     "jasmine-reporters": "2.1.1",
-    "jest-cli": "12.1.1",
-    "jest-webpack-alias": "2.3.0",
+    "jest-cli": "^16.0.2",
+    "jest-webpack-alias": "^3.3.0",
```
rcorral: @philipnrmn can you remove the `^`.
This way we avoid the "It works on my machine" thing.

wavesoft:
Love the way jest 16 reports the progress 😄 
LGTM! ✨  💯

PR #1320, opened 2016-11-02
Batch improvements

mlunoe:
This PR adds improvements to the Batch to keep the amount of events minimal, to optimize for reducer iterations.

### DIFF
--- a/src/js/structs/Batch.js
+++ b/src/js/structs/Batch.js
@@ -11,7 +11,7 @@
  * @example <caption>Using Batch</caption>
  * class Component extends React.Component() {
  *   constructor() {
- *     super();
+ *     super(...arguments);
  *     this.state = {
  *        batch: new Batch()
  *     }
@@ -21,7 +21,7 @@
  *     let {batch} = this.state;
  *     batch.add({
  *           type: 'set',
- *           key: 'foo',
+ *           path: ['foo'],
  *           value: 'bar'
  *        });
  *     this.setState({batch});
@@ -42,10 +42,9 @@
 
 class Batch {
   constructor() {
-    const batch = [{action: 'INIT'}];
+    const batch = [];
 
     this.add = this.add.bind(batch);
-
     this.reduce = this.reduce.bind(batch);
   }
 
@@ -68,6 +67,11 @@ class Batch {
    * @returns {any} - The resulting state of the reduce function
    */
   reduce(callback, data) {
+    // Run at least once even if there are no actions in the batch
+    if (this.length === 0) {
+      return callback(data, {value: 'INIT'}, 0);
+    }
+
     return this.reduce(callback, data);
   };
 
--- a/src/js/structs/Batch.js
+++ b/src/js/structs/Batch.js
@@ -54,6 +54,28 @@ class Batch {
    * @param {Object} item
    */
   add(item) {
+    let batchIndex = this.length;
+
+    // As the while is faster then the Array.prototype.forEach and this
+    // function can potentially be called more often we choose while here.
+    while (--batchIndex >= 0) {
+      let {path, value} = this[batchIndex] || {};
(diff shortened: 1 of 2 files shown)


src/js/structs/Batch.js
```diff
@@ -55,6 +54,28 @@ class Batch {
...
+        if (value === item.value) {
+          return;
+        }
+        // Remove previous if path is the same as current to minimize
+        // number of actions
+        if (batchIndex === this.length - 1) {
+          this.pop();
+        }
+      }
+    }
```
Poltergeist: The while loop and looking back if the same path had the same value previously is a good idea, but it will not work. Here is one examples: 

```JS
batch.add({value: 'a', path: ['id']});
batch.add({value: 1, path: ['cpu']});
batch.add({value: 'b', path: ['id']});
batch.add({value: 1, path: ['mem']});
batch.add({value: 'a', path: ['id']});
```

I would expect to have 5 events in the batch but I only would have 4 because the last event with the `value: a` and the `path: ['id']` would have been omitted, because it is the same as the first one.

Another example for duplicatable rows:
- value 10, path `['foo', '0', 'bar']`
- add, path `['foo', '1', 'bar']`
- remove, path `['foo', '0', 'bar']`
- value 10, path `['foo', '0', 'bar']`

If we add a `value=10` for the path `['foo', '0', 'bar']` then we add `['foo', 1, 'bar']` then we remove `['foo', '0', 'bar']` and then change the the value to `value=10` for `['foo', '0', 'bar']` which used to be `['foo', '1', 'bar']` this event would be omitted. Which will lead to a false state. 

Also we want to only replace events / omit events if they have a certain type.
example:

- add, path `['foo', '1', 'bar']`
- add, path `['foo', '2', 'bar']`
- add, path `['foo', '3', 'bar']`
- remove, path `['foo', '0', 'bar']`
- remove, path `['foo', '0', 'bar']`
- remove, path `['foo', '0', 'bar']`

In this case we want to have only 1 row left. but by omitting the last 2 rows we are left with 3 rows.

I hope this makes sense in this abstract form.

I would suggest to only do this
```JS
if (item.type === TransactionTypes.SET) {
  let {path, value} = this[this.length - 1] || {};
  let hasEqualPaths = item && Array.isArray(path) &&
    Array.isArray(item.path) && path.join() === item.path.join();
  if(hasEqualPaths && value === item.value) {
    this.pop();
  }
}

**Comment by orlandohohmeier** [comment]
Please consider implementing this improvement in a subclass or renaming the batch into something form/transaction specific as this change is no longer generic but very specific to the service form use case.

File: src/js/structs/Batch.js (hunk 1 above)
Comment (mlunoe) [inline src/js/structs/Batch.js]: Good points! I need to add some more tests it seems, however, I think it could still work with some tweaks.
For removing the previous event looking at the path should be enough, no need to check the value too.
If we only look back to the first equal path (which I guess was the intention) and check if the value is equal that should handle all those cases too.
Code would look something like this:
```js
let done = false;
while (!done && --batchIndex >= 0) {
  let {path, value} = this[batchIndex] || {};
  let hasEqualPaths = item && Array.isArray(path) &&
    Array.isArray(item.path) && path.join() === item.path.join();
  if (hasEqualPaths) {
    // Last entry with same path holds same value,
    // do not add redundant action
    if (value === item.value) {
      return;
    }
    // Remove previous if path is the same as current to minimize
    // number of actions
    if (batchIndex === this.length - 1) {
      this.pop();
    }
    // break out of while
    done = true;
  }
}
```

mlunoe: We talked offline and decided to remove the look-back to remove older values as there might be added stuff in between that makes this action invalid, e.g. with duplicable fields, which might change one field path into another.

Poltergeist:
LGTM lets 🚢  this.

PR #1317, opened 2016-11-01
[Service create form 2 of 2] First simple form with JSON editor

mlunoe:
This PR introduces the first working form, with interaction between JSON editor and the form section(s).

Here's a video showing it working:

Here's how to activate the new create service form:

PR #1324, opened 2016-11-04
Environment section [1 of n] (Multiple Items)

Poltergeist:
env_vars mov

This is also *Renaming* Duplicatable Rows to Multiple Items, as this is a more fitting name.

This is adding the support for multiple Items.

As in #1317 You need to add some changes to get this working.

Since most of this has been achieved during the task of the Multiple Items(Duplicatable Rows) integration, however this is not the complete environment section but does contain some parts.

TODO:
- [x] Use Parsers on JSON change events
- [x] Tidy up reducers JSON & Form
- [x] Add Style


orlandohohmeier:
THIS IS AMAZING! ✨

jfurrow:
Wow! This looks awesome so far!

plugins/services/src/js/components/forms/EnvironmentFormSection.js
```diff
@@ -0,0 +1,220 @@
...
+  }
+
+  getEnvironmentLines(data) {
+    const {errors} = this.props;
+
+    return data.map((env, key) => {
+      let keyLabel = null;
+      let valueLabel = null;
+      if (key === 0) {
+        keyLabel = (
```
mlunoe: We can do this inline: `keyLabel = <FieldLabel>Key</FieldLabel>;`

plugins/services/src/js/components/forms/EnvironmentFormSection.js
```diff
@@ -0,0 +1,220 @@
...
+            required={false}
+            showError={Boolean(errors.env[key])}>
+            {keyLabel}
+            <FieldInput
+              name={`env.${key}.key`}
+              type="text"
+              value={env.key}/>
+            <FieldError>{errors.env[key]}</FieldError>
+            <span className="emphasis"
+              style={{
```
mlunoe: Let's not use inline styles like this

plugins/services/src/js/components/forms/EnvironmentFormSection.js
```diff
@@ -0,0 +1,220 @@
...
+    ); });
+  }
+
+  getLabelsLines(data) {
+    const {errors} = this.props;
+    return data.map((label, key) => {
+      let keyLabel = null;
+      let valueLabel = null;
+      if (key === 0) {
+        keyLabel = (
```
mlunoe: Let's do this inline

plugins/services/src/js/components/forms/EnvironmentFormSection.js
```diff
@@ -0,0 +1,220 @@
...
+            required={false}
+            showError={Boolean(errors.labels[key])}>
+            {keyLabel}
+            <FieldInput
+              name={`labels.${key}.key`}
+              type="text"
+              value={label.key}/>
+            <span className="emphasis"
+              style={{
+                position: 'absolute',
```
mlunoe: No inline styles, please

plugins/services/src/js/components/forms/EnvironmentFormSection.js
```diff
@@ -0,0 +1,220 @@
...
+            {valueLabel}
+            <FieldInput
+              name={`labels.${key}.value`}
+              type="text"
+              value={label.value}/>
+            <FieldError>{errors.labels[key]}</FieldError>
+          </FormGroup>
+          <div className="form-group flex flex-item-align-end column-4">
+            <a className="column-3 button button-primary-link button-flush"
+              onClick={(event) => {
```
mlunoe: Let's create a `handleRemoveItem` function to separate view declaration and logic. Also, so we are not passing down a new function on each render

plugins/services/src/js/components/forms/EnvironmentFormSection.js
```diff
@@ -0,0 +1,220 @@
...
+            Environment Variables
+          </h2>
+          <p>
+            Set up environment variables for each task your service launches.
+          </p>
+        </div>
+        {this.getEnvironmentLines(data.env)}
+        <div>
+          <a className="button button-primary-link button-flush"
+            onClick={(event) => {
```
mlunoe: Let's create a handler

plugins/services/src/js/components/forms/EnvironmentFormSection.js
```diff
@@ -0,0 +1,220 @@
...
+          </h2>
+          <p>
+            Attach metadata to expose additional information to other services.
+          </p>
+        </div>
+        {this.getLabelsLines(data.labels)}
+        <div>
+          <a
+            className="button button-primary-link button-flush"
+            onClick={(event) => {
```
mlunoe: Let's create a handler

plugins/services/src/js/reducers/form/EnvironmentVariables.js
```diff
@@ -0,0 +1,106 @@
...
+
+        return this.env.reduce((memo, item) => {
+          if (item.key != null || item.value != null) {
+            memo[item.key] = item.value;
+          }
+          return memo;
+        }, {});
+      }
+
+      const joinedPath = path.join('.');
```
mlunoe: Let' move this up. I see three `path.join('.')`s now.
You could do `const joinedPath = Array.isArray(path) && path.join('.');` (or something similar)

plugins/services/src/js/reducers/form/EnvironmentVariables.js
```diff
@@ -0,0 +1,106 @@
...
+  ADD_ITEM,
+  REMOVE_ITEM,
+  SET
+} from '../../../../../../src/js/constants/TransactionTypes';
+import Transaction from '../../../../../../src/js/structs/Transaction';
+
+module.exports = {
+  JSONReducer(state, {type, path, value}) {
+    if (this.env == null) {
+      this.env = [];
```
mlunoe: Not exactly sure what is going on here, what is this `env` array? Could you maybe add some comments? This is hard to decipher.

plugins/services/src/js/reducers/form/EnvironmentVariables.js
```diff
@@ -0,0 +1,106 @@
...
+      return memo;
+    }, {});
+  },
+  JSONParser(state) {
+    if (state.env == null) {
+      return [];
+    }
+
+    return Object.keys(state.env).reduce(function (memo, key, index) {
+      memo.push(new Transaction(['env'], index, ADD_ITEM));
```
mlunoe: Why is this necessary?

plugins/services/src/js/reducers/form/EnvironmentVariables.js
```diff
@@ -0,0 +1,106 @@
...
+          case REMOVE_ITEM:
+            state = state.filter((item, index) => {
+              return index !== value;
+            });
+            break;
+        }
+        return state;
+      }
+
+      let joinedPath = path.join('.');
```
mlunoe: Let's do the same with this joinedPath

plugins/services/src/js/reducers/form/EnvironmentVariables.js
```diff
@@ -0,0 +1,106 @@
...
+
+      return memo;
+    }, []);
+  },
+  FormReducer(state = [], {type, path, value}) {
+    if (path != null && path.join('.').search('env') !== -1) {
+      if (path.join('.') === 'env') {
+        switch (type) {
+          case ADD_ITEM:
+            state.push({key: null, value: null});
```
mlunoe: Are we ok with modifying `state` here? Maybe we should return a new object?

plugins/services/src/js/reducers/form/Labels.js
```diff
@@ -0,0 +1,105 @@
...
+  ADD_ITEM,
+  REMOVE_ITEM,
+  SET
+} from '../../../../../../src/js/constants/TransactionTypes';
+import Transaction from '../../../../../../src/js/structs/Transaction';
+
+module.exports = {
+  JSONReducer(state, {type, path, value}) {
+    if (this.labels == null) {
+      this.labels = [];
```
mlunoe: Let's get a comment about the purpose of this variable

plugins/services/src/js/reducers/form/Labels.js
```diff
@@ -0,0 +1,105 @@
...
+      memo[item.key] = item.value;
+      return memo;
+    }, {});
+  },
+  JSONParser(state) {
+    if (state.labels == null) {
+      return [];
+    }
+    return Object.keys(state.labels).reduce(function (memo, key, index) {
+      memo.push(new Transaction(['labels'], index, ADD_ITEM));
```
mlunoe: I'm getting confused about this again

src/js/components/form/FormGroup.js
```diff
@@ -8,7 +8,7 @@ const FormGroup = (props) => {
   let {children, className, errorClassName, showError, required} = props;
 
   let clonedChildren = React.Children.map(children, (child) => {
-    if (!showError && child.type === FieldError) {
+    if (child == null || (!showError && child.type === FieldError)) {
```
mlunoe: 👍

plugins/services/src/js/components/forms/EnvironmentFormSection.js
Poltergeist: done

plugins/services/src/js/components/forms/EnvironmentFormSection.js
Poltergeist: done

plugins/services/src/js/components/forms/EnvironmentFormSection.js
Poltergeist: Done.

plugins/services/src/js/components/forms/EnvironmentFormSection.js
Poltergeist: done

plugins/services/src/js/components/forms/EnvironmentFormSection.js
Poltergeist: done

plugins/services/src/js/reducers/form/EnvironmentVariables.js
Poltergeist: added a comment, which explains what this is referring to and why it is used here.

plugins/services/src/js/reducers/form/EnvironmentVariables.js
Poltergeist: This is necessary as the Reducer is adding a object with `{key: null, value: null}` to the array. If this would not happen we could run into problems, like an Item is not an Object. Which we have to check on each render. I would rather be more verbose here but remove therefore complexity from views and reducers.

plugins/services/src/js/reducers/form/EnvironmentVariables.js
Poltergeist: We should be okay. As the input state from the combine reducers ist already a copy https://github.com/dcos/dcos-ui/blob/master/src/js/utils/ReducerUtil.js#L39

plugins/services/src/js/reducers/form/EnvironmentVariables.js
Poltergeist: check

plugins/services/src/js/reducers/form/EnvironmentVariables.js
Poltergeist: Also we need this to add a item / row in the form. Otherwise we would have logic in the form.  which should be in the Batch. But I am open to another solution here.

plugins/services/src/js/components/forms/EnvironmentFormSection.js
```diff
@@ -0,0 +1,213 @@
...
+    const {errors} = this.props;
+    return data.map((label, key) => {
+      let keyLabel = null;
+      let valueLabel = null;
+      if (key === 0) {
+        keyLabel = <FieldLabel>Key</FieldLabel>;
+        valueLabel = <FieldLabel>Value</FieldLabel>;
+      }
+      return (
+        <div key={key} className="flex row">
```
jfurrow: @mlunoe @Poltergeist Looks like we should create a `<FormRow>` component to do this for us; what do you think? (no need to do it in this PR)

plugins/services/src/js/components/forms/EnvironmentFormSection.js
```diff
@@ -0,0 +1,213 @@
...
+            required={false}
+            showError={Boolean(errors.env[key])}>
+            {valueLabel}
+            <FieldInput
+              name={`env.${key}.value`}
+              type="text"
+              value={env.value}/>
+            <FieldError>{errors.env[key]}</FieldError>
+          </FormGroup>
+          <div className="form-group flex flex-item-align-end column-4">
```
jfurrow: Let's use the `FormGroup` component here.

plugins/services/src/js/components/forms/EnvironmentFormSection.js
mlunoe: But how would the columns then work? We had issues with this in the old form. It is a good conversation to have, but let's take somewhere else?

plugins/services/src/js/components/forms/EnvironmentFormSection.js
```diff
@@ -0,0 +1,213 @@
...
+            <a className="column-3 button button-primary-link button-flush"
```
jfurrow: Let's remove the `column-3` class here. `button-flush` isn't removing the padding because we're on an outdated version of CNVS, but once that's up to date, the button should be aligned nicely :)

plugins/services/src/js/components/forms/EnvironmentFormSection.js
```diff
@@ -0,0 +1,213 @@
...
+              data-path="env"
```
jfurrow: I would much prefer us a more React friendly approach rather than the element's `dataset`. I would suggest making a `MultipleItemDelete` component (probably that's not a good name), to which we can pass these `path` and `value` and attach click handlers.

plugins/services/src/js/components/forms/EnvironmentFormSection.js
```diff
@@ -0,0 +1,213 @@
...
+          </p>
+        </div>
+        {this.getEnvironmentLines(data.env)}
+        <div>
+          <a className="button button-primary-link button-flush"
+            data-value={data.env.length}
+            data-path="env"
+            onClick={this.handleOnAddItem}>
+            + Add Environment Variable
+          </a>
```
jfurrow: I would like for us not to use the `dataset` for this add button, similar to my comment about the delete button.

jfurrow:
@Poltergeist This works REALLY nicely! I did request a couple of changes.

Also, what do you think about trying to restrict users from deleting items when it is the last of the multiple items? So, if there are two or more multiple items, they can delete any item they wish. But as soon as there is only one multiple item, they cannot delete it.

Here's a terrible gif!

plugins/services/src/js/components/forms/EnvironmentFormSection.js
mlunoe: done

plugins/services/src/js/components/forms/EnvironmentFormSection.js
mlunoe: done

plugins/services/src/js/components/forms/EnvironmentFormSection.js
mlunoe: done

PR #1326, opened 2016-11-07
DCOS-11291: Fix 'access denied' page when visiting a cluster for the first time

nLight:
This PR fixes a regression that user sees 'access denied' page when visiting a cluster for the first time. The solution is to move the root redirect under the `Authenticated` component so the UI will be able to check on the login status first and then redirect.

🙇

### DIFF
--- a/src/js/routes/index.js
+++ b/src/js/routes/index.js
@@ -21,6 +21,11 @@ let routeFactories = [Organization, Network];
 function getApplicationRoutes() {
   // Statically defined routes
   let routes = [].concat(
+    {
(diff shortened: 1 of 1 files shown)


rcorral:
Thanks @nLight!

nLight:
mesosphere-ci retest this please

nLight:
How do I rerun integration?

jfurrow:
@nLight Integration tests are failing right now, unfortunately. Looks like they have been removed from the required cehcks.

PR #1329, opened 2016-11-07
Improve table styling

jfurrow:
This PR:
* sets `vertical-align: top` for all `table` elements (as far as I can tell, all tables that rely on `vertical-align: middle` have explicitly declared this)
* adds a `table-break-word` class
* adds a `table-simple` style variant which looks like this:


mlunoe:
Tested this super small change out, and it works great. Since it is small, I will go ahead and merge it.

PR #1328, opened 2016-11-07
Improve ConfigurationMap styling

jfurrow:
This PR:
* applies the `short-top` class to all `ConfigurationMapHeading` components where `level === 2`
* applies the `@neutral` color to each `ConfigurationMapLabel` element
* assigns the width for the `ConfigurationMapLabel` elements to a variable for re-use in the neighboring tables

PR #1330, opened 2016-11-07
Introduce ServiceConfigDisplay component

jfurrow:
Sorry for the large PR! Most of it is the configuration for generating the display in `ServiceConfigDisplayUtil`. ~~(This PR will look better, visually in your browser, after these PRs merge: https://github.com/dcos/dcos-ui/pull/1328 and https://github.com/dcos/dcos-ui/pull/1329)~~ Marged 

---

This PR introduces a `ServiceConfigDisplay` component and accompanying util which accepts one prop (`appDefinition`) and returns a nicely rendered view of the service's configuration.

The `ServiceConfigDisplay` component iterates over the section arrays defined in the `${__section__}ServiceConfigSection` components. These arrays describe how the app config values should be displayed. Each item in the array results in a `ConfigurationMapHeading ` or `ConfigurationMapRow` (unless a custom `render` method is supplied).

You'll also notice that some items in the `${__section__}ServiceConfigSection.js` arrays contain a `transformValue` and/or `render` methods.
* `transformValue` accepts the value as plucked from the `appDefinition` and returns a modified value (ex: it uses the `Units` util to render disk sizes nicely).
* `render` will completely replace the default `ConfigurationMap`-related components as defined in the `ServiceConfigDisplay` component (ex: you would use this to render a `Table` component rather than the default `Label:Value` style).

---
### WIP Sections
The logic for `Networking`, `Container Runtime`, and potentially `Volumes` still need to be finalized by somebody else. I left `TODO` comments in these cases. I believe I know what logic needs to be implemented for `Container Runtime` and `Volumes` so let me know if you are the person that implements this and I'll provide what I know!

---

Here's a JSON blob you can use to generate the following screenshots: https://gist.github.com/jfurrow/c2907f56cdf25e1cd8a6f65c8fae1ac3


plugins/services/src/js/components/ServiceConfigDisplay.js
```diff
@@ -0,0 +1,124 @@
...
+import Util from '../../../../../src/js/utils/Util';
+
+const shouldExcludeItem = (row, appDefinition) => {
+  switch (row.key) {
+    case 'gpus':
+      return appDefinition.gpus === 0;
+    case 'container.volumes':
+      return appDefinition.container == null
+        || appDefinition.container.volumes == null
+        || appDefinition.container.volumes.length === 0;
```
mlunoe: You could do:
```js
return Util.findNestedPropertyInObject(
  appDefinition,
  'container.volumes.length'
) === 0;
```

plugins/services/src/js/components/ServiceConfigDisplay.js
```diff
@@ -0,0 +1,124 @@
...
+import ConfigurationMap from '../../../../../src/js/components/ConfigurationMap';
+import ConfigurationMapHeading from '../../../../../src/js/components/ConfigurationMapHeading';
+import ConfigurationMapLabel from '../../../../../src/js/components/ConfigurationMapLabel';
+import ConfigurationMapRow from '../../../../../src/js/components/ConfigurationMapRow';
+import ConfigurationMapSection from '../../../../../src/js/components/ConfigurationMapSection';
+import ConfigurationMapValue from '../../../../../src/js/components/ConfigurationMapValue';
+import ServiceConfigDisplayUtil from '../utils/ServiceConfigDisplayUtil';
+import Util from '../../../../../src/js/utils/Util';
+
+const shouldExcludeItem = (row, appDefinition) => {
```
mlunoe: Could we use `appConfig`? That's what we use elsewhere

plugins/services/src/js/components/ServiceConfigDisplay.js
```diff
@@ -0,0 +1,124 @@
...
+    (section, sectionIndex) => {
+      let configurationMapRows = section.values.reduce(
+        (memo, row, rowIndex) => {
+          // Some rows must be excluded if relevant data is missing.
+          if (shouldExcludeItem(row, appDefinition)) {
+            return memo;
+          }
+
+          let reactKey = `${sectionIndex}.${rowIndex}`;
+          let value = null;
```
mlunoe: `findNestedPropertyInObject` will return `null` if it doesn't find it so I think you can just do:
`let value = Util.findNestedPropertyInObject(appDefinition, row.key);`

plugins/services/src/js/components/ServiceConfigDisplay.js
```diff
@@ -0,0 +1,124 @@
...
+        || Object.keys(appDefinition.labels).length === 0;
+    default:
+      return false;
+  }
+};
+
+const ServiceConfigDisplay = ({appDefinition}) => {
+  const configurationMapSections = ServiceConfigDisplayUtil.displayValues.map(
+    (section, sectionIndex) => {
+      let configurationMapRows = section.values.reduce(
```
mlunoe: Instead of reduce, you could do a `filter` and then `map`, which I think is a little bit more readable, and makes the function easier to parse. Refactored it a bit here (didn't test it):
```js
const getDisplayValue = (type, value) => {
  // Otherwise we treat the row as "label:value" type display.
  let displayValue = null;

  // If the row's type is pre, we wrap it in a pre tag.
  if (type === 'pre') {
    return (
      <pre className="flush transparent wrap">{value}</pre>
    );
  }

  return ServiceConfigDisplayUtil.getDisplayValue(value);
}

const ServiceConfigDisplay = ({appDefinition}) => {
  const configurationMapSections = ServiceConfigDisplayUtil.displayValues.map(
    (section, sectionIndex) => {
      let configurationMapRows = section.values.filter((row) => {
          // Some rows must be excluded if relevant data is missing.
          return !shouldExcludeItem(row, appDefinition);
        })
        .map((row, rowIndex) => {
          let reactKey = `${sectionIndex}.${rowIndex}`;
          let value = Util.findNestedPropertyInObject(appDefinition, row.key);

          // If a transformValue was specified on the row, we use it.
          if (row.transformValue != null) {
            value = row.transformValue(value, appDefinition);
          }

          if (row.render != null) {
            // If a custom render method was specified on the row, we use that.
            return row.render(value, appDefinition);
          }

          if (row.heading != null) {
            // If the row is a heading, we render the heading.
            return (
              <ConfigurationMapHeading key={reactKey} level={row.headingLevel}>
                {row.heading}
              </ConfigurationMapHeading>
            );
          }

          return (
            <ConfigurationMapRow key={reactKey}>
              <ConfigurationMapLabel>
                {row.label}
              </ConfigurationMapLabel>
              <ConfigurationMapValue>
                {getDisplayValue(row.type, value)}
              </ConfigurationMapValue>
            </ConfigurationMapRow>
          );
        }
      );

      if (configurationMapRows.length === 0) {
        return null;
      }

      return (
        <ConfigurationMapSection key={sectionIndex}>
          {configurationMapRows}
        </ConfigurationMapSection>
      );
    }
  );
  ...
}
```

plugins/services/src/js/utils/ServiceConfigDisplayUtil.js
```diff
@@ -0,0 +1,691 @@
...
+import HostUtil from './HostUtil';
+import Networking from '../../../../../src/js/constants/Networking';
+import Units from '../../../../../src/js/utils/Units';
+import Util from '../../../../../src/js/utils/Util';
+
+const renderMillisecondsFromSeconds = (prop, row) => {
+  let value = row[prop];
+
+  if (value != null) {
+    value = `${value * 1000} ms`;
```
mlunoe: Shouldn't we use `DateUtil` for displaying time values?

plugins/services/src/js/utils/ServiceConfigDisplayUtil.js
```diff
@@ -0,0 +1,691 @@
...
+            if (value == null) {
+              return ServiceConfigDisplayUtil.getDisplayValue(value);
+            }
+
+            return Units.formatResource('disk', value);
+          }
+        },
+        {
+          key: 'gpus',
+          label: 'GPU'
```
mlunoe: I think we should do something similar to CPU here

plugins/services/src/js/utils/ServiceConfigDisplayUtil.js
```diff
@@ -0,0 +1,691 @@
...
+    );
+  };
+};
+
+const ServiceConfigDisplayUtil = {
+  displayValues: [
+    {
+      values: [
+        {
+          heading: 'General',
```
mlunoe: What do you say we divide these configurations into separate files by category like it is done here: https://github.com/dcos/dcos-ui/pull/1324/files#diff-db6f93bce675a7029b88eb07056b4e52

`/DisplayConfigs/General.js`
`/DisplayConfigs/Network.js`
...
etc.

Not exactly sure if `/DisplayConfigs` is best place for those though, but something along those lines

plugins/services/src/js/components/ServiceConfigDisplay.js
jfurrow: We also need to check for a `null` value, which looks like this:

```javascript
return Util.findNestedPropertyInObject(
    appConfig,
    'container.volumes'
  ) == null
  || Util.findNestedPropertyInObject(
    appConfig, 'container.volumes.length'
  ) === 0;
```

So I actually think it looks a little cleaner to do the triple checks:
```javascript
return appConfig.container == null
  || appConfig.container.volumes == null
  || appConfig.container.volumes.length === 0;
```

What do you think?

plugins/services/src/js/utils/ServiceConfigDisplayUtil.js
jfurrow: Yeah, good point. The designs showed the number in milliseconds, but I remember having a discussion about being consistent with times in the app. So I've implemented `DateUtil.getDuration` here:

plugins/services/src/js/utils/ServiceConfigDisplayUtil.js
jfurrow: Sorry, I'm not sure what you mean by this comment; `gpus` and `cpus` are treated similarly.

plugins/services/src/js/components/ServiceConfigDisplay.js
jfurrow: Done!

plugins/services/src/js/components/ServiceConfigDisplay.js
jfurrow: I agree, this is much more readable. I chose the `map` to reduce the number of times we iterate over items in the array, but I changed it as you suggested. So, done!

plugins/services/src/js/components/ServiceConfigDisplay.js
jfurrow: Done!

plugins/services/src/js/utils/ServiceConfigDisplayUtil.js
mlunoe: Oh dang, I was looking at `disk`. Sorry! _hides away in the corner_

plugins/services/src/js/utils/ServiceConfigDisplayUtil.js
jfurrow: This does make sense, and would make this file a lot more readable! Working on this now.

plugins/services/src/js/components/ServiceConfigDisplay.js
mlunoe: Anyways, up to you. You can keep it the way it is too :)

plugins/services/src/js/utils/ServiceConfigDisplayUtil.js
mlunoe: Awesome!

plugins/services/src/js/components/ServiceConfigDisplay.js
jfurrow: I like your solution. Done!

jfurrow:
@mlunoe I believe I have addressed all of your comments, let me know if there's anything else you'd like to see.

plugins/services/src/js/service-configuration/ServiceConfigDisplay.js
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  return ServiceConfigDisplayUtil.getDisplayValue(value);
+};
+
+const ServiceConfigDisplay = ({appConfig}) => {
+  const configurationMapSections = serviceConfigDisplayList.map(
+    (section, sectionIndex) => {
+      let configurationMapRows = section.values.filter((row) => {
+        // Some rows must be excluded if relevant data is missing.
```
MatApple: I wonder if we should have a warning and link to go edit this section? @leemunroe @ashenden

MatApple:
This looks great @jfurrow - really nice work!

plugins/services/src/js/service-configuration/ServiceConfigDisplay.js
leemunroe: I think it's fine to not show a row or section if there is nothing configured, or no default e.g. if no volume configured then don't show volumes section. You're right that there could also be an advantage showing it's not configured, but I would start with hiding it then based on feedback decide if we actually want to show everything.

PR #1331, opened 2016-11-08
Add SSE handling for logs

mlunoe:
This PR introduces Action and Store file for handling Server Sent Events for logs.

You can test the action and store out here (Not complete implementation, but gives you the ability to play around with it): https://github.com/dcos/dcos-ui/tree/mlunoe/feature/add-new-stream-logs-view.

You need to start a cluster with `testing/pull/923` to test it out.


src/js/stores/SystemLogStore.js
```diff
@@ -0,0 +1,194 @@
...
+        case REQUEST_PREVIOUS_SYSTEM_LOG_ERROR:
+          this.processLogPrependError(subscriptionID, data);
+          break;
+      }
+
+      return true;
+    });
+  }
+
+  addEntries(logData, subscriptionID, entries, eventType) {
```
mlunoe: I have not written tests for this guy yet

mlunoe:
@orlandohohmeier @jfurrow Could I get some 👀 on this?

src/js/events/SystemLogActions.js
```diff
@@ -0,0 +1,118 @@
...
+const SystemLogActions = {
+  subscribe(nodeID, options = {}) {
+    let {
+      subscriptionID
+    } = options;
+
+    let url = SystemLogUtil.getUrl(nodeID, options);
+    subscriptionID = subscriptionID || global.btoa(url);
+    let source = new EventSource(url, {
+      withCredentials: Boolean(CookieUtils.getUserMetadata())
```
orlandohohmeier: Let's not use the Boolean object to cast non-falsy values to booleans and rather add a an additional condition as it's easier to reason about.

src/js/events/SystemLogActions.js
```diff
@@ -0,0 +1,118 @@
...
+    });
+
+    source.addEventListener('message', function ({data, origin, eventPhase} = {}) {
+      if (origin !== global.location.origin && eventPhase !== EventSource.OPEN) {
+        // Event is not from same origin, or is not open anymore
+        return false;
```
orlandohohmeier: Shouldn't we also unsubscribe (remove all listener) in  this case?

src/js/events/SystemLogActions.js
```diff
@@ -0,0 +1,118 @@
...
+      }
+
+      let parsedData = JSON.parse(data);
```
orlandohohmeier: Let's wrap this in try/cate block if there is a chance that the received data is malformed?

src/js/events/SystemLogActions.js
```diff
@@ -0,0 +1,118 @@
...
+// Store of current open connections
+let sources = {};
+const SystemLogActions = {
+  subscribe(nodeID, options = {}) {
+    let {
+      subscriptionID
+    } = options;
+
+    let url = SystemLogUtil.getUrl(nodeID, options);
+    subscriptionID = subscriptionID || global.btoa(url);
```
orlandohohmeier: Why do we need to encode the url? Can't we use a `Symbol` instead to create a new ID?

src/js/events/SystemLogActions.js
```diff
@@ -0,0 +1,118 @@
...
+    source.addEventListener('error', (event = {}) => {
+      AppDispatcher.handleServerAction({
+        type: REQUEST_SYSTEM_LOG_ERROR,
+        data: event,
+        subscriptionID
+      });
+    }, false);
+
+    // Unsubscribe if any open connection exists with the same ID
+    this.unsubscribe(subscriptionID);
```
orlandohohmeier: Shouldn't we do this as a first step, before create a new event source?

src/js/events/SystemLogActions.js
```diff
@@ -0,0 +1,118 @@
...
+    sources[subscriptionID] = {source};
```
orlandohohmeier: Please remove the extra object if we don't need to store more than the source itself.

src/js/events/SystemLogActions.js
```diff
@@ -0,0 +1,118 @@
...
+
+    return subscriptionID;
+  },
+
+  unsubscribe(subscriptionID) {
+    if (sources[subscriptionID]) {
+      sources[subscriptionID].source.close();
```
orlandohohmeier: Is `close` also removing any registered event listener so that stuff gets garbage collected?

src/js/events/SystemLogActions.js
```diff
@@ -0,0 +1,118 @@
...
+      delete sources[subscriptionID];
+    }
+  },
+
+  fetchLogRange(nodeID, options = {}) {
```
orlandohohmeier: This method is in large parts very similar to the `subscribe` one. any chance we can separate the respective event source handling so that the two methods only handle the data?

src/js/events/SystemLogActions.js
```diff
@@ -0,0 +1,118 @@
...
+    } = options;
+
+    let url = SystemLogUtil.getUrl(nodeID, options, false);
+    subscriptionID = subscriptionID || global.btoa(url);
+    let source = new EventSource(url, {
+      withCredentials: Boolean(CookieUtils.getUserMetadata())
+    });
+
+    let items = [];
+    let hasReachedTop = false;
```
orlandohohmeier: Let's not use names that reflect our way of representing the data in a store  #nit

src/js/events/SystemLogActions.js
```diff
@@ -0,0 +1,118 @@
...
+      delete sources[subscriptionID];
+    }
+  },
+
+  fetchLogRange(nodeID, options = {}) {
+    let {
+      cursor,
+      subscriptionID,
+      skip_prev
+    } = options;
```
orlandohohmeier: Please document these options

src/js/events/SystemLogActions.js
```diff
@@ -0,0 +1,118 @@
...
+      sources[subscriptionID].source.close();
+      delete sources[subscriptionID];
+    }
+  },
+
+  fetchLogRange(nodeID, options = {}) {
+    let {
+      cursor,
+      subscriptionID,
+      skip_prev
```
orlandohohmeier: As there a reason for using snake_case instead of camelCase?

orlandohohmeier:
I haven't found the time to review the `SystemLogUtil`, will continue tomorrow.

src/js/utils/SystemLogUtil.js
```diff
@@ -0,0 +1,44 @@
...
+
+      return memo + paramItem;
+    }, '');
+
+    let endpoint = 'logs';
+    if (isStream) {
+      endpoint = 'stream';
+    }
+
+    let timestamp = `timestamp=${Date.now()}`;
```
orlandohohmeier: Does the API require a timestamp or is it just the usual cache busting?

MatApple:
@mlunoe I think this code should live in `services` alongside all the task specific code. I can move it for you after we merge this.

src/js/events/SystemLogActions.js
```diff
@@ -0,0 +1,118 @@
...
+    let hasReachedTop = false;
+
+    source.addEventListener('message', function ({data, origin, eventPhase} = {}) {
+      if (origin !== global.location.origin && eventPhase !== EventSource.OPEN) {
+        // Event is not from same origin, or is not open anymore
+        return false;
+      }
+
+      let parsedData = JSON.parse(data);
+      if (!hasReachedTop && skip_prev > 0 && parsedData.cursor === cursor) {
```
MatApple: It takes a while to digest what is happening here. Could you please add a comment about how we're deciding we're at the top.

src/js/utils/SystemLogUtil.js
```diff
@@ -0,0 +1,44 @@
+const SystemLogUtil = {
+  getUrl(nodeID, options, isStream = true) {
+    let {params = {}} = options;
+
+    let range = ['cursor', 'limit', 'skip_prev', 'skip_next'].reduce(function (memo, key) {
+      if (options[key] === '' || options[key] == null) {
+        return memo;
+      }
+      let rangeItem = `${key}=${encodeURIComponent(options[key])}`;
+      if (memo !== '') {
```
MatApple: Instead of reducing on a string, we could reduce onto an Array - `memo.push(rangeItem)` and then after the reduce function has completed we can do `range.join('&')`. This circumvent this `if (memo !== '') {}` check. It's also much more efficient (although for a small number of iterations it doesn't really matter).

src/js/utils/SystemLogUtil.js
```diff
@@ -0,0 +1,44 @@
...
+      return memo + rangeItem;
+    }, '');
+
+    let paramString = Object.keys(params).reduce(function (memo, key) {
+      if (params[key] === '' || params[key] == null) {
+        return memo;
+      }
+
+      let paramItem = `filter=${encodeURIComponent(key)}:${encodeURIComponent(params[key])}`;
+      if (memo !== '' || range !== '') {
```
MatApple: Same here - reduce on an Array.

src/js/utils/SystemLogUtil.js
```diff
@@ -0,0 +1,44 @@
...
+    if (isStream) {
+      endpoint = 'stream';
+    }
+
+    let timestamp = `timestamp=${Date.now()}`;
+    if (range !== '' || paramString !== '') {
+      timestamp = `&${timestamp}`;
+    }
+
+    return `/system/logs/v1/agent/${nodeID}/${endpoint}?${range}${paramString}${timestamp}`;
```
MatApple: This API prefix should be in the config file.

src/js/events/SystemLogActions.js
mlunoe: Why not? I think this is very easy to parse

src/js/events/SystemLogActions.js
```diff
@@ -0,0 +1,118 @@
...
+/**
+ * Implementation of server sent event handling for
+ * https://github.com/dcos/dcos-log
+ */
+
+// Store of current open connections
+let sources = {};
+const SystemLogActions = {
+  subscribe(nodeID, options = {}) {
+    let {
```
mlunoe: done

src/js/events/SystemLogActions.js
mlunoe: Yup! Using a Symbol now

src/js/events/SystemLogActions.js
mlunoe: Nope. The comment is misleading and was based on a previous perception I had. We should just ignore these events. It does not indicate that the event stream is not open anymore

src/js/events/SystemLogActions.js
mlunoe: Yes, done!

src/js/events/SystemLogActions.js
mlunoe: Yes, this was leftover from when I needed more options, I actually still needed it to store the listeners, so we could make sure to remove them, when unsubscribing.

src/js/events/SystemLogActions.js
mlunoe: added functionality to remove listeners

src/js/events/SystemLogActions.js
mlunoe: Decided to postpone this as it is not crucial for this task

src/js/events/SystemLogActions.js
mlunoe: done

src/js/events/SystemLogActions.js
mlunoe: Not quite sure what you mean, but changed the name to `firstEntry`, hope this makes it clearer.

src/js/utils/SystemLogUtil.js
```diff
@@ -0,0 +1,44 @@
+const SystemLogUtil = {
+  getUrl(nodeID, options, isStream = true) {
+    let {params = {}} = options;
+
+    let range = ['cursor', 'limit', 'skip_prev', 'skip_next'].reduce(function (memo, key) {
```
mlunoe: Changed them into more simple reducers, don't see a need for them to be a general url util, other than the reducer part

src/js/utils/SystemLogUtil.js
```diff
@@ -0,0 +1,44 @@
...
+      }
+      let rangeItem = `${key}=${encodeURIComponent(options[key])}`;
+      if (memo !== '') {
+        rangeItem = `&${rangeItem}`;
+      }
+
+      return memo + rangeItem;
+    }, '');
+
+    let paramString = Object.keys(params).reduce(function (memo, key) {
```
mlunoe: done

src/js/events/SystemLogActions.js
mlunoe: done

src/js/utils/SystemLogUtil.js
mlunoe: done

src/js/utils/SystemLogUtil.js
mlunoe: done

src/js/utils/SystemLogUtil.js
mlunoe: done

jest/setupEnv.js
```diff
@@ -41,3 +42,22 @@ RequestUtil.json = function () {};
     };
   }
 }());
+
```
philipnrmn: I like the separate IIFE for clarity, but can we add a short explanation of what this block is doing?

philipnrmn:
💯  awesome work!

philipnrmn:
Three addressed reviews so I'm going to go ahead and merge :)

PR #1334, opened 2016-11-08
DCOS-11326: Implement Navigation Service

nLight:
ℹ️  ~~This PR is not feature complete yet, but you can get a feeling of how the API will look like.~~

----

This PR introduces `NavigationService` module that lives in `foundation-ui` and will be used by packages to register themselves in the Navigation/Sidebar.

It's a simple EventEmitter that will emit change events so any other package could subscribe to it. ATM there will be only one _external_ listener - Sidebar.js

There's also a utility module `NavigationServiceUtil` that lives in DCOS-UI and maps application routes into NavigationService's API calls to support legacy until we migrate all the packages.

I decided to go with a simple declarative API as I did in the `RoutingService`.

🙇


foundation-ui/navigation/EventTypes.js
```diff
@@ -0,0 +1,3 @@
+export default {
```
rcorral: I'm surprised `export default` isn't causing any issues here. It was in the past, where we would have to do:
`import Default as EventTypes from '...'`

foundation-ui/navigation/NavigationService.js
```diff
@@ -0,0 +1,111 @@
...
+
+  registerCategory(category) {
+    if (!this.definition.find((element) => element.category === category)) {
+      this.definition.push({
+        category,
+        children: []
+      });
+    }
+
+    this.emit(EventTypes.NAVIGATION_CHANGE);
```
rcorral: Should we only emit when we've actually made a change, rather than on every call to `#registerCategory`?

foundation-ui/navigation/NavigationService.js
```diff
@@ -0,0 +1,111 @@
+import {EventEmitter} from 'events';
+import EventTypes from './EventTypes';
```
rcorral: Let's do `import {NAVIGATION_CHANGE} from './EventTypes';`
This pattern is better as an error will occur if we mistype a constant, rather than accidentally fat-fingering it when we access the object eg. `EventTypes.FOo_BAR`

src/js/components/Sidebar.js
```diff
@@ -228,6 +163,42 @@ var Sidebar = React.createClass({
     return <ul className="sidebar-menu">{groupMenuItems}</ul>;
   },
 
+  getGroupSubmenu({path, children = []}, {currentPath, isParentActive}) {
```
nLight: I will, though this is an old piece of code I mostly copy/pasted from the Sidebar.js

foundation-ui/navigation/EventTypes.js
nLight: I'll change it, but I'd even prefer having multiple named exports like 
```
export const NAVIGATION_CHANGE = 'NAVIGATION_CHANGE';
```
but then we have eslint rule prefer-default-export :trollface:

foundation-ui/navigation/NavigationService.js
```diff
@@ -0,0 +1,111 @@
...
+      this.definition.push({
+        category,
+        children: []
+      });
+    }
+
+    this.emit(NAVIGATION_CHANGE);
+  }
+
+  registerPrimary(path, link, options = {}) {
```
orlandohohmeier: Let's combine the primary and secondary link registration so that we can compose every level of navigation.

foundation-ui/navigation/NavigationService.js
```diff
@@ -0,0 +1,111 @@
...
+    const {category = 'root'} = options;
+    const categoryElement = this.definition
+      .find((element) => element.category === category);
+
+    if (!categoryElement) {
+      return this.defer('registerPrimary', arguments);
```
orlandohohmeier: Let's pass the method instead of a string if we really need dynamic pointers.

foundation-ui/navigation/NavigationService.js
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    const existingElement = categoryElement.children.find((element) => {
+      return element.path === path;
+    });
+
+    if (existingElement) {
```
orlandohohmeier: Shouldn't we throw an error in this case (at least in development)?

src/js/components/PrimarySidebarLink.js
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import {Link} from 'react-router';
+import NotificationStore from '../stores/NotificationStore';
+
+export default ({to, children, icon}) => {
```
orlandohohmeier: Let's please always name components

orlandohohmeier: Please you the CommonJS syntax for exporting.

foundation-ui/navigation/index.js
```diff
@@ -0,0 +1,5 @@
+import NavigationService from './NavigationService';
+import EventTypes from './EventTypes';
+
+export default NavigationService;
+export {EventTypes};
```
orlandohohmeier: Please you the CommonJS syntax for exporting, as discussed

foundation-ui/navigation/NavigationService.js
```diff
@@ -0,0 +1,111 @@
...
+      path: `${parentPath}/${path}`,
+      children: []
+    });
+
+    this.emit(NAVIGATION_CHANGE);
+  }
+};
+
+const service = new NavigationService();
+export default service;
```
orlandohohmeier: Please use the CommonJS syntax for exporting

src/js/utils/NavigationServiceUtil.js
```diff
@@ -0,0 +1,51 @@
...
+
+const NavigationServiceUtil = {
+  /**
+   * routesToDefinition - declares Navigation Elements based on routes using
+   * NavigationService API
+   *
+   * @param  {Array} routes Application routes
+   * @return {undefined}
+   */
+  routesToDefinition(routes) {
```
orlandohohmeier: Please rename this util so that it's clear that it actually registers categories and links on the navigation service

src/js/utils/NavigationServiceUtil.js
```diff
@@ -0,0 +1,51 @@
+import React from 'react';
+import NavigationService from '../../../foundation-ui/navigation';
+
+const NavigationServiceUtil = {
```
orlandohohmeier: Please add proper tests for this util

src/js/utils/NavigationServiceUtil.js
nLight: `registerRoutesInNavigation` ?

foundation-ui/navigation/NavigationService.js
nLight: maybe warn?

foundation-ui/navigation/NavigationService.js
nLight: 😩  I wanted to avoid any guessing/splitting/regexing with the path

orlandohohmeier: Shouldn't something like the following be rather straight forward?
 
```
 registerLink(path, component, parent = null, category = 'root') {
   // find category
      // defer if not found
      // find parent if defined
           // defer if not found
  // register link
 }
```

nLight: I thought you mean
```js
registerLink(path, component, options)
```
and then split the path and try to guess which level it is.

nLight: In my taste it would be again a function with 4 arguments and declaration would be rather hard to read. Right now it's straight forward what you want to register, category, primary, secondary

foundation-ui/navigation/NavigationService.js
Poltergeist: Warn is probably better.

src/js/utils/NavigationServiceUtil.js
Poltergeist: sounds good to me.

Poltergeist:
image
I am getting a warning I can't figure out what exactly the problem is because in all places I was expecting a key there also was a key. 

Otherwise this PR is in a good state and only the two outstanding changes from @orlandohohmeier are missing.

src/js/components/Sidebar.js
```diff
@@ -228,6 +168,39 @@ var Sidebar = React.createClass({
     return <ul className="sidebar-menu">{groupMenuItems}</ul>;
   },
 
+  getGroupSubmenu(children) {
+    let {pathname} = this.props.location;
+    let isChildActive = false;
+
+    let menuItems = children.reduce(function (children, currentChild) {
+      let isActive = pathname.startsWith(currentChild.path);
```
MatApple: Instead of `startsWith` should we have more robust check like splitting on `/` and matching the array indices from 0 -> number of url "parts"? That would prevent false positives.

foundation-ui/navigation/NavigationService.js
```diff
@@ -0,0 +1,116 @@
...
+
+  registerCategory(category) {
+    if (!this.definition.find((element) => element.category === category)) {
+      this.definition.push({
+        category,
+        children: []
+      });
+    }
+
+    this.instance.emit(NAVIGATION_CHANGE);
```
nLight: 👍

nLight:
@Poltergeist @MatApple I addressed all the comments! Thank you 🙇

nLight:
@orlandohohmeier 🙏

src/js/components/Sidebar.js
MatApple: @nLight - just this comment outstanding but I'm happy to defer this discussion.

MatApple:
Great work @nLight!

nLight:
I'll extract this one from the PluginSDK as discussed with Mat.

nLight:
rebased!

foundation-ui/navigation/NavigationService.js
nLight: Done!

nLight:
retest this please

nLight:
retest this please

orlandohohmeier:

nLight:
It's green \o/ 💚

src/js/pages/Index.js
```diff
@@ -41,6 +42,8 @@ var Index = React.createClass({
     MetadataStore.init();
     SidebarStore.init();
 
+    NavigationServiceUtil.registerRoutesInNavigation(this.props.routes);
```
wavesoft: Not a blocker issue, but I have a feeling there is something conceptually wrong here.

I would expect a utility function to operate only on the arguments it receives and not to perform any behind-the-scenes access to instances or singletons.

For me, I would suggest either:
1. `NavigationService.registerRoutesInNavigation` or
2. `NavigationServiceUtil.registerRoutesInNavigation( NavigationService, this.props.routes );`

But again, that's not a blocker. If you agree, let's make a JIRA issue.

wavesoft: After discussion, with @nLight we agreed that this utility is more like a "Legacy Glue" and therefore kind of makes sense to be separated. 👍

PR #1335, opened 2016-11-08
Introduce Alert component

jfurrow:
This PR introduces the `Alert` component. It will be implemented in the upcoming rejected offers PR.

### DIFF
--- a/src/js/components/Alert.js
+++ b/src/js/components/Alert.js
@@ -10,28 +10,14 @@ const Alert = ({children, showIcon, type}) => {
   let icon = null;
 
   if (showIcon) {
-    const className = 'alert-icon';
-    const family = 'mini';
-    const size = 'mini';
+    const ids = {
+      danger: 'ring-exclamation',
+      success: 'checkmark'
+    };
 
-    switch (type) {
-      case 'danger':
-        icon = (
-          <Icon className={className}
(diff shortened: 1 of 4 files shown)


src/js/components/Alert.js
```diff
@@ -0,0 +1,58 @@
...
+    [`alert-${type}`]: type != null
+  });
+  let icon = null;
+
+  if (showIcon) {
+    const className = 'alert-icon';
+    const family = 'mini';
+    const size = 'mini';
+
+    switch (type) {
```
nLight: What do we think about this pattern?
```js
const id = {
  danger: 'ring-exclamation', 
  success: 'checkmark'
}[type];
```

We could also go with a private component:
```js
const AlertIcon = ({type}) => {
  const id = {
    danger: 'ring-exclamation', 
    success: 'checkmark'
  }[type];

  return <Icon className="alert-icon" family="mini" id={id} size="mini" />;
};

Comment (nLight) [inline src/js/components/Alert.js]: Or maybe even extract `danger` and `success` as constants and operate this constants and not bare string. In this case we don't need a switch at all.
```
const DANGER = 'ring-exclamation';
const SUCCESS = 'checkmark';

....

<Alert type={Alert.SUCCESS}>
```

PR #1336, opened 2016-11-09
DCOS-11391: Fix Links to Volume Detail

nLight:
This PR fixes links to the Volume Detail view from all locations:
- Services
- Service/Volumes
- Nodes/Task/Volumes

PS. I do agree that we should refactor the way we detect current location, I'd rather see HOC pushing this piece of config down to `VolumeTable`

🙇

### DIFF
--- a/plugins/nodes/src/js/routes/nodes.js
+++ b/plugins/nodes/src/js/routes/nodes.js
@@ -156,18 +156,18 @@ let nodesRoutes = {
               }
             },
             {
-              type: Route,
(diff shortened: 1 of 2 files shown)


wavesoft:
Besides the funky way of checking the current path, looks good to me! Great! ✨  💯 
I agree on the idea of specialised high-order components, as discussed

jfurrow:
Nice work @nLight!

nLight:
I think 2👍  + 🎉  === `reviewer` ;D

PR #1337, opened 2016-11-09
ObjectUtil: Adding walkObjectsByPath function

wavesoft:
This PR introduces the `walkObjectsByPath` function that can now be used for performing arbitrary operations on objects, based on array paths. I found myself repeating a couple of times, working on the new data validator, so I thought of separating the core functionality as a utility function.

The idea is quite simple:
* Use a `resolution` function to walk every path component and resolve the appropriate nested child
* Use an `operation` function that performs the decided operation on the children of the final leaf

This separation simplifies the way we handle missing object keys, since the `resolution` function is responsible for this (check [walkCreateMissing](https://github.com/dcos/dcos-ui/pull/1337/files#diff-a5ea432b2de66adf76cd2f82b8072e65R54) or  [walkSkipMissing](https://github.com/dcos/dcos-ui/pull/1337/files#diff-a5ea432b2de66adf76cd2f82b8072e65R71)).

This makes the `operation` to always operate on a flat object, allowing any kind of operations to be cleanly applied.

## Example

For example, you can create a utility that assigns values to an object by path like so:
Ps. [`walkObjectByPath`](https://github.com/dcos/dcos-ui/pull/1337/files#diff-a5ea432b2de66adf76cd2f82b8072e65R134) is a shorthand for `walkObjectsByPath` for only one object.

```js
function setByPath(object, path, value) {
  ObjectUtil.walkObjectByPath(path, object, function(leafObject, leafKey) {
    leafObject[leafKey] = value;
  });
}

var object = {};
setByPath(object, ['a', 'b'], 1);

// object = {'a': {'b': 1}};
```

### DIFF
--- a/src/js/utils/ObjectUtil.js
+++ b/src/js/utils/ObjectUtil.js
@@ -1,4 +1,18 @@
-module.exports = {
+/**
+ * Return a new array or a new object, if the key given is a number or string.
+ *
+ * @param {String} forKey - The key to create a new object/array for
+ * @returns {Object|Array} Returns an object or array, according to type
+ */
+function newTypeFor(forKey) {
+  if (isNaN(forKey)) {
+    return {};
+  } else {
+    return [];
+  }
+}
+
+var ObjectUtil = {
 
   /**
    * This function adds the given mark to the provided object.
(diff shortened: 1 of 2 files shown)


wavesoft:
As @nLight pointed out, this might be a bit specific to a particular case, so it's not a good idea to separate it as a utility function.
As a matter of fact we are discussion possible introduction of https://github.com/nLight/objektiv instead of paths.

wavesoft:
We are now using @nLight 's `objektiv` + functional lenses instead of walking objects by path.

PR #1338, opened 2016-11-09
[Rejected Offers: 2/2] DCOS-10994: Implement rejected offers view components

jfurrow:
~~Depends on https://github.com/dcos/dcos-ui/pull/1373 (this PR will be much smaller when that merges)~~

This PR implements the rejected offers view components.


mlunoe:
This needs a rebase!

jfurrow:
@mlunoe Rebased!

plugins/services/src/js/components/DeclinedOffersTable.js
```diff
@@ -0,0 +1,266 @@
...
+      let icon = null;
+      if (isOfferDeclined) {
+        icon = <Icon color="red" family="mini" id="close" size="mini" />;
+      } else {
+        icon = <Icon color="neutral" family="mini" id="check" size="mini" />;
+      }
+
+      const tooltipContent = (
+        <div>
+          <div className="">
```
philipnrmn: Empty className could be removed, I guess?

plugins/services/src/js/components/DeclinedOffersTable.js
```diff
@@ -0,0 +1,266 @@
...
+            <strong>Requested</strong>{': '}
+            {summary[prop].requested}
+          </div>
+          <div className="">
```
philipnrmn: And again

plugins/services/src/js/components/RecentOffersSummary.js
```diff
@@ -0,0 +1,161 @@
...
+
+const columns = [
+  {
+    heading: '',
+    prop: 'resource',
+    render: (prop, row) => {
+      const resource = row[prop];
+      let tooltipContent = null;
+
+      // TODO: Implement actual tooltip content.
```
philipnrmn: Are we tracking this TODO?

plugins/services/src/js/containers/pod-debug/PodDebugContainer.js
```diff
@@ -81,14 +123,88 @@ class PodDebugTabView extends React.Component {
...
+    } else {
+      const {declinedOffers: {summary}} = queue;
+      const {roles: {offers = 0}} = summary;
+      const docsURL = MetadataStore.buildDocsURI(
+        '/overview/concepts/#mesos-resource-offer'
+      );
+
+      introText = (
+        <span>
+          When you attempt to deploy a service, DC/OS waits for offers to match the resources your service requires. If the offer does not satisfy the requirement, it is declined and DC/OS retries. <a href={docsURL} target="_blank">Learn more</a>.
```
philipnrmn: I guess we're deliberately abstracting Marathon and Mesos away here? Feels kinda weird to be talking about DC/OS itself doing either of these things.

plugins/services/src/js/containers/service-debug/ServiceDebugContainer.js
```diff
@@ -83,6 +98,81 @@ class ServiceDebugContainer extends React.Component {
...
+    } else {
+      const {declinedOffers: {summary}} = queue;
+      const {roles: {offers = 0}} = summary;
+      const docsURL = MetadataStore.buildDocsURI(
+        '/overview/concepts/#mesos-resource-offer'
+      );
+
+      introText = (
+        <span>
+          When you attempt to deploy a service, DC/OS waits for offers to match the resources your service requires. If the offer does not satisfy the requirement, it is declined and DC/OS retries. <a href={docsURL} target="_blank">Learn more</a>.
```
philipnrmn: This block of text is repeated exactly in both Pods and Services debug container, reckon it's worth moving it to a constant somewhere?

plugins/services/src/js/components/DeclinedOffersTable.js
jfurrow: Yikes... thanks!

plugins/services/src/js/components/DeclinedOffersTable.js
jfurrow: Thanks! Removed!

plugins/services/src/js/components/RecentOffersSummary.js
jfurrow: Good call, I added a JIRA link here :)

plugins/services/src/js/containers/service-debug/ServiceDebugContainer.js
jfurrow: Good call. I ended up creating a constants file which actually contains JSX... not sure how to feel about it. Do you think it should be an actual component? https://github.com/dcos/dcos-ui/pull/1338/files#diff-0eac48c221e8d29951e9ffac25121e75

orlandohohmeier:
There were some issues with the `DeclinedOffersUtil` which resulted in wrong outputs that #1425 addresses. Please rebase this one once that PR is merged. 

Unfortunately there is also another case that needs to be handled. Please "x" all the other rejection reasons, if an offer doesn't match the required role as resources are always matched against the role first and thus never match in those cases. Otherwise the result is somewhat confusing as in the following screenshot:



Please also add the requested and offered role to the tooltip in that case. 

/cc @leemunroe @pyronicide @unterstein

jfurrow:
@orlandohohmeier So are you saying I should remove the tooltip and force the red "x" for any rejection reason that follows the role column, if the role was rejected?

orlandohohmeier:
@jfurrow I wouldn't remove the tooltip but extend it to include the role information.

jfurrow:
@orlandohohmeier This is done! After working with @leemunroe, here's what we came up with:

orlandohohmeier:
@jfurrow Looks great, and works well. Great job! 💯 ✨

PR #1339, opened 2016-11-10
Fix environment section

Poltergeist:
This is removing some remains from the implementation of the multiple items. As the concept of the form reducers changed once but not all the associated code has been removed this is fixed now.


plugins/services/src/js/components/forms/EnvironmentFormSection.js
```diff
@@ -4,20 +4,10 @@ import FieldError from '../../../../../../src/js/components/form/FieldError';
...
-import ReducerUtil from '../../../../../../src/js/utils/ReducerUtil';
 import {FormReducer as env} from '../../reducers/serviceForm/EnvironmentVariables';
 import {FormReducer as labels} from '../../reducers/serviceForm/Labels';
 
 class EnvironmentFormSection extends Component {
-  constructor() {
-    super(...arguments);
-
-    this.state = {
-      reducers: ReducerUtil.combineReducers(
```
mlunoe: 👍 I was wondering about this!

PR #1348, opened 2016-11-11
Add collapsible component

Poltergeist:
With this component it is possible to create a
section which is collapsible and is hidden behind
a label this label can be configured with the
label prop. The label prop expects a string

Example:

```
<CollapsibleContainer label="Advanced Setting">
  ... Some Content
</CollapsibleContainer>
```

Closed:
image

Open:
image

### DIFF
--- a/src/js/components/CollapsibleContainer.js
+++ b/src/js/components/CollapsibleContainer.js
@@ -1,0 +1,57 @@
+import React, {Component} from 'react';
+
+import Icon from './Icon';
+
+class CollapsibleContainer extends Component {
+  constructor() {
+    super(...arguments);
+
+    this.state = {
+      toggle: false
+    };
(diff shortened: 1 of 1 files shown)


src/js/components/CollapsibleContainer.js
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+}
+
+CollapsibleContainer.defaultProps = {
+  label: ''
+};
+
+CollapsibleContainer.propTypes = {
+  label: React.PropTypes.string
```
nLight: Should we require a label?

orlandohohmeier: Let's make this a required prop!

Poltergeist:
@orlandohohmeier @nLight Comment haz been addressed.

orlandohohmeier:
LGTM! 🚢

PR #1349, opened 2016-11-12
[GraphQL 1] Initial GraphQL project scaffold

MatApple:
## Initial scaffold of GraphQL service

Please see https://github.com/dcos/dcos-ui/pull/1350 for an overview of the GraphQL service.

In this PR:

- Configuration for various runtimes
- New npm packages needed for the service. We use Koa for the server which will come in another PR. I've added all dependencies in this PR so we avoid tracking and rebasing multiple shrinkwrap files across PRs.
- Updates to jest preprocessor for testing GraphQL files
- New scripts for exporting the schemas

PR #1353, opened 2016-11-12
[GraphQL 3] Adding data connectors for GraphQL service

MatApple:
## Adds connectors to data for various runtimes.
**:warning: Depends on https://github.com/dcos/dcos-ui/pull/1349**

Structure:
```
data/
  mesos/
    state/
      client.js -> connects to client-side MesosStateActions (TODO)
      server.js -> connects to cluster '/mesos/master/state'
      mock.js -> connects to fixture
      state.js -> exports connector for specified runtime
    MesosConnector.js -> exports all mesos connectors
  marathon/
    deployments/
      client.js -> connects to client-side MarathonActions (TODO)
      server.js -> connects to cluster '/marathon/v2/deployments'
      mock.js -> connects to fixture
      deployments.js -> exports connector for specified runtime
    groups/
      client.js -> connects to client-side MarathonActions (TODO)
      server.js -> connects to cluster '/marathon/v2/groups'
      mock.js -> connects to fixture
      group.js -> exports connector for specified runtime
    MarathonConnector.js -> exports all marathon connectors for runtime
  EndpointsConnector.js -> exports all connectors
```
We use **\_\_mocks__** in each directory to provide a convenient testing strategy for components closer to the query such a resolvers, models, and schema introspection.

**Server connectors use [DataLoader](https://github.com/facebook/dataloader) for per-request caching. A new set of connectors are created on every request.**


graphql/data/marathon/deployments/server.js
```diff
@@ -0,0 +1,19 @@
...
+
+    const getDeployments = fetchWithAuth(authToken, {
+      baseURI: config.endpoints.deployments
+    });
+
+    this.deploymentsLoader = new DataLoader(getDeployments, { batch: false });
+  }
+
+  get() {
+    return this.deploymentsLoader.load('');
```
nLight: So by providing an empty string as a key we will be returning the same value for each request, right?

graphql/data/marathon/groups/server.js
```diff
@@ -0,0 +1,40 @@
...
+
+    const getGroup = fetchWithAuth(authToken, {
+      baseURI: config.endpoints.marathon.groups,
+      query
+    });
+
+    this.groupsLoader = new DataLoader(getGroup, { batch: false });
+  }
+
+  get(groupId = '/') {
```
nLight: having '/' by default and then clearing the value right at the beginning of the function looks weird,
but then most probably you can either call it either with `/` or without any arguments, right?

graphql/data/marathon/deployments/client.js
```diff
@@ -0,0 +1,14 @@
...
+/**
+ * Currently using a mock response.
+ *
+ * This will connect to our MarathonActions file to trigger requests
+ * and listen to action responses via the AppDispatcher.
+ */
+
+export default class ClientDeploymentsConnector {
+  get() {
+    return Promise.resolve(mockResponse);
```
MatApple: Thanks @nLight! Good catch. This file is a placeholder for the time being because I'm going to wire it up to the MarathonActions. But we'll still use clone so I'll fix it for consistency.

graphql/data/marathon/deployments/server.js
MatApple: Correct. The value we pass in here also gets appended to the base URL. So we could do `getById` and pass in 'deploymentId` and it will cache that value for the `deploymentId` too.

graphql/data/marathon/groups/server.js
MatApple: Yes the reason I put the default argument in is so it can be called with no arguments. Otherwise we explicitly have to query with `groupId: '/'` which might not be such a bad thing anyway. I can remove the default.

MatApple:
Thanks @nLight!

MatApple:
Closing to reduce repo noise and defer for future. Do not delete branch.

PR #1359, opened 2016-11-14
[GraphQL 6] Adding Schemas and Resolvers for GraphQL

MatApple:
Adds schemas and resolvers for:
- Agents
- Cluster
- Applications
- Frameworks
- Groups
- Tasks

More Types to come and more fields on the above types...


MatApple:
Closing to reduce repo noise and defer for future. Do not delete branch.

PR #1361, opened 2016-11-14
JSONUtil: Adding `position` attribute on JSON tokens

wavesoft:
This PR adds a `pos` attribute on the JSON tokens, containing the starting, ending offset of each token in the source string. This is particularly useful on tokens that are on the same line.

### DIFF
--- a/src/js/utils/JSONUtil.js
+++ b/src/js/utils/JSONUtil.js
@@ -12,10 +12,6 @@ const TOKENIZER_TOKENS = {
   'null': /null/g,
   'boolean': /(true|false)/g,
   'number': /-?\d+(\.\d+)?([eE]-?\d+)?/g,
-  'maybe-decimal-number': /-?\d+\./g,
-  'maybe-negative-number': /-/g,
-  'maybe-exponential-number': /-?\d+(\.\d+)?([eE])?/g,
-  'maybe-exponential-number-negative': /-?\d+(\.\d+)?([eE]-)?/g,
   'symbol': /\w+/g
 };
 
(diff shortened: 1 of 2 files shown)


src/js/utils/JSONUtil.js
```diff
@@ -155,23 +154,31 @@ module.exports = {
         case 'begin-object':
           // Keep track of keyed objects
           if (objectKey) {
-            keyLines.push({
+            let blockToken = {
```
nLight: I see a repeating pattern here. Massive Switch/case is a different story, but maybe we could think a way to generalize this token creation, how do you think?

src/js/utils/JSONUtil.js
```diff
@@ -155,23 +154,31 @@ module.exports = {
...
               path: [].concat(path, objectKey),
               line: lineNo,
-              type: 'object'
-            });
+              type: 'object',
+              pos: [offset, i]
```
nLight: Let's not shorten `position`

src/js/utils/JSONUtil.js
wavesoft: I am open to suggestions, since I am currently focused on another task. From a quick look that I had it's a bit tricky to extract a re-usable function without carrying almost every variable in scope along...

wavesoft:
Addressed the `position` issue, however I couldn't come up with a summarising solution for the first comment. If there are no other ideas or objections, please approve for commit because it's needed for others 😄

wavesoft:

src/js/utils/JSONUtil.js
```diff
@@ -224,6 +239,12 @@ module.exports = {
           } else {
             arrayIndex = -1;
           }
+
+          // Update the ending position of the last block token on stack
+          let lastBlockToken = blockTokens.pop();
```
MatApple: Curious why we are creating a stack to track the current `blockToken` (aka `lastBlockToken`) and why it's not just assigned to a variable?

wavesoft: Because if you open two or more block tokens, you will loose track of them since you will always look at the last one :wink: (Unless I got the question wrong)

MatApple: Ahh you're right - I missed that this was inside the while loop 👍

PR #1362, opened 2016-11-14
DCOS-9827: Add a new auth package

orlandohohmeier:
---

⚠️ This PR depends on changes that will get introduced with #1130 

---

Introduce an auth service as well as an auth component.

* The service provides a method to authorize the current user, using the registered `Authorizer`, if the required permission was granted. 
* The component will show and hide elements based on user permissions.

**Example** 

```
// Custom Authorizer 
  class SimpleAuthorizer extends Authorizer {
    authorize(permission) {
      return userPermissions.indexOf(permission) > -1;
    }
  }

// Registration
  AuthService.registerAuthorizer(new SimpleAuthorizer());

// Authorize
  if(AuthService.authorize('some:permission')){
    console.log('authorized')
  };

// Component
  <Auth permission="*">
      ... Some Content
  </Auth>
```

### DIFF
--- a/foundation-ui/auth/AuthService.js
+++ b/foundation-ui/auth/AuthService.js
@@ -30,6 +30,14 @@ class AuthService extends EventEmitter {
       return;
     }
 
+    if (authorizers.includes(authorizer)) {
+      if (global.__DEV__) {
+        throw new Error('Provided authorizer is already registered');
+      }
+
+      return;
+    }
+
     authorizers.push(authorizer);
     this.emit(CHANGE);
   }
--- a/foundation-ui/auth/AuthService.js
+++ b/foundation-ui/auth/AuthService.js
@@ -1,0 +1,8 @@
+import {EventEmitter} from 'events';
+
+/**
+ * AuthService
+ */
+class AuthService extends EventEmitter {}
+
+module.exports = new AuthService();
--- a/foundation-ui/auth/AuthService.js
+++ b/foundation-ui/auth/AuthService.js
@@ -1,5 +1,7 @@
 import {EventEmitter} from 'events';
+
 import Authorizer from './Authorizer';
+import {CHANGE} from './AuthEvent';
 
 /**
  * List of authorizers
@@ -29,6 +31,7 @@ class AuthService extends EventEmitter {
     }
 
     authorizers.push(authorizer);
+    this.emit(CHANGE);
   }
 
   /**
@@ -44,6 +47,8 @@ class AuthService extends EventEmitter {
         break;
       }
     }
+
+    this.emit(CHANGE);
   }
 
   /**
--- a/foundation-ui/auth/AuthService.js
+++ b/foundation-ui/auth/AuthService.js
@@ -31,6 +31,21 @@ class AuthService extends EventEmitter {
     authorizers.push(authorizer);
   }
 
+  /**
+   * Unregisters authorizer
+   *
+   * @param  {Authorizer} authorizer
+   */
+  unregisterAuthorizer(authorizer) {
+    let i = authorizers.length;
+    while (--i >= 0) {
+      if (authorizers[i] === authorizer) {
+        authorizers.splice(i, 1);
+        break;
+      }
+    }
+  }
+
 }
 
 module.exports = new AuthService();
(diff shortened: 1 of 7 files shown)


orlandohohmeier:

foundation-ui/auth/__tests__/Auth-test.js
```diff
@@ -0,0 +1,120 @@
...
+    AuthService.registerAuthorizer(unauthorized);
+    const result = TestUtils.renderIntoDocument(
+        <Auth permission="*">
+          <span>foo</span>
+        </Auth>
+    );
+
+    AuthService.unregisterAuthorizer(unauthorized);
+
+    expect(TestUtils.findRenderedDOMComponentWithTag(result, 'span'))
```
MatApple: Lets check the result before before unregistering the authorizer to prove something changed :)

foundation-ui/auth/AuthService.js
```diff
@@ -0,0 +1,67 @@
...
+   *
+   * @returns {boolean} authorized
+   */
+  authorize(permission = '') {
+    return authorizers.every((authorizer) => authorizer.authorize(permission));
+  }
+
+}
+
+module.exports = new AuthService();
```
MatApple: I don't think we shouldn't export a singleton here. That way we can run tests against new instances of AuthService. Otherwise there is no safe way to flush all the authorizers from the system for robust testing. We should export the singleton in `foundation-ui/auth/index.js` and perhaps export the `Auth` component wrapped in a stateless functional component that pushes this singleton "AuthService" (and other props) into the `Auth` component.

foundation-ui/auth/AuthService.js
```diff
@@ -0,0 +1,67 @@
...
+  /**
+   * Register authorizer
+   *
+   * @param {Authorizer} authorizer
+   */
+  registerAuthorizer(authorizer) {
+    if (!authorizer || !(authorizer instanceof Authorizer)) {
+      if (global.__DEV__) {
+        throw new TypeError('Provided component must be a ' +
+            'React.Component constructor or a stateless functional component.');
```
orlandohohmeier: Whoops, I completely forgot to adjust the error message...

foundation-ui/auth/AuthService.js
```diff
@@ -0,0 +1,67 @@
...
+ */
+class AuthService extends EventEmitter {
+
+  /**
+   * Register authorizer
+   *
+   * @param {Authorizer} authorizer
+   */
+  registerAuthorizer(authorizer) {
+    if (!authorizer || !(authorizer instanceof Authorizer)) {
```
orlandohohmeier: It ensures the all  authorizer are objects with a method named `authorize` which will return a boolean.

foundation-ui/auth/AuthService.js
orlandohohmeier: You are right, this singleton pattern makes it rather hard to test the service properly. I thought about introducing a new pattern but decided to follow our current "standard" way for consistency.

I started a document back in May illustrating a few alternative patterns and shared it with @dcos/frontend and would love your feedback as we finally need to agree on a new pattern due to the mentioned reasons.

foundation-ui/auth/AuthService.js
```diff
@@ -0,0 +1,67 @@
...
+   * Unregisters authorizer
+   *
+   * @param  {Authorizer} authorizer
+   */
+  unregisterAuthorizer(authorizer) {
+    let i = authorizers.length;
+    while (--i >= 0) {
+      if (authorizers[i] === authorizer) {
+        authorizers.splice(i, 1);
+        break;
```
orlandohohmeier: Good point, I'll make sure that it is not possible to register an authorizer multiple times.

foundation-ui/auth/AuthService.js
MatApple: The `authorize` method doesn't have to return a boolean, just like a regular function, it could return anything too. I'm ok with the proposal - just wanted to understand the extra layer required of extending from an existing class :)

MatApple:
Looks great @orlandohohmeier - nice work!

orlandohohmeier:
ℹ️ I'm going to align service with the singleton pattern used in the navigation and routing service as it's a good first step towards a better pattern.

orlandohohmeier:
Going to close this as this new package isn't properly solving our problems and we might opt for a different/better solution.

PR #1365, opened 2016-11-15
Tying new service create edit flow together

mlunoe:
This PR ties current efforts on the create/edit flow for services together, see it in action here:

Here is the changes you need to make to `ServiceModals` to activate the new flow:
```
- import ServiceFormModal from './ServiceFormModal';
+ import NewCreateServiceModal from './NewCreateServiceModal';
@@ -65,7 +65,7 @@ class ServiceModals extends React.Component {
    };

    return (
-      <ServiceFormModal
+      <NewCreateServiceModal
@@ -102,10 +102,10 @@ class ServiceModals extends React.Component {
    }

    return (
-      <ServiceFormModal
+      <NewCreateServiceModal
-        isEdit={false}
+        isEdit={true}
```

### DIFF
--- a/plugins/services/src/js/components/modals/NewCreateServiceModal.js
+++ b/plugins/services/src/js/components/modals/NewCreateServiceModal.js
@@ -13,6 +13,7 @@ import Util from '../../../../../../src/js/utils/Util';
 
 const METHODS_TO_BIND = [
   'handleGoBack',
+  'handleClose',
   'handleJSONToggle',
   'handleServiceSelection',
   'handleServiceReview',
@@ -23,18 +24,7 @@ class NewServiceFormModal extends Component {
   constructor() {
     super(...arguments);
 
-    this.state = {
-      isJSONModeActive: false,
-      serviceFormActive: false,
-      servicePickerActive: true,
-      serviceReviewConfig: null
-    };
-
-    // Switch directly to form if edit
-    if (this.props.isEdit) {
-      this.state.servicePickerActive = false;
-      this.state.serviceFormActive = true;
-    }
+    this.state = this.getResetState();
 
     METHODS_TO_BIND.forEach((method) => {
       this[method] = this[method].bind(this);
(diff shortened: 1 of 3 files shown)


plugins/services/src/js/components/modals/NewCreateServiceModal.js
```diff
@@ -1,51 +1,166 @@
...
+            actions={this.getPrimaryActions()}
+            type="primary" />
+        </FullScreenModalHeader>
+      );
+    }
+
+    let title = 'Run a Service';
+    let {isEdit, service} = this.props;
+    let serviceName = service.getName() || 'Service';
+    if (serviceName) {
```
MatApple: This will always be `true` because of the `||` above. So let's get rid of some code here.

MatApple: Also, since we only use serviceName if `isEdit`, let's put this `serviceName` code inside the `if (isEdit)` block

plugins/services/src/js/components/modals/NewCreateServiceModal.js
```diff
@@ -47,7 +47,7 @@ class NewServiceFormModal extends Component {
       serviceReviewConfig
     } = this.state;
 
-    // Close if clicker is open, or if editing a service in the form
+    // Close if picker is open, or if editing a service in the form
```
mlunoe: 👍

plugins/services/src/js/components/modals/NewCreateServiceModal.js
mlunoe: Removed the `|| 'Service'` as that was a mistake. Had a conversation offline and I decided to keep the `serviceName` outside of the `if (isEdit)` to keep it easier to parse.

MatApple:
Once the conflicts have been fixed and tests pass, lets 🚢

PR #1369, opened 2016-11-15
Introducing AdvancedSection component (with helper components)

mlunoe:
This PR introduces an AdvancedSection component to be used with forms:

Here are the changes you need to make to `ServiceModals` to activate the new flow:
```
import NewCreateServiceModal from './NewCreateServiceModal';
...
     <NewCreateServiceModal
...
     <NewCreateServiceModal
...
       isEdit={true}
```


Poltergeist:
I created a collapsible container component https://github.com/dcos/dcos-ui/pull/1348
Which has far less option then this component I do :heart: your component but I do think that the name is to specific it could also be used for different things. Could we change the name to collapsible content / section and remove the "old" component.

mlunoe:
Removed the old, had a conversation offline with @Poltergeist about keeping the name, since it is going to be only used in the form for now. If we need it elsewhere, we can rename the component

plugins/services/src/js/components/forms/GeneralServiceFormSection.js
```diff
@@ -189,15 +140,34 @@ class GeneralServiceFormSection extends Component {
...
-            size="tiny" />
-          Advanced Service Settings
-        </a>
-        {this.getContainerRuntime(data, errors.container)}
+        <AdvancedSection>
+          <AdvancedSectionLabel>
+            Advanced Service Settings
+          </AdvancedSectionLabel>
+          <AdvancedSectionContent>
+          <h3 className="short-top short-bottom">
```
jfurrow: This should be indented once more.

src/styles/components/advanced-section/styles.less
```diff
@@ -0,0 +1,16 @@
...
+      margin-top: @pod-margin-top;
+    }
+
+    &-label,
+    &-label:hover {
+      text-decoration: none;
+
+      .icon {
+        margin-right: @base-spacing-unit * 1/4;
+      }
```
jfurrow: Let's adjust this block as follows:

```less
&-label {
  
  &,
  &:hover {
    text-decoration: none;
  }

  .icon {
    margin-right: @base-spacing-unit * 1/4;
  }
}
```
This way we're not unnecessarily duplicating the properties for the nested `.icon` selector.

src/styles/components/advanced-section/styles.less
```diff
@@ -0,0 +1,16 @@
+& when (@advanced-section-enabled) {
+  .advanced-section {
+    &-content {
```
jfurrow: Let's get a newline above this. The linter should catch this, I wonder if it's not running...

src/styles/components/advanced-section/styles.less
```diff
@@ -0,0 +1,16 @@
+& when (@advanced-section-enabled) {
+  .advanced-section {
```
jfurrow: Let's get a newline above this. The linter should catch this, I wonder if it's not running...

src/styles/components/advanced-section/styles.less
```diff
@@ -0,0 +1,16 @@
...
+    &-label,
+    &-label:hover {
+      text-decoration: none;
+
+      .icon {
+        margin-right: @base-spacing-unit * 1/4;
+      }
+    }
+  }
+}
```
jfurrow: Let's add the media queries to this and adjust the `.advanced-section-content` margin with the proper `@pod-margin-top-screen-*` and `@base-spacing-unit-screen-*` variables.

mlunoe: Oh dang - :facepalm:. Thanks for catching this

src/styles/components/advanced-section/styles.less
mlunoe: Done!

src/styles/components/advanced-section/styles.less
mlunoe: Done!

plugins/services/src/js/components/forms/GeneralServiceFormSection.js
mlunoe: Done!

mlunoe:
@jfurrow All comments should be addressed!

Poltergeist:
This looks good to me.

PR #1372, opened 2016-11-16
Activating new create/edit service flow

mlunoe:
This PR activates the new create/edit service flow:


MatApple:
@mlunoe - I think we should rename the component and get rid of the leading `New`. What do you think?

mlunoe:
We're keeping the old modal in here, so we can switch back if we need to, until it is removed, let's keep the name

rcorral:
Let's keep the old one for now, until we have full functionality with the new create flow.

MatApple:
@mesosphere-ci test this please

MatApple:
@mesosphere-ci test this please

PR #1373, opened 2016-11-16
[Rejected Offers: 1/2] DCOS-10994: Implement declined offers API response transformation

jfurrow:
This PR implements the transformation of the `/queue` API response into a structure that can be consumed by https://github.com/dcos/dcos-ui/pull/1338


MatApple:
@mesosphere-ci test this please

plugins/services/src/js/utils/DeclinedOffersUtil.js
```diff
@@ -0,0 +1,79 @@
...
+        offers: matchedConstraintOffers,
+        matched: matchedPortOffers
+      }
+    };
+  },
+
+  getOffersFromQueue(queue) {
+    let {lastUnusedOffers} = queue;
+
+    if (lastUnusedOffers == null || lastUnusedOffers.length === 0) {
```
MatApple: What do you think about defaulting `lastUnsedOffers` to `[]` like:
```
let {lastUnsedOffers = []} = queue;

if (!lastUnusedOffers.length) {
  return null;
}
```

plugins/services/src/js/utils/DeclinedOffersUtil.js
```diff
@@ -0,0 +1,79 @@
...
+      return null;
+    }
+
+    return lastUnusedOffers.map((declinedOffer) => {
+      return {
```
MatApple: You can also do
```
lastUnusedOffers.map((declinedOffer) => ({hostname... etc}));
```
so you don't explicitly need the return statement :)

plugins/services/src/js/utils/__tests__/DeclinedOffersUtil-test.js
```diff
@@ -0,0 +1,132 @@
...
+      expect(summary.mem.offers).toEqual(9);
+      expect(summary.mem.matched).toEqual(9);
+      expect(summary.disk.offers).toEqual(9);
+      expect(summary.disk.matched).toEqual(9);
+      expect(summary.ports.offers).toEqual(9);
+      expect(summary.ports.matched).toEqual(9);
+    });
+
+    it('transforms the API response into the expected format', function () {
+      const summary = DeclinedOffersUtil.getSummaryFromQueue(QueueFixture.queue[1]);
```
MatApple: It would be really nice if we didn't use the fixtures for testing and specified a slice of the fixture here in the test. That way test code is tightly coupled to the fields it expects as inputs and the fields it returns as an output. It's also a little more future proof because these fixtures might disappear with GraphQL. I also think it's more readable to look at the tests and see what the example input looks like. What do you think?

plugins/services/src/js/utils/DeclinedOffersUtil.js
```diff
@@ -0,0 +1,79 @@
...
+
+    if (processedOffersSummary == null
+      || processedOffersSummary.unusedOffersCount === 0
+      || lastUnusedOffers == null
+      || lastUnusedOffers.length === 0) {
+      return null;
+    }
+
+    const {processedOffersCount = 0, rejectReason = {}} = processedOffersSummary;
+    const matchedRoleOffers = processedOffersCount -
```
jfurrow: The API for this has changed, so we don't have to do this funnel-down math anymore (it all happens in the backend). We still have to do some math to figure out the number of matched offers, but it's more simple.

We take the number of processed offers and subtract the number of declined offers to get the number of matched offers.

plugins/services/src/js/utils/DeclinedOffersUtil.js
```diff
@@ -0,0 +1,79 @@
...
+      (rejectReason[DeclinedOffersReasons.UNFULFILLED_ROLE] || 0);
+    const matchedConstraintOffers = matchedRoleOffers -
+      (rejectReason[DeclinedOffersReasons.UNFULFILLED_CONSTRAINT] || 0);
+    const matchedCPUOffers = matchedConstraintOffers -
+      (rejectReason[DeclinedOffersReasons.INSUFFICIENT_CPU] || 0);
+    const matchedMemOfffers = matchedConstraintOffers -
```
jfurrow: Hahaha, good catch... this awesome variable is no longer needed after the API changes :)

plugins/services/src/js/utils/DeclinedOffersUtil.js
jfurrow: I like it! Done!

plugins/services/src/js/utils/DeclinedOffersUtil.js

plugins/services/src/js/utils/__tests__/DeclinedOffersUtil-test.js
jfurrow: Agreed, done!

jfurrow:
Thanks for the review @MatApple, I addressed all your comments and updated this to work with the latest API changes.

mlunoe:
Looks good! Let's not let this be held back by @MatApple being out. @philipnrmn mind giving this a look, if you have time?

jfurrow:
Making a few additional changes — sorry! Adding `hold merge` for now.

jfurrow:
@mlunoe @philipnrmn This should be ready for review, now :)

philipnrmn:
This is great!

tobilg:
As a question, will the `/queue` "only" cover the tasks that are started by Marathon directly? Or also the case like the Kafka framework, where Marathon starts the scheduler, and the scheduler starts its own broker tasks? Thanks!

mlunoe:
@jfurrow ^

jfurrow:
@tobilg It is only for tasks started by Marathon directly.

PR #1374, opened 2016-11-16
Fix ServiceScaleFormModal textbox

jfurrow:
This PR fixes the width of the input field in the `ServiceScaleFormModal`. The `column-2` class was constraining the width of the element which was already inside a `column-12` element.

Before:

After:


mlunoe:
:shipit: once tests pass

PR #1375, opened 2016-11-16
Exclude null reducer values from state

jfurrow:
This PR will exclude null values from state in `#combineReducers`.

For the networking configuration, in some instances we want to remove keys from the root of the configuration and move them elsewhere (e.g. `portMappings` exists at the root in some situations and other times it exists at `container.docker`).

@Poltergeist Could you please review this and see if this is reasonable? If it is not reasonable, do you have other ideas for how we could exclude some keys from the app config?

### DIFF
--- a/src/js/utils/ReducerUtil.js
+++ b/src/js/utils/ReducerUtil.js
@@ -56,7 +56,11 @@ module.exports = {
         // reducer.bind(context.get(reducer))(state[key], action);
         // but it will not copy the function, which in this case is a huge
         // increase in performance.
-        localState[key] = reducer.call(context.get(reducer), localState[key], action);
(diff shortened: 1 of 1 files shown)


jfurrow:
retest this please

philipnrmn:
I'm OK with the code but don't really love mutating the config object - I get why we're doing it here but it seems somehow inelegant. I'd be interested to hear @Poltergeist 's take.

wavesoft:
@jfurrow I had exactly the same idea while playing with the form yesterday. A proper solution is a bit more complicated, I will also crate a PR with it today.

jfurrow:
Thanks @philipnrmn and @wavesoft! I also spoke with @Poltergeist offline and will happily close this PR :)

PR #1377, opened 2016-11-17
Remove obsolete `system-organization-tabs` filter

nLight:
Since we've split system page into several pages and it doesn't exist anymore we don't need this filter.


nLight:
retest this please

PR #1382, opened 2016-11-18
DCOS-10991 [2 of 4] Add ExternalVolumes Reducer

Poltergeist:
This is part of DCOS-10991 and adds the reducer
and parser for external volumes.

### DIFF
--- a/plugins/services/src/js/reducers/serviceForm/ExternalVolumes.js
+++ b/plugins/services/src/js/reducers/serviceForm/ExternalVolumes.js
@@ -1,0 +1,129 @@
+import {
+  ADD_ITEM,
+  REMOVE_ITEM,
+  SET
+} from '../../../../../../src/js/constants/TransactionTypes';
+import Transaction from '../../../../../../src/js/structs/Transaction';
+
+module.exports = {
+  JSONParser(state) {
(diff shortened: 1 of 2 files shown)


plugins/services/src/js/reducers/serviceForm/ExternalVolumes.js
```diff
@@ -0,0 +1,129 @@
...
+  },
+
+  FormReducer(state = [], {type, path, value}) {
+    if (path == null) {
+      return state;
+    }
+
+    let joinedPath = path.join('.');
+
+    if (joinedPath.search('externalVolumes') !== -1) {
```
wavesoft: What about `path[0] !== 'externalVolumes'`?

plugins/services/src/js/reducers/serviceForm/ExternalVolumes.js
```diff
@@ -0,0 +1,129 @@
...
+            state = state.filter((item, index) => {
+              return index !== value;
+            });
+            break;
+        }
+
+        return state;
+      }
+
+      let index = joinedPath.match(/\d+/)[0];
```
wavesoft: What about `index = path[1]`?

Poltergeist:
Addressed all the comments

PR #1384, opened 2016-11-18
DCOS-11605: Introducing applyPatch utility function to the form

wavesoft:
This PR introduces a utility function used by the `NewCreateServiceModalForm` to update the JSON content in the smartest way possible. This happens in two phases:

1. The form uses the batch reducers to create a "UI patch" to the JSON
2. It uses the `applyPatch` function apply the patch to the current JSON state

This function follows these principles:

1. An "empty" value in the patch removes the field.
2. The previous statement is not applied when the respective `data` field already contains an "empty" value. In that case, the original "empty" data value is perserved.
3. If both `data` and `patch` have a value in a field, but with different types, the value from `data` is preferred.

### DIFF
--- a/plugins/services/src/js/utils/CreateServiceModalFormUtil.js
+++ b/plugins/services/src/js/utils/CreateServiceModalFormUtil.js
@@ -1,5 +1,6 @@
-const CreateServiceModalFormUtil = {
+import ValidatorUtil from '../../../../../src/js/utils/ValidatorUtil';
 
+const CreateServiceModalFormUtil = {
   /**
    * Apply patch data on the given source data, following the principles:
    *
@@ -39,8 +40,8 @@ const CreateServiceModalFormUtil = {
       // If the patch contains an empty value we have to remove the item,
       // with the only exception of the source value having already an empty
       // value in place.
-      if (CreateServiceModalFormUtil.isEmptyValue(patch[key])) {
-        if (!CreateServiceModalFormUtil.isEmptyValue(memo[key])) {
+      if (ValidatorUtil.isEmpty(patch[key])) {
+        if (!ValidatorUtil.isEmpty(memo[key])) {
           delete memo[key];
         }
 
@@ -51,24 +52,7 @@ const CreateServiceModalFormUtil = {
 
       return memo;
     }, Object.assign({}, data));
-  },
-
-  /**
-   * Tests if the given value is "empty". This means:
-   *
-   * - `null` or `undefined`
-   * - Empty string
-   * - Empty object
-   * - Empty array
-   *
-   * @param {*} value - A value to test
-   * @return {Boolean} True if the value is empty
-   */
-  isEmptyValue(value) {
-    return (value == null) || (value === '') ||
-           ((typeof value === 'object') && (Object.keys(value).length === 0));
   }
-
 };
 
 module.exports = CreateServiceModalFormUtil;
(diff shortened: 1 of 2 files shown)


plugins/services/src/js/utils/CreateServiceModalFormUtil.js
```diff
@@ -0,0 +1,74 @@
...
+   *
+   * 3. If both `data` and `patch` have a value in a field, but with different
+   *    types, the value from `data` is preferred.
+   *
+   * @param {Object} data - The source data to patch
+   * @param {Object} patch - The patch to apply with
+   * @returns {Object} The patched data response
+   */
+  applyPatch(data, patch) {
+    // If we don't have this in data, prefer patch
```
nLight: Clarification -> `// If we don't have data, prefer patch`

plugins/services/src/js/utils/CreateServiceModalFormUtil.js
```diff
@@ -0,0 +1,74 @@
...
+      return data;
+    }
+
+    // Non-object types just pass through
+    if (typeof patch !== 'object') {
+      return patch;
+    }
+
+    // Walk object types
+    let keys = Object.keys(patch);
```
nLight: No need in extra variable. Does it fit 80 chars?

plugins/services/src/js/utils/CreateServiceModalFormUtil.js
```diff
@@ -0,0 +1,74 @@
...
+   *
+   * - `null` or `undefined`
+   * - Empty string
+   * - Empty object
+   * - Empty array
+   *
+   * @param {*} value - A value to test
+   * @return {Boolean} True if the value is empty
+   */
+  isEmptyValue(value) {
```
nLight: Let's reuse `ValidatorUtils.isEmpty()`

plugins/services/src/js/utils/CreateServiceModalFormUtil.js
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    // Walk object types
+    let keys = Object.keys(patch);
+    return keys.reduce(function (memo, key) {
+
+      // If the patch contains an empty value we have to remove the item,
+      // with the only exception of the source value having already an empty
+      // value in place.
+      if (CreateServiceModalFormUtil.isEmptyValue(patch[key])) {
```
nLight: Let's reuse `ValidatorUtils.isEmpty()`

plugins/services/src/js/utils/__tests__/CreateServiceModalFormUtil-test.js
```diff
@@ -0,0 +1,159 @@
...
+    it('should preserve null source type if {} is given', function () {
+      let data = {a: null};
+      let patch = {a: {}};
+      let patched = CreateServiceModalFormUtil.applyPatch(data, patch);
+      expect(patched).toEqual({a: null});
+    });
+
+  });
+
+  describe('#isEmptyValue', function () {
```
nLight: :fire: :)

plugins/services/src/js/utils/CreateServiceModalFormUtil.js
wavesoft: 😮  You are right, I forgot about it.

wavesoft:
Good catch, rebased & addressed all comments! 👍

wavesoft:
retest this please 🙇

PR #1391, opened 2016-11-21
Add the capability to save HTTPS for healthChecks

Poltergeist:
If you switch to HTTP protocol and check the HTTPS checkbox and then switch to protocol COMMAND and then beck to HTTP the HTTPS setting would be lost. This PR adds the feature to keep the value saved.


wavesoft:
Works like a charm! ❤️  ✨

Poltergeist:
I added two tests.

PR #1392, opened 2016-11-21
DCOS-11638: Fixing recursive application of applyPatch

wavesoft:
This PR makes sure `CreateServiceModalFormUtil.applyPatch` is called recursively in order to remove objects that become empty after the first pass. For example:

For example given an input:
```js
{
  "object": {
    "value": ""
  }
}
```

Previously was producing:
```js
{
  "object": {
  }
}
```

But now it produces:
```js
{
}
```


nLight:
LGTM 🚢

PR #1394, opened 2016-11-21
Add z-index to FormGroupContainer's remove button

jfurrow:
This PR:
* Renames `.form-remove` to `.form-group-container-action-button`
* Gives a z-index to `.form-group-container-action-button` to prevent it from rendering below sibling elements

### DIFF
--- a/src/js/components/form/FormGroupContainer.js
+++ b/src/js/components/form/FormGroupContainer.js
@@ -6,7 +6,7 @@ const FormGroupContainer = (props) => {
   let removeButton = null;
   if (props.onRemove != null) {
     removeButton = (
-      <div className="form-group-container-action-button">
+      <div className="form-group-container-action-button-group">
         <a className="button button-primary-link"
           onClick={props.onRemove}>
           <Icon id="close" color="grey" size="tiny"/>
--- a/src/js/components/form/FormGroupContainer.js
+++ b/src/js/components/form/FormGroupContainer.js
@@ -6,7 +6,7 @@ const FormGroupContainer = (props) => {
   let removeButton = null;
   if (props.onRemove != null) {
     removeButton = (
-      <div className="form-remove">
+      <div className="form-group-container-action-button">
         <a className="button button-primary-link"
           onClick={props.onRemove}>
           <Icon id="close" color="grey" size="tiny"/>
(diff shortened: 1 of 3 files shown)


src/js/components/form/FormGroupContainer.js
```diff
@@ -6,7 +6,7 @@ const FormGroupContainer = (props) => {
   let removeButton = null;
   if (props.onRemove != null) {
     removeButton = (
-      <div className="form-remove">
+      <div className="form-group-container-action-button">
```
mlunoe: Should this be `form-group-actions-container`?

src/styles/components/form-elements/styles.less
```diff
@@ -35,9 +41,20 @@
     position: absolute;
   }
 
-  .form-remove {
+  .form-group-container-action-button {
```
mlunoe: `form-group-actions-container`?

src/styles/variables/variables-z-index.less
```diff
@@ -5,6 +5,7 @@
 @z-index-dygraph-chart: @z-index-side-panel + 10;
 @z-index-dygraph-hover-label: 100;
 @gemini-scrollbar-handle: @z-index-header-tabs + 1;
+@z-index-form-group-container-remove-button: 1;
```
mlunoe: Should this be `@z-index-form-group-actions-container`?

src/styles/components/form-elements/styles.less
jfurrow: It's a bit funky to read but it's actually used inside of the `FormGroupContainer` component, so I do think it makes sense... do you still think it should be changed?

mlunoe: Ah, ok. But it is not on the action button? It's a wrapper around the action button, right?

jfurrow: That's true, the element with this class is not the actual button. And I could see it containing more than one button.

What do you think about `.form-group-container-action-button-group`? It's a bit wordy but it does make sense to me...

philipnrmn:
I can't actually force the remove button to render under its siblings without these changes, but this looks good to me in any case.

orlandohohmeier:
@mlunoe @jfurrow I've adjusted the class and var names as suggested.

PR #1396, opened 2016-11-21
Pass checked prop to checkbox

jfurrow:
This PR passes the `checked` prop when `FieldInput`'s type is `checkbox`.

This is required for maintaining a checkbox's state if React reuses the `input` DOM node where the new checkbox's value differs from the previous.


jfurrow:
After speaking with @mlunoe offline, we've decided to close this PR and just pass in the `checked` attribute.

PR #1400, opened 2016-11-22
DCOS-9609: Pass target param down to the auth service

nLight:
This PR allows us to provide a specific `target` when we perform a login action.
Which will be handled by the `IAM` service

### DIFF
--- a/src/js/events/AuthActions.js
+++ b/src/js/events/AuthActions.js
@@ -10,9 +10,14 @@ import AppDispatcher from '../events/AppDispatcher';
 import Config from '../config/Config';
 
 const AuthActions = {
-  login(credentials) {
+  login(credentials, target) {
+    let query = '';
+    if (target) {
+      query = `?target=${encodeURIComponent(target)}`;
+    }
+
     RequestUtil.json({
-      url: `${Config.rootUrl}${Config.acsAPIPrefix}/auth/login`,
+      url: `${Config.rootUrl}${Config.acsAPIPrefix}/auth/login${query}`,
       method: 'POST',
       data: credentials,
       success() {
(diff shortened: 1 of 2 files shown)


mlunoe:
Could you describe a concrete example where this would be used? Not sure I follow the description exactly.
Code looks good

wavesoft:
Hey @mlunoe , there is some discussion on the JIRA issue, you can have a look. The idea is to use the UI as an SSO login target for bouncer. If I got this correctly:

* The user requests to log-in via the CLI
* The CLI opens the browser (or provides a link) with the `target=` parameter present.
* The user performs log-in on the UI as usual
* The bouncer handles the `target` parameter & the authentication credentials & passes them to the CLI

From what I see the PR does what it's supposed to do, however like I said on the related plugins PR, we should run this through @jgehrcke to do a dry-run before we merge this.

You got my :label: !

nLight:
A little correction to the explanation from @wavesoft:

* IAM (Bouncer) detects target and generates an HTML page with the auth token
* as the final step, a user should copy that token and paste it into CLI

We need this mechanism to be able to use SSO from the CLI. 
`target` parameter tells IAM to emit the html page in the end
If there's no `target` provided the auth flow goes normally

jgehrcke:
:rocket: Thanks for that work, people! 

Some background from my side: a high level view onto the architecture this feature is part of can be found in the design doc(s) linked in the epic DCOS-9603.

This part here is just one piece of a puzzle. At the core, it is really about _enabling_ the UI to "forward" a certain piece of information from the fragment part of the URL to the IAM during the initiating login request (via a query parameter in the request URL). The piece of information I am referring to can be looked at as if it were an "authentication flow target" in a generic sense.

This method of forwarding the authentication flow target from the fragment part of the URL to the IAM enables us to do many things. CLI-triggered OpenID Connect SSO is just _one_ example! :guitar: 

I will help with manual testing as much as I can! Note that the core parts of the outlined architecture are already covered by extensive unit tests, and will shortly be covered by integration tests.

philipnrmn:
SGTM 👍

PR #1403, opened 2016-11-22
Add JSON thumbnails

leemunroe:
@ashenden @wavesoft Icon for choosing JSON mode. This may or may not be the final icon we want to use. If the later then good enough for a placeholder for now.


leemunroe:
For this screen 
image

PR #1405, opened 2016-11-23
Fix issue where routes might be undefined

mlunoe:
This PR fixes a couple of issues that emerges on an open cluster:
`Uncaught TypeError: Cannot read property 'reduce' of undefined(…)` in Sidebar.js
`Uncaught TypeError: Cannot read property 'length' of undefined(…)` in Breadcrumbs.js

### DIFF
--- a/src/js/components/Breadcrumbs.js
+++ b/src/js/components/Breadcrumbs.js
@@ -119,7 +119,8 @@ class Breadcrumbs extends React.Component {
 
 Breadcrumbs.defaultProps = {
   // Remove root '/' by default
-  shift: 0
+  shift: 0,
+  routes: []
 };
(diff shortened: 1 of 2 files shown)


nLight:
Very good catch Michael! ❤️  
But instead of empty array we should fall back to the initial set of paths. Please see my commit.

wavesoft:
Indeed, good catch Michael! However I aggree on @nLight , you are fixing an edge-case but you are not addressing the problem. Ideally, the `Hooks.applyFilter` should not modify the input data if the hook is missing.

mlunoe:
Yup, this looks good. I just needed a quick fix. Thanks for following up!

PR #1406, opened 2016-11-23
TaskDirectoryActions cleanup; move fetchNodeState to MesosStateActions

mlunoe:
This PR does not change functionality but moves fetchNodeState action to MesosStateActions from TaskDirectoryActions.


mlunoe:

plugins/services/src/js/events/TaskDirectoryActions.js
```diff
@@ -4,56 +4,23 @@ import ActionTypes from '../constants/ActionTypes';
...
-    if (node) {
-      pid = node.pid;
-    }
-
-    if (pid) {
-      nodePID = pid.substring(0, pid.indexOf('@'));
-    }
-
-    return `${Config.rootUrl}/agent/${task.slave_id}/${nodePID}/state`;
+  fetchNodeState(task, node, callback) {
```
orlandohohmeier: Can't we completely remove this from the task directory actions?

mlunoe: Actually, you are right. Will update when I have time

src/js/events/MesosStateActions.js
```diff
@@ -37,8 +51,31 @@ var MesosStateActions = {
...
 
+  fetchNodeState: RequestUtil.debounceOnError(
+    Config.getRefreshRate(),
+    function (resolve, reject) {
+      return function (task, node, success, error) {
+        return RequestUtil.json({
+          url: getNodeStateURL(task, node),
+          success(response) {
+            resolve();
+            success(response);
```
MatApple: This is deviating from Flux. Curious why we can't rely on the AppDispatcher to emit the results?

mlunoe: We could. I was just moving code around at this point. It didn't make sense to me that it was listed where it was.

MatApple: So this code was already in our repo?

rcorral:
I've addressed the comments from @MatApple and @orlandohohmeier. Also cleaned up the actions so that we fire different events.
@mlunoe please take a look!

rcorral:
@mlunoe if the changes look good to you, then let's merge!

PR #1409, opened 2016-11-23
DCOS-10991 [4 of 4]: Volumes form section

Poltergeist:
volumes mov

This adds the form component to add volumes.

### DIFF
--- a/plugins/services/src/js/reducers/JSONConfigReducers.js
+++ b/plugins/services/src/js/reducers/JSONConfigReducers.js
@@ -1,8 +1,9 @@
-import {JSONReducer as container} from './serviceForm/Container';
 import {JSONReducer as env} from './serviceForm/EnvironmentVariables';
 import {JSONReducer as labels} from './serviceForm/Labels';
+import {JSONReducer as volumes} from './serviceForm/Volumes';
 import {JSONReducer as healthChecks} from './serviceForm/HealthChecks';
 import {
+  combineReducers,
   simpleFloatReducer,
   simpleIntReducer,
   simpleReducer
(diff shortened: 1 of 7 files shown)


src/js/utils/__tests__/ReducerUtil-test.js
```diff
@@ -86,6 +86,62 @@ describe('ReducerUtil', function () {
         .toEqual([{id: 'foo', vip: 'foo:8080'}, {id: 'bar', vip: undefined}]);
     });
 
+    it('should use context in nested combineReducers', function () {
+      const reducers = ReducerUtil.combineReducers({
+        id: idReducer,
+        container: ReducerUtil.combineReducers({
+          vip(state = undefined, action) {
```
mlunoe: the default `state` assignment to `undefined` has no effect

plugins/services/src/js/reducers/JSONConfigReducers.js
```diff
@@ -11,7 +12,15 @@ import {
 module.exports = {
   id: simpleReducer('id'),
   instances: simpleIntReducer('instances'),
-  container,
+  container: combineReducers({
```
mlunoe: We probably need to re-consider this for when we introduce the multiple-container, but this is a good stepping stone

src/js/utils/__tests__/ReducerUtil-test.js
Poltergeist: Addressed this issue in all cases.

plugins/services/src/js/reducers/JSONConfigReducers.js
mlunoe: Ignore this, as it is handled by switching reducers!

mlunoe:
Ok, going to merge this in. I added a parser to make sure we are parsing integers for sizes.

A few follow-ups:
- We are missing the option to add a "Host" container, which in the old create form was referred to as "Docker Container Volumes". See image:

- External volumes can take size, but let's handle that in a separate issue (created a JIRA), which we can prioritise and add after the release.
- Since there is only one selection for external volumes, let's remove the select box, with only one option

PR #1408, opened 2016-11-23
DCOS-10987: [1/2] Separate JSON Flow

wavesoft:
This is the first PR for the separate JSON flow that cleans-up the `NewCreateServiceModalServicePicker`, keeping only the 3 acceptable options for now, and adding a placeholder for the JSON-only view.

@jfurrow I tried to center the buttons for making them look like the design, but I didn't want to mess-up the CSS. Can you have a look? https://mesosphere.invisionapp.com/d/main#/console/9042004/193660846/preview

PR #1410, opened 2016-11-23
DCOS-10987: [2/2] Separate JSON Flow

wavesoft:
---
⚠️  Depends on https://github.com/dcos/dcos-ui/pull/1408

---

This PR introduces the first draft of the json-only editor view. There is proper error validation and integration with the review screen.

new-json-only mov

The following things are missing:

- [x] Use the correct classes to make the JSON editor occupy the full size -- @jfurrow I might need some help here
- [x] Fix the floating error message (caused by the hack required on the side-by-side view)
- [x] Fix an annoying bug that displays an error the first time the JSON editor is shown

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "prettycron": "0.10.0",
     "query-string": "4.1.0",
     "react": "0.14.7",
-    "react-ace": "3.4.1",
+    "react-ace": "3.7.0",
     "react-addons-css-transition-group": "0.14.7",
     "react-addons-pure-render-mixin": "0.14.7",
     "react-addons-test-utils": "0.14.7",
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -6573,9 +6573,9 @@
       }
     },
     "react-ace": {
-      "version": "3.4.1",
-      "from": "react-ace@latest",
-      "resolved": "https://registry.npmjs.org/react-ace/-/react-ace-3.4.1.tgz"
+      "version": "3.7.0",
+      "from": "react-ace@3.7.0",
+      "resolved": "https://registry.npmjs.org/react-ace/-/react-ace-3.7.0.tgz"
     },
     "react-addons-css-transition-group": {
       "version": "0.14.7",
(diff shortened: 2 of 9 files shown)


jfurrow:
@wavesoft do you know if theres a reason we can't update `react-ace` to the most recent version? I tested this quickly and am receiving an error `Uncaught TypeError: Super expression must either be null or a function, not undefined` which seems to come from the `JobFormModal`. I did not investigate any further than this, sorry!

The new version allows us to pass custom styles into the editor itself, so the attribute `style={{height: '100%', width: '100%', position: 'absolute'}}` on `JSONEditor` in `CreateServiceJsonOnly.js` would do the trick!

If there is a problem with upgrading `react-ace`, then we will have to override some styles with the `!important`. So if you go this route, just add these styles to `components/modal/full-screen` and then pass `className="modal-full-screen-fill-body"` to `JSONEditor`:
```less
.modal-full-screen-fill-body {
    height: 100%;
    left: 0;
    position: absolute;
    top: 0;
    width: 100%;

    &.ace_editor {
      // Override inline styles on the Ace editor.
      height: 100% !important;
      position: absolute !important;
    }
  }
```

Feel free to ping me if you'd like any additional help with this!

wavesoft:
Hey @jfurrow . Thanks for the tips. This version also supports providing `width` and `height`, check the `JSONEditor` component. The problem with the latest version is that it also needs the latest `React`. Apparently they have changed something on their API.

I will try your suggestions and check how it goes.

wavesoft:
Btw, I was referring to the `4.x` branch. I actually updated to the latest `3.x` branch (that still uses the old React API) and it works fine.

package.json
```diff
@@ -43,7 +43,7 @@
     "prettycron": "0.10.0",
     "query-string": "4.1.0",
     "react": "0.14.7",
-    "react-ace": "3.4.1",
+    "react-ace": "3.7.0",
```
mlunoe: Have you checked the implications of this update with the old flow?
Does that still work?

wavesoft: Yes, there is a big change on the `4.x` branch. `3.x` is still compatible like before. In addition we now have a single place the ACE editor is used (the `JSONEditor` component) so if something goes wrong, that's the only place we have to touch.

orlandohohmeier:
validation-issue

For some reason the validation isn't working when I paste a app/pod definition.

wavesoft:
Everything should be addressed 👍

PR #1411, opened 2016-11-23
Fix health-check advanced section label

philipnrmn:
Before:

After:


Poltergeist:
👈  His Fault 👎  😢  Thanks for fixing.

PR #1412, opened 2016-11-23
[Sidebar: 1 of 4] Implement user account menu

jfurrow:
This PR removes the old `UserDropup` component and implements a new `UserAccountDropdown` component.


plugins/oauth/hooks.js
```diff
@@ -47,7 +45,7 @@ module.exports = Object.assign({}, StoreMixin, {
     'applicationRoutes',
     'delayApplicationLoad',
     'organizationRoutes',
-    'sidebarFooter',
+    'userDropdownItems',
```
orlandohohmeier: Not part of this PR but we should use the `Mount` component instead of these filters...

orlandohohmeier:
LGTM! 🚢

PR #1420, opened 2016-11-25
Update CNVS and remove overrides

jfurrow:
This PR updates CNVS to the latest release which includes a number of bug fixes and style changes.

I'm also removing all but a few CNVS variable overrides that @ashenden originally introduced in DC/OS UI. @ashenden feel free to remove any remaining overrides that you don't think we need.


src/styles/components/typeahead/styles.less
```diff
@@ -114,19 +114,3 @@
...
-
-      .typeahead {
-
-        .dropdown-menu {
-          margin-left: calc(-1 * (@icon-mini-width + @form-control-border-width) ~' - ' (-2 * @form-control-padding-left-screen-large));
-        }
-      }
-    }
-  }
-}
```
jfurrow: Because the new variable doesn't have corresponding screen sizes (it's a `rem` unit), we don't need to change our declaration at the `large` breakpoint.

orlandohohmeier:
LGTM 💯 ✨

PR #1432, opened 2016-11-29
Adjust config display component registration

orlandohohmeier:
Change the default config display component registration to better explain the component priorities and ensure that the component registration happens only once.

/cc @wavesoft @nLight @Poltergeist

### DIFF
--- a/plugins/services/src/js/components/ServiceConfigDisplay.js
+++ b/plugins/services/src/js/components/ServiceConfigDisplay.js
@@ -9,29 +9,30 @@ import ServiceConfigEnvironmentVariablesSectionDisplay from './ServiceConfigEnvi
 import ServiceConfigLabelsSectionDisplay from './ServiceConfigLabelsSectionDisplay';
 import ServiceConfigHealthChecksSectionDisplay from './ServiceConfigHealthChecksSectionDisplay';
 
-// To pad priorities, so we can add components in between
+const MOUNT_TYPE = 'CreateService:ServiceConfigDisplay';
 const PRIORITIES_PAD_NUMBER = 100;
-
-const SERVICE_CONFIG_DISPLAY_LIST = [
+const DEFAULT_DISPLAY_COMPONENTS = [
   ServiceConfigGeneralSectionDisplay,
   ServiceConfigNetworkingSectionDisplay,
   ServiceConfigStorageSectionDisplay,
   ServiceConfigEnvironmentVariablesSectionDisplay,
   ServiceConfigLabelsSectionDisplay,
   ServiceConfigHealthChecksSectionDisplay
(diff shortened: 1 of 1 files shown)


plugins/services/src/js/components/ServiceConfigDisplay.js
```diff
@@ -9,29 +9,32 @@ import ServiceConfigEnvironmentVariablesSectionDisplay from './ServiceConfigEnvi
...
   ServiceConfigEnvironmentVariablesSectionDisplay,
   ServiceConfigLabelsSectionDisplay,
   ServiceConfigHealthChecksSectionDisplay
-].reverse(); // MountService sorts by priority but DESC ¯\_(ツ)_/¯
-
-class ServiceConfigDisplay extends React.Component {
-  componentWillMount() {
-    SERVICE_CONFIG_DISPLAY_LIST.forEach((component, index) => {
+]
+    .forEach((component, index, componentArray) => {
```
wavesoft: #nit The indentation here looks ind of funny. I would say either indent left the entire `forEach` block, or wrap it around some block.

plugins/services/src/js/components/ServiceConfigDisplay.js
```diff
@@ -9,29 +9,32 @@ import ServiceConfigEnvironmentVariablesSectionDisplay from './ServiceConfigEnvi
...
+      // Define  component priorities based on index (highest prio goes first)
+      // and pad priorities, so that we can add components in between.
+      let priority = (componentArray.length - index) * PRIORITIES_PAD_NUMBER;
```
nLight: I think this will break the order I'm relying on in https://github.com/mesosphere/dcos-ui-plugins-private/pull/462/files#diff-f1e16cc6041efa19213bfb8863789802R71 I'm ok changing it in the plugin.

plugins/services/src/js/components/ServiceConfigDisplay.js
nLight: Let's stick with a redundant constant

plugins/services/src/js/components/ServiceConfigDisplay.js
nLight: I think this will break the order I'm relying on in https://github.com/mesosphere/dcos-ui-plugins-private/pull/462/files#diff-f1e16cc6041efa19213bfb8863789802R71 I'm ok changing it in the plugin.

orlandohohmeier: Do you want me to adjust this here?

nLight: I'll create a pr fix for the plugin

nLight:
✨

PR #1437, opened 2016-11-29
dcos-11760: Placement constraints

nLight:
This PR introduces placement constraints:

⚠️  I moved "Container Runtime" section from the foldable Advanced Settings as it is supposed to be according to the design. So it's always visible, while Placement Constraints go under Advanced Settings section

screen shot 2016-11-29 at 19 10 17

screen shot 2016-11-29 at 19 12 21

### DIFF
--- a/plugins/services/src/js/components/forms/GeneralServiceFormSection.js
+++ b/plugins/services/src/js/components/forms/GeneralServiceFormSection.js
@@ -31,7 +31,66 @@ const containerRuntimes = {
 };
 
 class GeneralServiceFormSection extends Component {
+  getPlacementConstraints(data = []) {
+    const errors = this.props.errors || [];
 
+    return data.map((constraint, index) => {
+      let fieldLabel = null;
+      let operatorLabel = null;
+      let parameterLabel = null;
+      if (index === 0) {
+        fieldLabel = <FieldLabel>Field</FieldLabel>;
+        operatorLabel = <FieldLabel>Operator</FieldLabel>;
+        parameterLabel = <FieldLabel>Parameter</FieldLabel>;
+      }
+
+      return (
+        <div key={index} className="flex row">
+          <FormGroup
+            className="column-3"
+            required={false}
+            showError={Boolean(errors[index])}>
+            {fieldLabel}
+            <FieldInput
+              name={`constraints.${index}.field`}
+              type="text"
+              value={constraint.field}/>
+            <FieldError>{errors[index]}</FieldError>
+          </FormGroup>
+          <FormGroup
+            className="column-3"
+            required={false}
+            showError={Boolean(errors[index])}>
+            {operatorLabel}
+            <FieldInput
+              name={`constraints.${index}.operator`}
+              type="text"
+              value={constraint.operator}/>
+            <FieldError>{errors[index]}</FieldError>
+          </FormGroup>
+          <FormGroup
+            className="column-3"
+            required={false}
+            showError={Boolean(errors[index])}>
+            {parameterLabel}
+            <FieldInput
+              name={`constraints.${index}.value`}
+              type="text"
+              value={constraint.value}/>
+            <FieldError>{errors[index]}</FieldError>
+          </FormGroup>
+
+          <FormGroup className="flex flex-item-align-end column-2">
+            <a
+              className="button button-primary-link button-flush"
+              onClick={this.props.onRemoveItem.bind(this, {value: index, path: 'constraints'})}>
+              Delete
+            </a>
+          </FormGroup>
+        </div>
+      );
+    });
+  }
   getRuntimeSelections(data = {}) {
     let {container = {}, cmd, gpus} = data;
     let isDisabled = {};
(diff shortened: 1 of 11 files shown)


plugins/services/src/js/components/forms/GeneralServiceFormSection.js
```diff
@@ -30,7 +30,66 @@ const containerRuntimes = {
...
+        fieldLabel = <FieldLabel>Field</FieldLabel>;
+        operatorLabel = <FieldLabel>Operator</FieldLabel>;
+        parameterLabel = <FieldLabel>Parameter</FieldLabel>;
+      }
+
+      return (
+        <div key={index} className="flex row">
+          <FormGroup
+            className="column-3"
+            required={false}
```
Poltergeist: Just omit `required` here. We only need to provide `require` if it is required.

plugins/services/src/js/reducers/serviceForm/Constraints.js
```diff
@@ -0,0 +1,125 @@
...
+          case REMOVE_ITEM:
+            this.constraints = this.constraints.filter((item, index) => {
+              return index !== value;
+            });
+            break;
+        }
+
+        return this.constraints.filter((item) => {
+          return item.field != null && item.operator != null;
+        }).map(({field, operator, value}) => {
```
Poltergeist: As we use this multiple times it would be better to have this as a constant function. And could we also filter before the map in the `JSONReducer` to only output something if field and operator are populated.

plugins/services/src/js/reducers/serviceForm/Constraints.js
```diff
@@ -0,0 +1,125 @@
...
+        return state;
+      }
+
+      let index = joinedPath.match(/\d+/)[0];
+      let key = ['field', 'operator', 'value'].find(function (name) {
+        return `constraints.${index}.${name}` === joinedPath;
+      });
+      if (key !== -1) {
+        state[index][key] = value;
+      }
```
Poltergeist: 😲

plugins/services/src/js/reducers/serviceForm/Constraints.js
```diff
@@ -0,0 +1,125 @@
...
+  },
+
+  FormReducer(state = [], {type, path, value}) {
+    if (path == null) {
+      return state;
+    }
+
+    let joinedPath = path.join('.');
+
+    if (joinedPath.search('constraints') !== -1) {
```
mlunoe: We could just use `path.includes('constraints')` here

plugins/services/src/js/reducers/serviceForm/Constraints.js
```diff
@@ -0,0 +1,125 @@
...
+      });
+      if (key !== -1) {
+        this.constraints[index][key] = value;
+      }
+    }
+
+    return this.constraints.filter((item) => {
+      return item.field != null && item.operator != null;
+    }).map(({field, operator, value}) => {
+      return value ? [field, operator, value] : [field, operator];
```
mlunoe: We didn't decide to use the ternary operator yet. Let's hold off until the team has decided this.

plugins/services/src/js/reducers/serviceForm/Constraints.js
```diff
@@ -0,0 +1,125 @@
...
+        }
+
+        return state;
+      }
+
+      if (type !== SET) {
+        return state;
+      }
+
+      let index = joinedPath.match(/\d+/)[0];
```
mlunoe: You always know that `index` is going to be the second item in the array, so you can do this: `const index = path[1];`

plugins/services/src/js/reducers/serviceForm/Constraints.js
```diff
@@ -0,0 +1,125 @@
...
+      let key = ['field', 'operator', 'value'].find(function (name) {
```
mlunoe: I think this makes it a little bit more concise what is going on:
```js
let name = path[2];
if (['field', 'operator', 'value'].includes(name)) {
  state[index][key] = value;
}

File: plugins/services/src/js/reducers/serviceForm/Constraints.js [hunk 8]
```diff
@@ -0,0 +1,125 @@
...
+        }).map(({field, operator, value}) => {
+          return value ? [field, operator, value] : [field, operator];
+        });
+      }
+
+      if (type !== SET) {
+        return state;
+      }
+
+      let index = joinedPath.match(/\d+/)[0];
```
mlunoe: Same thing here:
```js
let index = path[1];
let name = path[2];
if (path[1] === 'constraints' && ['field', 'operator', 'value'].includes(name)) {
  state[index][key] = value;
}

File: plugins/services/src/js/reducers/serviceForm/Constraints.js (hunk 3 above)
Comment (nLight) [inline plugins/services/src/js/reducers/serviceForm/Constraints.js]: :trollface:

File: plugins/services/src/js/reducers/serviceForm/Constraints.js (hunk 5 above)
Comment (nLight) [inline plugins/services/src/js/reducers/serviceForm/Constraints.js]: :sadtroll:

**Comment by nLight** [comment]
@Poltergeist @mlunoe I heard the message and addressed the comments! 🙇

**Comment by Poltergeist** [comment]
Nice Work, Let's merge this.

File: plugins/services/src/js/reducers/serviceForm/Constraints.js [hunk 9]
```diff
@@ -0,0 +1,99 @@
...
+    if (!isEmpty(value)) {
+      return [field, normalisedOperator, value];
+    }
+
+    return [field, normalisedOperator];
+  });
+}
+
+function processTransaction(state, {type, path, value}) {
+  const [field, index, name] = path;
```
mlunoe: ❤️

PR #1440, opened 2016-11-30
Reduce tooltip font size

jfurrow:
These tooltips were obnoxiously huge IMO. @ashenden @leemunroe What do you think?

Before:

After:

### DIFF
--- a/src/styles/components/tooltip/styles.less
+++ b/src/styles/components/tooltip/styles.less
@@ -7,6 +7,7 @@
       background: @tooltip-background-color;
       border-radius: 2px;
       color: @white;
+      font-size: 0.9rem;
       padding: 0.8rem 1rem;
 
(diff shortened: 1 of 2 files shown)


ashenden:
About it! Looks good.  Might be worth playing with the line-height a bit, or at least at the var so we can fuck with it later.

leemunroe:
+1 Yes please. Been a pet peeve of mine too.

ashenden:
I'd also like to see `.tooltip` as a component included CNVS.  Just created an issue for this -- https://github.com/mesosphere/cnvs/issues/65

jfurrow:
Adding `hold merge` until I can come back around and add the font-size and line-height variables.

jfurrow:
Added the variables, ready for 🚢

PR #1441, opened 2016-11-30
dcos-11762: Container Artifacts section

nLight:
This PR introduces container artifacts section. 

Designs are a bit confusing, because the review screen shows all possible settings of an artifact while the form and other issue description tells we should implement only URIs. So I did only URI but all reducers are aware that there could be other fields in near future.

⚠️  I added my first CSS \o/ please verify if I did this in a right place.

## Form
screen shot 2016-11-30 at 15 54 24

## Review screen
screen shot 2016-11-30 at 15 53 58

### DIFF
--- a/plugins/services/src/js/components/ServiceConfigBaseSectionDisplay.js
+++ b/plugins/services/src/js/components/ServiceConfigBaseSectionDisplay.js
@@ -10,19 +10,8 @@ import Util from '../../../../../src/js/utils/Util';
 
 class ServiceConfigBaseSectionDisplay extends React.Component {
 
-  shouldExcludeItem(row) {
-    const {appConfig} = this.props;
-
-    switch (row.key) {
-      case 'gpus':
-        return !Util.findNestedPropertyInObject(appConfig, 'gpus');
-      case 'container.volumes':
-        return !Util.findNestedPropertyInObject(
-          appConfig, 'container.volumes.length'
-        );
-      default:
-        return false;
-    }
+  shouldExcludeItem() {
+    return false;
   }
 
   getDisplayValue(type, value) {
--- a/plugins/services/src/js/components/forms/ContainerServiceFormSection.js
+++ b/plugins/services/src/js/components/forms/ContainerServiceFormSection.js
@@ -170,7 +170,6 @@ class ContainerServiceFormSection extends Component {
           </FormGroup>
         </div>
         {this.getArtifactsInputs(data.fetch)}
-        <FieldError>{errors.fetch}</FieldError>
         <div>
           <a
             className="button button-primary-link button-flush"
--- a/plugins/services/src/js/components/forms/ContainerServiceFormSection.js
+++ b/plugins/services/src/js/components/forms/ContainerServiceFormSection.js
@@ -31,7 +31,7 @@ const getArtifactsLabel = () => {
     <FieldLabel>
       {'Artifact URI '}
       <Tooltip
-        content="Provided URIs are passed to Mesos fetcher module and resolved in runtime."
+        content="Provided URIs are passed to Mesos fetcher module and resolved at runtime."
         interactive={true}
         maxWidth={300}
         scrollContainer=".gm-scroll-view"
--- a/plugins/services/src/js/components/forms/ContainerServiceFormSection.js
+++ b/plugins/services/src/js/components/forms/ContainerServiceFormSection.js
@@ -96,6 +96,7 @@ class ContainerServiceFormSection extends Component {
     if (data.container.type === 'DOCKER') {
       return [
         <Tooltip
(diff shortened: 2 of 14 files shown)


nLight:
retest this please 🙏

plugins/services/src/js/components/forms/ContainerServiceFormSection.js
```diff
@@ -26,7 +26,72 @@ const containerSettings = {
   }
 };
 
+const getArtifactsLabel = () => {
+  return (
+    <FieldLabel>
+      {'Artifact URI '}
+      <Tooltip
+        content="Provided URIs are passed to Mesos fetcher module and resolved in runtime."
```
philipnrmn: 'at runtime?'

plugins/services/src/js/components/ServiceConfigBaseSectionDisplay.js
```diff
@@ -10,19 +10,8 @@ import Util from '../../../../../src/js/utils/Util';
 
 class ServiceConfigBaseSectionDisplay extends React.Component {
```
philipnrmn: Wonder whether it's worth marking this class `@abstract`

plugins/services/src/js/reducers/serviceForm/Artifacts.js
```diff
@@ -0,0 +1,75 @@
+import {
+  ADD_ITEM,
+  REMOVE_ITEM,
+  SET
+} from '../../../../../../src/js/constants/TransactionTypes';
```
philipnrmn: 👍 really enjoying our use of symbols

src/styles/components/artifacts-section/styles.less
```diff
@@ -0,0 +1,3 @@
+.artifacts-section {
+  margin-top: 18px;
```
philipnrmn: can we make use of the spacing unit variables here instead of `px` please?

plugins/services/src/js/components/forms/ContainerServiceFormSection.js
```diff
@@ -102,6 +167,15 @@ class ContainerServiceFormSection extends Component {
             <FieldError>{errors.disk}</FieldError>
           </FormGroup>
         </div>
+        {this.getArtifactsInputs(data.fetch)}
+        <FieldError>{errors.fetch}</FieldError>
```
MatApple: Do we need this error if we are already showing them inside the jsx from `this.getArtifactsInput`?

src/styles/components/artifacts-section/styles.less
jfurrow: As part of a separate PR, I think we need to go back and consistently apply margins to different sections of the form. I created a JIRA for it here: https://mesosphere.atlassian.net/browse/DCOS-11803

Would you mind including the ticket number and a `TODO` in this file?

orlandohohmeier:
We need to make sure that respective parser/reducer pair isn't overwriting unknown (to the form) fields. Please the respective schema to for the complete list of possible values:

https://github.com/mesosphere/marathon/blob/6db4d9980af15b45bc3c8647335f39deef3f49bf/docs/docs/rest-api/public/api/v2/types/artifact.raml

src/styles/components/artifacts-section/styles.less
jfurrow: I made some changes...
* I used the spacing variable and its screen sizes here
* I moved the styles to `src/styles/views/create-service-modal/`
* I added the TODO for creating a `FormSection` component

plugins/services/src/js/reducers/serviceForm/Artifacts.js
```diff
@@ -0,0 +1,75 @@
...
+
+  if (type === ADD_ITEM) {
+    newState.push({uri: null});
+  }
+  if (type === REMOVE_ITEM) {
+    newState = newState.filter((item, index) => {
+      return index !== value;
+    });
+  }
+  if (type === SET) {
```
mlunoe: I believe we need to check name here as well:
```js
  const [field, index, name] = path;

  if (type === SET && name === 'uri') {
    newState[index].uri = value;
  }
```

rcorral:
Took care of @orlandohohmeier's and @mlunoe's comments :)

plugins/services/src/js/components/forms/ContainerServiceFormSection.js
rcorral: done!

plugins/services/src/js/components/forms/ContainerServiceFormSection.js
philipnrmn: Sure enough, they appear twice. Thanks for the catch.

PR #1450, opened 2016-11-30
Use JSON only editor when editing a pod

rcorral:
refs: DCOS-11793

gif

### DIFF
--- a/plugins/services/src/js/components/modals/NewCreateServiceModal.js
+++ b/plugins/services/src/js/components/modals/NewCreateServiceModal.js
@@ -69,9 +69,21 @@ class NewServiceFormModal extends Component {
    */
   componentWillReceiveProps(nextProps) {
     if (!ServiceUtil.isEqual(this.props.service, nextProps.service)) {
-      this.setState({
-        serviceConfig: nextProps.service.getSpec()
-      });
+      let newState = {serviceConfig: nextProps.service.getSpec()};
+
+      if (nextProps.isEdit) {
+        newState.servicePickerActive = false;
+
+        if (nextProps.service instanceof Pod) {
+          newState.serviceJsonActive = true;
+          newState.serviceFormActive = false;
+        } else {
+          newState.serviceJsonActive = false;
+          newState.serviceFormActive = true;
+        }
+      }
+
+      this.setState(newState);
     }
   }
 
@@ -83,18 +95,19 @@ class NewServiceFormModal extends Component {
       serviceReviewActive
     } = this.state;
 
-    // Close if picker is open, or if editing a service in the form
-    if (servicePickerActive || (serviceFormActive && this.props.isEdit)) {
-      this.handleClose();
-      return;
-    }
-
     if (serviceReviewActive) {
       // Just hide review screen. Form or JSON mode will be
       // activated automaticaly depending on their last state
       this.setState({
         serviceReviewActive: false
       });
+      return;
+    }
+
+    // Close if picker is open, or if editing a service in the form
+    if (servicePickerActive ||
+      ((serviceFormActive || serviceJsonActive) && this.props.isEdit)) {
+      this.handleClose();
       return;
     }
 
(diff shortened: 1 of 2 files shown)


MatApple:
retest this please

nLight:
After rebasing this is what I see, only couple of rows. But the logic works perfectly

screen shot 2016-12-01 at 10 17 35

nLight:
I think this visual bug is unrelated, so you have my label

nLight:
I merged the fix and JSON editor is fullscreen again \o/

plugins/services/src/js/components/modals/NewCreateServiceModal.js
```diff
@@ -69,9 +69,21 @@ class NewServiceFormModal extends Component {
...
-      });
+      let newState = {serviceConfig: nextProps.service.getSpec()};
+
+      if (nextProps.isEdit) {
+        newState.servicePickerActive = false;
+
+        if (nextProps.service instanceof Pod) {
+          newState.serviceJsonActive = true;
+          newState.serviceFormActive = false;
+        } else {
```
orlandohohmeier: Let's remove the `else` case by moving the respective lines to the top.

plugins/services/src/js/components/modals/NewCreateServiceModal.js
```diff
@@ -98,6 +104,13 @@ class NewServiceFormModal extends Component {
       return;
     }
 
+    // Close if picker is open, or if editing a service in the form
+    if (servicePickerActive ||
```
orlandohohmeier: Any chance that we can simplify this condition?

plugins/services/src/js/components/modals/NewCreateServiceModal.js
```diff
@@ -371,16 +384,28 @@ class NewServiceFormModal extends Component {
     // Switch directly to form if edit
     if (nextProps.isEdit) {
       newState.servicePickerActive = false;
-      newState.serviceFormActive = true;
+
+      if (nextProps.service instanceof Pod) {
+        newState.serviceJsonActive = true;
+      } else {
+        newState.serviceFormActive = true;
```
orlandohohmeier: Let's remove the `else` case here as well

plugins/services/src/js/components/modals/NewCreateServiceModal.js
```diff
@@ -371,16 +384,28 @@ class NewServiceFormModal extends Component {
...
-    let {servicePickerActive, serviceFormActive} = this.state;
+    let {
+      serviceFormActive,
+      servicePickerActive,
+      serviceReviewActive,
+      serviceJsonActive
+    } = this.state;
     let label = 'Back';
-    if (servicePickerActive || (serviceFormActive && this.props.isEdit)) {
+    if (servicePickerActive ||
```
orlandohohmeier: Please simplify this condition or document the different cases

plugins/services/src/js/components/modals/NewCreateServiceModal.js
```diff
@@ -98,6 +104,13 @@ class NewServiceFormModal extends Component {
       return;
     }
 
+    // Close if picker is open, or if editing a service in the form
+    if (servicePickerActive ||
+      ((serviceFormActive || serviceJsonActive) && this.props.isEdit)) {
+      this.handleClose();
+      return;
```
MatApple: New line

plugins/services/src/js/components/modals/NewCreateServiceModal.js
MatApple: Let's also add a newline above here - this block of `let`s and `if` is hard to parse.

plugins/services/src/js/components/modals/NewCreateServiceModal.js
orlandohohmeier: Thought we don't do a newline in this particular case...

orlandohohmeier:
@MatApple addressed all comments...

PR #1453, opened 2016-12-01
Remove restart option from Pods as this isn’t supported by the API.

rcorral:


plugins/services/src/js/containers/services/ServicesTable.js
```diff
@@ -193,6 +188,16 @@ class ServicesTable extends React.Component {
       }
     ];
 
+    if (!isPod) {
+      dropdownItems.splice(1, 0, {
```
orlandohohmeier: Let's push the items separately so that we don't need to use splice to add the restart item.

MatApple: actually it's much simpler - just have to add the hidden class like we do if `isGroup || instancesCount === 0`

MatApple:
@orlandohohmeier - addressed the comment

orlandohohmeier:
@MatApple Let's please properly remove (or even better not add) the button instead of hiding it with CSS.

rcorral:
@orlandohohmeier are you cool with us merging this in for now and then having a discussion about the best approach to this. There's probably arguments for both approaches and I'm not sure what's best right now.

orlandohohmeier:
@rcorral I'm ok with merging it, but let's please create an issue so that we don't forget to revisit this decision.

rcorral:
Added to our sync.

PR #1454, opened 2016-12-01
Fix force action and error message handling for modals

MatApple:
Fixing force action and error handling on modals.


rcorral:
Thanks for this fix @MatApple - I pushed a commit too.

MatApple:
Thanks for the extra cleanup @rcorral.

rcorral:
@MatApple for some reason the force delete doesn't work.
Could you check it out?
gif

MatApple:
Fixed @rcorral - good find!

PR #1456, opened 2016-12-01
DCOS-11813: Fix edit pod interaction

orlandohohmeier:
---

⚠️  #1450 needs to land first

---

This PR changes the service `isEqual` util, to only compare the service specs as everything else is status information and bound to frequent changes. Without this change we end up with unwanted updates of the service form which sometimes resulted in data loss.

Closes DCOS-11813


orlandohohmeier:

PR #1458, opened 2016-12-01
DCOS-10470: [1/4] Introducing the . dropdown-panel-group class

wavesoft:
This PR introduces the `.dropdown-panel-group` class that is needed in order to implement the drop-down styling options for the DSL.

image

### DIFF
--- a/src/styles/components/dropdown-panels/styles.less
+++ b/src/styles/components/dropdown-panels/styles.less
@@ -20,15 +20,15 @@
       }
 
       &.dropdown-panel-animated {
-        margin-top: -10px;
         opacity: 0;
         pointer-events: none;
-        transition: margin-top @modal-animation-duration @modal-animation-easing, opacity @modal-animation-duration @modal-animation-easing;
+        transform: @dropdown-panels-animation-transform;
+        transition: transform @modal-animation-duration @modal-animation-easing, opacity @modal-animation-duration @modal-animation-easing;
 
         &.is-visible {
-          margin-top: 0;
           opacity: 1;
           pointer-events: auto;
+          transform: @dropdown-panels-rest-transform;
         }
       }
     }
--- a/src/styles/components/dropdown-panels/styles.less
+++ b/src/styles/components/dropdown-panels/styles.less
@@ -1,0 +1,36 @@
+& when (@dropdown-panels-enabled) {
+
+  .dropdown-panel-group {
+    position: relative;
+
+    > .dropdown-panel {
+      box-shadow: @modal-shadow;
+      max-width: 100%;
+      position: absolute;
+      right: 0;
+      z-index: @z-index-overlay;
+
+      &.dropdown-panel-full-width {
+        width: 100%;
+      }
+
+      &.dropdown-panel-left {
+        left: 0;
+        right: auto;
+      }
+
+      &.dropdown-panel-animated {
+        margin-top: -10px;
+        opacity: 0;
+        pointer-events: none;
+        transition: margin-top @modal-animation-duration @modal-animation-easing, opacity @modal-animation-duration @modal-animation-easing;
+
+        &.is-visible {
+          margin-top: 0;
+          opacity: 1;
+          pointer-events: auto;
+        }
+      }
+    }
+  }
+}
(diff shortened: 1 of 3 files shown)


nLight:
I'd reduce paddings a bit, still ✨!

wavesoft:
The paddings are from the panel-styling, this PR only introduces the fact that a panel can now float under an element without much of fancy hacking.

orlandohohmeier:
Looks great. ✨ @jfurrow what do you think?

jfurrow:
I'm not sure creating a new element is the right approach. Although it's hacky, I think it would be best to use the actual `Dropdown` component. This gives us a few advantages out of the box:
* it will render on top of neighboring elements, always
* it will render outside of `Modal` boundaries, should that situation arise
* it will have the exact same styling and animation as the current `Dropdown`

For this to work, we would pass the `Dropdown` only two items: one would be the "trigger" element (the caret icon), and the other would be the component that contains all of the filtering options. Then we could set the `persistentID` prop on the `Dropdown` to display the caret icon in all cases.

In the future it would be best to create a more abstract component that gives us this "panel popup" beahavior and styling so that we can use it in multiple places, but I don't think we'll have time for that in this release.

@orlandohohmeier @wavesoft what do you think of the `Dropdown` approach? I'll leave some comments on the styling just in case you opt to move forward with this PR.

src/styles/components/dropdown-panels/styles.less
```diff
@@ -0,0 +1,36 @@
...
+      &.dropdown-panel-left {
+        left: 0;
+        right: auto;
+      }
+
+      &.dropdown-panel-animated {
+        margin-top: -10px;
+        opacity: 0;
+        pointer-events: none;
+        transition: margin-top @modal-animation-duration @modal-animation-easing, opacity @modal-animation-duration @modal-animation-easing;
```
jfurrow: Let's animate `transform` instead of `margin-top` for best performance. https://developers.google.com/web/fundamentals/design-and-ui/animations/animations-and-performance

wavesoft:
Hey @jfurrow , my two pieces of mind:

1. the truth is that I had a look on the `Dropdown` component but it felt a bit of an overkill. (Plus I saw that it's using absolute positioning + javascript to align the pop-up)
2. I wanted to re-use this panel component for also providing typeahead suggestions when the user is typing (hence the `full-width` class).
3. I saw that we have some amazing less variables, and therefore could address most of your bullet points.

I am also not opposing against `Dropdown`, but I felt it would cause more problems than solving them at this particularly short time.

src/styles/components/dropdown-panels/styles.less
wavesoft: Ah, true. I am also always using `transform` for this reason, but I thought it would screw-up the alignment and cause the same problems that I had with the JSON editor + modals.

But actually in this particular case we don't have that issue, so I could definitely use `transform` instead.

jfurrow: The complications with `transform` only occur when a direct descendent uses `position: fixed` as well. I can't see a reason for a child of this component to use `position: fixed`, and in the rare case that a child did need `position: fixed`, we would just use the `Portal` component to render at the root of `body`

jfurrow:
@wavesoft That's three pieces of mind! 😄 

Your points are valid, let's move forward with this for now. If we do need some of the functionality we've needed from the `Dropdown`, we can address that when we get there!

Nice work!

src/styles/components/dropdown-panels/styles.less
```diff
@@ -0,0 +1,36 @@
...
+        width: 100%;
+      }
+
+      &.dropdown-panel-left {
+        left: 0;
+        right: auto;
+      }
+
+      &.dropdown-panel-animated {
+        margin-top: -10px;
```
jfurrow: The `Dropdown` currently uses a hard-coded value of `30px`, maybe we should use this here? And I do think it would be nice to make use of `transform` here, as addressed here: https://github.com/dcos/dcos-ui/pull/1458/files#r90519960

wavesoft:
So, I am now using `transform` and I exposed two more variables in order for us to configure this easily in the future.

PR #1459, opened 2016-12-01
DCOS-11725: Display Marathon errors in create service form

nLight:
This PR introduces a Modal that displays marathon errors on the new create service form.

dec-01-2016 17-07-16

Please tell me if I should be doing this type of error display instead 😅 
https://mesosphere.invisionapp.com/d/main#/console/9042004/196037823/preview


orlandohohmeier:
@nLight unfortunately yes... we should render the errors in the review screen.

nLight:
@orlandohohmeier implementing the design would be a non-trivial thing, esp since we've decided to ditch definition style declaration 😓

orlandohohmeier:
@nLight how about using the error component from the form for now?
 


I think the besides that we also want to parse the error paths so that we can feed the error to the form as well.

nLight:
Replaced modal with Alert

screen shot 2016-12-01 at 18 19 29

leemunroe:
The ideal state, in priority order from my pov :-)

1. Highlighting the issues on the [review screen](https://mesosphere.invisionapp.com/share/969KCBASB) so the user knows they can't proceed until they fix these things
2. Highlighting the [field contextually](https://mesosphere.invisionapp.com/share/J2913TAVT#/196005505_Error-1) where the issue is so the user can easily find and fix the issue
3. Highlighting the tab with a badge and count so the user knows where the issues are and can find and fix them

I also prefer the alert you suggested vs a modal, but only on the review screen if it's something we're not able to highlight in the config details like invalid JSON.

Example
image

* The services tab would have a red badge with "1" in it
* The alert that is currently there would not be there
* The field would be red with a line of text below the input (above the help text) "A Service ID is required"

Related: `Service Name` should be `Service ID`

nLight:
Aggregated errors on the review screen. The best we can do now.

screen shot 2016-12-02 at 15 32 03

nLight:
@leemunroe sadly we have to reduce the scope, I'll create tickets for all those steps that you've mentioned

nLight:
https://mesosphere.atlassian.net/browse/DCOS-11849
https://mesosphere.atlassian.net/browse/DCOS-11850
https://mesosphere.atlassian.net/browse/DCOS-11851

orlandohohmeier:
Please reset the errors if the user update the service spec.

nLight:
@orlandohohmeier unfortunately, we can't keep the track of whether we've shown the review screen or not inside the component itself because we remount it every time we show it. That means we should keep that info in the Modal which makes it a bit more complex.

rcorral:
@nLight could we simply clear the errors whenever we leave the review page? So there should never be errors when coming from the form/json to the review page.

nLight:
@rcorral 👍  reused a function that was there already.
Added: Clearing errors on leaving the review screen
@orlandohohmeier & @wavesoft 🙏

nLight:
dec-05-2016 10-03-32

orlandohohmeier:
LGTM! 🚢

wavesoft:
LGTM too ✨ 
(Keep in mind that might be a good idea to create a marathon error normalisation utility function)

mlunoe:
@wavesoft please delete branches when you merge PRs, so we don't clutter our repository

PR #1466, opened 2016-12-02
DCOS-11823: Fix PodInstancesTable layout

jfurrow:
This PR fixes the `PodInstancesTable` layout. I've applied widths and responsiveness in a way that we haven't done before, so I'm interested to see what you guys think.

Instead of applying `hidden-*` classes and using inline `style` props, it uses classnames and external CSS. I think it's easier to reason about and it would allow us to change the width of certain columns in media queries. There would be a little extra work when creating new tables, but I think it's worth it. I'd be happy to apply this to the rest of the tables in the UI, if you all like it! :)

---

In an unrelated note, for some reason when I apply the `pod-instances-table-column-primary` class to the related `col` in the `colgroup`, the sorting of the table gets reversed in one of the tests. WTF??? Am I missing something? We don't actually need the `colgroup` if we use this approach, but I left it here for now.

### DIFF
--- a/plugins/services/src/js/containers/pod-instances/PodInstancesTable.js
+++ b/plugins/services/src/js/containers/pod-instances/PodInstancesTable.js
@@ -16,6 +16,18 @@ import PodUtil from '../../utils/PodUtil';
 import TimeAgo from '../../../../../../src/js/components/TimeAgo';
 import Units from '../../../../../../src/js/utils/Units';
 
+const tableColumnClasses = {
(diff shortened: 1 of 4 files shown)


jfurrow:
retest this please

orlandohohmeier:
✨  This looks great. I very much like your CSS approach and think it's the way we should have done it from the beginning.

rcorral:
Looks great!

PR #1467, opened 2016-12-02
Fix empty service tree view

orlandohohmeier:
Add missing onCreateGroup property.


nLight:
I've checked it out and it fixes the issue 👍

PR #1468, opened 2016-12-02
Fix ContainerServiceFormSection

Poltergeist:
This fixes an issue in the ContainerServiceFormSection which resulted in throwing errors, because there where no keys on jsx elements in an array.

image


nLight:
😅  my fault, thanks for fixing that one!

PR #1469, opened 2016-12-02
Fix icon leftover in the Alert component

nLight:
This PR brings back an icon in the Alert component that went missing after the icons update


wavesoft:
Yahoo ✨  💯

PR #1470, opened 2016-12-02
Run the complete validation on each blur. Always.

nLight:
This PR removes flags and checks that we don’t have time to do properly. So we will run the complete ramp validation on every form change. Otherwise I could run a service without providing ID etc. And the form validation didn’t work for me in general, only JSON editor validation worked.


rcorral:
This looks good for now, but we need to be careful as I already notice this slowing down the performance of the form.

PR #1471, opened 2016-12-02
Upgrading raml-validator-loader to v0.1.6

wavesoft:
## v0.1.6 Changelog

- **Change**: The path of missing property errors now points to the actual missing field.
- **Change**: This feature is controlled through the `missingPropertiesOnTheirPath` option flag

PR #1474, opened 2016-12-02
JSONEditor: Better handling of error paths

wavesoft:
This PR is fixing some issues the #1471 is introducing. In detail, when the JSON editor cannot find a token on the given path, it will try the parent paths until it reaches the root.


nLight:
This is so awesome!

nLight:
screen shot 2016-12-02 at 17 57 11
✨ 💯

orlandohohmeier:
✨ Awesome

PR #1480, opened 2016-12-02
Changes to Task health and state (DCOS-11264)

MatApple:
Feature reference: [DCOS-11264](https://mesosphere.atlassian.net/browse/DCOS-11264)

Changes:
1. Split up Status and Health into separate columns
2. Health dots only apply to active tasks
3. Health dots should have tooltips that explain what they mean e.g. Healthy, Unhealthy, No health checks available
4. Completed task rows are visually greyed out
5. States that are errors or considered bad are colored red

The new table-row state contained in the directory `src/styles/content/tables/table-cell-inactive` will be moved to CNVS (once approved) where it will join table-cell-danger, table-cell-warning, table-cell-success and table-cell-selected.

Pod Instances Table:

Task Table:

Tooltips:


src/styles/content/tables/table-cell-inactive/styles.less
```diff
@@ -0,0 +1,131 @@
+& when (@table-cell-inactive-enabled) {
+  /* Table */
+
+  .table {
+
+    tr.inactive td,
+    td.inactive {
+      .element-text-style(table-cell-inactive);
```
jfurrow: Let's not use these mixins; instead just write the actual CSS properties. It will reduce this down to just a few lines of code.

src/styles/content/tables/styles.less
```diff
@@ -11,6 +11,29 @@
...
+        .status-text,
+        .table-cell-link-primary,
+        .table-cell-link-secondary,
+        .table-cell-value {
+          color: @table-cell-inactive-color;
+          text-decoration: none;
+        }
+
+        .expanding-table-primary-cell.is-expandable:before {
+          border-left-color: @table-cell-inactive-color;
```
jfurrow: Let's write this as:
```less
.expanding-table-primary-cell {

  &.is-expandable {

    &:before {
      border-left-color: @table-cell-inactive-color;
    }
  }
}
```

src/styles/content/tables/table-cell-inactive/variables.less
```diff
@@ -0,0 +1,373 @@
...
+
+/*
+ * Table Cell Styling
+ */
+
+/* Default */
+
+// Normal
+
+@table-cell-inactive-color: color-lighten(@black, 75);
```
jfurrow: Without the mixins, this would be only 13 variables :)

plugins/services/src/js/containers/pod-instances/PodInstancesTable.js
```diff
@@ -363,16 +374,40 @@ class PodInstancesTable extends React.Component {
...
+      tooltipContent = 'No health checks available';
+    }
+
+    return this.renderWithClickHandler(rowOptions, (
+      <div className="flex-box flex-box-align-vertical-center
+        table-cell-flex-box flex-align-items-center
+        flex-direction-top-to-bottom">
+        <div className="table-cell-icon">
+          <Tooltip anchor="center" content={tooltipContent}>
+            <span className={status.dotClassName}></span>
```
jfurrow: We also need to add `flush` to this `span`, otherwise the tooltip won't be centered properly.

Also let's make it a self-closing element, e.g. `<span />`

jfurrow:
Nice! This might be for a different PR, but the status dots look different for tasks/pods. One has a thin stroke while the other is filled. Do we know which is correct?

src/styles/content/tables/table-cell-inactive/variables.less
rcorral: can't we use the mixins to generate this?

rcorral:
Awesome!

I think we should make the health/status columns hide together.
gif

rcorral:
All of the tooltips say "Healthy", I think it's because we cache the column.
img

src/styles/content/tables/table-cell-inactive/styles.less
MatApple: I matched the CNVS format for this. The goal is to remove this from here and merge into CNVS so I think we should leave this as is unless @ashenden suggests otherwise. Thoughts?

src/styles/content/tables/table-cell-inactive/variables.less
MatApple: see https://github.com/dcos/dcos-ui/pull/1480#discussion_r90745837

plugins/services/src/js/containers/pod-instances/PodInstancesTable.js
MatApple: done!

src/styles/content/tables/styles.less
MatApple: done!

MatApple:
@jfurrow - addressed all comments. The health dots with just a border will be present on tasks that are in a completed state (i.e. killed, complete, error, etc). A grey health dot with fill represents a task that is still active but has no health checks.

@rcorral - fixed the caching and hiding of health/status so they are in sync.

jfurrow:
@MatApple Thanks for the explanation! It looks great, good catch!!!

src/styles/content/tables/table-cell-inactive/styles.less
orlandohohmeier: I'm with @jfurrow and would vote against using those mixins due to the named reasons. The CSS that we write should be in line with our needs not those of CNVS.

ashenden:
FWIW, I suggested this additional row state be a PR against CNVS not dcos-UI.    
  
  

  
  
  

  
  
>

src/styles/content/tables/table-cell-inactive/styles.less
MatApple: I'm in favor of the suggestion. The intention was to put this code in CNVS so in that case it would have to follow the best practices of CNVS. Making this change means we keep this code and maintain it ourselves.

MatApple:
This PR also fixes a bug where active PODs weren't previously shown on the "active" filter selection. That also meant that active PODs reported the wrong health status. Fixed this here https://github.com/dcos/dcos-ui/pull/1480/files#diff-ff5234fa70f403ffc708e3774cac23a9R27

MatApple:
@ashenden  cc/ @jfurrow, @rcorral, @orlandohohmeier  that was the intention. I wanted to get it merged here so that it works and is testable for reviewers before migrating it to a CNVS PR.

src/styles/content/tables/table-cell-inactive/styles.less
jfurrow: @MatApple I reduced it to the bare minimum in this commit (Rafa asked me to help out to keep the PR moving, I hope you don't mind): https://github.com/dcos/dcos-ui/pull/1480/commits/3b4b6bcb9feb23cae4ef0c15e59eac0091ddaf65.

We'll just remove it when it's released in CNVS, so we won't be maintaining it in both places.

@ashenden If you'd like you can copy the mixins and variables definitions from this commit: https://github.com/dcos/dcos-ui/pull/1480/commits/5526d450c2b04ea91987b84a90aa1e88930caa3b

jfurrow:
I just noticed the table is displaying strangely for me, but it's not from this PR. I'll fix in master.

jfurrow:
retest this please

PR #1481, opened 2016-12-02
Introduce PageHeaderActionsMenu component

jfurrow:
This PR introduces a component called `PageHeaderActionsMenu` which is really just a wrapper for the `Dropdown`. Here's how I propose it works:

```jsx
<PageHeaderActionsMenu>
  <span clickHandler={() => console.log('hello')}>
    Hello!
  </span>
  <span clickHandler={() => console.log('goodbye')}>
    Goodbye!
  </span>
</PageHeaderActionsMenu>
```

Each child of `PageHeaderActionsMenu` will be rendered as an item in the `Dropdown` without being modified. When selected, it will call the child's `clickHandler` prop.

Sneak peek:

---

@philipnrmn In the new `PageHeader` component, this should be wrapped inside `div.button-collection.flush-bottom`, which will provide the proper layout for a neighboring action button. Here's an example of how that markup might look:

```jsx
<div className="page-header-actions">
  <div className="button-collection flush-bottom">
    <button className="button button-link">
      <Icon id="plus" size="mini" />
    </button>
    <PageHeaderActionsMenu>
      <span clickHandler={() => console.log('hello')}>
        Hello!
      </span>
      <span clickHandler={() => console.log('goodbye')}>
        Goodbye!
      </span>
    </PageHeaderActionsMenu>
  </div>
</div>
```

### DIFF
--- a/src/js/components/PageHeaderActionsMenu.js
+++ b/src/js/components/PageHeaderActionsMenu.js
@@ -28,18 +28,17 @@ const getDropdownItemFromComponent = (child, index) => {
   };
 };
 
-const PageHeaderActionsMenu = (props) => {
+const PageHeaderActionsMenu = ({children}) => {
   return (
     <Dropdown
       buttonClassName="button button-link"
-      items={getMenuItems(props.children)}
+      items={getMenuItems(children)}
       onItemSelection={handleItemSelection}
       persistentID="trigger"
       transition={true}
       dropdownMenuClassName="dropdown-menu"
       dropdownMenuListClassName="dropdown-menu-list"
-      wrapperClassName="dropdown"
-      {...props} />
+      wrapperClassName="dropdown" />
   );
 };
 
--- a/src/js/components/PageHeaderActionsMenu.js
+++ b/src/js/components/PageHeaderActionsMenu.js
@@ -20,11 +20,11 @@ const handleItemSelection = (item) => {
   }
 };
 
-const getDropdownItemFromComponent = (child) => {
+const getDropdownItemFromComponent = (child, index) => {
   return {
     clickHandler: child.props.clickHandler,
     html: child,
-    id: child.key
+    id: index
   };
 };
 
@@ -53,8 +53,7 @@ PageHeaderActionsMenu.propTypes = {
     React.PropTypes.shape({
       props: React.PropTypes.shape({
         clickHandler: React.PropTypes.func
-      }),
-      key: React.PropTypes.string.isRequired
+      })
     })
   )
 };
--- a/src/js/components/PageHeaderActionsMenu.js
+++ b/src/js/components/PageHeaderActionsMenu.js
@@ -15,8 +15,8 @@ const getMenuItems = (children) => {
 };
 
 const handleItemSelection = (item) => {
-  if (item.clickHandler) {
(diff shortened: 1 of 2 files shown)


src/js/components/PageHeaderActionsMenu.js
```diff
@@ -0,0 +1,62 @@
...
+      html: <Icon id="ellipsis-vertical" size="mini" />,
+      id: 'trigger'
+    },
+    ...React.Children.map(children, getDropdownItemFromComponent)
+  ];
+};
+
+const handleItemSelection = (item) => {
+  if (item.clickHandler) {
+    item.clickHandler();
```
rcorral: Would it be better if we just called this `onClick` instead of `clickHandler`?

orlandohohmeier: I'm not sure if I completely understand your reasoning, but I would argue that in case we extend the functionality we should also add new handlers not reuse the`clickHandler`.  Let's follow our best practices and rename the handler to `onClick`.

jfurrow: My concern with using `onClick` is that the element which gets rendered inside the dropdown menu will have padding surrounding it. See the following screenshots:

This element has the native `onClick` handler:


This element is wrapping the child passed to `PageHeaderActionsMenu`:


So if the user clicks on the `span` element, the `span`'s native event handler function will get called. If they click outside the `span` element, say in the area that has padding, then the `PageHeaderActionsDropdown` will be the callee of the `onClick` function.

Additionally, if we use `onClick` we would need to proxy it in order to stop propagation. Otherwise this `onClick` method will be called twice if the user clicks on the `span` element (once for the native event handler, and again for when `handleItemSelect` calls it). This is certainly not difficult to implement, I just wanted to point that out.

Maybe `onItemSelect` could be a better prop name? Or do you think we should go with `onClick` still?

orlandohohmeier: Ok, now I get it... `onItemSelect`sounds good!

src/js/components/PageHeaderActionsMenu.js
```diff
@@ -0,0 +1,62 @@
...
+    <Dropdown
+      buttonClassName="button button-link"
+      items={getMenuItems(props.children)}
+      onItemSelection={handleItemSelection}
+      persistentID="trigger"
+      transition={true}
+      dropdownMenuClassName="dropdown-menu"
+      dropdownMenuListClassName="dropdown-menu-list"
+      wrapperClassName="dropdown"
+      {...props} />
```
orlandohohmeier: Let's not pass all the props to the dropdown. We should only use this technique in very rare cases as it is really hard to track down those props.

orlandohohmeier:
@jfurrow I've renamed the item select handler and adjusted the implementation to not pass all props along. Please take a look.

jfurrow:
Thanks @orlandohohmeier!

philipnrmn:
Looks like we missed a bit [here](https://github.com/dcos/dcos-ui/pull/1481/files#diff-753ec596a66950e535e5334fcfeba7e0R25), we should have converted clickHandler to onItemSelect. I'll open a fix-up PR.

PR #1486, opened 2016-12-04
Two small fixes for the kebab menu

philipnrmn:
This builds on the work in #1481 and fixes two small errors:

- the `anchorRight` property was not being passed to the `Dropdown`, so the dropdown on the right of the page header was appearing off the screen.
- the `clickHandler` child property was not updated to `onItemSelect` everywhere, so it defined actions were not working.


rcorral:
Thanks Phil!

PR #1482, opened 2016-12-03
[1/11] Page template headers

philipnrmn:
This PR introduces the new Page Header components. In order to keep noise low, I've stayed relatively faithful to the original interface of the components. 

The page header is wrapped in a utility component and added to the Page namespace in order to allow subclasses of Page to override it. 

The intent is to allow every view to control its header by defining itself as a page, eg

```
<Page>
  <Page.Header breadcrumbs={...} />
  {usual page content here}
</Page>
```

PR #1487, opened 2016-12-05
[3/11] Implement new header in Packages pages

rcorral:
---
⚠️   _This branch depends on features that will be introduced with #1482_

---

### DIFF
--- a/src/js/pages/UniversePage.js
+++ b/src/js/pages/UniversePage.js
@@ -1,72 +1,17 @@
-import mixin from 'reactjs-mixin';
-/* eslint-disable no-unused-vars */
 import React from 'react';
-/* eslint-enable no-unused-vars */
-import {routerShape} from 'react-router';
 
 import Icon from '../components/Icon';
-import RouterUtil from '../utils/RouterUtil';
-import SidebarActions from '../events/SidebarActions';
-import TabsMixin from '../mixins/TabsMixin';
 
-class UniversePage extends mixin(TabsMixin) {
-  constructor() {
-    super();
-
-    this.tabs_tabs = {};
-    this.state = {};
-  }
-
-  componentWillMount() {
-    this.updateCurrentTab();
-  }
-
-  componentWillReceiveProps(nextProps) {
-    this.updateCurrentTab(nextProps);
-  }
-
-  updateCurrentTab(nextProps) {
-    let {routes} = nextProps || this.props;
-    let currentTab = RouterUtil.reconstructPathFromRoutes(routes);
-
-    // Universe Tabs
-    this.tabs_tabs = {
-      '/universe/packages': 'Packages',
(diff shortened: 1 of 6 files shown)


philipnrmn:
Rebased with the latest state of #1482

philipnrmn:
Looks great! I'm happy with the code, one question about the breadcrumbs - do you think this should be `Packages > Installed` rather than just `Installed Packages`?

screen shot 2016-12-04 at 9 07 56 pm

rcorral:
Good question. Let's ask design, but not let it block the PR.

ashenden:
@philipnrmn is correct.  Sections like "Packages" and "Installed" are really tier-1 sections, but we are choosing to nest them under the "Packages" nav item from a category POV.  So, these sections would read "Installed", not "Packages > Installed" -- or "Users", not "Organization > Users".  They would all use the icon of their category.

BTW, the naming should be 1:1 with the sidebar label.  So "Installed" not "Installed Packages".  We will likely tweak these names over the weeks leading up to GA, so be prepared for that.

src/js/pages/UniversePage.js
```diff
@@ -52,13 +51,7 @@ class UniversePage extends mixin(TabsMixin) {
   }
 
   render() {
-    return (
-      <Page
```
nLight: Having deleted this we can also delete all the methods from the class and inherit it directly from `React.Component` I tried and it still worked :)

src/js/pages/universe/PackagesTab.js
```diff
@@ -13,9 +13,18 @@ import FilterInputText from '../../components/FilterInputText';
 import Image from '../../components/Image';
 import InstallPackageModal from '../../components/modals/InstallPackageModal';
 import Loader from '../../components/Loader';
+import Page from '../../components/Page';
 import Panel from '../../components/Panel';
 import StringUtil from '../../utils/StringUtil';
 
+const UniverseBreadcrumbs = () => {
+  const crumbs = [
+    <Link to="universe/packages" key={-1}>Universe</Link>
```
nLight: Going back to John's comment, do we want to change it to `Packages` so it matches with the Sidebar link?

jfurrow: I made this change :)

src/js/pages/UniversePage.js
jfurrow: Done! Good catch :)

jfurrow:
I implemented both of @nLight's requested changes, so dismissing his review!

This looks & works great!

MatApple:
retest this please

MatApple:
The Jenkins reporter seems to be hanging on this PR. Could be due to the Jenkins maintenance that just wrapped up. Tests are passing so I'm going to merge:

PR #1488, opened 2016-12-05
[4/11] Page Header for components health

rcorral:
---
~~⚠️   _This branch depends on features that will be introduced with #1482_~~ Merged

---

### DIFF
--- a/src/js/pages/system/UnitsHealthDetail.js
+++ b/src/js/pages/system/UnitsHealthDetail.js
@@ -1,19 +1,36 @@
 import mixin from 'reactjs-mixin';
+import {Link} from 'react-router';
 /* eslint-disable no-unused-vars */
 import React from 'react';
 /* eslint-enable no-unused-vars */
 import {StoreMixin} from 'mesosphere-shared-reactjs';
 
-import Breadcrumbs from '../../components/Breadcrumbs';
-import DetailViewHeader from '../../components/DetailViewHeader';
 import FilterBar from '../../components/FilterBar';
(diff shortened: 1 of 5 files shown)


jfurrow:
I pushed a commit with a few modifications; otherwise it LGTM!

src/js/pages/system/UnitsHealthDetail.js
```diff
@@ -138,49 +146,52 @@ class UnitsHealthDetail extends mixin(StoreMixin) {
...
+    let content = null;
+    let unit = null;
 
     if (hasError) {
-      return this.getErrorNotice();
-    }
-
-    if (isLoadingUnit || isLoadingNodes) {
-      return this.getLoadingScreen();
+      content = this.getErrorNotice();
```
mlunoe: Could we turn this into a separate function (e.g. `this.getContent()`) so we can `return` and keep them as simple `if`-statements?

src/js/pages/system/UnitsHealthNodeDetail.js
```diff
@@ -68,36 +100,40 @@ class UnitsHealthNodeDetail extends mixin(StoreMixin) {
 
   render() {
     let {hasError, isLoadingNode, isLoadingUnit} = this.state;
+    let content = null;
+    let node = null;
+    let unit = null;
 
     if (hasError) {
-      return this.getErrorNotice();
+      content = this.getErrorNotice();
```
mlunoe: Same here, let's create a separate function to get this

MatApple:

PR #1489, opened 2016-12-05
[5/11] Headers for jobs

philipnrmn:
---
~~⚠️   _This branch depends on features that will be introduced with #1482_~~ Merged!

---

This is a WIP PR which adds headers to the jobs section. 

The TaskDetail section is still missing, this is awkward because there are several valid routes for the view via services, jobs, and apparently network also. The answer to this is probably to write individual pages for each route which share the common component.

### DIFF
--- a/src/js/pages/__tests__/JobsTab-test.js
+++ b/src/js/pages/__tests__/JobsTab-test.js
@@ -1,6 +1,7 @@
 jest.dontMock('../jobs/JobsTab');
 jest.dontMock('../../mixins/InternalStorageMixin');
 jest.dontMock('../../mixins/SaveStateMixin');
+jest.dontMock('../../components/Page');
 jest.mock('../../stores/UserSettingsStore');
 jest.mock('../../mixins/QueryParamsMixin');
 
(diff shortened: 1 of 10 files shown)


src/js/pages/__tests__/JobsTab-test.js
```diff
@@ -93,7 +94,7 @@ describe('JobsTab', function () {
       ).toBeDefined();
     });
 
-    it('renders correct empty panel', function () {
+    xit('renders correct empty panel', function () {
```
mlunoe: I'm guessing this is the reason why this is WIP?

mlunoe:
Code looks good, so it gets a label, but let's fix that test

philipnrmn:
The hold merge is because of the missing taskdetail which will be coming soon. I'd forgotten that test, thanks for picking it up :)

orlandohohmeier:
Unfortunately there is no header on the empty job view and the job actions are missing...

Empty Job Tree | Job Detail
--- | --- 
  | screen shot 2016-12-07 at 12 03 53

nLight:
Action buttons are missing

master | current
--- | --- 
screen shot 2016-12-07 at 12 13 57 |  screen shot 2016-12-07 at 12 11 58

orlandohohmeier:

orlandohohmeier:
@nLight I've addressed all issues...

nLight:
💪  Let's get this in

PR #1492, opened 2016-12-05
DCOS-11871: Pass Create Service Form errors to all the missing partials

nLight:
In this PR I took care of errors in
Environment
Health checks
Volumes

ℹ️  I took the risk and used some lenses here and there, slap my hands if I should stop doing this.

### DIFF
--- a/plugins/services/src/js/components/forms/HealthChecksFormSection.js
+++ b/plugins/services/src/js/components/forms/HealthChecksFormSection.js
@@ -22,59 +22,59 @@ class HealthChecksFormSection extends Component {
     }
 
     return (
-        <AdvancedSection>
-          <AdvancedSectionLabel>
-            Advanced Health Check Settings
-          </AdvancedSectionLabel>
-          <AdvancedSectionContent>
-            <div className="flex row">
-              <FormGroup
-                  className="column-3"
-                  showError={Boolean(healthChecks[key])}>
-                <FieldLabel>Grace Period (s)</FieldLabel>
-                <FieldInput
-                    name={`healthChecks.${key}.gracePeriodSeconds`}
-                    type="number"
-                    min="0"
-                    value={healthCheck.gracePeriodSeconds}/>
-                <FieldError>{healthChecks[key]}</FieldError>
-              </FormGroup>
-              <FormGroup
-                  className="column-3"
-                  showError={Boolean(healthChecks[key])}>
-                <FieldLabel>Interval (s)</FieldLabel>
-                <FieldInput
-                    name={`healthChecks.${key}.intervalSeconds`}
-                    type="number"
-                    min="0"
-                    value={healthCheck.intervalSeconds}/>
-                <FieldError>{healthChecks[key]}</FieldError>
-              </FormGroup>
-              <FormGroup
-                  className="column-3"
-                  showError={Boolean(healthChecks[key])}>
-                <FieldLabel>Timeout (s)</FieldLabel>
-                <FieldInput
-                    name={`healthChecks.${key}.timeoutSeconds`}
-                    type="number"
-                    min="0"
-                    value={healthCheck.timeoutSeconds}/>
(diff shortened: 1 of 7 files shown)


plugins/services/src/js/components/forms/HealthChecksFormSection.js
```diff
@@ -13,89 +14,99 @@ import FormGroupContainer from '../../../../../../src/js/components/form/FormGro
 
 import {FormReducer as healthChecks} from '../../reducers/serviceForm/HealthChecks';
 
+const errorsLens = Objektiv.attr('healthChecks', []);
+
 class HealthChecksFormSection extends Component {
+  getErrors(key) {
```
orlandohohmeier: Let's pleas be consistent and either remove this method or adjust it so that we can use it in L:94 as well.

plugins/services/src/js/reducers/serviceForm/Volumes.js
```diff
@@ -75,8 +75,11 @@ module.exports = {
             break;
         }
 
-        return [].concat(this.externalVolumes,
-          this.localVolumes.filter(filterHostVolumes.bind(this)).map(mapLocalVolumes));
+        return [].concat(
```
orlandohohmeier: Let's add a comment here and in line 120 to document the operation.

plugins/services/src/js/components/forms/HealthChecksFormSection.js
```diff
@@ -13,89 +14,99 @@ import FormGroupContainer from '../../../../../../src/js/components/form/FormGro
...
+  }
+
   getAdvancedSettings(healthCheck, key) {
-    const {healthChecks} = this.props.errors;
     if (healthCheck.protocol !== 'COMMAND' && healthCheck.protocol !== 'HTTP' &&
       healthCheck.protocol !== 'HTTPS') {
       return null;
     }
 
+    const errors = this.getErrors(key);
```
wavesoft: Just to check, since I am still not used on the lens logic. Is this roughly equal to:
```
this.props.errors.healthChecks[key] || {} right?
```

Then can we either use lenses everywhere or the approach in EnvironmentFormSection.js (or the other way around)?

orlandohohmeier:
I like how you used lenses to retrieve the errors, but we should make sure it's consistent (at least throughout the form).

plugins/services/src/js/reducers/serviceForm/Volumes.js
nLight: I wish I knew it, /cc @Poltergeist

plugins/services/src/js/components/forms/HealthChecksFormSection.js
nLight: it is more like
`(this.props.errors || {healthChecks: []})[key] || {}`

nLight: I mean, it's more complex. I use default value for every lens so that we can go deeper even when we don't have the full path in `errors`

plugins/services/src/js/reducers/serviceForm/Volumes.js
nLight: I had to change order here to match the form, otherwise errors came in improper order

Poltergeist:
```JS
/**
 * This is concatinating the contents of localVolumes and externalVolumes
 * we need to filter localVolumes for hostVolumes if no docker is available that 
 * is happening in the FilterHostVolumes method. This is because in the ui local
 * Volumes and Host Volumes are represented in the same space.
 * Because both Volumes are in the same reducer we also need to transform them
 * to either a localVolume or a hostVolume because they both have a different format.
 * External Volumes is just external volumes.
 */
```

I hope this is good enough.

nLight:
Addressed all the comments 🙇

wavesoft:
🎉

PR #1497, opened 2016-12-05
DCOS-11875: Fix 404 -> multiple sidebar registrations

nLight:
This PR addresses the issue when a user hits 404 and goes back Navigation would re-register all the routes.


orlandohohmeier:
LGTM! 🚢

mlunoe:
Pushed a small fix where the PageNotFound route was nested under the wrong parent.

Before:


After:

PR #1501, opened 2016-12-06
Fix 404 for all the plugins

nLight:
I kept NotFoundPage Route being a child of Index while keeping it away from being in the main routes array.

Otherwise plugins routes go after the glob * NotFoundPage Route and therefore become unreachable.

### DIFF
--- a/src/js/routes/index.js
+++ b/src/js/routes/index.js
@@ -1,3 +1,4 @@
+import React from 'react';
 import {Route, Redirect} from 'react-router';
 import {Hooks} from 'PluginSDK';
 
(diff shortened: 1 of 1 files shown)


src/js/routes/index.js
```diff
@@ -49,12 +55,20 @@ function getApplicationRoutes() {
...
         {
           type: Route,
           id: 'index',
-          component: Index,
-          children: routes
+          children: routes,
+          component: PassThroughComponent
+        },
+        {
+          // Please make sure this is THE LAST route and is not inside of id: 'index'
```
orlandohohmeier: Let's explain the "why" as well...

nLight: // Please make sure this is THE LAST route and is not inside of id: 'index'
// as this is a glob route and matching stops here
// making all the routes going after this point unreachable

orlandohohmeier:
LGTM 💯 ✨

Poltergeist:
LGTM 💯

PR #1505, opened 2016-12-06
Improve the git hook helper

orlandohohmeier:
Adjust the install hook helper to properly check if a hook is already "linked" and only run the install if the hook isn't linked.

### DIFF
--- a/scripts/utils/git
+++ b/scripts/utils/git
@@ -22,35 +22,41 @@ install_hooks() {
   for hook in $1
   do
     # Check if hook is executable
-    if [ -x scripts/${hook} ]
+    if [ ! -x scripts/${hook} ]
     then
-      header "Create symlink for ${hook} hook"
-
-      # Don't replace the same file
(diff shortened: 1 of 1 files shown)


scripts/utils/git
```diff
@@ -22,36 +22,42 @@ install_hooks() {
   for hook in $1
   do
     # Check if hook is executable
-    if [ -x scripts/${hook} ]
+    if [! -x scripts/${hook} ]
```
nLight: Early returns are good 👍

nLight: but then spaces are mandatory around an expr so it should be 
`if [ ! -x scripts/${hook} ]` or it throws `[!: command not found`

orlandohohmeier: Oh dang, I must have removed the space during the rebase...

orlandohohmeier: ✅  Fixed

PR #1510, opened 2016-12-06
DCOS-11728: Force hardware acceleration for tooltips

jfurrow:
This PR implements two changes:
* The `will-change` property on the tooltip itself which should force hardware acceleration even when the tooltip is not being animated
* A specific anti-aliasing property to prevent the browser from switching on the fly

You may be wondering what the vendor prefixes are doing here! Autoprefixer doesn't currently handle these, so the vendor prefixes are sanctioned by Stylelint.

### DIFF
--- a/src/styles/components/tooltip/styles.less
+++ b/src/styles/components/tooltip/styles.less
@@ -1,6 +1,7 @@
 & when (@tooltip-enabled) {
 
   .tooltip {
+    will-change: opacity, transform, visibility;
     z-index: @z-index-tooltip;
 
     .tooltip-content {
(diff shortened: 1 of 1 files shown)


src/styles/components/tooltip/styles.less
```diff
@@ -1,14 +1,18 @@
...
   .tooltip {
+    will-change: opacity, transform, visibility;
     z-index: @z-index-tooltip;
 
     .tooltip-content {
       background: @tooltip-background-color;
       border-radius: @tooltip-border-radius;
       color: @white;
       font-size: @tooltip-font-size;
+      -moz-osx-font-smoothing: grayscale;
```
mlunoe: Let's add a comment for this, so we know why it is here, when we need to change it (when it gets added as a rule in autoprefixer)

mlunoe:
Otherwise this looks good. Let's :shipit: as soon as the comment is added

src/styles/components/tooltip/styles.less
jfurrow: Good call, done!

rcorral:
very impressed :notbad:

PR #1512, opened 2016-12-06
Renamed function. More stringent prop types.

rcorral:
This is a continuation from #1509. Github had a brain fart and wouldn't see the commit that I added to this branch...so now I need to put up another PR...
img


orlandohohmeier:
L👀ks good!

PR #1516, opened 2016-12-07
Page header styling updates

jfurrow:
This PR cleans up the page header styles and applies them to the new components. It also adds the `SidebarToggle` component from the old page header. :arrow_left: :arrow_right:

### DIFF
--- a/src/styles/components/page-header/variables.less
+++ b/src/styles/components/page-header/variables.less
@@ -21,11 +21,11 @@
 @page-header-sidebar-toggle-padding-screen-large: @base-spacing-unit-screen-large * 1/2;
 @page-header-sidebar-toggle-padding-screen-jumbo: @base-spacing-unit-screen-jumbo * 1/2;
 
-@page-header-sidebar-toggle-margin-right: @base-spacing-unit * 1/8;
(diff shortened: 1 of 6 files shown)


philipnrmn:
Looks nice 👍

nLight:
So g👀 d! ✨

src/styles/components/page-header/variables.less
```diff
@@ -34,71 +34,90 @@
...
+@page-header-breadcrumbs-margin-right-screen-large: @base-spacing-unit-screen-large * 1/2;
+@page-header-breadcrumbs-margin-right-screen-jumbo: @base-spacing-unit-screen-jumbo * 1/2;
+
+/*
+ * Page Header Content Sections
+ */
+
+// Layout
+
+@page-header-content-section-primary-padding-left: @icon-mini-width + @page-header-sidebar-toggle-padding + @page-header-sidebar-toggle-margin-right;
```
orlandohohmeier: Not blocking this PR but those variable assignments look funky and I wonder if we really need to calculate all of these paddings.

orlandohohmeier: I wonder if there is way to not use absolute positioning (e.g using flex) so that we don't need to calculate all of these values by hand. In anyway nothing that we need to change right now.

PR #1506, opened 2016-12-06
DCOS-10982:  Add healthchecks for multi containers

Poltergeist:
This introduces HealthChecks for Multi Containers
healthchecksforpods mov

This does include commits from #1502

PR #1502, opened 2016-12-06
DCOS-10983: Add multiple containers to the service create form

Poltergeist:
As requested by @orlandohohmeier here is a gif:
giphy

and here is a screen capture:

pods-multi-container mov

PR #1519, opened 2016-12-07
[multiContainer] add service picker and env vars

Poltergeist:
---
ℹ️   _This branch depends on features that will be introduced with ~~#1502~~ and ~~#1506~~._

---

This reintroduces the pod option in the picker and excludes the mount point for multi containers.


jfurrow:
This looks good to me! Nice work.

PR #1508, opened 2016-12-06
DCOS-10470: [6/10] Integrating plain-text DSL input field

wavesoft:
---
⚠️  ~~Depends on #1496~~ Merged

---

[Diff between 1496 & This](https://github.com/dcos/dcos-ui/compare/ic/feat/DCOS-10470-new-dsl-components-5...ic/feat/DCOS-10470-new-dsl-components-6)

This PR introduces the `DSLFilterField` and it integrates it in the `ServiceTreeView`:

dslfield mov

PR #1521, opened 2016-12-07
DCOS-10470: [7/10] Introducing DSLExpressionPart

wavesoft:
---
⚠️  ~~Depends on #1508~~ Merged

---

[Diff between #1508 and this](https://github.com/dcos/dcos-ui/pull/1521/commits/9bee05d5e2b0382ec0894bcdb154f5fdb2682214)

This PR introduces the `DSLExpressionPart` specifications that will be used by the `DSLForm` component (coming in a future PR) in order to specify which attributes it will be operating upon.

### DIFF
--- a/src/js/structs/DSLPropTypes.js
+++ b/src/js/structs/DSLPropTypes.js
@@ -1,0 +1,45 @@
+import DSLFilterTypes from '../constants/DSLFilterTypes';
+import {FilterNode} from './DSLASTNodes';
+
+/**
+ * The DSL PropTypes are used by components inserted in a `DSLInputForm` in
+ * order to specify the AST nodes they are managing.
+ *
+ * The functions and getters in this file are just shorthands to create virtual
+ * AST nodes (without any position information) that are used as comparision
+ * targets for selecting the correct nodes from the paresed AST nodes.
+ *
+ * @example <caption>How to use DSLPropTypes</caption>
+ *
+ * MyFormComponent.dslPropTypes = {
+ *
+ *   // Create a boolean property that is defined when the attribute is:healthy
+ *   // exists in the parsed DSL
+ *   is_healthy: DSLPropTypes.attrib('is', 'healthy'),
+ *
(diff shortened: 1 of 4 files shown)


nLight:
Naming it PropTypes could be rather confusing, as I understood you not gonna use it like:
```
Component.propTypes = {
   prop: DSLPropType
}
```

wavesoft:
Well, It's going to be like this:

```
Component.dslPropTypes = {
   prop: DSLPropTypes.fuzzy
}
```

And this is going to deliver the value of the matching DSL filter down to the component. Currently the idea is to pass them all down as an object in the `data` property. But I am open on ideas.

nLight:
So it's not React.PropTypes?

orlandohohmeier:
Discussing the concepts and ideas that this PR is introducing offline with @wavesoft

orlandohohmeier:
We've decided to not use introduce new concepts (prop type like static props...), but we still want these DSLPropTypes as they are going to be used as helpers inside of the DSL form.

src/js/structs/DSLPropTypes.js
```diff
@@ -0,0 +1,45 @@
...
+ * The DSL PropTypes are used by components inserted in a `DSLInputForm` in
+ * order to specify the AST nodes they are managing.
+ *
+ * The functions and getters in this file are just shorthands to create virtual
+ * AST nodes (without any position information) that are used as comparision
+ * targets for selecting the correct nodes from the paresed AST nodes.
+ *
+ * @example <caption>How to use DSLPropTypes</caption>
+ *
+ * MyFormComponent.dslPropTypes = {
```
orlandohohmeier: Please update the example as discussed...

wavesoft:
Renamed to `DSLExpressionPart` in order to make it clear that this is a way of targeting some particular part of the DSL expression.

orlandohohmeier:
🚂  Whoot Whoot

PR #1522, opened 2016-12-07
DCOS-10470: [8/10] Introducing utilities to operate on DSL expressions

wavesoft:
This PR introduces the `DSLUtil` for processing prop types and high-level DSL AST operations and the `DSLUpdateUtil` that provides the utilities for dynamically updating DSL expressions.

### DIFF
--- a/src/js/utils/DSLUpdateUtil.js
+++ b/src/js/utils/DSLUpdateUtil.js
@@ -107,9 +107,11 @@ const DSLUpdateUtil = {
       if ((src[start - 1] === ':') && (endingRegex.exec(src.substr(end)))) {
         start = position[0][0] + offset;
         end = position[1][1] + offset;
+
       // Otherwise, bleed left to remove the comma if we are part of multi-value
       } else if (src[start - 1] === ',') {
         start -= 1;
+
       // Or bleed right if we were the first item
       } else if (src[end] === ',') {
         end += 1;
@@ -121,6 +123,11 @@ const DSLUpdateUtil = {
       start -= 1;
     } else if (src[end] === ' ') {
       end += 1;
+    }
+
+    // Bleed left one more time if we have a lingering tailing ','
+    if ((src[start - 1] === ',') && (end >= src.length)) {
+      start -= 1;
     }
 
     // Return the striped result
--- a/src/js/utils/DSLUpdateUtil.js
+++ b/src/js/utils/DSLUpdateUtil.js
@@ -1,4 +1,5 @@
 import Config from '../config/Config';
+import DSLCombinerTypes from '../constants/DSLCombinerTypes';
 import DSLExpression from '../structs/DSLExpression';
 import DSLFilterTypes from '../constants/DSLFilterTypes';
 import DSLUtil from './DSLUtil';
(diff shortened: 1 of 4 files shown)


src/js/utils/DSLUpdateUtil.js
```diff
@@ -0,0 +1,197 @@
...
+
+  /**
+   * Delete the given list of nodes from the expression
+   *
+   * @param {DSLExpression} expression - The expression to update
+   * @param {Array} nodes - The node(s) to delete
+   * @returns {DSLExpression} expression - The updated expression
+   */
+  applyDelete(expression, nodes) {
+    var offset = 0;
```
nLight: `var` :trollface:

src/js/utils/DSLUpdateUtil.js
```diff
@@ -0,0 +1,197 @@
...
+    let expressionValue = nodes.reduce((value, deleteNode) => {
```
nLight: I'd still call it `node`

src/js/utils/DSLUpdateUtil.js
nLight: you could reduce `nodes` to `{expression, offset}` to keep `offset` in.

src/js/utils/DSLUpdateUtil.js
```diff
@@ -0,0 +1,197 @@
...
+   * @param {Array} newNodes - The node(s) to update with
+   * @returns {DSLExpression} expression - The updated expression
+   */
+  applyUpdate(expression, nodes, newNodes) {
+    const updateCount = Math.min(nodes.length, newNodes.length);
+    let expressionValue = expression.value;
+    let offset = 0;
+
+    // First update nodes
+    for (let i=0; i<updateCount; ++i) {
```
nLight: #nit spaces around operators

src/js/utils/DSLUpdateUtil.js
```diff
@@ -0,0 +1,197 @@
...
+        //
+        // position[0] contains the `label:` part
+        // position[1] contains the `value` part
+        //
+        const labelStart = position[0][0] + offset;
+        const labelEnd = position[0][1] + offset;
+        const valueStart = position[1][0] + offset;
+        const valueEnd = position[1][1] + offset;
+
+        return src.substr(0, labelStart) + `${label}:` +
```
nLight: Not sure if I like it...

src/js/utils/DSLUpdateUtil.js
```diff
@@ -0,0 +1,197 @@
...
+   * @param {String} src - The source expression string
+   * @param {ASTNode} node - The ast node to replace
+   * @param {ASTNode} newNode - The ast node to replace with
+   * @param {number} offset - The offset to apply on position indices
+   * @returns {String} Returns the updated string
+   */
+  updateNodeString(src, node, newNode, offset=0) {
+    const {position, filterType} = node;
+    let {filterParams: {text, label}} = newNode;
+    if ((Config.environment === 'development') &&
```
nLight: shall we `return` in production, or what will exactly happen?

wavesoft: We are throwing an exception 😄

src/js/utils/DSLUpdateUtil.js
wavesoft: It was kind of silly to represent the entire string with a template since I was adding more characters around the substr's making things more difficult to read.

The only place it actually did make sense was in the label, that's why the little template over there is on it's own.

src/js/utils/DSLUpdateUtil.js
wavesoft: `let` is supposed to be block-scoped, meaning that it won't be visible to the function inside (even though babel just converts it into `var`) in this case `var` has a reason for being there.

But yeah, I will use reduce with `{expression, offset}` instead.

src/js/utils/DSLUpdateUtil.js
wavesoft: Yup, it was copy-paste from the update function above 👍

src/js/utils/DSLUpdateUtil.js
nLight: Somewhere deep in the code?

nLight:
Shall we start? Retest this please.

wavesoft:
Slight pause until I fix some issues arised form a discussion with @orlandohohmeier yesterday.

src/js/utils/DSLUpdateUtil.js
wavesoft: Yeah, that's an invalid operation it should not be assumed to be ok to continue (my opinion).

wavesoft: Ok, let's return the source value for now. But I opt for a global error handler.

src/js/utils/DSLUpdateUtil.js
```diff
@@ -0,0 +1,308 @@
...
+        src += ', ';
+      }
+
+      return src + DSLUtil.getNodeString(node);
+    }
+
+    // On attributes with OR operator we first check if we already have an
+    // attribute with such label and we prefer creating multi-value expression
+    const attribNodes = DSLUtil.reduceAstFilters(fullAst,
+      (memo, filter) => {
```
orlandohohmeier: shouldn't this fit in one line? #nit

src/js/utils/DSLUtil.js
```diff
@@ -0,0 +1,238 @@
...
+   * This function walks the given AST and extracts all the nodes matching the
+   * given filter node.
+   *
+   * @param {ASTNode} ast - The AST node to walk on
+   * @param {FilterNode} filter - The reference AST node to compare against
+   * @returns {Array} Returns an array of FilterNodes tha match filter
+   */
+  findNodesByFilter(ast, filter) {
+    return DSLUtil.reduceAstFilters(ast,
+      (memo, astFilter) => {
```
orlandohohmeier: shouldn't this fit in one line? #nit

wavesoft:
Addressed the #nits, but there is a small bug on the delete function, fixing...

orlandohohmeier:
🚢 🚢 🚢

PR #1523, opened 2016-12-07
Fix access denied page styling

jfurrow:
This PR simplifies the classes applied in `AccessDeniedPage` and adds `application-wrapper` which is meant to wrap everything in DC/OS UI.

💩 Before:

✨ After:


nLight:
giphy

PR #1534, opened 2016-12-07
Removing Destroy action for DCOS Frameworks

MatApple:
https://mesosphere.atlassian.net/browse/DCOS-9723


rcorral:
You can still destroy from here
img

rcorral:
could you point this PR to `master`?

MatApple:
@rcorral good catch with dropdown in collection view. I'd rather not point it at master because it will be nearly as much work to resolve conflicts as doing this work itself ;)

MatApple:
@rcorral fixed your catch.

PR #1544, opened 2016-12-08
DCOS-11958: Add missing task states

nLight:
This PR adds missing task states.

This is a quickfix™

We should consider having a function returning `{stateTypes, displayName}` instead of object lookup so we could work around a new missing task state. Sadly proxies are still missing in IE.

### DIFF
--- a/plugins/services/src/js/constants/TaskStates.js
+++ b/plugins/services/src/js/constants/TaskStates.js
@@ -52,6 +52,31 @@ const TaskStates = {
   TASK_ERROR: {
     stateTypes: ['completed', 'failure'],
     displayName: 'Error'
+  },
+
+  TASK_GONE: {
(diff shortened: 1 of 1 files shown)


wavesoft:
LGTM! ✨

orlandohohmeier:
Let's add a null check to the task view so that it doesn't fail the next time if someone updates the task states.

nLight:
@orlandohohmeier there are bunch of places to add the check, that's why I'd opt-in for a getter instead of `[]` lookup

orlandohohmeier:
@nLight sounds like a good plan, but also like a rather large refactoring. Let's land this one and the take care of the refactoring. Please open an JIRA to not forget.

orlandohohmeier:
Code l👀ks good!

PR #1546, opened 2016-12-08
DCOS-11960: Add missing leading slash

nLight:
This PR adds a leading slash to breadcrumbs, because otherwise one can notice a jump in the address bar.


orlandohohmeier:
LGTM! 🚢

jfurrow:
Good catch! I noticed this but didn't realize the fix was so simple :)

PR #1554, opened 2016-12-08
Adds metadata for tracking js errors.

rcorral:


plugins/tracking/hooks.js
```diff
@@ -52,10 +54,18 @@ module.exports = {
   pluginsConfigured() {
     DOMUtils.appendScript(document.head, segmentScript);
 
-    window.analytics.ready(function () {
-      function updateTrackJSConfiguration() {
+    window.analytics.ready(() => {
+      let updateTrackJSConfiguration = () => {
         global.trackJs.configure({version: MetadataStore.version});
-      }
+        global.trackJs.addMetadata('version', MetadataStore.version);
```
jfurrow: It's possible for the metadata endpoint not to be called if the user's permissions aren't met. Are we concerned about this?

rcorral:
Merging this in as @jfurrow's concern is already being accounted for.

PR #1555, opened 2016-12-08
Bug fix: Fixing log and file viewer routes, ref. DCOS-11948

mlunoe:
This PR fixes the task detail routes

NB: @jfurrow @rcorral You should be added to an _open_ cluster that has logs, if you want to check it out


jfurrow:
:ship: Nice!

philipnrmn:
Awesome :)

PR #1559, opened 2016-12-09
DCOS-11939: Pass update callback to ClusterHeader

jfurrow:
This PR allows an `onUpdate` callback function to be passed to the `ClusterHeader` component. This is necessary for the `Sidebar` because the `ClusterHeader`'s size will update when it receives the IP address of the cluster. When this happens, the height of the neighboring Gemini container needs to be updated.

A symptom of this was the Sidebar's dock icon rendering slightly off the viewport.

Before:

After:


rcorral:
Thanks!

PR #1560, opened 2016-12-09
Bug fix: Duplicate logs from range request, ref. DCOS-11818, DCOS-11819, DCOS-11820

mlunoe:
This PR fixes a few bugs for the log view, that should now:
1. Close open connections when done viewing a log file.
2. Handle when the logs take up _more_ vertical space than container better
3. Handle when the logs take up _less_ vertical space than container (including requesting to see if we have reached the top)
4. Handle receiving duplicate logs (requesting previous logs with `skip_prev` was buggy, so we are using `read_reverse` instead).


rcorral:
Works grrrreat! 🙌

PR #1561, opened 2016-12-09
DCOS-12009: Don't lock JSON editor when user types

nLight:
Turns out our JSON editor is smart enough to detect where it should update and where not.
So this block is:
a) redundant
b) causes bugs described in the ticket


nLight:
retest this please

nLight:
I'm dropping this one for now

PR #1549, opened 2016-12-08
DCOS-10470: [9/10] Introducing form-dropdown to DSLFilterField

wavesoft:
This PR introduces the form UI to the `DSLFilterField`

image

PR #1564, opened 2016-12-09
DCOS-10470: [10/10] Introducing the "Other" section

wavesoft:
---
~~:warning: This PR depends on #1549~~

---

Final parts of the DSL

image

### DIFF
--- a/plugins/services/src/js/components/dsl/ServiceOtherDSLSection.js
+++ b/plugins/services/src/js/components/dsl/ServiceOtherDSLSection.js
@@ -10,9 +10,9 @@ import FieldLabel from '../../../../../../src/js/components/form/FieldLabel';
 import FormGroup from '../../../../../../src/js/components/form/FormGroup';
 
 const PARTS = {
-  is_pod: DSLExpressionPart.attrib('is', 'pod'),
-  is_universe: DSLExpressionPart.attrib('is', 'universe'),
-  has_volumes: DSLExpressionPart.attrib('has', 'volumes')
+  is_pod: DSLExpressionPart.attribute('is', 'pod'),
+  is_universe: DSLExpressionPart.attribute('is', 'universe'),
+  has_volumes: DSLExpressionPart.attribute('has', 'volumes')
 };
 
 class ServiceOtherDSLSection extends React.Component {
--- a/plugins/services/src/js/components/dsl/ServiceOtherDSLSection.js
+++ b/plugins/services/src/js/components/dsl/ServiceOtherDSLSection.js
@@ -1,0 +1,79 @@
+import React, {PropTypes} from 'react';
+
+import DSLCombinerTypes from '../../../../../../src/js/constants/DSLCombinerTypes';
+import DSLExpression from '../../../../../../src/js/structs/DSLExpression';
+import DSLExpressionPart from '../../../../../../src/js/structs/DSLExpressionPart';
+import DSLFormWithExpressionUpdates from '../../../../../../src/js/components/DSLFormWithExpressionUpdates';
+import DSLUtil from '../../../../../../src/js/utils/DSLUtil';
+import FieldInput from '../../../../../../src/js/components/form/FieldInput';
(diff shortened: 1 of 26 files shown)


wavesoft:
retest this please

wavesoft:
let's try to retest this again please

wavesoft:
let's try again to retest this please

wavesoft:
Rebased

mlunoe:
Found this issue:


Steps to reproduce:
- Click filter dropdown
- Check 'Volumes' 
- Check 'Pod'
- Uncheck 'Volumes'

mlunoe:
Code looks good though, and works great otherwise!

rcorral:
@wavesoft I think the "reset" button should work when you're in a bad state. Even if that means clearing everything. Otherwise the user still has to do it themselves. What do you think?
cc/ @leemunroe 
img

wavesoft:
@rcorral ok, I thought that "Reset" was "Undo my changes so far", meaning that it will bring the UI to the state it was before the dropdown. However yes, that should be addressed.

orlandohohmeier:

Poltergeist:

wavesoft:
let's retest this please again

src/js/utils/DSLUpdateUtil.js
```diff
@@ -3,11 +3,74 @@ import DSLCombinerTypes from '../constants/DSLCombinerTypes';
...
+    src = src.replace(/^\s*,/g, ''); // Orphan commas (left-side)
+    src = src.replace(/,\s*$/g, ''); // Orphan commas (right-side)
+    src = src.replace(/\(\s*,\s*/g, '('); // Orphan commas (left-paren)
+    src = src.replace(/\s*,\s*\)/g, ')'); // Orphan commas (right-paren)
+    src = src.replace(/:,+/g, ':'); // Orphan commas (multi-value)
+    src = src.replace(/\s+,\s+/g, ', '); // Commas with surrounding whitespace
+
+    // Put strings back
+    /* eslint-disable no-control-regex */
+    src = src.replace(/\x01/g, () => {
```
Poltergeist: Why do we need this? I mean why do we enable rules if we disable them somewhere in the code?

plugins/services/src/js/components/dsl/ServiceOtherDSLSection.js
```diff
@@ -0,0 +1,80 @@
...
+import DSLCombinerTypes from '../../../../../../src/js/constants/DSLCombinerTypes';
+import DSLExpression from '../../../../../../src/js/structs/DSLExpression';
+import DSLExpressionPart from '../../../../../../src/js/structs/DSLExpressionPart';
+import DSLFormWithExpressionUpdates from '../../../../../../src/js/components/DSLFormWithExpressionUpdates';
+import DSLUtil from '../../../../../../src/js/utils/DSLUtil';
+import FieldInput from '../../../../../../src/js/components/form/FieldInput';
+import FieldLabel from '../../../../../../src/js/components/form/FieldLabel';
+import FormGroup from '../../../../../../src/js/components/form/FormGroup';
+
+const PARTS = {
```
Poltergeist: This is probably a leftover. Please rename PARTS to something more catchy.

src/js/utils/DSLUpdateUtil.js
wavesoft: > Control characters are special, invisible characters in the ASCII range 0-31. These characters are rarely used in JavaScript strings so a regular expression containing these characters is most likely a mistake.

Not in this particular case. Non-visible character `\x01` is used as a special placeholder for extracted string expressions.

plugins/services/src/js/components/dsl/ServiceOtherDSLSection.js
wavesoft: Whoops, indeed. Fixing

wavesoft:
retest this please

PR #1566, opened 2016-12-09
DCOS-10693: Prevent stop action for scheduler tasks

MatApple:
Scheduler tasks don't like to be stopped so we disable the stop button if one is selected.

~~@leemunroe - should we hide this button or disable it? If we disable it, I presume we'd also want a tooltip to explain why.~~

**UPDATE**: The stop button is now disabled with a tooltip.

https://mesosphere.atlassian.net/browse/DCOS-10693

### DIFF
--- a/plugins/services/src/js/containers/tasks/TasksView.js
+++ b/plugins/services/src/js/containers/tasks/TasksView.js
@@ -153,6 +153,20 @@ class TasksView extends mixin(SaveStateMixin) {
     const hasSchedulerTask = tasks
       .some((task) => task.id in checkedItems && task.schedulerTask);
 
+    // Using Button's native "disabled" prop prevents onMouseLeave from
+    // being correctly called, preventing correct dismissal of the Tooltip.
+    //
+    // So we manually disable the button.
+    let onStopClick = () => {};
+
+    if (!hasSchedulerTask) {
+      onStopClick = this.handleKillClick.bind(this, 'stop');
+    }
+
+    let stopButtonClasses = classNames('button button-link', {
+      disabled: hasSchedulerTask
+    });
+
     return (
       <div className="button-collection flush-bottom">
         <button
@@ -161,17 +175,17 @@ class TasksView extends mixin(SaveStateMixin) {
           <Icon id="repeat" size="mini" />
           <span>Restart</span>
         </button>
-        <button
-          className="button button-link"
-          disabled={hasSchedulerTask}
-          onClick={this.handleKillClick.bind(this, 'stop')}>
-          <Tooltip
-            content="Stopping a scheduler task is not supported."
-            suppress={!hasSchedulerTask}>
-            <Icon id="circle-close" size="mini" />
-            <span>Stop</span>
-          </Tooltip>
-        </button>
+        <Tooltip
+          wrapperClassName="button-group"
+          content="Stopping a scheduler task is not supported."
+          suppress={!hasSchedulerTask}>
+          <button
+            className={stopButtonClasses}
+            onClick={onStopClick}>
+              <Icon id="circle-close" size="mini" />
+              <span>Stop</span>
+          </button>
+        </Tooltip>
       </div>
     );
   }
--- a/plugins/services/src/js/containers/tasks/TasksView.js
+++ b/plugins/services/src/js/containers/tasks/TasksView.js
@@ -67,7 +67,7 @@ class TasksView extends mixin(SaveStateMixin) {
     this.setState({checkedItems});
   }
 
-  handleKillClick(killAction) {
+  handleStopClick(killAction) {
     const {checkedItems} = this.state;
 
     if (!Object.keys(checkedItems).length) {
@@ -142,7 +142,7 @@ class TasksView extends mixin(SaveStateMixin) {
     );
   }
 
-  getKillButtons() {
+  getStopButtons() {
     const {tasks} = this.props;
     const {checkedItems} = this.state;
 
@@ -157,10 +157,10 @@ class TasksView extends mixin(SaveStateMixin) {
     // being correctly called, preventing correct dismissal of the Tooltip.
     //
     // So we manually disable the button.
-    let onStopClick = () => {};
+    let handleStopClick = function () {};
 
     if (!hasSchedulerTask) {
-      onStopClick = this.handleKillClick.bind(this, 'stop');
+      handleStopClick = this.handleStopClick.bind(this, 'stop');
     }
 
     let stopButtonClasses = classNames('button button-link', {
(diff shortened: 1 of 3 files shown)


leemunroe:
@MatApple Good point. I think disabled with a tooltip is a better UX in this case. `Stopping a scheduler task is not supported.`

MatApple:
Thanks @leemunroe - will make the appropriate changes.

MatApple:
@jfurrow - I'm having trouble getting the tooltip to work with the "Stop" button. I've tried wrapping the button with the Tooltip and also having the Tooltip as the immediate child of the button. Either way, the button doesn't like it and the Tooltip doesn't work correctly - especially with the "disabled" attribute on the button. Any suggestions? Thanks

MatApple:

plugins/services/src/js/containers/tasks/TasksView.js
```diff
@@ -141,25 +142,50 @@ class TasksView extends mixin(SaveStateMixin) {
     );
   }
 
-  getKillButtons(hasCheckedTasks) {
-    if (!hasCheckedTasks) {
+  getKillButtons() {
```
orlandohohmeier: Let's rename this as well as the `handleKillClick` methods to match the new wording...

plugins/services/src/js/containers/tasks/TasksView.js
```diff
@@ -141,25 +142,50 @@ class TasksView extends mixin(SaveStateMixin) {
...
     }
+    // Only show Stop if a scheduler task isn't selected
+    const hasSchedulerTask = tasks
+      .some((task) => task.id in checkedItems && task.schedulerTask);
+
+    // Using Button's native "disabled" prop prevents onMouseLeave from
+    // being correctly called, preventing correct dismissal of the Tooltip.
+    //
+    // So we manually disable the button.
+    let onStopClick = () => {};
```
orlandohohmeier: I'd use a good old `function` here, as the arrow functiion looks a little funky...  #nit

src/js/stores/MesosStateStore.js
```diff
@@ -179,10 +187,21 @@ class MesosStateStore extends GetSetBaseStore {
     let services = this.get('lastMesosState').frameworks || [];
     let memberTasks = {};
 
-    services.forEach(function (service) {
+    const schedulerTaskIds = services.reduce((memo, service) => {
+      const serviceName = service.name.split('.').reverse().join('/');
```
orlandohohmeier: This part looks familiar... I'm pretty sure we have a util to get the name

src/js/stores/MesosStateStore.js
```diff
@@ -179,10 +187,21 @@ class MesosStateStore extends GetSetBaseStore {
...
+      const schedulerTask = this.getSchedulerTaskFromServiceName(serviceName);
+
+      if (schedulerTask) {
```
orlandohohmeier: Let's also check `id` if there is a chance that it's undefined.

src/js/stores/MesosStateStore.js
```diff
@@ -179,10 +187,21 @@ class MesosStateStore extends GetSetBaseStore {
...
+      }
+
+      return memo;
+    }, []);
+
+    services.forEach((service) => {
       service.tasks.forEach(function (task) {
         if (task.slave_id === nodeID) {
-          memberTasks[task.id] = task;
+          memberTasks[task.id] = isSchedulerTask(task, schedulerTaskIds);
```
orlandohohmeier: wonder if there is way to flag the scheduler tasks in one go instead of iterating same list over and over again

src/js/stores/MesosStateStore.js
MatApple: Nope - if we weren't going to deprecate this code, I'd put it in a util.

src/js/stores/MesosStateStore.js
MatApple: I don't think it's possible to have a task without an `id`.

src/js/stores/MesosStateStore.js
MatApple: We are already iterating that same list every time we fetch tasks by nodeId :) All I do is add a check to see if the task is a scheduler task. I think it's out of scope for this PR to refactor the way this store works. These issues will be solved when this logic moves to GraphQL.

plugins/services/src/js/containers/tasks/TasksView.js
MatApple: done!

src/js/stores/MesosStateStore.js
```diff
@@ -34,6 +34,14 @@ function stopPolling() {
   }
 }
 
+function isSchedulerTask(task, schedulerTaskIds) {
```
philipnrmn: Can we make this method name a bit cleaner? `isSchedulerTask` sounds to me like it should return a boolean, not mutate an arg.

leemunroe:
@MatApple Functionally looks good but why does the disabled state have a grey background and hover state? It makes it stand out more 😪

We should remove the background if disabled. And no hover state needed (i.e. no underline).

src/js/stores/MesosStateStore.js
orlandohohmeier: Yeah, this store desperately needs a refactoring and I can't wait to 🔥  it once we have GraphQL in place. However, I wasn't referring to the general problems but to the helper and methods you've introduced.
You added functions to first create a list of scheduler task ids by iterating over the complete list of services. Then we iterate over the same list of service where you add the call to flag tasks as scheduler tasks. By using an array instead of a Map to store the the scheduler ids add even more unnecessary calls.
Please combine the `reduce` and `forech` and change the data structure (do we even need to store the ids?), to improve the performance.

MatApple:
@leemunroe - completely agree, the button is weird. Talked to @ashenden about this. Instead of adding custom CSS, we want to handle this by updating the button styles for the disabled state in CNVS.

src/js/stores/MesosStateStore.js
MatApple: We can't combine the `reduce` and the `forEach` in a way that would make sense. All the scheduler tasks belong to `marathon`. So first we need to go through all the frameworks, and identify their scheduler tasks. Then we need to find those tasks within marathon.

orlandohohmeier: Ok, now I get it. Thanks for the explanation! Please add a comment explaining why things need to happen in that order and consider changing the data structure as using an array for the IDs is still suboptimal. 
Once we've merged this PR I'd like to have a close look at our `getSchedulerTaskFromServiceName` method, even though that we hopefully burn this store soon, as it is also iterating over the complete list of frameworks and the whole constructs feels a lot like a hack on top of a hack, on top of a hack...

orlandohohmeier: Ok, I went ahead and fixed most the of performance issues in #1630 thus we should be ok to merge this branch as is.

PR #1567, opened 2016-12-09
[1 of 3]: Create FluidGeminiContainer component

jfurrow:
This PR introduces a wrapper component for Gemini which essentially does one thing: creates CSS to override Gemini's inline `width` property. It also renames the `GeminiUtil` to `ScrollbarUtil` and adds a method to get the width of the browser's scrollbar.

This is necessary when the width of the container of a Gemini element is animated. A good example of is the `Page`'s width, when you toggle between a docked and undocked sidebar.

Without this change, the width of Gemini's view container is set to a value in pixels, and while the container's width animates, the native scrollbar is visible. With this change, the width of Gemini's view container will always match its container's width, plus an offset equal to the width of the browser's scrollbar.

If you'd like to test this, replace `GeminiScrollbar` with `FluidGeminiScrollbar` in `Page.js` and toggle the sidebar's docked state. You'll see the width of `Page`'s children animate along with the Sidebar.


rcorral:
Pushed a few changes @jfurrow

jfurrow:
Thanks @rcorral! I just pushed another small change (I didn't set the `. styleSheet.cssText` property correctly, though it seems to work either way).

orlandohohmeier:
I'm ok with this component but wonder if it is actually worth all the effort, simply to have a "better" looking scrollbar. Every once in a while, we have serious issues with the `GeminiScrollbar` like the form state loss, unwanted updates, wrong dimensions, and visual issues like the one you're fixing. 

Let's discuss this with design team before we continue to fix the `GeminiScrollbar` as I don't want us to waste time with something that cost so much but brings so little. It's tiny and decreasing number of users that is still doomed to see ugly scrollbars.

/cc @dcos/design @dcos/frontend

jfurrow:
@orlandohohmeier This is a good point. Right now I am aware of at least one fatal bug with the scrollbar on Ubuntu (both FF and Chrome) that still needs to be fixed.

PR #1569, opened 2016-12-09
DCOS-10878: Update CLI Install Instructions

MatApple:
Splice `dcos auth login` command step into the CLI Install instructions when authentication is enabled.

https://mesosphere.atlassian.net/browse/DCOS-10878

### DIFF
--- a/plugins/oauth/hooks.js
+++ b/plugins/oauth/hooks.js
@@ -65,14 +65,15 @@ module.exports = Object.assign({}, StoreMixin, {
   },
 
   dcosInstallCommand(instructions) {
-    let steps = instructions.split('&&');
-    steps.splice(-1, 0, ' \n dcos auth login ');
+    let steps = [...instructions];
 
-    return steps.join('&&');
+    steps.splice(-1, 0, 'dcos auth login');
+
+    return steps;
   },
 
   dcosInstallSteps(steps) {
-    let newSteps = steps.map((step) => step);
+    let newSteps = [...steps];
 
     newSteps.splice(-1, 0, 'dcos auth login');
 
--- a/plugins/oauth/hooks.js
+++ b/plugins/oauth/hooks.js
@@ -44,8 +44,7 @@ module.exports = Object.assign({}, StoreMixin, {
   filters: [
     'applicationRoutes',
     'delayApplicationLoad',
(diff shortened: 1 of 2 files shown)


plugins/oauth/hooks.js
```diff
@@ -62,6 +64,21 @@ module.exports = Object.assign({}, StoreMixin, {
...
+    let steps = instructions.split('&&');
+    steps.splice(-1, 0, ' \n dcos auth login ');
+
+    return steps.join('&&');
+  },
+
+  dcosInstallSteps(steps) {
+    let newSteps = steps.map((step) => step);
+
+    newSteps.splice(-1, 0, 'dcos auth login');
```
mlunoe: Could we get a comment here? something like: "Add login as the step second to last, before printing command options"

plugins/oauth/hooks.js
```diff
@@ -62,6 +64,21 @@ module.exports = Object.assign({}, StoreMixin, {
     }]);
   },
 
+  dcosInstallCommand(instructions) {
+    let steps = instructions.split('&&');
+    steps.splice(-1, 0, ' \n dcos auth login ');
```
mlunoe: Could we get a comment here? something like: "Add login as the step second to last, before printing command options"

orlandohohmeier: This seems highly error prone to me and I suggest to refactor this using the `mount` package.

MatApple: Made some changes so we don't split a string and pass in an Array. I'm not sure a Mount would be more beneficial without code duplication. Do you have a particular example in mind @orlandohohmeier?

orlandohohmeier: I'd much rather duplicate code/data than having a hard to maintain and error prone implementation. DRY isn't absolute.

mlunoe:
I would still like some comments

mlunoe:
Run unit tests

mlunoe:
Feel free to :shipit: once tests pass

rcorral:
run unit tests

rcorral:
run integration tests

rcorral:
legit unit test errors.

nLight:
Fixed the linting errors, merging it.

PR #1573, opened 2016-12-12
[Multi-container] endpoint name

Poltergeist:
As it turned out the endpoint name is not optional, this pr adds the field back for the hose mode.

image

https://www.youtube.com/watch?v=bupagiROLV8

### DIFF
--- a/plugins/services/src/js/reducers/serviceForm/Containers.js
+++ b/plugins/services/src/js/reducers/serviceForm/Containers.js
@@ -192,14 +192,29 @@ function containersParser(state) {
 
     if (item.endpoints != null && item.endpoints.length !== 0) {
       item.endpoints.forEach((endpoint, endpointIndex) => {
-        memo.push(
-            new Transaction(['containers', index, 'endpoints'], endpointIndex, ADD_ITEM)
-        );
-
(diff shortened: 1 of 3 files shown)


nLight:
This could be a separate issue, but when I try to change hostPort from JSON it always sets `automatically assign` checkbox no matter what value I have in JSON

Poltergeist:
@nLight Added parsers for automaticPort

nLight:
Awesome work! 💯 
I wonder how many of optional yet required fields we still have 😅

plugins/services/src/js/reducers/serviceForm/Containers.js
```diff
@@ -195,10 +196,13 @@ function containersParser(state) {
             new Transaction(['containers', index, 'endpoints'], endpointIndex, ADD_ITEM)
         );
 
+        memo.push(new Transaction(['containers', index, 'endpoints', endpointIndex, 'hostPort'], endpoint.hostPort));
```
mlunoe: Could we use concat here?

mlunoe: Done

PR #1576, opened 2016-12-13
DCOS-12045: New funnel graphic for declined offers summary

jfurrow:
This PR implements the new design for the declined offers summary:

### DIFF
--- a/plugins/services/src/js/components/RecentOffersSummary.js
+++ b/plugins/services/src/js/components/RecentOffersSummary.js
@@ -15,7 +15,7 @@ const displayedResourceValues = {
   ports: 'Ports'
 };
 
-const maxBarHeight = 200;
+const MAX_BAR_HEIGHT = 200;
 
 const getGraphBar = (resource, data, index) => {
   const resourceOfferSummary = data[resource];
@@ -45,7 +45,7 @@ const getGraphBar = (resource, data, index) => {
     }
   );
 
-  const offeredHeight = Math.ceil(maxBarHeight * percentageOffered);
+  const offeredHeight = Math.ceil(MAX_BAR_HEIGHT * percentageOffered);
   const matchedHeight = Math.ceil(offeredHeight * percentageMatched);
 
   return (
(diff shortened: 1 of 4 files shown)


plugins/services/src/js/components/RecentOffersSummary.js
```diff
@@ -10,170 +10,166 @@ const displayedResourceValues = {
...
   ports: 'Ports'
 };
 
-const getColumnClassNameFn = (classes) => {
-  return (prop, sortBy) => {
-    return classNames(classes, {
-      'active': prop === sortBy.prop
-    });
-  };
+const maxBarHeight = 200;
```
nLight: I know we haven't discussed that but I try to go `MAX_BAR_HEIGHT` for constant top level constants.

nLight:
Is it a correct behaviour? We're requesting any role but display `role: slave_public` as rejected.

screen shot 2016-12-14 at 19 45 45

jfurrow:
@nLight I believe this is correct, given the data we have from the API. The logic for that is as follows:
* if the received offer does not match the request (ie. requested 6 cpus, received 4), then don't display the role data
* if the received offer matches or exceeds the resource request (ie. requested 4 cpus, received 4), but the API tells us that this resource is a reason for rejection, then display the role information in the tooltip

In this case, it does look like the API shouldn't have returned `InsufficientCpus` as a reason for this offer being rejected, but it did, so we have to display it as such.

orlandohohmeier:
@jfurrow @nLight we don't have role "globbing" meaning it's always an exact match, therefore `slave_public` is reject if you request `*`... http://mesos.apache.org/documentation/latest/roles/

plugins/services/src/js/components/RecentOffersSummary.js
orlandohohmeier: 💯 let's always capitalize top level constants.

jfurrow: This makes sense! Done!

jfurrow:
Thanks for the clarification @orlandohohmeier!

jfurrow:
retest this please

jfurrow:
retest this please

PR #1578, opened 2016-12-13
Display not supported not configured for empty fields, ref. DCOS-11676

mlunoe:
This PR fixes a few small issues with the config review and displays 'Not Configured'/'Not Supported' for empty values, rather than `--`.

Before:

After:

### DIFF
--- a/plugins/services/src/js/utils/ServiceConfigDisplayUtil.js
+++ b/plugins/services/src/js/utils/ServiceConfigDisplayUtil.js
@@ -42,14 +42,16 @@ const ServiceConfigDisplayUtil = {
     );
   },
 
-  getDisplayValue(value) {
-    // Return the emdash character.
-    if (value == null || value === '') {
-      return String.fromCharCode(8212);
+  getDisplayValue(value, isDisabled = false) {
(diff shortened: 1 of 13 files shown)


plugins/services/src/js/utils/ServiceConfigDisplayUtil.js
```diff
@@ -42,14 +42,16 @@ const ServiceConfigDisplayUtil = {
     );
   },
 
-  getDisplayValue(value) {
-    // Return the emdash character.
-    if (value == null || value === '') {
-      return String.fromCharCode(8212);
+  getDisplayValue(value, isDisabled = false) {
+    if (!isDisabled && (value == null || value === '')) {
+      return <i>Not Configured</i>;
```
jfurrow: I think these should be `em` instead of `i`: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/em

jfurrow:
@mlunoe I pushed a small change, hope that's okay!

mlunoe:
Do I have your label, or do you want other changes?

jfurrow:
@mlunoe You've got my label!

rcorral:
Looks great!

PR #1579, opened 2016-12-13
DCOS-12036: De-bounce DSL input

wavesoft:
This PR de-bounces the user typing on the input field for 500ms before applying the filter, optimising the update performance.

### DIFF
--- a/src/js/components/DSLInputField.js
+++ b/src/js/components/DSLInputField.js
@@ -3,8 +3,9 @@ import React, {PropTypes} from 'react';
 
 import Icon from './Icon';
 import DSLExpression from '../structs/DSLExpression';
+import Util from '../utils/Util';
 
-const DEBOUNCE_TIMEOUT = 500;
+const DEBOUNCE_TIMEOUT = 250;
 
 const METHODS_TO_BIND = [
   'handleBlur',
@@ -47,6 +48,10 @@ class DSLInputField extends React.Component {
     METHODS_TO_BIND.forEach((method) => {
       this[method] = this[method].bind(this);
     });
+
+    this.handleDebounceUpdate = Util.debounce(
+      this.handleDebounceUpdate, DEBOUNCE_TIMEOUT
+    );
   }
 
   /**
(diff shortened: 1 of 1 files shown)


src/js/components/DSLInputField.js
```diff
@@ -72,7 +91,21 @@ class DSLInputField extends React.Component {
...
+    clearTimeout(this.debounceTimer);
+    this.debounceTimer = setTimeout(
+      this.handleDebounceUpdate, DEBOUNCE_TIMEOUT
+    );
+  }
+
+  /**
+   * Forward property change update after the debounce timer expired
+   */
+  handleDebounceUpdate() {
```
nLight: Is it make sense to have a `debounce` utility function and hide all the timers inside?

```Javascript
handleDebounceUpdate() {
  Util.debounce(
    () => this.props.onChange(this.state.expression),
    DEBOUNCE_TIMEOUT
  );
}
```

wavesoft: I had the same thinking. Having a unique ID to somehow identify your de-bounce timer ID. That would actually be perfect. But this introduces a new concept that I would like to run through the team first.

nLight: We could use lodash.debounce :trollface:

nLight: how convenient that we have it already! 👍 I gotta peak into Utils more often

src/js/components/DSLInputField.js
```diff
@@ -72,7 +96,18 @@ class DSLInputField extends React.Component {
    * @param {SyntheticEvent} event - The change event
    */
   handleChange({target}) {
-    this.props.onChange(new DSLExpression(target.value));
+    this.setState({
+      expression: new DSLExpression(target.value)
```
mlunoe: Why are we storing this in state as well? Couldn't we create call a callback to pass it to the parent component that holds the expression?
This way we don't have two sources of truth and are sure for it to not get out of sync

wavesoft: We always have one source of truth. The one that comes from the parent of the `DSLFilterField` component, and the one updated through the `onChange` callback. That's why the `componentWillReceiveProps` function always prefers props against state.

The problem is that if we want the UI to be editable until the time the timeout kicks-in, we must keep the current state of the expression somewhere. Keeping the `expression` instance instead of just the text value also allows us to update the error state of the field more interactively than waiting for the timeout.

PR #1581, opened 2016-12-13
DCOS-11844: Fixing resources in pod containers

wavesoft:
This PR fixes the resource display in the pod container review section (and it also does some minor clean-ups in other places too)

image

### DIFF
--- a/plugins/services/src/js/service-configuration/PodContainerConfigSection.js
+++ b/plugins/services/src/js/service-configuration/PodContainerConfigSection.js
@@ -28,7 +28,7 @@ function getCommand(containerConfig) {
 module.exports = ({containerConfig, appConfig}) => {
   let fields = {
     command: getCommand(containerConfig),
-    resources: appConfig.resources || {},
+    resources: containerConfig.resources || {},
     user: containerConfig.user || appConfig.user
   };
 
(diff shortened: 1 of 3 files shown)


plugins/services/src/js/service-configuration/PodContainerConfigSection.js
```diff
@@ -54,39 +54,39 @@ module.exports = ({containerConfig, appConfig}) => {
       </Row>
 
       {/* Resources */}
-      {fields.resources.cpus && (
+      {!!fields.resources.cpus && (
```
orlandohohmeier: I'd vote for verbose conditions instead of the `!!value` notation to improve readability/maintainability. Let's change this condition to `(fields.resources.cpus != null && fields.resources.cpus >  0)`.

plugins/services/src/js/service-configuration/PodContainerConfigSection.js
```diff
@@ -54,39 +54,39 @@ module.exports = ({containerConfig, appConfig}) => {
       </Row>
 
       {/* Resources */}
-      {fields.resources.cpus && (
+      {(fields.resources.cpus != null && fields.resources.cpus > 0) && (
```
mlunoe: I'm really not a fan of this merging logic and view declaration :/
Not a show stopper here, but let's have a conversation about it

orlandohohmeier:
@wavesoft please rebase this branch

plugins/services/src/js/service-configuration/PodContainerConfigSection.js
wavesoft: Hey @mlunoe , this actually follows the [react best practices for in-line conditional rendering](https://facebook.github.io/react/docs/conditional-rendering.html#inline-if-with-logical-ampamp-operator). And I wouldn't say that's really "logic" there. This is more like conditional rendering.

A more complicate logic would definitely need a separate function, or even perhaps a local component.

What do you think /cc @orlandohohmeier ?

wavesoft:
Rebased

PR #1582, opened 2016-12-13
DCOS-12147: Allow ternary operator

nLight:
This PR enables 2 eslint rules to tame ternary:
`no-nested-ternary` - ensures that we don't nest them
`no-unneeded-ternary` - ensures that we don't do `value ? true : false` stuff
`multiline-ternary` is off to explicitly show that we don't yet have a decision on whether we always break a ternary into multiple lines them or we always use single line. So both are allowed. http://eslint.org/docs/rules/multiline-ternary


orlandohohmeier:
💯 ✨ Thanks for enabling the linting rules

PR #1583, opened 2016-12-13
DCOS-12149: Remove underscore dependency

nLight:
Looks like we don't use it


nLight:
@mlunoe yep, so we still depend on underscore while we not depending on underscore :)

mlunoe:
Exactly! Great stuff `npm`... great stuff...

PR #1584, opened 2016-12-13
[multi container] endpoint tests

Poltergeist:
Based on #1573

This Introduces the endpoints tests for `JSONReducers` and `FormReducers`.

And also the network reducer got some tests.

On the way I fixed some issues which would never occur except in test cases.

For example set the path for the initial transaction in the batch.

### DIFF
--- a/plugins/services/src/js/reducers/serviceForm/__tests__/MultiContainerNetwork-test.js
+++ b/plugins/services/src/js/reducers/serviceForm/__tests__/MultiContainerNetwork-test.js
@@ -1,0 +1,42 @@
+const MultiContainerNetwork = require('../MultiContainerNetwork');
+const Networking = require('../../../../../../../src/js/constants/Networking');
+const Batch = require('../../../../../../../src/js/structs/Batch');
+const Transaction = require('../../../../../../../src/js/structs/Transaction');
+
+describe('MultiContainerNetwork', function () {
+  describe('#JSONReducer', function () {
+    it('should be host default type', function () {
+      let batch = new Batch();
+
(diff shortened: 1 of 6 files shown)


wavesoft:
Looks good! 👍 :shipit:

plugins/services/src/js/reducers/serviceForm/__tests__/MultiContainerNetwork-test.js
```diff
@@ -0,0 +1,42 @@
...
+    it('should return a host object', function () {
+      let batch = new Batch();
+
+      batch = batch.add(new Transaction(['network'], Networking.type.HOST));
+
+      expect(batch.reduce(MultiContainerNetwork.JSONReducer.bind({}), {}))
+        .toEqual({mode: 'host'});
+    });
+
+    it('should return a container object', function () {
```
orlandohohmeier: Please update the test description to better describe the expected output

plugins/services/src/js/reducers/serviceForm/__tests__/MultiContainerNetwork-test.js
```diff
@@ -0,0 +1,42 @@
...
+      let batch = new Batch();
+
+      batch = batch.add(new Transaction(['network'], `${Networking.type.CONTAINER}.foo`));
+
+      expect(batch.reduce(MultiContainerNetwork.JSONReducer.bind({}), {}))
+      .toEqual({mode: 'container', name: 'foo'});
+    });
+
+    it('should return a container object', function () {
```
orlandohohmeier: Please update this test description as well to better describe the expected output and make sure it's "unique" to make it easier to fine the respective test in case of failure.

PR #1586, opened 2016-12-13
DCOS-12018: Maintain label copy for enabling/disabling LB service addresses and port mapping

jfurrow:


jfurrow:
@mlunoe Oops, thanks for that!

mlunoe:
Fixed it for ya 😉

PR #1590, opened 2016-12-13
Adding filter for mesos state error in Services

MatApple:

### DIFF
--- a/plugins/services/src/js/components/DeclinedOffersTable.js
+++ b/plugins/services/src/js/components/DeclinedOffersTable.js
@@ -1,5 +1,6 @@
 import classNames from 'classnames';
 import {Link} from 'react-router';
+import {MountService} from 'foundation-ui';
 import React from 'react';
 import {Table, Tooltip} from 'reactjs-components';
 
@@ -292,11 +293,13 @@ class RejectedOffersTable extends React.Component {
     const {offers} = this.props;
 
     return (
-      <Table className="table table-simple table-header-nowrap table-break-word flush-bottom"
-        colGroup={this.getColGroup()}
-        columns={this.getColumns()}
-        data={offers}
-        sortBy={{prop: 'timestamp', order: 'desc'}} />
+      <MountService.Mount type="DeclinedOffersTable">
+        <Table className="table table-simple table-header-nowrap table-break-word flush-bottom"
+          colGroup={this.getColGroup()}
+          columns={this.getColumns()}
+          data={offers}
+          sortBy={{prop: 'timestamp', order: 'desc'}} />
+      </MountService.Mount>
     );
   }
 }
--- a/plugins/services/src/js/containers/tasks/ServiceTasksContainer.js
+++ b/plugins/services/src/js/containers/tasks/ServiceTasksContainer.js
@@ -1,5 +1,5 @@
-import {Hooks} from 'PluginSDK';
 import mixin from 'reactjs-mixin';
+import {MountService} from 'foundation-ui';
 import React from 'react';
 import {StoreMixin} from 'mesosphere-shared-reactjs';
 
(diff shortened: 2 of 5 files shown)


rcorral:
retest this please

rcorral:
retest this please

rcorral:
retest this please

orlandohohmeier:
Let's use the `mount` package instead of filters to allow plugins to hook into the service error view.

plugins/services/src/js/containers/tasks/ServiceTasksContainer.js
```diff
@@ -38,19 +40,33 @@ class ServiceTasksContainer extends mixin(StoreMixin) {
...
+        errorMessage: null,
         lastUpdate: Date.now(),
         mesosStateErrorCount: 0
       });
     }
   }
 
-  onStateStoreError() {
+  onStateStoreError(xhr) {
+    const errorMessage = Hooks.applyFilter(
```
orlandohohmeier: Let's use the mount package instead of filters to allow plugins to hook into the service error view.

orlandohohmeier:

plugins/services/src/js/components/DeclinedOffersTable.js
```diff
@@ -292,11 +293,13 @@ class RejectedOffersTable extends React.Component {
     const {offers} = this.props;
 
     return (
-      <Table className="table table-simple table-header-nowrap table-break-word flush-bottom"
-        colGroup={this.getColGroup()}
-        columns={this.getColumns()}
-        data={offers}
-        sortBy={{prop: 'timestamp', order: 'desc'}} />
+      <MountService.Mount type="DeclinedOffersTable">
```
orlandohohmeier: Wonder if it makes sense to choose a less specific *types* (think CSS class), as the intention seems to mount the same type of component here as well as in the service page in case of errors. It might also worth thinking about ways to add a general page level error mount.

MatApple:
@orlandohohmeier I like the suggestion of adding an Array variant for types on a Mount. However, I'd prefer it doesn't hold up this PR because I can see some discussion required. Ordering and prioritization of components being one point of debate. There are several ways we could do this. Do we follow similar rules to how CSS is applied, do we have something similar to !important or do we create another prop with rules that can be customized, like block all 'MarathonErrors' together and block all 'MesosErrors' together. I also like the dynamic namespaces example, lets think about how we can apply this throughout our pages/containers.

I've opened an issue and assigned it to @orlandohohmeier for these improvements. It also includes updating these Mount types once the functionality is in place. https://mesosphere.atlassian.net/browse/DCOS-12315

orlandohohmeier:
@MatApple I agree the potential `Mount` refactoring should by no means block this PR. I simply wanted to post my thoughts in this context as this PR spawned the discussion and I thinks it's valuable to have a proper context when discussion such refactorings.

MatApple:
Sounds good @orlandohohmeier - think I addressed all your comments so perhaps you could dismiss your requested changes?

src/js/events/MesosStateActions.js
```diff
@@ -21,10 +21,11 @@ var MesosStateActions = {
             });
             resolve();
           },
-          error(e) {
+          error(xhr) {
             AppDispatcher.handleServerAction({
               type: ActionTypes.REQUEST_MESOS_STATE_ERROR,
-              data: e.message
+              data: xhr.message,
```
mlunoe: Shouldn't we parse this through `RequestUtil.getErrorFromXHR(xhr)` like we do other places?

mlunoe: Sounds good man

plugins/services/src/js/components/DeclinedOffersTable.js
rcorral: I think this would make sense in this case, but usually I would prefer for more IDs over a generic one as in the future refactoring involves refactoring everywhere in the codebase, rather than just the packages that implement the mounts.

PR #1592, opened 2016-12-13
DCOS-12205: Use defs instead of symbol in SVG sprite

jfurrow:
This PR uses `<defs>` instead of `<symbol>` for the SVG sprite. This fixes SVGs that use gradients in Firefox and doesn't break them in other browsers (even IE11 and Edge)!


rcorral:
run integration

wavesoft:
Looks good in firefox, safari, chrome. Great!! ✨

PR #1593, opened 2016-12-13
DCOS-12130: Use inline-block for dropdown-toggle elements

jfurrow:
Firefox doesn't treat pseudo-elements as regular flex elements, so I've decided to change `.dropdown-toggle` elements to `inline-block` rather than `inline-flex`. Here's a codepen to illustrate the issue: http://codepen.io/yeahjohnnn/pen/ObBOBG


rcorral:
yeahjohnnn!

wavesoft:
Greeeat!! ✨

PR #1596, opened 2016-12-14
DCOS-12145: Fix force edit for service

MatApple:
Closes DCOS-12145


wavesoft:
This is great! ✨  💯 

However I have a tiny suggestion to add here: It feels that this error message is regarding something wrong on the user input, which is not the case. So I would recommend to:

* Either use a warning-styled alert message
* Or change change the button name from "Run service" to something like "Run service (Forced)"

What do you think?

MatApple:
@wavesoft - thanks for the suggestion. I pushed a change and updated the GIF in the description to illustrate how it works now.

PR #1600, opened 2016-12-14
DCOS-12206: Exposing some hooks needed by the plugin

wavesoft:
This PR exposes some hooks and modules needed by the DCOS-12206 plugin


plugins/services/src/js/pages/task-details/ServiceTaskDetailPage.js
```diff
@@ -16,11 +17,11 @@ class ServiceTaskDetailPage extends React.Component {
     const {id, taskID} = params;
 
     let routePrefix = `/services/overview/${encodeURIComponent(id)}/tasks/${encodeURIComponent(taskID)}`;
-    const tabs = [
+    const tabs = Hooks.applyFilter('serviceTaskDetailPageTabs', [
```
nLight: What I really would like to see here is `Mount`
This is not something you could do right now, but let's imagine:
```javascript
const tabs = [
  <TabLink label="Details" routePath={routePrefix + '/details'} />,
  <TabLink label="Files" routePath={routePrefix + '/files'} />,
  <TabLink label="Logs" routePath={routePrefix + '/logs'} />,
  <Mount type="ServiceTaskDetailPageTabs" />
];
```

I'm down to implement this with backwards compatibility if @orlandohohmeier gives a 💚

orlandohohmeier: @nLight using `Mount` instead of hooks sounds great, but I'm not sure about your example. Shouldn't we use the same approach as we used for the config views?

nLight: The thing is that it goes like 2-3 levels deep so how'd you define a mounting point there? We would have to override all the nesting chain. Or am I missing something? Here I'd like to pass a ready to use set of elements instead of just configuration.

MatApple: +1 for using the NavigationService

mlunoe:
@wavesoft could we get an update on this PR, otherwise close it?

mlunoe:
@orlandohohmeier do you know what is to happen here?
Do we resolve the conflicts?

orlandohohmeier:
Let's close this one as we don't need the hooks anymore...

PR #1601, opened 2016-12-14
Fix Marathon residency validator

orlandohohmeier:
Adjust the residency rule to properly look for persistent volumes.

PR #1602, opened 2016-12-14
[local volumes] add residency if needed

Poltergeist:
If one adds a local persistent volume also the residency has to be set. As we do not have any field in the form representing this. We only have a `JSONReducer` and `JSONParser` for this.

residency mov

The erros which you see in the animation is addressed by: #1601


wavesoft:
Tested against #1601 and works perfectly! ✨  💯

nLight:
JSON - to - From reducer (I still can't remember names) switches the Form to `UCR` from `Mesos Runtime`, which is the same thing yet can be confusing for a user. Just hit space bar anywhere in JSON editor to see what I mean

dec-14-2016 14-44-06

orlandohohmeier:
@nLight I fear there is no simple way of fixing that issue, but I have a few ideas...  anyway, let's not block this PRs as it's not an issue related to the introduced changes. Please file an JIRA so that we don't forget to fix the parsing.

orlandohohmeier:
LGTM 💯 ✨

PR #1606, opened 2016-12-14
DCOS-11803: Improve consistency of service creation modal spacing

jfurrow:
This PR aims to improve the consistency of the spacing in the service creation modal. Originally I intended to create a `FormSection` component, but it is @ashenden's intentions for the margin on headings in the form to create the desired visual spacing.

I created a `FormRow` component which should be used to wrap `FormGroup` elements. Doing so will apply a negative bottom margin on the `FormRow` element to offset the margins applied by `FormGroup` children. This means that `FormRow` must always contain `FormGroup` elements.

The structure of all form sections should be as follows (of course with any number of `FormRow` or `FormGroup` elements):

```jsx
/* Notice the .flush-top on the first heading. */
<h2 className="short-bottom flush-top">Top-level Heading</h2>
<FormRow>
  <FormGroup />
</FormRow>
/* Subsequent headings require the margin from the heading's style. */
<h3 className="short-bottom">Sub-heading of first section</h3>
<FormRow>
  <FormGroup />
</FormRow>
<h2 className="short-bottom">Top-level Heading</h2>
<FormRow>
  <FormGroup />
</FormRow>
```

It might be worth introducing a `FormSectionHeading` component to apply the proper classes for us, but we would still need to provide some configuration options, so I'm not sure how useful it would be.

### DIFF
--- a/plugins/services/src/js/components/forms/ContainerServiceFormSection.js
+++ b/plugins/services/src/js/components/forms/ContainerServiceFormSection.js
@@ -89,11 +89,7 @@ class ContainerServiceFormSection extends Component {
       );
     }
 
-    return (
-      <div className="artifacts-section">
-        {content}
-      </div>
-    );
+    return content;
   }
 
   getGPUSInput(data) {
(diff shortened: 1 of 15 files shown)


Poltergeist:
nLight is speaking of my mind! Love the improvement.

PR #1608, opened 2016-12-15
DCOS-12274: Show rejected offers warning in service breadcrumb

jfurrow:
This PR displays the rejected offers warning icon & tooltip in the service breadcrumb and moves the corresponding code to a separate component.

### DIFF
--- a/plugins/services/src/js/components/ServiceBreadcrumbs.js
+++ b/plugins/services/src/js/components/ServiceBreadcrumbs.js
@@ -4,6 +4,7 @@ import {Link} from 'react-router';
 import {DCOSStore} from 'foundation-ui';
 import HealthBar from './HealthBar';
 import PageHeaderBreadcrumbs from '../../../../../src/js/components/NewPageHeaderBreadcrumbs';
+import ServiceStatusWarning from './ServiceStatusWarning';
 
 function getHealthStatus(serviceID) {
   if (serviceID == null) {
@@ -25,6 +26,7 @@ function getHealthStatus(serviceID) {
   return (
     <div className="service-page-header-status page-header-breadcrumb-content-secondary muted">
       {`${serviceStatus} (${runningTasksCount} of ${instancesCount})`}
+      <ServiceStatusWarning item={service} showDebugInstruction={true} />
       <HealthBar isDeploying={isDeploying}
         tasksSummary={tasksSummary}
         instancesCount={instancesCount}/>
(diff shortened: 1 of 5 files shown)


nLight:
retest this please

plugins/services/src/js/components/ServiceStatusWarning.js
```diff
@@ -0,0 +1,60 @@
...
+    queue = item.getQueue();
+  }
+
+  if (queue != null) {
+    const waitingSince = DateUtil.strToMs(queue.since);
+    const timeWaiting = Date.now() - waitingSince;
+
+    // If the item has been waiting for less than five minutes, we don't
+    // display the warning.
+    if (timeWaiting >= 1000 * 60 * 5) {
```
nLight: Let's extract a constant

plugins/services/src/js/components/ServiceBreadcrumbs.js
```diff
@@ -25,6 +26,7 @@ function getHealthStatus(serviceID) {
   return (
     <div className="service-page-header-status page-header-breadcrumb-content-secondary muted">
       {`${serviceStatus} (${runningTasksCount} of ${instancesCount})`}
+      <ServiceStatusWarning item={service} showDebugInstruction={true} />
```
nLight: I'm not sure I like an inversion of control here.
Do you think it's better to have two of them one with Debug instructions and one without?

MatApple:
LGTM after addressing @nLight's comments!

plugins/services/src/js/components/ServiceBreadcrumbs.js
jfurrow: I think the reason for adding the debug instruction in this case (when displayed in the breadcrumb) is because the debug tab is available to the user at this point, but would not be available when displayed in the services table.

@leemunroe What do you think?

plugins/services/src/js/components/ServiceStatusWarning.js
jfurrow: Done!

wavesoft:
Since I would like to use this functionality also on a new filter (`has:rejectedoffers`) does it make sense to move the check into a utility file? (In addition to the new component)? Otherwise I will do so in a future PR when I introduce the filter...

nLight:
run unit tests

nLight:
@Poltergeist :1st_place_medal:

Poltergeist:
I addressed the last comment from nLight. @jfurrow

nLight:
run unit tests

PR #1611, opened 2016-12-15
Adjust the pre-commit task order

orlandohohmeier:
Change the task order to first lint all files and then run the tests to catch stylistic issues faster, as the tests might take some time to run.

/cc @mlunoe


Poltergeist:
SGTM ! let's hit merge

PR #1524, opened 2016-12-07
Introduce multiple container form

orlandohohmeier:
This PR introduce a nice form for creating and editing  multiple containers (pods).

PR #1612, opened 2016-12-15
[Multi Container] stylistic adjustments

Poltergeist:
This pr contains changes which @mlunoe requested in another PR.

#1524

### DIFF
--- a/plugins/services/src/js/components/modals/NewCreateServiceModalForm.js
+++ b/plugins/services/src/js/components/modals/NewCreateServiceModalForm.js
@@ -93,11 +93,11 @@ class NewCreateServiceModalForm extends Component {
     let nextJSON = ServiceUtil.getServiceJSON(nextProps.service);
     let isPod = nextProps.service instanceof PodSpec;
 
-    // Note: We ignore changes that might derrive from the `onChange` event
-    //       handler. In that case the contents of nextJSON would be the same
-    //       as the contents of the last rendered appConfig in the state.
+    // Note: We ignore changes that might derive from the `onChange` event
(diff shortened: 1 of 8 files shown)


plugins/services/src/js/components/modals/NewCreateServiceModalForm.js
```diff
@@ -314,16 +323,18 @@ class NewCreateServiceModalForm extends Component {
     if (this.state.isPod) {
       return [
```
wavesoft: Since we are rerurning an array we need `key=` here

plugins/services/src/js/components/modals/NewCreateServiceModalForm.js
```diff
@@ -314,16 +323,18 @@ class NewCreateServiceModalForm extends Component {
...
-        <TabButton id="healthChecks" label="Health Checks" key="multihealthChecks"/>,
-        <TabButton id="volumes" label="Volumes" key="multivolumes"/>
+        <TabButton id="environment" label="Environment"
+          key="multienvironment" />,
+        <TabButton id="healthChecks" label="Health Checks"
+          key="multihealthChecks" />,
+        <TabButton id="volumes" label="Volumes" key="multivolumes" />
       ];
     }
     return [
```
wavesoft: The same here, we need `key=`

plugins/services/src/js/components/modals/NewCreateServiceModalForm.js
```diff
@@ -334,33 +345,33 @@ class NewCreateServiceModalForm extends Component {
       return [
```
wavesoft: Also here, `key=`

PR #1613, opened 2016-12-15
Re-adding missing reverse of data, when reading reverse

mlunoe:
This PR fixes a bug where it would not reverse the received data, when reading in reverse order.
It also makes sure to always check if we are at the top of the logs, so we don't get stuck at a perceived infinite load at the top.

To test this out, deploy some service that has more than a page of logs and try to page to the top

### DIFF
--- a/plugins/services/src/js/components/LogView.js
+++ b/plugins/services/src/js/components/LogView.js
@@ -132,9 +132,9 @@ class LogView extends React.Component {
       return;
     }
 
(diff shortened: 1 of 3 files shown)


philipnrmn:
Works for me 👍

MatApple:
@mlunoe when I click through to the stderr tab it never loads. Any ideas?

MatApple:
Otherwise LGTM!

PR #1614, opened 2016-12-15
Restoring '!!service.getQueue' in ServicesTable.js

wavesoft:
~Boy-scout clean-up went wrong :rotating_light:~
Boy-scout clean-up done right 👮

### DIFF
--- a/plugins/services/src/js/containers/services/ServicesTable.js
+++ b/plugins/services/src/js/containers/services/ServicesTable.js
@@ -232,7 +232,11 @@ class ServicesTable extends React.Component {
       verboseOverview = ` (${tasksRunning} of ${instancesCount} Instances)`;
     }
 
-    let queue = service.getQueue();
+    let queue = null;
+    if (!!service.getQueue) {
+      queue = service.getQueue();
+    }
+
     if (queue != null) {
       const waitingSince = DateUtil.strToMs(queue.since);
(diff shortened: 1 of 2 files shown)


plugins/services/src/js/containers/services/ServicesTable.js
```diff
@@ -232,7 +232,11 @@ class ServicesTable extends React.Component {
       verboseOverview = ` (${tasksRunning} of ${instancesCount} Instances)`;
     }
 
-    let queue = service.getQueue();
+    let queue = null;
+    if (!!service.getQueue) {
```
orlandohohmeier: Let's add `getQueue` to the service struct instead of doing this check...

jfurrow: We'll also need to add it to `ServiceTree` if we go down this route.

jfurrow:
This fixes the bug and LGTM after @orlandohohmeier's comment is addressed! :ship:

plugins/services/src/js/containers/services/ServicesTable.js
wavesoft: Ok, that sounds like a better 'clean-up' will do 👍

rcorral:
LGTM! Thanks!

orlandohohmeier:
LGTM 💯 ✨

PR #1615, opened 2016-12-15
Formatting messages to be equal to CLI format

mlunoe:
This PR introduces a new format for log lines that follows the CLI (see here https://github.com/dcos/dcos-cli/pull/817/files#diff-8f3b06e62cf338c8e4e2ac6414447d26R260):

### DIFF
--- a/src/js/stores/SystemLogStore.js
+++ b/src/js/stores/SystemLogStore.js
@@ -19,7 +19,8 @@ import {
 import BaseStore from './BaseStore';
 import SystemLogActions from '../events/SystemLogActions';
 import {APPEND, PREPEND} from '../constants/SystemLogTypes';
-import Util from '../utils/Util';
+import {findNestedPropertyInObject} from '../utils/Util';
+import {msToCTime} from '../utils/DateUtil';
 
 // Max data storage, i.e. maximum 5MB per log stream
 const MAX_FILE_SIZE = 500000;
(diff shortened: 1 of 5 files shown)


MatApple:
@mlunoe legit failing tests

mlunoe:
Oh man, different time on the CI than locally :/

src/js/stores/SystemLogStore.js
```diff
@@ -117,13 +118,52 @@ class SystemLogStore extends BaseStore {
...
+      const {fields = {}} = entry;
+      let lineData = [];
+      // entry.realtime_timestamp returns a unix time in microseconds
+      // https://www.freedesktop.org/software/systemd/man/sd_journal_get_realtime_usec.html
+      if (typeof entry.realtime_timestamp === 'number') {
+        lineData.push(msToCTime(entry.realtime_timestamp/1000));
+      }
+
+      // Optional fields
+      ['_HOSTNAME', 'SYSLOG_IDENTIFIER'].forEach((optionalField) => {
```
MatApple: Let's put these these in a constants file since we reuse them.

mlunoe: Ok, moved them to a constants file

jfurrow:
To me this seems like this makes the logs really hard to parse because the length of the prefix is so long. Maybe the prefix could be on a new line of its own, or could be a lighter shade of grey to offset it from the output? Just an idea, maybe this isn't a problem!

mlunoe:
I cannot change the format. It has to follow the CLI exactly, as it does now

jfurrow:
@mlunoe Okay cool, thanks!

jfurrow:
Works great, thanks for adding tests!

PR #1617, opened 2016-12-15
Handle empty log by sending update if nothing has happened

mlunoe:

### DIFF
--- a/plugins/services/src/js/components/LogView.js
+++ b/plugins/services/src/js/components/LogView.js
@@ -20,7 +20,7 @@ class LogView extends React.Component {
     this.state = {
       isAtBottom: true,
       userScroll: true,
-      fullLog: null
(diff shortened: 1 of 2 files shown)


MatApple:
@mlunoe I think we should handle this by still showing the log viewer, either as empty or with a message conveying the fact that the logs are empty.

mlunoe:
Isn't that what we hare doing?

MatApple:
I guess we are, perhaps it's just taking a long time to fetch the error log before we realize it's empty. Is that what's happening?

mlunoe:
Yup I can set the timeout to be slower, like 4 seconds?

mlunoe:
Done

jfurrow:
This makes sense to me and works well. :ship:

jfurrow:
Feel free to merge when tests pass!

PR #1619, opened 2016-12-15
DCOS-12280: Move version modal to cluster config screen

jfurrow:
This PR adds the `VersionsModal` to the cluster configuration screen. It also removes the actions from the `SidebarActions` and into the `MetadataActions`.


wavesoft:
Looks good! 👍 ✨

wavesoft:
TBH, I would prefer if that link was in the header, aligned with the "Cluster" headline.
It feels kind of weird if it's part of the description table...

jfurrow:
@wavesoft That's a good point, it probably would look better in the page header!

@ashenden What do you think?

rcorral:
run integration

ashenden:
In general, we want to reserve the dropdown in the two main buckets of info User Actions/Context and Support.  In fact, we are looking to change the info there from "Cluster Name/IP" to "Cluster Name/Username".

Information about your cluster, settings for it, etc... should live under the `System` section in the sidebar.  With that in mind, the content in this modal should find it's home there.  FWIW, I don't think this should even be a modal, but a more parseable set of data available in the Cluster section -- or at the very least, just a file you can download.

jfurrow:
@ashenden We were actually talking about putting it in this page's header actions dropdown, not the overall user account dropdown.

See this screenshot as an example:


So rather than being a large button that's in a row of its own, it would be in that dropdown.

This is just a JSON file that we're fetching and displaying in the modal, so we could just link to it instead. Which would you prefer?

jfurrow:
After speaking with @ashenden outside of this PR, we are going to ditch the modal and just display a download link the page header:

wavesoft:
That's what i'd call a "middle ground solution" 👍 
run integration

wavesoft:
run integration

jfurrow:
This is rebased and should be good to go! :)

src/js/events/MetadataActions.js
```diff
@@ -7,6 +7,27 @@ import Config from '../config/Config';
 
 var MetadataActions = {
 
+  fetchDCOSBuildInfo() {
```
nLight: 👍

PR #1621, opened 2016-12-16
DCOS-12327: Removing toJSON from ApplicationSpec

wavesoft:
This PR removes an old functionality from the `ApplicationSpec.toJSON` that prohibits instances of type `MESOS` and having an `docker.image` property to scale properly.


mlunoe:
:shipit:

PR #1625, opened 2016-12-16
DCOS-12292: Display all errors on the review screen

nLight:
This PR adds support for top-level error messages with an empty path.

### DIFF
--- a/plugins/services/src/js/service-configuration/ServiceConfigDisplay.js
+++ b/plugins/services/src/js/service-configuration/ServiceConfigDisplay.js
@@ -77,13 +77,24 @@ class ServiceConfigDisplay extends React.Component {
   }
 
   getRootErrors() {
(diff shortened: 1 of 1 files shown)


orlandohohmeier:
✨ Great

wavesoft:
Is it possible to somehow detect unhandled paths? Because what would happen for example if you get a `expecting non-null value` on a property we don't know?

wavesoft:
Otherwise this looks good 👍

nLight:
@wavesoft I didn't quite get what you mean 🤔

wavesoft:
Ok, just expose all the errors you receive now (not only top-level) just in case we don't handle some of the paths in the review sections.

nLight:
@orlandohohmeier @wavesoft Changed the implementation to display all the errors while we don't highlight `em line by line

nLight:
run integration tests

PR #1630, opened 2016-12-16
Opimize mesos state store scheduler handling

orlandohohmeier:
---
⚠️  ~~_This branch depends on changes that will be introduced with #1566_~~

---

Improve `MesosStateStore` scheduler task handling by introducing a new method to retrieve the list of scheduler tasks and simplifying the existing methods. These changes will reduce the number of `framework`/`task` list iterations. 

/cc @MatApple

### DIFF
--- a/src/js/stores/MesosStateStore.js
+++ b/src/js/stores/MesosStateStore.js
@@ -245,19 +245,13 @@ class MesosStateStore extends GetSetBaseStore {
   }
 
   getSchedulerTaskFromServiceName(serviceName) {
-    const tasks = this.getTasksFromServiceName('marathon');
-    return tasks.find(function (task) {
-      let labels = task.labels;
+    const tasks = this.getSchedulerTasks();
 
-      if (!labels) {
-        return false;
-      }
+    return tasks.find(function ({labels}) {
 
-      const frameworkName = labels.find(function (label) {
-        return label.key === 'DCOS_PACKAGE_FRAMEWORK_NAME';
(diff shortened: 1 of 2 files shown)


src/js/stores/__tests__/MesosStateStore-test.js
```diff
@@ -180,6 +180,52 @@ describe('MesosStateStore', function () {
...
+                id: 1,
+                labels: [{key: 'DCOS_PACKAGE_FRAMEWORK_NAME', value: 'foo'}]
+              },
+              {
+                id: 2
+              },
+              {
+                id: 3,
+                labels: [{key: 'DCOS_PACKAGE_FRAMEWORK_NAME', value: 'bar'}]
+              }
```
mlunoe: Could we also add some tasks that will not be included, to test out that scenario as well?

src/js/stores/MesosStateStore.js
```diff
@@ -229,36 +220,30 @@ class MesosStateStore extends GetSetBaseStore {
...
     });
+  }
 
-    if (!framework || !framework.tasks) {
-      return null;
-    }
-
-    let result = framework.tasks.find(function (task) {
-      let labels = task.labels;
+  getSchedulerTaskFromServiceName(serviceName) {
```
mlunoe: Same thing for the existing tests for this function

orlandohohmeier:
@mlunoe addressed your comments...

nLight:
✨ 💯

PR #1634, opened 2016-12-16
move the rest of the files

aeyakovenko:
https://mesosphere.atlassian.net/browse/DCOS-10936


rcorral:
merging as the same failing integration tests are failing on all PRs. We're looking into this right now.

PR #1636, opened 2016-12-16
Fix Volumes tab bug

jfurrow:
This PR fixes a small bug where the Volumes tab would throw an error when selected. It also adds the `Page` and `Page.Header` component to the `VolumeDetail` page.


rcorral:
run integration

PR #1637, opened 2016-12-17
Removes integration test that doesn't pass consistently as application loads too fast.

rcorral:
Open to suggestion on how to fix this, but right now the Config.js file gets bundled with the application and that's where we tell the application how much time to wait before rendering.


rcorral:
run integration

rcorral:
run integration

rcorral:
retest this please

PR #1638, opened 2016-12-17
Fix badge styling on PackageDetail header

jfurrow:
The `flush-left` class was removing both margin and padding from the badge.


jfurrow:
run integration tests

mlunoe:
Before:


After:


Great stuff! :shipit:

PR #1640, opened 2016-12-19
DCOS-12224: Handle missing protocol in app definition

nLight:
This PR adds a check for the situation when a protocol field is missing from the port mappings

### DIFF
--- a/plugins/services/src/js/service-configuration/ServiceNetworkingConfigSection.js
+++ b/plugins/services/src/js/service-configuration/ServiceNetworkingConfigSection.js
@@ -79,8 +79,10 @@ class ServiceNetworkingConfigSection extends ServiceConfigBaseSectionDisplay {
                 prop: keys.protocol,
                 className: getColumnClassNameFn(),
                 render(prop, row) {
-                  return getDisplayValue(row[prop])
-                    .split(',').join(', ');
+                  let protocol = row[prop] || '';
+                  protocol = protocol.replace(',', ', ');
+
+                  return getDisplayValue(protocol);
                 },
                 sortable: true
               },
--- a/plugins/services/src/js/service-configuration/ServiceNetworkingConfigSection.js
+++ b/plugins/services/src/js/service-configuration/ServiceNetworkingConfigSection.js
@@ -80,7 +80,7 @@ class ServiceNetworkingConfigSection extends ServiceConfigBaseSectionDisplay {
                 className: getColumnClassNameFn(),
                 render(prop, row) {
                   let protocol = row[prop] || '';
-                  protocol = protocol.replace(',', ', ');
+                  protocol = protocol.replace(/,\s*/g, ', ');
 
                   return getDisplayValue(protocol);
                 },


nLight:
run integration tests

rcorral:
run integration

rcorral:
run integration tests

plugins/services/src/js/service-configuration/ServiceNetworkingConfigSection.js
```diff
@@ -79,8 +79,10 @@ class ServiceNetworkingConfigSection extends ServiceConfigBaseSectionDisplay {
                 prop: keys.protocol,
                 className: getColumnClassNameFn(),
                 render(prop, row) {
-                  return getDisplayValue(row[prop])
-                    .split(',').join(', ');
```
wavesoft: Here you first `getDisplayValue` and then `split` ...

plugins/services/src/js/service-configuration/ServiceNetworkingConfigSection.js
```diff
@@ -79,8 +79,10 @@ class ServiceNetworkingConfigSection extends ServiceConfigBaseSectionDisplay {
...
+                  let protocol = row[prop] || '';
+                  protocol = protocol.replace(',', ', ');
+
+                  return getDisplayValue(protocol);
```
wavesoft: But here you first `split` and then `getDisplayValue`. What's correct?

plugins/services/src/js/service-configuration/ServiceNetworkingConfigSection.js
```diff
@@ -79,8 +79,10 @@ class ServiceNetworkingConfigSection extends ServiceConfigBaseSectionDisplay {
                 prop: keys.protocol,
                 className: getColumnClassNameFn(),
                 render(prop, row) {
-                  return getDisplayValue(row[prop])
-                    .split(',').join(', ');
+                  let protocol = row[prop] || '';
+                  protocol = protocol.replace(',', ', ');
```
wavesoft: Use `.replace(/,/g, ', ')` because `.replace('string')` will replace only the first occurrence. And while at it, why don't you try `/,.*/g` just to be safe?

plugins/services/src/js/service-configuration/ServiceNetworkingConfigSection.js
Poltergeist: Good question.

nLight: Preparing a value first and then feeding it to `getDisplayValue` is correct.

plugins/services/src/js/service-configuration/ServiceNetworkingConfigSection.js
nLight: Currently it is an `enum` value can be `udp`, `tcp` or `tcp,udp` and that guy `/,.*/g` will just eat the `,udp` part entirely. So we could go with `//g` flag but I wouldn't change the expression itself.

Poltergeist: I would still stick to `/,(?=\S)/g`, because it could always change and then we need to find those locations where we did not think about having more then two items.

nLight: Implemented `/,\s*/g` version

nLight:
run integration tests

wavesoft:
Great! Thanks! ✨ 💯

orlandohohmeier:
LGTM! 🚢

PR #1643, opened 2016-12-19
Adds flag to docker container.

rcorral:
To use host resources, this avoids Chrome crashing in container due to it not GCing on time.


orlandohohmeier:
@rcorral could you please add an empty line in your commit messages to properly separate the subject from the message body? This will make it easier to brows thought the list of logs.

orlandohohmeier:
L👀ks good!

PR #1646, opened 2016-12-20
DCOS-11886: Prevent .row styles from mucking with .panel-grid children

jfurrow:
This PR fixes a bug where the `:before` and `:after` pseudo elements from `.row` were rendering in Safari and causing the `.panel-grid` children to wrap in Safari.

Before:

After:

### DIFF
--- a/src/styles/components/panel/styles.less
+++ b/src/styles/components/panel/styles.less
@@ -48,6 +48,14 @@
     .panel {
       margin-bottom: @panel-grid-margin-vertical;
     }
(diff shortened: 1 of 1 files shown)


nLight:
LGTM!

Unrelated: It still messes the login button tho
screen shot 2016-12-20 at 11 31 27

wavesoft:
LGTM too! I also see the same issue with @nLight , but that's outside the scope of this PR.

PR #1647, opened 2016-12-20
Fixes two bugs with task health and status

rcorral:
1. We separated task health and status, but didn’t think about how other parts of the application use task health. This caused an issue where all tasks were shown as healthy.
2. We weren’t polling marathon in the Nodes page, which caused task to show as “Unkown” health status until we went to a page that polled marathon.

img

Closes DCOS-12888


rcorral:
Needs integration tests.

rcorral:
Integration tests added 🌮

MatApple:
run integration tests

orlandohohmeier:
@rcorral any chance you can simplify the included fixtures?

orlandohohmeier:
Tasks from an app with health checks should be marked as "unknown" until we have the respective health results and not rendered as "No health checks available".

rcorral:
Aren't we certain that there is no health check set up and therefore it would be OK to say "No health checks available"?

orlandohohmeier:
I'm pretty sure that Marathon UI as well as the "old" implementation properly marked tasks as unknown... Anyway, how about rephrasing the tooltip to something like "no health check results available"?

nLight:
This can be related https://mesosphere.atlassian.net/browse/DCOS-12888

rcorral:
@orlandohohmeier cleaned up fixtures.

src/js/config/Config.test.js
```diff
@@ -4,7 +4,7 @@ var ConfigDev = {
   rootUrl: '',
   historyServer: '',
   analyticsKey: '39uhSEOoRHMw6cMR6st9tYXDbAL3JSaP',
-  stateRefresh: 2000,
+  stateRefresh: 20000,
```
jfurrow: Why is this increased to 20 seconds?

jfurrow:
I wasn't able to reproduce #2, but #1 is working as expected. I had one question, but otherwise this LGTM!

philipnrmn:
I'm happy to see it 🚢 d

src/js/config/Config.test.js
rcorral: it just helps when running integration tests. It doesn't pollute the heck out of the history that cypress keeps. 

I'm OK with removing this change if you want. I guess it could cause issues in the future, but only when running `npm run testing` which won't run in CI.

rcorral:
@jfurrow you have to load the Nodes page directly, and not load any other page that polls marathon first, otherwise we will already have the data.

src/js/config/Config.test.js
jfurrow: That makes good sense! I'm fine with you leaving it, I was just curious why.

mlunoe: There is some sense in keeping it the same as in the application, but I also see the benefit of it.

PR #1659, opened 2016-12-20
Delete ZeroClipboard.swf

philipnrmn:
As far as I'm able to determine, we don't use zeroclipboard anywhere, not even as a fallback. 

Let's kill the swf file!


MatApple:
Nice find @philipnrmn!!

PR #1662, opened 2016-12-20
Bug fix: Make edit button point to correct view for multiple containers, ref. DCOS-11676

mlunoe:


plugins/services/src/js/service-configuration/PodContainerArtifactsConfigSection.js
```diff
@@ -57,7 +57,11 @@ class PodContainerArtifactsConfigSection extends React.Component {
   }
 
   render() {
-    const {artifacts, onEditClick} = this.props;
+    const {artifacts, index, onEditClick} = this.props;
+    let tabViewID = 'services';
```
MatApple: ```const tabViewID = index != null ? `container${index}` : 'services';``` ;p

mlunoe: We decided not to use this before we had the eslint rule

PR #1665, opened 2016-12-21
Bug fix: Minor issues in multiple container feature branch

mlunoe:
This PR fixes an inconsistency issue, where we would toggle `Enabled`/`Disabled`, instead of having it be `Enabled` all the time.
It fixes the interaction with numbers to be a little more user-friendly, and an issue, where users could not enter a small number like 0.03.
Finally, it passes down errors to container sub-sections correctly.

### DIFF
--- a/plugins/services/src/js/components/forms/ContainerServiceFormSection.js
+++ b/plugins/services/src/js/components/forms/ContainerServiceFormSection.js
@@ -98,7 +98,9 @@ class ContainerServiceFormSection extends Component {
     const inputFiled = (
       <FieldInput
         key="gpus-input"
+        min="0.001"
         name="gpus"
+        step="any"
         type="number"
         disabled={disabled}
         value={data.gpus} />
@@ -189,7 +191,9 @@ class ContainerServiceFormSection extends Component {
           <FormGroup className="column-4" showError={Boolean(errors.disk)}>
             <FieldLabel>Disk (MiB)</FieldLabel>
             <FieldInput
+              min="0.001"
               name="disk"
+              step="any"
               type="number"
               value={data.disk} />
             <FieldError>{errors.disk}</FieldError>
@@ -337,9 +341,10 @@ class ContainerServiceFormSection extends Component {
               CPUs
             </FieldLabel>
             <FieldInput
+              min="0.001"
               name="cpus"
+              step="any"
               type="number"
-              step="0.01"
               value={data.cpus} />
             <FieldError>{errors.cpus}</FieldError>
           </FormGroup>
@@ -352,7 +357,9 @@ class ContainerServiceFormSection extends Component {
               Memory (MiB)
             </FieldLabel>
             <FieldInput
+              min="0.001"
               name="mem"
+              step="any"
               type="number"
               value={data.mem} />
             <FieldError>{errors.mem}</FieldError>
--- a/plugins/services/src/js/components/forms/ContainerServiceFormSection.js
+++ b/plugins/services/src/js/components/forms/ContainerServiceFormSection.js
@@ -98,7 +98,7 @@ class ContainerServiceFormSection extends Component {
     const inputFiled = (
       <FieldInput
         key="gpus-input"
-        min="0.001"
+        min="0"
         name="gpus"
         step="any"
         type="number"
--- a/plugins/services/src/js/components/forms/MultiContainerNetworkingFormSection.js
+++ b/plugins/services/src/js/components/forms/MultiContainerNetworkingFormSection.js
@@ -126,11 +126,6 @@ class MultiContainerNetworkingFormSection extends mixin(StoreMixin) {
       address = `${hostname}${Networking.L4LB_ADDRESS}${port}`;
     }
 
-    let loadBalancedLabel = 'Disabled';
-    if (loadBalanced) {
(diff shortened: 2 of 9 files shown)


plugins/services/src/js/reducers/serviceForm/Containers.js
```diff
@@ -672,6 +680,9 @@ module.exports = {
     }
 
     if (type === SET && field === 'resources') {
+      // Do not parse numbers, as user might be in the middle of
+      // entering a number. 0.0 would then equal 0, yielding it impossible to,
```
mlunoe: I English well!

mlunoe:
run integration tests

plugins/services/src/js/components/forms/MultiContainerNetworkingFormSection.js
```diff
@@ -152,7 +147,7 @@ class MultiContainerNetworkingFormSection extends mixin(StoreMixin) {
               checked={loadBalanced}
               name={`containers.${containerIndex}.endpoints.${index}.loadBalanced`}
               type="checkbox" />
-            {loadBalancedLabel}
+            Enabled
```
orlandohohmeier: Did we fixed this for the single container as well?

plugins/services/src/js/components/forms/ContainerServiceFormSection.js
```diff
@@ -352,7 +357,9 @@ class ContainerServiceFormSection extends Component {
               Memory (MiB)
             </FieldLabel>
             <FieldInput
+              min="0.001"
```
orlandohohmeier: Mesos requires 32 min... Let's adjust this value.

orlandohohmeier: Hm... The API spec actually specifies `0.001` as minimum, thus I'd say we should keep it as it is. 

https://github.com/mesosphere/marathon/blob/master/docs/docs/rest-api/public/api/v2/types/resources.raml#L18

plugins/services/src/js/components/forms/ContainerServiceFormSection.js
```diff
@@ -98,7 +98,9 @@ class ContainerServiceFormSection extends Component {
     const inputFiled = (
       <FieldInput
         key="gpus-input"
+        min="0.001"
```
orlandohohmeier: Let's change this to `0` as it is specified as an int32... 

https://github.com/mesosphere/marathon/blob/master/docs/docs/rest-api/public/api/v2/types/resources.raml

philipnrmn: This will cause the value to default to 0, which strikes me as a bad value for number of cpus...

plugins/services/src/js/components/forms/ContainerServiceFormSection.js
philipnrmn: Let's respect the RAML defaults - 1.0 CPUs and 128m memory.

plugins/services/src/js/components/forms/ContainerServiceFormSection.js
orlandohohmeier: It's GPUs and we shouldn't have a default other then `0` as  gpu support is still experimental.

plugins/services/src/js/components/forms/ContainerServiceFormSection.js
orlandohohmeier: Isn's this the `min` and not the `default` value? I thought we pre-populate the form with reasonable defaults.

philipnrmn: yes.

plugins/services/src/js/components/forms/MultiContainerNetworkingFormSection.js
mlunoe: Yes, it was. This is the inconsistency

plugins/services/src/js/components/forms/ContainerServiceFormSection.js
philipnrmn: but the populated value is also 0.001.

philipnrmn: It is! But looks like default memory is 32, not 128.

plugins/services/src/js/components/forms/ContainerServiceFormSection.js
mlunoe: Done!

philipnrmn:
I fixed up a serious bug which was causing the whole form to error out, and am adding my label. Can somebody else give this a last pass?

plugins/services/src/js/components/forms/ContainerServiceFormSection.js
mlunoe: Updated the defaults to match the raml files

mlunoe:
run integration tests

philipnrmn:
Thanks for taking care of the defaults @mlunoe

philipnrmn:
run integration tests 💃

mlunoe:
run integration tests

mlunoe:
run integration tests

PR #1670, opened 2016-12-21
Add Support for health checks with unknown protocols

Poltergeist:
This adds the support for healthchecks without a known protocol. We will omit those from the form

image

PR #1671, opened 2016-12-21
show healthchecks with unknown values

Poltergeist:
~~BASED on #1670~~ Merged

Added support for healthchecks with unknown protocol in the ui. There might come design suggestion for this feature. With this PR it will be possible to delete healthchecks with unknown protocols from the form.

image

\cc @leemunroe


Poltergeist:
Run integration tests

nLight:
Rebased that one, otherwise plugins threw an exception.

mlunoe:
Same error (as in the base PR) appears here:

mlunoe:
Problem fixed in base PR, which merged.

PR #1672, opened 2016-12-21
DCOS-9310: Adjust default job memory

orlandohohmeier:
Introduce job resource defaults and use the new constants instead of Mesos constants to properly define job defaults that match the API.

Closes DCOS-9310


rcorral:
merging as the same failing integration tests are failing on all PRs. We're looking into this right now.

PR #1674, opened 2016-12-21
DCOS-12404: Add missing requirePorts field

nLight:
This PR adds missing `requirePorts` field when a user specifies a hostPort for her Docker app with a Bridge network. This field is important because otherwise Marathon ignores specified hostPort and uses a random one.


nLight:
retest this please

PR #1675, opened 2016-12-21
Add Drag and drop component

Poltergeist:
This has been the work of @gargs0n

This adds the possibility to drop a JSON file onto the form / JSON editor and then uses the contents to populate the form and the JSON editor.

draganddropfiles mov

### DIFF
--- a/plugins/services/src/js/components/DragAndDrop.js
+++ b/plugins/services/src/js/components/DragAndDrop.js
@@ -1,0 +1,71 @@
+import React, {PropTypes, Component} from 'react';
+
+const METHODS_TO_BIND = [
+  'handleDragEvents',
+  'handleFileResult',
+  'handleOnDrop'
+];
+
+class DragAndDrop extends Component {
+  constructor() {
+    super(...arguments);
+
+    this.fileReader = new FileReader();
+
+    METHODS_TO_BIND.forEach((method) => {
+      this[method] = this[method].bind(this);
+    });
+  }
+
+  componentDidMount() {
+    this.fileReader.addEventListener('load', this.handleFileResult, false);
+
+  }
+
+  componentWillUnmount() {
+    this.fileReader.removeEventListener('load', this.handleFileResult, false);
+  }
+
+  handleFileResult() {
+    this.props.handleFileContents(this.fileReader.result);
+  }
+
+  // In order to have the drop event occur on a div element, you must cancel
+  // the ondragenter and ondragover events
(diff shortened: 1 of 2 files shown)


plugins/services/src/js/components/DragAndDrop.js
```diff
@@ -0,0 +1,71 @@
...
+  // the ondragenter and ondragover events
+  handleDragEvents(event) {
+    event.preventDefault();
+  }
+
+  handleOnDrop(event) {
+
+    event.preventDefault();
+
+    const files = event.dataTransfer ? event.dataTransfer.files : event.target.files;
```
mlunoe: > 80 chars

plugins/services/src/js/components/modals/NewCreateServiceModalForm.js
```diff
@@ -184,6 +187,10 @@ class NewCreateServiceModalForm extends Component {
     });
   }
 
+  handleFileContents(appString) {
+    this.handleJSONChange(JSON.parse(appString));
```
mlunoe: We need a `try`-`catch` around this and a fallback if it fails, e.g. a modal

plugins/services/src/js/components/modals/NewCreateServiceModalForm.js
```diff
@@ -258,11 +265,15 @@ class NewCreateServiceModalForm extends Component {
...
+      <DragAndDrop className="flex flex-item-grow-1"
+        handleFileContents={this.handleFileContents}>
+        <div
+          className="flex flex-item-grow-1 modal-body-offset gm-scrollbar-container-flex">
           <FluidGeminiScrollbar>
-            <div className="container flex flex-direction-top-to-bottom modal-body-padding-surrogate">
-              <form onChange={this.handleFormChange} onBlur={this.handleFormBlur}>
+            <div
+              className="container flex flex-direction-top-to-bottom modal-body-padding-surrogate">
+              <form onChange={this.handleFormChange}
```
mlunoe: #nit separate lines for the attributes

mlunoe:
Also, we should probably have a modal to accept that you will overrode already entered data (if you have made a change).
How does this work, if you drag in a pod spec on the regular form and the other way around?
This doesn't seem to work with the JSON editor either

wavesoft:
Ps. Might be a good idea to show a "Drop file here" message on hover. We can also show there a warning like, "dropping this file here will overwrite your existing data"

PR #1681, opened 2016-12-22
Cleanup: Using global consistently

mlunoe:
This PR cleans up the way we use global


mlunoe:
run unit tests

plugins/banner/hooks.js
```diff
@@ -49,7 +49,7 @@ module.exports = {
   },
```
mlunoe: Webpack exposes `global` as the global object by default.
Unfortunately, I couldn't find an official description of this, but here is an issue where it is mentioned: https://github.com/webpack/webpack/issues/260.

Also, if you compile our code with `build-assets` without setting `NODE_ENV=production` you will see it happen too:
`/* WEBPACK VAR INJECTION */(function(global) {'use strict';`

wavesoft:
Looks good 💯 ✨

PR #1683, opened 2016-12-22
Using lower case when using task in a sentence, ref. DCOS-11931

mlunoe:
Here are the places that have been changed:


Poltergeist:
LGTM. 🎆

PR #1684, opened 2016-12-22
Make sure to not uppercase resource labels

mlunoe:
Before:

After:


jfurrow:
I know my comment is two weeks too late, but I just wanted to add this comment for reference:

I would prefer for us to handle this as follows:
```jsx
<FieldLabel>
Size <span className="unit">MiB</span>
</FieldLabel>
```

Then we can just remove the `text-transform` property for the unit class instead of having to hard-code the uppercase labels.

PR #1690, opened 2016-12-22
Improving the package search, ref. DCOS-5779

mlunoe:
Improving package search, so `ch` does not match Apache, since we are using startsWith:


Poltergeist:
LGTM 🎆

PR #1694, opened 2016-12-27
Service name > Service ID in service create form

MatApple:
Changed copy from `SERVICE NAME` to `SERVICE ID` which now aligns with the config section (REVIEW & RUN).

### DIFF
--- a/plugins/services/src/js/components/forms/GeneralServiceFormSection.js
+++ b/plugins/services/src/js/components/forms/GeneralServiceFormSection.js
@@ -302,7 +302,7 @@ class GeneralServiceFormSection extends Component {
           Services
         </h2>
         <p>
-          Configure your service below. Start by giving your service an Id.
+          Configure your service below. Start by giving your service an ID.
         </p>
 
         <FormRow>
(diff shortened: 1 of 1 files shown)


plugins/services/src/js/components/forms/GeneralServiceFormSection.js
```diff
@@ -302,7 +302,7 @@ class GeneralServiceFormSection extends Component {
           Services
         </h2>
         <p>
-          Configure your service below. Start by giving your service a name.
+          Configure your service below. Start by giving your service an Id.
```
mlunoe: Let's use ID (uppercase) in the paragraph as well

MatApple: I disagree - It looks strange with ID. I did try that first. @leemunroe @ashenden anyone want to weigh in here?

ashenden: It should be All Caps - http://english.stackexchange.com/questions/101248/how-should-the-abbreviation-for-identifier-be-capitalized

MatApple:
@mlunoe @ashenden thanks - updated the copy.

mlunoe:
Let's ⛵️ this thing :)

PR #1695, opened 2016-12-27
DCOS-12421: Add sensible defaults to healthcheck settings

MatApple:
https://mesosphere.atlassian.net/browse/DCOS-12421

Add defaults for health check settings.


orlandohohmeier:
Let's not add any defaults for `gracePeriodSeconds`, `intervalSeconds`, `timeoutSeconds` or `maxConsecutiveFailures` as all of them are optional (see [healthCheck.raml](https://github.com/mesosphere/marathon/blob/master/docs/docs/rest-api/public/api/v2/types/healthCheck.raml) for details) and adding those only increases maintenance without providing any value. Having those defaults also makes it harder to read the JSON generated by the form.

PR #1689, opened 2016-12-22
Adjusting the scroll container listener to be the page content

mlunoe:
---
ℹ️ This is the same fix as for release 1.8: https://github.com/dcos/dcos-ui/pull/1688

---

Before:

After:

PR #1696, opened 2016-12-27
Revert "Adjusting the scroll container listener to be the page content"

mlunoe:
Reverts dcos/dcos-ui#1689

Fix being applied here: https://github.com/dcos/dcos-ui/pull/1699


rcorral:
The real fix is being applied to reactjs-components.

PR #1702, opened 2016-12-27
DCOS-12289: Fix json generation for host volumes

Poltergeist:
Host volumes should only be rendered if the
runtime type is MESOS or DOCKER or a docker image
has been set

close DCOS-12289

### DIFF
--- a/plugins/services/src/js/reducers/serviceForm/Volumes.js
+++ b/plugins/services/src/js/reducers/serviceForm/Volumes.js
@@ -60,6 +60,10 @@ module.exports = {
       this.docker = value !== '';
     }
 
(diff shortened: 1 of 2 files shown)


nLight:
The problem is still there, I'll reassign the ticket to myself @orlandohohmeier

nLight:
OK, so I misunderstood the problem, but while debugging it I've discovered that we display 'Not Configured' for an empty value where we should actually say 'Not supported' that is fixed now.

screen shot 2017-01-05 at 16 09 47

jfurrow:
Seems to work for me!

wavesoft:
Looks good! ✨  👍

PR #1711, opened 2016-12-28
Add a set of app specs for manual testing

orlandohohmeier:
This PR adds a set of app/service specs for manual testing.


wavesoft:
I assume these are copy-paste-able references, not really used by the code, right?

In that case, would it make sense to move them in the `docs/examples` folder, or in a wiki page?
Because `resources` contains files that are supposed to be part of the final build, directly or indirectly.

wavesoft:
Not blocking the PR though.

orlandohohmeier:
@wavesoft good point! Going to close this and the related PRs.

PR #1716, opened 2017-01-03
DCOS-11957: Add redirect after destroying a service

nLight:
This PR adds a delayed redirect after a service got destroyed via the `ServiceDestroyModal`.
We have to delay the redirect because Router redirects faster than our code manages to close the modal which results in the Modal inviting you to destroy the root `/` group as well.

### DIFF
--- a/plugins/services/src/js/components/modals/ServiceDestroyModal.js
+++ b/plugins/services/src/js/components/modals/ServiceDestroyModal.js
@@ -10,7 +10,8 @@ import Pod from '../../structs/Pod';
 import Service from '../../structs/Service';
 import ServiceTree from '../../structs/ServiceTree';
 
-const REDIRECT_DELAY = 500;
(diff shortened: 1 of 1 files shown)


nLight:
run unit tests

orlandohohmeier:
L👀ks good!

orlandohohmeier:
🥗  All test pass locally and the test failure is unrelated to the introduced changes.

plugins/services/src/js/components/modals/ServiceDestroyModal.js
```diff
@@ -9,6 +10,11 @@ import Pod from '../../structs/Pod';
 import Service from '../../structs/Service';
 import ServiceTree from '../../structs/ServiceTree';
 
+const REDIRECT_DELAY = 500;
```
wavesoft: This feels a bit random. It's a good idea to comment what's the reasoning behind this number.

PR #1717, opened 2017-01-03
DCOS-12464: Display seconds in the health checks review section

nLight:
This PR addresses the issue with seconds not being displayed in the Health checks review section.
The reason was that `moment.js` library doesn't display the amount of seconds by default like it's not that relevant.

ℹ️  Please note `.humanize()` method doesn't only display seconds by default but it also makes some assumptions on how to round up the numbers. I.e. it will display `a minute` for a value > 45 seconds and `an hour` for a value > 50 minutes. We might want to correct that as well.

Before:
image 2016-12-13 at 3 16 43 pm

After:
screen shot 2017-01-03 at 16 51 03

### DIFF
--- a/plugins/services/src/js/service-configuration/ServiceHealthChecksConfigSection.js
+++ b/plugins/services/src/js/service-configuration/ServiceHealthChecksConfigSection.js
@@ -4,10 +4,10 @@ import {Table} from 'reactjs-components';
 import {
   getColumnClassNameFn,
   getColumnHeadingFn,
-  getDisplayValue,
-  renderMillisecondsFromSeconds
+  getDisplayValue
 } from '../utils/ServiceConfigDisplayUtil';
 import ConfigurationMapEditAction from '../components/ConfigurationMapEditAction';
+import ConfigurationMapDurationValue from '../components/ConfigurationMapDurationValue';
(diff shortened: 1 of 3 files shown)


orlandohohmeier:
How about adding a tooltip including the original value?

orlandohohmeier:
🍏  All test pass locally and the test failure is unrelated to the introduced changes.

orlandohohmeier:
Code l👀ks good!

nLight:
@wavesoft awesome! I'll fix the pr then!

nLight:
@wavesoft Done! But that was someone else duplicating the code :notsureif: 😜

wavesoft:
Well, the plan was to move towards declarative design and move away from the definition-driven views. The `PodXXX` config sections have the base components that should be used in the Services. After the 1.9 hype is gone, I am going to use the sam principles on the `Service` views too.

wavesoft:
Looks good, works good ✨ 💯 
/cc @orlandohohmeier to remove _hold merge_ if he is OK with this change.

wavesoft:
Ok, we can merge this and if there are issues, we resolve them in a different PR.

PR #1719, opened 2017-01-04
Fix deep copy util references

orlandohohmeier:
Change `isObject` and `deepCopy` references from `this` to `Util`, so that they're "always" pointing to the correct method. Without this change it's impossible to only import the `deepCopy` (`import {deepCopy} from 'Util'`) method as the respective methods are undefined due to broken references.

/cc @nLight


nLight:
Funny:
/jenkins/workspace/public-dcos-ui-pulls/plugins/services/src/js/utils/ServiceUtil.js
  460:5  error  Expected newline before return statement  newline-before-return

unrelated eslint error

orlandohohmeier:
🤣  How is that even possible?

orlandohohmeier:
Fixed the linting error in #1720

wavesoft:
Looks good :shipit: !

wavesoft:
Going to rebase this because the errors seem to be purely on code styling.

PR #1723, opened 2017-01-04
DCOS-12847: Use var instead of let in JISON parser

jfurrow:
This PR uses `var` instead of `let` in the JISON parser generator because this won't be transpiled by Babel. This fixes an error thrown in any browser that doesn't support the use of `let` variable declarations.


rcorral:
@wavesoft just a self note to not use ES6 here, or to make sure we transpile it somehow.

wavesoft:
😮 Good catch @jfurrow , @rcorral yes, it must have slipped 😞

PR #1724, opened 2017-01-04
DCOS-12431: Apply flex class to AccessDeniedPage

jfurrow:
This fixes a bug in Safari where the absence of `flex-grow: 1` prevented the content from being visible, so it was just a blank white page.

After:


wavesoft:
Looks good! :shipit:

PR #1725, opened 2017-01-04
DCOS-12774: [1/2] Fix multi-container health-check parsers & reducers

wavesoft:
This PR fixes the multi-container health-check parsers & reducers

### DIFF
--- a/plugins/services/src/js/reducers/serviceForm/Containers.js
+++ b/plugins/services/src/js/reducers/serviceForm/Containers.js
@@ -9,7 +9,10 @@ import {
   simpleFloatReducer,
   simpleReducer
 } from '../../../../../../src/js/utils/ReducerUtil';
-import {JSONReducer as MultiContainerHealthChecks} from './MultiContainerHealthChecks';
+import {
+  JSONSegmentReducer as multiContainerHealthCheckReducer,
+  JSONSegmentParser as multiContainerHealthCheckParser
+} from './MultiContainerHealthChecks';
 import {findNestedPropertyInObject} from '../../../../../../src/js/utils/Util';
 import Networking from '../../../../../../src/js/constants/Networking';
 import {FormReducer as volumeMountsReducer} from './MultiContainerVolumes';
@@ -26,13 +29,6 @@ const containerFloatReducer = combineReducers({
   mem: simpleFloatReducer('resources.mem'),
   disk: simpleFloatReducer('resources.disk')
 });
-
-const advancedContainerSettings = [
-  'gracePeriodSeconds',
-  'intervalSeconds',
-  'timeoutSeconds',
-  'maxConsecutiveFailures'
-];
 
 const defaultEndpointsFieldValues = {
   automaticPort: true,
@@ -141,52 +137,10 @@ function containersParser(state) {
       ], item.privileged));
     }
 
-    if (item.healthChecks != null && item.healthChecks.length !== 0) {
-      memo = item.healthChecks.reduce(function (memo, item, HealthCheckIndex) {
-        if (item.protocol == null) {
-          return memo;
-        }
-        memo.push(new Transaction([
-          'containers',
-          index,
-          'healthChecks'
-        ], HealthCheckIndex, ADD_ITEM));
-        memo.push(new Transaction([
-          'containers',
-          index,
-          'healthChecks',
-          HealthCheckIndex,
-          'protocol'
-        ], item.protocol.toUpperCase(), SET));
-
-        if (item.protocol.toUpperCase() === 'COMMAND') {
-          if (item.command != null && item.command.command != null) {
-            memo.push(new Transaction([
-              'containers',
-              index,
-              'healthChecks',
-              HealthCheckIndex,
-              'command'
-            ], item.command.command, SET));
-          }
-        }
-
-        advancedContainerSettings.filter((key) => {
-          return item[key] != null;
-        }).forEach((key) => {
-          if (item[key] != null) {
-            memo.push(new Transaction([
-              'containers',
-              index,
-              'healthChecks',
-              HealthCheckIndex,
-              key
-            ], item[key], SET));
-          }
-        });
-
-        return memo;
-      }, memo);
+    if (item.healthCheck != null) {
+      memo = multiContainerHealthCheckParser(
+        item.healthCheck, ['containers', index, 'healthCheck']
+      );
     }
 
     if (item.artifacts != null && item.artifacts.length !== 0) {
@@ -348,10 +302,6 @@ module.exports = {
       this.cache = [];
     }
 
-    if (this.healthChecks == null) {
-      this.healthChecks = [];
-    }
-
     if (this.endpoints == null) {
       this.endpoints = [];
     }
@@ -456,18 +406,13 @@ module.exports = {
       return container;
     });
 
-    if (field === 'healthChecks') {
-      if (newState[index].healthChecks == null) {
-        newState[index].healthChecks = [];
+    if (field === 'healthCheck') {
+      if (newState[index].healthCheck == null) {
+        newState[index].healthCheck = {};
       }
 
-      if (this.healthChecks[index] == null) {
-        this.healthChecks[index] = {};
-      }
-
-      newState[index].healthChecks = MultiContainerHealthChecks.call(
-        this.healthChecks[index],
-        newState[index].healthChecks,
+      newState[index].healthCheck = multiContainerHealthCheckReducer.call(
+        newState[index].healthCheck,
         {type, path: path.slice(2), value}
       );
     }
@@ -540,10 +485,6 @@ module.exports = {
       this.cache = [];
     }
 
-    if (this.healthChecks == null) {
-      this.healthChecks = [];
-    }
-
     if (!state) {
       state = [];
     }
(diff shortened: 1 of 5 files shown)


wavesoft:
retest this please

nLight:
I think the ADD/REMOVE_ITEM solution should be copied here

mlunoe:
Tested this out with the final PR https://github.com/dcos/dcos-ui/pull/1728, and it works well

plugins/services/src/js/reducers/serviceForm/MultiContainerHealthChecks.js
```diff
@@ -1,81 +1,354 @@
...
+          break;
+      }
+
+      return newState;
+    }
+
+    // Assign properties
+    switch (group) {
+      case 'exec':
+        reduceCommandHealthCheck(newState, secondField, value);
```
mlunoe: Comment from @nLight:
https://github.com/dcos/dcos-ui/pull/1728#discussion_r94931809
"I'd rather not mutate newState here
then we could even return reduceFormCommandHealthCheck(...) right away"

wavesoft:
* Imported ADD_ITEM/REMOVE_ITEM solution here
* Added tests in order to be able to validate it
* Addressed the mutation issue

orlandohohmeier:

wavesoft:
retest this please

plugins/services/src/js/reducers/serviceForm/MultiContainerHealthChecks.js
```diff
@@ -3,79 +3,423 @@ import {
...
+    }
+
+    // Add an ADD_ITEM transaction
+    memo.push(new Transaction(path, null, ADD_ITEM));
+
+    // Parse detailed fields according to type
+    if (healthCheck.http != null) {
+      memo.push(new Transaction(path.concat(['protocol']), HTTP, SET));
+      memo.push(
+        ...parseHttpHealthCheck(healthCheck.http, path.concat(['http']), memo)
```
mlunoe: #nit why not use `concat` for this? Shouldn't hold up the PR

wavesoft: To avoid creating yet another array instance. But not really sure if that's helping...

nLight: It doesn't really
http://babeljs.io/repl/#?babili=false&evaluate=true&lineWrap=false&presets=latest%2Creact%2Cstage-2&experimental=false&loose=false&spec=false&code=memo.push(%0A%20%20...parse(h.http%2C%20path.concat(%5B'http'%5D)%2C%20memo)%0A)%3B&playground=true

wavesoft:
Let's get over with this 😄

wavesoft:
run unit tests

plugins/services/src/js/reducers/serviceForm/MultiContainerHealthChecks.js
```diff
@@ -3,79 +3,423 @@ import {
...
   SET
 } from '../../../../../../src/js/constants/TransactionTypes';
-import Util from '../../../../../../src/js/utils/Util';
+import {COMMAND, HTTP, HTTPS, TCP} from '../../constants/HealtCheckProtocols';
+import Transaction from '../../../../../../src/js/structs/Transaction';
+import ValidatorUtil from '../../../../../../src/js/utils/ValidatorUtil';
 
-function mapHealthChecks(item) {
-  const newItem = Util.omit(item, ['command', 'protocol']);
+function intOrNull(value) {
```
orlandohohmeier: Any chance we can change this function to only "replace" `NaN` with `null`?  It kinda feels wrong to have a function that parses the input into an integer if the input isn't an empty string esp. as you can have strings (e.g. "e2") that still result in `NaN`.

plugins/services/src/js/reducers/serviceForm/Containers.js
```diff
@@ -609,19 +560,15 @@ module.exports = {
...
+        this.healthCheckState[index] = {};
       }
 
-      newState[index].healthChecks = MultiContainerHealthChecks.call(
-        this.healthChecks[index],
-        newState[index].healthChecks,
-        {type, path: path.slice(2), value}
+      newState[index].healthCheck = multiContainerHealthFormReducer.call(
+        this.healthCheckState[index],
+        newState[index].healthCheck || {},
```
orlandohohmeier: Let's change `multiContainerHealthFormReducer` to handle undefined healthChecks instead, of defaulting the health check to an empty object here.

plugins/services/src/js/reducers/serviceForm/Containers.js
```diff
@@ -456,19 +411,15 @@ module.exports = {
...
-    if (field === 'healthChecks') {
-      if (newState[index].healthChecks == null) {
-        newState[index].healthChecks = [];
-      }
-
-      if (this.healthChecks[index] == null) {
-        this.healthChecks[index] = {};
+    if (field === 'healthCheck') {
+      if (this.healthCheckState[index] == null) {
+        this.healthCheckState[index] = {};
```
orlandohohmeier: *Note to self: We need to refactor the container reducer as those indexes as well as the crazy number of conditions makes it really hard to follow through.*

plugins/services/src/js/reducers/serviceForm/MultiContainerHealthChecks.js
```diff
@@ -3,79 +3,423 @@ import {
...
+ * @param {Array} path - The path prefix to the transaction
+ * @returns {Array} - Returns an array with the new transactions to append
+ */
+function parseCommandHealthCheck(healthCheck, path) {
+  const memo = [];
+  const {command={}} = healthCheck;
+
+  if (command.shell != null) {
+    memo.push(new Transaction(
+      path.concat(['command', 'shell']),
```
orlandohohmeier: Let's rename this to `command.type` to make obvious that this is toggling the command type

plugins/services/src/js/reducers/serviceForm/MultiContainerHealthChecks.js
```diff
@@ -3,79 +3,423 @@ import {
...
+      true,
+      SET
+    ));
+
+    memo.push(new Transaction(
+      path.concat(['command', 'string']),
```
orlandohohmeier: Please rename this one to `command.value` to make this view agnostic.

plugins/services/src/js/reducers/serviceForm/Containers.js
mlunoe: Yes, in general we need to split it up into multiple reducers, that we combine instead of this mega-file of logic

plugins/services/src/js/reducers/serviceForm/MultiContainerHealthChecks.js
mlunoe: Let's make sure that we are following the app configuration here, and don't something different. We've been burnt by this in the passed.

plugins/services/src/js/reducers/serviceForm/MultiContainerHealthChecks.js
mlunoe: Let's make sure that we are following the app configuration here, and don't something different. We've been burnt by this in the passed.

plugins/services/src/js/reducers/serviceForm/MultiContainerHealthChecks.js
orlandohohmeier: Where exactly do you see the problem/risk? The app and pod spec have a very different structure thus we shouldn't get burnt if the respective parses/reducers look different. If we want to align them, as initially planned, we need to finally move away from mirroring the app and pod spec JSON and find a better/abstract representation for the data. I'll create a JIRA to keep track of that.

plugins/services/src/js/reducers/serviceForm/MultiContainerHealthChecks.js
wavesoft: Good point. Also renamed it to `parseIntOrNull` for clarity.

plugins/services/src/js/reducers/serviceForm/MultiContainerHealthChecks.js
wavesoft: Renamed to `command.type`.

@mlunoe , I agree with @orlandohohmeier that whatever we keep in the batch should **never** be considered as an 1:1 representation of the data. The batch should contain whatever high-level data structures we might need on the UI in order to properly display and process the data. I also agree that we don't have a clean way of describing the data structures inside the batch, but this is something we have to address separately.

plugins/services/src/js/reducers/serviceForm/MultiContainerHealthChecks.js
wavesoft: Renamed to `command.value`

plugins/services/src/js/reducers/serviceForm/Containers.js
wavesoft: Actually, it's already handled, so I will just remove the `|| {}`

rcorral:
@orlandohohmeier could you review this again please?

PR #1726, opened 2017-01-04
Rename page header components

jfurrow:
This PR renames the `NewPageHeader*` components to `PageHeader*` and removes two related but unused components.


wavesoft:
I love cleanups! Looks good! ✨ 💯

PR #1734, opened 2017-01-05
DCOS-12415: Fixed clipped descenders in elements that use text-overflow

jfurrow:
This PR resets the `line-height` for all elements that use the `text-overflow` property. This prevents the descenders from being clipped in some browsers.

Before:

After:

Before:

After:


wavesoft:
Looks good ✨  💯

rcorral:
run integration tests

PR #1736, opened 2017-01-05
[Logs 1 of 2] DCOS-12769: Make paging in log view a bit snappier

mlunoe:
Logs 1 of 2 PR: https://github.com/dcos/dcos-ui/pull/1736
Logs 2 of 2 PR: https://github.com/dcos/dcos-ui/pull/1745

This PR also resolves: DCOS-12737

This PR fixes a few issues related to the same thing - making the paging become smoother:
- It initializes the logs no matter what response comes from the server, so we do not see 'Empty Log' when there is an error.
- It stop debouncing `componentDidUpdate` in  the log view, so it actually goes to bottom initially
- It removes log window so the position calculation in `componentDidUpdate` continues to work
- It lets the container handle when to retrieve a previous page, instead of assuming an ongoing request when at the top.

Initially, the paging goes pretty fast, but eventually the journald bug manifests and loading becomes increasingly slower. This is not handled by this PR.

### DIFF
--- a/plugins/services/src/js/components/LogView.js
+++ b/plugins/services/src/js/components/LogView.js
@@ -176,19 +176,12 @@ class LogView extends React.Component {
   }
 
   checkIfCloseToTop(container) {
-    const {closeToTop} = this.state;
     const distanceFromTop = DOMUtils.getDistanceFromTop(container);
-    if (distanceFromTop < 100 && !closeToTop) {
-      this.setState({closeToTop: true}, () => {
-        const {hasLoadedTop, fetchPreviousLogs} = this.props;
-        if (!hasLoadedTop) {
-          fetchPreviousLogs();
-        }
-      });
-    }
-
-    if (distanceFromTop > 100 && closeToTop) {
-      this.setState({closeToTop: false});
+    if (distanceFromTop < 100) {
+      const {hasLoadedTop, fetchPreviousLogs} = this.props;
+      if (!hasLoadedTop) {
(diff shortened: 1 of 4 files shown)


mlunoe:
Run unit tests

mlunoe:
Run unit tests

plugins/services/src/js/components/LogView.js
```diff
@@ -28,18 +28,12 @@ class LogView extends React.Component {
...
 
     this.handleWindowResize = Util.debounce(
       this.handleWindowResize.bind(this), 100
     );
-
-    // Since we can receive a lot of updates (from a stream potentially),
-    // let's debounce this function to not call it too often
-    this.componentDidUpdate = Util.debounce(
-      this.componentDidUpdate.bind(this), 100
-    );
```
jfurrow: Are we no longer concerned about receiving a lot of updates at once?

jfurrow:
Works well!

plugins/services/src/js/components/LogView.js
mlunoe: I am working on a performance PR that will take care of this. More or less I am using `shouldComponentUpdate` a little more actively

jfurrow: Okay cool, thanks for the clarification!

mlunoe: Removed the commit from this PR

PR #1739, opened 2017-01-06
Increase width of Service ID input field on create flow

leemunroe:
Increases the width of the Service ID field. Why?

* More space for the help text so it doesn't wrap
* Vertical alignment with the fields further down
* More space when working with IDs that contain long group names
* Instances doesn't need to be that wide

Before
image

After
image

cc @jfurrow


wavesoft:
Looks good :+1:

PR #1740, opened 2017-01-06
Fixing conditional rendering on pod sections

wavesoft:
I am quite confident this was fixed in one of my past PRs but I could have forgotten to push... this PR makes sure no `0`s appear when a conditional rendering case fails to match because of number 0.

### DIFF
--- a/plugins/services/src/js/service-configuration/PodContainerConfigSection.js
+++ b/plugins/services/src/js/service-configuration/PodContainerConfigSection.js
@@ -65,7 +65,7 @@ const PodContainerConfigSection = ({containerConfig, appConfig, onEditClick, ind
       </ConfigurationMapRow>
 
       {/* Resources */}
-      {fields.resources.cpus && (
+      {!!fields.resources.cpus && (
         <ConfigurationMapRow>
           <ConfigurationMapLabel>CPUs</ConfigurationMapLabel>
           <ConfigurationMapValue value={fields.resources.cpus} />
@@ -74,7 +74,7 @@ const PodContainerConfigSection = ({containerConfig, appConfig, onEditClick, ind
             tabViewID={tabViewID} />
         </ConfigurationMapRow>
       )}
-      {fields.resources.mem && (
+      {!!fields.resources.mem && (
         <ConfigurationMapRow>
           <ConfigurationMapLabel>Memory</ConfigurationMapLabel>
           <ConfigurationMapSizeValue value={fields.resources.mem} />
@@ -83,7 +83,7 @@ const PodContainerConfigSection = ({containerConfig, appConfig, onEditClick, ind
             tabViewID={tabViewID} />
         </ConfigurationMapRow>
       )}
-      {fields.resources.disk && (
+      {!!fields.resources.disk && (
         <ConfigurationMapRow>
           <ConfigurationMapLabel>Disk</ConfigurationMapLabel>
           <ConfigurationMapSizeValue value={fields.resources.disk} />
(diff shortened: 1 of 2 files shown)


nLight:
Good to go!

plugins/services/src/js/service-configuration/PodContainerConfigSection.js
```diff
@@ -65,7 +65,7 @@ const PodContainerConfigSection = ({containerConfig, appConfig, onEditClick, ind
       </ConfigurationMapRow>
 
       {/* Resources */}
-      {fields.resources.cpus && (
+      {!!fields.resources.cpus && (
```
mlunoe: Aren't all of these numbers? This would not render `0` values.

wavesoft: And that's correct behaviour :)

mlunoe: Ok, I would prefer us casting it to booleans with `Boolean()`, but in general, I am still not sold on this pattern.
It was not introduced here, so it should not hold this PR up, but I feel like this was just introduced before we had a discussion about it.

rcorral: I prefer the use of `Boolean` over the bang bang. And would echo @mlunoe's comment and add, that, we should try to be mindful about new patterns and add them to our coding styles first rather than add them to the codebase and then to the code styles.

mlunoe: Fixed them for you @wavesoft

mlunoe:
Run unit tests

mlunoe:
Run unit tests

mlunoe:
Run integration tests

mlunoe:
Run unit tests

plugins/services/src/js/service-configuration/PodContainerConfigSection.js
orlandohohmeier: I'm a little to late to the game, but let's please always try to avoid type casting esp. if it's changing the type representation as the result could be very different to the expected outcome. 

*Examples*
```
Boolean("0") === true
Boolean(0) === false
Boolean(Number("")) === true
Boolean(String(Number(null))) === true
Boolean(parseInt("1.1.1")) === true
Boolean(Number("1.1.1")) === false
// ...
```

If possible please always prefer using an expressive function/util that properly checks the values types. In this case we could use the `isNumberInRange` util as we actually want to check if it a number an if that number is greater than zero.

We might also want to add an eslint rule that at least warns if someone if casting type due to the named reasons.

/cc @dcos/frontend

nLight: TL;DR use `get<Section>` or extract private components that is. Let's ban this syntax.

... or just write those private components and put all the logic there. To me this discussion went to far already. I'd say we can use the syntax only and only if we have a boolean flag. Otherwise we always will put a type checking / casting function that would nullify all the beauty of the approach.

wavesoft: @mlunoe thanks! However I was going to go with something that satisfies both @nLight and @orlandohohmeier and do something like:

```js
function isValueVisible(value) {
   return (isNumber(value) && (parseInt(value) !== 0) || isDefined(value);
}

...

      {/* Resources */}
      {isValueVisible(fields.resources.cpus) && (
        <ConfigurationMapRow>
          <ConfigurationMapLabel>CPUs</ConfigurationMapLabel>
          <ConfigurationMapValue value={fields.resources.cpus} />
          <ConfigurationMapEditAction
```

mlunoe: Then let's create a follow-up PR for this

PR #1741, opened 2017-01-06
Display Pod container image and forcePull param

nLight:
the data is on container level not on the app level.

screen shot 2017-01-06 at 16 08 01


wavesoft:
Looks good ✨  👍 

However I think that we should not display all these "Not Configured" fields, since they are bloating the display without adding any useful information. But this can be done in a separate PR.

PR #1742, opened 2017-01-06
Fix TaskTable column responsiveness

jfurrow:
This PR fixes the following:
* the `TaskTable`'s key for the `host` column was incorrect, so the class wasn't being applied properly
* the responsiveness has been adjusted so that the `host` column appears at screen size medium and up, rather than large and up


nLight:
\o/ :shipit:

PR #1746, opened 2017-01-06
Fix propType warning in TabButtonList

jfurrow:


rcorral:
Merging as this is tiny.

PR #1721, opened 2017-01-04
DCOS-12405: Conditionally show declined offers funnel &table

jfurrow:
This PR hides the declined offers funnel & table data from a service's debug tab until the service has been waiting or over 5 minutes.

PR #1749, opened 2017-01-06
DCOS-12892: Display declined offers for ServiceTree instances in ServicesTable

jfurrow:
This PR adds the declined offers warning for groups of apps if any apps inside the group have declined offers.


plugins/services/src/js/components/ServiceStatusWarning.js
```diff
@@ -7,29 +7,48 @@ import Icon from '../../../../../src/js/components/Icon';
...
+
+      if (appCount > 0) {
+        const appsNoun = StringUtil.pluralize('service', appCount);
+
+        return this.getTooltip(`DC/OS is waiting for resources and is unable to complete the deployment of ${appCount} ${appsNoun} in this group.`);
+      }
+    } else {
+      const queue = item.getQueue();
+
+      if (queue != null
```
mlunoe: No need to check if `queue != null`

plugins/services/src/js/components/ServiceStatusWarning.js
```diff
@@ -7,29 +7,48 @@ import Icon from '../../../../../src/js/components/Icon';
...
+        && DeclinedOffersUtil.shouldDisplayDeclinedOffersWarning(queue)) {
+        const waitingSince = DateUtil.strToMs(queue.since);
```
mlunoe: Don't we need same calculation as here? https://github.com/dcos/dcos-ui/pull/1721/files#diff-05d538b8dad1163c5f63426574d0d46aR223

plugins/services/src/js/components/ServiceStatusWarning.js
```diff
@@ -7,29 +7,48 @@ import Icon from '../../../../../src/js/components/Icon';
...
-        <Tooltip
-          content={this.getTooltipContent(timeWaiting)}
-          width={250}
-          wrapText={true}
-          wrapperClassName="tooltip-wrapper status-waiting-indicator">
-          <Icon color="red" id="yield" size="mini" />
-        </Tooltip>
-      );
+
+    if (item instanceof ServiceTree) {
```
mlunoe: Couldn't we simplify this a bit and try to avoid those nested `if`-statements?
Didn't test this and not sure we actually need the `instance`-check if we could hit groups here, maybe it makes sense to have `getQueue` implemented on the `Group` struct to return `null` as we do other places so we don't need to check the instance type.
```js
class ServiceStatusWarning extends Component {
  getTooltip(content) {
    return (
      <Tooltip
        content={content}
        width={250}
        wrapText={true}
        wrapperClassName="tooltip-wrapper status-waiting-indicator">
        <Icon color="red" id="yield" size="mini" />
      </Tooltip>
    );
  }

  render() {
    const {item} = this.props;

    if (getAppsWithDeclinedOffersFromServiceTree(item).length > 0) {
      const appsNoun = StringUtil.pluralize('service', appCount);

      return this.getTooltip(`DC/OS is waiting for resources and is unable to complete the deployment of ${appCount} ${appsNoun} in this group.`);
    }

    if (item instanceof Service &&
      shouldDisplayDeclinedOffersWarning(item.getQueue())) {
      const queue = item.getQueue();
      const waitingSince = DateUtil.strToMs(queue.since);
      const timeWaiting = Date.now() - waitingSince;

      return this.getTooltip(`DC/OS has been waiting for resources and is unable to complete this deployment for ${DateUtil.getDuration(timeWaiting, null)}.`);
    }

    return <noscript />;
  }
};
```

jfurrow: In order for this to work, we would need `getAppsWithDeclinedOffersFromServiceTree` to accept anything and return an empty array of it's not an instance of `ServiceTree`. We would also need to perform the same logic more than once (`item.getQueue()`). To me that doesn't really seem worth it. What do you think?

plugins/services/src/js/components/ServiceStatusWarning.js
jfurrow: Good catch, removed!

plugins/services/src/js/components/ServiceStatusWarning.js
jfurrow: Yes, good call! I'll make the change in that PR and then handle the conflicts when they arise.

jfurrow:

plugins/services/src/js/components/ServiceStatusWarning.js
jfurrow: I created a method called `#getTimeWaiting` in `DeclinedOffersUtil` to handle this. It's in this PR: https://github.com/dcos/dcos-ui/pull/1721

plugins/services/src/js/components/ServiceStatusWarning.js
mlunoe: Alright, but could we at least get rid of that `getTooltipContent` function, that doesn't really do anything for us we could not just do in this function?

jfurrow: Sure, I'll remove it. I created it to avoid duplicating the Tooltip configuration since the only thing that changes is the text content.

mlunoe: I am not talking about `getTooltip`, but `getTooltipContent`!

jfurrow: My apologies! I removed it :)

mlunoe: No worries. Looks good man!

plugins/services/src/js/utils/DeclinedOffersUtil.js
```diff
@@ -35,6 +36,26 @@ const constraintsReducer = (accumulator, constraint) => {
 };
 
 const DeclinedOffersUtil = {
+  getAppsWithDeclinedOffersFromServiceTree(serviceTree = {}) {
+    if (!serviceTree.list || serviceTree.list.length === 0) {
```
orlandohohmeier: Please don't operate on internal "private" properties and use the public interface instead!

plugins/services/src/js/utils/DeclinedOffersUtil.js
```diff
@@ -35,6 +36,26 @@ const constraintsReducer = (accumulator, constraint) => {
...
+      return [];
+    }
+
+    return serviceTree.list.reduce((accumulator, item) => {
+      if (item instanceof ServiceTree) {
+        accumulator = accumulator.concat(
+          DeclinedOffersUtil.getAppsWithDeclinedOffersFromServiceTree(item)
```
orlandohohmeier: Please use `Tree.prototype.reduce()` instead of implementing your own recursive function.

plugins/services/src/js/components/ServiceStatusWarning.js
```diff
@@ -7,27 +7,42 @@ import Icon from '../../../../../src/js/components/Icon';
...
+      const appsWithDeclinedOffers = DeclinedOffersUtil
```
orlandohohmeier: There is no need of a new util if you only want to reduce the service tree to a list of apps with declined offers...

*List*
```
    const appsWithDeclinedOffers = item.filterItems(function (item) {
        if (!(item instanceof ServiceTree)) {
          return DeclinedOffersUtil
            .shouldDisplayDeclinedOffersWarning(item.getQueue());
        }

        return false;
      }).flattenItems();
```

plugins/services/src/js/utils/DeclinedOffersUtil.js

plugins/services/src/js/utils/DeclinedOffersUtil.js
jfurrow: Done! I removed this util method in favor of the one-off filter method.

plugins/services/src/js/components/ServiceStatusWarning.js
jfurrow: Done!

jfurrow:
@orlandohohmeier Thanks for catching this stuff, I've addressed all of your comments

plugins/services/src/js/components/ServiceStatusWarning.js
```diff
@@ -7,27 +7,52 @@ import Icon from '../../../../../src/js/components/Icon';
...
+        content={content}
+        width={250}
+        wrapText={true}
+        wrapperClassName="tooltip-wrapper status-waiting-indicator">
+        <Icon color="red" id="yield" size="mini" />
+      </Tooltip>
+    );
   }
+
+  getAppsWithDeclinedOffers(serviceTree) {
```
orlandohohmeier: Let's either inline the service tree call or move it to a util.

jfurrow: I don't understand what you mean by this, are you saying I shouldn't add a `getAppsWithDeclinedOffers` method to this component? I think that's what you mean, so that's what I did here: https://github.com/dcos/dcos-ui/pull/1749/commits/94218f6b7b584bac22e987d1b21369622ebfb383

Could you expand more on the reasoning for this? To me the method call made the logic easier to follow.

MatApple:
@orlandohohmeier - going to merge because I know you're busy. I'll take on the additional changes if what john did doesn't satisfy your request.

PR #1753, opened 2017-01-10
[1 of 2: Service Create Modal] DCOS-12463: Remove full-screen styling when JSON-only mode is active

jfurrow:
Before:

After:


wavesoft:
Feels a bit empty, but it's definitely cleaner, nice! ✨  💯

PR #1755, opened 2017-01-10
DCOS-12642: Fixing lost characters when typing in Jobs filter

wavesoft:
This PR removes the `QueryParamsMixin` and `SaveStateMixin` (for consistency) from the `JobsTab`, and fixes an issue where letters were dropped when typing in the Jobs search filter.

In detail:
* `QueryParamsMixin` is not using the `nextProps` coming from the `componentWillReceiveProps` in order to synchronize the component state from the query parameters, and
* Mix-ins are considered deprecated, therefore I didn't fix the problem there, rather I re-used the same approach as from the `ServicesContainer`
* `SaveStateMixin` was racing the `QueryParamsMixin` and causing another set of problems. However I do agree that we are losing some of the functionality and perhaps has to be re-introduced in another way (without using mix-ins).

### DIFF
--- a/src/js/pages/jobs/JobsTab.js
+++ b/src/js/pages/jobs/JobsTab.js
@@ -11,13 +11,12 @@ import FilterHeadline from '../../components/FilterHeadline';
 import JobsBreadcrumbs from '../../components/breadcrumbs/JobsBreadcrumbs';
 import JobsTable from './JobsTable';
 import JobSearchFilter from '../../components/JobSearchFilter';
+import JobFilterTypes from '../../constants/JobFilterTypes';
 import JobFormModal from '../../components/modals/JobFormModal';
 import JobTree from '../../structs/JobTree';
 import Loader from '../../components/Loader';
 import Page from '../../components/Page';
-import QueryParamsMixin from '../../mixins/QueryParamsMixin';
 import ServiceFilterTypes from '../../../../plugins/services/src/js/constants/ServiceFilterTypes';
-import SaveStateMixin from '../../mixins/SaveStateMixin';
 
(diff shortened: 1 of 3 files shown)


jfurrow:
Works great!

src/js/pages/jobs/JobsTab.js
```diff
@@ -81,11 +72,17 @@ class JobsTab extends mixin(StoreMixin, QueryParamsMixin, SaveStateMixin) {
     this.setState({isJobFormModalOpen: true});
   }
 
-  handleFilterChange(filterValues, filterType) {
-    var stateChanges = Object.assign({}, this.state);
-    stateChanges[filterType] = filterValues;
+  handleFilterChange(filterValue) {
+    const {router} = this.context;
+    const {location: {pathname}} = this.props;
+    const query = {};
```
nLight: nit: You could do it right here `const query = {[JobFilterTypes.TEXT]: filterValue};`

src/js/pages/jobs/JobsTab.js
```diff
@@ -104,6 +101,17 @@ class JobsTab extends mixin(StoreMixin, QueryParamsMixin, SaveStateMixin) {
     this.setState(state, this.resetFilterQueryParams);
   }
 
+  updateFromProps(props) {
+    const {location={query:{}}} = props;
```
nLight: nit: spaces around `=`

src/js/pages/jobs/JobsTab.js
wavesoft: Ow, neat 😍

PR #1758, opened 2017-01-10
DCOS-12661: Differentiate between Host and Container Port

jfurrow:
* Uses `Host Port` for the column heading rather than the vague `Port`
* Adds a `Container Port` column when network type is anything but `HOST`
* Fixes the service address string concatenation when network type is anything but `HOST`

### DIFF
--- a/plugins/services/src/js/service-configuration/ServiceNetworkingConfigSection.js
+++ b/plugins/services/src/js/service-configuration/ServiceNetworkingConfigSection.js
@@ -12,6 +12,18 @@ import ServiceConfigUtil from '../utils/ServiceConfigUtil';
 import ServiceConfigBaseSectionDisplay from './ServiceConfigBaseSectionDisplay';
 import {findNestedPropertyInObject} from '../../../../../src/js/utils/Util';
 
+const getNetworkType = (networkType, appDefinition) => {
+  networkType = networkType || Networking.type.HOST;
+  const networkName = findNestedPropertyInObject(
+    appDefinition,
+    'ipAddress.networkName'
+  );
+
+  return networkName != null ?
+    Networking.type.USER :
+    networkType;
+};
+
 class ServiceNetworkingConfigSection extends ServiceConfigBaseSectionDisplay {
   /**
    * @override
@@ -43,17 +55,7 @@ class ServiceNetworkingConfigSection extends ServiceConfigBaseSectionDisplay {
         {
           key: 'container.docker.network',
           label: 'Network Type',
-          transformValue(networkType, appDefinition) {
-            networkType = networkType || Networking.type.HOST;
-            const networkName = findNestedPropertyInObject(
-              appDefinition,
-              'ipAddress.networkName'
-            );
-
-            return networkName != null ?
-              Networking.type.USER :
-              networkType;
-          }
+          transformValue: getNetworkType
         },
         {
           key: 'ipAddress.networkName',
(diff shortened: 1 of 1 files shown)


nLight:
You stole my task 😱

plugins/services/src/js/service-configuration/ServiceNetworkingConfigSection.js
```diff
@@ -12,6 +12,18 @@ import ServiceConfigUtil from '../utils/ServiceConfigUtil';
 import ServiceConfigBaseSectionDisplay from './ServiceConfigBaseSectionDisplay';
 import {findNestedPropertyInObject} from '../../../../../src/js/utils/Util';
 
+const getNetworkType = (networkType, appDefinition) => {
```
nLight: I was about to ask why not using `function` here and decided to research a bit...
Here what I've found and I don't know how to live my life anymore

http://stackoverflow.com/a/28135120

tea


Do we really want a top level arrow functions bound to a `const`?
// cc @rcorral @orlandohohmeier

wavesoft: It's true that functions should not be used as a "fancier syntax", but have a reason for being there.
In this particular case I have to agree with @nLight , and a `function` approach would be better:

```js
function getNetworkType(networkType, appDefinition) {
```

plugins/services/src/js/service-configuration/ServiceNetworkingConfigSection.js
```diff
@@ -73,13 +75,20 @@ class ServiceNetworkingConfigSection extends ServiceConfigBaseSectionDisplay {
...
+              findNestedPropertyInObject(
+                appDefinition, 'container.docker.network'
+              ),
+              appDefinition
+            );
+
             const containerPortMappings = findNestedPropertyInObject(
               appDefinition, 'container.docker.portMappings'
             );
             if ((portDefinitions == null || portDefinitions.length === 0) &&
```
nLight: That conditional causes a bug, that even when we have BRIDGE/USER network it still works like it's a HOST network. The thing is that Marathon will add `portDefinitinos` section with a `servicePort` therefore this conditional always will be true and you will see something like

screen shot 2017-01-11 at 16 18 08

I.e. take this JSON, launch an app and go to it's configuration tab and notice that even you don't see `portDefinitions` in the form it will be in the appDefinition you are getting from the server ¯\_(ツ)_/¯

so in my case I went for just checking
```javascript
if( containerPortMappings != null && containerPortMappings.length !== 0` ) {
...
}
```

```json
{
  "id": "/dima",
  "instances": 1,
  "cpus": 1,
  "mem": 128,
  "disk": 0,
  "gpus": 0,
  "backoffSeconds": 1,
  "backoffFactor": 1.15,
  "container": {
    "type": "DOCKER",
    "docker": {
      "image": "nginx",
      "network": "BRIDGE",
      "portMappings": [
        {
          "name": "memeememme",
          "hostPort": 8091,
          "containerPort": 2020,
          "protocol": "tcp",
          "servicePort": 10001,
          "labels": {
            "VIP_0": "/dima:2020"
          }
        }
      ],
      "privileged": false,
      "forcePullImage": false
    }
  },
  "upgradeStrategy": {
    "minimumHealthCapacity": 1,
    "maximumOverCapacity": 1
  },
  "unreachableStrategy": {
    "inactiveAfterSeconds": 900,
    "expungeAfterSeconds": 604800
  },
  "killSelection": "YoungestFirst",
  "requirePorts": true
}
```

plugins/services/src/js/service-configuration/ServiceNetworkingConfigSection.js
orlandohohmeier: Even though that it's more a stylistic question than actually an issue in this case,  I think that we should revisit our decision to allow arrow functions (actually only in cases where one want to "bind" `this`) and prohibit the use of arrow functions again, after reading through the Stack Overflow discussion.
Let's discuss this in our next team sync and decide wether we want to allow or prohibit the use and adjust our linting rules accordingly. 

/cc @dcos/frontend

jfurrow: Thanks for the research & feedback, @nLight @wavesoft @orlandohohmeier! That was a good discussion to read. I switched it to a regular `function`.

plugins/services/src/js/service-configuration/ServiceNetworkingConfigSection.js
jfurrow: Good catch! I implemented the conditional you provided and it works well. Thanks!

jfurrow:
@nLight All comments addressed!

orlandohohmeier:
Looks like we're good to merge this one once we've verified that the integration test pass locally.

PR #1759, opened 2017-01-10
DCOS-12740: Move task resources to Configuration section

jfurrow:
This PR moves the resource allocation into the Configuration section.

~~I'm adding `hold merge` because Configuration doesn't seem like the best place for this, so I'm waiting for clarification from @ashenden.~~ This has been blessed!


mlunoe:
Alright, makes sense. Looks good and works great, so you get my label

jfurrow:

PR #1763, opened 2017-01-11
DCOS-12971: Add heading to Containers section

jfurrow:
This PR adds a heading to the Containers section in a multi-container form.

Before:

After:


mlunoe:
Looks good, works great

PR #1767, opened 2017-01-11
DCOS-12449: Add Public IP to cluster details page

jfurrow:
This PR adds the `General` heading and a `Public IP` entry to the cluster details page.

### DIFF
--- a/src/js/pages/system/OverviewDetailTab.js
+++ b/src/js/pages/system/OverviewDetailTab.js
@@ -49,7 +49,7 @@ class OverviewDetailTab extends mixin(StoreMixin) {
       },
       {
         name: 'metadata',
-        events: ['dcosBuildInfoChange', 'dcosSuccess']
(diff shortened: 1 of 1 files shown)


src/js/pages/system/OverviewDetailTab.js
```diff
@@ -123,6 +129,18 @@ class OverviewDetailTab extends mixin(StoreMixin) {
     }];
   }
 
+  getPublicIP() {
+    const metadata = MetadataStore.get('metadata');
+
+    if ((typeof metadata !== 'object') ||
```
mlunoe: This should handle the same cases:
```js
const publicIPs = findNestedPropertyInObject(
  MetadataStore.get('metadata'),
  'PUBLIC_IPV4'
) || [];
if (!publicIPs.length) {
  return null;
}

return publicIPs;
```

jfurrow: Done!

PR #1769, opened 2017-01-11
DCOS-12867: Fix un-minified JS in production build

jfurrow:
This PR removes the desctructuring assignment which was causing `UglifyJS` to throw an error. This file is not transpiled so we can't use ES6 features here.

### DIFF
--- a/src/resources/grammar/SearchDSL.jison
+++ b/src/resources/grammar/SearchDSL.jison
@@ -2,9 +2,9 @@
 /*              common filter description language.                           */
 
 %{
-const DSLParserUtil = require('../../js/utils/DSLParserUtil');
-const Merge = DSLParserUtil.Merge;
-const Operator = DSLParserUtil.Operator;
+var DSLParserUtil = require('../../js/utils/DSLParserUtil');
+var Merge = DSLParserUtil.Merge;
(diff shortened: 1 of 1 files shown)


src/resources/grammar/SearchDSL.jison
```diff
@@ -2,7 +2,9 @@
 /*              common filter description language.                           */
 
 %{
-const {Merge, Operator} = require('../../js/utils/DSLParserUtil');
+const DSLParserUtil = require('../../js/utils/DSLParserUtil');
```
mlunoe: Should we use `var`s here as well, to be safe?

rcorral: done.

jfurrow: Sorry I meant to reply to this — we're actually good to use `const` http://caniuse.com/#search=const

mlunoe: Yes, but no reason to just exclude everything below for that

jfurrow: Yeah that's fair!

jfurrow:

mlunoe:
Run integration tests

mlunoe:
Works great dude!

PR #1773, opened 2017-01-11
DCOS-9159: Disable tasks' restart button while its service is deploying

jfurrow:
This PR disables the task restart button when the service is deploying.

Here's the tooltip that appears:

### DIFF
--- a/plugins/services/src/js/containers/tasks/TasksView.js
+++ b/plugins/services/src/js/containers/tasks/TasksView.js
@@ -10,6 +10,7 @@ import FilterInputText from '../../../../../../src/js/components/FilterInputText
 import GraphQLTaskUtil from '../../utils/GraphQLTaskUtil';
 import Icon from '../../../../../../src/js/components/Icon';
 import SaveStateMixin from '../../../../../../src/js/mixins/SaveStateMixin';
+import Service from '../../structs/Service';
 import ServiceStatusTypes from '../../constants/ServiceStatusTypes';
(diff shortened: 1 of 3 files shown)


mlunoe:
When I am not hovering it looks like this: 

Is that intended? It looks kinda weird to me

jfurrow:
@mlunoe I suspect your version of CNVS is out of date, this is what the button's disabled static state should look like:

plugins/services/src/js/containers/tasks/TasksView.js
```diff
@@ -143,42 +145,63 @@ class TasksView extends mixin(SaveStateMixin) {
...
     // Using Button's native "disabled" prop prevents onMouseLeave from
     // being correctly called, preventing correct dismissal of the Tooltip.
     //
-    // So we manually disable the button.
+    // So we overwrite the click handlers manually.
+    let handleRestartClick = function () {};
     let handleStopClick = function () {};
 
+    if (!isDeploying) {
+      handleRestartClick = this.handleStopClick.bind(this, 'restart');
```
mlunoe: This might also need a rename:

jfurrow: Thanks! Fixed!

jfurrow:

jfurrow:
Rebased with integration test fixes 🎉 v2

philipnrmn:
Nice! Thanks for you patience on the merge @jfurrow :)

PR #1774, opened 2017-01-12
Fix Mount's wrapper propType

jfurrow:
This PR fixes the `Mount`'s `wrapper` `propTypes` definition. I believe this is the intended behavior, as the `wrapper` prop is passed as the first argument to `React.createElement` which can only be a string, function, or class (in this case, `PropTypes.func` covers both a function and class).

cc @mlunoe @orlandohohmeier

### DIFF
--- a/foundation-ui/mount/Mount.js
+++ b/foundation-ui/mount/Mount.js
@@ -102,7 +102,7 @@ Mount.propTypes = {
   limit: PropTypes.number,
   children: PropTypes.element,
   type: PropTypes.string.isRequired,
-  wrapper: PropTypes.node
+  wrapper: PropTypes.oneOfType([PropTypes.string, PropTypes.func])
 };
 
 module.exports = Mount;


wavesoft:
Good catch @jfurrow ! ✨

foundation-ui/mount/Mount.js
```diff
@@ -102,7 +102,7 @@ Mount.propTypes = {
   limit: PropTypes.number,
   children: PropTypes.element,
   type: PropTypes.string.isRequired,
-  wrapper: PropTypes.node
+  wrapper: PropTypes.oneOfType([PropTypes.string, PropTypes.func])
```
mlunoe: Yes, this is what React does itself: https://github.com/facebook/react/blob/c78464f8ea9a5b00ec80252d20a71a1482210e57/src/isomorphic/classic/element/ReactElementValidator.js#L186

PR #1777, opened 2017-01-12
DCOS-13184: Moving cleanServiceJSON to the correct place

wavesoft:
This PR tries to move the `cleanServiceJSON` to the most appropriate place, in order for it to be called only when it makes sense and not every time.

Ideally, I would put it in `createServiceFromResponse` , but I thought that `Application` **can** contain additional ("state"?) information, but `ApplicationSpec` **should not**.


wavesoft:
run integration tests

PR #1778, opened 2017-01-12
DCOS-10182: Adding deprecation warning for `uris`

wavesoft:
This PR introduces a validator that emits an error when `uris` and `fetch` is defined, prompting the users to use `fetch` in favour of `uris`.

### DIFF
--- a/plugins/services/src/js/validators/MarathonAppValidators.js
+++ b/plugins/services/src/js/validators/MarathonAppValidators.js
@@ -137,6 +137,22 @@ const MarathonAppValidators = {
 
     // No errors
     return [];
+  },
+
+  /**
+   * @param {Object} app - The data to validate
+   * @returns {Array} Returns an array with validation errors
+   */
+  mustNotContainUris(app) {
+    if (ValidatorUtil.isDefined(app.uris)) {
+      const message = '`uris` are deprecated. Please use `fetch` instead';
+
+      return [
+        {path: ['uris'], message}
+      ];
+    }
+
+    return [];
   }
 };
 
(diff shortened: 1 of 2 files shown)


plugins/services/src/js/validators/MarathonAppValidators.js
```diff
@@ -137,6 +137,23 @@ const MarathonAppValidators = {
...
     return [];
+  },
+
+  /**
+   * @param {Object} app - The data to validate
+   * @returns {Array} Returns an array with validation errors
+   */
+  mustNotContainUris(app) {
+    if (ValidatorUtil.isDefined(app.uris) &&
+        ValidatorUtil.isDefined(app.fetch)) {
```
nLight: So you are checking if both `uris` and `path` are present?

wavesoft: Yes. Because:

* If only `uris` is present, that's a a legit old way of pasting stuff (still accepted by mesos) so it must be accepted
* If only `fetch` is present, that's the correct way of doing things
* Both `fetch` and `uris` will appear ONLY if you are editing a service and you try to add `uris`. This time, marathon will not accept a JSON with both fields present. Therefore I am prompting the user to remove the `uris` field in favor of `fetch`

nLight: 👍

orlandohohmeier:
LGTM 💯 ✨

orlandohohmeier:
run integration tests

orlandohohmeier:
run integration tests

nLight:
run integration tests

nLight:
run integration tests

rcorral:
run integration tests

PR #1780, opened 2017-01-12
Prevent Mount component from wrapping singular registered child components

jfurrow:
This PR prevents `Mount` from wrapping a singular registered component component.

This is necessary for me to predictably render the grid of options in the create service modal. I'm fairly confident it's desirable everywhere because it doesn't force any component to be wrapped, unless there are more than 1 components to render, or the owner of `Mount` has explicitly declared it.

cc @orlandohohmeier I discussed this offline with @mlunoe, but it would be nice if you could review this as well.

### DIFF
--- a/foundation-ui/mount/__tests__/Mount-test.js
+++ b/foundation-ui/mount/__tests__/Mount-test.js
@@ -148,8 +148,7 @@ describe('Mount', function () {
       </Mount>
     );
 
-    expect(renderer.getRenderOutput().props.children[0].props)
-        .toEqual({message:'hello world'});
+    expect(renderer.getRenderOutput().props).toEqual({message:'hello world'});
   });
 
 });
--- a/foundation-ui/utils/ReactUtil.js
+++ b/foundation-ui/utils/ReactUtil.js
@@ -15,6 +15,10 @@ const ReactUil = {
       return null;
     }
 
+    if (Array.isArray(elements) && elements.length === 1 && !alwaysWrap) {
+      return elements[0];
+    }
+
     if (React.isValidElement(elements) && !alwaysWrap) {
       return elements;
     }
@@ -24,4 +28,3 @@ const ReactUil = {
 };
 
 module.exports = ReactUil;
-
--- a/foundation-ui/utils/ReactUtil.js
+++ b/foundation-ui/utils/ReactUtil.js
@@ -4,6 +4,9 @@ const ReactUil = {
   /**
    * Wrap React elements
    *
+   * If elements is an array with a single element, it will not be wrapped
+   * unless alwaysWrap is true.
+   *
    * @param {Array.<ReactElement>|ReactElement} elements
    * @param  {function|String} wrapper component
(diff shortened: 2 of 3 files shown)


foundation-ui/mount/__tests__/Mount-test.js
```diff
@@ -148,8 +148,7 @@ describe('Mount', function () {
       </Mount>
     );
 
-    expect(renderer.getRenderOutput().props.children[0].props)
-        .toEqual({message:'hello world'});
+    expect(renderer.getRenderOutput().props).toEqual({message:'hello world'});
```
jfurrow: This test no longer has to look for a nested child.

rcorral:
run unit tests

jfurrow:

nLight:
If @orlandohohmeier is fine I'd merge id

foundation-ui/utils/ReactUtil.js
```diff
@@ -15,6 +15,10 @@ const ReactUil = {
       return null;
     }
 
+    if (Array.isArray(elements) && elements.length === 1 && !alwaysWrap) {
```
orlandohohmeier: I have the feeling that we, with this addition, should change the procedure to avoid unnecessary checks, but that shouldn't block this PR.

orlandohohmeier: Please document this new behavior in the util description.

orlandohohmeier: Please add check to verify that the element is a valid react element to avoid errors.

jfurrow: @mlunoe I decided against it so that strings, numbers, and `null` could still be returned without being wrapped. If we add the check, won't the same error occur when the non-renderable element is passed as children in the resulting call to `React.createElement`?

jfurrow: Done!

jfurrow:
@orlandohohmeier Thanks for the review! I added tests as requested.

foundation-ui/utils/ReactUtil.js
orlandohohmeier: Yeah, good point...  We would need to check all elements and that's probably a little over the top and definitely not blocking this PR.

jfurrow: Agreed, here's a JIRA! https://mesosphere.atlassian.net/browse/DCOS-13209

mlunoe: Come to think of it, it sort adds some inconsistency, since we are allowing some types of elements here and only ReactElements in the next statement.
Talked with @jfurrow offline about it and we decided to add the check and if we start having the need for the other types of elements, let's change that in a separate PR.

jfurrow: I just pushed this change. Thanks guys!

mlunoe:
Going to merge, as the tests have been added and comment have been added.

PR #1786, opened 2017-01-13
DCOS-13075: Keep empty values on `docker` field

nLight:
Why:

* otherwise these's no way applyPatch can delete portMappings. General idea that it's applyPatch's job to remove empty values, reducers should return all of them

DCOS-13075 #close


rcorral:
retest this please

looks like tests didn't get triggered, so running them all again.

mlunoe:
Looks good man, I was thinking about the same thing, but never actually got to fixing it

wavesoft:
Looks good! :shipit:

mlunoe:
@wavesoft please delete the branches after merge, so we don't pollute the repository

PR #1791, opened 2017-01-16
Tooltips and hint texts in service creation modal (healthchecks and networking)

philipnrmn:
This pull request adds tooltip texts in the create service form for healthchecks, and brings networking hint texts into line with our current design documents.

### DIFF
--- a/plugins/services/src/js/components/forms/HealthChecksFormSection.js
+++ b/plugins/services/src/js/components/forms/HealthChecksFormSection.js
@@ -373,21 +373,13 @@ class HealthChecksFormSection extends Component {
                       interactive={true}
                       maxWidth={300}
                       scrollContainer=".gm-scroll-view"
-                      wrapperClassName="tooltip-wrapper text-align-center pull-right"
+                      wrapperClassName="tooltip-wrapper text-align-center"
                       wrapText={true}>
                       <Icon color="grey" id="circle-question" size="mini" />
                     </Tooltip>
                   </FormGroupHeadingContent>
                 </FormGroupHeading>
               </FieldLabel>
-              <p>
-                {'Protocol of the requests to be performed. '}
-                <a
-                  href="https://mesosphere.github.io/marathon/docs/health-checks.html"
-                  target="_blank">
-                  More Information
-                </a>.
-              </p>
               <FieldSelect
                 name={`healthChecks.${key}.protocol`}
                 value={healthCheck.protocol && healthCheck.protocol.replace(MESOS_HTTPS, MESOS_HTTP)}>
@@ -428,17 +420,19 @@ class HealthChecksFormSection extends Component {
             <FormGroupHeadingContent primary={true}>
               Health Checks
             </FormGroupHeadingContent>
+            <FormGroupHeadingContent>
+              <Tooltip
+                content={tooltipContent}
+                interactive={true}
+                maxWidth={300}
+                scrollContainer=".gm-scroll-view"
+                wrapperClassName="tooltip-wrapper text-align-center pull-right"
+                wrapText={true}>
+                <Icon color="grey" id="circle-question" size="mini" />
+              </Tooltip>
+            </FormGroupHeadingContent>
           </FormGroupHeading>
         </h2>
-        <Tooltip
-          content={tooltipContent}
-          interactive={true}
-          maxWidth={300}
-          scrollContainer=".gm-scroll-view"
-          wrapperClassName="tooltip-wrapper text-align-center pull-right"
-          wrapText={true}>
-          <Icon color="grey" id="circle-question" size="mini" />
-        </Tooltip>
         <p>
           Health checks may be specified per application to be run against
           the application{'\''}s instances.
@@ -451,14 +445,6 @@ class HealthChecksFormSection extends Component {
               )}>
               Add Health Check
             </AddButton>
-            <Tooltip
-              content={tooltipContent}
-              interactive={true}
-              maxWidth={300}
-              scrollContainer=".gm-scroll-view"
-              wrapText={true}>
-              <Icon color="grey" id="circle-question" size="mini" />
-            </Tooltip>
           </FormGroup>
         </FormRow>
       </div>
--- a/plugins/services/src/js/components/forms/HealthChecksFormSection.js
+++ b/plugins/services/src/js/components/forms/HealthChecksFormSection.js
@@ -77,22 +77,24 @@ class HealthChecksFormSection extends Component {
                   <FormGroupHeadingContent primary={true}>
                     Grace Period (s)
                   </FormGroupHeadingContent>
+                  <FormGroupHeadingContent>
+                    <Tooltip
+                      content={gracePeriodTooltipContent}
+                      interactive={true}
+                      maxWidth={300}
+                      scrollContainer=".gm-scroll-view"
+                      wrapText={true}>
+                      <Icon color="grey" id="circle-question" size="mini" />
+                    </Tooltip>
+                  </FormGroupHeadingContent>
                 </FormGroupHeading>
               </FieldLabel>
-              <Tooltip
-                content={gracePeriodTooltipContent}
-                interactive={true}
-                maxWidth={300}
-                scrollContainer=".gm-scroll-view"
-                wrapperClassName="tooltip-wrapper tooltip-block-wrapper text-align-center"
-                wrapText={true}>
-                <FieldInput
-                  name={`healthChecks.${key}.gracePeriodSeconds`}
-                  type="number"
-                  min="0"
-                  placeholder="300"
-                  value={healthCheck.gracePeriodSeconds}/>
-              </Tooltip>
+              <FieldInput
+                name={`healthChecks.${key}.gracePeriodSeconds`}
+                type="number"
+                min="0"
+                placeholder="300"
+                value={healthCheck.gracePeriodSeconds}/>
               <FieldError>{errors.gracePeriodSeconds}</FieldError>
             </FormGroup>
             <FormGroup
@@ -103,22 +105,24 @@ class HealthChecksFormSection extends Component {
                   <FormGroupHeadingContent primary={true}>
                     Interval (s)
                   </FormGroupHeadingContent>
+                  <FormGroupHeadingContent>
+                    <Tooltip
+                      content={intervalTooltipContent}
+                      interactive={true}
+                      maxWidth={300}
+                      scrollContainer=".gm-scroll-view"
+                      wrapText={true}>
+                      <Icon color="grey" id="circle-question" size="mini" />
+                    </Tooltip>
+                  </FormGroupHeadingContent>
                 </FormGroupHeading>
               </FieldLabel>
-              <Tooltip
-                content={intervalTooltipContent}
-                interactive={true}
-                maxWidth={300}
-                scrollContainer=".gm-scroll-view"
-                wrapperClassName="tooltip-wrapper tooltip-block-wrapper text-align-center"
-                wrapText={true}>
-                <FieldInput
-                  name={`healthChecks.${key}.intervalSeconds`}
-                  type="number"
-                  min="0"
-                  placeholder="60"
-                  value={healthCheck.intervalSeconds}/>
-              </Tooltip>
+              <FieldInput
+                name={`healthChecks.${key}.intervalSeconds`}
+                type="number"
+                min="0"
+                placeholder="60"
+                value={healthCheck.intervalSeconds}/>
               <FieldError>{errors.intervalSeconds}</FieldError>
             </FormGroup>
             <FormGroup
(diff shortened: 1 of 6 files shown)


plugins/services/src/js/components/forms/HealthChecksFormSection.js
```diff
@@ -237,6 +359,14 @@ class HealthChecksFormSection extends Component {
               onClick={this.props.onAddItem.bind(this, {value: data.healthChecks.length, path: 'healthChecks'})}>
               <Icon color="purple" id="plus" size="tiny" /> Add Health Check
             </a>
+            <Tooltip
```
mlunoe: I believe this tooltip should be adjacent to the section heading: "Health Checks" (at least that's is how I read the docs).

plugins/services/src/js/components/forms/HealthChecksFormSection.js
```diff
@@ -127,27 +190,53 @@ class HealthChecksFormSection extends Component {
...
           className="column-6"
           showError={false}>
           <FieldLabel>Service Endpoint</FieldLabel>
-          <FieldSelect
-            name={`healthChecks.${key}.portIndex`}
-            value={String(healthCheck.portIndex)}>
-            <option value="">Select Endpoint</option>
-            {this.getEndpoints()}
-          </FieldSelect>
+          <Tooltip
```
mlunoe: In other places, we have added these with a question mark icon next to the label text (inside the label). I would suggest that we do the same here.

plugins/services/src/js/components/forms/HealthChecksFormSection.js
philipnrmn: It looks a bit weird on the title, so I've added it next to the intro text instead.

philipnrmn:
Addressed all your feedback @mlunoe

plugins/services/src/js/components/forms/HealthChecksFormSection.js
jfurrow: I'm with @mlunoe so let's get a designer's input on this. @leemunroe what do you think?

In the title:


In the help text:

philipnrmn:
run integration tests

plugins/services/src/js/components/forms/HealthChecksFormSection.js
leemunroe: +1 for title

plugins/services/src/js/components/forms/MultiContainerNetworkingFormSection.js
```diff
@@ -160,15 +168,36 @@ class MultiContainerNetworkingFormSection extends mixin(StoreMixin) {
       address = `${hostname}${Networking.L4LB_ADDRESS}${port}`;
     }
 
+    const loadBalancerTooltipContent = (
+      <span>
+        {`Load balance the service internally (layer 4), and create a service
+        address. For external (layer 7) load balancing, create an external
+        load balancer and attach this service. `}
+        <a href="https://docs.mesosphere.com/1.8/usage/service-discovery/load-balancing-vips/"
```
MatApple: I'm a bit worried about release specific links being hard coded. I fear we might end up with links to wrong documentation for future releases. Can't think of a simple way to manage this though. Doc links could also change between releases. Any ideas?

plugins/services/src/js/components/forms/MultiContainerNetworkingFormSection.js
```diff
@@ -86,13 +85,20 @@ class MultiContainerNetworkingFormSection extends mixin(StoreMixin) {
 
     let tooltipContent = (
```
MatApple: Perhaps we should rename this to more specific amidst the other tooltip declarations.

plugins/services/src/js/components/forms/NetworkingFormSection.js
```diff
@@ -137,15 +145,36 @@ class NetworkingFormSection extends mixin(StoreMixin) {
       address = `${hostname}${Networking.L4LB_ADDRESS}${port}`;
     }
 
+    const loadBalancerTooltipContent = (
```
MatApple: This looks like the same message used elsewhere. I'm wondering if all these tooltips should be extracted to a single file and all the help texts into another. That would make it easier to update the docs links too

plugins/services/src/js/components/forms/MultiContainerNetworkingFormSection.js
philipnrmn: That's a good point. For the moment, I propose we use the generic endpoint (eg in this case https://docs.mesosphere.com/usage/service-discovery/load-balancing-vips).

plugins/services/src/js/components/forms/NetworkingFormSection.js
philipnrmn: Yeah, I think we are expecting to roll tooltips out further and those will definitely want to be centralised. Is it worth spending some time now trying to fix this, or shall we press on with what we've got? @mlunoe , what's your feeling?

plugins/services/src/js/components/forms/MultiContainerNetworkingFormSection.js
```diff
@@ -160,15 +168,36 @@ class MultiContainerNetworkingFormSection extends mixin(StoreMixin) {
       address = `${hostname}${Networking.L4LB_ADDRESS}${port}`;
     }
 
+    const loadBalancerTooltipContent = (
+      <span>
+        {`Load balance the service internally (layer 4), and create a service
+        address. For external (layer 7) load balancing, create an external
+        load balancer and attach this service. `}
+        <a href="https://docs.mesosphere.com/usage/service-discovery/load-balancing-vips/"
```
mlunoe: We do have a `MetadataStore.buildDocsURI` function to build doc links with the correct version!

plugins/services/src/js/components/forms/MultiContainerNetworkingFormSection.js
```diff
@@ -370,6 +399,13 @@ class MultiContainerNetworkingFormSection extends mixin(StoreMixin) {
       </span>
     );
 
+    const serviceEndpointsTooltipContent = (
+      <a href="https://docs.mesosphere.com/usage/service-discovery/load-balancing-vips/virtual-ip-addresses/"
```
mlunoe: Same here: We do have a `MetadataStore.buildDocsURI` function to build doc links with the correct version!

plugins/services/src/js/components/forms/NetworkingFormSection.js
```diff
@@ -137,15 +145,36 @@ class NetworkingFormSection extends mixin(StoreMixin) {
...
+      <span>
+        {`Load balance the service internally (layer 4), and create a service
+        address. For external (layer 7) load balancing, create an external
+        load balancer and attach this service. `}
+        <a href="https://docs.mesosphere.com/usage/service-discovery/load-balancing-vips/"
```
mlunoe: Same here: We do have a `MetadataStore.buildDocsURI` function to build doc links with the correct version!

plugins/services/src/js/components/forms/NetworkingFormSection.js
mlunoe: I would say; Not worth the effort. What we should do is create smaller components that we could use in both sections. That was a clear tradeoff we made when we were working on this feature.

plugins/services/src/js/components/forms/NetworkingFormSection.js
```diff
@@ -408,6 +438,24 @@ class NetworkingFormSection extends mixin(StoreMixin) {
     const isMesosRuntime = !type || type === NONE;
     const isUserNetwork = networkType && networkType.startsWith(USER);
 
+    const serviceEndpointsTooltipContent = (
+      <a href="https://docs.mesosphere.com/usage/service-discovery/load-balancing-vips/virtual-ip-addresses/"
```
mlunoe: Same here

plugins/services/src/js/schemas/service-schema/Networking.js
```diff
@@ -11,7 +11,7 @@ const Networking = {
   title: 'Network',
   description: (
     <span>
-      Configure the networking for your service. We will automatically generate a Service Address to connect to for each of your load balanced endpoints. <a href="https://docs.mesosphere.com/1.8/usage/service-discovery/load-balancing-vips/virtual-ip-addresses/" target="_blank">Read more about load balancing</a>.
+      Configure the networking for your service. We will automatically generate a Service Address to connect to for each of your load balanced endpoints. <a href="https://docs.mesosphere.com/usage/service-discovery/load-balancing-vips/virtual-ip-addresses/" target="_blank">Read more about load balancing</a>.
```
mlunoe: 1. Dynamic docs link
2. break this into multiple lines. You are going to hit errors when this thing lands: https://github.com/dcos/dcos-ui/pull/1866

plugins/services/src/js/components/forms/MultiContainerNetworkingFormSection.js
mlunoe: We do have a MetadataStore.buildDocsURI function to build doc links with the correct version!

philipnrmn:
Thanks for the feedback, fixed up this PR accordingly.

philipnrmn:
Rebased 🔨

philipnrmn:
run unit tests

jfurrow:
Ping @MatApple @mlunoe let's get this merged!

I made some changes to this PR (I hope you don't mind @philipnrmn):
* moved the tooltips to be inline with headings and labels
* applied spacing with whitespace character (this spacing will be fixed by https://github.com/dcos/dcos-ui/pull/1887, and until that merges I'd like us to be consistent with the rest of the form, even though it's a bad solution)

leemunroe:
* The help text, or hint text, for Protocol breaks convention here. It should be below the field. Although I don't think this field needs both a help tooltip AND a line of help text. I would side with just help tooltip. @sascala It looks like this is what we suggested in the content doc. Was that intended? i.e. remove `Protocol of the requests to be performed. More Information.`

* What is the help tooltip next to the add health check button? Not sure if that should be there.

plugins/services/src/js/components/forms/HealthChecksFormSection.js
```diff
@@ -39,7 +71,9 @@ class HealthChecksFormSection extends Component {
             <FormGroup
               className="column-3"
               showError={Boolean(errors.gracePeriodSeconds)}>
-              <FieldLabel>Grace Period (s)</FieldLabel>
+              <FieldLabel helpText={gracePeriodHelpText}>
```
mlunoe: I am not sure I understand what is going on here?
FieldLabel does not have a property `helpText`.
This needs to be added inside the `FormGroup` using a `FieldHelp` component, like so:
```js
<FormGroup
  className="column-3"
  showError={Boolean(errors.gracePeriodSeconds)}>
  <FieldInput
    name={`healthChecks.${key}.gracePeriodSeconds`}
    type="number"
    min="0"
    placeholder="300"
    value={healthCheck.gracePeriodSeconds} />
  <FieldHelp>
    (Optional. Default: 300): Health check failures are ignored within this number of seconds or until the instance becomes healthy for the first time.
  </FieldHelp>
  <FieldError>{errors.gracePeriodSeconds}</FieldError>
</FormGroup>
```

This needs to change for all of them

philipnrmn: Hi @mlunoe this PR actually adds the property `helpText` to FieldLabel.

mlunoe: Arh! I missed that part. However, I feel this is going against our pattern of composing using smaller components. I would like us to do this as suggested above

philipnrmn: Mm, OK. I'm probably missing some context, but I believed the advantage of composition was to avoid components with flexible interfaces (like the `HashMapList`), which tend to introduce bugs. 

`FieldLabel` is itself a convenience instead of using multiple divs, so introducing a parameter which optionally adds a tooltip doesn't seem evil to me.

mlunoe: Yeah, I still don't feel this is the declarative approach. @jfurrow has worked on this I think that is a superior solution: https://github.com/dcos/dcos-ui/pull/1887/files#diff-82fc08fe0a4224982c1e31ce7424c416R97

What do you say to reverting to what we do elsewhere and let that PR handle it?

MatApple:
@jfurrow @mlunoe - has anyone verified @leemunroe's comments and the fact that part of this PR was only a temp fix till https://github.com/dcos/dcos-ui/pull/1887 merged (which it now has) - see comment https://github.com/dcos/dcos-ui/pull/1791#issuecomment-282374702

mlunoe:
@MatApple good catch! Will go through with @leemunroe now to make sure we are following his direction (which I agree with).

jfurrow:
Good catch @MatApple, I either didn't see or forgot about @leemunroe's comments... let's work with him to figure out the best solution to this.

mlunoe:
Adjusted the Network Section:

Adjusted the Health Checks Section:

Adjusted the Multi Container Network Section:

Adjusted the Multi Container Health Checks Section:

mlunoe:
@leemunroe I've made the requested changes; please confirm or reject as you see fit!

mlunoe:
@MatApple how do you feel about the change?

mlunoe:
I have worked with @leemunroe and @silverem2 to make sure the sections follow their direction.
Here are updated screenshots:
Networking Form Section

Multi Container Networking Form Section

Health Check Form Section

Multi Container Health Check Form Section

jfurrow:
That looks great! 👏

plugins/services/src/js/components/forms/MultiContainerNetworkingFormSection.js
```diff
@@ -183,42 +184,51 @@ class MultiContainerNetworkingFormSection extends mixin(StoreMixin) {
...
-                Load Balanced Service Address
+                Enable Load Balanced Service Address
+              </FormGroupHeadingContent>
+              <FormGroupHeadingContent>
+                <Tooltip
+                  content={loadBalancerTooltipContent}
+                  interactive={true}
+                  maxWidth={300}
+                  scrollContainer=".gm-scroll-view"
+                  wrapperClassName="tooltip-wrapper text-align-center pull-right"
```
jfurrow: Let's remove the `pull-right` class on this tooltip.

mlunoe: Thank you for checking up on that! They have now been removed!

mlunoe:
Rebased and updated the environment variable reference to what Marathon actually creates:
https://github.com/mesosphere/marathon/blob/master/src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala#L410
Created a pod with service endpoint name `one`, port 80 and printed out the variables in the command:

mlunoe:
Here is a visual of the update (look for `ENDPOINT_{NAME}`:

mlunoe:
@jfurrow is on vacation, but I did take care of his points, so I hope he is ok with the PR at this stage

Poltergeist:
I'll have a look here now.

Poltergeist:
This looks good to me and I will add my label as the other label is quite old as in 24 of February I'll remove it.

PR #1792, opened 2017-01-16
DCOS-13238: Ensuring Job.getSchedules always returns an array

wavesoft:
This PR fixes a bug that occurs if somebody pastes a wrongly-formatted job spec, that uses `schedules: { }` instead of `schedules: [ {} ]`.

For example, as seen [here](https://dcos.io/docs/1.9/usage/jobs/getting-started/):
```json
 {
     "id": "myjob",
     "description": "A job that sleeps regularly",
     "run": {
         "cmd": "sleep 20000",
         "cpus": 0.01,
         "mem": 32,
         "disk": 0
     },
     "schedules": {
         "id": "sleep-schedule",
         "enabled": true,
         "cron": "20 0 * * *",
         "concurrencyPolicy": "ALLOW"
     }
 }
```

### DIFF
--- a/src/js/structs/Job.js
+++ b/src/js/structs/Job.js
@@ -102,7 +102,13 @@ module.exports = class Job extends Item {
   }
 
   getSchedules() {
-    return this.get('schedules') || [];
+    const schedules = this.get('schedules');
+
+    if (!Array.isArray(schedules)) {
+      return [];
+    }
+
+    return schedules;
   }
 
   getScheduleStatus() {


src/js/structs/Job.js
```diff
@@ -102,7 +102,13 @@ module.exports = class Job extends Item {
   }
 
   getSchedules() {
-    return this.get('schedules') || [];
+    const schedules = this.get('schedules');
+
+    if (!Array.isArray(schedules)) {
+      return [];
```
nLight: I'd rather test if it's an Object and wrap it with an array.

Because here we just discard user input without any message/error
so a user is able to create a job that will be unscheduled. Everybody knows how it is frustrating when a service just throws your input away without telling you anything. Let's not be that service 💪

wavesoft: I agree, but doing so we are accepting an invalid API request. That's not correct.

I would vote for adding RAML (or other type of) validation to the jobs when it's available, rather than trying to be smart on what we accept.

jfurrow:
run integration tests

src/js/structs/Job.js
wavesoft: @nLight can you approve the changes, unless you have any other objection?

nLight: Let's create a ticket to track both 
1. documentation that should be fixed
2. raml validation that should be added to jobs

wavesoft: Done 👍

wavesoft:
run integration tests

jfurrow:
LGTM! :ship:

PR #1793, opened 2017-01-16
Fix service actions tests

orlandohohmeier:
Provide an alias for all "default" routes to enable tests to wait for the respective responses and refactor the service action test using the aliases instead of arbitrary timeouts. These changes will resolve the service action test reliability issues.

### DIFF
--- a/tests/pages/services/ServiceActions-cy.js
+++ b/tests/pages/services/ServiceActions-cy.js
@@ -2,6 +2,7 @@ describe('Service Actions', function () {
 
   function clickHeaderAction(actionText) {
     cy.get('.page-header-actions .dropdown').click();
+
     cy.get('.dropdown-menu-items').contains(actionText).click();
   }
 
@@ -11,7 +12,8 @@ describe('Service Actions', function () {
       nodeHealth: true
     });
 
-    cy.visitUrl({url: '/services/overview/%2Fcassandra-healthy'});
+    cy.visitUrl({url: '/services/overview/%2Fcassandra-healthy'})
+      .wait(['@marathon/groups', '@marathon/deployments', '@state']);
   });
 
   context('Edit Action', function () {
@@ -29,9 +31,9 @@ describe('Service Actions', function () {
         .route({
           method: 'PUT',
           url: /marathon\/v2\/apps\/\/cassandra\-healthy/,
-          response: [],
-          delay: 0
+          response: []
         });
+
       cy.get('.modal .modal-header .button')
         .contains('Review & Run')
         .click();
@@ -59,7 +61,8 @@ describe('Service Actions', function () {
           nodeHealth: true
         });
 
-        cy.visitUrl({url: '/services/overview/%2Fsleep'});
+        cy.visitUrl({url: '/services/overview/%2Fsleep'})
+          .wait(['@marathon/groups', '@marathon/deployments', '@state']);
         clickHeaderAction('Destroy');
       });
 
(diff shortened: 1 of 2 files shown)


wavesoft:
I have some flacky tests...

screen shot 2017-01-17 at 14 56 52

orlandohohmeier:
@wavesoft it seems that in very rare(!) cases our modal fails to have a reference to the respective DOM element, but I wasn't able to reproduce the issue. I've slightly adjust the test as I saw a few test fail every now in then as crypress wasn't quick enough to check css classes.

orlandohohmeier:

tests/pages/services/ServiceActions-cy.js
```diff
@@ -127,14 +130,13 @@ describe('Service Actions', function () {
             url: /marathon\/v2\/apps\/\/sleep/,
             response: {
               message: {message: 'Not Authorized to perform this action!'}
-            },
-            delay: 100
```
rcorral: We should leave this delay here. It's part of the test.
Basically, the button should be briefly disabled and we check that bellow. Then we check that the class is removed.

tests/pages/services/ServiceActions-cy.js
```diff
@@ -219,13 +219,14 @@ describe('Service Actions', function () {
...
           url: /marathon\/v2\/apps\/\/cassandra\-healthy/,
-          response: [],
-          delay: 50
-        });
+          response: []
+        }).as('marathon/apps/cassandra-healthy:scaleService');
+
       cy.get('.modal-footer .button-collection .button-primary')
         .as('primaryButton').click();
-      cy.get('@primaryButton').should('have.class', 'disabled');
```
rcorral: Same here, we're actually testing that the button is disabled when waiting for the request response.

tests/pages/services/ServiceActions-cy.js
```diff
@@ -59,7 +61,8 @@ describe('Service Actions', function () {
           nodeHealth: true
         });
 
-        cy.visitUrl({url: '/services/overview/%2Fsleep'});
+        cy.visitUrl({url: '/services/overview/%2Fsleep'})
+          .wait(['@marathon/groups', '@marathon/deployments', '@state']);
```
rcorral: I feel we should just rely on what's on the page to tell if it's ready or not rather than relying on endpoints responding. Can't we adjust the page to only render once we've received data from these different endpoints and showing the loading indicator until this data has been received?

rcorral:
@orlandohohmeier I've created this PR #1816
The problem was that the delay on the requests wasn't delayed enough, causing flaky tests.
Let me know what you think.

orlandohohmeier:
@rcorral we're disabling the button immediately, thus I don't think that we need to increase the delay. It seems that cypress isn't fast enough if the operations are written as separate directives... `cy.get('...').click(); cy.get('...').should('...');` vs. `cy.get('...').click().should('...');`

tests/pages/services/ServiceActions-cy.js
orlandohohmeier: There is no need for the delay as the button is disabled immediately.

tests/pages/services/ServiceActions-cy.js
orlandohohmeier: I'm happy to add the check again. I only removed it to reduce the number of checks to narrow down the flaky ones and didn't added it again as we have an extra test to verify that the button is properly disabled.

mlunoe:
I talked with @rcorral about this offline. The delays feel less intrusive and will be easier to change to a better solution down the road.
The route alias solution reminds me of the unmock/mock situation in Jest where you will end up getting out of sync; some request might not even be required anymore and some needs to be added, but it is hard to tell which.

All in all, I think we need a pattern we can recognise and a solution we can apply to handle it. I was thinking some way of controlling when a server responds, e.g. store the result somewhere and only deliver it to the client once you are done with your assessment.
@rcorral agreed to talk with the guys from Cypress to see if there is a better way to write these tests or (even better) if there is a way we could set this control of response up. Otherwise, we could work on a solution ourselves.

rcorral:
closing this PR as #1816 merged.

PR #1811, opened 2017-01-18
DCOS-11160: Adjust the DCOSStore data received flags

orlandohohmeier:
---
ℹ️   _This PR basically cherry-picks a DC/OS UI 1.8 fix (#1303) into master, which wasn't forwarded earlier due to the planned GraphQL refactoring._

---

Rename `dataProcessed` to `serviceDataReceived` and introduce a new job specific data received getter to entangle the Job and Service data received/processed behavior. This changes will make sure that we don't block the Job table rendering in case we didn't get any Service data,
which in some cases resulted in an unusable UI.

Closes DCOS-11160

### DIFF
--- a/plugins/services/src/js/pages/__tests__/DeploymentsTab-test.js
+++ b/plugins/services/src/js/pages/__tests__/DeploymentsTab-test.js
@@ -43,7 +43,7 @@ describe('DeploymentsTab', function () {
         }
       ]
     });
-    DCOSStore.dataProcessed = true;
+    DCOSStore.serviceDataReceived = true;
     DCOSStore.deploymentsList = deployments;
     this.container = global.document.createElement('div');
     this.instance = ReactDOM.render(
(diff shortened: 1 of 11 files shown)


orlandohohmeier:
run integration tests

orlandohohmeier:
run integration tests

orlandohohmeier:
run integration tests

mlunoe:
Works great :)

plugins/services/src/js/pages/__tests__/DeploymentsTab-test.js
```diff
@@ -43,7 +43,7 @@ describe('DeploymentsTab', function () {
         }
       ]
     });
-    DCOSStore.dataProcessed = true;
+    DCOSStore.serviceDataReceived = true;
```
wavesoft: You have only defined a getter, so wouldn't this be ignored? Or is this just mocked?

wavesoft: Since you have only defined a getter for `serviceDataReceived`, assigning a value to it should not be a legit operation. It currently works because we are mocking the `DCOSStore` component.

I am not going to block this PR since we need to get moving and this is not really prohibiting it from working properly, but please create an issue so we can tackle it at some point. 👍

PR #1813, opened 2017-01-18
DCOS-13246: Disabling portMapping when using MESOS + Virtual Network

wavesoft:
Apparently mesos does not support MESOS containerizer and the `portMapping` property to be present in the `docker` object. This PR ensures this does not happen


jfurrow:
Could we disable the corresponding form, or display a message to the user? The form allows me to continue adding service endpoints but the resulting JSON displays nothing:

wavesoft:
Good catch @jfurrow , fixed 👍

orlandohohmeier:
run integration tests

jfurrow:
Works great! 🥇

orlandohohmeier:
LGTM! 🚢

PR #1814, opened 2017-01-18
DCOS-12442: Adding containerVolmesPath validator

wavesoft:
This validator ensures that the container volumes do not contain '/' in the
containerPath property.

### DIFF
--- a/plugins/services/src/js/validators/MarathonAppValidators.js
+++ b/plugins/services/src/js/validators/MarathonAppValidators.js
@@ -139,6 +139,35 @@ const MarathonAppValidators = {
     return [];
   },
 
+  /*
+   * This validator tests two fields:
+   *
+   * - container.volumes.X.containerPath should not contain '/'
+   *
+   * @param {Object} app - The data to validate
+   * @returns {Array} Returns an array with validation errors
+   */
+  containerVolmesPath(app) {
+    const volumes = findNestedPropertyInObject(app, 'container.volumes');
+    if (ValidatorUtil.isEmpty(volumes)) {
+      return [];
+    }
+
+    return volumes.reduce(function (memo, volume, index) {
+      const containerPath = volume.containerPath || '';
+
+      // Local volumes should not contain '/'
+      if (containerPath.indexOf('/') !== -1) {
+        memo.push({
+          path: ['container', 'volumes', index, 'containerPath'],
+          message: 'Should not contain "/"'
+        });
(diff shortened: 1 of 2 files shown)


plugins/services/src/js/validators/MarathonAppValidators.js
```diff
@@ -137,6 +137,38 @@ const MarathonAppValidators = {
...
+    }
+
+    return volumes.reduce(function (memo, volume, index) {
+      const containerPath = volume.containerPath || '';
+
+      // Local volumes should not contain '/'
+      if (containerPath.indexOf('/') !== -1) {
+        memo.push({
+          path: ['container', 'volumes', index, 'containerPath'],
+          message: 'Should not contain "/"'
```
rcorral: What about other characters?
Could it be best to just say use alphanumeric characters?

wavesoft: Marathon matching regex (at least from the error response) is `^[^/]$', so it only checks for '/'

nLight:
⚔️ flicts!

plugins/services/src/js/validators/MarathonAppValidators.js
```diff
@@ -137,6 +137,38 @@ const MarathonAppValidators = {
 
     // No errors
     return [];
+  },
+
+  /**
+   * This validator tests two fields:
```
nLight: I see only one field

plugins/services/src/js/validators/MarathonAppValidators.js
```diff
@@ -137,6 +137,38 @@ const MarathonAppValidators = {
...
+  containerVolmesPath(app) {
+    const volumes = findNestedPropertyInObject(app, 'container.volumes');
+    if (ValidatorUtil.isEmpty(volumes)) {
+      return [];
+    }
+
+    return volumes.reduce(function (memo, volume, index) {
+      const containerPath = volume.containerPath || '';
+
+      // Local volumes should not contain '/'
```
nLight: Comment says 'Local volumes' but during my test I observed that the check is being triggered for 'External volumes' as well

wavesoft:
Rebased + Addressed comments 👍

wavesoft:
run integration tests

jfurrow:
run integration tests

jfurrow:
Works perfectly!

PR #1822, opened 2017-01-19
[1 of 3: User Account Menu] DCOS-12462: Extend ClipboardTrigger functionality

jfurrow:
1 of 3: #1822 
2 of 3: #1826 — [diff between 1 and 2](https://github.com/dcos/dcos-ui/compare/john/feature/extend-clipboard-trigger...john/DCOS-12462/new-user-account-menu-components?expand=1)
3 of 3: #1827  — [diff between 2 and 3](https://github.com/dcos/dcos-ui/compare/john/DCOS-12462/new-user-account-menu-components...john/DCOS-12462/sidebar-user-account-menu?expand=1)

This PR extends the functionality of the `ClipboardTrigger` component. It allows custom trigger buttons instead of just the clipboard icon and adds an optional `onTextCopy` callback.

I also removed the Safari exclusion because Safari 10 supports the Clipboard API. As of December 2016, Safari 10 is at 2.92% global usage while Safari 9.0 + 9.1 is at 1.37% combined (and dropping quickly: http://gs.statcounter.com/#desktop-browser_version_partially_combined-ww-monthly-201512-201612).

### DIFF
--- a/src/js/components/ClipboardTrigger.js
+++ b/src/js/components/ClipboardTrigger.js
@@ -46,22 +46,24 @@ class ClipboardTrigger extends React.Component {
   getTriggerContent() {
     const {children, className} = this.props;
 
+    let childNode = (
+      <Icon
+        id="clipboard"
+        size="mini"
+        className={`clickable icon-clipboard ${className}`}
+        color="purple" />
+    );
+
     if (children != null) {
-      return (
-        <span className={className} onMouseEnter={this.handleCopyIconMouseEnter}
-          ref="copyButton">
-          {children}
-        </span>
-      );
+      childNode = children;
     }
 
     return (
-      <span onMouseEnter={this.handleCopyIconMouseEnter} ref="copyButton">
-        <Icon
-          id="clipboard"
-          size="mini"
-          className={`clickable icon-clipboard ${className}`}
-          color="purple" />
+      <span
+        className={className}
+        onMouseEnter={this.handleCopyIconMouseEnter}
+        ref="copyButton">
+        {childNode}
       </span>
     );
   }
@@ -80,13 +82,10 @@ class ClipboardTrigger extends React.Component {
 
   render() {
     const {copiedText, tooltipContent, useTooltip} = this.props;
+    const {hasCopiedToClipboard} = this.state;
 
     if (useTooltip) {
-      let text = tooltipContent;
-
-      if (this.state.hasCopiedToClipboard) {
-        text = copiedText;
-      }
+      const text = hasCopiedToClipboard ? copiedText : tooltipContent;
 
       return (
         <Tooltip position="bottom" content={text}>
--- a/src/js/components/ClipboardTrigger.js
+++ b/src/js/components/ClipboardTrigger.js
@@ -56,13 +56,13 @@ class ClipboardTrigger extends React.Component {
     }
 
     return (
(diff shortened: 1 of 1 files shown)


src/js/components/ClipboardTrigger.js
```diff
@@ -44,39 +43,59 @@ class ClipboardTrigger extends React.Component {
...
+      return (
+        <span className={className} onMouseEnter={this.handleCopyIconMouseEnter}
+          ref="copyButton">
+          {children}
+        </span>
+      );
+    }
+
+    return (
+      <Icon
```
wavesoft: Apparently the `Clipboard` constructor will not accept an `<svg />` element as a first argument. This means that we should wrap the `<Icon>` with a `<span>` or something else, and put the `ref=` there.

jfurrow: Good catch! Done!

src/js/components/ClipboardTrigger.js
```diff
@@ -44,39 +43,59 @@ class ClipboardTrigger extends React.Component {
     }
   }
 
+  getTriggerContent() {
+    const {children, className} = this.props;
+
+    if (children != null) {
+      return (
+        <span className={className} onMouseEnter={this.handleCopyIconMouseEnter}
```
mlunoe: #nit All attributes on a separate line. Related PR: https://github.com/dcos/dcos-ui/pull/1866

src/js/components/ClipboardTrigger.js
```diff
@@ -44,39 +43,59 @@ class ClipboardTrigger extends React.Component {
...
+          ref="copyButton">
+          {children}
+        </span>
+      );
+    }
+
+    return (
+      <span onMouseEnter={this.handleCopyIconMouseEnter} ref="copyButton">
+        <Icon
```
mlunoe: Couldn't we just default assign children to this icon?

src/js/components/ClipboardTrigger.js
```diff
@@ -44,39 +43,59 @@ class ClipboardTrigger extends React.Component {
...
-      <Icon
-        id="clipboard"
-        size="mini"
-        className="clickable icon-clipboard"
-        color="purple"
-        onMouseEnter={this.handleCopyIconMouseEnter} />
-    );
-    let text = tooltipContent;
+    if (useTooltip) {
+      let text = tooltipContent;
```
mlunoe: Could be written like so (to loose the nested if-statement):
```js
const {hasCopiedToClipboard} = this.state;

if (useTooltip) {
  const text = hasCopiedToClipboard ? copiedText : tooltipContent;
```

src/js/components/ClipboardTrigger.js
jfurrow: Done!

src/js/components/ClipboardTrigger.js
jfurrow: I did this, but this means we introduce an unnecessary render cycle for the `Icon` component in some cases. It's not a big deal in this component, but we should be cautious about doing this everywhere, IMO.

src/js/components/ClipboardTrigger.js
jfurrow: Done!

PR #1823, opened 2017-01-19
Align location and start-time in deployments table

philipnrmn:
Before:


jfurrow:
This works for now, but at some point we should refactor this table so as not to rely on these magic pixel values to achieve the vertical alignment. I created a JIRA for that here: https://mesosphere.atlassian.net/browse/DCOS-13425

PR #1824, opened 2017-01-19
[1 of 2: Dockerized UI] Adds proper Dockerized DC/OS UI

rcorral:
To try this out, check out this branch and follow the instructions on the [README](https://github.com/dcos/dcos-ui/blob/rafa/dcos-ui-docker/docker/README.md).

**I'd like feedback on the [instructions](https://github.com/dcos/dcos-ui/blob/rafa/dcos-ui-docker/docker/README.md) and ease of use.**


rcorral:
run integration tests

docker/README.md
```diff
@@ -0,0 +1,40 @@
...
+  # --build-arg PLUGINS_REPO="" \ # Optional
+    -t dcos-ui:latest \
+    -f docker/Dockerfile.dev .
+  ```
+  If you have a repository of plugins to enhance DC/OS UI, you may configure this by passing a URL to the git repo to be cloned. For example: `https://TOKEN:x-oauth-basic@github.com/my-company/dcos-ui-plugins.git`
+
+3. Run the dcos-ui docker image
+
+  ```sh
+  docker run --ipc=host \
```
mlunoe: Same thing here, not sure how to do this best, though

docker/README.md
```diff
@@ -0,0 +1,40 @@
...
+1. Start docker on your machine
+2. Build the dcos-ui docker image. Run this command inside the the cloned `dcos-ui` repo.
+
+  ```sh
+  docker build \
+  # --build-arg PLUGINS_REPO="" \ # Optional
+    -t dcos-ui:latest \
+    -f docker/Dockerfile.dev .
+  ```
+  If you have a repository of plugins to enhance DC/OS UI, you may configure this by passing a URL to the git repo to be cloned. For example: `https://TOKEN:x-oauth-basic@github.com/my-company/dcos-ui-plugins.git`
```
mlunoe: Let's make it obvious that `TOKEN` should be replaced

mlunoe: Add example: https://github.com/settings/tokens

docker/Dockerfile.dev
```diff
@@ -0,0 +1,43 @@
...
+LABEL node-version="4.4.4"
+
+ARG PLUGINS_REPO
+EXPOSE 4200
+WORKDIR /usr/src
+
+ENV JAVA_HOME /usr/lib/jvm/java-8-openjdk-amd64
+
+# Basic dependencies
+RUN echo 'deb http://httpredir.debian.org/debian jessie-backports main' > /etc/apt/sources.list.d/jessie-backports.list
```
mlunoe: The mirrors seem to fail consistently. Adding this line before `apt-get upgrade` seemed to do the trick:
https://github.com/rgeissert/http-redirector/issues/74#issuecomment-241682578

```sh
# Insert this line before "RUN apt-get update" to dynamically
# replace httpredir.debian.org with a single working domain
# in attempt to "prevent" the "Error reading from server" error.
RUN sed -i "s/httpredir.debian.org/`curl -s -D - http://httpredir.debian.org/demo/debian/ | awk '/^Link:/ { print $2 }' | sed -e 's@<http://\(.*\)/debian/>;@\1@g'`/" /etc/apt/sources.list
```

mlunoe: Done!

docker/README.md
mlunoe: Done

docker/README.md
```diff
@@ -0,0 +1,40 @@
+## Running DCOS US in Docker
```
rcorral: lol

rcorral:
@mlunoe I updated the readme a bit, let me know what you think.

rcorral:
run integration tests

ashenden:
Instructions weren't too hard to follow. Some random thoughts though:

* It might be good to give a high-level description of what this is actually doing.
* Maybe separate the Build steps from the Run steps?
* Do I need to go through the build steps every time or is this a one time thing?  Would be good to get to explicitly define when you'd do the build/run.
* I went through the whole build process (5-10mins) before being alerted that I had the wrong plugins repo URL.  Communicate errors like this earlier?
* For step #3 ("Set cluster URL") - I thought the whole point of this was to avoid needing a cluster. Saving setup complexity and money. Maybe I misunderstood.
* Make a video tutorial?
* Autolaunch UI plz

ngehani:
Here are some recommendations:

1. Put a link to pre-requisite to install a dcos cluster since CCM is not option for customers
2. Webpack: Build is now available doesn't tell you what to do next. Would be good to say something regarding the UI is available on http://localhost:4200
3. The docker commands or any command line should be written so a user can just copy and paste on the command line. I would suggest splitting the docker run command into two docker run commands. One for the required and one for the optional version.

mlunoe:
Looks good man!

docker/README.md
```diff
@@ -0,0 +1,79 @@
...
+
+2. Set cluster URL. The address of a DC/OS cluster to be used as the backend. The cluster version must be compatible with the UI assets.
+
+  ```sh
+  export CLUSTER_ADDRESS=https://a-dcos-cluster.us-west-2.elb.amazonaws.com
+  ```
+
+3. Run the dcos-ui docker image
+
+  **Option 1**. Run without a plugins repo
```
MatApple: Instead of having option 1, option 2, maybe we could have:
### Run the dcos-ui docker image
`docker run --ipc=host -it dcos-ui-browse`

**Options**: (configure options by adding `-e "option_name=option_value"` to command) e.g. to add a cluster address: `docker run --ipc=host -e "CLUSTER_ADDRESS=https://blah/blah.com" -it dcos-ui-browse`

**CLUSTER_ADDRESS**: (default: none - UI will use fixtures) https address of cluster.
**APPLICATION_CHECKOUT**: (default: master) Branch or commit hash to checkout
...

docker/build-dcos-ui
```diff
@@ -0,0 +1,22 @@
...
+cd /usr/src/dcos-ui
+git pull
+git checkout $APPLICATION_CHECKOUT
+
+npm set progress=false
+npm run --unsafe-perm preinstall
+npm i
+npm run scaffold
+
+echo "module.exports = {'*': '$CLUSTER_ADDRESS'};" >> webpack/proxy.dev.js
```
MatApple: If no `$CLUSTER_ADDRESS` let's set UI to use fixtures.

docker/build-dcos-ui
```diff
@@ -0,0 +1,22 @@
...
+
+if [[ ! -z "$PLUGINS_CHECKOUT" && -d "./dcos-ui-plugins" ]]; then
+  cd ./dcos-ui-plugins
+  git pull
+  git checkout $PLUGINS_CHECKOUT
+fi
+
+cd /usr/src/dcos-ui
+git pull
+git checkout $APPLICATION_CHECKOUT
```
MatApple: Let's default to master if no `$APPLICATION_CHECKOUT`

docker/Dockerfile.browse
```diff
@@ -0,0 +1,19 @@
...
+
+ARG PLUGINS_REPO
+EXPOSE 4200
+WORKDIR /usr/src
+
+# Update npm
+RUN npm install -g npm@3.9.0
+
+# Clone repos
+RUN git clone https://github.com/dcos/dcos-ui.git /usr/src/dcos-ui
```
orlandohohmeier: Why are we downloading the complete repository instead of copying it into the container?

docker/Dockerfile.browse
```diff
@@ -0,0 +1,19 @@
+FROM node:4.4.4
+LABEL node-version="4.4.4"
```
orlandohohmeier: Do we really need the label? The node base image should provide a `NODE_VERSION` env if I'm not mistaken.

docker/README.md
```diff
@@ -0,0 +1,79 @@
...
+  **Option 1**. Build without a plugins repo
+    ```sh
+    docker build \
+      -t dcos-ui-browse:latest \
+      -f docker/Dockerfile.browse .
+    ```
+
+  **Option 2**. Build with a plugins repo
+
+    If you have a repository of plugins to enhance DC/OS UI, you may configure this by passing a URL to the git repo. For example: `https://{AUTH_TOKEN}:x-oauth-basic@github.com/my-company/dcos-ui-plugins.git`, where `{AUTH_TOKEN}` would be your generated token from, e.g. https://github.com/settings/tokens.
```
rcorral: @MatApple I'm trying to make it as agnostic to GH has possible. But you have a good point, it could be more verbose. I'll make a few changes.

docker/README.md
```diff
@@ -0,0 +1,79 @@
...
+
+    ```sh
+    docker build \
+      --build-arg PLUGINS_REPO="" \
```
rcorral: This is where I'm hoping that the text below comes in to help. Typically people just like to copy/paste code block and hit enter. So my thought was to make it easy for them to do that and then they just need to copy and paste their repo URL which could be a GH repo, but doesn't have to. This is also why I split things into Opt 1 and Opt 2, so that the user has to think the least amount as possible.

docker/README.md
```diff
@@ -0,0 +1,79 @@
...
+### Run step
+
+**Pre-requisite:** You must have an already running DC/OS cluster.
+
+1. Start docker on your machine
+
+2. Set cluster URL. The address of a DC/OS cluster to be used as the backend. The cluster version must be compatible with the UI assets.
+
+  ```sh
+  export CLUSTER_ADDRESS=https://a-dcos-cluster.us-west-2.elb.amazonaws.com
```
rcorral: That's a good idea for v2!

docker/README.md
rcorral: Check out Neil's comment above https://github.com/dcos/dcos-ui/pull/1824#issuecomment-274692807
This is why I split things out into Opt1, Opt2.

docker/README.md
```diff
@@ -0,0 +1,79 @@
...
+  ```
+
+3. Run the dcos-ui docker image
+
+  **Option 1**. Run without a plugins repo
+  ```sh
+  docker run --ipc=host \
+    -p 127.0.0.1:4200:4200 \
+    -e "CLUSTER_ADDRESS=$CLUSTER_ADDRESS" \
+    -e "APPLICATION_CHECKOUT=master" \
```
rcorral: I was hoping the instructions would describe that this is configurable.
Again, I'm making it as easy as possible for people to be successful, rather than having to think/guess what are possible values and what they should be entering on their first time trying this out. ref: https://github.com/dcos/dcos-ui/pull/1824#issuecomment-274692807

docker/build-dcos-ui
rcorral: But then, if the user misspelled `APPLICATION_CHECKOUT` and didn't want master, they would be confused. I much rather this just fail and they have to fix.
Do you think having a default adds much value?

docker/build-dcos-ui
rcorral: We don't have fixtures for the entire app. This feels way out of scope.

rcorral:
Hey @orlandohohmeier, I agree that mounting the local FS would be helpful and we could reuse this container for developing DC/OS UI among other things. Unfortunately this is not what I set out to solve with this specific container. You can see PR #1880 adds another container to run integration tests. You can see these two containers do different things and have different dependencies. I'm weary about creating a single container that does all sorts of things, instead I would prefer we create a new container for each problem we try to solve.
With this container, I wanted to give Design and Product the ability to build the bleeding edge of the DC/OS UI without having to start a DC/OS cluster or have us create a new pull request against dcos/dcos. I also wanted to make sure that this would withstand the test of time with little to no support to the people using this container. With your suggestion of mounting the local FS, I can see problems that would arise with this. I think a nice adition to the solution I have now, is to add a shell script that will only copy the latest version of the files needed to build the image, rather than having to clone dcos-ui. We could add this enhancement later.
I didn't use docker compose because the only way to pass environment variables to a container in a docker compose file is to have a separate file (at build time, you can pass them as args at run). I felt this added unecessary steps for users, just for the sake of using docker compose. I actually built this using docker compose the first time, but realized it was too much and if I only have one container, I don't need to use it. To add to this, as an example, users would have to write private keys to a file if they're using a private GH repository. You also can't remove build-time environment variables at run-time unless you specifically override them.

Could you expand what you don't like about the number of configuration options and why? I'm trying to make it easy for non-technical users to spin up DC/OS UI. My aim is for Design/Product to easily view progress on Pull Requests so that they can provide feedback.

docker/Dockerfile.browse
rcorral: You're right, no label needed!

docker/Dockerfile.browse
rcorral: I explain this in the comment below, but in short, I want to ensure this container will stand the test of time on user's computers. The goal for this container is to make it easy for non-technical users to build the UI off a given branch that they can get from a PR.

docker/README.md
rcorral: @MatApple I updated the readme. Let me know what you think!

orlandohohmeier:
@rcorral thanks for clarifying the use case! 

Considering the purpose of this container I would vote to move it to a different repository (e.g. `mesosphere/tools`) esp. as the `dcos-ui` repository is already somewhat bloated and the container doesn't really belong to DC/OS UI. Besides that I'm not sure if adding such a container really solves the issue Design and Product have and would much rather invest time in tooling and processes to continuously deliver the latest and greatest instead of providing tools to work around the rather slow integration cycle.

Regarding the configuration I'd prefer a setup where people don't need to configure anything esp. if the audience might not be that tech savvy. The Marathon UI docker-compose setup for examples configures and starts all required components with one simple `docker-compose up`.

How about changing slightly changing the container so that we can run it in the cluster itself similar to the PR preview that @philipnrmn once build for the marathon project.

rcorral:
Closing this PR in favor to move this container into another repo where it makes more sense.

PR #1825, opened 2017-01-20
Duplicable row delete button improvements

philipnrmn:
This PR
 - vertically centres the delete icons on duplicable rows in the service create
form
 - adds a 'delete' tooltip to form group containers in the service create form

Before centering the delete icons:

Before adding tooltips to the group containers:
https://cl.ly/0V46200O1L08

### DIFF
--- a/plugins/services/src/js/components/forms/GeneralServiceFormSection.js
+++ b/plugins/services/src/js/components/forms/GeneralServiceFormSection.js
@@ -1,3 +1,4 @@
+import classNames from 'classnames';
 import React, {Component} from 'react';
 import {Confirm, Tooltip} from 'reactjs-components';
 
@@ -220,6 +221,11 @@ class GeneralServiceFormSection extends Component {
         );
       }
 
+      const deleteRowButtonClassNames = classNames(
+        'column-2 flush-left',
+        {'form-group-without-top-label': padDeleteButton}
+      );
+
       return (
         <FormRow key={index}>
           <FormGroup
@@ -276,10 +282,7 @@ class GeneralServiceFormSection extends Component {
             </FieldError>
           </FormGroup>
 
-          <FormGroup className={{
-            'column-2 flush-left': true,
-            'form-group-without-top-label': padDeleteButton
-          }}>
+          <FormGroup className={deleteRowButtonClassNames}>
             <DeleteRowButton
               onClick={this.props.onRemoveItem.bind(this,
                 {value: index, path: 'constraints'})} />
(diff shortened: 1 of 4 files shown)


src/js/components/form/FormGroupContainer.js
```diff
@@ -7,10 +8,16 @@ const FormGroupContainer = (props) => {
   if (props.onRemove != null) {
     removeButton = (
       <div className="form-group-container-action-button-group">
-        <a className="button button-primary-link"
-          onClick={props.onRemove}>
-          <Icon id="close" color="grey" size="tiny" family="tiny"/>
-        </a>
+        <Tooltip content="Delete"
+          interactive={true}
```
jfurrow: Let's remove the `interactive` property (it defaults to `false`), since there's nothing in the tooltip to interact with.

src/styles/components/form-elements/form-row/styles.less
```diff
@@ -14,6 +14,11 @@
       margin-bottom: 0;
     }
   }
+
+  // Add top-padding equivalent to a label
+  .form-group-without-top-label {
+    padding-top: 1.82rem;
+  }
```
jfurrow: I'd like it if we could construct this value based on the CNVS variables. So that would be as follows: `@label-line-height + @label-margin-bottom`. Let's also make sure to add the breakpoints as necessary.

CNVS doesn't change the label line height for any breakpoint, but it does change the margin for breakpoints `medium` and `large`, so we just need to account for those two.

Line-height variable: https://github.com/mesosphere/cnvs/blob/develop/styles/components/form/label/variables.less#L107
Margin variables: https://github.com/mesosphere/cnvs/blob/develop/styles/components/form/label/variables.less#L129-L133

plugins/services/src/js/components/forms/GeneralServiceFormSection.js
```diff
@@ -274,7 +276,10 @@ class GeneralServiceFormSection extends Component {
             </FieldError>
           </FormGroup>
 
-          <FormGroup className="flex flex-item-align-end column-2 flush-left">
+          <FormGroup className={{
```
MatApple: It's nice that we get this natively now! 🎉

nLight: It's rather confusing actually. Because it is not a native behavior, but `FormGroup` component that uses `classNames` library underneath. I'm not convinced it is a good idea to keep it like that.

MatApple: Ahh, yes I agree @nLight - thought @philipnrmn had discovered some React wizardry (it would be nice if it was native). I think we should avoid this pattern too if we're enabling it in another component.

src/styles/components/form-elements/form-row/styles.less
philipnrmn: Ah, this makes more sense (I didn't much like the magic number either). 

Thanks for the very clear explanation and references! 🙌

philipnrmn:
Feedback is addressed :)

jfurrow:
Looks great!

MatApple:
run integration tests

PR #1828, opened 2017-01-20
[Service Modal Routes 1 of 3] DCOS-12423: Adding routes to the service create modal

mlunoe:
Service Modal Routes 1 of 3: https://github.com/dcos/dcos-ui/pull/1828
Service Modal Routes 2a of 3: https://github.com/dcos/dcos-ui/pull/1853
Service Modal Routes 2b of 3: https://github.com/dcos/dcos-ui/pull/1871

Partly closes: DCOS-12423

Essentially this PR should not make a difference in functionality, other than the `NewCreateServiceModal` now is routed with `/create` and `/edit:version` routes.

One thing to note is that it is possible to go to a group route and then add `/edit` at the end, which will open the modal in an edit mode. It is possible to change the group into an application if it is empty, otherwise Marathon will complain and provide a reasonable error message:

Since this is actually possible, I chose to leave it as is. Also, I don't expect many people will enter that route manually. If they do and feel like what happens is not what they expect we can create an issue and take it from there.

### DIFF
--- a/plugins/services/src/js/components/modals/NewCreateServiceModal.js
+++ b/plugins/services/src/js/components/modals/NewCreateServiceModal.js
@@ -74,12 +74,17 @@ class NewCreateServiceModal extends Component {
     const isEdit = location.pathname.includes('/edit');
     const serviceID = decodeURIComponent(params.id || '/');
     const service = isEdit ? DCOSStore.serviceTree.findItemById(serviceID) : null;
+    const isSpecificVersion = service instanceof Application && params.version;
     let serviceConfig = new Application(
       Object.assign({id: getBaseID(serviceID)}, DEFAULT_APP_SPEC)
     );
 
-    if (isEdit && service instanceof Service) {
+    if (isEdit && service instanceof Service && !isSpecificVersion) {
       serviceConfig = service.getSpec();
+    }
+
+    if (isEdit && isSpecificVersion) {
+      serviceConfig = service.getVersions().get(params.version);
     }
 
     this.state = {
--- a/plugins/services/src/js/components/modals/NewCreateServiceModal.js
+++ b/plugins/services/src/js/components/modals/NewCreateServiceModal.js
@@ -197,6 +197,8 @@ class NewCreateServiceModal extends Component {
       MARATHON_SERVICE_EDIT_SUCCESS,
       this.onMarathonStoreServiceEditSuccess
     );
+
+    // Also remove DCOS change listener, if still subscribed
     DCOSStore.removeChangeListener(DCOS_CHANGE, this.handleStoreChange);
   }
 
--- a/plugins/services/src/js/components/modals/NewCreateServiceModal.js
+++ b/plugins/services/src/js/components/modals/NewCreateServiceModal.js
@@ -71,7 +71,7 @@ class NewCreateServiceModal extends Component {
     super(...arguments);
 
     const {location, params} = this.props;
-    const isEdit = location.pathname.includes('/edit');
+    const isEdit = this.isLocationEdit(location);
     const serviceID = decodeURIComponent(params.id || '/');
     const service = isEdit ? DCOSStore.serviceTree.findItemById(serviceID) : null;
     const isSpecificVersion = service instanceof Application && params.version;
@@ -139,7 +139,7 @@ class NewCreateServiceModal extends Component {
       return;
     }
 
-    const isEdit = nextProps.location.pathname.includes('/edit');
+    const isEdit = this.isLocationEdit(nextProps.location);
     const serviceID = decodeURIComponent(nextProps.params.id || '/');
     const service = isEdit ? DCOSStore.serviceTree.findItemById(serviceID) : null;
     let serviceConfig = new Application(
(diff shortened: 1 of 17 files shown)


mlunoe:
Run integration tests

mlunoe:
Run integration tests

mlunoe:
Run integration tests

mlunoe:
Run integration tests

mlunoe:
Run integration tests

mlunoe:
Run integration tests

mlunoe:
Retest this please

MatApple:
After successful creation of a service, the modal closes. For some reason polling for updates to the services page stops. So your new service doesn't show up unless you navigate away and back again.

mlunoe:
Yup, I noticed that myself a little earlier. Working on a fix

mlunoe:
Found and fixed problem @MatApple!

mlunoe:
Run unit tests

plugins/services/src/js/components/modals/NewCreateServiceModal.js
```diff
@@ -74,14 +74,19 @@ class NewCreateServiceModal extends Component {
     const isEdit = location.pathname.includes('/edit');
     const serviceID = decodeURIComponent(params.id || '/');
     const service = isEdit ? DCOSStore.serviceTree.findItemById(serviceID) : null;
+    const isSpecificVersion = service instanceof Application && params.version;
```
MatApple: Are you sure we need to check if it's an instance of Application?

mlunoe: Yup, only `Application` has the `getVersions` method. Wanted to make sure that was the case before we called it. `Service` does not, and this could also be a `ServiceTree` (group).

plugins/services/src/js/components/modals/NewCreateServiceModal.js
```diff
@@ -123,7 +260,8 @@ class NewServiceFormModal extends Component {
     }
 
     // Close if picker is open, or if editing a service in the form
-    if (servicePickerActive || (!serviceReviewActive && this.props.isEdit)) {
+    if (servicePickerActive ||
+      (!serviceReviewActive && location.pathname.includes('/edit'))) {
```
nLight: wonder if it's possible to forge such an app name so that it will give false positives here 🤔

plugins/services/src/js/components/modals/NewCreateServiceModal.js
```diff
@@ -255,15 +399,11 @@ class NewServiceFormModal extends Component {
...
-      serviceName = `"${serviceName}"`;
-    } else {
-      serviceName = 'Service';
-    }
+    const {location} = this.props;
+    const {service} = this.state;
+    const serviceName = service ? `"${service.getName()}"` : 'Service';
 
-    if (isEdit) {
+    if (location.pathname.includes('/edit')) {
```
nLight: do we want `isLocationEdit()` function?

nLight:
giphy-2

plugins/services/src/js/components/modals/NewCreateServiceModal.js
MatApple: ➕  1 (seeing it's used > 1 times)

mlunoe: Fixed!

plugins/services/src/js/components/modals/NewCreateServiceModal.js
mlunoe: Fixed!

MatApple:
run integration tests

mlunoe:
run integration tests

mlunoe:
The original solution I tried with a regex capturing the end of the url didn't work as we have these random hashes at the end `?_k=0dtyxm`, which will be hard to capture, so decided to go back to the `includes` solution as you have to encode a the service name, so it can never contain a slash.

nLight:
Hey @mlunoe,
this is the value I get from `this.props.location` for `http://0.0.0.0:4200/#/services/overview?_k=pm0m5u`, note that random hash goes to `key` while `pathname` doesn't have it
 
```json
{
"pathname":"/services/overview",
"search":"",
"hash":"",
"state":null,
"action":"POP",
"key":"pm0m5u",
"query":{},"$searchBase":{"search":"","searchBase":""}}
```

nLight:
Holding merge so we don't accidentally merge it until @mlunoe checks `location.pathname`.
Feel free to un-hold it.

mlunoe:
I had an integration test fail on me because we weren't capturing all scenarios. Also, it is still fragile since `/edit` and `/edit/` are both valid. I'd rather go with less restrictive and make sure it works. We don't know if we'll have additions to this route in the near future.

jfurrow:
It seems we've lost the closing animation of the modal here:
https://cl.ly/2K3K183k471i/Screen%20Recording%202017-01-30%20at%2003.53%20PM.mov

mlunoe:
Good catch! I guess I must have removed my solution for it. Re-added the timer for the animation to finish before navigating away

PR #1830, opened 2017-01-20
DCOS-13362: [1/7] Error Normalisation: Introduce MarathonErrorUtil

wavesoft:
This utility exposes the `MarathonErrorUtil.parseErrors` function that normalises whatever marathon error response into an internal representation of an array of error messages.

_This PR implements the `Marathon Error Parsing` task in the spec document: https://docs.google.com/document/d/1kwdPxBo3UlRZvMxLLIUqSZqLlUJitEuIAWderUgWY1c/edit#heading=h.thdpdh4mx6uj_

### DIFF
--- a/plugins/services/src/js/utils/MarathonErrorUtil.js
+++ b/plugins/services/src/js/utils/MarathonErrorUtil.js
@@ -1,0 +1,121 @@
+import ValidatorUtil from '../../../../../src/js/utils/ValidatorUtil';
+
+const MarathonErrorUtil = {
+
+  /**
+   * Try to guess an error type from it's contents
+   *
+   * @param {String} message - The marathon error string
+   * @returns {String} Returns the message type
+   */
+  /* eslint-disable no-unused-vars */
+  getErrorType(message) {
+  /* eslint-enable no-unused-vars */
+
+    // TODO: This will be implemented on the next PR
+    return 'MARATHON_ERROR';
+
+  },
+
+  /**
+   * Parses errros from the given input format into the common error format
+   * used internally
+   *
+   * For reference, marathon responds an error according to the exception that
+   * occurred and it's translated with the following checks:
+   *
+   * https://github.com/mesosphere/marathon/blob/master/src/main/scala/mesosphere/marathon/api/MarathonExceptionMapper.scala#L63
+   *
+   * @param {Object|String} error - The marathon error(s) to process
+   * @returns {Array} Returns an array of error objects
+   */
+  parseErrors(error) {
+    if (error == null) {
+      return [];
+    }
+
+    // `error` could be a string, if the error has no details
+    if (typeof error === 'string') {
+      return [
+        {
+          path: [],
+          message: error,
+          type: MarathonErrorUtil.getErrorType(error),
+          variables: {}
+        }
+      ];
+    }
+
+    // `details` can be missing
+    if (!error.details) {
+      if (!error.message) {
+        return [];
+      }
+
+      return [
+        {
+          path: [],
+          message: error.message,
+          type: MarathonErrorUtil.getErrorType(error.message),
+          variables: {}
+        }
+      ];
+    }
+
+    // `details` can be a string
+    if (typeof error.details === 'string') {
+      return [
+        {
+          path: [],
+          message: error.details,
+          type: MarathonErrorUtil.getErrorType(error.details),
+          variables: {}
+        }
+      ];
+
+    }
+
+    // `details` can be an array of errors
+    return error.details.reduce(function (memo, {errors, path}) {
+
+      // Convert marathon path components to a dot-separated string
+      // and then split it into an array
+      //
+      // A marathon path looks like: /container(0)/containerPath
+      //
+      const pathString = path
+        .replace(/\//g, '.')
+        .replace(/\((\d+)\)/g, (_, id) => `.${id}`)
+        .slice(1);
+
+      // Don't create array with empty first item when we have an empty path
+      let pathComponents = [];
+      if (pathString !== '') {
+        pathComponents = pathString
+          .split('.')
+          .map(function (component) {
+            if (ValidatorUtil.isNumber(component)) {
+              return Number(component);
+            }
+
+            return component;
+          });
+      }
+
+      // For every error, create the correct message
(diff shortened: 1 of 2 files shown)


wavesoft:
run integration tests

wavesoft:
run unit tests

plugins/services/src/js/utils/MarathonErrorUtil.js
```diff
@@ -0,0 +1,121 @@
...
+          message: error,
+          type: MarathonErrorUtil.getErrorType(error),
+          variables: {}
+        }
+      ];
+    }
+
+    // `details` can be missing
+    if (!error.details) {
+      if (!error.message) {
```
mlunoe: Let's separate these with one level `if`-statements:

```js
if (!error.details && !error.message) {
  return [];
}

if (!error.details && error.message) {
  return [
    {
      path: [],
      message: error.message,
      type: this.getErrorType(error.message),
      variables: {}
    }
  ];
}
```

plugins/services/src/js/utils/MarathonErrorUtil.js
```diff
@@ -0,0 +1,121 @@
...
+    // `details` can be an array of errors
+    return error.details.reduce(function (memo, {errors, path}) {
+
+      // Convert marathon path components to a dot-separated string
+      // and then split it into an array
+      //
+      // A marathon path looks like: /container(0)/containerPath
+      //
+      const pathString = path
+        .replace(/\//g, '.')
```
mlunoe: Is this part really necessary? Couldn't we just split on `/`?

plugins/services/src/js/utils/MarathonErrorUtil.js
```diff
@@ -0,0 +1,121 @@
...
+            return component;
+          });
+      }
+
+      // For every error, create the correct message
+      return errors.reduce(function (memo, errorMessage) {
+        memo.push({
+          path: pathComponents,
+          message: errorMessage,
+          type: MarathonErrorUtil.getErrorType(errorMessage),
```
mlunoe:
```js
return errors.reduce((memo, errorMessage) => {
  ...
  type: this.getErrorType(errorMessage),
  ...
}, []);
```

mlunoe:
Other than that, I think it looks good!

plugins/services/src/js/utils/MarathonErrorUtil.js
```diff
@@ -0,0 +1,121 @@
...
+      return [];
+    }
+
+    // `error` could be a string, if the error has no details
+    if (typeof error === 'string') {
+      return [
+        {
+          path: [],
+          message: error,
+          type: MarathonErrorUtil.getErrorType(error),
```
wavesoft: Hey @mlunoe , I would vote against this principle in all of our static utilities. These are not methods, but static functions and `this` should be considered "undefined".

plugins/services/src/js/utils/MarathonErrorUtil.js
```diff
@@ -0,0 +1,121 @@
...
+
+const MarathonErrorUtil = {
+
+  /**
+   * Try to guess an error type from it's contents
+   *
+   * @param {String} message - The marathon error string
+   * @returns {String} Returns the message type
+   */
+  /* eslint-disable no-unused-vars */
```
wavesoft: Didn't know that. Nice one 👍

plugins/services/src/js/utils/MarathonErrorUtil.js
wavesoft: Good point 👍

plugins/services/src/js/utils/MarathonErrorUtil.js
wavesoft: D-oh! Good point...

plugins/services/src/js/utils/MarathonErrorUtil.js
```diff
@@ -0,0 +1,121 @@
...
+  getErrorType(message) {
+  /* eslint-enable no-unused-vars */
+
+    // TODO: This will be implemented on the next PR
+    return 'MARATHON_ERROR';
```
wavesoft: We don't have actual types yet, but will definitely do so. 👍

plugins/services/src/js/utils/MarathonErrorUtil.js
wavesoft: Not going to fix this, per my last comment, unless you have any objection.

wavesoft:
run unit tests

wavesoft:
run unit tests

plugins/services/src/js/utils/MarathonErrorUtil.js
```diff
@@ -0,0 +1,120 @@
...
+  // eslint-disable-next-line no-unused-vars
+  getErrorType(message) {
+
+    // TODO: This will be implemented on the next PR
+    return 'MARATHON_ERROR';
+
+  },
+
+  /**
+   * Parses errros from the given input format into the common error format
```
nLight: spelling `errors`

mlunoe:
Right on! Looks good man!

plugins/services/src/js/utils/MarathonErrorUtil.js
nLight: Using `this` will make it impossible importing just that function without the whole package like 
`import {parseErrors} from ...;` so let's keep `MarathonErrorUtil` as it is.

plugins/services/src/js/utils/MarathonErrorUtil.js
```diff
@@ -0,0 +1,120 @@
...
+          path: [],
+          message: error,
+          type: MarathonErrorUtil.getErrorType(error),
+          variables: {}
+        }
+      ];
+    }
+
+    // Both `details` and `errors` can be missing
+    if (!error.details && !error.message) {
```
mlunoe: Argh, dang. I missed this the first time I ran through this.
Apparently, Marathon can emit errors with the form `{message: null}` (thanks, Marathon!) this means that there was an error, we just don't have a description for it.
We need to handle this case @wavesoft. I created an issue here: https://jira.mesosphere.com/browse/DCOS_OSS-653

PR #1831, opened 2017-01-20
Fix Ace editor tooltip positioning

orlandohohmeier:
Change the fullscreen modal side panel positioning to use position `right` instead of transform `translateX` as `transform` creates a new stacking context and the containing block is changing the context which results in "wrong" positions.

For more details please see:
https://drafts.csswg.org/css-transforms-1/#transform-rendering

With this change in place we no longer need the hacks to correct the Ace editor tooltip position, as position fixed (used to position the tooltip) is working as expected.

### DIFF
--- a/plugins/services/src/js/components/modals/CreateServiceJsonOnly.js
+++ b/plugins/services/src/js/components/modals/CreateServiceJsonOnly.js
@@ -114,12 +114,6 @@ class CreateServiceJsonOnly extends React.Component {
   render() {
     let {appConfig, errorList} = this.state;
 
-    // Note: The `transform` parameter is just a hack to properly align the
-    // error message.
-    let editorStyles = {
-      transform: 'translateX(0)'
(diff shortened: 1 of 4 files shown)


wavesoft:
Take that hack out!! 💯  ✨  

Cleanup

jfurrow:
+1000 thanks for doing this!

orlandohohmeier:
@jfurrow why do you want to use pixel values instead of percent?

jfurrow:
@orlandohohmeier I pushed a commit, hope you don't mind. The `right: -100%` value for the side panel needs to change at each breakpoint, because the `100%` refers to the whole width of the modal, rather than the width of the side panel itself.

jfurrow:
I'll let someone else merge since I had a hand in this as well :)

wavesoft:
Looks good 👍

PR #1834, opened 2017-01-20
DCOS-13362: [4/7] Error Normalisation: Concentrate error handling

wavesoft:
---
~⚠️  This PR depends on #1833 - ( [See the diff](https://github.com/dcos/dcos-ui/compare/ic/feat/error-message-refactoring-3...ic/feat/error-message-refactoring-4?expand=1) )~

~⚠️ This PR depends on https://github.com/dcos/dcos-ui/pull/1848 (raml-validator-loader v0.1.8)~

---
 
This commit includes the following:

- Adds NewCreateServiceModal.getAllErrors function that is collecting
  all errors, including validation and marathon errors
- Removes validation from NewCreateServiceModalForm & CreateServiceJsonOnly
- Fixes syling of the alert dialog in form and the review screen

**Note that this PR makes the error validation more "aggressive", meaning that you will instantly see an error when you edit a form field. This will be fixed on the next PR**

_This PR implements the `Normalize internal error structure` task in the spec document: https://docs.google.com/document/d/1kwdPxBo3UlRZvMxLLIUqSZqLlUJitEuIAWderUgWY1c/edit#heading=h.thdpdh4mx6uj_

PR #1836, opened 2017-01-20
DCOS-13362: [5/7] Error Normalisation: Show errors only when needed

wavesoft:
---
⚠️  ~~This PR depends on #1834 - ( [See the diff](https://github.com/dcos/dcos-ui/compare/ic/feat/error-message-refactoring-4...ic/feat/error-message-refactoring-5?expand=1) )~~

---
 
new-form-errors mov

This commit introduces a series of changes in order to show errors
in the most optimal way. In a bit more detail:

- Marathon errors now persist after the review screen, back to the form, until the user has edited the form.
- The "Review & Run" button is now always enabled, even if there are errors in the form.
- NewCreateServiceModalForm tracks which fields the user has edited and shows real-time validation errors only on them. The rest of the errors are muted.
- When the user clicks "Review & Run", a showAllErrors= bypass flag is set and forces the form to display all errors. Not only in-line, next to the fields, but also to the alert on top.
- When the user edits a field with error the validation is temporarily disabled until the user blurs away.
- When the form is changed, the showAllErrors= bypass flag is reset, along with all marathon errors and it returns on it's normal behaviour.

_This PR implements a logic described in the spec document: https://docs.google.com/document/d/1kwdPxBo3UlRZvMxLLIUqSZqLlUJitEuIAWderUgWY1c/edit#heading=h.thdpdh4mx6uj_

Closes DCOS-12422

### DIFF
--- a/plugins/services/src/js/components/modals/NewCreateServiceModal.js
+++ b/plugins/services/src/js/components/modals/NewCreateServiceModal.js
@@ -410,9 +410,7 @@ class NewCreateServiceModal extends Component {
    * @returns {Array} - An array of error objects
    */
   getAllErrors() {
-    const {apiErrors} = this.state;
-
-    return apiErrors.concat(this.getFormErrors());
+    return this.state.apiErrors.concat(this.getFormErrors());
   }
 
   getHeader() {
--- a/plugins/services/src/js/components/modals/NewCreateServiceModal.js
+++ b/plugins/services/src/js/components/modals/NewCreateServiceModal.js
@@ -205,11 +205,18 @@ class NewCreateServiceModal extends Component {
     } = this.state;
 
     if (serviceReviewActive) {
+      // Remove the 'Application is deploying' error when we havigate back
+      // since it's not related to the form
+      const apiErrors = this.state.apiErrors.filter(function (error) {
+        return error.type !== ServiceErrorTypes.SERVICE_DEPLOYING;
+      });
+
       // Just hide review screen. Form or JSON mode will be
       // activated automatically depending on their last state
       this.setState({
-        serviceReviewActive: false,
-        activeTab: tabViewID
+        activeTab: tabViewID,
+        apiErrors,
+        serviceReviewActive: false
       });
 
       return;
@@ -248,7 +255,10 @@ class NewCreateServiceModal extends Component {
   }
 
   handleClearError() {
-    this.setState({apiErrors: []});
+    this.setState({
+      apiErrors: [],
+      showAllErrors: false
+    });
   }
 
   handleClose() {
@@ -269,7 +279,11 @@ class NewCreateServiceModal extends Component {
   }
 
   handleServiceChange(newService) {
-    this.setState({serviceConfig: newService});
+    this.setState({
+      apiErrors: [],
+      serviceConfig: newService,
+      showAllErrors: false
+    });
   }
 
   handleServiceErrorsChange(errors) {
@@ -324,6 +338,8 @@ class NewCreateServiceModal extends Component {
     const errors = this.getAllErrors();
     if (errors.length === 0) {
       this.setState({serviceReviewActive: true});
+    } else {
+      this.setState({showAllErrors: true});
     }
   }
 
@@ -438,7 +454,6 @@ class NewCreateServiceModal extends Component {
             <ServiceConfigDisplay
               onEditClick={this.handleGoBack}
               appConfig={serviceConfig}
-              clearError={this.handleClearError}
               errors={this.getAllErrors()} />
           </div>
         </div>
@@ -454,6 +469,7 @@ class NewCreateServiceModal extends Component {
 
     if (serviceFormActive) {
       const {location} = this.props;
+      const {showAllErrors} = this.state;
 
       const SECTIONS = [
         ContainerServiceFormSection,
@@ -496,15 +512,16 @@ class NewCreateServiceModal extends Component {
           jsonConfigReducers={jsonConfigReducers}
           handleTabChange={this.handleTabChange}
           inputConfigReducers={inputConfigReducers}
+          isEdit={this.isLocationEdit(location)}
           isJSONModeActive={isJSONModeActive}
           ref={(ref) => {
             return this.createComponent = ref;
           }}
-          service={serviceConfig}
           onChange={this.handleServiceChange}
           onConvertToPod={this.handleConvertToPod}
           onErrorsChange={this.handleServiceErrorsChange}
-          isEdit={this.isLocationEdit(location)} />
+          service={serviceConfig}
+          showAllErrors={showAllErrors} />
       );
     }
 
@@ -557,7 +574,7 @@ class NewCreateServiceModal extends Component {
     }
 
     if (serviceFormActive) {
-      const errors = this.getAllErrors();
+      // const errors = this.getAllErrors();
 
       return [
         {
@@ -575,20 +592,22 @@ class NewCreateServiceModal extends Component {
         {
           className: 'button-primary flush-vertical',
           clickHandler: this.handleServiceReview,
-          disabled: errors.length !== 0,
+          disabled: false,
+          // disabled: errors.length !== 0,
           label: 'Review & Run'
         }
       ];
     }
 
     if (serviceJsonActive) {
-      const errors = this.getAllErrors();
+      // const errors = this.getAllErrors();
 
       return [
         {
           className: 'button-primary flush-vertical',
           clickHandler: this.handleServiceReview,
-          disabled: errors.length !== 0,
+          disabled: false,
+          // disabled: errors.length !== 0,
           label: 'Review & Run'
         }
       ];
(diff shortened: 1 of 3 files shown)


mlunoe:
This is a more robust diff link compares the two branches: https://github.com/dcos/dcos-ui/compare/ic/feat/error-message-refactoring-4...ic/feat/error-message-refactoring-5?expand=1

wavesoft:
run integration tests

rcorral:
@wavesoft I have a couple questions for you:
* **The "Review & Run" button is now always enabled, even if there are errors in the form.** – Could you confirm this is what we want with Design if you haven't already?
* On your gif, it looks like the error on the "Instances" field shows up before the errors show up on top.

wavesoft:
Hey @rcorral:

1. Yes, that's the idea. Since (1) we don't want to show the errors in the field unless the user has actually visited it and (2) we want somehow the trigger to perform full validation: the Review & Run is now always enabled and it always shows the full set of errors in the stack when clicked -if any-.
2. When the user edits a field, its errors are now visible. This means that when the user blurs away from the field, the errors show up instantly. Regardless, if you try to submit a form with errors, the error will also appear on the global error section. (This is particularly helpful atm since we don't have  error indicators on the form tabs)

rcorral:
1. Makes sense. Would it also make sense that if there are errors, to keep the user in the form and not allow them to move forward to Review? Potentially also disabling the button since we know errors exist.
2. Can we get the errors to show up in the General section whenever they show up on the field? It seems weird for them not to show up at the same time.

wavesoft:
1. Already the user is not allowed to move away from the form. Clicking "Review" is just the trigger for the enforced validation. Disabling the button (as we do now) is going to be a bit cryptic, since it's going to be disabled, but no errors will be highlighted in the form (since the user has to use the field first).
2. I would vote against this, since this would cause an annoying shifting of the contents when the user blurs away from a field with errors.

wavesoft:
run integration tests

wavesoft:
So, after a discussion with @rcorral and @mlunoe we realised that the problem is mainly when the alert bubble is shown. We agreed that for consistency, after the user has clicked "Review", the bubble should persist until all the errors are resolved. API errors should be reset only when the user clicks "Review" and there are no validation errors.

plugins/services/src/js/components/modals/NewCreateServiceModalForm.js
```diff
@@ -422,10 +451,38 @@ class NewCreateServiceModalForm extends Component {
...
+      const errorPath = error.path.join('.');
+
+      // Always mute the error on the field we are editing
+      if ((editingFieldPath != null) && (errorPath === editingFieldPath)) {
+        return false;
+      }
+
+      // Never mute fields in the CONSTANTLY_UNMUTED_ERRORS fields
+      const isUnmuted = CONSTANTLY_UNMUTED_ERRORS.some(function (rule) {
+        return rule.exec(errorPath);
```
orlandohohmeier: Please use `RegExp.prototype.test()` if you only need to know wether there was a match or not.

plugins/services/src/js/utils/MarathonErrorUtil.js
```diff
@@ -11,9 +12,12 @@ const MarathonErrorUtil = {
   // eslint-disable-next-line no-unused-vars
   getErrorType(message) {
 
-    // TODO: This will be implemented on the next PR
-    return 'MARATHON_ERROR';
+    // Check for 'service is deploying' error messages
+    if (/force=true/.exec(message)) {
```
orlandohohmeier: Please use`RegExp.prototype.test()` to avoid unnecessary casting.

plugins/services/src/js/components/modals/NewCreateServiceModalForm.js
```diff
@@ -422,10 +451,38 @@ class NewCreateServiceModalForm extends Component {
     ];
   }
 
+  /**
+   * This function filters the error list in order to keep only the
+   * errors that should be displayed to the UI.
+   *
+   * @returns {Array} - Returns an array of errors that passed the filter
+   */
+  getUnmutedErrors() {
```
wavesoft: Hmm.. this is very specific case to `NewCreateServiceModalForm`. Which part do you think might be good to be separated?

wavesoft: Plus, `CONSTANTLY_UNMUTED_ERRORS` is a temporary solution and re-using it should not be encouraged.

orlandohohmeier: I was hoping that we can generalize it (e.g. `filterErrorsByPaths`) and get rid of the `CONSTANTLY_UNMUTED_ERRORS` solution. This definitely not blocking this PR but we should only add methods to a component if the method actually belongs to the component and not if it's only used by the component. This will help us to better separate things.

wavesoft: I would vote against `filterErrorsByPaths` because it goes towards the wrong direction. I think we should focus on properly processing the errors through the reducers first.

When we get rid of `CONSTANTLY_UNMUTED_ERRORS`, having a `getRootErrors` utility function might make sense, even though it's logic is quite minimal.

orlandohohmeier:
Adding a `hold merge` until #1834 is on master... v

orlandohohmeier:
✨ Awesome job!

orlandohohmeier:
run integration tests

orlandohohmeier:
run unit tests

wavesoft:
retest this please

orlandohohmeier:

plugins/services/src/js/components/modals/NewCreateServiceModal.js
```diff
@@ -371,11 +400,19 @@ class NewCreateServiceModal extends Component {
...
+  }
+
+  /**
+   * This function combines the errors received from marathon and the errors
+   * produced by the form into a unified error array
+   *
+   * @returns {Array} - An array of error objects
+   */
+  getAllErrors() {
+    const {apiErrors} = this.state;
```
nLight: I wouldn't destruct here (nit)

plugins/services/src/js/components/modals/NewCreateServiceModal.js
```diff
@@ -350,8 +379,8 @@ class NewCreateServiceModal extends Component {
    *
    * @returns {Array} - An array of error objects
    */
-  getAllErrors() {
-    const {apiErrors, serviceFormErrors, serviceConfig} = this.state;
+  getFormErrors() {
```
nLight: ❤️  the rename

plugins/services/src/js/components/modals/NewCreateServiceModal.js
```diff
@@ -575,20 +611,18 @@ class NewCreateServiceModal extends Component {
         {
           className: 'button-primary flush-vertical',
           clickHandler: this.handleServiceReview,
-          disabled: errors.length !== 0,
+          disabled: false,
```
nLight: What if we won't pass any value here? Shall we maybe refactor `FullScreenModalHeaderActions` so that we don't need to pass `disabled: false`

plugins/services/src/js/components/modals/NewCreateServiceModal.js
```diff
@@ -575,20 +611,18 @@ class NewCreateServiceModal extends Component {
...
 
     if (serviceJsonActive) {
-      const errors = this.getAllErrors();
-
       return [
         {
           className: 'button-primary flush-vertical',
           clickHandler: this.handleServiceReview,
-          disabled: errors.length !== 0,
+          disabled: false,
```
nLight: and here

plugins/services/src/js/components/modals/NewCreateServiceModalForm.js
```diff
@@ -90,8 +111,13 @@ class NewCreateServiceModalForm extends Component {
     this.props.onConvertToPod(this.getAppConfig());
   }
 
-  componentDidUpdate() {
-    this.props.onChange(new this.props.service.constructor(this.state.appConfig));
+  componentDidUpdate(prevProps, prevState) {
+    const {editingFieldPath} = this.state;
```
nLight: 🤔  you don't destruct `appConfig` that is being used twice but you do `editingFieldPath` that is being used once

plugins/services/src/js/components/modals/NewCreateServiceModalForm.js
```diff
@@ -90,8 +111,13 @@ class NewCreateServiceModalForm extends Component {
...
+
+    if ((editingFieldPath === null) &&
+       !deepEqual(this.state.appConfig, prevState.appConfig)) {
```
nLight: Let's destruct stuff so that we fit 80 chars rule 💪

plugins/services/src/js/components/modals/NewCreateServiceModal.js
wavesoft: Same like above, fixed on the 6/7

plugins/services/src/js/components/modals/NewCreateServiceModal.js
wavesoft: Michael found this out and is fixed on the next PR in chain: https://github.com/dcos/dcos-ui/pull/1839#discussion_r98755089

plugins/services/src/js/components/modals/NewCreateServiceModal.js
wavesoft: Yup, true. This was the result of a rebasing fight

wavesoft:
@nLight should be fixed 👍

PR #1839, opened 2017-01-20
DCOS-13362: [6/7] Error Normalisation: Introducing error message translation

wavesoft:
---
⚠️  This PR depends on #1836 - ( [See the diff](https://github.com/dcos/dcos-ui/compare/ic/feat/error-message-refactoring-5...ic/feat/error-message-refactoring-6?expand=1) )

---
 
image

image

This commit introduces the ErrorMessageUtil utility that can translate an array of error messages, with the same input and output format.

It also introduces the basic translation tables for system-wide and service-wide error messages

_This PR implements the `Error Translation` task in the spec document: https://docs.google.com/document/d/1kwdPxBo3UlRZvMxLLIUqSZqLlUJitEuIAWderUgWY1c/edit#heading=h.thdpdh4mx6uj_


mlunoe:
This is a more robust diff link compares the two branches: https://github.com/dcos/dcos-ui/compare/ic/feat/error-message-refactoring-5...ic/feat/error-message-refactoring-6?expand=1

wavesoft:
run integration tests

wavesoft:
run integration tests

src/js/constants/DefaultErrorMessages.js
```diff
@@ -0,0 +1,164 @@
+const DefaultErrorMessages = [
+
+  //
+  // The following rule will match every message path, behaving
+  // like the default sink if a path is not found.
+  //
+  // Remember to concat these error messages when you are composing
```
nLight: Will good old inheritance solve the remembering issue?
Yes, it's not gonna be a plain array, but one less thing to remember

nLight: or a factory that remembers to add Defaults for you?

src/js/utils/__tests__/ErrorMessageUtil-test.js
```diff
@@ -0,0 +1,185 @@
+jest.dontMock('../ErrorMessageUtil');
+
+const ErrorMessageUtil = require('../ErrorMessageUtil');
+
+describe('ErrorMessageUtil', function () {
```
nLight: all the tests work on an empty `path: []` shall we add some diversity?

src/js/constants/DefaultErrorMessages.js
wavesoft: Been there, done that, and it was considered bad idea to write more complicated high-level functions for such a simple task :|

src/js/utils/__tests__/ErrorMessageUtil-test.js
wavesoft: The function is never touching `path` so I didn't thought it would make sense. But you might have a point for actually ensuring that it's not touching it 😄

wavesoft:
run integration tests

mlunoe:
@wavesoft flicts!

plugins/services/src/js/components/modals/NewCreateServiceModal.js
```diff
@@ -28,6 +33,7 @@ import HealthChecksFormSection from '../forms/HealthChecksFormSection';
 import NetworkingFormSection from '../forms/NetworkingFormSection';
 import MultiContainerNetworkingFormSection from '../forms/MultiContainerNetworkingFormSection';
 import MultiContainerVolumesFormSection from '../forms/MultiContainerVolumesFormSection';
+// import ValidatorUtil from '../../../../../../src/js/utils/ValidatorUtil';
```
mlunoe: Let's just remove this then

plugins/services/src/js/components/modals/NewCreateServiceModal.js
```diff
@@ -158,6 +177,11 @@ class NewServiceFormModal extends Component {
     this.setState(this.getResetState());
   }
 
+  handleClearErrors() {
```
mlunoe: Where is this being used?

plugins/services/src/js/components/modals/NewCreateServiceModal.js
```diff
@@ -158,6 +177,11 @@ class NewServiceFormModal extends Component {
...
+    this.props.clearError();
+    this.setState({showAllErrors: false});
```
mlunoe: Let's not set this unless we are navigating back to the form from picking another option or editing a service

plugins/services/src/js/components/modals/NewCreateServiceModal.js
```diff
@@ -238,6 +272,44 @@ class NewServiceFormModal extends Component {
...
   }
 
+  getFormErrors() {
+    const {serviceFormErrors, serviceConfig} = this.state;
+    let validationErrors = [];
+
+    // Validate Application or Pod according to the contents
+    // of the serviceConfig property.
+
+    if (serviceConfig instanceof ApplicationSpec) {
```
mlunoe: Since `serviceConfig` is a spec maybe we should rename it?

plugins/services/src/js/components/modals/NewCreateServiceModal.js
```diff
@@ -460,7 +509,7 @@ class NewServiceFormModal extends Component {
         {
           className: 'button-primary flush-vertical',
           clickHandler: this.handleServiceReview,
-          disabled: this.state.serviceFormHasErrors,
+          disabled: false,
```
mlunoe: If it's always going to be false, let's just remove it

plugins/services/src/js/components/modals/NewCreateServiceModal.js
```diff
@@ -471,7 +520,7 @@ class NewServiceFormModal extends Component {
...
```
mlunoe: same here

plugins/services/src/js/components/modals/NewCreateServiceModalForm.js
```diff
@@ -265,22 +279,46 @@ class NewCreateServiceModalForm extends Component {
...
+        </li>
+      );
+    });
+
+    return (
+      <Alert>
+        <strong>There is an error with your configuration</strong>
+        <div className="pod pod-narrower-left pod-shorter-top flush-bottom">
+          <ul className="list-unstyled short flush-bottom">
+          {errorItems}
```
mlunoe: #nit indentation

src/js/utils/ErrorMessageUtil.js
```diff
@@ -0,0 +1,41 @@
+const MESSAGE_VARIABLE = /\{\{([^}]+)\}\}/g;
```
orlandohohmeier: Please change the pattern to `/\{\{([^{}]+)\}\}/g` to properly handle additional curly braces (e.g `{{{variable}}}`)

src/js/utils/ErrorMessageUtil.js
```diff
@@ -0,0 +1,41 @@
...
+
+      // If there is no translation rule, pass-through the message
+      if (rule == null) {
+        return error;
+      }
+
+      // Return the translated message
+      return {
+        message: rule.message.replace(MESSAGE_VARIABLE, function (match) {
+          return ''+variables[match.slice(2, -2)] || '';
```
orlandohohmeier: Instead of slicing the match to remove the curly braces you could you the capture group directly, as illustrated in the following example:

``` 
function (_match, variable) {
        return variables[variable] || '';
}
```

src/js/utils/ErrorMessageUtil.js
wavesoft: It was initially one brace, then I change it to two braces, but why now you think three is better? I seriously doubt an expression with two braces will appear on an error message...

orlandohohmeier: The problem with the pattern is that it catch malformed variable names like `{{{foo}}}` or `{{bar {foo}}` thus I would change the character group to not match `{` as well as `}`.

wavesoft: Ah, now I see. Sure, why not.

wavesoft:
run tests

wavesoft:
retest this please

wavesoft:
Rebased on latest master

orlandohohmeier:
LGTM! 🚢

PR #1843, opened 2017-01-20
Correct usage of word wrapping style properties

jfurrow:


jfurrow:
run unit tests

philipnrmn:
Sorry to pick at a one-line PR but `break-word` is a nonstandard webkit-only value for `word-break` and causes a warning in Firefox:
screen shot 2017-01-31 at 4 59 15 pm


I don't think this is a big deal, just as long as we're aware of and OK with it.

jfurrow:
@philipnrmn Thanks for catching that! I did more research on these properties and determined we actually don't need to alter the `word-break` property at all. Instead we just need to use `overflow-wrap` and `word-wrap` (which are essentially [the same property](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap)) to tell the browser to wrap text when desired. I fixed this in all instances in our codebase.

nLight:
I think there's an eslint plugin to check on that

nLight:
https://www.npmjs.com/package/eslint-plugin-caniuse

philipnrmn:
lint errors :hammer:

philipnrmn:
Thanks for taking care of the `word-break` thing, @jfurrow! Good to know :)

PR #1844, opened 2017-01-20
Waits till we receive mesos state.json before rendering Node detail.

rcorral:
Sometimes going to the Node detail page would throw an exception because the request for mesos state took too long and we weren't waiting for it before rendering.


rcorral:
run unit tests

wavesoft:
Looks good to me 👍  💯

PR #1845, opened 2017-01-21
Change alignment of services toggle button text

philipnrmn:
This PR updates the toggle button label's alignment, moving it from the right
of the button to the left.

This is implemented as an optional mode as it's still right-aligned in the jobs
modal.

Additionally the button label colour animates grey->black on activation.


jfurrow:
Looks great! Nice touch on the color change, too!

PR #1846, opened 2017-01-21
Fix jobs tooltip styles

philipnrmn:
Before, note overlarge font and unreadable grey colour:

After:


philipnrmn:
Run unit tests

PR #1855, opened 2017-01-24
DCOS-13017: Prevent labels in forms from wrapping to multiple lines

jfurrow:
This PR adds a class to prevent text in labels from overflowing.

I decided not to do this globally for all `labels` because it might not always be desirable (e.g. in the `ToggleButton`, the `overflow: hidden` property clips the toggle indicator). So this is implemented only on elements that make use of `FieldLabel`, which ideally will be all form fields in the future.

Before (look at `Grace Period (s)`):

After (look at `Grace Period (s)`):

### DIFF
--- a/src/js/components/form/FieldLabel.js
+++ b/src/js/components/form/FieldLabel.js
@@ -13,6 +13,7 @@ const FieldLabel = (props) => {
     }
   });
   const classes = classNames(
+    'form-control-label',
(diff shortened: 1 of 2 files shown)


nLight:
LGTM! 💯  💖

jfurrow:
Closing for now because I discovered a problem with tooltips in labels being clipped (I should have noticed this sooner). This will require a lot more effort...

PR #1856, opened 2017-01-24
DCOS-13348: Increase Environment form section column widths

jfurrow:
Before:

After:

### DIFF
--- a/plugins/services/src/js/components/forms/EnvironmentFormSection.js
+++ b/plugins/services/src/js/components/forms/EnvironmentFormSection.js
@@ -30,7 +30,7 @@ class EnvironmentFormSection extends Component {
       return (
         <FormRow key={key}>
           <FormGroup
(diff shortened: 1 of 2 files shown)


nLight:
LGTM! 💯  💖

philipnrmn:
Can we update the widths of the fields in the pods environment section to match?

image

jfurrow:
@philipnrmn they actually use the same component, so the classes are applied to both. The multi-container env section was more narrow because the form's parent didn't have the right `flex-grow` property set, but that was fixed by #1861. I just rebased this to include those changes :)

philipnrmn:
Ah, I see. I was wondering why the same component had two different widths.

philipnrmn:
lint errors :hammer:

jfurrow:
Ohh thank oyu @philipnrmn!

PR #1858, opened 2017-01-26
DCOS-11559: Keep host ports when switching a container network to HOST

nLight:
⚠️  Some 1.8 codebase here, PR targets 1.8

---

When we have a container app with a Bridge network and want to go to HOST network the UI would just discard ports from `portDefinitions` and put `containerPort` instead which can result in a conflict.

This PR ensures that the UI, as requested, keeps host ports and discards container ports instead.
Since we don't have access to the `portDefinitions` field in the `ServiceForm` I'm using `servicePort` which is always equal to a `portDefinitions[*].port`

### DIFF
--- a/src/js/components/ServiceForm.js
+++ b/src/js/components/ServiceForm.js
@@ -373,6 +373,19 @@ class ServiceForm extends SchemaForm {
           return port;
         });
       }
+
+      let previousNetworkType = this.props.model.networking.networkType;
+      // When switching a container network from anything to HOST
+      // we want to keep portDefinitions[*].port which is equal to servicePort
+      if (model.networking.networkType === 'host'
+        && previousNetworkType !== 'host') {
+        model.networking.ports.map(function (port) {
+          port.lbPort = port.servicePort;
+
+          return port;
+        });
+      }
+
       let {dockerVolumesDefinition = null} = this.internalStorage_get();
       this.internalStorage_set({model, dockerVolumesDefinition});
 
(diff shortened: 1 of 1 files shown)


nLight:
Looks like 1.8 isn't compatible with our integration tests :(

MatApple:
@nLight you're right - It's possible that by changing the cypress config we might be able to make them work though I expect there will still be more issues because they were switched off for so long and backports may have broken some tests.

MatApple:
So glad we moved away from this implementation! Nice fix @nLight!

rcorral:
Hey @nLight – The fix works in cases where I've set a port using BRIDGE networking, switch to JSON mode, switch to the Form and then select HOST. But if I select BRIDGE, enter a port, and then go to HOST without switching to JSON mode, then the fix doesn't work.
See this gif:
gif

nLight:
@rcorral thanks for pointing out that edge case! Fixed.

src/js/components/ServiceForm.js
```diff
@@ -40,6 +40,8 @@ const FIELDS_TO_WATCH = {
   }
 };
 
+let networkTypeChanged = false;
```
rcorral: @MatApple consider that this fix is for a patch version and this code probably doesn't exist in the same form on our `master` branch.

MatApple: I still think it's important to ship quality for patches :) Since we're keeping a component's state outside of the component, couldn't we potentially end up with an issue after mounting/unmounting lifecycles? I just don't think this is a good practice, unless there is a strong argument for it in this case.

rcorral: I fixed it so that we can all move on.

rcorral:
run unit tests

PR #1859, opened 2017-01-26
Remove obsolete links from README

nLight:
Let's re-add them once we've modularized everything

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,20 +109,6 @@ If you want to add a new npm package to 'node_modules':
     }
     ```
 
-## Module examples
-
-Check out these module types that you can use as a blueprint.
-
-* [Table](https://github.com/dcos/dcos-ui/blob/master/src/js/components/ServicesTable.js)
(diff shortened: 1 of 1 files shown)


nLight:
run integration tests

README.md
```diff
@@ -109,20 +109,6 @@ If you want to add a new npm package to 'node_modules':
     }
     ```
 
-## Module examples
-
-Check out these module types that you can use as a blueprint.
-
-* [Table](https://github.com/dcos/dcos-ui/blob/master/src/js/components/ServicesTable.js)
```
mlunoe: Why do we just remove these instead of updating them?
https://github.com/dcos/dcos-ui/blob/master/plugins/services/src/js/containers/services/ServicesTable.js

nLight: we could but then we're in the middle of modularization and new links won't give a comprehensive picture. If you still think it's better to keep them - I can do so.

mlunoe: I guess we haven't been using them much since they have been out of date for a while.
Ok, let's just remove them, then

PR #1861, opened 2017-01-26
DCOS-13349: Let the form grow freely

nLight:
Before:
screen shot 2017-01-26 at 16 51 06

After:
screen shot 2017-01-26 at 16 50 33

Me:
b7e

### DIFF
--- a/src/styles/views/create-service-modal/styles.less
+++ b/src/styles/views/create-service-modal/styles.less
@@ -19,6 +19,7 @@
 
     &-form {
       display: flex;
+      flex: 1 1 auto;
 
       &-container {
         display: flex;


nLight:
run unit tests

nLight:
run unit tests

jfurrow:
LOL @ image!

This is perfect! This element is a child of a `display: flex` parent, so we need it to grow and shrink in accordance with its siblings (the first and second values of the `flex` shorthand property, respectively). The third value is the `flex-basis` property, and `auto` is exactly what we want because we don't want to make any assumptions about the content width.

nLight:
run unit tests

orlandohohmeier:
✨ Awesome

PR #1864, opened 2017-01-26
DCOS-13456: Fix fluid gemini scrollbar stylesheet handling

orlandohohmeier:
Change the `componentWillUnmount` method to use `Node.prototype.removeChild` as `ChildNode.prototype.remove` is still experimental and not supported by the Internet Explorer.

Closes DCOS-13456


jfurrow:
Nice fix!

PR #1867, opened 2017-01-27
DCOS-13017: Extend FieldLabel's wrapping logic

jfurrow:
This PR extends the `FieldLabel` component in the following ways:
* accepts a `tooltipContent` prop which will render the tooltip after the main content
* accepts a `tooltipIconID` prop to determine the tooltip icon to render (defaults to `circle-question`)
* accepts a `wordWrap` prop to control text wrapping (and if the text wraps, it doesn't get truncated) (defaults to `false`)

Ideally all field label tooltips should be applied in the same manner for consistency.

The following screenshot shows the scenarios that make these props useful:

* `Container Image` and `CPUs` labels are truncated but the tooltip & required indicators are visible (narrow widths applied for the screenshot, normally they don't truncate)
* The two checkboxes make use of the `wordWrap` property to prevent the text from being truncated. A nested `FieldLabel` wraps only the "primary" label and does not use the `wordWrap` property

I wrapped the toggle options in `FieldRow` and `FieldGroup` to give them the proper margin. I also edited the `FieldRow` styles to remove the bottom margin on all `FieldGroup`s which are children of `FieldRow` when it is the last child in its container.

### DIFF
--- a/plugins/services/src/js/components/forms/ContainerServiceFormSection.js
+++ b/plugins/services/src/js/components/forms/ContainerServiceFormSection.js
@@ -1,5 +1,4 @@
 import React, {Component} from 'react';
-import {Tooltip} from 'reactjs-components';
 
 import {FormReducer as ContainerReducer} from '../../reducers/serviceForm/Container';
 import {FormReducer as ContainersReducer} from '../../reducers/serviceForm/Containers';
@@ -48,16 +47,8 @@ class ContainerServiceFormSection extends Component {
     );
 
     return (
-      <FieldLabel>
-        {'Artifact URI '}
-        <Tooltip
-          content={tooltipContent}
-          interactive={true}
-          maxWidth={300}
-          scrollContainer=".gm-scroll-view"
-          wrapText={true}>
-          <Icon color="grey" id="circle-question" size="mini" />
-        </Tooltip>
+      <FieldLabel tooltipContent={tooltipContent}>
+        Artifact URI
       </FieldLabel>
     );
   }
@@ -116,16 +107,12 @@ class ContainerServiceFormSection extends Component {
 
     if (disabled) {
       return (
-        <FieldLabel className="text-no-transform">
-          {'GPUs '}
-          <Tooltip
-            content="Docker Engine does not support GPU resources, please select Universal Container Runtime if you want to use GPU resources."
-            interactive={true}
-            maxWidth={300}
-            scrollContainer=".gm-scroll-view"
-            wrapText={true}>
-            <Icon color="grey" id="lock" size="mini" />
-          </Tooltip>
+        <FieldLabel
+          className="text-no-transform"
+          key="gpus-input-tooltip"
+          tooltipContent="Docker Engine does not support GPU resources, please select Universal Container Runtime if you want to use GPU resources."
+          tooltipIconID="lock">
+          GPUs
         </FieldLabel>
       );
     }
@@ -134,23 +121,6 @@ class ContainerServiceFormSection extends Component {
       <FieldLabel className="text-no-transform">
         GPUs
       </FieldLabel>
-    );
-  }
-
-  getTooltipIfContent(content) {
-    if (!content) {
-      return null;
-    }
-
-    return (
-      <Tooltip
-        content={content}
-        interactive={true}
-        maxWidth={300}
-        scrollContainer=".gm-scroll-view"
-        wrapText={true}>
-        <Icon color="grey" id="lock" size="mini" />
-      </Tooltip>
     );
   }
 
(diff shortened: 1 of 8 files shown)


plugins/services/src/js/components/forms/ContainerServiceFormSection.js
```diff
@@ -290,16 +257,8 @@ class ContainerServiceFormSection extends Component {
...
-        {'Container Image '}
-        <Tooltip
-          content={tooltipContent}
-          interactive={true}
-          wrapText={true}
-          maxWidth={300}
-          scrollContainer=".gm-scroll-view">
-          <Icon color="grey" id={iconID} size="mini" />
-        </Tooltip>
+      <FieldLabel tooltipContent={tooltipContent} tooltipIconID={iconID} key="image-label">
```
mlunoe: Let's break these attributes into separate lines

plugins/services/src/js/components/forms/ContainerServiceFormSection.js
```diff
@@ -169,23 +139,28 @@ class ContainerServiceFormSection extends Component {
...
-            type="checkbox"
-            disabled={containerType !== DOCKER}
-            value={settingName} />
-          {label}
-          {this.getTooltipIfContent(dockerOnly)}
-          <FieldHelp>{helpText}</FieldHelp>
-        </FieldLabel>
+        <FormRow>
+          <FormGroup className="column-12">
+            <FieldLabel key={index} wordWrap>
```
mlunoe: Shouldn't `wordWrap` be set to `={true}`?

plugins/services/src/js/components/forms/GeneralServiceFormSection.js
```diff
@@ -351,16 +343,22 @@ class GeneralServiceFormSection extends Component {
...
-            disabled={isDisabled[runtimeName]}
-            name="container.type"
-            type="radio"
-            value={runtimeName} />
-          {label}
-          <FieldHelp>{helpText}</FieldHelp>
-        </FieldLabel>
+        <FormRow>
+          <FormGroup className="column-12">
+            <FieldLabel className="text-align-left" key={index} wordWrap>
```
mlunoe: Shouldn't `wordWrap` be set to `={true}`?

plugins/services/src/js/components/forms/MultiContainerNetworkingFormSection.js
```diff
@@ -163,7 +154,7 @@ class MultiContainerNetworkingFormSection extends mixin(StoreMixin) {
     return [
       <FormRow key="title">
         <FormGroup className="column-9">
-          <FieldLabel>
+          <FieldLabel wordWrap>
```
mlunoe: Shouldn't `wordWrap` be set to `={true}`?

plugins/services/src/js/components/forms/NetworkingFormSection.js
```diff
@@ -140,7 +132,7 @@ class NetworkingFormSection extends mixin(StoreMixin) {
     return [
       <FormRow key="title">
         <FormGroup className="column-9">
-          <FieldLabel>
+          <FieldLabel wordWrap>
```
mlunoe: Shouldn't `wordWrap` be set to `={true}`?

src/js/components/form/FieldLabel.js
```diff
@@ -1,32 +1,82 @@
...
+    );
   }
 
   const label = (
     <label className={classes}
       {...omit(props, Object.keys(FieldLabel.propTypes))}>
-      {children}
-      {requiredNode}
+      <span className={contentWrapperClasses}>
+        {childNodes}
```
mlunoe: If the children are being cut off by `overflow: hidden` we should display a `title` or something with the full label.

src/js/components/form/FieldLabel.js
```diff
@@ -41,19 +91,27 @@ const FieldLabel = (props) => {
...
+  tooltipContent: React.PropTypes.node,
 
   // Classes
   className: React.PropTypes.oneOfType([
     React.PropTypes.array,
     React.PropTypes.object,
     React.PropTypes.string
-  ])
+  ]),
+  wordWrap: React.PropTypes.bool
```
mlunoe: I think this should go above `className`, or we should move `className` up

plugins/services/src/js/components/forms/ContainerServiceFormSection.js
jfurrow: As it turns out, when you're setting boolean props to `true` in JSX, you don't need to provide the value. Similar to HTML's `disabled`.

jfurrow:
@mlunoe That's a very valid point, because one of my tasks this week is to move the `Tooltip` from the icon to the `FieldGroup`. I think I'll close this PR, but let's discuss how to best implement the tooltip for the `FieldGroup` to make sure I address the concerns you brought up here!

PR #1869, opened 2017-01-27
DCOS-13451: Simplify the runtime configuration

wavesoft:
This PR simplifies the runtime configuration, by reordering some fields in the form and hiding runtime selection in the advanced section.

image

image

### DIFF
--- a/plugins/services/src/js/components/forms/ContainerServiceFormAdvancedSection.js
+++ b/plugins/services/src/js/components/forms/ContainerServiceFormAdvancedSection.js
@@ -1,0 +1,257 @@
+import React, {Component} from 'react';
+import {Tooltip} from 'reactjs-components';
+
+import {FormReducer as ContainerReducer} from '../../reducers/serviceForm/Container';
+import {FormReducer as ContainersReducer} from '../../reducers/serviceForm/Containers';
+import {findNestedPropertyInObject} from '../../../../../../src/js/utils/Util';
+import ContainerConstants from '../../constants/ContainerConstants';
+import DeleteRowButton from '../../../../../../src/js/components/form/DeleteRowButton';
+import FieldError from '../../../../../../src/js/components/form/FieldError';
+import FieldHelp from '../../../../../../src/js/components/form/FieldHelp';
+import FieldInput from '../../../../../../src/js/components/form/FieldInput';
+import FieldLabel from '../../../../../../src/js/components/form/FieldLabel';
+import FormGroup from '../../../../../../src/js/components/form/FormGroup';
+import FormRow from '../../../../../../src/js/components/form/FormRow';
+import Icon from '../../../../../../src/js/components/Icon';
+
+const {DOCKER, MESOS} = ContainerConstants.type;
+
+const containerSettings = {
+  privileged: {
+    label: ' Grant Runtime Privileges',
+    helpText: 'By default, containers are “unprivileged” and cannot, for example, run a Docker daemon inside a Docker container.',
+    dockerOnly: 'Grant runtime privileges is only supported in Docker Runtime.'
+  },
+  forcePullImage: {
+    label: ' Force Pull Image On Launch',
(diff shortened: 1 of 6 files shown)


nLight:
run integration tests

plugins/services/src/js/components/forms/GeneralServiceFormSection.js
```diff
@@ -52,7 +53,7 @@ const containerRuntimes = {
     )
   },
   [MESOS]: {
-    label: <span>{labelMap[MESOS]} <em>(experimental)</em></span>,
+    label: <span>{labelMap[MESOS]} <span className="badge badge-rounded badge-success">Experimental</span></span>,
```
nLight: ❤️  it

plugins/services/src/js/components/forms/GeneralServiceFormSection.js
```diff
@@ -126,23 +124,43 @@ class GeneralServiceFormSection extends Component {
     }
 
     return (
-      <p>
-        {'Need to run a service with multiple containers? '}
-        <a className="clickable" onClick={this.handleOpenConvertToPodModal}>
-          Add another container
-        </a>.
-      </p>
+      <div className="pod pod-narrow pod-short">
```
nLight: Don't like how `pod pod-narrow pod-short` add padding here. Is it by design?

screen shot 2017-01-27 at 12 24 34

wavesoft: Not really, but it looks better when collapsed
image

It also helps differentiate it from the rest of the form when expanded

image

jfurrow: I think it's fine to add margin to the top of this `Need to run a service...` copy, for now. However I think it should only be on the top, so your classes would look like this:
`pod pod-short flush-horizontal flush-bottom`.

In other parts of the form, we're relying on the top margin of heading tags to give separation to the different sections. In the future, I think we should create a `FormSection` component (or something similar) to provide consistent margins, but definitely not part of this PR!

plugins/services/src/js/components/forms/ContainerServiceFormAdvancedSection.js
```diff
@@ -0,0 +1,257 @@
+import React, {Component} from 'react';
```
jfurrow: Good call on separating this to its own component! 👍

jfurrow:
Nice changes! There are conflicts but otherwise I think it looks great!

plugins/services/src/js/components/forms/GeneralServiceFormSection.js
leemunroe: It seems odd that "Need to run a service..." isn't left aligned.

image

jfurrow: @leemunroe That will be fixed with the classes I provided above (specifically the `flush-horizontal` class).

rcorral:
Adding hold merge as we want to user test this before merging.

wavesoft:
@rcorral let me know if user testing is OK with this.

leemunroe:
@wavesoft @sascala Can we remove the `More information` link from the `Mesos` option? Seems odd to have a link for this when we don't have one beside the other options. Ideally there is something about the Mesos option in the [docs](https://docs.mesosphere.com/1.9/usage/containerizers/) that we link to.

rcorral:
@wavesoft - @leemunroe will be running it by a few people in the company and report back.

sascala:
@rcorral that doc _should_ say more about the Mesos containerizer. Once it does, then that page can be linked to from the `Container Runtime` tooltip. 

I added this to the Jira; will be done by 1.9 GA at the latest: https://mesosphere.atlassian.net/browse/DOCS-1664

wavesoft:
This PR is now closed in favour of #1930

PR #1870, opened 2017-01-27
Use width instead of flex-basis for vertical tab menu

jfurrow:
This PR uses `width` instead of `flex-basis` to set the menu's width.

This is necessary because we don't want the width of the menu to ever change depending on its content or its siblings' content.


jfurrow:
run unit tests

nLight:
run unit tests

nLight:
LGTM! 💯  💖

PR #1853, opened 2017-01-24
[Service Modal Routes 2a of 3] DCOS-12423: Show marathon errors in service config display 

mlunoe:
~~Service Modal Routes 1 of 3: https://github.com/dcos/dcos-ui/pull/1828~~ Merged!
Service Modal Routes 2a of 3: https://github.com/dcos/dcos-ui/pull/1853
Service Modal Routes 2b of 3: https://github.com/dcos/dcos-ui/pull/1871

This PR ensures that we do display marathon errors on the service configuration page, when they emerge.

~~Diff for review can be found here: https://github.com/dcos/dcos-ui/compare/mlunoe/feature/DCOS-12423-add-routes-for-the-new-create-service-modal...mlunoe/feature/DCOS-12423-add-routes-for-the-new-create-service-modal-2?expand=1~~ Merged!

PR #1871, opened 2017-01-27
[Service Modal Routes 2b of 3] DCOS-12423: Add confirm modals when navigating away from route

mlunoe:
~~Service Modal Routes 1 of 3: https://github.com/dcos/dcos-ui/pull/1828~~ Merged!
Service Modal Routes 2a of 3: https://github.com/dcos/dcos-ui/pull/1853
Service Modal Routes 2b of 3: https://github.com/dcos/dcos-ui/pull/1871

This PR adds a confirm modal to route changes from service create/edit modal and when navigating away if there are any changes to the form.

Diff for review can be found here: https://github.com/dcos/dcos-ui/compare/mlunoe/feature/DCOS-12423-add-routes-for-the-new-create-service-modal...mlunoe/feature/DCOS-12423-add-confirm-modals-when-navigating-away-from-route?expand=1

### DIFF
--- a/plugins/services/src/js/components/modals/NewCreateServiceModal.js
+++ b/plugins/services/src/js/components/modals/NewCreateServiceModal.js
@@ -96,15 +96,17 @@ class NewCreateServiceModal extends Component {
   constructor() {
     super(...arguments);
 
-    const {route} = this.props;
-    const {router} = this.context;
+    this.state = this.getResetState(this.props);
 
     METHODS_TO_BIND.forEach((method) => {
       this[method] = this[method].bind(this);
     });
+  }
 
-    this.state = this.getResetState(this.props);
-
+  componentDidMount() {
+    const {location, route} = this.props;
+    const {service} = this.state;
+    const {router} = this.context;
     // Add store change listeners the traditional way as React Router is
     // not able to pass down correct props if we are using StoreMixin
     MarathonStore.addChangeListener(
@@ -128,6 +130,11 @@ class NewCreateServiceModal extends Component {
       route,
       this.handleRouterWillLeave
     );
+
+    // Only add change listener if we didn't receive our service in first try
+    if (!service && this.isLocationEdit(location)) {
+      DCOSStore.addChangeListener(DCOS_CHANGE, this.handleStoreChange);
+    }
   }
 
   componentWillReceiveProps(nextProps) {
@@ -262,7 +269,6 @@ class NewCreateServiceModal extends Component {
 
   handleGoBack(event) {
     const {tabViewID} = event;
-    const {location} = this.props;
     const {
       hasChangesApplied,
       serviceFormActive,
@@ -734,11 +740,6 @@ class NewCreateServiceModal extends Component {
       serviceFormHasErrors: false,
       showAllErrors: false
     };
-
-    // Only add change listener if we didn't receive our service in first try
-    if (!service && isEdit) {
-      DCOSStore.addChangeListener(DCOS_CHANGE, this.handleStoreChange);
-    }
 
     return newState;
   }
--- a/plugins/services/src/js/components/modals/NewCreateServiceModal.js
+++ b/plugins/services/src/js/components/modals/NewCreateServiceModal.js
@@ -1,6 +1,7 @@
 import classNames from 'classnames';
 import deepEqual from 'deep-equal';
 import React, {Component, PropTypes} from 'react';
+import {Confirm} from 'reactjs-components';
 import {Hooks} from 'PluginSDK';
 import {routerShape} from 'react-router';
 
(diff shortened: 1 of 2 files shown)


wavesoft:
run integration tests

mlunoe:
Run integration tests

mlunoe:
Run integration tests

sascala:
For text, I suggest: "Are you sure you want to leave this page? Any data you entered will be lost."

MatApple:
@mlunoe I'm not sure this is the right UX to solve the problem. Did we consider saving form data to local storage and bootstrapping the form with that when you navigate back and invalidating it under some conditions?

mlunoe:
@MatApple Yes. We only show this modal if you have made changes. This solution that protects the user from losing their data/changes. The was the solution proposed by design.
We could discuss this further, but let's not hold this PR back because of it. It is crucial to get a solution in, since we have already had reports from users being frustrated about this.

@leemunroe would you comment?

plugins/services/src/js/components/modals/NewCreateServiceModal.js
```diff
@@ -1,100 +1,237 @@
...
+      this.onMarathonStoreServiceEditError
+    );
+    MarathonStore.addChangeListener(
+      MARATHON_SERVICE_EDIT_SUCCESS,
+      this.onMarathonStoreServiceEditSuccess
+    );
 
-    if (shouldClose) {
-      this.handleClose();
+    router.setRouteLeaveHook(route, this.handleRouterWillLeave);
```
mlunoe: This is how to do it according to react-router:
https://github.com/ReactTraining/react-router/blob/master/docs/guides/ConfirmingNavigation.md

mlunoe: I will move it to didMount

plugins/services/src/js/components/modals/NewCreateServiceModal.js
```diff
@@ -524,42 +699,52 @@ class NewServiceFormModal extends Component {
...
-          header={this.getHeader()}
-          onClose={this.handleClose}
-          useGemini={useGemini}
-          {...Util.omit(props, Object.keys(NewServiceFormModal.propTypes))}>
-          {this.getModalContent()}
-        </FullScreenModal>
-      </div>
+      <FullScreenModal
+        header={this.getHeader()}
+        onClose={this.handleClose.bind(this, hasChangesApplied)}
```
mlunoe: Done

plugins/services/src/js/components/modals/NewCreateServiceModal.js
mlunoe: Actually, going to keep it as this is where we usually bind functions

mlunoe:
Run integration tests

mlunoe:
Run integration tests

leemunroe:
@MatApple @mlunoe I like Mat's suggestion too but this solution is great and 100x better than what currently happens. Happy to relook at other ideas in future like restoring the last used config.

@mlunoe Regarding copy, I like @sascala's suggestion for the text. I think we need better clarity for the buttons. Currently it's not quite clear what "Close" and "Continue" mean e.g. at a glance a user might think "Continue" means to continue editing.

How about adding a ? to the title and removing "current" i.e. `Discard Changes?` and making the primary button `Discard` instead of continue, and secondary `Cancel` instead of close?

plugins/services/src/js/components/modals/NewCreateServiceModal.js
wavesoft: Hey @mlunoe , that code snippet looked a bit weird, since it never de-registers the hook, so I started digging into the react-router internals. And found that this:

* https://github.com/ReactTraining/react-router/blob/6eeb7ad358f987520f5b519e48bdd31f725cbade/modules/RouterUtils.js#L4
* leading to this https://github.com/ReactTraining/react-router/blob/6eeb7ad358f987520f5b519e48bdd31f725cbade/modules/createTransitionManager.js#L188

That looks it returns a function that should be called to unregister the callback. So, to be correct, you should do something like:

```js
componentDidMount() {
  this.unregisterLeaveHook = router.setRouteLeaveHook(route, this.handleRouterWillLeave);
}

componentWillUnmount() {
  if (this.unregisterLeaveHook) {
    this.unregisterLeaveHook();
  }
}
```

mlunoe: Wow! That is some good digging right there! So sorry I missed this, and happy that you didn't! Updated!

mlunoe:
@leemunroe updated to follow your recommendations!

plugins/services/src/js/components/modals/NewCreateServiceModal.js
mlunoe: I didn't wrap it in the `if`-statement as we are adding it in the `constructor`.

wavesoft: Hey @mlunoe , I would also vote against putting it in the constructor for the following reason: I am not sure if constructing an element means also mounting. This means that I am not sure if `constructor` and `componentDidMount` are 100% sure to be always called (at least in our case).

This means that there are two corner cases: 
* Construct the component but never mount it (therefore never de-registering the hook).
* Unmount and re-mount the same instance (therefore calling the unregister function twice, which means this is an invalid operation).

mlunoe: That is a very good point. I updated to register the listeners in the `componentDidMount`. This will also prevent it from trying to setting the state on an unmounted component if the call fires an event before the component has mounted.

jfurrow:
This is awesome!!! :ship:

mlunoe:
Ok, rebased this guy! @wavesoft would ❤️  some 👀 on this :one: so we can prevent users from losing data unnecessarily!

silverem2:
I'd love for us to push this out to other create/edit experiences beyond just services in a later release.

wavesoft:
👷   LET'S MERGE THIS ! 👷 

Merge

PR #1850, opened 2017-01-24
[1 of 3: Mobile Page Header Navigation] DCOS-12558: Create page header navigation dropdown menu

jfurrow:
1 of 3: #1850 
2 of 3: #1872 — [diff between 1 and 2](https://github.com/dcos/dcos-ui/compare/john/DCOS-12558/mobile-page-header-navigation...john/feature/separate-page-header-navigation-styles?expand=1)
3 of 3: #1873 — [diff between 2 and 3](https://github.com/dcos/dcos-ui/commit/83fb381a92b8c2369ac59f3193783ef9c51d63b1)

This PR implements a dropdown menu for the page header navigation at viewports narrower than `768px` (the medium breakpoint).

PR #1872, opened 2017-01-27
[2 of 3: Mobile Page Header Navigation] Separate page header navigation styles

jfurrow:
~~1 of 3: #1850~~
2 of 3: #1872
3 of 3: #1873

This PR simply separates the styles for the page header navigation from the rest of the page header styles, as it was getting a little unruly.

### DIFF
--- a/src/styles/components/page-header/navigation/styles.less
+++ b/src/styles/components/page-header/navigation/styles.less
@@ -1,0 +1,243 @@
+& when (@page-header-enabled) {
+
+  .page-header {
+
+    &-navigation {
+      align-items: flex-start;
+      display: flex;
+      flex: 1 1 auto;
+
+      .menu-tabbed {
+        margin-bottom: 0;
+      }
+
+      & when not (@pod-margin-top = null) {
+        padding-top: @pod-margin-top;
+      }
+
+      & when not (@pod-margin-bottom = null) {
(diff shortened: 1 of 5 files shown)


MatApple:
@jfurrow flicts! :)

jfurrow:
@MatApple I think I'll wait until 1 of 3 merges before fixing this up if you don't mind!

mlunoe:
Still flicts!

jfurrow:
@mlunoe Flicts fixed!

mlunoe:
I don't think #1850 has been rebased off the same point, so the diff includes a lot more than it should, which makes it harder to review.

mlunoe:
But looking at the complete change to master actually makes sense. The base PR makes sense and this PR makes sense

jfurrow:
@mlunoe I'm not sure why the diff links are not working properly, but they are all rebased off one another and contain the same commits. Perhaps something went awry when I rebased them...

jfurrow:
Ping @mlunoe @MatApple Since 1 of 2 merged this should be a lot easier to review :)

mlunoe:
Yup, my label persists!

jfurrow:
Sorry @mlunoe for some reason I didn't think you had added your label... thanks :)

PR #1874, opened 2017-01-28
DCOS-13462: Add link to Universe via MountService

jfurrow:
This PR adds a Universe link to the service create modal via `MountService`.

Without Universe permissions:

With Universe permissions:


src/js/components/CreateServiceModalUniversePanelOption.js
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+
+import packageServiceImage from '../../img/service-picker-options/service-image-package-large@2x.png';
+import CreateServiceModalServicePickerOption from './CreateServiceModalServicePickerOption';
+import CreateServiceModalServicePickerOptionContent from './CreateServiceModalServicePickerOptionContent';
+import CreateServiceModalServicePickerOptionImage from './CreateServiceModalServicePickerOptionImage';
+
+class CreateServicePickerUniverseOption extends React.Component {
+  render() {
+    const {columnClasses, onOptionSelect} = this.props;
```
MatApple: Could we get propTypes for these guys please! :)

src/js/pages/universe/PackagesTab.js
```diff
@@ -30,6 +33,18 @@ const METHODS_TO_BIND = [
   'handleSearchStringChange'
 ];
 
+const shouldRenderUniverseOption = Hooks.applyFilter(
```
MatApple: Let's move this to a lifecycle method. There is no guarantee of execution order here. It's possible the appropriate filter hasn't yet registered.

src/js/pages/universe/PackagesTab.js
```diff
@@ -30,6 +33,18 @@ const METHODS_TO_BIND = [
...
+  'hasCapability', false, 'packageAPI'
+);
+
+if (shouldRenderUniverseOption) {
```
MatApple: I think we should just have a mount in `NewCreateServiceModal` that provides a way for additional options to be injected but the injection (component registration against a mount point) should be handled in a plugin and this open source code shouldn't need to know anything about `hasCapability`.

src/js/components/CreateServiceModalUniversePanelOption.js
jfurrow: Done!

src/js/pages/universe/PackagesTab.js
jfurrow: I don't think a lifecycle method will work here because we can't rely on this component having been mounted before registering the component. I'm not sure what the best way to handle this, what do you think?

src/js/pages/universe/PackagesTab.js
jfurrow: Can you expand on how this could be accomplished? This option should be available to all Open DC/OS users, so it can't be a added via a plugin (I implemented it in a plugin by mistake initially; this PR is fixing that mistake).

@rcorral and I discussed this and decided `PackagesTab` was the best place to handle this.

rcorral: @MatApple keep in mind that we already use the `hasCapability` filter in other areas. We should abstract the call of the filter event because it's odd to have these littered all over the place. This is out of scope though.

MatApple: Now that we have the Mount Service we should in most cases only use that in open DC/OS. Because `hasCapabilities` is an enterprise feature and not available in open DC/OS I think it shouldn't live here.

The list of Service create options should be wrapped in a Mount Component and there should be a plugin that listens to that Mount Component and adds an extra option, in this case - Universe, if the user `hasCapability`. Does that make sense or am I missing something? :)

jfurrow: @MatApple Then the Universe option is only available to enterprise customers; this option should be available to all customers.

mlunoe:
Something was missing and this wasn't working properly with the latest plugins repo, so I rebased it for you and that seemed to fix the issue.

jfurrow:
ping @MatApple

jfurrow:
run integration tests

src/js/pages/universe/PackagesTab.js
philipnrmn: @jfurrow In that case, I think we can remove the ACL check altogether. EE users without the appropriate permissions will be taken to an access denied page though.

philipnrmn:
Apart from the ACL confusion, this LGTM :)

src/js/pages/universe/PackagesTab.js
jfurrow: @philipnrmn That would definitely be easier, but I think hiding the option altogether (if they don't have access to Packages) is a better user experience... @leemunroe what do you think?

leemunroe: Not a fan of teasing the user with an option then telling them they can't use it. We should hide it if they don't have permission to access it.

philipnrmn: Yep, as I thought. In that case we should move the option to within a Mount, and remove it in a plugin in EE.

jfurrow: For those following along, we discussed this offline and will leave as-is for now. :)

I did create a JIRA to track the larger issue of conditional rendering based on ACLs: https://mesosphere.atlassian.net/browse/DCOS-13720

jfurrow:
ping @MatApple again :)

src/js/pages/universe/PackagesTab.js
MatApple: @jfurrow I'm going to approve for now just so we can get this in but I side with @philipnrmn that we should use a Mount and remove the option in a plugin. It's a little bizarre having this conditional in this file, especially in the JS module scope. Let's follow up on that JIRA :)

jfurrow: Sounds good, thanks! Please chat with @rcorral about this as we have discussed this multiple times and keep coming to the conclusion that this is the best solution for now :)

src/js/components/CreateServiceModalServicePickerOptionWrapper.js
```diff
@@ -1,17 +1,14 @@
 import React from 'react';
 
 function OptionsWrapper({children}) {
-  let columnClasses = null;
-  let containerClasses = null;
+  let columnClasses = 'column-12 column-small-6';
+  let containerClasses = 'column-12 column-small-10 column-small-offset-1 column-large-8 column-large-offset-2';
```
nLight: ❤️

src/js/pages/universe/PackagesTab.js
```diff
@@ -30,6 +33,18 @@ const METHODS_TO_BIND = [
...
+  MountService.MountService.registerComponent(
```
nLight: I'm about to merge it, there's one thing I can't agree with though.

I'd rather not put this code into `src/js/pages/universe/PackagesTab.js`. It's just not the right place. 
Let's create a `universe` dir in the `plugins` with `index.js` and `hooks.js` and we put this code into `hooks.js`
Then all the universe related code will eventually be moved there as well.

//cc @MatApple @orlandohohmeier

jfurrow: @nLight I understand your concern and I do agree it would make sense. However, I started implementing your suggestion and then ran into a road block: how do we initialize this new plugin? Right now the only way to initialize plugins is to add them to the `ui-config.json` (or hard-code an `import` somewhere else in the app). I chatted with @rcorral about this and since we are moving away from the "plugins" approach, we didn't think it made sense to create a plugin for this until we can move toward the new "package" approach.

What do you think about this decision?

nLight: You're right, but let's not forget a tech debt ticket or something.

jfurrow: Ahh yeah, meant to share the JIRA ticket: https://jira.mesosphere.com/browse/DCOS-13463

MatApple: Updated the ticket to include links to this discussion.

PR #1875, opened 2017-01-28
Change CPU default from 1 to 0.1 for service create

leemunroe:
This changes the default value for CPU from 1 to 0.1. Users find themselves deploying demo apps with the defaults then quickly run out of resources. This will help with that, as well as educating the user that fractional values are allowed.

image

### DIFF
--- a/plugins/services/src/js/components/forms/ContainerServiceFormSection.js
+++ b/plugins/services/src/js/components/forms/ContainerServiceFormSection.js
@@ -452,7 +452,7 @@ class ContainerServiceFormSection extends Component {
             <FieldLabel className="text-no-transform">CPUs</FieldLabel>
             <FieldInput
               min="0.001"
-              name="cpus"
+              name={cpusPath}
               step="any"
               type="number"
               value={findNestedPropertyInObject(data, cpusPath)} />
--- a/plugins/services/src/js/components/forms/ContainerServiceFormSection.js
+++ b/plugins/services/src/js/components/forms/ContainerServiceFormSection.js
@@ -453,7 +453,7 @@ class ContainerServiceFormSection extends Component {
             <FieldInput
               min="0.001"
               name="cpus"
-              step="0.1"
+              step="any"
               type="number"
               value={findNestedPropertyInObject(data, cpusPath)} />
             <FieldError>{cpusErrors}</FieldError>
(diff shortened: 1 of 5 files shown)


rcorral:
@mlunoe or @jfurrow could one of you fix the integration tests please?

jfurrow:
@rcorral @mlunoe Integration tests fixed!

wavesoft:
Hey @leemunroe that's great! ✨  💯 

I would vote to do the same thing to the multi-container definition, and I also found a tiny #nit : CPU now is 0.1, but the spinner increments are on every full unit. This means clicking "up" in the input goes to 1,1. I would prefer it to go to 0.2 😄

leemunroe:
@wavesoft Good call on Pod default. Updated. I couldn't figure out how to make the increments 0.1 while retaining the minimum 0.001. Can it be done? https://jsfiddle.net/jbsbm2ya/

mlunoe:
Do a search for `step="any"`, that will give you all the resource fields. From here you can change the `step` attribute to your preference. I initially set it to any to not limit the user, but I think it might make sense to have it in increments of the `min` attribute:
> If this attribute is not set to any, the control accepts only values at multiples of the step value greater than the minimum. ref.: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#attr-step

jfurrow:
run unit tests

jfurrow:
@leemunroe That's interesting! I suggest we leave `step="any"`. FWIW @ashenden voted to hide the native controls on `input[type="number"]` elements.

leemunroe:
@jfurrow Yeh I'm not opposed to that. Or having our own custom controls. Don't find the native controls useful. Do you need anything else for this PR?

jfurrow:
@leemunroe Sounds good!

This PR is ready, we just need someone else to give it a review and click the "merge" button. :)

plugins/services/src/js/components/forms/ContainerServiceFormSection.js
```diff
@@ -456,7 +456,7 @@ class ContainerServiceFormSection extends Component {
             <FieldLabel className="text-no-transform">CPUs</FieldLabel>
             <FieldInput
               min="0.001"
-              name={cpusPath}
+              name="cpus"
```
nLight: it should stay `name={cpusPath}` otherwise it doesn't allow any changes for pods

jfurrow: Fixed!

jfurrow:
@nLight This should be good to go!

PR #1876, opened 2017-01-28
DCOS-12565: Add chosen route to expandedItems at mount

jfurrow:
Previously, if you loaded the UI at a route other than `/`, and that route was a nested item (e.g. `/universe/packages`), the sidebar would not show the parent item as expanded.

Now, the Sidebar adds the parent route to `state.expandedItems` in `componentWillMount` so it will be expanded when it renders.

### DIFF
--- a/src/js/components/Sidebar.js
+++ b/src/js/components/Sidebar.js
@@ -53,8 +53,7 @@ var Sidebar = React.createClass({
   },
 
   componentWillMount() {
-    const {pathname} = this.props.location;
-    const pathnameSegments = pathname.split('/');
+    const pathnameSegments = this.props.location.pathname.split('/');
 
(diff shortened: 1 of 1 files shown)


src/js/components/Sidebar.js
```diff
@@ -52,6 +52,17 @@ var Sidebar = React.createClass({
     return {expandedItems: []};
   },
 
+  componentWillMount() {
+    const {pathname} = this.props.location;
+    const pathnameSegments = pathname.split('/');
```
nLight: we could split it right here `this.props.location.pathname.split('/')`

jfurrow: Of course, done!

mlunoe:
Word! ⛵️

PR #1837, opened 2017-01-20
DCOS-13195: Adjust the JSON Editor change handling

orlandohohmeier:
---
~⚠️  _This branch depends on changes that will be introduced with #1831._~

---

Adjust the JSON Editor change handling to properly handle interfering prop updates triggered by the surrounding components.

* Introduces a new method to get the JSON "state" from the JSON text
* Change the update JSON state method to only assign new properties
* Reorder the handle change statements to prevent interfering updates

Closes DCOS-13195

PR #1877, opened 2017-01-30
DCOS-13526: Properly calling updateLocalJsonState on JSONEditor constructor

wavesoft:
This PR fixes a tiny bug that slipped through the #1837 review.

### DIFF
--- a/src/js/components/JSONEditor.js
+++ b/src/js/components/JSONEditor.js
@@ -84,7 +84,7 @@ class JSONEditor extends React.Component {
     this.externalErrors = (this.props.errors || []).slice();
     this.jsonError = null;
     this.jsonMeta = [];
-    this.jsonText = '{}';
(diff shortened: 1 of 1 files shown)


nLight:
Could you give some info on what this tiny bug is about?
I wasn't able to reproduce the initial race condition bug BTW.

wavesoft:
@nLight , the `updateLocalJsonState` function is not used properly. Previously, a single function was de-composing the input string to a set of local properties, but since Orlando's changes, the logic is now separated in two functions that needs to be chained `this.updateLocalJsonState(this.getNewJsonState(initialText));`

Currently, we pass down a string to the `updateLocalJsonState` function, that makes all local state properties undefined, causing errors like this when opening the form:

image

jfurrow:
:ship:

PR #1884, opened 2017-01-31
DCOS-13579: Avoid URI encoding JSX in ServiceBreadcrumbs

jfurrow:
This PR fixes an issue where a variable `id` is reassigned to a JSX object. This is problematic because immediately after this, we URI encode the `id` and then pass it to the `Link` component.


nLight:
LGTM! 💯  💖

wavesoft:
LGTM! ✨ ✨ 
Heavy approves

PR #1886, opened 2017-02-01
[1 of 2: Fix form label wrapping] Introduce FormGroupHeading components

jfurrow:
1 of 2: #1886 
2 of 2: #1887

This PR introduces two components: `FormGroupHeading` and `FormGroupHeadingContent`.

`FormGroupHeading` is now responsible for applying the required indicator (`*`) to the fields. This was previously handled by `FormGroup`, so that logic has been removed.

`FormGroupHeading` can take any number of `FormGroupHeadingContent` elements and will render them on a single line. Setting `primary={true}` on a `FormGroupHeadingContent` element allows that element to shrink and for its content to be truncated.

Example usage:
```jsx
<FormGroupHeading>
  <FormGroupHeadingContent primary={true}>
    I am a piece of content that is reasonably long.
  </FormGroupHeadingContent>
  <FormGroupHeadingContent>
    <Tooltip ... />
  </FormGroupHeadingContent>
</FormGroupHeading>
```

In this example, the tooltip will always be directly adjacent to the string. If the viewport is too narrow, the string will be truncated while the tooltip remains visible.

### DIFF
--- a/src/js/components/form/FormGroup.js
+++ b/src/js/components/form/FormGroup.js
@@ -5,14 +5,14 @@ import FieldError from './FieldError';
 import {omit} from '../../utils/Util';
 
 const FormGroup = (props) => {
-  const {children, className, errorClassName, showError, required} = props;
+  const {children, className, errorClassName, showError} = props;
 
   const clonedChildren = React.Children.map(children, (child) => {
-    if (child == null || (!showError && child.type === FieldError)) {
+    if (child != null && !showError && child.type === FieldError) {
       return null;
     }
 
-    return React.cloneElement(child, {required});
+    return child;
   });
 
   const classes = classNames(
@@ -44,9 +44,6 @@ FormGroup.propTypes = {
   children: React.PropTypes.node,
   // Optional boolean to display error
   showError: React.PropTypes.bool,
-
-  // Optional boolean to make field required
-  required: React.PropTypes.bool,
 
   // Classes
   className: classPropType,
--- a/src/js/components/form/FormGroupHeading.js
+++ b/src/js/components/form/FormGroupHeading.js
@@ -3,7 +3,7 @@ import React from 'react';
 
 import FormGroupHeadingContent from './FormGroupHeadingContent';
 
-function getModifiedChildren(children) {
+function injectAsteriskNode(children) {
   const asteriskNode = (
     <FormGroupHeadingContent className="text-danger" key="asterisk">
       *
@@ -21,7 +21,7 @@ function FormGroupHeading({className, children, required}) {
 
   return (
     <div className={classes}>
-      {required ? getModifiedChildren(children) : children}
+      {required ? injectAsteriskNode(children) : children}
     </div>
   );
 }
--- a/src/js/components/form/FormGroupHeading.js
+++ b/src/js/components/form/FormGroupHeading.js
@@ -1,0 +1,56 @@
+import classNames from 'classnames';
+import React from 'react';
+
+import FormGroupHeadingContent from './FormGroupHeadingContent';
+
+function getChildren(children, required) {
+  if (!required) {
+    return children;
+  }
+
+  // We add the required field indicator after the first child.
+  return React.Children.toArray(children).reduce(
+    function (accumulator, child, index) {
+      accumulator.push(child);
+
+      if (index === 0) {
+        accumulator.push(
+          <FormGroupHeadingContent
+            className="text-danger"
+            key={`${index}-asterisk`}>
+            *
+          </FormGroupHeadingContent>
+        );
+      }
+
+      return accumulator;
+    },
+    []
+  );
+}
+
+function FormGroupHeading({className, children, required}) {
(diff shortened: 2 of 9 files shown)


src/js/components/form/FormGroupHeading.js
```diff
@@ -0,0 +1,56 @@
...
+
+import FormGroupHeadingContent from './FormGroupHeadingContent';
+
+function getChildren(children, required) {
+  if (!required) {
+    return children;
+  }
+
+  // We add the required field indicator after the first child.
+  return React.Children.toArray(children).reduce(
```
nLight: can we do something like 
```
const newChildren = React.Children.toArray(children);
newChildren.splice(1, 0, formGroupHeading);
```
so we avoid empty cycles?

src/js/components/form/FormGroupHeading.js
```diff
@@ -0,0 +1,56 @@
...
+  );
+}
+
+FormGroupHeading.defaultProps = {
+  required: false
+};
+
+FormGroupHeading.propTypes = {
+  children: React.PropTypes.node.isRequired,
+  className: React.PropTypes.oneOfType([
```
nLight: yep, we definitely need a propType for that, let's create a small tech debt jira

src/js/components/form/FormGroup.js
```diff
@@ -5,14 +5,14 @@ import FieldError from './FieldError';
 import {omit} from '../../utils/Util';
 
 const FormGroup = (props) => {
-  const {children, className, errorClassName, showError, required} = props;
+  const {children, className, errorClassName, showError} = props;
 
   const clonedChildren = React.Children.map(children, (child) => {
-    if (child == null || (!showError && child.type === FieldError)) {
```
nLight: so we're changing the logic here with no replacement, meaning that if we merge it we break existing functionality?

src/js/components/form/FormGroupHeading.js
```diff
@@ -0,0 +1,56 @@
+import classNames from 'classnames';
+import React from 'react';
+
+import FormGroupHeadingContent from './FormGroupHeadingContent';
+
+function getChildren(children, required) {
+  if (!required) {
```
nLight: Let's remove inversion of control here, see below

src/js/components/form/FormGroupHeading.js
```diff
@@ -0,0 +1,56 @@
...
+    []
+  );
+}
+
+function FormGroupHeading({className, children, required}) {
+  const classes = classNames('form-group-heading', className);
+
+  return (
+    <div className={classes}>
+      {getChildren(children, required)}
```
nLight: To remove inversion of control from `getChildren`
`{required ? getChildren(children) : children}`

nLight: and then it is maybe not `getChildren` anymore, I mean the name

src/js/components/form/FormGroup.js
jfurrow: Yes, until "2 of 2" merges, the red asterisk will not show up in the form.

src/js/components/form/FormGroupHeading.js
jfurrow: Good call, done!

src/js/components/form/FormGroupHeading.js
jfurrow: I implemented this, but I don't understand the concept of "inversion of control." I researched it briefly, but could you please explain why this is preferable to my original solution?

src/js/components/form/FormGroupHeading.js
jfurrow: JIRA: https://mesosphere.atlassian.net/browse/DCOS-13617

src/js/components/form/FormGroupHeading.js
nLight: I could have mixed up the terms, what I really meant was using boolean arguments to control the internal flow of a function. It's not really a problem but rather a thing that one should avoid. Because it's hard to reason about such functions.

The question is: why don't you extract this argument into an `if` conditional and call that function conditionally rather then running it always and conditionally getting a result :) 

How is `getChildren(children, true)` different from `getChildren(children, false)` or `getChildren(children, isRequired)`? You can't tell without looking into the function.
But something like `injectAsteriskNode(children)` immediately tells you what's going to happen.

There's a good short article: https://silkandspinach.net/2004/07/15/avoid-boolean-parameters/

Please note it is different from a React function that internally checks `this.state.arg` or `this.props` because in this case you're using an internal state of a React component. But even then I'd rather put this control logic on the upper level.

Feel free to bring this topic to one of our sync meetings so we could discuss it more.

nLight:
Thank you for addressing the review comments! 👍

src/js/components/form/FormGroupHeading.js
jfurrow: Ahh, this makes very good sense. Thank you for the thorough explanation!

I really like your function name too, I'll use it :)

PR #1882, opened 2017-01-31
DCOS-13481: Feature (service form): allow GPUs for Mesos runtime

nLight:
This PR unblocks GPUs field for Mesos Runtime as it is the same thing as UCR and confirmed to support GPUs as well.

as GPUs field is still under the Container section we might want to reconsider its location in the form @leemunroe 

screen shot 2017-01-31 at 16 32 27

PR #1890, opened 2017-02-01
DCOS-12589: Apply tooltips to disabled fields

jfurrow:
This PR moves the disabled tooltip notification to the disabled fields, rather than a lock icon near the disabled field.

Before:

After:

_In the above example, a bug in the `Tooltip` prevents us from aligning the tooltip with the left edge of the checkbox. This needs to be fixed in `reactjs-components`._

### DIFF
--- a/plugins/services/src/js/components/forms/ContainerServiceFormSection.js
+++ b/plugins/services/src/js/components/forms/ContainerServiceFormSection.js
@@ -158,30 +158,6 @@ class ContainerServiceFormSection extends Component {
     });
   }
 
-  getGPUSLabel() {
-    if (this.isGpusDisabled()) {
-      return (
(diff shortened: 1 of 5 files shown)


nLight:
Flicts! Please don't forget that we've (you've :trollface: ) merged my PR https://github.com/dcos/dcos-ui/pull/1882 so GPUs should be available for Mesos Runtime as well

and

Warning: Each child in an array or iterator should have a unique "key" prop. Check the render method of `ContainerServiceFormSection`. 

That isn't the case on master

jfurrow:
@nLight Fixed conflict & the missing `key` prop!

nLight:
Awesome stuff! LGTM! 💯  💖

PR #1892, opened 2017-02-01
Bug: Fix page header tab alignment

jfurrow:
This PR fixes a bug I introduced in #1870. These overflow properties should only be applied to vertical tabs.

Before:

After:


nLight:
LGTM! 💯  💖

PR #1893, opened 2017-02-02
Bug: Fix .expanding-table's "row" misalignment

jfurrow:
This PR forces `line-height` inheritance on all children of `.expanding-table td`s. This is necessary to maintain proper alignment among the expanded "rows" in each cell (I was tempted to add an `!important` here but opted not to...).

Before:

After:


philipnrmn:
👌  perfect, very elegant fix as well. Much nicer than my attempt.

rcorral:
Works great!

PR #1907, opened 2017-02-03
Bug fix: Logs unsubscribing on file load

mlunoe:
Before:

After:

### DIFF
--- a/plugins/services/src/js/components/LogView.js
+++ b/plugins/services/src/js/components/LogView.js
@@ -26,18 +26,22 @@ class LogView extends React.Component {
       this[method] = this[method].bind(this);
     });
 
-    this.handleLogContainerScroll = Util.throttleScroll(
-      this.handleLogContainerScroll, 50
+    // Make sure to run this on the leading edge to capture this event as soon
+    // as possible. This has impact on both checkIfCloseToTop and
+    // checkIfAwayFromBottom. Trailing edge will call them too late, i.e. when
+    // the data is out of data and possibly scrolled passed the trigger range.
+    // We need them to be called as soon as we hit the range wherein these will
+    // trigger handlers.
+    this.handleLogContainerScroll = Util.throttle(
+      this.handleLogContainerScroll,
+      50,
+      {leading: true, trailing: false}
     );
 
-    this.handleWindowResize = Util.debounce(
-      this.handleWindowResize.bind(this), 100
-    );
-
-    // Since we can receive a lot of updates (from a stream potentially),
-    // let's debounce this function to not call it too often
-    this.componentDidUpdate = Util.debounce(
-      this.componentDidUpdate.bind(this), 100
+    this.handleWindowResize = Util.throttle(
+      this.handleWindowResize.bind(this),
+      50,
+      {leading: true, trailing: false}
     );
   }
 
@@ -57,11 +61,11 @@ class LogView extends React.Component {
 
     // Prevent updates to fullLog, if it has not changed
     if (this.state.fullLog !== nextProps.fullLog) {
-      this.setState({fullLog: nextProps.fullLog}, function () {
+      this.setState({fullLog: nextProps.fullLog}, () => {
         // This allows the user to stay at the place of the log they were at
         // before the prepend.
         if (nextProps.direction === PREPEND && previousScrollHeight
-          && logContainer) {
+          && logContainer && !this.state.isAtBottom) {
           const currentScrollHeight = logContainer.scrollHeight;
           const heightDifference = currentScrollHeight - previousScrollHeight;
           logContainer.scrollTop = previousScrollTop + heightDifference;
@@ -132,13 +136,8 @@ class LogView extends React.Component {
     }
 
     // Cap animation time between 500 and 3000
-    const animationTime = Math.max(
-      Math.min(
-        logContainer.scrollHeight - DOMUtils.getDistanceFromTop(logContainer),
-        3000
-      ),
-      500
-    );
+    const scrollDistance = logContainer.scrollHeight - logContainer.scrollTop;
+    const animationTime = Math.max(500, Math.min(scrollDistance, 3000));
 
     DOMUtils.scrollTo(
       logContainer,
@@ -153,8 +152,7 @@ class LogView extends React.Component {
   }
 
   checkIfCloseToTop(container) {
-    const distanceFromTop = DOMUtils.getDistanceFromTop(container);
-    if (distanceFromTop < 2000) {
+    if (container.scrollTop < 2000) {
       const {hasLoadedTop, fetchPreviousLogs} = this.props;
       if (!hasLoadedTop) {
         fetchPreviousLogs();
@@ -163,8 +161,7 @@ class LogView extends React.Component {
   }
 
   checkIfAwayFromBottom(container) {
-    const distanceFromTop = DOMUtils.getDistanceFromTop(container);
-    const isAtBottom = container.offsetHeight + distanceFromTop
+    const isAtBottom = container.offsetHeight + container.scrollTop
       >= container.scrollHeight;
 
     if (isAtBottom !== this.state.isAtBottom) {
--- a/plugins/services/src/js/components/LogView.js
+++ b/plugins/services/src/js/components/LogView.js
@@ -43,12 +43,6 @@ class LogView extends React.Component {
 
   componentDidMount() {
     global.addEventListener('resize', this.handleWindowResize);
-
-    const {logContainer} = this;
-    if (logContainer) {
-      // Make sure to check if we are at the top on mount
-      this.checkIfCloseToTop(logContainer);
-    }
   }
 
   componentWillReceiveProps(nextProps) {
--- a/plugins/services/src/js/components/LogView.js
+++ b/plugins/services/src/js/components/LogView.js
@@ -152,6 +152,8 @@ class LogView extends React.Component {
   }
 
   checkIfCloseToTop(container) {
+    // This number has been determined by trail and error to be a good
+    // measurement for close to the top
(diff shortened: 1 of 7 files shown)


plugins/services/src/js/pages/task-details/TaskLogsTab.js
```diff
@@ -21,7 +21,7 @@ const METHODS_TO_BIND = [
   'handleItemSelection'
 ];
 
-const PAGE_ENTRY_COUNT = 200;
+const PAGE_ENTRY_COUNT = 400;
```
jfurrow: What does this value represent? The number of lines in a "page"?

plugins/services/src/js/pages/task-details/TaskLogsTab.js
```diff
@@ -168,13 +178,11 @@ class TaskLogsTab extends mixin(StoreMixin) {
     const {task} = this.props;
     const {subscriptionID} = this.state;
 
-    // Stop tailing before fetching preiovus logs
-    SystemLogStore.stopTailing(this.state.subscriptionID);
-    // Fetch two pages previous log entries to gain more leverage to explore
+    // Fetch a full page previous log entries to gain more leverage to explore
```
jfurrow: I notice the value of the `PAGE_ENTRY_CONT` doubled, but why are we doubling it there and not here anymore?

jfurrow:
Sorry @mlunoe but I'm experiencing couple of bugs.
1. It doesn't seem to "tail" the logs automatically. It seems I have to scroll the mousewheel or click the mouse button to get a few lines of the bottom of the log to show up (and then they show up one-by-one). _This seems to be the case on `master` also._
2. It seems like it's eating a ton of CPU; the browser is struggling to respond to my mouse input (hover effects aren't registering, button clicks take a long time before the UI responds). _This does not seem to be the case on `master` also._
3. On initial load, the loading indicator is visible at the top of the log screen. Is this intentional? It seems this should be hidden until the user scrolls up in the log viewer. _This seems to be the case on `master` also._

plugins/services/src/js/pages/task-details/TaskLogsTab.js
mlunoe: Yes, it probably shouldn't have been committed like that (the doubling), I guess I was testing out the size of the page and landed at 400, so I updated it to that.

mlunoe:
1. and 2. seems related. It should scroll to the bottom, and therefore keep connected to the stream (it disconnects if you scroll too far from the bottom). The fact that it doesn't scroll to the bottom initially is a bug in this PR.
@MatApple maybe you could take a look?

mlunoe:
Talking with @jfurrow offline and decided to make 2. and 3. separate issues.

orlandohohmeier:
@nLight could you please check this one as well?

nLight:
Scroll back throws an error:
`cannot proxy to https://frontend-elasticl-1bg30u9sxj6kn-790264852.eu-central-1.elb.amazonaws.com/ (socket hang up)`

Reproducible on `master` as well.

nLight:
Kinda flaky, sometimes it unsubscribes after a page load, but then resubscribes again and tails logs. And sometimes it just works™.

mlunoe:
Ok, so the first error `cannot proxy to https://frontend-elasticl-1bg30u9sxj6kn-790264852.eu-central-1.elb.amazonaws.com/ (socket hang up)` is a backend issue, which is being fixed here: https://github.com/dcos/adminrouter/pull/52

The unsubscribe was due to first of all an error which I introduced here, where I would delete the EventSource reference, where I should not have. This has now been fixed.

Secondly, there was another issue where the scroll event would fire on the trailing edge, which would cause some issues. This has also been fixed.

Finally, we were throttling the `componentDidUpdate` (which is totally not a good idea), which would make it so that we would miss certain important updates to set the height of the scroll container, which would in turn also make the TaskLogsTab unsubscribe from the stream. This has also been fixed.

plugins/services/src/js/components/LogView.js
```diff
@@ -159,8 +152,7 @@ class LogView extends React.Component {
   }
 
   checkIfCloseToTop(container) {
-    const distanceFromTop = DOMUtils.getDistanceFromTop(container);
```
mlunoe: We need the `scrollTop` here and not ever the `pageYOffset` (difference in scroll not to the top of the page)

plugins/services/src/js/pages/task-details/TaskLogsTab.js
```diff
@@ -21,7 +21,8 @@ const METHODS_TO_BIND = [
   'handleItemSelection'
 ];
 
-const PAGE_ENTRY_COUNT = 200;
+// Number of lines (entries) we asses to be a page
+const PAGE_ENTRY_COUNT = 400;
```
mlunoe: This is an arbitrary number that describes how many lines that goes into one page. I landed on this number by trial and error and doubled it this time because we are no longer requesting "two half" pages initially, just one and 400 lines seemed to be a good number.

mlunoe:
Run unit tests

jfurrow:
@mlunoe After testing again, I'm not sure we should leave the performance problems for a separate PR. Or if we do, we should start working on it immediately because it freezes the UI and makes the logs tab unusable. I really think we should get this in before the 1.9 GA.

mlunoe:
To add to this: I am currently implementing a solution that will default to the Mesos File API to use for logs and you have to configure your cluster to select the new journald API for logs.

Also, yes! The performance issues should be addressed and in my opinion ASAP. I have created a ticket to address this. The first idea is to move the throttling that we removed from `componentDidUpdate` here to be handled in the actions file, where we can accumulate data and then send to the store as we are throttling. This should have the experience that we have today, except it will only call a function in the actions file, and not try to re-render the view all the time.

nLight:
This is my app, I see `blip` if I view stdout file, but logs screen tells me 
`STDOUT Log is Currently Empty. Please try again later.`

```
{
  "id": "/log",
  "cmd": "echo \"blip\" && sleep 10000",
  "instances": 1,
  "cpus": 0.1,
  "mem": 128,
  "disk": 0,
  "gpus": 0,
  "backoffSeconds": 1,
  "backoffFactor": 1.15,
  "container": {
    "type": "DOCKER",
    "docker": {
      "image": "alpine",
      "privileged": false,
      "forcePullImage": false
    }
  },
  "upgradeStrategy": {
    "minimumHealthCapacity": 1,
    "maximumOverCapacity": 1
  },
  "unreachableStrategy": {
    "inactiveAfterSeconds": 300,
    "expungeAfterSeconds": 600
  },
  "portDefinitions": [
    {
      "port": 10000,
      "protocol": "tcp",
      "name": "default"
    }
  ],
  "requirePorts": true
}
```

nLight:
The same app with cmd
`while true; do echo 'blip' && sleep 10; done`
starts showing logs after some time. Not sure it's a UI problem though

mlunoe:
Looking into it. It looks to be a backend issue, so far. Will keep you updated

mlunoe:
Update, this is a backend issue. Will keep posting on progress here.

mlunoe:
Update: Logs issue has been fixed in the current frontend cluster, so you can test it out there and there is a fix up for the issue.

plugins/services/src/js/components/LogView.js
```diff
@@ -159,8 +152,7 @@ class LogView extends React.Component {
...
-    if (distanceFromTop < 2000) {
+    if (container.scrollTop < 2000) {
```
MatApple: Is this `2000` arbitrary? If so, maybe we could get a comment about why 2000

src/js/utils/Util.js
```diff
@@ -92,37 +92,66 @@ const Util = {
...
-
-    const resetCall = function () {
-      if (callWhenReady) {
-        setTimeout(resetCall.bind(this, arguments[0]), wait);
-        callWhenReady = false;
-        func.apply(this, arguments);
-      } else {
-        canCall = true;
+  throttle(func, wait, options) {
+    var context;
```
MatApple: Is there a reason we're using `var`?

src/js/utils/Util.js
```diff
@@ -92,37 +92,66 @@ const Util = {
...
+   * see http://underscorejs.org/docs/underscore.html#section-82
    * @param {Function} func A callback function to be called
    * @param {Number} wait How long to wait
-   * @returns {Function} A function, that, after triggered the first time, will
-   * wait for a period of time before the next trigger. If triggered during the
-   * wait, the function will be invoked immediately after the wait is over.
-   * This function is specifically made for React events, hence the nativeEvent
-   * lookup.
+   * @param {Object} [options]
+   * @param {Object} [options.leading = false]
```
MatApple: Could we add some comments about what `leading` and `trailing` options are? Based on the description below in `@returns` I'm still not sure what they do

src/js/utils/Util.js
```diff
@@ -92,37 +92,66 @@ const Util = {
...
+    }
+
+    var later = function () {
+      previous = Date.now();
+      if (options.leading === false) {
+        previous = 0;
+      }
+      timeout = null;
+      result = func.apply(context, args);
+      if (!timeout) {
```
MatApple: Won't this always be `true` because we set `timeout = null;` just above? I see this is an implementation from `underscore` so I guess we could just leave as is.

mlunoe: I believe this is a timing situation because multiple functions could be called right after each other

src/js/utils/Util.js
mlunoe: Added some more comments!

src/js/utils/Util.js
mlunoe: Updated to use `const` and `let`!

PR #1912, opened 2017-02-06
Add pull request template

rcorral:


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,11 @@
+Description of what this PR fixes, steps to reproduce.
+
+```
+Codeblock to help reproduce issue.
+```
+
```
nLight: Shall we add "Consider adding an image or gif to demonstrate visual change"?

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,11 @@
...
+**Checklist**
+[ ] Added JIRA issue in the message of one of your commits?
```
wavesoft: Did you name your branch in a way to include the JIRA issue?

rcorral: Could we say "Did you add the JIRA issue in a commit message or as part of the branch name"?
I feel like having this strong rule can result in unnecessary extra work. For example, if there's no JIRA and then someone creates a JIRA, but you've already opened up a PR. In this case you'd have to rename, your branch, push and update the PR, and also delete the old branch locally and from the remote.

If you forgot to put the JIRA as part of a commit message you can just amend the commit, and force push which is a lot less effort.

What do you think?

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,11 @@
+Description of what this PR fixes, steps to reproduce.
```
orlandohohmeier: Let's start the doc with a "thank you" note

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,11 @@
...
+
+```
```
orlandohohmeier: Let's wrap hints, comments, and helpers like this code block in comments (`<!--- hint/helper -->`) so that we don't need to delete them every time we're opening a PR.

orlandohohmeier:
Let's not move the `CONTRIBUTING.md` to the _hidden_ github directory, as it should be easy people to find the file even after cloning the repository.

.github/PULL_REQUEST_TEMPLATE.md
rcorral: Won't this mean that if someone wants to use this they would then need to remove the code block comment? Which would require the removal of the beginning/end of the comment tag.

.github/PULL_REQUEST_TEMPLATE.md
rcorral: Good idea. Done!

.github/PULL_REQUEST_TEMPLATE.md
rcorral: I added them in some places, let me know what you think!

rcorral:
@orlandohohmeier I've made some changes, let me know what you think. Feel free to make change of your own and I can review. Thanks!

nLight:
Let's get this in 💪

orlandohohmeier:
@rcorral thanks a bunch for addressing my comments - the templates looks good to me! 💯 ✨  Let's land the template and see how it feels and then adjust if needed.

PR #1913, opened 2017-02-06
Fix sidebar scrollbar bug

jfurrow:
This PR rounds one of the variables in the user account menu because it was resulting the component's height ending in a half pixel. This in turn causes Gemini to be off by one pixel when it calculates its own height.

Before (note the height of the user account dropdown button is `63.5px`, and the 1px sliver of native scrollbar at the bottom of the scroll panel):

After:


jfurrow:
run integration tests

rcorral:
Works grrrreat! Thanks for the fix!

PR #1915, opened 2017-02-06
Update copy in empty dashboard services health panel

philipnrmn:
Before:

screen shot 2017-02-06 at 11 46 45 am

After:
screen shot 2017-02-06 at 11 46 16 am

### DIFF
--- a/plugins/services/src/js/components/ServiceList.js
+++ b/plugins/services/src/js/components/ServiceList.js
@@ -4,7 +4,6 @@ import {List, Tooltip} from 'reactjs-components';
 import React from 'react';
 import {routerShape} from 'react-router';
 
-import Config from '../../../../../src/js/config/Config';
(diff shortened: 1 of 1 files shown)


jfurrow:
:shipit:

plugins/services/src/js/components/ServiceList.js
```diff
@@ -113,7 +112,7 @@ const ServiceList = React.createClass({
     return (
       <div>
         <h3 className="flush-top text-align-center">No Services Running</h3>
-        <p className="flush text-align-center">Use the {Config.productName} command line tools to find and install services.</p>
```
rcorral: I wonder if we should link to the Services page. What do you think?

philipnrmn: Makes sense :)

philipnrmn:
Run integration tests

PR #1917, opened 2017-02-07
[1 of 2] DCOS-12713: Emitting xhr object in action error handlers

MatApple:
This is part 1 of several PRs.

This PR alters action files and their associated tests to ensure the xhr object is emitted in action error handlers.

The goal is to allow `<RequestErrorMsg />` to customize the error message based on xhr properties.

### DIFF
--- a/src/js/events/MetronomeActions.js
+++ b/src/js/events/MetronomeActions.js
@@ -49,15 +49,6 @@ const MetronomeActions = {
     Config.getRefreshRate(),
     function (resolve, reject) {
       return function () {
-
-        const handleError = function (e) {
-          AppDispatcher.handleServerAction({
-            type: REQUEST_METRONOME_JOBS_ERROR,
-            data: e.message
-          });
-          reject();
-        };
-
         RequestUtil.json({
           url: `${Config.metronomeAPI}/v1/jobs`,
           data: [
@@ -74,10 +65,21 @@ const MetronomeActions = {
               });
               resolve();
             } catch (error) {
-              handleError(error);
+              AppDispatcher.handleServerAction({
+                type: REQUEST_METRONOME_JOBS_ERROR,
+                data: error
+              });
+              reject();
             }
           },
-          error: handleError,
+          error(xhr) {
+            AppDispatcher.handleServerAction({
+              type: REQUEST_METRONOME_JOBS_ERROR,
+              data: xhr.message,
+              xhr
+            });
+            reject();
+          },
           hangingRequestCallback() {
             AppDispatcher.handleServerAction({
               type: REQUEST_METRONOME_JOBS_ONGOING
--- a/plugins/services/src/js/events/MarathonActions.js
+++ b/plugins/services/src/js/events/MarathonActions.js
@@ -303,10 +303,11 @@ var MarathonActions = {
               });
               resolve();
             },
-            error(e) {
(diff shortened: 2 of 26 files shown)


jfurrow:
Works as expected! :ship:

wavesoft:
Lengthy, but looks good :+1: 💯  ✨

src/js/events/MetronomeActions.js
```diff
@@ -74,10 +65,21 @@ const MetronomeActions = {
               });
               resolve();
             } catch (error) {
-              handleError(error);
+              AppDispatcher.handleServerAction({
+                type: REQUEST_METRONOME_JOBS_ERROR,
+                data: error
+              });
```
wavesoft: What about the XHR object in this case? Shall we assume that the consumer of this API should always check if `xhr` is undefined? What about providing an empty object, or an XHR-like object?

What about:

```js
{
  status: 600,
  statusText: 'Exception'
}
```

nLight:
It actually looks like a non-intrusive 900 loc PR, I'd merge it

src/js/events/MetronomeActions.js
```diff
@@ -74,10 +65,21 @@ const MetronomeActions = {
...
+              reject();
             }
           },
-          error: handleError,
+          error(xhr) {
+            AppDispatcher.handleServerAction({
+              type: REQUEST_METRONOME_JOBS_ERROR,
+              data: xhr.message,
```
orlandohohmeier: Please adjust the `REQUEST_METRONOME_JOBS_ERROR` event as well as similar cases to always provide the same type of data as varying types result in unnecessary complex handling logic. It might be worth to think about a general error event object (`{code, message}`)  to provide a consistent interface for error handling.

src/js/events/MetronomeActions.js
MatApple: I think the consumer should always check if `xhr == null` first. That would be the safest option.

src/js/events/MetronomeActions.js
MatApple: I understand where you're coming from and agree that would be nice. I do think it's out of scope for this PR though and perhaps not necessary as we migrate to GraphQL. That's tackling technical debt that already existed. I just took the same code and put it in a different place so it was easier to read.

nLight: @orlandohohmeier are you ok going as it is for now?

nLight:
Let's get this in, guys

src/js/events/MetronomeActions.js
orlandohohmeier: @MatApple @nLight Let's invest the time to properly clean things up as we don't have a schedule for the GraphQL refactoring and I don't see an urgent need for merging this PR.

rcorral: @orlandohohmeier it seems like the work you're asking for Mat to complete is not part of the work he set out to do, nor part of this task, but it would be really nice to have this consistency throughout the codebase. Let's merge this in and complete the task you're referring to as a separate effort?
@MatApple please create a JIRA and then feel free to merge this in.

MatApple: 👍 JIRA ticket here - https://jira.mesosphere.com/browse/DCOS-14153

PR #1918, opened 2017-02-07
DCOS-12247: Fix for flipping runtime type Mesos to UCR

nLight:
This PR fixes the situation when a Mesos app with local volumes will be switched to UCR as you edit JSON.

I added additional check into parser so that it looks not only for `container.type === MESOS` but also for `container.docker.image` not to be empty.

In addition to that I added a fix that remove Residency as we remove the last localVolume.

Please refer to the original issue https://mesosphere.atlassian.net/browse/DCOS-12247 to come up with a test plan.

### DIFF
--- a/plugins/services/src/js/reducers/serviceForm/Container.js
+++ b/plugins/services/src/js/reducers/serviceForm/Container.js
@@ -207,7 +207,12 @@ module.exports = {
 
   JSONParser: combineParsers([
     function (state) {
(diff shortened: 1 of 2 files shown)


nLight:
run unit tests

orlandohohmeier:
LGTM! 🚢

jfurrow:
Works perfectly! 🚢

PR #1919, opened 2017-02-07
DCOS-13707: Safer alternative to parseInt for reducers

wavesoft:
In some places in the reducers we are using plain `parseInt` when we are processing a value. This results in some unwanted `NaN` values being collected to the output. 

This PR introduces the parseIntValue function that applies parseInt to the given value and it returns it's numerical representation only if it is indeed a number. Otherwise it passes through the value.

It allso replaces `parseInt(value, 10)` in all other places it was used


nLight:
I have been thinking on implementing that one by myself 👍

nLight:
LGTM! 💯  💖

orlandohohmeier:

jfurrow:
Works great! Nice job! 🏆 🏅

PR #1923, opened 2017-02-08
DCOS-13661: Forbid destructuring in methods

nLight:
As we've agreed that we want to keep our class interface as clear as possible it is beneficial to have a name for a param and you can always destructure it in a method body.

It is still okay to destructure in params in higher order functions and in standalone functions.


jfurrow:
Works as expected!

PR #1924, opened 2017-02-09
DCOS-12608: Collapsing repeating errors in ActionsModal

wavesoft:
This PR enables collapsing of repeating errors in the ActionsModal

image

Related plugins-private PR: https://github.com/mesosphere/dcos-ui-plugins-private/pull/543

### DIFF
--- a/src/js/components/modals/ActionsModal.js
+++ b/src/js/components/modals/ActionsModal.js
@@ -186,16 +186,41 @@ class ActionsModal extends mixin(StoreMixin) {
 
   getRequestErrorMessage(errors) {
     if (errors.length > 0) {
-      // Only show 5 first errors
-      const errorMessages = errors.slice(0, 5).map(function (error, index) {
-        return (
-          <p className="text-error-state" key={index}>{error}</p>
-        );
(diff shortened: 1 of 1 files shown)


nLight:
Do we want to show what exactly failed to be added? There could be a situation when you select all, click add and one half is added successfully and the other half hasn't been and we say just 10x Is already part of the Group, without providing any info.

wavesoft:
Unfortunately marathon does not respond with more details about the error:
```
{
    "title": "Bad Request",
    "description": "User is already part of group",
    "code": "ERR_USER_ALREADY_IN_GROUP"
}
```

If we have the XHR object along with the error (when Matt's PRs gets merged), we could extract the request URL and from there the token. But currently it's not very easy 😑

nLight:
I get both :(

feb-10-2017 15-33-53

wavesoft:
@nLight , you also need the mesosphere/dcos-ui-plugins-private#543

nLight:
LGTM! 💯  💖

jfurrow:
Nice work!

PR #1927, opened 2017-02-09
Add hover and focus states to breadcrumbs in page headers

leemunroe:
Adds hover effect to page header breadcrumbs. This gives the user immediate feedback making it more obvious that these are clickable. It also gives keyboard users feedback with a focus state. The hover style is consistent with that of the links in tables.

image

cc @jfurrow


jfurrow:
Love it! :ship:

PR #1928, opened 2017-02-09
Add resource table to node detail page

philipnrmn:
This PR adds the resource table to the nodes detail page.

screen shot 2017-02-09 at 3 21 57 pm

### DIFF
--- a/plugins/nodes/src/js/pages/nodes/NodeDetailTab.js
+++ b/plugins/nodes/src/js/pages/nodes/NodeDetailTab.js
@@ -2,6 +2,7 @@ import PureRender from 'react-addons-pure-render-mixin';
 import React from 'react';
 
 import ConfigurationMap from '../../../../../../src/js/components/ConfigurationMap';
+import ConfigurationMapHeading from '../../../../../../src/js/components/ConfigurationMapHeading';
 import ConfigurationMapLabel from '../../../../../../src/js/components/ConfigurationMapLabel';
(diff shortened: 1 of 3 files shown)


philipnrmn:
run unit tests

nLight:
⚔️  flicts!

philipnrmn:
:hammer:

wavesoft:
Shall we expand this in the same way @jfurrow has done in his PRs like #1904 ?

philipnrmn:
Can do! What's the intention of this? Just more expressive?

philipnrmn:
Hmm, I'm actually not sure that it's any better, unless we want to deprecate the HashMapDisplay entirely?

jfurrow:
@philipnrmn My take on it is that `HashMapDisplay` should be used when we don't know exactly what keys might exist in an object, but we want to display whatever the API gives us. When we do know what keys we want to display (e.g. the resources in the second `HashMapDisplay` instance you added in this PR), then we should use the `ConfigurationMap` components.

I don't feel super strongly about this and don't want it to hold up this PR, so I'm happy to add my label!

MatApple:
@philipnrmn I do agree with @jfurrow here - would you have time to update? If not, @jfurrow - how would you feel about making the change?

jfurrow:
@MatApple @philipnrmn I'm on it!

jfurrow:
@MatApple @philipnrmn Done!

nLight:
run unit tests

PR #1932, opened 2017-02-14
Updating raml-validator-loader to v0.1.10

wavesoft:
This PR updates `raml-validator-loader` dependency to `0.1.10`

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -6612,41 +6612,43 @@
       "resolved": "https://registry.npmjs.org/querystringify/-/querystringify-0.0.3.tgz"
     },
     "raml-1-parser": {
-      "version": "1.1.13",
+      "version": "1.1.12",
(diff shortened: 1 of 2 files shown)


npm-shrinkwrap.json
```diff
@@ -6612,41 +6612,43 @@
       "resolved": "https://registry.npmjs.org/querystringify/-/querystringify-0.0.3.tgz"
     },
     "raml-1-parser": {
-      "version": "1.1.13",
+      "version": "1.1.12",
```
orlandohohmeier: Why are we downgrading "raml-1-parser"?

wavesoft: Starting from `1.1.13` there are some issues raising. We haven't noticed so far because we already had a previous version of `raml-1-parser` install that satisfied the constraint.

nLight:
I like what you got! 👍

jfurrow:
This works well for me! @orlandohohmeier what do you think?

PR #1933, opened 2017-02-14
DCOS-13910: Re-enable keystroke-updates to the JSON

wavesoft:
This PR re-enables the per-keystroke updates of the JSON editor to the `NewCreateServiceModalForm`


jfurrow:
This does what it says, but are we sure this is what we want? The performance in the form is already noticeably laggy, and I can't imagine this helps?

wavesoft:
@jfurrow keep in mind that this sluggish feeling is due to propType checks and it only happens in development build. In production the feeling should be much smoother. 

And I aggree here with Orlando, where he mentions that this kind of feature adds a value to the overall form, since it gives an overall responsive feeling.

nLight:
funny-two-fat-guys-in-avocado-costumes-14724916

nLight:
Alright, let's get this in 💪

PR #1934, opened 2017-02-14
[1 of 2: DCOS-12561] Page Header Breadcrumbs

jfurrow:
This is a refactor of the `ServicesBreadcrumbs` components. It adds some new features:
* If the user is navigating through groups, only the last two groups will appear in the breadcrumb. All other breadcrumbs get "collapsed" into the ellipsis. Clicking the ellipsis navigates one level up.
* When there is not enough horizontal space to display the service health, it gets removed from the DOM.

I also introduced four new breadcrumb components:
* `Breadcrumb`: this wraps the rest of the Breadcrumb components
* `BreadcrumbCaret`: this will render a caret
* `BreadcrumbSupplementalContent`: this content will not shrink by default, and is meant for icons or other supplemental content
* `BreadcrumbTextContent`: this is the primary content; it will shrink and display an ellipsis when truncated

A follow-up PR will implement these new components in the rest of the app. Until then the breadcrumb styling will incorrect.

This styling is intentional for this PR:

**Checklist**

### DIFF
--- a/src/js/components/PageHeaderBreadcrumbs.js
+++ b/src/js/components/PageHeaderBreadcrumbs.js
@@ -1,30 +1,53 @@
+import {Link} from 'react-router';
 import React from 'react';
+import {Tooltip} from 'reactjs-components';
 
 import Icon from './Icon';
 
 class PageHeaderBreadcrumbs extends React.Component {
-
-  getCaret(index) {
+  getCaret(key) {
     return (
       <li className="page-header-breadcrumb page-header-breadcrumb-caret"
-        key={`caret-${index}`}>
+        key={`caret-${key}`}>
         <Icon color="light-grey" id="caret-right" size="mini" />
       </li>
     );
   }
 
   render() {
-    const {props: {iconID, breadcrumbs}} = this;
-
+    const {props: {breadcrumbs, iconID, iconRoute}} = this;
+    const breadcrumbCount = breadcrumbs.length;
     const containerIcon = (
       <li
         className="page-header-breadcrumb page-header-breadcrumb-icon"
         key="-1">
-        <Icon family="product" id={iconID} size="small" />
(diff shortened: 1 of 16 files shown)


wavesoft:
Hey @jfurrow , is something like this expected with this PR?

image

jfurrow:
Sorry @wavesoft the screenshot in the description showed something quite different... that is correct, and https://github.com/dcos/dcos-ui/pull/1935 will fix it to look pretty again. Perhaps we shouldn't split up the PRs at all?

wavesoft:
I was thinking that every PR that targets master should not break functionality. I am not sure how this can be achieved here, possibly by splitting this into a different place, such that each part is not changing anything on master?

But personally, I am also OK if we merge this and the next one in the same day, but I don't know how we can enforce this. Therefore it gets my label 👍

src/js/components/PageHeaderBreadcrumbs.js
```diff
@@ -1,30 +1,58 @@
...
+    const shouldTruncateBreadcrumbs = breadcrumbCount > 3;
 
     const breadcrumbElements = breadcrumbs.reduce((memo, breadcrumb, index) => {
+      if (index === 0) {
+        memo.push(this.getCaret('first-caret'));
+      }
+
+      if (shouldTruncateBreadcrumbs && index === breadcrumbCount - 3) {
+        memo.push(
+          this.getCaret('first-caret-forreal'),
```
nLight: Let's change the name and not use `forreal` 🙏

jfurrow: Ughhh thank you for catching this, I did not intend to actually commit this, but here we are... 😳

jfurrow: Done!

PR #1937, opened 2017-02-16
DCOS-13386: Add Resume option for suspended services

jfurrow:
This PR adds a "Resume" option for suspended services (meaning their instance count is 0).

When the modal appears, the user can choose how many instances they'd like to resume to. If the app label `MARATHON_SINGLE_INSTANCE_APP` is present, we do not present the instance count textbox, and instead scale the service back to a single instance.

**Checklist**

### DIFF
--- a/plugins/services/src/js/components/modals/ServiceResumeModal.js
+++ b/plugins/services/src/js/components/modals/ServiceResumeModal.js
@@ -72,10 +72,11 @@ class ServiceResumeModal extends React.Component {
   }
 
   handleConfirmation() {
-    this.props.resumeService(
-      this.state.instancesFieldValue || 0,
-      this.shouldForceUpdate()
-    );
+    const instances = this.state.instancesFieldValue == null
+      ? 1
+      : this.state.instancesFieldValue;
+
+    this.props.resumeService(instances, this.shouldForceUpdate());
   }
 
   handleInstancesFieldChange(event) {
--- a/plugins/services/src/js/components/modals/ServiceResumeModal.js
+++ b/plugins/services/src/js/components/modals/ServiceResumeModal.js
@@ -1,0 +1,179 @@
+import {Confirm} from 'reactjs-components';
+import React, {PropTypes} from 'react';
+import PureRender from 'react-addons-pure-render-mixin';
+
+import AppLockedMessage from './AppLockedMessage';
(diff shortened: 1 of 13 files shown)


rcorral:
@jfurrow could you please add an integration test!

plugins/services/src/js/components/modals/ServiceResumeModal.js
```diff
@@ -0,0 +1,179 @@
...
+    if (!errorMsg || !errorMsg.length) {
+      errorMsg = null;
+    }
+
+    this.setState({errorMsg});
+  }
+
+  handleConfirmation() {
+    this.props.resumeService(
+      this.state.instancesFieldValue || 0,
```
wavesoft: Perhaps you mean `|| 1` ? Otherwise when the `MARATHON_SINGLE_INSTANCE_APP` is present it tries to scale it to `0`

wavesoft: Unless the resume request is handled in a different way in Marathon... (for some reason I cannot view the JIRA issue)

jfurrow: You are correct, `|| 1` will be a better user experience (and I'll adjust it to check for `0` accordingly).

jfurrow:
@rcorral Yes you're right, tests incoming!

wavesoft:
retest this please

jfurrow:
ping @wavesoft I believe all of your comments have been addressed, could you please review again? Thank you!

wavesoft:
Approved 😎 

Chunk norris approves

tests/_fixtures/marathon-1-task/groups.json
```diff
@@ -41,7 +41,7 @@
         "maximumOverCapacity": 1
       },
       "labels": {
-        
+        "MARATHON_SINGLE_INSTANCE_APP": true
```
MatApple: Let's make the value here a string `"true"` since that's what we'll get.

jfurrow:
Thanks @MatApple, it should be good to go now!

jfurrow:
Ping @MatApple

MatApple:
Looks great, works great! 💯  For some reason certain frameworks cannot be suspended which requires more investigation (e.g. Cassandra)

PR #1939, opened 2017-02-16
DCOS-13152: Add service validators hooks

nLight:
Two things:
- JSON parser has been adjusted to let all the values in the batch so that we don't delete something we don't support in the form
- Form validators moved from the module level inside `getFormErrors` method so that we could apply hooks and extend validators

**Checklist**

### DIFF
--- a/plugins/services/src/js/components/modals/NewCreateServiceModal.js
+++ b/plugins/services/src/js/components/modals/NewCreateServiceModal.js
@@ -388,20 +388,28 @@ class NewCreateServiceModal extends Component {
     const {serviceFormErrors, serviceSpec} = this.state;
     let validationErrors = [];
 
+    const appValidators = APP_VALIDATORS.concat(
+      Hooks.applyFilter('appValidators', [])
+    );
+
+    const podValidators = POD_VALIDATORS.concat(
+      Hooks.applyFilter('podValidators', [])
+    );
+
     // Validate Application or Pod according to the contents
     // of the serviceSpec property.
 
     if (serviceSpec instanceof ApplicationSpec) {
       validationErrors = DataValidatorUtil.validate(
         getServiceJSON(serviceSpec),
-        APP_VALIDATORS
+        appValidators
       );
     }
 
     if (serviceSpec instanceof PodSpec) {
       validationErrors = DataValidatorUtil.validate(
         getServiceJSON(serviceSpec),
-        POD_VALIDATORS
+        podValidators
       );
     }
 
--- a/plugins/services/src/js/reducers/serviceForm/__tests__/EnvironmentVariables-test.js
+++ b/plugins/services/src/js/reducers/serviceForm/__tests__/EnvironmentVariables-test.js
@@ -15,6 +15,7 @@ describe('Environment Variables', function () {
       expect(batch.reduce(EnvironmentVariables.JSONReducer.bind({}), {}))
         .toEqual({key: 'value'});
     });
+
     it('should keep the last value if they have the same key', function () {
       let batch = new Batch();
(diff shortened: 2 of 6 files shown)


plugins/services/src/js/components/modals/NewCreateServiceModal.js
```diff
@@ -388,6 +375,25 @@ class NewCreateServiceModal extends Component {
     const {serviceFormErrors, serviceSpec} = this.state;
     let validationErrors = [];
 
+    const APP_VALIDATORS = Hooks.applyFilter(
```
wavesoft: Love it ❤️

wavesoft:
Looks good 👍

nLight:
@wavesoft `VALUE_SHAPE` error message added 🏹

wavesoft:
I actually like the idea and the naming 👍

plugins/services/src/js/components/modals/NewCreateServiceModal.js
orlandohohmeier: We we really want to allow the removal of validators? Let's use and _action_ hook instead.

plugins/services/src/js/reducers/serviceForm/__tests__/EnvironmentVariables-test.js
```diff
@@ -92,5 +92,12 @@ describe('Environment Variables', function () {
         {type: SET, value: 'value', path: ['env', 0, 'value']}
       ]);
     });
+    it('should keep complex values', function () {
```
orlandohohmeier: 🎨  Please add an empty line to separate the tests.

nLight: :notsureif: the documentation comment for `applyHook` and `doAction` is the same :trollface:

plugins/services/src/js/components/modals/NewCreateServiceModal.js
nLight: :notsureif: the documentation comment for applyHook and doAction is the same :trollface:

orlandohohmeier: 🙀

nLight: I think here it's better to use a filter. Since `doAction` returns `undefined` and I'd rather not mutating anything. I wouldn't care so much that there will be an evil plugin that will add a filter that removes all the validators.

orlandohohmeier: Please use the filter to get a list of additional filter and append them as discussed offline.

nLight:
@orlandohohmeier that one as well 🏹

nLight:
run unit tests

plugins/services/src/js/components/modals/NewCreateServiceModal.js
```diff
@@ -388,20 +388,28 @@ class NewCreateServiceModal extends Component {
     const {serviceFormErrors, serviceSpec} = this.state;
     let validationErrors = [];
 
+    const appValidators = APP_VALIDATORS.concat(
+      Hooks.applyFilter('serviceValidators', [])
```
orlandohohmeier: Please use `app` and `pod` to distinguish the different service types.

nLight: done!

orlandohohmeier:
🥑 👏

orlandohohmeier:
run integration tests

orlandohohmeier:
@nLight Feel free to merge once the test pass

PR #1941, opened 2017-02-17
DCOS-13770: Pass patchUpdate=false to Marathon on edit action

nLight:
This PR adds `patchUpdate=false` to all service edit requests the UI sends to Marathon. So that Marathon doesn't do `patch` but proper `put` update. That hopefully will fix ~it doesn't~ some related issues

**Checklist**

### DIFF
--- a/plugins/services/src/js/events/MarathonActions.js
+++ b/plugins/services/src/js/events/MarathonActions.js
@@ -211,19 +211,21 @@ var MarathonActions = {
     }
 
     let url = buildURI(`/apps/${service.getId()}`);
+    const params = {
+      force,
+      patchUpdate: false // Switching Marathon edit endpoint into proper PUT
+    };
 
     if (service instanceof Pod) {
       url = buildURI(`/pods/${service.getId()}`);
     }
 
-    if (force === true) {
-      url += '?force=true';
-    }
+    url = url + Util.objectToGetParams(params);
 
     RequestUtil.json({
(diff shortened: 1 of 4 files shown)


wavesoft:
LGTM 💯  ✨

plugins/services/src/js/events/MarathonActions.js
```diff
@@ -211,19 +211,21 @@ var MarathonActions = {
...
+    };
 
     if (service instanceof Pod) {
       url = buildURI(`/pods/${service.getId()}`);
     }
 
-    if (force === true) {
-      url += '?force=true';
-    }
+    url = url + Util.objectToGetParams(params);
```
jfurrow: #nit We could use template literals here :)

src/js/utils/Util.js
```diff
@@ -197,6 +197,18 @@ const Util = {
...
+  objectToGetParams(obj) {
+    const queryString = Object.keys(obj).filter(function (param) {
+      return obj[param] != null;
+    }).map(function (param) {
+      return `${encodeURIComponent(param)}=${encodeURIComponent(obj[param])}`;
+    }).join('&');
+
+    return queryString
+      ? `?${queryString}`
+      : '';
```
jfurrow: Nice!

jfurrow:
Nice work @nLight! Let's 🚢

PR #1942, opened 2017-02-20
DCOS-397: Adjust serivce task handling

orlandohohmeier:
Add a service id getter to the Task struct, which returns the corresponding service id. Refactor `TaskDetail` and `TaskVolumeContainer` to sue the new getter in combination with the
`ServiceTree.findItemById()` method instead of the old `ServiceTree.getServiceFromTaskID()` to retrieve the corresponding service. The change address issues with wrong task health reports and potentially wrong volume mappings which were caused be inexact matches.

**Checklist**

### DIFF
--- a/plugins/services/src/js/pages/task-details/__tests__/TaskDetail-test.js
+++ b/plugins/services/src/js/pages/task-details/__tests__/TaskDetail-test.js
@@ -1,3 +1,5 @@
+jest.dontMock('foundation-ui');
+jest.dontMock('foundation-ui/stores/DCOSStore');
 jest.dontMock('../TaskFilesTab');
 jest.dontMock('../TaskDetail');
 jest.dontMock('../../../stores/MarathonStore');
--- a/plugins/services/src/js/containers/volume-detail/TaskVolumeContainer.js
+++ b/plugins/services/src/js/containers/volume-detail/TaskVolumeContainer.js
@@ -1,6 +1,7 @@
 import {DCOSStore} from 'foundation-ui';
 import React from 'react';
 
+import MesosStateStore from '../../../../../../src/js/stores/MesosStateStore';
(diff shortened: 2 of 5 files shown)


orlandohohmeier:
Forgot to remove `getTaskFromTaskID` ... will update the respective commit.

orlandohohmeier:
run unit tests

wavesoft:
run integration tests

orlandohohmeier:
I dropped 903cfd4dfeefc8463374f7d722616d52fb13077d  as there were more component relying on than I was aware of.

plugins/services/src/js/pages/task-details/__tests__/TaskDetail-test.js
```diff
@@ -19,13 +19,18 @@ const TestUtils = require('react-addons-test-utils');
...
+const DCOSStore = require('foundation-ui').DCOSStore;
 const MesosStateStore = require('../../../../../../../src/js/stores/MesosStateStore');
 const Task = require('../../../structs/Task');
 const TaskDirectory = require('../../../structs/TaskDirectory');
 const TaskDirectoryStore = require('../../../stores/TaskDirectoryStore');
 const TaskDetail = require('../TaskDetail');
 
 describe('TaskDetail', function () {
+  DCOSStore.serviceTree = {
+    findItemById: jest.fn()
```
wavesoft: Since you are mocking this function, would it make sense checking if it's called?

nLight:
run integration tests

plugins/services/src/js/pages/task-details/__tests__/TaskDetail-test.js
orlandohohmeier: Good catch! I intended to write such test but decided not to... going to drop the respective lines.

PR #1943, opened 2017-02-21
DCOS-13072: Proper access denied when user has no permissions in services

wavesoft:
This commit ensures that the user has `mesosAPI` and `marathonAPI` permissions in order to grant access to the `/services` page. If not, it redirects to the `/access-denied` page. 

**Checklist**


wavesoft:
run integration tests

MatApple:
@wavesoft this should not be handled here. Please move to plugins.

MatApple:
@orlandohohmeier - how do you feel about this?

weblancaster:
@orlandohohmeier is this something we want still?

nLight:
Should be reimplemented without exposing the hook in the open version.

PR #1944, opened 2017-02-22
DCOS-13474: Fix invisible service form in Safari

jfurrow:
This PR fixes the missing service form in Safari. 

To reproduce this, you'll need to make sure that the native macOS scrollbars are visible. You can force this by setting your "Show scroll bar" settings to "Always":

Before:

After:

**Checklist**


nLight:
Awesome! LGTM! 💯  💖

PR #1945, opened 2017-02-22
DCOS-14088: Fix sidebar styling in IE11

jfurrow:
This PR changes the sidebar's `max-width` property to `width` to avoid any inconsistent behavior with the its width. Fixes a bug in IE11 where the sidebar container element wasn't as wide as the actual sidebar.

This also fixes a bug where the user account dropdown menu was overflowing its container.

Before:

After:

**Checklist**

### DIFF
--- a/src/styles/components/user-account-dropdown/styles.less
+++ b/src/styles/components/user-account-dropdown/styles.less
@@ -37,6 +37,7 @@
       flex: 1 1 auto;
       flex-direction: column;
       min-width: 0;
+      overflow: hidden;
       text-align: left;
 
       &-primary {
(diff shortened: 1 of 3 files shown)


rcorral:
@jfurrow could you please write some integration tests!

wavesoft:
Looks good, but I aggree with @rcorral . It would be good to have some integration tests in order to ensure the text is wrapped correctly.

(Though I don't know how we can use cypress to test on IE)

jfurrow:
@wavesoft Cypress doesn't work with other browsers yet, but it will in the future! I will add some integration tests anyway, good call 👍

jfurrow:
@wavesoft @rcorral Added integration tests to catch this in the future

jfurrow:
run integration tests

PR #1946, opened 2017-02-23
DCOS-13422: Fix TaskTable's status sorting method

jfurrow:
Before:

After: 

**Checklist**


jfurrow:
@leemunroe Can you please verify that the ordering of these states is correct? I chose them based on what I think is the most important... Here are the values (lower values mean they will be higher in the list when in ascending order): https://github.com/dcos/dcos-ui/blob/ae468987c3fbaeef9c259e94abadeaeb739870d9/plugins/services/src/js/constants/TaskStatusSorting.js

foundation-ui/utils/ResourceTableUtil.js
```diff
@@ -45,6 +46,8 @@ var ResourceTableUtil = {
 
   getSortFunction(tieBreakerProp) {
```
nLight: How about a small refactoring? The name of the file is `ResourceTableUtil` but `ResourceTable` does't exist. That's ok, the `getSortFunction` is being used only by the `TaskTable` if I'm not mistaken. Let's extract this function into `TaskTableUtil`. Then I won't be doing a hardcore soul searching justifying presence of a specific constant `TaskStatusSorting` inside a generic function `getSortFunction`. Does it make sense? 

Happy to discuss this today.

jfurrow: Good idea! Will work on this now :)

plugins/services/src/js/constants/TaskStatusSorting.js
```diff
@@ -0,0 +1,20 @@
+const TaskStatusSorting = {
```
jfurrow: Yes, that's a much better name. Done!

foundation-ui/utils/ResourceTableUtil.js
jfurrow: Done!

nLight:
run unit tests

nLight:
John you are awesome!

giphy-2

jfurrow:
No *you're* awesome!

rcorral:
Works great, thanks @jfurrow!!

PR #1947, opened 2017-02-23
DCOS-13960: Remove modal body content's bottom margins

jfurrow:
This PR removes the `margin-bottom` from the last paragraph tags inside `modal-body`. I think it should be safe as it only targets `.modal-body > p:last-child` and `.modal-body > div:last-child > p:last-child`.

Before:

After:

**Checklist**


nLight:
LGTM! 💯  💖  🙌🏻

PR #1654, opened 2016-12-20
Fixing integration tests for JobDetails

MatApple:
⚠️  ~~Depends on a fix for https://mesosphere.atlassian.net/browse/DCOS-11920~~

PR #1948, opened 2017-02-24
DCOS-14157: Properly pass the JobRunHistoryTable sort properties

jfurrow:
This PR fixes the way we were passing the `sortBy` props to the `CheckboxTable`.

I can add integration tests when #1654 is fixed up :)

**Checklist**


nLight:
LGTM! 💯  💖

PR #1949, opened 2017-02-24
DCOS-10426: Disable the Form for custom VIPs

nLight:
🚨  **1.8 fix** 🚨 

This PR fixes the way ServiceFormModal detects that VIPs are invalid, so that we disable the form and allow JSON only, os that the Form won't rewrite VIPs.

**Checklist**


jfurrow:
Works perfectly! :ship:

PR #1951, opened 2017-02-24
DCOS-14022: Adding validate-build script to ensure minification

wavesoft:
This PR introduces the `scripts/validate-build.js` script and the `validate-build` command in `package.json` that checks if the files in `dist/*.js` are minified.

In order to efficiently and correctly process the javscript AST the script uses the `esprima` ecmascript parser and ensures that every comment found in the source is a legit one. Such comments are:

* Source maps
* Comments kept by default by the [UglifyJSPlugin](https://webpack.github.io/docs/list-of-plugins.html#uglifyjsplugin)

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,11 +78,13 @@
     "eslint": "3.14.1",
     "eslint-config-airbnb": "10.0.1",
     "eslint-plugin-compat": "1.0.1",
+    "eslint-plugin-destructuring": "2.1.0",
     "eslint-plugin-import": "1.13.0",
     "eslint-plugin-jsx-a11y": "2.1.0",
     "eslint-plugin-jsx-max-len": "1.0.0",
     "eslint-plugin-react": "6.9.0",
-    "eslint-plugin-destructuring": "2.1.0",
+    "esprima": "3.1.3",
+    "esprima-walk": "0.1.0",
     "extract-text-webpack-plugin": "1.0.1",
     "file-loader": "0.8.5",
     "glob": "7.0.5",
@@ -128,7 +130,7 @@
     "prebuild-assets": "./node_modules/.bin/gulp checkDependencies && npm install compression-webpack-plugin@latest",
     "build-assets": "npm run clean && NODE_ENV=production webpack --config ./webpack/webpack.production.babel.js",
     "postbuild-assets": "rm -rf node_modules/compression-webpack-plugin",
-    "build": "npm run lint && npm test && npm run build-assets",
+    "build": "npm run lint && npm test && npm run build-assets && npm run validate-build",
     "build-with-external-plugins": "npm run lint && npm run lint-plugins && npm test && npm run build-assets",
     "lint": "npm run stylelint && npm run eslint",
     "eslint": "bash -c 'opts=$@ ; set -f; if [[ $# -lt 1 ]]; then opts=\"{src/js,plugins,foundation-ui}/**/*.js\"; fi ; eslint --quiet ${opts}' 0",
@@ -140,7 +142,8 @@
     "scaffold": "./node_modules/.bin/gulp ensureConfig && ./node_modules/.bin/gulp ensureDevProxy",
     "start": "npm run check-env && NODE_ENV=development webpack-dev-server --config ./webpack/webpack.dev.babel.js --hot --progress --colors --content-base ./build --host 0.0.0.0 --port $npm_package_config_port",
     "test": "node jest/gen-config.js && NODE_PATH=node_modules jest --config=jest/config.json --no-cache",
-    "testing": "NODE_ENV=testing webpack-dev-server --config ./webpack/webpack.dev.babel.js --progress --colors --content-base ./build --port $npm_package_config_port"
+    "testing": "NODE_ENV=testing webpack-dev-server --config ./webpack/webpack.dev.babel.js --progress --colors --content-base ./build --port $npm_package_config_port",
+    "validate-build": "node scripts/validate-build.js dist/*.js"
   },
   "jest-webpack-alias": {
     "configFile": "./webpack/webpack.test.js"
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -140,7 +142,8 @@
     "scaffold": "./node_modules/.bin/gulp ensureConfig && ./node_modules/.bin/gulp ensureDevProxy",
     "start": "npm run check-env && NODE_ENV=development webpack-dev-server --config ./webpack/webpack.dev.babel.js --hot --progress --colors --content-base ./build --host 0.0.0.0 --port $npm_package_config_port",
     "test": "node jest/gen-config.js && NODE_PATH=node_modules jest --config=jest/config.json --no-cache",
-    "testing": "NODE_ENV=testing webpack-dev-server --config ./webpack/webpack.dev.babel.js --progress --colors --content-base ./build --port $npm_package_config_port"
+    "testing": "NODE_ENV=testing webpack-dev-server --config ./webpack/webpack.dev.babel.js --progress --colors --content-base ./build --port $npm_package_config_port",
+    "validate-build": "node scripts/validate-build.js dist/*.js"
```
rcorral: If you added the executable bit to `scripts/validate-build.js`, then you could just remove the `.js` extension and then change this line to just this: `scripts/validate-build dist/*.js` that way it looks and acts more as a binary as you already have `#!/usr/bin/env node` at the top of the file

scripts/validate-build.js
```diff
@@ -0,0 +1,113 @@
+#!/usr/bin/env node
+// eslint-disable-next-line strict
+'use strict';
+
+const fs = require('fs');
+const esprima = require('esprima');
+const walk = require('esprima-walk');
+const colors = require('colors');
```
rcorral: 🔤  is always nice :)

MatApple: Let's make sure these packages are in our dev dependencies. Looks like `colors` is a dependency of a dependency.

scripts/validate-build.js
```diff
@@ -0,0 +1,113 @@
...
+
+/**
+ * Iterate over the command-line arguments (them being file names) and check
+ * if the files given are minified.
+ */
+process.exit(
+  process.argv.slice(2).reduce(function (code, file) {
+    const isOk = isMinified(file);
+    if (isOk) {
+      console.error(
```
MatApple: `console.log` here?

scripts/validate-build.js
```diff
@@ -0,0 +1,113 @@
...
+/**
+ * Collect all comments from the AST
+ *
+ * @param {ASTTree} ast - The AST tree
+ * @returns {Array<ASTNode>} Returns the comment AST nodes
+ */
+function loadComments(ast) {
+  let comments = [];
+
+  walk(ast, function (node) {
```
ariya: You do not need to traverse the AST to collect the comments. Every single comment is available from the top-level object already.
For more details, read http://esprima.readthedocs.io/en/3.1/syntactic-analysis.html#comment-collection.

package.json
wavesoft: Good point 👍

scripts/validate-build.js
wavesoft: 😊  👍

wavesoft: Damn, I thought `colors` was a direct dependency. Thanks for noticing this!

scripts/validate-build.js
wavesoft: Yup, fixed 👍

scripts/validate-build.js
wavesoft: Hey @ariya , I can't confirm this. For me the comments go right above of the AST node they target.

Do you also need to specify any additional flag? Or are you using a different `esprima` version?

wavesoft:
Addressed all comments 👍

scripts/validate-build.js
ariya: @wavesoft The behavior has been like this across all versions of Esprima.

To illustrate _comment collection_, run this with Node.js REPL (assuming `esprima` module is installed):

```js
> var esprima = require('esprima')
> esprima.parse('x + y // hello', { comment: true })
Program {
  type: 'Program',
  body: [ ExpressionStatement { type: 'ExpressionStatement', expression: [Object] } ],
  sourceType: 'script',
  comments: [ { type: 'Line', value: ' hello' } ] }
```
Only the top-level node _Program_ will ever have the `comments` array. You can prove this by traversing the node top-to-bottom, no other node will have such an array.

This is not to be confused with _comment attachment_, as demonstrated below:

```js
> var esprima = require('esprima')
> esprima.parse('x + y // hello', { attachComment: true })
Program {
  type: 'Program',
  body: [ ExpressionStatement { type: 'ExpressionStatement', expression: [Object] } ],
  sourceType: 'script' }
> esprima.parse('x + y // hello', { attachComment: true }).body[0].expression
BinaryExpression {
  type: 'BinaryExpression',
  operator: '+',
  left: Identifier { type: 'Identifier', name: 'x' },
  right: Identifier { type: 'Identifier', name: 'y' },
  trailingComments: [ { type: 'Line', value: ' hello', range: [Object] } ] }
```
Look at how `trailingComments` array exists only for the _BinaryExpression_ node (because that particular comment is attached to the node).

I believe that in your particular use case, what you want is to _collect_ every single comment, not to have it attached to various node. Therefore, it is sufficient to inspect the top-level Program node and grab the `comments` array and look at its element, **without** the need to traverse the whole AST.

Does this make sense?

wavesoft:
retest this please

scripts/validate-build.js
wavesoft: You are right... for some reason the `comment` property was not appearing on the AST on my tests, but now that I re-tested it on a fresh installation it seems to work... I really don't know why 😄 

Thanks for this 👍

wavesoft:
Rebased and using @ariya 's suggestion

jfurrow:
Awesome work @wavesoft!

scripts/validate-build.js
ariya: @wavesoft Glad to know it's now working correctly! 👍

PR #1956, opened 2017-02-28
DCOS-14064: Remove "Variables" from volume section headers in service form

mlunoe:
This PR removes "Variables" from volume form section headers.
Before:
Local Volumes Variables
External Volumes Variables

After:
Local Volumes
External Volumes

**Checklist**


jfurrow:
:ship:

PR #1960, opened 2017-03-01
DCOS-14090: Allow menu tabbed content to shrink in IE11

jfurrow:
This PR fixes a bug in IE11 where flex children don't shrink as they should. It's present anywhere we use `.menu-tabbed`, but most notably in the service create form:

Before:

After:

**Checklist**

### DIFF
--- a/src/styles/components/menu-tabbed/styles.less
+++ b/src/styles/components/menu-tabbed/styles.less
@@ -6,7 +6,7 @@
 
     .menu-tabbed-view-container {
       flex: 1 1 auto;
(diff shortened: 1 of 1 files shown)


nLight:
Those selects are amazing! :D

nLight:
unrelated, but unsure if it's Virtualbox acting up or IE

mar-03-2017 12-23-52

MatApple:
@nLight - could you please investigate the issue you observed above to see if it's a valid bug? 💯

PR #1962, opened 2017-03-01
Refactoring CosmosErrorMessages to be more flexible

mlunoe:
… and not use `CollapsibleErrorMessage` component.

This PR updates the cosmos error message handling, so all details are displayed without having to click "view details". It uses `Alert` component rather than the `CollapsibleErrorMessage` component.

PackageTab page:

Install:

Uninstall:

**Checklist**

### DIFF
--- a/src/js/components/Alert.js
+++ b/src/js/components/Alert.js
@@ -3,9 +3,10 @@ import React from 'react';
 
 import Icon from './Icon';
 
-const Alert = ({children, showIcon, type}) => {
+const Alert = ({children, flushBottom, showIcon, type}) => {
   const classes = classNames('alert', {
-    [`alert-${type}`]: type != null
+    [`alert-${type}`]: type != null,
+    'flush-bottom': flushBottom === true
   });
   let icon = null;
 
@@ -31,12 +32,14 @@ const Alert = ({children, showIcon, type}) => {
 };
 
 Alert.defaultProps = {
+  flushBottom: false,
   showIcon: true,
   type: 'danger'
 };
 
 Alert.propTypes = {
   children: React.PropTypes.node.isRequired,
+  flushBottom: React.PropTypes.bool,
   showIcon: React.PropTypes.bool,
   type: React.PropTypes.oneOf(['danger', 'success'])
 };
--- a/src/js/components/modals/UninstallPackageModal.js
+++ b/src/js/components/modals/UninstallPackageModal.js
@@ -97,7 +97,7 @@ class UninstallPackageModal extends mixin(StoreMixin) {
     return (
       <CosmosErrorMessage
         error={packageUninstallError}
-        flushBottom={false} />
+        flushBottom={true} />
     );
   }
 
--- a/src/js/components/modals/UninstallPackageModal.js
+++ b/src/js/components/modals/UninstallPackageModal.js
@@ -96,9 +96,8 @@ class UninstallPackageModal extends mixin(StoreMixin) {
 
     return (
       <CosmosErrorMessage
-        className="text-error-state text-overflow-break-word"
         error={packageUninstallError}
-        header="" />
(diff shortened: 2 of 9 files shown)


mlunoe:
Run unit tests

mlunoe:
Run integration tests

mlunoe:
Run integration tests

mlunoe:
Run integration tests

src/js/pages/universe/PackagesTab.js
```diff
@@ -109,9 +110,9 @@ class PackagesTab extends mixin(StoreMixin) {
     const {errorMessage} = this.state;
 
     return (
-      <CosmosErrorMessage
-        error={errorMessage}
-        headerClass="h3 text-align-center flush-top" />
+      <AlertPanel title="An Error Occurred">
+        <CosmosErrorMessage error={errorMessage} />
```
jfurrow: I think we should add `flushBottom={true}` here, what do you think?

Without `flushBottom`:


With `flushBottom`:

src/js/components/modals/UninstallPackageModal.js
```diff
@@ -96,9 +96,8 @@ class UninstallPackageModal extends mixin(StoreMixin) {
 
     return (
       <CosmosErrorMessage
-        className="text-error-state text-overflow-break-word"
         error={packageUninstallError}
-        header="" />
+        flushBottom={false} />
```
jfurrow: `false` is the default value for this prop so I suggest we remove this.

jfurrow:
Looks great overall, just some minor comments. Nice work!

src/js/components/modals/UninstallPackageModal.js
mlunoe: Changed it to true as it should have been!

src/js/pages/universe/PackagesTab.js
mlunoe: Yes! Added!

mlunoe:
@jfurrow comments addressed!

nLight:
flicts!

mlunoe:
Rebased!

nLight:
Let's handle overflows 💪 

screen shot 2017-03-06 at 09 51 10

src/js/components/Alert.js
```diff
@@ -3,9 +3,10 @@ import React from 'react';
 
 import Icon from './Icon';
 
-const Alert = ({children, showIcon, type}) => {
+const Alert = ({children, flushBottom, showIcon, type}) => {
   const classes = classNames('alert', {
-    [`alert-${type}`]: type != null
+    [`alert-${type}`]: type != null,
+    'flush-bottom': flushBottom === true
```
nLight: Both checks look redundant. 
Not empty string in `type` will render `alert-<type>`
Truthy value in `fulshBottom` will render `flush-bottom`

nLight:
Apart from these two small comments it looks good!

src/js/components/Alert.js
mlunoe: 1. Yes, that is the point. a `type` of 'foo' will add the class 'alert-foo'. Not sure I understand your comment? The PropTypes will make sure it is either type: 'danger' or 'success'. Also, I did not change this part.
2. Truthy value for `flushBottom` will not render `flush-bottom`, only `true` will apply the class: `1 === true` will equal `false`

mlunoe:
I will handle the overflow!

jfurrow:
Good catch on the overflow. Once we handle that I'll add my label!

jfurrow:
I added the overflow behavior, here it is in action:

mlunoe:
Thanks @jfurrow!

rcorral:
Code looks good, and it works as expected!

PR #1964, opened 2017-03-01
DCOS-14091: Apply flex workaround for service create modal in IE11

jfurrow:
This PR implements a `flex` workaround to force the service form to fill its parent in IE11.

The workaround is described here: https://github.com/philipwalton/flexbugs#3-min-height-on-a-flex-container-wont-apply-to-its-flex-items

Before:

After:

**Checklist**


nLight:
LGTM! 💯  💖

PR #1965, opened 2017-03-01
DCOS-14089: Use height instead of min-height for .modal-body

jfurrow:
This PR changes the `min-height` property on the full screen's `.modal-body` to `height` to make the flexbox children vertically centered in IE11.

Before:

After:

**Checklist**


jfurrow:
run integration tests

jfurrow:
run integration tests

jfurrow:
This actually causes a bug with the config review screen, so I'm going to close this until I can find a better solution.

PR #1969, opened 2017-03-03
DCOS-14281: Handle `null` patch values in applyPatch

wavesoft:
This PR ensures that `null` patch value is properly handled in `applyPatch`.

**Checklist**


nLight:
Awesome! 👍 💯

PR #1970, opened 2017-03-03
DCOS_OSS-702: Remove API errors on CreateServiceJsonOnly when changed

wavesoft:
This PR ensures that the API errors are removed when the user uses the `CreateServiceJsonOnly` form and there are marathon errors after review.

Without this PR the API errors would always persist, giving the feeling that the JSON is never correct.

**Checklist**


nLight:
LGTM! 💯  💖

jfurrow:
Works perfectly! :ship:

PR #1973, opened 2017-03-07
DCOS-14307: Fix Discard dialogue behavior

nLight:
This PR addresses the issue when discarding changes would have just closed the form which cause a glitch with Universe install button.

The issue is addressed by resetting the flag `hasChangesApplied` every time we discard changes.

**Checklist**


jfurrow:
Nice work! Works great. :ship:

rcorral:
Thanks Dima!

PR #1978, opened 2017-03-08
DCOS-14386: Remove styleguide

jfurrow:
I checked with @ashenden and nobody is using this, so let's remove it for now. In the future we'll replace this with the design system docs/resources that will live alongside, but not inside of, DC/OS UI.

**Checklist**


rcorral:
Works great. Tested this out with `npm start` and `npm run build`. 
Thanks John!

PR #1979, opened 2017-03-09
DCOS-14277: Fix pod container status string

wavesoft:
This PR fixes a typo in the `PodContainerStatus.js` and introduces the new `PodInstanceStatus.js` state `FINISHED` that will be used when the pod is in `terminal` state, yet all of it's containers are in `FINISHED` state.

Before:

image

After:

image

**Checklist**


nLight:
LGTM! 💯  💖

PR #1982, opened 2017-03-09
Change LogView container offset

rcorral:
When I was testing out the log viewer I noticed that if I scrolled up a  little and a new log line would come through, the autoscroll would kick in, thinking that I was at the bottom, but I scrolled up just a bit. This confused me and frustrated me, so I've lowered this buffer that we have to something really small.

**Checklist**

### DIFF
--- a/plugins/services/src/js/components/LogView.js
+++ b/plugins/services/src/js/components/LogView.js
@@ -21,6 +21,9 @@ class LogView extends React.Component {
   constructor() {
     super(...arguments);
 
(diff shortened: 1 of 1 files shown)


mlunoe:
Run integration tests

rcorral:
run unit tests

mlunoe:
Looks good mah dudes :sunglasses:

jfurrow:
Works well for me.

PR #1986, opened 2017-03-10
DCOS_OSS-733: Corrupted log files

mlunoe:
This PR fixes issues that will corrupt the logs.

To test this out create a task that echos something every 1s, page upwards through the file. If you pause paging upwards and then continue you will see one issue that will corrupt the log files, since the offset would not be updated in the mesos logs.

The second issue emerges at the top. Whenever you have paged all the way to the top you will see corrupt log files since we would not properly update the requested page length at the top.

**Checklist**


jfurrow:
Nice work! :ship:

rcorral:
run integration tests

PR #1987, opened 2017-03-10
Adding missing .bind in LogView constructor

wavesoft:
This PR adds a missing `.bind(this)` argument to the `handleLogContainerScroll` function. This should fix the `setState` issues.

### DIFF
--- a/plugins/services/src/js/components/LogView.js
+++ b/plugins/services/src/js/components/LogView.js
@@ -40,7 +40,7 @@ class LogView extends React.Component {
     // We need them to be called as soon as we hit the range wherein these will
     // trigger handlers.
     this.handleLogContainerScroll = Util.throttle(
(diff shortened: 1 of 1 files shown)


nLight:
retest this please

nLight:
The issue comes from `reactjs-components` VirtualList. 
You can reproduce it on http://localhost:4200/#/services/overview
by making your window small enough so you have to scroll the table

nLight:
So apparently the problem is that VirtualList doesn't properly remove it's onscroll subscription and usually your task tables not that long so there's nothing to scroll. But since listener is still there and you go from the task detail to log view and start scrolling event fires. Once you reload page so that you on the log view right away there will be no error

nLight:
I'm rejecting this one 😢

PR #1988, opened 2017-03-10
DCOS-14431: Fix memory leak where events would not be closed properly

orlandohohmeier:
Refactor the `SystemLogActions` to improve the log view performance. 

* Introduce a general (un)subscribe functions
* Change the default subscription ID not to change if the URL is the same
* Unify the (un)subscribe handling to make sure event listeners are properly removed

Closes DCOS-14431

**Checklist**

### DIFF
--- a/src/js/events/SystemLogActions.js
+++ b/src/js/events/SystemLogActions.js
@@ -19,8 +19,8 @@ import SystemLogUtil from '../utils/SystemLogUtil';
  */
 
 // Store of current open connections
-const sources = {};
-const tails = {};
+const URLToEventSourceMap = {};
+const subscriptionIDtoURLMap = {};
 
 function subscribe(url, onMessage, onError) {
   // Unsubscribe if any open connection exists with the same ID
@@ -34,7 +34,7 @@ function subscribe(url, onMessage, onError) {
   source.addEventListener('error', onError, false);
 
   // Store listeners along with EventSource reference, so we can clean up
-  sources[url] = {
+  URLToEventSourceMap[url] = {
     errorListener: onError,
     messageListener: onMessage,
     source
@@ -44,18 +44,18 @@ function subscribe(url, onMessage, onError) {
 }
 
 function unsubscribe(url) {
-  if (!sources[url]) {
+  if (!URLToEventSourceMap[url]) {
     return;
   }
 
-  const { errorListener, messageListener, source } = sources[url];
+  const { errorListener, messageListener, source } = URLToEventSourceMap[url];
 
   source.removeEventListener('message', messageListener);
   source.removeEventListener('error', errorListener);
 
   source.close();
 
-  delete sources[url];
+  delete URLToEventSourceMap[url];
 }
 
 const SystemLogActions = {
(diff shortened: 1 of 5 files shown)


orlandohohmeier:
run integration tests

orlandohohmeier:
run integration tests

nLight:
giphy-2

src/js/events/SystemLogActions.js
```diff
@@ -39,17 +76,14 @@ const SystemLogActions = {
...
     // Unsubscribe if any open connection exists with the same ID
-    this.unsubscribe(subscriptionID);
+    this.stopTail(subscriptionID);
 
     const url = SystemLogUtil.getUrl(nodeID, options);
-    subscriptionID = subscriptionID || Symbol(url + Date.now());
-    const source = new EventSource(url, {
-      withCredentials: Boolean(CookieUtils.getUserMetadata())
-    });
+    subscriptionID = subscriptionID || Symbol.for(url);
```
mlunoe: The `Date.now()` was to keep them unique, if you have multiple requests with the same parameters. I'm not sure this edge would occur, but if the client expects a response and never gets one that would have a bad impact

src/js/events/SystemLogActions.js
```diff
@@ -181,16 +203,11 @@ const SystemLogActions = {
...
-      // Close connection,
-      // no need to delete source as the reference is not stored
-      source.close();
+
+      unsubscribe(subscriptionID);
     }
 
-    source.addEventListener('message', messageListener, false);
-    source.addEventListener('error', errorListener, false);
+    subscribe(url, subscriptionID, messageListener, errorListener);
```
mlunoe: When using range, we don't need to store the source! I guess this will work, but again I would add the `Date.now()` to ensure uniqueness

src/js/events/SystemLogActions.js
orlandohohmeier: `Symbol("debug name")` is always unique.

mlunoe: Yup, I do feel it is important even if it is a rare edge-case!

mlunoe: Thinking about it a bit more… when I am requesting the same stream again, I guess it would make sense to close any outgoing connection, if there is one open. I think you are right, that it is not needed

orlandohohmeier:
Pushed a fix to address issues to with accidental unsubscribes due to reusing the `subscriptionID` to handle the connection.

src/js/events/SystemLogActions.js
```diff
@@ -123,7 +124,9 @@ const SystemLogActions = {
       });
     }
 
-    subscribe(url, subscriptionID, messageListener, errorListener);
+    tials[subscriptionID] = url;
```
nLight: Typo? tials -> tails

orlandohohmeier: yep 😊

jfurrow:
Works well for me! Great work, everyone! :ship:

PR #1989, opened 2017-03-10
Wrap job command in <pre> tag

jfurrow:
This PR wraps the job's command in a `pre` tag to match the service configuration view.

Before:

After:

**Checklist**


rcorral:
Looks good, tested manually.

PR #1991, opened 2017-03-10
Don't allow flex shrinking on page header

jfurrow:
This PR fixes a weird bug in IE11 where the `.page-header` element would grow a little bit when the `.page-body-content` content changed. I think it's due to the overflowing children of `.page-body-content`; the bug seems to occur when the content was overflowing, then isn't (due to its content changing to a loading indicator), and then back to overflowing (due to the content being loaded).

Reproduction steps (in IE11 only):
1. Navigate to the service detail page
2. Select a task to view the task detail page
3. Navigate between the tabs

Before:

After:

**Checklist**

### DIFF
--- a/src/styles/components/page-header/styles.less
+++ b/src/styles/components/page-header/styles.less
@@ -4,7 +4,7 @@
 
   .page-header {
     .clearfix();
-    flex: 0 1 auto;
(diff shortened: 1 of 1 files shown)


nLight:
Works like a charm! 💯

PR #1605, opened 2016-12-14
DCOS-11750: add size to external volumes

Poltergeist:
This introduces size to external volumes.
Fixes volumes on root level in the reducer.
And also fixes container type to be set with external volumes.

PR #1992, opened 2017-03-13
DCOS-14449: Remove outdated volume marathon app validation

mlunoe:
This PR reverts the introduction of an appValidator introduced here: https://github.com/dcos/dcos-ui/pull/1814

The validator is now out of date and the latest requirement is that it should just be a string: https://github.com/mesosphere/marathon/blob/master/docs/docs/rest-api/public/api/v2/types/volumes.raml#L80

There was also a bug with parsing, where it would always try to parse an integer if passed to any of the volume fields. Here is where it was introduced:
https://github.com/dcos/dcos-ui/pull/1605/files#diff-7df350c7661c694dcce5e1552dcde9feR53

Before (checks if it contains `/` which is ok now):
Before

After (only validates for string):
After

**Checklist**

### DIFF
--- a/plugins/services/src/js/reducers/serviceForm/ExternalVolumes.js
+++ b/plugins/services/src/js/reducers/serviceForm/ExternalVolumes.js
@@ -3,6 +3,7 @@ import {
   REMOVE_ITEM,
   SET
 } from '../../../../../../src/js/constants/TransactionTypes';
+import {parseIntValue} from '../../../../../../src/js/utils/ReducerUtil';
 import Transaction from '../../../../../../src/js/structs/Transaction';
 
 module.exports = {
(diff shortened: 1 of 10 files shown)


jfurrow:
Works well! :ship

plugins/services/src/js/reducers/serviceForm/ExternalVolumes.js
```diff
@@ -116,22 +117,23 @@ module.exports = {
 
       const index = path[1];
       if (type === SET && `externalVolumes.${index}.provider` === joinedPath) {
-        state[index].provider = value;
+        state[index].provider = String(value);
```
orlandohohmeier: Why do we need to cast the value to strings?

PR #1995, opened 2017-03-14
Make serviceID prop not required

nLight:
Why:

* we don't have any on the loading screen and yet want to render breadcrumbs

This change addresses the need by:

* removing isRequired from the prop

### DIFF
--- a/plugins/services/src/js/components/ServiceBreadcrumbs.js
+++ b/plugins/services/src/js/components/ServiceBreadcrumbs.js
@@ -158,10 +158,6 @@ class ServiceBreadcrumbs extends React.Component {
 
   render() {
     const {serviceID, taskID, taskName, extra} = this.props;
-    const trimmedServiceID = decodeURIComponent(serviceID).replace(/^\//, '');
-    const ids = trimmedServiceID.split('/');
-
-    let aggregateIDs = '';
 
     let crumbs = [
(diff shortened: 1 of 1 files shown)


plugins/services/src/js/components/ServiceBreadcrumbs.js
```diff
@@ -236,7 +236,7 @@ class ServiceBreadcrumbs extends React.Component {
 
 ServiceBreadcrumbs.propTypes = {
   extra: React.PropTypes.arrayOf(React.PropTypes.node),
-  serviceID: React.PropTypes.string.isRequired,
+  serviceID: React.PropTypes.string,
```
wavesoft: Since this is not required shouldn't we provide some reasonable defaults?
Right now if it's missing, the `trimmedServiceID` would be `undefined`, right?

MatApple:
LGTM after @wavesoft's comment so I give it my label 👍

plugins/services/src/js/components/ServiceBreadcrumbs.js
nLight: Added an empty string as a default

wavesoft:
Great! 👍  Merge when tests pass 👍

nLight:
run unit tests

PR #1997, opened 2017-03-14
DCOS-14120: Optimise logs (CPU)

mlunoe:
This PR optimises the `LogView` and the SSE stream of logs by throttling the event stream and by being cautious of reading/writing to properties that causes reflow.

Before:             |  After:
:-------------------------:|:-------------------------:
Before  |  After

Notice the difference in execution time of the `handleUpdateScrollPosition` function, but also the difference in frequency

**Checklist**

### DIFF
--- a/src/js/utils/SystemLogUtil.js
+++ b/src/js/utils/SystemLogUtil.js
@@ -67,6 +67,44 @@ const SystemLogUtil = {
     }, []);
 
     return `${Config.logsAPIPrefix}/${nodeID}/logs/v1/${base}/${idArray.join('/')}${endpoint}?${paramsArray.join('&')}`;
+  },
+
+  /**
+   * A throttle function that will make sure callback is not called more than
+   * once per `wait` miliseconds. It will accumulate arguments and pass it to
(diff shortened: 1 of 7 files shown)


src/js/utils/SystemLogUtil.js
```diff
@@ -67,6 +67,46 @@ const SystemLogUtil = {
     }, []);
 
     return `${Config.logsAPIPrefix}/${nodeID}/logs/v1/${base}/${idArray.join('/')}${endpoint}?${paramsArray.join('&')}`;
+  },
+
+  accumulatedThrottle(callback, wait, useAccumulation = true) {
```
nLight: Since it is _accumulated_ throttle I'd drop `useAccumulation` argument, we have a normal `throttle` for that or am I missing something?

mlunoe: Done!

nLight:
LGTM! 💯  💖

mlunoe:
Ping @jfurrow

jfurrow:
Great job @mlunoe! This feels way more snappy.

PR #2000, opened 2017-03-15
DCOS-13032: Aggregate child tasks resources

nLight:
With this PR we stop counting a framework's child tasks as its instances and properly aggregate their resources instead.

Note! It works only one level deep, so if you run MoM and then Kafka on that MoM and then something on that Kafka - only Kafka's resource will be aggregated.

Testing

Easy mode: start confluent-kafka, wait until broker task starts, verify
Harder: start MoM, start a bunch of tasks on MoM, verify

**Checklist**

### DIFF
--- a/plugins/services/src/js/containers/services/ServicesContainer.js
+++ b/plugins/services/src/js/containers/services/ServicesContainer.js
@@ -15,6 +15,7 @@ import ServiceModals from '../../components/modals/ServiceModals';
 import ServiceTree from '../../structs/ServiceTree';
 import ServiceTreeView from './ServiceTreeView';
 
+import MesosStateStore from '../../../../../../src/js/stores/MesosStateStore';
 import AppDispatcher from '../../../../../../src/js/events/AppDispatcher';
 import ContainerUtil from '../../../../../../src/js/utils/ContainerUtil';
 import Icon from '../../../../../../src/js/components/Icon';
@@ -34,7 +35,8 @@ import ServiceAttributeIsUniverseFilter from '../../filters/ServiceAttributeIsUn
 import ServiceAttributeHasVolumesFilter from '../../filters/ServiceAttributeHasVolumesFilter';
 
 import {
-  DCOS_CHANGE
+  DCOS_CHANGE,
+  MESOS_STATE_CHANGE
 } from '../../../../../../src/js/constants/EventTypes';
 
 import {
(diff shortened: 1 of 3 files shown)


mlunoe:
PR number 2000 🙌

plugins/services/src/js/structs/Framework.js
```diff
@@ -51,4 +45,40 @@ module.exports = class Framework extends Application {
 
     return {value};
   }
+
+  getResources() {
+    // FIXME: Circular reference workaround
+    const MesosStateStore = require('../../../../../src/js/stores/MesosStateStore');
```
nLight: Previous behavior, counting child tasks as instances, doesn't make sense and is wrong. 
In case MesosState is not there it will calc only the framework itself without its children.

I could get MesosStateStore as an argument but that would break the interface and I'd rather not break it.

plugins/services/src/js/containers/services/ServicesContainer.js
```diff
@@ -151,6 +153,9 @@ class ServicesContainer extends React.Component {
   componentDidMount() {
     DCOSStore.addChangeListener(DCOS_CHANGE, this.onStoreChange);
 
+    // Don't block the whole screen if Mesos state is not there
+    MesosStateStore.addChangeListener(MESOS_STATE_CHANGE, function () {});
```
mlunoe: I am not sure why this is here? Could you elaborate a bit?

plugins/services/src/js/structs/Framework.js
mlunoe: Could we do a `TODO` and add a Jira ticket?

plugins/services/src/js/containers/services/ServicesContainer.js
nLight: We need to subscribe to the store so that we get tasks for frameworks.
Notice that we're using `MesosStateStore` in Framework.js

mlunoe: Ah ok. Could we add that explicitly to the comment?

nLight:
@wavesoft @mlunoe addressed!

plugins/services/src/js/structs/Framework.js
```diff
@@ -51,4 +45,40 @@ module.exports = class Framework extends Application {
 
     return {value};
   }
+
+  getResources() {
+    // TODO: Circular reference workaround DCOS_OSS-783
+    const MesosStateStore = require('../../../../../src/js/stores/MesosStateStore');
```
nLight: or we could find different approach of checking on an instance class. Currently we import Framework in the Store to check if an instance is a Framework. :man_shrugging:

nLight:
@wavesoft :poke:

wavesoft:
Looks great! 👍

PR #2001, opened 2017-03-15
DCOS-13950: Fix sorting by health

weblancaster:
This PR fixes the issue when sorting by health in service table, adds minor refactoring (naming, code structure), unit test improvement based on code changes.

After talking to @mlunoe we decided that code fix was needed for the sorting to work properly as well.

As a effort to move from this [reusable sorting](https://github.com/dcos/dcos-ui/blob/2e63bbbc94875485c6dadfa9298649b42a6f213f/plugins/services/src/js/utils/TaskTableUtil.js#L19) because of reasons, code complexity, hard to follow code flow and hard to create more precise unit tests. 

That being said code refactoring was added, removing the sorting work from the `reusable sorting` function health and created a new method targeting specifically health sorting.

**Checklist**

### DIFF
--- a/plugins/services/src/js/utils/TaskTableUtil.js
+++ b/plugins/services/src/js/utils/TaskTableUtil.js
@@ -3,7 +3,6 @@ import React from 'react';
 /* eslint-enable no-unused-vars */
 
 import HealthSorting from '../constants/HealthSorting';
-import MarathonStore from '../stores/MarathonStore';
 import TableUtil from '../../../../../src/js/utils/TableUtil';
 import TaskStatusSortingOrder from '../constants/TaskStatusSortingOrder';
 import Util from '../../../../../src/js/utils/Util';
@@ -15,18 +14,72 @@ function getUpdatedTimestamp(model) {
 }
 
 const TaskTableUtil = {
+  /**
+   * Normalize param received and try mapping to HealthSorting
+   * if not able to mapping default to HealthSorting.NA
+   *
+   * @param {String} name
+   * @returns {Number} HealthSorting value
+   */
+  getHealthValueByName(name) {
+    let match;
+    name = Util.toUpperCaseIfString(name);
+    match = HealthSorting[name];
+
+    // defaults to NA if can't map to a HealthTypes key
+    if ( typeof match !== 'number' ) {
+      match = HealthSorting.NA;
+    }
+
+    return match;
+  },
+
+  /**
+   * Order health status
+   * based on HealthSorting mapping value
+   * where lowest (0) (top of the list) is most important for visibility
+   * and highest (3) (bottom of the list) 3 is least important for visibility
+   *
+   * @param {Object} a
+   * @param {Object} b
+   * @returns {Number} item position
+   */
+  sortHealthValues(a, b) {
+    const aTieBreaker = Util.toLowerCaseIfString(a.id);
+    const bTieBreaker = Util.toLowerCaseIfString(b.id);
+
+    a = TaskTableUtil.getHealthValueByName(a.health);
+    b = TaskTableUtil.getHealthValueByName(b.health);
+
+    if (a === b) {
+      a = aTieBreaker;
+      b = bTieBreaker;
+    }
+
+    if (a > b) {
+      return 1;
+    }
+    if (a < b) {
+      return -1;
+    }
+
+    return 0;
+  },
+
+/**
+ *
+ * @returns {Function} sortHealthValues
+ */
+  getHealthSorting() {
+    return TaskTableUtil.sortHealthValues;
+  },
+
   getSortFunction(tieBreakerProp) {
     return TableUtil.getSortFunction(tieBreakerProp, function (item, prop) {
       const hasGetter = typeof item.get === 'function';
 
       if (prop === 'updated') {
         return getUpdatedTimestamp(item) || 0;
-      }
-
-      if (prop === 'health') {
-        return HealthSorting[
-          MarathonStore.getServiceHealth(item.get('name')).key
-        ];
       }
 
       if (prop === 'status' && !hasGetter) {
--- a/plugins/services/src/js/utils/__tests__/TaskTableUtil-test.js
+++ b/plugins/services/src/js/utils/__tests__/TaskTableUtil-test.js
@@ -1,17 +1,12 @@
 jest.dontMock('../../constants/HealthSorting');
 jest.dontMock('../TaskTableUtil');
 
-const MarathonStore = require('../../stores/MarathonStore');
 const Service = require('../../structs/Service');
 const TaskTableUtil = require('../TaskTableUtil');
+const HealthSorting = require('../../constants/HealthSorting');
 
 describe('TaskTableUtil', function () {
   beforeEach(function () {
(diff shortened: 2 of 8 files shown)


.gitignore
```diff
@@ -17,3 +17,4 @@ jest/config.json
 jest/test-results
 webpack/proxy.dev.js
 src/resources/raml
+.nvmrc
```
weblancaster: Makes sense, will do next time if fine with everyone else.

mlunoe: Actually, let's remove it. I'll put up a PR to add it.
We don't want this in the `.gitignore`, if we want to add it to the project.

mlunoe: Created a PR here: https://github.com/dcos/dcos-ui/pull/2005

weblancaster: ok cool, thanks.

Poltergeist:
How can I reproduce the initial issue? I am trying to check the PR, and cannot reproduce the issue.

plugins/services/src/js/utils/__tests__/TaskTableUtil-test.js
```diff
@@ -9,13 +9,14 @@ describe('TaskTableUtil', function () {
   beforeEach(function () {
     this.getServiceHealth = MarathonStore.getServiceHealth;
     MarathonStore.getServiceHealth = function (prop) {
-      return this[prop].get('health');
+      return this[prop]['health'];
```
Poltergeist: The item should be of type `Item`. Which includes the get function so this line is fixing the issue but not the problem. Let's rather try to find the source of the problem. And ensure that all tasks are of type `Item`

plugins/services/src/js/utils/__tests__/TaskTableUtil-test.js
```diff
@@ -52,7 +54,7 @@ describe('TaskTableUtil', function () {
       }
     });
 
-    this.sortFunction = TaskTableUtil.getSortFunction('name');
+    this.getComparator = TaskTableUtil.getSortFunction('name');
```
Poltergeist: If we decide to call the sort functions now comparator, which I do like, we should do this in a separate PR and apply it to the complete code base.

plugins/services/src/js/utils/__tests__/TaskTableUtil-test.js
mlunoe: I don't see us using tasks as `Items` anywhere. We do this for Services.
Either we should make the effort and make everywhere we use tasks to structs or we keep this change to do as we do elsewhere when working with tasks.
This spawns a larger discussion about the future of structs in our code base and what the plan is to get where we want.
I am ok with this change as it follows what we do elsewhere.

plugins/services/src/js/utils/__tests__/TaskTableUtil-test.js
mlunoe: I think we can still do this change in this PR. This helps readability of what is going on, but I do agree we should rename throughout the code base

plugins/services/src/js/utils/__tests__/TaskTableUtil-test.js
weblancaster: @Poltergeist my main reason to remove the structs is that in this case is only creating more complexity in the code base because the get method is only a abstraction of the native notation which objects are being extended to use the method.

I'm totally fine changing to use the struct but I believe we want to move out using it.

plugins/services/src/js/utils/__tests__/TaskTableUtil-test.js
weblancaster: I believe that changing incrementally as we go is a better option at the moment as developers work on files that needs to be refactor in a way to make more readable but I'm also fine removing from this PR and creating a Jira ticket to refactor the naming.

plugins/services/src/js/utils/__tests__/TaskTableUtil-test.js
MatApple: @Poltergeist I agree with you that we shouldn't have inconsistencies. However, as @mlunoe points out - we've never enforced structs for tasks. In fact IMO, our structs (at least how they currently work via `Item.get('property')` causes far more complexity than was intended. We will be deprecating structs with Relay/GraphQL. Let's keep what @weblancaster has proposed as this will follow a concept that will persist (i.e. no structs).

plugins/services/src/js/utils/__tests__/TaskTableUtil-test.js
MatApple: How about we leave this file as is AND create a JIRA ticket for renaming the comparator's? We should also be mindful that as we scale, it will be more difficult to enforce "soft" rules like this. We want to delegate as much as possible to the linter and outside of that we will have to pick our battles, keeping in mind that we also want to move fast and with community contribution. If something is easier to reason about let's use it and count it as an improvement.

plugins/services/src/js/utils/__tests__/TaskTableUtil-test.js
Poltergeist: Okay I didn't know that we do not use structs on tasks my bad. I just saw the `item.get` and concluded that it was a struct. My bad then forget what I said here. Nice catch!

plugins/services/src/js/utils/__tests__/TaskTableUtil-test.js
Poltergeist: 👍 MatApple, Let's keep the change and add the JIRA ticket.

plugins/services/src/js/utils/TaskTableUtil.js
```diff
@@ -23,9 +23,9 @@ const TaskTableUtil = {
         return getUpdatedTimestamp(item) || 0;
       }
 
-      if (prop === 'health') {
+      if (prop === 'health' && !hasGetter ) {
```
mlunoe: Tiny #nit: You have an extra space inside the if-statement at the end

weblancaster: good catch

mlunoe:
I might be missing something, but I think it is not working as expected:

weblancaster:
Just fyi I have been working on a change for this PR based on my conversation with @mlunoe I'm also fixing the sorting besides the error thrown (the current changes here are only because I changed computers so I pushed what I had).

weblancaster:
**before:**
https://cl.ly/3s0x430O0z1Y/Screen%20Recording%202017-03-22%20at%2005.14%20PM.gif

**after:**
https://cl.ly/0I1M2k2Z2A33/download/Screen%20Recording%202017-03-22%20at%2005.24%20PM.gif

orlandohohmeier:
For you next PR, please don't squash all changes into one commit, so that we have a complete log of the things changed and the reasoning behind those changes.

weblancaster:
Sounds good @orlandohohmeier I will start doing micro-commits going forward for big PRs changes, but you still get a complete log view of the changes looking the files changed and the reasons behind should be on the description.

weblancaster:
retest this please

wavesoft:
Looks, works great! Good work ✨  👍

mlunoe:
🎉 Yaaaay!

PR #2008, opened 2017-03-16
DCOS-13710 [preparation]: Fix the code that prettier can't process 

nLight:
Currently prettier can't prettify pieces of code where return statements have comments in it.
```javascript
return (
  // Comment
  someLogic !== here
);
```

So I decided to fix those as a separate PR so that these changes could be tested separately.

### DIFF
--- a/plugins/services/src/js/components/LogView.js
+++ b/plugins/services/src/js/components/LogView.js
@@ -64,23 +64,25 @@ class LogView extends React.Component {
   }
 
   shouldComponentUpdate(nextProps, nextState) {
-    const {hasLoadedTop, highlightText, logName, watching} = this.props;
-    const {fullLog, isAtBottom} = this.state;
+    const curState = this.state;
+    const curProps = this.props;
+    const propsToCheck = [
+      'hasLoadedTop',
+      'highlightText',
+      'logName',
+      'watching'
+    ];
 
-    return (
-      // Check hasLoadedTop
-      (hasLoadedTop !== nextProps.hasLoadedTop) ||
-      // Check highlightText
-      (highlightText !== nextProps.highlightText) ||
-      // Check logName
-      (logName !== nextProps.logName) ||
-      // Check watching
-      (watching !== nextProps.watching) ||
-      // Check isAtBottom
-      (isAtBottom !== nextState.isAtBottom) ||
-      // Check fullLog at the end, as this could be a long string
-      (fullLog !== nextState.fullLog)
-    );
+    const stateToCheck = [
+      'isAtBottom',
+      'fullLog' // Check fullLog at the end, as this could be a long string
+    ];
+
+    return propsToCheck.some(function (key) {
+      return curProps[key] !== nextProps[key];
+    }) || stateToCheck.some(function (key) {
+      return curState[key] !== nextState[key];
+    });
   }
 
   componentDidUpdate(prevProps) {
--- a/plugins/services/src/js/components/MesosLogContainer.js
+++ b/plugins/services/src/js/components/MesosLogContainer.js
@@ -79,46 +79,35 @@ class MesosLogContainer extends mixin(StoreMixin) {
   }
 
   shouldComponentUpdate(nextProps, nextState) {
-    const {
-      filePath,
-      highlightText,
-      logName,
-      onCountChange,
-      task,
-      watching
-    } = this.props;
-    const {
-      direction,
-      fullLog,
-      hasLoadingError,
-      isFetchingPrevious,
-      isLoading
-    } = this.state;
+    const curProps = this.props;
+    const curState = this.state;
 
-    return (
-      // Check filePath
-      (filePath !== nextProps.filePath) ||
-      // Check logName
-      (logName !== nextProps.logName) ||
-      // Check highlightText
-      (highlightText !== nextProps.highlightText) ||
-      // Check watching
-      (watching !== nextProps.watching) ||
-      // Check watching
-      (onCountChange !== nextProps.onCountChange) ||
-      // Check task (slave_id is the only property being used)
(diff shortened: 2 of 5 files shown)


plugins/services/src/js/components/LogView.js
```diff
@@ -64,23 +64,29 @@ class LogView extends React.Component {
...
-      (isAtBottom !== nextState.isAtBottom) ||
-      // Check fullLog at the end, as this could be a long string
-      (fullLog !== nextState.fullLog)
-    );
+    const stateToCheck = [
+      'isAtBottom',
+      'fullLog' // Check fullLog at the end, as this could be a long string
+    ];
+
+    const isPropsChanged = propsToCheck.some(function (key) {
```
MatApple: My only #nit is that perhaps we use `did` instead of `is`? e.g. `didPropsChange`. What do you think?

nLight: Sure! That's a good point!

I miss the possibility to use `?` in function names as you can do in ruby. With that trick you can get away with not using a popper grammar `propsChanged?` :D

nLight:
@MatApple fixed!

nLight:
Ping @orlandohohmeier @jfurrow 🙇

plugins/services/src/js/components/LogView.js
```diff
@@ -64,23 +64,29 @@ class LogView extends React.Component {
...
+
+    const didPropsChange = propsToCheck.some(function (key) {
+      return curProps[key] !== nextProps[key];
+    });
+
+    const isStateChanged = stateToCheck.some(function (key) {
+      return curState[key] !== nextState[key];
+    });
+
+    return didPropsChange || isStateChanged;
```
jfurrow: This is definitely more legible, but it does mean that we might be unnecessarily checking for changes in `state` if we have already found changes in `props`. One solution would be to write this as follows:

```js
return propsToCheck.some(function (key) {
  return curProps[key] !== nextProps[key];
}) || stateToCheck.some(function (key) {
  return curState[key] !== nextState[key];
});
```

What do you think?

plugins/services/src/js/components/MesosLogContainer.js
```diff
@@ -79,46 +79,37 @@ class MesosLogContainer extends mixin(StoreMixin) {
...
-      // Check isLoading
-      (isLoading !== nextState.isLoading) ||
-      // Check fullLog at the end, as this could be a long string
-      (fullLog !== nextState.fullLog)
-    );
+    const isStateChanged = stateToCheck.some(function (key) {
+      return curState[key] !== nextState[key];
+    });
+
+    return didPropsChange || isTaskChanged || isStateChanged;
```
jfurrow: Same comment as https://github.com/dcos/dcos-ui/pull/2008/files#r109492478

plugins/services/src/js/components/LogView.js
orlandohohmeier: Let's split this into separate condition to early exit as suggest by @jfurrow

plugins/services/src/js/components/LogView.js
```diff
@@ -64,23 +64,29 @@ class LogView extends React.Component {
...
+    const stateToCheck = [
+      'isAtBottom',
+      'fullLog' // Check fullLog at the end, as this could be a long string
+    ];
+
+    const didPropsChange = propsToCheck.some(function (key) {
+      return curProps[key] !== nextProps[key];
+    });
+
+    const isStateChanged = stateToCheck.some(function (key) {
```
MatApple: Perhaps we could change cases of `isStateChanged` to `didStateChange` too as I think it makes more sense.

nLight:
Thanks for the suggestion @jfurrow! Addressed all comments, please give it another shot
@MatApple @orlandohohmeier

orlandohohmeier:
I think that we should reduce the complexity of our `shouldComponentUpdate` methods as there way to many conditions but that is a different issue and not part of this PR. Great job! 💯 🥇

jfurrow:
LGMT @nLight! I'll hold off on merging until @MatApple has had a chance to review the changes.

mlunoe:
@MatApple can we get an approval?

mlunoe:
I rebased this as it's following PR had conflicts and I am resolving those

plugins/services/src/js/components/modals/NewCreateServiceModalForm.js
```diff
@@ -159,12 +159,17 @@ class NewCreateServiceModalForm extends Component {
       return true;
     }
 
+    const isBaseConfigChanged = this.state.baseConfig !== nextState.baseConfig;
```
MatApple: @jfurrow @mlunoe could one of you please update to use `did` instead of `is`? I think it's very confusing when we mix present and past tense in a variable name.

mlunoe: I am actually unsure how you want it, so I left it for you to do. Or, could you be a bit more specific, so I know what to do here? Do you want _every_ variable in this change renamed to `did`?

MatApple: Thanks @mlunoe - I won't get a chance to do this for a couple of days so I appreciate the help if someone can make the changes. Instead of `isBaseConfigChanged` it should be `didBaseConfigChange`. We are mixing tenses and it's confusing.

MatApple: All variables that look like `is...Changed` should be changed to `did...Change`

jfurrow: Done!

mlunoe:
Run integration tests!

MatApple:
run unit tests

PR #2010, opened 2017-03-17
Restyle table row actions dropdown

leemunroe:
This PR restyles the actions dropdown for service collection rows and moves it to the right. 

Why? We want a consistent style and location for row actions throughout DC/OS. Using the ellipsis and positioning it to the right will keep us consistent and also keeps us open to adding more actions here. A tooltip has also been added for extra clarity.

Before
image

After
image

### DIFF
--- a/plugins/services/src/js/containers/services/ServicesTable.js
+++ b/plugins/services/src/js/containers/services/ServicesTable.js
@@ -1,5 +1,5 @@
 import classNames from 'classnames';
-import {Dropdown, Table} from 'reactjs-components';
+import {Dropdown, Table, Tooltip} from 'reactjs-components';
 import {Link} from 'react-router';
 import React, {PropTypes} from 'react';
 import {ResourceTableUtil} from 'foundation-ui';
@@ -28,14 +28,16 @@ const columnClasses = {
   status: 'service-table-column-status',
   cpus: 'service-table-column-cpus',
   mem: 'service-table-column-mem',
-  disk: 'service-table-column-disk'
+  disk: 'service-table-column-disk',
+  actions: 'service-table-column-actions'
 };
 
 const METHODS_TO_BIND = [
   'onActionsItemSelection',
   'renderHeadline',
   'renderStats',
-  'renderStatus'
+  'renderStatus',
+  'renderServiceActions'
 ];
 
 class ServicesTable extends React.Component {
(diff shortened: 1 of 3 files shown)


plugins/services/src/js/containers/services/ServicesTable.js
```diff
@@ -210,21 +210,24 @@ class ServicesTable extends React.Component {
...
-        dropdownMenuListItemClassName="clickable"
-        wrapperClassName="dropdown flush-bottom table-cell-icon"
-        items={dropdownItems}
-        persistentID={ServiceActionItem.MORE}
-        onItemSelection={this.onActionsItemSelection.bind(this, service)}
-        scrollContainer=".gm-scroll-view"
-        scrollContainerParentSelector=".gm-prevented"
-        transition={true}
-        transitionName="dropdown-menu" />
+      <Tooltip interactive={true} content={'More actions'}>
```
jfurrow: Let's adjust this a little bit to: `<Tooltip content="More actions">`

We don't want `interactive={true}` in this case (the component defaults to `interactive={false}`. The `interactive` property toggles the ability for the user's cursor to enter the tooltip vs being dismissed when the user's cursor leaves the trigger element. We set `interactive={true}` when the tooltip contains links or actions that the user needs to interact with.

plugins/services/src/js/containers/services/ServicesTable.js
```diff
@@ -210,21 +210,24 @@ class ServicesTable extends React.Component {
...
+        <Dropdown
+          key="actions-dropdown"
```
jfurrow: Let's kill the `key` property. It was needed previously because this element was an element in an array, whereas now it's being rendered directly in JSX.

plugins/services/src/js/containers/services/ServicesTable.js
mlunoe: I think you want `interactive={false}` here. There are no links in the tooltip.

plugins/services/src/js/containers/services/ServicesTable.js
mlunoe: Oh, good catch! I missed that!

mlunoe:
Haha! Double review 😝

MatApple:
@leemunroe @mlunoe @jfurrow - I applied the requested changes on behalf of @leemunroe. Nice work @leemunroe!

mlunoe:
We have some legit integration tests failing:

leemunroe:
@mlunoe @jfurrow @MatApple Failing because I removed the `dropdown-toggle` class (which added a dropdown caret to the icon). 

What's the protocol for this situation and tests? e.g. I could change the test and add a class with a `js-` prefix so the class is only for JS and not styling. Or add a custom data attribute?

jfurrow:
@leemunroe We'll change the test to check for the dropdown in the proper position in the DOM. I'll help you out with getting the tests set up today and then we can quickly adjust this test.

jfurrow:
retest this please

MatApple:
Since this has 2 reviewers approving and my label, I added an extra label for admin tracking.

PR #2012, opened 2017-03-17
DCOS-14630: Split up artifacts reducer

Poltergeist:
**Checklist**

Move FormReducer, JSONReducer/Parser to
their own file, being `FormReducer/Artifacts.js` and `JSONReducer/Artifacts.js`.

Close DCOS-14630


plugins/services/src/js/reducers/JSONMultiContainerReducers.js
```diff
@@ -24,7 +24,7 @@ module.exports = {
       }
     };
   },
-  fetch,
```
nLight: so was it a bug?

plugins/services/src/js/reducers/serviceForm/FormReducers/Artifacts.js
```diff
@@ -0,0 +1,44 @@
+import {
+  ADD_ITEM,
+  REMOVE_ITEM,
+  SET
+} from '../../../../../../../src/js/constants/TransactionTypes';
+
+function processTransaction(state, {type, path, value}) {
```
nLight: Do we duplicate the code 🚨 ?

Poltergeist: You mean duplication of the code in the JSON reducer? Well I had an interessting discussion about that the other day. Basically I was also thinking this is duplication, but it actually isn't. With one function we are creating the json output and with the other one we are creating the Form output. In this case these both happen to be the same. So if one changes the other one will still work.

nLight: ok, that makes sense

jfurrow:
Nice work @Poltergeist!

PR #2014, opened 2017-03-17
Sort eslin rules

nLight:
This PR does nothing else but sorts eslint rules. So that when I start moving them into dcos/javascript it's easier to trace changes.


wavesoft:
Confirmed with `deep-equal` 👍  😎 

Chunk norris approves

PR #2015, opened 2017-03-17
DCOS-14171: Clean up usage and styling of DeleteRowButton

jfurrow:
This PR cleans up the styling of the `DeleteRowButton` component. It removes all horizontal padding on the delete row button, uses `column-auto` instead of `column-2`, and increases the column width of all sibling `FormGroups` to ensure that the delete button takes up only as much space as is necessary. I also moved all classes to the `FormGroup` element so that we don't have to remember which classes to use in the future.

This adds two new props to the `FormGroup` component:
`isDeleteButton` — when true will apply the correct `column` and `flush` classes
`applyLabelOffset` — when true will apply the correct class, adds padding to the top of the form group so that it aligns vertically with its siblings who have labels

Before (note the real estate used by the button's column and its alignment with other delete buttons):

After:

**Checklist**

### DIFF
--- a/plugins/services/src/js/components/forms/ArtifactsSection.js
+++ b/plugins/services/src/js/components/forms/ArtifactsSection.js
@@ -62,7 +62,7 @@ class ArtifactsSection extends Component {
       return (
         <FormRow key={`${path}.${index}`}>
           <FormGroup
-            className="column-10"
+            className="column-12"
             showError={Boolean(error)}>
             {label}
             <FieldInput
@@ -71,7 +71,7 @@ class ArtifactsSection extends Component {
               value={item.uri}/>
             <FieldError>{error}</FieldError>
           </FormGroup>
-          <FormGroup className="flex flex-item-align-end column-2 flush-left">
+          <FormGroup isDeleteButton={true} applyLabelOffset={index === 0}>
             <DeleteRowButton
               onClick={this.props.onRemoveItem.bind(
                 this,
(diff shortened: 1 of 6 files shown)


plugins/services/src/js/components/forms/ArtifactsSection.js
```diff
@@ -71,7 +71,7 @@ class ArtifactsSection extends Component {
               value={item.uri}/>
             <FieldError>{error}</FieldError>
           </FormGroup>
-          <FormGroup className="flex flex-item-align-end column-2 flush-left">
+          <FormGroup isDeleteButton={true} applyLabelOffset={index === 0}>
```
wavesoft: I have some concerns for the `isDeleteButton` name in the `FormGroup` since the two concepts are a bit different. I would vote for `narrowMargins` or something like this, that would describe the styling and not the functionality.

jfurrow: Good idea! I think that's much better.

tests/pages/services/ServiceFormModal-cy.js
```diff
@@ -401,6 +401,30 @@ describe('Service Form Modal', function () {
           .contains('Add Placement Constraint');
       });
 
+      it.only('Should vertically align the placement constraint delete row button', function () {
```
MatApple: `.only`

plugins/services/src/js/components/forms/ArtifactsSection.js
jfurrow: Done!

tests/pages/services/ServiceFormModal-cy.js
jfurrow: Removed!

PR #2017, opened 2017-03-17
DCOS-13209: Reduce complex logic in #wrapElement

jfurrow:
This PR refactors the `ReactUtil#wrapElements` method to be more readable. This comes from a comment in a PR: https://github.com/dcos/dcos-ui/pull/1780#discussion_r95982491

**Checklist**

### DIFF
--- a/foundation-ui/utils/ReactUtil.js
+++ b/foundation-ui/utils/ReactUtil.js
@@ -1,5 +1,18 @@
 import React from 'react';
 
+function getUnwrappedElements(elements) {
+  if (Array.isArray(elements) && elements.length === 1
+    && React.isValidElement(elements[0])) {
+    return elements[0];
+  }
+
+  return elements;
+}
+
+function shouldWrapElements(elements, alwaysWrap) {
+  return alwaysWrap || (Array.isArray(elements) && elements.length > 1);
+}
+
 const ReactUtil = {
   /**
    * Wrap React elements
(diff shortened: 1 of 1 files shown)


jfurrow:
run integration tests

MatApple:
Nice @jfurrow!

orlandohohmeier:
@MatApple why did you merged this one? Every PR must have two labels before it's ok to hit the merge button.

foundation-ui/utils/ReactUtil.js
```diff
@@ -18,16 +31,11 @@ const ReactUtil = {
...
     }
 
-    if (Array.isArray(elements) && elements.length === 1
-      && React.isValidElement(elements[0]) && !alwaysWrap) {
-      return elements[0];
-    }
-
-    if (React.isValidElement(elements) && !alwaysWrap) {
-      return elements;
+    if (shouldWrapElements(elements, alwaysWrap)) {
```
orlandohohmeier: What is the reasoning for moving the condition into a function?

foundation-ui/utils/ReactUtil.js
```diff
@@ -1,5 +1,18 @@
 import React from 'react';
 
+function getUnwrappedElements(elements) {
+  if (Array.isArray(elements) && elements.length === 1
```
orlandohohmeier: Please document this case so that it's cleat to everyone that the util simply renders single item arrays.

foundation-ui/utils/ReactUtil.js
jfurrow: We talked about this briefly in person, but I'll respond here for visibility:

I moved this to a separate function solely to help the readability of this method. I'm not married to this approach by any means, so let's discuss a better way to do this!

foundation-ui/utils/ReactUtil.js
jfurrow: Makes sense! I've done this offline, but I'll wait to push this until you and I can sync about the overall approach.

PR #2020, opened 2017-03-21
DCOS-14699: Support null value for placement constraints

nLight:
This PR replaces destructuring syntax with a default with a regular conditional assignment
so that `null` value also gets replaced by the default value

**Checklist**


wavesoft:
Ah. Nasty... Looks good to me ✨  👍

jfurrow:
👍 Let's merge when integration tests pass!

PR #2021, opened 2017-03-21
[System Tests]: Adding development environment dockerfile

wavesoft:
This PR introduces a `Dockerfile` that is used to build the [dcos-ui docker image](https://hub.docker.com/r/mesosphere/dcos-ui/) for use in development and build environments.

**NOTE:** Due to some bugs on the previous versions of `webpack` and `babel-loader` I also had to update the respective versions to the next available one.

This container **does not contain the dcos-ui sources**. It contains only the tooling required by the developer, the build system or the system integration tests.

In detail the docker image contains:
* Node 4.4.7 & npm 3.9
* The current branch's dependencies pre-installed
* Automatic initialisation script that runs `npm run scaffold` if requested to run on a clean repo

### How to use

You must mount the `dcos-ui` repository manually under the `/dcos-ui` volume. For example:

```
docker run -it --rm \
  -v `pwd`:/dcos-ui \
  -p 4200:4200 \
  mesosphere/dcos-ui:latest \
  npm start
```

_If you are developing your own plugins you can mount their directory in the `/dcos-ui-plugins` volume. The container will automatically set-up the environment to include them_ 

**Checklist**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,59 @@
+FROM mesosphere/dcos-system-test-driver:latest
+
+# Specify the component versions to use
+ENV CYPRESS_VERSION="0.19.1" \
+    NODE_VERSION="4.4.7" \
+    NPM_VERSION="3.9"
+
+# Expose the 4200 port
+EXPOSE 4200
+
+# Copy required files in order to be able to do npm install
+WORKDIR /dcos-ui
+COPY package.json npm-shrinkwrap.json scripts /dcos-ui/
+
+# Copy the entrypoint script that takes care of executing run-time init tasks,
+# such as creating the scaffold in the user's repository
+COPY scripts/docker-entrypoint /usr/local/bin/dcos-ui-docker-entrypoint
+
+# Install required components & prepare environment
+RUN set -x \
+
+  # Install node 4.4.7 & npm 3.9
+  && curl -o- https://nodejs.org/dist/v${NODE_VERSION}/node-v${NODE_VERSION}-linux-x64.tar.gz | tar -C /usr/local --strip-components=1 -zx \
+  && npm install -g npm@${NPM_VERSION} \
+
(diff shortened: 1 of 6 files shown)


webpack/webpack.production.babel.js
```diff
@@ -1,6 +1,7 @@
 import CompressionPlugin from 'compression-webpack-plugin';
 import ExtractTextPlugin from 'extract-text-webpack-plugin';
 import HtmlWebpackPlugin from 'html-webpack-plugin';
+import path from 'path';
```
nLight: Looks like it is not being used anywhere

Dockerfile
```diff
@@ -0,0 +1,59 @@
...
+# Copy required files in order to be able to do npm install
+WORKDIR /dcos-ui
+COPY package.json npm-shrinkwrap.json scripts /dcos-ui/
+
+# Copy the entrypoint script that takes care of executing run-time init tasks,
+# such as creating the scaffold in the user's repository
+COPY scripts/docker-entrypoint /usr/local/bin/dcos-ui-docker-entrypoint
+
+# Install required components & prepare environment
+RUN set -x \
```
nLight: Do you think it makes sense to split it so that we don't invalidate the layer of npm if we change something in scripts?

webpack/webpack.production.babel.js
wavesoft: Whoops, right 👍

Dockerfile
wavesoft: I would prefer to avoid creating an extra layer. I don't think we are saving much by doing so...

wavesoft:
run unit tests

Dockerfile
nLight: :okay:

jfurrow:
Works as expected! Nice work!

PR #2026, opened 2017-03-22
DCOS-14688: split up constraints

Poltergeist:
**Checklist**


jfurrow:
This seems to work exactly the same as master. Aside from the conflicts it's good to go!

PR #2032, opened 2017-03-23
DCOS-14696: Move Labels reducer

Poltergeist:
**Checklist**

Split up the labels reducer. As the JSON part should be together but the Form Reducer should stand alone as it is the logic for the form and not for JSON output.

close DCOS-14696


wavesoft:
LGTM! ✨  👍

PR #2033, opened 2017-03-23
DCOS-14697: split LocalVolumes Reducer

Poltergeist:
**Checklist**

Split local volumes form reducer and JSON parser.

close DCOS-14967


nLight:
Flicts!

nLight:
LGTM! 💯  💖

plugins/services/src/js/reducers/serviceForm/FormReducers/LocalVolumes.js
```diff
@@ -0,0 +1,52 @@
...
+            state = state.filter((item, index) => {
+              return index !== value;
+            });
+            break;
+        }
+
+        return state;
+      }
+
+      const index = joinedPath.match(/\d+/)[0];
```
mlunoe: We could make this a bit nicer:
```js
const [_base, index, name] = path;
if (type === SET && name === 'type') {
  state[index].type = String(value);
}
...
```

Poltergeist: I am okay with the change request but wouldn't it be better to change this in all places where this pattern has been used instead of only for this reducer? However I will change this in this PR for all the places where this pattern has been used in this PR and create an issue for the rest.

Poltergeist: As mentioned in the other PR I am not going to fix it in this PR but in DCOS-15145 I will tackle the issue for all other reducers.

PR #2040, opened 2017-03-24
[System Tests]: DCOS-12808: Add SI tests for pod with virtual network

jfurrow:
This PR introduces the automated tests for [Create a pod with a virtual network](https://wiki.mesosphere.com/pages/viewpage.action?pageId=101253133)

**The following test is expected to fail**, because some volume labels have changed (`MiB` to `GiB`, or vice versa, I can't remember) and they have not yet been released in the DCOS release branches:
```
✘ Services Applications Create an app with external volume
```

To run this test, use:
<pre>
docker pull mesosphere/dcos-ui && docker run -it --rm --ipc=host \
  -e CLUSTER_URL=$CLUSTER_URL  \
  -v `pwd`:/dcos-ui \
  ./system-tests/driver-config/cluster.sh
</pre>

### DIFF
--- a/system-tests/services/test-pods.js
+++ b/system-tests/services/test-pods.js
@@ -757,5 +757,181 @@ describe("Services", function() {
         .getTableRowThatContains(serviceName)
         .should("exist");
     });
+
+    it("Create a pod with virtual network", function() {
+      const serviceName = "pod-with-virtual-network";
+      const command = "python3 -m http.server 8080";
+      const containerImage = "python:3";
+
+      cy.contains("Multi-container (Pod)").click();
+
+      cy
+        .root()
+        .getFormGroupInputFor("Service ID *")
+        .type(`{selectall}{rightarrow}${serviceName}`);
+
+      cy.get(".menu-tabbed-item").contains("container-1").click();
+
+      // TODO: Due to a bug in cypress you cannot type values with dots
+      // cy
+      //   .root()
+      //   .getFormGroupInputFor('CPUs *')
+      //   .type('{selectall}0.5');
+
+      cy.root().getFormGroupInputFor("Memory (MiB) *").type("{selectall}32");
+
+      cy.root().getFormGroupInputFor("Container Image").type(containerImage);
+
+      cy.root().getFormGroupInputFor("Command").type(command);
+
+      cy.get(".menu-tabbed-item").contains("Networking").click();
+
+      cy
+        .root()
+        .getFormGroupInputFor("Network Type")
+        .select("Virtual Network: dcos");
+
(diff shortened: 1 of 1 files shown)


wavesoft:
@jfurrow flicts also here. :)

wavesoft:
Rebased this for you 👍

jfurrow:
Thanks @wavesoft!

Ping @MatApple @orlandohohmeier

jfurrow:
Sorry to be obnoxious! Ping @MatApple @orlandohohmeier

MatApple:
Not obnoxious at all @jfurrow! Thanks :-)

MatApple:
@jfurrow I'm wondering if we should change our pattern here:

**What we're doing now:** Testing for all our service configuration before we submit to the server, which essentially tests our form/JSON editor.

**What we're missing:** Testing the configuration of the service that was created, to make sure we sent the right data to the server and that Marathon created a service with the correct network configuration.

**What I suggest:** We move the service configuration testing from service creation and focus it on the configuration of the created service.

Thoughts?

jfurrow:
@MatApple That does make sense. These system test PRs are meant to automate the manual steps that we currently perform [as outlined by this document](https://wiki.mesosphere.com/pages/viewpage.action?pageId=101253133), but I'd be happy to add the additional steps.

So for this test, you're saying we should add tests that navigate to the Networks section and test that our network actually shows up in the table — is that right?

MatApple:
Thanks @jfurrow - I understand you're automating the manual steps and thanks so much for doing that! The manual test really is an integration test and not a system test. I do think it's highly valuable to have a system test here though. I understand my request technically changes the scope of your original task but I think we should take the time to do this right. cc/ @orlandohohmeier @rcorral (please see https://github.com/dcos/dcos-ui/pull/2040#issuecomment-295893797 for more context)

jfurrow:
Thanks @MatApple! That makes sense and I'm happy to do it in this PR. I still need some clarification, though. What would make this a system test? Is it simply navigating to the Networks section and testing that the table shows the new virtual network?

MatApple:
Hey @jfurrow, thanks for your willingness :) I spoke with @orlandohohmeier offline and Julian is going to add the extra steps to these tests to give us the additional coverage, so I think we are good to move forward without you altering this test.

To answer your question above, to complete this system test we should navigate to the service detail once the service was successfully created and we should then open the edit modal to check the JSON is correct. I.e. test that Marathon correctly interpreted what we sent it and created a service with the correct configuration.

jfurrow:
Thanks @MatApple! I appreciate the clarification, that makes good sense.

PR #2041, opened 2017-03-24
[foundation 3 of 3] DCOS-14745: Extract foundation

nLight:
ℹ️  License adds 200LOC, don't be intimidated by +333/-134 :)

## Local packages

- Introduced /packages dir for all the local packages we're not ready to completely extract.
As discussed offline with @orlandohohmeier it is considered beneficial since we don't need to mess around with `npm link` nor `npm install` as we change code inside those packages
- Rewired dcos-ui to catch everything from this dir and treat as an npm package while also including them in the `npm test` and `npm run lint` routines to make our life easier

## DCOSStore and ResourceTableUtil

- Moved from `foundation-ui` to DCOS since they are highly DCOS-UI specific

## PluginSDK

- Removed all the foundation bits from the PluginSDK, plugins now can just use the package
- Exposed DCOSStore and ResourceTableUtil

~Should go together with https://github.com/mesosphere/dcos-ui-plugins-private/pull/565~
~List of PRs: https://github.com/pulls?utf8=%E2%9C%93&q=is%3Aopen+is%3Apr+author%3AnLight+foundation~

🙇


jfurrow:
When testing this, I'm receiving an error and a warning. Is this to be expected?

nLight:
For some reason webpack wants to include LICENSE in the build. I suspect that's because foundation-ui package sits right there and maybe we need to exclude it somehow from the build.

Do we have webpack Gods on the team?

/cc @orlandohohmeier

foundation-ui/src/mount/MountService.js
```diff
@@ -43,7 +43,7 @@ class MountService extends EventEmitter {
     let {components, instance, rank} = this;
 
     if (typeof component !== 'function') {
-      if (global.__DEV__) {
+      if (process.env.NODE_ENV !== 'production') {
```
mlunoe: ❤️

src/js/routes/index.js
```diff
@@ -92,7 +92,7 @@ function getRoutes() {
 
   // Register packages
   indexRoute.children = indexRoute.children.concat(
-    RoutingService.getDefinition()
+    RoutingService.RoutingService.getDefinition()
```
mlunoe: 🤔 Could we not get rid of the duplicated name?

src/js/utils/NavigationServiceUtil.js
```diff
@@ -28,14 +29,14 @@ const NavigationServiceUtil = {
         const {path, category, childRoutes = []} = route;
         const primaryPath = `/${path}`;
 
-        NavigationService.registerCategory(category);
+        NavigationService.NavigationService.registerCategory(category);
```
mlunoe: Same here, could we not get rid of that duplicated name?

foundation-ui/index.js
```diff
@@ -1,17 +1,13 @@
-import DCOSStore from './stores/DCOSStore';
-import ResourceTableUtil from './utils/ResourceTableUtil';
-import RoutingService from './routing';
-import NavigationService from './navigation';
-import {MountService} from './mount';
-import Mount from './mount/Mount';
+import RoutingService from './src/routing';
+import NavigationService from './src/navigation';
```
mlunoe:
```js
import {NavigationService, EventTypes as NavigationEventTypes} from './src/navigation';
...
module.exports = {
  NavigationService,
  NavigationEventTypes,
};
```
Or something?

gulpfile.js
```diff
@@ -25,7 +25,14 @@ gulp.task('ensureDevProxy', function () {
 
 function buildDependenciesArray(dependencies) {
   return Object.keys(dependencies).map(function (dependency) {
-    return dependency + '@' + dependencies[dependency];
+    var version = dependencies[dependency];
+
+    if (version.startsWith('./')) {
```
mlunoe: Let's get a comment here. Something like:
`// Module is local, include dependencies from its package.json`

foundation-ui/index.js
nLight: How about:

```
module.exports = {
  navigation: {...},
  routing: {...}
};
```

```
import {navigation: {EventTypes}} from 'foundation-ui';
```

nLight: Technically those are sub-packages...

mlunoe: Yup, let's do it!

nLight:
@mlunoe @orlandohohmeier @jfurrow please give it another shot. I addressed all the comments and improved the original solution with some guidance from Orlando 🙇

mlunoe:
This looks really good @nLight! I do get this error a lot navigating around:

nLight:
@mlunoe good one! Looks like the change made the problem visible, we shouldn't ideally register a mount point in a constructor, rather using the react life cycle methods.

plugins/services/src/js/components/modals/NewCreateServiceModalServicePicker.js
```diff
@@ -49,27 +49,30 @@ function JSONOption({columnClasses, onOptionSelect}) {
...
 
-    MountService.MountService.registerComponent(
-      MultiContainerOption,
-      'CreateService:ServicePicker:GridOptions',
-      2
-    );
+class NewCreateServiceModalServicePicker extends React.Component {
+  componentDidMount() {
+    OPTIONS.forEach(function (component, index) {
+      MountService.MountService.registerComponent(
```
mlunoe: 👍

mlunoe:
Looking good! Works great!

jfurrow:
It works great! Nice work!

PR #2044, opened 2017-03-24
Remove unnecessary withCredentials from EventSource

mlunoe:
Remove unnecessary withCredentials from EventSource. This should only be used for CORS, not for authentication.

I removed functionality, adjusted some tests and removed tests that explicitly was testing for this.

**Checklist**


orlandohohmeier:
Code looks good to me. Feel free to merge once verified against a cluster.

mlunoe:
Verified on a cluster.

PR #2045, opened 2017-03-24
Move icon to top of alert

leemunroe:
Moves the icon that is part of the alert to the top of the alert so it is not floating arbitrarily in the middle.

Before
image

After
image

### DIFF
--- a/src/styles/components/alert/styles.less
+++ b/src/styles/components/alert/styles.less
@@ -1,7 +1,7 @@
 & when (@alert-enabled) {
 
   .alert {
-    align-items: center;
+    align-items: flex-start;
     border-radius: @alert-border-radius;
     display: flex;
     margin-bottom: @alert-margin-bottom;
--- a/src/js/components/Alert.js
+++ b/src/js/components/Alert.js
@@ -17,7 +17,9 @@ const Alert = ({children, flushBottom, showIcon, type}) => {
     };
 
     icon = (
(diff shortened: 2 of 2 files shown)


jfurrow:
Should we offset the icon by a few pixels? When there's a single line (e.g. Declined Offers warning), the icon isn't perfectly aligned with the neighboring text.

In the following gif I toggle between `center` and `flex-start`

leemunroe:
@jfurrow Good call. What do you think of this solution? http://stackoverflow.com/a/32119917/557828

src/styles/components/alert/styles.less
```diff
@@ -1,7 +1,7 @@
 & when (@alert-enabled) {
 
   .alert {
-    align-items: center;
+    align-items: flex-start;
```
nLight: Are IE11 and Edge cool with that?

jfurrow: This is supported by IE11 and Edge, and we actually use this quite a bit throughout the codebase. So we're good to go!

jfurrow:
Just to follow up, @leemunroe and I discussed this outside of the PR. We solved the alignment problem by wrapping the `svg` in a `div` and moving the `alert-icon` class to the `div`. This works well in this scenario because the `div` will inherit the `line-height` from the flex container, meaning the `svg` will be vertically centered within the inherited `line-height`.

Poltergeist:
LGTM 💯 👍  Let's 🚢  this.

PR #2046, opened 2017-03-24
Check for the proper key containing environment variables

jfurrow:
This PR fixes a bug where environment variables weren't displayed on the pod's configuration screen. We were looking for the `environment` key but the environment variables are defined on the `env` key.

I didn't add a test in this PR because I'm currently writing SI tests for this in a subsequent PR.

**Checklist**


MatApple:
💯

wavesoft:
Ouch, good one 👍  ✨

PR #2048, opened 2017-03-27
fix(DCOS-14146): Fix sorting for health status in a couple of places.

weblancaster:
This PR fixes a variety of health sorting issues included in the Jira ticket (DCOS-14146), where now those places the health sorting weight is based on visibility importance, meaning `Unhealthy, NA, Idle/Warn, Healthy`.

Be aware that some places at first loading the sort is selected to be in another field e.g `name`, that being said you will need to select the health field.

- Dashboard pane
- Node Components tab
- Components page
- Node components
- Services (task detail)

To accomplish the fix a couple of refactoring work was needed to be done around how the higher order functions handle the sorting, In summary I have split/removed the complexity in to smaller functions easier to understand and enabling more precise unit tests.

The work in this PR also refactors the work in #2001 which was/is related.

**Checklist**

### DIFF
--- a/src/js/constants/UnitHealthStatus.js
+++ b/src/js/constants/UnitHealthStatus.js
@@ -1,23 +1,52 @@
+/**
+ * Values for health value defined in the server
+ */
+const SERVER_HEALTHY = 0;
+const SERVER_UNHEALTHY = 1;
+const SERVER_WARN = 2;
+const SERVER_NA = 3;
+
+/**
+ * sortingValue = Order health types by it's label and number value
+ * This will depend on the sorting method/function
+ * suggested use is ascending 0 meaning ((top of the list)) more important
+ * visibility and 3 meaning least important (bottom of the order)
+ */
 const UnitHealthStatus = {
-  HEALTHY: {
+  [SERVER_HEALTHY]: {
     title: 'Healthy',
-    value: 0,
-    classNames: 'text-success'
+    key: 'HEALTHY',
+    classNames: 'text-success',
+    sortingValue: 3,
+    value: SERVER_HEALTHY
   },
-  UNHEALTHY: {
+  [SERVER_UNHEALTHY]: {
     title: 'Unhealthy',
-    value: 1,
-    classNames: 'text-danger'
+    key: 'UNHEALTHY',
+    classNames: 'text-danger',
+    sortingValue: 0,
+    value: SERVER_UNHEALTHY
   },
-  WARN: {
+  [SERVER_WARN]: {
     title: 'Warning',
-    value: 2,
-    classNames: 'text-warning'
+    key: 'WAR',
+    classNames: 'text-warning',
+    sortingValue: 2,
+    value: SERVER_WARN
+  },
+  [SERVER_NA]: {
+    title: 'N/A',
+    key: 'NA',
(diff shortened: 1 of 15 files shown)


MatApple:
Hey @weblancaster - I've added [WIP] to the beginning of the PR title :) That's how we usually indicate to others that this is WIP 👍

weblancaster:
Ok cool @MatApple thanks.

weblancaster:
retest this please

jfurrow:
I think we need to implement a tie breaker for the dashboard's Component Health panel, as I'm seeing the order change arbitrarily. Check out the following gif:

weblancaster:
@jfurrow you are right. I have pushed a new commit fixing that.

screen shot 2017-04-04 at 9 27 15 am

jfurrow:
Nice @weblancaster! I just noticed the same thing is happening in the Components table also (at the `/components` route), so we'll need the tie breaker there too.

weblancaster:
Ops I thought the id was sufficient for the tie breaker. @jfurrow PR updated and good catches.

screen shot 2017-04-04 at 1 17 19 pm

weblancaster:
retest this please

jfurrow:
Hmm, now I'm getting the following error from the plugins :(

weblancaster:
Looks like @jfurrow screenshot isn't related to the PR. We are investigating.

jfurrow:
Great work @weblancaster!

src/js/constants/UnitHealthStatus.js
```diff
@@ -1,23 +1,52 @@
...
+    value: SERVER_WARN
+  },
+  [SERVER_NA]: {
+    title: 'N/A',
+    key: 'NA',
+    classNames: 'text-mute',
+    sortingValue: 1,
+    value: SERVER_NA
   },
   NA: {
```
mlunoe: Shouldn't this also be `[SERVER_NA]`?

src/js/utils/TableUtil.js
```diff
@@ -71,6 +73,73 @@ var TableUtil = {
...
+   * @returns {Number} HealthSorting value
+   */
+  getHealthSortingValue(healthValue) {
+    let match;
+
+    if (typeof healthValue == 'number') {
+      match = UnitHealthStatus[healthValue].sortingValue;
+    }
+
+    if ( typeof healthValue == 'string' ) {
```
mlunoe: #nit let remove the extra spaces inside the `if`-statement
```
if ( typeof healthValue == 'string' ) {
    ^                              ^
```

mlunoe:
This looks and works great! Great job @weblancaster! Thanks for making this change!
Two small comments, I'll wait with merging until they are done

src/js/utils/TableUtil.js
weblancaster: This is addressed

src/js/constants/UnitHealthStatus.js
weblancaster: This is addressed

mlunoe:
Nice work! ⛵️

weblancaster:
thanks for the reviews @mlunoe @jfurrow

PR #2051, opened 2017-03-27
DCOS-13710: Unleash prettier

nLight:
NO MERCY
------

144425-3530-smyhr-a

Review commit by commit ignoring: style(prettier): unleash prettier
-----

~https://github.com/dcos/javascript/pull/2~
~https://github.com/dcos/javascript/pull/1~
~https://github.com/dcos/dcos-ui/pull/2008~


package.json
```diff
@@ -123,26 +124,27 @@
     "webpack-notifier": "1.3.0"
   },
   "scripts": {
-    "check-env": "./scripts/validate-engine-versions",
```
nLight: I sorted them

package.json
```diff
@@ -123,26 +124,27 @@
...
-    "stylelint": "bash -c 'opts=$@ ; set -f; if [[ $# -lt 1 ]]; then opts=\"{src/styles,plugins,foundation-ui}/**/*.less\"; fi ; stylelint ${opts}' 0",
     "lint-plugins": "PLUGINS=$npm_config_externalplugins; npm run eslint -- \"$PLUGINS/**/*.js\" && npm run stylelint -- \"$PLUGINS/**/*.less\"",
+    "lint": "npm run stylelint && npm run eslint",
+    "postbuild-assets": "rm -rf node_modules/compression-webpack-plugin",
+    "prebuild-assets": "./node_modules/.bin/gulp checkDependencies && npm install compression-webpack-plugin@latest",
     "prebuild": "./scripts/validate-tests",
-    "serve": "npm run build-assets && ./node_modules/.bin/gulp serve",
-    "serve-notify": "export NOTIFY=true && npm run serve",
+    "preinstall": "./scripts/pre-install",
+    "prettier": "prettier --write",
```
nLight: This is what I've added

orlandohohmeier:
@nLight Why don't we simply update the whole code base and then enforce the new styling? The approach feels very error prone.

nLight:
@orlandohohmeier I've been thinking about that and since it is still in beta it won't be safe to upgrade the whole think. No one will ever review that PR and the fact is there are bugs in prettier. So having it in small chunks feels safer. After we churn through it for a release or so we could upgrade the rest in one take. Does it make sense?

orlandohohmeier:
@nLight Hm... Changing files in a commit hook simply feels wrong and highly error prone. How about enforcing (linting) prettier for staged files only?

wavesoft:
I would vote with forcing all the changes before commit, yet using some flag to optionally disable it (since it's in beta).
For example `touch .prettier.disable` if you want to disable this feature (in which case you enforce the linting).

nLight:
@orlandohohmeier they will be changed true, but not staged. So that you can commit logic and style changes separately while linting will prevent you from committing, forcing us to have a mess of logic changes melt with style changes.

@wavesoft  so that everyone just disables it.

jfurrow:
This works for me! :ship:

scripts/pre-commit
```diff
@@ -26,6 +26,13 @@ then
   info "Staged Less looks good"
 fi
 
+# Prettiefy staged JavaScript files
+if [ -n "${staged_javascript_files}" ]
+then
+  header "Prettiefy staged JavaScript files.."
```
jfurrow: #nit I think it should say `Prettify` instead of `Prettiefy`

orlandohohmeier:
@nLight I got that but still don't think that any of the git hooks should ever change files automatically! I'd add prettier as a linter (for staged files only) that warns instead of errors out. This would allow us to commit/push logic changes without mixing them with stylistic changes while increasing awareness of the new stylistic rules and encouraging people to fix the styling. What do you think?

mlunoe:
Prettier 1.0 is OUT! http://jlongster.com/prettier-1.0

mlunoe:
> "What it means is that prettier is safe to use for production. We've fixed a ton of bugs and made sure the output is stable and semantically correct. Prettier is ready to be used from a few files in your side project to converting your entire codebase. It's up to you to figure out where you are in the journey of letting go of a particular way to formatting your code."

mlunoe:
Rebased this, since it had conflicts

nLight:
Hey @orlandohohmeier, it's one swipe! \o/

nLight:
...and it's 💚 \o/

nLight:
Sorry @jfurrow I'm removing your label because it changed a lot

mlunoe:
Should we use single quotes?

nLight:
@mlunoe I tried to make it work with single quotes on our codebase and it was tough. When prettier encounters a single quote inside a string it would use double quotes producing inconsistency. The reason for that is that it would always try to strip an escaping slash from a single quote inside a string

```javascript
it('doesn\'t do anything') // <--- that ugliness
```
and we have a lot of those in tests.

In conclusion I think we should embrace double quotes and stick to the very default settings of prettier.

Everyone is free to continue using single quotes it will be prettified :)

package.json
```diff
@@ -78,13 +78,14 @@
     "babel-runtime": "6.9.2",
     "colors": "1.1.2",
     "css-loader": "0.23.1",
-    "eslint": "3.14.1",
-    "eslint-config-airbnb": "10.0.1",
-    "eslint-plugin-compat": "1.0.1",
+    "eslint": "3.15.0",
+    "eslint-config-dcos": "github:dcos/javascript#drozhkov/prettier-1.0",
```
mlunoe: I guess the plan is to update this when the javascript PR has merged?

mlunoe:
Hmmm... I like using the default settings. On the other hand, I like the difference between an attribute string and a variable string.

At this point I could probably make the case for both. I just remember us having discussed this _a lot_ when we moved from double back to single quotes, so I will expect a reaction from others.

nLight:
I know, Orlando already gave me a head up on the single/double quotes situation. Yet I believe we should be less married to this kind of things and be leaner.

jfurrow:
Although I also would prefer single quotes, consistency is more important, so I'm adding my label. Thanks for doing all the leg work @nLight!

mlunoe:
I am happy to go with this too, but we should probably notify the team @dcos/frontend before this gets merged.

mlunoe:
It has konfliktz now, though

nLight:
run integration tests

orlandohohmeier:
There seem to be an issue with the prettier dependency... 

> Dependency length doesn't match.
> [ 'prettier@^1.2.2' ]

rcorral:
I'm good with the change to double quotes.
Always a fan of single quotes but the whole problem with apostrophes makes it annoying.

mlunoe:
💚 ! Let's revert! https://github.com/dcos/dcos-ui/pull/2133

PR #2053, opened 2017-03-28
DCOS-14842: Change size label from MiB to GiB for external vols

nLight:
This PR changes label from MiB to GiB for external volumes. Please not just review it, test it 🙇 

Marathon treats size value for External volumes as it is in GiB and sends this data to Mesos.
This is going to be fixed (changed to MiB) in 1.10 to align it with other volumes. 

But in the 1.9 we want to show the right label.

https://jira.mesosphere.com/browse/DCOS-14842

### DIFF
--- a/plugins/services/src/js/containers/volume-detail/VolumeDetail.js
+++ b/plugins/services/src/js/containers/volume-detail/VolumeDetail.js
@@ -15,6 +15,15 @@ import ServiceBreadcrumbs from '../../components/ServiceBreadcrumbs';
 import VolumeStatus from '../../constants/VolumeStatus';
 
 class VolumeDetail extends React.Component {
+
+  getSizeLabel() {
+    if (this.props.volume.type === 'External') {
+      return 'Size (GiB)';
+    }
+
+    return 'Size (MiB)';
+  }
+
   renderSubHeader() {
     const {volume} = this.props;
 
@@ -81,7 +90,7 @@ class VolumeDetail extends React.Component {
             </ConfigurationMapRow>
             <ConfigurationMapRow>
               <ConfigurationMapLabel>
-                Size (MiB)
+                {this.getSizeLabel()}
               </ConfigurationMapLabel>
               <ConfigurationMapValue>
                 {volume.getSize()}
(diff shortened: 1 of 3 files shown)


nLight:
Found one more place

plugins/services/src/js/containers/volume-detail/VolumeDetail.js
```diff
@@ -54,6 +54,10 @@ class VolumeDetail extends React.Component {
       />
     );
 
+    const sizeLabel = volume.type === 'External'
```
orlandohohmeier: Please use a proper condition or private util to get the respective size label instead of a ternary to improve readability.

orlandohohmeier:
Let's add a tooltips to this list to clarify the difference.

nLight:
@orlandohohmeier I'd extract the table tooltip as an improvement jira

plugins/services/src/js/containers/volume-detail/VolumeDetail.js
nLight: Sorry, I'm not convinced that
```
let sizeLabel = 'Size (MiB)';

if (volume.type === 'External') {
  sizeLabel = 'Size (GiB)';
}
```
is more readable

nLight: changed to a function

MatApple: I agree with @nLight - a ternary in this case is the easiest to read and cleanest IMO.

orlandohohmeier: Consider the following example which would introduce a server bug that is rather hard to spot during a simple code review. 

```javascript
var foo = bar = 'baz' 
  ? 'qux' : 'quux'
```

Let's please not use ternaries until we have a solid set of eslint rules to cover the bad parts.

nLight: Airbnb/javascript gives a warn here
screen shot 2017-03-28 at 14 36 05

We've agreed on ternaries and we have eslint rules in place

nLight: https://github.com/dcos/dcos-ui/blob/master/.eslintrc#L77
https://github.com/dcos/dcos-ui/blob/master/.eslintrc#L59
https://github.com/dcos/dcos-ui/blob/master/.eslintrc#L73

orlandohohmeier:
Great Job! ⚒ 💯

PR #1957, opened 2017-02-28
Adding missing property

mlunoe:
A property was missing to be added to the `ServiceConfigDisplay`. This PR amends that.

Warning from console:
```
Warning: Failed propType: Required prop `onEditClick` was not specified in `ServiceConfigDisplay`. Check the render method of `ServiceConfiguration`.
warning @ warning.js?8a56:45
checkPropTypes @ ReactElementValidator.js?a599:189
validatePropTypes @ ReactElementValidator.js?a599:208
createElement @ ReactElementValidator.js?a599:242
render @ ServiceConfiguration.js?dc74:283
(anonymous) @ makeAssimilatePrototype.js?c7d8:15
_renderValidatedComponentWithoutOwnerOrContext @ ReactCompositeComponent.js?cd59:587
_renderValidatedComponent @ ReactCompositeComponent.js?cd59:607
ReactCompositeComponent__renderValidatedComponent @ ReactPerf.js?ef93:66
_updateRenderedComponent @ ReactCompositeComponent.js?cd59:560
_performComponentUpdate @ ReactCompositeComponent.js?cd59:544
updateComponent @ ReactCompositeComponent.js?cd59:473
ReactCompositeComponent_updateComponent @ ReactPerf.js?ef93:66
performUpdateIfNecessary @ ReactCompositeComponent.js?cd59:421
performUpdateIfNecessary @ ReactReconciler.js?6bfa:102
runBatchedUpdates @ ReactUpdates.js?ce09:129
perform @ Transaction.js?6dff:136
perform @ Transaction.js?6dff:136
perform @ ReactUpdates.js?ce09:86
flushBatchedUpdates @ ReactUpdates.js?ce09:147
ReactUpdates_flushBatchedUpdates @ ReactPerf.js?ef93:66
closeAll @ Transaction.js?6dff:202
perform @ Transaction.js?6dff:149
batchedUpdates @ ReactDefaultBatchingStrategy.js?ef70:62
enqueueUpdate @ ReactUpdates.js?ce09:176
enqueueUpdate @ ReactUpdateQueue.js?fd2c:24
enqueueForceUpdate @ ReactUpdateQueue.js?fd2c:143
ReactComponent.forceUpdate @ ReactComponent.js?702a:86
store_onStoreChange @ StoreMixin.js?a69d:181
EventEmitter.emit @ events.js?7c71:92
```

**Checklist**
- [ ] Did you add a JIRA issue in a commit message or as part of the branch name?
- [ ] Did you add new unit tests?
[...]

PR #2058, opened 2017-03-29
[Route Service Tabs 2 of 2] DCOS-14508: Make service detail tabs routed

mlunoe:
This PR makes the service detail tabs routed. A bug emerged while implementing this that was introduced here: https://github.com/dcos/dcos-ui/pull/1957. This PR also takes care of that.

~~Route Service Tabs 1 of 2: #2079~~ Merged
Route Service Tabs 2 of 2: https://github.com/dcos/dcos-ui/pull/2058

~~Diff between 1 and 2: https://github.com/dcos/dcos-ui/compare/mlunoe/DCOS-14508-rearrange-service-routes-to-divide-overview-and-detail...mlunoe/feature/DCOS-14508-make-service-detail-tabs-routed~~ PR 1 has merged!

**Checklist**


plugins/services/src/js/containers/service-detail/ServiceDetail.js
```diff
@@ -182,59 +154,78 @@ class ServiceDetail extends mixin(TabsMixin) {
...
 
     tabs.push({
       label: 'Instances',
       callback: () => {
-        this.setState({currentTab: 'tasks'});
+        this.context.router.push(`${routePrefix}/tasks`);
       },
-      isActive: activeTab === 'tasks'
+      isActive: currentTab === `${routeDefinitionPrefix}/tasks`
     });
```
jfurrow: Since all of these tabs are routed, you don't need to supply a `callback` method to handle the route changes. Instead you can just pass an array of objects that have the properties `label` and `routePath`. From there, the tabs will use the `Link` component for each tab, which gives us the active indicator for free, so we can remove the `isActive` property and the associated `currentTab` logic.

Check out an example in `NodesTaskDetailPage`: https://github.com/dcos/dcos-ui/blob/master/plugins/nodes/src/js/pages/nodes/NodesTaskDetailPage.js#L36

plugins/services/src/js/containers/service-detail/ServiceDetail.js
```diff
@@ -182,59 +154,78 @@ class ServiceDetail extends mixin(TabsMixin) {
...
+      volumes: service.getVolumes().getItems()
+    };
+
+    const clonedChildren = React.Children.map(children, function (child) {
+      // Only add props to children that are not ServiceModals
+      if (child != null && child.type === ServiceModals) {
+        return child;
+      }
+
+      return React.cloneElement(child, clonedProps);
```
jfurrow: Why do we need to clone the children now?

mlunoe: We need to pass down the props that was passed down directly in the `renderxxxTab` functions

plugins/services/src/js/containers/service-detail/ServiceDetail.js
mlunoe: Thanks! Pushed the change!

mlunoe:
The integration tests I created are failing for some reason. Need to look into it

plugins/services/src/js/containers/service-detail/ServiceDetail.js
jfurrow: Makes sense, thanks!

jfurrow:
I'm getting an error when I try to navigate to a grouping of services via the collection view:

mlunoe:
Good catch! I'll look into it!

mlunoe:
Ok, this was a really good find. As it seems there needs to be some bulk work in place to release the newly added tab children from the ServiceTreeView (which does not need them).
I am going to create another base PR, where I will separate the ServiceDetail/PodDetail and ServiceTreeView to have separate containers.

mlunoe:
Alright, created a base branch that splits up the `/overview` and `/detail` routes, which handles the issue that you were seeing @jfurrow

plugins/services/src/js/containers/service-detail/ServiceDetail.js
```diff
@@ -181,60 +153,54 @@ class ServiceDetail extends mixin(TabsMixin) {
...
+      volumes: service.getVolumes().getItems()
+    };
+
+    const clonedChildren = React.Children.map(children, function (child) {
+      // Only add props to children that are not ServiceModals
+      if (child.type === ServiceModals) {
+        return child;
+      }
+
+      return React.cloneElement(child, clonedProps);
```
weblancaster: Why are us having to clone the element and use it later this way?

plugins/services/src/js/containers/services/ServicesContainer.js
```diff
@@ -474,34 +475,21 @@ class ServicesContainer extends React.Component {
...
-          service={item}>
-          {this.getModals(item)}
-          {children}
-        </ServiceDetail>
-      );
-    }
-
     // Show Tree
-    if (item instanceof ServiceTree) {
+    const currentroutePath = reconstructPathFromRoutes(routes);
```
weblancaster: shouldn't it be `currentRoutePath `?

plugins/services/src/js/containers/services/ServicesContainer.js
```diff
@@ -474,34 +475,21 @@ class ServicesContainer extends React.Component {
...
+    if (currentroutePath.startsWith('/services/overview')) {
+
+      if (!(item instanceof ServiceTree)) {
+        // Not found
+        return (
```
weblancaster: Would be nice to start extracting this conditions block from the render and putting in a method function to make more clear (helps with unit test too) and possibly even re-use in case it's necessary.

e.g `notFoundComponent` inside you would have the render, if possible without passing any argument but accessing in props (only if possible).

```
// Not found
        return (
          <Page>
            <Page.Header breadcrumbs={<ServiceBreadcrumbs />} />
            <ServiceItemNotFound
              message={`The group with the ID of "${itemId}" could not be found.`} />
          </Page>
        );
```

plugins/services/src/js/containers/services/ServicesContainer.js
```diff
@@ -522,12 +510,38 @@ class ServicesContainer extends React.Component {
...
           serviceTree={item}>
-          {this.getModals(item)}
           {children}
+          {this.getModals(item)}
         </ServiceTreeView>
       );
     }
 
+    if (item instanceof Pod) {
+      return (
```
weblancaster: basically same comment here, extracting the condition block to a method will help a bunch.

plugins/services/src/js/containers/services/ServicesTable.js
```diff
@@ -95,6 +95,11 @@ class ServicesTable extends React.Component {
 
   getServiceLink(service) {
     const id = encodeURIComponent(service.getId());
+    const isGroup = service instanceof ServiceTree;
+    let urlPrefix = '/services/detail/';
+    if (isGroup) {
+      urlPrefix = '/services/overview/';
+    }
```
weblancaster: shorthand?

```
const isGroup = service instanceof ServiceTree;
const urlPrefix = (isGroup) ? '/services/overview/' : '/services/detail/';
```

plugins/services/src/js/containers/services/ServicesTable.js
```diff
@@ -109,7 +114,7 @@ class ServicesTable extends React.Component {
     return (
       <Link
         className="table-cell-link-primary text-overflow"
-        to={`/services/overview/${id}`}>
+        to={urlPrefix + id}>
```
weblancaster: I believe storing in a variable will help debug. So adding right below `const urlPrefix` something like `const serviceLinkId = urlPrefix + id`

then in the return

```
<Link
        className="table-cell-link-primary text-overflow"
        to={serviceLinkId}>
        {service.getName()}
      </Link>
```

plugins/services/src/js/containers/services/ServicesTable.js
```diff
@@ -137,13 +142,18 @@ class ServicesTable extends React.Component {
 
   renderHeadline(prop, service) {
     const id = encodeURIComponent(service.getId());
+    const isGroup = service instanceof ServiceTree;
+    let urlPrefix = '/services/detail/';
+    if (isGroup) {
+      urlPrefix = '/services/overivew/';
+    }
```
weblancaster: same comment as above

src/js/utils/NavigationServiceUtil.js
```diff
@@ -40,13 +40,19 @@ const NavigationServiceUtil = {
           route.component.routeConfig.label,
           { category, icon }
         );
-
         childRoutes
           .filter(({isInSidebar}) => isInSidebar)
           .forEach((childRoute) => {
+            // TODO: Temporary fix to allow routes to register under a different
```
weblancaster: what is the effort to not making this only temporary?

tests/pages/services/ServiceDetail-cy.js
```diff
@@ -0,0 +1,74 @@
...
+
+    it('shows volumes tab when clicked', function () {
+      cy.visitUrl({url: '/services/detail/%2Fsleep'});
+
+      cy.get('.menu-tabbed-item').contains('Volumes').click();
+
+      cy.get('.menu-tabbed-item .active').contains('Volumes');
+
+      cy.get('.table')
+        .contains('tr', 'volume-1')
```
weblancaster: would be great to having the rest as this chaining in each line, not big deal though.

weblancaster:
I'm seeing the same issue as the other PR, not sure if it's expected as they are in separated PRs. so please let me know.

screen recording 2017-04-10 at 09 17 am

plugins/services/src/js/containers/service-detail/ServiceDetail.js
mlunoe: Because we need to pass down extra props to the elements. Props that was being passed down in these render functions that we turned into routes: https://github.com/dcos/dcos-ui/pull/2058/files#diff-fc8d30a093163d01bb787f96fdf40542L94

src/js/utils/NavigationServiceUtil.js
mlunoe: Because this function has been deprecated. When we make the effort to remove this function (and start registering routes using the `NavigationService` directly) we should also remove the added sidebarPath that was added to the services routes in this PR. This `TODO` is just to remind whoever will take this on to also do that

tests/pages/services/ServiceDetail-cy.js
mlunoe: like
```js
cy
  .get('.table')
  .contains('tr', 'volume-1')
  ...
```
?

weblancaster: yep. every call in a new line. 

Like I said though not big deal.

src/js/utils/NavigationServiceUtil.js
weblancaster: got it, thank you for the clarification.

plugins/services/src/js/containers/service-detail/ServiceDetail.js
weblancaster: got it

tests/pages/services/ServiceDetail-cy.js
mlunoe: Done

mlunoe:
@jfurrow and @weblancaster comments addressed!

jfurrow:
I'm still seeing the error... am I doing something incorrectly here?

weblancaster:
Looks good but the https://github.com/dcos/dcos-ui/pull/2079 still need some updates you have made here.

mlunoe:
Change backported to depending PR. Depending PR merged and this have now been rebased off of it

weblancaster:
Tested with the rebase and works fine.

jfurrow:
Yay!!!

PR #2059, opened 2017-03-29
[System Tests]: DCOS-12813: Add SI tests for pod with service address

jfurrow:
This PR introduces the automated tests for [Create a simple pod with service address](https://wiki.mesosphere.com/pages/viewpage.action?pageId=99615053)

To run this test, use:

<pre>
  -e CLUSTER_URL=$CLUSTER_URL  \
  -v `pwd`:/dcos-ui \
  ./system-tests/driver-config/cluster.sh
</pre>

### DIFF
--- a/system-tests/services/test-pods.js
+++ b/system-tests/services/test-pods.js
@@ -250,6 +250,246 @@ describe('Services', function () {
 
     });
 
+    it('Create a pod with service address', function () {
+      const serviceName = 'pod-with-service-address';
+      const command = 'python3 -m http.server 8080';
+      const containerImage = 'python:3';
+
+      cy
+        .contains('Multi-container (Pod)')
+        .click();
+
+      cy
+        .root()
+        .getFormGroupInputFor('Service ID *')
+        .type(`{selectall}{rightarrow}${serviceName}`);
+
+      cy
+        .get('.menu-tabbed-item')
+        .contains('container-1')
+        .click();
+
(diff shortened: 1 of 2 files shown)


system-tests/services/test-pods.js
```diff
@@ -0,0 +1,256 @@
...
+      const serviceName = 'pod-with-service-address';
+      const command = 'python3 -m http.server 8080';
+      const containerImage = 'python:3';
+
+      cy
+        .contains('Multi-container (Pod)')
+        .click();
+
+      cy
+        .root()
```
rcorral: I don't think this `.root` call is needed here or anywhere on this file. It's only needed if you need to reset the context on a chain of commands, but if you're already starting with `cy` it already starts with root.

jfurrow: Actually `.root()` is required for the command `getFormGroupInputFor`, otherwise its first argument (`elements`) will be `undefined`. One alternative might be to adjust `getFormGroupInputFor` to default to `document` if `elements` is `undefined`?

jfurrow:
@rcorral I'm okay with holding on this if you'd like, but since we already have tests that look for elements based on their labels, it would make sense to me to handle the i18 in tests as a separate task. But I'll be happy with whatever you think is best!

mlunoe:
I would rather get this in and not hold up the PR. This adds value now. When we at internationalisation we will have to update these tests, but that should not be too big of a burden.

rcorral:
Merged :)

PR #2064, opened 2017-03-31
DCOS-12861: Enable force Pull Image setting in UCR

Poltergeist:
**Checklist**

image

close DCOS-12861


nLight:
Had to rebase to make it working with the plugins, but it works now 👍

plugins/services/src/js/components/forms/ContainerServiceFormAdvancedSection.js
```diff
@@ -23,11 +23,6 @@ const containerSettings = {
     label: 'Grant Runtime Privileges',
     helpText: 'By default, containers are “unprivileged” and cannot, for example, run a Docker daemon inside a Docker container.',
     dockerOnly: 'Grant runtime privileges is only supported in Docker Runtime.'
-  },
-  forcePullImage: {
```
mlunoe: Since we now only have `privileged` in the settings couldn't we just add that with its view declaration instead of doing it dynamically?

nLight:
@mlunoe fair point! I addressed the issue, please give it another try 🙇

mlunoe:
Works great!

PR #2069, opened 2017-04-03
fix(DeleteRowButton): fix alignment of DeleteRowButton instances

jfurrow:
This PR fixes the lack of label offset for `DeleteRowButton` and removes a comment I accidentally committed.

Before:

After:


Poltergeist:
LGTM 👍  💯

PR #2070, opened 2017-04-04
test(DCOS-14460): Implement integration tests for Service modal.

weblancaster:
This PR implements integration tests for Service modal.
**Jira:** DCOS-14460

**Tests:**
- Modal is full screen
- Horizontally/Vertically centered
- Panels contain same size (width/height).

**Checklist**

### DIFF
--- a/tests/pages/services/ServiceFormModal-cy.js
+++ b/tests/pages/services/ServiceFormModal-cy.js
@@ -273,7 +273,7 @@ describe('Service Form Modal', function () {
         });
     });
 
-    it.only('should be horizontally and vertically centered in the modal container', function () {
+    it('should be horizontally and vertically centered in the modal container', function () {
       cy.get('.modal-body-wrapper')
         .should(function ($modalWrapper) {
           const modalWrapperRect = $modalWrapper[0].getBoundingClientRect();
@@ -309,7 +309,7 @@ describe('Service Form Modal', function () {
         });
     });
 
-    it('should panes be the same width and height', function () {
+    it('should contain panes with the same width and height', function () {
       var isPanesSameSize = true;
 
       cy.get('.create-service-modal-service-picker-option')
--- a/tests/pages/services/ServiceFormModal-cy.js
+++ b/tests/pages/services/ServiceFormModal-cy.js
@@ -258,6 +258,41 @@ describe('Service Form Modal', function () {
       cy.visitUrl({url: '/services/overview/%2F/create'});
     });
 
+    it('should fill the entire viewport', function () {
+      var isModalFullScreen = true;
+
+      cy.window()
+        .then(function ($window) {
+          cy.get('.modal-full-screen')
+            .should(function ($element) {
+              if ($element[0].clientHeight !== $window.innerHeight || $element[0].clientWidth !== $window.innerWidth) {
+                isModalFullScreen = false;
+              }
+              expect(isModalFullScreen).to.be.equal(true);
+            });
+        });
+    });
+
+    it.only('should be horizontally and vertically centered in the modal container', function () {
+      cy.get('.modal-body-wrapper')
+        .should(function ($modalWrapper) {
+          const modalWrapperRect = $modalWrapper[0].getBoundingClientRect();
+          const modalWrapperTop = Math.floor(modalWrapperRect.top + (modalWrapperRect.height / 2));
+          const modalWrapperLeft = Math.floor(modalWrapperRect.left + (modalWrapperRect.width / 2));
+
+          const $modalContent = $modalWrapper.find('.row.panel-grid');
+          const modalContentRect = $modalContent[0].getBoundingClientRect();
+          const modalContentTop = Math.floor(modalContentRect.top + (modalContentRect.height / 2));
+          const modalContentLeft = Math.floor(modalContentRect.left + (modalContentRect.left / 2));
+
+          const topPosDifference = Math.floor(modalWrapperTop - modalContentTop);
+          const leftPosDifference = Math.floor(modalWrapperLeft - modalContentLeft);
+
+          expect(topPosDifference <= modalContentTop).to.be.equal(true);
+          expect(leftPosDifference <= modalContentTop).to.be.equal(true);
+        });
+    });
+
     it('should have four options to choose from', function () {
       cy.get('.panel-grid h5')
         .should(function (items) {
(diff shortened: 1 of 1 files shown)


Poltergeist:
LGTM 💯  👍

tests/pages/services/ServiceFormModal-cy.js
```diff
@@ -258,6 +258,41 @@ describe('Service Form Modal', function () {
...
+          const $modalContent = $modalWrapper.find('.row.panel-grid');
+          const modalContentRect = $modalContent[0].getBoundingClientRect();
+          const modalContentTop = Math.floor(modalContentRect.top + (modalContentRect.height / 2));
+          const modalContentLeft = Math.floor(modalContentRect.left + (modalContentRect.left / 2));
+
+          const topPosDifference = Math.floor(modalWrapperTop - modalContentTop);
+          const leftPosDifference = Math.floor(modalWrapperLeft - modalContentLeft);
+
+          expect(topPosDifference <= modalContentTop).to.be.equal(true);
+          expect(leftPosDifference <= modalContentTop).to.be.equal(true);
```
jfurrow: I think we should compare these values to something other than `modalContentTop`. This value should represent a reasonable threshold that the delta of the midpoints can be within to still be considered "centered."

tests/pages/services/ServiceFormModal-cy.js
```diff
@@ -258,6 +258,41 @@ describe('Service Form Modal', function () {
...
+            });
+        });
+    });
+
+    it('should be horizontally and vertically centered in the modal container', function () {
+      cy.get('.modal-body-wrapper')
+        .should(function ($modalWrapper) {
+          const modalWrapperRect = $modalWrapper[0].getBoundingClientRect();
+          const modalWrapperTop = Math.floor(modalWrapperRect.top + (modalWrapperRect.height / 2));
+          const modalWrapperLeft = Math.floor(modalWrapperRect.left + (modalWrapperRect.width / 2));
```
jfurrow: I think the variables `modalWrapperLeft` and `modalWrapperTop` actually represent the x,y coordinates of the midpoint of these elements, so what do you think about naming the variables something like `modalWrapperMidpointX` and `modalWrapperMidpointY`?

tests/pages/services/ServiceFormModal-cy.js
```diff
@@ -258,6 +258,41 @@ describe('Service Form Modal', function () {
...
+
+          const $modalContent = $modalWrapper.find('.row.panel-grid');
+          const modalContentRect = $modalContent[0].getBoundingClientRect();
+          const modalContentTop = Math.floor(modalContentRect.top + (modalContentRect.height / 2));
+          const modalContentLeft = Math.floor(modalContentRect.left + (modalContentRect.left / 2));
```
jfurrow: Same here, let's rename these variables to reflect the fact that we're finding the mid-point of these elements.

tests/pages/services/ServiceFormModal-cy.js
```diff
@@ -258,6 +258,41 @@ describe('Service Form Modal', function () {
...
+
+          const topPosDifference = Math.floor(modalWrapperTop - modalContentTop);
+          const leftPosDifference = Math.floor(modalWrapperLeft - modalContentLeft);
```
jfurrow: I think we'll want to use `Math.abs` instead of `Math.floor` here.

tests/pages/services/ServiceFormModal-cy.js
weblancaster: Good point. Addressed.

tests/pages/services/ServiceFormModal-cy.js
weblancaster: Addressed.

tests/pages/services/ServiceFormModal-cy.js
weblancaster: Addressed.

tests/pages/services/ServiceFormModal-cy.js
weblancaster: what you think about the container midpoint? @jfurrow

jfurrow: I think a threshold of `5` would be sufficient for this.

I just checked again, and I think maybe we shouldn't check the midpoint of `.row.panel-grid` because, because it has a negative margin. We could account for the negative margin, but that makes the test a little more complicated to understand. Perhaps the element `.create-service-modal-service-picker-options` would be good for the vertical midpoint; I just did this manually and the midpoints of the two elements is the exact same.

weblancaster: Ok sounds good.

tests/pages/services/ServiceFormModal-cy.js
```diff
@@ -258,6 +258,41 @@ describe('Service Form Modal', function () {
...
+          const $modalContent = $modalWrapper.find('.row.panel-grid');
+          const modalContentRect = $modalContent[0].getBoundingClientRect();
+          const modalContentMidpointY = Math.abs(modalContentRect.top + (modalContentRect.height / 2));
+          const modalContentMidpointX = Math.abs(modalContentRect.left + (modalContentRect.left / 2));
+
+          const topPosDifference = Math.abs(modalWrapperMidpointY - modalContentMidpointY);
+          const leftPosDifference = Math.abs(modalWrapperMidpointX - modalContentMidpointX);
+
+          expect(topPosDifference <= modalContentMidpointY).to.be.equal(true);
+          expect(leftPosDifference <= modalContentMidpointX).to.be.equal(true);
```
jfurrow: I think we are still comparing the wrong values here. In an ideal world, `topPosDifference` and `leftPosDifference` would be `0`. But it's acceptable for the mid-points to be within ~5 pixels of one another. So instead of comparing `topPosDifference <= modalContentMidpointY` and `leftPosDifference <= modalContentMidpointX` we should be comparing `leftPosDifference <= 5` and `leftPosDifference <= 5`. The value `5` represents the alignment threshold that we consider acceptable.

I defined an `alignmentTheshold` variable elsewhere in this file (for the test `Should vertically align the placement constraint delete row button`), so we could move that variable's declaration and reuse it here.

jfurrow:
Sorry @weblancaster one more request! Otherwise this looks great, nice work! 🥇

tests/pages/services/ServiceFormModal-cy.js
weblancaster: No problem. And makes all sense.

Thank you for the review.

weblancaster: it's addressed @jfurrow

weblancaster:
np @jfurrow should be addressed, let me know if there is something else. thanks.

jfurrow:
Thanks @weblancaster this is great!

weblancaster:
✌️

PR #2073, opened 2017-04-05
DCOS-14844: Fix local/external volumes mixup

nLight:
This PR addresses the diffing of arrays of objects bug in the `applyPatch`

To reproduce the bug try create a Single container app:

- Add 1 local volume
- Add 1 external volume
- ! Change something in the JSON, this is important to update the base config
- Try adding one more local volume
- Notice that it gained the `external` field in the JSON

Then pull the branch and validate that the problem is gone.

On top of that, since we have a battle tested diffing algo we don't need deleting `KEY_VALUE_FIELDS` anymore, `jsondiffpatch` handles that as well. The issue was that an attempt of editing an env var in the json would duplicate the var.

Less magic is better. Less tests in the test suite is better as well.

🙇


nLight:
run integration tests

plugins/services/src/js/utils/CreateServiceModalFormUtil.js
```diff
@@ -31,168 +31,6 @@ const CreateServiceModalFormUtil = {
 
       return memo;
     }, baseObject);
-  },
```
weblancaster: I like we are using `jsondiffpatch` and removing the code below :) 

_(ps: have you looked this one https://github.com/kpdecker/jsdiff)_

PR #2074, opened 2017-04-05
feat(DCOS-14986): i18n poc implementation.

weblancaster:
This PR is a poc implementation of i18n where two libs were explored (React-intl and i18next).

React-intl was chosen for the following reasons, support all acceptance criteria defined in `DCOS-14986`,
has been proven internally and externally at Yahoo, community support, follows React-like pattern where components are used instead imperative way (calling functions),
Good documentation/API/source code and lastly follows the ECMA standards for i18n. Unfortunately both contain only one main contributor which ins't ideal but since Yahoo has a better open source history (e.g YUI) make it "safer" to use.

We also have other team members (@jfurrow) familiar with React-intl which makes it faster to adopt.

This PR also include the documentation guide to get started on how to add new translations (files or key/values) and a couple of other points where to find more information about the React-Intl.

Steps to test:
`npm i react-intl` then `npm start` when the build is ready you can check the `dashboard` page `panels heading` and `view all` buttons

**before**
before

**after**
after

**Checklist**


weblancaster:
retest this please

src/js/translations/en-US.json
```diff
@@ -0,0 +1,10 @@
+{
+  "DASHBOARD.VIEW_ALL": "View all {servicesCount} Services",
```
MatApple: Let's think about how we could handle the pluralization. i.e. 1 Service or X Services

weblancaster: @MatApple the lib also support plural but in this case it's Services one or many based on the UI and the code we had in place so I only "replicated" using the localization.

That being said though I have seen two ways to do that so far. One is to use the
[string formatting](https://github.com/yahoo/react-intl/wiki/Components#string-formatting-components) and the other is by composing `FormattedMessage` component with `FormattedPlural` where we would add another key/value in value object.

package.json
```diff
@@ -54,6 +54,7 @@
     "react-addons-test-utils": "0.14.7",
     "react-dom": "0.14.7",
     "react-gemini-scrollbar": "2.1.0",
+    "react-intl": "2.2.3",
```
rcorral: When adding dependencies, please run `npm run shrinkwrap`

weblancaster: @rcorral it's addressed.

src/js/translations/en-US.json
```diff
@@ -0,0 +1,10 @@
+{
```
rcorral: @weblancaster can you amend our Contributing guidelines to add something around using the `FormattedMessage` component for any piece of copy?

Please include the naming scheme for IDs.

weblancaster: @rcorral not sure if there is anything I can add because the [documentation](https://github.com/yahoo/react-intl/wiki/Components#string-formatting-components) in the React-intl repo is pretty clear.

Now about the IDs scheme I was going to start another PR for it since this one is only the code for the first experimental work.

weblancaster: When this is merged I'm going to add a PR suggestion to`CONTRIBUTING` where will explain how to add i18n IDs.

rcorral: Let's add it all in this PR. I think it belongs together.
If you want to do it in a different PR then I think we should have it in a feature branch. Once we all agree on it we can merge everything into master.

Feel free to point to the React-intl repo as it's not going to be obvious for people.

weblancaster: Ok cool, so I will do in this one.

weblancaster: It's addressed @rcorral

jfurrow:
Nice work @weblancaster!

nLight:
We're off to a good start here! LGTM 💯 ✨ 
Sorry it took me too long to look into it.

PR #2080, opened 2017-04-07
[System Tests]: Re-trying teardown

wavesoft:
Retry 3 times when tearing down services group, since some times
the operation with the cluster times out.

**Checklist**

### DIFF
--- a/system-tests/services/_scripts/teardown
+++ b/system-tests/services/_scripts/teardown
@@ -1,4 +1,11 @@
 #!/bin/bash
 
-# Remove the service group for the test
-dcos marathon group remove --force /$TEST_UUID
+# Retry 3 times to remove the service (for an unknown reason this operation
+# times out some times)
+for I in {1..3}; do
+  dcos marathon group remove --force /$TEST_UUID
+  [ $? -eq 0 ] && exit 0
+done
+
+# Fail
+exit 1


Poltergeist:
I don't like the approach, but I guess there is nothing we can do about it right?

mlunoe:
Can we get some confidence how robust this is?

mlunoe:
Like run it 10 times in a row and see if it fails at any point in time to see whether we need to increase the number?
And I agree this is not ideal. Is there a reason why this command fails?

wavesoft:
Actually i couldn't pinpoint the reason 😞 I just get a "response timeout" from the cluster.
Honestly, I couldn't reproduce this myself. I think @orlandohohmeier and @Poltergeist saw it quite some times.

mlunoe:
Sad panda... Hmm... Could you investigate some time @Poltergeist?

Poltergeist:
I did not had the problem of tear down, I had the problem with docker. And latency to the cluster. But not with tearing down the group. @wavesoft @mlunoe

mlunoe:
🤔 Ok, @wavesoft or @orlandohohmeier could elaborate a bit more on the error?

wavesoft:
To be honest I never saw this error occurring again. It seemed like a docker issue because when somebody restarts the docker agent this goes away...

wavesoft:
Actually, I am going to close this issue until somebody complains about it again...

PR #2082, opened 2017-04-07
test(DCOS-14457): Integration test for header components

weblancaster:
This PR (DCOS-14457) adds integration test for header components where breadcrumb, header tab are included.

AC:
* When there are more than 3 breadcrumb items:
-Breadcrumbs should collapse so that only the last two breadcrumb items are visible
-An ellipsis should take place of all breadcrumbs except the last two
-Hovering the ellipsis should render a tooltip that displays the path that is hidden
-The Services icon should route back to the services overview

- When the viewport is narrow, the breadcrumbs should truncate and not overflow their container
- When the viewport is narrow, the page header actions should always be visible
- When a page header's tab is active, its activity indicator should be flush with the bottom of the page header.

**Checklist**

### DIFF
--- a/tests/components/Header-cy.js
+++ b/tests/components/Header-cy.js
@@ -55,7 +55,6 @@ describe('Header Component', function () {
         if ( firstItem.classList.value.indexOf('breadcrumb--is-ellipsis') !== -1
         && beforeLastItem.classList.value.indexOf('breadcrumb--is-ellipsis') !== -1
         && lastItem.classList.value.indexOf('breadcrumb--is-ellipsis') !== -1) {
-          console.log('called');
           isValid = false;
         }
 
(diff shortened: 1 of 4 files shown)


weblancaster:
retest this please

weblancaster:
retest this please

Poltergeist:
run integration tests

tests/components/Header-cy.js
```diff
@@ -0,0 +1,131 @@
+describe('Header Component', function () {
```
jfurrow: Could we name this "Page Header Component" instead (and then rename the file to `PageHeader-cy`)? Just to be explicit about exactly which component we're testing.

jfurrow:
Nice tests! I have one minor comment but otherwise this looks great!

tests/components/Header-cy.js
weblancaster: It's addressed @jfurrow thanks for the review 👍

wavesoft:
Looks good, works great. Great tests @weblancaster !

PR #2083, opened 2017-04-07
fix(integration-test): Fix flaky integration test.

weblancaster:
This PR fixes a flaky integration test where in some cases was failing because the selector is returning multiple elements instead of one in which is causing the test to not being able to click.

**Checklist**


mlunoe:
Looks good man!

Poltergeist:
LGTM Let's 🚢  this.

PR #2091, opened 2017-04-12
DCOS-15137: Fix pods environment variables handling

nLight:
Turns out multi container environment variables field is `environment` and not `env` as it is for single container services. This PR addresses the issue by splitting reducers and adjusting review screen.

Bonus track: fixing reducers uncovered one more bug in: `plugins/services/src/js/components/modals/NewCreateServiceModal.js`
where we applied single container hooks for multi container reducers 😅

**Checklist**

### DIFF
--- a/plugins/services/src/js/components/modals/NewCreateServiceModal.js
+++ b/plugins/services/src/js/components/modals/NewCreateServiceModal.js
@@ -604,35 +604,46 @@ class NewCreateServiceModal extends Component {
         MultiContainerNetworkingFormSection
       ];
 
-      let jsonParserReducers = combineParsers(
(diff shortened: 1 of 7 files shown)


jfurrow:
This looks & works perfectly! Good catch! 🏆 🥇 🎉

mlunoe:
Awesome work dude! I the error message that marathon provides is not really very explanatory in that the keys needs to be uppercased. I'll submit a ticket with them to update it

orlandohohmeier:
@nLight please open a PR to back-port the fix to DC/OS UI 1.9

nLight:
@orlandohohmeier already did ;) https://github.com/dcos/dcos-ui/pull/2097

PR #2092, opened 2017-04-12
[System Tests] Using correct dist folder on jenkins config

wavesoft:
This PR ensures that the locally-served files are correctly found by the `http-server`.

The bugs caused by this are quite tricky, since if `http-server` fails to locate the `dist` folder it will just forward every requests to the upstream server. Until this is merged, the Jenkins job will be effectively using the UI served from the server and not the development version.

### DIFF
--- a/system-tests/driver-config/jenkins.sh
+++ b/system-tests/driver-config/jenkins.sh
@@ -29,7 +29,7 @@ targets:
       PATH: "/usr/local/bin:$PATH"
 
     scripts:
-      proxy: http-server --proxy-secure=false -p 4201 -P \$CLUSTER_URL dist
+      proxy: http-server --proxy-secure=false -p 4201 -P \$CLUSTER_URL ../../dist
       auth: ../_scripts/auth-open.py
 
 secrets:


system-tests/driver-config/jenkins.sh
```diff
@@ -29,7 +29,7 @@ targets:
       PATH: "/usr/local/bin:$PATH"
 
     scripts:
-      proxy: http-server --proxy-secure=false -p 4201 -P \$CLUSTER_URL dist
+      proxy: http-server --proxy-secure=false -p 4201 -P \$CLUSTER_URL ../../dist
```
Poltergeist: This sounds like a major error source could we use something like `npm dir` to determine the correct path like `$(npm dir)/../../dist` ?

wavesoft: The `dcos-system-test-driver` always ensures that the paths are relative to the file being processed. However we moved the file from the root directory (previously called `driver-config.sh`) into `system-tests/driver-config`.

We could however use a variable in the file, like `BASE_DIR=../..` and then use `$BASE_DIR}/dist` in the script :thinking:

Poltergeist: We talked offline, the Jenkins driver is always ensuring that the path's in this file are relative to the file location. So it will always hit the right directory. Unless we move the file, I can live with this drawback.

PR #2097, opened 2017-04-13
[release 1.9] DCOS-15137: Fix pods environment

nLight:
⚠️  It's a back port of the https://github.com/dcos/dcos-ui/pull/2091 ⚠️ 

Copy/paste

Turns out multi container environment variables field is `environment` and not `env` as it is for single container services. This PR addresses the issue by splitting reducers and adjusting review screen.

Bonus track: fixing reducers uncovered one more bug in: `plugins/services/src/js/components/modals/NewCreateServiceModal.js`
where we applied single container hooks for multi container reducers 😅


jfurrow:
Perfect!

PR #2102, opened 2017-04-13
fix(page header navigation): Fix page header navigation flex props

jfurrow:
This PR fixes a bug where the dropdown menu in the page header didn't fill its `flex` parent. It only applies to the dropdown within the `page-header-content-section` because of its `flex-direction`.

Before

After:


mlunoe:
Looks great dude!

PR #2105, opened 2017-04-14
DCOS-14184: Always show declined offers data if available

jfurrow:
This PR changes the behavior of declined offers to always show the declined offers summary & table when the data is available. Previously the summary & table were hidden until the resource had been waiting more than 5 minutes.

**Checklist**

### DIFF
--- a/plugins/services/src/js/containers/pod-debug/PodDebugContainer.js
+++ b/plugins/services/src/js/containers/pod-debug/PodDebugContainer.js
@@ -15,7 +15,6 @@ import TimeAgo from "#SRC/js/components/TimeAgo";
 
 import DeclinedOffersHelpText from "../../constants/DeclinedOffersHelpText";
 import DeclinedOffersTable from "../../components/DeclinedOffersTable";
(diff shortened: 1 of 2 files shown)


rcorral:
Works well for me!

plugins/services/src/js/containers/pod-debug/PodDebugContainer.js
```diff
@@ -41,8 +40,7 @@ class PodDebugTabView extends React.Component {
     const {pod} = this.props;
     const queue = pod.getQueue();
 
-    if (!DeclinedOffersUtil.shouldDisplayDeclinedOffersWarning(pod)
```
mlunoe: If we are not using `shouldDisplayDeclinedOffersWarning` anywhere I think we should remove it

jfurrow: We are still using it in the `ServiceStatusWarning` component, here: https://github.com/dcos/dcos-ui/blob/6709773bd605e91d19c7cc3d59221168b444618f/plugins/services/src/js/components/ServiceStatusWarning.js#L16

PR #2106, opened 2017-04-14
DCOS-15162: Remove old service plan code

jfurrow:
This PR just removes all of the old old service upgrade code from the codebase. I thought there was more, but it looks like we removed it a while ago. :fire:

**Checklist**


weblancaster:
LGTM 👍

mlunoe:
Clean slate! ⛵️ when you have rebased and tests pass 💪

PR #2108, opened 2017-04-17
DCOS-15164: Add #isSDKService to ServiceUtil

jfurrow:
This PR adds a method to `ServiceUtil` to determine when a service is an SDK service. It does this by checking for the existence of the label `DCOS_COMMONS_API_VERSION`.

cc @mlunoe 

**Checklist**


Poltergeist:
LGTM Let's merge this.

PR #2109, opened 2017-04-17
DCOS-14755: Fix 'read only' value alignment

jfurrow:
This PR fixes the alignment of the "read only" value in `PodStorageConfigSection`. Before we were rendering an element with class `.configuration-map-value` inside a table cell which already has padding, so the padding was doubled up.

Before:

After:

**Checklist**

### DIFF
--- a/plugins/services/src/js/service-configuration/PodStorageConfigSection.js
+++ b/plugins/services/src/js/service-configuration/PodStorageConfigSection.js
@@ -5,8 +5,6 @@ import ConfigurationMapHeading
 import ConfigurationMapSection
   from "#SRC/js/components/ConfigurationMapSection";
 
-import ConfigurationMapBooleanValue
-  from "../components/ConfigurationMapBooleanValue";
 import ConfigurationMapTable from "../components/ConfigurationMapTable";
 import ServiceConfigDisplayUtil from "../utils/ServiceConfigDisplayUtil";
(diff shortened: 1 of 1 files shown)


plugins/services/src/js/service-configuration/PodStorageConfigSection.js
```diff
@@ -26,11 +25,11 @@ class PodStorageConfigSection extends React.Component {
         heading: 'Read Only',
         prop: 'readOnly',
         render(prop, row) {
-          return (
-            <ConfigurationMapBooleanValue
-              options={BOOLEAN_OPTIONS}
-              value={row[prop]} />
-          );
+          if (row[prop]) {
+            return BOOLEAN_OPTIONS.truthy;
```
weblancaster: just making sure we are expecting to send the `TRUE` and `FALSE` as string and not booleans.

jfurrow: Yes, we want the strings here because we just want to show the words "TRUE" or "FALSE" in the UI.

weblancaster: 👍

mlunoe:
Let's rebase this and merge it when tests pass ⛵️

nLight:
Summing up labels/ship-it comments I assume it's good to be merged. I tested it and it works.

PR #2110, opened 2017-04-17
DCOS-14845: Default to 0 for undefined scalar values in DeclinedOffersUtil

jfurrow:
This PR fixes a bug where an offered resource's `scalar` value might be displayed as `NaN` when the API omits this value. Now the `scalar` value is default assigned to `0`.

Before:

**Checklist**


rcorral:
LGTM!

PR #2111, opened 2017-04-17
DCOS-15229: Fix undefined route redirect

jfurrow:
This PR fixes a bug where the we redirect to a route with `undefined` after deploying a service.

**Checklist**


mlunoe:
Thanks for fixing this!

MatApple:
LGTM! 👍  Thanks for fixing!

PR #2079, opened 2017-04-06
[Route Service Tabs 1 of 2] DCOS-14508: Splitting routes into overview and detail

mlunoe:
This PR will split routes into services/overview and services/detail routes, which will allow for different children for these routes, which is necessary when routing tabs.

Route Service Tabs 1 of 2: https://github.com/dcos/dcos-ui/pull/2079
Route Service Tabs 2 of 2: https://github.com/dcos/dcos-ui/pull/2058

**Checklist**
- [x] Did you add a JIRA issue in a commit message or as part of the branch name?
- [ ] Did you add new unit tests?
- [ ] Did you add new integration tests?
- [ ] If this is a regression, did you write a test to catch this in the future?

PR #2112, opened 2017-04-17
DCOS-14508: Service breadcrumbs: Use overview route for service groups

mlunoe:
This bug was introduced in this PR: #2079

This PR makes sure that we display `overview` route for service groups and a `detail` route for a service:

**Checklist**


weblancaster:
Tested and worked fine.

jfurrow:
Looks good, works well! Nice test, too!

PR #2113, opened 2017-04-18
Modularization: Introduce an alias for /src

nLight:
This is a part of modularization effort. 

With this PR we transition from relative imports in our `plugins` to a webpack alias. 
As the first step we stop relying on relative paths and unify everything. 
Second step would be identifying common modules and moving them to SDK.

This PR makes it quite prominent to see how coupled our plugins are.

Now `import Something from '#SRC/js/...'` will import a module from the `src` dir. 

Decision to alias `src` and not `src/js` has been made to enable importing `img` and `styles` on the one hand and making it less confusing on the other.

Discussed offline with @orlandohohmeier we concluded to go with a `#` because `@` is being used in some `npm` packages and thus confusing. If you have a better suggestion I'm happy to rebase the PR. 

I had to reorder imports because of the `eslint-plugin-import` rule demanding absolute imports to go before relative.

I decided not to change anything inside the `/src` as relative imports are ok until you stay within a scope.

It is a lot of changes, but those are only rename/reorder changes.

🙏 

PS. see, how I'm selling a +1200/-1100 PR? :trollface:


nLight:
...and it's 💚

mlunoe:
One concern I have about being able to include is that if allowed we are going to reference images and styles elsewhere that should really live alongside the package.
A way for us to get around this is to have a formalised approach to isolating a package. Do we have a place where we can list these details?

mlunoe:
Works great!

nLight:
@mlunoe we could alias `/src/js` and no `/src` effectively forbidding anything else, of course one still will be able to `import ... from '#SRC/../img/...'` but that's another story.

Yet if you think it makes more sense to go `/src/js` way I'm happy to do so

orlandohohmeier:
@mlunoe I'm not too worried about referencing images as I'd say that this is only a temporary solution and the first step towards properly isolating the packages. We should reject new packages that rely on directory aliases as well as changes to the current packages that introduce new dependencies. In a second step we're going to remove the current dependencies including all aliases.

mlunoe:
Sounds good, man! Thanks for following up!

PR #2117, opened 2017-04-19
DCOS-14894: Aggregate resources for running tasks only

nLight:
This PR fixes the resources aggregation issue for Frameworks by taking into account running tasks only.

To test just run MoM and launch a task on Marathon that finishes in 5 seconds or so.
Observe that without the fix it will report growing number of resources in DC/OS UI

**Checklist**

### DIFF
--- a/plugins/services/src/js/structs/Framework.js
+++ b/plugins/services/src/js/structs/Framework.js
@@ -67,6 +67,9 @@ module.exports = class Framework extends Application {
     // resources of child frameworks won't be aggregated
     return tasks
       .filter(function (task) {
+        return task.state === 'TASK_RUNNING';
+      })
+      .filter(function (task) {
         return !task.isStartedByMarathon;
       })
       .reduce(function (memo, task) {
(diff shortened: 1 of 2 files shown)


Poltergeist:
LGTM 👍  💯

plugins/services/src/js/structs/Framework.js
```diff
@@ -67,6 +67,9 @@ module.exports = class Framework extends Application {
     // resources of child frameworks won't be aggregated
     return tasks
       .filter(function (task) {
+        return task.state === 'TASK_RUNNING';
```
weblancaster: Why not `task.state === 'TASK_RUNNING' && !task.isStartedByMarathon`?

jfurrow: I wondered this too. Is the additional `filter` introduced for readability? With @weblancaster's suggestion we would eliminate an iteration of the array.

orlandohohmeier:
@nLight please open a PR to back-port the fix to DC/OS UI 1.9 once this one is green. 💚

MatApple:
@nLight @orlandohohmeier - I'm back porting now so we can get this in. Will open a PR in a few minutes.

PR #2119, opened 2017-04-19
DCOS-12290: (UCR) - Add bridge network support

Poltergeist:
**Checklist**

Add Bridge mode network support for ucr
single container apps.

Closes DCOS-12290

### DIFF
--- a/plugins/services/src/js/reducers/serviceForm/__tests__/Container-test.js
+++ b/plugins/services/src/js/reducers/serviceForm/__tests__/Container-test.js
@@ -40,7 +40,7 @@ describe('Container', function () {
             image: 'foo',
             forcePullImage: null,
             privileged: null,
-            network: undefined,
+            network: null,
             portMappings: null
           },
           type: 'DOCKER',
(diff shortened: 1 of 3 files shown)


plugins/services/src/js/reducers/serviceForm/__tests__/Container-test.js
```diff
@@ -40,7 +40,7 @@ describe('Container', function () {
             image: 'foo',
             forcePullImage: null,
             privileged: null,
-            network: undefined,
```
mlunoe: Could we add tests for UCR and BRIDGE to test that the output is as expected?

nLight:
What am I doing wrong?

screen shot 2017-04-20 at 12 55 32

Poltergeist:
This feature is yet not implemented into Marathon, yet. Next version should contain this feature after that we can also move this feature to master.

plugins/services/src/js/reducers/serviceForm/__tests__/Container-test.js
Poltergeist: Added and actually fixed the tests. Unfortunately a lot of the container tests were not executed.

nLight:
@mlunoe could you please give it another shot?

mlunoe:
Looks good and works well!

PR #2126, opened 2017-04-20
⬅️ DCOS-14842: Change size label

orlandohohmeier:
---
⚠️  _This PR back-ports a fix to release/1.9 introduced with #2053._

---

This PR updates the size label for external volumes due to the way marathon handle the value.

> Marathon treats size value for External volumes as it is in GiB and sends this data to Mesos.
> This is going to be fixed (changed to MiB) in 1.10 to align it with other volumes.


orlandohohmeier:
run integration tests

jfurrow:
Yep, this works great!

PR #2078, opened 2017-04-06
DCOS-11781: fix scroll view not set correctly in packages table

mlunoe:
This PR fixes so the bottom packages in the table are always visible:

**Checklist**
- [x] Did you add a JIRA issue in a commit message or as part of the branch name?
- [ ] Did you add new unit tests?
- [x] Did you add new integration tests?
- [x] If this is a regression, did you write a test to catch this in the future?

PR #2128, opened 2017-04-20
Fix task table keys (Mesos task IDs are not unique)

mlunoe:

**Checklist**


mlunoe:
Adding hold merge as we are currently investigating with Mesos whether they can implement a fix for us instead of having this issue and performance hit.

MatApple:
We will not get a fix for https://github.com/dcos/dcos-ui/pull/2128#issuecomment-295980536 in 1.10 so removing the hold label.

mlunoe:
😢 Ok, well, let's get this reviewed then

rcorral:
Sounds good :)

PR #2129, opened 2017-04-21
[release/1.8] Fix task table keys

mlunoe:
This PR is the same as this just pointing to release 1.8: https://github.com/dcos/dcos-ui/pull/2128

**Checklist**


jfurrow:
Works great! Let's merge when tests pass!

PR #2130, opened 2017-04-21
↩️ Fix task table keys (Mesos task IDs are not unique)

mlunoe:
This PR is the same as this just pointing to release 1.9: https://github.com/dcos/dcos-ui/pull/2128

**Checklist**


mlunoe:
Run integration tests

mlunoe:
Run integration tests

nLight:
Since the master fix is on `hold` feel free to discard if we want to go less performant way for the 1.9 release

rcorral:
Approving this so that we can merge this into 1.9. Let's keep holding on the fix for `master` in hopes that we can have this fixed in the API.

PR #2131, opened 2017-04-21
DCOS-14508: Make service (and pod) detail tabs routed 

mlunoe:
This PR is an extension of this, as the original got reverted due to the lack of support for pod detail tabs: https://github.com/dcos/dcos-ui/pull/2058

This PR makes the service detail tabs routed. A bug emerged while implementing this that was introduced here: https://github.com/dcos/dcos-ui/pull/1957. This PR also takes care of that.

I moved every detail tab view to its own route. In order to decide whether to display the pod or regular detail view, I created high-order components to make this decision.

**Checklist**

P.S. Sorry for the big size. We missed the pod part of the change. The good news is that the service part was already reviewed here: https://github.com/dcos/dcos-ui/pull/2058
Everything after the `Revert "Revert "[Route Service Tabs 2 of 2] DCOS-14508: Make service …` commit is additional code to make pods work.


mlunoe:
Rebased! ✨

mlunoe:
Run integration tests

mlunoe:
There is a bug I discovered with TaskDetail, when tasks are not defined. That is handled here: https://github.com/dcos/dcos-ui/pull/2135

plugins/services/src/js/components/HighOrderServiceConfiguration.js
```diff
@@ -0,0 +1,31 @@
+import React from "react";
+
+import Pod from "../structs/Pod";
+import PodConfigurationContainer
+  from "../containers/pod-configuration/PodConfigurationContainer";
+import Service from "../structs/Service";
+import ServiceConfigurationContainer
+  from "../containers/service-configuration/ServiceConfigurationContainer";
+
+const HighOrderServiceConfiguration = props => {
```
orlandohohmeier: Please adjust the component name as it's not an good practice to name components after patterns. How about something like `GeneralServiceConfigurationContainer`. 

Side note: The naming of the `ServiceConfigurationContainer` seems off as pods are services as well...

orlandohohmeier: Please use old school functions for function components to make sure the the context of the surrounding module don't get mixed with the component. For details regarding arrow functions please see: http://stackoverflow.com/questions/22939130/when-should-i-use-arrow-functions-in-ecmascript-6

jfurrow: #nit A while back we decided not to use fat arrow functions.

plugins/services/src/js/components/HighOrderServiceDebug.js
```diff
@@ -0,0 +1,23 @@
+import React from "react";
+
+import Pod from "../structs/Pod";
+import PodDebugContainer from "../containers/pod-debug/PodDebugContainer";
+import Service from "../structs/Service";
+import ServiceDebugContainer
+  from "../containers/service-debug/ServiceDebugContainer";
+
+const HighOrderServiceDebug = props => {
```
jfurrow: Same here, let's nix the fat arrow function.

plugins/services/src/js/components/HighOrderServiceInstances.js
```diff
@@ -0,0 +1,25 @@
+import React from "react";
+
+import Pod from "../structs/Pod";
+import PodInstancesContainer
+  from "../containers/pod-instances/PodInstancesContainer";
+import Service from "../structs/Service";
+import ServiceInstancesContainer
+  from "../containers/service-instances/ServiceInstancesContainer";
+
+const HighOrderServiceInstances = props => {
```
jfurrow: Same here, let's nix the fat arrow function.

plugins/services/src/js/components/HighOrderServiceConfiguration.js
```diff
@@ -0,0 +1,31 @@
...
+  return (
+    <ServiceConfigurationContainer
+      errors={errors}
+      onEditClick={onEditClick}
+      service={service}
+    />
+  );
+};
+
+HighOrderServiceConfiguration.propTypes = {
```
orlandohohmeier: I would change the component interface to only accept one service property that could be either a pod or  an app or frame work. It's a good practice to reduce the number of inputs to create an easy to use component. Please note the Pod inherits all it's properties from Service.

plugins/services/src/js/containers/pod-configuration/PodConfigurationContainer.js
```diff
@@ -9,6 +9,7 @@ class PodConfigurationTabView extends React.Component {
     const spec = this.props.pod.getSpec();
     const localeVersion = new Date(spec.getVersion()).toLocaleString();
 
+    // TODO: Implement ability to edit a Pod
```
jfurrow: Could we add a JIRA ticket number here?

plugins/services/src/js/components/HighOrderServiceDebug.js
orlandohohmeier: Please use an old school function and adjust the name here as well.

plugins/services/src/js/components/HighOrderServiceDebug.js
```diff
@@ -0,0 +1,23 @@
...
+  if (pod != null) {
+    return <PodDebugContainer pod={pod} />;
+  }
+
+  return <ServiceDebugContainer service={service} />;
+};
+
+HighOrderServiceDebug.propTypes = {
+  pod: React.PropTypes.instanceOf(Pod),
+  service: React.PropTypes.instanceOf(Service)
```
orlandohohmeier: As with the configuration component I suggest to reduce the number of properties.

plugins/services/src/js/components/HighOrderServiceInstances.js
orlandohohmeier: Please use an old school function and adjust the name.

plugins/services/src/js/routes/services.js
```diff
@@ -171,44 +180,117 @@ const serviceRoutes = [
           },
           {
             type: Route,
+            path: "configuration",
+            title: "Configuration",
+            component: HighOrderServiceConfiguration,
+            buildBreadCrumb() {
```
jfurrow: We actually don't use the `#buildBreadCrumb` method anymore, right? As a separate task, we should remove all of these methods, but let's not add any more.

plugins/services/src/js/components/HighOrderServiceInstances.js
```diff
@@ -0,0 +1,25 @@
...
+  const { pod, service, params } = props;
+  if (pod != null) {
+    return <PodInstancesContainer pod={pod} />;
+  }
+
+  return <ServiceInstancesContainer service={service} params={params} />;
+};
+
+HighOrderServiceInstances.propTypes = {
```
orlandohohmeier: Please reduce the number of properties. As a user of the component I shouldn't need to worry wether my service is a pod or an app.

plugins/services/src/js/containers/pod-configuration/PodConfigurationContainer.js
orlandohohmeier: Is this a regression? Please file a JIRA!

plugins/services/src/js/containers/service-detail/ServiceDetail.js
```diff
@@ -187,50 +160,43 @@ class ServiceDetail extends mixin(TabsMixin) {
...
 
     return tabs;
   }
 
   render() {
-    const { children, service: { id } } = this.props;
-    const breadcrumbs = <ServiceBreadcrumbs serviceID={id} />;
+    const { children, actions, errors, params, routes, service } = this.props;
+    const breadcrumbs = <ServiceBreadcrumbs serviceID={service.id} />;
+    const clonedProps = {
```
orlandohohmeier: Why do we need to *clone* the props?

jfurrow:
This works well! I suggested some minor changes, but functionally it's perfect!

plugins/services/src/js/containers/service-detail/ServiceDetail.js
```diff
@@ -187,50 +160,43 @@ class ServiceDetail extends mixin(TabsMixin) {
...
+      params,
+      routes,
+      service,
+      errors: MarathonErrorUtil.parseErrors(errors[ActionKeys.SERVICE_EDIT]),
+      onClearError: this.handleEditClearError,
+      onEditClick: actions.editService,
+      volumes: service.getVolumes().getItems()
+    };
+
+    const clonedChildren = React.Children.map(children, function(child) {
```
orlandohohmeier: Why do we need to clone the children? This seems wrong.

plugins/services/src/js/routes/services.js
mlunoe: I know, but I didn't want this PR to grow anymore. We have a different task for this already. Hope you don't mind me not including that in this already rather large PR

jfurrow: Oh yeah I don't think you should remove all of the `buildBreadCrumb` methods here, but shouldn't we avoid adding new ones, since they're not in use anyway?

plugins/services/src/js/components/HighOrderServiceConfiguration.js
mlunoe: It is _not_ a container since it doesn't use a store. It is a high-order component, hence I chose that name. We do this for containers as well, but please come up with some better suggestions for this, so I can use that. I do not have better suggestions myself.

plugins/services/src/js/containers/pod-detail/PodDetail.js
```diff
@@ -162,8 +153,7 @@ class PodDetail extends mixin(TabsMixin) {
             tabs={this.tabs_getUnroutedTabs()}
```
MatApple: Should we still use this method now that tabs are routed?

orlandohohmeier:
Please discuss the proposed solution with @philipnrmn and @nLight as they though about different way enable routed tabs using the `NavigationService` in combination with the `Page` skeleton.

plugins/services/src/js/containers/pod-detail/PodDetail.js
```diff
@@ -146,6 +127,16 @@ class PodDetail extends mixin(TabsMixin) {
 
     const breadcrumbs = <ServiceBreadcrumbs serviceID={pod.id} />;
 
+    const clonedProps = { pod };
+    const clonedChildren = React.Children.map(children, function(child) {
```
MatApple: Could you please add a comment for why we have to do this?

plugins/services/src/js/containers/pod-configuration/PodConfigurationContainer.js
mlunoe: There was never an implementation for this. I will create a Jira

plugins/services/src/js/containers/service-detail/ServiceDetail.js
MatApple: Here too please :-)

plugins/services/src/js/containers/service-instances/ServiceInstancesContainer.js
```diff
@@ -77,16 +77,16 @@ class ServiceTasksContainer extends mixin(StoreMixin) {
 
   render() {
     return (
-      <MountService.Mount type="ServiceTasksContainer:TasksContainer">
+      <MountService.Mount type="ServiceInstancesContainer:TasksContainer">
```
MatApple: Does this affect any Mounts registered for the old type, plugins inclusive?

plugins/services/src/js/containers/service-detail/ServiceDetail.js
mlunoe: I have already answered that before, but I'll answer it again;
Yes, we "need" to clone props since we are passing them down to their route.
It would be nice to get these as parameters in the route, but that requires us to change around logic in each individual tab, and in an effort to try to contain this change in as small chunks as possible, I figured this was the path of least resistance.

And I'll add a comment about it.

plugins/services/src/js/routes/services.js
MatApple: Do we still need to `buildBreadCrumb`?

plugins/services/src/js/containers/service-detail/ServiceDetail.js
mlunoe: see my reply here:
https://github.com/dcos/dcos-ui/pull/2131#discussion_r113984366

plugins/services/src/js/components/HighOrderServiceConfiguration.js
mlunoe: Done

plugins/services/src/js/components/HighOrderServiceDebug.js
mlunoe: Done

plugins/services/src/js/components/HighOrderServiceInstances.js
mlunoe: Done

plugins/services/src/js/components/HighOrderServiceConfiguration.js
orlandohohmeier: I assume you're wrapping them to provide a constant/simplified interface thus I'd say a name like `GeneralServiceConfigurationContainer` isn't wrong though I think there are propbably better names. BTW: I'd this is a facade not a higher-order component. 

I suggest that we double check if this is the right architecture pattern (including our extensive use of container), but that shouldn't block this PR. I'm fine if you're not addressing the name, but

plugins/services/src/js/containers/pod-configuration/PodConfigurationContainer.js
mlunoe: Done

plugins/services/src/js/containers/service-detail/ServiceDetail.js
orlandohohmeier: Let's file a JIRA to change the logic so that we get the respective bits as parameters.

plugins/services/src/js/components/HighOrderServiceConfiguration.js
mlunoe: Done

plugins/services/src/js/components/HighOrderServiceDebug.js
mlunoe: Done

plugins/services/src/js/containers/pod-configuration/PodConfigurationContainer.js
mlunoe: Done

plugins/services/src/js/containers/service-detail/ServiceDetail.js
mlunoe: Done

plugins/services/src/js/containers/service-detail/ServiceDetail.js
mlunoe: Done

plugins/services/src/js/containers/pod-detail/PodDetail.js
mlunoe: Done

plugins/services/src/js/containers/pod-detail/PodDetail.js
mlunoe: Good catch! Done

plugins/services/src/js/containers/service-detail/ServiceDetail.js
mlunoe: Done

plugins/services/src/js/containers/service-instances/ServiceInstancesContainer.js
mlunoe: Good catch! Created a PR to update affected plugins

plugins/services/src/js/routes/services.js
mlunoe: They have been removed from newly added routes now. Thanks

mlunoe:
Ok, all comments addressed!

We should have a conversation about how to move forward with routing tabs as it seems there conflicting ideas about the approach, but I would like that to not hold up this PR.

I still see this providing a step in the right direction and hopefully makes it easier to build upon than it was before.

plugins/services/src/js/components/HighOrderServiceDebug.js
mlunoe: Done

plugins/services/src/js/components/HighOrderServiceInstances.js
mlunoe: Done

plugins/services/src/js/components/HighOrderServiceConfiguration.js
mlunoe: Done

jfurrow:
@mlunoe This PR still has newly added `buildBreadCrumb` methods — maybe you forgot to push a commit?

mlunoe:
I indented routes, so it might look like that, but there shouldn't be any new crumbs added in the route definitions

PR #2134, opened 2017-04-21
DCOS-14845: DeclinedOffers now default to 0 for undefined scalar values

jfurrow:
**This is the same commit as https://github.com/dcos/dcos-ui/pull/2110, but now Prettier** 😄 

This PR fixes a bug where an offered resource's `scalar` value might be displayed as `NaN` when the API omits this value. Now the `scalar` value is default assigned to `0`.

Before:

**Checklist**

### DIFF
--- a/plugins/services/src/js/utils/DeclinedOffersUtil.js
+++ b/plugins/services/src/js/utils/DeclinedOffersUtil.js
@@ -201,14 +201,14 @@ const DeclinedOffersUtil = {
         unmatchedResource: declinedOffer.reason,
         offered: resources.reduce(
           (accumulator, resource) => {
-            const { name, role } = resource;
+            const { name, role, scalar = 0 } = resource;
 
             if (name === "ports") {
               const { ranges = [] } = resource;
 
               accumulator[name] = ranges.reduce(rangeReducer, []);
             } else {
-              accumulator[name] = resource.scalar;
+              accumulator[name] = scalar;
             }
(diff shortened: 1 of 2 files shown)


plugins/services/src/js/utils/DeclinedOffersUtil.js
```diff
@@ -201,14 +201,14 @@ const DeclinedOffersUtil = {
         unmatchedResource: declinedOffer.reason,
         offered: resources.reduce(
           (accumulator, resource) => {
-            const { name, role } = resource;
+            const { name, role, scalar = 0 } = resource;
```
nLight: Is it possible to get `null` in `resource.scalar`? Because in this case the default value won't be assigned.

jfurrow: Good call! And to be honest, I don't know. I'll adjust this to make sure `null` values are replaced with `0` as well.

mlunoe:
I'll wait until @jfurrow has pushed his update to handle `null` before merging

plugins/services/src/js/utils/DeclinedOffersUtil.js

jfurrow:
Thanks @mlunoe, this is done!

jfurrow:
run integration tests

nLight:
run integration tests

nLight:
run integration tests

PR #2136, opened 2017-04-24
Upgrade node to 8 and npm to 5.0.3

nLight:
New LTS version 8 is out \o/ why not using it?

I went ahead and updated all the places I could find.
I also had to fix one test (version dependent tests /o\)
and a webpack config bug

## Dependencies

✅ ~https://github.com/dcos/dcos-ui/pull/2148~
✅ ~https://github.com/dcos/dcos-ui/pull/2153~
🕐 https://github.com/mesosphere/reactjs-components/pull/406


nLight:
Looks like our Jenkins should be adjusted

mlunoe:
There is a bug in node 6 that makes jest not run and proposed solution is to downgrade to node 5.11.1: https://github.com/webpack/webpack/issues/2463#issuecomment-218164083

Sad face 😢

mlunoe:
I could not get it to work with node 6 :(

mlunoe:
However, I did put up a PR for prettier in reactjs-components: https://github.com/mesosphere/reactjs-components/pull/402

nLight:
Does it work with node 7? 
Since node is just a build tool, meaning we bundle the app and don't run anything on node.js in production, for us we could go with a non LTS release.

Anyways, this is not a priority, because of the reason I mentioned above, let's move forward with more important PRs rather then investing in this one right now. 🙏

mlunoe:
Yup, I think node 7 is the way to go. Did not try it out though

nLight:
Another proposed solution was to upgrade babel to 6.10.x which looks like the right thing to do choosing between downgrading to node 5 (non-LTS) and upgrading babel to a recent version :)

weblancaster:
@nLight let's just use Node 7+, as you noted we are only using as a tool and not shipping to prod.

nLight:
@weblancaster yep, still we should upgrade babel :)

weblancaster:
@nLight yes

weblancaster:
@nLight please change `validate-engine-versions.js` to accomodate the node changes.

weblancaster:
@nLight #2148 worked greatly with `node v7.8.0` `npm 4.2.0` and my build was noticeably faster.

mlunoe:
Ok, related:
I put up a PR to update reactjs-components that makes the prettier linting rule happy: https://github.com/dcos/dcos-ui/pull/2153

nLight:
I observe strange bugs with this setup:
when running `npm run build` it errors out performing
`./node_modules/.bin/gulp checkDependencies && npm install compression-webpack-plugin@latest` saying `Cannot find module internal/fs`

Then trying to run `npm shrinkwrap --dev` it enters into an infinite loop

😩

nLight:
Turns out the loop of shrinkwrap was caused by `npm run shrinkwrap` task that calls `npm shrinkwrap --dev` I suspect npm 4 now doesn't require `run` keyword for overrides this is why it used to work with npm 3 and stopped with npm 4. 

Moreover I think we don't need fixShrinkwrap anymore as new setup doesn't depend on `fsevent` and there's no need in removing it from the npm-shrinkwrap.json

nLight:
Closing this one for now

PR #2138, opened 2017-04-25
fix(test-apps.js): Fix external volume assertions

jfurrow:
This PR fixes an assertion in the system tests which looks for the wrong units for the volume size.

**Checklist**


jfurrow:
Actually, this is not necessary. The problem is that the release branches haven't been updated yet, and these system tests are running against that.

PR #2143, opened 2017-04-28
feat(DCOS-13976): Add effort to protect from clickjacking attack

weblancaster:
(**DCOS-13976**) This solution is outdated and there is many ways around it BUT we have no harm in having it (because it's so small having in the head is fine).
The ideal solution was implemented here https://github.com/dcos/dcos/pull/1506

For more info about it [Clickjacking Defense](https://www.owasp.org/index.php/Clickjacking_Defense_Cheat_Sheet)

To test this create a index.html page and use iframe markup pointing to `https://52.41.194.202`. You should see this in the console `Refused to display 'https://52.41.194.202/' in a frame because it set 'X-Frame-Options' to 'deny'.`

Because security.
security

**Checklist**

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -38,7 +38,7 @@
 
     <script type="text/JavaScript">
     // old school Frame busting solution
-    // even though this solution has other way around it
+    // even though this solution has ways around it
     // the X-Frame-Options set in the HEADERS is the ideal
     if (top != self) { top.location.replace(self.location.href); }
     </script>
--- a/src/index.html
+++ b/src/index.html
@@ -35,6 +35,13 @@
     <link rel="apple-touch-icon-precomposed" sizes="152x152" href="./img/components/icons/favicon-apple-152x152.png">
     <script type="text/javascript">(function () {var PAGE_LOADED_TIME = Date.now(); window.getPageLoadedTime = function () {return PAGE_LOADED_TIME;}})();</script>
     <!--[if BOOTSTRAP-CSS]><![endif]-->
+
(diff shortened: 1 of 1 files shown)


src/index.html
```diff
@@ -35,6 +35,13 @@
     <link rel="apple-touch-icon-precomposed" sizes="152x152" href="./img/components/icons/favicon-apple-152x152.png">
     <script type="text/javascript">(function () {var PAGE_LOADED_TIME = Date.now(); window.getPageLoadedTime = function () {return PAGE_LOADED_TIME;}})();</script>
     <!--[if BOOTSTRAP-CSS]><![endif]-->
+
+    <script type="text/JavaScript">
+    // old school Frame busting solution
+    // even though this solution has ways around it
+    // the X-Frame-Options set in the HEADERS is the ideal
+    if (top != self) { top.location.replace(self.location.href); }
```
MatApple: I suspect this might destroy our integration testing.

rcorral: I think this is also going to break our banner plugin. I think it needs to also check that the location isn't the same as the top frame or check that it's using the banner plugin, somehow.

weblancaster: Do we even need this after adding the X-FRAME? this is basically useless, it's more to keep our souls clean because technically based on the research I did it's pretty easy to by pass it and well know how to by pass.

MatApple: Ok, I suggest we close this if you're Ok with that @weblancaster - looks like your other fixes (https://github.com/dcos/dcos/pull/1506) will take care of this 👍

PR #2147, opened 2017-05-02
feat(DCOS-15151): Add Mesos details to overview page

weblancaster:
(**DCOS-15151**)This PR adds Mesos details to overview page.
To test, go to overview page and then you should be able to see Mesos details like so.

screen shot 2017-05-01 at 9 42 51 pm

**Checklist**

### DIFF
--- a/src/js/pages/system/OverviewDetailTab.js
+++ b/src/js/pages/system/OverviewDetailTab.js
@@ -6,6 +6,7 @@ import React from "react";
 /* eslint-enable no-unused-vars */
 import { StoreMixin } from "mesosphere-shared-reactjs";
 import { FormattedMessage } from "react-intl";
+import moment from "moment";
 
 import Breadcrumb from "../../components/Breadcrumb";
 import BreadcrumbTextContent from "../../components/BreadcrumbTextContent";
@@ -26,7 +27,6 @@ import MetadataStore from "../../stores/MetadataStore";
 import MesosStateStore from "../../stores/MesosStateStore";
 import Page from "../../components/Page";
 import VersionsModal from "../../components/modals/VersionsModal";
-import { msToRelativeTime } from "../../utils/DateUtil";
 import { isEmpty } from "../../utils/ValidatorUtil";
 
 const METHODS_TO_BIND = [
@@ -260,7 +260,7 @@ class OverviewDetailTab extends mixin(StoreMixin) {
             <FormattedMessage id="COMMON.BUILT" />
           </ConfigurationMapLabel>
           <ConfigurationMapValue>
-            {msToRelativeTime(mesosBuilt)} {this.getMesosBuildUser()}
+            {moment(Date(mesosBuilt)).fromNow()} {this.getMesosBuildUser()}
           </ConfigurationMapValue>
         </ConfigurationMapRow>
         <ConfigurationMapRow key="started">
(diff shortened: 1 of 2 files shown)


weblancaster:
retest this please

weblancaster:
retest this please

weblancaster:
run integration tests

src/js/pages/system/OverviewDetailTab.js
```diff
@@ -22,8 +23,11 @@ import Loader from "../../components/Loader";
 import MarathonStore
   from "../../../../plugins/services/src/js/stores/MarathonStore";
 import MetadataStore from "../../stores/MetadataStore";
+import MesosStateStore from "../../stores/MesosStateStore";
 import Page from "../../components/Page";
 import VersionsModal from "../../components/modals/VersionsModal";
+import { msToRelativeTime } from "../../utils/DateUtil";
```
nLight: `msToRelativeTime` operates on milliseconds but we get a unix timestamp in _seconds_ from the API thus you see 47 ago on the screen

src/js/pages/system/OverviewDetailTab.js
```diff
@@ -179,6 +188,101 @@ class OverviewDetailTab extends mixin(StoreMixin) {
...
+  /**
+   * Get user if existent otherwise
+   * return null
+   *
+   * @returns {Component|Null} User string or null
+   *
+   * @memberOf OverviewDetailTab
+   */
+  getMesosBuildUser() {
+    const mesosConfig = MesosStateStore.get("lastMesosState");
```
nLight: Let's pass `mesosConfig` as an argument into this function

weblancaster: @nLight thanks for the reviews.. quick question though, why would you rather have this being passed?

Having the method `getMesosBuildUser()` to be the only to know about the `build_user` makes it easy to understand and follow the path if needed, we are not unit testing here but this would also help make better/more precise unit tests. (we wont gain any performance improvement here if passed instead since the data is store in a obj and "always available" for retrieving)

The method `getMesosDetails()` don't need to have logic for this particular data, it only need the data to be displayed. Another point is that the "clean code" patterns suggest not passing arguments to methods/funcs when avoidable.

All that being said if you have strong feelings against it I can move the logic to `getMesosDetails()`

src/js/pages/system/OverviewDetailTab.js
weblancaster: This is addressed @nLight 

thanks for reviewing.

src/js/pages/system/OverviewDetailTab.js
rcorral: @nLight could you comment or approve please?

nLight: It feels wrong to fetch this data once again after we've got it in `getMesosDetails` which is _the only_ function that calls `getMesosBuildUser` so my suggestion would be to pass `mesosConfig` or even `mesosConfig.build_user` as an argument while keeping the logic in `getMesosBuildUser`. 

To you argument about unit testing it would be much easier to test a function that depends on one argument rather that a function that has to mock a Store.

If my arguments aren't strong enough feel free to leave it as it is and merge.

weblancaster: @nLight it's addressed thanks.

rcorral:
run integration tests

PR #2151, opened 2017-05-02
[Deployments 2 of 3]: DCOS-15161: Add supplemental content to PageHeaderActions

jfurrow:
~~[1 of 3](https://github.com/dcos/dcos-ui/pull/2150)~~
__2 of 3__
[3 of 3](https://github.com/dcos/dcos-ui/pull/2152)

This PR adds the ability to have supplemental content in PageHeaderActions. This will be used for the new deployments feature.

**Checklist**
- [x] Did you add a JIRA issue in a commit message or as part of the branch name?
- [ ] Did you add new unit tests?
- [ ] Did you add new integration tests?
- [ ] If this is a regression, did you write a test to catch this in the future?

PR #2152, opened 2017-05-02
[Deployments 3 of 3]: DCOS-15161: Refactor deployments into modal

jfurrow:
~~[1 of 3](https://github.com/dcos/dcos-ui/pull/2150)~~
~~[2 of 3](https://github.com/dcos/dcos-ui/pull/2151)~~
__3 of 3__

This PR moves the deployments tab from the sidebar into a modal. The modal can be accessed via the page header on the services overview page when a deployment is active.

You'll notice I removed some unit tests for the `DeploymentsModal` (previously named `DeploymenstTab`); these tests are now covered by the integration tests.

**Checklist**


leemunroe:
@jfurrow Nice! Couple things I noticed:

1. Can we auto expand the rows? That is consistent with the behavior today. They are actually not that useful when collapsed as the ID doesn't tell us anything about what is deploying.

2. Today I see more information about the deployment status e.g. `Deploying (ScaleApplication)` In this solution I only see `Deploying`. I would expect to see the same information. Or if we're concerned about spacing, then can we show the more granular part of the status e.g. `ScaleApplication`

3. Today I can click on the location/group name that the service is in. In this solution I cannot. Was that removed on purpose?

4. This is nitpicky but the loading animation is [rotating anti-clockwise](https://cl.ly/3Q3P110l2n2J). I would expect this to be clockwise. Consistent with most loading animations. 

Before/After for reference
image
image

cc @ashenden

jfurrow:
Thanks for the review @leemunroe!

1. Good point, I'll add this
2. No reason to not show the additional details, I just omitted it to match the mockups. I'll add this
3. There was no reason for removing this, either. I'll add this
4. This is just the default styling from [the spinner component we use](https://connoratherton.com/loaders), but I'll change this

jfurrow:
@leemunroe All of your comments have been addressed. To prevent this PR from getting larger, I put the expanded by default behavior and reversed loading indicator animation in separate PRs, so you'll need to cherry-pick those commits to see this all working together. Those PRs are https://github.com/dcos/dcos-ui/pull/2165 and https://github.com/dcos/dcos-ui/pull/2173

weblancaster:
Great work @jfurrow worked fine locally and the code seems 💯

PR #2155, opened 2017-05-03
DCOS-15287: Improve app tests

juliangieseke:
Improve the app system tests by adjusting and improving the test cases.

* Add additional checks to verify that the Marathon response is as expected. To do so the test will open the service detail, hit edit and check the form content.
* Introduce new test to check that Marathon throws an error in case an app with an existig ID already exists.
* Adjust create simple app test to bail early in case of an error.

### DIFF
--- a/system-tests/services/test-apps.js
+++ b/system-tests/services/test-apps.js
@@ -8,7 +8,12 @@ describe("Services", function() {
   describe("Applications", function() {
     beforeEach(function() {
       cy.visitUrl(`services/overview/%2F${Cypress.env("TEST_UUID")}/create`);
+      cy.server().route("POST", /\/service\/marathon\/v2\/apps/).as("appsReq");
     });
+
+    function clickElementByText(text) {
+      cy.contains(text).click();
+    }
 
     afterEach(() => {
       cy.window().then(win => {
@@ -17,11 +22,11 @@ describe("Services", function() {
     });
 
     function selectMesosRuntime() {
-      cy.contains("More Settings").click();
-      cy.contains("Mesos Runtime").click();
+      clickElementByText("More Settings");
+      clickElementByText("Mesos Runtime");
     }
 
-    it("Create a simple app", function() {
+    it("should create a simple app", function() {
       const serviceName = "app-with-inline-shell-script";
       const cmdline = "while true; do echo 'test' ; sleep 100 ; done";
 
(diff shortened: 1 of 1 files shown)


system-tests/services/test-apps.js
```diff
@@ -638,9 +888,52 @@ describe("Services", function() {
...
+      cy
+        .root()
+        .getFormGroupInputFor("Service ID *")
+        .should("have.value", `/${Cypress.env("TEST_UUID")}/${serviceName}`);
+
+      cy
+        .root()
+        .getFormGroupInputFor("Memory (MiB) *")
+        .should("have.value", "10");
+      cy.root().getFormGroupInputFor("Command").should("have.value", cmdline);
```
weblancaster: Let's try keep consistent throughout files and break a line in every chaining, it makes a easier to read and catch any human error.

```
cy.get('blabla')
    .foo()
    .bar()
```

system-tests/services/test-apps.js
```diff
@@ -938,9 +1269,79 @@ describe("Services", function() {
         .getTableRowThatContains(serviceName)
         .get(".bar.healthy", { timeout: Timeouts.SERVICE_DEPLOYMENT_TIMEOUT })
         .should("exist");
+
+      // Now click on the name
+      cy
+        .get(".page-body-content table")
+        .getTableRowThatContains(serviceName)
+        .get("a.table-cell-link-primary")
+        .contains(`${serviceName}`)
```
weblancaster: you can always use the reference `serviceName ` instead template strings. Not a big deal though.

system-tests/services/test-apps.js
```diff
@@ -92,6 +110,13 @@ describe("Services", function() {
       // Run service
       cy.get("button.button-primary").contains("Run Service").click();
 
+      // Also no Error should exist
+      cy
+        .wait("@appsReq")
+        .wait(1000)
```
juliangieseke: without that second wait() cypress tries to get `.alert-danger` before the page is loaded/rendered and then assumes everything is fine. 
But TBH I'm not sure anymore if this check is needed at all. if there is an error, the next check will fail anyway.

system-tests/services/test-apps.js
juliangieseke: I agree. Please have a look at DCOS-15564 regarding this.

system-tests/services/test-apps.js
```diff
@@ -100,14 +125,81 @@ describe("Services", function() {
...
+      // Click Review and Run
+      cy.contains("button", "Review & Run").click();
+
+      // Run service
+      cy.contains("button", "Run Service").click();
+
+      // Also no Error should exist
+      cy
+        .wait("@appsReq")
+        .wait(2000)
```
juliangieseke: ok, lets try without it, if tests are passings thats okay for me. but we have to keep that in mind when cypres starts to fail randomly again

system-tests/services/test-apps.js
juliangieseke: actually prettier thinks one line is better:

/.../dcos-ui/system-tests/services/test-apps.js
  899:9  error  Follow `prettier` formatting (expected '.' but found '\n')  prettier/prettier

weblancaster: must the the formatting on the tabs/spaces because if we look the other files we are breaking the line and `prettier` isn't complaining.

juliangieseke: I have asked @nLight about it, he also says prettier doesnt allow short lines to be broken up.

I'm totally with you that especially the test-apps.js file is too big and hard to read - I'm already working on refactoring the tests and split up these big files. (Again, see DSCO-15564)

nLight: One the whole chain fits in one line prettier runs it in one line.

nLight:
I tried 

```
  -e CLUSTER_URL=$OPEN_CLUSTER_URL  \
  -v `pwd`:/dcos-ui \
  -v ./system-tests/driver-config/dev.sh
```

and it doesn't work, I'm still behind how the whole system-test thingie should be working

nLight:
Jobs tests are failing, unrelated to the changes though

system-tests/services/test-apps.js
nLight: Hey @weblancaster would you please approve the PR, or is there anything left?

juliangieseke:
Hm, which one exactly?

nLight:
Like all of them

nLight:
Though I might have tested it wrong. It is not always clear which magic shell command I need to run 😩

system-tests/services/test-apps.js
weblancaster: yep I noticed that when writing tests again after `prettier` was merged.

Thanks @juliangieseke @nLight

weblancaster:
retest this please

weblancaster:
Can be merged once the tests are all ✅

PR #2156, opened 2017-05-03
DCOS-15287: Add checks to verify marathon response (jobs)

juliangieseke:
Add additional checks to verify that the Marathon response is as expected. To do so the test will open the service detail, hit edit and check the form content.

### DIFF
--- a/system-tests/jobs/test-jobs.js
+++ b/system-tests/jobs/test-jobs.js
@@ -7,7 +7,7 @@ describe("Jobs", function() {
     });
   });
 
-  it("Create a simple job", function() {
+  it("should create a simple job", function() {
     const jobName = "job-with-inline-shell-script";
     const fullJobName = `${Cypress.env("TEST_UUID")}.${jobName}`;
     const cmdline = "while true; do echo 'test' ; sleep 100 ; done";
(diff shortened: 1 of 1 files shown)


system-tests/jobs/test-jobs.js
```diff
@@ -7,7 +7,7 @@ describe("Jobs", function() {
     });
   });
 
-  it("Create a simple job", function() {
+  it("should create a simple job", function() {
```
weblancaster: is it really necessary? I believe the wording `should` is not a worth "pattern". Not a big deal though.

weblancaster:
is this a duplicated of #2155?

juliangieseke:
We've split them up into three PRs, one per file.

system-tests/jobs/test-jobs.js
juliangieseke: @orlandohohmeier

orlandohohmeier: @weblancaster using `should` here makes perfect sense as it aligns the system tests with our unit and integration tests. If you feel we should abandon the pattern, please add an item to our team sync agenda to discuss the topic.

orlandohohmeier:
@juliangieseke looks like you lost parts of the changes during the rebase as there aren't any additional checks to verify the marathon response.

juliangieseke:
here they are. sorry.

system-tests/jobs/test-jobs.js
```diff
@@ -1,13 +1,7 @@
 const { Timeouts } = require("../_support/constants");
 
 describe("Jobs", function() {
-  afterEach(() => {
-    cy.window().then(win => {
-      win.location.href = "about:blank";
-    });
-  });
```
juliangieseke: dang it! i'll add it again.

system-tests/jobs/test-jobs.js
weblancaster: ok cool.

nLight: I'm with @weblancaster 

I think we could learn from ruby http://www.betterspecs.org/ I also tend to write what a system _does_ instead of what it _should_ do. So I'd write `it("creates a simple job")` let's team sync on this one.

PR #2157, opened 2017-05-03
[marathon 1.5] Make docker network transactions JSON paths agnostic

nLight:
**TL;DR** Batch transaction paths `container.docker.network` -> `networks.0.mode`

This PR prepares the base for the new Network API.

Currently Batch transactions paths reflect JSON structure and with this PR we are switching them to be more generic.

As a start of the migration, we switch from a single network to an array of networks hardcoded to handle the first element only.

## Background

Currently the UI supports only one network for a single container app, defined in `container.docker.network` there's a migration coming with marathon 1.5, networking API will be unified with Pods allowing a user to have more than one network per app. To achieve this, Marathon team reorganized the app definition moving networking information from `container.docker.network` on the top level and making it to be an array.

Ioannis created a [doc](https://docs.google.com/document/d/1CpUq51q5_AsZubmcWVXu5pAGaS2D7NXXhrwmWtPAyXU) to describe all the differences.

> While transactions still look to be bound to the JSON structure the JSON structure itself will become more generic, therefore transactions are not about _where_ the data resides but about _what_ it describes.

### DIFF
--- a/plugins/services/src/js/components/forms/NetworkingFormSection.js
+++ b/plugins/services/src/js/components/forms/NetworkingFormSection.js
@@ -473,7 +473,7 @@ class NetworkingFormSection extends mixin(StoreMixin) {
       .join(", ");
 
     const selections = (
-      <FieldSelect name="container.docker.network" value={network}>
+      <FieldSelect name="networks.0.mode" value={network}>
         <option disabled={Boolean(disabledMap[HOST])} value={HOST}>
           Host
         </option>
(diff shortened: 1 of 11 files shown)


MatApple:
From my limited understanding of the problem, this looks reasonable to me. @nLight - are we able to create a cluster to test this? If so, perhaps we could create a long running cluster or post the details here so others can test PRs. Thoughts?

nLight:
Actually you've got me thinking @MatApple. So this PR doesn't change anything and still works with Marathon 1.4 shall we merge it to master maybe as a general improvement and rebase `integrate/marathon-1.5`? I also wrote a bit more background in the description.

@orlandohohmeier

Poltergeist:
LGTM! 💯  🎆

PR #2144, opened 2017-04-29
[SDK Actions 1 of 2] DCOS-14910: Disable sdk service actions

mlunoe:
This PR introduces new modals for SDK services and applies them only to the services in the service table.

1 of 2: https://github.com/dcos/dcos-ui/pull/2144
2 of 2: https://github.com/dcos/dcos-ui/pull/2160

Test this by adding this repository for packages and installing the Hello Wolrd package:
https://infinity-artifacts.s3.amazonaws.com/autodelete7d/hello-world/20170501-114415-H2EXdmN1BlhErgpw/stub-universe-hello-world.zip

This PR merges to a feature branch and is not concerned with:
- group actions
- service detail actions
- pod detail actions
- service configuration
These will come in later PRs

(Preview doesn't seem to work 😢  Here is the link: https://cl.ly/kDyf/Screen%20Recording%202017-04-28%20at%2010.22.gif)

**Checklist**
- [x] Did you add a JIRA issue in a commit message or as part of the branch name?
- [ ] Did you add new unit tests?
- [x] Did you add new integration tests?
- [ ] If this is a regression, did you write a test to catch this in the future?

P.S. Don't be scared by the huge amount of additions. I took the liberty and created test for the service table with the same action tests that we already have for service detail.

PR #2160, opened 2017-05-04
[SDK Actions 2 of 3] DCOS-14910: Disable sdk group actions

mlunoe:
This PR introduces new modals for groups containing SDK services in the service table.

~~1 of 3: https://github.com/dcos/dcos-ui/pull/2144~~ Merged
2 of 3: https://github.com/dcos/dcos-ui/pull/2160
3 of 3: https://github.com/dcos/dcos-ui/pull/2163

Test this by adding this repository for packages and installing the Hello Wolrd package using the advanced install, changing the id to be installed in a group:
https://infinity-artifacts.s3.amazonaws.com/autodelete7d/hello-world/20170501-114415-H2EXdmN1BlhErgpw/stub-universe-hello-world.zip

After that, you can install a regular application in the same group.

This PR merges to a feature branch and is not concerned with:
- service detail actions
- pod detail actions
- service configuration
These will come in later PRs

**Checklist**

~~Diff between the two PRs: https://github.com/dcos/dcos-ui/compare/mlunoe/DCOS-14910-disable-sdk-service-actions...mlunoe/DCOS-14910-disable-sdk-group-actions~~ Merged


plugins/services/src/js/components/modals/ServiceActionDisabledModal.js
```diff
@@ -209,20 +217,229 @@ class ServiceActionDisabledModal extends React.Component {
...
+        memo.otherServices.push(item);
+      }
+
+      return memo;
+    }, { selectedServices: [], otherServices: [] });
+  }
+
+  getServiceList(servies) {
+    return servies.map(function(item, index) {
+      const comma = index === servies.length - 1 ? "" : ", ";
```
jfurrow: `servies` is a cute name but we should probably uses `services`

mlunoe: Done

jfurrow:
I think it'd be nice if we remove the `modal-large` class from the modal, so that it's a little more narrow:

src/js/translations/en-US.json
```diff
@@ -25,13 +25,34 @@
   "SERVICE_ACTIONS.SCALE_BY": "Scale By",
   "SERVICE_ACTIONS.SUSPEND": "Suspend",
 
+  "SERVICE_ACTIONS.SDK_GROUP_DELETE_1": "Copy and paste this command into the CLI",
+  "SERVICE_ACTIONS.SDK_GROUP_DELETE_2": "Then, run the",
+  "SERVICE_ACTIONS.SDK_GROUP_DELETE_3": "for each service to clean up reserved resources",
+  "SERVICE_ACTIONS.SDK_GROUP_DELETE_4": "Then",
+  "SERVICE_ACTIONS.SDK_GROUP_DELETE_5": "the group",
+  "SERVICE_ACTIONS.SDK_GROUP_DELETE_6": "Copy the command below into the DC/OS CLI",
```
jfurrow: For anyone wondering why template strings aren't used, it's because we can't use the `FormattedMessage` component here because this text gets rendered inside a Modal, and the Modal is rendered in a separate context than the rest of the app. Therefore we need to use the `formatMessage` function provided by the `injectIntl` HOC, which doesn't allow React components to be rendered in a template string.

jfurrow:
This

jfurrow:
This works great! :ship:

jfurrow:
Still looks good, let's merge when everything's 💚

src/js/utils/__tests__/ServiceUtil-test.js
```diff
@@ -23,4 +32,34 @@ describe("ServiceUtil", function() {
       expect(ServiceUtil.isSDKService(service)).toEqual(false);
     });
   });
+
+  describe("#isPackage", function() {
+    it("should return true if service does not have the proper label", function() {
```
MatApple: Don't think this test it titled correctly for what it's testing.

plugins/services/src/js/components/modals/ServiceActionDisabledModal.js
```diff
@@ -363,7 +363,7 @@ class ServiceActionDisabledModal extends React.Component {
   getGroupDeleteMessage() {
     const { actionID, service, intl } = this.props;
     const serviceID = service ? service.getId() : "";
-    const { selectedServices } = this.getServiceTypes(isPackage);
+    const { selectedServices } = this.getServiceTypes(isSDKService);
```
mlunoe: You need to use the isPackage since we need a different delete command for packages, i.e. clean out zookeeper as well, and for other services, i.e. stuff that has been installed directly with Marathon.
I guess you saw an empty command because we added the flags artificially to a marathon app, and we are here assuming that an SDK service is installed through Cosmos. Maybe we can't assume that, but I think that is ok.

mlunoe: ping @rcorral

MatApple: Thanks for catching this @mlunoe. @orlandohohmeier @nLight - can you please check this before we put up PRs to DC/OS?

orlandohohmeier: @MatApple not sure I understand what I need to check...

MatApple: @mlunoe - could you provide more details as to why `isSDKService` won't work here? It seems reasonable so I'd like to understand what we're missing. Thanks!

mlunoe: @MatApple here is some more information, hope it makes sense:
The user action here is to delete a group. In the group where we want to delete services installed from Cosmos (Packages) and apps (Marathon) differently. Packages installed from Cosmos needs to be deleted using the `dcos package uninstall ${packageName} --app-id=${serviceID}`-command and run the cleaner script. Only applications installed directly through Marathon can be removed with the group delete command; `dcos marathon group remove ${groupID}`.

This deviates slightly from what we are doing in general since we just run the Marathon delete command on everything (which we should change), but this is the right thing to do.

PR #2162, opened 2017-05-05
Add table row hover and remove highlighted column styling

leemunroe:
Proposal to improve table styling.

1. Add a row hover style. Why?  Improves usability by visually highlighting the row the user is looking at so they can see the related data and columns.

2. Remove the highlighted active column. Why? Don't think this is adding value (as the caret already shows the active column) and it looks odd whenever there are just two columns.

3. Show the sorting caret on hover. Why? So the user knows they can interact with and sort the table.

Before
https://cl.ly/3Z40263T2c1y

After
https://cl.ly/0U05022x2M3p

cc @ashenden

### DIFF
--- a/src/styles/content/tables/styles.less
+++ b/src/styles/content/tables/styles.less
@@ -1,12 +1,30 @@
 & when (@table-enabled) {
 
+  .table {
+
+    th {
+
+      &.active {
+        background-color: inherit;
+      }
+    }
+
+    td {
+
+      &.active {
+        background-color: inherit;
+      }
+    }
+  }
+
   tr,
   .table-row {
 
     &:hover {
 
-      td {
-        background: @table-cell-hover-color;
+      td,
+      td.active {
+        background-color: @table-cell-hover-color;
       }
 
       .table-display-on-row-hover {
@@ -65,6 +83,16 @@
     font-size: 0.9rem;
     text-transform: uppercase;
     vertical-align: middle;
+
+    &.clickable {
+
+      &:hover {
+
+        .caret {
+          display: inline-block;
+        }
+      }
+    }
   }
 
   td {
@@ -72,6 +100,10 @@
     // property. http://stackoverflow.com/questions/9789723/css-text-overflow-in-a-table-cell
     max-width: 0;
     vertical-align: middle;
+
+    &.active {
+      background-color: inherit;
+    }
   }
 
   .table-header-title {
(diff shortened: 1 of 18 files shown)


ashenden:
Fully support these changes.

src/styles/content/tables/styles.less
```diff
@@ -1,10 +1,26 @@
...
+    }
+  }
+
   tr,
   .table-row {
 
     &:hover {
 
+      td,
+      td.active {
```
jfurrow: Sorry for the nit-pick but let's write this is as follows:

```LESS
td {

  &,
  &.active {
    background-color: @table-cell-hover-color;
  }
}
```

leemunroe: Sounds good. Done.

ashenden:
@leemunroe @jfurrow you should be able to achieve the row hover behavior without modifying the CSS.  CNVS provides a class for this called `.table-hover` (http://mesosphere.github.io/cnvs/content/tables/)

leemunroe:
That would mean adding another class to every instance of a table (~10 JS files) and developers having to remember to do that in the future for new tables VS changing one CSS file. Seems like bad practice, no? Isn't the ideal world a reusable semantic class like  `table-collection` or `table-data` instead of adding more classes?

jfurrow:
I agree with @leemunroe that this should be added by default to all tables.

ashenden:
@jfurrow @leemunroe agree, but aren't developers already today using a generalized table component?  One where they don't have to think about classes?  If so, on this generalized table component, could we not define `table-hover` once and then it is just applied to all instances of that table going forward?

What I'm suggesting we avoid is having to manually change the style definition, essentially adding more style, when this is already support and available to us.

jfurrow:
The generalized table component is in `reactjs-components` and we shouldn't add the class there IMO.

ashenden:
@jfurrow think we could talk about this in person tomorrow? I believe we want the same thing (maybe?).

jfurrow:
For sure! Grab me when you're available

MatApple:
@jfurrow @leemunroe what was the outcome of your offline discussion?

leemunroe:
Added `table-hover` class to all relevant tables as per offline discussion.

jfurrow:
@MatApple We decided, for now, to just add the `table-hover` classes to all `Table` components with interactive elements.

jfurrow:
Thanks @leemunroe! Works great.

PR #2166, opened 2017-05-08
DCOS-15410: Support new Networking API

nLight:
~~ℹ️ Depends on #2157~~ Merged!

## TL;DR
* This PR drops support for the old Networking API and adds support for the new version. 
* `ipAddresses` field is deprecated thus this PR drops the support.
* The new API adds support for multiple networks, but the UI will support only one network. 

## Background

Ioannis created a [doc](https://docs.google.com/document/d/1CpUq51q5_AsZubmcWVXu5pAGaS2D7NXXhrwmWtPAyXU) explaining all the differences.

Currently the UI supports only one network for a single container app, defined in `container.docker.network` there's a migration coming with marathon 1.5, networking API will be unified with Pods allowing a user to have more than one (user-)network per app. To achieve this, Marathon team reorganized the app definition moving networking information from `container.docker.network` on the top level and making it to be an array.


nLight:
run integration tests

jfurrow:
This works as expected! Great work @nLight.

PR #2167, opened 2017-05-08
DCOS-15568: show warning on multiple networks

Poltergeist:
Disable the networking form if more then one
network is set. If this happens the form section
will only show the JSON from `networks`.

image

**Checklist**

Closes DCOS-15568

### DIFF
--- a/plugins/services/src/js/components/forms/NetworkingFormSection.js
+++ b/plugins/services/src/js/components/forms/NetworkingFormSection.js
@@ -608,6 +608,8 @@ class NetworkingFormSection extends mixin(StoreMixin) {
       "networks"
     );
 
+    let { networks } = this.props.data;
+
     const tooltipContent = (
       <span>
         {
@@ -621,6 +623,29 @@ class NetworkingFormSection extends mixin(StoreMixin) {
         </a> for more information.
       </span>
     );
+
+    if (networks != null && networks.length > 1) {
+      networks = networks.map(function({ mode, name }) {
+        return {
+          name,
+          mode: Networking.internalToJson[mode]
+        };
+      });
+
+      return (
+        <div>
+          <h2 className="flush-top short-bottom">
+            Networking
+          </h2>
+          <FieldLabel>
+            Unable to edit Networking
+          </FieldLabel>
+          <pre>
+            {JSON.stringify(networks, null, 2)}
+          </pre>
+        </div>
+      );
+    }
 
     return (
       <div>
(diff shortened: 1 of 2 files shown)


Poltergeist:
@leemunroe @tobilg Could we maybe find a better description text for this issue?

image

leemunroe:
@Poltergeist Let me make sure I understand this scenario. To get to this point, the user has entered some JSON that contains two networks (which we now support) but this isn't typical, which is why we're only supporting it in JSON mode. Is that accurate?

I would suggest:

1. Change the message to `This service has advanced networking configuration, which we don't currently support in the UI. Use the JSON editor to make changes.`

2. Change the format of the message to a simple text paragraph. Looks like it is currently formatted as an input label which is why it is all caps.

Poltergeist:
Thanks lee this is exactly what is happening here.

I added your suggestions this is an updated screenshot:
image

plugins/services/src/js/components/forms/NetworkingFormSection.js
```diff
@@ -622,6 +624,31 @@ class NetworkingFormSection extends mixin(StoreMixin) {
...
+      });
+
+      return (
+        <div>
+          <h2 className="flush-top short-bottom">
+            Networking
+          </h2>
+          <p>
+            {"This service has advanced networking configuration, which we" +
+              " don't currently support in the UI. Use the JSON editor to make" +
```
nLight: Let's be more polite: "Please use the JSON editor..." ❤️

plugins/services/src/js/components/forms/NetworkingFormSection.js
```diff
@@ -622,6 +624,31 @@ class NetworkingFormSection extends mixin(StoreMixin) {
...
+        };
+      });
+
+      return (
+        <div>
+          <h2 className="flush-top short-bottom">
+            Networking
+          </h2>
+          <p>
+            {"This service has advanced networking configuration, which we" +
```
nLight: Shall we use `i18n` here? @weblancaster is doing great job pushing it forward 💪

nLight: as discussed off line: there's an issue with our setup so we couldn't just add i18n here, but since this is a quick fix to be removed in the following sprints we decided not to waste any more time here.

jfurrow: Thanks for following up! Makes sense.

weblancaster: 👍

jfurrow:
run integration tests

PR #2170, opened 2017-05-09
🔥 dead old ServiceFormModal code

nLight:
Rationale: while migrating Networking, I stumbled upon dead code I had to wrap my head around if I have to adjust it or not. So I'd say it's time to delete it. 

I'm pretty sure I didn't remove all the related dead code, but let's say we start here and see how much more we can delete.


plugins/services/src/js/utils/ServiceUtil.js
```diff
@@ -95,387 +61,6 @@ const ServiceUtil = {
...
-        memo[key] = null;
-      }
-
-      return memo;
-    }, {});
-
-    return ServiceUtil.createSpecFromDefinition(definition);
-  },
-
   createFormModelFromSchema(schema, service = new Application()) {
```
nLight: I think Jobs section still uses it

orlandohohmeier: Ok. Then let's merge this one!

PR #2141, opened 2017-04-28
Refactor reducers

nLight:
An attempt to rescue #2076 
I squashed all the commits and rebased.

-----

Move FormReducer, JSONReducer and JSONParser to
their own file.

Close DCOS-14630

Move JSONParser to JSONReducer of artifacts

Move JSONParser to JSONReducer file to have one
complete Workflow in one file.

Move IpAddress reducer

In the process create tests for the ipAddress
reducer

close DCOS-14695

Move External Volumes reducer

Split up Constraints reducer

Move form Reducer to form reducer

Move Labels reducer

close DCOS-14696

Move Healtchecks Reducers

Move require ports reducer

close DCOS-14729

Move Residency Reducer

close DCOS-14728

Refactor volumes reducer

Remove `FormReducer` as the form uses the specific
typed reducers like external volumes, host volumes
and local volumes.

close DCOS-14727

Move Environment variables reducer

style(ServiceFormReducers): Remove typos

This PR removes typs from the Environment
Variables reducer test cases.

split local volumes reducer

close DCOS-14967

PR #2171, opened 2017-05-09
🚨 Regression: Support pods environment

nLight:
This PR reenables support of Pods `environment` parsing that went missing after #2141


jfurrow:
Works well! Let's :shipit: when tests pass!

plugins/services/src/js/reducers/serviceForm/JSONReducers/EnvironmentVariables.js
```diff
@@ -59,11 +59,13 @@ module.exports = {
   },
 
   JSONParser(state) {
-    if (state.env == null) {
+    if (state.env == null && state.environment == null) {
       return [];
     }
 
-    return Object.keys(state.env).reduce(function(memo, key, index) {
+    const env = state.env || state.environment;
```
orlandohohmeier: Why do we need to parse `environment` and `env`?

jfurrow: This is used for both multi-container and single-container services, and environment variables for multi-containers use the `environment` key while single-container services use the `env` key.

nLight: @orlandohohmeier the original solution utilized separate reducer for Multicontainer. Basically, John nailed it, Pods use `environment` where's apps `env` Since this reducer is being used in both cases we need to check both. We actually should have an integration test for that no idea how has this regression happened. 

But since the solution is trivial, you're right, we split reducers and write some integration tests around 💪

nLight:
@orlandohohmeier thanks for blocking this one. Turns out there was a separate file `MultiContainerEnvironmentVariables `, only import was incorrect. I also added an integrity test that parses a definition, reduces it back and compares JSON. JSON shouldn't change for the fields reducers are aware of.

plugins/services/src/js/reducers/serviceForm/JSONReducers/EnvironmentVariables.js
nLight: Done!

plugins/services/src/js/reducers/__tests__/JSONMultiContainer-test.js
```diff
@@ -0,0 +1,84 @@
+const Batch = require("#SRC/js/structs/Batch");
+const { combineParsers } = require("#SRC/js/utils/ParserUtil");
+const { combineReducers } = require("#SRC/js/utils/ReducerUtil");
+const JSONMultiContainerParser = require("../JSONMultiContainerParser");
+const JSONMultiContainerReducers = require("../JSONMultiContainerReducers");
+
+describe("JSONMultiContainer", function() {
+  describe("integrity", function() {
+    it("parse and reduce doesn't change JSON", function() {
```
orlandohohmeier: Please adjust the test description to be inline with the rest of out test cases... For Example: "it **should** parse and reduce without changing the JSON". Not a blocker though...

jfurrow: Done!

plugins/services/src/js/reducers/__tests__/JSONMultiContainer-test.js
```diff
@@ -0,0 +1,84 @@
+const Batch = require("#SRC/js/structs/Batch");
+const { combineParsers } = require("#SRC/js/utils/ParserUtil");
+const { combineReducers } = require("#SRC/js/utils/ReducerUtil");
+const JSONMultiContainerParser = require("../JSONMultiContainerParser");
+const JSONMultiContainerReducers = require("../JSONMultiContainerReducers");
+
+describe("JSONMultiContainer", function() {
```
jfurrow: I do not see a `JSONMultiContainerParser`, so I left the tests in the current file.

As for your comment about going from `Batch` to `JSON` to `Batch` again... I added a test for what I think this means, but I could be wrong. Effectively the test now does this: `JSON` -> `Batch` -> `JSON` -> `Batch` -> `JSON`

nLight: I'd rather not have it as a part of `JSONMultiContainerParser` because then we should check transactions and my goal was to check how Parser and Reducer work together and if JSON stays intact. 

We could have a separate test for `JSONMultiContainerParser` though.

orlandohohmeier: I was thinking that it might make sense to check the parse to reducer (JSON>Batch>JSON) flow in one file and the reducer to parser (Batch>JSON>Batch) in another. But I'm fine with whatever you favor.

orlandohohmeier:
Great Job! ⚒ 💯

PR #2175, opened 2017-05-09
DCOS-15173: Fix incorrect running task count

jfurrow:
This PR...
* introduces a `getTaskCount` method on the `Service` and `ServiceTree` structs which tallies the number of tasks as reported by Marathon
* implements the `getTaskCount` method for the `ServiceTable` and `ServiceBreadcrumbs` components

Before:

After:

**Checklist**

### DIFF
--- a/plugins/services/src/js/components/ServiceBreadcrumbs.js
+++ b/plugins/services/src/js/components/ServiceBreadcrumbs.js
@@ -129,13 +129,19 @@ class ServiceBreadcrumbs extends React.Component {
       return null;
     }
 
+    let taskCountDetails = `(${runningTasksCount})`;
+
+    if (runningTasksCount !== instancesCount) {
+      taskCountDetails = `(${runningTasksCount} of ${instancesCount})`;
+    }
+
     return (
       <BreadcrumbSupplementalContent
         ref={ref => (this.breadcrumbStatusRef = ref)}
       >
         <BreadcrumbSupplementalContent>
           <span className="muted">
-            {serviceStatus} ({runningTasksCount} of {instancesCount})
+            {serviceStatus} {taskCountDetails}
           </span>
           <ServiceStatusWarningWithDebugInformation item={service} />
         </BreadcrumbSupplementalContent>
--- a/plugins/services/src/js/components/ServiceBreadcrumbs.js
+++ b/plugins/services/src/js/components/ServiceBreadcrumbs.js
@@ -129,11 +129,9 @@ class ServiceBreadcrumbs extends React.Component {
       return null;
     }
 
-    let taskCountDetails = `(${runningTasksCount})`;
-
-    if (runningTasksCount !== instancesCount) {
-      taskCountDetails = `(${runningTasksCount} of ${instancesCount})`;
-    }
+    const taskCountDetails = runningTasksCount === instancesCount
+      ? `(${runningTasksCount})`
+      : `(${runningTasksCount} of ${instancesCount})`;
 
     return (
       <BreadcrumbSupplementalContent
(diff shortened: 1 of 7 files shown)


jfurrow:
Sorry @weblancaster I pushed one more commit based on feedback from @leemunroe.

plugins/services/src/js/components/ServiceBreadcrumbs.js
```diff
@@ -121,21 +121,27 @@ class ServiceBreadcrumbs extends React.Component {
...
     const instancesCount = service.getInstancesCount();
     const isDeploying = serviceStatus === "Deploying";
 
     if (instancesCount === 0) {
       return null;
     }
 
+    let taskCountDetails = `(${runningTasksCount})`;
+
+    if (runningTasksCount !== instancesCount) {
```
weblancaster: Works great @jfurrow !

What do you think about using ternary? I believe less explicit conditions keep the code cleaner. not a big deal though.
```
const taskCountDetails = runningTasksCount !== instancesCount
      ? `(${runningTasksCount} of ${instancesCount})`
      : `(${runningTasksCount})`;
```

plugins/services/src/js/containers/services/ServicesTable.js
```diff
@@ -266,13 +266,14 @@ class ServicesTable extends React.Component {
...
-
+    const tasksRunning = service.getTaskCount();
     const isDeploying = serviceStatus === "Deploying";
 
-    const conciseOverview = ` (${tasksRunning}/${instancesCount})`;
+    let conciseOverview = ` (${tasksRunning})`;
     let verboseOverview = ` (${tasksRunning} ${StringUtil.pluralize("Instance", tasksRunning)})`;
+
     if (tasksRunning !== instancesCount) {
+      conciseOverview = ` (${tasksRunning}/${instancesCount})`;
```
weblancaster: The code was like this already here but would be nice to refactor to ternary too.
```
const conciseOverview = tasksRunning !== instancesCount
      ? ` (${tasksRunning}/${instancesCount})`
      : ` (${tasksRunning})`;
    const verboseOverview = tasksRunning !== instancesCount
      ? ` (${tasksRunning} of ${instancesCount} Instances)`
      : ` (${tasksRunning} ${StringUtil.pluralize("Instance", tasksRunning)})`;
```

jfurrow:
@weblancaster Thanks for the review! I implemented the ternaries you suggested, which dismissed your previous approval.

plugins/services/src/js/structs/ServiceTree.js
```diff
@@ -388,6 +388,10 @@ module.exports = class ServiceTree extends Tree {
 
   getTaskCount() {
     return this.flattenItems().getItems().reduce(function(memo, service) {
+      if (service instanceof ServiceTree) {
```
MatApple: `#flattenItems` should recursively flatten the tree. Did you run in to an error or are you being extra cautious? :)

jfurrow: `#flattenItems` does flatten it into a single array, but the array also includes the `ServiceTree` instances, so the accumulated total was incorrect.

MatApple: Ah I see - you are correct. :+1:

plugins/services/src/js/structs/__tests__/Service-test.js
```diff
@@ -32,6 +32,32 @@ describe("Service", function() {
...
+  describe("#getTaskCount", function() {
+    it("returns the number of reported tasks", function() {
+      const service = new Service({
+        tasks: [{ foo: "bar" }, { bar: "baz" }]
+      });
+
+      expect(service.getTaskCount()).toEqual(2);
+    });
+
+    it("returns the number of reported tasks", function() {
```
MatApple: Let's rename this test to something different from the one above. It'll be easier to read the test from the test reporter to understand what it was testing and to find the exact test in code (being unique).

Perhaps `returns 0 when task array is empty`. We already have context for the method given the parent `describe` so we should name this something relevant to the test case.

jfurrow: Good catch! I totally agree, it's fixed now!

PR #2180, opened 2017-05-11
DCOS-15073: add external volume system test

juliangieseke:
This PR adds a new system test that creates a single container service with external volume and tests if the volume is persistent after the service has been suspended and resumed. A separate test suite for external volumes was created. A script is used to delete all the external volumes after all the tests have run.

To run:
```
  -v `pwd`:/dcos-ui \
  -e REGION=$REGION \
  -e ACCESS_KEY_ID=$ACCESS_KEY_ID \
  -e SECRET_ACCESS_KEY=$SECRET_ACCESS_KEY \
  mesosphere/dcos-ui dcos-system-test-driver -v --only "external-volumes"  \
```

`REGION`: AWS Region
`ACCESS_KEY_ID`: AWS Access Key ID
`SECRET_ACCESS_KEY`: AWS Secret Access Key
~~`NUM_VOLS_DELETE`: Number of Volumes to Delete (More or less corresponds to the number of test cases)~~

See `IntegrationTests` secure notes for credentials. `REGION` is as it appears in the `CLUSTER_URL`.

To test:

- Run the above command and verify the results of the test cases
- Observe that the volumes are deleted from the "Removing XX Volumes in Region..." message during the teardown phase


system-tests/services/test-external-volumes.js
```diff
@@ -0,0 +1,62 @@
...
+        .get(".dropdown-menu-items")
+        .contains("Resume")
+        .click();
+
+      cy.contains("button", "Resume Service").click();
+
+      // link is partly covered by another one, so we have to force it.
+      cy
+        .contains(`${Cypress.env("TEST_UUID")}_external-volumes-single`)
+        .click({ force: true });
```
weblancaster: Makes deeply sad we need to use `force` this takes as to more flaky tests.

system-tests/services/test-external-volumes.js
```diff
@@ -0,0 +1,62 @@
...
+   */
+  describe("External Volumes", function() {
+    afterEach(() => {
+      cy.window().then(win => {
+        win.location.href = "about:blank";
+      });
+    });
+
+    it("should be persistent after suspension in single container services", function() {
+      cy.visitUrl(
```
weblancaster: let's move this to `beforeEach`

system-tests/services/test-external-volumes.js
```diff
@@ -0,0 +1,62 @@
...
+      // link is partly covered by another one, so we have to force it.
+      cy
+        .contains(`${Cypress.env("TEST_UUID")}_external-volumes-single`)
+        .click({ force: true });
+
+      cy.contains("Logs").click();
+
+      cy.contains(`TEST_OUTPUT_${Cypress.env("TEST_UUID")}`);
+
+      cy.visitUrl(
```
weblancaster: seems like this entire `it` unit should be split into multiple units.

system-tests/services/test-external-volumes.js
juliangieseke: in this case both visitUrl calls are shortcuts. I agree that this testcase could be split up, but as it is written for now its only that one block. when we add more to that file it might be a good idea to move it though.

nLight:
The test is failing, but I think you are on it already.

system-tests/services/_scripts/setup-external-volumes
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+
+# create simple app
+cat <<EOF | dcos marathon app add
```
nLight: Can we add this output to results? Feels like dcos fails to start the app and then the tests fails because there's no app, but no way to debug this.

juliangieseke: hm, @wavesoft ?!

wavesoft: When running in verbose mode, the driver logs the STDOUT/STDERR in the console. I think we are already running the job in verbose mode in Jenkins.

Look for `Running lifecycle script "setup"` and the `E:` (stderr) or `O:` (stdout) lines.

juliangieseke: unfortunately that didnt help. script is executed and no error is reported, but it doesnt work.

nLight: Could it be a Marathon error?

juliangieseke:
well, looks like I was trying to load the wrong url.. now all tests are passing locally. please try again & sorry.

system-tests/services/_scripts/setup-external-volumes
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
```
tnguyen0: I don't think we need the script since we can use the setJSON helper.

system-tests/services/test-external-volumes.js
tnguyen0: Can you elaborate?

system-tests/services/test-external-volumes.js
tnguyen0: I tried a bunch of things to avoid using `force: true`, but it seems unavoidable in this case.

system-tests/services/test-external-volumes.js
tnguyen0: I'm not convinced we need a separate test suite for external volumes here. I feel a more general test suite for just volumes would be better.

tnguyen0:
This depends on: https://github.com/dcos/dcos-ui/pull/2240

system-tests/services/_scripts/setup-external-volumes
orlandohohmeier: @tnguyen0 Let's please not use the `setJSON` helper to provision/set up the cluster for the respective test.

system-tests/services/_scripts/external-volumes/volclean.py
```diff
@@ -0,0 +1,213 @@
+#!/usr/bin/env python3
```
tnguyen0: Taken from here: https://github.com/lloesche/awsvolclean

system-tests/services/_scripts/external-volumes/teardown
```diff
@@ -0,0 +1,35 @@
...
+
+cmd="./services/_scripts/external-volumes/volclean.py \
+-y -r $REGION -k $ACCESS_KEY_ID -s \"$SECRET_ACCESS_KEY\" \
+--ignore-metrics --tags 'Name:^integration-test-dcos-ui-$TEST_UUID'"
+
+MATCH_STRING="Removing"
+TIMEOUT=30
+counter=0
+
+num_tests=$(($NUM_TESTS))
```
tnguyen0: The purpose of `NUM_TESTS` is to iteratively delete each volume. The volumes must be detached before the script can find the volume. This is done instead of having an arbitrary delay to wait for the volumes to detach.

tnguyen0:
@juliangieseke @weblancaster @ahoskins et. al can you guys take a look?

system-tests/services/_scripts/external-volumes/volclean.py
tnguyen0: Taken from here: https://github.com/lloesche/awsvolclean

system-tests/services/_scripts/external-volumes/teardown
```diff
@@ -0,0 +1,36 @@
...
+    if [[ "$counter" -ge $TIMEOUT ]]; then
+      echo "Timeout exceeded: External volume deletion"
+      exit 1
+    fi
+
+    [[ $result != $MATCH_STRING ]]
+  do
+    :
+  done
+done
```
tnguyen0: The purpose of `NUM_VOLS_DELETE` is to iteratively delete each volume. The volumes must be detached before the script can find the volume. It is done this way instead of having an arbitrary delay to wait for the volumes to detach.

system-tests/services/_scripts/external-volumes/teardown
```diff
@@ -0,0 +1,36 @@
...
+
+cmd="./services/_scripts/external-volumes/volclean.py \
+-y -r $REGION -k $ACCESS_KEY_ID -s \"$SECRET_ACCESS_KEY\" \
+--ignore-metrics --tags 'Name:^integration-test-dcos-ui-$TEST_UUID'"
+
+MATCH_STRING="*Removing*Volumes*"
+TIMEOUT=30
+counter=0
+
+num_vols_delete=$NUM_VOLS_DELETE
```
tnguyen0: The purpose of NUM_VOLS_DELETE is to iteratively delete each volume. The volumes must be detached before the script can find the volume. It is done this way instead of having an arbitrary delay to wait for the volumes to detach.

MatApple:
@juliangieseke @tnguyen0 can you guys please work together to get this in good shape? If you need reviewers to take another look then please ping them 👍

system-tests/services/_scripts/external-volumes/teardown
MatApple: Let's just hard code this and write a comment about what it is. Currently we need to clean up 2 volumes because we create 2 for the tests but if we write more tests we'll need to clean up more.

system-tests/services/_scripts/external-volumes/volclean.py
ahoskins: We can just direct copy-paste something like this with no citation or reference to original source? @MatApple

system-tests/services/_scripts/external-volumes/teardown
```diff
@@ -0,0 +1,39 @@
...
+while [ "$num_vols_delete" -gt 0 ]
+do
+  echo "$num_vols_delete volume(s) left to delete"
+  while
+    result=$(eval "$cmd 2>&1")
+    echo $result
+    remove_string=$(echo $result | sed -n 's/^.*\(\Removing.*[0-9].*\Volumes\).*/\1/p')
+    num_removed=$(echo $remove_string | sed 's/[^0-9]*//g')
+    num_vols_delete=$(($num_vols_delete-${num_removed:-0}))
+    sleep 1s
```
ahoskins: Is this an arbitrary amount of time to sleep? Maybe add a comment to explain why it's here.

system-tests/services/_scripts/external-volumes/volclean.py
tnguyen0: I will add a reference to the original source. Good idea @ahoskins!

system-tests/services/_scripts/external-volumes/teardown
tnguyen0: Yes, I am running the script to check for volumes every second. I will add a comment, thanks @ahoskins!

tnguyen0:
Run integration tests

system-tests/services/_scripts/external-volumes/volclean.py
```diff
@@ -0,0 +1,219 @@
...
```
weblancaster: This file concerns me because we have copied and pasted here, what happen when there is a change/update/security fix to the original file? we won't update this ever.

Can't we create a python file that import the original and uses it? That would be the ideal.

system-tests/services/test-external-volumes.js
```diff
@@ -0,0 +1,257 @@
...
+   * WHENEVER YOU ADD OR REMOVE A TEST CASE
+   */
+  describe("External Volumes", function() {
+    afterEach(() => {
+      cy.window().then(win => {
+        win.location.href = "about:blank";
+      });
+    });
+
+    it("should create an app with external volume", function() {
```
weblancaster: we are removing `should` from tests when writing new ones, this is not a big deal but would be nice not to use it anymore.

system-tests/services/_scripts/external-volumes/volclean.py
tnguyen0: Your concern is valid @weblancaster. I think we can just merge it as it is right now and revisit it later in the future. I will need to speak to @lloesche in order for him to make changes on his side. For now, I will add a TODO comment to the script.

MatApple: While it's a valid point, I think we should merge this as is. Let's just make sure we put attribution at the top of the file and the authors credentials (in this case Lucas - who works at Mesosphere). That way we know who to contact.

system-tests/services/_scripts/external-volumes/setup
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash
+
+if [ -z "$REGION" ]; then
```
MatApple: I really like that you've checked for these early and throw an error.

system-tests/services/_scripts/external-volumes/teardown
```diff
@@ -0,0 +1,63 @@
...
+-y -r $REGION -k $ACCESS_KEY_ID -s \"$SECRET_ACCESS_KEY\" \
+--ignore-metrics --tags 'Name:^integration-test-dcos-ui-$TEST_UUID'"
+
+# Used to match against the output of the script
+MATCH_STRING="*Removing*Volumes*"
+
+TIMEOUT=30
+counter=0
+
+# This is the number of volumes we want to delete.
```
MatApple: Comments are 💯

system-tests/services/_scripts/external-volumes/volclean.py
weblancaster: sounds good.

weblancaster: Opened a issue in the repo https://github.com/lloesche/awsvolclean/issues/1

PR #2183, opened 2017-05-12
fix(DCOS-15332): Validate redirect is the same domain as the original hostname

weblancaster:
(**DCOS-15332**) This PR removes the ability to use the redirect in the client side for security reasons.

This is like-a-proposal because I personally did not find any use for it yet after looking in the code and in the jira tickets so I'm assuming this is safe to delete.

**Checklist**

### DIFF
--- a/plugins/oauth/hooks.js
+++ b/plugins/oauth/hooks.js
@@ -201,8 +201,9 @@ module.exports = Object.assign({}, StoreMixin, {
 
   userLoginSuccess() {
     const redirectTo = RouterUtil.getRedirectTo();
+    const isValidRedirect = RouterUtil.isValidRedirect(redirectTo);
 
-    if (redirectTo) {
+    if (isValidRedirect) {
       global.location.href = redirectTo;
     } else {
       ApplicationUtil.beginTemporaryPolling(() => {
--- a/src/js/utils/RouterUtil.js
+++ b/src/js/utils/RouterUtil.js
@@ -17,12 +17,10 @@ function findRedirect(queryString) {
 }
 
 const RouterUtil = {
-  isValidRedirect(domain) {
-    if (!Util.isString(domain)) {
-      return false;
-    }
+  isValidRedirect(url) {
+    const parsedUrl = Util.parseUrl(url);
 
-    return domain.includes(global.location.hostname);
+    return parsedUrl.hostname === global.location.hostname;
   },
 
   getRedirectTo() {
--- a/src/js/utils/RouterUtil.js
+++ b/src/js/utils/RouterUtil.js
@@ -17,6 +17,14 @@ function findRedirect(queryString) {
 }
 
 const RouterUtil = {
+  isValidRedirect(domain) {
+    if (!Util.isString(domain)) {
+      return false;
+    }
+
+    return domain.includes(global.location.hostname);
+  },
+
   getRedirectTo() {
     let redirectTo = false;
 
--- a/src/js/utils/Util.js
+++ b/src/js/utils/Util.js
@@ -212,6 +212,38 @@ const Util = {
 
   isString(str) {
     return typeof str === "string";
+  },
+
+  /**
+   * Parse url string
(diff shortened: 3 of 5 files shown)


MatApple:
@weblancaster - we should definitely keep the redirect parameter. Could we just reject redirect locations that belong to an external domain?

weblancaster:
yes @MatApple that would be the second option.

What's the use for currently (use case)?

nLight:
We might want to add some tests for Utils

weblancaster:
Thanks for caching it @nLight It's addressed.

weblancaster:
retest this please

weblancaster:
@MatApple @nLight tests were added.

plugins/oauth/hooks.js
```diff
@@ -201,8 +201,9 @@ module.exports = Object.assign({}, StoreMixin, {
 
   userLoginSuccess() {
     const redirectTo = RouterUtil.getRedirectTo();
+    const isValidRedirect = RouterUtil.isValidRedirect(redirectTo);
```
nLight: `isValidRedirect` function expects a `domain` as an argument but we provide the whole thing

src/js/utils/RouterUtil.js
```diff
@@ -17,6 +17,14 @@ function findRedirect(queryString) {
 }
 
 const RouterUtil = {
+  isValidRedirect(domain) {
+    if (!Util.isString(domain)) {
+      return false;
+    }
+
+    return domain.includes(global.location.hostname);
```
nLight: Then here we check if the whole thing we've provided _includes_ `location.hostname`

Imagine this:
`global.location.hostname === "my-cloud.acme.com"`
`domain === "http://malicious.domain.com/pwned?my-cloud.acme.com"`

src/js/utils/__tests__/RouterUtil-test.js
```diff
@@ -133,4 +133,28 @@ describe("RouterUtil", function() {
...
+      // Overwrite jsdom global/window location mock
+      Object.defineProperty(global.location, "hostname", {
+        writable: true,
+        value: "localhost"
+      });
+    });
+
+    it("domain in redirect is valid", function() {
+      const expectedResult = true;
+      const domain = "http://localhost:4200/";
```
nLight: Sorry it's not a `domain` that string also contains schema and a port.

weblancaster:
@nLight `address the outstanding potential vulnerability` which are?

plugins/oauth/hooks.js
weblancaster: Correct. We need to pass the entire domain/url that was extracted from redirect query

nLight: There's a difference between a domain and a url, generally speaking I think that we can't really redirect to a _domain_ only to a URL thus we should change the name

Examples:
dcos.io is a domain name
http://dcos.io is a URL

https://en.wikipedia.org/wiki/URL
https://en.wikipedia.org/wiki/Domain_name

weblancaster:
@nLight all comments addressed. If there is a better naming convention you want to follow for the url please suggest.

Thanks for reviewing.

src/js/utils/Util.js
```diff
@@ -208,6 +208,42 @@ const Util = {
...
+  parseUrl(url) {
+    if (!Util.isString(url)) {
+      return null;
+    }
+
+    // TODO: replace below with browser native API <new URL()>
+    // when all browsers support are available
+    const aElement = document.createElement("a");
+    aElement.href = url;
+
```
weblancaster: the reason I'm returning the object here in this schema is because when moving to use the browser native API it's going to return a obj that way we don't break the API for those using `parseUrl` down the road

rcorral:
run integration tests

src/js/utils/Util.js
jfurrow: This is sweet! Nice work.

PR #2186, opened 2017-05-15
↩️ [release/1.9] DCOS_OSS-981: Move ports auto assign checkbox at the top level

nLight:
Detailed explanation: https://jira.mesosphere.com/browse/DCOS_OSS-981
One more issue I fixed accidentally along with it: https://jira.mesosphere.com/browse/DCOS_OSS-1107 😆 

Turned out we were wrong assuming that ports auto assign works per endpoint, in fact it works on the app level. This PR moves the checkbox from the endpoint form to the network form and changes the way this checkbox affects `port` field in a `portDefinition`. When auto assign is checked JSONReducer replaces any value for all the ports with 0, when the checkbox is not checked it emits previous value or 0.

We might want to create a custom validator that will force a user to enter a port in case the auto assign feature is off.

Testing:
Apart from the form interaction. Launch an 1 instance app with auto-assugned ports. Wait until it is running. Edit the app *with the form* bump instances to an arbitrary number, say 6. Save. Verify that all the instances are running, not Waiting.

Visual:

screen shot 2017-05-15 at 14 47 54

/cc @timcharper


Poltergeist:
Assigning a specific port is not working 😢 
ports mov

plugins/services/src/js/components/forms/NetworkingFormSection.js
```diff
@@ -578,6 +571,35 @@ class NetworkingFormSection extends mixin(StoreMixin) {
     );
   }
 
+  getHostNetworkPortsAutoAssignSestion() {
```
Poltergeist: This should be `getHostNetworkPortsAutoAssignSection`

MatApple:
2 tests are still failing

tests/pages/services/ServiceFormModal-cy.js
```diff
@@ -211,11 +211,11 @@ describe('Service Form Modal', function () {
       killSelection: 'YOUNGEST_FIRST',
       portDefinitions: [
         {
-          port: 10000,
+          port: 0,
           protocol: 'tcp'
         }
       ],
-      requirePorts: true
+      requirePorts: false
```
nLight: We should never change `requirePorts` basing on a port value,
in `_fixtures/marathon-1-task-with-columes/groups.json`
we see 
```
{
...
      "ports": [
        10000
      ],
      "portDefinitions": [
        {
          "port": 10000,
          "protocol": "tcp",
          "labels": {
            
          }
        }
      ],
      "requirePorts": false,
...
}
```

Port value 10000 has been auto assigned because `"requirePorts": false` and Marathon will return it.
One shouldn't even take into consideration those ports and use env vars $PORT0...$PORT<n>

nLight:
`Hold merge` because it causes a regression for other network types. I'd assume we should emit `requirePorts` only for HOST network. Though I'm not sure 😩

nLight:
network != HOST issue should be addressed now.

nLight:
Hold merge once again to address:
* Host networking: all ports are auto-assigned, or none of them are.
* Bridge networking: host ports can individually be specified as auto-assigned, or not
* Container networking: host ports can individually be specified as auto-assigned, or not

nLight:
@orlandohohmeier @Poltergeist I addressed all the issues and rewrote it 🙇

orlandohohmeier:
@tobilg @leemunroe can you please have a look and check if the wording is ok?

plugins/services/src/js/reducers/JSONSingleContainerParser.js
```diff
@@ -13,8 +13,9 @@ import {JSONParser as scaling} from './serviceForm/MultiContainerScaling';
 import {JSONParser as network} from './serviceForm/Network';
 import {JSONParser as multiContainerNetwork} from './serviceForm/MultiContainerNetwork';
 import {JSONParser as volumeMounts} from './serviceForm/MultiContainerVolumes';
-import {simpleParser} from '../../../../../src/js/utils/ParserUtil';
 import {JSONParser as containers} from './serviceForm/Containers';
+import {JSONParser as requriePorts} from './serviceForm/RequirePorts';
```
orlandohohmeier: 🔡 `requirePorts`

orlandohohmeier:
Great Job! ⚒ 💯

PR #2191, opened 2017-05-17
fix(DCOS-15234): Display service ports only when available

weblancaster:
**DCOS-15234** Display service ports only when available under service configuration tab.

**Checklist**

### DIFF
--- a/plugins/services/src/js/service-configuration/ServiceNetworkingConfigSection.js
+++ b/plugins/services/src/js/service-configuration/ServiceNetworkingConfigSection.js
@@ -71,7 +71,8 @@ class ServiceNetworkingConfigSection extends ServiceConfigBaseSectionDisplay {
               name: "name",
               port: "port",
               protocol: "protocol",
-              labels: "labels"
+              labels: "labels",
+              service: "servicePort"
             };
 
             const networkType = getNetworkType(
@@ -86,6 +87,7 @@ class ServiceNetworkingConfigSection extends ServiceConfigBaseSectionDisplay {
               appDefinition,
               "container.docker.portMappings"
             );
+
             if (
               containerPortMappings != null &&
               containerPortMappings.length !== 0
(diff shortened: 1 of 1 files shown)


plugins/services/src/js/service-configuration/ServiceNetworkingConfigSection.js
```diff
@@ -169,6 +171,33 @@ class ServiceNetworkingConfigSection extends ServiceConfigBaseSectionDisplay {
               });
             }
 
+            const containsServicePorts = portDefinitions.filter(function(
```
nLight: Let's use `some` since we don't really need `portDefinitions` that contain a `servicePort`

plugins/services/src/js/service-configuration/ServiceNetworkingConfigSection.js
```diff
@@ -169,6 +171,33 @@ class ServiceNetworkingConfigSection extends ServiceConfigBaseSectionDisplay {
...
+              portDefinition
+            ) {
+              const result = portDefinition.servicePort
```
nLight: So we need to detect if there's at least one `portDefinition` with a `servicePort` that is not `0`
will just `return portDefinition.servicePort != null && portDefinition.servicePort !== 0` work?

plugins/services/src/js/service-configuration/ServiceNetworkingConfigSection.js
weblancaster: Agree.

plugins/services/src/js/service-configuration/ServiceNetworkingConfigSection.js
weblancaster: won't work because `portDefinition.servicePort` is undefined instead of null, the reason I went with the ternary checking for `portDefinition.servicePort` first is to address both cases `null` `undefined`.

weblancaster: but I could I also just do `portDefinition.servicePort && portDefinition.servicePort !== 0`

nLight: `portDefinition.servicePort != null` checks both `null` and `undefined` please note `!=` not `!==`

Wheres `portDefinition.servicePort && portDefinition.servicePort !== 0` will return not strictly boolean value but Truthy and Falsy values instead.

weblancaster: It was addressed already.

nLight:
Feel free to go ahead and merge once the change is implemented and it gets ✅ and I'm not online

PR #2194, opened 2017-05-18
refactor(remove-duplicated): Remove duplicated system integration test

weblancaster:
While working in one of my tasks last week I noticed that we had a lot of identical system integration tests. So this PR removes the system integration duplication.

The identical code/or same test can be found in `ServiceTable-cy.js` and `ServiceActions-cy.js`

**Checklist**


jfurrow:
Good catch!

PR #2197, opened 2017-05-19
Fix incorrect rounding of resource allocation.

MatApple:
Before:

After:

**Checklist**


orlandohohmeier:
Looks good, though I think we should change the rounding to allow at least two decimals in the next iteration.

PR #2198, opened 2017-05-19
DCOS_OSS-981: Host port auto-assign issue

nLight:
This is the master fix of #2186 Host port auto-assign issue 

Detailed explanation: https://jira.mesosphere.com/browse/DCOS_OSS-981
One more issue I fixed accidentally along with it: https://jira.mesosphere.com/browse/DCOS_OSS-1107 😆 

Turned out we were wrong assuming that ports auto assign works per endpoint, in fact it works on the app level. This PR moves the checkbox from the endpoint form to the network form and changes the way this checkbox affects `port` field in a `portDefinition`. When auto assign is checked JSONReducer replaces any value for all the ports with 0, when the checkbox is not checked it emits previous value or 0.

We might want to create a custom validator that will force a user to enter a port in case the auto assign feature is off.

Testing:
Apart from the form interaction. Launch an 1 instance app with auto-assigned ports. Wait until it is running. Edit the app *with the form* bump instances to an arbitrary number, say 6. Save. Verify that all the instances are running, not Waiting.

Visual:

screen shot 2017-05-15 at 14 47 54


nLight:
retest this please

MatApple:
Should we reset the `hostPort` in the form when toggling `assign ports automatically`? The old auto-assigned port value is kept. There is inconsistency between the JSON and the form.

nLight:
@MatApple good catch! 
Though this is not how I'd assume `<FieldInput value={null} />` to work. I had to set the value to an empty string to clear the field. In my understanding `null` _is_ a value, wheres `undefined` _is not_ so I'd expect providing null to the `<FieldInput />` clears it's value.

orlandohohmeier:
auto-assign

Not a blocker but it would be nice to show an error when auto assign is selected and a port other then 0 is defined instead of replacing the value. What do you think?

nLight:
@orlandohohmeier I think that edge cases like that is the another level of UX already. Because now JSON has contradicting values: `requirePorts: false` and a `port: 9999` so possible solutions is to: 1. Say "Hey, fix your JSON" and block everything
2. Uncheck Ports auto-assign
3. Reset the port value

How do we exactly know whats right?

I'm growing in concern that having the split edit view is a good idea. If a user wants/needs to edit JSON the Form should disappear at this point. It should be either Form or JSON editor, not both together @leemunroe

plugins/services/src/js/reducers/serviceForm/FormReducers/SingleContainerPortMappingsReducer.js
```diff
@@ -0,0 +1,76 @@
...
+ * @param {String[]} action.path - location of value
+ * @param {*} action.value - to perform action with
+ * @return {Object[]} new portDefinitions with action performed on it
+ */
+function SingleContainerPortMappingsReducer(state = [], action) {
+  const { type, path = [], value } = action;
+  const [base, index, field, subfield] = path;
+  const joinedPath = path.join(".");
+
+  if (base === "portDefinitions") {
```
orlandohohmeier: Not blocking as you "only" separated the reducers, but we need to reduce the complexity of these reducers as it's really hard to read and understand them.

leemunroe:
Reiterating that port auto assign does not deserve its own heading and section as it is closely related to "Service Endpoints". Is moving this being tracked separately (per [this JIRA](https://jira.mesosphere.com/browse/DCOS_OSS-956?focusedCommentId=103396&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-103396))?

image

leemunroe:
nvm found it :)
https://github.com/dcos/dcos-ui/pull/2216

PR #2199, opened 2017-05-19
test(DCOS-14465): Add multi container tests

weblancaster:
This PR adds system integration tests defined in **DCOS-14465**

**Checklist**

### DIFF
--- a/tests/pages/services/ServiceFormModal-cy.js
+++ b/tests/pages/services/ServiceFormModal-cy.js
@@ -1611,5 +1611,75 @@ describe("Service Form Modal", function() {
         });
       });
     });
+
+    context("Multi-container (pod)", function() {
+      beforeEach(function() {
+        cy
+          .get(".menu-tabbed-item-label")
+          .eq(0)
+          .click()
+          .get(".menu-tabbed-view h2")
+          .contains("Service");
+      });
+
+      it("Should add new container", function() {
+        cy.get(".menu-tabbed-view .button.button-primary-link").eq(3).click();
+      });
+
+      it("Should contain two containers navigation under Services tab", function() {
+        cy.get(".menu-tabbed-view .button.button-primary-link").eq(3).click();
+
+        cy
+          .get(".menu-tabbed-item-label")
+          .eq(0)
+          .siblings()
+          .should("to.have.length", 2);
+      });
+
+      it("Should be right aligned of the parent", function() {
+        cy.get(".menu-tabbed-view .button.button-primary-link").eq(3).click();
+
+        cy
+          .get(".menu-tabbed-item-label")
+          .eq(0)
+          .siblings()
+          .each(function($element) {
+            const $parentLeftPosition = $element
+              .parent()[0]
+              .getBoundingClientRect().left;
+            const $elementLeftPosition = $element[0].getBoundingClientRect()
+              .left;
+            const threshold = 10;
+            const difference = $elementLeftPosition - $parentLeftPosition;
+
+            expect(difference >= threshold).to.equal(true);
+          });
+      });
+    });
+
+    context("Multi-container (pod) Review & Run", function() {
+      beforeEach(function() {
+        // Fill in SERVICE ID
+        cy.get('.form-control[name="id"]').clear().type("/test-review-and-run");
+
+        // Add a second container
+        cy.get(".menu-tabbed-view .button.button-primary-link").eq(3).click();
+
+        // Click review and run
+        cy
+          .get(".modal-full-screen-actions")
+          .contains("button", "Review & Run")
+          .click();
+      });
+
+      it("Should add new container", function() {
+        cy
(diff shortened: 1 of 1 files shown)


weblancaster:
retest this please

tests/pages/services/ServiceFormModal-cy.js
```diff
@@ -1648,5 +1648,75 @@ describe("Service Form Modal", function() {
...
+        cy
+          .get(".menu-tabbed-item-label")
+          .eq(0)
+          .click()
+          .get(".menu-tabbed-view h2")
+          .contains("Service");
+      });
+
+      it("Should add new container", function() {
+        cy.get(".menu-tabbed-view .button.button-primary-link").eq(3).click();
```
nLight: Am I right that here we test if clicking on the button doesn't throw an Exception? Description says "Should add new container" but we don't check that container is added anywhere.

tests/pages/services/ServiceFormModal-cy.js
```diff
@@ -1648,5 +1648,75 @@ describe("Service Form Modal", function() {
...
+      });
+
+      it("Should contain two containers navigation under Services tab", function() {
+        cy.get(".menu-tabbed-view .button.button-primary-link").eq(3).click();
+
+        cy
+          .get(".menu-tabbed-item-label")
+          .eq(0)
+          .siblings()
+          .should("to.have.length", 2);
```
nLight: If we really want to test that clicking that button adds a container we should verify the initial state first.
Like before clicking the button in should `'to.have.length', 1`

Looks like we can leverage chai's `change` assertion 
```javascript
expect(function() {
  cy.get(".menu-tabbed-view .button.button-primary-link").eq(3).click();
}).to.change(cy.get(".menu-tabbed-item-label").eq(0).siblings(), 'length')
``` 

or better `increases`

```javascript
expect(function() {
  cy.get(".menu-tabbed-view .button.button-primary-link").eq(3).click();
}).to.increase(cy.get(".menu-tabbed-item-label").eq(0).siblings(), 'length')
``` 

https://docs.cypress.io/docs/making-assertions#section-available-assertions

tests/pages/services/ServiceFormModal-cy.js
```diff
@@ -1648,5 +1648,75 @@ describe("Service Form Modal", function() {
...
+      });
+
+      it("Should be right aligned of the parent", function() {
```
nLight: Even though it is a requirement in the ticket, I'd rather avoid such tests @jfurrow 
Shall we wait/introduce tests that will match screenshots?

tests/pages/services/ServiceFormModal-cy.js
```diff
@@ -1648,5 +1648,75 @@ describe("Service Form Modal", function() {
...
+        cy.get(".menu-tabbed-view .button.button-primary-link").eq(3).click();
+
+        // Click review and run
+        cy
+          .get(".modal-full-screen-actions")
+          .contains("button", "Review & Run")
+          .click();
+      });
+
+      it("Should add new container", function() {
```
nLight: Test description doesn't match the content. Easiest would be to change the description to something like "Should contain two containers" since that it exactly what this specific test is verifying.

weblancaster: true I had copied the describe from the above test and forgot to change the description.

tests/pages/services/ServiceFormModal-cy.js
weblancaster: I believe this is fine for now since we have been writing this kind of tests for other UI components and specially because we might not even introduce screenshots soon.

tests/pages/services/ServiceFormModal-cy.js
weblancaster: The suggestion doesn't work.

I also believe that having expect it's more like a helper and not something we should be looking to use all the time because cypress API give us enough for majority of tests.

weblancaster:
Your comments are addresse @nLight

tests/pages/services/ServiceFormModal-cy.js
nLight: Yeah, the suggestion was kind of a pseudocode 😅  based on the docs. I'd like to research these API so we could use it. Thanks for pointing that out explicitly.

tests/pages/services/ServiceFormModal-cy.js
nLight: :thisisfine: though fragile and feels wrong. If there's no JIRA for screenshots I'd like to have one. What do our JIRA masters @orlandohohmeier @rcorral think?

orlandohohmeier: @nLight we don't have a JIRA yet, but I'm more than happy to create one :)

weblancaster: @orlandohohmeier you left a comment but you didn't approve or disapprove 😢

PR #2201, opened 2017-05-23
Improve UX for Package Install Beta opt-in

MatApple:
Last minute UX cleanup with some hacky bits to solve SDK's beta optin issue.

Test by installing any beta packages from the catalog.

[See video for demo](https://cl.ly/300i1i423l1I/Screen%20Recording%202017-05-22%20at%2009.01%20PM.gif)

**Needs tests on Utils but please review in meantime as it's a blocker**

**Also need to backport to 1.9**

**Checklist**

### DIFF
--- a/src/js/components/modals/InstallPackageModal.js
+++ b/src/js/components/modals/InstallPackageModal.js
@@ -4,6 +4,7 @@ import mixin from "reactjs-mixin";
 import React from "react";
 import { StoreMixin } from "mesosphere-shared-reactjs";
 
+import BetaOptInUtil from "../../utils/BetaOptInUtil";
 import CosmosErrorHeader from "../CosmosErrorHeader";
 import CosmosErrorMessage from "../CosmosErrorMessage";
 import CosmosPackagesStore from "../../stores/CosmosPackagesStore";
@@ -24,6 +25,7 @@ const PREINSTALL_NOTES_CHAR_LIMIT = 140;
 
 const METHODS_TO_BIND = [
   "getAdvancedSubmit",
+  "handleAcceptBetaTerms",
   "handleChangeTab",
   "handleInstallPackage",
   "handleAdvancedFormChange",
@@ -37,6 +39,7 @@ class InstallPackageModal
     super(...arguments);
 
     this.tabs_tabs = {
+      betaTerms: "BetaTerms",
       defaultInstall: "DefaultInstall",
       advancedInstall: "AdvancedInstall",
       reviewAdvancedConfig: "ReviewAdvancedConfig",
@@ -89,19 +92,29 @@ class InstallPackageModal
         currentTab: "defaultInstall",
         truncatedPreInstallNotes: true
       });
+
+      return;
     }
-    // If Opening
-    if (!props.open && nextProps.open) {
-      // If default - auto-install
-      if (!nextProps.advancedConfig) {
-        this.handleInstallPackage();
-      } else {
-        // Show advanced install
-        this.setState({
-          currentTab: "advancedInstall"
-        });
-      }
+
+    if (nextProps.isBetaPackage) {
+      this.setState({
+        betaTermsAccepted: false,
+        betaOptInProperties: BetaOptInUtil.getProperty(
+          nextProps.cosmosPackage.getConfig()
+        ),
+        currentTab: "betaTerms"
+      });
+
+      return;
     }
+
+    if (nextProps.advancedConfig) {
+      this.setState({ currentTab: "advancedInstall" });
+
+      return;
+    }
+
+    this.handleInstallPackage();
   }
 
   componentDidUpdate() {
@@ -147,6 +160,20 @@ class InstallPackageModal
     this.setState({ currentTab: "packageInstalled" });
   }
 
+  handleAcceptBetaTerms() {
+    if (this.props.advancedConfig) {
+      // Accept terms and move to advanced install tab
+      this.setState({
+        betaTermsAccepted: true,
+        currentTab: "advancedInstall"
+      });
+
+      return;
+    }
+    // Auto-install
+    this.setState({ betaTermsAccepted: true }, this.handleInstallPackage);
+  }
+
   handleAdvancedFormChange(formObject) {
     this.internalStorage_update({ hasFormErrors: !formObject.isValidated });
     this.forceUpdate();
@@ -174,10 +201,15 @@ class InstallPackageModal
   }
 
   handleInstallPackage() {
-    const { cosmosPackage } = this.props;
+    const { betaTermsAccepted } = this.state;
+    const { cosmosPackage, isBetaPackage } = this.props;
     const name = cosmosPackage.getName();
     const version = cosmosPackage.getCurrentVersion();
-    const configuration = this.getPackageConfiguration();
+    let configuration = this.getPackageConfiguration();
+
+    if (isBetaPackage && betaTermsAccepted) {
+      configuration = BetaOptInUtil.setBetaOptIn(configuration);
+    }
 
     CosmosPackagesStore.installPackage(name, version, configuration);
     this.internalStorage_update({ pendingRequest: true });
@@ -293,6 +325,38 @@ class InstallPackageModal
       <span>
         {ellipses} {text}
       </span>
+    );
+  }
+
+  renderBetaTermsTabView() {
+    const { title, description } = this.state.betaOptInProperties;
+
+    return (
+      <div>
+        <div className="modal-body">
+          <div className="horizontal-center">
+            <h2 className="flush-top">
+              {title}
+            </h2>
+            <p className="text-align-center">
+              {description}
+            </p>
+          </div>
+        </div>
+        <div className="modal-footer">
+          <div className="button-collection button-collection-align-horizontal-center flush-bottom">
+            <button className="button" onClick={this.props.onClose}>
+              Cancel
+            </button>
+            <button
+              className="button button-success"
+              onClick={this.handleAcceptBetaTerms}
+            >
+              Accept
+            </button>
+          </div>
+        </div>
+      </div>
     );
   }
 
@@ -498,10 +562,17 @@ class InstallPackageModal
 
   getModalContents() {
     const { currentTab, schemaIncorrect } = this.state;
-    const { cosmosPackage } = this.props;
+    const { cosmosPackage, isBetaPackage } = this.props;
 
     if (schemaIncorrect) {
       return this.getIncorrectSchemaWarning(cosmosPackage);
+    }
+
+    let schema = cosmosPackage.getConfig();
+
+    if (isBetaPackage) {
+      // Remove beta opt-in from schema for better UX
+      schema = BetaOptInUtil.filterProperty(cosmosPackage.getConfig());
     }
 
     const name = cosmosPackage.getName();
@@ -520,7 +591,7 @@ class InstallPackageModal
             packageIcon={cosmosPackage.getIcons()["icon-small"]}
             packageName={name}
             packageVersion={version}
-            schema={cosmosPackage.getConfig()}
+            schema={schema}
             onChange={this.handleAdvancedFormChange}
             getTriggerSubmit={this.getAdvancedSubmit}
           />
@@ -571,6 +642,7 @@ class InstallPackageModal
 
 InstallPackageModal.defaultProps = {
   advancedConfig: false,
+  isBetaPackage: false,
   onClose() {},
   open: false
 };
@@ -578,6 +650,7 @@ InstallPackageModal.defaultProps = {
 InstallPackageModal.propTypes = {
   advancedConfig: React.PropTypes.bool,
   cosmosPackage: React.PropTypes.instanceOf(UniversePackage).isRequired,
+  isBetaPackage: React.PropTypes.bool,
   open: React.PropTypes.bool,
   onClose: React.PropTypes.func
 };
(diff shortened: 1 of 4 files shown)


src/js/utils/BetaOptInUtil.js
```diff
@@ -0,0 +1,151 @@
+import Util from "./Util";
```
MatApple: Need tests...

nLight:
Why this feature requires tree traversal? /o\

MatApple:
@nLight - removed it after confirmation that SDK will guarantee property to exist at a fixed location.

src/js/utils/BetaOptInUtil.js
```diff
@@ -0,0 +1,104 @@
...
+    if (!Util.isObject(config)) {
+      return config;
+    }
+
+    const betaConfig = Util.deepCopy(config);
+
+    let keyIndex = 0;
+    let pointer = betaConfig;
+
+    while (keyIndex < BETA_OPTIN_CONFIG_PATH.length) {
```
MatApple: Sure, we can simplify but it feels strange to hard code this path in this method but have the other paths configured by CONST's at the top. I'll hard code all methods in this case.

src/js/components/modals/InstallPackageModal.js
```diff
@@ -92,6 +95,18 @@ class InstallPackageModal
     }
     // If Opening
     if (!props.open && nextProps.open) {
+      if (nextProps.isBetaPackage) {
```
orlandohohmeier: Let's adjust these condition to reduce the nesting... 

```js
 if (props.open || !nextProps.open) {
  return;
}

if (nextProps.advancedConfig) {
  // ...
  return;
}

if (nextProps.isBetaPackage) {
  // ...
  return;
}

this.handleInstallPackage();
```

src/js/components/modals/InstallPackageModal.js
```diff
@@ -147,6 +162,25 @@ class InstallPackageModal
     this.setState({ currentTab: "packageInstalled" });
   }
 
+  handleAcceptBetaTerms() {
+    // Set accept terms on config model
+
+    if (this.props.advancedConfig) {
```
orlandohohmeier: This method is unnecessary complex. Please adjust the conditions and setState calls to reduce the complexity.

src/js/utils/BetaOptInUtil.js
```diff
@@ -0,0 +1,104 @@
...
+   * @param  {Object} configSchema - cosmos package config schema
+   * @return {Object} Returns new config Object without opt-in property
+   */
+  filterProperty(configSchema) {
+    const filteredConfig = Util.deepCopy(configSchema);
+
+    let keyIndex = 0;
+    let pointer = filteredConfig;
+
+    while (keyIndex < BETA_OPTIN_SCHEMA_PATH.length - 1) {
```
orlandohohmeier: Why are we using a while loop here?

orlandohohmeier: Let's not read the path length over and over again. Please use a variable to _cache_ the value.

orlandohohmeier: Please always update a the variables of a loop condition right away to make sure that we don't create an infinite loop by accident.

src/js/utils/BetaOptInUtil.js
```diff
@@ -0,0 +1,104 @@
...
+      pointer = pointer[BETA_OPTIN_SCHEMA_PATH[keyIndex++]];
+    }
+    if ("required" in pointer) {
```
orlandohohmeier: Does this condition rely on the loop result?

src/js/utils/BetaOptInUtil.js
orlandohohmeier: Please adjust this loop as well.

src/js/utils/BetaOptInUtil.js
```diff
@@ -0,0 +1,104 @@
+import Util from "./Util";
+
+const BETA_OPTIN_CONFIG_PATH = ["service"];
```
orlandohohmeier: I don't think that these constants actually help and would argue that it's these "options" make it even harder to reason about the respective code.

src/js/utils/BetaOptInUtil.js
MatApple: done!

src/js/components/modals/InstallPackageModal.js
```diff
@@ -296,6 +328,36 @@ class InstallPackageModal
...
   }
 
+  renderBetaTermsTabView() {
+    const { title, description } = this.state.betaOptInProperties;
+
+    return (
+      <div>
+        <div className="modal-body">
+          <div className="horizontal-center">
+            <h2>{title}</h2>
```
jfurrow: Let's add `className="flush-top"` to this guy.

src/js/components/modals/InstallPackageModal.js
```diff
@@ -89,19 +92,29 @@ class InstallPackageModal
...
-      if (!nextProps.advancedConfig) {
-        this.handleInstallPackage();
-      } else {
-        // Show advanced install
-        this.setState({
-          currentTab: "advancedInstall"
-        });
-      }
+
+    if (nextProps.isBetaPackage) {
```
rcorral: This could potentially cause a double render if a different prop of the component changed and `nextProps.isBetaPackage` was true and is true with updated props. Can you do something to prevent this?

MatApple:
@nLight @orlandohohmeier - hardcoded all values and removed loops to simplify.

@orlandohohmeier - yes I think removing the option from the schema is the right way to go. If we keep it in there, it does nothing and a user can deselect it if they wish which would submit a value of `false` to cosmos. This is where some of the weirdness comes in - cosmos will still accept `false` because it just checks the field exists. SDK also doesn't check the `value` and only checks for presence. So keeping the property in the schema is strange because a user could deselect it, effectively unaccepting the terms but still successfully deploy.

SDK will need to update it's validation mechanism, because obviously the UI is only one way to submit this data. But for now, this is the only viable way to fix the UX in the UI and that's where most users are running into issues.

src/js/components/modals/InstallPackageModal.js
```diff
@@ -296,6 +328,36 @@ class InstallPackageModal
...
+        <div className="modal-footer">
+          <div className="button-collection button-collection-align-horizontal-center flush-bottom">
+            <button className="button" onClick={this.props.onClose}>
+              Cancel
+            </button>
+            <button
+              className="button button-success"
+              onClick={this.handleAcceptBetaTerms}
+            >
+              Accept
```
rcorral: Could you internationalize these strings, please?

MatApple:
@rcorral @jfurrow - spacing fixed!

src/js/components/modals/InstallPackageModal.js
```diff
@@ -89,19 +92,29 @@ class InstallPackageModal
...
+      return;
     }
+
+    if (nextProps.advancedConfig) {
+      this.setState({ currentTab: "advancedInstall" });
+
+      return;
+    }
+
+    this.handleInstallPackage();
```
rcorral: It feels a little odd to call this method at the end of this lifecycle method. Could you write a comment about why this is needed? This method also ends up calling `#forceUpdate` which we should avoid calling throughout the lifecycle.

MatApple: This modal should be burned and it will be soon. It has a lot of anti-patterns and bad design including the use of `forceUpdate` which is a legacy part of our codebase. We need this method call here auto-install when a user clicks the `deploy` button.

Calling this already existed - I just refactored this method to untangle it a little.

src/js/components/modals/InstallPackageModal.js
MatApple: Thinking about this, I think we might be over-engineering here. If any of the props change, we actually want to fall into this code block again to force a reset in beta-terms related state.

src/js/components/modals/InstallPackageModal.js
MatApple: If feels strange to me having some internationalized strings in this file and some that are not. I vote for consistency here and I also vote for not internationalizing this file now since it is tagged for a heavy refactor.

MatApple:
@rcorral agree that we should write system tests for the entire package install flow and all it's permeations. However, as we have a significant overhaul on the horizon I'd suggest we wait. We currently don't have any system tests for package install.

leemunroe:
Spacing looks good now.

Confirming we don't have a publicly available beta agreement that we can link to?

src/js/components/modals/InstallPackageModal.js
rcorral: A comment would be super helpful.

src/js/components/modals/InstallPackageModal.js
rcorral: `#componentWillReceiveProps` will get called anytime props change. Should you remove the conditional, if what you say is true?

src/js/components/modals/InstallPackageModal.js
MatApple: Ahh I originally read as a Github comment - I'll put a comment in the code.

src/js/components/modals/InstallPackageModal.js
MatApple: No, if `isBetaPackage`, always fall into this block if props have changed.

orlandohohmeier:
@MatApple I don't like that we once more fix API shortcomings on the frontend. I'm ok with this fix as a short term workaround if we have a JIRA to track the needed SDK validation updates and commitment from the SDK team to get this done in the foreseeable future.

/cc @triclambert @nickbp @ssk2

src/js/utils/BetaOptInUtil.js
```diff
@@ -0,0 +1,85 @@
+import Util from "./Util";
+
+// SDK sets/reads this property within package config.
+const BETA_OPTIN_PROPERTY_NAME = "beta-optin";
+const DOT_NOTATION_SCHEMA_PATH = "properties.service.properties.beta-optin";
```
orlandohohmeier: I don't think the `DOT_NOTATION_` prefix is necessary, it's also probably ok to omit `BETA_OPTIN_` as both constants are part of the **BetaOptIn**Util. #style #naming

src/js/utils/BetaOptInUtil.js
```diff
@@ -0,0 +1,85 @@
...
+
+const BetaOptInUtil = {
+  /**
+   * Finds beta-optin property within cosmos package config schema.
+   *
+   * @param  {Object} configSchema - cosmos package config schema
+   * @return {Boolean} Returns true if configSchema contains beta-optin property
+   */
+  isBeta(configSchema) {
```
orlandohohmeier: Let's rename this method to emphasize that this util checks wether a package is beta. Please adjust the comment as well.  #style #naming

src/js/utils/BetaOptInUtil.js
```diff
@@ -0,0 +1,85 @@
...
+    );
+  },
+
+  /**
+   * Removes beta-optin property within config schema
+   *
+   * @param  {Object} configSchema - cosmos package config schema
+   * @return {Object} Returns new config Object without opt-in property
+   */
+  filterProperty(configSchema) {
```
orlandohohmeier: Please rename this util to clearly state that it will "remove" (create a copy omitting the beta property) the beta property from the passed schema. #style #naming

src/js/utils/BetaOptInUtil.js
```diff
@@ -0,0 +1,85 @@
...
+  },
+
+  /**
+   * Inserts beta-optin property at path within config model
+   *
+   * @param  {Object} config - cosmos package config model
+   * @return {Object} Returns new cosmos package config with beta
+   * opt-in property set to true
+   */
+  setBetaOptIn(config) {
```
orlandohohmeier: Please align the name with `filterProperty` as it should be obvious that this method does the opposite. #style #naming

src/js/utils/BetaOptInUtil.js
```diff
@@ -0,0 +1,85 @@
...
+    if (!BetaOptInUtil.isBeta(configSchema)) {
+      return configSchema;
+    }
+
+    const filteredConfig = Util.deepCopy(configSchema);
+    const service = filteredConfig["properties"]["service"];
+
+    // Filter out required property as this will cause validation
+    // errors in the schema form if we remove the beta-optin schema Object
+    if (Array.isArray(service.required)) {
```
orlandohohmeier: Please check if it's possible to simplify this method by utilizing `Util.omit`

orlandohohmeier:
@MatApple thanks for simplifying the implementation by removing the unnecessary loops. Added a few stylistic comments and requests to improve the usability, though none of those should block this PR.

MatApple:
@orlandohohmeier thanks for the thorough review and stylistic suggestions. I agree that this Util could be much cleaner and follow better naming conventions. I would however suggest that too much time has been sunk into this short-term patch and we move forward "as is" to close the immediate blocker. Chris is happy to apply the appropriate fix in the backend but I'm still awaiting a firm commitment and fixVersion. Till then let's keep the `hold-merge`.

MatApple:
There are many outstanding comments on this PR and I do believe in fixing things that are important to the team. Because these are stylistic suggestions and this PR is to provide a temporary patch, I'm going to move forward with the merge. We have 2 accepted reviews and @orlandohohmeier has accepted that his comments should not hold this PR up. We also have an Epic to track the proper fix and improved UX around Beta packages to ensure we can back this diff out once those land.

PR #2202, opened 2017-05-23
test(pods-communication) add tests for basic pod communication

juliangieseke:
Add test case which creates a pod with two containers and tests if a http-client on one can access nginx on the other container

Closes DCOS_15051


juliangieseke:
hm, external tests alre included in this one? thats wrong.

PR #2204, opened 2017-05-23
DCOS_OSS-1132: Adjust the constraints validation to fix validation errors

orlandohohmeier:
Change the placements constraints validation to not return errors for
optional fields.

Closes DCOS_OSS-1132


plugins/services/src/js/validators/__tests__/MarathonAppValidators-test.js
```diff
@@ -449,6 +449,13 @@ describe("MarathonAppValidators", function() {
       ]);
     });
 
+    it("shouldn't return an error for empty optional fields", function() {
```
Poltergeist:
```JS
it("returns an error for empty optional fields", function() {`
```

PR #2205, opened 2017-05-24
↩️ [release/1.9] DCOS_OSS-956: Implement a field for a load balanced port

nLight:
This PR adds a field for Load Balanced port that has meaningful default values. For HOST network it will be hostPort and for Bridge/User networks it will be container port

screen shot 2017-05-24 at 13 39 02
screen shot 2017-05-24 at 13 39 13

### DIFF
--- a/plugins/services/src/js/components/forms/NetworkingFormSection.js
+++ b/plugins/services/src/js/components/forms/NetworkingFormSection.js
@@ -199,7 +199,11 @@ class NetworkingFormSection extends mixin(StoreMixin) {
     const defaultVipPort = this.isHostNetwork()
       ? hostPort
       : containerPort;
-    const placeholder = defaultVipPort;
+
+    // clear placeholder when HOST network portsAutoAssign is true
+    const placeholder = this.isHostNetwork() && portsAutoAssign
+      ? ''
+      : defaultVipPort;
 
     let vipPortError = null;
     let loadBalancedError = findNestedPropertyInObject(
--- a/plugins/services/src/js/components/forms/NetworkingFormSection.js
+++ b/plugins/services/src/js/components/forms/NetworkingFormSection.js
@@ -25,8 +25,9 @@ import FormRow from '../../../../../../src/js/components/form/FormRow';
 import Icon from '../../../../../../src/js/components/Icon';
 import MetadataStore from '../../../../../../src/js/stores/MetadataStore';
 import Networking from '../../../../../../src/js/constants/Networking';
-import ServiceConfigUtil from '../../utils/ServiceConfigUtil';
 import VirtualNetworksStore from '../../../../../../src/js/stores/VirtualNetworksStore';
+import ValidatorUtil from '../../../../../../src/js/utils/ValidatorUtil';
+import ServiceConfigUtil from '../../utils/ServiceConfigUtil';
 
 const {BRIDGE, HOST, USER} = Networking.type;
 const {MESOS, NONE} = ContainerConstants.type;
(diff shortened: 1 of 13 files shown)


orlandohohmeier:
Any chance we can display the error (port number) in the form as well?

nLight:
@orlandohohmeier works for me 🤔  screen shot 2017-05-24 at 15 52 23

nLight:
@orlandohohmeier I addressed the issue with not working validator. Had to 'unmute' those errors since error path in JSON doesn't match the Batch path of the value.
I also specified the lower border for a port to be valid == 1

nLight:
retest this please

MatApple:
When editing an existing app I get a validation issue before I have a chance to do anything. What's more, the port value seems like it would be valid although it's not clear as a user if this value is a placeholder or an actual value. Error is also strange for an input field.

plugins/services/src/js/components/forms/NetworkingFormSection.js
```diff
@@ -209,11 +186,105 @@ class NetworkingFormSection extends mixin(StoreMixin) {
...
           </FieldLabel>
-          <FieldError>{vipPortError || loadBalancedError}</FieldError>
+        </FormGroup>
+      </FormRow>,
+      (loadBalanced && this.getLoadBalancedPortField(endpoint, index))
+    ];
+  }
+
+  getLoadBalancedPortField(endpoint, index) {
+    const {errors, data: {id, portsAutoAssign}} = this.props;
```
nLight: It won't as we have it in the `defaultProps` that will also prevent destruction error.

nLight:
@MatApple this is a tricky one... I agree that the UX is broken here. Not sure if I can remap that error to something meaningful for the form while keeping the original error for the JSON @wavesoft ❓
Will try to address the second part of the UX issue regarding the confusing placeholder.

wavesoft:
Discussed this a bit with @nLight . I would recommend adding support of two "contexts" to the error message translation rules:

* `json-editor` : Targets the error messages of the JSON editor
* `form` : Targets the form 

Then you could use the following rules:

<pre lang="js">
  {
    path: /^portDefinitions\.[0-9]+\.name$/,
    type: "STRING_PATTERN",
    message: "May only contain digits (0-9), dashes (-) and " +
      "lowercase letters (a-z) e.g. web-server",
    <strong>context: "json-editor"</strong>
  }
</pre>

nLight:
@MatApple interesting, so in your case the form picked up a servicePort 10001 to use as a host port? That should be fixed on release/1.9 and this branch should be based on the fix.

I addressed the misleading placeholder issue, could you please re-check and, if the problem reappears, post your app definition?

nLight:
@wavesoft @MatApple we can think about error system improvements in scope of 1.11

Please keep in mind this is a fix for 1.9. We want to patch a generally broken feature. Can we agree on polishing after we also land the fix in 1.10?

nLight:
Hey @orlandohohmeier, is the change sufficient for you?

plugins/services/src/js/components/forms/NetworkingFormSection.js
```diff
@@ -209,11 +186,105 @@ class NetworkingFormSection extends mixin(StoreMixin) {
...
-            </FieldHelp>
           </FieldLabel>
-          <FieldError>{vipPortError || loadBalancedError}</FieldError>
+        </FormGroup>
+      </FormRow>,
+      (loadBalanced && this.getLoadBalancedPortField(endpoint, index))
+    ];
+  }
+
+  getLoadBalancedPortField(endpoint, index) {
```
nLight: I tried to keep the style in this 1.9 fix though the first intention was to extract components. That would have resulted in too many changes that don't have anything to do with the bug. So I decided not to sneak a refactoring along with this bug fix.

orlandohohmeier:
@nLight did you only rebase this one?

nLight:
@orlandohohmeier one small addition:
`feat(NetworkingFormSection): conditionally clear vip port placeholder`

orlandohohmeier:
L👀ks good!

PR #2207, opened 2017-05-25
feat(DCOS-15919): Design System Table of Contents

tnguyen0:
This PR adds the pages needed for the design system.

**Checklist**


plugins/design-system/pages/ui-components/banners/tabs/OverviewTab.js
```diff
@@ -0,0 +1,87 @@
+import React, { Component } from "react";
+
+class OverviewTab extends Component {
+  render() {
+    return (
```
weblancaster: Can we extract some parts to components so that way we can re use them.

example: `DesignSystemTable.js` `DesignSystemDescription.js` and etc.

What you think @jfurrow

MatApple:
@tnguyen0 let's sit together and go through this on Tuesday. We will need to break this down into reviewable chunks. Typically a PR should not be bigger than ~500 lines of changes. This ensures reviewers can thoroughly comb through the code and test. Please think about how you can break this PR into ~10 PRs. Each PR should be testable. It's Ok if a PR depends on another PR (i.e. there is a merge order). Generally it is best if you reduce cross PR dependencies as much as possible. Please reach out to me if you have any questions, otherwise we'll go through this on Tuesday.

tnguyen0:
Thanks for the feedback @MatApple. Closing this PR.

PR #2208, opened 2017-05-26
fix (JobForm) Fixes issue where the enabled flag would always be truthy

rcorral:
The issue would occur even if the schedule had been manually disabled.
closes DCOS-15615

gif/image

**Checklist**


orlandohohmeier:
LGTM 💯 ✨

PR #2209, opened 2017-05-26
DCOS_OSS-1050: Remove image if no id is provided

Poltergeist:
Remove image if a container id has been removed
but was available earlier.

Closes DCOS_OSS-1050

**Checklist**

### DIFF
--- a/plugins/services/src/js/reducers/serviceForm/Containers.js
+++ b/plugins/services/src/js/reducers/serviceForm/Containers.js
@@ -539,6 +539,9 @@ module.exports = {
       newState[index] = Object.assign({}, newState[index], {
         image: { id: value, kind: "DOCKER" }
       });
+      if (value === "") {
+        delete newState[index].image;
+      }
     }
 
     return newState;
--- a/plugins/services/src/js/reducers/serviceForm/__tests__/Containers-test.js
+++ b/plugins/services/src/js/reducers/serviceForm/__tests__/Containers-test.js
@@ -25,6 +25,62 @@ describe("Containers", function() {
       ]);
     });
 
+    describe("container with image", function() {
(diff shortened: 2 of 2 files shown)


plugins/services/src/js/reducers/serviceForm/Containers.js
```diff
@@ -539,6 +539,9 @@ module.exports = {
       newState[index] = Object.assign({}, newState[index], {
         image: { id: value, kind: "DOCKER" }
       });
+      if (value === "") {
```
orlandohohmeier: Would love if we could get away without adding another condition... #style

orlandohohmeier:
I'm ok with merging this one as is, but would love if you could check if there is away to not introduce another condition.

plugins/services/src/js/reducers/serviceForm/Containers.js
MatApple: can we move condition to outer condition @Poltergeist?
```
if (type === SET && joinedPath === `containers.${index}.image.id` && value !== "") {
```

Poltergeist: Problem is that we operate on the last state, so we need to remove the image attribute if it is not set, so I fear we are not able to remove the condition. Except, we always call the delete before we do the object assign.

nLight: can't we do:
 
```javascript
const image = value === ""
  ? {}
  : {image: { id: value, kind: "DOCKER" }};

newState[index] = Object.assign({}, newState[index], image);
```

Poltergeist: We would still have image: `null` in the resulting object.

PR #2210, opened 2017-05-26
fix(DCOS_OSS-1108): Fix jobs status display

weblancaster:
**DCOS_OSS-1108** Remove class on status label which was blending the color with background and making the text not visible to users.

**before**
screen shot 2017-05-26 at 10 39 42 am

**after**
screen shot 2017-05-26 at 10 37 34 am

**Checklist**

### DIFF
--- a/src/js/pages/jobs/JobsTable.js
+++ b/src/js/pages/jobs/JobsTable.js
@@ -226,8 +226,7 @@ class JobsTable extends React.Component {
 
     const statusClasses = classNames({
       "text-success": jobState.stateTypes.includes("success"),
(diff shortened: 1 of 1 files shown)


src/js/pages/jobs/JobsTable.js
```diff
@@ -226,8 +226,7 @@ class JobsTable extends React.Component {
 
     const statusClasses = classNames({
       "text-success": jobState.stateTypes.includes("success"),
-      "text-danger": jobState.stateTypes.includes("failure"),
-      "text-color-white": jobState.stateTypes.includes("active")
```
MatApple: Oh man...

PR #2212, opened 2017-05-26
fix(DCOS-14334): Update CLI setup script

weblancaster:
**DCOS-14334** Well, the title explains a lot.

**before**
screen shot 2017-05-26 at 1 58 33 pm

**after**
screen shot 2017-05-26 at 1 53 23 pm

**Checklist**

### DIFF
--- a/src/js/components/modals/CliInstallModal.js
+++ b/src/js/components/modals/CliInstallModal.js
@@ -77,11 +77,10 @@ class CliInstallModal extends React.Component {
 
     const instructions = [
       `curl ${downloadUrl} -o dcos`,
-      "sudo mv dcos /usr/local/bin",
-      "sudo chmod +x /usr/local/bin/dcos",
-      `dcos config set core.dcos_url ${clusterUrl}`,
-      Hooks.applyFilter("dcosInstallCommandExtraSteps", undefined),
-      "dcos"
+      `sudo mv dcos /usr/local/bin`,
+      `sudo chmod +x /usr/local/bin/dcos`,
+      `dcos cluster setup ${clusterUrl}`,
+      `dco`
     ].filter(instruction => instruction !== undefined);
 
     return (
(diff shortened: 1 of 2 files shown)


src/js/components/modals/CliInstallModal.js
```diff
@@ -77,11 +77,10 @@ class CliInstallModal extends React.Component {
...
       `curl ${downloadUrl} -o dcos`,
-      "sudo mv dcos /usr/local/bin",
-      "sudo chmod +x /usr/local/bin/dcos",
-      `dcos config set core.dcos_url ${clusterUrl}`,
-      Hooks.applyFilter("dcosInstallCommandExtraSteps", undefined),
-      "dcos"
+      `sudo mv dcos /usr/local/bin`,
+      `sudo chmod +x /usr/local/bin/dcos`,
+      `dcos cluster setup ${clusterUrl}`,
+      `dco`
```
MatApple: This should be `dcos`.

Also wondering why we have stripped out `Hooks.applyFilter("dcosInstallCommandExtraSteps", undefined),`?

weblancaster: good catch. misspelled addressed.

Now the extra steps part looks like we don't need anymore based on the Jira ticket https://jira.mesosphere.com/browse/DCOS-14334?

MatApple: @weblancaster I'm not sure that is the case. We shouldn't assume a change in implementation and should always question cases like this. If you search for where `dcosInstallCommandExtraSteps` is defined, please note the additional commands that are added and please verify with @tamarrow. Thanks.

weblancaster:
@tamarrow Can you review and confirm we have removed the login command please? 

`dcos auth login`

tamarrow:
LGTM

weblancaster:
No problem @MatApple thanks for the review

PR #2200, opened 2017-05-22
Integrate Marathon 1.5 🏃 

orlandohohmeier:
---
⚠️  _Apparently there is a GH bug thus this PR integrates the changes from #2178._

---

This is an integration branch that contains Marathon 1.5 compatible changes:

✅  new Networking API #2166 
✅  Multi Network Form Warning #2167 
✅  UCR Improvements #2098 

Please use https://github.com/dcos/dcos/pull/1505 for testing.

PR #2213, opened 2017-05-29
Integration/1.5 refactoring extraction

nLight:
This is an extraction from #2200 

Since integration takes a bit longer and creates conflicts, let's merge this refactoring first to ease the pain of rebasing #2200 

- The first commit unties Batch transaction paths from JSON representation.
- The second one separates single and multi container form sections so that we effectively separate input reducers.


orlandohohmeier:
Great Job! ⚒ 💯

PR #2215, opened 2017-05-29
DCOS-15173: fix(frameworks): Show task count running

Poltergeist:
This fix changes the way taskcount is displayedfor frameworks. For Frameworks it is better to display all the tasks running instead of instances.

image
image

Closes DCOS-15173

**Checklist**

### DIFF
--- a/plugins/services/src/js/structs/Framework.js
+++ b/plugins/services/src/js/structs/Framework.js
@@ -41,6 +41,17 @@ module.exports = class Framework extends Application {
     return tasksSummary;
   }
 
+  getTaskCount() {
+    // TODO: Circular reference workaround DCOS_OSS-783
+    const MesosStateStore = require("#SRC/js/stores/MesosStateStore");
+
+    const tasks = MesosStateStore.getTasksByService(this) || [];
+
+    return tasks.filter(function(task) {
+      return task.state === "TASK_RUNNING";
+    }).length;
+  }
+
   getUsageStats(resource) {
     const value = this.get("used_resources")[resource];
 
--- a/plugins/services/src/js/components/ServiceBreadcrumbs.js
+++ b/plugins/services/src/js/components/ServiceBreadcrumbs.js
@@ -10,6 +10,7 @@ import BreadcrumbTextContent from "#SRC/js/components/BreadcrumbTextContent";
 import PageHeaderBreadcrumbs from "#SRC/js/components/PageHeaderBreadcrumbs";
 import Util from "#SRC/js/utils/Util";
 
+import Framework from "../structs/Framework";
 import HealthBar from "./HealthBar";
 import ServiceStatusWarningWithDebugInformation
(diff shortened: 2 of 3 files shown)


nLight:
If we show number of tasks maybe we should say tasks and not instances?

screen shot 2017-05-30 at 16 51 25

orlandohohmeier:
@nLight An (Marathon) instance is a running tasks thus I'd say it's fine to stick with "instances" for now.

/cc @leemunroe

nLight:
@orlandohohmeier still confusing: imagine 5 instances of marathon running 5 tasks each. What should we say then: 30 instances? In the app definition will be `instances: 5`

nLight:
The problem I see that by trying to resolve one confusion we're opening a door for another. Right now we have a term Instance, an instance of an app can have multiple tasks, still we show 1 instance. For a framework we have instances of a framework and tasks running by the framework. I.e. Kafka launches 2 tasks to support 1 instance of Kafka.

orlandohohmeier:
@nLight let's not block this PR because of an already confusing task/instance naming. I agree that we should revisit the naming to finally define a comprehensive nomenclature for services. I create https://jira.mesosphere.com/browse/DCOS_OSS-1161 to track the respective work.

plugins/services/src/js/structs/Framework.js
```diff
@@ -41,6 +41,17 @@ module.exports = class Framework extends Application {
     return tasksSummary;
   }
 
+  getTaskCount() {
+    // TODO: Circular reference workaround DCOS_OSS-783
+    const MesosStateStore = require("#SRC/js/stores/MesosStateStore");
```
nLight: Let's extract this as a method i.e. `getAllTasks` and reuse it in `getTaskCount` and `getResources`

```javascript
const MesosStateStore = require("#SRC/js/stores/MesosStateStore");
return MesosStateStore.getTasksByService(this) || [];
```

plugins/services/src/js/structs/Framework.js
```diff
@@ -41,6 +41,17 @@ module.exports = class Framework extends Application {
...
+
+    const tasks = MesosStateStore.getTasksByService(this) || [];
+
+    return tasks.filter(function(task) {
```
nLight: ...and maybe add a test 👮

leemunroe:
Per offline discussion, stick to instance count in the collection table and next to the service name. In the status bar tooltip add the word "tasks" to help add clarity.

Example: marathon-user has 5 instances and each instance has 10 healthy tasks.
* In the Service collection view, the status bar would say "50 Healthy tasks (100%)" in the tooltip, and "Running (5 Instances)" in the status column
* In the Service detail view, the title should read "marathon-user | Running (5)" and the bar should be green with "50 Healthy tasks (100%)" in the tooltip, same as collection view

@orlandohohmeier Did I catch everything?

PR #2221, opened 2017-05-31
DCOS-14844, DCOS-14756: refactor(services): remove applyPatch function

Poltergeist:
This is mostly the work of @nLight from #2081 Which has been rebased to current Master and squashed into one commit. 

As this PR is introducing a MAYOR refactoring. Please give this PR some serious testing before approving to the changes.

This PR will introduce a mayor change in the way we parse / reduce application configurations in the service form. Now the ADD_ITEM transaction will hold null or a complete object of the item instead of the index. We do this so that we can keep unknown fields from the JSON. Right now this also is
creating redundancy because all the transactions after the ADD_ITEM are just reseting the same
value.

E.G. :

The following JSON is parsed:
```JSON
{
  "array": [{
    "id": "1234",
    "unknownField": "withoutTransaction"
  }]
}
```
And will create the following Transactions:
 - `ADD_ITEM, {id: "1234", unknownField:
   "withoutTransaction", ["array"]}`
 - `SET, ['array', 0, 'id'], "1234"`

But in this case this redundancy is good,in some places we are depending on the separate Transactions.

Closes DCOS-14844, DCOS-14756

**Checklist**

### DIFF
--- a/plugins/services/src/js/reducers/serviceForm/Containers.js
+++ b/plugins/services/src/js/reducers/serviceForm/Containers.js
@@ -102,7 +102,7 @@ function containersParser(state) {
   }
 
   return state.containers.reduce((memo, item, index) => {
-    memo.push(new Transaction(["containers"], index, ADD_ITEM));
+    memo.push(new Transaction(["containers"], item, ADD_ITEM));
     if (item.name) {
       memo.push(new Transaction(["containers", index, "name"], item.name));
     }
@@ -161,7 +161,7 @@ function containersParser(state) {
         memo.push(
           new Transaction(
             ["containers", index, "artifacts"],
-            artifactIndex,
+            artifact,
             ADD_ITEM
           )
         );
@@ -187,7 +187,7 @@ function containersParser(state) {
         memo = memo.concat([
           new Transaction(
             ["containers", index, "endpoints"],
-            endpointIndex,
+            endpoint,
             ADD_ITEM
           ),
           new Transaction(
@@ -306,7 +306,10 @@ function containersParser(state) {
 }
 
 module.exports = {
-  JSONReducer(state = [], { type, path = [], value }) {
+  JSONReducer(state = [], { type, path = [], value }, containerIndex) {
+    if (containerIndex === 0) {
+      state = [];
+    }
     const [base, index, field, secondIndex, name, subField] = path;
 
     if (this.networkType == null) {
@@ -333,11 +336,17 @@ module.exports = {
 
     if (!path.includes("containers") && !path.includes("volumeMounts")) {
       return state.map((container, index) => {
-        container.endpoints = mapEndpoints(
-          this.endpoints[index].endpoints,
-          this.networkType,
-          this.appState
-        );
+        if (
+          this.endpoints &&
+          this.endpoints[index] &&
+          this.endpoints[index].endpoints
+        ) {
+          container.endpoints = mapEndpoints(
+            this.endpoints[index].endpoints,
+            this.networkType,
+            this.appState
+          );
+        }
 
         return container;
       });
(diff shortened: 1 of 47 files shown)


nLight:
testses...

nLight:
run integration tests

nLight:
As discussed: Pod containers are multiplying as you jump between JSON and form 🐑  🐑  🐑

orlandohohmeier:
There seems to be an issue with the health checks json reducer...

```
HealthChecks.js?9d97:96 Uncaught TypeError: Cannot create property 'protocol' on number '1'
```

jfurrow:
I tried for quite some time to break the UI and it all works great! I also couldn't seem to get the health check error that @orlandohohmeier found, so I'm going to approve.

Great work everyone!

plugins/services/src/js/reducers/serviceForm/FormReducers/__tests__/Artifacts-test.js
```diff
@@ -7,9 +7,9 @@ describe("Artifacts", function() {
   describe("#FromReducer", function() {
     it("emits correct JSON", function() {
       const batch = new Batch([
-        new Transaction(["fetch"], 0, ADD_ITEM),
+        new Transaction(["fetch"], { uri: "http://mesosphere.io" }, ADD_ITEM),
         new Transaction(["fetch", 0, "uri"], "http://mesosphere.io", SET),
-        new Transaction(["fetch"], 1, ADD_ITEM),
+        new Transaction(["fetch"], { uri: "http://mesosphere.com" }, ADD_ITEM),
```
weblancaster: is this correct? should we use `https` instead if so?

orlandohohmeier: This is just a test – no need for https, but we might want to use `example.com` to avoid confusion.

plugins/services/src/js/reducers/serviceForm/Containers.js
```diff
@@ -365,8 +370,9 @@ module.exports = {
       switch (type) {
         case ADD_ITEM:
           const name = `container-${newState.length + 1}`;
+          const newItem = Object.assign({}, DEFAULT_POD_CONTAINER, { name });
 
-          newState.push(Object.assign({}, DEFAULT_POD_CONTAINER, { name }));
+          newState.push(value || newItem);
```
orlandohohmeier: I'm pretty sure that we never want to add any falsy values but we probably should adjust the line to avoid confusions and clearly state the expected type. Good catch!

orlandohohmeier:
@Poltergeist @nLight Great Job! ⚒ 💯  I'm ok with merging it as it, but would love it one of you could address @weblancaster comments to improve the maintainability.

Poltergeist:
Found some Bugs regarding the containers. Will add some tests for that.

plugins/services/src/js/reducers/serviceForm/JSONReducers/ExternalVolumes.js
```diff
@@ -29,7 +29,8 @@ module.exports = {
        * 6) Set the mode from `volume.mode` on the path
        *    `externalVolumes.${index}.mode`
        */
-        memo.push(new Transaction(["externalVolumes"], index, ADD_ITEM));
+        // memo.push(new Transaction(["externalVolumes"], JSON.parse(JSON.stringify(item)), ADD_ITEM));
```
nLight: Comment 👮

Poltergeist: Fixed

plugins/services/src/js/reducers/serviceForm/Containers.js
Poltergeist: I addressed it for the containers file. But this is a pattern we used at all ADD_ITEM cases. I will fix the other files (reducers) in a separate PR.

rcorral:
Looks like there's something going on with the networking page for pods
gif

Poltergeist:
The issue from @rcorral is fixed.

nLight:
run integration tests

PR #2224, opened 2017-05-31
feat(DCOS-15919): Design System Introduction and Wrapper Pages

tnguyen0:
This PR adds the wrapper design system and introduction pages to the design system.

**Checklist**

### DIFF
--- a/plugins/design-system/pages/DesignSystemPage.js
+++ b/plugins/design-system/pages/DesignSystemPage.js
@@ -1,0 +1,14 @@
+import React, { Component } from "react";
+import "../styles/shame.less";
+
+const SDK = require("../SDK").getSDK();
+
+const { Page } = SDK.get(["Page"]);
+
+class DesignSystemPage extends Component {
+  render() {
+    return <Page>{this.props.children}</Page>;
+  }
+}
+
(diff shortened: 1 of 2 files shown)


plugins/design-system/pages/DesignSystemPage.js
```diff
@@ -0,0 +1,14 @@
...
+
+const SDK = require("../SDK").getSDK();
+
+const { Page } = SDK.get(["Page"]);
+
+class DesignSystemPage extends Component {
+  render() {
+    return <Page>{this.props.children}</Page>;
+  }
+}
```
jfurrow: Just curious, what's the purpose of this component?

tnguyen0: You wrote it I believe :P

jfurrow: LOL, so it's obviously very important...

jfurrow: I'm fine with keeping it for now. Maybe there was a good reason, but I can't think of any right now except that it prevents us from having to import the SDK in every design system page, maybe.

PR #2229, opened 2017-06-01
fix(DCOS-12243): Redirect user back to original relative path after logged-in

weblancaster:
**DCOS-12243** If the user is logged out and try to go to a path e.g `http://localhost:4200/#/services/detail/%2Fsome-service/configuration`
we will redirect the user to login page but after the user log-in we do not send back to where the original path was.

This PR add the feat to redirect the user back to the relative path after logged-in.

**Checklist**

### DIFF
--- a/plugins/oauth/hooks.js
+++ b/plugins/oauth/hooks.js
@@ -74,7 +74,13 @@ module.exports = Object.assign({}, StoreMixin, {
   },
 
   redirectToLogin(nextState, replace) {
-    replace("/login");
+    const redirectTo = RouterUtil.getRedirectTo();
+    // Ignores relative path if redirect is present
+    if (redirectTo) {
+      replace(`/login?redirect=${redirectTo}`);
+    } else {
+      replace(`/login?relativePath=${nextState.location.pathname}`);
+    }
   },
 
   AJAXRequestError(xhr) {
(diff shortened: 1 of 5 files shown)


src/js/utils/RouterUtil.js
```diff
@@ -17,28 +17,63 @@ function findRedirect(queryString) {
 }
 
 const RouterUtil = {
+  /**
+   * Parse the url and find the query string (?)
+   * before or after the #
+   *
+   * @returns {Boolean|Object} false or query string object
+   */
+  getQueryStringInUrl() {
```
MatApple: Please make sure we write tests for these methods in RouterUtil. Looks like we could do with more coverage 👍

weblancaster: It's addressed.

jfurrow:
This works great! Nice work.

weblancaster:
that's what happen when you work in two Jiras at the same time. I added the correct Jira number @Poltergeist

Poltergeist:
I am going to merge this. The tests requested by @MatApple were added. And the Issue number is now correct. 

Thanks for the changes.

plugins/oauth/hooks.js
```diff
@@ -74,7 +74,13 @@ module.exports = Object.assign({}, StoreMixin, {
   },
 
   redirectToLogin(nextState, replace) {
```
orlandohohmeier: Do we really need the whole new state? Please try to only pass arguments and properties that you actually need, as it's otherwise really hard to maintain the respective function/component.

PR #2234, opened 2017-06-07
DCOS-15999: Support unknown volumes

nLight:
ℹ️  **Dependencies**:
1. ~#2221~
2. https://github.com/mesosphere/dcos-ui-plugins-private/pull/601

✅  **Checklist**

This PR adds support for unknown volumes, they won't be shown in the Form, but the Batch now is fully aware of them, parses them and reduces them back.

Decided not to extract chores and Volumes-test move, they are isolated and small. 

**Testing**

Please test along with: https://github.com/mesosphere/dcos-ui-plugins-private/pull/598
and follow the instructions provided there.

### DIFF
--- a/plugins/services/src/js/components/forms/MultiContainerVolumesFormSection.js
+++ b/plugins/services/src/js/components/forms/MultiContainerVolumesFormSection.js
@@ -94,6 +94,10 @@ class MultiContainerVolumesFormSection extends Component {
         path: "volumeMounts"
       });
 
+      if (volumes.type === VolumeConstants.type.unknown) {
+        return null;
+      }
+
       return (
         <FormGroupContainer onRemove={removeHandler} key={key}>
           <FormRow>
--- a/plugins/services/src/js/reducers/serviceForm/JSONReducers/__tests__/UnknownVolumes-test.js
+++ b/plugins/services/src/js/reducers/serviceForm/JSONReducers/__tests__/UnknownVolumes-test.js
@@ -1,0 +1,137 @@
+const { ADD_ITEM } = require("#SRC/js/constants/TransactionTypes");
+const UnknownVolumes = require("../UnknownVolumes");
+
+describe("UnknownVolumes", function() {
+  describe("#JSONParser", function() {
+    describe("with empty input", function() {
+      it("returns an empty array", function() {
+        expect(UnknownVolumes.JSONParser({})).toEqual([]);
+      });
+    });
+
+    describe("with known volumes", function() {
+      describe("external volume", function() {
+        it("returns an empty array", function() {
+          const externalVolume = {
+            container: {
+              volumes: [
+                {
+                  containerPath: "/mnt/volume",
+                  external: {
+                    name: "someVolume",
+                    provider: "dvdi",
+                    options: {
+                      "dvdi/driver": "rexray"
+                    }
+                  },
+                  mode: "RW"
+                }
+              ]
+            }
+          };
+          expect(UnknownVolumes.JSONParser(externalVolume)).toEqual([]);
+        });
+      });
+
+      describe("persistent volume", function() {
+        it("returns an empty array", function() {
+          const persistentVolume = {
+            container: {
+              volumes: [
+                {
+                  containerPath: "/dev/null",
+                  persistent: { size: 1024 },
+                  mode: "RW"
+                }
+              ]
+            }
+          };
+          expect(UnknownVolumes.JSONParser(persistentVolume)).toEqual([]);
+        });
+      });
+
+      describe("docker volume", function() {
+        it("returns an empty array", function() {
+          const dockerVolume = {
+            container: {
+              volumes: [
+                {
+                  hostPath: "some/path",
+                  containerPath: "/mnt/volume",
+                  mode: "RW"
+                }
+              ]
+            }
+          };
+          expect(UnknownVolumes.JSONParser(dockerVolume)).toEqual([]);
+        });
+      });
+    });
+
+    describe("with unknown volume", function() {
+      it("extracts the volume", function() {
+        const unknownVolumes = {
+          container: {
+            volumes: [
+              {
+                containerPath: "/dev/null"
+              }
+            ]
+          }
+        };
+        expect(UnknownVolumes.JSONParser(unknownVolumes)).toEqual([
+          {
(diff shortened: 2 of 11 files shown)


jfurrow:
run unit tests

plugins/services/src/js/components/forms/MultiContainerVolumesFormSection.js
```diff
@@ -94,6 +94,10 @@ class MultiContainerVolumesFormSection extends Component {
         path: "volumeMounts"
       });
 
+      if (volumes.type === VolumeConstants.type.unknown) {
+        return <noscript />;
```
weblancaster: please return `null` since is explicit and suggested by react docs

weblancaster:
retest this please

plugins/services/src/js/reducers/serviceForm/JSONReducers/__tests__/UnknownVolumes-test.js
```diff
@@ -0,0 +1,137 @@
+const { ADD_ITEM } = require("#SRC/js/constants/TransactionTypes");
+const UnknownVolumes = require("../UnknownVolumes");
+
+describe("UnknownVolumes", function() {
+  describe("#JSONParser", function() {
```
weblancaster: Do we need all `describe` in here? Seems like adding more complexity than clearly.

When trying to read and understand the units tests I'm getting confused would be clearer if the `it` description contained the info needed to understand what needs to be tested.

```
describe("UnknownVolumes", function() {
  describe("#JSONParser", function() {
    it("returns an empty array with empty input", function() {
      expect(UnknownVolumes.JSONParser({})).toEqual([]);
    });
  });
});
```

plugins/services/src/js/reducers/serviceForm/JSONReducers/__tests__/UnknownVolumes-test.js
```diff
@@ -0,0 +1,137 @@
...
+        };
+        expect(UnknownVolumes.JSONParser(unknownVolumes)).toEqual([
+          {
+            type: ADD_ITEM,
+            value: {
+              containerPath: "/dev/null"
+            },
+            path: ["unknownVolumes"]
+          }
+        ]);
```
weblancaster: Would be great if we follow `AAA` Arrange Act Assert. In this particular case.

```
const expectedResult = [
  {
    type: ADD_ITEM,
    value: {
      containerPath: "/dev/null"
    },
    path: ["unknownVolumes"]
  }
];
```

rcorral:
@nLight I'm trying to find a good way to test this but the validation keeps failing for me.
What's a valid JSON? I'm using the following:
```
{
  "id": "/foo",
  "instances": 1,
  "container": {
    "type": "DOCKER",
    "docker": {
      "image": "bar"
    },
    "volumes": [
      {
        "containerPath": "/foo/bar"
      }
    ]
  },
  "cpus": 0.1,
  "mem": 128,
  "requirePorts": false,
  "cmd": "foo"
}
```

plugins/services/src/js/reducers/serviceForm/JSONReducers/__tests__/UnknownVolumes-test.js
nLight: Following http://www.betterspecs.org/#contexts extracting a common context which is `with empty input` should be done via contexts or nested describe blocks in our case. While it seems not making any sense in this particular example it draws the line between `with empty input` and `with any other inputs` that can have multiple expectations as well as multiple contexts.

plugins/services/src/js/components/forms/MultiContainerVolumesFormSection.js
nLight: Thank you for noticing! Somehow I thought we still use an older React, I had a rough day... 😅

plugins/services/src/js/reducers/serviceForm/JSONReducers/__tests__/UnknownVolumes-test.js
nLight: Makes sense but following the link it actually should be:

```javascript
it("extracts the volume", function() {
  // Arrange
  const input = {
    container: {
      volumes: [
        {
          containerPath: "/dev/null"
        }
      ]
    }
  };

  // Act
  const result = UnknownVolumes.JSONParser(input);

  // Assert
  expect(result).toEqual([
    {
      type: ADD_ITEM,
      value: {
        containerPath: "/dev/null"
      },
      path: ["unknownVolumes"]
    }
  ]);
});
```
What do you think?

weblancaster: I think it's good I would only move the `expectedResult` to be part of the arrange as `input` is

```
const expectedResult = [
    {
      type: ADD_ITEM,
      value: {
        containerPath: "/dev/null"
      },
      path: ["unknownVolumes"]
    }
  ]
```

jfurrow:
@rcorral This worked for me:
```json
{
  "id": "app-with-secrets",
  "cmd": "sleep 100",
  "container": {
    "volumes": [
      {
        "containerPath": "path",
        "secret": "secretpassword"
      }
    ]
  },
  "secrets": {
    "secretpassword": {
      "source": "databasepassword"
    }
  }
}
```

orlandohohmeier:
@nLight this needs another rebase...

jfurrow:
@nLight Is this validation error expected? `Volume mode must be defined`

orlandohohmeier:
@jfurrow yeah, you need to define `"mode": "RW"` or `"mode": "RO"`for every volume...

jfurrow:
Thanks @orlandohohmeier!

MatApple:
@orlandohohmeier @nLight I have spoken with @jdef and @kensipe about FBS and they are currently blocked (by https://jira.mesosphere.com/browse/DCOS-16109) when it comes to adding the latest FBS bits. So until they land or at least end up in a PR, we cannot validate this PR works as expected.

nLight:
@MatApple @jfurrow on one hand this PR doesn't break anything because it introduces a volume type that one cannot create via the Form. On the other hand I understand you that raml is blocking you from verifying the fix. We can wait until the blocker is resolved.

PR #2235, opened 2017-06-07
Adjust job breadcrumb generation

orlandohohmeier:
Utilize the  job tree structure for the breadcrumb generation instead of relying on a rather error prone string based mechanism. This change is a prerequisite for the job tree structure   adjustments to align the UI with the way Metronome handles IDs and namespaces.

### DIFF
--- a/src/js/components/breadcrumbs/JobsBreadcrumbs.js
+++ b/src/js/components/breadcrumbs/JobsBreadcrumbs.js
@@ -3,6 +3,10 @@ import prettycron from "prettycron";
 import React from "react";
 import { Tooltip } from "reactjs-components";
 
+import Job from "#SRC/js/structs/Job";
+import JobTree from "#SRC/js/structs/JobTree";
+import TaskStates from "#PLUGINS/services/src/js/constants/TaskStates";
+
 import Breadcrumb from "../../components/Breadcrumb";
 import BreadcrumbSupplementalContent
   from "../../components/BreadcrumbSupplementalContent";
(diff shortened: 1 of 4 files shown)


orlandohohmeier:

src/js/pages/jobs/JobTaskDetailPage.js
```diff
@@ -20,19 +26,25 @@ class JobTaskDetailPage extends React.Component {
       { label: "Logs", routePath: routePrefix + "/logs" }
     ];
 
+    const job = MetronomeStore.getJob(id);
     const task = MesosStateStore.getTaskFromTaskID(taskID);
 
-    let breadcrumbs;
+    let breadcrumbs = (
+      <JobsBreadcrumbs tree={DCOSStore.jobTree} item={job} details={false} />
```
nLight: Do you think it makes sense to have 2 components and pick one when we need details and another when we don't? Right now `details=false` doesn't tell me anything. Let's think how we can remove this flag.

ℹ️ https://martinfowler.com/bliki/FlagArgument.html

src/js/components/breadcrumbs/JobsBreadcrumbs.js
```diff
@@ -3,104 +3,110 @@ import prettycron from "prettycron";
...
+const JobsBreadcrumbs = ({ tree, item, children, details = true }) => {
+  const elements = tree
+    .filterItems(function(currentItem) {
+      return item && item.getId() === currentItem.getId();
+    })
+    .reduceItems(function(acc, currentItem) {
+      return acc.concat(getBreadcrumb(currentItem, details));
+    }, []);
+
+  const breadcrumbs = [getBreadcrumb(tree, details)].concat(
```
nLight: Nitpicking. Looks like ObjectiveC :trollface:  I'd rather see:
 
```javascript
[].concat(
  getBreadcrumb(tree, details),
  ...
)
```

src/js/components/breadcrumbs/JobsBreadcrumbs.js
```diff
@@ -3,104 +3,110 @@ import prettycron from "prettycron";
...
-  if (jobStatus != null) {
-    return (
-      <BreadcrumbSupplementalContent>
-        <div className="service-page-header-status muted">
-          ({jobStatus})
-        </div>
-      </BreadcrumbSupplementalContent>
-    );
+function getItemStatus(item) {
+  if (!(item instanceof Job)) {
```
nLight: both `getItemSchedule` and `getItemStatus` perform this check `!(item instanceof Job)` does it make sense to extract the check on the upper level?

src/js/components/breadcrumbs/JobsBreadcrumbs.js
```diff
@@ -3,104 +3,110 @@ import prettycron from "prettycron";
...
+    return null;
+  }
+
+  let longestRunningTask = null;
```
nLight: Shall we extract a method here?
```javascript
const longestRunningTask = getLongestRunningTask();
```

orlandohohmeier: I only moved the respective function and didn't invested time to adjust the respective implementation, but I'm happy to optimize and improve.

src/js/components/breadcrumbs/JobsBreadcrumbs.js
orlandohohmeier: Nope. We should rather align the job and jobtree interface so that there is no need for a type check.

src/js/pages/jobs/JobTaskDetailPage.js
orlandohohmeier: Yeah, this is a  cheap _workaround_ to align the functionality with the old implementation. I'm planning to completely redo the breadcrumbs (remove the type specific components and introduce utils to generate the list of crumbs), but went for a less intrusive refactoring as this change is urgently needed to eventually fix DCOS_OSS-1033. I will file a JIRA to properly refactor the breadcrumbs and leave this as is fi that's ok with you.

orlandohohmeier:
@nLight @Poltergeist I've addressed your comments... thanks for the review!

PR #2237, opened 2017-06-07
fix(TimeSeriesGraphs): Hide mouse coordinate axis lines on mouse out

jfurrow:
This PR hides the mouse coordinate axis lines on initial render and when the cursor leaves the time series graph. Previously the lines were rendered on the graph at all times; at mount they were visible at the top and left edges, and were not hidden when the mouse left the graph.

Before:

After:

**Checklist**


rcorral:
I've been wanting to fix this for quite some time now. Thanks, @jfurrow!

PR #2239, opened 2017-06-08
test(DCOS-12815): Create an app with virtual network

tnguyen0:
This PR adds a system test for creating an app with virtual network.

Run the test:
```
    -v `pwd`:/dcos-ui \
    mesosphere/dcos-ui dcos-system-test-driver --only "apps" \
```

**Checklist**

### DIFF
--- a/system-tests/services/test-apps.js
+++ b/system-tests/services/test-apps.js
@@ -1868,5 +1868,196 @@ describe("Services", function() {
         .getTableRowThatContains(serviceName)
         .should("exist");
     });
+
+    it("should create an app with virtual network", function() {
+      const serviceName = "app-with-virtual-network";
+      const command = "python3 -m http.server 8080";
+      const containerImage = "python:3";
+
+      cy.contains("Single Container").click();
+
+      cy
+        .root()
+        .getFormGroupInputFor("Service ID *")
+        .type(`{selectall}{rightarrow}${serviceName}`);
+
+      // Hack to allow entering decimals in number fields
+      cy.root().getFormGroupInputFor("CPUs *").invoke("attr", "type", "text");
+      cy.root().getFormGroupInputFor("CPUs *").type("{selectall}0.5");
+      cy.root().getFormGroupInputFor("CPUs *").invoke("attr", "type", "number");
+
+      cy.root().getFormGroupInputFor("Memory (MiB) *").type("{selectall}32");
+      cy.root().getFormGroupInputFor("Container Image").type(containerImage);
+      cy.root().getFormGroupInputFor("Command").type(command);
+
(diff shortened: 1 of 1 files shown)


system-tests/services/test-apps.js
```diff
@@ -1677,5 +1677,186 @@ describe("Services", function() {
...
+                {
+                  name: "mapped",
+                  hostPort: 4200,
+                  containerPort: 8080,
+                  protocol: "tcp"
+                }
+              ]
+            }
+          },
+          ipAddress: {
```
wavesoft: This looks like the old network configuration. Perhaps you should migrate this to the new version.
Have a look here: https://github.com/mesosphere/marathon/blob/master/docs/docs/upgrade/network-api-migration.md

Even if the tests are passing now, the will definitely fail when we make the new release 😄

system-tests/services/test-apps.js
```diff
@@ -1677,5 +1677,186 @@ describe("Services", function() {
...
+        .configurationSection("Network")
+        .configurationMapValue("Network Name")
+        .contains("dcos");
+
+      cy
+        .root()
+        .configurationSection("Service Endpoints")
+        .then(function($section) {
+          expect($section.get().length).to.equal(1);
+          const numOfCols = 6;
```
wavesoft: In general I wanted to avoid keeping such a generic code inside the tests. For this kind of uses, I introduced the `.getTableColumn()` function (check `_support/index.js`). 

You can use it like so:

```js
$section.find("table").getTableColumn(0).should("deep.equal", [
  "http",
  "Not configured",
  ...
]);
```

And if that function is not working for you, I would recommend to create a new utility function that could (for example) extract all cells and rows of the table.

system-tests/services/test-apps.js
tnguyen0: Thanks for the heads up! I will hold off getting this PR merged until 1.5

system-tests/services/test-apps.js
tnguyen0: Sounds good! Much better :)

tnguyen0:
Run integration tests

ahoskins:
LGTM, nice and comprehensive 👍

weblancaster:
run tests

tnguyen0:
@ahoskins, @weblancaster: Can you guys take another look?

weblancaster:
@wavesoft are you planning to take a look on this?

wavesoft:
Looks good 👍  your test is passing.
(Why the rest is failing so badly though? 😞 )

tnguyen0:
I do not have the latest master on my branch. Those tests have been fixed.

PR #2203, opened 2017-05-23
test(pods-communication) add tests for basic pod communication

juliangieseke:
Add test case which creates a pod with two containers and tests
if a http-client on one can access nginx on the other container

Closes DCOS_15051

PR #2240, opened 2017-06-08
test(DCOS-15051): Create a pod with communicating services

tnguyen0:
This PR adds a system test for communication between two services in a pod.

It branches off and addresses the comments in: https://github.com/dcos/dcos-ui/pull/2203

Run the test:
```
    -v `pwd`:/dcos-ui \
    mesosphere/dcos-ui dcos-system-test-driver --only "pods" \
```

**Checklist**


system-tests/_support/index.js
```diff
@@ -159,6 +159,30 @@ Cypress.addChildCommand("contents", function(elements) {
...
 });
 
+/* Sets given JSONString as Value for ACE Editor
+ * 
+ * This function will set the given text as value for ACE Editor
+ *
+ * @param {jQuery.Element} elements - The DOM scope to search within
+ * @param {String} JSONString - JSON Code as String
+ */
+Cypress.addChildCommand("setJSON", function(elements, JSONString) {
```
weblancaster: Don't we have this util already?

tnguyen0: I don't see it. I only see one to get the contents.

system-tests/_support/index.js
```diff
@@ -159,6 +159,30 @@ Cypress.addChildCommand("contents", function(elements) {
     .get();
 });
 
+/* Sets given JSONString as Value for ACE Editor
```
rcorral: We normally begin JDOC comment blocks with `/**` and on the next line you start document. Check the example above.

system-tests/_support/index.js
```diff
@@ -159,6 +159,30 @@ Cypress.addChildCommand("contents", function(elements) {
...
+ * @param {jQuery.Element} elements - The DOM scope to search within
+ * @param {String} JSONString - JSON Code as String
+ */
+Cypress.addChildCommand("setJSON", function(elements, JSONString) {
+  if (elements.length != null && JSONString) {
+    elements.map(function(index, element) {
+      const doc = element.ownerDocument;
+      const win = doc.defaultView || doc.parentWindow;
+      if (element.classList.contains("ace_editor")) {
+        win.ace.edit(element.id).setValue(JSONString);
```
rcorral: if you're doing other conditional checks above, why not also check that `win.ace` is defined, rather than assuming it will be because an element with class `ace_editor` exists.

rcorral:
The tests are also not passing
img

system-tests/_support/index.js
weblancaster: hm I thought we had this util already

system-tests/_support/index.js
weblancaster: @tnguyen0 since you are using VS code you can use https://marketplace.visualstudio.com/items?itemName=joelday.docthis

system-tests/_support/index.js
```diff
@@ -160,6 +160,30 @@ Cypress.addChildCommand("contents", function(elements) {
...
+ * Sets given JSONString as value for ACE Editor
+ *
+ * This function will set the given text as value for ACE Editor
+ *
+ * @param {jQuery.Element} elements - The DOM scope to search within
+ * @param {String} JSONString - JSON Code as String
+ */
+Cypress.addChildCommand("setJSON", function(elements, JSONString) {
+  if (elements.length != null && JSONString) {
+    elements.map(function(index, element) {
```
ahoskins: You aren't modifying the elements, should use `jQuery.each()` instead (and remove the return also).

tnguyen0: @ahoskins Good call. Fixed!

ahoskins: Great - changes look good now 👍

PR #2244, opened 2017-06-09
DCOS-16148: Default port definitions to empty array

nLight:
This PR changes the way we remove empty values from an app definition. So that we keep empty arrays and empty objects.

Turns out there's a difference for Marathon between `null` and `[]` so when we don't have `portDefinitions` in the initial JSON Marathon will allocate a port by default, but when we send an empty array Marathon won't give us a default port.

ℹ️  `applyPatch` tests were failing and I decided to finish the cleanup and remove `applyPatch` from the codebase.

Closes DCOS-16148


jfurrow:
I don't seem to be able to submit an app definition that uses the Mesos runtime anymore:

nLight:
@jfurrow thanks for catching that one, should be fixed now.

jfurrow:
Nice work @nLight, LGTM! :shipit: 🚢

ahoskins:
Working for me, as well 👍

PR #2245, opened 2017-06-09
DCOS-11685: Disallow leading/trailing whitespace chars in label keys

jfurrow:
This PR disallows leading & trailing whitespace characters in label keys.

**Checklist**

### DIFF
--- a/plugins/services/src/js/validators/MarathonAppValidators.js
+++ b/plugins/services/src/js/validators/MarathonAppValidators.js
@@ -324,6 +324,34 @@ const MarathonAppValidators = {
 
       return errors;
     }, []);
+  },
+
+  /**
+   * @param {Object} app - The data to validate
+   * @returns {Array} Returns an array with validation errors
+   */
+  validateLabels(app) {
+    if (ValidatorUtil.isDefined(app.labels)) {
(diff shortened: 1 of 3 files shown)


plugins/services/src/js/validators/MarathonAppValidators.js
```diff
@@ -324,6 +324,34 @@ const MarathonAppValidators = {
...
+          if (/^\s|\s$/.test(labelKey)) {
+            accumulator.push(`labels.${labelKey}`);
+          }
+
+          return accumulator;
+        }, [])
+        .map(function(labelPath) {
+          return {
+            path: [labelPath],
+            message: "Keys must not start or end with whitespace characters",
```
weblancaster: Should we i18n this?

jfurrow: I decided not to because the rest of the file isn't using it. But if you think we should do it for this case, I'll be happy to do it!

rcorral: The consensus was to do the internationalization in a follow-up PR

weblancaster: Sounds good.

PR #2246, opened 2017-06-12
↩️ [1.9] DCOS-16147: default portDefinitions to an empty array

nLight:
This is a back port of https://github.com/dcos/dcos-ui/pull/2244
I had to adjust `applyPatch` tests since it uses `stripEmptyProperties` that I had to update so it doesn't remove empty arrays and objects.

To test: launch 1.9 cluster, go and create an app, verify that `portDefinitions: []`. Without the fix you won't see `portDefinitions` at all and because of that Marathon will create a port by default.


orlandohohmeier:
LGTM! 👓 👍

PR #2248, opened 2017-06-12
fix(DCOS-16008): Improve service confirmation modal.

weblancaster:
**DCOS-16008** This PR improve the service confirmation modal where now we only display the path/service.
This PR also adds a couple of minor fixes to typos and input autofocus.

How to test:
Create a service container then click delete de service container which will bring the modal to type the service name. before the entire path was required to be typed now only the service name
e.g `path/to/service/name` in case you are in the name service you will only need to type `name`.

If in the `/service/` path you will need to type only `service` but whatever is under service will also be deleted.

**Checklist**

### DIFF
--- a/plugins/services/src/js/components/modals/ServiceDestroyModal.js
+++ b/plugins/services/src/js/components/modals/ServiceDestroyModal.js
@@ -103,8 +103,7 @@ class ServiceDestroyModal extends React.Component {
 
   getIsRightButtonDisabled() {
     return (
-      this.getServiceName(this.props.service.getId()) !==
-      this.getServiceName(this.state.serviceNameConfirmationValue)
+      this.props.service.getName() !== this.state.serviceNameConfirmationValue
     );
   }
 
@@ -197,7 +196,7 @@ class ServiceDestroyModal extends React.Component {
 
   getDestroyServiceModal() {
     const { open, service, intl } = this.props;
-    const serviceName = this.getServiceName(service.getId());
+    const serviceName = service.getName();
 
     let itemText = `${StringUtil.capitalize(UserActions.DELETE)}`;
 
@@ -260,26 +259,6 @@ class ServiceDestroyModal extends React.Component {
       <p className="text-align-center flush-bottom">
         {this.props.subHeaderContent}
       </p>
-    );
-  }
-
-  /**
-   * Format serviceId to only the service name
-   * e.g path/to/service/name will only return name
-   *
-   * @param {String} serviceId
-   * @returns {String} formatted serviceId
-   *
-   * @memberof ServiceDestroyModal
-   */
-  getServiceName(serviceId) {
-    if (typeof serviceId !== "string") {
-      return null;
-    }
-
-    return serviceId.substring(
-      serviceId.lastIndexOf("/") + 1,
-      serviceId.length
     );
   }
 
--- a/plugins/services/src/js/components/modals/ServiceDestroyModal.js
+++ b/plugins/services/src/js/components/modals/ServiceDestroyModal.js
@@ -103,8 +103,8 @@ class ServiceDestroyModal extends React.Component {
 
   getIsRightButtonDisabled() {
     return (
-      this.formatServiceName(this.props.service.getId()) !==
-      this.formatServiceName(this.state.serviceNameConfirmationValue)
+      this.getServiceName(this.props.service.getId()) !==
+      this.getServiceName(this.state.serviceNameConfirmationValue)
     );
   }
 
@@ -197,7 +197,7 @@ class ServiceDestroyModal extends React.Component {
 
   getDestroyServiceModal() {
     const { open, service, intl } = this.props;
-    const serviceName = this.formatServiceName(service.getId());
+    const serviceName = this.getServiceName(service.getId());
 
     let itemText = `${StringUtil.capitalize(UserActions.DELETE)}`;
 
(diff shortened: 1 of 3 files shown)


plugins/services/src/js/components/modals/ServiceDestroyModal.js
```diff
@@ -103,7 +103,8 @@ class ServiceDestroyModal extends React.Component {
 
   getIsRightButtonDisabled() {
     return (
-      this.props.service.getId() !== this.state.serviceNameConfirmationValue
+      this.formatServiceName(this.props.service.getId()) !==
+      this.formatServiceName(this.state.serviceNameConfirmationValue)
```
jfurrow: It looks like we can reuse [this `#getName` method](https://github.com/dcos/dcos-ui/blob/master/plugins/services/src/js/structs/Service.js#L18-L20) on the `Service` struct, what do you think?

jfurrow:
It looks like we've got a legit test failure here:
```
1) Service Actions Destroy Action Application opens the correct service destroy dialog:
     CypressError: Timed out retrying: Expected to find content: '/sleep' within the element: <strong> but never did.
```

plugins/services/src/js/components/modals/ServiceDestroyModal.js
```diff
@@ -261,6 +263,26 @@ class ServiceDestroyModal extends React.Component {
...
+   * @returns {String} formatted serviceId
+   *
+   * @memberof ServiceDestroyModal
+   */
+  formatServiceName(serviceId) {
+    if (typeof serviceId !== "string") {
+      return null;
+    }
+
+    return serviceId.substring(
```
ahoskins: Do trailing slash services, ex : `"foo/"` never exist?

plugins/services/src/js/components/modals/ServiceDestroyModal.js
```diff
@@ -235,6 +236,7 @@ class ServiceDestroyModal extends React.Component {
           onChange={this.handleChangeInputFieldDestroy}
           type="text"
           value={this.state.serviceNameConfirmationValue}
+          autoFocus
```
ahoskins: Should there be an added unit test asserting this attribute on the input?

plugins/services/src/js/components/modals/ServiceDestroyModal.js
weblancaster: good question, Correct `foo/` should be always a path and not a service.

plugins/services/src/js/components/modals/ServiceDestroyModal.js
weblancaster: I believe that everything that is part of the browser default behavior should not be tested since browser vendors has put the feats through tests, that being said no need for test on our end but good question though.

ahoskins: Yeah, makes sense. 👍

plugins/services/src/js/components/modals/ServiceDestroyModal.js
```diff
@@ -261,6 +263,26 @@ class ServiceDestroyModal extends React.Component {
...
+  /**
+   * Format serviceId to only the service name
+   * e.g path/to/service/name will only return name
+   *
+   * @param {String} serviceId
+   * @returns {String} formatted serviceId
+   *
+   * @memberof ServiceDestroyModal
+   */
+  formatServiceName(serviceId) {
```
nLight: So... the name of the function is `formatServiceName` and we're passing in `serviceId`. 
From a function named `format*` I'd expect formatting the input, where this function just extracts a part of the input. So my suggestion would be to rename it `extractServiceName` or `getServiceName`

weblancaster: agree `getServiceName` is a more clear. It's addressed.

plugins/services/src/js/components/modals/ServiceDestroyModal.js
```diff
@@ -261,6 +263,26 @@ class ServiceDestroyModal extends React.Component {
...
+   * @returns {String} formatted serviceId
+   *
+   * @memberof ServiceDestroyModal
+   */
+  getServiceName(serviceId) {
+    if (typeof serviceId !== "string") {
+      return null;
+    }
+
+    return serviceId.substring(
```
Poltergeist: Also possible:
```JS
return serviceId.split("/").pop();
```
But as mentioned earlier we might not need this function.

plugins/services/src/js/components/modals/ServiceDestroyModal.js
```diff
@@ -196,7 +197,7 @@ class ServiceDestroyModal extends React.Component {
 
   getDestroyServiceModal() {
     const { open, service, intl } = this.props;
-    const serviceName = service.getId();
+    const serviceName = this.getServiceName(service.getId());
```
weblancaster: hey @Poltergeist take a look at https://github.com/dcos/dcos-ui/pull/2248#discussion_r121544548

plugins/services/src/js/components/modals/ServiceDestroyModal.js
weblancaster: Nice this is simpler. I have changed to your suggestion.

weblancaster:
great catch @Poltergeist fixing that enabled me to remove the `getServiceName` completely. Thanks testing. it's addressed.

Poltergeist:
Thanks @weblancaster this looks really great. Thanks for the changes. This makes the code a lot cleaner.

PR #2252, opened 2017-06-14
↩️[1️⃣.9️⃣] DCOS_OSS-1237: Remove Size from ExternalVolumes with docker runtime

Poltergeist:
This commit disables the possibility to set size
on an external volumes for Docker containers.

Closes DCOS_OSS-1237

**Checklist**

### DIFF
--- a/plugins/services/src/js/reducers/serviceForm/Volumes.js
+++ b/plugins/services/src/js/reducers/serviceForm/Volumes.js
@@ -21,6 +21,17 @@ const mapLocalVolumes = function (volume) {
   };
 };
 
+const mapExternalVolumes = function (volume) {
+  if (this.docker &&
+    this.dockerType === 'DOCKER' && volume.external.size != null) {
+    return Object.assign({},
+      volume,
+      {external: Object.assign({}, volume.external, {size: null})});
+  }
+
+  return volume;
+};
+
 const filterHostVolumes = function (volume) {
   return volume.type !== 'HOST' || this.docker;
 };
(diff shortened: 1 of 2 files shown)


plugins/services/src/js/reducers/serviceForm/Volumes.js
```diff
@@ -21,6 +22,17 @@ const mapLocalVolumes = function (volume) {
   };
 };
 
+const mapExternalVolumes = function (volume) {
+  if (this.docker &&
+    this.dockerType === 'DOCKER' && volume.external.size != null) {
+    return Object.assign({},
```
orlandohohmeier: Please respect the `omit` interface and remove the second `Object.assign` as it isn't needed.

`Object.assign({},  volume, {external: omit(volume.external, ['size']))});`

plugins/services/src/js/reducers/serviceForm/Volumes.js
```diff
@@ -167,7 +186,9 @@ function reduceVolumes(state, {type, path, value}) {
     this.localVolumes
       .filter(filterHostVolumes.bind(this))
       .map(mapLocalVolumes),
-    this.externalVolumes);
+    this.externalVolumes.map(
+      mapExternalVolumes.bind(this)
```
orlandohohmeier: Wonder if it makes sense to only pass the needed properties instead of the entire internal state...

plugins/services/src/js/reducers/serviceForm/Volumes.js
Poltergeist: Is addressed I had to `omit` the `omit` but instead do the second Object.assign. because we need the `size: null`

Poltergeist:
run integration tests

PR #2253, opened 2017-06-14
DCOS-16190: Skip non-string values in env var parsers

nLight:
This PR makes DC/OS parsers to parse string values for environment variables and let the plugins parse/reduce the rest.

To test: 
Create a secret and give it a hard time with a lot of Env vars and secrets. Switching between JSON editor and form. Add a var in the form edit in the JSON, delete in the form, add in the JSON. You know the grind.

### DIFF
--- a/plugins/services/src/js/reducers/serviceForm/JSONReducers/EnvironmentVariables.js
+++ b/plugins/services/src/js/reducers/serviceForm/JSONReducers/EnvironmentVariables.js
@@ -63,20 +63,26 @@ module.exports = {
       return [];
     }
 
-    return Object.keys(state.env).reduce(function(memo, key, index) {
-      /**
-       * For the environment variables which are a key => value based object
-       * we want to create a new item and fill it with the key and the
-       * value. So we need 3 transactions for each key value pair.
-       * 1) Add a new Item to the path with the index equal to index.
-       * 2) Set the key on the path `env.${index}.key`
-       * 3) Set the value on the path `env.${index}.value`
-       */
-      memo.push(new Transaction(["env"], index, ADD_ITEM));
-      memo.push(new Transaction(["env", index, "key"], key, SET));
-      memo.push(new Transaction(["env", index, "value"], state.env[key], SET));
+    return Object.keys(state.env)
+      .filter(function(key) {
+        return typeof state.env[key] === "string";
+      })
+      .reduce(function(memo, key, index) {
+        /**
+         * For the environment variables which are a key => value based object
+         * we want to create a new item and fill it with the key and the
+         * value. So we need 3 transactions for each key value pair.
+         * 1) Add a new Item to the path with the index equal to index.
+         * 2) Set the key on the path `env.${index}.key`
+         * 3) Set the value on the path `env.${index}.value`
+         */
+        memo.push(new Transaction(["env"], index, ADD_ITEM));
+        memo.push(new Transaction(["env", index, "key"], key, SET));
+        memo.push(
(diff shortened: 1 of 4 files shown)


plugins/services/src/js/reducers/serviceForm/JSONReducers/EnvironmentVariables.js
```diff
@@ -63,20 +63,26 @@ module.exports = {
       return [];
     }
 
-    return Object.keys(state.env).reduce(function(memo, key, index) {
-      /**
+    return Object.keys(state.env)
+      .filter(function(key) {
+        return typeof state.env[key] === "string";
```
Poltergeist: We could instead of the filter do this in the reduce like 

```JS
if(typeof state.env[key] === "string") {
  return memo;
}
```
With this we would remove half of the cycles above the env vars. With the same result.

plugins/services/src/js/reducers/serviceForm/JSONReducers/MultiContainerEnvironmentVariables.js
```diff
@@ -7,22 +7,26 @@ module.exports = {
       return [];
     }
 
-    return Object.keys(state.environment).reduce(function(memo, key, index) {
-      /**
+    return Object.keys(state.environment)
+      .filter(function(key) {
+        return typeof state.environment[key] === "string";
```
Poltergeist: Same like the other comment.

plugins/services/src/js/reducers/serviceForm/JSONReducers/__tests__/MultiContainerEnvironmentVariables-test.js
```diff
@@ -19,16 +19,12 @@ describe("Environment Variables", function() {
       ]);
     });
 
-    it("should keep complex values", function() {
+    it("should skip complex values", function() {
```
Poltergeist:
```JS
it("skips complex values", function() {
```

plugins/services/src/js/reducers/serviceForm/JSONReducers/MultiContainerEnvironmentVariables.js
```diff
@@ -7,22 +7,26 @@ module.exports = {
...
+      })
+      .reduce(function(memo, key, index) {
+        /**
```
Poltergeist: Is the comment below also right indented?

plugins/services/src/js/reducers/serviceForm/JSONReducers/EnvironmentVariables.js
nLight: I prefer a combination of a higher order functions instead of nested conditionals.

plugins/services/src/js/reducers/serviceForm/JSONReducers/MultiContainerEnvironmentVariables.js
nLight: oops

plugins/services/src/js/reducers/serviceForm/JSONReducers/__tests__/MultiContainerEnvironmentVariables-test.js
nLight: I'll change all of them then

nLight:
Hey @Poltergeist I addressed your comments!

nLight:
run integration tests

nLight:
@Poltergeist pong!

PR #2254, opened 2017-06-14
DCOS_OSS-1238: fix(externalVolumes): Remove size if docker

Poltergeist:
This commit will introduce that only external
volumes in the MESOS container type are allowed to
have a size.

Closes DCOS_OSS-1238

**Checklist**

### DIFF
--- a/plugins/services/src/js/components/forms/VolumesFormSection.js
+++ b/plugins/services/src/js/components/forms/VolumesFormSection.js
@@ -7,6 +7,7 @@ import FieldError from "#SRC/js/components/form/FieldError";
 import FieldInput from "#SRC/js/components/form/FieldInput";
 import FieldLabel from "#SRC/js/components/form/FieldLabel";
 import FieldSelect from "#SRC/js/components/form/FieldSelect";
+import { findNestedPropertyInObject } from "#SRC/js/utils/Util";
 import FormGroup from "#SRC/js/components/form/FormGroup";
 import FormGroupContainer from "#SRC/js/components/form/FormGroupContainer";
 import FormGroupHeading from "#SRC/js/components/form/FormGroupHeading";
@@ -16,12 +17,16 @@ import FormRow from "#SRC/js/components/form/FormRow";
 import Icon from "#SRC/js/components/Icon";
 import MetadataStore from "#SRC/js/stores/MetadataStore";
 
+import ContainerConstants from "../../constants/ContainerConstants";
+
 import {
   FormReducer as externalVolumes
 } from "../../reducers/serviceForm/FormReducers/ExternalVolumes";
 import {
   FormReducer as localVolumes
 } from "../../reducers/serviceForm/FormReducers/LocalVolumes";
+
+const { type: { DOCKER } } = ContainerConstants;
 
 const errorsLens = Objektiv.attr("container", {}).attr("volumes", []);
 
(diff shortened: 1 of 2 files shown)


nLight:
Let's adjust the tooltip wording to align with https://mesosphere.github.io/marathon/docs/external-volumes.html and maybe even put the link in there.

Poltergeist:
Does 
```
Docker Runtime only supports the default size for implicit volumes, please select Mesos Runtime if you want to modify the size.
```
Sound better to you?

plugins/services/src/js/components/forms/VolumesFormSection.js
```diff
@@ -242,6 +243,38 @@ class VolumesFormSection extends Component {
         .at(key + offset, {})
         .get(this.props.errors).containerPath;
 
+      const dockerType = findNestedPropertyInObject(
```
nLight: name suggestion `runtimeType`

plugins/services/src/js/components/forms/VolumesFormSection.js
```diff
@@ -242,6 +243,38 @@ class VolumesFormSection extends Component {
...
+        this.props.data,
+        "container.type"
+      );
+
+      let sizeField = (
+        <Tooltip
+          content="Docker Runtime only supports the default size for implicit volumes, please select Mesos Runtime if you want to modify the size."
```
nLight: Do we want to i18n that one?

plugins/services/src/js/components/forms/VolumesFormSection.js
```diff
@@ -242,6 +243,38 @@ class VolumesFormSection extends Component {
...
+          <FieldInput
+            name={`externalVolumes.${key}.size`}
+            type="number"
+            disabled={true}
+            value={""}
+          />
+        </Tooltip>
+      );
+
+      if (dockerType !== "DOCKER") {
```
nLight: Let's use a constant here

plugins/services/src/js/reducers/serviceForm/JSONReducers/Volumes.js
```diff
@@ -21,6 +22,20 @@ const mapLocalVolumes = function(volume) {
   };
 };
 
+const mapExternalVolumes = function(volume) {
+  if (
+    this.docker &&
+    this.dockerType === "DOCKER" &&
```
nLight: Let's use a constant `DOCKER`

plugins/services/src/js/reducers/serviceForm/JSONReducers/Volumes.js
```diff
@@ -48,6 +63,10 @@ function reduceVolumes(state, { type, path, value }) {
     this.docker = false;
   }
 
+  if (this.dockerType == null) {
+    this.dockerType = "DOCKER";
```
nLight: Let's use a constant `DOCKER`

plugins/services/src/js/reducers/serviceForm/JSONReducers/Volumes.js
```diff
@@ -48,6 +63,10 @@ function reduceVolumes(state, { type, path, value }) {
     this.docker = false;
   }
 
+  if (this.dockerType == null) {
```
nLight: name suggestion `runtime` or `runtimeType`

plugins/services/src/js/reducers/serviceForm/JSONReducers/Volumes.js
nLight: Does this check still make sense since we've merged UCR and MESOS? Feels weird to double check on `dockerType` (runtime type) and on a boolean `docker`

plugins/services/src/js/components/forms/VolumesFormSection.js
weblancaster: would be nice if we could

ahoskins:
I checked it out, and worked as expected for my fresh eyes. I agree with all of @nLight 's comments. 👍

plugins/services/src/js/components/forms/VolumesFormSection.js
Poltergeist: Well since it is in the create service modal which is not below the provider in the react tree it is not so simple. We should fix this in another PR.

orlandohohmeier:
LGTM! 🚢

PR #2257, opened 2017-06-14
fix(DCOS-15288): override uppercase key for mesos agent attributes

ahoskins:
**DCOS-15288**: fixes capitalized mesos agent attributes. It does this by introducing a new class `"configuration-map-attribute-label"` that resets the CSS text-transform.

**before**:

screen shot 2017-06-14 at 1 40 23 pm

**after**:

screen shot 2017-06-14 at 1 39 21 pm

**Checklist**

### DIFF
--- a/src/js/components/ConfigurationMapLabel.js
+++ b/src/js/components/ConfigurationMapLabel.js
@@ -1,12 +1,10 @@
+import classNames from "classnames";
 import React from "react";
 
 const ConfigurationMapLabel = props => {
-  let labelClasses;
-  if (props.isAttribute) {
-    labelClasses = "configuration-map-label configuration-map-attribute-label";
-  } else {
-    labelClasses = "configuration-map-label";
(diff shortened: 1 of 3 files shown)


src/js/components/ConfigurationMapLabel.js
```diff
@@ -1,11 +1,22 @@
 import React from "react";
 
 const ConfigurationMapLabel = props => {
+  let labelClasses;
+  if (props.isAttribute) {
+    labelClasses = "configuration-map-label configuration-map-attribute-label";
+  } else {
+    labelClasses = "configuration-map-label";
+  }
```
ahoskins: Yes, very nice! 😄  Changed in next commit

ahoskins:
@Poltergeist I agree. Changed prop name to `keepTextCase`. It's the best I can think of - open to other suggestions.

PR #2261, opened 2017-06-15
fix(MARATHON_EE-1538): Fix service containers host/port empty space.

weblancaster:
**MARATHON_EE-1538** This PR fixes the issue when no host/port available the "table" space will be blank causing misalignment.

To test create a service with the following json config then go to the service and expand the service:

```
{
  "id": "/rpgqa/qa1/champion-mastery-memcached",
  "version": "2017-06-07T23:43:24.644Z",
  "containers": [
    {
      "name": "sidecar",
      "exec": {
        "command": {
          "shell": "sleep 10000000000"
        }
      },
      "resources": {
        "cpus": 0.1,
        "mem": 32,
        "disk": 0,
        "gpus": 0
      },
      "image": {
        "kind": "DOCKER",
        "id": "alpine"
      }
    },
    {
      "name": "memcached",
      "exec": {
        "command": {
          "shell": "sleep 1000000000"
        }
      },
      "resources": {
        "cpus": 0.1,
        "mem": 32,
        "disk": 0,
        "gpus": 0
      },
      "endpoints": [
        {
          "name": "main",
          "containerPort": 11211,
          "hostPort": 11211,
          "protocol": [
            "tcp"
          ]
        }
      ],
      "image": {
        "kind": "DOCKER",
        "id": "alpine"
      },
      "user": "root"
    }
  ],
  "networks": [
    {
      "name": "dcos",
      "mode": "container"
    }
  ],
  "scaling": {
    "kind": "fixed",
    "instances": 1
  },
  "scheduling": {
    "backoff": {
      "backoff": 1,
      "backoffFactor": 1.15,
      "maxLaunchDelay": 3600
    },
    "upgrade": {
      "minimumHealthCapacity": 1,
      "maximumOverCapacity": 1
    },
    "unreachableStrategy": {
      "inactiveAfterSeconds": 300,
      "expungeAfterSeconds": 600
    }
  },
  "executorResources": {
    "cpus": 0.1,
    "mem": 32,
    "disk": 10
  }
}
```

**Checklist**

### DIFF
--- a/src/styles/components/expanding-table/styles.less
+++ b/src/styles/components/expanding-table/styles.less
@@ -85,6 +85,10 @@
 
   .expanding-table-child {
 
+    > div {
+      height: 25px;
+    }
+
     &:last-child {
 
       .expanding-table-primary-cell-heading {


src/styles/components/expanding-table/styles.less
```diff
@@ -85,6 +85,10 @@
 
   .expanding-table-child {
 
+    > div {
```
weblancaster: the js solution in place isn't working I have tried to use another ways through js but none of them worked so I moved to the simple fix since this is a very particular html structure (one depending on another) specifying the child div is sufficient.

ahoskins: should this use the `@expanding-table-line-height` variable instead of 25px hardcoded?

weblancaster: I believe we shouldn't use since it's specifically the line height for `.expanding-table td` and it's probably higher height than the `25px` but if there is enough people asking to use that base after design agreement I'm fine switching.

ahoskins: I see. Yeah, `@expanding-table-line-height` is a bit bigger...I was just curious where the 25px came from or if there was a better constant to use.

weblancaster: Came from the rendered height which when using the Chrome inspector it tells you the height.

nLight:
The affected version in the ticket is 1.9, are we going to back port it?

src/styles/components/expanding-table/styles.less
nLight: for logs column we just render 
```
if (rowOptions.isParent) {
  // Because elements are just stacked we need a spacer
  return <span>&nbsp;</span>;
}
```
a bit uncomfortable to style all the divs with `> div`

weblancaster: @nLight please take a look at this comment. https://github.com/dcos/dcos-ui/pull/2261#discussion_r122276957

I have tried that solution and also other options with js but none seemed.

The solution in place using css is very specific/edge (take a look in the selector hierarchy) case and won't bleed out to another markup. I can always add more specificity using `> div::first-child`.

weblancaster:
@nLight https://github.com/dcos/dcos-ui/pull/2263

src/styles/components/expanding-table/styles.less
nLight: I didn't quite realize that you've been discussing that solution. I ok merging it, thanks for trying out different solutions here 💪

PR #2262, opened 2017-06-16
DCOS-16274:  Add Intl polyfill

orlandohohmeier:
Add a Intl (ECMA-402) polyfill to fix application failures due to the missing API.

Closes DCOS-16274

### DIFF
--- a/src/js/index.js
+++ b/src/js/index.js
@@ -125,10 +125,23 @@ RequestUtil.json = function(options = {}) {
     );
   }
 
-  // Plugins events
-  PluginSDK.Hooks.addAction("pluginsConfigured", onPluginsLoaded);
-  ConfigStore.addChangeListener(CONFIG_ERROR, onConfigurationError);
+  function startApplication() {
+    // Plugins events
+    PluginSDK.Hooks.addAction("pluginsConfigured", onPluginsLoaded);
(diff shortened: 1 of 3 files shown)


nLight:
I was able to login to OSS Cluster using Firefox 28 that shouldn't have intl api

orlandohohmeier:
Thanks for the review! Will merge once the system tests pass.

orlandohohmeier:
ℹ️  Please note that I changed the base to `release/1.10`

src/js/index.js
```diff
@@ -125,10 +125,23 @@ RequestUtil.json = function(options = {}) {
...
+    ConfigStore.addChangeListener(CONFIG_ERROR, onConfigurationError);
 
-  // Load configuration
-  ConfigStore.fetchConfig();
+    // Load configuration
+    ConfigStore.fetchConfig();
+  }
+
+  if (!global.Intl) {
+    require.ensure(["intl", "intl/locale-data/jsonp/en.js"], function(require) {
```
weblancaster: we can store `intl/locale-data/jsonp/en.js` and re-use it. not big deal though.

orlandohohmeier: Good point. It might even makes sense to create a list of needed locals and use that to require the respective changes.

PR #2264, opened 2017-06-16
fix(DCOS-16286): Fix task detail view marathon task status

ahoskins:
This fixes [DCOS-16286](https://jira.mesosphere.com/browse/DCOS-16286) by changing the field name from `status` to `state`, and using the `TaskStates` map to render the proper display name.

**After**:
screen shot 2017-06-16 at 2 31 22 pm

**Checklist**

### DIFF
--- a/plugins/services/src/js/components/MarathonTaskDetailsList.js
+++ b/plugins/services/src/js/components/MarathonTaskDetailsList.js
@@ -20,11 +20,11 @@ class MarathonTaskDetailsList extends React.Component {
   }
 
   getTaskStatus(task) {
-    if (task == null || task.status == null) {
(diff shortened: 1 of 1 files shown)


plugins/services/src/js/components/MarathonTaskDetailsList.js
```diff
@@ -20,11 +20,11 @@ class MarathonTaskDetailsList extends React.Component {
   }
 
   getTaskStatus(task) {
-    if (task == null || task.status == null) {
+    if (task == null || task.state == null) {
       return "Unknown";
     }
 
-    return task.status;
+    return task.state;
```
weblancaster: need to be mapped to `TaskStates.js`

PR #2268, opened 2017-06-19
Adjust the Metronome data handling

orlandohohmeier:
Change the `MetronmeStore` and `MetronmeActions` to fix issue with the data handling. All data received from the respective actions in now stored in a single array to keep the data always up to date and in sync. The store then parses the data into a `JobTree` instead of the actions providing pre-processed data.

BREAKING CHANGE:  The `REQUEST_METRONOME_JOBS_SUCCESS` as well as the `REQUEST_METRONOME_JOB_DETAIL_SUCCESS` event emitted by `MetronomeActions` have changed to fix issue with the data handling.

To migrate event handlers please  follow the example below:

Before:

```
function handleJobsSuccess(event) {
     console.log(new JobTree(event.data));
}
```

```
function handleJobDetailSuccess(event) {
     console.log(new Job(event.data));
}
```

After:

```
function handleJobsSuccess(event) {
     console.log(new JobTree(MetronomeUtil.parseJobs(event.data)));
}
```

```
function handleJobDetailSuccess(event) {
     console.log(new Job(MetronomeUtil.parseJob(event.data)));
}
```

### DIFF
--- a/src/js/stores/MetronomeStore.js
+++ b/src/js/stores/MetronomeStore.js
@@ -44,6 +44,7 @@ import {
 } from "../constants/EventTypes";
 import Config from "../config/Config";
 import MetronomeActions from "../events/MetronomeActions";
+import MetronomeUtil from "../utils/MetronomeUtil";
 import Job from "../structs/Job";
 import JobTree from "../structs/JobTree";
 import VisibilityStore from "./VisibilityStore";
@@ -80,8 +81,8 @@ class MetronomeStore extends EventEmitter {
     super(...arguments);
 
     this.data = {
-      jobDetail: {},
-      jobTree: { id: "/", items: [] }
+      jobMap: new Map(),
+      jobTree: null
     };
 
     PluginSDK.addStoreConfig({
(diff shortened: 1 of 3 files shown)


src/js/stores/MetronomeStore.js
```diff
@@ -130,7 +130,7 @@ class MetronomeStore extends EventEmitter {
           this.emit(METRONOME_JOB_DELETE_SUCCESS, action.jobID);
           break;
         case REQUEST_METRONOME_JOB_DETAIL_SUCCESS:
-          this.data.jobDetail[action.jobID] = action.data;
+          this.data.jobs = [].concat(this.data.jobs, action.data);
```
weblancaster: This can be also written using spread operator `[...this.data.jobs, ...action.data]`

nLight:
The same 9 Jobs related tests are failing is it expected?

orlandohohmeier:
@nLight the test actually fail due to an issue with the data merging which occurs when the data provided by `/jobs` isn't inline with `/jobs/detail`. I'll adjust the `MetronomeStore` to properly handle that case.

orlandohohmeier:
@nLight I've adjusted the internal data handling to solve the respective edge case.

PR #2269, opened 2017-06-19
DCOS-13054: click container in pod takes you to container config

ahoskins:
Clicking a container in a pod now takes you to that container's config page. This was brought up in user testing [here](https://jira.mesosphere.com/browse/DCOS-13054). 

After:
https://cl.ly/2O1W453N072V

**Checklist**

### DIFF
--- a/plugins/services/src/js/components/forms/GeneralServiceFormSection.js
+++ b/plugins/services/src/js/components/forms/GeneralServiceFormSection.js
@@ -183,10 +183,9 @@ class GeneralServiceFormSection extends Component {
           key={index}
           onRemove={this.props.onRemoveItem.bind(this, {
             value: index,
-            path: "containers",
-            noPropagate: true
+            path: "containers"
           })}
-          onClick={this.props.onClickContainer.bind(this, `container${index}`)}
+          onClick={this.props.onClickItem.bind(this, `container${index}`)}
         >
           {item.name || `container-${index + 1}`}
         </FormGroupContainer>
@@ -671,7 +670,7 @@ GeneralServiceFormSection.propTypes = {
   errors: React.PropTypes.object,
   onAddItem: React.PropTypes.func,
   onRemoveItem: React.PropTypes.func,
-  onClickContainer: React.PropTypes.func
+  onClickItem: React.PropTypes.func
 };
 
 GeneralServiceFormSection.configReducers = General;
--- a/plugins/services/src/js/components/forms/GeneralServiceFormSection.js
+++ b/plugins/services/src/js/components/forms/GeneralServiceFormSection.js
@@ -185,6 +185,7 @@ class GeneralServiceFormSection extends Component {
             value: index,
             path: "containers"
           })}
+          onClick={this.props.onClickContainer.bind(this, `container${index}`)}
         >
           {item.name || `container-${index + 1}`}
         </FormGroupContainer>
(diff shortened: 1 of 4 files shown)


plugins/services/src/js/components/forms/GeneralServiceFormSection.js
```diff
@@ -183,7 +183,8 @@ class GeneralServiceFormSection extends Component {
           key={index}
           onRemove={this.props.onRemoveItem.bind(this, {
             value: index,
-            path: "containers"
+            path: "containers",
+            noPropagate: true
```
ahoskins: I want the `onRemoveItem()` implementation to stop event propagation when invoked from this context (because otherwise it will trigger the container click event too). There might be a better way of doing this though.

plugins/services/src/js/components/modals/NewCreateServiceModalForm.js
```diff
@@ -680,6 +689,7 @@ class NewCreateServiceModalForm extends Component {
                         onConvertToPod={onConvertToPod}
                         service={service}
                         onRemoveItem={this.handleRemoveItem}
+                        onClickContainer={this.handleClickContainer}
```
weblancaster: Let's keep it consistent with the other names `onRemoveItem` `onAddItem ` so I would name `onClickItem` and `handleClickItem`

plugins/services/src/js/components/modals/NewCreateServiceModalForm.js
```diff
@@ -279,17 +280,25 @@ class NewCreateServiceModalForm extends Component {
     this.setState({ batch, appConfig: this.getAppConfig(batch) });
   }
 
-  handleRemoveItem(event) {
-    const { value, path } = event;
+  handleRemoveItem(event, e) {
+    const { value, path, noPropagate } = event;
     let { batch } = this.state;
 
+    if (noPropagate) {
```
weblancaster: I believe the `noPropagate` condition is not needed here since the browser know how to handle the click on the specific element.

plugins/services/src/js/components/forms/GeneralServiceFormSection.js
```diff
@@ -183,8 +183,10 @@ class GeneralServiceFormSection extends Component {
...
```
weblancaster: based on my comment below this is not needed.

weblancaster:
This is great! really liked the UX.

I just left a couple comments @ahoskins

plugins/services/src/js/components/modals/NewCreateServiceModalForm.js
ahoskins: Yup, good idea.

plugins/services/src/js/components/modals/NewCreateServiceModalForm.js
weblancaster: what you can do is

```
onRemoveItem={(options, event) => {
  event.stopPropagation();
  this.handleRemoveItem(options);
}}
```

jfurrow:
Awesome! Good idea!

ahoskins:
Thanks for the review @weblancaster. I changed those names to match the others (`onClickItem`, etc), and put the `noPropagate` inline as suggested in the anonymous function!

weblancaster:
retest this please

PR #2276, opened 2017-06-21
bug(Tables): Clean up table styling

jfurrow:
This PR fixes three small bugs related to table styling:
* toggle `visibility` instead of `display` property of sorting indicator carets
* right-align resource values in `TaskTable`
* remove sorting indicator for `logs` column

Before:

After:

**Checklist**


ahoskins:
Played with it on different screen sizes and looks good 🌞

PR #2277, opened 2017-06-21
DCOS-16009: fix(ServiceActionsModal): Remove duplicated word 'documentation'

jfurrow:
This PR just removes the duplicated word `documentation` in the service actions modal.

Before:

After:

**Checklist**


ahoskins:
Unless they meant it was documentation for the documentation? 😉  

LGTM 🚢

PR #2278, opened 2017-06-21
fix(DCOS-16363): Content in table cells overflowing

tnguyen0:
This PR wraps content in table cells.

Before:
https://cl.ly/3Z0V3v260Q31

After:
https://cl.ly/3d2h0h2r3X3n

**Checklist**


tnguyen0:
Thoughts @leemunroe?

leemunroe:
Spoke offline with @tnguyen0 and in the case of the permissions I agree it makes sense to wrap.

For other instances I would hold off until we learn if is an issue or not e.g. for IDs truncation works

PR #2281, opened 2017-06-21
DCOS-14185: chore(ServicesTable): Animate Waiting status

jfurrow:
This PR will render the animated health bar when a service's status is either `deploying` or `waiting`.

I also adjusted the way we determine the `isDeploying` property that's passed to `HealthBar`. Previously, we were comparing the displayed status string to a hard-coded string (`Deploying`). I've changed this to compare the status' `key` property to a constants file that maps to these keys.

It seems like we need to refactor the service status constants, as they are spread across multiple files, but this solves the issue and gets us closer to a more desirable solution, IMO.

Before:

After:

```
<!-- Suplemental codeblock -->
```

**Checklist**

### DIFF
--- a/plugins/services/src/js/containers/services/ServicesTable.js
+++ b/plugins/services/src/js/containers/services/ServicesTable.js
@@ -29,6 +29,7 @@ import {
   SUSPEND
 } from "../../constants/ServiceActionItem";
 import ServiceActionLabels from "../../constants/ServiceActionLabels";
+import ServiceStatusTypes from "../../constants/ServiceStatusTypes";
 import ServiceStatusWarning from "../../components/ServiceStatusWarning";
 import ServiceTableHeaderLabels from "../../constants/ServiceTableHeaderLabels";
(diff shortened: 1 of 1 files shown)


plugins/services/src/js/containers/services/ServicesTable.js
```diff
@@ -329,11 +330,14 @@ class ServicesTable extends React.Component {
   renderStatus(prop, service) {
     const instancesCount = service.getInstancesCount();
     const serviceId = service.getId();
-    const serviceStatus = service.getStatus();
-    const serviceStatusClassSet = StatusMapping[serviceStatus] || "";
+    const serviceStatusText = service.getStatus();
+    const serviceStatusClassSet = StatusMapping[serviceStatusText] || "";
+    const { key: serviceStatusKey } = service.getServiceStatus();
```
ahoskins: Cool, didn't know you could unpack object + rename.

PR #2283, opened 2017-06-23
test: fix broken apps system tests

ahoskins:
This PR fixes the apps system tests (prior to this, 11/15 were failing).  This PR is a good pre-requisite to my main sprint tasks of adding a couple new system tests. 

The tests were failing because the JSON structure was changed, and a couple text fields on the UI have been re-labeled. **After this change, all these tests should be passing**. However, these tests are still pretty brittle because they depend on JSON structure, and many string values of labels on the pages. They look suspiciously like integration tests, and we should discuss how to approach system tests moving forward.  

**Checklist**

### DIFF
--- a/system-tests/services/test-apps.js
+++ b/system-tests/services/test-apps.js
@@ -60,7 +60,17 @@ describe("Services", function() {
           cmd: cmdline,
           cpus: 0.1,
           mem: 10,
-          instances: 1
+          instances: 1,
+          portDefinitions: [],
+          container: {
+            type: "MESOS",
+            volumes: []
+          },
+          requirePorts: false,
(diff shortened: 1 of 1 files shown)


orlandohohmeier:
@ahoskins could you please align the commit message with our contribution guideline?

orlandohohmeier:
The `pod` and `job` test still fail but all the `app` ones pass. Great Job! ⚒ 💯

ahoskins:
Updated the commit name! @orlandohohmeier, github seems to have automatically dismissed your review since the commit is updated. Also, @tnguyen0, can you take a look at this please?

PR #2285, opened 2017-06-26
test(DCOS-15049): add two UCR container tests using docker and command

ahoskins:
This adds system test coverage for a UCR runtime in two scenarios: 

1) using a Docker container and command 
2) using just a command

Most of this is direct copy-paste from the `"should create an app with docker config"` test, and just modifies by adding `selectMesosRuntime()` and changes a couple assertions to expect the mesos runtime. 

Run with:
```
    -v `pwd`:/dcos-ui \
    mesosphere/dcos-ui dcos-system-test-driver --only "apps" -v \
```

**Checklist**

### DIFF
--- a/system-tests/services/test-apps.js
+++ b/system-tests/services/test-apps.js
@@ -771,6 +771,410 @@ describe("Services", function() {
         .should("have.value", "http");
     });
 
+    it("should create an app with ucr config and docker container", function() {
+      const serviceName = "app-with-ucr-config-and-docker-container";
+      const cmdline = "python3 -m http.server 8080";
+
+      // Select 'Single Container'
+      cy.contains("Single Container").click();
+
+      // Fill-in the input elements
+      cy
+        .root()
+        .getFormGroupInputFor("Service ID *")
+        .type(`{selectall}{rightarrow}${serviceName}`);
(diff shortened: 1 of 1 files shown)


system-tests/services/test-apps.js
```diff
@@ -771,6 +771,410 @@ describe("Services", function() {
...
+      cy.root().get(".menu-tabbed-item").contains("Networking").click();
+
+      // Select "Bridge"
+      cy
+        .root()
+        .getFormGroupInputFor("Network Type")
+        .should("have.value", "BRIDGE");
+
+      // Click "Add Service Endpoint"
+      cy.contains("Add Service Endpoint").click();
```
tnguyen0: Extra action here that is not required.

system-tests/services/test-apps.js
```diff
@@ -771,6 +771,410 @@ describe("Services", function() {
...
```
tnguyen0: Here as well.

system-tests/services/test-apps.js
ahoskins: Yes, good catch.

ahoskins:
run integration tests

PR #2289, opened 2017-06-28
fix(InstallPackageModal): add tighter constraints for auto-install

MatApple:
**Problem:**
Package will attempt to auto-install if window is resized while viewing the package detail page.

**Solution:**
InstallPackageModal `#componentwillReceiveProps` get's called when browser resize occurs. We only care about setting new component state if modal is opening or closing. We do **not** care if the modal is not in a transition. So we return early in the method if the component is not in a state of "transition" [opening/closing].

**Testing:**
Navigate to package detail page from catalog and resize window. Package previously attempted to send an API request for install when resizing but it **should no longer do this**.

**Note:**
I would love to clean this component up including the logic in `#componentwillReceiveProps`, however we don't have time to boy-scout this one.

**Checklist**

### DIFF
--- a/src/js/components/modals/InstallPackageModal.js
+++ b/src/js/components/modals/InstallPackageModal.js
@@ -79,6 +79,12 @@ class InstallPackageModal
   componentWillReceiveProps(nextProps) {
     super.componentWillReceiveProps(...arguments);
     const { props } = this;
+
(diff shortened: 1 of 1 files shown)


ahoskins:
run integration tests

PR #2291, opened 2017-06-29
docs(PR): Improve PR template to contain i18n reminder/checkbox

weblancaster:
**Checklist**


ahoskins:
As I've been working on all these systems tests related tasks, I also was wondering if we should add a `"Did you add new system tests?"` checkbox? Would be a good reminder that system tests exist, and have to be updated in certain cases.

MatApple:
@ahoskins sounds like a good idea 👍

weblancaster:
I'm closing in favor of keeping the PRs clean. Whenever we decide i18n is part the workflow we can put a PR back.

PR #2293, opened 2017-07-03
System tests: adjust JSON expectations to reality

nLight:
I've changed the way we remove empty values from JSON our expectations went broken. This PR fixes that.

### DIFF
--- a/system-tests/services/test-pods.js
+++ b/system-tests/services/test-pods.js
@@ -66,7 +66,14 @@ describe("Services", function() {
             {
               mode: "host"
             }
-          ]
+          ],
+          volumes: [],
+          fetch: [],
+          scheduling: {
+            placement: {
+              constraints: []
+            }
+          }
         }
       ]);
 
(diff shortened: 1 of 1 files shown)


system-tests/services/test-pods.js
```diff
@@ -1887,7 +1942,14 @@ describe("Services", function() {
...
             }
-          ]
+          ],
+          "volumes": [],
+          "fetch": [],
+          "scheduling": {
+            "placement": {
+              "constraints": []
+            }
+          }
```
tnguyen0: No quotes :)

nLight: Please expand the piece of code it belongs to. In this particular place we actually need quotes.

nLight:
I bet those are failing because of #2295 I will rebase and see if it resolves the issue.

orlandohohmeier:
LGTM 💯 ✨

system-tests/services/test-pods.js
tnguyen0: You are correct!

PR #2297, opened 2017-07-05
DCOS-12615: autofocus service form

bstavroulakis:
This component is introduced for autofocusing input/ textarea fields.

Usage
```
<FieldAutofocus>
  <FieldInput name="id" type="text" value={data.id} />
</FieldAutofocus>
```

Once the FieldInput, FieldTextarea, input or textarea element is mounted, the document will focus on this element.
If another element with the FieldAutofocus container is already focused it won't focus on that element.

The FieldAutofocus is implemented on the NewCreateServiceModalForm for the Services Form Modal.

Closes DCOS-12615, DCOS-12613

**Checklist**

### DIFF
--- a/src/js/components/form/FieldAutofocus.js
+++ b/src/js/components/form/FieldAutofocus.js
@@ -11,7 +11,7 @@ class FieldAutofocus extends React.Component {
       return;
     }
     const input = DOMUtils.getInputElement(ReactDOM.findDOMNode(this));
-    if (!input) {
+    if (!input || input === document.activeElement) {
       return;
     }
     lockFieldAutofocus = true;
--- a/src/js/components/form/FieldAutofocus.js
+++ b/src/js/components/form/FieldAutofocus.js
@@ -3,7 +3,6 @@ import ReactDOM from "react-dom";
 import DOMUtils from "#SRC/js/utils/DOMUtils";
 
 let lockFieldAutofocus = false;
-const SUPPORTED_TYPES = ["FieldInput", "FieldTextarea", "input", "textarea"];
 
 class FieldAutofocus extends React.Component {
   componentDidMount() {
@@ -11,7 +10,7 @@ class FieldAutofocus extends React.Component {
       return;
     }
     const input = DOMUtils.getInputElement(ReactDOM.findDOMNode(this));
-    if (!input || input === document.activeElement) {
+    if (!input) {
       return;
     }
     lockFieldAutofocus = true;
@@ -27,19 +26,5 @@ class FieldAutofocus extends React.Component {
     return this.props.children;
   }
 }
-
-FieldAutofocus.propTypes = {
-  children(props, propName, componentName) {
-    const prop = props[propName];
-    if (
-      React.Children.count(prop) !== 1 ||
-      !SUPPORTED_TYPES.includes(prop.type.name)
-    ) {
-      return new Error(
-        `${componentName} should have a single child of the following types: ${SUPPORTED_TYPES.join(", ")}.`
-      );
-    }
-  }
-};
 
 module.exports = FieldAutofocus;
--- a/src/js/components/form/FieldAutofocus.js
+++ b/src/js/components/form/FieldAutofocus.js
@@ -1,0 +1,45 @@
+import React from "react";
+import ReactDOM from "react-dom";
+import DOMUtils from "#SRC/js/utils/DOMUtils";
+
+let lockFieldAutofocus = false;
+const SUPPORTED_TYPES = ["FieldInput", "FieldTextarea", "input", "textarea"];
+
+class FieldAutofocus extends React.Component {
+  componentDidMount() {
+    if (lockFieldAutofocus) {
+      return;
+    }
+    const input = DOMUtils.getInputElement(ReactDOM.findDOMNode(this));
+    if (!input) {
+      return;
+    }
+    lockFieldAutofocus = true;
+    input.focus();
+
+    input.addEventListener("blur", function onBlurInput() {
+      lockFieldAutofocus = false;
+      this.removeEventListener("blur", onBlurInput);
+    });
+  }
+
+  render() {
+    return this.props.children;
+  }
+}
+
+FieldAutofocus.propTypes = {
+  children(props, propName, componentName) {
+    const prop = props[propName];
+    if (
+      React.Children.count(prop) !== 1 ||
+      !SUPPORTED_TYPES.includes(prop.type.name)
+    ) {
+      return new Error(
+        `${componentName} should have a single child of the following types: ${SUPPORTED_TYPES.join(", ")}.`
+      );
+    }
+  }
+};
+
+module.exports = FieldAutofocus;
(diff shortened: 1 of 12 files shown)


src/js/components/form/FieldAutofocus.js
```diff
@@ -0,0 +1,45 @@
...
+    });
+  }
+
+  render() {
+    return this.props.children;
+  }
+}
+
+FieldAutofocus.propTypes = {
+  children(props, propName, componentName) {
```
nLight: I'm pretty sure this check can be implemented using PropTypes, but since prop types checking is pretty much useless and we're going to roll with either Flow or Typescript I'm ok with it.

nLight:
I think those integration tests failures are legit and should be fixed.

nLight:
Please implement the same solution in plugins

weblancaster:
@bstavroulakis if you don't mind explaining the reason behind abstracting `autofocus` into a component rather than using `autofocus` attribute?

When adding `autofocus` attribute React/jsx uses it to call `focus()` already behind the scenes so I believe we are only reinventing the wheel.
Using `findDOMNode` is also discouraged https://facebook.github.io/react/docs/react-dom.html#finddomnode

That being said, I support the solution proposed if we have had issues using `autofocus` attribute (I have used on the delete service modal without any problem).

bstavroulakis:
@weblancaster I tried the autofocus attribute as a first implementation for the stateless FieldInput. 

The first issue that came up was that if we have multiple elements on our page the document would focus on the last input of the page with this feature. In our case we would like to autofocus on the first input of the page that has that autofocus "feature". Also, when you would click on a checkbox that would render a new area with input text fields it wouldn't focus on those text inputs.

My second solution was to use the Autofocus attribute mentioned and as a boolean include a module that would either return "true" or "false". It would return true the first time, then false for the following times it was called. Once all of the elements were rendered it should switch to true again. This way it would autofocus on the first element and not on the rest. But, how do we know that all of the elements are rendered? This could probably work with a timeout. Also, this doesn't solve the other issue with the input/links that reveal new areas with inputs.

To avoid using a timeout and other non-deterministic methods or lists with items that are focused, I created a new container component that saves the state of the focused item. Once, these items are loaded I attach a "blur" event and once you navigate somewhere else the "focus" lock is reset and the document will autofocus on the next item. This way I've solved the issue of focusing on the first input only and when you click around it would aggressively find the next item.

I know it is not the best solution since we introduce a new component and use the actual DOM for the inputs but this way we don't touch any of the stateless FieldInputs and we can have some logic with the state necessary to select which element to focus on. It also aggressively focuses on the following input when you select a checkbox or a link for example that opens a new area with input fields. With the autofocus attribute, in my testing with this, this didn't work. Lastly, with this way we can remove it and include it in other areas of the site very quickly.

If you reached this part of the comment I salute you for your patience! 👍

weblancaster:
Great. Thanks for the answer @bstavroulakis

src/js/utils/__tests__/DOMUtils-test.js
```diff
@@ -283,4 +283,65 @@ describe("DOMUtils", function() {
...
+      expect(returnValue.nodeName.toLowerCase()).toEqual("textarea");
+    });
+    it("returns input element if DOM element has an input", function() {
+      const returnValue = DOMUtils.getInputElement(input);
+      expect(returnValue.nodeName.toLowerCase()).toEqual("input");
+    });
+    it("returns textarea element if DOM element has an textarea", function() {
+      const returnValue = DOMUtils.getInputElement(textarea);
+      expect(returnValue.nodeName.toLowerCase()).toEqual("textarea");
+    });
```
ahoskins: Maybe add a test case to make sure nothing breaks when multiple input or textarea elements.

src/js/components/form/FieldAutofocus.js
```diff
@@ -0,0 +1,45 @@
+import React from "react";
+import ReactDOM from "react-dom";
+import DOMUtils from "#SRC/js/utils/DOMUtils";
+
+let lockFieldAutofocus = false;
```
ahoskins: It's unclear to me what impact this field is having. I can't reproduce a scenario where it changes the behavior. In what scenario is this needed?

bstavroulakis: The lockFieldAutofocus is needed to lock the autofocus functionality to the first element that triggers it.

In our page we may have multiple fields with the autofocus functionality. For example in my Services Form every time I click "ADD ENVIRONMENT VARIABLE" I would like to focus on the first input field entered. In this case, this is the Key input field. 

This can be easily done now if we wrap the input with the FieldAutofocus

```
<FieldAutofocus>
    <FieldInput
        name={`env.${key}.key`}
        type="text"
        value={env.key}
    />
</FieldAutofocus>
``` 

image

We can wrap this new container component anywhere we would like to autofocus once the inner input/ textarea is mounted on the page.

The problem now is that if we have multiple components with the FieldAutofocus wrapper the document would focus on the last input that triggers a focus. 

In the example above if we click on the "Service" tab and go back to the "Environment" tab we would like to autofocus on the first Key input field and not the second one. To do this we set the lockFieldAutofocus to "true" and once we blur the component it is set to "false" again.

src/js/utils/__tests__/DOMUtils-test.js
bstavroulakis: Included an additional test for textareas a9490f93394eade4b4b456626981ec75fd7a51b8

bstavroulakis:
@weblancaster "ps: for tab/focus order we can use the attribute tabIndex"

I believe that the general Accessibility of the Service Form should be addressed at some point. The general flow of an HTML page and a form should be the order of its elements I think. So once you click on the tab button, or someone who uses the form through a device that doesn't support a mouse you should move over to the next element on the page.

The form now doesn't quite follow this flow. The tabindex could be a fix but I believe that we should use the tabindex as an exception in only the cases where something can't be rendered on the right spot and not as the rule. Also, the tabindex could maybe be used when we have errors to set those with a high priority index, so when you move around you can go through the errors first. 

:globe_with_meridians: :globe_with_meridians:

src/js/components/form/FieldAutofocus.js
ahoskins: I understand. That makes sense, and I was able to easily reproduce that use case locally. Thanks for clearing it up!

bstavroulakis:
Alfred please retest this 🥇

ahoskins:
Thanks for addressing my comments, and adding the extra test. To me, this works and looks fine. The custom children propType check isn't something we do in the rest of the codebase, but that doesn't seem like a bad thing IMO. But I wasn't added as reviewer so I'll leave it to the others for approval.

src/js/components/form/FieldAutofocus.js
weblancaster: I believe we can remove this, since we are moving away fro propTypes and this is a custom one so I vote to remove it.

bstavroulakis:
Alfred please retest

bstavroulakis:
Alfred please retest this

bstavroulakis:
Alfred please retest this one

PR #2300, opened 2017-07-05
fix(DCOS-11519): correct alignment of form labels with tooltips

ahoskins:
This PR fixes a spacing issue with form elements with tooltips. 

**Fix:**
I think this is the best solution. It targets the specific use of tooltips in form headings
and sets the max-height to match the label font-size, which looks like it's always 0.9rem set by canvas. 

**Before: (misalignment of service id and container image sections)**
screen shot 2017-07-05 at 10 53 02 am

**After:**
screen shot 2017-07-05 at 10 51 02 am

**Testing**:
Open create service form. Look around the rest of the UI for regression. Toggle browser window size, and check mobile user agent.

**Checklist**

### DIFF
--- a/src/styles/components/form-elements/form-group/styles.less
+++ b/src/styles/components/form-elements/form-group/styles.less
@@ -29,6 +29,10 @@
           text-overflow: ellipsis;
           white-space: nowrap;
         }
+
+        .tooltip-wrapper {
+          max-height: 0.9rem;
+        }
       }
     }
   }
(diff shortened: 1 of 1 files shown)


orlandohohmeier:

orlandohohmeier:

src/styles/components/form-elements/form-group/styles.less
```diff
@@ -29,6 +29,10 @@
           text-overflow: ellipsis;
           white-space: nowrap;
         }
+
+        .tooltip-wrapper {
+          max-height: 0.9rem;
```
weblancaster: I would only try to use a variable if there is any available with the value `0.9rem`

PR #2303, opened 2017-07-06
fix(react@15): Address unsupported attributes warnings

nLight:
This is a dcos-ui part of addressing react@15 wrong html attributes warnings.
`reactjs-components` part: https://github.com/mesosphere/reactjs-components/pull/407

They can totally go separately.

### DIFF
--- a/src/js/components/Image.js
+++ b/src/js/components/Image.js
@@ -2,8 +2,6 @@ import classNames from "classnames/dedupe";
 import React from "react";
 import ReactDOM from "react-dom";
 
-import Util from "../utils/Util";
-
 const METHODS_TO_BIND = ["onImageError"];
 
 class Image extends React.Component {
@@ -57,17 +55,13 @@ class Image extends React.Component {
     );
 
     return (
-      <img
-        className={classes}
-        onError={this.onImageError}
-        {...Util.omit(props, ["className"])}
-      />
+      <img src={props.src} className={classes} onError={this.onImageError} />
     );
   }
 }
 
 Image.propTypes = {
-  src: React.PropTypes.string,
+  src: React.PropTypes.string.required,
   fallbackSrc: React.PropTypes.string,
 
   // Classes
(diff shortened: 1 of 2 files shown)


nLight:
Alfred please retest this one

nLight:
Alfred please retest this one

src/js/components/TimeAgo.js
```diff
@@ -73,7 +73,7 @@ class TimeAgo extends React.Component {
       <time
         title={DateUtil.msToUTCDate(time)}
         dateTime={DateUtil.msToUTCDate(time)}
-        {...Util.omit(this.props, ["prefix", "suppressSuffix", "time"])}
+        {...Util.omit(this.props, Object.keys(TimeAgo.propTypes))}
```
ahoskins: Isn't this kind of strange removing all the `propTypes` from the read-only `props`? If this always returns `{}` why even have this line?

src/js/components/Image.js
```diff
@@ -60,7 +60,7 @@ class Image extends React.Component {
       <img
         className={classes}
         onError={this.onImageError}
-        {...Util.omit(props, ["className"])}
+        {...Util.omit(props, ["className", "fallbackSrc"])}
```
ahoskins: It seems like we would only ever want to pass the `src` prop, so why not do `src={props.src}` directly?

src/js/components/TimeAgo.js
nLight: This is a good question, I was just following a pattern, applying it to all the cases. I'll research how we're using the `TimeAgo` component and if we don't pass anything except what is in the propTypes I'd say we remove the line 👍

src/js/components/Image.js
nLight: I'll research if we're not missing anything 👍

nLight:
@ahoskins looks like we can just remove it in both cases. Thanks for pointing this out!

nLight:
retest this please

nLight:
Alfred please retest this one

PR #2304, opened 2017-07-07
DCOS-16606: Fix the multi-container network config edit link

bstavroulakis:
Fix for the multi-container network config edit link.

Network "Edit" links don't work on the Multi-container Review screen.

Closes DCOS-16606

review-network-edit-link-broken mov

**Checklist**

### DIFF
--- a/tests/pages/services/ServiceFormModal-cy.js
+++ b/tests/pages/services/ServiceFormModal-cy.js
@@ -1678,5 +1678,19 @@ describe("Service Form Modal", function() {
         .siblings()
         .should("to.have.length", 2);
     });
+
+    it("Should show network form when clicking on Network Configuration Edit", function() {
+      cy.get(".menu-tabbed-item").contains("Networking").click();
+      cy.get(".menu-tabbed-view .button.button-primary-link").first().click();
+      cy.get('input[name="containers.0.endpoints.0.name"]').type("test");
+      // Click review and run
+      cy
+        .get(".modal-full-screen-actions")
+        .contains("button", "Review & Run")
+        .click();
+      // Click edit to view form
+      cy.get("a.button.button-link").eq(-1).click({ force: true });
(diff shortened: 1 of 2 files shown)


tests/pages/services/ServiceFormModal-cy.js
```diff
@@ -1678,5 +1678,19 @@ describe("Service Form Modal", function() {
...
+      cy.get(".menu-tabbed-item").contains("Networking").click();
+      cy.get(".menu-tabbed-view .button.button-primary-link").first().click();
+      cy.get('input[name="containers.0.endpoints.0.name"]').type("test");
+      // Click review and run
+      cy
+        .get(".modal-full-screen-actions")
+        .contains("button", "Review & Run")
+        .click();
+      // Click edit to view form
+      cy.get("a.button.button-link").eq(-1).click({ force: true });
```
weblancaster: is the button hidden or something? because the use of `({ force: true })` is going to make our tests flaky.

bstavroulakis: Yes, the button is hidden and appears when you hover over the area.
Do you recommend I run the invoke('show') command and then click?

https://docs.cypress.io/api/commands/hover.html#Workaround

weblancaster: In that case makes sense to use it. thanks.

bstavroulakis: Yes, we will go with force then, since:

1) The invoke("show") actually uses jQuery's show method :(
2) We'll have to access the DOM and either change the visibility to visible or create a custom mouseover event and dispatch that event.

We don't use jQuery and the second method is sketchy as we move away from the abstraction of Cypress.

bstavroulakis:
@nLight The issue has been addressed. 
We will go with the "force" then! Nothing needs to change.

PR #2305, opened 2017-07-08
Removing unnecessary CNVS variable overrides.

ashenden:
Notable changes:
* Remove CNVS var overrides in `styles/vendors` directory that were doing nothing.
* Moved `icon-tiny` to main application styles and out of `styles/vendors/cnvs` directory.


src/styles/components/icons/shapes/icon-tiny/styles.less
```diff
@@ -0,0 +1,78 @@
+& when (@icon-tiny-enabled) {
```
nLight: We're missing a top level `& when (@icon-enabled)` condition here. Is it by intention?

src/styles/components/typography/styles.less
```diff
@@ -114,10 +114,6 @@
     color: @link-color;
   }
 
-  .text-small {
-    font-size: 0.7rem;
```
nLight: .small in CNVS is actually a bit bigger `@small-font-size: 0.8rem;` are you fine with having it a bit bigger in this place?

nLight:
I'd love to merge this one! Please answer my questions.

src/styles/components/icons/shapes/icon-tiny/styles.less
ashenden: @nLight yes, that is intentional.  The check is only on the component variant itself, following the pattern set for similar icons in CNVS.

src/styles/components/typography/styles.less
ashenden: Good question.  Let's stick with the 0.8rem def.

ashenden:
@nLight followed up on your comments.

PR #2306, opened 2017-07-10
fix(DCOS-16539): Fix incorrect tooltip state on health check

weblancaster:
**DCOS-16539**

Fix tooltip on health status when state is `killing`.

Steps to reproduce the issue:
Create a service with a health check where is `healthy` 
Then suspend the service
Now the `status` should be `killing` and the health should be `red dot` but with `Healthy` on the tooltip.

**before**
screen shot 2017-06-29 at 11 26 47

**after**
screen shot 2017-07-10 at 5 07 37 pm

**Checklist**

### DIFF
--- a/plugins/services/src/js/containers/tasks/TaskTable.js
+++ b/plugins/services/src/js/containers/tasks/TaskTable.js
@@ -322,15 +322,11 @@ class TaskTable extends React.Component {
 
     const dangerState = TaskStates[state].stateTypes.includes("failure");
     const activeState = TaskStates[state].stateTypes.includes("active");
+    const transitional = ["TASK_KILLING"].includes(state);
 
     const healthy = task.health === TaskHealthStates.HEALTHY;
     const unhealthy = task.health === TaskHealthStates.UNHEALTHY;
     const unknown = task.health === TaskHealthStates.UNKNOWN;
-
-    const failing = ["TASK_ERROR", "TASK_FAILED", "TASK_KILLING"].includes(
-      state
-    );
-    const running = ["TASK_RUNNING", "TASK_STARTING"].includes(state);
 
     let tooltipContent = TaskHealthStates.HEALTHY;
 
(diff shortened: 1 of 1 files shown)


plugins/services/src/js/containers/tasks/TaskTable.js
```diff
@@ -327,19 +327,21 @@ class TaskTable extends React.Component {
     const unhealthy = task.health === TaskHealthStates.UNHEALTHY;
     const unknown = task.health === TaskHealthStates.UNKNOWN;
 
+    const failing = ["TASK_ERROR", "TASK_FAILED", "TASK_KILLING"].includes(
```
nLight: `TASK_KILLING` doesn't mean it is failing, does it? Let's create a separate shortcut for `killing`

nLight: Moreover on the line 351 we're using `failing` variable to assign `danger` css class, I'm not sure if it's correct

plugins/services/src/js/containers/tasks/TaskTable.js
```diff
@@ -322,6 +322,7 @@ class TaskTable extends React.Component {
 
     const dangerState = TaskStates[state].stateTypes.includes("failure");
     const activeState = TaskStates[state].stateTypes.includes("active");
+    const transitional = ["TASK_KILLING", "TASK_STARTING", "TASK_STAGING"].includes(state);
```
nLight: Please run prettier, that line is a bit longer.

/jenkins/workspace/Frontend/public-dcos-ui-pulls/plugins/services/src/js/containers/tasks/TaskTable.js
  325:27  error  Follow `prettier` formatting (expected '\n' but found '"')  prettier/prettier

��� 1 problem (1 error, 0 warnings)

weblancaster: I forgot to push the changes.

PR #2307, opened 2017-07-10
fix(DCOS_OSS-1272): retain port value if requirePorts is set to false

bstavroulakis:
I removed the this.appState.portsAutoAssign check.

To reproduce:

1) Copy the following JSON in a new Service JSON Form.
2) Run the service.
3) Edit the service, the port should remain 10001 and not reset to 0 even though requirePorts is false.

```
{
    "id": "/chronos",
    "backoffFactor": 1.15,
    "backoffSeconds": 1,
    "container": {
      "type": "DOCKER",
      "docker": {
        "forcePullImage": false,
        "image": "nginx"
      }
    },
    "cpus": 1,
    "disk": 0,
    "instances": 1,
    "mem": 1024,
    "gpus": 0,
    "networks": [
      {
        "mode": "host"
      }
    ],
    "portDefinitions": [
      {
        "port": 10001,
        "protocol": "tcp"
      }
    ],
    "requirePorts": false
}
```

Closes DCOS_OSS-1272

**Checklist**


nLight:
Hey Bill, I'm sorry that slipped my mind, I think we should also change the review screen. So that when there's a port value along with `requirePorts: false` we should say `Host Port = value`

screen shot 2017-07-11 at 11 41 32

So we either want to stick to 0, or say that it's going to be assigned automatically in this particular case.

bstavroulakis:
@nLight The review screen has been addressed as well.

nLight:
Hey @gkleiman @timcharper please tell us what is correct in case of having `port` set to something and `requirePorts` set to `false`. What do we show as a Host Port? Is it `0` or something like `Auto Assigned` or `N/A`? https://github.com/dcos/dcos-ui/pull/2307#issuecomment-314393919

gkleiman:
@nLight: in my opinion something like `Auto Assigned` would be more user-friendly, but I'll let Tim chime in =).

bstavroulakis:
@nLight @gkleiman @timcharper: When the `requirePorts` is set to `false` in the configuration now instead of 0 we show the "Auto Assigned" message.

plugins/services/src/js/service-configuration/ServiceNetworkingConfigSection.js
```diff
@@ -114,7 +114,9 @@ class ServiceNetworkingConfigSection extends ServiceConfigBaseSectionDisplay {
                 prop: keys.port,
                 className: getColumnClassNameFn(),
                 render(prop, row) {
-                  return getDisplayValue(row[prop]);
+                  return appDefinition.requirePorts
+                    ? getDisplayValue(row[prop])
+                    : <em>Auto Assigned</em>;
```
weblancaster: @bstavroulakis can you confirm `em` (emphasis) was approved by design please? other than that works perfectly.

bstavroulakis: @ashenden I've sent you a message with an image snapshot of this.

screen shot 2017-07-17 at 10 17 02 am

bstavroulakis:
@nLight @weblancaster All changes done, tested and approved by the design team.

PR #2310, opened 2017-07-11
fix(appValidators): Remove Healthchecks validators

Poltergeist:
Remove healthcheck validators, since we do not have any warning type and currently the user is not
able to create / edit an app with a marathon healthcheck

**Checklist**


nLight:
@Poltergeist rebase this please

nLight:
Is there  a jira for that?

PR #2311, opened 2017-07-12
fix(TasksView): improve is deploying to accommodate Node tasks

weblancaster:
This PR fixes the TypeError when selecting a task (via checkbox) on the node tasks page.

**DCOS_OSS-1350**

**Details**:
Remove dependency on the use of service struct for Services and Nodes instead start using require`DCOSStore.serviceTree` in `TasksView`.

**How to test:**
When selecting a task on Service/Node page if it's not deploying you should be able to see the `restart` and `stop` buttons, if one of the services selected is deploying `restart` should be disabled. Should not throw a `TypeError`.

**Checklist**


weblancaster:
alfred retest this please

PR #2312, opened 2017-07-12
DCOS-15921 (1 of 4): Packages needed for design system component example

tnguyen0:
This PR adds the packages and library files that are needed for the design system code example component.

See branch `tommy/design-system/component-example-ace-editor` for the complete implementation.
In the UI: `Design System` --> `Components` --> `Checkboxes` --> `Code Tab`

**Checklist**


package.json
```diff
@@ -39,6 +39,7 @@
     "dcos-dygraphs": "1.1.0-beta.3",
```
weblancaster: Hey @tnguyen0 did you follow the docs when adding this packages? https://github.com/dcos/dcos-ui/blob/master/CONTRIBUTING.md#adding-npm-package-dependencies

Just want to make sure because I had issues the first time adding a new package to the.

bstavroulakis:
What we did with @MatApple was to rebase the feature/design-system with master because it was outdated. This created the conflict with the package.json
After that we fixed any issues with shrinwrap, ran the command and included the new npm-shrinkwrap.json file as well.
The conflict remains, that we can fix with an additional rebase.

MatApple:
@bstavroulakis @tnguyen0 we somehow got into a funk with the rebasing - cleaned up this branch and the feature branch and think we're good to go!

MatApple:
@tnguyen0 can you please verify this works as expected. We had to downgrade the codemirror dependency from latest to latest - 1 to play nicely with our version of React.

tnguyen0:
Thanks @MatApple @bstavroulakis! Did a quick test and everything works fine. Just looked into AceEditor and I can definitely get it working soon.

tnguyen0:
@bstavroulakis I was able to get it working with ace editor, however it is not so easy to get custom scrollbars for the editor whereas codemirror provides it to you.

MatApple:
Great @tnguyen0 - let's use Ace.

plugins/design-system/styles/vendor/simplescrollbars.css
```diff
@@ -0,0 +1,66 @@
...
+.CodeMirror-simplescroll-vertical {
+  right: 0; top: 0;
+  width: 8px;
+}
+.CodeMirror-simplescroll-vertical div {
+  right: 0;
+  width: 100%;
+}
+
+
```
Poltergeist: Please use one line in between css selectors.

plugins/design-system/styles/vendor/simplescrollbars.css
```diff
@@ -0,0 +1,66 @@
...
+}
+.CodeMirror-simplescroll-horizontal div {
+  bottom: 0;
+  height: 100%;
+}
+
+.CodeMirror-simplescroll-vertical {
+  right: 0; top: 0;
+  width: 8px;
+}
```
Poltergeist: Please use one line in between css selectors.

plugins/design-system/styles/vendor/simplescrollbars.css
```diff
@@ -0,0 +1,66 @@
...
+
+.CodeMirror-overlayscroll-horizontal, .CodeMirror-overlayscroll-vertical {
+  position: absolute;
+  z-index: 6;
+}
+
+.CodeMirror-overlayscroll-horizontal {
+  bottom: 0; left: 0;
+  height: 8px;
+}
```
Poltergeist: Please use one line in between css selectors.

plugins/design-system/styles/vendor/simplescrollbars.css
```diff
@@ -0,0 +1,66 @@
...
+.CodeMirror-overlayscroll-horizontal div {
+  bottom: 0;
+  height: 100%;
+}
+
+.CodeMirror-overlayscroll-vertical {
+  right: 0; top: 0;
+  width: 8px;
+}
```
Poltergeist: Please use one line in between css selectors.

tnguyen0:
@weblancaster @bstavroulakis: can you guys take another look?

npm-shrinkwrap.json
```diff
@@ -237,7 +237,591 @@
         "chokidar": {
           "version": "1.6.1",
           "from": "chokidar@>=1.6.1 <2.0.0",
-          "resolved": "https://registry.npmjs.org/chokidar/-/chokidar-1.6.1.tgz"
+          "resolved": "https://registry.npmjs.org/chokidar/-/chokidar-1.6.1.tgz",
+          "dependencies": {
+            "fsevents": {
```
weblancaster: I believe this is not supposed to be here the script `shrinkwrap` should remove the `fsevents`. Please take a look at gulpfile `fixShrinkwrap`

tnguyen0: @weblancaster you're right. I did it again and it removed it.

package.json
```diff
@@ -100,6 +100,7 @@
     "jest-cli": "16.0.2",
     "jest-webpack-alias": "3.3.0",
     "jison-loader": "1.0.0",
+    "js-beautify": "1.6.14",
```
Poltergeist: Hey @tnguyen0 thanks for the changes regarding codemirror, but I am still not a fan of `js-beautify`. You are just using this for HTML formation. I would like us to omit a package if we only need a simple function from it, because one package more means one package more we need to care for and update. Even if the update is not affecting (read improving) the way we use it. 

I kind of went forward and created a short example for the indentation of HTML:
https://jsfiddle.net/onq7ooyj/6/

Also the Javascript code could easily be refactored I just grabbed it from stackoverflow from this question and added the feature that the first level will not have a newline character before it:
https://stackoverflow.com/questions/26360414/javascript-how-to-correct-indentation-in-html-string

Poltergeist: Hey @tnguyen0 thanks for the changes regarding codemirror, but I am still not a fan of `js-beautify`. You are just using this for HTML formation. I would like us to omit a package if we only need a simple function from it, because one package more means one package more we need to care for and update. Even if the update is not affecting (read improving) the way we use it. 

I kind of went forward and created a short example for the indentation of HTML:
https://jsfiddle.net/onq7ooyj/6/

Also the Javascript code could easily be refactored I just grabbed it from stackoverflow from this question and added the feature that the first level will not have a newline character before it:
https://stackoverflow.com/questions/26360414/javascript-how-to-correct-indentation-in-html-string

tnguyen0: I agree with what you're saying @Poltergeist. Just from playing around with the jsfiddle, I feel it is sufficient for our needs. Thanks for pointing it out! 👍

tnguyen0:
Changes made. Can you please have another look @Poltergeist.

Poltergeist:
This looks good for me, thanks for the effort.

PR #2318, opened 2017-07-13
Style Updates 

MatApple:
This is a feature branch for @ashenden's style adjustments.

**Make sure to squash all commits that don't follow our commit message guide into 1 before we merge to master**.

**Checklist**


weblancaster:
@MatApple is this still a WIP?

MatApple:
@ashenden do you have more style updates you want to add? If not, we'll get this merged. cc/ @weblancaster

ashenden:
I do, but I don't know when I'll get to them. So, if you'd like to merge this, please go ahead. I can always start a new feature branch.
  

  
  

  
  
>

MatApple:
Sweet @ashenden - altered all your commit messages and we're good to get this merged once tests pass.

ashenden:
@tnguyen0 please use the new font change and adjust accordingly.  Not sure why that Deployments action would spill to a new line unless something is defining a fixed width -- which seems like bad practice anyways.

weblancaster:
@MatApple @tnguyen0 the loader misalignment is fixed.

The issue is that we have the class `horizontal-center` being applied to the `Loader` component and removing fixes the issue but breaks for everything else using the class, so after trying to fix using minimum css possible to not affect the elements using the same component I decided to just introduce a prop where we can remove the class.

_ps let me know if you rather adding specificity to css where targets the deploying and loader only_

ashenden:
@weblancaster I'd go with your solution for now. We can go back behind and adjust these as a global component later.

src/js/components/Loader.js
```diff
@@ -52,7 +52,9 @@ class Loader extends React.Component {
   render() {
     const { className, innerClassName, flip, size, type } = this.props;
     const config = typeMap[type] || typeMap.ballScale;
-    const classes = classNames("loader horizontal-center", className);
+    const classes = this.props.supressHorizontalCenter
```
MatApple: @weblancaster @tnguyen0 @bstavroulakis let's make sure we add a new prop to the propTypes, including a default. I think a BOOL would be sensible here.

MatApple: Also just a #nit language thing but let's add another "p" to suppress > `suppressHorizontalCenter`

src/js/components/Loader.js
```diff
@@ -52,7 +52,9 @@ class Loader extends React.Component {
   render() {
     const { className, innerClassName, flip, size, type } = this.props;
     const config = typeMap[type] || typeMap.ballScale;
-    const classes = classNames("loader horizontal-center", className);
+    const classes = this.props.suppressHorizontalCenter
+      ? classNames("loader", className)
+      : classNames("loader horizontal-center", className);
```
tnguyen0: Slight nit, but you could do something like:
```
    const classes = classNames("loader", className, {
      "horizontal-center": !this.props.suppressHorizontalCenter
    });
```

PR #2320, opened 2017-07-14
↩️  DCOS-17242: Adjust MarathonErrorUtil to handle missing detail errors

orlandohohmeier:
---
↩️  _This PR back-ports a fix to release/1.9 introduced with #2319_

---

Adjust the `parseErrors` util to translates messages with empty or missing detail errors.

DCOS-17242

### DIFF
--- a/plugins/services/src/js/utils/MarathonErrorUtil.js
+++ b/plugins/services/src/js/utils/MarathonErrorUtil.js
@@ -99,7 +99,7 @@ const MarathonErrorUtil = {
     }
 
     // `details` can be an array of errors
-    return error.details.reduce(function(memo, { errors, path }) {
+    return error.details.reduce(function(memo, { errors = [], path }) {
       // Convert marathon path components to a dot-separated string
       // and then split it into an array
       //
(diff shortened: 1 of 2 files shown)


orlandohohmeier:
run integration tests

nLight:
retest this please

nLight:
run unit tests

plugins/services/src/js/utils/MarathonErrorUtil.js
```diff
@@ -121,6 +121,15 @@ const MarathonErrorUtil = {
         });
       }
 
+      if (!Array.isArray(errors) || errors.length === 0) {
```
nLight: We also could populate errors and have the following reduce do the work, why do you think this is a better solution?

orlandohohmeier: It's by no means a better solutions... I've changed the implementation to reuse the reducer.

orlandohohmeier:
@ahoskins Please note that Marathon 1.5 (DC/OS 1.10) responds with the following error messages, which we simple parse and display. Though I have to admit that the two messages are confusing. 

```json
{
	"message": "Object is not valid",
	"details": [{
		"path": "/",
		"errors": ["must not be root"]
	}, {
		"path": "/id",
		"errors": ["Path must contain at least one path element"]
	}]
}
```

Please retest using Marathon 1.4 (DC/OS 1.9).

PR #2321, opened 2017-07-14
DCOS-17262: fix(NetworkValidatorUtil): add support for 0 vip port

Poltergeist:
Add support for 0 at the vip port in the network validator util.

Closes DCOS-17262

**Checklist**

### DIFF
--- a/src/js/utils/__tests__/NetworkValidatorUtil-test.js
+++ b/src/js/utils/__tests__/NetworkValidatorUtil-test.js
@@ -19,17 +19,17 @@ describe("NetworkValidatorUtil", function() {
     });
 
     it("should handle number like inputs", function() {
-      expect(NetworkValidatorUtil.isValidPort(0)).toBe(false);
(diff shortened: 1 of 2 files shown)


src/js/utils/__tests__/NetworkValidatorUtil-test.js
```diff
@@ -38,5 +38,9 @@ describe("NetworkValidatorUtil", function() {
       expect(NetworkValidatorUtil.isValidPort(65535)).toBe(true);
       expect(NetworkValidatorUtil.isValidPort(65536)).toBe(false);
     });
+
+    it("verifies that port 0 is valid", function() {
```
MatApple: Probably don't need this test since we already have it https://github.com/dcos/dcos-ui/pull/2321/files#diff-d085bc6b0af09b2f1baa5824f77f9a01R22 and https://github.com/dcos/dcos-ui/pull/2321/files#diff-d085bc6b0af09b2f1baa5824f77f9a01R32 but I won't argue with good coverage ;)

PR #2327, opened 2017-07-18
chore(*): make compression-webpack-plugin a dependency

nLight:
I discovered that our build depends on `compression-webpack-plugin` in a funny manner. And on the `@latest` version which is even funnier. So `compression-webpack-plugin@latest` hit version 1.0.0 and now it depends on `webpack` 2 or 3 Ooops 😊


orlandohohmeier:
LGTM! 🚢

PR #2328, opened 2017-07-18
Update npm 3.10 🙏🏻

nLight:
To address our npm shrikwrap in a less intrusive way we need to update npm to at least 3.10

**Deps**:
https://github.com/mesosphere/reactjs-components/pull/410

**Please Note**:
`npm run shrinkwrap` become `npm run fix-shrinkwrap` because you can't override core commands anymore.


orlandohohmeier:
@nLight @MatApple let's please discuss how to move forward with this one as this potential breaks our build.

bstavroulakis:
It looks like this update may need a time slot of its own, as it is not seamless and introduces 81 failures as I see with the integration tests.

PR #2330, opened 2017-07-18
DCOS-16705: fix(ServiceForm): allow unknown values for container images

Poltergeist:
This PR introduces the support for unknown values in container(s) images. They will now be saved
and not wiped with the next generation.

*How to test*

in the tests there is a good example which can be copied over. Apply it to the service form and try to wipe the custom / unknown value.

Closes DCOS-16705

**Checklist**

### DIFF
--- a/plugins/services/src/js/reducers/serviceForm/Container.js
+++ b/plugins/services/src/js/reducers/serviceForm/Container.js
@@ -87,6 +87,11 @@ const containerJSONReducer = combineReducers({
     );
 
     const joinedPath = path && path.join(".");
+
+    if (type === SET && joinedPath === "container.docker") {
+      this.internalState = Object.assign({}, this.internalState, value);
+    }
+
     if (type === SET && joinedPath === "container.type") {
       this.containerType = value;
     }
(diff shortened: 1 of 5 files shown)


weblancaster:
@Poltergeist while testing this PR I noticed that I wasn't able to see the custom config I saved originally.

I introduced `"some": "value"` inside the `container` and was able to save without any validation error but after running and then clicking to edit again the custom key/value wasn't there anymore, is this expected?

```
{
  "id": "/batman",
  "backoffFactor": 1.15,
  "backoffSeconds": 1,
  "cmd": "echo \"holla como estas\"",
  "container": {
    "docker": {
      "image": "nginx",
      "forcePullImage": false,
      "privileged": false,
      "parameters": []
    },
   "some": "value",
    "type": "DOCKER",
    "volumes": []
  },
  "cpus": 0.1,
  "disk": 0,
  "instances": 1,
  "mem": 128,
  "gpus": 0,
  "networks": [
    {
      "mode": "host"
    }
  ],
  "portDefinitions": [],
  "requirePorts": false,
  "upgradeStrategy": {
    "maximumOverCapacity": 1,
    "minimumHealthCapacity": 1
  },
  "unreachableStrategy": {
    "inactiveAfterSeconds": 300,
    "expungeAfterSeconds": 600
  },
  "healthChecks": [],
  "fetch": [],
  "constraints": []
}
```

Poltergeist:
Hi @weblancaster so the thing is that marathon only saves values which it knows. The purpose of this pull request is to enable people with a certain plugin for marathon to save their configuration. So the purpose of this PR is that the value stays in the JSON and gets sent to marathon.

orlandohohmeier:
Great Job! ⚒ 💯

PR #2333, opened 2017-07-21
DCOS-14469: fix container duplicates for multi-container-pod

bstavroulakis:
To reproduce

1) Create a multi-container-pod with the following settings

```
{
  "id": "/test-multi",
  "version": "2017-07-20T20:29:38.654Z",
  "containers": [
    {
      "name": "container-1",
      "resources": {
        "cpus": 0.1,
        "mem": 128,
        "disk": 0
      },
      "image": {
        "kind": "DOCKER",
        "id": "nginx"
      }
    },
    {
      "name": "container-2",
      "resources": {
        "cpus": 0.1,
        "mem": 128,
        "disk": 0
      },
      "image": {
        "kind": "DOCKER",
        "id": "python"
      }
    }
  ],
  "networks": [
    {
      "mode": "host"
    }
  ],
  "scaling": {
    "instances": 1,
    "kind": "fixed"
  },
  "scheduling": {
    "placement": {
      "constraints": []
    }
  },
  "executorResources": {
    "cpus": 0.1,
    "mem": 32,
    "disk": 10
  },
  "volumes": [],
  "fetch": []
}
```

2) Once you save, you can click on the services tab, click on the service details and check the containers of the instance.
It should show container-1 and container-2 twice.

With this update it should show container-1 and container-2 once.

The PodUtil was looking for unique containers to merge them. But some containers had a containerId of null and it thought that they were different. That is why I included a check for null.

Closes DCOS-14469

**Checklist**

### DIFF
--- a/plugins/services/src/js/utils/PodUtil.js
+++ b/plugins/services/src/js/utils/PodUtil.js
@@ -90,7 +90,7 @@ var PodUtil = {
       // Filter combined container list to remove potential duplicates
       const containerIds = new Map();
       combinedContainers = combinedContainers.filter(function({ containerId }) {
-        if (!containerIds.has(containerId)) {
+        if (containerId != null && !containerIds.has(containerId)) {
           containerIds.set(containerId);
 
           return true;


plugins/services/src/js/utils/PodUtil.js
```diff
@@ -90,7 +90,7 @@ var PodUtil = {
       // Filter combined container list to remove potential duplicates
       const containerIds = new Map();
       combinedContainers = combinedContainers.filter(function({ containerId }) {
-        if (!containerIds.has(containerId)) {
+        if (!containerIds.has(containerId) && containerId != null) {
```
tnguyen0: What do you think about the null check being first? We could have one less check.

bstavroulakis:
@tnguyen0 @MatApple @nLight Good catch! 👍  Fixed.

ahoskins:
Why do these containers have a `containerId` of null in the first place?

PR #2334, opened 2017-07-21
DCOS-17408: fix(serviceForm): do not set container if it is not necessary

Poltergeist:
volumes was returning an empty array which resulted in an error of undefined type for non docker
apps

Closes DCOS-17408

**Checklist**

### DIFF
--- a/plugins/services/src/js/reducers/serviceForm/Container.js
+++ b/plugins/services/src/js/reducers/serviceForm/Container.js
@@ -188,6 +188,14 @@ module.exports = {
       return null;
     }
 
+    if (
+      newState.type == null &&
+      newState.docker == null &&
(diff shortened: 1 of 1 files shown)


plugins/services/src/js/reducers/serviceForm/Container.js
```diff
@@ -188,6 +188,14 @@ module.exports = {
       return null;
     }
 
+    if (
+      newState.type == null &&
+      newState.docker == null &&
+      ValidatorUtil.isEmpty(newState.volumes)
```
nLight: We decided to go with `isEmpty` just to make sure that we cover `volumes` being `undefined` and `null`

nLight:
run integration tests

PR #2337, opened 2017-07-21
DCOS-16400: remove SDK.get in favor of direct import

nLight:
Deps:
⚠️ Should go together with: https://github.com/mesosphere/dcos-ui-plugins-private/pull/617

This PR replaces SDK.get() with direct imports utilizing the #SRC webpack alias so that we aren't bound to a specific location.

This is an interim step for modularization. Direct imports will be replaced by a proper SKD/Loader mechanism.

Closes DCOS-16400


src/js/plugin-bridge/PluginSDK.js
```diff
@@ -297,14 +255,10 @@ const getSDK = function(pluginID, config) {
  * @param  {Object} config Plugin configuration
  * @param  {String} pluginType - One of 'internal' or 'external'.
```
tnguyen0: Should remove `pluginType` from the function documentation

tnguyen0: Should remove `pluginType` from the function documentation. Also, there are traces to `pluginType` that should be removed as well. There's a call to `bootstrapPlugin(pluginID, path, pluginsConfig[pluginID], pluginType);` in the `initialize` function.

Poltergeist: Remove this line because of what @tnguyen0 said.

Poltergeist:
I updated this PR as, @nLight is not able to update this pr in the next days.

PR #2338, opened 2017-07-25
fix(serviceForm): Remove docker object if no image is provided

Poltergeist:
Remove docker object if no image is provided and the container type is unequal to `"DOCKER"`

For testing create:
```
{
  "id": "/sleeper",
  "instances": 1,
  "portDefinitions": [],
  "container": {
    "type": "DOCKER",
    "volumes": [],
    "docker": {
      "image": "alpine"
    }
  },
  "cpus": 0.1,
  "mem": 128,
  "requirePorts": false,
  "networks": [],
  "healthChecks": [],
  "fetch": [],
  "constraints": [],
  "cmd": "sleep 1000;"
}
```

Wait till the app is deployed then go in remove the `alpine` image and set the container runtime to
MESOS. The Result is positive if in the JSON the docker object is not present anymore. And you
should be able to deploy the new version.

Closes DCOS-17284

**Checklist**

### DIFF
--- a/plugins/services/src/js/reducers/JSONMultiContainerParser.js
+++ b/plugins/services/src/js/reducers/JSONMultiContainerParser.js
@@ -1,6 +1,5 @@
 import { simpleParser } from "#SRC/js/utils/ParserUtil";
 
-import { JSONParser as container } from "./serviceForm/Container";
 import {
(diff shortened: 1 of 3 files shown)


ahoskins:
run integration tests

PR #2339, opened 2017-07-25
test(Dashboard): move app.json to inline, fixing failing jenkins system tests

ahoskins:
**Description**:
In the `test-dashboard.js` system test, we created services by using an `app.json` file. The tests are broken on [Jenkins](https://jenkins.mesosphere.com/service/jenkins/job/Frontend/job/dcos-ui-ci-master-ee/145/artifact/dcos-ui/results/results.html) because this file is not found. Since this file is small, move it inline to the cypress test file. This will fix the issue on Jenkins.

Closes DCOS-17383.

**Testing**:

Everyday I usually create an open cluster named "andrew-(random-id)-open". Feel free to use it for testing.

```
    -v `pwd`:/dcos-ui \
    mesosphere/dcos-ui dcos-system-test-driver --only "dashboard" \
```
**Checklist**


ahoskins:
run integration tests

ahoskins:
run integration tests

system-tests/_support/index.js
```diff
@@ -219,23 +219,27 @@ Cypress.addChildCommand("triggerHover", function(elements) {
...
- * @param {String} serviceName - The service name which will launch
- * @param {String} serviceJsonPath - The path to JSON file for service definition
+ * @param {Object} serviceDefinition - The service JSON definition file
  *
  */
-export function createService(serviceName, serviceJsonPath) {
+export function createService(serviceDefinition) {
   cy.exec(
-    `sed 's/${serviceName}/${Cypress.env("TEST_UUID")}\\/${serviceName}/g' ${serviceJsonPath} | dcos marathon app add`
+    `echo '${JSON.stringify(serviceDefinition)}' | dcos marathon app add`
```
MatApple: Wouldn't it only create the service in this group if the `id` inside `serviceDefinition` started with `TEST_UUID`? Feel like the description is misleading because it depends on the input. Is there a better way to do this? L231 is also relying on TEST_UUID being in the `id` field for it to work. Let's rethink this function.

system-tests/_support/index.js
```diff
@@ -245,15 +249,15 @@ export function createService(serviceName, serviceJsonPath) {
...
- * @param {String} serviceName - The service name which it will delete
+ * @param {String} serviceId - The service id which it will delete
  *
  */
-export function deleteService(serviceName) {
-  cy.exec(
-    `dcos marathon app remove /${Cypress.env("TEST_UUID")}/${serviceName}`
-  );
+export function deleteService(serviceId) {
+  cy.exec(`dcos marathon app remove ${serviceId}`);
```
MatApple: Same here. We are relying on `serviceId` being prefixed with `TEST_UUID` but it's possible that it wouldn't be and we should avoid logic like this.

system-tests/_support/index.js
ahoskins: Yeah, I agree. These are very general helpers that will likely re-used by future tests. I'll change this to be more flexible.

system-tests/_support/index.js
```diff
@@ -219,45 +219,56 @@ Cypress.addChildCommand("triggerHover", function(elements) {
...
+ * Launches a new service using the dcos CLI. Service must be created
+ * in the TEST_UUID group.
  *
  * @param {Object} serviceDefinition - The service JSON definition file
  *
  */
 export function createService(serviceDefinition) {
+  const idParts = serviceDefinition.id.split("/");
+  if (idParts.length !== 3) {
+    throw new Error("Must include leading slash in service id");
```
MatApple: if `id = /$TEST_UUID/group/app` the error will be a little odd. Perhaps a better check for throwing this error is `id[0] !== '/'`.

We could then insert another check on L236:
`
if (idParts.length !== 3) {
  throw new Error("Must deploy service directly in TEST_UUID group");
}
`

ahoskins: Ah yes, good catch! 👍

tnguyen0: @ahoskins (Nit): What do you think about doing `!id.startsWith("/")` for better readability?

ahoskins: Sure, I'll change it to that.

system-tests/_support/index.js
tnguyen0: One option is to extract the `cy.exec` calls into their own functions.

system-tests/_support/index.js
```diff
@@ -219,23 +219,46 @@ Cypress.addChildCommand("triggerHover", function(elements) {
...
  *
- * @param {String} serviceName - The service name which will launch
- * @param {String} serviceJsonPath - The path to JSON file for service definition
+ * @param {String} id - serviceId to validate
  *
  */
-export function createService(serviceName, serviceJsonPath) {
+function validateServiceId(id) {
+  if (!id.startsWith("/")) {
+    throw new Error("Must include leading slash in service id");
```
Poltergeist: 👍

system-tests/_support/index.js
```diff
@@ -219,23 +219,46 @@ Cypress.addChildCommand("triggerHover", function(elements) {
...
+  }
+  const idParts = id.split("/");
+  if (idParts[1] !== Cypress.env("TEST_UUID")) {
+    throw new Error("Service must be in the TEST_UUID group");
```
Poltergeist: 👍

system-tests/_support/index.js
```diff
@@ -219,23 +219,46 @@ Cypress.addChildCommand("triggerHover", function(elements) {
...
+  }
+  if (idParts.length !== 3) {
+    throw new Error("Must deploy service directly in TEST_UUID group");
```
Poltergeist: 👍

ahoskins:
run integration tests

PR #2340, opened 2017-07-25
fix(ContainerServiceFormAdvancedSection): Fix fraction in GPU

weblancaster:
This PR fixes the issue when creating/editing a service where the GPU input would not "reset" when adding fraction,
Now GPU accepts fraction but the validator won't allow to save it, this validation was also added to Disk since we only accept integers.

closes: DCOS-14850

**How to test**
Create a new service or edit a existent one
Go to the GPU field (`MESOS RUNTIME` need to be checked)
Change the value to faction
_before_ would 

**Checklist**


weblancaster:
🎉

PR #2341, opened 2017-07-26
DCOS-17301: not all tasks were shown under the service

bstavroulakis:
If you create a service inside a folder, it won't show all of the tasks under the service.

In the MesosStateStore we had the following check

```
const serviceName = service.getId().split("/").pop();

if(framework.name === serviceName) { ... }

```
with

```
service: { id: "/test/elastic" }
framework { name: "test/elastic" }
```

The problem is that if we have a service in a folder the service id is "/test/elastic" and with the split pop() we lose the full id.

Also the framework's name is "test/elastic". So it would compare framework.name = "test/elastic" with serviceName = "elastic" eventually.

> So you would see this on one DCOS dashboard
image
> and this on the Mesos dashboard
image

I created 2 new methods one **getFullId()** to return **"test/elastic"** from the service
 and one **getServiceName()** to return **"test/elastic"** from the framework.

> Note concerning the naming for "getFullId() and getServiceName()"
> 
> I don't like the following method in the Application.js. Although it says getName() it returns part of the Id.
> But we can't change it now easily as a defect since it is used in numerous areas.

```
  getName() {
    return this.getId().split("/").pop();
  }
```

Closes DCOS-17301

**Checklist**

### DIFF
--- a/src/js/stores/MesosStateStore.js
+++ b/src/js/stores/MesosStateStore.js
@@ -283,14 +283,18 @@ class MesosStateStore extends GetSetBaseStore {
     }
 
     const schedulerTasks = this.getSchedulerTasks();
+    const serviceIsFramework = service && service instanceof Framework;
+    const serviceFrameworkName = serviceIsFramework
+      ? service.getFrameworkName()
+      : serviceName;
 
     // Combine framework (if matching framework was found) and filtered
     // Marathon tasks. This will give you a list of framework tasks including
     // the scheduler tasks or a list of Marathon application tasks.
     return frameworks.reduce(function(serviceTasks, framework) {
       const { tasks = [], completed_tasks = {}, name } = framework;
-      // Include tasks from framework match, if service is a Framework
-      if (service instanceof Framework && name === serviceName) {
+
+      if (serviceIsFramework && serviceFrameworkName === name) {
         return serviceTasks
           .concat(tasks, completed_tasks)
           .map(task => assignSchedulerTaskField(task, schedulerTasks));


bstavroulakis:
run integration tests

plugins/services/src/js/structs/Framework.js
```diff
@@ -20,6 +20,32 @@ module.exports = class Framework extends Application {
     return this.get("slave_ids");
   }
 
+  getFullId() {
```
MatApple: `#getFullId` is a little misleading here because it's actually the full Id but without the leading `/`. I think we should change this method to `#getId` so it overrides the super method and instead of pruning the leading `/` we should make sure it is always present.

plugins/services/src/js/structs/Framework.js
```diff
@@ -20,6 +20,32 @@ module.exports = class Framework extends Application {
...
+    const frameworkId = this.getId();
+    if (!frameworkId) {
+      return frameworkId;
+    }
+
+    return frameworkId.replace(/^\//, "");
+  }
+
+  getServiceName() {
```
MatApple: Let's rename this to `getFrameworkName` since this concept is very specific to a `Framework`. That will help alleviate confusion between `getServiceName` and `getName`.

We should probably also ensure this has a leading `/` as well.

plugins/services/src/js/structs/Framework.js
ahoskins: Very related to the create [createService()](https://github.com/dcos/dcos-ui/pull/2339/files#diff-361072f18fbf1e35155c80dd51343244R230) helper in my system test PR. In `createService()` I made it enforce the leading slash (even though the backend would work without slash also, I think).

**SO**, I agree with Mat, would be nice to be consistent with enforcing that slash.

src/js/utils/StringUtil.js
```diff
@@ -215,6 +215,16 @@ const StringUtil = {
     }, []);
 
     return this.idToTitle(splitID, splitBy, replace, removeConsecutive);
+  },
+
+  lrtrim(str, trimStr = " ") {
```
MatApple: Native JS `.trim()` will trim all whitespace characters (space, tab, no-break space, etc.) and all the line terminator characters (LF, CR, etc.). We could simulate this with `\s\uFEFF\xA0` - what do you think about changing to `trimStr = "\s\uFEFF\xA0"`?

plugins/services/src/js/structs/Framework.js
```diff
@@ -20,6 +20,32 @@ module.exports = class Framework extends Application {
...
+    const frameworkName = this.get("name");
+    if (!frameworkName) {
+      return frameworkName;
+    }
+
+    return frameworkName.replace(/^\//, "");
+  }
+
+  getTasks() {
```
bstavroulakis: I removed that one completely

plugins/services/src/js/structs/Framework.js
bstavroulakis: I removed that one completely

plugins/services/src/js/structs/Framework.js
bstavroulakis: I removed that one completely. Will go with another util method named lrtrim where we can trim a string with another one. In this case I compare the two strings and trim the "/" string.

plugins/services/src/js/structs/__tests__/Framework-test.js
```diff
@@ -85,6 +85,33 @@ describe("Framework", function() {
 
       expect(service.getFrameworkName()).toEqual(undefined);
     });
+
+    it("returns correct full name for service comparison with slash", function() {
```
bstavroulakis: You are correct, it slipped through. I actually removed those test but included some tests for the getTasks and getCompletedTasks

src/js/utils/StringUtil.js
```diff
@@ -215,6 +215,16 @@ const StringUtil = {
...
+    if (typeof str === "string") {
+      return str
```
MatApple: Some comments -
1. Think we should use `+` instead of `*` since `+` will only match if there is at least 1 char. `*` will always match.
2. We could combine into a single RegExp: 
`.replace(new RegExp("^[" + trimStr +" ]+|[" + trimStr + "]+$", "g"), "");`

src/js/utils/StringUtil.js
MatApple: Let's also add documentation :)

bstavroulakis: Or we could also run .trim() at the end of the statement and leave the trimStr with no default value?

bstavroulakis: Added :)

MatApple: Wouldn't that require a check for `null`?

src/js/utils/StringUtil.js
MatApple: Using a template literal instead of the string concatenation would be clean.

bstavroulakis: Yeap, bad es5 habits die hard

src/js/utils/StringUtil.js
bstavroulakis: I improved this. So now we include the default trim values as well as the one we would like to lrtrim. Also, included more test cases.

src/js/utils/StringUtil.js
```diff
@@ -215,6 +215,29 @@ const StringUtil = {
...
+   * @return {String}        - The trimmed string. If a non-string type is
+   *                         - entered it will just return the input.
+   */
+  lrtrim(str, trimStr) {
+    if (typeof str === "string") {
+      const defaultTrimCharacters = "\\s|\\uFEFF|\\xA0";
+
+      return str.replace(
+        new RegExp(
+          `^(${trimStr}|${defaultTrimCharacters})+|(${trimStr}|${defaultTrimCharacters})+$`,
```
MatApple: @bstavroulakis thanks for making some changes! However, I strongly believe we shouldn't trim both all white space AND a user defined char set. I think it is harder to reason. A sensible default that can be overridden is how I understand most `trim` utilities work.

weblancaster: Can you add a description too please? thanks

src/js/utils/StringUtil.js
```diff
@@ -215,6 +215,29 @@ const StringUtil = {
...
+
+  /**
+   * @param {String} str     - A string that we would like to trim.
+   * @param {String} trimStr - The string that we would like to trim from
+   *                         - the beginning and the end.
+   * @return {String}        - The trimmed string. If a non-string type is
+   *                         - entered it will just return the input.
+   */
+  lrtrim(str, trimStr) {
+    if (typeof str === "string") {
```
MatApple: Although it's not painfully obvious here because we only have a single `if` but we should avoid nesting the logic in such a condition because the nesting can get deep and messy if more conditions are required. We should instead account for all the edge cases first and return early. So in this case we could just have

```
if (typeof str !== "string") {
  return str;
}
```

src/js/utils/StringUtil.js
```diff
@@ -215,6 +215,29 @@ const StringUtil = {
...
+  },
+
+  /**
+   * @param {String} str     - A string that we would like to trim.
+   * @param {String} trimStr - The string that we would like to trim from
+   *                         - the beginning and the end.
+   * @return {String}        - The trimmed string. If a non-string type is
+   *                         - entered it will just return the input.
+   */
+  lrtrim(str, trimStr) {
```
weblancaster: I would rename this method to something else than `lrtrim`

plugins/services/src/js/structs/Framework.js
```diff
@@ -20,6 +20,14 @@ module.exports = class Framework extends Application {
     return this.get("slave_ids");
   }
 
+  getTasks() {
```
weblancaster: I believe `getkey` is a better option here just because if we think let's say a Queue, or Stack (even though we are not using them) they contain a `find` where we pass a value and returns the value if exists.

I propose `getKey("tasks")` and get rid of `getTasks`, `getCompletedTasks` and if there is not a lot of code using `getNodeIDs` I would also refactor that one to start using `getKey` instead

src/js/utils/StringUtil.js
weblancaster: Agree. I also have the practice to put the logic outside whenever possible.

plugins/services/src/js/structs/Framework.js
MatApple: These structs by design are not great in general. If we had a `getKey` method, we would likely either not need the struct at all since we could just access the key on the object directly, or we'd have a long switch statement checking the key for lookup. The reason why we have `getXXXX` is to provide a custom getter. Of course, now we can use getters much more elegantly but the structs are legacy. For consistency and discoverability of available methods on the object, let's keep what you have here.

weblancaster: @MatApple in that case we are only contributing for more legacy because the consistency here is only creating more of what we don't want instead moving forward as we go because it's very hard for us to change all together.

If we can, using the object.notation would still be consistent only not to structs and even easier to discover since the stack and path is direct.

weblancaster: But at the same time I'm indifferent if we need to keep following the legacy for stronger reasons.

src/js/stores/MesosStateStore.js
```diff
@@ -287,19 +288,23 @@ class MesosStateStore extends GetSetBaseStore {
...
+      const framework = new Framework(frameworkProperties);
       // Include tasks from framework match, if service is a Framework
-      if (service instanceof Framework && name === serviceName) {
+      if (
+        service instanceof Framework &&
+        lrtrim(framework.name, "/") === lrtrim(service.getId(), "/")
+      ) {
         return serviceTasks
-          .concat(tasks, completed_tasks)
+          .concat(framework.getTasks(), framework.getCompletedTasks())
```
weblancaster: @MatApple As a example, what is the implications when accessing direct `framework ["tasks"]` instead of using a getter here?

src/js/utils/StringUtil.js
bstavroulakis: Renamed to stripLeftRight and included a description

src/js/utils/StringUtil.js
bstavroulakis: @MatApple I agree. Now if the trim is not defined it will default to the default values.

src/js/utils/StringUtil.js
bstavroulakis: @weblancaster @MatApple Yes, I agree, it is fixed now

plugins/services/src/js/structs/Framework.js
MatApple: I hear you and you have a good argument. However, consistency should trump most arguments, especially when the tradeoffs are low. Our codebase has become rather inconsistent and hence, complex and that adds significant overhead to onboarding and reviewing. I agree we should change the legacy, but let's plan that change effectively.

weblancaster: Sounds good to me then.

orlandohohmeier: @weblancaster @MatApple  Introducing an abstract getter like `getKey("tasks")` provides no guarantees and potentially introduces a high number of bugs thus should be avoided as discussed in on of our team syncs (reflection). 

I agree that we need to revisit our structs, but not because of the new (much more elegant) getter syntax or the seemingly redundant code.  
A struct is a composite type, which is used to model the relationship between types (e.g., the relationship between tasks and services as well as the relationship between service IDs and services). It provides the getters and setters to retrieve and assign values and should safeguard data access to prevent accidental mutations. We also use subtyping (`extend`) to describe that a struct is of the type of its parent struct. This allows us to add detail to a general data type and create a more specific data type (e.g. `Service` and `Framework`). These structs allow us to render all the different kinds of service in one list (service overview) as they share the same interface. 
Structs shouldn't transform data, considering the principles described,  which is the actual reason why we need to revisit our structs.

src/js/stores/MesosStateStore.js
```diff
@@ -287,19 +288,24 @@ class MesosStateStore extends GetSetBaseStore {
     // Combine framework (if matching framework was found) and filtered
     // Marathon tasks. This will give you a list of framework tasks including
     // the scheduler tasks or a list of Marathon application tasks.
-    return frameworks.reduce(function(serviceTasks, framework) {
-      const { tasks = [], completed_tasks = {}, name } = framework;
+    return frameworks.reduce(function(serviceTasks, frameworkProperties) {
+      const framework = new Framework(frameworkProperties);
       // Include tasks from framework match, if service is a Framework
-      if (service instanceof Framework && name === serviceName) {
+      if (
```
orlandohohmeier: I'm super uncomfortable with all the name/id manipulation and pretty sure it should be enough to simple change `serviceName` (e6350a94a53261628755fad599c04c6e4c146b09) to  `service.getFrameworkName()` to address DCOS-17301.

orlandohohmeier:
@bstavroulakis I'm super sorry, that I haven found the time to review your PR earlier! I'm uncomfortable with the current solution as expressed in https://github.com/dcos/dcos-ui/pull/2341#discussion_r130352151 and would love if you could try fixing this one without adding yet another util to modify strings. @nLight just recently introduced new getters to retrieve the framework and package name which should be enough to fix DCOS-17301.

bstavroulakis:
@orlandohohmeier @MatApple I updated the commit with the minimal changes for this to work. framework.getFrameworkName() was returning null, so I just get the name and make sure that there is a "/" in front of it.

orlandohohmeier:
@bstavroulakis please use `service.getFrameworkName()` and only fallback to `service.getName()` as there is a chance that the framework name doesn't match the service name/ID. 
An instance of `Framework` should always have a framework name as the respective factory uses `ServiceValidatorUtil.isFrameworkResponse(response)` to determine if a response object is a framework, which uses existence of framework name (`DCOS_PACKAGE_FRAMEWORK_NAME`) to flag frameworks.

bstavroulakis:
@orlandohohmeier Ah, yes I included the service.getFrameworkName() in the statement and it actually worked with services in folders and single services as well. If you could please check it again. Thanks

src/js/stores/MesosStateStore.js
```diff
@@ -289,8 +289,11 @@ class MesosStateStore extends GetSetBaseStore {
     // the scheduler tasks or a list of Marathon application tasks.
     return frameworks.reduce(function(serviceTasks, framework) {
       const { tasks = [], completed_tasks = {}, name } = framework;
-      // Include tasks from framework match, if service is a Framework
-      if (service instanceof Framework && name === serviceName) {
+
+      if (
```
orlandohohmeier: Not blocking, but I'd actually get the framework name only once to reduce the number of call. We might also be able to do the `Framework` type check only once.

bstavroulakis: You are correct, applied those changes as well.

bstavroulakis:
@orlandohohmeier @MatApple @weblancaster Applied the check out of the loop and changed the commit description.

orlandohohmeier:
@bstavroulakis thanks for relentlessly improving the fix. Great Job! 💯

PR #2342, opened 2017-07-27
DCOS-16225: change the update cli dialog message.

bstavroulakis:
Change the Update Dialog message to the following

"To Update a framework's configuration, please update "options.json" <make this a link to documentation i the service docs?>.  Run the following CLI commands:"

Before
image

After
image

Closes DCOS-16225

**Checklist**

### DIFF
--- a/plugins/services/src/js/components/modals/ServiceActionDisabledModal.js
+++ b/plugins/services/src/js/components/modals/ServiceActionDisabledModal.js
@@ -235,9 +235,9 @@ class ServiceActionDisabledModal extends React.Component {
             )}
             target="_blank"
           >
-            {intl.formatMessage({ id: "DOCS.MORE_INFORMATION" })}
+            'options.json'
           </a>
-          {" "}
+          .{" "}
           {intl.formatMessage({
             id: "SERVICE_ACTIONS.SDK_SERVICE_UPDATE_PART_2"
           })}
--- a/plugins/services/src/js/components/modals/ServiceActionDisabledModal.js
+++ b/plugins/services/src/js/components/modals/ServiceActionDisabledModal.js
@@ -49,7 +49,7 @@ class ServiceActionDisabledModal extends React.Component {
 
     // For everything SDK, not other packages
     if (isSDKService(service)) {
-      return `dcos ${packageName} --name=${serviceID} update start --options=${packageName}-options.json`;
+      return `dcos ${packageName} --name=${serviceID} update start --options=options.json`;
     }
 
     return `dcos marathon app update ${serviceID} options.json`;
@@ -235,7 +235,7 @@ class ServiceActionDisabledModal extends React.Component {
             )}
             target="_blank"
           >
-            'options.json'
+            options.json
           </a>
           .{" "}
           {intl.formatMessage({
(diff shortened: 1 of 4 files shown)


ahoskins:
Very related to: https://jira.mesosphere.com/browse/DCOS_OSS-1320

I was told the above issue a few weeks ago, but it never got prioritized or investigated further. Not sure if still a problem. We should address it though as part of this!

bstavroulakis:
@ahoskins @Poltergeist Made a small supplementary commit to fix related issue DCOS_OSS-1320 as well.

ahoskins:
Also related to DCOS-16528, which is the same issue as DCOS_OSS-1320. Both of these are mentioned in Jira as related issues.

image

ahoskins:
The copy seems like it could be improved:
_"To Update a framework's configuration, please update 'options.json'. Run the following CLI commands:"_

1) why is "update" capitalized?
2) "CLI commands", but it seems that only one command is suggested. This is a bit confusing why this is plural.

plugins/services/src/js/components/modals/ServiceActionDisabledModal.js
```diff
@@ -49,7 +49,7 @@ class ServiceActionDisabledModal extends React.Component {
 
     // For everything SDK, not other packages
     if (isSDKService(service)) {
-      return `dcos ${packageName} --name=${serviceID} update --options=${packageName}-options.json`;
+      return `dcos ${packageName} --name=${serviceID} update start --options=${packageName}-options.json`;
```
MatApple: Let's just make `--options=options.json` and remove the packageName from it. I think that's easier to follow given the copy above the command.

plugins/services/src/js/components/modals/ServiceActionDisabledModal.js
```diff
@@ -235,9 +235,9 @@ class ServiceActionDisabledModal extends React.Component {
             )}
             target="_blank"
           >
-            {intl.formatMessage({ id: "DOCS.MORE_INFORMATION" })}
+            'options.json'
```
MatApple: I think we can remove the quotes.

src/js/translations/en-US.json
```diff
@@ -68,8 +68,8 @@
   "SERVICE_ACTIONS.SDK_SERVICE_DELETE_PART_1": "You must delete",
   "SERVICE_ACTIONS.SDK_SERVICE_DELETE_PART_2": "via the DC/OS CLI. Copy and paste this command into the CLI. See the",
   "SERVICE_ACTIONS.SDK_SERVICE_DELETE_PART_3": "To clean up resources reserved by your service, run the",
-  "SERVICE_ACTIONS.SDK_SERVICE_UPDATE_PART_1": "Update your service's target configuration. Include an `options.json` file with your changes.",
-  "SERVICE_ACTIONS.SDK_SERVICE_UPDATE_PART_2": "on how to create an 'options.json' file.",
+  "SERVICE_ACTIONS.SDK_SERVICE_UPDATE_PART_1": "To Update a framework's configuration, please update",
```
MatApple: Please change to - `To update this service's configuration, please update` instead of calling out `framework`. Thats a term we don't use in the UI for end users.

bstavroulakis: Yeap, changed

plugins/services/src/js/components/modals/ServiceActionDisabledModal.js
bstavroulakis: Quotes removed

plugins/services/src/js/components/modals/ServiceActionDisabledModal.js
bstavroulakis: Changed it to --options=options.json and updated the tests as well

PR #2343, opened 2017-07-28
fix(PageHeaderActions): Remove restart action when is SDK

weblancaster:
Removes restart action button from services and service detail page when service is SDK.

A code refactoring was also done to follow `ServiceDetail` patterns for the dropdown items, meaning when testing this PR we need to do a sanity check for the other buttons besides `restart`

Closes DCOS-16564

**Checklist**

### DIFF
--- a/plugins/services/src/js/containers/service-detail/ServiceDetail.js
+++ b/plugins/services/src/js/containers/service-detail/ServiceDetail.js
@@ -1,4 +1,5 @@
 import { injectIntl } from "react-intl";
+import classNames from "classnames";
 import mixin from "reactjs-mixin";
 import React, { PropTypes } from "react";
 import { routerShape } from "react-router";
@@ -161,6 +162,7 @@ class ServiceDetail extends mixin(TabsMixin) {
   getActions() {
     const { service } = this.props;
     const instanceCount = service.getInstancesCount();
+    const isSDK = isSDKService(service);
 
     const actions = [];
 
@@ -185,7 +187,10 @@ class ServiceDetail extends mixin(TabsMixin) {
     if (instanceCount > 0) {
       actions.push({
         label: "Restart",
-        onItemSelect: this.onActionsItemSelection.bind(this, RESTART)
+        onItemSelect: this.onActionsItemSelection.bind(this, RESTART),
+        className: classNames({
+          hidden: isSDK
+        })
       });
     }
     if (!service.getLabels().MARATHON_SINGLE_INSTANCE_APP) {
--- a/plugins/services/src/js/containers/service-detail/ServiceDetail.js
+++ b/plugins/services/src/js/containers/service-detail/ServiceDetail.js
@@ -1,5 +1,4 @@
 import { injectIntl } from "react-intl";
-import classNames from "classnames";
 import mixin from "reactjs-mixin";
 import React, { PropTypes } from "react";
 import { routerShape } from "react-router";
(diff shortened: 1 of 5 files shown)


plugins/services/src/js/containers/service-detail/ServiceDetail.js
```diff
@@ -185,7 +187,10 @@ class ServiceDetail extends mixin(TabsMixin) {
     if (instanceCount > 0) {
```
MatApple: I think instead of hiding with a classname, we should not push the action if SDK. We can add that check to this `if` statement. We can then get rid of the new `prop` required on `PageHeaderActionsMenu`.

weblancaster: Ok cool.

plugins/services/src/js/containers/services/ServicesTable.js
```diff
@@ -239,6 +239,16 @@ class ServicesTable extends React.Component {
...
+    const restartObj = {
+      className: classNames({
+        hidden: isPod || isGroup || instancesCount === 0
+      }),
+      id: RESTART,
+      html: this.props.intl.formatMessage({
+        id: ServiceActionLabels[RESTART]
+      })
+    };
+    const isSDK = isSDKService(service) ? [] : [restartObj];
```
MatApple: I think we should take this opportunity to clean up this method. We should remove all the `className` properties and follow the same pattern of appending actions to an array as we do in `ServiceDetail`.

ahoskins: I actually was going to suggest the same, like:

```
    if (!isSDKService(service)) {
      dropdownItems.push({
        className: classNames({
          hidden: isPod || isGroup || instancesCount === 0
        }),
        id: RESTART,
        html: this.props.intl.formatMessage({
          id: ServiceActionLabels[RESTART]
        })
      });
    }
```

Poltergeist:
run integration tests

Poltergeist:

plugins/services/src/js/containers/services/ServicesTable.js
weblancaster: this was addressed in the last commit yesterday

weblancaster:
retest this please

weblancaster:
I have made updates to the PR to address the failing integration tests.

plugins/services/src/js/containers/services/ServicesTable.js
```diff
@@ -239,71 +239,78 @@ class ServicesTable extends React.Component {
...
         html: this.props.intl.formatMessage({ id: ServiceActionLabels.edit })
-      },
-      {
-        className: classNames({
-          hidden: (isGroup && instancesCount === 0) || isSingleInstanceApp
-        }),
+      });
+    }
+
+    if ((!isGroup && instancesCount > 0) || !isSingleInstanceApp) {
```
MatApple: Shouldn't this be `&&` instead of `||`?

plugins/services/src/js/containers/services/ServicesTable.js
```diff
@@ -239,71 +239,78 @@ class ServicesTable extends React.Component {
...
           id: ServiceActionLabels[SUSPEND]
         })
-      },
-      {
+      });
+    }
+
+    if (!isGroup && instancesCount === 0) {
+      actions.push({
         className: classNames({
```
MatApple: We don't need this property anymore.

plugins/services/src/js/containers/services/ServicesTable.js
weblancaster: This is the only case that we need to use `||` because will not add the `scale` since won't match `!isGroup && instancesCount > 0) && !isSingleInstanceApp` when actually need the `scale`.

bstavroulakis: Maybe it should be `(!isGroup || instancesCount > 0) && !isSingleInstanceApp`

[https://www.wolframalpha.com/input/?i=NOT((A+AND+B)+OR+C),+((NOT+A+OR+NOT+B)+AND+NOT+C)](https://www.wolframalpha.com/input/?i=NOT((A+AND+B)+OR+C),+((NOT+A+OR+NOT+B)+AND+NOT+C))

MatApple: 2 cases where scale should be added as an option:
- isGroup and instanceCount > 0
- !isSingleInstanceApp and !isGroup

PR #2344, opened 2017-07-31
fix(ServiceEditModal): show service edit instructions only in enterprise

ahoskins:
**Description**:
As described in [DCOS-17205](https://jira.mesosphere.com/browse/DCOS-17205), this modal content on Open DC/OS should be as below:
screen shot 2017-08-01 at 10 33 57 am

**Checklist**

### DIFF
--- a/plugins/services/src/js/components/modals/ServiceActionDisabledModal.js
+++ b/plugins/services/src/js/components/modals/ServiceActionDisabledModal.js
@@ -1,5 +1,6 @@
 import React, { PropTypes } from "react";
 import { injectIntl } from "react-intl";
+import { MountService } from "foundation-ui";
 import { Modal } from "reactjs-components";
 
 import ClipboardTrigger from "#SRC/js/components/ClipboardTrigger";
@@ -219,31 +220,23 @@ class ServiceActionDisabledModal extends React.Component {
   }
 
   getServiceEditMessage() {
-    const { intl } = this.props;
-    const command = this.getUpdateCommand();
+    const { intl, service } = this.props;
 
     return (
-      <div>
-        <p>
-          {intl.formatMessage({
-            id: "SERVICE_ACTIONS.SDK_SERVICE_UPDATE_PART_1"
-          })}
+      <MountService.Mount
+        type={"ServiceEditMessage:Modal"}
+        intl={intl}
+        service={service}
+      >
+        <div className="center">
+          Editing this service is only available on
           {" "}
-          <a
-            href={MetadataStore.buildDocsURI(
-              "/usage/managing-services/config-universe-service"
-            )}
-            target="_blank"
-          >
-            options.json
+          <a href="https://mesosphere.com/product/" target="_blank">
+            Mesosphere Enterprise DC/OS
           </a>
-          .{" "}
-          {intl.formatMessage({
-            id: "SERVICE_ACTIONS.SDK_SERVICE_UPDATE_PART_2"
-          })}
-        </p>
-        {this.getClipboardTrigger(command)}
-      </div>
+          .
+        </div>
+      </MountService.Mount>
     );
   }
 
--- a/tests/pages/services/ServiceTable-cy.js
+++ b/tests/pages/services/ServiceTable-cy.js
@@ -231,9 +231,9 @@ describe("Service Table", function() {
         .should("to.have.length", 1);
 
       cy
-        .get(".modal pre")
+        .get(".modal-body")
         .contains(
-          "dcos test --name=/services/sdk-sleep update start --options=options.json"
+          "Editing this service is only available on Mesosphere Enterprise DC/OS."
         );
 
       cy.get(".modal button").contains("Close").click();
--- a/tests/pages/services/ServiceTable-cy.js
+++ b/tests/pages/services/ServiceTable-cy.js
@@ -221,26 +221,6 @@ describe("Service Table", function() {
       cy.get(".modal").should("not.exist");
     });
 
-    it("opens the edit dialog", function() {
-      openDropdown("sdk-sleep");
-      clickDropdownAction("Edit");
-
-      cy
-        .get(".modal-header")
-        .contains("Edit Service")
-        .should("to.have.length", 1);
-
-      cy
-        .get(".modal-body")
-        .contains(
-          "Editing this service is only available on Mesosphere Enterprise DC/OS."
-        );
-
-      cy.get(".modal button").contains("Close").click();
-
-      cy.get(".modal").should("not.exist");
-    });
-
     it("opens the scale dialog", function() {
       openDropdown("sdk-sleep");
       clickDropdownAction("Scale");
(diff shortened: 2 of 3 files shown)


plugins/services/src/js/components/modals/ServiceActionDisabledModal.js
```diff
@@ -223,27 +224,16 @@ class ServiceActionDisabledModal extends React.Component {
...
-        {this.getClipboardTrigger(command)}
-      </div>
+      <MountService.Mount
+        type={"ServiceEditMessage:Modal"}
+        getClipboardTrigger={this.getClipboardTrigger}
+        command={command}
+        intl={intl}
+      >
+        <div>
+          Only available on EE.
```
MatApple: Although we spoke offline - I'm making a note to get appropriate copy from @ashenden. My suggestion would be:
```
Editing this service is only available on <a href="https://mesosphere.com/product/">Mesosphere Enterprise DC/OS</a>.
```

plugins/services/src/js/components/modals/ServiceActionDisabledModal.js
```diff
@@ -223,27 +224,21 @@ class ServiceActionDisabledModal extends React.Component {
...
-          {intl.formatMessage({
-            id: "SERVICE_ACTIONS.SDK_SERVICE_UPDATE_PART_1"
-          })}
+      <MountService.Mount
+        type={"ServiceEditMessage:Modal"}
+        getClipboardTrigger={this.getClipboardTrigger}
+        command={command}
+        intl={intl}
+      >
+        <div>
```
MatApple: Let's center align this content.

plugins/services/src/js/components/modals/ServiceActionDisabledModal.js
```diff
@@ -223,27 +224,21 @@ class ServiceActionDisabledModal extends React.Component {
...
     return (
-      <div>
-        <p>
-          {intl.formatMessage({
-            id: "SERVICE_ACTIONS.SDK_SERVICE_UPDATE_PART_1"
-          })}
+      <MountService.Mount
+        type={"ServiceEditMessage:Modal"}
+        getClipboardTrigger={this.getClipboardTrigger}
+        command={command}
```
MatApple: Let's rethink this. I don't think it makes sense to build the command here when it's not shown by default.

Poltergeist:
run integration tests

plugins/services/src/js/components/modals/ServiceActionDisabledModal.js
ahoskins: Done

plugins/services/src/js/components/modals/ServiceActionDisabledModal.js
ahoskins: Updated to pass `getUpdateCommand` instead of the command directly.

plugins/services/src/js/components/modals/ServiceActionDisabledModal.js
```diff
@@ -220,30 +225,23 @@ class ServiceActionDisabledModal extends React.Component {
...
+          Editing this service is only available on
           {" "}
-          <a
-            href={MetadataStore.buildDocsURI(
-              "/usage/managing-services/config-universe-service"
-            )}
-            target="_blank"
-          >
-            options.json
+          <a href="https://mesosphere.com/product/" target="_blank">
```
weblancaster: Is this link correct?

tests/pages/services/ServiceTable-cy.js
```diff
@@ -234,9 +234,9 @@ describe("Service Table", function() {
         .should("to.have.length", 1);
 
       cy
-        .get(".modal pre")
+        .get(".modal-body")
```
weblancaster: I would just remove this test entirely because it's duplicated, but not bug deal for me to block from merging.

plugins/services/src/js/components/modals/ServiceActionDisabledModal.js
ahoskins: It works, and is the link @MatApple suggested to include.

tests/pages/services/ServiceTable-cy.js
ahoskins: Agree. This modal is fully tested in`ServiceAction-cy.js`.

ahoskins: Actually, I'm not going to remove this because there are others like destroy, scale, that are in the exact same situation with duplicated tests. I don't want to change this one and then have inconsistency with the other modal states.

weblancaster: I don't think removing one will create inconsistency because otherwise we could apply the same argument to the other tests that don't have duplicated.

plugins/services/src/js/components/modals/ServiceActionDisabledModal.js
weblancaster: ok

plugins/services/src/js/components/modals/ServiceActionDisabledModal.js
```diff
@@ -220,30 +225,23 @@ class ServiceActionDisabledModal extends React.Component {
...
 
     return (
-      <div>
-        <p>
-          {intl.formatMessage({
-            id: "SERVICE_ACTIONS.SDK_SERVICE_UPDATE_PART_1"
-          })}
+      <MountService.Mount
+        type={"ServiceEditMessage:Modal"}
+        getClipboardTrigger={this.getClipboardTrigger}
```
MatApple: Sorry, I should've been more clear with my last comment! I don't think the `getClipboardTrigger` and `getUpdateCommand` should live here. We shouldn't be passing commands into the MountService. Something is very wrong if we have to be doing that.

tests/pages/services/ServiceTable-cy.js
ahoskins: True, will remove this test.

plugins/services/src/js/components/modals/ServiceActionDisabledModal.js
ahoskins: I don't understand what would be a better alternative. `getUpdateCommand` has other consumers in this file, and relies on `isSDKService`. Isn't the alternative to duplicate this into private-plugins?

MatApple: Yes. This is a case where code duplication is ok. We want to avoid binding components together like this.

ahoskins: I've changed this to just pass `intl` and `service` into the mounted component. I think it's a lot better and more clear interface now.

plugins/services/src/js/components/modals/ServiceActionDisabledModal.js
```diff
@@ -219,31 +220,23 @@ class ServiceActionDisabledModal extends React.Component {
...
-    const command = this.getUpdateCommand();
+    const { intl, service } = this.props;
 
     return (
-      <div>
-        <p>
-          {intl.formatMessage({
-            id: "SERVICE_ACTIONS.SDK_SERVICE_UPDATE_PART_1"
-          })}
+      <MountService.Mount
```
MatApple: Much cleaner :)

orlandohohmeier:
LGTM! 🚢

PR #2346, opened 2017-08-01
fix(MesosStateStore): include unreachable tasks

Poltergeist:
Include unreachable tasks into the task list. Also make them selectable in the Task table in the
Service detail page.

To check the issue you have to deploy an application and then shut down a agent node where one task
is currently running. This does take a couple of steps

- Find out the master IP
- find out the internal agent IP
- ssh into the master best with -A option so that you can ssh into the agent with the same keyfile
- ssh from the master to the agent node
- run `sudo systemctl stop dcos-mesos-slave.service`
- wait about 10 minutes.

Your service will be unreachable at this point and it will still be in the task table.

Also a unreachable tasks are like Schroedingers, completed and active at the same time we decieded to
rather see them as active so we can kill / restart them from the UI.

Closes DCOS_OSS-940

**Checklist**

### DIFF
--- a/src/js/stores/MesosStateStore.js
+++ b/src/js/stores/MesosStateStore.js
@@ -134,8 +134,13 @@ class MesosStateStore extends GetSetBaseStore {
     const taskIndex = {};
 
     lastMesosState.frameworks.forEach(function(service) {
-      const tasks = service.tasks.concat(service.completed_tasks);
-      tasks.forEach(function(task) {
+      const {
+        tasks = [],
+        completed_tasks = [],
+        unreachable_tasks = []
+      } = service;
+
+      tasks.concat(completed_tasks, unreachable_tasks).forEach(function(task) {
         taskIndex[task.id] = task;
       });
     });
@@ -264,8 +269,9 @@ class MesosStateStore extends GetSetBaseStore {
     if (framework) {
       const activeTasks = framework.tasks || [];
       const completedTasks = framework.completed_tasks || [];
+      const unreachableTasks = framework.unreachable_tasks || [];
 
-      return activeTasks.concat(completedTasks);
+      return activeTasks.concat(completedTasks, unreachableTasks);
     }
 
     return [];
@@ -292,8 +298,13 @@ class MesosStateStore extends GetSetBaseStore {
     // Marathon tasks. This will give you a list of framework tasks including
     // the scheduler tasks or a list of Marathon application tasks.
     return frameworks.reduce(function(serviceTasks, framework) {
-      const { tasks = [], completed_tasks = {}, name } = framework;
-
+      const {
+        tasks = [],
+        completed_tasks = [],
+        unreachable_tasks = [],
+        name
+      } = framework;
+      // Include tasks from framework match, if service is a Framework
       if (serviceIsFramework && serviceFrameworkName === name) {
         return serviceTasks
           .concat(tasks, completed_tasks)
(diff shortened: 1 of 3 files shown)


src/js/stores/MesosStateStore.js
```diff
@@ -288,7 +297,12 @@ class MesosStateStore extends GetSetBaseStore {
     // Marathon tasks. This will give you a list of framework tasks including
     // the scheduler tasks or a list of Marathon application tasks.
     return frameworks.reduce(function(serviceTasks, framework) {
-      const { tasks = [], completed_tasks = {}, name } = framework;
+      const {
+        tasks = [],
+        completed_tasks = {},
```
ahoskins: It's strange how this default value is an object, but above it's an array, and all other task data types are arrays.

Poltergeist: I am going to find out why it is an object I speculated as well but then figured that there was some reason. I didn't changed this tho'.

Poltergeist: It actually makes no sense, I'll adjust. to default `[]`

Poltergeist:
run integration tests

src/js/stores/MesosStateStore.js
```diff
@@ -300,6 +314,7 @@ class MesosStateStore extends GetSetBaseStore {
       if (name === "marathon") {
         return tasks
           .concat(completed_tasks)
+          .concat(unreachable_tasks)
```
weblancaster: why not re-use the first `concat(completed_tasks, unreachable_tasks)` as one operation?

src/js/stores/MesosStateStore.js
```diff
@@ -264,8 +272,9 @@ class MesosStateStore extends GetSetBaseStore {
     if (framework) {
       const activeTasks = framework.tasks || [];
       const completedTasks = framework.completed_tasks || [];
+      const unreachableTasks = framework.unreachable_tasks || [];
 
-      return activeTasks.concat(completedTasks);
+      return activeTasks.concat(completedTasks).concat(unreachableTasks);
```
weblancaster: why not re-use the first `concat(completed_tasks, unreachable_tasks)` as one operation?

src/js/stores/MesosStateStore.js
weblancaster: same as previous comment.

src/js/stores/MesosStateStore.js
```diff
@@ -134,10 +134,18 @@ class MesosStateStore extends GetSetBaseStore {
...
-      });
+      const {
+        tasks = [],
+        completed_tasks = [],
+        unreachable_tasks = []
+      } = service;
+
+      tasks
+        .concat(completed_tasks)
+        .concat(unreachable_tasks)
```
weblancaster: We can just use one concat and keep same operation

Poltergeist: You are right I changed the behaviour.

src/js/stores/MesosStateStore.js
Poltergeist: You are right I changed the behaviour.

orlandohohmeier:
LGTM 💯 ✨

MatApple:
Rebased this because it had conflicts. @weblancaster @orlandohohmeier @ahoskins - need another approval 👍

PR #2347, opened 2017-08-01
DCOS-15921 (1 of 2): Add ComponentExample component

tnguyen0:
**Design System**

This PR adds the `ComponentExample` component, which is made up of the `CodeExampleHeader` and `CodeExampleFooter` components. It is the primary component for the code example component.

To test:

- See branch `tommy/design-system/component-example-ace-editor` for the complete implementation or merge in #2349
- In the UI: Design System --> Components --> Checkboxes --> Code Tab
- React code is shown on one tab and HTML on the other
- Clipboard function works for both tabs
- Collapse/Expand of code

Options:

Implement ComponentExample component used to display code blocks for design system components. Code can either be in React or HTML. Props can be added to the component to override the generated React and HTML Code. These optional props are: reactCode and htmlCode and are set  with the desired custom content. The component preview is set through the preview prop. It is a mandatory prop unless reactCode and htmlCode are both set.  There are additional props to show/hide these parts: showPreview, showHtml and showReact. These props are set to true by default.

There are several different combinations possible:

    * All (Default with Tabs): preview={preview}
    * Preview Only: preview={preview}, showReact={false}, showHtml={false}
    * Generated React Only: preview={preview}, showHtml={false}
    * Generated HTML Only: preview={preview}, showReact={false}
    * Custom React Only: preview={preview}, reactCode={reactCode}, showHtml={false}
    * Custom HTML Only: preview={preview}, htmlCode={htmlCode}, showReact={false}
    * Custom React: preview={preview}, reactCode={reactCode}
    * Custom HTML: preview={preview}, htmlCode={htmlCode}
    * Custom React and HTML: preview={preview}, reactCode={reactCode}, htmlCode={htmlCode}
    * Omit Preview: showPreview={false}

Closes DCOS-15921

**Checklist**

### DIFF
--- a/plugins/design-system/components/component-example/CodeExampleFooter.js
+++ b/plugins/design-system/components/component-example/CodeExampleFooter.js
@@ -1,0 +1,73 @@
+import classNames from "classnames/dedupe";
+import React, { Component } from "react";
+
+const defaultClasses = {
+  panel: "code-example-wrapper footer",
+  content: "code-example-footer-content"
+};
+
+class CodeExampleFooter extends Component {
+  getNode(nodeName) {
+    const { props } = this;
+    const node = props[nodeName];
+
+    if (!node) {
+      return null;
+    }
+
+    const classes = classNames(
+      defaultClasses[nodeName],
+      props[nodeName + "Class"]
+    );
+
+    return (
+      <div className={classes}>
+        {node}
+      </div>
+    );
+  }
+
+  render() {
+    const { props } = this;
+    const contentClasses = classNames(
+      defaultClasses.content,
+      props.contentClass
+    );
+    const panelClasses = classNames(defaultClasses.panel, props.className);
+
+    return (
+      <div className={panelClasses} onClick={this.props.onClick}>
+        {this.getNode("heading")}
+        <div className={contentClasses}>
+          {props.children}
+        </div>
+        {this.getNode("footer")}
+      </div>
+    );
+  }
+}
+
+CodeExampleFooter.propTypes = {
+  heading: React.PropTypes.node,
+  footer: React.PropTypes.node,
+
+  contentClass: React.PropTypes.oneOfType([
+    React.PropTypes.array,
+    React.PropTypes.object,
+    React.PropTypes.string
+  ]),
+  headingClass: React.PropTypes.oneOfType([
+    React.PropTypes.array,
+    React.PropTypes.object,
+    React.PropTypes.string
+  ]),
+  footerClass: React.PropTypes.oneOfType([
+    React.PropTypes.array,
+    React.PropTypes.object,
+    React.PropTypes.string
+  ]),
+
+  onClick: React.PropTypes.func
+};
+
+module.exports = CodeExampleFooter;
--- a/plugins/design-system/components/component-example/ComponentExample.js
+++ b/plugins/design-system/components/component-example/ComponentExample.js
@@ -1,0 +1,248 @@
+import React, { Component } from "react";
+import ReactDOMServer from "react-dom/server";
+import reactElementToJSXString from "react-element-to-jsx-string";
+
+import StringUtil from "#SRC/js/utils/StringUtil";
+import ClipboardTrigger from "#SRC/js/components/ClipboardTrigger";
+import Icon from "#SRC/js/components/Icon";
+
+import CodeExample from "./CodeExample";
+import CodeExampleFooter from "./CodeExampleFooter";
+import CodeExampleHeader from "./CodeExampleHeader";
+
+import CodeExampleTabButtonList from "./CodeExampleTabButtonList";
+import CodeExampleTabs from "./CodeExampleTabs";
+import CodeExampleTabView from "./CodeExampleTabView";
+import CodeExampleTabViewList from "./CodeExampleTabViewList";
+import CodeExampleTabButton from "./CodeExampleTabButton";
+
+import ExpandButton from "./ExpandButton";
+import CollapseButton from "./CollapseButton";
+
+const DEFAULT_HEIGHT = 205;
+
+const REACT_STRING_OPTIONS = {
+  showDefaultProps: false,
+  useBooleanShorthandSyntax: false
+};
+
+const TABS = {
+  React: {
+    label: "REACT",
+    id: "react"
+  },
+  Html: {
+    label: "HTML",
+    id: "html"
+  }
+};
+
+const METHODS_TO_BIND = [
+  "toggleExpandCollapse",
+  "handleTabChange",
+  "handleCodeCopy",
+  "handleCanExpand"
+];
+
+class ComponentExample extends Component {
+  constructor() {
+    super(...arguments);
+
+    const activeTab = this.isTabbedMode()
+      ? this.props.activeTab || TABS.React.id
+      : null;
+
+    this.state = {
+      isExpanded: false,
+      isCodeCopied: false,
+      canExpand: false,
+      activeTab
+    };
+
+    METHODS_TO_BIND.forEach(method => {
+      this[method] = this[method].bind(this);
+    });
+  }
+
+  showPreview() {
+    return this.props.showPreview && this.props.preview;
+  }
+
+  isSingleBlockMode() {
+    return (
+      (this.props.showReact || this.props.showHtml) &&
+      this.props.showHtml !== this.props.showReact
+    );
+  }
+
+  isTabbedMode() {
+    return this.props.showReact && this.props.showHtml;
+  }
+
+  isReactMode() {
+    return this.isSingleBlockMode() && this.props.showReact;
+  }
+
+  isShowingReactCode() {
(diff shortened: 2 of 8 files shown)


plugins/design-system/components/component-example/ComponentExample.js
```diff
@@ -0,0 +1,249 @@
...
+
+  handleCodeCopy() {
+    this.setState({ isCodeCopied: true });
+  }
+
+  generateCodeExample(lang, code) {
+    return (
+      <CodeExample
+        lang={lang}
+        height={this.state.isExpanded ? "100%" : this.defaultHeight}
```
weblancaster: would be nice to extract this to a `const`

plugins/design-system/components/component-example/ComponentExampleOverrides.js
```diff
@@ -0,0 +1,7 @@
+const ComponenxtExampleOverrides = {};
+
+["ReactCode", "HtmlCode", "Preview"].forEach(function(element) {
```
weblancaster: I see the Higher order components did not work?

tnguyen0: I haven't dived too deeply into it yet. My initial impression is that it is more than what I need for my use case.

plugins/design-system/components/component-example/ComponentExample.js
```diff
@@ -0,0 +1,249 @@
...
+  "handleTabChange",
+  "handleCodeCopy",
+  "handleCanExpand"
+];
+
+class ComponentExample extends Component {
+  constructor() {
+    super(...arguments);
+
+    this.state = {
```
weblancaster: what do you think about

```
this.state = {
      isExpanded: false,
      isCodeCopied: false,
      canExpand: false,
      activeTab: this.isTabbedMode() ? this.props.activeTab || ReactCode : null
    };
```

plugins/design-system/components/component-example/ComponentExample.js
```diff
@@ -0,0 +1,249 @@
...
+      return child.type !== ReactCode && child.type !== HtmlCode;
+    });
+  }
+
+  expandCollapseToggle() {
+    this.setState({ isExpanded: !this.state.isExpanded });
+  }
+
+  handleTabChange(activeTab) {
+    this.setState({ activeTab });
```
weblancaster: Let's use only one `setState` here.

```
this.setState({
      activeTab,
      isExpanded: false
    });
```

weblancaster: whenever we `setState` that triggers React lifecycle

tnguyen0: Didn't know you could mix it like that. 👍

plugins/design-system/components/component-example/ComponentExampleOverrides.js
```diff
@@ -0,0 +1,7 @@
+const ComponenxtExampleOverrides = {};
```
nLight: I think there's a type here `Componen*x*tExampleOverrides `

tnguyen0: Good catch, fixed!

plugins/design-system/components/component-example/ComponentExample.js
nLight: I'd rather see this conditional extracted:
```
activeTab: this.isTabbedMode() ? this.props.activeTab || ReactCode : null
```

plugins/design-system/components/component-example/ComponentExample.js
```diff
@@ -0,0 +1,244 @@
...
+
+  handleCanExpand(canExpand) {
+    this.setState({ canExpand });
+  }
+
+  handleCodeCopy() {
+    this.setState({ isCodeCopied: true });
+  }
+
+  generateCodeExample(lang, code) {
```
nLight: I'd replaced this function with a `getCodeBody` function and moved everything there.

plugins/design-system/components/component-example/ComponentExample.js
```diff
@@ -0,0 +1,244 @@
...
+    };
+
+    this.defaultHeight = this.props.defaultHeight || DEFAULT_HEIGHT;
+
+    METHODS_TO_BIND.forEach(method => {
+      this[method] = this[method].bind(this);
+    });
+  }
+
+  findChildByType(type) {
```
nLight: Wouldn't it be better introducing props for `Preview`, `ReactCode` and `HtmlCode` instead of having a rather weird looking `ComponentExampleOverrides` and constantly looping though the children to find a one of them?

plugins/design-system/components/component-example/ComponentExample.js
```diff
@@ -0,0 +1,244 @@
...
+        height={height}
+        handleCanExpand={this.handleCanExpand}
+      >
+        {`${code}`}
+      </CodeExample>
+    );
+  }
+
+  generateFooter() {
+    const expandButton = (
```
nLight: How about we extract two private components?
```
const toggleButton = this.state.isExpanded ? <CollapseButton onClick={...} /> : <ExpandButton onClick={...} />;
```

plugins/design-system/components/component-example/ComponentExample.js
```diff
@@ -0,0 +1,244 @@
...
+  getReactCode(jsxCode) {
+    let reactCode = this.findChildByType(ReactCode);
+    if (reactCode) {
+      reactCode = reactCode.props.children;
+    }
+
+    return reactElementToJSXString(reactCode || jsxCode, REACT_STRING_OPTIONS);
+  }
+
+  expandCollapseToggle() {
```
nLight: Name suggestion `toggleExpandCollapse`

plugins/design-system/components/component-example/CodeExampleFooter.js
```diff
@@ -0,0 +1,73 @@
+import classNames from "classnames/dedupe";
+import React, { Component } from "react";
+
+const defaultClasses = {
```
nLight: So `CodeExampleFooter` and `CodeExampleHeader` are essentially the same only `defaultClasses` are different. Thinking if it is a ok duplication or not.

plugins/design-system/components/component-example/ComponentExample.js
```diff
@@ -0,0 +1,244 @@
...
+    super(...arguments);
+
+    this.state = {
+      isExpanded: false,
+      isCodeCopied: false,
+      canExpand: false,
+      activeTab: this.isTabbedMode() ? this.props.activeTab || ReactCode : null
+    };
+
+    this.defaultHeight = this.props.defaultHeight || DEFAULT_HEIGHT;
```
nLight: Would `ComponentExample.defaultProps = {defaultHeight: DEFAULT_HEIGHT}` do the trick?

plugins/design-system/components/component-example/CodeExampleFooter.js
weblancaster: I was thinking the same.. but it's a real paradox because we should be creating new components when has something different trying to avoid custom config down the road but at the same time is so minimum that maybe makes sense to have one for both

plugins/design-system/components/component-example/ComponentExample.js
weblancaster: seems redundant.. why no only `toggle`? otherwise I would go with @nLight suggestions `toggleExpandCollapse `

plugins/design-system/components/component-example/ComponentExample.js
weblancaster: that is also a good option.

plugins/design-system/components/component-example/ComponentExample.js
tnguyen0: Yes indeed! :)

plugins/design-system/components/component-example/ComponentExample.js
tnguyen0: I will go with `toggleExpandCollapse` since it is more specific. Thanks guys!

plugins/design-system/components/component-example/ComponentExample.js
tnguyen0: Much better 👍

plugins/design-system/components/component-example/ComponentExample.js
tnguyen0: Renamed!

plugins/design-system/components/component-example/CodeExampleFooter.js
tnguyen0: It's a decision it seems I have to make often when using React. I think it is okay in this case.

tnguyen0:
@weblancaster @nLight can you guys take another look?

nLight:
Thank you for addressing those comments!

PR #2348, opened 2017-08-01
fix(CreateServiceModalCatalogPanelOption): fix redirect to catalog

ahoskins:
**Problem**:
Clicking the "install a package" options redirects to an improper location and hence directs to the dashboard page. Demo:

https://cl.ly/0K0m0f2g0O2i

**Fix**:
Adding a slash before the "catalog" name fixes the problem. Now clicking the "install a package" will direct to the catalog page.

**Checklist**

### DIFF
--- a/src/js/components/CreateServiceModalCatalogPanelOption.js
+++ b/src/js/components/CreateServiceModalCatalogPanelOption.js
@@ -17,7 +17,7 @@ class CreateServicePickerCatalogOption extends React.Component {
       <CreateServiceModalServicePickerOption
         columnClasses={columnClasses}
         onOptionSelect={onOptionSelect.bind(null, {
-          route: "catalog",
+          route: "/catalog",
           type: "redirect"
         })}
       >


src/js/components/CreateServiceModalCatalogPanelOption.js
```diff
@@ -17,7 +17,7 @@ class CreateServicePickerCatalogOption extends React.Component {
       <CreateServiceModalServicePickerOption
         columnClasses={columnClasses}
         onOptionSelect={onOptionSelect.bind(null, {
-          route: "catalog",
+          route: "/catalog",
```
weblancaster: As I discussed with @ahoskins this was the only solution found without breaking the other panel buttons but perhaps there is a bug somewhere else.

ahoskins:
As @weblancaster mentioned in the comment, this PR is a reluctant patch after quite a bit of time spent trying to find why this issue was actually happening.

The strange thing is, we could only reproduce this with an Enterprise configuration - toggling the enterprise flag causes this issue to disappear. This suggests something in the Enterprise routing was different, but we couldn't find what was happening, so we arrived at this simple fix.

PR #2349, opened 2017-08-01
DCOS-15921 (2 of 2): Use ComponentExample for design system checkbox

tnguyen0:
Uses the ComponentExample code preview component to show the React and HTML code
representation of the design system checkbox component.

:warning: This PR depends on: #2347

See branch `tommy/design-system/component-example-ace-editor` for the complete implementation.

In the UI: `Design System` --> `Components` --> `Checkboxes` --> `Code Tab`

**Checklist**


src/styles/index.less
```diff
@@ -5,6 +5,7 @@
 @import '../../node_modules/cnvs/styles/cnvs.less'; // Source CNVS
 @import 'vendor/cnvs/styles.less'; // CNVS Overrides
 @import 'vendor/loaders.less';
+@import '../../plugins/design-system/styles/styles.less'; // TODO: Tommy figure out where to include this.
```
nLight: What's the problem with that one?

plugins/design-system/styles/shame.less
```diff
@@ -1,5 +1,4 @@
 // Shame CSS
-// A place to put quick CSS iterations and cleanup later
```
nLight: Can we delete this file? In this PR https://github.com/dcos/dcos-ui/pull/2388/files you've introduced a separate file for the `do-dont` component. Doesn't matter how messy css is it should all go there.

src/styles/index.less
tnguyen0: We want to be able to use CNVS styles for plugins. The point was discussed and it was decided this way is fine. I will remove the comment.

plugins/design-system/styles/shame.less
tnguyen0: Yes :)

nLight:
Alfred, retest this please

PR #2350, opened 2017-08-02
fix(TasksView): disable restart from tasksViews

weblancaster:
Disable restart in TasksView when is SDK, this is part of #2343

Closes DCOS-16564

**Checklist**

### DIFF
--- a/plugins/services/src/js/containers/tasks/TasksView.js
+++ b/plugins/services/src/js/containers/tasks/TasksView.js
@@ -180,7 +180,7 @@ class TasksView extends mixin(SaveStateMixin) {
     let handleRestartClick = function() {};
     let handleStopClick = function() {};
 
-    if (!isDeploying || !isSDK) {
+    if (!isDeploying && !isSDK) {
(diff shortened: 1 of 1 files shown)


plugins/services/src/js/containers/tasks/TasksView.js
```diff
@@ -161,6 +162,13 @@ class TasksView extends mixin(SaveStateMixin) {
       return service.getServiceStatus().key === ServiceStatusTypes.DEPLOYING;
     });
 
+    const isSDK = Object.keys(checkedItems).some(function(taskId) {
+      const service = DCOSStore.serviceTree.getServiceFromTaskID(taskId);
+      const isSDK = isSDKService(service);
```
MatApple: #nit but we don't need to assign to a variable here. We can just `return isSDKService(service);`

weblancaster: true

MatApple:
Nice catch @bstavroulakis @weblancaster 👍

PR #2351, opened 2017-08-03
Adjust Health Reporting for SDK Tasks

MatApple:
All tasks for an SDK service have been changed to report "healthy" by default, if in RUNNING state and they haven't been reported as "unhealthy". Previously tasks matching this criteria reported "unknown".

This change affects the tasks associated with an SDK service (service detail page for SDK service, and node detail page, where SDK tasks are running on that node).

SDK service detail before:

SDK service detail after:

Node detail before:

Node detail after:

**Checklist**

### DIFF
--- a/plugins/services/src/js/structs/Framework.js
+++ b/plugins/services/src/js/structs/Framework.js
@@ -1,4 +1,5 @@
 import { cleanServiceJSON } from "#SRC/js/utils/CleanJSONUtil";
+import { isSDKService } from "#SRC/js/utils/ServiceUtil";
 
 import {
   ROUTE_ACCESS_PREFIX,
@@ -32,6 +33,7 @@ module.exports = class Framework extends Application {
   }
 
   getTasksSummary() {
+    const isSDK = isSDKService(this);
     // TODO: Circular reference workaround DCOS_OSS-783
     const MesosStateStore = require("#SRC/js/stores/MesosStateStore");
 
@@ -47,7 +49,7 @@ module.exports = class Framework extends Application {
       }
       if (task.statuses != null) {
         return task.statuses.reduce(function(memo, status) {
-          if (status.healthy) {
+          if (status.healthy || (isSDK && status.healthy === undefined)) {
             memo.tasksHealthy++;
             memo.tasksUnknown--;
           }
--- a/plugins/services/src/js/structs/ServiceTree.js
+++ b/plugins/services/src/js/structs/ServiceTree.js
@@ -119,7 +119,10 @@ module.exports = class ServiceTree extends Tree {
    */
   findServiceByName(name) {
     return this.findItem(function(item) {
-      return item.getName() === name && item instanceof Service;
+      return (
+        (item instanceof Framework && item.getFrameworkName() === name) ||
+        (item instanceof Service && item.getName() === name)
+      );
     });
   }
 
--- a/plugins/services/src/js/structs/ServiceTree.js
+++ b/plugins/services/src/js/structs/ServiceTree.js
@@ -113,19 +113,6 @@ module.exports = class ServiceTree extends Tree {
     });
   }
 
-  /**
-   * @param {string} name
-   * @return {Service} matching Service
-   */
-  findServiceByName(name) {
-    return this.findItem(function(item) {
-      return (
-        (item instanceof Framework && item.getFrameworkName() === name) ||
-        (item instanceof Service && item.getName() === name)
-      );
-    });
-  }
-
   // TODO @pierlo-upitup MARATHON-1030: refactor for more generic usage
   filterItemsByFilter(filter) {
     let services = this.getItems();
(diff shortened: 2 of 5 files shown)


plugins/services/src/js/structs/Framework.js
```diff
@@ -47,7 +49,7 @@ module.exports = class Framework extends Application {
       }
       if (task.statuses != null) {
         return task.statuses.reduce(function(memo, status) {
-          if (status.healthy) {
+          if (status.healthy || (isSDK && status.healthy === undefined)) {
```
weblancaster: `status.healthy === undefined` seems like code smell here because I believe `status.healthy` should always be a defined type.

MatApple: This field can either be `true`, `false` or `undefined` because the property doesn't exist. I do agree that is SHOULD be a defined type - just like all our primitive types.

weblancaster: Got it.

src/js/stores/MesosStateStore.js
```diff
@@ -194,18 +211,24 @@ class MesosStateStore extends GetSetBaseStore {
...
     const schedulerTasks = this.getSchedulerTasks();
 
-    services.forEach(service => {
-      service.tasks.forEach(function(task) {
+    frameworks.forEach(framework => {
+      framework.tasks.forEach(function(task) {
         if (task.slave_id === nodeID) {
-          memberTasks[task.id] = assignSchedulerTaskField(task, schedulerTasks);
+          // Need to get service from Marathon because we need the labels
+          const service = DCOSStore.serviceTree.findServiceByName(
```
orlandohohmeier: Please introduce a custom find as `findServiceByName` could return the wrong service.  

```js
serviceTree.findItem((item) => item instanceof Framework && item.getFrameworkName() === framework.name)
```

src/js/stores/MesosStateStore.js
```diff
@@ -49,6 +52,20 @@ function assignSchedulerTaskField(task, schedulerTasks) {
   return task;
 }
 
+/**
+ * Assigns a property to task if it belongs to an SDK service.
+ * @param  {Object} task
+ * @param  {Array} service task belongs to
+ * @return {Object} task
+ */
+function flagSDKTask(task, service) {
```
orlandohohmeier: Not blocking but it might make sense to move this one along with a `isSDKTask` to `TaskUtil`.

plugins/services/src/js/structs/ServiceTree.js
```diff
@@ -119,7 +119,10 @@ module.exports = class ServiceTree extends Tree {
    */
   findServiceByName(name) {
     return this.findItem(function(item) {
-      return item.getName() === name && item instanceof Service;
+      return (
+        (item instanceof Framework && item.getFrameworkName() === name) ||
```
MatApple: The only time this could happen is if someone creates an App and adds their own custom label `DCOS_PACKAGE_FRAMEWORK_NAME`. `id` and `DCOS_PACKAGE_FRAMEWORK_NAME` will at most differ by a leading slash and that is the whole reason we need `getFrameworkName`. Cosmos derives the marathon app `id` from `name` at install so if you install with `name: "foo/bar"` or `name: "/foo/bar"`  you'll get `id: "/foo/bar"` in both cases.

MatApple: I do agree that they are semantically different, but if anything we should never be using what we previously had since that can also lead to finding the wrong service: 
```js
const serviceTree = new ServiceTree({items:[{
    id: "foo/bar"
},
{
    id: "baz/bar"
}]})

serviceTree.findServiceByName("bar"); // -> finds id = foo/bar but really we wanted id = baz/bar
```

MatApple: Since this method is only being used once in `MesosStateStore` and only needs the case where we find by `getFrameworkName`, I think the former behavior should be completely removed.

src/js/stores/MesosStateStore.js
MatApple: Since `findServiceByName` is only being used here, only needs the case where we find by `getFrameworkName` and was fundamentally broken previously, I have removed the former behavior in favor of the new behavior.

orlandohohmeier: That just one more reason to introduce a custom find. Introducing a custom/specific find will also allow us to eventually remove the old "fundamentally broken" `findServiceByName` method.

plugins/services/src/js/structs/ServiceTree.js
ahoskins: In this above example @MatApple gave, what would we ideally want as output? A list of matching services, or first match (how it is now), or prioritize matching frameworks or services?

orlandohohmeier:
LGTM! 🚢

plugins/services/src/js/structs/ServiceTree.js
ahoskins: (comment no longer relevant, method removed) 👍

PR #2353, opened 2017-08-07
feat(object-utilities) introduce a new object-utilities package

juliangieseke:
This Package will hold object related utilities.
Feel free to add more and to refactor/resort stuff in here.

The first utility inside this package will be `__protected`:

Using private variables is relatively straightforward with local context
variables, but protected is a bit more tricky. This Utility enables
us to use protected variables like this:

```
class Something {
  constructor() {
    __protected(this, { foo: "bar" });
  }
  func() {
    __protected(this).foo =  __protected(this).foo + "baz" // barbaz
  }
}
class SubSomething {
  constructor() {
    __protected(this).foo = "subbaz";
  }
}
```

**Under the Hood:**
After some fiddling, I came to an approach where we simply add a __protected__ property to the class/object. Its not a real protected property, but as it is freezed and "__"-prefixed we can say its "protected by definition".

**Please discuss / give feedback.**

### DIFF
--- a/packages/object-utilities/index.js
+++ b/packages/object-utilities/index.js
@@ -1,0 +1,1 @@
+export { default } from "./src/js/__protected.js";
--- a/packages/object-utilities/src/js/__protected.js
+++ b/packages/object-utilities/src/js/__protected.js
@@ -1,0 +1,44 @@
+/**
+ * Creates a non-enumerable, non-configurable
+ * and non-writeable property in the given object.
+ * Use this for properties you want to be considered
+ * as protected, functions are public for now.
+ *
+ * Usage:
+ * class WithProtected {
+ *   constructor() {
+ *     // set multiple properties
+ *     __protected(this, { foo: "bar", some: "thing"});
+ *     // set/change one
+ *     __protected(this).foo = "baz";
+ *     // read
+ *     alert(__protected(this).foo);
+ *   }
+ * }
+ *
+ * @param {Object} object – given object
+ * @param {Object} [value] – given value
+ * @return {Object} protected property / context
+ */
+export default function __protected(object, value) {
+  if (object === undefined || object === null) {
+    throw new Error("no object given or null");
+  }
+  if (object.__protected__ === undefined) {
+    Object.defineProperty(object, "__protected__", {
+      configurable: false,
+      enumerable: false,
+      writeable: false,
+      value: {}
+    });
+  }
+  if (value !== undefined) {
+    if (typeof value === "object") {
+      Object.assign(object.__protected__, value);
+    } else {
+      throw new Error("value must be an object");
+    }
+  }
+
+  return object.__protected__;
+}
--- a/packages/object-utilities/src/js/__tests__/__protected-test.js
+++ b/packages/object-utilities/src/js/__tests__/__protected-test.js
@@ -1,0 +1,40 @@
+jest.dontMock("../__protected");
+
+const __protected = require("../__protected").default;
+
+describe("__protected", () => {
+  it("initializes", () => {
+    const obj = {};
+    expect(__protected(obj)).toEqual({});
+  });
+
+  it("throws error without object", () => {
+    expect(() => {
+      __protected();
+    }).toThrowError();
+  });
+
+  it("throws error with invalid value", () => {
+    const obj = {};
+    expect(() => {
+      __protected(obj, "foobar");
+    }).toThrowError();
+  });
+
+  it("saves values through value parameter", () => {
+    const obj = {};
+    expect(__protected(obj, { foo: "bar" }).foo).toEqual("bar");
+  });
+
+  it("saves value through returned object", () => {
+    const obj = {};
+    __protected(obj).foo = "bar";
+    expect(__protected(obj).foo).toEqual("bar");
+  });
+
+  it("changes value through value parameter", () => {
+    const obj = {};
+    __protected(obj, { foo: "bar" });
+    expect(__protected(obj, { foo: "baz" }).foo).toEqual("baz");
+  });
+});


orlandohohmeier:

packages/object-utilities/src/js/__tests__/__protected-test.js
```diff
@@ -0,0 +1,42 @@
...
+const __protected = require("../__protected").default;
+
+describe("ProtectedUtil", () => {
+  it("throw error without object", () => {
+    expect(() => {
+      __protected();
+    }).toThrowError();
+  });
+
+  it("call without value creates empty protected property and returns that object", () => {
```
orlandohohmeier: Please adjust the test description according to the guidelines http://www.betterspecs.org/

packages/object-utilities/src/js/__tests__/__protected-test.js
```diff
@@ -0,0 +1,42 @@
+jest.dontMock("../__protected");
+
+// import { __protected } from "../__protected";
```
orlandohohmeier: Please always remove unused lines.

packages/object-utilities/index.js
```diff
@@ -0,0 +1 @@
+export { default } from "./src/js/__protected.js";
```
orlandohohmeier: Could you please check that the import works as follows?

```js
import {__protected} from "object-utilities"
```

packages/object-utilities/src/js/__protected.js
```diff
@@ -0,0 +1,37 @@
+/**
+ * Creates an protected property in the given object
```
orlandohohmeier: Please clearly state that the "protected" property is just non-enumerable and not "really" protected.

packages/object-utilities/src/js/__protected.js
```diff
@@ -0,0 +1,37 @@
...
+ * Use this for protected properties, functions are public for now.
+ *
+ * First Usage:
```
orlandohohmeier: Let's remove the distinction between first and any other usage.

packages/object-utilities/src/js/__protected.js
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * After that:
+ * read: var foo = __protected(this).foo;
+ * write: __protected(this).foo = "baz";
+ *
+ * @param {Object} object – given object
+ * @param {Object} [value] – given value
+ * @return {Object} protected property / context
+ */
+export default function __protected(object, value) {
```
orlandohohmeier: Wonder if it make sense to drop the `value` param to simplify the usage.

orlandohohmeier:
This is a great way of "flagging" properties as _protected_ without introducing any new function bind craziness.

packages/object-utilities/index.js
juliangieseke: Yes, it works as expected in the Connection Manager.

ahoskins:
Interesting stuff! A couple general thoughts/questions:

1. Do we have well defined use-cases or guidelines for when to use this utility? I know the JS world has **LOTS** of thoughts on this kind of thing, but curious how it'll fit into our codebase.
2. It's nicer to read/write with dot notation rather than like (read) `__protected(obj).val` and (write) `__protected(obj).val = "baz"`. The way it is now, it won't show up in the normal grep (`obj.`) for all accesses on the object. Would be nice to read/write a value like (read) `obj.__protected__.val` and (write) `obj.__protected__.val = "baz"`. But then we have the issue of setting up the `__protected__` field the first time.

packages/object-utilities/src/js/__protected.js
ahoskins: Very related to my comment.

juliangieseke:
**re 1**: For us it was important to keep it simple and set up some kind of definition how we handle protected stuff in the future. Using a constructor-local private context variable is good for stuff, that _really should be private_, but using it for everything which _shouldn't be public_ can get us into some trouble. For example binding getter/setter to a private context variable isn't possible (AFAIK), because it is private to the constructor and not private to the class. Also its hard to subclass such a class because you always have to check which function uses which context and stuff like that. Thats why we wanted to introduce a simple way of using `protected`

**re 2 / orlando**: Ditching the `value` parameter is perfectly fine for me, also only using the `__protected` function for initialisation in the constructor is (but then I'd rename it to something like `initProtectedContext(this)`).  
As stated in **re 1**, its more about a definition of using protected as for the exact implementation. 

_Moar Feedback please!_

juliangieseke:
update on 2. after removing the value parameter we have no longer the possibility to initialize it with default/init values because `__protected(obj) = {foo: "bar"};` isnt allowed.

packages/object-utilities/index.js
juliangieseke: npm run start works - so yes?! :)

weblancaster:
I want to understand why we need the protected function? I understand what is happening but why? I'm just asking because I don't see us using as much and this is only going to increase code complexity with that code quality and productivity goes down.

packages/object-utilities/src/js/__protected.js
weblancaster: the naming convention here is breaking the practices we see throughout the code base. I would name something more clear and without the `__`

packages/object-utilities/src/js/__protected.js
```diff
@@ -0,0 +1,44 @@
...
+ *     alert(__protected(this).foo);
+ *   }
+ * }
+ *
+ * @param {Object} object – given object
+ * @param {Object} [value] – given value
+ * @return {Object} protected property / context
+ */
+export default function __protected(object, value) {
+  if (object === undefined || object === null) {
```
weblancaster: we have been using `!= null` instead

packages/object-utilities/src/js/__protected.js
juliangieseke: yeah, the name isnt perfect, but sadly "protected" is a reserved word and we wanted to make sure it is a "system utility". if you have a better idea - please tell us :)

juliangieseke:
@weblancaster I was writing my Connection Manager Package "the OOP way", yes. But I'm not saying that we have to do it the OOP way – FP is perfectly fine. Nevertheless IMHO _if_ we're doing OOP stuff, then it should work as expected. Using `class {}` and then creating a local context inside the constructor felt wrong for me, so I was looking for a different approach. Which we're discussing here :) Its just another workaround if you want to see it that way.

packages/object-utilities/src/js/__protected.js
```diff
@@ -0,0 +1,44 @@
...
+      enumerable: false,
+      writeable: false,
+      value: {}
+    });
+  }
+  if (value !== undefined) {
+    if (typeof value === "object") {
+      Object.assign(object.__protected__, value);
+    } else {
+      throw new Error("value must be an object");
```
Poltergeist: A error message should be a complete sentence. And at best also provide some information for example what is the current value.

packages/object-utilities/src/js/__protected.js
```diff
@@ -0,0 +1,44 @@
...
+  }
+  if (object.__protected__ === undefined) {
+    Object.defineProperty(object, "__protected__", {
+      configurable: false,
+      enumerable: false,
+      writeable: false,
+      value: {}
+    });
+  }
+  if (value !== undefined) {
```
Poltergeist: The following block looks very complex, please either provide a comment or what do you think about this structure:

```JS
if(value !== undefined && typeof value === 'object') {
  Object.assign(object.__protected__, value);
}

if (value !== undefined && typeof value !== 'object') {
  throw new Error("value must be an object");
}

return object.__protected__;
```

or

```JS
if(value !== undefined) {
  if (typeof value === 'object') {
    Object.assign(object.__protected__, value);

    return object.__protected__;
  }
  throw new Error("value must be an object");
}

return object.__protected__;
```

packages/object-utilities/src/js/__protected.js
Poltergeist: To clarify this, a complete sentence starts with a Capital letter and ends with a full stop.

packages/object-utilities/src/js/__protected.js
```diff
@@ -0,0 +1,44 @@
...
+    if (typeof value === "object") {
```
MatApple: What happens when value is an `Array` or `null`?

MatApple: Think we need many more test cases.

MatApple:
In general, I like the clean interface to the private namespace.

packages/object-utilities/src/js/__protected.js
juliangieseke: As `value` in this case is our protected context, is has to be an object.

juliangieseke:
@MatApple `__private` would be misleading, because in ´class sub extends parent {}´ ´this´is the same in both classes. but a private context would be different in both of them. AFAIK there is no possibility to get private variables into javascript "classes".

and yes, technically there is no difference. but it tells us which variable shouldnt be changed directly from outside the class. in your example, the second one would be public (and ok) and the first one shouldnt be done this way. `__protected` should only be called from inside the class.
the correct way for your first example would be 
```
//main.js
import obj from "obj"
obj.foo = "bar";

//obj.js
import __protected from "object-utilities"
class obj {
...
set foo(val) { __protected(this).foo = val; }
...
}
```

packages/object-utilities/src/js/__protected.js
nLight: equal to `object == null`

packages/object-utilities/src/js/__protected.js
```diff
@@ -0,0 +1,44 @@
...
+ * @param {Object} [value] – given value
+ * @return {Object} protected property / context
+ */
+export default function __protected(object, value) {
+  if (object === undefined || object === null) {
+    throw new Error("no object given or null");
+  }
+  if (object.__protected__ === undefined) {
+    Object.defineProperty(object, "__protected__", {
+      configurable: false,
```
nLight: all the config options are `false` by default, so why don't we use `Object.defineProperty(this,"protectedVar", {value})` instead of the util?

packages/object-utilities/src/js/__protected.js
```diff
@@ -0,0 +1,44 @@
...
+  }
+  if (value !== undefined) {
+    if (typeof value === "object") {
+      Object.assign(object.__protected__, value);
+    } else {
+      throw new Error("value must be an object");
+    }
+  }
+
+  return object.__protected__;
```
nLight: at the end we just return the prop
`__protected(this).bla` vs `this.__protected__.bla`
former does a function call, a bit more work for Webpack to resolve the import and one more module to go through.

packages/object-utilities/src/js/__protected.js
juliangieseke: awesome, yeah!

packages/object-utilities/src/js/__protected.js
juliangieseke: yup, but I prefer `===` over `==`, as you know :)

juliangieseke:
As @nLight stated we can simply use `Object.defineProperty(this,"__protected__", {value})` instead of this util. Actually the only reason for this one was to shorten the code a bit, so if we dont have to shorten it, we dont need it.

then we only have agree that no one will access this `__protected__` variable (and its contents) from outside the class. 
and: which name would you like to use? You guys have a better insight into the codebase and as @weblancaster alrady mentioned my approach doesnt follow the current naming styles.

orlandohohmeier:
@juliangieseke thanks a lot for proposing this solution and starting a discussion! I'd close this PR and opt to  simply use `Object.defineProperty` for every single _protected_ property, after reading into the `defineProperty` spec. Every property defined is by default not configurable, enumerable, nor writable, thus there is very little overhead.

**Example**
```
class User {
  constructor(name){
   Object.defineProperty(this, "role", {value:"user"});
   Object.defineProperty(this, "tasks", {value: new Array()}); 
   Object.defineProperty(this, "username", {get(){return name}}); 
 }
}
```

This doesn't protect anyone from accessing these properties from the "outside" but it actually protects the values. What do you think?

nLight:
There's also `Object.defineProperties()` when you need to define many.
```
Object.defineProperties(this, {
  role: {value: "user"},
  tasks: {value: []},
  username: {...}
});
```

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperties

PR #2357, opened 2017-08-08
DCOS-16280: include forcePull parameter in image object for multi-pod containers

bstavroulakis:
Include the forcePull parameter in the image object for the multi-container service.
You can find this option over here => https://mesosphere.github.io/marathon/docs/pods.html#containerizers

Closes DCOS-16280

To reproduce:

1) Open the multi-container create form
2) Once you open the container's options you can see a new option named "Force pull image on launch"

multi-container-forcepull

3) There are some interesting cases you can check
     a) If the forcePull = false and the containerImageName == "" then the image object should be           
         deleted
     b) If the forcePull = true and the containerImageName == "" then the image object doesn't show in the JSON Editor but still exists, so if you set something at the container image name it should have forcePull = true.
     c) If you first create a container and add only the container image name but don't specifically set the state of the forcePull it doesn't exist which is ok. If you deliberately set it to true or false then it should keep the value you set.

image

**Checklist**

### DIFF
--- a/plugins/services/src/js/components/forms/MultiContainerFormAdvancedSection.js
+++ b/plugins/services/src/js/components/forms/MultiContainerFormAdvancedSection.js
@@ -1,0 +1,52 @@
+import React from "react";
+
+import { findNestedPropertyInObject } from "#SRC/js/utils/Util";
+import FieldHelp from "#SRC/js/components/form/FieldHelp";
+import FieldInput from "#SRC/js/components/form/FieldInput";
+import FieldLabel from "#SRC/js/components/form/FieldLabel";
+
+const getForcePullSection = function(data, path) {
+  const imageForcePullPath = `${path}.image.forcePull`;
+  const imageForcePull = findNestedPropertyInObject(data, imageForcePullPath);
+
+  return (
+    <FieldLabel>
+      <FieldInput
+        checked={imageForcePull}
+        name={imageForcePullPath}
+        type="checkbox"
+        value={imageForcePull}
+      />
+      Force Pull Image On Launch
+      <FieldHelp>
+        Force Docker to pull the image before launching each instance.
+      </FieldHelp>
+    </FieldLabel>
+  );
+};
+
+const MultiContainerFormAdvancedSection = ({ data, path }) => {
+  return (
+    <div>
+      <h3 className="short-top short-bottom">
+        Advanced Settings
+      </h3>
+      <p>Advanced settings of the container.</p>
+      <p>
+        {getForcePullSection(data, path)}
+      </p>
+    </div>
+  );
+};
+
+MultiContainerFormAdvancedSection.defaultProps = {
+  data: {},
+  path: "container"
+};
+
+MultiContainerFormAdvancedSection.propTypes = {
+  data: React.PropTypes.object,
+  path: React.PropTypes.string
+};
+
+module.exports = MultiContainerFormAdvancedSection;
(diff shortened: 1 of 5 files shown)


plugins/services/src/js/components/forms/MultiContainerFormAdvancedSection.js
```diff
@@ -0,0 +1,60 @@
+import React, { Component } from "react";
+
+import { findNestedPropertyInObject } from "#SRC/js/utils/Util";
+import FieldHelp from "#SRC/js/components/form/FieldHelp";
+import FieldInput from "#SRC/js/components/form/FieldInput";
+import FieldLabel from "#SRC/js/components/form/FieldLabel";
+
+class MultiContainerFormAdvancedSection extends Component {
```
ahoskins: No component state, so can just do a pure function component like [ConfigurationMapEditAction](https://github.com/dcos/dcos-ui/blob/master/plugins/services/src/js/components/ConfigurationMapEditAction.js).

plugins/services/src/js/reducers/serviceForm/JSONReducers/Containers.js
```diff
@@ -287,6 +295,17 @@ function containersParser(state) {
   }, []);
 }
 
+function deleteContainerImageCheck(image) {
+  if (image == null) {
+    return true;
+  }
+  if (!image.id && !image.forcePull) {
```
weblancaster: What do you think about combining this condition with line 299?

plugins/services/src/js/reducers/serviceForm/JSONReducers/Containers.js
```diff
@@ -287,6 +295,17 @@ function containersParser(state) {
   }, []);
 }
 
+function deleteContainerImageCheck(image) {
```
ahoskins: I think this is awkwardly named. Maybe `shouldDeleteContainerImage` will read better.

ahoskins: Also, this does behave different than the forcePull for single container services. Using just `if (!image.id) return true` would make it work that same!

plugins/services/src/js/components/forms/MultiContainerFormAdvancedSection.js
```diff
@@ -0,0 +1,58 @@
...
+  );
+};
+
+const MultiContainerFormAdvancedSection = ({ data, path }) => {
+  return (
+    <div>
+      <h3 className="short-top short-bottom">
+        Advanced Settings
+      </h3>
+      <p>Advanced settings of the container</p>
```
bstavroulakis: Added!

plugins/services/src/js/components/forms/MultiContainerFormAdvancedSection.js
```diff
@@ -0,0 +1,58 @@
...
+        {getForcePullSection(data, path)}
+      </p>
+    </div>
+  );
+};
+
+MultiContainerFormAdvancedSection.defaultProps = {
+  data: {},
+  errors: {},
+  onAddItem() {},
```
bstavroulakis: Removed. I added those for consistency with the other components. But we can remove them for now.

plugins/services/src/js/reducers/serviceForm/JSONReducers/Containers.js
bstavroulakis: Renamed and yes (image == null || !image.id) now makes it work exactly like the single container

plugins/services/src/js/reducers/serviceForm/JSONReducers/Containers.js
```diff
@@ -287,6 +295,14 @@ function containersParser(state) {
   }, []);
 }
 
+function shouldDeleteContainerImage(image) {
+  if (image == null || !image.id) {
```
bstavroulakis: Cool catch

bstavroulakis:
run integration tests

PR #2358, opened 2017-08-08
Adjust external volume test teardown

orlandohohmeier:
Adjust the external volume system test to avoid failures due to timeouts and unwanted side effects.

1. Remove the test `teardown` and all related scripts as the external volume clean up is an infrastructure related task that shouldn't be part of the respective test definition. 

2. Change the external volume name to match the pattern defined by the system test janitor so that they're properly removed after running the tests.


MatApple:
💯

PR #2361, opened 2017-08-08
Connection Manager

juliangieseke:
~This branch depends on changes that will be introduced with #2403~

---
The Connection Manager will allow us to manage our requests in the future. We will be able to queue them up with different priorities and even do stuff like repriorizing and more.

---
@nLight change log:

1. Export additional modules from the package
2. Split EventTypes and StateConstants
3. Removed prefixes from EventTypes and StateConstants
4. Changed EventTypes to be strings for compatibility
5. Changed `XHRConnection` interface
6. Deleted WebSocketConnection for now
7. Added some tests here and there
8. Various cleanups
9. Packagefied the package: package.json, license, readme
10. Changed the package to export `ConnectionManager` as a default export

### DIFF
--- a/packages/connection-manager/README.md
+++ b/packages/connection-manager/README.md
@@ -1,0 +1,39 @@
+# Connection Manager
+
+This package allows to manage connections inside the browser and order them in a queue by priority.
+
+Browsers can open only a limited amount of connections so when you urgently need one there may not be a slot. This package solves the issue by providing a manager that enqueues all requested connections and priorities them accordingly.
+
+## Usage
+
+```javascript
+import ConnectionManager, { XHRConnection } from "connection-manager";
+
+const connection = new XHRConnection("http://127.0.0.1");
+connection.on("close", function() {
+  // doing something with connection.response
+});
+ConnectionManager.queue(connection);
+```
+
+Connection will be opened by the manager once there's a slot available.
+
+## Connection types
+
+At the moment `connection-manager` package exposes only one connection `XHRConnection` that wraps `XmlHttpRequest`
+
+```javascript
+const connection = new XHRConnection("http://127.0.0.1", {
+  method: "POST",
+  body: "{}",
+  headers: {
+    "Content-Type": "application/json"
+  }
+});
+```
+
+## Creating your own connection type
+
+If you want to create your own connection type that will be compatible with the manager please extend `AbstractConnection`.
+Have a look at `Connections/AbstractConnection.js` for more info.
+
--- a/packages/connection-manager/index.js
+++ b/packages/connection-manager/index.js
@@ -1,0 +1,12 @@
+import {
+  default as ConnectionManagerClass
+} from "./src/js/ConnectionManager/ConnectionManager";
+import XHRConnection from "./src/js/Connection/XHRConnection";
+import AbstractConnection from "./src/js/Connection/AbstractConnection";
+import * as EventTypes from "./src/js/Connection/EventTypes";
+import * as StateConstants from "./src/js/Connection/StateConstants";
+
+const ConnectionManager = new ConnectionManagerClass();
+
+export default ConnectionManager;
+export { AbstractConnection, EventTypes, StateConstants, XHRConnection };
--- a/packages/connection-manager/src/js/Connection/AbstractConnection.js
+++ b/packages/connection-manager/src/js/Connection/AbstractConnection.js
@@ -1,0 +1,91 @@
+import { EventEmitter } from "events";
+// import ConnectionEvent from "../ConnectionEvent/ConnectionEvent";
+
+/**
+ * AbstractConnection provides some default properties, methods, states and event-definitions used by Connection Manager
+ *
+ * Events which MUST be fired by all SubClasses:
+ * OPEN: when the connection actually blocks a pipe
+ * DATA: everytime data is recieved from the server
+ * CLOSE: when the connection frees its pipe
+ * ERROR: when an error occurs
+ */
+export default class AbstractConnection extends EventEmitter {
+  constructor(url) {
+    super();
+
+    if (this.constructor === AbstractConnection) {
+      throw new Error("Can't instantiate abstract class!");
+    }
+
+    if (!url) {
+      throw new Error("Can't instantiate without given URL!");
+    }
+    Object.defineProperty(this, "url", {
+      get: () => url
+    });
+
+    const created = Date.now();
+    Object.defineProperty(this, "created", {
+      get: () => created
+    });
+
+    let state = AbstractConnection.INIT;
+    Object.defineProperty(this, "state", {
+      get: () => state,
+      set: _state => {
+        if (
+          ![
+            AbstractConnection.INIT,
+            AbstractConnection.STARTED,
+            AbstractConnection.DONE,
+            AbstractConnection.CANCELLED
+          ].includes(_state)
+        ) {
+          throw new Error("Cant set Connection into unknown state.");
+        }
+        state = _state;
+      }
+    });
+
+    const symbol = Symbol("Connection:" + this.url + this.created);
+    Object.defineProperty(this, "symbol", {
+      get: () => symbol
+    });
+  }
+  static get INIT() {
+    return Symbol.for("DCOS.ConnectionManager.AbstractConnection.INIT");
+  }
+  static get STARTED() {
+    return Symbol.for("DCOS.ConnectionManager.AbstractConnection.STARTED");
+  }
+  static get DATA() {
+    return Symbol.for("DCOS.ConnectionManager.AbstractConnection.DATA");
+  }
+  static get DONE() {
+    return Symbol.for("DCOS.ConnectionManager.AbstractConnection.DONE");
+  }
+  static get CANCELLED() {
+    return Symbol.for("DCOS.ConnectionManager.AbstractConnection.CANCELLED");
+  }
+
+  // Abstract Methods
+  /* eslint-disable no-unused-vars */
+  /**
+   * Opens the connection
+   * This method should only be called by Connection Manager, if you want to call it by yourself be sure you know what you're doing.
+   * Also this method should be the point where a Connection actually starts using a "Connection Pipe" of the browser.
+   * @param {string} token – Authentication token
+   */
+  open(token) {}
+  /**
+   * Closes the connection
+   * This method should free the "Connection Pipe" of the browser and do everything needed to tear down the connection
+   */
+  close() {}
+  /**
+   * Resets the connection
+   * This method should reset the Connection in a way that open() can be called again.
+   */
+  reset() {}
+}
--- a/packages/connection-manager/src/js/Connection/XHRConnection.js
+++ b/packages/connection-manager/src/js/Connection/XHRConnection.js
@@ -1,0 +1,229 @@
+import AbstractConnection from "./AbstractConnection";
+import ConnectionEvent from "../ConnectionEvent/ConnectionEvent";
+
+export const ALLOWED_METHODS = ["GET", "POST", "PUT", "DELETE"];
+
+/**
+ * Basic XHR Connection
+ * @todo add/remove headers
+ */
+export default class XHRConnection extends AbstractConnection {
+  /**
+   * Initializes an Instance of XHRConnection
+   * @constructor
+   * @param {string} url – URL to be fetched
+   * @param {object} [options]
+   * @param {string} [options.method=GET] – used method
+   * @param {*} [options.body] – payload for request
+   * @param {string} [options.headers] – additional headers (like content-type)
+   */
+  constructor(url, options = {}) {
+    super(url);
+
+    const {
+      method = "GET",
+      body = null,
+      headers = null,
+      responseType = "json"
+    } = options;
+
+    let _method;
+    Object.defineProperty(this, "method", {
+      get: () => _method,
+      set: __method => {
+        if (ALLOWED_METHODS.includes(__method) === -1) {
+          throw new Error("Invalid method for XHR Request.");
+        }
+        _method = __method;
+      }
+    });
+    this.method = method;
+
+    let _body;
+    Object.defineProperty(this, "body", {
+      get: () => _body,
+      set: __body => {
+        _body = __body;
+      }
+    });
+    this.body = body;
+
+    let _headers = {};
+    Object.defineProperty(this, "headers", {
+      get: () => _headers,
+      set: __headers => {
+        _headers = Object.assign(_headers, __headers);
+      }
+    });
+    this.headers = headers;
+
+    let _responseType = {};
+    Object.defineProperty(this, "responseType", {
+      get: () => _responseType,
+      set: __responseType => {
+        _responseType = __responseType;
+      }
+    });
+    this.responseType = responseType;
+
+    let xhr = null;
+    Object.defineProperty(this, "xhr", {
+      get: () => xhr,
+      set: _xhr => {
+        if (_xhr === null) {
+          throw new Error(
+            "Given XMLHttpRequest has to be an instance of XMLHttpRequest"
+          );
+        }
+        if (
+          this.state !== XHRConnection.INIT &&
+          this.state !== XHRConnection.CANCELLED
+        ) {
+          xhr.abort();
+          this.emit(ConnectionEvent.ERROR);
+        }
+        xhr = _xhr;
+        this.state = XHRConnection.INIT;
+      }
+    });
+  }
+
+  /**
+   * more xhr related getter
+   * @todo make them more robust / do we need more?
+   */
+  get response() {
+    return this.xhr && this.xhr.response;
+  }
+  get readyState() {
+    return this.xhr && this.xhr.readyState;
+  }
+  get status() {
+    return this.xhr && this.xhr.status;
+  }
+
+  /**
+   * create, prepare, open and send the xhr request
+   * @param {string} [token] – authentication token
+   */
+  open(token) {
+    super.open(token);
+
+    /**
+     * handle open  of native xhr
+     */
+    const handleProgress = function() {
+      this.emit(
+        ConnectionEvent.DATA,
+        new ConnectionEvent(this, ConnectionEvent.DATA)
+      );
+    }.bind(this);
+    /**
+     * handle abort  of native xhr
+     */
+    const handleAbort = function() {
+      this.state = XHRConnection.DONE;
+      this.emit(
+        ConnectionEvent.ABORT,
+        new ConnectionEvent(this, ConnectionEvent.ABORT)
+      );
+      this.emit(
+        ConnectionEvent.CLOSE,
+        new ConnectionEvent(this, ConnectionEvent.CLOSE)
+      );
+    }.bind(this);
+    /**
+     * handle error  of native xhr
+     */
+    const handleError = function() {
+      this.state = XHRConnection.DONE;
+      this.emit(
+        ConnectionEvent.ERROR,
+        new ConnectionEvent(this, ConnectionEvent.ERROR)
+      );
+      this.emit(
+        ConnectionEvent.CLOSE,
+        new ConnectionEvent(this, ConnectionEvent.CLOSE)
+      );
+    }.bind(this);
+    /**
+     * handle load of native xhr
+     */
+    const handleLoad = function() {
+      this.state = XHRConnection.DONE;
+      if (this.status >= 400) {
+        this.emit(
+          ConnectionEvent.ERROR,
+          new ConnectionEvent(this, ConnectionEvent.ERROR)
+        );
+      } else {
+        this.emit(
+          ConnectionEvent.SUCCESS,
+          new ConnectionEvent(this, ConnectionEvent.SUCCESS)
+        );
+      }
+      this.emit(
+        ConnectionEvent.CLOSE,
+        new ConnectionEvent(this, ConnectionEvent.CLOSE)
+      );
+    }.bind(this);
+    /**
+     * handle timeout of native xhr
+     */
+    const handleTimeout = function() {
+      handleError();
+    };
+
+    if (this.xhr !== null) {
+      throw new Error("cannot open XHR Connection a second time!");
+    }
+    this.xhr = new XMLHttpRequest();
+
+    this.xhr.addEventListener("progress", handleProgress);
+    this.xhr.addEventListener("abort", handleAbort);
+    this.xhr.addEventListener("error", handleError);
+    this.xhr.addEventListener("load", handleLoad);
+    this.xhr.addEventListener("timeout", handleTimeout);
+    this.xhr.open(this.method, this.url);
+
+    this.state = XHRConnection.STARTED;
+    this.emit(
+      ConnectionEvent.OPEN,
+      new ConnectionEvent(this, ConnectionEvent.OPEN)
+    );
+
+    Object.keys(this.headers).forEach(key => {
+      if (this.headers[key] !== undefined && this.headers[key] !== null) {
+        this.xhr.setRequestHeader(key, this.headers[key]);
+      }
+    });
+
+    if (token !== undefined && token !== "") {
+      this.xhr.setRequestHeader("Authorization", "Bearer " + token);
+    }
+
+    this.xhr.responseType = this.responseType;
+    this.xhr.send(this.body);
+  }
+  /**
+   * aborts native xhr
+   */
+  close() {
+    super.close();
+
+    if (this.xhr === null || this.state === XHRConnection.INIT) {
+      throw new Error("XMLHttpRequest not open, cant close it.");
+    }
+    this.xhr.abort();
+  }
+  /**
+   *
+   * @todo abort here before delete?
+   */
+  reset() {
+    super.reset();
+
+    this.state = XHRConnection.CANCELLED;
+    this.xhr = null;
+  }
+}
--- a/packages/connection-manager/src/js/Connection/__tests__/XHRConnection-test.js
+++ b/packages/connection-manager/src/js/Connection/__tests__/XHRConnection-test.js
@@ -1,0 +1,266 @@
+import ConnectionEvent from "../../ConnectionEvent/ConnectionEvent";
+import XHRConnection from "../XHRConnection";
+
+const token = "Token";
+const url = "http://example.com/url.json";
+const originalXhr = global.XMLHttpRequest;
+
+describe("XHRConnection", () => {
+  beforeEach(function() {
+    const xhr = function() {
+      this.events = {
+        progress: [],
+        abort: [],
+        error: [],
+        load: [],
+        timeout: []
+      };
+      this.open = jest.fn();
+      this.setRequestHeader = jest.fn();
+      this.send = jest.fn();
+      this.abort = jest.fn();
+      this.addEventListener = (type, callback) => {
+        this.events[type].push(callback);
+      };
+      this.__emit = type => {
+        this.events[type].forEach(callback => {
+          callback();
+        });
+      };
+    };
+
+    global.XMLHttpRequest = jest.fn(() => new xhr());
+  });
+
+  afterEach(function() {
+    global.XMLHttpRequest = originalXhr;
+  });
+
+  describe("#initialization", () => {
+    it("initializes with url", () => {
+      expect(() => new XHRConnection(url)).not.toThrow();
+    });
+    it("initializes with url and method", function() {
+      expect(() => {
+        return new XHRConnection(url, { method: "POST" });
+      }).not.toThrowError();
+    });
+    it("initializes with url, method and body", function() {
+      expect(() => {
+        return new XHRConnection(
+          url,
+          "POST",
+          JSON.stringify({ foo: "bar", baz: 0 })
+        );
+      }).not.toThrowError();
+    });
+    it("initializes with url, method, body and headers", function() {
+      expect(() => {
+        return new XHRConnection(url, {
+          method: "POST",
+          body: JSON.stringify({ foo: "bar", baz: 0 }),
+          headers: { "content-type": "text/json" }
+        });
+      }).not.toThrowError();
+    });
+    it("fails to initialize without url", () => {
+      expect(() => new XHRConnection()).toThrow();
+    });
+  });
+  describe("#open", () => {
+    it("opens without token", function() {
+      expect(() => {
+        return new XHRConnection(url).open();
+      }).not.toThrowError();
+    });
+    it("opens connection with token", function() {
+      expect(() => {
+        return new XHRConnection(url).open(token);
+      }).not.toThrowError();
+    });
+    it("opens correctly with given token", function() {
+      const req = new XHRConnection(url);
+      req.open(token);
+      expect(req.xhr.setRequestHeader.mock.calls).toContainEqual([
+        "Authorization",
+        "Bearer " + token
+      ]);
+    });
+    it("opens connection with correct url", function() {
+      const req = new XHRConnection(url);
+      req.open(token);
+      expect(req.xhr.open.mock.calls).toEqual([["GET", url]]);
+    });
+    ["GET", "POST", "DELETE", "PUT"].forEach(method => {
+      it(`opens connection with given method (${method})`, function() {
+        const req = new XHRConnection(url, { method });
+        req.open(token);
+        expect(req.xhr.open.mock.calls).toEqual([[method, url]]);
+      });
+    });
+    it("opens connection with correct body", () => {
+      const body = { foo: "bar" };
+      const req = new XHRConnection(url, { method: "POST", body });
+      req.open();
+      expect(req.xhr.send.mock.calls).toEqual([[body]]);
+    });
+    it("opens connection with correct header", () => {
+      const headers = { foo: "bar" };
+      const req = new XHRConnection(url, {
+        method: "POST",
+        body: null,
+        headers
+      });
+      req.open();
+      expect(req.xhr.setRequestHeader.mock.calls).toEqual([["foo", "bar"]]);
+    });
+  });
+  describe("state", () => {
+    it("has correct state on init", () => {
+      const req = new XHRConnection(url);
+
+      expect(req.state).toEqual(XHRConnection.INIT);
+    });
+    it("sets correct state on open", () => {
+      const req = new XHRConnection(url);
+
+      req.open(token);
+
+      expect(req.state).toEqual(XHRConnection.STARTED);
+    });
+
+    it("sets correct state on abort", () => {
+      const req = new XHRConnection(url);
+
+      req.open(token);
+      req.xhr.__emit("abort");
+
+      expect(req.state).toEqual(XHRConnection.DONE);
+    });
+
+    it("sets correct state on error", () => {
+      const req = new XHRConnection(url);
+      const cb = jest.fn();
+
+      req.open(token);
+      req.addListener(ConnectionEvent.ERROR, cb);
+      req.xhr.__emit("error");
+
+      expect(req.state).toEqual(XHRConnection.DONE);
+    });
+
+    it("sets correct state on load", () => {
+      const req = new XHRConnection(url);
+
+      req.open(token);
+      req.xhr.status = 200;
+      req.xhr.__emit("load");
+
+      expect(req.state).toEqual(XHRConnection.DONE);
+    });
+
+    it("sets correct state on timeout", () => {
+      const req = new XHRConnection(url);
+      const cb = jest.fn();
+
+      req.open(token);
+      req.addListener(ConnectionEvent.ERROR, cb);
+      req.xhr.status = 400;
+      req.xhr.__emit("error");
+
+      expect(req.state).toEqual(XHRConnection.DONE);
+    });
+  });
+  describe("events", () => {
+    it("emits open event on open", () => {
+      const req = new XHRConnection(url);
+      const cb = jest.fn();
+
+      req.on(ConnectionEvent.OPEN, cb);
+      req.open(token);
+
+      expect(cb).toHaveBeenCalled();
+    });
+
+    it("emits abort event on abort", () => {
+      const req = new XHRConnection(url);
+      const cb = jest.fn();
+
+      req.open(token);
+      req.addListener(ConnectionEvent.ABORT, cb);
+      req.xhr.__emit("abort");
+
+      expect(cb).toHaveBeenCalled();
+    });
+
+    it("emits close event on abort", () => {
+      const req = new XHRConnection(url);
+      const cb = jest.fn();
+
+      req.open(token);
+      req.addListener(ConnectionEvent.CLOSE, cb);
+      req.xhr.__emit("abort");
+
+      expect(cb).toHaveBeenCalled();
+    });
+
+    it("emits error event on error", () => {
+      const req = new XHRConnection(url);
+      const cb = jest.fn();
+
+      req.open(token);
+      req.addListener(ConnectionEvent.ERROR, cb);
+      req.xhr.__emit("error");
+
+      expect(cb).toHaveBeenCalled();
+    });
+
+    it("emits close event on error", () => {
+      const req = new XHRConnection(url);
+      const cb = jest.fn();
+
+      req.open(token);
+      req.addListener(ConnectionEvent.CLOSE, cb);
+      req.addListener(ConnectionEvent.ERROR, () => {});
+      req.xhr.__emit("error");
+
+      expect(cb).toHaveBeenCalled();
+    });
+
+    it("emits error event on load (error)", () => {
+      const req = new XHRConnection(url);
+      const cb = jest.fn();
+
+      req.open(token);
+      req.addListener(ConnectionEvent.ERROR, cb);
+      req.xhr.status = 400;
+      req.xhr.__emit("load");
+
+      expect(cb).toHaveBeenCalled();
+    });
+
+    it("emits success event on load (success)", () => {
+      const req = new XHRConnection(url);
+      const cb = jest.fn();
+
+      req.open(token);
+      req.addListener(ConnectionEvent.SUCCESS, cb);
+      req.xhr.status = 200;
+      req.xhr.__emit("load");
+
+      expect(cb).toHaveBeenCalled();
+    });
+
+    it("emits close event on load (always)", () => {
+      const req = new XHRConnection(url);
+      const cb = jest.fn();
+
+      req.open(token);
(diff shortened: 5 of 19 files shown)


packages/object-utilities/src/js/__protected.js
```diff
@@ -0,0 +1,44 @@
+/**
```
weblancaster: I want to understand why we need the `protected` function? what is the problem we are trying to solve here? I understand what is happening but why?

MatApple:
@juliangieseke I know this is a WIP so I apologize for the "requested changes". Please treat them as general comments :) Just providing feedback.

juliangieseke:
perfectly fine - the more feedback, the better :) 
no reaction = you're right, i'll change it.

packages/authenticator/src/js/AuthStore/AuthStore.js
```diff
@@ -0,0 +1,9 @@
+/*
```
juliangieseke: when requiring (and importing) files, `AuthStore`, `AuthStore.js` and `authstore` aren't recognized as the same file, so when importing a singleton this way, you'll have different instances. this file is more or less an experiment if proxying will help (it does). its not needed because we do the same thing in `index.js`

packages/connection-manager/src/js/ConnectionManager/ConnectionManager.js
```diff
@@ -0,0 +1,82 @@
+import { AuthStore } from "authenticator";
+import ConnectionStore from "../ConnectionStore/ConnectionStore.js";
+
+/**
+ * The Connection Manager which is responsible for
+ * queueing Connections into the ConnectionStore and
```
orlandohohmeier: 🔤  queuing

packages/connection-manager/src/js/ConnectionManager/__tests__/ConnectionManager-test.js
```diff
@@ -0,0 +1,53 @@
...
+  beforeEach(() => {
+    ConnectionManager = new ConnectionManagerClass();
+  });
+  it("add listeners to new connection", () => {
+    var connection = new ConnectionTemplate();
+
+    ConnectionManager.queue(connection);
+    expect(connection.on).toBeCalled();
+  });
+  it("dont add listeners second time", () => {
```
orlandohohmeier: 🔤  don't

packages/connection-manager/src/js/ConnectionStore/ConnectionStore.js
```diff
@@ -0,0 +1,146 @@
+/**
+ * Connection Queue which holds queued connections.
+ *
+ * @todo check if optimisation of storage is needed (performance tests needed)
```
orlandohohmeier: 🇺🇸  optimizations

packages/connection-manager/src/js/ConnectionStore/ConnectionStore.js
```diff
@@ -0,0 +1,146 @@
...
+   */
+  delete(connection) {
+    var index = this.indexOf(connection), rconnection;
+    if (index === -1) {
+      throw new Error("Connection is not stored.");
+    }
+    rconnection = this.store.splice(index, 1)[0].connection;
+    if (rconnection !== connection) {
+      throw new Error(
+        "To-Be-Removed and actually removed Connection arent the same."
```
orlandohohmeier: 🔤  aren't

packages/connection-manager/src/js/Connection/XHRConnection.js
```diff
@@ -0,0 +1,149 @@
+import AbstractConnection from "./AbstractConnection";
+
+/**
+ * Basic XHR Connection
+ */
+export default class XHRConnection extends AbstractConnection {
+  /**
+   * Initialises an Instance of XHRConnection
```
orlandohohmeier: :us: Initializes

packages/connection-manager/src/js/Connection/__tests__/XHRConnection-test.js
```diff
@@ -0,0 +1,264 @@
...
+    it("uses contentType=null as default (does not set header)", function() {
+      const req = new XHRConnection(url);
+      req.open(token);
+      expect(
+        req.__protected__.native.setRequestHeader.mock.calls
+      ).not.toContainEqual(["Content-Type", null]);
+    });
+  });
+
+  describe("initialisation", () => {
```
orlandohohmeier: 🇺🇸  initialization

packages/connection-manager/src/js/ConnectionStore/ConnectionStore.js
```diff
@@ -0,0 +1,146 @@
+/**
+ * Connection Queue which holds queued connections.
+ *
+ * @todo check if optimisation of storage is needed (performance tests needed)
+ */
+export default class ConnectionStore {
+  /**
+   * Initialises an instance of ConnectionStore
```
orlandohohmeier: :us: Initializes

packages/connection-manager/src/js/ConnectionManager/ConnectionManager.js
```diff
@@ -0,0 +1,82 @@
...
+ * actually starting them, when they are head of
+ * waiting line.
+ */
+export default class ConnectionManager {
+  /**
+   * Initialises an Instance of CM
```
orlandohohmeier: :us: Initializes

orlandohohmeier: 🎨  [...] of ConnectionManager

packages/connection-manager/src/js/ConnectionManager/ConnectionManager.js
```diff
@@ -0,0 +1,82 @@
...
+   * @param {Integer} maxConnections – max open connections
+   */
+  constructor(maxConnections = 6) {
+    var store = new ConnectionStore();
+
+    Object.defineProperty(this, "maxConnections", {
+      get: () => maxConnections,
```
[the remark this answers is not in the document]
juliangieseke: hm, eslint says I should write it like this.

packages/connection-manager/src/js/ConnectionManager/ConnectionManager.js
```diff
@@ -0,0 +1,82 @@
...
+      set: mc => {
+        maxConnections = Math.max(0, Math.min(20, mc));
+      }
+    });
+
+    Object.defineProperty(this, "store", {
+      get: () => store
```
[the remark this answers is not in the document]
juliangieseke: its basically `store` vs `connectionStore` vs `connections` - for me all three would work, `store` was the shortest.

packages/connection-manager/src/js/ConnectionManager/ConnectionManager.js
```diff
@@ -0,0 +1,82 @@
...
+   * Queues given connection with given priority
+   * @param {AbstractConnection} connection – connection to queue
+   * @param {Integer} [priority] – optional change of priority, will be updated inside connection
+   */
+  queue(connection, priority = 0) {
+    if (
+      connection.state < 2 &&
+      connection
+        .listeners("close")
+        .indexOf(this.handleConnectionClosingEvents) === -1
```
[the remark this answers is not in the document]
juliangieseke: if someone adds a connection a second time we dont want to add the listener again and as the ConnectionManager doesnt have different functions for adding and updating, we need to check it.

packages/connection-manager/src/js/ConnectionManager/ConnectionManager.js
```diff
@@ -0,0 +1,82 @@
...
+    ) {
+      connection.on("close", this.handleConnectionClosingEvents);
+    }
+    if (!this.store.includes(connection)) {
```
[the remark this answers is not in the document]
juliangieseke: Had it like that before (only one `add` method for everything), but I find `add` not the best name for a an `addingOrUpdating` Method. 
IMHO its okay to have one `ConnectionManager.queue`, but on the store I'd split them up into different functions. 

_To be discussed._

packages/connection-manager/src/js/ConnectionStore/ConnectionStore.js
```diff
@@ -0,0 +1,146 @@
...
+    Object.defineProperty(this, "store", {
+      get: () => store
+    });
+  }
+  /**
+   * Adds given connection with given (or default) priority to queue
+   * @param {AbstractConnection} connection – Connection to be added to queue
+   * @param {Integer} priority – priority of connection
+   */
+  add(connection, priority) {
```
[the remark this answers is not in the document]
juliangieseke: We don't, its removed from connection because they dont need to know what priority they have.

if we add it to connection, we also need to add some kind of listener if the priority of a connection changes, because then we need to reorder the store. but we could do that if we are adding events to the store anyways..

packages/connection-manager/src/js/ConnectionStore/ConnectionStore.js
```diff
@@ -0,0 +1,146 @@
+/**
+ * Connection Queue which holds queued connections.
+ *
+ * @todo check if optimisation of storage is needed (performance tests needed)
+ */
+export default class ConnectionStore {
```
nLight: So is it a `Connection Queue` or a `ConnectionStore`?

packages/connection-manager/src/js/ConnectionStore/ConnectionStore.js
```diff
@@ -0,0 +1,146 @@
...
+    if (rconnection !== connection) {
+      throw new Error(
+        "To-Be-Removed and actually removed Connection arent the same."
+      );
+    }
+
+    return true;
+  }
+
+  waitingCount() {
```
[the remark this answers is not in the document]
nLight: Could be

packages/connection-manager/src/js/ConnectionStore/__tests__/ConnectionStore-test.js
```diff
@@ -0,0 +1,140 @@
...
+  priority,
+  connection
+});
+
+describe("ConnectionStore", () => {
+  beforeEach(() => {
+    ConnectionStore = new ConnectionStoreClass();
+  });
+  describe("waitingCount", () => {
+    it("no waiting after init", () => {
```
nLight: doesn't have

packages/connection-manager/src/js/ConnectionStore/__tests__/ConnectionStore-test.js
```diff
@@ -0,0 +1,140 @@
...
+      expect(ConnectionStore.waitingCount()).toEqual(0);
+    });
+
+    it("has waiting after added waiting", () => {
+      ConnectionStore.store.push(new StoreItemTemplate());
+      expect(ConnectionStore.waitingCount()).toEqual(1);
+    });
+
+    it("no waiting after added open", () => {
```
nLight: doesn't have

packages/connection-manager/src/js/ConnectionStore/__tests__/ConnectionStore-test.js
```diff
@@ -0,0 +1,140 @@
...
+      ConnectionStore.store.push(
+        new StoreItemTemplate(0, new ConnectionTemplate({ state: 1 }))
+      );
+      expect(ConnectionStore.waitingCount()).toEqual(0);
+    });
+  });
+  describe("openCount", () => {
+    it("no open after init", () => {
```
nLight: doesn't have

packages/connection-manager/src/js/ConnectionStore/__tests__/ConnectionStore-test.js
```diff
@@ -0,0 +1,140 @@
...
+    });
+
+    it("has open after added open", () => {
+      ConnectionStore.store.push(
+        new StoreItemTemplate(0, new ConnectionTemplate({ state: 1 }))
+      );
+      expect(ConnectionStore.openCount()).toEqual(1);
+    });
+
+    it("no open after added waiting", () => {
```
nLight: doesn't have

packages/connection-manager/src/js/ConnectionStore/__tests__/ConnectionStore-test.js
```diff
@@ -0,0 +1,140 @@
...
+    });
+  });
+  describe("delete", () => {
+    it("deletes connection", () => {
+      const connection = new ConnectionTemplate();
+
+      ConnectionStore.store.push(new StoreItemTemplate(0, connection));
+      expect(ConnectionStore.delete(connection)).toEqual(true);
+    });
+    it("cant delete unknown connection", () => {
```
nLight: deletes

juliangieseke:
I've rebuilt Store & Manager and removed Authenticator Stuff. 
Still have to update tests.

juliangieseke:
cleaned up, made tests pass, changed a few things.

tests are still kind of WIP

orlandohohmeier:

packages/connection-manager/src/js/Connection/XHRConnection.js
```diff
@@ -0,0 +1,198 @@
...
+   */
+  handleOpen() {
+    this.state = CONNECTION_STATE_STARTED;
+    this.emit(CONNECTION_EVENT_OPEN, this);
+  }
+  /**
+   * handle open  of native xhr
+   */
+  handleProgress() {
+    this.emit(CONNECTION_EVENT_PROGRESS, this);
```
nLight: I tried to use it and got a question: why do we want to pass `this` to an event listener? Usually `ProgressEvent` goes there. Had to modify it like this to make it work:
```
handleProgress(event) {
  this.emit(CONNECTION_EVENT_PROGRESS, event);
}
```

packages/connection-manager/index.js
```diff
@@ -0,0 +1,5 @@
+import {
+  default as ConnectionManagerClass
+} from "./src/js/ConnectionManager/ConnectionManager";
+
+export const ConnectionManager = new ConnectionManagerClass();
```
nLight: I think the package should export a lot more. Otherwise how'd I create a connection? Had to export additional stuff to make it working with the `mesos-client` package I'm working on.
```
import {
  default as ConnectionManagerClass
} from "./src/js/ConnectionManager/ConnectionManager";
import XHRConnection from "./src/js/Connection/XHRConnection";
import * as ConnectionConstants from "./src/js/Connection/ConnectionConstants";

const ConnectionManager = new ConnectionManagerClass();

export { ConnectionConstants, ConnectionManager, XHRConnection };
```

nLight:
Commit messages don't pass the lint. Should be feat(...)**:** bla bla

so that colon is missing

packages/connection-manager/src/js/Connection/XHRConnection.js
```diff
@@ -0,0 +1,198 @@
...
+export default class XHRConnection extends AbstractConnection {
+  /**
+   * Initializes an Instance of XHRConnection
+   * @constructor
+   * @param {string} url – URL to be fetched
+   * @param {string} [_method=GET] – used method
+   * @param {mixed} [_data] – payload for request
+   * @param {string} [_header] – additional headers (like content-type)
+   */
+  constructor(url, _method = "GET", _data = null, _header = null) {
```
orlandohohmeier: Let's pass all of the optional arguments as an options object as suggested in   https://github.com/dcos/dcos-ui/pull/2409#discussion_r138210249, to align the interface with known request libs as well as the coming [`Request`](https://developer.mozilla.org/en-US/docs/Web/API/Request/Request ) (fetch) in the standard lib.

## XHRConnection

```js
var myRequest = new XHRConnection(url, options);

```` 

### Parameter

* `url` - Defines the resource that you wish to request
* `options` [optional] - Optional options object containing any custom settings that you want to apply to the request.

#### Options

* `method` - Request method, e.g., GET, POST.
* `headers` -  Any headers you want to add to your request, contained within a Headers object or an object literal with ByteString values.
* body: Any body that you want to add to your request: this can be a Blob, BufferSource, FormData, URLSearchParams, or USVString object. Note that a request using the GET or HEAD method cannot have a body.

What do you think?

/cc @dcos/frontend-foundation

File: packages/connection-manager/src/js/Connection/ConnectionConstants.js [hunk 26]
```diff
@@ -0,0 +1,11 @@
+export const CONNECTION_EVENT_OPEN = Symbol("CONNECTION_EVENT_OPEN");
```
orlandohohmeier: Let's please split these into state constants and event types and remove the unnecessary prefix as discussed offline and questioned in https://github.com/dcos/dcos-ui/pull/2409#discussion_r138212722

packages/connection-manager/src/js/Connection/XHRConnection.js
```diff
@@ -0,0 +1,198 @@
...
+    super(url);
+
+    var method;
```
orlandohohmeier: Please always use `const` and `let` instead of `var`... @dcos/frontend-foundation can we add a eslint rule for this one?

orlandohohmeier:
@nLight are you ok with squashing 5f092d8 and 5a0f204?

nLight:
@dcos/frontend-foundation I adjusted the implementation and squashed commits 🙇

packages/connection-manager/src/js/Connection/XHRConnection.js
```diff
@@ -0,0 +1,187 @@
...
+/**
+ * Basic XHR Connection
+ * @todo add/remove headers
+ */
+export default class XHRConnection extends AbstractConnection {
+  /**
+   * Initializes an Instance of XHRConnection
+   * @constructor
+   * @param {string} url – URL to be fetched
+   * @param {object} options
```
orlandohohmeier: Please indicate the these arguments are optional.

```
@param {object} [options]
@param {string} [options.method=GET] – used method
@param {*} [options.body] – payload for request
@param {string} [options.headers] – additional headers (like content-type)
```

http://usejsdoc.org/tags-param.html#optional-parameters-and-default-values

packages/connection-manager/src/js/Connection/XHRConnection.js
```diff
@@ -0,0 +1,187 @@
...
+   * @param {string} options.method=GET – used method
+   * @param {mixed} options.body – payload for request
```
orlandohohmeier: Please use `{*}` to indicate that any type is allowed.

packages/connection-manager/src/js/ConnectionManager/__tests__/ConnectionManager-test.js
```diff
@@ -0,0 +1,48 @@
...
+  beforeEach(() => {
+    ConnectionManager = new ConnectionManagerClass();
+  });
+  it("adds listeners to new connection", () => {
+    const connection = new Connection(url);
+
+    ConnectionManager.queue(connection);
+    expect(connection.on).toBeCalled();
+  });
+  /* it("doesnt add listeners second time", () => {
```
orlandohohmeier: Please always remove unused code.

packages/connection-manager/src/js/ConnectionQueue/__tests__/ConnectionQueue-test.js
```diff
@@ -0,0 +1,210 @@
...
+import ConnectionQueue from "../ConnectionQueue.js";
+// import ConnectionQueueItem from "../ConnectionQueueItem.js";
+
+jest.mock("../../Connection/AbstractConnection");
+// jest.mock("../ConnectionQueueItem");
+
+/* per default, jest reads /__mocks__/immutable.js even with automock disabled. */
+jest.unmock("immutable");
+
+// Object.assign(global[Symbol.for("$$jest-matchers-object")].matchers, {
```
orlandohohmeier: Please remove all unused code.

packages/connection-manager/src/js/ConnectionQueue/__tests__/ConnectionQueue-test.js
```diff
@@ -0,0 +1,210 @@
...
+        .dequeue(connection);
+
+      expect(queue.size).toEqual(0);
+    });
+  });
+  describe("#contains", () => {});
+});
+
+/*
+var ConnectionStore;
```
orlandohohmeier: These test seem old. Please remove the old test and add a sufficient set for the `ConnectionQueue`

packages/connection-manager/src/js/Connection/StateConstants.js
```diff
@@ -0,0 +1,4 @@
+export const INIT = Symbol("CONNECTION_STATE_INIT");
```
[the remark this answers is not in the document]
nLight: https://www.youtube.com/watch?v=GxBSyx85Kp8

packages/connection-manager/src/js/Connection/EventConstants.js
```diff
@@ -0,0 +1,6 @@
+export const OPEN = Symbol("CONNECTION_EVENT_OPEN");
```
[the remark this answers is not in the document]
nLight: I'm not sure about the Symbols actually

nLight:
Done! ✅

packages/connection-manager/src/js/Connection/StateConstants.js
```diff
@@ -0,0 +1,4 @@
...
+export const STARTED = Symbol("CONNECTION_STATE_STARTED");
+export const DONE = Symbol("CONNECTION_STATE_DONE");
+export const CANCELED = Symbol("CONNECTION_STATE_CANCELED");
```
MatApple: #nit `CANCELED` > `CANCELLED`

nLight:
@orlandohohmeier @MatApple @weblancaster please give it another look!

packages/connection-manager/src/js/ConnectionQueue/ConnectionQueue.js
```diff
@@ -0,0 +1,91 @@
...
+  head() {
+    return this.connections.first().connection;
+  }
+
+  /**
+   * returns a new List without the first Element
+   * @return {ConnectionQueue} – new ConnectionQueue without first Connection
+   */
+  tail() {
+    return new ConnectionQueue(this.connections.delete(0));
```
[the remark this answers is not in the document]
MatApple: What happens if `this.connections` is empty?

packages/connection-manager/src/js/ConnectionQueue/ConnectionQueue.js
```diff
@@ -0,0 +1,91 @@
...
+   */
+  get size() {
+    return this.connections.size;
+  }
+
+  /**
+   * returns head of Set - no removal
+   * @return {AbstractConnection} – head Connection
+   */
+  head() {
```
MatApple: Let's make sure we handle the edge cases in these methods to create a robust implementation. Handling the edge cases for a method should be done in the method otherwise it is hard to reason about and follow.

packages/connection-manager/src/js/ConnectionManager/ConnectionManager.js
```diff
@@ -0,0 +1,102 @@
...
+   * @param {AbstractConnection} connection
+   * @return {void}
+   */
+  next() {
+    if (this.open.size >= this.maxConnections || this.waiting.size === 0) {
+      return;
+    }
+
+    const connection = this.waiting.head();
+    this.waiting = this.waiting.tail();
```
MatApple: Let's revise `this.waiting`. The methods we're using on this `Queue` are a little strange. A Queue has `enqueue`, `dequeue`, and a few other helpers for peeking and checking size/empty. I don't understand why we've introduced `head` and `tail` methods. A `head` and  `tail` are usually reserved for internal use by the queue to track position and depend on the `queue`s implementation. Because we use a `List` as the core storage mechanism, I don't see why we even need to track head/tail since a `List` in Javascript provides out of the box methods for this.

packages/connection-manager/src/js/Connection/XHRConnection.js
```diff
@@ -0,0 +1,187 @@
...
+  /**
+   * Initializes an Instance of XHRConnection
+   * @constructor
+   * @param {string} url – URL to be fetched
+   * @param {object} [options]
+   * @param {string} [options.method=GET] – used method
+   * @param {*} [options.body] – payload for request
+   * @param {string} [options.headers] – additional headers (like content-type)
+   */
+  constructor(url, options = {}) {
```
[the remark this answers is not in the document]
nLight: I'd rather have this in the `resource-service`, Observables will give us this for free

packages/connection-manager/src/js/ConnectionQueue/ConnectionQueue.js
```diff
@@ -0,0 +1,91 @@
...
+   * @param {AbstractConnection} connection – Connection to be added to queue
+   * @param {Integer} priority – priority of connection
+   * @return {ConnectionQueue} – new ConnectionQueue containing the added Connection
+   */
+  enqueue(connection, priority) {
+    return new ConnectionQueue(
+      this.connections
+        .push(new ConnectionQueueItem(connection, priority))
+        .sort((a, b) => {
+          return Math.min(Math.max(a.priority >= b.priority, -1), 1);
```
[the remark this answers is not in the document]
juliangieseke: it could also be `return (a.priority > b.priority) ? 1 : (a.priority < b.priority) : -1 : 0`
but, eslint complains, its harder to read (IMHO) and I didnt want to write it expanded.

packages/connection-manager/src/js/ConnectionQueue/ConnectionQueue.js
```diff
@@ -0,0 +1,91 @@
...
+        })
+    );
+  }
+
+  /**
+   * removes a connection from store
+   * @param {AbstractConnection} connection – connection to be deleted
+   * @return {AbstractConnection} removed connection
+   */
+  dequeue(connection) {
```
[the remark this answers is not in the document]
juliangieseke: uh, looks like i forgot the `remove` method in CM 🤔 
we should be able to remove connections from the queue.

packages/connection-manager/src/js/ConnectionManager/ConnectionManager.js
```diff
@@ -0,0 +1,102 @@
...
+
+    if (connection.state === INIT) {
+      connection.open(); // AuthStore.getTokenForURL(connection.url)
+    }
+
+    if (connection.state === INIT || connection.state === STARTED) {
+      this.open = this.open.push(connection);
```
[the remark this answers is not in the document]
juliangieseke: true, we should move line 81 to 86 or something

packages/connection-manager/src/js/ConnectionQueue/ConnectionQueue.js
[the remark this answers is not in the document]
juliangieseke: as stated before, we had some trouble naming/implementing theses methods. but it actually returns a new queue without the first Item, which is basically a pointer to the tail (its immutable too, so you still have the old one in your code)

packages/connection-manager/src/js/ConnectionQueue/ConnectionQueue.js
```diff
@@ -0,0 +1,91 @@
...
+
+    return new ConnectionQueue(this.connections.delete(index));
+  }
+
+  /**
+   * checks if a connection is already enqueued
+   * @param {AbstractConnection} connection – given connection to check for
+   * @return {bool} – true: queued, false: unknown
+   */
+  contains(connection) {
```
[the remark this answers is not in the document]
juliangieseke: * re 1: we had and `update` functionality before, but removed it. also if we want to remove items from the Queue from CM, we need this. (I think)
* re 2: We could do it, yes. I forgot why we decided no to do it 🤔

packages/connection-manager/README.md
```diff
@@ -0,0 +1,37 @@
+# Connection Manager
+
+## Why
```
orlandohohmeier: Let's remove this headline.

packages/connection-manager/src/js/Connection/XHRConnection.js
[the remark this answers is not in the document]
juliangieseke: I have added this to JIRA for further discussion, but we should leave it as it is now for this PR.

packages/connection-manager/src/js/Connection/XHRConnection.js
```diff
@@ -0,0 +1,187 @@
...
+   * more xhr related getter
+   * @todo make them more robust / do we need more?
+   */
+  get response() {
+    return this.xhr.response;
+  }
+  get readyState() {
+    return this.xhr.readyState;
+  }
+  get responseType() {
```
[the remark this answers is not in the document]
juliangieseke: I have added this to JIRA for further discussion, but we should leave it as it is now for this PR.

packages/connection-manager/src/js/Connection/XHRConnection.js
```diff
@@ -0,0 +1,187 @@
...
+  }
+  get status() {
+    return this.xhr.status;
+  }
+
+  /**
+   * create, prepare, open and send the xhr request
+   * @param {string} [token] – authentication token
+   */
+  open(token) {
```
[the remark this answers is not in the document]
juliangieseke: I have added this to JIRA for further discussion, but we should leave it as it is now for this PR.

packages/connection-manager/src/js/Connection/XHRConnection.js
```diff
@@ -0,0 +1,187 @@
...
+    if (this.xhr !== null) {
+      throw new Error("cannot open XHR Connection a second time!");
+    }
+    this.xhr = new XMLHttpRequest();
```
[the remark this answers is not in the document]
juliangieseke: I have added this to JIRA for further discussion, but we should leave it as it is now for this PR.

packages/connection-manager/src/js/Connection/XHRConnection.js
```diff
@@ -0,0 +1,187 @@
...
+   * aborts native xhr
+   */
+  close() {
+    this.xhr.abort();
+  }
+  /**
+   *
+   * @todo abort here before delete?
+   */
+  reset() {
```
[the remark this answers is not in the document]
juliangieseke: I have added this to JIRA for further discussion, but we should leave it as it is now for this PR.

packages/connection-manager/src/js/Connection/XHRConnection.js
```diff
@@ -0,0 +1,187 @@
...
+    }
+    this.emit(CLOSE, this);
+  }
+  /**
+   * handle timeout  of native xhr
+   */
+  handleTimeout() {
+    this.handleError();
+  }
+}
```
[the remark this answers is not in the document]
juliangieseke: I have added this to JIRA for further discussion, but we should leave it as it is now for this PR.

packages/connection-manager/src/js/ConnectionManager/ConnectionManager.js
```diff
@@ -0,0 +1,102 @@
...
+ * actually starting them, when they are head of
+ * waiting line.
+ */
+export default class ConnectionManager {
+  /**
+   * Initializes an Instance of ConnectionManager
+   * @param {Integer} maxConnections – max open connections
+   */
+  constructor(maxConnections = 6) {
+    Object.defineProperty(this, "maxConnections", {
```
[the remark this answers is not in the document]
juliangieseke: I have added this to JIRA for further discussion, but we should leave it as it is now for this PR.

packages/connection-manager/src/js/ConnectionManager/ConnectionManager.js
juliangieseke: I have added this to JIRA for further discussion, but we should leave it as it is now for this PR.

packages/connection-manager/src/js/ConnectionQueue/ConnectionQueue.js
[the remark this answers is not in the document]
juliangieseke: I have added this to JIRA for further discussion, but we should leave it as it is now for this PR.

packages/connection-manager/src/js/ConnectionQueue/ConnectionQueue.js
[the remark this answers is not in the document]
juliangieseke: I have added this to JIRA for further discussion, but we should leave it as it is now for this PR.

packages/connection-manager/src/js/ConnectionQueue/ConnectionQueue.js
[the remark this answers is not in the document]
juliangieseke: I have added this to JIRA for further discussion, but we should leave it as it is now for this PR.

packages/connection-manager/src/js/Connection/XHRConnection.js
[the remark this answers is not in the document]
orlandohohmeier: Please note that the connection manager is a low level package and all theses operation are going to be abstract by other packages like the proposed resource service.

packages/connection-manager/src/js/Connection/XHRConnection.js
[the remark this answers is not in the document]
orlandohohmeier: No! There is no way I'm going to ever green light anything like this. Overwriting or adjusting any of the native APIs comes with a huge high risk of introducing bugs that are nearly impossible to fix.

packages/connection-manager/src/js/Connection/XHRConnection.js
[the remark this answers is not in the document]
orlandohohmeier: Functionality like `retry` is application specific and must not be part of the connection.

packages/connection-manager/src/js/ConnectionManager/ConnectionManager.js
orlandohohmeier: The respective functions _mirror_ the immutable [Queue](http://www.scala-lang.org/api/2.12.0/scala/collection/immutable/Queue.html) (part of Scala standard lib) implementation and follow a well known pattern thus I don't think we need to revise them. But we might want to rethink the general naming as one could argue that's not a queue due to the sorting we're applying which violates the FIFO principal.

packages/connection-manager/src/js/ConnectionQueue/ConnectionQueue.js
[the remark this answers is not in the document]
orlandohohmeier: There is nothing wrong with the name and behavior, but we might want add a null check as suggest. 

Compare to haskell or Scala:

> tail takes a list and returns its tail. In other words, it chops off a list's head.

http://learnyouahaskell.com/starting-out

> returns a [...] collection consisting of all elements of this  [...]  collection except the first one.

http://www.scala-lang.org/api/2.12.0/scala/collection/GenTraversableLike.html#tail:Repr

packages/connection-manager/src/js/Connection/StateConstants.js
```diff
@@ -0,0 +1,4 @@
+export const INIT = Symbol("INIT");
```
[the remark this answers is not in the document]
juliangieseke: it removed the need of importing the constants explicitly.. like that.

packages/connection-manager/src/js/Connection/XHRConnection.js
```diff
@@ -0,0 +1,187 @@
...
+  constructor(url, options = {}) {
+    super(url);
+
+    const { method = "GET", body = null, headers = null } = options;
+
+    let _method;
+    Object.defineProperty(this, "method", {
+      get: () => _method,
+      set: __method => {
+        if (ALLOWED_METHODS.indexOf(__method) === -1) {
```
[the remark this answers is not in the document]
juliangieseke: `includes`

packages/connection-manager/src/js/Connection/StateConstants.js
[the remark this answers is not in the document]
juliangieseke: => https://github.com/dcos/dcos-ui/pull/2361/files#diff-b9e18da2e01a60e4b2c5941fa751bd95R48

juliangieseke:
Everything should be addressed or moved into JIRA, we should be able to merge this one now. 🤞

juliangieseke:
one more thing: we have to add a ConnectionEvent before we can merge it. I'll implement it tomorrow

packages/connection-manager/src/js/ConnectionManager/ConnectionManager.js
orlandohohmeier: The function names sparked quite some discussion and thus even though they follow well known patterns we're going revisit the names and try to align them with non immutable implementations. /cc @dcos/frontend-foundation

MatApple: @juliangieseke I'm sorry to bring up more PITA things and appreciate your patience and diligence on this. The immutable aspect is what makes this tricky. My suggestion would be to keep `tail()` as long as we rename `head()` to `peek()`.

orlandohohmeier: @MatApple Having a `peek` and `tail` method sounds wrong to me as that would really follow any pattern. We're going to either drop immutable.js and rewrite the queue to provide mutable operations similar to those found in Java or expose the immutable.js interface (exposing `first` and `shift` should be sufficient) instead of following any known implementation. /cc @dcos/frontend-foundation @juliangieseke

juliangieseke: changed `head` to `first` and `tail` to `shift`. 👍

nLight:
Alfred retest this please

packages/connection-manager/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+Browsers can open only a limited amount of connections so when you urgently need one there may not be a slot. This package solves the issue by providing a manager that enqueues all requested connections and priorities them accordingly.
+
+## Usage
+
+```javascript
+import ConnectionManager, { XHRConnection } from "connection-manager";
+
+const connection = new XHRConnection("http://127.0.0.1");
+connection.on("close", function() {
```
orlandohohmeier: Please update the doc as the implementation has changed.

packages/connection-manager/src/js/Connection/XHRConnection.js
```diff
@@ -0,0 +1,229 @@
...
+    } = options;
+
+    let _method;
+    Object.defineProperty(this, "method", {
+      get: () => _method,
+      set: __method => {
+        if (ALLOWED_METHODS.includes(__method) === -1) {
+          throw new Error("Invalid method for XHR Request.");
+        }
+        _method = __method;
```
orlandohohmeier: Not blocking, but do really need all these underscores?  Let's please not use a hungarian notation to flag them as private.

packages/connection-manager/src/js/Connection/XHRConnection.js
```diff
@@ -0,0 +1,229 @@
...
+      method = "GET",
+      body = null,
+      headers = null,
+      responseType = "json"
+    } = options;
+
+    let _method;
+    Object.defineProperty(this, "method", {
+      get: () => _method,
+      set: __method => {
```
orlandohohmeier: We shouldn't expose any setters for these properties!

packages/connection-manager/src/js/ConnectionQueue/ConnectionQueue.js
```diff
@@ -0,0 +1,95 @@
...
+   * @param {AbstractConnection} connection – Connection to be added to queue
+   * @param {Integer} priority – priority of connection
+   * @return {ConnectionQueue} – new ConnectionQueue containing the added Connection
+   */
+  enqueue(connection, priority) {
+    return new ConnectionQueue(
+      this.connections
+        .push(new ConnectionQueueItem(connection, priority))
+        .sort((a, b) => {
+          return Math.min(Math.max(a.priority >= b.priority, -1), 1);
```
orlandohohmeier: This doesn't work as `Math.min(Math.max(1 >= 1, -1), 1);` will return `1` not `0` and thus result in wrong order. You can user the following snippet if you don't want to use conditions:

 ```js
const delta = b.priority - a.priority;

 return delta / Math.abs(delta || 1);
```

File: packages/connection-manager/src/js/Connection/XHRConnection.js [hunk 57]
```diff
@@ -0,0 +1,229 @@
...
+      this.state = XHRConnection.DONE;
+      if (this.status >= 400) {
+        this.emit(
+          ConnectionEvent.ERROR,
+          new ConnectionEvent(this, ConnectionEvent.ERROR)
+        );
+      } else {
+        this.emit(
+          ConnectionEvent.SUCCESS,
+          new ConnectionEvent(this, ConnectionEvent.SUCCESS)
```
orlandohohmeier: This event doesn't exists.

packages/connection-manager/src/js/Connection/XHRConnection.js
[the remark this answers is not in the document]
orlandohohmeier: Let's remove this method as it adds quite some complexity and I like to be sure that we really need this feature before we add all this extra code.

packages/connection-manager/src/js/Connection/AbstractConnection.js
```diff
@@ -0,0 +1,91 @@
...
+      get: () => symbol
+    });
+  }
+  static get INIT() {
+    return Symbol.for("DCOS.ConnectionManager.AbstractConnection.INIT");
+  }
+  static get STARTED() {
+    return Symbol.for("DCOS.ConnectionManager.AbstractConnection.STARTED");
+  }
+  static get DATA() {
```
orlandohohmeier: There is no `DATA` state.

packages/connection-manager/src/js/Connection/AbstractConnection.js
```diff
@@ -0,0 +1,91 @@
...
+      }
+    });
+
+    const symbol = Symbol("Connection:" + this.url + this.created);
+    Object.defineProperty(this, "symbol", {
+      get: () => symbol
+    });
+  }
+  static get INIT() {
+    return Symbol.for("DCOS.ConnectionManager.AbstractConnection.INIT");
```
orlandohohmeier: Let's define the symbols using constants and only pass them along instead of using `Symbol.for` to avoid recreation.

packages/connection-manager/src/js/Connection/AbstractConnection.js
```diff
@@ -0,0 +1,91 @@
+import { EventEmitter } from "events";
```
orlandohohmeier: Let's please reduce the number of possible states to make it easier to reason about a connections state.

juliangieseke:
closing in favor of #2418 (connections) and #2425 (manager).

PR #2362, opened 2017-08-08
fix(ServiceDestroyModal): improve wording on delete/restart/suspend

weblancaster:
Improve wording/sentence when deleting/suspending/restarting a service/pod/group

Closes DCOS-17296 and DCOS-17534

Be aware that the wording was updated here in comparison to the Jira comment after I discussed "offline" with @ashenden 

**Now:**
screen shot 2017-08-08 at 3 35 43 pm
screen shot 2017-08-08 at 3 36 03 pm
screen shot 2017-08-08 at 3 36 15 pm

**Checklist**

### DIFF
--- a/plugins/services/src/js/components/modals/ServiceDestroyModal.js
+++ b/plugins/services/src/js/components/modals/ServiceDestroyModal.js
@@ -198,9 +198,7 @@ class ServiceDestroyModal extends React.Component {
     const { open, service } = this.props;
     const serviceName = service.getName();
     const serviceLabel = this.getServiceLabel();
-
-    let itemText = `${StringUtil.capitalize(UserActions.DELETE)}`;
-    itemText += ` ${serviceLabel}`;
+    const itemText = `${StringUtil.capitalize(UserActions.DELETE)} ${serviceLabel}`;
 
     return (
       <Confirm
--- a/plugins/services/src/js/components/modals/ServiceDestroyModal.js
+++ b/plugins/services/src/js/components/modals/ServiceDestroyModal.js
@@ -223,7 +223,7 @@ class ServiceDestroyModal extends React.Component {
           <strong>{serviceName}</strong>
           {" "}
           {serviceLabel.toLowerCase()}.
-          Please type ("
+          Type ("
           <strong>{serviceName}</strong>
           ") below to confirm you want to delete the
           {" "}
--- a/plugins/services/src/js/components/modals/ServiceDestroyModal.js
+++ b/plugins/services/src/js/components/modals/ServiceDestroyModal.js
@@ -195,18 +195,12 @@ class ServiceDestroyModal extends React.Component {
   }
 
   getDestroyServiceModal() {
-    const { open, service, intl } = this.props;
+    const { open, service } = this.props;
     const serviceName = service.getName();
+    const serviceLabel = this.getServiceLabel();
 
     let itemText = `${StringUtil.capitalize(UserActions.DELETE)}`;
-
-    if (service instanceof Pod) {
-      itemText += " Pod";
-    }
-
-    if (service instanceof ServiceTree) {
-      itemText += " Group";
-    }
+    itemText += ` ${serviceLabel}`;
 
     return (
       <Confirm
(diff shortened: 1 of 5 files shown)


plugins/services/src/js/components/modals/ServiceDestroyModal.js
```diff
@@ -195,18 +195,12 @@ class ServiceDestroyModal extends React.Component {
...
     let itemText = `${StringUtil.capitalize(UserActions.DELETE)}`;
-
-    if (service instanceof Pod) {
-      itemText += " Pod";
-    }
-
-    if (service instanceof ServiceTree) {
-      itemText += " Group";
-    }
+    itemText += ` ${serviceLabel}`;
```
bstavroulakis: Maybe this would be more concise `let itemText = `${StringUtil.capitalize(UserActions.DELETE)} ${serviceLabel}`;`

plugins/services/src/js/components/modals/ServiceRestartModal.js
```diff
@@ -79,38 +80,64 @@ class ServiceRestartModal extends React.Component {
     );
   }
 
-  render() {
-    const { isPending, onClose, open, service, restartService } = this.props;
+  getServiceLabel() {
```
bstavroulakis: It looks like this is used in all three cases and is the same code. Would it make sense to add it somewhere centrally?

plugins/services/src/js/components/modals/ServiceSuspendModal.js
```diff
@@ -80,34 +80,46 @@ class ServiceSuspendModal extends React.Component {
...
     );
+  }
+
+  render() {
+    const { isPending, onClose, open, service, suspendItem } = this.props;
+
+    let itemText = "";
+    let serviceName = "";
+    const serviceLabel = this.getServiceLabel();
+    itemText += ` ${serviceLabel}`;
```
bstavroulakis: itemText looks like it is the same with serviceLabel and doesn't change. We could probably remove it altoghere and use only the serviceLabel

plugins/services/src/js/components/modals/ServiceDestroyModal.js
bstavroulakis: Maybe this would be more concise ``let itemText =`${StringUtil.capitalize(UserActions.DELETE)} ${serviceLabel}` ``;

plugins/services/src/js/components/modals/ServiceSuspendModal.js
weblancaster: Agree.

plugins/services/src/js/components/modals/ServiceDestroyModal.js
weblancaster: yes.

plugins/services/src/js/components/modals/ServiceRestartModal.js
```diff
@@ -79,38 +80,64 @@ class ServiceRestartModal extends React.Component {
...
       </ModalHeading>
     );
+  }
+
+  render() {
+    const { isPending, onClose, open, service, restartService } = this.props;
     let serviceName = "";
+    const serviceLabel = this.getServiceLabel();
 
     if (service) {
```
MatApple: Since you've cleaned this up a lot already - maybe we could remove this `if` statement too since service is a required prop!

plugins/services/src/js/components/modals/ServiceSuspendModal.js
```diff
@@ -80,47 +80,62 @@ class ServiceSuspendModal extends React.Component {
...
       </ModalHeading>
     );
+  }
+
+  render() {
+    const { isPending, onClose, open, service, suspendItem } = this.props;
+    const serviceLabel = this.getServiceLabel();
+    let serviceName = "";
+
+    if (service) {
```
MatApple: Same here!

Poltergeist:
Looks good thanks for the effort I will merge this PR Soon™️

PR #2363, opened 2017-08-09
DCOS-17361: round cpu allocation used and total value

bstavroulakis:
In the dashboard, in the cpu card under the percentage the (usedNumber) to (totalTotal) shares wasn't rounding up correctly.

Closes DCOS-17361

screen shot 2017-08-08 at 5 03 06 pm

**Checklist**

### DIFF
--- a/src/js/components/charts/ResourceTimeSeriesChart.js
+++ b/src/js/components/charts/ResourceTimeSeriesChart.js
@@ -24,19 +24,20 @@ var ResourceTimeSeriesChart = React.createClass({
   },
 
   getData() {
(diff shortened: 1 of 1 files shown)


ahoskins:
Can you give us a cluster scenario to test this? (Like some specific cluster resource loads.)

Or maybe we should add unit tests for these labels rendering the proper number of digits? This kind of issue seems to have come up multiple times. Thoughts?

weblancaster:
@ahoskins I just hard coded the value (same as screenshot) instead of taking from the prop.

PR #2365, opened 2017-08-09
fix(JobsList): add column sorting for Status and Last Run

ahoskins:
**Description**: 
Previously, jobs were only able to be sorted by Job Name. Nothing happened when clicking on the "Status" and "Last Run" columns.

This change adds sorting by "Status" and "Last Run". Both these columns are sorted in order of importance to the average user. For example, jobs who's last run failed are higher importance than successful jobs, and will be sorted to the top of the list.

The sort order is determined by the `sortOrder` fields in `JobStates.js` and `JobStatus.js`. This is similar to the existing [HealthSorting.js](https://github.com/dcos/dcos-ui/blob/master/plugins/services/src/js/constants/HealthSorting.js), which helps sort services.

**Testing**:
Go to the jobs page, and turn on fixtures, and click around on the column headers to see the sorting. Demo: https://cl.ly/3E1h143I160l

Here is my `jobs.json` fixture contents to make testing easier (added a few more jobs of different states): https://gist.github.com/ahoskins/7c316a86f3332b27c098be5243e2997e

Closes DCOS-16744.

**Checklist**

### DIFF
--- a/src/js/pages/jobs/JobsTable.js
+++ b/src/js/pages/jobs/JobsTable.js
@@ -6,12 +6,13 @@ import { Table, Tooltip } from "reactjs-components";
 
 import Icon from "../../components/Icon";
 import JobStates from "../../constants/JobStates";
+import JobStatus from "../../constants/JobStatus";
 import JobTableHeaderLabels from "../../constants/JobTableHeaderLables";
 import ResourceTableUtil from "../../utils/ResourceTableUtil";
 import TableUtil from "../../utils/TableUtil";
 import Tree from "../../structs/Tree";
 
-const METHODS_TO_BIND = ["renderHeadline"];
+const METHODS_TO_BIND = ["renderHeadline", "jobSortFunction"];
 
 class JobsTable extends React.Component {
   constructor() {
@@ -44,7 +45,7 @@ class JobsTable extends React.Component {
         prop: "name",
         render: this.renderHeadline,
         sortable: true,
-        sortFunction: this.sortJobNames
+        sortFunction: this.jobSortFunction
       },
       {
         className,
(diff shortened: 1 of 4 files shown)


ahoskins:
Put the fixture in a gist at: https://gist.github.com/ahoskins/7c316a86f3332b27c098be5243e2997e. Thanks @Poltergeist

src/js/pages/jobs/JobsTable.js
```diff
@@ -96,7 +100,22 @@ class JobsTable extends React.Component {
...
+        return (a, b) =>
+          JobStates[a.status].sortOrder - JobStates[b.status].sortOrder;
+      case "lastRun":
+        return (a, b) =>
+          JobStatus[a.lastRun.status].sortOrder -
+          JobStatus[b.lastRun.status].sortOrder;
+    }
+  }
+
+  jobSortFunction(prop, direction) {
```
weblancaster: what do you think about naming `jobSort` and remove the `function`

ahoskins: I named this way to conform to the same naming as [ServiceTableUtil.js](https://github.com/dcos/dcos-ui/blob/master/plugins/services/src/js/utils/ServiceTableUtil.js).

weblancaster: Yes I figured I believe that is a bad name convention being used there and would be nice if we stop doing that.

Not that we do that but for example purposes e.g `var tasksArray = []` so why `jobSortFunction ` it seems redundant to me, let's see what others think about it.

ahoskins: After me and @weblancaster discussed offline, we'll keep this naming as-is because in other parts of the codebase we follow this convention for functions we pass to `Array#sort()` (although probably a strange naming pattern) 👍

Poltergeist: Let's discuss this in the team sync please put it to the agenda. I am happy to change this.

PR #2368, opened 2017-08-10
[WIP] feat(MesosEventManager): implemented event stream based on Mesos V1 API.

suanmiao:
Implemented field-wise Mesos event stream using Bacon.js and Immutable.js.

To test:
- Create a custom open cluster using `testing/pull/1795`
- Import `MesosEventManager` and call `MesosEventManager.startConnection()`. This can be done in the MesosStateStore.
- You will see console logs from the stream when you run it


package.json
```diff
@@ -20,13 +20,17 @@
...
+    "last 2 versions",
+    "ie 11"
+  ],
   "config": {
     "port": 4200,
     "externalplugins": ""
   },
   "dependencies": {
     "babel-polyfill": "6.23.0",
+    "baconjs": "0.7.95",
```
orlandohohmeier: Not sure about the latest numbers but I'd use Rx or Kefir instead of Beacon due to performance issues.  https://github.com/Reactive-Extensions/RxJS/blob/master/doc/mapping/bacon.js/whyrx.md

package.json
```diff
@@ -59,7 +64,9 @@
     "reactjs-components": "0.16.0-beta.23",
     "reactjs-mixin": "0.0.2",
     "redux": "3.3.1",
-    "tv4": "1.2.7"
+    "tv4": "1.2.7",
+    "winston": "2.3.1",
```
orlandohohmeier: For what do we need winston?

src/js/components/MesosOperatorApiClient.js
```diff
@@ -0,0 +1,489 @@
...
+const http = require("http");
+
+const helpers = require("../utils/mesos-operator-api/MesosOperatorApiUtil");
+
+/**
+ * Represents a mesos Event Bus Client
+ * @constructor
+ * @param {object} options - The option map object.
+ */
+function MesosOperatorApiClient(options) {
```
orlandohohmeier: I'm not sure I understand where and how this manager will be used.

tnguyen0:
@bstavroulakis is taking over this. Closing.

PR #2372, opened 2017-08-16
DCOS-17939: Blacklist historySummary

nLight:
historySummary field prevents you from saving a metronome job, so we should just blacklist it as it is a meta information. Please refer to the ticket to come up wit a testing scenario.


orlandohohmeier:
LGTM! 👓 👍

PR #2374, opened 2017-08-17
fix(Services): remove "open service" buttons for SDK services

ahoskins:
This action is not supported for SDK service, so remove it from three places: two links in the service table row, and one link in the service detail header.

Closes DCOS-17979.

**Checklist**

### DIFF
--- a/tests/pages/services/ServiceActions-cy.js
+++ b/tests/pages/services/ServiceActions-cy.js
@@ -7,16 +7,13 @@ describe("Service Actions", function() {
   }
 
   context("Open Service Action", function() {
-    beforeEach(function() {
+    it('displays the "Open Service" option for services that have a web UI', function() {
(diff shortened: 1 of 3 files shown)


tests/pages/services/ServiceActions-cy.js
```diff
@@ -37,6 +38,22 @@ describe("Service Actions", function() {
           expect($menuItem.length).to.equal(0);
         });
     });
+
+    it('does not display the "Open Service" option for SDK services', function() {
```
ahoskins: Just added a text case specifically for SDK services, and slight refactor removing the `beforeEach()`.

MatApple:
@ahoskins minor nit but I don't think the "scope" in the commit message is quite right :)

ahoskins:
run integration tests

PR #2375, opened 2017-08-18
feat(PackageDetailTab): implement package version dropdown

weblancaster:
This commit includes the work for epic DCOS_OSS-1528 where as a user now
is able to select a version in package detail page.

What is included:
- package detail page UI dropdown element
- Actions to retrieve data for package versions

How to test:
- Go to catalog page click on a package
- Notice the dropdown at the top where the package name is
- When choosing a version the content should change (if there is any difference) to the chosen version
- When deployed the version selected will be installed

**before:**
screen shot 2017-08-21 at 2 55 25 pm

**after:**
screen shot 2017-08-21 at 2 58 39 pm

**Checklist**


orlandohohmeier:
@weblancaster please squash 0049d92 before moving on with this one, to keep the history clean.

leemunroe:
@weblancaster Works great functionality wise. Visually could do with a tweak. Currently looks like the dropdown is floating in an odd position. Please sync with @lilysamimi on visual treatment.

weblancaster:
All this is ready to review.

@orlandohohmeier I will squash that commit.

weblancaster:
@leemunroe this is the new deisgn version discussed with @lilysamimi.

We need to discuss more since I notice after implementing that if the package name is too long what should we do.

weblancaster:
retest this please

src/js/events/CosmosPackagesActions.js
```diff
@@ -127,9 +129,8 @@ const CosmosPackagesActions = {
       data: JSON.stringify({ packageName, packageVersion }),
       success(response) {
         const cosmosPackage = response.package;
-        if (!cosmosPackage.currentVersion && cosmosPackage.version) {
-          cosmosPackage.currentVersion = cosmosPackage.version;
-        }
+
+        cosmosPackage.currentVersion = cosmosPackage.version;
```
weblancaster: This is the first version sent by the response or the version selected in the dropdown.

src/js/pages/catalog/PackageDetailTab.js
```diff
@@ -78,15 +79,31 @@ class PackageDetailTab extends mixin(StoreMixin) {
     });
   }
 
-  componentDidMount() {
-    super.componentDidMount(...arguments);
-
+  retrievePackageInfo() {
```
weblancaster: instead of having the same methods being called in `componentDidMount` and `componentDidUpdate` I consolidated the methods under `retrievePackageInfo`

src/js/pages/catalog/PackageDetailTab.js
```diff
@@ -78,15 +79,31 @@ class PackageDetailTab extends mixin(StoreMixin) {
...
 
+  componentDidMount() {
+    super.componentDidMount(...arguments);
+    this.retrievePackageInfo();
+  }
+
+  componentDidUpdate(prevProps) {
+    const { version } = this.props.location.query;
+
+    if (version === prevProps.location.query.version) {
```
weblancaster: If this is not here we will be calling the API all the time because of the react lifecycle and how we architecture the UI, this is improving performance by only calling the server when there's actually a change.

src/js/pages/catalog/__tests__/PackageDetailTab-test.js
```diff
@@ -6,21 +6,22 @@ jest.dontMock("../../../mixins/InternalStorageMixin");
...
 jest.dontMock("../../../../../tests/_fixtures/cosmos/package-describe.json");
-
-// Setting useFixtures for when we load CosmosPackagesStore/CosmosPackageActions
-/* eslint-disable import/newline-after-import */
-const Config = require("../../../config/Config");
-var configUseFixtures = Config.useFixtures;
-Config.useFixtures = true;
-require("../../../stores/CosmosPackagesStore");
-Config.useFixtures = configUseFixtures;
-/* eslint-enable import/newline-after-import */
```
weblancaster: No need for the lines above in the unit tests. We only needed the fixtures.

tests/_fixtures/cosmos/package-describe.json
```diff
@@ -1,18 +1,19 @@
 {
```
weblancaster: This file was one year old therefore outdated. I took the updated response for the same package.

tests/_fixtures/cosmos/package-list-versions.json
```diff
@@ -0,0 +1,7 @@
+{
```
weblancaster: Needed for unit tests and most important, integration/system tests.

src/js/components/PackageDetailVersionsDropdown.js
```diff
@@ -0,0 +1,51 @@
...
+class PackageDetailVersionsDropdown extends Component {
+  constructor() {
+    super();
+
+    METHODS_TO_BIND.forEach(method => {
+      this[method] = this[method].bind(this);
+    });
+  }
+
+  handleVersionChange(event) {
```
weblancaster: Here I use the mechanism already in place which is based on the query `version` so I only change the location that way triggering the rest of the actions.

leemunroe:
@weblancaster @lilysamimi In that screenshot it looks like we're doubling up on the version. If we have the version displayed in the dropdown we don't also need it beside the certified label, right?

Visually, the fact it is styled like a form element looks odd. Can we introduce a class for text style dropdowns? Basically the same type of style as a `.button-link` which would sit better beside the heading.

src/js/components/PackageDetailVersionsDropdown.js
```diff
@@ -0,0 +1,51 @@
...
+    const selectedVersion = `${originalPath}?version=${packageVersion}`;
+
+    global.location.replace(selectedVersion);
+  }
+
+  render() {
+    const { cosmosPackage, cosmosPackageVersions } = this.props;
+    const currentVersion = cosmosPackage.getCurrentVersion();
+    const packageVersions = cosmosPackageVersions.getAllVersions();
+    const sortedPackageVersions = Object.keys(packageVersions).sort();
```
tnguyen0: It looks like the intention here is that we want to sort such that the latest version is last. Then, we should short using the values (instead of keys) from the response object. For example, chronos is 

`{2.5.0: "5", 3.0.1: "4", 3.0.0: "2", 2.4.0: "0"}` (case here is 2.5.0 is actually latest)

Also, if you look at the sort order for confluent-kafka package, the sort order is not correct since `1.1.4-3.0.0` should come before `1.1.19.1-3.2.2`.

src/js/stores/CosmosPackagesStore.js
```diff
@@ -263,6 +280,15 @@ class CosmosPackagesStore extends GetSetBaseStore {
     return null;
   }
 
+  getPackageVersions() {
+    const packageVersions = this.get("packageVersions");
+    if (packageVersions) {
+      return new UniversePackage(packageVersions);
```
tnguyen0: I personally don't think `UniversePackage` wrapper is needed here and we can just return the object directly.

src/js/events/CosmosPackagesActions.js
```diff
@@ -149,6 +150,35 @@ const CosmosPackagesActions = {
...
+        AppDispatcher.handleServerAction({
+          type: REQUEST_COSMOS_PACKAGE_LIST_VERSIONS_SUCCESS,
+          data: packageVersions
+        });
+      },
+      error(xhr) {
+        AppDispatcher.handleServerAction({
+          type: REQUEST_COSMOS_PACKAGE_LIST_VERSIONS_ERROR,
+          data: RequestUtil.getErrorFromXHR(xhr),
+          xhr
```
tnguyen0: You can also include `packageName` in the error object here.

src/js/events/__tests__/CosmosPackagesActions-test.js
```diff
@@ -220,6 +220,126 @@ describe("CosmosPackagesActions", function() {
...
+
+    it("sends query in request body", function() {
+      expect(JSON.parse(this.configuration.data)).toEqual({
+        packageName: "foo",
+        includePackageVersions: false
+      });
+    });
+
+    it("sends query in request body, even if it is undefined", function() {
+      CosmosPackagesActions.fetchPackageDescription();
```
tnguyen0: It looks like you meant to call `CosmosPackageActions.fetchPackageVersions()` here.

src/js/stores/__tests__/CosmosPackagesStore-test.js
```diff
@@ -200,6 +200,106 @@ describe("CosmosPackagesStore", function() {
     });
   });
 
+  describe("#fetchPackageVersions", function() {
+    const packageVersions = {
```
tnguyen0: #nit: You could put this in a fixture file.

src/js/components/PackageDetailVersionsDropdown.js
weblancaster: Correct, I'm going to change to sort by the values. Thanks.

src/js/events/CosmosPackagesActions.js
weblancaster: why though?

tnguyen0: I see it being done in the other tests. So I figure there is reason for it.

src/js/events/__tests__/CosmosPackagesActions-test.js
weblancaster: Yes, good catch.

src/js/stores/__tests__/CosmosPackagesStore-test.js
weblancaster: yes, the ideal in unit tests are to keep them on the file but because our mock data/fixtures are big and we use them for integration/system test.

And the more dependencies the worse.

that being said this mock being small I preferred to keep it only for the`describe` here.

src/js/events/__tests__/CosmosPackagesActions-test.js
weblancaster: fixed.

src/js/events/CosmosPackagesActions.js
weblancaster: yes I saw that too, I found no reason to be passing.

src/js/components/PackageDetailVersionsDropdown.js
weblancaster: fixed. even though the `values` can also be not in "order"

weblancaster:
@leemunroe @lilysamimi

design updated: turns out we had the ability to accomplish what you guys wanted without "extra" work.

screen shot 2017-08-28 at 3 18 40 pm

weblancaster:
please head to https://github.com/dcos/dcos-ui/pull/2397 since the PR was split into 3 PRs

PR #2377, opened 2017-08-18
DCOS-17705: fix(JSONSingleContianerParser): remove duplicated fetch parser

Poltergeist:
Remove duplicated fetch parser in the JSONSingleContainerParser.

As it happens the fetch parser was duplicated in the Parser and it was creating duplicated transactions for artifacts.

Closes DCOS-17705

### DIFF
--- a/plugins/services/src/js/reducers/JSONSingleContainerParser.js
+++ b/plugins/services/src/js/reducers/JSONSingleContainerParser.js
@@ -49,6 +49,5 @@ module.exports = [
   portMappings, // Note: must come after portDefinitions, as it uses its information!
   requirePorts,
   residency,
-  fetch,
   unknownVolumes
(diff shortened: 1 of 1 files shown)


plugins/services/src/js/reducers/JSONSingleContainerParser.js
```diff
@@ -49,6 +49,5 @@ module.exports = [
   portMappings, // Note: must come after portDefinitions, as it uses its information!
   requirePorts,
   residency,
-  fetch,
```
orlandohohmeier: 👯

nLight: I bet it sneaked in when I sorted them and then merged/rebased not careful enough...

PR #2378, opened 2017-08-18
DCOS-17997: revert icon var adjustments

orlandohohmeier:
Revert the icon variable adjustments introduced with 291ed72e as they result in rendering issues due to wrong `icon-mini` class specificity.

Closes DCOS-17997

/cc @ashenden

### DIFF
--- a/src/styles/components/icons/styles.less
+++ b/src/styles/components/icons/styles.less
@@ -109,6 +109,62 @@
       flex: 0 0 auto;
     }
   }
+
+  & when (@icon-tiny-enabled) {
+
+    .icon.icon-tiny {
+      height: @icon-tiny-height;
+      width: @icon-tiny-width;
+    }
+  }
+
+  & when (@icon-mini-enabled) {
+
+    .icon.icon-mini {
+      height: @icon-mini-height;
+      width: @icon-mini-width;
+    }
+  }
+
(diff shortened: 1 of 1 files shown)


orlandohohmeier:
@ashenden I'm happy to remove these overrides again once we've fixed CNVS

ashenden:
Before we go off an undo this, can you point me where this issue is happening?  It's probably a simple fix we can make in markup, as opposed to changing the CSS.

orlandohohmeier:
@ashenden the issue is that `.icon svg` is more specific than `.icon-mini` which is why the icon mini styles are overwritten by the default icon svg styles.

ashenden:
The problem is we are wrapping `svg.icon` in another element of class `.icon`.  This doubling down of `.icon` is what's creating the issue.

If you remove the `.icon` from the wrapping element, the size issue goes away. However, this is also a class called `.icon-margin-left` which adds spacing to the left of the icon (stupid approach if you ask me, but whatever).  Anyways, that class is dependent on being coupled to `.icon`.  So you need to change `.icon.icon-margin-left` to `.icon-margin-left`.

PROBLEM SOLVED!

VIDEO: https://cl.ly/3h172D1e1e0i

orlandohohmeier:
Closing in favor of #2379

PR #2381, opened 2017-08-21
↩️[release/1.9] chore(jenkins): add system tests config stub

nLight:
As we'd like to run our system tests for every PR that means running them for 1.9 PRs as well. But 1.9 didn't have any system test, so we need at least the config in place otherwise Jenkins build will fail

The config is a copy/paste from master with one exception:
```
channel: testing/1.9.3
```

Hardcoded channel will ensure that we launch proper cluster in case we want to back port some of the tests.

### DIFF
--- a/system-tests/driver-config/jenkins.sh
+++ b/system-tests/driver-config/jenkins.sh
@@ -1,0 +1,34 @@
+#!/bin/bash
+#
+# This is a configuration script to the system-test-driver that runs the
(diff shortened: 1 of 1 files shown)


orlandohohmeier:
System test setup fails due to missing files... 

```sh
cat: */tests.yml: No such file or directory
```

nLight:
...and it's 💚  \o/

orlandohohmeier:
LGTM! 🚢

PR #2383, opened 2017-08-23
refactor(DOMUtils): remove isElementOnTop util

orlandohohmeier:
Remove the `isElementOnTop` DOM util as the util is only used to check whether the `Chart` component should be updated. The checks performed by the utility force layout/reflow (for a
comprehensive list of properties or methods that force layout/reflow see:  https://gist.github.com/paulirish/5d52fb081b3570c81e3a) and as such shouldn't be used to check whether a component should be updated. The util also always returns true if an element is off the screen which completely renders the utility useless if used as a should component
update condition.

### DIFF
--- a/src/js/components/charts/Chart.js
+++ b/src/js/components/charts/Chart.js
@@ -48,10 +48,6 @@ var Chart = React.createClass({
     global.addEventListener("resize", this.updateWidth);
   },
 
-  shouldComponentUpdate() {
-    return DOMUtils.isElementOnTop(ReactDOM.findDOMNode(this));
-  },
-
   componentWillUnmount() {
     global.removeEventListener("resize", this.updateWidth);
   },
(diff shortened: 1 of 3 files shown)


nLight:
Look legit as well :D 
✘ Dashboard Create service tests "before each" hook for "increments CPU usage when service is started"
✘ Dashboard Create service tests "after each" hook for "increments CPU usage when service is started"
✘ Dashboard Delete service tests "before each" hook for "decrements CPU usage when service is stopped"

nLight:
And that one just looks to be flaky
✘ Universe uses advanced install to deploy a community package

orlandohohmeier:
The test failures seem unrelated... will trigger another run.

```
Error: App '/c7e6d1c373974be8bebcbd02d70307e4/dashboard-test-service' does not exist
```

orlandohohmeier:

nLight:
Alfred retest this one please

nLight:
Alfred retest this one please

PR #2395, opened 2017-08-28
feat(CosmosPackagesActions): [1 of 3] select package version actions/events/types

weblancaster:
_ps: to test out the whole feat you will need #2397 and #2396_

Commit including only the actions, event types and unit test for them.

Closes DCOS_OSS-1560 DCOS_OSS-1559 DCOS_OSS-1558

**Checklist**
- [x] Did you add a JIRA issue in a commit message or as part of the branch name?
- [x] Did you add new unit tests?
- [ ] Did you add new integration tests?
- [ ] If this is a regression, did you write a test to catch this in the future?

PR #2396, opened 2017-08-28
Select package version on deploy

weblancaster:
~## ⚠️  Depends on https://github.com/dcos/dcos-ui/pull/2395~

**Checklist**


tests/_fixtures/cosmos/package-list-versions.json
```diff
@@ -0,0 +1,7 @@
+{
```
weblancaster: leaving this here even though added to #2395 so tests can pass.

src/js/stores/CosmosPackagesStore.js
```diff
@@ -299,6 +325,18 @@ class CosmosPackagesStore extends GetSetBaseStore {
     this.emit(COSMOS_DESCRIBE_ERROR, error, name, version);
   }
 
+  processPackageListVersionsSuccess(packageVersions) {
+    this.set({ packageVersions });
+
+    this.emit(COSMOS_LIST_VERSIONS_CHANGE, name);
```
MatApple: `name` is undefined.

src/js/stores/CosmosPackagesStore.js
```diff
@@ -299,6 +325,18 @@ class CosmosPackagesStore extends GetSetBaseStore {
     this.emit(COSMOS_DESCRIBE_ERROR, error, name, version);
   }
 
+  processPackageListVersionsSuccess(packageVersions) {
+    this.set({ packageVersions });
```
MatApple: I think we should store `packageVersions` by `name` because there is a chance that when `getPackageVersions` is called, that is returns versions for a different package. It would be nice if in `getPackageVersions` you pass in the package name and if the versions exist, no API request is made. Currently, if you click around and select different packages, a request will be made every time. The way we store `packageDetails` is also unfortunate because this also should be stored by name.

MatApple:
@weblancaster this likely needs a rebase.

src/js/stores/CosmosPackagesStore.js
weblancaster: Its addressed, some parts of the same work is to address was added to https://github.com/dcos/dcos-ui/pull/2397

src/js/stores/CosmosPackagesStore.js
weblancaster: addressed

weblancaster:
retest this please

src/js/stores/CosmosPackagesStore.js
```diff
@@ -299,6 +325,24 @@ class CosmosPackagesStore extends GetSetBaseStore {
...
+      }
+    });
+
+    this.set({ packages });
+
+    this.emit(COSMOS_LIST_VERSIONS_CHANGE, packageName);
+  }
+
+  processPackageListVersionsError(error) {
+    this.set({ packageVersions: null });
```
MatApple: @weblancaster I'm not sure I follow why we set `packageVersions: null` on error, but on success we store:
```
packages: {
  foo: {
    1.0.1: 0,
    1.1.0: 1
  },
  bar: {}
}
```

src/js/stores/CosmosPackagesStore.js
```diff
@@ -263,6 +280,15 @@ class CosmosPackagesStore extends GetSetBaseStore {
     return null;
   }
 
+  getPackages() {
```
MatApple: Won't this return an instance of `UniversePackage` with properties that equal `packageNames` and values that equal the versions we've fetched for those `packageNames`?

ahoskins: I find it really strange it returns an single instance of `UniversePackage` with all the packages contained inside, instead of an array of `UniversePackage`s for each package.

bstavroulakis:
Alfred please retest this

src/js/stores/CosmosPackagesStore.js
weblancaster: That was a bad choice I made. When I weighted creating a new struct or reusing what `UniversePackages` struct I decided to reuse.

I'm addressing this creating a new struct and renaming a couple things. thanks.

src/js/stores/CosmosPackagesStore.js
weblancaster: I forgot to update the code after changing the code to cache all retrieved packages.

I have addressed to only be `null` for the failed package.

```
const packagesVersions = Object.assign({}, this.get("packagesVersions"), {
    [packageName]: null
});

this.set({ packagesVersions });
```

weblancaster:
retest this please

orlandohohmeier:
@MatApple please update you commit message to be inline with the contributing guidelines. Every commit messages should have a meaningful subject and the correct type. Messages like "edge cases" don't mean a thing.

orlandohohmeier:
package-describe-issue

The current implementation triggers a package describe loop resulting in many request. Please adjust!

src/js/pages/catalog/PackageDetailTab.js
```diff
@@ -78,15 +84,73 @@ class PackageDetailTab extends mixin(StoreMixin) {
...
+
+    this.retrievePackageInfo();
+  }
+
+  componentWillReceiveProps(nextProps) {
+    const isLoadingVersions =
+      CosmosPackagesStore.getPackageVersions(nextProps.params.packageName) ==
+      null;
+
+    this.setState({ isLoadingVersions });
```
orlandohohmeier: If possible please create a general `isLoading` helper that check both the versions as well as the details.

src/js/pages/catalog/PackageDetailTab.js
```diff
@@ -78,15 +84,73 @@ class PackageDetailTab extends mixin(StoreMixin) {
...
+  componentDidUpdate(prevProps) {
+    const { isLoadingVersions } = this.state;
+    const { packageName } = this.props.params;
+    const { version } = this.props.location.query;
+
+    // If packageName changed, fetch new data
+    if (prevProps.params.packageName !== packageName) {
+      this.retrievePackageInfo();
+
+      return false;
```
orlandohohmeier: Please always return the same type (check L143).

src/js/pages/catalog/PackageDetailTab.js
```diff
@@ -78,15 +84,73 @@ class PackageDetailTab extends mixin(StoreMixin) {
...
+      // Redirect to latest version for package
+      const packageVersions = CosmosPackagesStore.getPackageVersions(
+        this.props.params.packageName
+      );
+
+      if (!packageVersions) {
+        return;
+      }
+
+      this.redirectToPackageVersion(packageVersions.getVersions()[0]);
```
orlandohohmeier: Why do we need to redirect here?

src/js/pages/catalog/PackageDetailTab.js
```diff
@@ -78,15 +84,73 @@ class PackageDetailTab extends mixin(StoreMixin) {
...
+
+  componentWillReceiveProps(nextProps) {
+    const isLoadingVersions =
+      CosmosPackagesStore.getPackageVersions(nextProps.params.packageName) ==
+      null;
+
+    this.setState({ isLoadingVersions });
+  }
+
+  componentDidUpdate(prevProps) {
```
orlandohohmeier: I'm afraid I don't fully understand why we need all these cases. Please try to adjust the implementation to simplify the data flow.

MatApple:
@orlandohohmeier yes I will do - I pushed it last minute before I had to leave so I could follow up at home :) Will fix it up today.

src/js/pages/catalog/PackageDetailTab.js
```diff
@@ -107,6 +175,20 @@ class PackageDetailTab extends mixin(StoreMixin) {
     this.setState({ openInstallModal: true, advancedConfig: true });
   }
 
+  handlePackageVersionChange(selection) {
```
orlandohohmeier: This might be a stupid questions, but why do we need to "manually" redirect users? Can't we simply provide a list of links which will in turn trigger a rout update?

src/js/stores/CosmosPackagesStore.js
```diff
@@ -263,6 +281,15 @@ class CosmosPackagesStore extends GetSetBaseStore {
     return null;
   }
 
+  getPackageVersions(packageName) {
+    const packagesVersions = this.get("packagesVersions");
```
orlandohohmeier: I know that this method is following the pattern `getPackageDetails` is using to fetch and store data and thus I'm ok moving forward with the current implementation, but we must refactor this store! Storing and accessing the data in this way is going to create issues sooner than later as there is a very high chance that the data is mutated in unintended ways and there is no guarantee the versions match the package.

src/js/pages/catalog/PackageDetailTab.js
weblancaster: This is part of the lifecycle so whenever we have a change (update) we need this otherwise there is no way to know when to change fetch data.

src/js/pages/catalog/PackageDetailTab.js
weblancaster: We tried that but since we have no route context access the link did not work.

src/js/stores/CosmosPackagesStore.js
MatApple: @weblancaster and I talked about this because we recognize the store is a mess. This is one of the first stores I think dcos-ui had. Perfect example of where we should file an issue that we can follow up on, but the scope of adjustment shouldn't be included in this feature work 👍

src/js/pages/catalog/PackageDetailTab.js
MatApple: We must have a version in the URL for linking and state management. A user can navigate to `/catalog/packages/kafka` without a version and it will default to selecting the "latest" version of kafka. This redirect ensures we update the URL appropriately else we can be in an inconsistent state.

MatApple:
Alfred retest this please

MatApple:
Alfred retest this please

src/js/pages/catalog/PackageDetailTab.js
MatApple: We don't need to block the page for both requests when we only need to wait for one to render 99% of the contents.

src/js/pages/catalog/PackageDetailTab.js
orlandohohmeier: I'm not sure I understand the issue. Could you please elaborate?

src/js/pages/catalog/PackageDetailTab.js
orlandohohmeier: This should be part of the routing logic and shouldn't be handled within a view!

src/js/pages/catalog/PackageDetailTab.js
orlandohohmeier: I understand then case but would like to reduce the complexity.

src/js/events/CosmosPackagesActions.js
```diff
@@ -165,14 +165,15 @@ const CosmosPackagesActions = {
...
-          data: packageVersions
+          data: packageVersions,
+          packageName
         });
       },
       error(xhr) {
         AppDispatcher.handleServerAction({
           type: REQUEST_COSMOS_PACKAGE_LIST_VERSIONS_ERROR,
           data: RequestUtil.getErrorFromXHR(xhr),
-          xhr
```
orlandohohmeier: Are we sure that we don't need the xhr anymore? @MatApple didn't you add the native xhr to almost all actions to handle permission issues?

src/js/pages/catalog/PackageDetailTab.js
MatApple: @orlandohohmeier our dropdown's render outside of the React context using https://github.com/mesosphere/reactjs-components/blob/master/src/Portal/Portal.js which means a `<Link />` does not work.

src/js/pages/catalog/PackageDetailTab.js
```diff
@@ -78,21 +82,76 @@ class PackageDetailTab extends mixin(StoreMixin) {
...
+
+    const packageVersions = CosmosPackagesStore.getPackageVersions(packageName);
+
+    // Make sure version is in URL
+    if (
+      version == null &&
+      packageVersions &&
+      packageVersions.getVersions().length
+    ) {
+      this.redirectToPackageVersion(packageVersions.getVersions()[0]);
```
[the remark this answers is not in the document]
MatApple: I don't see a way of doing this within the routing logic but perhaps it just isn't needed.

src/js/events/CosmosPackagesActions.js
MatApple: We're not actually using the xhr object. The permissions error handling was never completed and I opposed the way we attempted to complete it.

MatApple: Added it here for consistency.

src/js/pages/catalog/PackageDetailTab.js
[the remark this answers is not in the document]
MatApple: Removed the logic

src/js/stores/CosmosPackagesStore.js
orlandohohmeier: @weblancaster @MatApple can one of you please file an issue?

orlandohohmeier:
Great Job! ⚒ 💯

PR #2397, opened 2017-08-28
feat(PackageDetailTab): [3 of 3] implement select package version

weblancaster:
## ⚠️   Depends on https://github.com/dcos/dcos-ui/pull/2395 and https://github.com/dcos/dcos-ui/pull/2396

This commit includes the work for epic DCOS_OSS-1528 where as a user now you
are able to select a version in package detail page.

How to test:
- Go to catalog page click on a package
- Notice now at the top where the version used to be displayed
- You will see a dropdown with the versions available to install
- When choosing a version the description should change to the chosen version
- When deployed the version selected will be installed

**before:**
screen shot 2017-08-21 at 2 55 25 pm

**after:**
screen shot 2017-08-28 at 3 18 40 pm

**Checklist**

### DIFF
--- a/src/js/pages/catalog/PackageDetailTab.js
+++ b/src/js/pages/catalog/PackageDetailTab.js
@@ -5,7 +5,7 @@ import { Link } from "react-router";
 import React from "react";
 /* eslint-enable no-unused-vars */
 import { StoreMixin } from "mesosphere-shared-reactjs";
-import { Dropdown } from "reactjs-components";
+import { Dropdown, Tooltip } from "reactjs-components";
 
 import { replaceQueryInPathString } from "#SRC/js/utils/RouterUtil";
 import BetaOptInUtil from "../../utils/BetaOptInUtil";
@@ -78,6 +78,13 @@ class PackageDetailTab extends mixin(StoreMixin) {
     METHODS_TO_BIND.forEach(method => {
       this[method] = this[method].bind(this);
     });
+  }
+
+  isSelectedVersionLoading() {
+    const { version } = this.props.location.query;
+    const cosmosPackage = CosmosPackagesStore.getPackageDetails();
+
+    return cosmosPackage.getVersion() !== version;
   }
 
   retrievePackageInfo() {
@@ -226,6 +233,8 @@ class PackageDetailTab extends mixin(StoreMixin) {
   }
 
   getInstallButtons(cosmosPackage) {
+    const tooltipContent = "loading selected version";
+
     if (cosmosPackage.isCLIOnly()) {
       return (
         <div>
@@ -247,18 +256,34 @@ class PackageDetailTab extends mixin(StoreMixin) {
 
     return (
       <div className="button-collection">
-        <button
-          className="button button-outline"
-          onClick={this.handleConfigureInstallModalOpen}
+        <Tooltip
+          wrapperClassName="button-group"
+          wrapText={true}
+          content={tooltipContent}
+          suppress={!this.isSelectedVersionLoading()}
         >
-          Configure
-        </button>
-        <button
-          className="button button-primary"
-          onClick={this.handleInstallModalOpen}
+          <button
+            disabled={this.isSelectedVersionLoading()}
+            className="button button-outline"
+            onClick={this.handleConfigureInstallModalOpen}
+          >
+            Configure
+          </button>
+        </Tooltip>
+        <Tooltip
+          wrapperClassName="button-group"
+          wrapText={true}
+          content={tooltipContent}
+          suppress={!this.isSelectedVersionLoading()}
         >
-          Deploy
-        </button>
+          <button
+            disabled={this.isSelectedVersionLoading()}
+            className="button button-primary"
+            onClick={this.handleInstallModalOpen}
+          >
+            Deploy
+          </button>
+        </Tooltip>
       </div>
     );
   }
@@ -313,6 +338,15 @@ class PackageDetailTab extends mixin(StoreMixin) {
         transition={true}
         wrapperClassName="dropdown"
       />
+    );
+  }
+
+  getPackageDescription(definition, cosmosPackage) {
+    return (
+      <div className="pod flush-horizontal flush-bottom">
+        {this.getItems(definition, this.getItem)}
+        <ImageViewer images={cosmosPackage.getScreenshots()} />
+      </div>
     );
   }
 
@@ -406,10 +440,9 @@ class PackageDetailTab extends mixin(StoreMixin) {
               </div>
             </div>
           </div>
-          <div className="pod flush-horizontal flush-bottom">
-            {this.getItems(definition, this.getItem)}
-            <ImageViewer images={cosmosPackage.getScreenshots()} />
-          </div>
+          {this.isSelectedVersionLoading()
+            ? this.getLoadingScreen()
+            : this.getPackageDescription(definition, cosmosPackage)}
         </div>
         <InstallPackageModal
           open={state.openInstallModal}
--- a/src/js/pages/catalog/PackageDetailTab.js
+++ b/src/js/pages/catalog/PackageDetailTab.js
@@ -296,7 +296,7 @@ class PackageDetailTab extends mixin(StoreMixin) {
   getPackageVersionsDropdown() {
     const cosmosPackage = CosmosPackagesStore.getPackageDetails();
     const cosmosPackageVersions = CosmosPackagesStore.getPackageVersions();
-    const selectedVersion = cosmosPackage.getCurrentVersion();
+    const selectedVersion = cosmosPackage.getVersion();
     const packageVersions = cosmosPackageVersions.getAllVersions();
     const formattedPackageVersions = formatPackageVersions(packageVersions);
 
--- a/src/js/pages/catalog/PackageDetailTab.js
+++ b/src/js/pages/catalog/PackageDetailTab.js
@@ -52,7 +52,12 @@ const METHODS_TO_BIND = [
   "handlePackageVersionChange"
 ];
 
-const formatPackageVersions = versions => {
+const formatPackageVersions = selectedPackage => {
+  if (selectedPackage.packageVersions == null) {
+    return [];
+  }
+  const versions = selectedPackage.packageVersions;
+
   return Object.keys(versions)
     .sort((a, b) => {
       if (+versions[a] > +versions[b]) {
@@ -103,9 +108,14 @@ class PackageDetailTab extends mixin(StoreMixin) {
   retrievePackageInfo() {
     const { packageName } = this.props.params;
     const { version } = this.props.location.query;
+    const cosmosPackages = CosmosPackagesStore.getPackages();
+    const selectedPackage = cosmosPackages.getPackageByName(packageName);
 
-    // fetch package versions available
-    CosmosPackagesStore.fetchPackageVersions(packageName);
+    // fetch package versions available only if not cached
+    if (Object.keys(selectedPackage).length < 1) {
+      CosmosPackagesStore.fetchPackageVersions(packageName);
+    }
+
     // Fetch package description
     CosmosPackagesStore.fetchPackageDescription(packageName, version);
   }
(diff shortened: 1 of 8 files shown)


src/js/pages/catalog/PackageDetailTab.js
```diff
@@ -249,6 +285,35 @@ class PackageDetailTab extends mixin(StoreMixin) {
     }
   }
 
+  handlePackageVersionChange(packageOpt) {
+    const packageVersion = packageOpt.id;
+    const originalPath = global.location.hash.split("?")[0];
+    const selectedVersion = `${originalPath}?version=${packageVersion}`;
+
+    global.location.replace(selectedVersion);
```
MatApple: Could we use react router tools to do this?
```
// To set a parameter:
<Link to={{
    pathname: router.location.pathname,
    query: {
        ...(router.location.query),
        myProp: 'foo',
    }
 }} />
```

weblancaster: `<Link />` returns a component which expects a click, so in that case doesn't work for our case. https://github.com/ReactTraining/react-router/blob/master/packages/react-router-dom/modules/Link.js

MatApple: Ok, maybe we can do the same as this https://github.com/dcos/dcos-ui/blob/c693793411634bbd23e452648cb5f33dfdb20852/src/js/pages/catalog/PackagesTab.js#L95

src/js/pages/catalog/PackageDetailTab.js
```diff
@@ -249,6 +285,35 @@ class PackageDetailTab extends mixin(StoreMixin) {
     }
   }
 
+  handlePackageVersionChange(packageOpt) {
```
ahoskins: This seems like the kind of thing we'd have a util for? It's just a function that sets a value from the query string.

src/js/pages/catalog/PackageDetailTab.js
```diff
@@ -78,15 +99,31 @@ class PackageDetailTab extends mixin(StoreMixin) {
...
     // Fetch package description
     CosmosPackagesStore.fetchPackageDescription(packageName, version);
   }
 
+  componentDidMount() {
+    super.componentDidMount(...arguments);
+    this.retrievePackageInfo();
+  }
+
+  componentDidUpdate(prevProps) {
```
ahoskins: Why add this indirection between selecting a version on the dropdown and fetching new package info? I think it's more clear to just directly call `this.retrievePackageInfo()` after updating the query string.

src/js/pages/catalog/PackageDetailTab.js
```diff
@@ -62,14 +84,13 @@ class PackageDetailTab extends mixin(StoreMixin) {
...
-        },
-        listenAlways: false,
-        suppressUpdate: true
+        events: [
+          "descriptionError",
+          "descriptionSuccess",
+          "listVersionsSuccess",
+          "listVersionsError"
+        ],
+        suppressUpdate: false
```
ahoskins: I'm curious why this got changed? The `suppressUpdate` boolean is now false, `unmountWhen()` and `listenAlways` is removed.

src/js/pages/catalog/PackageDetailTab.js
weblancaster: I argue against a util because promotes reusability and we want this being used only for this component.

ahoskins: I'm more just surprised that there wouldn't be somewhere else in the codebase we do this exact same type of thing.

src/js/pages/catalog/PackageDetailTab.js
weblancaster: because this is part of the React lifecycle, no need to call directly since the mechanism is there already

ahoskins: I know this works, I'm just saying that `componentDidUpdate` is being introduced only for this very specific url hash update...so I don't see why we would add more abstraction by using React's lifecycle for this.

weblancaster: discussed offline

src/js/pages/catalog/PackageDetailTab.js
```diff
@@ -257,12 +322,12 @@ class PackageDetailTab extends mixin(StoreMixin) {
     }
 
     const cosmosPackage = CosmosPackagesStore.getPackageDetails();
-    if (state.isLoading || !cosmosPackage) {
+    const cosmosPackageVersions = CosmosPackagesStore.getPackageVersions();
+    if (state.isLoading || !cosmosPackage || !cosmosPackageVersions) {
```
MatApple: I wonder if we could avoid blocking the entire page and progressively load this, especially since the `cosmosPackageVersions` are only required/shown once you click the dropdown.

src/js/pages/catalog/PackageDetailTab.js
weblancaster: `suppressUpdate` will force to update React when false.
`unmountWhen` it's called when an event happen.
`listenAlways` remove listener on the store change.

src/js/pages/catalog/PackageDetailTab.js
weblancaster: It's addressed @MatApple

src/js/structs/UniversePackage.js
```diff
@@ -84,7 +84,9 @@ class UniversePackage extends Item {
   }
 
   getVersion() {
-    return this.get("version");
+    // backport because some actions are removing
+    // version and replacing the package obj with currentVersion
+    return this.get("version") || this.get("currentVersion");
```
ahoskins: Maybe it's better to change the places that use `.currentVersion` instead of having this inconsistency.

src/js/utils/RouterUtil.js
```diff
@@ -18,6 +18,28 @@ function findRedirect(queryString) {
...
+   *    query: "version",
+   *    value: "test"
+   * }
+   *
+   * Output: http://localhost:4200/#/catalog/packages/arangodb3?version=test&_k=ed535h
+   *
+   * @param {Object} options object config
+   * @returns {String} string with new query value
+   */
+  replaceQueryInPathString(options) {
```
ahoskins: There is also the case when a query string has no value. Example: `?key=` or `?key`. This is probably not an expected case (but is valid and can happen). Since this is a general util we may want to make this function more general to handle all kinds of keys, but just keys with values.

src/js/utils/__tests__/RouterUtil-test.js
```diff
@@ -208,4 +208,33 @@ describe("RouterUtil", function() {
       expect(RouterUtil.getQueryStringInUrl()).toEqual(expectedResult);
     });
   });
+
+  describe("#replaceQueryInPathString", function() {
```
ahoskins: Or keep the method as it is, and add a test that shows what happens when calling on a url with a key with no value like:
`http://localhost:4200/#/catalog/packages/arangodb3?version=` (expect no-op)

src/js/utils/RouterUtil.js
weblancaster: This method was created to only replace the value when existent nothing else, but I'm open to add that if more people feel like that will happen.

I rather see a unit test confirming this was wanted.

src/js/pages/catalog/PackageDetailTab.js
```diff
@@ -46,9 +48,35 @@ const PackageDetailBreadcrumbs = ({ cosmosPackage }) => {
...
+
+  return Object.keys(versions)
+    .sort((a, b) => {
+      if (+versions[a] > +versions[b]) {
+        return 1;
+      }
+      if (+versions[a] < +versions[b]) {
+        return -1;
+      }
+
```
bstavroulakis: I think it would make more sense to reverse these
```
      if (+versions[a] > +versions[b]) {
        return -1;
      }
      if (+versions[a] < +versions[b]) {
        return 1;
      }
```
I'm used to previewing versions from latest to oldest. I think it makes more sense logically as the user will probably look for the latest version and it is more standard to promote and bubble up the latest.

image

src/js/pages/catalog/PackageDetailTab.js
```diff
@@ -103,9 +108,14 @@ class PackageDetailTab extends mixin(StoreMixin) {
   retrievePackageInfo() {
     const { packageName } = this.props.params;
     const { version } = this.props.location.query;
+    const cosmosPackages = CosmosPackagesStore.getPackages();
+    const selectedPackage = cosmosPackages.getPackageByName(packageName);
+
+    // fetch package versions available only if not cached
+    if (Object.keys(selectedPackage).length < 1) {
```
ahoskins: Nit: `Object.keys(selectedPackage).length === 0` is clearer for checking empty object.

weblancaster: This is minor to be a concern and block the PR because `Object.keys(selectedPackage).length < 1` is clear and not really different than `Object.keys(selectedPackage).length === 0`

ahoskins: That's why I left as a comment and not a request 😄

src/js/pages/catalog/PackageDetailTab.js
weblancaster: Agree, let me just confirm with design team

src/js/utils/RouterUtil.js
weblancaster: @ahoskins have we decided what to do here?

src/js/structs/UniversePackage.js
weblancaster: that would be a refactoring and I would need to test other parts besides only the feat implemented here.. should be added on another Jira/tasks so for that reason I included both as of now.

src/js/structs/UniversePackagesVersions.js
```diff
@@ -0,0 +1,27 @@
...
+    if (+versions[a] < +versions[b]) {
+      return 1;
+    }
+
+    return 0;
+  });
+};
+
+class UniversePackagesVersions extends Item {
+  getPackageByName(packageName) {
```
ahoskins: The thing that this method is "getting" is an sorted array of package versions, but it's named like it's getting a a package. I think something like `getVersionsForPackage()` is much better.

src/js/pages/catalog/PackageDetailTab.js
```diff
@@ -310,18 +287,28 @@ class PackageDetailTab extends mixin(StoreMixin) {
   getPackageVersionsDropdown() {
     const cosmosPackage = CosmosPackagesStore.getPackageDetails();
     const packageName = cosmosPackage.getName();
-    const cosmosPackages = CosmosPackagesStore.getPackages();
+    const cosmosPackagesVersions = CosmosPackagesStore.getPackagesVersions();
```
ahoskins: It is possible for the `cosmosPackagesVersions` to be null and it's accessed later without checking or handling this.

src/js/structs/UniversePackagesVersions.js
weblancaster: Was a design decision to have the sort by default.

I believe `getPackageByName` is more clear than `getVersionsForPackage` since that's what it's doing, but I'm happy to change if that's a blocker for you.

src/js/pages/catalog/PackageDetailTab.js
weblancaster: I don't understand what you trying to say here.

src/js/structs/UniversePackagesVersions.js
ahoskins: Talked offline: will change to `getPackageVersionsByName`

src/js/pages/catalog/PackageDetailTab.js
ahoskins: Talked offline: will not return `UniversePackagesVersions({})` from `getPackagesVersions()` when packagesVersions field is undefined.

src/js/utils/RouterUtil.js
ahoskins: Talked offline: we will add more unit tests as commented below, and leave this method as-is.

src/js/pages/catalog/PackageDetailTab.js
```diff
@@ -80,6 +80,13 @@ class PackageDetailTab extends mixin(StoreMixin) {
     });
   }
 
+  isSelectedVersionLoading() {
```
weblancaster: Instead of using `setState` which will trigger lifecycle and re-render.

src/js/pages/catalog/PackageDetailTab.js
```diff
@@ -226,6 +233,8 @@ class PackageDetailTab extends mixin(StoreMixin) {
   }
 
   getInstallButtons(cosmosPackage) {
+    const tooltipContent = "loading selected version";
```
weblancaster: Revised by @leemunroe

MatApple:
Closing this as we rebased this onto the branch for https://github.com/dcos/dcos-ui/pull/2396 for quicker edits.

PR #2399, opened 2017-08-31
DCOS_OSS-1582: adjust Marathon RAML install the version

orlandohohmeier:
Adjust the Marathon RAML URL pattern and update the version to fetch the latest Marathon 1.5 API Spec and ensure that the UI is compatible with
Marathon.

The latest spec includes `unreachableStrategy` adjustments that if not applied prevents users from creating/editings apps that are configured to expunge immediately or flag apps as inactive.

Closes DCOS_OSS-1582

### DIFF
--- a/scripts/utils/marathon
+++ b/scripts/utils/marathon
@@ -29,8 +29,7 @@ function install_marathon_raml() {
 
   # Download and expand on-the-fly the marathon tarball into a temporary folder
   header "Downloading marathon ${VERSION} archive"
-
-  local URL="https://downloads.mesosphere.io/marathon/snapshots/marathon-docs-${VERSION}.tgz"
+  local URL="https://downloads.mesosphere.com/marathon/snapshots/marathon-raml-${VERSION}.tgz"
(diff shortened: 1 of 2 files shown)


scripts/utils/marathon
```diff
@@ -29,8 +29,7 @@ function install_marathon_raml() {
 
   # Download and expand on-the-fly the marathon tarball into a temporary folder
   header "Downloading marathon ${VERSION} archive"
-
-  local URL="https://downloads.mesosphere.io/marathon/snapshots/marathon-docs-${VERSION}.tgz"
+  local URL="https://downloads.mesosphere.io/marathon/snapshots/marathon-raml-${VERSION}.tgz"
```
Poltergeist: Let's use `mesosphere.com` here. As this is the right TLD.

Poltergeist:
I would still like to see the TLD changed but I would not block  the PR for this. The Validation does work now.

orlandohohmeier:
@Poltergeist  @nLight  =Changed the TLD. Pleas re-review.

PR #2401, opened 2017-08-31
fix(GeneralServiceFormSection): rename Mesos runtime -> UCR

ahoskins:
In the new marathon service create form, instead of calling the checkbox Mesos Runtime, always call it Universal Container Runtime (UCR).

**Before**:
screen shot 2017-08-31 at 3 04 43 pm

**After**:
screen shot 2017-08-31 at 3 03 32 pm

Closes DCOS-18312.

**Checklist**

### DIFF
--- a/plugins/services/src/js/constants/ContainerConstants.js
+++ b/plugins/services/src/js/constants/ContainerConstants.js
@@ -5,6 +5,6 @@ module.exports = {
   },
   labelMap: {
     DOCKER: "Docker Engine",
-    MESOS: "Mesos Runtime"
+    MESOS: "Universal Container Runtime (UCR)"
   }
 };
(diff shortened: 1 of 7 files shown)


orlandohohmeier:
@ahoskins please adjust the system test as well

plugins/services/src/js/constants/ContainerConstants.js
```diff
@@ -5,6 +5,6 @@ module.exports = {
   },
   labelMap: {
     DOCKER: "Docker Engine",
-    MESOS: "Mesos Runtime"
+    MESOS: "Universal Container Runtime (UCR)"
```
bstavroulakis: Since we use this in so many places. Could we probably use a template literal when we use this?
E.x. Docker Runtime only supports the default size for implicit volumes, please select ${Container.MESOS} if you want to modify the size.

I can see @weblancaster not agreeing with this a mile away :)

ahoskins: I agree that refactoring to use template strings is a more maintainable approach. When I made this simple string change, it broke tests across many many files. Our cypress tests are written in a very "always repeat yourself" way. But I think in this case, since this string constant already exists, we should use the constant in a template string.

ahoskins:
run tests

ahoskins:
run alfred

ahoskins:
Alfred retest this please

PR #2402, opened 2017-09-01
Uninstall 01 | DCOS-18006: Uninstall Modal for Frameworks and SDK services

bstavroulakis:
## Uninstall Modal for Frameworks and SDK services

Until now when the user tried to delete a Framework or SDK Service from the UI we displayed a message that was referencing the CLI commands.

With this PR, we can delete frameworks and SDK Services from the UI.

Before:
image
After:
image

To test, please try to delete a Framework/ SDK Service either from the details or the table view page.

### Case 1: Group has frameworks 

When we delete a group the standard services are correctly deleted as well. However, Cosmos and Marathon don't communicate with each other, so **if we have Framework/ SDK Services in the group these aren't correctly uninstalled**.
Solution: We show a popup that lists the frameworks running in the group so the user will have to first uninstall them separately before deleting the group.

screen shot 2017-09-12 at 4 01 41 pm

### Case 2: Empty Group

If the group doesn't have services we can delete it like before

### Case 3: Group with non-framework services

If the group doesn't have framework services we can delete it but with a force delete

image

**Checklist**


bstavroulakis:
Please retest this

bstavroulakis:
Alfred please retest this

plugins/services/src/js/events/MarathonActions.js
```diff
@@ -71,30 +134,37 @@ var MarathonActions = {
...
-    RequestUtil.json({
-      url,
-      method: "DELETE",
-      success() {
-        AppDispatcher.handleServerAction({
-          type: REQUEST_MARATHON_GROUP_DELETE_SUCCESS
-        });
-      },
-      error(xhr) {
+    const deleteSDKServices = groupFrameworks.map(function(framework) {
```
ahoskins: Isn't it possible that some of these `framework` are not SDK services?

plugins/services/src/js/events/MarathonActions.js
```diff
@@ -50,6 +58,61 @@ function buildURI(path) {
...
+
+const handleSDKDelete = function(
+  payload,
+  frameworkId,
+  dispatcher,
+  resolve,
+  reject
+) {
+  const { action, xhr } = payload;
+  const eventIsForThisFramework = action && action.appId
```
ahoskins: I'm pretty sure it's logically equivalent as `const eventIsForThisFramework = action && action.appId === frameworkId`

plugins/services/src/js/events/MarathonActions.js
```diff
@@ -50,6 +58,61 @@ function buildURI(path) {
...
+    ? action.appId === frameworkId
+    : false;
+  if (!eventIsForThisFramework) return;
+  if (action.type === REQUEST_COSMOS_PACKAGE_UNINSTALL_SUCCESS) {
+    AppDispatcher.unregister(dispatcher);
+    resolve();
+  } else if (action.type === REQUEST_COSMOS_PACKAGE_UNINSTALL_ERROR) {
+    AppDispatcher.unregister(dispatcher);
+    if (xhr && xhr.response) {
+      var errorMessage = JSON.parse(xhr.response);
```
ahoskins: `let` 🙃

plugins/services/src/js/events/MarathonActions.js
```diff
@@ -50,6 +58,61 @@ function buildURI(path) {
...
+      if (errorMessage && errorMessage.message) {
+        reject(errorMessage);
```
ahoskins: It's probably a good idea to `return` here to prevent continued execution of this function.

bstavroulakis:
Alfred please retest this

nLight:
@bstavroulakis cross posting my comment from the ticket:

After you uninstall i.e. beta-confluent-kafka the API returns a message:

```
Apache Kafka by Confluent has been uninstalled.
Please follow the instructions at https://docs.mesosphere.com/current/usage/service-guides/kafka/uninstall to remove any persistent state if required.
```

and if you go to that page there's an instruction what to do afterwards. **Shall we show that message to the user**?

To back this point, after I uninstalled beta-confluent-kafta and went to exhibitor and kafka's stuff is still there in zookeeper, though I was able to re-instal it afterwards.

nLight:
Alfred please retest this

nLight:
I think those Alfred failures are legit, 4 tests are keep failing.

plugins/services/src/js/events/ServiceDeleteActions.js
```diff
@@ -0,0 +1,106 @@
...
+        reject(errorMessage.message);
+
+        return;
+      }
+    }
+    reject("One or more services in group weren't deleted.");
+  }
+};
+
+function createFrameworkDeletePromise(serviceId, serviceName) {
```
orlandohohmeier: Please rename this function to reflect that it actually uninstalls a package.

orlandohohmeier:
We need to decide wether we'er ok with exposing promises or stick to the Flux pattern for now, before we move forward with this PR. /cc @dcos/frontend-foundation

bstavroulakis:
Alfred please retest this

src/js/events/CosmosPackagesActions.js
```diff
@@ -205,19 +205,18 @@ const CosmosPackagesActions = {
     });
   },
 
-  uninstallPackage(packageName, packageVersion, appId, all = false) {
+  uninstallPackage(packageName, appId, all = false) {
```
ahoskins: Nice catch removing this.

plugins/services/src/js/events/ServiceDeleteActions.js
```diff
@@ -0,0 +1,106 @@
...
+    .catch(function(error) {
+      AppDispatcher.handleServerAction({
+        type: REQUEST_MARATHON_SERVICE_DELETE_ERROR,
+        data: error
+      });
+    });
+}
+
+const ServiceDeleteActions = {
+  // This part is responsible for uninstalling the frameworks before deleting
```
ahoskins: Maybe add a note/link on why it's disabled for now. And the conditions in which it could be enabled? Or remove this commented out code all together?

bstavroulakis: I removed this altogether

bstavroulakis:
Alfred test this

bstavroulakis:
Hey Alfredo cappuccino, retest this.

bstavroulakis:
Alfredo cappuccino, retest this.

plugins/services/src/js/components/modals/ServiceDestroyModal.js
```diff
@@ -146,51 +162,27 @@ class ServiceDestroyModal extends React.Component {
...
-            })}
-          >
-            {intl.formatMessage({
-              id: "COMMON.DOCUMENTATION"
-            })}
-          </a>
-          {intl.formatMessage({
-            id: "SERVICE_ACTIONS.DELETE_SERVICE_FRAMEWORK_2"
-          })}
+          This group needs to be empty to delete it. Please delete any services in the group first.
```
ahoskins: This is a bit confusing/misleading. It's not true that it needs to be empty to delete it. It's only true that it needs no SDK service to delete it. Maybe change to:

"Cannot delete group containing SDK services. Please delete these services before deleting the group."

MatApple: Actually, we want to keep it generic like this so we don't have to call out specific types of services that you can/cannot delete etc.

ahoskins: Hm, okay fair enough. 👍

bstavroulakis:
Alfreddo cappuccino test this please

plugins/services/src/js/components/modals/ServiceDestroyModal.js
bstavroulakis: Sounds good, @ahoskins if you could remove the block and/ or approve the PR if we don't have any more issues, that would be greeeeat!

plugins/services/src/js/components/modals/GroupContainsFrameworksModal.js
```diff
@@ -0,0 +1,37 @@
...
+      open={props.isOpen}
+      showFooter={true}
+      showHeader={true}
+    >
+      <div className="modal-service-delete-center">
+        This group needs to be empty to delete it. Please delete any services in the group first.
+      </div>
+    </Modal>
+  );
+};
```
MatApple: Please add propTypes and sensible defaults.

plugins/services/src/js/components/modals/ServiceModals.js
```diff
@@ -11,6 +11,7 @@ import ServiceResumeModal from "./ServiceResumeModal";
 import ServiceScaleFormModal from "./ServiceScaleFormModal";
 import ServiceSpecUtil from "../../utils/ServiceSpecUtil";
 import ServiceSuspendModal from "./ServiceSuspendModal";
+import GroupContainsFrameworksModal from "./GroupContainsFrameworksModal";
```
MatApple: Please consider how we could rename this to prevent binding to the reason it's invoked. Perhaps `DisabledGroupDestroyModal`.

plugins/services/src/js/components/modals/ServiceModals.js
```diff
@@ -39,6 +40,19 @@ class ServiceModals extends React.Component {
     );
   }
 
+  getCannotDestroyGroupModal(groupFrameworks) {
```
MatApple: Also please align this method with the revised name for the modal :)

plugins/services/src/js/containers/services/ServicesContainer.js
```diff
@@ -424,6 +436,10 @@ class ServicesContainer extends React.Component {
 
     if (!modalProps.service) {
       modalProps.service = service;
+    } else {
+      modalProps.service = DCOSStore.serviceTree.findItemById(
```
MatApple: Let's comment why this is needed.

plugins/services/src/js/containers/services/ServicesContainer.js
```diff
@@ -438,7 +454,7 @@ class ServicesContainer extends React.Component {
     );
   }
 
-  getEmpyPage(itemType) {
+  getEmptyPage(itemType) {
```
MatApple: 👍

plugins/services/src/js/events/ServiceDeleteActions.js
```diff
@@ -0,0 +1,23 @@
+import CosmosPackagesActions from "#SRC/js/events/CosmosPackagesActions";
+import MarathonActions from "./MarathonActions";
+import Framework from "../structs/Framework";
+
+const ServiceDeleteActions = {
```
MatApple: Please add tests for this file!

src/js/translations/en-US.json
```diff
@@ -38,41 +38,58 @@
   "SERVICE_ACTIONS.SCALE_BY": "Scale By",
   "SERVICE_ACTIONS.SUSPEND": "Suspend",
 
-  "SERVICE_ACTIONS.SDK_GROUP_DELETE_1": "Copy and paste this command into the CLI",
+  "SERVICE_ACTIONS.SDK_GROUP_DELETE_1":
```
MatApple: Let's remove the things we no longer need.

plugins/services/src/js/components/modals/GroupContainsFrameworksModal.js
```diff
@@ -0,0 +1,37 @@
+import React from "react";
+import { Modal } from "reactjs-components";
+import ModalHeading from "#SRC/js/components/modals/ModalHeading";
+
+const GroupContainsFrameworksModal = props => {
```
MatApple: Please see other comment about renaming this component.

plugins/services/src/js/components/modals/ServiceModals.js
```diff
@@ -39,6 +40,19 @@ class ServiceModals extends React.Component {
...
+    const { onClose, modalProps } = this.props;
+    const key = ActionKeys.GROUP_DELETE;
+
+    return (
+      <GroupContainsFrameworksModal
+        services={groupFrameworks}
```
MatApple: Looks like we don't use the `services` prop in this component so let's remove.

plugins/services/src/js/components/modals/ServiceDestroyModal.js
```diff
@@ -101,6 +111,12 @@ class ServiceDestroyModal extends React.Component {
     });
   }
 
+  handleChangeInputForceDeleteGroupWithServices(event) {
```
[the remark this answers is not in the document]
bstavroulakis: Changed to handleChangeInputGroupDestroy

src/js/translations/en-US.json
bstavroulakis: That is currently being used. I've removed the strings not used.

plugins/services/src/js/components/modals/ServiceModals.js
bstavroulakis: removed

plugins/services/src/js/components/modals/ServiceModals.js
bstavroulakis: updated

plugins/services/src/js/components/modals/ServiceModals.js
bstavroulakis: Updated the name

plugins/services/src/js/components/modals/GroupContainsFrameworksModal.js
bstavroulakis: Changed the name

plugins/services/src/js/components/modals/GroupContainsFrameworksModal.js
bstavroulakis: Added

plugins/services/src/js/containers/services/ServicesContainer.js
bstavroulakis: comment added

plugins/services/src/js/events/ServiceDeleteActions.js
bstavroulakis: tests locked and loaded!

bstavroulakis:
Alfred test this

plugins/services/src/js/components/modals/ServiceDestroyModal.js
```diff
@@ -146,51 +162,27 @@ class ServiceDestroyModal extends React.Component {
     );
   }
 
-  getDestroyFrameworkModal() {
-    const { open, service, intl } = this.props;
-    const packageName = service.getPackageName();
+  getGroupHeader() {
+    if (!this.isGroupWithServices()) {
+      return;
```
weblancaster: please return `null` for consistency and best practices suggested by React.

plugins/services/src/js/components/modals/ServiceDestroyModal.js
```diff
@@ -213,27 +205,39 @@ class ServiceDestroyModal extends React.Component {
...
-          className="form-control filter-input-text"
-          onChange={this.handleChangeInputFieldDestroy}
-          type="text"
-          value={this.state.serviceNameConfirmationValue}
-          autoFocus
-        />
+        {this.getGroupHeader()}
+        {(!this.isGroupWithServices() ||
+          this.state.forceDeleteGroupWithServices) &&
+          <div className="modal-service-delete-center">
```
weblancaster: As discussed I was confused to see the markup part of the condition test, even though works @bstavroulakis agreed to change to not cause confusion.

bstavroulakis: @weblancaster Addressed issues, that you for reviewing this

plugins/services/src/js/components/modals/ServiceDestroyModal.js
```diff
@@ -39,6 +40,12 @@ class ServiceDestroyModal extends React.Component {
     });
   }
 
+  isGroupWithServices() {
+    const { service } = this.props;
+
+    return service instanceof ServiceTree && service.list.length > 0;
```
orlandohohmeier: Let's please add a `length` getter to the `Tree` resp. `List` struct instead of accessing the internals here.

plugins/services/src/js/components/modals/ServiceDestroyModal.js
```diff
@@ -39,6 +40,12 @@ class ServiceDestroyModal extends React.Component {
     });
   }
 
+  isGroupWithServices() {
```
orlandohohmeier: Please move this helper to a dedicated util or add a meaning full getter to the `Tree` struct.

plugins/services/src/js/components/modals/ServiceModals.js
```diff
@@ -49,8 +62,15 @@ class ServiceModals extends React.Component {
     let key = ActionKeys.SERVICE_DELETE;
 
     if (isGroup) {
-      deleteItem = force =>
-        this.props.actions.deleteGroup(service.getId(), force);
+      const groupFrameworks = "getFrameworks" in service
```
orlandohohmeier: Why do we need to check if `service` has a `getFrameworks` method? Please adjust the implication to make so that checks like this one are not needed.

plugins/services/src/js/events/ServiceDeleteActions.js
orlandohohmeier: Why do we add service **delete** actions? Let's please stick with the pattern and add general service actions.

plugins/services/src/js/components/modals/ServiceDestroyModal.js
bstavroulakis: Created a new util named ServiceTreeUtil.js. First I added this method in the ServiceUtil.js but it was creating a loop because ServiceTree.js included the ServiceUtil.js and if the ServiceUtil would include ServiceTree.js it would reference ServiceTree to {} to break the recursion.

plugins/services/src/js/components/modals/ServiceDestroyModal.js
bstavroulakis: Added in the tree

plugins/services/src/js/components/modals/ServiceModals.js
bstavroulakis: Removed, that was added just to be sure but yes the isGroup check already makes sure that is there.

plugins/services/src/js/events/ServiceDeleteActions.js
bstavroulakis: @orlandohohmeier renamed to serviceActions

plugins/services/src/js/utils/ServiceTreeUtil.js
```diff
@@ -0,0 +1,13 @@
+import ServiceTree from "../structs/ServiceTree";
+
+const ServiceTreeUtil = {
+  isGroupWithServices(service) {
+    if (service instanceof ServiceTree && service.getListCount() > 0) {
```
weblancaster: not big deal but can also be written as `return service instanceof ServiceTree && service.getListCount() > 0`

plugins/services/src/js/structs/__tests__/ServiceTree-test.js
```diff
@@ -997,4 +997,41 @@ describe("ServiceTree", function() {
...
+            }
+          },
+          new Application({ id: "a" }),
+          new Framework({ id: "b" })
+        ]
+      });
+      expect(this.instance.getListCount()).toEqual(6);
+    });
+
+    it("returns 0 for ServiceTree with no list items", function() {
```
weblancaster: Wondering if we should also add a unit to test when is greater than `0`

PR #2404, opened 2017-09-06
feat(CosmosPackageActions): implement service/describe action

ahoskins:
**Epic**: Actions for [Edit Configuration Settings](https://jira.mesosphere.com/browse/DCOS-17721).

**Summary**: Adds support for `service/describe` endpoint, used for getting information of a running cosmos service. This is very similar to the `package/describe`, which is used to get package data in catalog. 

**Quick description**: The full new endpoint path is `/cosmos/service/describe`. This is different than the rest of the Cosmos API that is under the `/package` root path. The `contentType` and `Accept` header are also slightly different than the rest of the cosmos API. The endpoint accepts a body that looks like this: `{  appId: /hdfs }`. 

Closes DCOS-18199. 

**Checklist**
- [x] Did you add a JIRA issue in a commit message or as part of the branch name?
- [x] Did you add new unit tests?
- [ ] Did you add new integration tests?
- [ ] If this is a regression, did you write a test to catch this in the future?

PR #2405, opened 2017-09-06
feat(CosmosPackagesStore): implement service/describe in store

ahoskins:
~⚠️  *Depends on #2404 (provides the actions)*~

**Epic**: Store support for [Edit Configuration Settings](https://jira.mesosphere.com/browse/DCOS-17721).

**Summary**: Adds support to cosmos store for `service/describe` endpoint.

**Quick description**: expose a new method `.fetchServiceDescription()` on the cosmos packages store. A successful result of this will be just like `.fetchPackageDescription()` - it will set the `packageDetails` field in the store. The component (not built yet) can then use this the populate its form.
  
Closes DCOS-18200. 

**Checklist**


src/js/stores/CosmosPackagesStore.js
```diff
@@ -78,7 +82,10 @@ class CosmosPackagesStore extends GetSetBaseStore {
         repositoryAddSuccess: COSMOS_REPOSITORY_ADD_SUCCESS,
         repositoryAddError: COSMOS_REPOSITORY_ADD_ERROR,
         repositoryDeleteSuccess: COSMOS_REPOSITORY_DELETE_SUCCESS,
-        repositoryDeleteError: COSMOS_REPOSITORY_DELETE_ERROR
+        repositoryDeleteError: COSMOS_REPOSITORY_DELETE_ERROR,
+
+        cosmosServiceDescriptionSuccess: COSMOS_SERVICE_DESCRIBE_CHANGE,
```
ahoskins: I will rename this. With this name, the StoreMixin names the callback `onCosmosPackagesStoreCosmosServiceDescriptionSuccess()`...which is incredibly wordy.

ahoskins:
Alfred please test this

src/js/stores/CosmosPackagesStore.js
```diff
@@ -87,7 +91,10 @@ class CosmosPackagesStore extends GetSetBaseStore {
         repositoryAddSuccess: COSMOS_REPOSITORY_ADD_SUCCESS,
         repositoryAddError: COSMOS_REPOSITORY_ADD_ERROR,
         repositoryDeleteSuccess: COSMOS_REPOSITORY_DELETE_SUCCESS,
-        repositoryDeleteError: COSMOS_REPOSITORY_DELETE_ERROR
+        repositoryDeleteError: COSMOS_REPOSITORY_DELETE_ERROR,
+
+        serviceDescriptionSuccess: COSMOS_SERVICE_DESCRIBE_CHANGE,
```
weblancaster: not big deal but based on the other constants shouldn't this be `COSMOS_SERVICE_DESCRIBE_SUCCESS`?

src/js/stores/CosmosPackagesStore.js
```diff
@@ -355,6 +372,18 @@ class CosmosPackagesStore extends GetSetBaseStore {
     this.emit(COSMOS_REPOSITORIES_SUCCESS);
   }
 
+  processServiceDescriptionSuccess(cosmosPackage, name) {
```
weblancaster: Why are we using this method to do the same thing as `processPackageDescriptionSuccess`?

src/js/stores/CosmosPackagesStore.js
```diff
@@ -355,6 +372,18 @@ class CosmosPackagesStore extends GetSetBaseStore {
...
+    this.set({ packageDetails: cosmosPackage });
+
+    this.emit(COSMOS_SERVICE_DESCRIBE_CHANGE, name);
+  }
+
+  processServiceDescriptionError(error, name) {
```
weblancaster: seems like you have this method `processPackageDescriptionError `

src/js/stores/CosmosPackagesStore.js
ahoskins: The only different is the emitted event. I think they should be kept using their own separate methods.

src/js/stores/CosmosPackagesStore.js
ahoskins: Same as above. The emitted event is different.

src/js/stores/CosmosPackagesStore.js
ahoskins: `COSMOS_SERVICE_DESCRIBE_CHANGE` follows the same naming as the existing `COSMOS_DESCRIBE_CHANGE`. These are "sister" endpoints, so I think current naming makes sense.

ahoskins: Judging from this file, we seem to use "success" when an action was performed that affects the cluster, and we use "change" when fetching some kind of state.

src/js/stores/CosmosPackagesStore.js
ahoskins: Talked offline and decided we will introduce a new field `serviceDetails` to make this nicer.

src/js/components/modals/InstallPackageModal.js
```diff
@@ -61,7 +61,7 @@ class InstallPackageModal
         name: "cosmosPackages",
         events: [
           "descriptionError",
```
MatApple: We need to change the `descriptionError` too don't we?

ahoskins: Yes, good catch

ahoskins:
@MatApple I only grepped for `descriptionSuccess` and `descriptionError` within `dcos-ui` and `dcos-ui-plugins-private`. There is no other repo that would use these events, right?

MatApple:
@ahoskins no other repos 👍

MatApple:
💯

PR #2406, opened 2017-09-07
DCOS-12802: Introduce Jenkinsfile

orlandohohmeier:
Introduce Jenkinsfile to use Jenkins Pipelines and track and test different build/test configurations.
The current configuration will provision an agent using the `Dockerfile` and then lint, build and test the UI.

Closes DCOS-12802

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,142 @@
+#!/usr/bin/env groovy
+pipeline {
+  agent {
+    dockerfile {
+      args  '--shm-size=1g'
+    }
+  }
+
+  environment {
+    JENKINS_VERSION = 'yes'
+    NODE_PATH = 'node_modules'
+  }
+
+  options {
+    timeout(time: 1, unit: 'HOURS')
+  }
+
+  stages {
+    stage('Initialization') {
+      steps {
+        ansiColor('xterm') {
+          retry(2) {
+            sh '''npm --unsafe-perm install'''
+          }
+
+          sh '''npm run scaffold'''
+        }
+      }
+    }
+
+    stage('Lint') {
+      steps {
+        ansiColor('xterm') {
+          sh '''npm run lint'''
+        }
+      }
+    }
+
+    stage('Unit Test') {
+      steps {
+        ansiColor('xterm') {
+          sh '''npm run test -- --coverage'''
+        }
+      }
+
+      post {
+        always {
+          junit 'jest/test-results/*.xml'
+          step([$class             : 'CoberturaPublisher',
+                autoUpdateHealth   : false,
+                autoUpdateStability: false,
+                coberturaReportFile: 'coverage/cobertura-coverage.xml',
+                failUnhealthy      : true,
+                failUnstable       : true,
+                maxNumberOfBuilds  : 0,
+                onlyStable         : false,
+                sourceEncoding     : 'ASCII',
+                zoomCoverageChart  : false])
+        }
+      }
+    }
+
+    stage('Build') {
+      steps {
+        ansiColor('xterm') {
+          sh '''npm run build-assets'''
+        }
+      }
+
+      post {
+        always {
+          stash includes: 'dist/*', name: 'dist'
+        }
+      }
+
+    }
+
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -0,0 +1,115 @@
...
+    stage('Build') {
+      steps {
+        ansiColor('xterm') {
+          sh '''npm run build-assets'''
+        }
+        stash includes: 'dist/*', name: 'dist'
+      }
+    }
+
+    stage('Integration Tests') {
```
weblancaster: I believe we would benefit a lot in performance if we use the `parallel` feat from jenkins pipeline.
The places I believe we can use is `Lint` `Unit test` `Integration tests` and `System test`

What you think @orlandohohmeier 

for more info: https://www.cloudbees.com/blog/parallelism-and-distributed-builds-jenkins

Jenkinsfile
```diff
@@ -0,0 +1,115 @@
...
+    dockerfile true
+  }
+
+  environment {
+    JENKINS_VERSION = 'yes'
+    NODE_PATH = 'node_modules'
+  }
+
+  options {
+    timeout(time: 2, unit: 'HOURS')
```
weblancaster: 2hrs is incredibly long, especially for our code base. I would expect 1hr max.

weblancaster: If there is a specific stage that we need to timeout we could only add to that stage.

weblancaster:
@orlandohohmeier I left some comments, other than that looks good and I'm excited for us to introduce Jenkinsfile pipeline (especially if we decrease the time it takes to run).

Jenkinsfile
orlandohohmeier: @weblancaster I had a version that used parallel steps, but ran into memory issues as all of these test are executed on the same machine if we don't use the `node` directive to schedule the work on a different nodes and I haven't figured out how to share the initially build docker image which results in rather long bootstrap times and thus is not faster than running these synchronously. Anyway, I'll continue investigating ways to improve the performance and build time and really appreciate your comments and ideas.

Jenkinsfile
orlandohohmeier: You're right that 2 hours is quite long and our builds should complete after 30 mins on average. I'll reduce the general timeout to 1 hour as it should be ok to flag a build that takes longer as a failure.

Jenkinsfile
weblancaster: Ok cool, thanks for the feedback.

Jenkinsfile
weblancaster: Thanks!

nLight:
What do we need to do to make this happen?

orlandohohmeier:
@nLight I need to further investigate why our jest workers (unit tests) or the chrome renderer (integration tests) fail. I thought it's a resource issue, but the agent should have enough resources. I'll continue to experiment with different configurations and will look into ways to better debug the failure. If anyone has an idea how to solve the issues please don't hesitate to adjust our Jenkins setup or this PR.

Dockerfile
```diff
@@ -33,6 +33,10 @@ RUN set -x \
   # Post-install java certificates
   /var/lib/dpkg/info/ca-certificates-java.postinst configure \
 
+ # Install npm dependencies
+  && cd /dcos-ui \
+  && npm install -g git://github.com/johntron/http-server.git#proxy-secure-flag \
```
nLight: I'd rather see python3 http-server here...

nLight:
`ENOMEM: not enough memory, read` for linting command

nLight:
Alfred test this

orlandohohmeier:

Jenkinsfile
```diff
@@ -0,0 +1,142 @@
...
+      post {
+        always {
+          archiveArtifacts 'cypress/**/*'
+          junit 'cypress/*.xml'
+        }
+      }
+    }
+
+
+//    stage('System Test') {
```
nLight: It's ok to have this part commented out

nLight:
@weblancaster please consider ✅

PR #2407, opened 2017-09-07
Add recordio reader package

nLight:
This PR introduces `recordio` package. The format is being used by Apache Mesos to encode events for the Event Stream at the HTTP Operator API. The package will be used to parse Mesos Even Stream.

Closes DCOS-18357

### DIFF
--- a/packages/recordio/README.md
+++ b/packages/recordio/README.md
@@ -1,0 +1,39 @@
+# RecordIO
+
+This package provides a function to read records in the RecordIO format from the input string.
+
+## Usage
+
+```javascript
+import { read } from "recordio";
+
+const [records, rest] = read(input);
+```
+
+Function `read` returns a tuple whose first element is an array of records and the second element is the rest part of the input that is either empty or contains partial records.
+
+## RecordIO format
+
+Prepends to a single record its length in bytes, followed by a newline and then the data:
+
+The BNF grammar for a RecordIO-encoded streaming response is:
+```
+records         = *record
+
+record          = record-size LF record-data
+
+record-size     = 1*DIGIT
+record-data     = record-size(OCTET)
+```
+`record-size` should be interpreted as an unsigned 64-bit integer (uint64).
+
+For example, a stream may look like:
+
+```
+121\n
+{"type": "SUBSCRIBED","subscribed": {"framework_id": {"value":"12220-3440-12532-2345"},"heartbeat_interval_seconds":15.0}20\n
+{"type":"HEARTBEAT"}675\n
+...
+```
+
+Further documentation can be found in the [Apache Mesos documentation](http://mesos.apache.org/documentation/latest/scheduler-http-api/#recordio-response-format).
--- a/packages/recordio/index.js
+++ b/packages/recordio/index.js
@@ -1,0 +1,2 @@
+// eslint-disable-next-line import/prefer-default-export
+export { default as read } from "./src/read";
--- a/packages/recordio/src/read.js
+++ b/packages/recordio/src/read.js
@@ -1,0 +1,26 @@
+const RECORD_PATTERN = /^\d+\n.+/;
+
+export default function read(input) {
+  const records = [];
+  let rest = input;
+
+  while (RECORD_PATTERN.test(rest)) {
+    const delimiterPosition = rest.indexOf("\n");
+
+    const recordLength = parseInt(rest.substring(0, delimiterPosition), 10);
+
+    const recordStartPosition = delimiterPosition + 1;
+    const recordEndPosition = recordStartPosition + recordLength;
+
+    if (isNaN(recordLength) || rest.length < recordEndPosition) {
+      return [records, rest];
+    }
+
+    const record = rest.substring(recordStartPosition, recordEndPosition);
+    rest = rest.substring(recordEndPosition);
+
+    records.push(record);
+  }
+
+  return [records, rest];
+}
--- a/packages/recordio/LICENSE
+++ b/packages/recordio/LICENSE
@@ -1,0 +1,201 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
+      control with that entity. For the purposes of this definition,
+      "control" means (i) the power, direct or indirect, to cause the
+      direction or management of such entity, whether by contract or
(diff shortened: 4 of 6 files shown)


packages/recordio/README.md
```diff
@@ -0,0 +1,48 @@
+# RecordIO
```
orlandohohmeier: As discussed offline, please rework the README so that it's easier to follow.

packages/recordio/index.js
```diff
@@ -0,0 +1,3 @@
+import read from "./src/read";
+
+export default read;
```
orlandohohmeier: Let's change the export to `export {default as read} form "./src/read"` so that we could potentially add a _write_ without breaking imports.

packages/recordio/src/read.js
```diff
@@ -0,0 +1,25 @@
+export default function read(input) {
+  const records = [];
+
+  while (/^\d+\n.+/.test(input)) {
```
orlandohohmeier: Please _cache_ the regular expression to not re-instantiate the expression over and over again.

packages/recordio/src/read.js
```diff
@@ -0,0 +1,25 @@
...
+    const delimiterPosition = input.indexOf("\n");
+    const recordPosition = delimiterPosition + 1;
+
+    const recordLength = parseInt(input.substring(0, delimiterPosition), 10);
+
+    if (recordLength > 0 && input.length >= recordPosition + recordLength) {
+      // Cut of the record length
+      input = input.substring(recordPosition, input.length);
```
orlandohohmeier: Don't reassign parameters even if they'er primitives and as such immutable, as this makes it harder to follow and maintain the code.

packages/recordio/src/read.js
```diff
@@ -0,0 +1,25 @@
+export default function read(input) {
```
orlandohohmeier: Let's invest a minute or two to adjust the implementation to further improve readability and performance.

packages/recordio/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+Package exports one function that parses RecordIO encoded records.
```
orlandohohmeier: How about: "This package provides a function to read records in the [RecordIO format](http://mesos.apache.org/documentation/latest/scheduler-http-api/#recordio-response-format) from the input string."

packages/recordio/README.md
```diff
@@ -0,0 +1,39 @@
...
+import read from "recordio";
+
+const [records, rest] = read(input);
+```
+
+Function `read` returns a tuple whose first element is an array of records and the second element is the rest part of the input that is either empty or contains partial records.
+
+## RecordIO format
+
+This documentation can be also found in the [Apache Mesos documentation](http://mesos.apache.org/documentation/latest/scheduler-http-api/#recordio-response-format).
```
orlandohohmeier: Let's rephrase this paragraph instead of "copying" it from Mesos. I'd extract only the important bits and then link to the details Mesos documentation.

orlandohohmeier:
@nLight please adjust the commit message to be inline with our contributing guidelines.

packages/recordio/index.js
nLight: Then we first have to readjust our eslint that insists I prefered default export.

MatApple: If we find ourselves adding a write in recordio format, I'd certainly have some questions.

orlandohohmeier: Point is, as long as this is a "general" `recordio` and not a `recordio-read` package, we should not prevent ourselves from adding new functions.

nLight: I'm going to switch to dynamic export for now

nLight:
@orlandohohmeier addressed your comments 🙇

nLight:
Alfred retest this please

packages/recordio/src/read.js
```diff
@@ -0,0 +1,22 @@
+const RECORD = /^\d+\n.+/;
+
+export default function read(input, records = []) {
```
orlandohohmeier: I like that you took the time to improve this function. Your implementation is now way more readable. Unfortunately the implementation is way slower than your initial one as your unnecessarily executing string operations.

I ran a benchmark using your initial, the current implementation as well as one I'd propose. 

> read (initial) x 1,921,278 ops/sec ±0.69% (87 runs sampled)
> read (current) 794,427 ops/sec ±0.55% (84 runs sampled)
> read (proposed) x 1,907,616 ops/sec ±1.42% (82 runs sampled)

The numbers vary a little but the following my proposal is either as good as imperative one or beats it by a few ops/sec. 

```js
const RECORD_PATTER = /^(\d+)\n([^]*)/

function read(input, records) {
  const [_, length, data] = RECORD_PATTER.exec(input) || [];

  if (!length || !data || data.length < length) {
    return [records, input];
  }

  const record = data.substring(0, length);
  const rest = input.substring(length + 1);

  return read(rest, records.concat(record));
}
```

nLight: The solution you've provided doesn't pass tests, but I got the idea! Thanks for the input, I'll try to improve the performance

nLight: I fixed the solution and eager to know what kind of a benchmark did you use?
```javascript
const RECORD_PATTERN = /^(\d+)\n([^]*)/;

export default function read(input, records = []) {
  let [_, length, data] = RECORD_PATTERN.exec(input) || [];
  length = parseInt(length, 10);

  if (!length || !data || data.length < length) {
    return [records, input];
  }

  const record = data.substr(0, length);
  const rest = data.substr(length);

  return read(rest, records.concat(record));
}
```

packages/recordio/index.js
```diff
@@ -0,0 +1,3 @@
+import read from "./src/read";
+
+module.exports = { read };
```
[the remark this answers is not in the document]
nLight: Default exports are good. I rather not changing the rule. Here we have a package that we are prematurely optimize API wise. I'd rather have one local hack, than loosening a general style guide.

nLight:
Alfred retest this please

nLight:
Commit message has been adjusted

packages/recordio/index.js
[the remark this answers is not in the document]
orlandohohmeier: How about `export default { read };` instead of `module.exports`?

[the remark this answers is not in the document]
nLight: It doesn't make any difference as it renders static analysis impossible in the very same way.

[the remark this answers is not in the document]
nLight: Rolled back to the improved imperative version.

nLight:
Rolled back to the improved imperative version.

packages/recordio/index.js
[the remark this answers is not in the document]
orlandohohmeier: That's true, but at least we'er not mixing the different module pattern.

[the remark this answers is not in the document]
nLight: How about suppressing eslint for that particular case? :trollface: 💯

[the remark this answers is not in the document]
nLight: I went with a conscious eslint override 🙏

nLight:
Alfred retest this please

packages/recordio/src/read.js
nLight: Invested

nLight:
Alfred retest this please

PR #2409, opened 2017-09-08
Introduce mesos-client package

nLight:
# mesos-client
It is a wrapper that uses `http-service` to establish a connection and `recordio` package to parse the incoming data and deliver it as a `Rx.Observable`.

## Dependencies
✅ #2407 `recordio` 
✅ #2415 `http-service` (former `resource-service`)

Closes DCOS-18358

### DIFF
--- a/packages/mesos-client/README.md
+++ b/packages/mesos-client/README.md
@@ -1,0 +1,24 @@
+# Apache Mesos Client
+
+This package provides a nice way of connecting to the Mesos Event Stream API. It is a wrapper that uses `http-service` to establish a connection and `recordio` package to parse the incoming data and deliver it as a `Rx.Observable`.
+
+## Example
+```javascript
+import { stream, request } from "mesos-client";
+
+stream({ type: "SUBSCRIBE" }).subscribe(
+  value => console.log(value),
+  error => console.log(error),
+  () => console.log("complete")
+);
+
+request({ type: "GET_FLAGS" }).subscribe(
+  value => console.log(value),
+  error => console.log(error),
+  () => console.log("complete")
+);
+```
+
+`stream` opens a persistent connection to [Mesos HTTP Operator Api](http://mesos.apache.org/documentation/latest/operator-http-api) Event Stream and returns `rxjs` Observable.
+
+`request` makes a call to the [Mesos HTTP Operator Api](http://mesos.apache.org/documentation/latest/operator-http-api)  and returns response as an `rxjs` Observable.
--- a/packages/mesos-client/index.js
+++ b/packages/mesos-client/index.js
@@ -1,0 +1,2 @@
+export { default as stream } from "./src/stream";
+export { default as request } from "./src/request";
--- a/packages/mesos-client/LICENSE
+++ b/packages/mesos-client/LICENSE
@@ -1,0 +1,201 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
+      control with that entity. For the purposes of this definition,
+      "control" means (i) the power, direct or indirect, to cause the
+      direction or management of such entity, whether by contract or
+      otherwise, or (ii) ownership of fifty percent (50%) or more of the
+      outstanding shares, or (iii) beneficial ownership of such entity.
+
+      "You" (or "Your") shall mean an individual or Legal Entity
+      exercising permissions granted by this License.
+
+      "Source" form shall mean the preferred form for making modifications,
+      including but not limited to software source code, documentation
+      source, and configuration files.
+
+      "Object" form shall mean any form resulting from mechanical
+      transformation or translation of a Source form, including but
+      not limited to compiled object code, generated documentation,
+      and conversions to other media types.
+
+      "Work" shall mean the work of authorship, whether in Source or
+      Object form, made available under the License, as indicated by a
+      copyright notice that is included in or attached to the work
+      (an example is provided in the Appendix below).
+
+      "Derivative Works" shall mean any work, whether in Source or Object
+      form, that is based on (or derived from) the Work and for which the
+      editorial revisions, annotations, elaborations, or other modifications
+      represent, as a whole, an original work of authorship. For the purposes
+      of this License, Derivative Works shall not include works that remain
+      separable from, or merely link (or bind by name) to the interfaces of,
+      the Work and Derivative Works thereof.
+
+      "Contribution" shall mean any work of authorship, including
+      the original version of the Work and any modifications or additions
+      to that Work or Derivative Works thereof, that is intentionally
+      submitted to Licensor for inclusion in the Work by the copyright owner
+      or by an individual or Legal Entity authorized to submit on behalf of
+      the copyright owner. For the purposes of this definition, "submitted"
+      means any form of electronic, verbal, or written communication sent
+      to the Licensor or its representatives, including but not limited to
+      communication on electronic mailing lists, source code control systems,
+      and issue tracking systems that are managed by, or on behalf of, the
+      Licensor for the purpose of discussing and improving the Work, but
+      excluding communication that is conspicuously marked or otherwise
+      designated in writing by the copyright owner as "Not a Contribution."
+
+      "Contributor" shall mean Licensor and any individual or Legal Entity
+      on behalf of whom a Contribution has been received by Licensor and
+      subsequently incorporated within the Work.
+
+   2. Grant of Copyright License. Subject to the terms and conditions of
+      this License, each Contributor hereby grants to You a perpetual,
+      worldwide, non-exclusive, no-charge, royalty-free, irrevocable
+      copyright license to reproduce, prepare Derivative Works of,
+      publicly display, publicly perform, sublicense, and distribute the
+      Work and such Derivative Works in Source or Object form.
+
+   3. Grant of Patent License. Subject to the terms and conditions of
+      this License, each Contributor hereby grants to You a perpetual,
+      worldwide, non-exclusive, no-charge, royalty-free, irrevocable
+      (except as stated in this section) patent license to make, have made,
+      use, offer to sell, sell, import, and otherwise transfer the Work,
+      where such license applies only to those patent claims licensable
+      by such Contributor that are necessarily infringed by their
+      Contribution(s) alone or by combination of their Contribution(s)
+      with the Work to which such Contribution(s) was submitted. If You
+      institute patent litigation against any entity (including a
+      cross-claim or counterclaim in a lawsuit) alleging that the Work
+      or a Contribution incorporated within the Work constitutes direct
+      or contributory patent infringement, then any patent licenses
+      granted to You under this License for that Work shall terminate
+      as of the date such litigation is filed.
+
+   4. Redistribution. You may reproduce and distribute copies of the
+      Work or Derivative Works thereof in any medium, with or without
+      modifications, and in Source or Object form, provided that You
+      meet the following conditions:
+
+      (a) You must give any other recipients of the Work or
+          Derivative Works a copy of this License; and
+
+      (b) You must cause any modified files to carry prominent notices
+          stating that You changed the files; and
+
+      (c) You must retain, in the Source form of any Derivative Works
+          that You distribute, all copyright, patent, trademark, and
+          attribution notices from the Source form of the Work,
+          excluding those notices that do not pertain to any part of
+          the Derivative Works; and
+
+      (d) If the Work includes a "NOTICE" text file as part of its
+          distribution, then any Derivative Works that You distribute must
+          include a readable copy of the attribution notices contained
+          within such NOTICE file, excluding those notices that do not
+          pertain to any part of the Derivative Works, in at least one
+          of the following places: within a NOTICE text file distributed
+          as part of the Derivative Works; within the Source form or
+          documentation, if provided along with the Derivative Works; or,
+          within a display generated by the Derivative Works, if and
+          wherever such third-party notices normally appear. The contents
+          of the NOTICE file are for informational purposes only and
+          do not modify the License. You may add Your own attribution
+          notices within Derivative Works that You distribute, alongside
+          or as an addendum to the NOTICE text from the Work, provided
+          that such additional attribution notices cannot be construed
+          as modifying the License.
+
+      You may add Your own copyright statement to Your modifications and
+      may provide additional or different license terms and conditions
+      for use, reproduction, or distribution of Your modifications, or
+      for any such Derivative Works as a whole, provided Your use,
+      reproduction, and distribution of the Work otherwise complies with
+      the conditions stated in this License.
+
+   5. Submission of Contributions. Unless You explicitly state otherwise,
+      any Contribution intentionally submitted for inclusion in the Work
+      by You to the Licensor shall be under the terms and conditions of
+      this License, without any additional terms or conditions.
+      Notwithstanding the above, nothing herein shall supersede or modify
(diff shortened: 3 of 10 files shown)


packages/mesos-client/src/createEventStreamConnection.js
```diff
@@ -0,0 +1,34 @@
...
+  // but we need the data chunk by chunk to emit records as they come
+  // thus we track the position in data that we've already processed
+  const chunk = data.substr(position);
+  const [records, rest] = read(buffer + chunk);
+
+  return { records, buffer: rest, position: position + chunk.length };
+}
+
+export default function createEventStreamConnection(baseUrl) {
+  const connection = new XHRConnection(
```
MatApple: Not relevant to your PR, but now that I'm seeing the multiple arguments approach to configuring an `XHRConnection` it doesn't feel right. We're used to a single options Object for configuring `XMLHttpRequest` (and most other request libraries) and it feels bizarre to do it this way. cc/ @orlandohohmeier @juliangieseke

packages/mesos-client/src/observableFromXhrConnection.js
```diff
@@ -0,0 +1,30 @@
+import { ConnectionConstants, ConnectionManager } from "connection-manager";
+import Rx from "rxjs";
+
+export default function observableFromXhrConnection(connection) {
+  return Rx.Observable.create(function(observer) {
+    connection.addListener(
+      ConnectionConstants.CONNECTION_EVENT_PROGRESS,
```
MatApple: Do we need to prefix these constants with `CONNECTION_`? @orlandohohmeier @juliangieseke

packages/mesos-client/src/observableFromXhrConnection.js
```diff
@@ -0,0 +1,30 @@
...
+      observer.error(error);
+    });
+    connection.addListener(
+      ConnectionConstants.CONNECTION_EVENT_CLOSE,
+      function() {
+        observer.complete();
+      }
+    );
+
+    ConnectionManager.queue(connection);
```
MatApple: I don't think this should live here - it pollutes the scope. It feels like magic because I would still expect to do this after I've defined my `XHRConnection` in `createEventStreamConnection.js`

packages/mesos-client/src/observableFromXhrConnection.js
```diff
@@ -0,0 +1,30 @@
...
+
+    return function() {
+      connection.close();
```
MatApple: Is there a chance the `connection` is already closed? Couldn't this function be invoked because `ConnectionConstants.CONNECTION_EVENT_CLOSE` was triggered?

packages/mesos-client/src/createEventStreamConnection.js
```diff
@@ -0,0 +1,34 @@
+import Rx from "rxjs";
+import { read } from "recordio";
+import { XHRConnection } from "connection-manager";
+
+import observableFromXhrConnection from "./observableFromXhrConnection";
+
+function extractRecords({ buffer = "", position = 0 } = {}, data = "") {
```
MatApple: This will need tests and as such we need to think about how we make it testable.

nLight: Yes and no, the read function is covered in the `recordio` package

packages/mesos-client/src/createEventStreamConnection.js
nLight: Adjusted in #2361

packages/mesos-client/src/observableFromXhrConnection.js
nLight: Adjusted in #2361

packages/mesos-client/src/observableFromXhrConnection.js
nLight: Yes, that can happen, but then `Observable` will be `completed`. This piece of logic is responsible for cleaning up resources after all Observers have unsubscribed from the Observable.

packages/mesos-client/src/observableFromXhrConnection.js
nLight: It may feel like a magic when you use it directly, but that's gonna be a utility function and the interface in general is that you pass the `baseUrl` and you have everything done for you. 

My first take was that one creates a connection and passes it to the observable wrapper that library provides. But then why would you call the library `mesos-client`, when you have to manually setup a connection? So after an offline discussion with @orlandohohmeier we've decided that it make much more sense to request only `baseUrl` and let the library do the rest.

orlandohohmeier: We probably have to move this helper to a different package and find a better name to make sure it feels less magic. How about introducing a resource(?) service that provides method similar to `fetch`  or`reqwest`which will internally create and enqueue a connection and return an observable.

This method needs to either clone the connection or create the connection (which is preferable) within the `Observable.create` closure as it otherwise probably doesn't make any sense to enqueue the connection. Only when we completely handing the connection management within the closure we can enable use cases like the following one:

```js
ResourceService
    .request("url", {method: "GET"})
    .retry(3)
    .subscribe((event) => console.log(event));
```

packages/mesos-client/src/createEventStreamConnection.js
orlandohohmeier: I think it still makes sense to test this function! How about separating all of the transformation functions into a separate file(s) so that we can properly test them?

packages/mesos-client/index.js
```diff
@@ -0,0 +1,3 @@
+export {
+  default as createEventStreamConnection
+} from "./src/createEventStreamConnection";
```
orlandohohmeier: We need to find a good way of sharing the created stream as we shouldn't open more than one stream. Please think about an elegant interface to do so.

nLight: I think that is out of the scope of the package. I'm not even convinced we should hardwire it with the connection manager, but making the connection a singleton out of the box will effectively merge this package with dcos ui. I planned it to be a reusable to a degree package.

nLight: So I'd say it is a job for DCOS UI, get the connection, and then make sure it is the only one and reusable.

packages/mesos-client/src/observableFromXhrConnection.js
orlandohohmeier: @nLight let's name this method to it is easier to understand.

nLight:
`resource-service` dependency implementation has diverged from latest `connection-manager` changes

packages/mesos-client/index.js
orlandohohmeier: Yeah, we shouldn't add the functionality to the mesos client, but we need to design a mechanism to share and manage the stream. We could introduce an IoC container to create and share the stream or build a singleton that provides an interface to subscribe to the stream. What do you think?

packages/mesos-client/src/rxjs-extensions/keepAlive.js
```diff
@@ -0,0 +1,111 @@
+import { Subscriber, Scheduler } from "rxjs";
```
orlandohohmeier: Let's move this operator to the resource service package

packages/mesos-client/src/rxjs-extensions/keepAlive.js
```diff
@@ -0,0 +1,111 @@
...
+  return source.lift(
+    new KeepAliveOperator(source, backoffInterval, maxRetries, scheduler)
+  );
+}
+
+class KeepAliveOperator {
+  constructor(source, backoffInterval, maxRetries, scheduler) {
+    this.source = source;
+    this.maxRetries = maxRetries;
+    this.backoffInterval = backoffInterval;
```
orlandohohmeier: The keepalive operartor shouldn't also apply the backoff. Every operator should only to one thing. 
 
It should be possible to add the backoff using `concat` as illustrated in the following diagram:

packages/mesos-client/src/rxjs-extensions/adaptiveTimeout.js
```diff
@@ -0,0 +1,89 @@
+import { Subscriber, Scheduler, TimeoutError } from "rxjs";
+
+// As Mesos documentation suggests, to make sure the connection is dead
+// we should wait for "an extended period of time" e.g., 5 x heartbeat_interval_seconds
+export const TIMEOUT_FACTOR = 5;
+const DEFAULT_TIMEOUT = 5000;
+
+export default function adaptiveTimeout(
```
orlandohohmeier: Do you think there is a way to write this operator so that it's generally applicable?

nLight: Sure, by exposing the update function as an argument. Do we really need this?
So the function can be called on every `next` passing the value inside. Whether use this value or not it's up to the function.

packages/mesos-client/src/rxjs-extensions/keepAlive.js
nLight: Then `keepAlive` is just a combination of `retry()` and `repeat()` :)
and, as you might remember, in the previous commit that I've kept on purpose, we've had the backoff extracted and it was rather confusing.

packages/mesos-client/src/rxjs-extensions/keepAlive.js
nLight: That actually makes sense!

packages/mesos-client/src/rxjs-extensions/keepAlive.js
nLight: Regarding the `concat` operator, isn't it that it should know all the observables upfront? We need to resubscribe to the stream dynamically.

orlandohohmeier: Yep, we don't even need  `retry` as we catch the connection errors.

```js 
  Rx.Observable.create((observer) => {
    console.log("subscribe")
    setTimeout(() => {
      if(Math.random() >= 0.5){
         console.log("complete")
         observer.complete()
      } else {
        console.log("error")
        observer.error()
      }
    }, 1000)
  })
  .catch((err) => {
      return Rx.Observable.empty();
  })
  .concat(
    Rx.Observable.create( (observer) => {
      console.log("backoff")
      setTimeout(function (){
        observer.complete();
      }, 1000)
    })
  )
  .repeat()
  .subscribe()
```

packages/mesos-client/src/rxjs-extensions/adaptiveTimeout.js
orlandohohmeier: rxjs actually had a `timeoutWithSelector` which was dropped in version 5 to reduce weight.

http://reactivex.io/documentation/operators/timeout.html

I'm fine if we don't generalize the operator as a simple `timeout` is probably sufficient in most cases, but then let's adjust the name so that it's obvious that this one checks the mesos heartbeat.

nLight: After some meditation on the class based operators I think it's not a good way to go. It exposes a lot of rxjs internals that 1) can change over time 2) one needs to know to understand what's happening.

how about something like this:
```javascript
function mesosTimeout(source) {
  let heartBeat;
  
  const sourceSubject = source.publish().refCount();
  
  return Rx.Observable.create(function subscribe(observer) {
    // First observable reads timeout and completes
    const sub = sourceSubject.do(function(value) {
      heartBeat = 1100;
      observer.next(value);
      observer.complete();
    }).subscribe();
    
    return function() {
      sub.unsubscribe();
    }
  }).concat(
    Rx.Observable.create(function (observer) {      
      // Second observable uses previously read timeout to control the stream
      sourceSubject
        .timeout(heartBeat)
        .subscribe(function(value) {
          observer.next(value);
        });
    })
  );
} 

mesosTimeout(Rx.Observable.interval(1000))
  .subscribe(function(value){ 
    console.log(value);
  });
```

packages/mesos-client/src/rxjs-extensions/keepAlive.js
nLight: @MatApple would like to get your view on both version.

packages/mesos-client/src/rxjs-extensions/adaptiveTimeout.js
nLight: @MatApple would like to get your view on both versions.

orlandohohmeier: I'd like to prose yet another solution to monitor the  "heartbeat" without any side effects. It also features a selector to dynamically adapt the timeout.

```js
// monitor heartbeat
function monitor(source, timeoutSelector = function(){}, initialTimeout=500){
   return source
     .scan(
        ({timeout}, message) => {
        const _timeout = timeoutSelector(message);
          
        if(!Number.isFinite(_timeout)){
          return {message, timeout: _timeout}
         }
          
        return {message, timeout};     
       }, {timeout: initialTimeout})
     .map(({timeout, message}) => { 
       return Rx.Observable.create(function create(observer){
         const timeoutId = setTimeout(()=> observer.error(), timeout);

         // pass message along 
         observer.next(message);

         return function disspose(){
           clearTimeout(timeoutId); 
         };                       
       })
     })
    .switch()
}
```

**Example**
```js
// Debug Source
const source = Rx.Observable.of(1000,500,1000,800,500,1000, 2000, 5000, 1000, 6000)
   .map((time)=>Rx.Observable.timer(time))
   .concatAll()
   .map((_, index ) => { 
     if(index === 2){
       return {index, heartbeat: 100}; 
     }
     return {index}; 
   });

    monitor(source, (message) => message.heatbeat * 5, 6000
     .subscribe(
          (message)=>console.log("message", message),
          (error)=>console.log("error", error)
      )
```

nLight: The implementation is elegant though it unfortunately doesn't work with an initial timeout, and I'd rather avoid `setTimeout` in favor of native timeouts that work on the rxjs scheduler thus can be tested with built-in tooling.

packages/mesos-client/src/rxjs-extensions/keepAlive.js
```diff
@@ -0,0 +1,3 @@
+export default function keepAlive(source, maxRetries) {
```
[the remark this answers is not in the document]
nLight: Do you want me to do this within this PR?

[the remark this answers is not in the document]
orlandohohmeier: How you like...

packages/mesos-client/src/rxjs-extensions/heartbeatTimeout.js
```diff
@@ -0,0 +1,43 @@
...
+          // so we don't run this logick any longer
+          observer.next(value);
+          observer.complete();
+        }
+      },
+      error: error => observer.error(error),
+      complete: () => observer.complete()
+    });
+
+    return function terminate() {
```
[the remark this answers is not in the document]
nLight: All the approaches feel wrong to me. I kinda like the idea of the class based operator, how it encapsulates the logic, the only problem was that it is to bound to the internals of rxjs. How about we discard listening timeout from the wire and stick with a _reasonable_ timeout?

packages/mesos-client/src/recordio-extensions/extractRecords.js
```diff
@@ -0,0 +1,14 @@
+import { read } from "recordio";
```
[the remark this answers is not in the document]
nLight: It's not really an `operator` in the `rxjs` terms, I'd rather not mixing terms.

nLight:
As discussed offline with @orlandohohmeier we're dropping the _smart_ timeout for now. Replacing it with a 10 seconds timeout. 

Restructured code, renamed stuff, wrote 1 more test.

🙇

packages/mesos-client/README.md
```diff
@@ -0,0 +1,73 @@
...
+```
+
+`stream` opens a persistent connection to [Mesos HTTP Operator Api](http://mesos.apache.org/documentation/latest/operator-http-api) Event Stream and returns `rxjs` `ReplaySubject` Observable.
+
+`request` makes a call to the Operator API and returns response as an `rxjs` Observable.
+
+## Why and how it works
+Mesos event stream always has the first message that contains its heartbeat interval. We need to use this interval as suggested by the documentation to reconnect if we don't get any messages within a timespan that is longer then a heartbeat.
+
+### `heartbeatTimeout`
```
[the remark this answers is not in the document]
nLight: Forgot to push 😅

orlandohohmeier:
✨ Awesome

PR #2415, opened 2017-09-14
Introduce http-service

nLight:
# http-service
Former `resource-service` is a helper package that leverages the `connection-manager` and `rxjs` to create a managed observable connection.

## Exports
Exports two functions: `request` and `stream`

### `request`
For a simple request like loading a `json` file one wants to get it when it's fully loaded.
So `request` fires only one `next` immediately followed by `complete` after the resource has been fully loaded.
```
XHRConnection DATA events
-o--o--o--o--o--|

request
-------------o--|
```

### `stream`
Fires `next` on every `DATA` event coming from the `XHRConnection`. Every value in next will have `XHRConnection#response` that will be growing, so if you need this data chunk by chunk you should run your own logic on top (as we do in `mesos-client`)
```
XHRConnection DATA events
-o--o--o--o--o--|

stream
-o--o--o--o--o--|
```

## Dependencies
✅  #2425  `connection-manager`
✅  #2418  `connections`

Closes DCOS-18687

### DIFF
--- a/packages/http-service/README.md
+++ b/packages/http-service/README.md
@@ -1,0 +1,31 @@
+# HTTP Service
+
+This package wraps connections managed by the `connection-manager` package into an Observable.
+
+## Usage
+```javascript
+import { request, stream } from "http-service";
+
+request("http://localhost:4200/payload.json")
+  .retry(3)
+  .subscribe({
+    next: response => console.log(response),
+    error: event => console.log(event),
+    complete: () => console.log("complete")
+  });
+
+stream("http://localhost:4200/mesos/api/v1", {
+  method: "POST",
+  responseType: "text",
+  body: JSON.stringify({ type: "SUBSCRIBE" }),
+  headers: {
+    "Content-Type": "application/json",
+    "Accept": "application/json"
+  }
+})
+  .subscribe({
+    next: data => console.log(data),
+    error: event => console.log(event),
+    complete: () => console.log("complete")
+  });
+```
--- a/packages/http-service/src/request.js
+++ b/packages/http-service/src/request.js
@@ -1,0 +1,29 @@
+import Rx from "rxjs";
+
+import ConnectionManager from "connection-manager";
+import { XHRConnection, ConnectionEvent } from "connections";
+
+export default function request(url, options = {}) {
+  return Rx.Observable.create(function(observer) {
+    const connection = new XHRConnection(url, options);
+
+    connection.addListener(ConnectionEvent.ERROR, function(event) {
+      observer.error({
+        code: event.target.xhr.status,
+        message: event.target.xhr.statusText
+      });
+    });
+
+    connection.addListener(ConnectionEvent.COMPLETE, function(event) {
+      observer.next(event.target.response);
+      observer.complete();
+    });
+
+    ConnectionManager.enqueue(connection);
+
+    // Closing the connection when there are no subscribers left
+    return function teardown() {
+      ConnectionManager.dequeue(connection);
+    };
+  });
+}
--- a/packages/http-service/src/stream.js
+++ b/packages/http-service/src/stream.js
@@ -1,0 +1,30 @@
+import Rx from "rxjs";
+
+import ConnectionManager from "connection-manager";
+import { XHRConnection, ConnectionEvent } from "connections";
+
+export default function stream(url, options = {}) {
+  return Rx.Observable.create(function(observer) {
+    const connection = new XHRConnection(url, options);
+
+    connection.addListener(ConnectionEvent.DATA, function(event) {
+      observer.next(event.target.response);
+    });
+    connection.addListener(ConnectionEvent.ERROR, function(event) {
+      observer.error({
+        code: event.target.xhr.status,
+        message: event.target.xhr.statusText
+      });
+    });
+    connection.addListener(ConnectionEvent.COMPLETE, function() {
+      observer.complete();
+    });
+
+    ConnectionManager.enqueue(connection);
+
+    // Closing the connection when there are no subscribers left
+    return function teardown() {
+      ConnectionManager.dequeue(connection);
+    };
+  });
+}
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -39,6 +39,18 @@
       "from": "acorn-jsx@>=3.0.0 <4.0.0",
       "resolved": "https://registry.npmjs.org/acorn-jsx/-/acorn-jsx-3.0.1.tgz"
     },
+    "ajv": {
+      "version": "4.11.8",
+      "from": "ajv@>=4.9.1 <5.0.0",
+      "resolved": "https://registry.npmjs.org/ajv/-/ajv-4.11.8.tgz",
+      "dependencies": {
+        "co": {
+          "version": "4.6.0",
+          "from": "co@>=4.6.0 <5.0.0",
+          "resolved": "https://registry.npmjs.org/co/-/co-4.6.0.tgz"
+        }
+      }
+    },
     "align-text": {
       "version": "0.1.4",
       "from": "align-text@>=0.1.3 <0.2.0",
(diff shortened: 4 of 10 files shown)


nLight:
Conneciton manager #2361 has changed since my last commit. Instead of continuously adjusting my implementation let's wait until it's merged

packages/resource-service/src/request.js
```diff
@@ -0,0 +1,28 @@
...
+    const connection = new XHRConnection(url, options);
+
+    connection.addListener(EventTypes.PROGRESS, function(target) {
+      observer.next(target.response);
+    });
+    connection.addListener(EventTypes.ERROR, function(error) {
+      observer.error(error);
+    });
+    connection.addListener(EventTypes.CLOSE, function() {
+      observer.complete();
```
orlandohohmeier: 🤔  wonder if we need to push another event before completing, to make sure we get the _complete_ response...

orlandohohmeier:
The current implementation makes it rather complicated when dealing with simple requests (e.g., static JSON file) as one could potentially receive several events and not only the "COMPLETE" event.  Please think of ways to make this interface as easy as using `fetch`.

orlandohohmeier:
We should adjust the interface and names as this service should eventually replace our [`RequestUtil`](https://github.com/mesosphere/mesosphere-shared-reactjs/blob/master/utils/RequestUtil.js) and as such must be super easy to use.

Currently performing a simple JSON request requires users to filter the stream for the "COMPLETE" event as the "DATA" event fires on every chunk received as mentioned in https://github.com/dcos/dcos-ui/pull/2415#issuecomment-330191588

```js
request(url)
  .filter(event => event.type === ConnectionEvent.COMPLETE)
  .subscribe(
     (event) => console.log(event), 
    (error) => console.log(event),
    () => console.log("complete")
  )
```
Requiring users to do so feels wrong thus I'd provide two different methods, one to stream data via HTTP and one to request data. Requesting data would push only once on "COMPLETE", and a such would be very similar to our current RequestUtil; stream, on the other hand, would continuously push top the stream till the connection is closed. 

```js
request(url).subscribe(
  (event) => console.log(event), 
  (error) => console.log(event),
  () => console.log("complete")
 )

stream(url).subscribe(
  (event) => console.log(event), 
  (error) => console.log(event), 
  () => console.log("complete")
 )
```

I also suggest renaming the package to `http-service` as it avoids confusion with other mechanisms to stream resources (e.g. WebSocket) that do come with there very own requirements and probably need a slightly different interface.

What do you think?

/cc @dcos/frontend-foundation

nLight:
Hey @orlandohohmeier, that makes sense. Implemented two separate functions as suggested.

packages/http-service/src/stream.js
```diff
@@ -0,0 +1,27 @@
...
+
+import ConnectionManager from "connection-manager";
+import { XHRConnection, ConnectionEvent } from "connections";
+
+export default function stream(url, options = {}) {
+  return Rx.Observable.create(function(observer) {
+    const connection = new XHRConnection(url, options);
+
+    connection.addListener(ConnectionEvent.DATA, function(event) {
+      observer.next(event.target.response);
```
orlandohohmeier: Let's always pass the event along and rather extend the event to also include the response.

packages/http-service/src/stream.js
```diff
@@ -0,0 +1,27 @@
...
+    });
+    connection.addListener(ConnectionEvent.ERROR, function(event) {
+      observer.error(event);
+    });
+    connection.addListener(ConnectionEvent.COMPLETE, function() {
+      observer.complete();
```
orlandohohmeier: We probably should to push the data on complete as well before closing the steam.

packages/http-service/src/stream.js
```diff
@@ -0,0 +1,27 @@
...
+    });
+
+    ConnectionManager.enqueue(connection);
+
+    // Closing the connection when there are no subscribers left
+    return function teardown() {
+      connection.close();
```
orlandohohmeier: Let's use the connection manager `dequeue` instead of directly interacting with the connection.

packages/http-service/README.md
```diff
@@ -0,0 +1,31 @@
+# HTTP Service
+
+This package wraps connections managed by the `connection-manager` package into an Observable.
+
+## Usage
+```javascript
+import { request, stream } from "http-service";
+
+request("http://localhost:4200/payload.json",)
```
orlandohohmeier: Please remove the dangling comma or add an options object.

packages/http-service/src/request.js
```diff
@@ -0,0 +1,25 @@
+import Rx from "rxjs";
+
+import ConnectionManager from "connection-manager";
+import { XHRConnection, ConnectionEvent } from "connections";
+
+export default function request(url, options = {}) {
```
orlandohohmeier: We should to either add a third argument or read the connection priority from the options provided to enable high priority requests and streams.

packages/http-service/src/stream.js
nLight: In this case we will be duplicating the last chunk and this is not something you want to have subscribing to a stream

orlandohohmeier: Is the data event also firing on complete?

packages/http-service/src/stream.js
nLight: Makes sense

packages/http-service/src/stream.js
nLight: On one hand that totally makes sense, but then I was designing it to be a certain level of abstraction, so on `next` you just get the data.

packages/http-service/src/stream.js
nLight: The thing is it's not `data` it's `progress`. Even when error occurs, first you get the `progress` event and then `error`

orlandohohmeier: Alright, thanks for checking.

packages/http-service/src/stream.js
orlandohohmeier: Following that logic we should adjust error and pass along only the error message, shouldn't we?

nLight: That was my idea, but since we don't have a clear `errorMessage` field and `ERROR` event has so many faces I opted in for `event` in this case.

packages/http-service/src/request.js
nLight: Fair point. Things to consider: maybe to control priority we should use raw ConnectionManager instead? Otherwise it feels like it is becoming a leaky abstraction. I'd say `http-service` gives you the default priority and if you need to override it, then you can just import `ConnectionManager` and roll everything manually.

orlandohohmeier: 💯  Convinced

packages/http-service/src/stream.js
orlandohohmeier: I'd love to always pass the same type if possible.

nLight: That means we would need to always map the happy path to unwrap the event.

orlandohohmeier: How about extending the connection event which always sends the data along?

nLight: Shall it also send xhr.status? ...and the status code? :)

orlandohohmeier: We're going to pass an abstraction instead of the event in case of an error and will keep the _next_ as is, as discussed offline.

packages/http-service/src/request.js
```diff
@@ -0,0 +1,25 @@
+import Rx from "rxjs";
```
nLight: `request` could be implemented by reshaping `stream` with some `rxjs` operators. That'd be DRY'er. I think it is better not to do this, but feel free to challenge that.

orlandohohmeier: :100:

nLight:
Alfred retest this please

packages/http-service/src/stream.js
juliangieseke: +1 `observer.next({data: event.target.data})` – it would be in line with the error case, why would we need `type` here?!

IMHO in the future we should pass `DataConnectionEvent` and `ErrorConnectionEvent` to the callback/subscriber. As both are TBD yet, we should at least pass an object to be able to exchange them later on.

nLight:
Alfred retest this please 💪

orlandohohmeier:
👏 🥑

packages/http-service/README.md
```diff
@@ -0,0 +1,31 @@
+# HTTP Service
+
+This package wraps connections managed by the `connection-manager` package into an Observable.
+
+## Usage
+```javascript
+import { request, stream } from "http-service";
+
+request("http://localhost:4200/payload.json")
+  .retry(3)
```
MatApple: @nLight @orlandohohmeier this isn't how we're going to use retry right?

nLight: We are free to use the built in retry with `http-service`, but you are correct we're going to implement the back-off retry. Just not for now.

PR #2416, opened 2017-09-14
fix(JSONReducers/Containers): delete protocol field on ADD_ITEM

nLight:
Since we've ditched `applyPatch` but still want to keep in stuff that we don't know about we started copying values via `ADD_ITEM`. The problem here was that the internal representation of `protocol` field was an object wheres the JSON representation is an array. 

That array threw a wrench into the parser-reducer logic, because an array _is_ an object in javascript.

To fix the issue I had to delete `protocol` field for the `ADD_ITEM` event and process it separately. I also added a missing bit to out integrity test to make sure we won't have it again.

Closes DCOS-18694


orlandohohmeier:
LGTM! 🚢

PR #2418, opened 2017-09-18
Introduce connections

orlandohohmeier:
---
ℹ️   _The respective features were originally implemented by @juliangieseke as part of #2361. This PR extracts the connections into their own package to enable different use cases and streamline the further development and review process._

---

Add a package that provides different connection types with a unified interface. For a detailed a list and example on how to use them please see: [README](https://github.com/dcos/dcos-ui/blob/898438d8a6ce629d2743672a77bc31fb81b92f88/packages/connections/README.md)

### DIFF
--- a/packages/connections/package.json
+++ b/packages/connections/package.json
@@ -1,0 +1,14 @@
+{
+  "name": "connections",
+  "main": "index.js",
+  "version": "0.1.0",
+  "license": "Apache-2.0",
+  "repository": "https://github.com/dcos/dcos-ui/",
+  "scripts": {
+    "test": "jest"
+  },
+  "devDependencies": {
+    "jest": "19.0.2",
+    "jest-cli": "16.0.2"
+  }
+}
--- a/packages/connections/src/AbstractConnection.js
+++ b/packages/connections/src/AbstractConnection.js
@@ -1,0 +1,130 @@
+import { EventEmitter } from "events";
+
+/**
+ * Connection init state
+ *
+ * @constant
+ * @type {Symbol}
+ */
+const INIT = Symbol("INIT");
+
+/**
+ * Connection started state
+ *
+ * @constant
+ * @type {Symbol}
+ */
+const OPEN = Symbol("OPEN");
+
+/**
+ * Connection done state
+ *
+ * @constant
+ * @type {Symbol}
+ */
+const CLOSED = Symbol("CLOSED");
+
+/**
+ * The `AbstractConnection` provides default properties, methods, states and
+ * event-definitions. Extend this class when creating a custom connection type
+ * to make sure it's compatible with the `ConnectionManager`.
+ *
+ * Use the following events to communicate state changes:
+ *
+ * - `OPEN` when the connection actually blocks a pipe
+ * - `DATA` every time data is received from the server
+ * - `ERROR` when an error occurred
+ * - `COMPLETE` when the connection frees its pipe
+ * - `ABORT` when the connection is aborted
+ */
+export default class AbstractConnection extends EventEmitter {
+  /**
+   * Abstract connection
+   * @param {string} url
+   */
+  constructor(url) {
+    super();
+
+    if (this.constructor === AbstractConnection) {
+      throw new Error("Can't instantiate abstract class!");
+    }
+
+    if (!url) {
+      throw new Error("Can't instantiate without given URL!");
+    }
+
+    /**
+     * @property {string}
+     * @name AbstractConnection#url
+     */
+    Object.defineProperty(this, "url", { value: url });
+
+    /**
+     * @property {number}
+     * @name AbstractConnection#created
+     */
+    Object.defineProperty(this, "created", { value: Date.now() });
+
+    /**
+     * @property {Symbol}
+     * @protected
+     * @name AbstractConnection#state
+     */
+    let state = AbstractConnection.INIT;
+    Object.defineProperty(this, "state", {
+      get: () => state,
+      set: value => {
+        if (![INIT, OPEN, CLOSED].includes(value)) {
+          throw new Error(
+            `Cant set connection into "${value}" state, allowed states are INIT, OPEN, CLOSED`
+          );
+        }
+        state = value;
+      }
+    });
+  }
+
+  /**
+   * @description Connection State after Initialization
+   * initial state, cant be set again by client
+   */
+  static get INIT() {
+    return INIT;
+  }
+
+  /**
+   * @description Connection State while occupiing one "browser pipe"
+   * active state, change is only allowed to AbstractConnection.CLOSED
+   */
+  static get OPEN() {
+    return OPEN;
+  }
+
+  /**
+   * @description Connection State after freeing the pipe.
+   * terminal state, no more state changed allowed
+   */
+  static get CLOSED() {
+    return CLOSED;
+  }
+
+  /**
+   * Opens the connection
+   *
+   * @description This method should only be called by Connection Manager,
+   * if you call it yourself, be sure you know what you're doing.
+   * Also this method should be the point where a Connection actually starts
+   * using a "Connection Pipe" of the browser.
+   *
+   * Furthermore, you have to bind the follwing events in this functions:
+   * - OPEN when this function is called
+   * - DATA everytime you get data from the server
+   * - ERROR, ABORT, _or_ COMPLETE after the Connection is closed
+   */
+  open() {}
+
+  /**
+   * Closes the connection
+   */
+  close() {}
+}
--- a/packages/connections/src/ConnectionEvent.js
+++ b/packages/connections/src/ConnectionEvent.js
@@ -1,0 +1,97 @@
+import AbstractConnection from "./AbstractConnection";
+
+/**
+ * Connection open event
+ *
+ * @constant
+ * @type {Symbol}
+ */
+const OPEN = Symbol("OPEN");
+
+/**
+ * Connection data event
+ *
+ * @constant
+ * @type {Symbol}
+ */
+const DATA = Symbol("DATA");
+
+/**
+ * Connection error event
+ *
+ * @constant
+ * @type {Symbol}
+ */
+const ERROR = Symbol("ERROR");
+
+/**
+ * Connection complete event
+ *
+ * @constant
+ * @type {Symbol}
+ */
+const COMPLETE = Symbol("COMPLETE");
+
+/**
+ * Connection abort event
+ *
+ * @constant
+ * @type {Symbol}
+ */
+const ABORT = Symbol("ABORT");
+
+/**
+ * Basic Connection Event Object
+ */
+export default class ConnectionEvent {
+  /**
+   * Connection event
+   * @param {AbstractConnection} target
+   * @param {Symbol} type
+   */
+  constructor(target, type) {
+    if (!(target instanceof AbstractConnection)) {
+      throw new Error(
+        `Invalid target, has to be an instance of Abstractconnection`
+      );
+    }
+
+    if (![OPEN, DATA, ERROR, COMPLETE, ABORT].includes(type)) {
+      throw new Error(
+        `Invalid type "${type}", allowed types are OPEN, DATA, ERROR, COMPLETE, ABORT`
+      );
+    }
+
+    /**
+     * @property {AbstractConnection}
+     * @name ConnectionEvent#target
+     */
+    Object.defineProperty(this, "target", { value: target });
+
+    /**
+     * @property {Symbol}
+     * @name ConnectionEvent#type
+     */
+    Object.defineProperty(this, "type", { value: type });
+  }
+
+  static get OPEN() {
+    return OPEN;
+  }
+
+  static get DATA() {
+    return DATA;
+  }
+
+  static get ERROR() {
+    return ERROR;
+  }
+
+  static get COMPLETE() {
+    return COMPLETE;
+  }
+
+  static get ABORT() {
+    return ABORT;
+  }
+}
--- a/packages/connections/src/XHRConnection.js
+++ b/packages/connections/src/XHRConnection.js
@@ -1,0 +1,216 @@
+import AbstractConnection from "./AbstractConnection";
+import ConnectionEvent from "./ConnectionEvent";
+
+const ALLOWED_METHODS = ["GET", "POST", "PUT", "DELETE"];
+const ALLOWED_RESPONSE_TYPES = [
+  "arraybuffer",
+  "blob",
+  "document",
+  "json",
+  "text"
+];
+
+/**
+ * Basic XHR Connection
+ */
+export default class XHRConnection extends AbstractConnection {
+  /**
+   * Initializes an Instance of XHRConnection
+   * @constructor
+   * @param {string} url – URL to be fetched
+   * @param {object} [options]
+   * @param {string} [options.method=GET] – used method
+   * @param {*} [options.body] – payload for request
+   * @param {object} [options.headers] – additional headers (like content-type)
+   * @param {string} [options.responseType]
+   * @param {int} [timeout]
+   */
+  constructor(url, options = {}) {
+    super(url);
+
+    const {
+      method = "GET",
+      body = null,
+      headers = {},
+      responseType = "json",
+      timeout = 0
+    } = options;
+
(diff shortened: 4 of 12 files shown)


packages/connections/src/XHRConnection.js
```diff
@@ -0,0 +1,170 @@
...
+    super(url);
+
+    const {
+      method = "GET",
+      body = null,
+      headers = {},
+      responseType = "json"
+    } = options;
+
+    if (ALLOWED_METHODS.includes(method) === -1) {
```
nLight: `includes` returns either `true` or `false` 👮

packages/connections/package.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "@dcos/connections",
```
nLight: Do we want to reflect the org structure in the folder structure?
```
packages
- @dcos
-- connections
```

packages/connections/package.json
```diff
@@ -0,0 +1,13 @@
...
+  "version": "0.1.0",
+  "license": "Apache-2.0",
+  "scripts": {
+    "test": "jest"
+  },
+  "devDependencies": {
+    "jest": "19.0.2",
+    "jest-cli": "16.0.2"
+  }
+}
```
nLight: Those `lerna.js` folks specify a main repo as a package repo in `package.json` we might want to follow that as well.

packages/connections/src/XHRConnection.js
```diff
@@ -0,0 +1,170 @@
...
+   * @param {string} [options.responseType]
+   */
+  constructor(url, options = {}) {
+    super(url);
+
+    const {
+      method = "GET",
+      body = null,
+      headers = {},
+      responseType = "json"
```
nLight: Shouldn't that be `application/json`?

orlandohohmeier: Nope, it's just `json`... https://github.com/dcos/dcos-ui/blob/898438d8a6ce629d2743672a77bc31fb81b92f88/packages/connections/README.md#options-optional

packages/connections/src/XHRConnection.js
orlandohohmeier: Good catch!

packages/connections/package.json
orlandohohmeier: I don't think we need to reflect the org structure as all packages belong to `@dcos`

nLight: Not all packages :) i.e. `recordio` and `mesos-client` aren't `@dcos` specific and could live freely. The thing is we either don't use `@dcos` prefix, or put it in a subdirectory as npm would put it in `node_modules`

orlandohohmeier: I'd simply "scope" all packages as they all belong to `dcos` including `recordio` and `mesos-client`.

orlandohohmeier: I'm going to remove the scope from the package for now to unblock the PR and will add the topic to our sync agenda.

packages/connections/src/AbstractConnection.js
```diff
@@ -0,0 +1,115 @@
...
+   * Opens the connection
+   *
+   * @description This method should only be called by Connection Manager,
+   * if you call it yourself be sure you know what you're doing.
+   * Also this method should be the point where a Connection actually starts
+   * using a "Connection Pipe" of the browser.
+   *
+   * @param {string} token – Authentication token
+   */
+  open(token) {}
```
MatApple: Let's remove the argument because it's not required here. If I follow the documentation and inherit from this class then I would expect I need to call `super.open(token)` but I don't need to do that. This documentation belongs on the override in `XHRConnection`

packages/connections/src/AbstractConnection.js
```diff
@@ -0,0 +1,115 @@
...
+  static get OPEN() {
+    return OPEN;
+  }
+
+  static get CLOSED() {
+    return CLOSED;
+  }
+
+  /* Abstract Methods */
+  /* eslint-disable no-unused-vars */
```
MatApple: Let's remove the eslint-disable

packages/connections/src/ConnectionEvent.js
```diff
@@ -0,0 +1,46 @@
...
+const COMPLETE = Symbol("COMPLETE");
+const ABORT = Symbol("ABORT");
+
+export default class ConnectionEvent {
+  /**
+   * Connection event
+   * @param {EventEmitter} target
+   * @param {Symbol} type
+   */
+  constructor(target, type) {
```
MatApple: Let's throw an error if these arguments aren't supplied. Otherwise we've created a type construct that still fails to guarantee structure.

packages/connections/src/XHRConnection.js
```diff
@@ -0,0 +1,170 @@
...
+     * @name XHRConnection#headers
+     */
+    Object.defineProperty(this, "headers", { value: headers });
+
+    /**
+     * @property {string}
+     * @protected
+     * @name XHRConnection#responseType
+     */
+    Object.defineProperty(this, "responseType", { value: responseType });
```
MatApple: Let's make this an ENUM similar to `method` and list the ENUMs in the error for a better debug experience.

packages/connections/src/XHRConnection.js
```diff
@@ -0,0 +1,170 @@
...
+
+    const {
+      method = "GET",
+      body = null,
+      headers = {},
+      responseType = "json"
+    } = options;
+
+    if (ALLOWED_METHODS.includes(method) === -1) {
+      throw new Error("Invalid method for XHR Request.");
```
MatApple: Let's append the allowed methods to the message for better debugging.

packages/connections/src/XHRConnection.js
```diff
@@ -0,0 +1,170 @@
...
+
+    /**
+     * @property {string}
+     * @protected
+     * @name XHRConnection#responseType
+     */
+    Object.defineProperty(this, "responseType", { value: responseType });
+
+    /**
+     * @property {XMLHttpRequest}                 ̵
```
MatApple: zombie `-`

packages/connections/src/XHRConnection.js
```diff
@@ -0,0 +1,170 @@
...
+  get status() {
+    return this.xhr.status;
+  }
+
+  /**
+   * create, prepare, open and send the xhr request
+   * @param {string} [token] – authentication token
+   */
+  open(token) {
+    super.open(token);
```
MatApple: We don't need to pass token to super. Please see other related comment on `AbstractConnection`.

packages/connections/src/XHRConnection.js
```diff
@@ -0,0 +1,170 @@
...
+    });
+
+    this.state = XHRConnection.OPEN;
+    this.xhr.open(this.method, this.url);
+    this.emit(
+      ConnectionEvent.OPEN,
+      new ConnectionEvent(this, ConnectionEvent.OPEN)
+    );
+
+    Object.keys(this.headers).forEach(key => {
```
MatApple: We should check `headers` is an Object before we set it in the constructor.

packages/connections/src/XHRConnection.js
orlandohohmeier: I'm reluctant to add these as some browsers allow types like `moz-blob` or `ms-stream` and I wouldn't want to maintain a list with all possible values. I'm happy to add a check if we restrict the available types and only allow the following types:

Value | Data type of response property
-- | --
"arraybuffer" | ArrayBuffer
"blob" | Blob
"document" | Document
"json" | JavaScript object, parsed from a JSON string returned by the server
"text" | DOMString

packages/connections/src/XHRConnection.js
orlandohohmeier: Yep, will add a check.

packages/connections/src/XHRConnection.js
orlandohohmeier: We should always pass all defined arguments to the super class, even if they'er not used in the super class to allow for future improvements and advancements.

packages/connections/src/AbstractConnection.js
orlandohohmeier: Yeah, this a leftover and you don't need to pass the token.

packages/connections/src/XHRConnection.js
MatApple: I think it's better to be explicit and choose our support. We shouldn't be allowing browser specific response values.

packages/connections/src/XHRConnection.js
MatApple: minor #nit but passing `...arguments` will be more future proof and require less maintenance.

packages/connections/src/AbstractConnection.js
juliangieseke: We had to keep it because of the token parameter in open - if that is removed, we can/should remove this too 👍

packages/connections/src/XHRConnection.js
juliangieseke: At least he didn't try `contains` :trollface: #scnr

packages/connections/src/XHRConnection.js
juliangieseke: After making this package standalone we should consider adding a simple hooks system again. We discarded it before because we were only thinking about the CM, but when these connections could potentionally used by anything (even without the CM), we should make the header injecion "more open".

As a first step I'd just add `addHeader` & `removeHeader` methods, everything else should be held back for a future "v2".

packages/connections/src/AbstractConnection.js
juliangieseke: +1 for removal, also see my comment on `XHRConnection` below.

packages/connections/src/ConnectionEvent.js
```diff
@@ -0,0 +1,46 @@
...
+  static get DATA() {
+    return DATA;
+  }
+
+  static get ERROR() {
+    return ERROR;
+  }
+
+  static get COMPLETE() {
+    return COMPLETE;
```
juliangieseke: I know thinking of an Request `COMPLETE` feels more natural, but as we are talking about Connections here, shouldnt this be `CLOSED`? Also it would be in line with `AbstractConnection.CLOSED`

packages/connections/src/XHRConnection.js
orlandohohmeier: ✔️  Done

packages/connections/src/ConnectionEvent.js
orlandohohmeier: There is a difference between the states (INIT, OPEN, CLOSED) and the events which convey reason.

packages/connections/src/XHRConnection.js
orlandohohmeier: ✔️  Done

juliangieseke:
I hope everything is addressed now 🙏

packages/connections/src/XHRConnection.js
```diff
@@ -0,0 +1,206 @@
...
+
+    if (typeof headers !== "object") {
+      throw new Error(
+        `Invalid headers. Type must be "object", is "${typeof headers}".`
+      );
+    }
+
+    if (!Number.isSafeInteger(timeout) || timeout < 0) {
+      throw new Error(
+        `Invalid timeout. Type must 0 or positive, is "${timeout}".`
```
MatApple: `Invalid timeout. Must be an integer >= 0 and less than Number.MAX_SAFE_INTEGER, is "${timeout}".`

packages/connections/src/XHRConnection.js
```diff
@@ -0,0 +1,206 @@
...
+    return this.xhr.response;
+  }
+
+  get status() {
+    return this.xhr.status;
+  }
+
+  /**
+   * create, prepare, open and send the xhr request
+   * @param {string} [token] – authentication token
```
MatApple: Let's remove the param docstring since we've removed the argument.

packages/connections/src/XHRConnection.js
```diff
@@ -0,0 +1,206 @@
...
+
+  /**
+   * Close the connection
+   * @description Close the connection and abort open requests
+   */
+  close() {
+    super.close();
+
+    if (this.state !== XHRConnection.OPEN) {
+      throw new Error(
```
MatApple: I like what this function did before - silently fail. Since state changes or invocations of close could be sitting on the call stack, the only safe way to ever call close will be to wrap it in a check for `Connection.state === XHRConnection.OPEN` first so we don't accidentally error, so I don't think this makes sense.

juliangieseke: Yes, thats exactly the idea behind this: to force the client to explicitly check what he is doing. 

If someone wants to deal with the Connections & the CM directly, he should know what he is doing and make sure that everything is handled as expected. Otherwise one should use the `Resource Service` (which is TBD).

Assumptions and silent fails are dangerous in some situations: For example, If `you` create a connection, and pass it to `me` (for whatever reason), I dont know if the connection is already closed or not. If I dont check it, but assume its not, add `on('complete',...)` and call `close()`, then it could happend that my added events are never fired.

But I also agree that we shouldnt worry about these details in later use, thats why we want to wrap all this with the `Resource Service` package.

MatApple: Thanks @juliangieseke, I understand your point about silent fails. I'd argue that `you` in your case above should check the `Connection.state` before you attach your `on('complete',...)` to make sure the connection is not `CLOSED` and if it is, handle accordingly. That is your responsibility to handle this and it isn't forced by throwing an error . Hence I think we should not throw an error here because calling close on a `CLOSED` connection doesn't hurt anybody.

juliangieseke: Okay, so we want to throw if a component isnt created correctly, but not if the "wrong" method is called. Also I dont want to have silent fails, which means I need to think about this for a second. Maybe just returning true/false or some errorcodes could be enough. :thinking_face:

juliangieseke:
requests addressed. Please stop using `git commit --amend` from now on, it makes it harder to follow the changes. i will squash everything to one commit before merge. (or whoever else will merge)

packages/connections/src/XHRConnection.js
```diff
@@ -0,0 +1,227 @@
...
+    Object.defineProperty(this, "xhr", { value: new XMLHttpRequest() });
+  }
+
+  get response() {
+    if (this.xhr.response && this.xhr.responseType === this.responseType) {
+      return this.xhr.response;
+    }
+
+    /**
+     * IE11 doesnt support responseType="json" correctly,
```
nLight: Have you tested in IE11?

packages/connections/src/XHRConnection.js
```diff
@@ -0,0 +1,227 @@
...
+    ) {
+      return JSON.parse(this.xhr.response);
+    }
+
+    /**
+     * make sure there is no other weird browser behaviour
+     */
+    if (process.env.NODE_ENV !== "production") {
+      throw new TypeError(
+        `Unexpected XHR responseType (expected "${this.responseType}", was "${this.xhr.responseType}").`
```
nLight: It throws in IE11 in dev mode because for whatever reason as soon as I add listener it calls `response` method and since the connection hasn't been open yet `this.xhr.response` is empty thus `IE11` specific code doesn't run, which results in the exception. Even though we're not developing in IE11 we still test in this browser time to time so it shouldn't happen.

packages/connections/src/XHRConnection.js
nLight: Tested in IE11 it works, but... see the following comment

packages/connections/src/XHRConnection.js
nLight: oh, I know why this happens! Looks like IE11 invokes the getter trying to dump the object in the console. But the thing is if you create an object and then try to use getter it will throw in _any_ browser, but actually it should just return `null`

juliangieseke: fixed 👍

packages/connections/src/ConnectionEvent.js
juliangieseke: done 👍

juliangieseke:
Alfred retest please

orlandohohmeier:

MatApple:
Alfred could you retest this PR?

PR #2419, opened 2017-09-19
Endpoints Tab | 01 | Add actions to request SDK Service Endpoints

bstavroulakis:
These are the new actions to receive the SDK Service Endpoints

Closes DCOS-18144

**Checklist**


plugins/services/src/js/events/SDKEndpointActions.js
```diff
@@ -0,0 +1,77 @@
...
+  REQUEST_SDK_ENDPOINTS_ERROR,
+  REQUEST_SDK_ENDPOINT_SUCCESS,
+  REQUEST_SDK_ENDPOINT_ERROR,
+  REQUEST_SDK_ENDPOINTS_LOADING
+} from "../constants/ActionTypes";
+
+const SDKEndpointsActions = {
+  fetchEndpoints(serviceId) {
+    AppDispatcher.handleServerAction({
+      type: REQUEST_SDK_ENDPOINTS_LOADING,
```
MatApple: I don't think we should introduce this pattern unless there is a strong argument for introducing it. Please follow our existing pattern in react components for identifying loading state or discuss in advance the need for such an introduction, before implementing it :)

bstavroulakis: Removed

plugins/services/src/js/events/SDKEndpointActions.js
```diff
@@ -0,0 +1,77 @@
...
+          },
+          xhr
+        });
+      }
+    });
+  },
+
+  fetchEndpoint(serviceId, endpointName) {
+    const url = `/service/${serviceId}/v1/endpoints/${endpointName}`;
+    const fetch = new XMLHttpRequest();
```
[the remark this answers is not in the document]
bstavroulakis: Renamed to request

plugins/services/src/js/events/SDKEndpointActions.js
```diff
@@ -0,0 +1,77 @@
...
+
+    fetch.open("GET", url);
+    fetch.onreadystatechange = function() {
+      if (fetch.readyState === 4) {
```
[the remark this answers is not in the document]
bstavroulakis: Changed

plugins/services/src/js/events/SDKEndpointActions.js
```diff
@@ -0,0 +1,77 @@
...
+        if (fetch.status === 200) {
+          const contentType = fetch.getResponseHeader("Content-Type");
+
+          AppDispatcher.handleServerAction({
+            type: REQUEST_SDK_ENDPOINT_SUCCESS,
+            data: {
+              serviceId,
+              endpointData: contentType.includes("json")
+                ? JSON.parse(fetch.response)
```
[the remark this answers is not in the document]
bstavroulakis: Yes, would be better probably to keep raw data and parse this in the store. But there are the only 2 cases for now. Maybe we can move over the logic if this changes in the future.

plugins/services/src/js/events/SDKEndpointActions.js
```diff
@@ -0,0 +1,74 @@
...
+const SDKEndpointsActions = {
+  fetchEndpoints(serviceId) {
+    const url = `/service/${serviceId}/v1/endpoints`;
+    RequestUtil.json({
+      url,
+      method: "GET",
+      success(xhr) {
+        AppDispatcher.handleServerAction({
+          type: REQUEST_SDK_ENDPOINTS_SUCCESS,
+          data: { serviceId, endpoints: xhr }
```
MatApple: Do we need to pass the xhr Object as part of the dispatch payload? Usually we only need to do that on the error case. The payload I'd expect to see here for endpoints would be more like `endpoints: RequestUtil.parseResponseBody(xhr)`

bstavroulakis: The parseResponseBody was created for responses that have a xhr.responseJSON obj or a xhr.responseText obj. In this case we return as a body the array of the endpoints so the parseResponseBody in the RequestUtil won't work for us and we can return the data immediately. 

- From the RequestUtil.parseResponseBody(xhr)
```
var responseJSON = xhr.responseJSON;
    var responseText = xhr.responseText;

    if (responseJSON) {
      return responseJSON;
    }
```

bstavroulakis: Renamed to "endpoints"

plugins/services/src/js/events/__tests__/SDKEndpointsActions-test.js
```diff
@@ -0,0 +1,137 @@
...
+        expect(this.configuration.url).toEqual(
+          `${Config.rootUrl}/service/${serviceId}/v1/endpoints`
+        );
+      });
+
+      it("uses GET for the request method", function() {
+        expect(this.configuration.method).toEqual("GET");
+      });
+
+      it("dispatches the correct action when successful", function() {
```
MatApple: Please make sure to test that all expected properties are emitted on both success and error - not just `type`. Sorry I missed this on the first review 😞

bstavroulakis: Yes, listening now to all properties.

plugins/services/src/js/events/__tests__/SDKEndpointsActions-test.js
```diff
@@ -0,0 +1,157 @@
+const RequestUtil = require("mesosphere-shared-reactjs").RequestUtil;
```
nLight: we actually can use `import` in tests as well

plugins/services/src/js/events/SDKEndpointActions.js
```diff
@@ -0,0 +1,74 @@
...
+      }
+    });
+  },
+
+  fetchEndpoint(serviceId, endpointName) {
+    const url = `/service/${serviceId}/v1/endpoints/${endpointName}`;
+    const request = new XMLHttpRequest();
+
+    request.open("GET", url);
+    request.onreadystatechange = function() {
```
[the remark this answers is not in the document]
bstavroulakis: Yes, good idea to be more explicit

plugins/services/src/js/events/__tests__/SDKEndpointsActions-test.js
bstavroulakis: Used the import, it is unfortunate that most of the tests use the require.

orlandohohmeier: Yeah, we should update all of the test till then let's please be consistent with at least the resp. package/plugin.

bstavroulakis: changed back

plugins/services/src/js/events/__tests__/SDKEndpointsActions-test.js
```diff
@@ -0,0 +1,156 @@
...
+
+    it("call open with GET request method and correct url", function() {
+      const url = `/service/${serviceId}/v1/endpoints/${endpointName}`;
+
+      SDKEndpointsActions.fetchEndpoint(serviceId, endpointName);
+
+      expect(mockXhr.open).toHaveBeenCalledWith("GET", url);
+    });
+
+    it("dispatches the correct action when successful", function() {
```
ahoskins: Add another test case for properly handling a response that is *not* JSON?

bstavroulakis: Included a test case for text.

plugins/services/src/js/events/__tests__/SDKEndpointsActions-test.js
```diff
@@ -0,0 +1,181 @@
...
+            endpointName: "arangodb",
+            serviceId: "foo"
+          }
+        });
+      });
+
+      mockXhr.success();
+    });
+
+    it.only(
```
MatApple: Please remove the `.only` ;)

bstavroulakis: Ooops, removed.

plugins/services/src/js/events/__tests__/SDKEndpointsActions-test.js
nLight: We could just do one by one, esp. for all new files.

orlandohohmeier: It's super easy to change them all at once. I'll file a JIRA and schedule for next sprint.

PR #2422, opened 2017-09-19
Endpoints Tab | 04 | DCOS-18156: Tab view components

bstavroulakis:
This is the presentation layer for the endpoints. It includes components that present the endpoint data in the new endpoint tab.

## Case 1 - Non-SDK Service
image

## Case 2 - Multi-container Service
image

## Case 3 - SDK Service
image

## Case 4 - Framework
image

**Checklist**


bstavroulakis:
#2421

plugins/services/src/js/containers/service-connection/SDKServiceConnectionContainer.js
```diff
@@ -0,0 +1,270 @@
...
+    return (
+      <ConfigurationMapRow>
+        <ConfigurationMapLabel>
+          {endpoint.getEndpointName()}
+        </ConfigurationMapLabel>
+        <ConfigurationMapValue>
+          <a
+            className="active endpoint-download"
+            download={endpoint.getEndpointName()}
+            href={`data:text/plain;content-disposition=attachment;filename=${endpoint.getEndpointName()};charset=utf-8,${encodeURIComponent(endpoint.getEndpointData())}`}
```
nLight: Not sure if it makes sense to extract this into a Util or a copmonent

plugins/services/src/js/containers/service-connection/SDKServiceConnectionContainer.js
```diff
@@ -0,0 +1,270 @@
...
+    if (fileEndpoints.length === 0) {
+      return null;
+    }
+
+    return (
+      <ConfigurationMapSection>
+        <ConfigurationMapHeading>
+          Files
+        </ConfigurationMapHeading>
+        {fileEndpoints.map(endpoint => {
```
nLight: I'd rather not mixing the logic and "template"

plugins/services/src/js/containers/service-connection/SDKServiceConnectionContainer.js
```diff
@@ -0,0 +1,270 @@
...
+
+  getAlertPanelNoEndpoints() {
+    return (
+      <AlertPanel>
+        <AlertPanelHeader>No Endpoints</AlertPanelHeader>
+        <p className="tall">
+          There are no endpoints currently configured for
+          {" "}
+          {this.props.service.getId()}
+          .You can edit the configuration to add service endpoints.
```
nLight: Trailing space is missing after the first dot

plugins/services/src/js/containers/service-connection/ServiceConnectionContainer.js
```diff
@@ -0,0 +1,189 @@
...
+const METHODS_TO_BIND = ["handleOpenEditConfigurationModal"];
+
+class ServiceConnectionContainer extends React.Component {
+  constructor() {
+    super(...arguments);
+
+    METHODS_TO_BIND.forEach(method => {
+      this[method] = this[method].bind(this);
+    });
+  }
```
nLight: Let's stick with either having a new line between functions or not 💪

plugins/services/src/js/containers/service-connection/ServiceConnectionContainer.js
```diff
@@ -0,0 +1,189 @@
...
+
+    if (!endpoints || endpoints.length === 0) {
+      return (
+        <AlertPanel>
+          <AlertPanelHeader>No Endpoints</AlertPanelHeader>
+          <p className="tall">
+            There are no endpoints currently configured for
+            {" "}
+            {this.props.service.getId()}
+            .You can edit the configuration to add service endpoints.
```
nLight: Please add the missing trailing space after the dot

plugins/services/src/js/components/HighOrderServiceConnection.js
```diff
@@ -0,0 +1,27 @@
...
+import { isSDKService } from "#SRC/js/utils/ServiceUtil";
+
+import Service from "../structs/Service";
+import Pod from "../structs/Pod";
+import ServiceConnectionContainer
+  from "../containers/service-connection/ServiceConnectionContainer";
+import SDKServiceConnectionContainer
+  from "../containers/service-connection/SDKServiceConnectionContainer";
+
+const HighOrderServiceConnection = function(props) {
```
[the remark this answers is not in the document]
orlandohohmeier: I remember having a rather long discussion about the name back then. Let's please not introduce any new components following this naming scheme. @nLight please file an issue to refactor the "old" bits, to avoid confusions.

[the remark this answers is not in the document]
nLight: https://jira.mesosphere.com/browse/DCOS-18864

plugins/services/src/js/containers/service-connection/ServiceConnectionContainer.js
bstavroulakis: Added

plugins/services/src/js/containers/service-connection/SDKServiceConnectionContainer.js
bstavroulakis: Added

plugins/services/src/js/containers/service-connection/SDKServiceConnectionContainer.js
bstavroulakis: Moved to another function

plugins/services/src/js/containers/service-connection/SDKServiceConnectionContainer.js
bstavroulakis: Created a new Util for this

plugins/services/src/js/components/HighOrderServiceConnection.js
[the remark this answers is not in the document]
bstavroulakis: Ok, renamed the component

bstavroulakis:
Alfred test this

bstavroulakis:
Alfred test this

plugins/services/src/js/containers/service-connection/SDKServiceConnectionEndpointList.js
```diff
@@ -0,0 +1,276 @@
...
+        <ConfigurationMap>
+          {this.getJSONEndpoints(sdkServiceEndpoints)}
+          {this.getFileEndpoints(sdkServiceEndpoints)}
+        </ConfigurationMap>
+      </div>
+    );
+  }
+}
+
+module.exports = SDKServiceConnectionEndpointList;
```
MatApple: Please add `propTypes` :)

plugins/services/src/js/containers/service-connection/SDKServiceConnectionEndpointList.js
```diff
@@ -0,0 +1,286 @@
...
+      copiedCommand: false
+    };
+
+    METHODS_TO_BIND.forEach(method => {
+      this[method] = this[method].bind(this);
+    });
+  }
+
+  handleTextCopy(copiedCommand) {
+    this.setState({ copiedCommand });
```
weblancaster: I noticed that the value for `copiedCommand`ins't boolean but a string, so I wonder if we should change the default state value from `false` to `null`

plugins/services/src/js/containers/service-connection/SDKServiceConnectionEndpointList.js
```diff
@@ -0,0 +1,286 @@
...
+  }
+
+  handleOpenEditConfigurationModal() {
+    this.setState({
+      actionDisabledModalOpen: true
+    });
+  }
+
+  handleActionDisabledModalClose() {
```
weblancaster: I would rather see we consolidate the setState for `actionDisabledModalOpen` in one method but receive the values as arguments. 
e.g `this.handleOpenEditConfigurationModal.bind(this, true)` and for `this.handleActionDisabledModalClose(this, false)`

What do you think?

plugins/services/src/js/containers/service-connection/SDKServiceConnectionEndpointList.js
```diff
@@ -0,0 +1,286 @@
...
+      this.props.service.getId()
+    );
+
+    const sdkServiceError = SDKEndpointStore.getServiceError(
+      this.props.service.getId()
+    );
+
+    const endpointsLoaded =
+      sdkServiceEndpoints &&
+      sdkServiceEndpoints.every(endpoint => {
```
weblancaster: Do we expect `endpoint.endpointData` to be `falsy` in some cases?

weblancaster: after discussing offline `endpoint.endpointData` can be `falsy`

src/styles/components/code/styles.less
```diff
@@ -38,6 +38,11 @@
           }
         }
       }
+
+      .code-copy-icon.tight {
```
MatApple: Let's please not introduce new modifiers like `tight`. Let's reuse what we have from CNVS.

src/styles/components/service-endpoints/styles.less
```diff
@@ -0,0 +1,12 @@
+.endpoint-download,
```
MatApple: Please avoid custom styles. We don't need this file because we can reuse existing classes from CNVS :) Please check other links for structure and behavior.

plugins/services/src/js/containers/service-connection/SDKServiceConnectionEndpointList.js
```diff
@@ -0,0 +1,265 @@
...
+    if (fileEndpoints.length === 0) {
+      return null;
+    }
+
+    return (
+      <ConfigurationMapSection>
+        <ConfigurationMapHeading>
+          Files
+        </ConfigurationMapHeading>
+        {this.getFileEndpointRows(fileEndpoints)}
```
orlandohohmeier: _Not_ saying that's wrong, but what's the benefit of extracting `getFileEndpointRows` into it's own function?

plugins/services/src/js/containers/service-connection/SDKServiceConnectionEndpointList.js
```diff
@@ -0,0 +1,265 @@
...
+    });
+  }
+
+  getFileEndpointRows(fileEndpoints) {
+    return fileEndpoints.map(endpoint => {
+      return this.getFileEndpoint(endpoint);
+    });
+  }
+
+  getFileEndpoints(endpoints) {
```
orlandohohmeier: Please add a check to verify that `endpoints` is actually an array.

plugins/services/src/js/containers/service-connection/SDKServiceConnectionEndpointList.js
```diff
@@ -0,0 +1,265 @@
...
+      sdkServiceEndpoints.every(endpoint => {
+        return endpoint.endpointData;
+      });
+
+    if (!endpointsLoaded && !sdkServiceError) {
+      return <Loader />;
+    }
+
+    if (sdkServiceError) {
+      return this.getAlertPanelSDKDeploying();
```
orlandohohmeier: It feels wrong to render the alert panel SDK *deploying* in case of an error. We should be able to use the service state to either render `getAlertPanelSDKDeploying` or a proper error message.

plugins/services/src/js/containers/service-connection/ServiceConnectionEndpointList.js
```diff
@@ -0,0 +1,244 @@
...
+
+  getProtocolValue(portDefinition) {
+    let protocol = portDefinition.protocol || "";
+    if (Array.isArray(protocol)) {
+      protocol = protocol.join(", ");
+    }
+    protocol = protocol.replace(/,\s*/g, ", ");
+
+    if (protocol instanceof Array) {
+      protocol = protocol.join(", ");
```
orlandohohmeier: Why do we need to join the array a second time?

plugins/services/src/js/containers/service-connection/ServiceConnectionEndpointList.js
```diff
@@ -0,0 +1,244 @@
...
+          </ClipboardTrigger>
+        </div>
+        {command}
+      </div>
+    );
+  }
+
+  getProtocolValue(portDefinition) {
+    let protocol = portDefinition.protocol || "";
+    if (Array.isArray(protocol)) {
```
orlandohohmeier: Why isn't protocol either always an array or an string? Wonder if we're able to remove these checks with the introduction of a service/endpoint portDefinition getter which returns the data always as an array.

plugins/services/src/js/containers/service-connection/ServiceConnectionEndpointList.js
```diff
@@ -0,0 +1,244 @@
...
+      protocol = protocol.join(", ");
+    }
+    protocol = protocol.replace(/,\s*/g, ", ");
+
+    if (protocol instanceof Array) {
+      protocol = protocol.join(", ");
+    }
+
+    if (protocol) {
```
orlandohohmeier: Please explicitly check for empty string as it's otherwise somewhat hard to reason about the respective lines.

plugins/services/src/js/containers/service-connection/ServiceConnectionEndpointList.js
```diff
@@ -0,0 +1,244 @@
...
+      return this.getClipboardTrigger(getDisplayValue(protocol));
+    }
+
+    return getDisplayValue(protocol);
+  }
+
+  getHostPortValue(portDefinition) {
+    const service = this.props.service;
```
orlandohohmeier: Why aren't we passing service (resp. `requirePorts`) to this function? Using props and state within helpers makes in unnecessarily hard to test and bare the risk of unwanted side effects.

plugins/services/src/js/containers/service-connection/ServiceConnectionEndpointList.js
```diff
@@ -0,0 +1,244 @@
...
+  }
+
+  handleOpenEditConfigurationModal() {
+    const { router } = this.context;
+    router.push(
+      `/services/detail/${encodeURIComponent(this.props.service.getId())}/edit/`
+    );
+  }
+
+  getClipboardTrigger(command) {
```
orlandohohmeier: Not a blocker, but I'd prefer local or _private_ components instead of all these helper methods. 

/cc @dcos/frontend-foundation

plugins/services/src/js/containers/service-connection/ServiceConnectionEndpointList.js
orlandohohmeier: Just saw that we've a dedicated `ServicePodConnectionEndpointList` meaning that we should always get a string, if I'm not mistaken. @Poltergeist @nLight any thoughts?

plugins/services/src/js/containers/service-connection/ServicePodConnectionEndpointList.js
```diff
@@ -0,0 +1,176 @@
...
+      </div>
+    );
+  }
+
+  getProtocolValue(portDefinition) {
+    let protocol = portDefinition.protocol || "";
+    if (Array.isArray(protocol)) {
+      protocol = protocol.join(", ");
+    }
+    protocol = protocol.replace(/,\s*/g, ", ");
```
orlandohohmeier: I'm pretty sure that protocol is always be an array for pods, thus there is no need to handle the _string_ case here.

orlandohohmeier:
@bstavroulakis please excuse all these late comments.

plugins/services/src/js/containers/service-connection/SDKServiceConnectionEndpointList.js
```diff
@@ -0,0 +1,265 @@
...
+              {" "}
+              Download
+            </span>
+          </a>
+        </ConfigurationMapValue>
+      </ConfigurationMapRow>
+    );
+  }
+
+  getJSONEndpoints(endpoints) {
```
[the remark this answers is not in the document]
bstavroulakis: Included

plugins/services/src/js/containers/service-connection/SDKServiceConnectionEndpointList.js
```diff
@@ -0,0 +1,265 @@
...
+    const jsonEndpoints = endpoints.filter(endpoint => {
+      return endpoint.isJSON();
+    });
+
+    if (jsonEndpoints.length === 0) {
```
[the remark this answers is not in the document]
bstavroulakis: Yes, it was kept from some previous changes like that. Made it more concise like you recommended, is better

plugins/services/src/js/containers/service-connection/SDKServiceConnectionEndpointList.js
```diff
@@ -0,0 +1,265 @@
...
+      return null;
+    }
+
+    return jsonEndpoints.map(endpoint => {
+      return this.getJSONEndpoint(endpoint);
```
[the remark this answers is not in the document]
bstavroulakis: Included

plugins/services/src/js/containers/service-connection/SDKServiceConnectionEndpointList.js
bstavroulakis: Added

plugins/services/src/js/containers/service-connection/SDKServiceConnectionEndpointList.js
bstavroulakis: Deleted the function and moved it in that area

plugins/services/src/js/containers/service-connection/SDKServiceConnectionEndpointList.js
```diff
@@ -0,0 +1,265 @@
...
+          {this.getFileEndpoints(sdkServiceEndpoints)}
+        </ConfigurationMap>
+      </div>
+    );
+  }
+}
+
+SDKServiceConnectionEndpointList.propTypes = {
+  service: React.PropTypes.oneOfType([
+    React.PropTypes.instanceOf(Pod),
```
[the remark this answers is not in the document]
bstavroulakis: Changed

plugins/services/src/js/containers/service-connection/ServiceConnectionEndpointList.js
bstavroulakis: Extracted to a separate component

src/styles/components/code/styles.less
bstavroulakis: I totally agree, but there is no other implementation or generic rule to do this as globally this image is attached to some default rules which is wrong!

src/styles/components/service-endpoints/styles.less
bstavroulakis: Totally agree, but there is no other implementation or generic rule to do this as much as I searched in the project or the cnvs library.

plugins/services/src/js/containers/service-connection/SDKServiceConnectionEndpointList.js
bstavroulakis: This is because when the server is deploying we receive an error state because it is not ready yet. The error message wouldn't be valuable for the user I think at this point.

bstavroulakis:
@orlandohohmeier No problem, thank you for your thorough review

plugins/services/src/js/containers/service-connection/SDKServiceConnectionEndpointList.js
```diff
@@ -0,0 +1,235 @@
...
+      return null;
+    }
+
+    return (
+      <ConfigurationMapSection>
+        <ConfigurationMapHeading>
+          Files
+        </ConfigurationMapHeading>
+        {fileEndpoints.map((endpoint, index) => {
+          return this.getFileEndpoint(endpoint, index);
```
orlandohohmeier: Not blocking as I believe that using the index in this case should be sufficient, but if possible always prefer an unique element identifier so that React can still identify the element even when you _mutate_ the array.

plugins/services/src/js/containers/service-connection/ServiceConnectionEndpointList.js
orlandohohmeier: Now that we have a `EndpointClipboardTrigger` I don't see any need for a `getClipboardTrigger` method. Anyway, feel free to address this in a later clean up PR.

plugins/services/src/js/containers/service-connection/ServiceConnectionEndpointList.js
```diff
@@ -0,0 +1,216 @@
...
+  }
+
+  render() {
+    const { service } = this.props;
+    let endpoints = [];
+
+    if (service.instances && Array.isArray(service.instances)) {
+      service.instances.forEach(instance => {
+        instance.containers.forEach(container => {
+          endpoints = endpoints.concat(container.endpoints);
```
nLight: Could be `reduce`

plugins/services/src/js/containers/service-connection/ServiceConnectionEndpointList.js
nLight: ...only if there will be a clean up pr following. Otherwise let's do it right away please :)

plugins/services/src/js/containers/service-connection/ServicePodConnectionEndpointList.js
```diff
@@ -0,0 +1,150 @@
...
+  render() {
+    const { service } = this.props;
+    let endpoints = [];
+
+    if (
+      service.spec &&
+      service.spec.containers &&
+      service.spec.containers.length > 0
+    ) {
+      service.spec.containers.forEach(container => {
```
nLight: could be `reduce`

nLight:
@orlandohohmeier you have to re-add your check.

PR #2424, opened 2017-09-19
Design system PRs grouped for review and merge to feature branch

MatApple:
- Move all the content from the tabs for each UI component and consolidate into a single page.
- Adds the page for the Buttons UI component.
- Since tabs are not being used for UI components anymore, the routes need to be updated to point to the single page.
- Previously, each UI component in the design system had 3 tabs: overview, code and style. This content is now consolidated into a single page.

**Checklist**


Fabs:
On the Lean spirit of reducing waste, I am closing this after a chat with team, because it looks to me it won't ever get merged.

PR #2425, opened 2017-09-19
Introduce connection manager

orlandohohmeier:
---
ℹ️   _The respective features were originally implemented by @juliangieseke as part of #2361. This PR extracts the connection manager to streamline the review process._
_This branch depends on changes that will be introduced with #2418. For a list of changes please see: https://github.com/dcos/dcos-ui/compare/orlandohohmeier/feat/connections...orlandohohmeier/feat/connection-manager_

---

The connection manager provides methods to manage connections by priority. This is useful as browsers can only open a limited amount of connections (http) and you might find you urgent request at the very end of a long queue without a dedicated manager.

### DIFF
--- a/packages/connection-manager/src/ConnectionManager.js
+++ b/packages/connection-manager/src/ConnectionManager.js
@@ -1,0 +1,168 @@
+import { AbstractConnection, ConnectionEvent } from "connections";
+import ConnectionQueue from "./ConnectionQueue.js";
+
+/**
+ * The Connection Manager which is responsible for
+ * queuing Connections into the ConnectionQueue and
+ * actually starting them, when they are head of
+ * waiting line.
+ */
+export default class ConnectionManager {
+  /**
+   * Initializes an Instance of ConnectionManager
+   *
+   * @param {int} maxConnections – max open connections
+   */
+  constructor(maxConnections = 1) {
+    /**
+     * Private Context
+     *
+     * @typedef {Object} ConnectionManager~Context
+     */
+    const context = {
+      /**
+       * @property {ConnectionManager} instance
+       * @description Current connection manager instance
+       * @name ConnectionManager~Context#instance
+       */
+      instance: this,
+
+      /**
+       * @property {ConnectionQueue} waitingConnections
+       * @description List of waiting connections ordered by priority
+       * @name ConnectionManager~Context#waitingConnections
+       */
+      waitingConnections: new ConnectionQueue(),
+
+      /**
+       * @property {List} openConnections
+       * @description List of open connections
+       * @name ConnectionManager~Context#next
+       */
+      openConnections: new ConnectionQueue(),
+
+      /**
+       * @property {function} next
+       * @description Opens the the connection if there's a free slot.
+       * @name ConnectionManager~Context#next
+       */
+      next() {
+        if (
+          context.openConnections.size >= maxConnections ||
+          context.waitingConnections.size === 0
+        ) {
+          return;
+        }
+
+        const connection = context.waitingConnections.first();
+
+        if (connection.state === AbstractConnection.INIT) {
+          connection.open(connection.url);
+        }
+
+        if (connection.state === AbstractConnection.OPEN) {
+          context.openConnections = context.openConnections.enqueue(connection);
+        }
+
+        // after added to open list, we can remove it from waiting
+        context.waitingConnections = context.waitingConnections.shift();
+        context.next();
+      },
+
+      /**
+       * @property {function} handleConnectionAbort
+       * @name ConnectionManager~Context#handleConnectionAbort
+       * @param {ConnectionEvent} event
+       */
+      handleConnectionAbort: event => {
+        this.dequeue(event.target);
+      },
+
+      /**
+       * @property {function} handleConnectionComplete
+       * @name ConnectionManager~Context#handleConnectionComplete
+       * @param {ConnectionEvent} event
+       */
+      handleConnectionComplete: event => {
+        this.dequeue(event.target);
+      },
+
+      /**
+       * @property {function} handleConnectionError
+       * @name ConnectionManager~Context#handleConnectionError
+       * @param {ConnectionEvent} event
+       */
+      handleConnectionError: event => {
+        this.dequeue(event.target);
+      }
+    };
+
+    this.enqueue = this.enqueue.bind(context);
+    this.dequeue = this.dequeue.bind(context);
+  }
+
+  /**
+   * Queues given connection with given priority
+   *
+   * @this ConnectionManager~Context
+   * @param {AbstractConnection} connection – connection to queue
+   * @param {Integer} [priority] – optional change of priority
+   */
+  enqueue(connection, priority) {
+    if (connection.state === AbstractConnection.CLOSED) {
+      return;
+    }
+
+    if (connection.state === AbstractConnection.INIT) {
+      this.waitingConnections = this.waitingConnections.enqueue(
+        connection,
+        priority
+      );
+    }
+
+    if (connection.state === AbstractConnection.OPEN) {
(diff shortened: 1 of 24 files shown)


packages/connection-manager/src/ConnectionQueue.js
```diff
@@ -0,0 +1,105 @@
...
+   * @return {ConnectionQueue} - ConnectionQueue containing the added connection
+   */
+  enqueue(connection, priority) {
+    return new ConnectionQueue(
+      this.connections
+        .push(new ConnectionQueueItem(connection, priority))
+        .sort((a, b) => {
+          const delta = b.priority - a.priority;
+
+          return delta / Math.abs(delta || 1);
```
MatApple: Let's enforce sensible bounds on priority in `ConnectionQueue` and remove this line.

packages/connection-manager/src/ConnectionQueueItem.js
```diff
@@ -0,0 +1,50 @@
...
+   * @param {int} [priority=DEFAULT_PRIORITY] – given priority
+   */
+  constructor(connection, priority = DEFAULT_PRIORITY) {
+    if (!(connection instanceof AbstractConnection)) {
+      throw new Error(
+        "Invalid Connection, has to be an instance of AbstractConnection."
+      );
+    }
+
+    if (typeof priority !== "number" || priority < MINIMUM_PRIORITY) {
```
MatApple: Let's use the polyfill for `Number.isInteger(priority)` https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isInteger since it's not available in IE and enforce integer, or change the parameter type to generic number.

packages/connection-manager/src/ConnectionQueueItem.js
```diff
@@ -0,0 +1,50 @@
...
+      throw new Error(
+        "Invalid Priority, has to be a number greater then " +
```
MatApple: greater than

packages/connection-manager/src/ConnectionQueue.js
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  /**
+   * removes a connection from queue
+   * @param {AbstractConnection} connection – connection to be deleted
+   * @return {ConnectionQueue} - ConnectionQueue without the resp. connection
+   */
+  dequeue(connection) {
+    const index = this.connections.findIndex(v =>
+      v.equals(new ConnectionQueueItem(connection))
```
MatApple: Let's avoid creating a new `ConnectionQueueItem` for every equality check.

packages/connection-manager/src/ConnectionQueue.js
```diff
@@ -0,0 +1,105 @@
...
+
+  /**
+   * checks if a connection is already enqueued
+   * @param {AbstractConnection} connection – given connection to check for
+   * @return {boolean} - true if this queue has a connection
+   *  that is equal to the resp. connection, false otherwise.
+   */
+  contains(connection) {
+    return this.connections.some(v =>
+      v.equals(new ConnectionQueueItem(connection))
```
MatApple: Here too.

packages/connection-manager/src/ConnectionQueue.js
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  /**
+   * Adds given connection with given (or default) priority to queue
+   * @param {AbstractConnection} connection – Connection to be added to queue
+   * @param {Integer} priority – priority of connection
+   * @return {ConnectionQueue} - ConnectionQueue containing the added connection
+   */
+  enqueue(connection, priority) {
+    return new ConnectionQueue(
```
MatApple: Let's return early if `this.contains(connection)`

packages/connection-manager/src/ConnectionQueue.js
orlandohohmeier: How does this relate to priority bounds? It the same as the following lines and what we do in [`ServiceTableUtil` ](https://github.com/dcos/dcos-ui/blob/master/plugins/services/src/js/utils/ServiceTableUtil.js#L11)

```js
.sort((a, b) => {
  if(a. priority > b. priority){
    return -1
  }

  if(a. priority < b. priority){
    return 1
  }

  return 0
}
```

juliangieseke:
I hope everything is addressed now 🙏

packages/connection-manager/src/ConnectionQueue.js
MatApple: Enforcing priority to be an integer means we don't have to think about bizarre floating point arithmetic and I think it's much cleaner - especially because we specified the type as integer. We already validate that priority should not be a sub-zero value. Then `return b.priority - a.priority;` is sufficient. We do not need to scale the values to return `-1`, `1` or `0` specifically. If I have missed something, please help me understand why this last line is needed.

packages/connection-manager/src/ConnectionQueueItem.js
MatApple: Edit on above comment: The polyfill ships with our babel toolset already.

juliangieseke:
changed this package to match the behaviour of `connections` (silent fails instead of exceptions)

packages/connection-manager/src/ConnectionManager.js
```diff
@@ -0,0 +1,168 @@
...
+
+    if (connection.state === AbstractConnection.INIT) {
+      this.waitingConnections = this.waitingConnections.enqueue(
+        connection,
+        priority
+      );
+    }
+
+    if (connection.state === AbstractConnection.OPEN) {
+      this.openConnections = this.openConnections.enqueue(connection);
```
nLight: So we can just open all the connections upfront and they go directly into `openConnections` skipping `maxConnections` check :trollface:?

packages/connection-manager/src/ConnectionQueue.js
```diff
@@ -0,0 +1,112 @@
...
+   */
+  enqueue(connection, priority) {
+    const item = new ConnectionQueueItem(connection, priority);
+    const index = this.connections.findIndex(arrayItem =>
+      arrayItem.equals(item)
+    );
+    let connections = null;
+
+    if (index !== -1) {
+      connections = this.connections.update(index, item);
```
nLight: Do we really want to update if `ConnectionQueueItem`s are the same, meaning they hold the same connection?

packages/connection-manager/src/ConnectionQueue.js
```diff
@@ -0,0 +1,112 @@
...
+
+    return new ConnectionQueue(
+      this.connections.filter(arrayItem => !arrayItem.equals(item))
+    );
+  }
+
+  includes(connection) {
+    const item = new ConnectionQueueItem(connection);
+
+    return this.connections.some(v => v.equals(item));
```
nLight: In sometimes there's `arrayItem` being used in this case it's `v` I'd strongly suggest not to use such short names and be consistent across the package. Moreover `arrayItem` is confusing because it's a List and not an array.

packages/connection-manager/src/ConnectionManager.js
juliangieseke: Yes, because if the connection is open, it is open.. it doesnt matter if it shouldnt be opened or not.

packages/connection-manager/src/ConnectionQueue.js
juliangieseke: yup, because the priority might have changed and we cant check it. (or in other words: checking for this would be as expensive as just updating)

packages/connection-manager/src/ConnectionQueue.js
juliangieseke: yup, it should be `listItem` everywhere

PR #2427, opened 2017-09-21
DCOS-18198: Trigger destroy with enter key while in input

bstavroulakis:
Trigger the delete command when you click on enter while in the input area.

image

**Checklist**


bstavroulakis:
Dependent on #2402

PR #2411, opened 2017-09-10
[1.10] DCOS-18055: Improvements for deployment behavior in Catalog

MatApple:
---
⚠️  _This branch depends on changes that are going be introduced with #2413. For a list of changes please see: https://github.com/dcos/dcos-ui/compare/orlandohohmeier/1.10/chore/remove-beta-opt-in...mat/review-and-run_ 

---

On **catalog > package detail** page.
- Consolidates "CONFIGURE" and "DEPLOY" buttons into a single "REVIEW & RUN" button.
- Opens review screen first with options to edit configuration or run service

**To better understand these changes, walk through the commits from top to bottom and read the commit messages.**

**To test**:
- Run Service without editing configuration. Attempt to run service again with same name and witness failure failure.
- Edit configuration for a package to change name and run service. Ensure service is then running.
- Try to run artifactory package without editing. Should fail, show option to edit config. Click option to edit and witness errors in the form.

**Checklist**
- [ ] Did you add a JIRA issue in a commit message or as part of the branch name?
- [ ] Did you add new unit tests?
- [ ] Did you add new integration tests?
- [ ] If this is a regression, did you write a test to catch this in the future?

PR #2428, opened 2017-09-22
feat(SchemaUtil): parse a partial definition

nLight:
We've fixed that in 1.10 as a part of #2411 https://github.com/dcos/dcos-ui/pull/2411/commits/552770ab47e3f849e7e3cabc870682ab0314708f and agreed not to fix on master since the form will be gone in 1.11

But people still stumble upon this bug, so let's fix it just in case.

to test try editing `spark-history` catalog package. Without the fix it throws with the fix it opens the form that looks broken but that's because the package doesn't provide anything to config.


orlandohohmeier:
LGTM 💯 ✨

PR #2430, opened 2017-09-22
fix(CosmosPackages): return entire response for service/describe

ahoskins:
These actions and stores were recently added, and it was a mistake to only return the `package` field of the response. The `package` field is just a static description of the cosmos package. But we also need the up to data options, available in the `resolvedOptions` field of the response. 

This changes these stores/actions to return the entire response from the API. 

**Checklist**

### DIFF
--- a/src/js/stores/__tests__/CosmosPackagesStore-test.js
+++ b/src/js/stores/__tests__/CosmosPackagesStore-test.js
@@ -5,6 +5,7 @@ const Config = require("../../config/Config");
 const EventTypes = require("../../constants/EventTypes");
 const CosmosPackagesStore = require("../CosmosPackagesStore");
 const packageDescribeFixture = require("./fixtures/MockPackageDescribeResponse.json");
+const serviceDescribeFixture = require("./fixtures/MockServiceDescribeResponse.json");
 const packagesListFixture = require("./fixtures/MockPackagesListResponse.json");
 const packagesSearchFixture = require("./fixtures/MockPackagesSearchResponse.json");
 const ActionTypes = require("../../constants/ActionTypes");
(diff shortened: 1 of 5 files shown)


src/js/stores/__tests__/CosmosPackagesStore-test.js
```diff
@@ -306,32 +307,41 @@ describe("CosmosPackagesStore", function() {
...
     afterEach(function() {
       RequestUtil.json = this.requestFn;
     });
 
-    it("should return an instance of UniversePackage", function() {
+    it("should return an instance of UniversePackage within response", function() {
       CosmosPackagesStore.fetchServiceDescription("foo");
-      var serviceDetails = CosmosPackagesStore.getServiceDetails();
+      var response = CosmosPackagesStore.getServiceDetails();
+      var serviceDetails = new UniversePackage(response.package);
```
MatApple: This is a strange test because it's always going to pass given you created the instance here. Can you provide more clarity about the various fields `userProvidedOptions`, `resolvedOptions` and `package`?  It seems to me `CosmosPackagesStore#getServiceDetails` should return an instance of `UniversePackage` where the `resolved` data has been merged. Is that something that makes sense based on what Cosmos is returning?

ahoskins: We talked offline a bit this morning. We want to make sure our store/action design is compatible with framework upgrades in the future. A few conclusions/facts:
1. Editing a framework in the UI involves populating a form with the JSON schema data. This data is available in the `package` field of the response from either `/service/describe` or the existing `package/describe`. The body of these endpoints is different because `service/describe` takes an `appId` in the body, whereas `package/describe` accepts the name and version.
2. `service/describe` also includes the additional data in the response under the key `resolvedOptions`. This is the resolution of the framework defaults and the user provided options. This is the form state when editing the framework in the UI.
3. The `Framework` struct is an appropriate place to hold the `resolvedOptions` data (it represents the current running configuration of the framework).
4. Ideally we only want a single action for getting the JSON schema for a package and returning a `UniversePackage`. This should be `package/describe`, because the naming semantically makes sense and avoids conflating the terms service and package.

At this point, we are still undecided how to move forward with this.

ahoskins:
I think we should go forward with this as-is for now. It's not ideal because we are just sending raw response to the edit component to let it parse what it wants. But for now, it's the simplest way. @MatApple @bstavroulakis 👁

orlandohohmeier:
@ahoskins do we have a JIRA to adjust the data handling as outlined in https://github.com/dcos/dcos-ui/pull/2430#discussion_r140889285?

ahoskins:
@orlandohohmeier To my understanding, it'll likely be something we'll encounter when implementing package upgrades, and/or think about more general store refactors. I don't think I can create a well-defined issue at this point.

ahoskins:
Alfreddo testo

ahoskins:
Alfreddo test this

ahoskins:
Alfreddo test this

PR #2431, opened 2017-09-25
DCOS-18858, DCOS-18949: feat(nodes) - Add Region and Zone information

Poltergeist:
Add a Region column to the nodes overview, displaying the region and zone information of the nodes.
Add Region and Zone information to the nodes detail page.

image

image

Closes DCOS-18858, DCOS-18949, DCOS-18859

**Checklist**

### DIFF
--- a/plugins/nodes/src/js/components/NodesTable.js
+++ b/plugins/nodes/src/js/components/NodesTable.js
@@ -34,9 +34,18 @@ var NodesTable = React.createClass({
     };
   },
 
+  renderRegion(_prop, node) {
+    const name = node.getRegionName() || "\u2014";
+
+    return (
+      <span>
+        {name}
+      </span>
+    );
+  },
+
   renderHeadline(prop, node) {
     let headline = node.get(prop);
-
     if (!node.isActive()) {
       headline = (
         <Tooltip anchor="start" content="Connection to node lost">
@@ -106,6 +115,10 @@ var NodesTable = React.createClass({
         return UnitHealthUtil.getHealthSorting(node);
       }
 
+      if (prop === "region") {
+        return node.getRegionName() || "\u2014";
+      }
+
       return node.get(prop);
     });
 
@@ -115,6 +128,15 @@ var NodesTable = React.createClass({
         headerClassName: className,
         prop: "hostname",
         render: this.renderHeadline,
+        sortable: true,
+        sortFunction,
+        heading
+      },
+      {
+        className,
+        headerClassName: className,
+        prop: "region",
+        render: this.renderRegion,
         sortable: true,
         sortFunction,
         heading
--- a/plugins/nodes/src/js/pages/nodes/NodeDetailTab.js
+++ b/plugins/nodes/src/js/pages/nodes/NodeDetailTab.js
@@ -62,6 +62,22 @@ class NodeDetailTab extends React.Component {
                 {MesosStateStore.get("lastMesosState").version}
               </ConfigurationMapValue>
             </ConfigurationMapRow>
+            <ConfigurationMapRow>
+              <ConfigurationMapLabel>
+                Region
+              </ConfigurationMapLabel>
+              <ConfigurationMapValue>
+                {node.getRegionName() || "\u2014"}
+              </ConfigurationMapValue>
+            </ConfigurationMapRow>
+            <ConfigurationMapRow>
+              <ConfigurationMapLabel>
+                Zone
+              </ConfigurationMapLabel>
+              <ConfigurationMapValue>
+                {node.getZoneName() || "\u2014"}
+              </ConfigurationMapValue>
+            </ConfigurationMapRow>
           </ConfigurationMapSection>
           <HashMapDisplay hash={node.attributes} headline="Attributes" />
           <ConfigurationMapSection>
--- a/src/js/constants/NodesTableHeaderLabels.js
+++ b/src/js/constants/NodesTableHeaderLabels.js
@@ -1,9 +1,10 @@
 var NodesTableHeaderLabels = {
-  health: "HEALTH",
   cpus: "CPU",
   disk: "DISK",
+  health: "HEALTH",
   hostname: "HOSTNAME",
   mem: "MEM",
+  region: "REGION",
   TASK_RUNNING: "TOTAL TASKS"
 };
 
--- a/src/js/structs/Node.js
+++ b/src/js/structs/Node.js
@@ -1,6 +1,7 @@
 import Item from "./Item";
 import TaskStates from "../../../plugins/services/src/js/constants/TaskStates";
 import UnitHealthUtil from "../utils/UnitHealthUtil";
+import { findNestedPropertyInObject } from "../utils/Util";
 
 class Node extends Item {
   getID() {
@@ -13,6 +14,24 @@ class Node extends Item {
 
   isActive() {
     return this.get("active");
+  }
+
+  getDomain() {
+    return this.get("domain");
+  }
+
+  getRegionName() {
+    return findNestedPropertyInObject(
+      this.getDomain(),
+      "fault_domain.region.name"
+    );
+  }
+
+  getZoneName() {
+    return findNestedPropertyInObject(
+      this.getDomain(),
+      "fault_domain.zone.name"
+    );
   }
 
   getUsageStats(resource) {
--- a/src/js/structs/__tests__/Node-test.js
+++ b/src/js/structs/__tests__/Node-test.js
@@ -41,6 +41,72 @@ describe("Node", function() {
     });
   });
 
+  describe("#getDomain", function() {
+    it("returns the domain object of the node", function() {
+      const node = new Node({
+        domain: {
+          fault_domain: {
+            region: {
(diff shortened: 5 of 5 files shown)


orlandohohmeier:
@Poltergeist please adjust the resp. test cases accordingly.

Poltergeist:
alfred please retest this

Poltergeist:

plugins/nodes/src/js/components/NodesTable.js
```diff
@@ -34,9 +34,22 @@ var NodesTable = React.createClass({
     };
   },
 
+  renderRegion(_prop, node) {
+    const domain = node.get("domain");
+    if (domain == null) {
+      return null;
+    }
+    const { region: { name } } = domain.fault_domain;
```
orlandohohmeier: We probably check that `fault_domain` is defined...

plugins/nodes/src/js/components/NodesTable.js
```diff
@@ -106,6 +119,10 @@ var NodesTable = React.createClass({
         return UnitHealthUtil.getHealthSorting(node);
       }
 
+      if (prop === "region") {
+        return node.get("domain").fault_domain.region.name;
```
orlandohohmeier: 🤔  I wonder if it would make sense to create a lens to get the region name /cc @nLight

nLight: Nah... 
```javascript
const lens = Objektiv.attr("fault_domain").attr("region").attr("name");
lens.get(node.get("domain"));
```
I should invest some time to make its interface friendlier

Poltergeist:
@orlandohohmeier I completely overhauled this PR! I also changed the Scope of the PR to also contain Zone column and add region and Zone information to the nodes detail page.

plugins/nodes/src/js/components/NodesTable.js
```diff
@@ -34,9 +34,28 @@ var NodesTable = React.createClass({
     };
   },
 
+  renderRegion(_prop, node) {
+    const name = node.getRegionName();
```
MatApple: I wonder what we should do in the case where Region or Zone is not set 🤔  I believe the direction from @leemunroe is to show `--` for cases where values are not set in tables/configmaps.

src/js/structs/Node.js
```diff
@@ -15,6 +16,24 @@ class Node extends Item {
     return this.get("active");
   }
 
+  getDomain() {
```
MatApple: Please add tests for the new methods.

plugins/nodes/src/js/components/NodesTable.js
leemunroe: Yes, when something isn't configured lets be consistent with what we show. And ideally this is a variable that is easily changed across the UI. I believe @ahoskins may already have something like this for service configuration. Not sure if `--` or `—`. Both work. I'm more concerned about consistency and using the same pattern.

ahoskins: For the framework configuration, I am showing the review screen using the `HashMapDisplay` component. I added a prop called `emptyValue` to this component. I use "\u2014" (&mdash;) as the empty value.

Poltergeist: As far as I understood the feature, this will always be set, and if the user did not set this value mesos is providing a default value. But to be sure I may add a &mdash; (`&mdash;`);

Poltergeist:
Argh forgot to include that in the PR, it was decided that the Zone column should be removed and only the Region column should be present I forgot to update the description accordingly.

plugins/nodes/src/js/pages/nodes/NodeDetailTab.js
```diff
@@ -62,6 +62,22 @@ class NodeDetailTab extends React.Component {
                 {MesosStateStore.get("lastMesosState").version}
               </ConfigurationMapValue>
             </ConfigurationMapRow>
+            <ConfigurationMapRow>
+              <ConfigurationMapLabel>
+                Region
+              </ConfigurationMapLabel>
+              <ConfigurationMapValue>
+                {node.getRegionName() || "\u2014"}
```
weblancaster: We should abstract this into a method where returns `node.getRegionName() || "\u2014"` because we are using multiple times

plugins/nodes/src/js/pages/nodes/NodeDetailTab.js
```diff
@@ -62,6 +62,22 @@ class NodeDetailTab extends React.Component {
...
+              </ConfigurationMapValue>
+            </ConfigurationMapRow>
+            <ConfigurationMapRow>
+              <ConfigurationMapLabel>
+                Zone
+              </ConfigurationMapLabel>
+              <ConfigurationMapValue>
+                {node.getZoneName() || "\u2014"}
```
weblancaster: related to my comment above

src/js/constants/NodesTableHeaderLabels.js
```diff
@@ -1,9 +1,10 @@
 var NodesTableHeaderLabels = {
-  health: "HEALTH",
   cpus: "CPU",
   disk: "DISK",
+  health: "HEALTH",
   hostname: "HOSTNAME",
   mem: "MEM",
+  region: "Region",
```
weblancaster: all `CAPS` for consistency.

plugins/nodes/src/js/pages/nodes/NodeDetailTab.js
Poltergeist: We are using it twice, once per file. I would feel uncomfortable to add it to the struct. Also to add a Util for 2 occurrences feels more like an overhead.

I see this type of change more as being a part from the `ConfigurationMapValue` component. To render a `\u2014` if there is no value. How do you feel about me removing the default value. And create an issue to add the default m dash behaviour to the component. I feel it belongs there.

src/js/constants/NodesTableHeaderLabels.js
Poltergeist: I am going to address this.

plugins/nodes/src/js/pages/nodes/NodeDetailTab.js
Poltergeist: We are using it twice, once per file. I would feel uncomfortable to add it to the struct. Also to add a Util for 2 occurrences feels more like an overhead.

I see this type of change more as being a part from the `ConfigurationMapValue` component. To render a `\u2014` if there is no value. How do you feel about me removing the default value. And create an issue to add the default m dash behaviour to the component. I feel it belongs there.

ahoskins: In my Framework Configuration PR, I added another prop to `HashMapDisplay` for [emptyValue](https://github.com/dcos/dcos-ui/pull/2442/files#diff-22971d009a414c242d15d6666a1e5155R80), where "\u2014" can be specified.

weblancaster: What I meant was to add in the same file not in structs nor a new/or existent util, but this is not a big deal though

Poltergeist: I do not see any value in that as we would have one function which would be called once in that file. Or we would have a util functions like:
```JS
function getValueOrEmptyValue(value) {
  return value || "\u2014";
}

// ...

<ConfigurationMapValue>
  {getValueOrEmptyValue(node.getRegionName())}
</ConfigurationMapValue>
// ...
<ConfigurationMapValue>
  {getValueOrEmptyValue(node.getZoneName())}
</ConfigurationMapValue>


// Or specific

function getRegionName(node) {
  return node.getRegionName() || "\u2014";
}
function getZoneName(node) {
  return node.getZoneName() || "\u2014";
}

// ...

<ConfigurationMapValue>
  {getRegionName(node)}
</ConfigurationMapValue>
// ...
<ConfigurationMapValue>
  {getZoneName(node)}
</ConfigurationMapValue>

```

I see what you want to change here, but I really think that change is better placed in the `ConfigurationMapValue` component. So that the user does not need to care about the empty value. This would result in

```JS
<ConfigurationMapValue>
  {node.getRegionName()}
</ConfigurationMapValue>
```

And regarding adding the emptyValue prop I would not go that far right now as the requirements right now say that empty values only need to be displayed as "\u2014", but I do like the idea to have it configurable as soon as this is becoming a requirement.

weblancaster: makes sense, thanks @Poltergeist

plugins/nodes/src/js/components/NodesTable.js
```diff
@@ -34,9 +34,18 @@ var NodesTable = React.createClass({
...
+
+    return (
+      <span>
+        {name}
```
ahoskins: This is missing the empty state `mdash`! Below, it is used in the sortFunction, but missing here. Should be `name || "\u2014"`

Poltergeist: Added this to line 38 but thanks that one slipped.

PR #2432, opened 2017-09-25
DCOS-18788: adjust port definition protocol handling

orlandohohmeier:
Change the reducers to properly handle missing protocols as they otherwise throw type errors which results in duplicated JSON definitions and a broken editor experience.

Closes DCOS-18788


orlandohohmeier:
Note to self: open a PR to back-port the fix to DC/OS UI 1.10

orlandohohmeier:
@Poltergeist not sure I understand what you're after...

Poltergeist:
Oh my bad ... the `SingleContainerPortDefinitionsReducer.js` does not have any tests. :/ And the PortDefinitions Reducer would just not return any value if tcp and udp is false.

Poltergeist:
💯

orlandohohmeier:
@Poltergeist yeah, though we should add test for respective reducers. Could you please file an JIRA to add the respective tests?

PR #2438, opened 2017-09-29
DCOS-18705: change cluster provisioning

orlandohohmeier:
Adjust the system-tests driver configuration to use `dcos-launch` to provision clusters. 

Please Note: That the systems for this PR will fail as the new setup required `dcos-launch` to be installed. You need to build the docker image locally and then run the `dcos-system-test-driver` after you've provided the respective environment variables. 

Closes DCOS-18705

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -41,6 +41,10 @@ RUN set -x \
   # Install cypress
   && cypress install --cypress-version ${CYPRESS_VERSION} \
 
+  # Install dcos-launch
+  && curl 'https://downloads.dcos.io/dcos-test-utils/bin/linux/dcos-launch' > /usr/local/bin/dcos-launch \
+  && chmod +x /usr/local/bin/dcos-launch \
+
   # Ensure entrypoint is executable
   && chmod +x /usr/local/bin/dcos-ui-docker-entrypoint \
 
--- a/system-tests/_scripts/delete-cluster.sh
+++ b/system-tests/_scripts/delete-cluster.sh
@@ -1,0 +1,11 @@
+#!/usr/bin/env bash
+
+CLUSTER_INFO=/tmp/cluster-info.json
+
+if ! dcos-launch delete -i ${CLUSTER_INFO} 1>&2; then
+  echo "Cluster deletion failed."
+  exit 1
+fi
+
+rm ${CLUSTER_INFO};
+exit 0
--- a/system-tests/_scripts/launch-cluster.sh
+++ b/system-tests/_scripts/launch-cluster.sh
@@ -1,0 +1,55 @@
+#!/usr/bin/env bash
+
+
+if [ -z "${1}" ] && [ -z "${INSTALLER_URL}" ]
+then
+  echo -e "Error: Please pass the installer url or" \
+    "specify the INSTALLER_URL environment variable"
+  exit 1
+fi
+
+CLUSTER_CONFIG=/tmp/cluster-config.yaml
+CLUSTER_INFO=/tmp/cluster-info.json
+
+# Create cluster config
+cat << EOF > ${CLUSTER_CONFIG}
+---
+launch_config_version: 1
+deployment_name: dcos-ui-system-test-$(date +%s)
+installer_url: ${1:-$INSTALLER_URL}
+platform: aws
+provider: onprem
+aws_region: us-west-2
+os_name: cent-os-7
+instance_type: m4.large
+dcos_config:
+  cluster_name: DC/OS UI System Integration Tests
+  resolvers:
+    - 8.8.4.4
+    - 8.8.8.8
+  dns_search: mesos
+  master_discovery: static
+num_masters: 1
+num_private_agents: 1
+num_public_agents: 1
+ssh_user: centos
+key_helper: true
+EOF
+
+if ! dcos-launch create -c ${CLUSTER_CONFIG} -i ${CLUSTER_INFO} 1>&2
+then
+  echo "Error: Cluster launch failed."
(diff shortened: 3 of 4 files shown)


orlandohohmeier:
Note to self: We need to back-port this change to 1.10 and 1.9.

system-tests/_scripts/launch-cluster.sh
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env bash
+
+if [ -z "${INSTALLER_URL}" ]; then
```
nLight: What is `INSTALLER_URL` and where do I take it?

orlandohohmeier: It's the installer dowload url you find on https://dcos.io/releases/ e.g., `https://downloads.dcos.io/dcos/stable/dcos_generate_config.sh`

system-tests/_scripts/launch-cluster.sh
```diff
@@ -0,0 +1,46 @@
...
+os_name: cent-os-7
+instance_type: m4.large
+num_masters: 1
+num_private_agents: 1
+num_public_agents: 1
+dcos_config:
+    cluster_name: DC/OS UI System Integration Tests
+EOF
+
+if ! dcos-launch create; then
```
wavesoft: Do you think it makes sense for clarity reasons to specify the full path to the config (`-c`) and the cluster info (`-i`)?

I would do something like:

```sh
CLUSTER_CONFIG=/tmp/dcos-ui-cluster.conf
CLUSTER_INFO=/tmp/dcos-ui-cluster.info

cat <<EOF > $CLUSTER_CONFIG
...
EOF

if ! dcos-launch create -c $CLUSTER_CONFIG -i $CLUSTER_INFO; then
...
```

system-tests/_scripts/launch-cluster.sh
```diff
@@ -0,0 +1,46 @@
...
+  exit 1
+fi
+
+if [ -z  "${SSH_KEY_PATH}" ]; then
+  echo "Error: Please specify the SSH_KEY_PATH environment variable"
+  exit 1
+fi
+
+# Create cluster config
+cat << EOF > config.yaml
```
wavesoft: I would recommend fixing the path somewhere.

```sh
# Either in the cwd
CLUSTER_CONFIG="./config.yml"
# Or somewhere known
CLUSTER_CONFIG=/tmp/dcos-ui-cluster.conf
# Or somwehre computed
CLUSTER_CONFIG="/var/run/${JENKINS_JOB}/cluster.conf"
```

system-tests/_scripts/launch-cluster.sh
```diff
@@ -0,0 +1,46 @@
...
+num_public_agents: 1
+dcos_config:
+    cluster_name: DC/OS UI System Integration Tests
+EOF
+
+if ! dcos-launch create; then
+  echo "Error: Cluster launch failed."
+  exit 1
+fi
+if ! dcos-launch wait; then
```
wavesoft: If you implement what I mentioned above you will now need to specify the cluster info file (`-i $CLUSTER_INFO`)

system-tests/_scripts/delete-cluster.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+set -euxo pipefail
+
+if ! dcos-launch delete; then
```
wavesoft: Like before, if you implement what I mentioned above you will now need to specify the cluster info file (`-i $CLUSTER_INFO`)

system-tests/_scripts/launch-cluster.sh
wavesoft: Also, be **extra careful** with what goes to STDOUT. Pipe everything to `/dev/null` if you must, but the only string that should be echoed on STDOUT is the path to the cluster!

For example, does `dcos-launch` display progress messages?

orlandohohmeier: Yeah, will adjust the script accordingly.

orlandohohmeier:
@wavesoft I've fixed the path as suggested and changed the logging to errors only so that nothing other then the url gets echoed to STDOUT.

Dockerfile
```diff
@@ -41,6 +41,10 @@ RUN set -x \
   # Install cypress
   && cypress install --cypress-version ${CYPRESS_VERSION} \
 
+  # Install dcos-launch
+  && curl 'https://downloads.dcos.io/dcos-test-utils/bin/linux/dcos-launch' > /usr/local/bin/dcos-launch \
```
weblancaster: Suggestion: Instead of using as `&&` we would benefit from Docker caching if we use multiple `RUN` that way docker will create a layer and re-use if possible.

orlandohohmeier: Yeah, I believe that @wavesoft decided to use only one `RUN` call to reduce the number of mounts which otherwise could potentially lead to issues. I think it should be fine in our case  though.  Anyway, let's not refactor our docker setup in this PR. 

Could you please file a JIRA to revisit our setup?

nLight: There are pros and cons to this. On the one hand we leverage the layer fs, on the other hand the resulting image size is bigger. In addition to that it usually makes sense to have multiple layers only when you expect frequent changes so you leverage caching and incremental updates.

nLight: FYI: created a jira to address empty continuation line warning https://jira.mesosphere.com/browse/DCOS-19026

weblancaster: Great feedback, thanks.

PR #2442, opened 2017-10-05
feat(FrameworkConfiguration): framework configuration components

ahoskins:
This adds a new JSON schema form for editing frameworks. It uses [this](https://github.com/mozilla-services/react-jsonschema-form) Mozilla library for the JSON schema itself.

**Testing**:
Look at #2443 to test edit framework configuration.
Look at #2444 to test deploy framework configuration.

**Description**
1. `FrameworkConfigurationForm.js`: the tabbed form itself, wrapping the Mozilla schema form library and gluing together with `SchemaField.js`, which injects our custom form components (more info below):
screen recording 2017-09-26 at 04 17 pm

2. `FrameworkConfiguration.js`: general coordinator for the configure framework flow. It is a child of the `EditFrameworkConfiguration.js` and `DeployFrameworkConfiguration.js` in the two PRs linked above.
screen recording 2017-09-26 at 04 20 pm

3. `SchemaField.js`: responsible for rendering each form element. We use our existing form components from `src/js/components`. We pass this SchemaField component into the Mozilla library, more information on how this works [here](https://github.com/mozilla-services/react-jsonschema-form#custom-schemafield).

**Smaller additions:**
1. Adds a `FullScreenModalHeaderSubTitle` component:
screen shot 2017-09-26 at 4 12 59 pm

2. Adds optional `ConfigurationMapEditAction` to HashMapDisplay:
screen shot 2017-09-26 at 4 24 08 pm

3. Adds optional error badge to `TabButton`:
screen shot 2017-09-26 at 4 25 28 pm

**Checklist**

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,6 +57,7 @@
     "react-dom": "15.4.1",
     "react-gemini-scrollbar": "2.1.5",
     "react-intl": "2.2.3",
+    "react-jsonschema-form": "0.50.0",
     "react-redux": "4.4.6",
     "react-router": "2.8.1",
     "react-transition-group": "1.2.1",
--- a/src/js/components/CosmosErrorMessage.js
+++ b/src/js/components/CosmosErrorMessage.js
@@ -29,6 +29,11 @@ class CosmosErrorMessage extends React.Component {
       return this.appendRepositoryLink(message);
     }
 
+    // make "Package is already installed error" better
+    if (error.type === "PackageAlreadyInstalled") {
+      return "A service with the same name already exists. Try a different name.";
+    }
+
     return message;
   }
 
--- a/src/js/components/FrameworkConfiguration.js
+++ b/src/js/components/FrameworkConfiguration.js
@@ -1,0 +1,388 @@
+import React, { PropTypes, Component } from "react";
+import { Confirm } from "reactjs-components";
+
+import FullScreenModal from "#SRC/js/components/modals/FullScreenModal";
+import FullScreenModalHeader
+  from "#SRC/js/components/modals/FullScreenModalHeader";
+import FullScreenModalHeaderActions
+  from "#SRC/js/components/modals/FullScreenModalHeaderActions";
+import FullScreenModalHeaderTitle
+  from "#SRC/js/components/modals/FullScreenModalHeaderTitle";
+import FullScreenModalHeaderSubTitle
+  from "#SRC/js/components/modals/FullScreenModalHeaderSubTitle";
+import ToggleButton from "#SRC/js/components/ToggleButton";
+import HashMapDisplay from "#SRC/js/components/HashMapDisplay";
+import ModalHeading from "#SRC/js/components/modals/ModalHeading";
+import UniversePackage from "#SRC/js/structs/UniversePackage";
+import Icon from "#SRC/js/components/Icon";
+import Util from "#SRC/js/utils/Util";
+import StringUtil from "#SRC/js/utils/StringUtil";
+import CosmosErrorMessage from "#SRC/js/components/CosmosErrorMessage";
+import RouterUtil from "#SRC/js/utils/RouterUtil";
+import FrameworkConfigurationForm
+  from "#SRC/js/components/FrameworkConfigurationForm";
+
+const METHODS_TO_BIND = [
+  "handleJSONToggle",
+  "handleGoBack",
+  "handleServiceReview",
+  "handleEditConfigurationButtonClick",
+  "handleActiveTabChange",
+  "handleFocusFieldChange",
+  "handleCloseConfirmModal",
+  "handleConfirmGoBack"
+];
+class FrameworkConfiguration extends Component {
+  constructor(props) {
+    super(props);
+
+    const { activeTab, focusField } = this.getFirstTabAndField();
+
+    this.state = {
+      reviewActive: props.isInitialDeploy,
+      activeTab,
+      focusField,
+      jsonEditorActive: false,
+      isConfirmOpen: false,
+      isOpen: true
+    };
+
+    METHODS_TO_BIND.forEach(method => {
+      this[method] = this[method].bind(this);
+    });
+  }
+
+  handleFocusFieldChange(activeTab, focusField) {
+    this.setState({ focusField, activeTab });
+  }
+
+  handleActiveTabChange(activeTab) {
+    const { packageDetails } = this.props;
+    const schema = packageDetails.getConfig();
+
+    const [focusField] = Object.keys(
+      Util.findNestedPropertyInObject(
+        schema,
+        `properties.${activeTab}.properties`
+      )
+    );
+
+    this.setState({ activeTab, focusField });
+  }
+
+  handleEditConfigurationButtonClick() {
+    const { activeTab, focusField } = this.getFirstTabAndField();
+
+    this.setState({ reviewActive: false, activeTab, focusField });
+  }
+
+  getHashMapRenderKeys(formData) {
+    if (!Util.isObject(formData)) {
+      return {};
+    }
+
+    let renderKeys = {};
+    Object.keys(formData).forEach(key => {
+      renderKeys[key] = StringUtil.capitalizeEveryWord(key);
+      renderKeys = Object.assign(
+        renderKeys,
+        this.getHashMapRenderKeys(formData[key])
+      );
+    });
+
+    return renderKeys;
+  }
+
+  getFirstTabAndField() {
+    const { packageDetails } = this.props;
+    const schema = packageDetails.getConfig();
+
+    const [activeTab] = Object.keys(schema.properties);
+    const [focusField] = Object.keys(
+      Util.findNestedPropertyInObject(
+        schema,
+        `properties.${activeTab}.properties`
+      )
+    );
+
+    return {
+      activeTab,
+      focusField
+    };
+  }
+
+  handleJSONToggle() {
+    this.setState({ jsonEditorActive: !this.state.jsonEditorActive });
+  }
+
+  handleCloseConfirmModal() {
+    this.setState({ isConfirmOpen: false });
+  }
+
+  handleGoBack() {
+    const { reviewActive } = this.state;
+
+    if (reviewActive) {
+      this.setState({ reviewActive: false });
+
+      return;
+    }
+
+    this.setState({ isConfirmOpen: true });
+  }
+
+  handleServiceReview() {
+    const { handleRun } = this.props;
+    const { reviewActive } = this.state;
+
+    if (reviewActive) {
+      handleRun();
+
+      return;
+    }
+
+    this.setState({ reviewActive: true });
+  }
+
+  handleConfirmGoBack() {
+    const { handleGoBack } = this.props;
+
+    this.setState({ isConfirmOpen: false, isOpen: false }, () => {
+      // Once state is set, start a timer for the length of the animation and
+      // navigate away once the animation is over.
+      setTimeout(handleGoBack, 300);
+    });
+  }
+
+  getSecondaryActions() {
+    const { reviewActive } = this.state;
+
+    return [
+      {
+        className: "button-stroke",
+        clickHandler: this.handleGoBack,
+        label: reviewActive ? " Back" : "Cancel"
+      }
+    ];
+  }
+
+  getPrimaryActions() {
+    const { jsonEditorActive, reviewActive } = this.state;
+    const { formErrors } = this.props;
+
+    const actions = [
+      {
+        className: "button-primary flush-vertical",
+        clickHandler: this.handleServiceReview,
+        label: reviewActive ? "Run Service" : "Review & Run",
+        disabled: Object.keys(formErrors).some(tab => formErrors[tab] > 0)
+      }
+    ];
+
+    if (!reviewActive) {
+      actions.unshift({
+        node: (
+          <ToggleButton
+            className="flush"
+            checkboxClassName="toggle-button toggle-button-align-left"
+            checked={jsonEditorActive}
+            onChange={this.handleJSONToggle}
+            key="json-editor"
+          >
+            JSON Editor
+          </ToggleButton>
+        )
+      });
+    }
+
+    return actions;
+  }
+
+  getReviewScreen() {
+    const {
+      packageDetails,
+      deployErrors,
+      isInitialDeploy,
+      formData
+    } = this.props;
+
+    const fileName = "config.json";
+    const configString = JSON.stringify(formData, null, 2);
+
+    const renderKeys = this.getHashMapRenderKeys(formData);
+
+    const preInstallNotes = packageDetails.getPreInstallNotes();
+    let preinstall = null;
+    if (preInstallNotes && isInitialDeploy) {
+      const preInstallNotesParsed = StringUtil.parseMarkdown(preInstallNotes);
+      preInstallNotesParsed.__html =
+        "<strong>Preinstall Notes: </strong>" + preInstallNotesParsed.__html;
+
+      preinstall = (
+        <div
+          dangerouslySetInnerHTML={preInstallNotesParsed}
+          className="pre-install-notes message message-warning"
+        />
+      );
+    }
+
+    let errorsAlert = null;
+    if (deployErrors) {
+      errorsAlert = <CosmosErrorMessage error={deployErrors} />;
+    }
+
+    return (
+      <div className="flex-item-grow-1">
+        <div className="container container-wide">
+          {errorsAlert}
+          {preinstall}
+          <div className="row">
+            <div className="column-4">
+              <h1 className="flush-top">Configuration</h1>
+            </div>
+            <div className="column-8 text-align-right">
+              <button
+                className="button button-primary-link button-inline-flex"
+                onClick={this.handleEditConfigurationButtonClick}
+              >
+                <Icon id="pencil" size="mini" family="system" />
+                <span className="form-group-heading-content">
+                  {"Edit Config"}
+                </span>
+              </button>
+              <a
+                className="button button-primary-link flush-right"
+                download={fileName}
+                onClick={RouterUtil.triggerIEDownload.bind(
+                  null,
+                  fileName,
+                  configString
+                )}
+                href={RouterUtil.getResourceDownloadPath(
+                  "attachment/json",
+                  fileName,
+                  configString
+                )}
+              >
+                <Icon id="download" size="mini" family="system" />
+                <span className="form-group-heading-content">
+                  {"Download Config"}
+                </span>
+              </a>
+            </div>
+          </div>
+          <HashMapDisplay
+            hash={formData}
+            renderKeys={renderKeys}
+            headlineClassName={"text-capitalize"}
+            emptyValue={"\u2014"}
+          />
+        </div>
+      </div>
+    );
+  }
+
+  getHeader() {
+    const { reviewActive } = this.state;
+    const { packageDetails } = this.props;
+
+    return (
+      <FullScreenModalHeader>
+        <FullScreenModalHeaderActions
+          actions={this.getSecondaryActions()}
+          type="secondary"
+        />
+        <FullScreenModalHeaderTitle className="modal-full-screen-header-with-sub-title">
+          {reviewActive ? "Review Configuration" : "Edit Configuration"}
+          <FullScreenModalHeaderSubTitle>
+            {StringUtil.capitalize(packageDetails.getName()) +
+              " " +
+              packageDetails.getVersion()}
+          </FullScreenModalHeaderSubTitle>
+        </FullScreenModalHeaderTitle>
+        <FullScreenModalHeaderActions
+          actions={this.getPrimaryActions()}
+          type="primary"
+        />
+      </FullScreenModalHeader>
+    );
+  }
+
+  render() {
+    const {
+      isConfirmOpen,
+      isOpen,
+      reviewActive,
+      jsonEditorActive,
+      focusField,
+      activeTab
+    } = this.state;
+    const {
+      packageDetails,
+      deployErrors,
+      formErrors,
+      formData,
+      onFormDataChange,
+      onFormErrorChange
+    } = this.props;
+
+    let pageContents;
+    if (reviewActive) {
+      pageContents = this.getReviewScreen();
+    } else {
+      pageContents = (
+        <FrameworkConfigurationForm
+          packageDetails={packageDetails}
+          jsonEditorActive={jsonEditorActive}
+          formData={formData}
+          formErrors={formErrors}
+          focusField={focusField}
+          activeTab={activeTab}
+          deployErrors={deployErrors}
+          onFormDataChange={onFormDataChange}
+          onFormErrorChange={onFormErrorChange}
+          handleActiveTabChange={this.handleActiveTabChange}
+          handleFocusFieldChange={this.handleFocusFieldChange}
+        />
+      );
+    }
+
+    return (
+      <FullScreenModal
+        header={this.getHeader()}
+        useGemini={reviewActive}
+        open={isOpen}
+      >
+        {pageContents}
+        <Confirm
+          closeByBackdropClick={true}
+          header={<ModalHeading>Discard Changes?</ModalHeading>}
+          open={isConfirmOpen}
+          onClose={this.handleCloseConfirmModal}
+          leftButtonText="Cancel"
+          leftButtonCallback={this.handleCloseConfirmModal}
+          rightButtonText="Discard"
+          rightButtonClassName="button button-danger"
+          rightButtonCallback={this.handleConfirmGoBack}
+          showHeader={true}
+        >
+          <p>
+            Are you sure you want to leave this page? Any data you entered will be lost.
+          </p>
+        </Confirm>
+      </FullScreenModal>
+    );
+  }
+}
+
+FrameworkConfiguration.propTypes = {
+  formData: PropTypes.object.isRequired,
+  onFormDataChange: PropTypes.func.isRequired,
+  packageDetails: PropTypes.instanceOf(UniversePackage).isRequired,
+  handleRun: PropTypes.func.isRequired,
+  handleGoBack: PropTypes.func.isRequired,
+  isInitialDeploy: PropTypes.bool.isRequired,
+  deployErrors: PropTypes.object
+};
+
+module.exports = FrameworkConfiguration;
--- a/src/js/components/FrameworkConfigurationForm.js
+++ b/src/js/components/FrameworkConfigurationForm.js
@@ -1,0 +1,321 @@
+import React, { Component, PropTypes } from "react";
+import classNames from "classnames";
+import SchemaForm from "react-jsonschema-form";
+import TabButton from "#SRC/js/components/TabButton";
+import TabButtonList from "#SRC/js/components/TabButtonList";
+import Tabs from "#SRC/js/components/Tabs";
+import Util from "#SRC/js/utils/Util";
+import JSONEditor from "#SRC/js/components/JSONEditor";
+import FluidGeminiScrollbar from "#SRC/js/components/FluidGeminiScrollbar";
+import PageHeaderNavigationDropdown
+  from "#SRC/js/components/PageHeaderNavigationDropdown";
+import UniversePackage from "#SRC/js/structs/UniversePackage";
+import CosmosErrorMessage from "#SRC/js/components/CosmosErrorMessage";
+import SchemaField from "#SRC/js/components/SchemaField";
+import StringUtil from "#SRC/js/utils/StringUtil";
+
+const METHODS_TO_BIND = [
+  "handleDropdownNavigationSelection",
+  "handleTabChange",
+  "handleFormChange",
+  "handleJSONChange",
+  "handleBadgeClick",
+  "validate"
+];
+class FrameworkConfigurationForm extends Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      errorSchema: null
+    };
+
+    METHODS_TO_BIND.forEach(method => {
+      this[method] = this[method].bind(this);
+    });
+  }
+
+  handleBadgeClick(activeTab, event) {
+    event.stopPropagation();
+
+    const { errorSchema } = this.state;
+    const { formData } = this.props;
+
+    const fieldsWithErrors = Object.keys(
+      errorSchema[activeTab]
+    ).filter(field => {
+      return (
+        errorSchema[activeTab][field].__errors &&
+        errorSchema[activeTab][field].__errors.length > 0
+      );
+    });
+
+    // first field with errors in the current tab
+    const fieldToFocus = Object.keys(formData[activeTab]).find(field => {
+      return fieldsWithErrors.includes(field);
+    });
+
+    this.props.handleFocusFieldChange(activeTab, fieldToFocus);
+  }
+
+  getFormTabList() {
+    const { formErrors, packageDetails } = this.props;
+    const schema = packageDetails.getConfig();
+
+    // the config will have 2-levels, we will render first level as the tabs
+    return Object.keys(schema.properties).map(tabName => {
+      return (
+        <TabButton
+          label={StringUtil.capitalizeEveryWord(tabName)}
+          labelClassName={"padded-right"}
+          id={tabName}
+          key={tabName}
+          showErrorBadge={formErrors[tabName] > 0}
+          count={formErrors[tabName]}
+          description={`${formErrors[tabName]} issues need addressing`}
+          onClickBadge={this.handleBadgeClick.bind(this, tabName)}
+        />
+      );
+    });
+  }
+
+  handleTabChange(activeTab) {
+    this.props.handleActiveTabChange(activeTab);
+  }
+
+  handleDropdownNavigationSelection(item) {
+    this.props.handleActiveTabChange(item.id);
+  }
+
+  getDropdownNavigationList() {
+    const { packageDetails, activeTab } = this.props;
+    const schema = packageDetails.getConfig();
+
+    return Object.keys(schema.properties).map(tabName => {
+      return {
+        id: tabName,
+        isActive: activeTab === tabName,
+        label: tabName
+      };
+    });
+  }
+
+  getUiSchema() {
+    const { formData, focusField, activeTab } = this.props;
+
+    const uiSchema = {
+      [activeTab]: {
+        [focusField]: { "ui:autofocus": true }
+      }
+    };
+
+    // hide all tabs not selected
+    Object.keys(formData).forEach(tabName => {
+      if (tabName !== activeTab) {
+        if (uiSchema[tabName] == null) {
+          uiSchema[tabName] = {};
+        }
+        uiSchema[tabName]["classNames"] = "hidden";
+      }
+    });
+
+    return uiSchema;
+  }
+
+  handleJSONChange(formData) {
+    this.props.onFormDataChange(formData);
+  }
+
+  writeErrors(formData, schema, isRequired, errors) {
+    if (Util.isObject(formData)) {
+      Object.keys(formData).forEach(property => {
+        this.writeErrors(
+          formData[property],
+          schema.properties[property],
+          schema.required && schema.required.includes(property),
+          errors[property]
+        );
+      });
+
+      return;
+    }
+
+    if (isRequired && formData === "") {
+      let message = `Expecting a ${schema.type} here`;
+      if (schema.default) {
+        message = `Expecting a ${schema.type} here, eg: ${schema.default}`;
+      }
+      errors.addError(message);
+    }
+  }
+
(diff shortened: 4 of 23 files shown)


orlandohohmeier:

src/js/components/FrameworkConfiguration.js
```diff
@@ -0,0 +1,381 @@
...
+export default class FrameworkConfiguration extends Component {
+  constructor(props) {
+    super(props);
+
+    const schema = props.packageDetails.config;
+    const activeTab = Object.keys(schema.properties)[0];
+    const firstField = Object.keys(schema.properties[activeTab].properties)[0];
+    const focusFieldPath = [activeTab, firstField];
+
+    this.state = {
```
orlandohohmeier: Not blocking as this is very much  in line with the `CreateServiceModal` component but we need to refactor the way we're handling state as the great number of entangled properties is highly error prone. For details please see [DCOS-13173](https://jira.mesosphere.com/browse/DCOS-13173).

ahoskins: @orlandohohmeier I agree and am not satisfied with complexity of state/props in these components. I read that issue, but can you elaborate / give-example of what this kind of "state machine" would look like?

ahoskins:
Alfreddo testo this

ahoskins:
Alfreddo test this

src/js/components/HashMapDisplay.js
```diff
@@ -31,7 +33,14 @@ class HashMapDisplay extends React.Component {
...
+      hash,
+      headingLevel,
+      renderKeys,
+      onEditClick,
+      keyPath,
+      emptyValue
+    } = this.props;
 
     return Object.keys(hash).map((key, index) => {
       let value = hash[key];
```
bstavroulakis: Maybe this could be improved with 
`let value = hash[key] || "";`
because if it is undefined or null it may throw an error with the join

package.json
```diff
@@ -57,6 +57,7 @@
     "react-dom": "15.4.1",
     "react-gemini-scrollbar": "2.1.5",
     "react-intl": "2.2.3",
+    "react-jsonschema-form": "0.50.0",
```
bstavroulakis: Don't we have to run shrinkwrap since we included a new package?

src/js/components/FrameworkConfigurationForm.js
```diff
@@ -0,0 +1,315 @@
...
+export default class FrameworkConfigurationForm extends Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      errorSchema: null
+    };
+  }
+
+  getFormattedSectionLabel(label) {
```
bstavroulakis: label = "" to check for null or undefined state?

src/js/components/FrameworkConfiguration.js
```diff
@@ -0,0 +1,381 @@
...
+      return;
+    }
+
+    this.setState({ focusFieldPath });
+  }
+
+  onActiveTabChange(activeTab) {
+    const { focusFieldPath } = this.state;
+    const { packageDetails } = this.props;
+    const currentActiveTab = focusFieldPath[0];
```
bstavroulakis: Do we know that the focusFieldPath will always be an array and have at least one item?

src/js/components/FrameworkConfiguration.js
```diff
@@ -0,0 +1,381 @@
...
+
+    if (deepEqual(activeTab, currentActiveTab)) {
+      return;
+    }
+
+    const schema = packageDetails.config;
+    const firstField = Object.keys(schema.properties[activeTab].properties)[0];
```
bstavroulakis: Again if there is no items this will become undefined, is that ok?

src/js/components/FrameworkConfiguration.js
```diff
@@ -0,0 +1,381 @@
...
+    const newFocusFieldPath = [activeTab, firstField];
+
+    this.setState({ focusFieldPath: newFocusFieldPath });
+  }
+
+  onReviewConfigurationRowClick(rowData) {
+    const { tabViewID } = rowData;
+
+    const focusFieldPath = tabViewID.split(".");
```
bstavroulakis: a check maybe to avoid errors tabViewID || "";
Or is the tabViewID always set?

src/js/components/SchemaField.js
```diff
@@ -0,0 +1,299 @@
...
+        onChange={event => onChange(event.target.value)}
+        onBlur={onBlur && (event => onBlur(name, event.target.value))}
+        onFocus={onFocus && (event => onFocus(name, event.target.value))}
+      >
+        {options}
+      </FieldSelect>
+    );
+    if (autofocus) {
+      field = (
+        <FieldAutofocus>
```
bstavroulakis: Awesome component here!

src/js/components/FrameworkConfigurationForm.js
orlandohohmeier: @bstavroulakis please note that defaults only apply if the value is undefined.

```js
const { a = "default", b = "default" } = { a: undefined, b: null};

console.log(a) //  default
console.log(b) //  null
```

src/js/components/FrameworkConfigurationForm.js
```diff
@@ -0,0 +1,315 @@
...
+            )}
+            items={this.getDropdownNavigationList()}
+          />
+          <div className="modal-body-padding-surrogate create-service-modal-form-container">
+            <div className="create-service-modal-form container container-wide">
+              <Tabs
+                activeTab={activeTab}
+                handleTabChange={this.handleTabChange.bind(this)}
+                vertical={true}
+                className={"menu-tabbed-container-fixed"}
```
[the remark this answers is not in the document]
orlandohohmeier: @Poltergeist I'm pretty sure that you or @jfurrow fixed this in the service form. Could you please help @ahoskins apply the _same_ fix?

src/js/components/FrameworkConfiguration.js
ahoskins: It should be. But I will extract this into an extra state item. I agree, it's a bit scary and strange how it's currently done.

src/js/components/FrameworkConfiguration.js
```diff
@@ -0,0 +1,381 @@
...
+                  <Icon id="pencil" size="mini" family="system" />
+                  <span className="form-group-heading-content">
+                    {"Edit Config"}
+                  </span>
+                </button>
+                <a
+                  className="button button-primary-link flush-right"
+                  download={fileName}
+                  onClick={ieDownloadConfig}
+                  href={`data:attachment/json;content-disposition=attachment;filename=${fileName};charset=utf-8,${encodeURIComponent(configString)}`}
```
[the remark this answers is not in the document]
ahoskins: This method is in @bstavroulakis [endpoints PR](https://github.com/dcos/dcos-ui/pull/2422). We expect his endpoints PR to merge first, and then I will rebase and use his helper.

src/js/components/SchemaField.js
```diff
@@ -0,0 +1,299 @@
...
+        </FormGroupHeadingContent>
+      </FormGroupHeading>
+    );
+  }
+
+  getFieldContent(errorMessage, autofocus) {
+    const { schema } = this.props;
+    const RADIO_SELECT_THRESHOLD = 4;
+
+    if (schema.type === "boolean") {
```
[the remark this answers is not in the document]
ahoskins: Sure, can change this.

src/js/components/SchemaField.js
```diff
@@ -0,0 +1,299 @@
...
+
+    const options = schema.enum.map(option => {
+      return (
+        <FieldLabel key={option}>
+          <FieldInput
+            type={"radio"}
+            value={option}
+            name={name}
+            checked={option === formData}
+            onChange={_ => onChange(option)}
```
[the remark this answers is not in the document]
ahoskins: I will change to () =>

src/js/components/FrameworkConfigurationForm.js
ahoskins: The `label` will not be undefined or null. This method is directly called for property names from the JSON schema. I would prefer to avoid this check, I believe it's overly defensive.

src/js/components/FrameworkConfiguration.js
```diff
@@ -0,0 +1,381 @@
...
+                  </span>
+                </a>
+              </div>
+            </div>
+            <HashMapDisplay
+              hash={formData}
+              renderKeys={renderKeys}
+              onEditClick={this.onReviewConfigurationRowClick.bind(this)}
+              headlineClassName={"text-capitalize"}
+              emptyValue={"\u2014"}
```
weblancaster: https://github.com/dcos/dcos-ui/pull/2447/files#diff-c7a540e5f0f308737396e460f6eb98cdR1

src/js/components/FrameworkConfiguration.js
```diff
@@ -0,0 +1,381 @@
...
+    this.setState({ focusFieldPath });
+  }
+
+  onActiveTabChange(activeTab) {
+    const { focusFieldPath } = this.state;
+    const { packageDetails } = this.props;
+    const currentActiveTab = focusFieldPath[0];
+
+    if (deepEqual(activeTab, currentActiveTab)) {
+      return;
```
weblancaster: Explicitly `return false` instead relying on engine coercion which will be undefined in this case.

src/js/components/FrameworkConfiguration.js
ahoskins: I added safety to this as you suggested.

src/js/components/FrameworkConfiguration.js
ahoskins: As talked offline today with @MatApple , we will assume that all packages will have at least two levels of config properties. I also talked to Cosmos team and they confirmed there are no packages they know of that don't have two-level object as the config.

src/js/components/FrameworkConfigurationForm.js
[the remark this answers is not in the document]
ahoskins: As far as I can tell, this has the same styling as the service form. It will ellipsis when a single word is longer than the space available. And it will wrap when there are many words it can split on. *But I've added a new class `padded-right` to the menu-tabbed, which gives it 5px of spacing.*  Now it looks like the below.

Ellipsis:
screen shot 2017-10-11 at 3 26 15 pm

Wrap:
screen shot 2017-10-11 at 3 25 17 pm

src/js/components/FrameworkConfiguration.js
ahoskins: I've changed this by adding a dedicated state item for the `activeTab`.

src/js/components/FrameworkConfigurationForm.js
```diff
@@ -0,0 +1,315 @@
...
+  }
+
+  handleBadgeClick(tabName, event) {
+    const { errorSchema } = this.state;
+
+    const newFocusFieldPath = this.getFirstErrorFieldPath(tabName, errorSchema);
+
+    this.props.onFocusFieldPathChange(newFocusFieldPath);
+
+    event.stopPropagation();
```
weblancaster: I'm going to guess that if you put `event.stopPropagation()` at the top doesn't work as expected?

src/js/components/HashMapDisplay.js
ahoskins: I don't this is needed because we are mapping over its keys. So when when access it's key it can't be undefined.

src/js/components/FrameworkConfiguration.js
ahoskins: Nice, I will use that component once your PR merges. For now, I'll add this as a todo in this code.

src/js/components/FrameworkConfiguration.js
```diff
@@ -0,0 +1,381 @@
...
+    const firstFieldName = Object.keys(
+      schema.properties[firstTabName].properties
+    )[0];
+
+    return [firstTabName, firstFieldName];
+  }
+
+  getHashMapRenderKeys(formData, renderKeys) {
+    if (!Util.isObject(formData)) {
+      return;
```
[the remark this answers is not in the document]
ahoskins: Thanks, will change all of these.

src/js/components/FrameworkConfigurationForm.js
ahoskins: I've moved this to the top, it works either way, but nicer at the top so I'll put it there.

src/js/components/FrameworkConfigurationForm.js
```diff
@@ -0,0 +1,315 @@
...
+      Object.keys(formData).forEach(property => {
+        this.writeErrors(
+          formData[property],
+          schema.properties[property],
+          schema.required && schema.required.includes(property),
+          errors[property]
+        );
+      });
+
+      return;
```
weblancaster: same suggestion as above about return

src/js/components/FrameworkConfigurationForm.js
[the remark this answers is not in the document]
jfurrow: 👋 Oh hi! Lookin' good!

I think I only added padding for the container names when configuring a multi-container service. Maybe I'm wrong about that — but in any case I should have applied it to all of these tabs!

nLight:
Does the JSON Editor component integration belongs to this PR? The editor behaves differently as opposed to how it works in the Services form. In service form it scrolls independently, wheres in this form they're connected which creates a lag when I'm hitting end of one of the containers.

oct-12-2017 14-46-14

src/js/components/FrameworkConfigurationForm.js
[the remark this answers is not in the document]
Poltergeist: Hey John Tanks for the Feedback. :-) This what you see there is basically the new advanced configuration view which Andrew is about to finish and I would have suggested the same. But Thanks for jumping in, that was unexpected.

src/js/components/FrameworkConfiguration.js
```diff
@@ -0,0 +1,387 @@
...
+    const focusField = keyPath[1];
+
+    this.setState({ reviewActive: false, focusField, activeTab });
+  }
+
+  onEditConfigurationButtonClick() {
+    const { packageDetails } = this.props;
+
+    const schema = packageDetails.getConfig();
+    const activeTab = Object.keys(schema.properties)[0];
```
Poltergeist: Another option to get the first element would be:
```JS
 const [activeTab] = Object.keys(schema.properties);
```
But this is personal taste. As it is not adding a lot of readability feel free to adjust in this and next line.

src/js/components/FrameworkConfiguration.js
```diff
@@ -0,0 +1,387 @@
...
+  }
+
+  handleCloseConfirmModal() {
+    this.setState({ isConfirmOpen: false });
+  }
+
+  handleGoBack() {
+    const { reviewActive, hasChangesApplied } = this.state;
+
+    if (reviewActive && hasChangesApplied) {
```
Poltergeist: Personal taste I would prefer early exit as it will remove complexity
```JS
if (reviewActive && hasChangesApplied) {
   this.setState({ reviewActive: false });

   return;
}
if (hasChangesApplied) {
  this.setState({ isConfirmOpen: true });

  return;
}
this.handleConfirmGoBack();
```

Feel free to adjust.

src/js/components/FrameworkConfiguration.js
```diff
@@ -0,0 +1,387 @@
...
+    } else {
+      this.handleConfirmGoBack();
+    }
+  }
+
+  handleServiceReview() {
+    const { handleRun } = this.props;
+    const { reviewActive } = this.state;
+
+    if (reviewActive) {
```
Poltergeist: Here we also could early exit and only have.
```JS
if (reviewActive) {
  handleRun();

  return;
}

this.setState({ reviewActive: true });
```

src/js/components/FrameworkConfigurationForm.js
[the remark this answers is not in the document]
MatApple: @jfurrow are you ready to come back yet?

src/js/components/FrameworkConfiguration.js
```diff
@@ -0,0 +1,387 @@
...
+    } = this.state;
+    const {
+      packageDetails,
+      deployErrors,
+      formErrors,
+      isInitialDeploy,
+      formData
+    } = this.props;
+
+    if (!reviewActive) {
```
Poltergeist: Could we negate this condition and put the code for the review screen into a helper function like `getReviewScreen`. This would increase the readability of this function.

src/js/components/FrameworkConfigurationForm.js
```diff
@@ -0,0 +1,308 @@
...
+      errorSchema: null
+    };
+  }
+
+  getFormattedSectionLabel(label) {
+    return label
+      .toLowerCase()
+      .split("_")
+      .map(word => word[0].toUpperCase() + word.slice(1))
+      .join(" ");
```
Poltergeist: 👍

src/js/components/FrameworkConfigurationForm.js
```diff
@@ -0,0 +1,308 @@
...
+              >
+                <TabButtonList>
+                  {this.getFormTabList()}
+                </TabButtonList>
+                <div className="menu-tabbed-view-container">
+                  {errorsAlert}
+                  <SchemaForm
+                    schema={schema}
+                    formData={formData}
+                    onChange={this.handleFormChange.bind(this)}
```
Poltergeist: This will always create a new function which might mess up the rerendering as the props are changing. To omit this we could bind this to the function in the constructor

```JS
this.handleFormChange = this.handleFormChange.bind(this);
```

This also applies for the `validate` prop.

src/js/components/FrameworkConfigurationForm.js
```diff
@@ -0,0 +1,308 @@
...
+            </div>
+          </div>
+        </div>
+        <div className={jsonEditorPlaceholderClasses} />
+        <div className={jsonEditorClasses}>
+          <JSONEditor
+            errors={jsonEditorErrors}
+            showGutter={true}
+            showPrintMargin={false}
+            onChange={this.handleJSONChange.bind(this)}
```
Poltergeist: Same as above let's bind this in the constructor.

src/js/components/SchemaField.js
```diff
@@ -0,0 +1,306 @@
...
+          {this.getFieldHeading(required, name, schema.description)}
+        </FieldLabel>
+        {field}
+        <FieldError>{errorMessage}</FieldError>
+      </div>
+    );
+  }
+
+  getFieldHeading(required, name = "", description) {
+    let asterisk = null;
```
Poltergeist: Very personal opinion about the name.
I would rather use something like `requiredSign` because if we decide to change the sign from an asterisk to an exclamation mark we also need to rename the variable. So I would rather opt for a more generic name which is describing what it does and not how it looks like.

Feel free to adjust.

src/js/components/FrameworkConfiguration.js
ahoskins: Yes, this will be much nicer. Thanks, will do this.

src/js/components/FrameworkConfigurationForm.js
ahoskins: Yeah, I was going to change all these to bind in the constructor! Thanks for reminder.

src/js/components/SchemaField.js
ahoskins: I agree with that reasoning 👍

ahoskins:
@nLight Thanks for pointing out that scrolling. I forgot to wrap the form section in the Gemini Scrollbar. Much better now 👍

src/js/components/FrameworkConfigurationForm.js
```diff
@@ -0,0 +1,329 @@
...
+      return (
+        <h2 className="flush-top short-bottom">
+          {this.getFormattedSectionLabel(props.title)}
+        </h2>
+      );
+    };
+
+    const jsonEditorPlaceholderClasses = classNames(
+      "modal-full-screen-side-panel-placeholder",
+      { "is-visible": jsonEditorActive }
```
bstavroulakis: There seems to be an issue with the JsonEditor toggle 

json-editor-toggle

ahoskins: I can't reproduce this, but I found another issue with the JSON editor and tabs on Firefox. I will test on IE / Edge this morning. This form/tabs/json-editor is equivalent to the marathon app form, with the only different being the form tabs are `position: fixed` in this framework form.

orlandohohmeier:
@ahoskins having commits like "changes from Philipps review"  (6319720), " fix scrolling of form/json and better duplicate package error" (102308d), or " fix minor tab focus and ordering issues" (543ca2c) don't provide much value and are not in line with our  contributing guidelines. Please squash all commits or rephrase following the guidelines?

ahoskins:
alfreddo test this

ahoskins:
Alfreddo test this

src/js/components/HashMapDisplay.js
```diff
@@ -114,7 +140,11 @@ HashMapDisplay.propTypes = {
...
   // replaced, and with corresponding values of the replacement to be rendered.
-  renderKeys: React.PropTypes.object
+  renderKeys: React.PropTypes.object,
+  showActions: React.PropTypes.bool,
+  onEditClick: React.PropTypes.func,
+  keyPath: React.PropTypes.array,
+  emptyValue: React.PropTypes.string
 };
 
 module.exports = HashMapDisplay;
```
MatApple: I think we should remove all the edits to this file and remove the edit button on each row. I thought we weren't going to add this till we could add it in the marathon config too? This `ConfigurationHashMap` component is problematic and we should (not in this PR) find a good way to get rid of it and instead just create the list of config key/value pairs like this:
```
item.map((index, item) => {   
   const [key, value] = item;
  
    return (
        <ConfigurationMapRow key={index}>
          <ConfigurationMapLabel keepTextCase={isAttribute}>
            {key}
          </ConfigurationMapLabel>
          <ConfigurationMapValue>{value}</ConfigurationMapValue>
        </ConfigurationMapRow>
      );
});
```
Until we do this, we have no flexibility for adding things like edit buttons without crazy props and key-path joining/splitting etc. cc/ @dcos/frontend-foundation

src/js/components/FrameworkConfiguration.js
```diff
@@ -0,0 +1,423 @@
...
+      deployErrors,
+      isInitialDeploy,
+      formData
+    } = this.props;
+
+    const fileName = "config.json";
+    const configString = JSON.stringify(formData, null, 2);
+
+    const renderKeys = {};
+    this.getHashMapRenderKeys(formData, renderKeys);
```
MatApple: Please update this method so it returns the `renderKeys` and doesn't mutate a previously set const.

src/js/components/FrameworkConfiguration.js
```diff
@@ -0,0 +1,423 @@
...
+        />
+      );
+    }
+
+    let errorsAlert = null;
+    if (deployErrors) {
+      errorsAlert = <CosmosErrorMessage error={deployErrors} />;
+    }
+
+    // todo use Michaels empty state component instead of hardcoded u2014
```
MatApple: Please remove all `todo`s :)

src/js/components/CosmosErrorMessage.js
```diff
@@ -29,6 +29,11 @@ class CosmosErrorMessage extends React.Component {
       return this.appendRepositoryLink(message);
     }
 
+    // make "Package is already installed error" better
+    if (error.type === "PackageAlreadyInstalled") {
+      return "A service with the same name already exists. Try a different name.";
```
bstavroulakis: I understand why this is necessary, we probably may want to open a ticket for the Cosmos team to brush some error messages.

ahoskins: Talked with @MatApple offline, and since this is overriding the message of a specific `error.type`, we are fine doing this manual translation in this component.

src/js/components/FrameworkConfigurationForm.js
bstavroulakis: This is fixed now, with the cross-browser updates!

src/js/components/SchemaField.js
```diff
@@ -0,0 +1,307 @@
...
+        showError={errorMessage.length > 0}
+        errorClassName="form-group-danger"
+      >
+        {this.getFieldContent(errorMessage, autofocus)}
+      </FormGroup>
+    );
+  }
+}
+
+export default SchemaField;
```
bstavroulakis: Let us change it to module.exports for consistency.

src/js/components/FrameworkConfiguration.js
```diff
@@ -0,0 +1,408 @@
...
+
+    return renderKeys;
+  }
+
+  getFirstTabAndField() {
+    const { packageDetails } = this.props;
+    const schema = packageDetails.getConfig();
+
+    const [activeTab] = Object.keys(schema.properties);
+    const [focusField] = Object.keys(schema.properties[activeTab].properties);
```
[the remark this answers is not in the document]
ahoskins: Added checking for this case when an empty second layer of configuration. Talked offline with @MatApple on it.

src/styles/components/modal/full-screen/styles.less
```diff
@@ -118,10 +118,17 @@
     text-align: center;
   }
 
+  .modal-full-screen-header-with-sub-title {
```
MatApple: I can't find where this class is used?!

src/js/components/FrameworkConfiguration.js
```diff
@@ -0,0 +1,417 @@
...
+      schema.properties[tab].properties &&
+      Object.keys(schema.properties[tab].properties).length >= 1
+    ) {
+      return Object.keys(schema.properties[tab].properties)[0];
+    }
+
+    return "NO-OP";
+  }
+
+  getFirstTabAndField() {
```
orlandohohmeier: 🎨  it would be nice to move _functions_ like this one into a util and adjust the implementation to not rely on local state...

src/styles/components/modal/full-screen/styles.less
ahoskins: It's used in `FrameworkConfiguration`...github doesn't open that file because too big, so it probably missed your cmd+f 😏

src/js/components/FrameworkConfiguration.js
```diff
@@ -0,0 +1,417 @@
...
+const METHODS_TO_BIND = [
+  "handleJSONToggle",
+  "handleGoBack",
+  "handleServiceReview",
+  "handleEditConfigurationButtonClick",
+  "handleActiveTabChange",
+  "handleFocusFieldChange",
+  "handleCloseConfirmModal",
+  "handleConfirmGoBack",
+  "onFormDataChange",
```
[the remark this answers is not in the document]
ahoskins: Fixed, passing now.

src/js/components/FrameworkConfiguration.js
```diff
@@ -0,0 +1,417 @@
...
+    });
+  }
+
+  onFormErrorChange(formErrors) {
+    this.props.onFormErrorChange(formErrors);
+  }
+
+  handleFocusFieldChange(activeTab, focusField) {
+    if (focusField === this.state.focusField) {
+      return false;
```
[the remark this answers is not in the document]
ahoskins: I will actually just remove this. Overly defensive, and not actually doing anything.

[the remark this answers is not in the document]
orlandohohmeier: I wouldn’t say that the early return is overly defensive... I just don’t understand why you return ‘false’.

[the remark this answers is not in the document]
ahoskins: Previously I was just doing `return;`. @weblancaster suggested to be explicit and actually return a value. I don't have any strong opinions on this, I think either way it doesn't matter for these kind of methods that are event handlers.

src/js/components/FrameworkConfiguration.js
ahoskins: I'm going to hold off doing this for now, but I understand potential benefit of moving these kind of things out of components into pure helpers. 

For this PR, I don't want to do any other major refactors that will require additional testing for regressions.

src/js/components/FrameworkConfigurationForm.js
```diff
@@ -0,0 +1,337 @@
...
+    }
+
+    return this.formatErrorForJSON(errorSchema, []);
+  }
+
+  // transforms the errors from validating against the JSON schema
+  transformErrors(errorSchema) {
+    const errorSchemaCopy = [];
+
+    errorSchema.map(error => {
```
orlandohohmeier: ☝️ Please adjust the transformation to either use `forEach` or even better to _properly_ use `map`. You're mutating `errorSchemaCopy` instead of mapping the values – I assume you did so to remove unwanted errors. You can either `filter` the list of errors before/after mapping or use reduce to do sone in one go. 

```js
  transformErrors(errorSchema) {
    // Filter "type" errors as they're caught by the validate() function and
    // capitalize error messages
    return errorSchema
      .filter(({ name }) => name !== "type")
      .map(({ message, ...rest }) => {
        return { message: StringUtil.capitalize(message), ...rest };
      });
  }
```

src/js/components/FrameworkConfigurationForm.js
```diff
@@ -0,0 +1,337 @@
...
+      // remove name: "type" errors because caught by the validate() function
+      if (error.name === "type") {
+        return;
+      }
+
+      // capitalize other errors (ex: regex errors
+      error.message = StringUtil.capitalize(error.message);
```
orlandohohmeier: ☝️ You're mutating the error object, which might result in unwanted behavior. Please use either `Object.assign` or the property spread operator to circumvent unwanted side-effects.

src/js/components/SchemaField.js
```diff
@@ -0,0 +1,315 @@
...
+        </FormGroupHeadingContent>
+      </FormGroupHeading>
+    );
+  }
+
+  getFieldContent(errorMessage, autofocus) {
+    const { schema } = this.props;
+    const RADIO_SELECT_THRESHOLD = 4;
+
+    switch (schema.type) {
```
orlandohohmeier: 🤔 wonder if we should add a `default` case for unknown fields...

src/js/components/FrameworkConfiguration.js
```diff
@@ -0,0 +1,403 @@
...
+      renderKeys = Object.assign(
+        renderKeys,
+        this.getHashMapRenderKeys(formData[key])
+      );
+    });
+
+    return renderKeys;
+  }
+
+  getFirstFieldInTab(schema, tab) {
```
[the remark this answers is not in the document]
ahoskins: Nice suggestion, I switched this to use `findNestedPropertyInObject`, very clean, forgot that util existed

src/js/components/FrameworkConfiguration.js
```diff
@@ -0,0 +1,403 @@
...
+    // for autofocus we return a field id name, if no fields return string "no-op"
+    if (
+      schema.properties[tab] &&
+      schema.properties[tab].properties &&
+      Object.keys(schema.properties[tab].properties).length >= 1
+    ) {
+      return Object.keys(schema.properties[tab].properties)[0];
+    }
+
+    return "NO-OP";
```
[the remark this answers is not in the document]
ahoskins: I switch to use findNestedPropertyInObject, and now it will just return `null` which is fine. When it's null, it will simply not focus any tab, which is not an issue.

src/js/components/FrameworkConfiguration.js
```diff
@@ -0,0 +1,403 @@
...
+        label: reviewActive ? " Back" : "Cancel"
+      }
+    ];
+  }
+
+  getPrimaryActions() {
+    const { jsonEditorActive, reviewActive } = this.state;
+    const { formErrors } = this.props;
+
+    const hasFormErrors = Object.keys(formErrors).reduce(
```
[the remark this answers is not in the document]
ahoskins: Will use `some` op 👍

src/js/components/FrameworkConfigurationForm.js
ahoskins: Yes, much nicer. It was funny for me to read my very badly implemented version using `map`... 👍

src/js/components/FrameworkConfigurationForm.js
ahoskins: Fixed this by using map properly 😉

src/js/components/SchemaField.js
```diff
@@ -0,0 +1,315 @@
...
+        <FormGroupHeadingContent className="text-danger" primary={false}>
+          *
+        </FormGroupHeadingContent>
+      );
+    }
+
+    return (
+      <FormGroupHeading>
+        <FormGroupHeadingContent primary={true}>
+          {name.split(/_|-/).join(" ")}
```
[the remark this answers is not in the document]
ahoskins: This name is directly from the json schema for the framework. It's either dash or underscore separated!

src/js/components/SchemaField.js
ahoskins: I think this a good idea. I've added a default that renders text input. Text input seems least likely to explode with a strange schema type:

```
default:
    return this.renderTextInput(errorMessage, autofocus, this.props);
```

src/js/components/SchemaField.js
```diff
@@ -0,0 +1,315 @@
...
+
+    if (schema.type === "object") {
+      return <DefaultSchemaField {...this.props} />;
+    }
+
+    const autofocus = uiSchema && uiSchema["ui:autofocus"];
+
+    let errorMessage;
+    if (errorSchema) {
+      errorMessage = errorSchema.__errors.map(error => {
```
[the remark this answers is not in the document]
ahoskins: I don't think there's a better way. The `errorSchema` object is delivered to the schema form onChange handler, and I also parse the `__errors` property in `FrameworkConfigurationForm`.
screen shot 2017-10-26 at 2 03 21 pm

src/styles/components/form-elements/form-components/field-select/styles.less
```diff
@@ -0,0 +1,17 @@
+& when (@field-select-enabled) {
+
+  // be more specific than the cnvs selector which speciies 100% width
```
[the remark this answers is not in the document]
ahoskins: I'm not sure what you mean by this comment.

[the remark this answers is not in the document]
MatApple: @ahoskins misspelled `speciies`

src/js/components/FrameworkConfigurationForm.js
```diff
@@ -0,0 +1,337 @@
...
+    this.state = {
+      errorSchema: null
+    };
+
+    METHODS_TO_BIND.forEach(method => {
+      this[method] = this[method].bind(this);
+    });
+  }
+
+  getFormattedSectionLabel(label) {
```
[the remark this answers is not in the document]
ahoskins: I moved this into StringUtil in a method called `capitalizeEveryWord(string)`.

src/js/components/SchemaField.js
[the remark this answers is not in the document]
orlandohohmeier: I've looked into the `react-jsonschema-form` implementation to understand why they're using a property named `__errors` as it looks a lot like a property meant for _internal_ use only. They're using `__errors` to avoid name collision with a possible sub field named "errors" (for details see: [mozilla-services/react-jsonschema-form/src/validate.js#L56](https://github.com/mozilla-services/react-jsonschema-form/blob/6322579b570336a398dc9faf0870875480455d33/src/validate.js#L56)), but they also provide util to transform `errorSchema` into an error list, so that we don't need to access  `__errors`  directly. Please use [`toErrorList`](https://github.com/mozilla-services/react-jsonschema-form/blob/6322579b570336a398dc9faf0870875480455d33/src/validate.js#L67) instead of directly accessing the property or create a custom util to at least contain the usage as there is a risk that they will change the variable name. I'm ok if you do so in a subsequent PR to no further block this PR.

/cc @dcos/frontend-foundation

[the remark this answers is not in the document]
ahoskins: Interesting. I looked into this a bit, and have created a Jira to prevent blocking this PR any further: https://jira.mesosphere.com/browse/DCOS-19332

src/styles/components/form-elements/form-components/field-select/styles.less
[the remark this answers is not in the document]
orlandohohmeier: ☑️  Fixed

src/js/components/SchemaField.js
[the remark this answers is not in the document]
orlandohohmeier: Alright, let's keep it as is for now. I'd vote to adjust the `FrameworkConfigurationForm ` in future iteration to transform schema headings before passing the data along and then remove the transformation logic from the `SchemaField` component. 

https://jira.mesosphere.com/browse/DCOS-19367

orlandohohmeier:
@ahoskins This is a splendid feature. Great Job! 💯  👏   Please squash your commits before merging so that they're in line with our guidelines.

PR #2443, opened 2017-10-05
feat(FrameworkConfiguration): framework configuration edit

ahoskins:
*Depends on #2442* ⚠️ 

This PR uses the framework configuration components to implement the framework edit flow.

**Testing:**
1. Run a service like cassandra
2. Click edit on this running service.

**Checklist**


plugins/services/src/js/components/modals/EditServiceModal.js
```diff
@@ -0,0 +1,28 @@
...
+import EditFrameworkConfiguration
+  from "#PLUGINS/services/src/js/pages/EditFrameworkConfiguration";
+import CreateServiceModal
+  from "#PLUGINS/services/src/js/components/modals/CreateServiceModal";
+import Framework from "#PLUGINS/services/src/js/structs/Framework";
+import DCOSStore from "#SRC/js/stores/DCOSStore";
+
+class EditServiceModal extends Component {
+  render() {
+    const { params } = this.props;
```
orlandohohmeier: Why don't you get the `id` directly? 

```js
 const { id  = "/" } = this.props.params;
```

plugins/services/src/js/components/modals/EditServiceModal.js
```diff
@@ -0,0 +1,28 @@
...
+
+    const serviceID = decodeURIComponent(params.id || "/");
+    const service = DCOSStore.serviceTree.findItemById(serviceID);
+
+    let child = null;
+    if (service instanceof Framework) {
+      child = <EditFrameworkConfiguration {...this.props} />;
```
orlandohohmeier: I'd simply return Instead of assigning the `child` to improve the readability.

```js
class EditServiceModal extends Component {
  render() {
    const { id  = "/" } = this.props.params;
    const service = DCOSStore.serviceTree.findItemById(decodeURIComponent(id));

    if (service instanceof Framework) {
      return (<FrameworkConfiguration {...this.props} />);
    } 
    
   return (<CreateServiceModal {...this.props} />);
  }
}
```

ahoskins:
Alfreddo test this

plugins/services/src/js/pages/EditFrameworkConfiguration.js
```diff
@@ -0,0 +1,131 @@
...
+import mixin from "reactjs-mixin";
+import { StoreMixin } from "mesosphere-shared-reactjs";
+import { routerShape } from "react-router";
+import deepEqual from "deep-equal";
+import CosmosPackagesStore from "#SRC/js/stores/CosmosPackagesStore";
+import UniversePackage from "#SRC/js/structs/UniversePackage";
+import FrameworkConfiguration from "#SRC/js/components/FrameworkConfiguration";
+import Loader from "#SRC/js/components/Loader";
+
+export default class EditFrameworkConfiguration extends mixin(StoreMixin) {
```
weblancaster: shouldn't we supposed to be using `module.exports` instead `export default` if I'm not wrong we might have issues related to jest and mocking.

ahoskins:
Alfreddo test this

ahoskins:
Alfreddo test this

ahoskins:
Alfreddo test this

ahoskins:
Alfred test this

plugins/services/src/js/pages/EditFrameworkConfiguration.js
```diff
@@ -0,0 +1,141 @@
...
+import { StoreMixin } from "mesosphere-shared-reactjs";
+import { routerShape } from "react-router";
+import deepEqual from "deep-equal";
+import CosmosPackagesStore from "#SRC/js/stores/CosmosPackagesStore";
+import UniversePackage from "#SRC/js/structs/UniversePackage";
+import FrameworkConfiguration from "#SRC/js/components/FrameworkConfiguration";
+import Loader from "#SRC/js/components/Loader";
+import RequestErrorMsg from "#SRC/js/components/RequestErrorMsg";
+
+export default class EditFrameworkConfiguration extends mixin(StoreMixin) {
```
bstavroulakis: We can add this export at the end of the file with module.exports

bstavroulakis:
Test failing looks legit. We'll probably have to update the tests.

plugins/services/src/js/pages/EditFrameworkConfiguration.js
```diff
@@ -0,0 +1,143 @@
...
+import { StoreMixin } from "mesosphere-shared-reactjs";
+import { routerShape } from "react-router";
+import deepEqual from "deep-equal";
+import CosmosPackagesStore from "#SRC/js/stores/CosmosPackagesStore";
+import UniversePackage from "#SRC/js/structs/UniversePackage";
+import FrameworkConfiguration from "#SRC/js/components/FrameworkConfiguration";
+import Loader from "#SRC/js/components/Loader";
+import RequestErrorMsg from "#SRC/js/components/RequestErrorMsg";
+
+class EditFrameworkConfiguration extends mixin(StoreMixin) {
```
orlandohohmeier: 🎨 I'm ok with the current implementation as it's inline with many of our components but moving forward please try to not use the `StoreMixin` or mixins in gernal. Mixins are super hard reason about and debug — using them is an anti-pattern and we should always use composition instead. /cc @dcos/frontend-foundation

plugins/services/src/js/pages/EditFrameworkConfiguration.js
```diff
@@ -0,0 +1,143 @@
...
+
+  onCosmosPackagesStoreServiceUpdateSuccess() {
+    this.handleGoBack();
+  }
+
+  onCosmosPackagesStoreServiceUpdateError(deployErrors) {
+    this.setState({ deployErrors });
+  }
+
+  // necessary to have review screen same order of tabs as the form
```
orlandohohmeier: 🎨 I'd move the comment to https://github.com/dcos/dcos-ui/pull/2443/files#diff-efd3f92415ec5298ecb20a1ff4642c0eR42 as it's not document the functionality but the why it needs to be applied.

plugins/services/src/js/pages/EditFrameworkConfiguration.js
```diff
@@ -0,0 +1,143 @@
...
+  reorderResolvedOptions(resolvedOptions, config) {
```
orlandohohmeier: 🤔 Out of curiosity, why are these `resolvedOptions` in a different order?Would it help to use a different data structure?

plugins/services/src/js/pages/EditFrameworkConfiguration.js
ahoskins: Sounds good thanks for bring to attention 👍

plugins/services/src/js/pages/EditFrameworkConfiguration.js
ahoskins: I agree and wish this wasn't necessary because it's quite weird to see in code. But when iterating over the keys in the parsed JSON response from Cosmos, the order is different.

We use the `.package` key (the schema) to render the form and tabs. And the `.resolvedOptions` key gives the current running framework configuration. So this method is resetting the resolvedOptions keys to match the order of the schema (the resolvedOptions is used to render the review screen because it's the source of truth of data values).

plugins/services/src/js/pages/EditFrameworkConfiguration.js
orlandohohmeier: Using `export default` is fine here, but there are a few cases in which we have to use `module.exports` due to issue with the current plugin implementation.

orlandohohmeier:
@weblancaster could you please approve this one?

orlandohohmeier:
@ahoskins please squash your commits before merging to align them with out guidelines.

PR #2445, opened 2017-10-05
feat(babel): allow stage-3 syntax

nLight:
Stage 3 brings a lot of nice things and means those proposals at the candidate level already which means little to no changes. Full list of stage 3 proposals can be found here https://github.com/tc39/proposals I'm especially looking forward to [rest and spread properties](https://github.com/tc39/proposal-object-rest-spread)

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,5 +1,5 @@
 {
   plugins: ['transform-runtime'],
-  presets: ['es2015', 'react']
+  presets: ['es2015', 'stage-3', 'react']
 }
 
--- a/jest/preprocessor.js
+++ b/jest/preprocessor.js
@@ -1,33 +1,39 @@
-var babel = require('babel-core');
-var jestPreset = require('babel-preset-jest');
-var jison = require('jison');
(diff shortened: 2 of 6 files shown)


jest/preprocessor.js
```diff
@@ -1,33 +1,39 @@
-var babel = require('babel-core');
-var jestPreset = require('babel-preset-jest');
-var jison = require('jison');
-var webpackAlias = require('jest-webpack-alias');
+var babel = require("babel-core");
```
orlandohohmeier: 😸  prettier is prettier

.babelrc
```diff
@@ -1,5 +1,5 @@
 {
   plugins: ['transform-runtime'],
-  presets: ['es2015', 'react']
+  presets: ['es2015', 'stage-3', 'react']
```
weblancaster: just slide `stage-0` silently here. 😉

ahoskins:
Nice, ran into a case of not being allowed to use object rest spread a few days ago, and this will include that capability 👍

PR #2447, opened 2017-10-07
feat(ServicesTable): Instances column

weblancaster:
Implement instance column in ServicesTable

Closes DCOS-17966

**before**
screen shot 2017-10-09 at 2 44 09 pm

**after**
screen shot 2017-10-09 at 2 05 04 pm

**Checklist**


plugins/services/src/js/containers/services/ServicesTable.js
```diff
@@ -391,6 +392,24 @@ class ServicesTable extends React.Component {
...
   }
 
+  renderInstances(prop, service) {
+    const instancesCount = service.getInstancesCount();
+    const tasksRunning = service.getTaskCount();
+    const overview = tasksRunning === instancesCount
+      ? ` ${tasksRunning}`
+      : ` ${tasksRunning}/${instancesCount}`;
+
+    const content = instancesCount == null || service.list
```
ahoskins: What is the meaning of `service.list`? It's not part of the struct.

weblancaster: whenever there is a path/folder `list` will exist containing the child paths.

I guess I can add to the struct since doesn't exist. Thanks.

weblancaster:
@ahoskins to answer your questions, 

1. that work is part of another Jira ticket.
2. Good catch, I forgot we sort of support responsive.

thanks.

plugins/services/src/js/structs/ServiceTree.js
```diff
@@ -429,4 +429,8 @@ module.exports = class ServiceTree extends Tree {
       return serviceTreeLabels;
     }, []);
   }
+
+  isFolder() {
```
orlandohohmeier: One can use the service instance type to differentiate between a service tree (folder) or an app, pod and framework. Let's please not add a method like this one.

plugins/services/src/js/structs/Application.js
```diff
@@ -230,4 +230,8 @@ module.exports = class Application extends Service {
 
     return `${Config.rootUrl}/service/${serviceName}/`;
   }
+
+  isFolder() {
+    return Array.isArray(this.list);
```
orlandohohmeier: `Application` doesn't feature a `list` property only a `ServiceTree` is a folder and features a _list_ of services and service trees – in other words, this method will always return false.

plugins/services/src/js/structs/Application.js
```diff
@@ -230,4 +230,8 @@ module.exports = class Application extends Service {
 
     return `${Config.rootUrl}/service/${serviceName}/`;
   }
+
+  isFolder() {
```
orlandohohmeier: ℹ️  If you ever want to add method to the _service interface_ please make sure to add it to the base `Service` struct and `ServiceTree` as all "services" should share the same interface.

plugins/services/src/js/containers/services/ServicesTable.js
```diff
@@ -393,15 +393,14 @@ class ServicesTable extends React.Component {
   }
 
   renderInstances(prop, service) {
+    const isFolder = service.isFolder();
```
orlandohohmeier: Please use the service type to check wether a service a _folder_ / tree of services or a plain service.

plugins/services/src/js/containers/services/ServicesTable.js
```diff
@@ -393,15 +393,14 @@ class ServicesTable extends React.Component {
...
     const instancesCount = service.getInstancesCount();
     const tasksRunning = service.getTaskCount();
     const overview = tasksRunning === instancesCount
       ? ` ${tasksRunning}`
       : ` ${tasksRunning}/${instancesCount}`;
 
-    const content = instancesCount == null || service.list
-      ? "\u2014"
-      : overview;
+    const content = instancesCount == null || isFolder ? "\u2014" : overview;
```
orlandohohmeier: Wonder why you _cache_ the result of `isFolder` in line 396 instead of calling the getter right in this expression?

plugins/services/src/js/containers/services/ServicesTable.js
```diff
@@ -391,6 +392,23 @@ class ServicesTable extends React.Component {
     );
   }
 
+  renderInstances(prop, service) {
+    const isFolder = service.isFolder();
```
orlandohohmeier: Please use the instance type to check wether it's a folder / `ServiceTree` or a plain `Service`.

plugins/services/src/js/containers/services/ServicesTable.js
```diff
@@ -391,6 +392,23 @@ class ServicesTable extends React.Component {
...
+    const instancesCount = service.getInstancesCount();
+    const tasksRunning = service.getTaskCount();
+    const overview = tasksRunning === instancesCount
+      ? ` ${tasksRunning}`
+      : ` ${tasksRunning}/${instancesCount}`;
+
+    const content = instancesCount == null || isFolder ? "\u2014" : overview;
```
orlandohohmeier: Wonder why you assign the return  value of `isFolder` in line 396 instead of simply calling it in the expression?

plugins/services/src/js/structs/ServiceTree.js
weblancaster: Correct, now do we want as a method in structs or inside `ServicesTable.js`? I see being part of the struct since this can be re-used by other parts of the application that has service, also because to keep consistent with what we already have.

plugins/services/src/js/containers/services/ServicesTable.js
weblancaster: Can you take a look at this comment please https://github.com/dcos/dcos-ui/pull/2447#discussion_r143761703

plugins/services/src/js/containers/services/ServicesTable.js
weblancaster: I found 1% cleaner to store in a variable having at the top the method and then using the variable. Nothing special here.

plugins/services/src/js/structs/ServiceTree.js
orlandohohmeier: If you really need the method for `convenience` please consider adding a util that checks the instance type.

plugins/services/src/js/containers/services/ServicesTable.js
orlandohohmeier: :)

weblancaster:
The failing integration test is passing locally so I'm assuming it's something flaky going on.

alfred test this please

weblancaster:
@ahoskins your review about smaller screens is addressed.
I sat with @leemunroe to come up the ideal display.

screen shot 2017-10-10 at 3 25 34 pm

weblancaster:
Fixing the failing integration tests

plugins/services/src/js/containers/services/ServicesTable.js
```diff
@@ -391,6 +392,22 @@ class ServicesTable extends React.Component {
...
   }
 
+  renderInstances(prop, service) {
+    const instancesCount = service.getInstancesCount();
+    const tasksRunning = service.getTaskCount();
+    const overview = tasksRunning === instancesCount
+      ? ` ${tasksRunning}`
+      : ` ${tasksRunning}/${instancesCount}`;
+
+    const content = instancesCount == null ? "\u2014" : overview;
```
weblancaster: ideally `instancesCount` will always have a number but to be safe and follow the design requirements we display `-`

weblancaster:
retest this please

src/js/components/LongDashPlaceholder.js
```diff
@@ -0,0 +1,16 @@
+import React from "react";
```
weblancaster: Component introduced so we don't need to hardcode Unicode all over the place.
This is also an approach towards component based UI.

Why not a more general naming? because as per component based UIs the components are disposable and whenever something new is needed we create a new component instead of extending or change. Until no more is in use that way we can just delete.

src/js/components/LongDashPlaceholder.js
```diff
@@ -0,0 +1,16 @@
...
+
+/**
+ * Component to display long dash placeholder
+ *
+ * @class LongDashPlaceholder
+ * @extends {React.Component}
+ * @returns {String} unicode dash
+ */
+class LongDashPlaceholder extends React.Component {
```
ahoskins: `export default LongDashPlaceHolder = props => (<span>{"\u2014"}</span>);`

weblancaster: I wrote like that first but then decided to follow the rest of our components, also is there any gain on this?

ahoskins: No gain as far as I know. We have have some though - like `ConfigurationMapLabel`.

nLight:
Is there a duplication going on? We have instance count in the Status col, don't we?

weblancaster:
@nlight that work will com in a separated PR.

plugins/services/src/js/structs/ServiceTree.js
nLight: I suggested the opposite in #2459 💃

orlandohohmeier: @nLight different case as this one refers to the service type and your comment was related to service state.

src/js/components/LongDashPlaceholder.js
orlandohohmeier: I understand the need for a means to not hardcode the same unicode everywhere but it feels strange to create component to only render one character. Please note that every component comes with an overhead.

weblancaster: Understood, those are good options too although I rather have a composable UI.

I'm changing to one of your suggestions though so we can merge the PR.

src/js/components/LongDashPlaceholder.js
orlandohohmeier: Functional components will eventually be more performant (there are plans to improve the rendering of stateless components) but at the moment they're internally converted to _class_ based components which means there is little to no gain. Though one can use functional stateless components like any other function which removes the entire react create element step and improves performance a lot. For details please see: https://medium.com/missive-app/45-faster-react-functional-components-now-3509a668e69f

src/js/components/LongDashPlaceholder.js
orlandohohmeier: I'm aspiring a declarative and compose-able UI as well, but honestly don't think that `<LongDashPlaceholder />` is any better than simply returning a long dash, which is why I'd go for a constant or a general purpose placeholder util or component.

You are right that some of our the components should, based on the presented logic, indeed be constants and helpers as well. I actually think that we should refactor those components as wouldn't say that writing these as react components is in any way more declarative or compose-able. There is a reason why the community is pushing functional components which are basically nothing else then simple helper functions.

src/js/components/LongDashPlaceholder.js
weblancaster: In the future will make sense to use them whenever needed but I'm talking about our case today and like I said in my first comment above functional component was my first option but I assumed someone was going to complain of not following the patterns we have in the code base for the other components since the majority is written using `class` so I went with `class`.

src/js/components/LongDashPlaceholder.js
weblancaster: addressed using util. thanks.

plugins/services/src/js/containers/services/ServicesTable.js
```diff
@@ -391,6 +392,24 @@ class ServicesTable extends React.Component {
...
   }
 
+  renderInstances(prop, service) {
+    const instancesCount = service.getInstancesCount();
+    const tasksRunning = service.getTaskCount();
+    const overview = tasksRunning === instancesCount
+      ? ` ${tasksRunning}`
+      : ` ${tasksRunning}/${instancesCount}`;
+
+    const content = typeof instancesCount !== "number"
```
weblancaster: In theory `instancesCount` should always contain a number, so in case isn't we display a long dash.

orlandohohmeier: As `NaN` is a "number" you either add `isNaN` or use `isFinite` or `isInteger` to ensure a it's actually a valid instance count.

src/js/utils/StringUtil.js
```diff
@@ -215,6 +215,10 @@ const StringUtil = {
     }, []);
 
     return this.idToTitle(splitID, splitBy, replace, removeConsecutive);
+  },
+
+  getLongDashPlaceholder() {
```
orlandohohmeier: This isn’t really what I had in mind and I’m sorry that I wasn’t precise. Having a util like this one is still not better then a constants as it comes with an overhead ans is be no means more declarative than using a constants. If you want to go with an util I’d opt for a more general util similar to an either functor that either returns the _value_ or the placeholder.

weblancaster: addressed, I went back with hard coded as the other PRs have.

plugins/services/src/js/containers/services/ServicesTable.js
```diff
@@ -391,6 +392,22 @@ class ServicesTable extends React.Component {
...
   }
 
+  renderInstances(prop, service) {
+    const instancesCount = service.getInstancesCount();
+    const tasksRunning = service.getTaskCount();
+    const overview = tasksRunning === instancesCount
+      ? ` ${tasksRunning}`
+      : ` ${tasksRunning}/${instancesCount}`;
+
+    const content = isNaN(instancesCount) ? "\u2014" : overview;
```
orlandohohmeier: If you want to be sure `instancesCount` is really a number you should also check `typeof` as you otherwise might get a false positive (e.g., `isNaN("")`).

plugins/services/src/js/containers/services/ServicesTable.js
```diff
@@ -391,6 +392,22 @@ class ServicesTable extends React.Component {
     );
   }
 
+  renderInstances(prop, service) {
+    const instancesCount = service.getInstancesCount();
```
orlandohohmeier: You could return early if instancesCount is not a number as there's no need to perform any of the other actions in that case. Similar to the following sketch: 

```js
renderInstances(prop, service) {
    
    if(!Number.isInteger(instancesCount)){
       return <span>\u2014</span>;
    } 
​
    const tasksRunning = service.getTaskCount();
​
    return (
      <span>
        { 
           tasksRunning === instancesCount 
              ? ` ${tasksRunning}` 
               : ` ${tasksRunning}/${instancesCount}`
         }
      </span>
    );
  }
```

orlandohohmeier:
@weblancaster there are to minor things you could improve but I'm ok to move forward as is. Thanks for all your hard work on this one!

plugins/services/src/js/containers/services/ServicesTable.js
weblancaster: I appreciate the feedback, I do rather see one return per method/functions because give us the impression of being cleaner with one flow because there is no use if condition (`if`)

plugins/services/src/js/containers/services/ServicesTable.js
weblancaster: Agree, I have changed to make sure is only `integer`. Thanks

plugins/services/src/js/containers/services/ServicesTable.js
orlandohohmeier: Why do you prefer to see only one return per method/function? Early returns help maintain readability esp. if you have a large number _preconditions_ as you don't need to nest conditions. Compare the following to examples:

```
if A is defined
  if B  is defined
    if  C  is defined
        // do something
    else 
      handle undefined C
  else 
     handle undefined B
else 
  handle undefined A
```

```
if A is undefined
  handle undefined A; 
  return;
if B is undefined
  handle undefined B; 
  return;
if C is  undefined
  handle undefined C; 
  return;

// do something
```

Please note that using early returns instead of nested conditions does't effect the cyclomatic complexity but improves readability. BTW: A ternary also affects the complexity and based on the metric (e.g., McGabe or NPATH) you're using it's even worse than a condition. 

Anyway, I'm ok with the current implementation and only wanted to provide an example of a different solution for educational purposes.

PR #2448, opened 2017-10-09
feat(DSLUtil): Add multi select capability

Poltergeist:
Add multi select capability to the DSLUtil this way a `<key>:a,b,c` is possible. This will be needed
for the nodes filtering since most of the filter values are dynamic.

### DIFF
--- a/src/js/utils/DSLUtil.js
+++ b/src/js/utils/DSLUtil.js
@@ -110,7 +110,11 @@ const DSLUtil = {
         // Require only testing properties to match
         const compareParamNames = Object.keys(compareFilterParams);
         const comparePropMatches = compareParamNames.every(prop => {
-          return filterParams[prop] === compareFilterParams[prop];
+          return (
+            filterParams[prop] === compareFilterParams[prop] ||
+            (compareFilterParams[prop] === undefined &&
+              compareFilterParams[prop] !== filterParams[prop])
+          );
         });
 
         if (compareParamNames.length === 0 || comparePropMatches) {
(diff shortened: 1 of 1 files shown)


src/js/utils/DSLUtil.js
```diff
@@ -110,7 +110,11 @@ const DSLUtil = {
         // Require only testing properties to match
         const compareParamNames = Object.keys(compareFilterParams);
         const comparePropMatches = compareParamNames.every(prop => {
-          return filterParams[prop] === compareFilterParams[prop];
+          return (
+            filterParams[prop] === compareFilterParams[prop] ||
+            (compareFilterParams[prop] === undefined &&
+              compareFilterParams[prop] !== filterParams[prop])
```
orlandohohmeier: This check is redundant... `compareFilterParams[prop] !== filterParams[prop]`

src/js/utils/DSLUtil.js
```diff
@@ -170,7 +174,13 @@ const DSLUtil = {
         // Properties created through attribute filter will get a boolean value
         //
         case DSLFilterTypes.ATTRIB:
-          memo[prop] = matchingNodes.length === 1;
+          if (matchAgainst.filterParams.text === undefined) {
```
orlandohohmeier: Wonder if we should introduce a dedicated symbol e.g., `Symbol("DYNAMIC")` instead of using `undefined` to indicate that we wan't to process _dynamic_ input / filter text.

nLight:
Hey @Poltergeist any update on this one?

orlandohohmeier:
@Poltergeist I've put this one on hold as it's still work in progress. Please remove once it's ready.

PR #2457, opened 2017-10-12
Uninstall 03 | Bump reactjs-components to beta.27

bstavroulakis:
Bump reactjs-components package version to beta.27.
Remove react-addons-css-transition-group as we don't use it anymore in reactjs-components.
Add react-transition-group as this is used now in reactjs-components.
Make adjustments to the code necessary for the react-transition-group library.

**Checklist**

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -1755,6 +1755,11 @@
       "from": "center-align@>=0.1.1 <0.2.0",
       "resolved": "https://registry.npmjs.org/center-align/-/center-align-0.1.3.tgz"
     },
+    "chain-function": {
+      "version": "1.0.0",
+      "from": "chain-function@>=1.0.0 <2.0.0",
+      "resolved": "https://registry.npmjs.org/chain-function/-/chain-function-1.0.0.tgz"
+    },
     "chalk": {
       "version": "1.1.1",
       "from": "chalk@1.1.1",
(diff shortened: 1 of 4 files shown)


tests/pages/jobs/JobActions-cy.js
```diff
@@ -13,9 +13,7 @@ describe("Job Actions", function() {
     });
 
     it("opens the correct jobs edit modal", function() {
-      cy
-        .get('.modal .form-panel input[name="id"]')
-        .should("to.have.value", "foo");
+      cy.get(".modal .form-panel form input").should("to.have.value", "foo");
```
bstavroulakis: @nLight @orlandohohmeier This is needed because after the bump I notice this behavior with the input fields. On the left is the before on the right is the after. It looks like it sanitizes the properties which is good in a way, is this expected behavior?

image

nLight: Hey @bstavroulakis! It is an artifact from an another update. So the filtering is indeed intentional, though we still want to keep the name attribute. I'm going to create a fix PR for the components. Sorry for any trouble.

nLight: https://github.com/mesosphere/reactjs-components/pull/416

nLight: We gonna have another bump

nLight:
Hope you're fine with me jumping in and fixing bugs caused by my updates 🙏

bstavroulakis:
@nLight Thank you looks great now!

PR #2440, opened 2017-10-04
[WIP] DCOS-18197: Delete status while framework is being deleted

bstavroulakis:
When you trigger to delete a framework we'll have to show the "deleting" status and disable the dropdowns.

## So If you delete elastic:
image

## Deleting Status and disabled dropdown 1:
image

## Deleting Status and disabled dropdown 2:
image

**Checklist**
- [ ] Did you add a JIRA issue in a commit message or as part of the branch name?
- [ ] Did you add new unit tests?
- [ ] Did you add new integration tests?
- [ ] If this is a regression, did you write a test to catch this in the future?

PR #2458, opened 2017-10-12
Uninstall 04 | Remove unnecessary inactive delete modal areas

bstavroulakis:
Remove unnecessary inactive delete modal areas since all of these are handled now in the Uninstall modal.
Related to #2402 

**Checklist**
- [ ] Did you add a JIRA issue in a commit message or as part of the branch name?
- [ ] Did you add new unit tests?
- [ ] Did you add new integration tests?
- [ ] If this is a regression, did you write a test to catch this in the future?

PR #2459, opened 2017-10-12
Uninstall 05 | DCOS-18197: add delete status while service is deleting and disable actions

bstavroulakis:
Add delete status while service is deleting and disable actions.

Closes DCOS-18197

**Checklist**


plugins/services/src/js/components/ServiceBreadcrumbs.js
```diff
@@ -119,7 +120,9 @@ class ServiceBreadcrumbs extends React.Component {
       return null;
     }
 
-    const serviceStatus = service.getStatus();
+    const serviceStatus = isFrameworkDeleting(service)
```
nLight: How do you think if we add `isDeleeting()` method to the `Service.js` struct so we never need to check if it's a framework or not?

Then we could just say `service.isDeleting()` and having everything else that is not a `Framework` returns `false`;

orlandohohmeier:
@bstavroulakis how does the PR relate to https://github.com/dcos/dcos-ui/pull/2440?

plugins/services/src/js/components/ServiceBreadcrumbs.js
nLight: @orlandohohmeier ^^

bstavroulakis:
@orlandohohmeier I closed #2440, since it is broken down into 3 now #2457, #2458, #2459

plugins/services/src/js/components/ServiceBreadcrumbs.js
bstavroulakis: @nLight @orlandohohmeier I thought about that, but I then thought we would be returning information that is not true. If someone called service.isDeleting() on a non-framework service it could return wrong information because it could be actually deleting but we're returning false.
So instead of creating an unreliable method and since this task is for frameworks only I chose to make it framework specific.

orlandohohmeier: I'm wondering why we're not adding a new ` isDeleting` getter instead of adding a new service statuts as mentioned in https://github.com/dcos/dcos-ui/pull/2440#discussion_r143800283

We could overwrite `getServiceStatus` to return _deleting_ in case a framework is flagged with the respective env. Using the approach would allow us to eventually add the same feature for all services as we don't add a special framework case. It's also not presenting wrong data as we're only returning _deleting_ for frameworks. Furthermore I'd say it's a much cleaner API as we present only one method to get the service status.

bstavroulakis: @nLight @orlandohohmeier I went with the isDeleting getter inside of the Service that we can extend in the future. My first approach was to make this very explicit but this is ok as well.

nLight: Hey @orlandohohmeier, we're ready to merge this one, feel free to press the button if the solution looks good to you.

orlandohohmeier: @bstavroulakis @nLight I hesitate to merge the PR, as I still feel that we should use `getServiceStatus` instead of adding a new `isDeleting` getter. Though I'm ok to move forward if one of you can answer my question and tell me why need the new getter and what the difference between "deleting" and "deploying" is.

orlandohohmeier:
Putting this one on hold till https://github.com/dcos/dcos-ui/pull/2459/files#discussion_r144799299 is addressed.

plugins/services/src/js/components/ServiceBreadcrumbs.js
bstavroulakis: @nLight @orlandohohmeier I removed the getter and I'm using only the getServiceStatus() now. You can give it a new swirl.

bstavroulakis:
@ahoskins Yeap, I agree

PR #2460, opened 2017-10-12
Change key to index as portDefinition name could be null

bstavroulakis:
Some frameworks in rare cases have a portDefinition name of null. Change key to index as portDefinition name could be null.

**Checklist**


orlandohohmeier:
@bstavroulakis Please shorten your commit message header (type, scope, subject) to 72 characters and make sure to always run both the `post-commit` as well as the `pre-push` hook. N.B. Moving forward we will validate commit messages in CI and fail PRs that don't adhere to the commit message guidelines.

PR #2461, opened 2017-10-16
refactor(StatusBar): rename statusBar to progressBar

weblancaster:
Rename statusBar to progressBar so it can become general and clear for usage.

Closes DCOS-18265S

**Checklist**

### DIFF
--- a/plugins/services/src/js/components/ServiceBreadcrumbs.js
+++ b/plugins/services/src/js/components/ServiceBreadcrumbs.js
@@ -62,14 +62,14 @@ class ServiceBreadcrumbs extends React.Component {
   checkBreadcrumbOverflow() {
     if (this.primaryBreadcrumbTextRef != null) {
       const availableWidth = this.getAvailableBreadcrumbWidth();
-      const statusBarWidth = this.getBreadcrumbStatusBarWidth();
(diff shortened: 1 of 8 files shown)


plugins/services/src/js/components/ServiceBreadcrumbs.js
```diff
@@ -62,14 +62,14 @@ class ServiceBreadcrumbs extends React.Component {
...
-      const statusBarWidth = this.getBreadcrumbStatusBarWidth();
+      const ProgressBarWidth = this.getBreadcrumbProgressBarWidth();
 
       if (availableWidth <= 0 && this.state.shouldRenderServiceStatus) {
         // Hide the status bar when the breadcrumbs are wider than the
         // container's width.
         this.setState({ shouldRenderServiceStatus: false });
       } else if (
-        availableWidth > statusBarWidth &&
+        availableWidth > ProgressBarWidth &&
```
bstavroulakis: Sorry for the detail, but better to keep camelCase for consistency.

weblancaster: not detail, you are right sir. I was on automatic changing everything and missed this guy is a variable

PR #2462, opened 2017-10-16
DCOS-18171: Add zone, region in the task Table and Details Page

bstavroulakis:
Add zone, region in the task Table and Details Page.

Closes DCOS-18171, DCOS-19136, DCOS-19137, DCOS-19138

**Checklist**

### DIFF
--- a/plugins/services/src/js/containers/tasks/TaskTable.js
+++ b/plugins/services/src/js/containers/tasks/TaskTable.js
@@ -20,6 +20,8 @@ const tableColumnClasses = {
   id: "task-table-column-primary",
   name: "task-table-column-name",
   host: "task-table-column-host-address",
+  zone: "task-table-column-zone-address",
+  region: "task-table-column-region-address",
   status: "task-table-column-status",
   health: "task-table-column-health",
   logs: "task-table-column-logs",
@@ -98,6 +100,24 @@ class TaskTable extends React.Component {
         heading,
         prop: "host",
         render: this.renderHost,
+        sortable: true,
+        sortFunction
+      },
+      {
+        className,
+        headerClassName: className,
+        heading,
+        prop: "zone",
+        render: this.renderZone,
+        sortable: true,
+        sortFunction
+      },
+      {
+        className,
+        headerClassName: className,
+        heading,
+        prop: "region",
+        render: this.renderRegion,
         sortable: true,
         sortFunction
       },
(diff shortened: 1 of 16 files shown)


plugins/services/src/js/containers/tasks/TaskTable.js
```diff
@@ -296,6 +317,38 @@ class TaskTable extends React.Component {
     );
   }
 
+  renderRegion(prop, task) {
+    if (!task.region) {
+      return "Primary";
```
orlandohohmeier: @bstavroulakis Let's use the same fallback as in [`NodesTable.js#L119`](https://github.com/dcos/dcos-ui/blob/master/plugins/nodes/src/js/components/NodesTable.js#L119) to be consistent. /cc @Poltergeist

orlandohohmeier:
@bstavroulakis Please squash the commits to ensure they work independently of any later commits.

plugins/services/src/js/containers/tasks/TaskTable.js
bstavroulakis: @orlandohohmeier @Poltergeist I'll have a definite answer from product to see what to do here.

src/js/structs/CompositeState.js
```diff
@@ -103,6 +104,10 @@ class CompositeState {
   getNodesList() {
     return new NodesList({ items: this.data.slaves });
   }
+
+  getNodeMaster() {
+    return new Node(this.data.leader_info);
```
nLight: We're missing this field in V1 😱 
There's a ticket for the mesos team https://issues.apache.org/jira/browse/MESOS-8078

nLight: @dcos/frontend-foundation

MatApple: @nLight have you spoken to Vinod about getting this prioritized?

nLight: Haven't spoken in person, I pinged him on slack, no reply though.

nLight: Ok, so _there is_ a call for that.

`GET_MASTER` call gives the same info and it also should include region and zone info if that master is configured in a such way.

As @vinodkone told me, in V1 only leading master responds to request, so the info we get via `GET_MASTER` is equal to `leader_info`

nLight: Please continue using `leader_info`, I will pick it up from three.

bstavroulakis:
Tested! Works as expected, with responsive priority column rules as well.

plugins/services/src/js/utils/TaskUtil.js
```diff
@@ -112,6 +113,71 @@ const TaskUtil = {
...
     });
+  },
+
+  getRegionName(task) {
+    const node = this.getNode(task);
+    if (!node) {
+      return "(Local)";
+    }
+
+    const masterNode = CompositeState.getMasterNode();
```
orlandohohmeier: Not blocking as this is following existing "patterns" but we should try hard to write _pure_ utils that don't access any kind of data "store" directly.

orlandohohmeier:
Im going to squash merge this PR as the commit message don't follow our guidelines.

PR #2463, opened 2017-10-17
Introduce DI/IoC container

nLight:
In order to loosen coupling in the codebase and set the foundation for the modularization. DI can help with singletons as the container will be the source of truth.

On top of that it's easier to test code with DI, just binding a mock to the container instead of relying on Jest or other tooling that isn't ideal.

So currently there's only npm dependencies and an empty container.

The first citizen of the container's going to be the MesosStream.

### DIFF
--- a/src/js/container.js
+++ b/src/js/container.js
@@ -1,0 +1,5 @@
+import { Container } from "inversify";
+
+const container = new Container();
+
+export default container;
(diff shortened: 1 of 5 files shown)


src/js/container.js
```diff
@@ -0,0 +1,8 @@
+import { Container } from "inversify";
+/* eslint-disable no-unused-vars */
+import Reflect from "reflect-metadata";
```
orlandohohmeier: Why do we need to import reflect?

nLight: Moved to webpack config.

nLight:
Please give it another shot!

orlandohohmeier:
L👀ks good!

PR #2468, opened 2017-10-18
DCOS 19137: Region/ Zones in tasks details

bstavroulakis:
Add Region/ Zones in task details.

**Checklist**


bstavroulakis:
Depends on #2462

PR #2472, opened 2017-10-24
DCOS 19139: DSL tasks filter

bstavroulakis:
Add dsl filter in tasks table page.

Integration test will be added in another PR.

Filter by:
text, zone, region, state

image

image

**Checklist**

### DIFF
--- a/plugins/nodes/src/js/pages/nodes/NodeDetailTaskTab.js
+++ b/plugins/nodes/src/js/pages/nodes/NodeDetailTaskTab.js
@@ -46,10 +46,10 @@ class NodeDetailTaskTab extends React.Component {
   }
 
   render() {
-    const { nodeID } = this.props.params;
+    const { location, params, params: { nodeID } } = this.props;
     const tasks = MesosStateStore.getTasksFromNodeID(nodeID);
 
-    return <TasksContainer params={this.props.params} tasks={tasks} />;
+    return <TasksContainer location={location} params={params} tasks={tasks} />;
   }
 }
 
--- a/plugins/services/src/js/components/dsl/FuzzyTextDSLSection.js
+++ b/plugins/services/src/js/components/dsl/FuzzyTextDSLSection.js
@@ -34,6 +34,7 @@ class FuzzyTextDSLSection extends React.Component {
   componentWillReceiveProps(nextProps) {
     const { expression } = nextProps;
     const data = DSLUtil.getPartValues(expression, EXPRESSION_PARTS);
+
     this.setState({ data });
   }
 
@@ -69,7 +70,7 @@ class FuzzyTextDSLSection extends React.Component {
             name="text"
             onChange={this.handleTextChange}
             placeholder="Enter words found in name"
-            value={data.text}
+            value={data.text || ""}
           />
         </FormGroup>
 
--- a/plugins/services/src/js/components/dsl/TaskRegionDSLSection.js
+++ b/plugins/services/src/js/components/dsl/TaskRegionDSLSection.js
@@ -1,0 +1,82 @@
+import React, { PropTypes } from "react";
+
+import DSLCombinerTypes from "#SRC/js/constants/DSLCombinerTypes";
+import DSLExpression from "#SRC/js/structs/DSLExpression";
+import DSLExpressionPart from "#SRC/js/structs/DSLExpressionPart";
+import DSLFormWithExpressionUpdates
+  from "#SRC/js/components/DSLFormWithExpressionUpdates";
+import DSLUtil from "#SRC/js/utils/DSLUtil";
+import FieldInput from "#SRC/js/components/form/FieldInput";
+import FieldLabel from "#SRC/js/components/form/FieldLabel";
+import FormGroup from "#SRC/js/components/form/FormGroup";
+
+const EXPRESSION_PARTS = {};
+
+class TaskRegionDSLSection extends React.Component {
+  render() {
+    const { expression, onChange, defaultData: { regions } } = this.props;
+
+    if (!regions || regions.length === 0) {
+      return null;
+    }
+
+    regions.forEach(function(region) {
+      EXPRESSION_PARTS[`region_${region}`] = DSLExpressionPart.attribute(
+        "region",
+        region
+      );
+    }, this);
+
+    const enabled = DSLUtil.canProcessParts(expression, EXPRESSION_PARTS);
+    const data = DSLUtil.getPartValues(expression, EXPRESSION_PARTS);
+
+    return (
+      <DSLFormWithExpressionUpdates
+        enabled={enabled}
+        expression={expression}
+        groupCombiner={DSLCombinerTypes.AND}
+        itemCombiner={DSLCombinerTypes.OR}
+        onChange={onChange}
+        parts={EXPRESSION_PARTS}
+      >
+
+        <label>Regions</label>
+        <div className="row">
+          <div className="column-12">
+            {regions.map(region => {
+              const regionName = `region_${region}`;
+
+              return (
+                <FormGroup key={regionName}>
+                  <FieldLabel>
+                    <FieldInput
+                      checked={data[regionName]}
+                      disabled={!enabled}
+                      name={regionName}
+                      type="checkbox"
+                    />
+                    {region}
+                  </FieldLabel>
+                </FormGroup>
+              );
+            })}
+          </div>
+        </div>
+      </DSLFormWithExpressionUpdates>
+    );
+  }
+}
+
+TaskRegionDSLSection.defaultProps = {
+  expression: new DSLExpression(""),
+  onChange() {},
+  defaultData: { regions: [] }
+};
+
+TaskRegionDSLSection.propTypes = {
+  expression: PropTypes.instanceOf(DSLExpression).isRequired,
+  onChange: PropTypes.func.isRequired,
+  defaultData: PropTypes.object.isRequired
+};
+
+module.exports = TaskRegionDSLSection;
--- a/plugins/services/src/js/components/dsl/TaskZoneDSLSection.js
+++ b/plugins/services/src/js/components/dsl/TaskZoneDSLSection.js
@@ -1,0 +1,82 @@
+import React, { PropTypes } from "react";
+
+import DSLCombinerTypes from "#SRC/js/constants/DSLCombinerTypes";
+import DSLExpression from "#SRC/js/structs/DSLExpression";
+import DSLExpressionPart from "#SRC/js/structs/DSLExpressionPart";
+import DSLFormWithExpressionUpdates
+  from "#SRC/js/components/DSLFormWithExpressionUpdates";
+import DSLUtil from "#SRC/js/utils/DSLUtil";
+import FieldInput from "#SRC/js/components/form/FieldInput";
+import FieldLabel from "#SRC/js/components/form/FieldLabel";
+import FormGroup from "#SRC/js/components/form/FormGroup";
+
+const EXPRESSION_PARTS = {};
+
+class TaskZoneDSLSection extends React.Component {
+  render() {
+    const { expression, onChange, defaultData: { zones } } = this.props;
+
+    if (!zones || zones.length === 0) {
+      return null;
+    }
+
+    zones.forEach(function(zone) {
+      EXPRESSION_PARTS[`zone_${zone}`] = DSLExpressionPart.attribute(
+        "zone",
+        zone
+      );
+    }, this);
+
+    const enabled = DSLUtil.canProcessParts(expression, EXPRESSION_PARTS);
+    const data = DSLUtil.getPartValues(expression, EXPRESSION_PARTS);
+
+    return (
+      <DSLFormWithExpressionUpdates
+        enabled={enabled}
+        expression={expression}
+        groupCombiner={DSLCombinerTypes.AND}
+        itemCombiner={DSLCombinerTypes.OR}
+        onChange={onChange}
+        parts={EXPRESSION_PARTS}
+      >
+
+        <label>Zones</label>
+        <div className="row">
+          <div className="column-12">
+            {zones.map(zone => {
+              const zoneName = `zone_${zone}`;
+
+              return (
+                <FormGroup key={zoneName}>
+                  <FieldLabel>
+                    <FieldInput
+                      checked={data[zoneName]}
+                      disabled={!enabled}
+                      name={zoneName}
+                      type="checkbox"
+                    />
+                    {zone}
+                  </FieldLabel>
+                </FormGroup>
+              );
+            })}
+          </div>
+        </div>
+      </DSLFormWithExpressionUpdates>
+    );
+  }
+}
+
+TaskZoneDSLSection.defaultProps = {
+  expression: new DSLExpression(""),
+  onChange() {},
+  defaultData: { zones: [] }
+};
+
+TaskZoneDSLSection.propTypes = {
+  expression: PropTypes.instanceOf(DSLExpression).isRequired,
+  onChange: PropTypes.func.isRequired,
+  defaultData: PropTypes.object.isRequired
+};
+
+module.exports = TaskZoneDSLSection;
--- a/plugins/services/src/js/containers/tasks/TasksContainer.js
+++ b/plugins/services/src/js/containers/tasks/TasksContainer.js
@@ -1,13 +1,26 @@
 import React, { PropTypes } from "react";
+import { routerShape } from "react-router";
 
 import AppDispatcher from "#SRC/js/events/AppDispatcher";
 import ContainerUtil from "#SRC/js/utils/ContainerUtil";
+import DSLExpression from "#SRC/js/structs/DSLExpression";
+import DSLFilterList from "#SRC/js/structs/DSLFilterList";
+import Tree from "#SRC/js/structs/Tree";
+
+import TasksStatusFilter
+  from "#PLUGINS/services/src/js/filters/TasksStatusFilter";
+import TasksZoneFilter from "#PLUGINS/services/src/js/filters/TasksZoneFilter";
+import TasksRegionFilter
+  from "#PLUGINS/services/src/js/filters/TasksRegionFilter";
+import TaskNameTextFilter
+  from "#PLUGINS/services/src/js/filters/TaskNameTextFilter";
 
 import ActionKeys from "../../constants/ActionKeys";
 import MarathonActions from "../../events/MarathonActions";
 import ServiceActionItem from "../../constants/ServiceActionItem";
 import TaskModals from "../../components/modals/TaskModals";
 import TasksView from "./TasksView";
+import TaskUtil from "../../utils/TaskUtil";
 
 import {
   REQUEST_MARATHON_TASK_KILL_ERROR,
@@ -17,6 +30,7 @@ import {
 const METHODS_TO_BIND = [
   "handleServerAction",
   "handleModalClose",
+  "handleExpressionChange",
   "clearActionError",
   "killTasks"
 ];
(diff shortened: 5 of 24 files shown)


plugins/services/src/js/components/dsl/TaskRegionDSLSection.js
```diff
@@ -0,0 +1,75 @@
...
+const EXPRESSION_PARTS = {};
+
+class TaskRegionDSLSection extends React.Component {
+  render() {
+    const { expression, onChange } = this.props;
+    const { regions } = this.props.defaultData || [];
+
+    if (regions) {
+      regions.forEach(function(region) {
+        EXPRESSION_PARTS["region_" + region] = DSLExpressionPart.attribute(
```
ahoskins: Use template string like below.

plugins/services/src/js/components/dsl/TaskRegionDSLSection.js
```diff
@@ -0,0 +1,75 @@
...
+          </div>
+        </div>
+      </DSLFormWithExpressionUpdates>
+    );
+  }
+}
+
+TaskRegionDSLSection.propTypes = {
+  onChange: PropTypes.func.isRequired,
+  expression: PropTypes.instanceOf(DSLExpression).isRequired
```
ahoskins: Add defaultData to propTypes.

plugins/services/src/js/components/dsl/TaskZoneDSLSection.js
```diff
@@ -0,0 +1,75 @@
...
+          </div>
+        </div>
+      </DSLFormWithExpressionUpdates>
+    );
+  }
+}
+
+TaskZoneDSLSection.propTypes = {
+  onChange: PropTypes.func.isRequired,
+  expression: PropTypes.instanceOf(DSLExpression).isRequired
```
ahoskins: defaultData

plugins/services/src/js/containers/tasks/TasksContainer.js
```diff
@@ -27,20 +41,31 @@ class TasksContainer extends React.Component {
...
+      defaultFilterData: { zones: [], regions: [] }
     };
 
     METHODS_TO_BIND.forEach(method => {
       this[method] = this[method].bind(this);
     });
   }
 
+  componentWillMount() {
+    this.propsToState(this.props);
```
ahoskins: https://twitter.com/dan_abramov/status/749710501916139520?lang=en 😏 .

I think this should just be calculated in the render method.

plugins/services/src/js/filters/TaskNameTextFilter.js
```diff
@@ -0,0 +1,31 @@
...
+   * @override
+   */
+  filterCanHandle(filterType) {
+    return (
+      filterType === DSLFilterTypes.EXACT || filterType === DSLFilterTypes.FUZZY
+    );
+  }
+
+  /**
+   * Keep only services whose status
```
ahoskins: Looks like a partially written comment

plugins/services/src/js/filters/TasksStatusFilter.js
```diff
@@ -0,0 +1,43 @@
...
+    );
+  }
+
+  /**
+   * Keep only tasks whose state matches the value of
+   * the `is` label
+   *
+   * @override
+   */
+  filterApply(resultset, filterType, filterArguments) {
```
ahoskins: Why is filterType passed into when not being used?

plugins/nodes/src/js/pages/nodes/NodeDetailTaskTab.js
```diff
@@ -46,10 +46,10 @@ class NodeDetailTaskTab extends React.Component {
   }
 
   render() {
-    const { nodeID } = this.props.params;
-    const tasks = MesosStateStore.getTasksFromNodeID(nodeID);
+    const { location, params } = this.props;
```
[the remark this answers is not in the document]
bstavroulakis: Good catch!

plugins/services/src/js/components/dsl/TaskRegionDSLSection.js
bstavroulakis: Yes

plugins/services/src/js/components/dsl/TaskRegionDSLSection.js
bstavroulakis: Added

plugins/services/src/js/components/dsl/TaskZoneDSLSection.js
bstavroulakis: Added

plugins/services/src/js/containers/tasks/TasksContainer.js
bstavroulakis: I agree with this in general, but in this case I think it is better to keep the filter and defaultFilterData in state. When they don't change from the props we want the *exact* same object, if they change this will trigger => https://github.com/dcos/dcos-ui/blob/master/src/js/components/DSLFilterField.js#L56 and reset everything.

plugins/services/src/js/containers/tasks/TasksContainer.js
```diff
@@ -77,6 +102,70 @@ class TasksContainer extends React.Component {
...
+
+          return node.getRegionName();
+        })
+      )
+    );
+
+    // If no region/ zones added from props return
+    if (
+      newRegions.length === regions.length &&
+      regions.every((v, i) => v === newRegions[i]) &&
```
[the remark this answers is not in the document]
bstavroulakis: I agree, included a better check

plugins/services/src/js/filters/TasksStatusFilter.js
bstavroulakis: Although not used, we cannot change this as the callback that triggers this will pass the parameters in the exact order specified. If we remove it the filterArguments will be the filterType.

plugins/services/src/js/containers/tasks/TasksContainer.js
ahoskins: Makes sense. So they are stored in state because you want to compare with the previous values, and decide if you update the `filters` and `defaultFilterData` triggering a full re-render. I agree, this is good to keep 👍

plugins/services/src/js/components/dsl/TaskRegionDSLSection.js
```diff
@@ -0,0 +1,80 @@
...
+import FieldInput from "#SRC/js/components/form/FieldInput";
+import FieldLabel from "#SRC/js/components/form/FieldLabel";
+import FormGroup from "#SRC/js/components/form/FormGroup";
+
+const EXPRESSION_PARTS = {};
+
+class TaskRegionDSLSection extends React.Component {
+  render() {
+    const { expression, onChange } = this.props;
+    const { regions } = this.props.defaultData || [];
```
weblancaster: Do we need to check `regions` like in some cases below (`if (regions) {...`) even though we have the "default" to be `[]` when falsey

plugins/services/src/js/components/dsl/TaskRegionDSLSection.js
```diff
@@ -0,0 +1,80 @@
...
+
+    if (regions) {
```
weblancaster: related to my comment above https://github.com/dcos/dcos-ui/pull/2472/files#r157902372

plugins/services/src/js/components/dsl/TaskZoneDSLSection.js
```diff
@@ -0,0 +1,80 @@
...
+import FieldInput from "#SRC/js/components/form/FieldInput";
+import FieldLabel from "#SRC/js/components/form/FieldLabel";
+import FormGroup from "#SRC/js/components/form/FormGroup";
+
+const EXPRESSION_PARTS = {};
+
+class TaskZoneDSLSection extends React.Component {
+  render() {
+    const { expression, onChange } = this.props;
+    const { zones } = this.props.defaultData || [];
```
weblancaster: related to my comment above https://github.com/dcos/dcos-ui/pull/2472/files#r157902372

plugins/services/src/js/containers/tasks/TasksContainer.js
```diff
@@ -77,6 +102,70 @@ class TasksContainer extends React.Component {
...
+    ]);
+
+    this.setState({
+      filterExpression: new DSLExpression(query),
+      filters,
+      defaultFilterData: { regions: newRegions, zones: newZones }
+    });
+  }
+
   fetchData() {
```
weblancaster: Can we delete this line please, whenever we see something small and not in use like this we should be able to remove without opening a new PR or creating new Jira for this.

plugins/services/src/js/containers/tasks/TasksView.js
```diff
@@ -225,19 +188,46 @@ class TasksView extends mixin(SaveStateMixin) {
...
+  getFilterBar() {
+    const {
+      filters,
+      filterExpression,
+      handleExpressionChange,
+      defaultFilterData
+    } = this.props;
+
+    const hostClasses = classNames({
+      "column-medium-5": !filterExpression.value,
```
weblancaster: `const filterExpression = filterExpression.value`; then you can re-use. Not big deal though.

plugins/services/src/js/containers/tasks/TasksView.js
```diff
@@ -1,44 +1,39 @@
...
 import StringUtil from "#SRC/js/utils/StringUtil";
 import { isSDKService } from "#SRC/js/utils/ServiceUtil";
 
+import TaskStatusDSLSection from "../../components/dsl/TaskStatusDSLSection";
+import TaskZoneDSLSection from "../../components/dsl/TaskZoneDSLSection";
+import TaskRegionDSLSection from "../../components/dsl/TaskRegionDSLSection";
+import FuzzyTextDSLSection from "../../components/dsl/FuzzyTextDSLSection";
+
 import ServiceStatusTypes from "../../constants/ServiceStatusTypes";
 import GraphQLTaskUtil from "../../utils/GraphQLTaskUtil";
```
weblancaster: kinda confusing having `GraphQLTaskUtil ` even though we don't use it. No need to change in this PR but would be nice a follow up PR renaming/refactoring (I can do that too if you want to)

plugins/services/src/js/containers/tasks/TasksView.js
```diff
@@ -246,36 +236,26 @@ class TasksView extends mixin(SaveStateMixin) {
       rightAlignLastNChildren = 1;
     }
 
-    filteredTasks = filteredTasks.map(GraphQLTaskUtil.mergeData);
+    const mergedTasks = tasks.map(GraphQLTaskUtil.mergeData);
```
weblancaster: related to https://github.com/dcos/dcos-ui/pull/2472/files#r157931754

plugins/services/src/js/containers/tasks/TasksView.js
```diff
@@ -246,36 +236,26 @@ class TasksView extends mixin(SaveStateMixin) {
...
         <FilterHeadline
-          currentLength={filteredTasks.length}
+          currentLength={mergedTasks.length}
           inverseStyle={inverseStyle}
-          isFiltering={filterByStatus !== "all" || searchString !== ""}
-          onReset={this.resetFilter}
           name={"task"}
-          totalLength={tasks.length}
+          totalLength={totalTasks}
+          onReset={() => {
```
weblancaster: btw, no need for the block in the arrow function here.

can be `onReset={() => handleExpressionChange({ value: "" })`

plugins/services/src/js/filters/TasksRegionFilter.js
```diff
@@ -0,0 +1,54 @@
...
+import DSLFilter from "#SRC/js/structs/DSLFilter";
+import TaskUtil from "../utils/TaskUtil";
+
+const LABEL = "region";
+
+/**
+ * This filter handles the `region:XXXX` for tasks
+ */
+class TasksRegionFilter extends DSLFilter {
+  constructor(regions) {
```
weblancaster: `regions = []` for default then `this.regions = regions` not big deal.

plugins/services/src/js/filters/TasksRegionFilter.js
```diff
@@ -0,0 +1,54 @@
...
+    super();
+    this.regions = [];
+
+    if (Array.isArray(regions)) {
```
weblancaster: Are we expecting always Array? because otherwise what happens if regions aren't Array but the method get called still, that will probably throw an error.

plugins/services/src/js/filters/TasksStatusFilter.js
```diff
@@ -0,0 +1,43 @@
...
+    const testStatus = LABEL_TO_STATUS[filterArguments.text.toLowerCase()];
+
+    return resultset.filterItems(task => {
```
weblancaster: Should this be `resultSet`?

plugins/services/src/js/filters/TasksZoneFilter.js
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * This filter handles the `zone:XXXX` for tasks
+ */
+class TasksZoneFilter extends DSLFilter {
+  constructor(zones) {
+    super();
+    this.zones = [];
+
+    if (Array.isArray(zones)) {
```
weblancaster: Same as https://github.com/dcos/dcos-ui/pull/2472/files#r157932788

plugins/services/src/js/filters/TasksZoneFilter.js
```diff
@@ -0,0 +1,54 @@
...
+
+  /**
+   * Keep only tasks whose zone matches the value of
+   * the `zone` label
+   *
+   * @override
+   */
+  filterApply(resultset, filterType, filterArguments) {
+    let zone = "";
+    if (this.zones.includes(filterArguments.text.toLowerCase())) {
```
weblancaster: why not just store in a variable `filterArguments.text.toLowerCase()`

plugins/services/src/js/filters/__tests__/TaskNameTextFilter-test.js
```diff
@@ -0,0 +1,41 @@
...
+      {
+        name: "node-1-server__1"
+      },
+      {
+        name: "node-0-server__2"
+      }
+    ];
+  });
+
+  it("Should match parts of task name", function() {
```
weblancaster: Remove the `should` from all tests please. We are not following this pattern anymore.

plugins/services/src/js/components/dsl/TaskRegionDSLSection.js
bstavroulakis: I've now added a check to return nothing when it is empty.

plugins/services/src/js/components/dsl/TaskZoneDSLSection.js
bstavroulakis: I've now added a check to return nothing when it is empty, so the UI won't show the zone filter if no zone filters exist.

orlandohohmeier:
filter-issue

There seems to be an issue with the filtering.

bstavroulakis:
Alfred please retest this

plugins/services/src/js/components/dsl/TaskRegionDSLSection.js
```diff
@@ -0,0 +1,81 @@
...
+import FieldInput from "#SRC/js/components/form/FieldInput";
+import FieldLabel from "#SRC/js/components/form/FieldLabel";
+import FormGroup from "#SRC/js/components/form/FormGroup";
+
+const EXPRESSION_PARTS = {};
+
+class TaskRegionDSLSection extends React.Component {
+  render() {
+    const { expression, onChange } = this.props;
+    const { regions } = this.props.defaultData || [];
```
orlandohohmeier: 😕  I'm not sure I understand what you we're trying, but this line looks wrong to me. I'd provide a default for `defaultData` using `defaultProps` and would ensure that's an object using `propTypes` so that it's save to simply destruct  the props. Use the following code in case you want to provide a default for `regions`.

```js
 const { regions=[] } = this.props.defaultData;
```

plugins/services/src/js/components/dsl/TaskZoneDSLSection.js
```diff
@@ -0,0 +1,81 @@
...
```
orlandohohmeier: ⬆️  Please see comments regarding https://github.com/dcos/dcos-ui/pull/2472/files#diff-a87db97c709e1b45a1ad9dc82ba6da1fR18

plugins/services/src/js/containers/tasks/TasksContainer.js
orlandohohmeier: 💭  Can't we achieve the same without the need to store this in state by using `shouldComponentUpdate`?

plugins/services/src/js/containers/tasks/TasksView.js
orlandohohmeier: @weblancaster would love if you could fix the name.

orlandohohmeier:
😇  please excuse the late review and change requests.

plugins/services/src/js/components/dsl/FuzzyTextDSLSection.js
```diff
@@ -33,7 +33,12 @@ class FuzzyTextDSLSection extends React.Component {
 
   componentWillReceiveProps(nextProps) {
     const { expression } = nextProps;
-    const data = DSLUtil.getPartValues(expression, EXPRESSION_PARTS);
+    let data = DSLUtil.getPartValues(expression, EXPRESSION_PARTS);
+
+    if (!data || !data.text) {
+      data = { text: "" };
```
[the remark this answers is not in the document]
bstavroulakis: If the text is not defined it doesn't reset the input field to an empty value when we reset the filters.

plugins/services/src/js/containers/tasks/TasksContainer.js
bstavroulakis: @orlandohohmeier I'm interested only in the filter changes. We receive from the properties the tasks and various other information.
So, if we receive some other tasks with the same region, zones I would like to update this component but keep the filter the same. If we return false when the filter doesn't change then it will become inactive and won't refresh with the new tasks. Also, it wouldn't be correct I think to compare the tasks and all of the other properties in the shouldComponentUpdate, it would become a bit complex.

plugins/services/src/js/components/dsl/FuzzyTextDSLSection.js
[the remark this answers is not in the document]
orlandohohmeier: Hm, ok. Then let's move that part to the util so that the util always provides correct and wellformed data.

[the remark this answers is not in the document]
bstavroulakis: I updated the check in the input since it is needed only over there.

I also responded to the comment for the shouldComponentUpdate over here => https://github.com/dcos/dcos-ui/pull/2472#discussion_r158072691

plugins/services/src/js/containers/tasks/TasksContainer.js
orlandohohmeier: Sounds like a data/control flow issues which we can fix by adjusting the components, token and filter mechanism. Anyway, let's move this implementation forward and create an issue so that we don't forget to revisit the implementation.

orlandohohmeier:
Please move https://github.com/dcos/dcos-ui/commit/c64e0b15decba4db94be7268890c3fab507e6658 to a different PR before merging this one.

orlandohohmeier:
✨ Awesome

bstavroulakis:
Alfred please retest this

PR #2474, opened 2017-10-24
DCOS_OSS 1784: Update messages to match UI Kit

leemunroe:
CNVS has introduced messages for styling feedback messages which affects components like validation alerts. This PR renames existing `alert` classes and removes any unneeded CSS.

Before
image
image

After
image
image


MatApple:
@leemunroe @bstavroulakis I've rebased this one and it's ready to go.

PR #2475, opened 2017-10-24
feat(servicesTable) status column redesign

weblancaster:
This is a redesign of the service status column, where we are removing the health checks from the progressBars and replacing to with running/total instances instead.

PR contains the necessary subtasks of DCOS-17968, and for more info please look at the service status documentation included on DCOS-17968.

Couple of things the PR include:
- new spinner icon
- improve stopped/minus icon size/color
- refactor/rename everything related to service healthStatus to serviceStatus
- remove the logic from the servicesTable > renderStatus method to separated components ServiceStatusIcon and ServiceStatusProgressBar
- update ServiceStatus constants to new wording
- update service application logic to factor in the new state/status recovering, suspended to stopped, delayed to deploying.

Subtasks:
- DCOS-18264
- DCOS-18268

**before**
screen shot 2017-11-01 at 12 46 43 am

**after**
screen shot 2017-11-01 at 12 24 40 am

**Checklist**


src/js/components/StatusIcon.js
```diff
@@ -0,0 +1,56 @@
...
+        color={iconState.color}
+        id={iconState.id}
+        size="mini"
+      />
+    );
+  }
+}
+
+StatusIcon.propTypes = {
+  label: PropTypes.string.isRequired
```
MatApple: I think this is supposed to be `state`. Please also ensure we use a `oneOf` propType here so we have an explicit enum of accepted states to prevent accidental misconfiguration.
e.g.
```
state: PropTypes.oneOfType([
    "RUNNING",
    "TRANSITION",
    "STOPPED",
    "WARNING"
  ])

File: src/js/components/StatusIcon.js [hunk 2]
```diff
@@ -0,0 +1,56 @@
...
+ * Status Icon is meant to be used across
+ * the application where it need to display any of the state below
+ * All green (RUNNING), transition (DEPLOYING, RECOVERING), Warning
+ *
+ * @class StatusIcon
+ * @extends {Component}
+ */
+class StatusIcon extends Component {
+  render() {
+    const state = this.props.state.toUpperCase();
```
MatApple: Please adjust so we don't transform the prop to uppercase - instead, be explicit about only accepting uppercase states which simplifies the component and provides a more explicit and clean API.

plugins/services/src/js/components/__tests__/ServiceStatusProgressBar-test.js
```diff
@@ -17,11 +17,14 @@ const testData = {
...
 
 const testInstanceCount = 4;
-describe("#HealthBar", function() {
+describe("#ServiceStatusProgressBar", function() {
   beforeEach(function() {
     this.container = global.document.createElement("div");
     this.instance = ReactDOM.render(
-      <HealthBar tasksSummary={testData} instancesCount={testInstanceCount} />,
+      <ServiceStatusProgressBar
+        tasksSummary={testData}
```
MatApple: This component only accepts one prop `service` :)

plugins/services/src/js/components/ServiceStatusProgressBar.js
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  render() {
+    const { service } = this.props;
+    const instancesCount = service.getInstancesCount();
+    const serviceStatus = service.getStatus();
+    const tasksRunning = service.getTaskCount();
+
+    if (instancesCount !== tasksRunning) {
+      return this.getProgressBar([
```
MatApple: Curious why you chose to split some of the rendering logic into `getProgressBar`?

Perhaps I'm missing something here but I see the logic could be much simpler if it was all contained in a single method as we have some duplication across the two - mostly same `const`s and same equality checks like `tasksRunning === instancesCount` (or it's inverse) and data is passed from one to the other that doesn't seem necessary.

plugins/services/src/js/components/ServiceStatusProgressBar.js
```diff
@@ -0,0 +1,88 @@
...
+          total={instancesCount}
+        />
+      </Tooltip>
+    );
+  }
+
+  render() {
+    const { service } = this.props;
+    const instancesCount = service.getInstancesCount();
+    const serviceStatus = service.getStatus();
```
MatApple: Why `getStatus` and not `getServiceStatus` like `StatusIcon`?

plugins/services/src/js/components/ServiceStatusProgressBar.js
```diff
@@ -0,0 +1,88 @@
...
+    if (instancesCount !== tasksRunning) {
+      return this.getProgressBar([
+        {
+          state: "success",
+          value: tasksRunning
+        }
+      ]);
+    }
+
+    if (serviceStatus === ServiceStatus.DELETING.displayName) {
```
MatApple: Please adjust to align with comments on `StatusIcon` so we compare:
```
const status = service.getServiceStatus();
if (status === ServiceStatus.DELETING) {

}
```

plugins/services/src/js/components/ServiceStatusProgressBar.js
```diff
@@ -0,0 +1,88 @@
...
+      return this.getProgressBar([
+        {
+          state: "",
+          value: 1
```
MatApple: Should this value be hardcoded?

plugins/services/src/js/components/ServiceStatusProgressBar.js
```diff
@@ -0,0 +1,88 @@
...
+          state: "success",
+          value: tasksRunning
+        }
+      ]);
+    }
+
+    if (serviceStatus === ServiceStatus.DELETING.displayName) {
+      return this.getProgressBar([
+        {
+          state: "",
```
MatApple: How come we don't pass a valid state?

plugins/services/src/js/components/ServiceStatusProgressBar.js
weblancaster: `getStatus` uses `getServiceStatus` but only returns the display name which is what I need.

plugins/services/src/js/components/ServiceStatusProgressBar.js
weblancaster: In this case `ServiceStatus.DELETING` is 
```
DELETING: {
    key: ServiceStatusTypes.DELETING,
    displayName: ServiceStatusLabels.DELETING
  }
```
which isn't what I want to compare

Did you mean `ServiceStatusLabels.DELETING`?

plugins/services/src/js/components/ServiceStatusProgressBar.js
weblancaster: This was WIP that's why you don't see any state.

plugins/services/src/js/components/ServiceStatusProgressBar.js
weblancaster: This was WIP that's why you see hardcoded.

plugins/services/src/js/components/__tests__/ServiceStatusProgressBar-test.js
weblancaster: Yes, this PR is a WIP.

src/js/components/StatusIcon.js
weblancaster: Correct. The PR is WIP.

plugins/services/src/js/components/ServiceStatusProgressBar.js
weblancaster: At first implementation/solution I ended up with that but I have refactored. thanks.

src/js/components/StatusIcon.js
```diff
@@ -0,0 +1,56 @@
+import React, { PropTypes, Component } from "react";
+
+import Icon from "#SRC/js/components/Icon";
+
+const ICONS_STATE = {
```
ahoskins: Since these states are also used in `ServiceStatusIcon` I think there should this should be moved to a constants file.

plugins/services/src/js/components/ServiceStatusProgressBar.js
```diff
@@ -0,0 +1,81 @@
...
+    const instancesCount = service.getInstancesCount();
+    const serviceStatus = service.getStatus();
+    const tasksRunning = service.getTaskCount();
+
+    if (
+      serviceStatus === ServiceStatus.RUNNING.displayName ||
+      serviceStatus === ServiceStatus.STOPPED.displayName
+    ) {
+      return null;
+    } else if (serviceStatus === ServiceStatus.DELETING.displayName) {
```
ahoskins: I think we like to fully flatten these kind of `if else` blocks.

```
if (RUNNING || STOPPED) {
   return ...
}

if (DELETING) {
  return ...
}
```

plugins/services/src/js/components/ServiceStatusWarningWithDebugInstruction.js
```diff
@@ -1,7 +1,7 @@
 import DateUtil from "#SRC/js/utils/DateUtil";
-import ServiceStatusWarning from "./ServiceStatusWarning";
+import ServiceStatusIcon from "./ServiceStatusIcon";
 
-class ServiceStatusWarningWithDebugInstruction extends ServiceStatusWarning {
+class ServiceStatusWarningWithDebugInstruction extends ServiceStatusIcon {
   getTooltipContent(timeWaiting) {
```
ahoskins: I know this PR is WIP, but I don't understand how this works because this method is not in the superclass.

src/img/components/icons/icon-system/spinner.svg
```diff
@@ -0,0 +1,33 @@
+<svg class="lds-spinner" width="200px"  height="200px"  xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" viewBox="0 0 100 100" preserveAspectRatio="xMidYMid" style="background: none;"><g transform="rotate(0 50 50)">
```
ahoskins: Nice, this does indeed spin 🌀

plugins/services/src/js/components/ServiceStatusIcon.js
```diff
@@ -95,28 +97,42 @@ class ServiceStatusWarning extends Component {
...
+      );
     }
 
     // Display the declined offers warning if that's causing a delay. If not,
     // we try to display an unable to launch warning, then default to nothing.
     return (
-      this.getDeclinedOffersWarning(item) ||
-      this.getUnableToLaunchWarning(item) ||
-      <noscript />
+      this.getDeclinedOffersWarning(service) ||
```
ahoskins: IMO changing these methods to just return JSX, and moving the conditional logic into the `render` is clearer. Less jumping around tracing things, and the methods will then have a consistent return type, rather than sometimes returning `null`:

```
render() {
 ...
 if (DeclinedOffersUtil.shouldDisplayDeclined()) {
     return this.getDeclinedOffersWarning()
 }

 if (this.shouldShowUnableToLaunch()) {
     return this.getUnableToLaunchWarning()
 }

 return <StatusIcon />
}

```

plugins/services/src/js/components/ServiceStatusProgressBar.js
weblancaster: why?

plugins/services/src/js/components/ServiceStatusWarningWithDebugInstruction.js
weblancaster: Good question, but if you look at the code base today we have there.

I will take a look at it to make sure we are not using anywhere once I have everything else 100%

src/js/components/StatusIcon.js
weblancaster: Not really used in `ServiceStatusIcon` because in `ServiceStatusIcon` we use the `serviceStatus` constants and here I only use the same state convention so we don't need to map serviceStatus to statusIcon. that being said only `transition` isn't a mapped state. 

I'm fine moving to constants though but from my conversation with Orlando and Matt looks like this is ideal since isn't a good idea to always use constants.

plugins/services/src/js/components/ServiceStatusIcon.js
weblancaster: When you don't want to return a component you would do `null` or `<noscript />` as was previously, so I believe returning `null` here is ok.

I also believe the change suggested don't change much.

weblancaster:
@ahoskins the breadcrumb isn't part of this PR. Will come next once this one is merged.

plugins/services/src/js/components/ServiceStatusIcon.js
ahoskins: Yes, I agree it's fine either way. It was just a suggestion for what I think is better grouping of the conditionals all in the same place.

src/js/components/StatusIcon.js
ahoskins: Okay, I just referring to this in particular: https://github.com/dcos/dcos-ui/pull/2475/files#diff-357cc9e62242bdfbc72ddfa7eaa929afR109 hardcoded constant string.

plugins/services/src/js/components/ServiceStatusProgressBar.js
ahoskins: I've been suggested this in PRs recently. And I've been convinced that it's the simpler of the two options because it makes it very clear that each block is completely independent.

plugins/services/src/js/components/ServiceStatusProgressBar.js
MatApple: It's OK to compare objects like this `status === ServiceStatus.DELETING` without digging down to the leaves of such an object tree to compare a string.

plugins/services/src/js/components/ServiceStatusProgressBar.js
MatApple: To @ahoskins point - you don't need the ` else if (` here because if it fell into the `if` statement above it, it would return anyway. IMO it's much cleaner to read multiple, separate `if` statements than a series of `if else` statements.

plugins/services/src/js/components/ServiceStatusProgressBar.js
MatApple: I left a comment about this over here https://github.com/dcos/dcos-ui/pull/2475#discussion_r148685344

plugins/services/src/js/components/ServiceStatusIcon.js
```diff
@@ -49,41 +67,45 @@ class ServiceStatusWarning extends Component {
...
-    let icon = <Icon color="red" id="yield" size="mini" />;
+    const { service } = this.props;
+    let icon = <StatusIcon state="WARNING" />;
 
-    if (this.props.item instanceof Service) {
-      const servicePath = encodeURIComponent(this.props.item.getId());
+    if (service instanceof Service) {
+      const servicePath = encodeURIComponent(service.getId());
+      content = getTooltipContent(service, content);
+      console.log(content);
```
MatApple: `console.log` :)

plugins/services/src/js/containers/services/ServicesTable.js
```diff
@@ -353,43 +351,18 @@ class ServicesTable extends React.Component {
   }
 
   renderStatus(prop, service) {
```
MatApple: This method is so much nicer now!

plugins/services/src/js/components/ServiceStatusIcon.js
weblancaster: ah good catch.

src/js/components/ProgressBar.js
```diff
@@ -79,22 +79,22 @@ class ProgressBar extends React.Component {
     stealPortion(barSizes, indexesLessThanThreshold, unassignedPortion);
 
     return data.map(function(status, index) {
-      let { className, style = {} } = status;
+      let { state, style = {} } = status;
```
ahoskins: I never like shadowing variables, and seems especially nasty when it's called `state` in a react component. I'd suggest a rename.

src/js/components/StatusIcon.js
```diff
@@ -0,0 +1,56 @@
...
+        color={iconState.color}
+        id={iconState.id}
+        size="mini"
+      />
+    );
+  }
+}
+
+StatusIcon.propTypes = {
+  state: PropTypes.oneOfType(["RUNNING", "TRANSITION", "STOPPED", "WARNING"])
```
ahoskins: Should be [oneOf](https://reactjs.org/docs/typechecking-with-proptypes.html) this is throwing an error in the console right now.

src/js/components/StatusIcon.js
```diff
@@ -0,0 +1,56 @@
...
+ * Status Icon is meant to be used across
+ * the application where it need to display any of the state below
+ * All green (RUNNING), transition (DEPLOYING, RECOVERING), Warning
+ *
+ * @class StatusIcon
+ * @extends {Component}
+ */
+class StatusIcon extends Component {
+  render() {
+    const state = this.props.state;
```
ahoskins: `this.props.state` 😕 I'm not a fan of overloading this word.

plugins/services/src/js/components/ServiceStatusProgressBar.js
Poltergeist: Let me add an example where an else if is needed or better used:

```JS
if (x > 1000) {
  ...
} else if (x > 100) {
  ...
} else if (x > 10) {
  ... 
}

return y;
```
In that case only one block will be executed but all blocks do have the same return that is a good example for a else if structure. Because each condition is simpler to read then x is bigger then z and smaller then z. But all conditions are connected to each other. 

But if your condition where `x === 3` ,`x  === 2`, `x === 1` you would most certainly not write your condition like `x === 3 && x !== 2 && x !== 1` so in that case it makes more sense to actually split up the if blocks as there are not related to each other. Each condition stands and works for it self. 

In this case although a switch would also work very fine:
```JS
switch(serviceStatus) {
  case ServiceStatus.RUNNING.displayName:
  case ServiceStatus.STOPPED.displayName:
    return null;
  case ServiceStatus.DELETING.displayName:
    ...
  default:
    ...
}
```
But I do think we are trying not to use switch blocks in dcos-ui for the reason that they are difficult to mental parse them. But I could be mistaken.

plugins/services/src/js/filters/__tests__/ServiceAttributeIsFilter-test.js
```diff
@@ -73,9 +73,9 @@ describe("ServiceAttributeIsFilter", function() {
     ]);
   });
 
-  it("Should correctly keep services in suspended state", function() {
+  it("Should correctly keep services in stopped state", function() {
```
Poltergeist: ❣️ If we adjust the message anyways let's remove the `Should` `it("keeps services in stopped state"), function() {`

src/js/components/ProgressBar.js
Poltergeist: ❣️ I do suggest `statusClassName`

plugins/services/src/js/components/HealthBar.js
```diff
@@ -1,121 +0,0 @@
-import classNames from "classnames";
-import React from "react";
-import { Tooltip } from "reactjs-components";
-
-import ProgressBar from "#SRC/js/components/ProgressBar";
-import StringUtil from "#SRC/js/utils/StringUtil";
-
-import HealthBarStates from "../constants/HealthBarStates";
-
-class HealthBar extends React.Component {
```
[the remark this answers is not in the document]
weblancaster: @Poltergeist please take a look at this commit https://github.com/dcos/dcos-ui/pull/2475/commits/4b87367e0f9f4f555c795941fab4b941c1f309b6

This was renamed to `ServiceStatusProgressBar` and was refactored so because git sees this as deleting and adding that's what we get for both files

`HealthBar.js` deleted
`ServiceStatusProgressBar.js` added

src/js/components/ProgressBar.js
weblancaster: @MatApple and I had decided to move to `state` instead of `className`.

I'm indifferent changing to `statusClassName`. Please let me know the verdict.

plugins/services/src/js/components/ServiceStatusIcon.js
```diff
@@ -3,22 +3,40 @@ import { Link } from "react-router";
...
 
-class ServiceStatusWarning extends Component {
-  getDeclinedOffersWarning(item) {
-    if (DeclinedOffersUtil.shouldDisplayDeclinedOffersWarning(item)) {
+const getTooltipContent = (service, content) => {
+  const servicePath = encodeURIComponent(service.getId());
+
+  return (
+    <span>
+      {content}
```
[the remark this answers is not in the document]
weblancaster: Agree. I have modified.

plugins/services/src/js/components/ServiceStatusIcon.js
```diff
@@ -49,41 +67,45 @@ class ServiceStatusWarning extends Component {
...
+        `DC/OS has been unable to complete this deployment for ${DateUtil.getDuration(Date.now() - DateUtil.strToMs(service.getQueue().since), null)}.`
       );
     }
 
     return null;
   }
 
-  shouldShowUnableToLaunchWarning(item) {
-    const queue = item.getQueue();
+  shouldShowUnableToLaunchWarning(service) {
```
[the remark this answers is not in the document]
weblancaster: that naming convention was already part of the code.. but I totally agree so I'm changing. thanks

plugins/services/src/js/filters/__tests__/ServiceAttributeIsFilter-test.js
weblancaster: yeah I thought about removing the `should` but if I do I need to removed all others `should` too to keep it consistent and since I'm trying to have as least loc changed I decided to keep.

if more people agree in removing all `should` in this file I'm in otherwise I wouldn't remove only for this guy.

@MatApple @ahoskins @bstavroulakis ?

src/img/components/icons/icon-system/spinner.svg
```diff
@@ -0,0 +1,33 @@
...
+  <rect x="43" y="2" rx="129" ry="6" width="14" height="28" fill="#1B2029">
+    <animate attributeName="opacity" values="1;0" times="0;1" dur="0.7s" begin="-0.6124999999999999s" repeatCount="indefinite"></animate>
```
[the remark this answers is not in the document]
weblancaster: @Poltergeist the css animations did not work since we are rendering the svg as shadow DOM and even trying to add scope css for some reason did not get parsed/used.

If you have a working suggestion following the same design required by @leemunroe please suggest

src/js/components/StatusIcon.js
weblancaster: Agree @ahoskins I'm indifferent using a separated file as constants or same file constants.

I could use the same argument though for many other cases where we hardcoded strings in our code base and potentially could be a separated constant file.

I believe @orlandohohmeier will have feedback on this case.

src/js/components/StatusIcon.js
weblancaster: what is your suggestion @ahoskins ?

src/js/components/StatusIcon.js
weblancaster: thanks

src/js/components/StatusIcon.js
weblancaster: Actually, @Poltergeist brought a great point and I'm moving it to constants. https://github.com/dcos/dcos-ui/pull/2475/files#r148768468

src/js/components/StatusIcon.js
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * Status Icon is meant to be used across
+ * the application where it need to display any of the state below
+ * All green (RUNNING), transition (DEPLOYING, RECOVERING), Warning
+ *
+ * @class StatusIcon
+ * @extends {Component}
+ */
+class StatusIcon extends Component {
```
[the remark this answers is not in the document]
weblancaster: The purpose of this component is to isolate the state/style and logic shaping the way we use only what is necessary, no more no less. That way keeping the style and logic consistent in all cases.

That being said you did bring a great point and suggestion to use a constant where I can accomplish the same goal. thanks.

Now, I do not want this to be specific to service status because these icons states will be used in other places.
The naming although is similar to serviceStatus I found that naming the state following a act rather than the color is a better approach.

I would change for something else but that was the best option at the time.

plugins/services/src/js/filters/__tests__/ServiceAttributeIsFilter-test.js
ahoskins: Whenever I write new tests, I try to not write `should` because I've been told not to many times.

plugins/services/src/js/components/ServiceStatusProgressBar.js
weblancaster: Now I understood what you meant @ahoskins

I thought you were talking about using `RUNNING || STOPPED` instead `ServiceStatus.RUNNING.displayName || ServiceStatus.STOPPED.displayName`

Next time will be easier to just say `no need to use "else if"`, I'm also an advocate for clean if only perhaps switch case statements in other cases, but you will see me advocating that on previous PRs.

Thanks for clarifying @MatApple @Poltergeist

src/js/components/StatusIcon.js
ahoskins: As @Poltergeist suggested below, what about `statusClassname`?

plugins/services/src/js/components/ServiceStatusIcon.js
```diff
@@ -49,41 +67,45 @@ class ServiceStatusWarning extends Component {
...
     }
 
-    return <noscript />;
+    return null;
   }
 
   getTooltip(content) {
-    let icon = <Icon color="red" id="yield" size="mini" />;
+    const { service } = this.props;
+    let icon = <StatusIcon state="WARNING" />;
```
ahoskins: It would be nice if "WARNING" came from a constants file so it was immediately clear what possible values can be passed into here.

plugins/services/src/js/components/ServiceStatusIcon.js
weblancaster: I have modified to use explicit conditions. thanks

plugins/services/src/js/filters/__tests__/ServiceAttributeIsFilter-test.js
weblancaster: Yes, I agree @ahoskins and I despise using `should` but the point here is SHOULD I remove **ALL** the `should`s?

plugins/services/src/js/components/ServiceStatusIcon.js
weblancaster: This is basically a duplication of:
https://github.com/dcos/dcos-ui/pull/2475#discussion_r148068634
https://github.com/dcos/dcos-ui/pull/2475#discussion_r148768468

Thanks for the reminder though.

plugins/services/src/js/filters/__tests__/ServiceAttributeIsFilter-test.js
ahoskins: ALL!

weblancaster: Well in that case based on both, we would have to also remove the should from this file https://github.com/dcos/dcos-ui/pull/2475/files#diff-06dacf8f6b57206ec2e005a8ebce14a3

orlandohohmeier: @weblancaster I value consistency but don't think you should change all test cases. Let's not change the scope of this PR.

weblancaster: @orlandohohmeier I agree I will keep the `should` for all for now.

src/js/constants/StatusIcon.js
```diff
@@ -0,0 +1,20 @@
+const STATUS_ICON = {
```
weblancaster: Even though a couple of naming here (STOPPED, WARNING) might have a coincidence with service status, this is not to match what we have there but the name was given because of the action of the icon.

plugins/services/src/js/components/ServiceStatusIcon.js
```diff
@@ -118,32 +117,48 @@ class ServiceStatusIcon extends Component {
...
-    if (
-      displayName === ServiceStatusLabels.DEPLOYING ||
-      displayName === ServiceStatusLabels.RECOVERING ||
-      displayName === ServiceStatusLabels.DELETING
-    ) {
-      iconState = "TRANSITION";
+    const serviceStatus = service.getServiceStatus();
+    let iconState;
+
+    switch (serviceStatus) {
```
weblancaster: I'm wondering if I should `if` statements instead that way I can combine the 3 `LOADING` checks in one if and save a couple lines.

plugins/services/src/js/filters/__tests__/ServiceAttributeIsFilter-test.js
orlandohohmeier: @weblancaster what ever you prefer. :)

weblancaster: @orlandohohmeier so I decided to open up a separated PR removing the `should` from this file and some others. https://github.com/dcos/dcos-ui/pull/2497

orlandohohmeier: @weblancaster awesome! 🙌

plugins/services/src/js/components/ServiceStatusIcon.js
MatApple: @weblancaster I think that makes sense!

MatApple: But I'm also good with it as is :)

bstavroulakis: I'm also good as it is now

weblancaster:
I have pushed a new commit addressing some feedback from Lee regarding the responsive part and other minor changes

plugins/services/src/js/components/ServiceStatusProgressBar.js
```diff
@@ -0,0 +1,80 @@
...
+          ]}
+          total={instancesCount}
+        />
+      );
+    }
+
+    return (
+      <Tooltip interactive={true} content={this.getTooltipContent()}>
+        <ProgressBar
+          className="status-bar--large staged"
```
MatApple: Did you mean to apply the `staged` class here?

src/styles/components/status-bar/styles.less
```diff
@@ -36,6 +51,10 @@
     text-align: left;
     width: @base-spacing-unit * 2;
 
+    &.staged {
+      .staged;
```
MatApple: @weblancaster I'm not sure what happens here - can you elaborate on what's happening for such a declaration?

src/styles/components/status-bar/styles.less
```diff
@@ -57,34 +76,32 @@
...
       }
 
-      &.color-6 {
+      &.color-6,
+      &.warning {
         background-color: @yellow;
       }
 
+      &.disabled {
+        background-color: #adafb4;
```
MatApple: Please use an existing variable for this!

src/img/components/icons/icon-system/spinner.svg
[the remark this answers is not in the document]
MatApple: @weblancaster if you haven't tested this animation across the browsers we support, we certainly should!

plugins/services/src/js/components/ServiceStatusIcon.js
```diff
@@ -0,0 +1,172 @@
...
+    if (DeclinedOffersUtil.displayDeclinedOffersWarning(service)) {
+      return this.getDeclinedOffersWarning(service);
+    }
+
+    // If not, we try to display an unable to launch warning, then default to nothing.
+    if (this.isUnableToLaunch(service)) {
+      return this.getUnableToLaunchWarning(service);
+    }
+
+    return <Icon {...iconState} size="mini" />;
```
ahoskins: In the current cluster, sometimes the `iconState` doesn't get set and so undefined is passed into the required prop `id` in `Icon`. See screenshot below:

screen shot 2017-11-07 at 4 28 56 pm

ahoskins: And this is because there is no `if` for the N/A service status. Also no `if` path for "Waiting" or "Delayed" status. Is this meant to be like this?

plugins/services/src/js/components/ServiceStatusProgressBar.js
weblancaster: correct, since it's "default" now for the bar.

plugins/services/src/js/components/ServiceStatusIcon.js
weblancaster: @ahoskins correct Lee and I are still figuring out what to do with the N/A case.

Now for `Waiting ` and `Delayed ` that shouldn't be the case since it's now `deploying` instead.

weblancaster: For the NA case I'm adding a condition to catch any other case that isn't defined before.

src/styles/components/status-bar/styles.less
weblancaster: It's basically a `Object.assign({...}, {...})` therefore extending the class `.staged` under `&.staged` because now I have the staged at the wrapper level and the child level (before was only child) so I can re-use the properties/values.

src/img/components/icons/icon-system/spinner.svg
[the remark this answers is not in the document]
weblancaster: the animation works fine in all browsers but IE edge but I wouldn't block the PR because of that.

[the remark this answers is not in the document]
MatApple: @weblancaster let's move forward with the PR. I've created an issue for us to revisit the spinner https://jira.mesosphere.com/browse/DCOS-19469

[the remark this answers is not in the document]
weblancaster: ok thanks!

plugins/services/src/js/components/ServiceStatusIcon.js
MatApple: @weblancaster do we need to capture that in a JIRA or have you already started the work?

weblancaster: @MatApple talking to @leemunroe seems like we do need a more long-term solution for `why do we have NA? and should we keep it?` but for now, to not block this PR I have added a check for all cases that don't match the states for icons and display no icon in that case.

weblancaster:
alfred retest this please

weblancaster:
Awesome! Thanks @MatApple

plugins/services/src/js/components/HealthBar.js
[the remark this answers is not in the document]
Poltergeist: Okay that makes sense now. I didn't drew the connection, that it actually was a rename.

Poltergeist:
Great Work on this PR @weblancaster. Thanks for all the extra effort.

PR #2477, opened 2017-10-26
Quickie: feat(Util.js): add isArray check

nLight:
During my MesosStream refactoring I stumbled upon the fact that we're just assume that we're always getting an array here, which might not always be true, so I decided adding `isArray` check.

### DIFF
--- a/src/js/utils/__tests__/Util-test.js
+++ b/src/js/utils/__tests__/Util-test.js
@@ -69,25 +69,53 @@ describe("Util", function() {
   });
 
   describe("#last", function() {
-    it("should return the last element in an array", function() {
-      var array = [0, 1, 2, 3];
-      var last = Util.last(array);
+    describe("with incorrect input", function() {
+      it("returns null for objects", function() {
+        var last = Util.last({});
 
-      expect(last).toEqual(3);
+        expect(last).toEqual(null);
+      });
+
+      it("returns null for strings", function() {
+        var last = Util.last("bla");
+
+        expect(last).toEqual(null);
+      });
+
+      it("returns null for Numbers", function() {
+        var last = Util.last(NaN);
+
+        expect(last).toEqual(null);
+      });
+
+      it("returns null for undefined", function() {
+        var last = Util.last(undefined);
+
+        expect(last).toEqual(null);
+      });
     });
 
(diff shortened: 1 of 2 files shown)


src/js/utils/__tests__/Util-test.js
```diff
@@ -69,6 +69,14 @@ describe("Util", function() {
   });
 
   describe("#last", function() {
+    describe("with not an array as an input", function() {
+      it("return null", function() {
```
orlandohohmeier: 🔤 "returns"

src/js/utils/__tests__/Util-test.js
```diff
@@ -69,6 +69,14 @@ describe("Util", function() {
   });
 
   describe("#last", function() {
+    describe("with not an array as an input", function() {
```
orlandohohmeier: 🤔 do we really need this describe block?

nLight: According to the better spec one should wrap input conditions in a block like:
- with valid input
- without `foo` argument
etc.

src/js/utils/__tests__/Util-test.js
nLight: ✅

src/js/utils/__tests__/Util-test.js
orlandohohmeier: Wouldn't that mean that you should move the other test into a "with valid input" block and the new one into an "with invalid input" block? You probably should also add at leas another case using an object as input.

nLight: Yes, but that's gonna be a refactoring already.

nLight: Shall I also add a string and a number? :)

orlandohohmeier: Yes, lets add a number and string as well

nLight: :sad_troll:

nLight: ✅

nLight:
Comments have been addressed 🙇

orlandohohmeier:
L👀ks good!

PR #2479, opened 2017-10-27
DCOS-19266: Disable delete group functionality fully

bstavroulakis:
Disable the ability to delete a group that contains services for every case.

Closes DCOS-19266

**Checklist**

### DIFF
--- a/plugins/services/src/js/components/modals/ServiceDestroyModal.js
+++ b/plugins/services/src/js/components/modals/ServiceDestroyModal.js
@@ -20,7 +20,6 @@ const METHODS_TO_BIND = [
   "handleChangeInputFieldDestroy",
   "handleModalClose",
   "handleRightButtonClick",
-  "handleChangeInputGroupDestroy",
   "handleFormSubmit"
 ];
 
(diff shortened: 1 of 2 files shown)


ahoskins:
Does this not need tests?

bstavroulakis:
@ahoskins There weren't any tests for the group delete. So since I removed it altogether I suppose we won't need tests for this.

plugins/services/src/js/components/modals/ServiceDestroyModal.js
```diff
@@ -82,14 +80,7 @@ class ServiceDestroyModal extends React.Component {
   }
 
   shouldForceUpdate() {
```
MatApple: We no longer need any force related commands/checks as it only applies when nested services are in deploying state. We can remove this method.

bstavroulakis: Couldn't a single service/pod be in deploying state? I'm pretty sure I needed this once with a pod

MatApple: Oh my mistake - you are correct. I thought this was a group delete modal.

PR #2480, opened 2017-10-27
Cluster Overview page MesosStateStore refactoring

nLight:
This PR replaces MesosStateStore with direct calls to the V1 API. The API lacks `elapsed_time` and `start_time`, there's a ticket for that in mesos.

Closes DCOS-19145

### DIFF
--- a/src/js/pages/system/OverviewDetailTab.js
+++ b/src/js/pages/system/OverviewDetailTab.js
@@ -7,6 +7,7 @@ import React from "react";
 import { StoreMixin } from "mesosphere-shared-reactjs";
 import { FormattedMessage } from "react-intl";
 import moment from "moment";
+import { request } from "@dcos/mesos-client";
 
 import Breadcrumb from "../../components/Breadcrumb";
 import BreadcrumbTextContent from "../../components/BreadcrumbTextContent";
@@ -24,7 +25,6 @@ import Loader from "../../components/Loader";
 import MarathonStore
   from "../../../../plugins/services/src/js/stores/MarathonStore";
 import MetadataStore from "../../stores/MetadataStore";
-import MesosStateStore from "../../stores/MesosStateStore";
 import Page from "../../components/Page";
 import VersionsModal from "../../components/modals/VersionsModal";
 import { isEmpty } from "../../utils/ValidatorUtil";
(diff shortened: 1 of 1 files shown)


src/js/pages/system/OverviewDetailTab.js
```diff
@@ -220,14 +250,29 @@ class OverviewDetailTab extends mixin(StoreMixin) {
...
+    const mesosMasterInfo = masterInfo || this.getLoading();
+    const mesosCluster = cluster || this.getLoading();
+    const mesosVersion = version || this.getLoading();
+    const mesosBuilt = buildTime
+      ? moment(buildTime * 1000).fromNow()
+      : this.getLoading();
+    const mesosStarted = startTime
+      ? moment(startTime * 1000).fromNow()
+      : this.getLoading();
+    const mesosElected = electedTime
```
bstavroulakis: When I ran this the started and elected time were endlessly loading. The info.start_time was null and I see that the electedTime is not set somewhere. Is this intended?

image

nLight: Yep, as I pointed in the description, V1 API lacks elapsed_time and start_time, there's a ticket for that in mesos.

src/js/pages/system/OverviewDetailTab.js
```diff
@@ -66,14 +71,39 @@ class OverviewDetailTab extends mixin(StoreMixin) {
...
         events: ["dcosBuildInfoChange", "dcosSuccess", "success"]
-      },
-      {
-        name: "state",
-        events: ["success"],
-        listenAlways: false
       }
     ];
 
+    request({ type: "GET_FLAGS" }).subscribe(message => {
```
MatApple: @nLight wondering if we should have some error handling/retry here? Looks like we don't have any for the `mesos-client` `request` and `subscribe` methods.

Poltergeist:
**Conflicting files**
npm-shrinkwrap.json
package.json

Please rebase.

src/js/pages/system/OverviewDetailTab.js
nLight: Do you think we should add like `retry(3)` just to be on a safe side?

nLight:
@bstavroulakis @MatApple @Poltergeist I rebased the branch, please re-review 🙇

src/js/pages/system/OverviewDetailTab.js
MatApple: I think we should and we should really show an error message after that. We need to architect our request error handling a little e.g. how we deal with 401, 403 vs 500's. A retry on 403 is not smart. cc/ @orlandohohmeier

MatApple: BTW - I don't think it should block this PR.

nLight: 👍  https://jira.mesosphere.com/browse/DCOS-19443

PR #2481, opened 2017-10-27
refactor(ConfigurationMapEditAction): delete this component

ahoskins:
**Summary**:
Delete this component because it does nothing useful, and just adds complexity. This component was looked at because action buttons will be added to the licensing information rows. We decided that rather than modify `ConfigurationMapEditAction` to work with the licensing rows, it is best to just remove this component.

**In this PR:**
1) Delete the `ConfigurationMapEditAction` component
2) Update all usages in the codebase to use the anchor directly

**Testing**: 
1) Checkout this branch and [private plugins branch](https://github.com/mesosphere/dcos-ui-plugins-private/pull/644)
2) No visual or DOM structure changes anywhere, and there is no expected regressions

Closes DCOS-19339.

**Checklist**


plugins/services/src/js/components/ConfigurationMapAction.js
```diff
@@ -0,0 +1,21 @@
+import React from "react";
+import classNames from "classnames";
+
+const ConfigurationMapAction = ({ isHover, onClick, children }) => {
+  const classes = classNames("button button-link flush", {
+    "table-display-on-row-hover": isHover
+  });
+
+  return (
+    <a className={classes} onClick={onClick}>
```
MatApple: @ahoskins I have a few questions because this component is effectively a <Button/> without all the props available on a <Button/> and with an isHover. This makes me curious about it's necessity. Are their any drawbacks or blockers for removing this entirely and just rendering the `children` directly?
e.g. instead of:
```
 <ConfigurationMapAction
    onClick={onEditClick.bind(this, tabViewID)}
    isHover={true}
  >
    Edit
  </ConfigurationMapAction>
```
we just do this:
```
  <button
    onClick={onEditClick.bind(this, tabViewID)}
    classNames="button button-link flush table-display-on-row-hover"
  >
    Edit
  </button>
```

ahoskins: I think since this component is always used beside the other components for configuration map, it should also be a component. ex:

```
<ConfigurationMapRow>
  <ConfigurationMapLabel>...
  <ConfigurationMapValue>...
  <ConfigurationMapAction> ...
</ConfigurationMapRow>
```

Agree?

But I do think that if a component, this should be able to render either a `<button>` or an `<a>`. There will be some usages when we want to pass it an `onClick` to execute some javascript, but other cases when we give it an `href`. For example in the node licensing rows, one of actions will be opening a link "Upgrade" that directs to a support.mesosphere.com.

Maybe this is too complex, and it's easier to just remove the component and render HTML tags directly. Thoughts @dcos/frontend-foundation

ahoskins:
I've changed this component to render either an `<a>` or `<button>` depending on whether the `href` prop is provided. For example:

```
<ConfigurationMapAction
     onClick={onEditClick.bind(this, tabViewID)}
     isHover={true}
>
  Edit
</ConfigurationMapAction>
``` 

Since the above doesn't give an href, it will be rendered as a button:

```
<button onClick={onEditClick} class="button button-link flush table-display-on-row-hover">
    Edit
</button>
```

But in the below example an href is provided:

```
<ConfigurationMapAction href="https://support.mesosphere.com" >
  Upgrade
</ConfigurationMapAction>
``` 

And it will be rendered as follows:

```
<a href="https://support.mesosphere.com" class="button button-link flush">
    Upgrade
</a>
```

More information on this guideline for anchors vs buttons at [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#Accessibility_recommendations).

juliangieseke:
After being generalized and used outside of the services plugin, shouldn't we move `ConfigurationMapAction` to `/src/js/components/`?

ahoskins:
@juliangieseke good question, I'm not sure. 

@orlandohohmeier @nLight what are your thoughts on my above comment? Another option is to remove this component entirely because it isn't doing anything special? (all it's doing is applying the `button button-link` class to an anchor or button).

ahoskins:
I have deleted this component because after discussions with @MatApple and @orlandohohmeier, we agree that this component does nothing useful and adds complexity.

juliangieseke:
Good solution! LGTM 🚢

PR #2483, opened 2017-10-31
feat(ToggleValue): add new toggle value component

ahoskins:
**Description**:
Add a `ToggleValue` component. It has an API like this:

```
<ToggleValue 
    primaryValue="5 days overdue" 
    secondaryValue="Feb 20, 2017" 
/>
```

This component will initially be used for the node licensing expiry row. In the future, it's intended to be rolled out as a general UX pattern when displaying dates/times anywhere in DC/OS UI (currently Marathon UI has this). 

**Testing**:
1. Checkout this branch.
2. In private plugins, check out the branch in this [PR](https://github.com/mesosphere/dcos-ui-plugins-private/pull/648).
3. Navigate to the `/overview` page and see the toggle on the row called "License Expiry"

**Checklist**


src/js/components/__tests__/ConfigurationMapToggleValue-test.js
```diff
@@ -0,0 +1,28 @@
+const React = require("react");
+const ReactDOM = require("react-dom");
+const TestUtils = require("react-addons-test-utils");
+
+const ConfigurationMapToggleValue = require("../ConfigurationMapToggleValue");
+
+describe("ConfigurationMapToggleValue", function() {
+  it("should toggle the text value when clicked", function() {
```
Poltergeist: Omit `should` here.

src/js/components/__tests__/ConfigurationMapToggleValue-test.js
```diff
@@ -0,0 +1,28 @@
...
+        secondaryValue="secondary"
+      />,
+      container
+    );
+    const content = TestUtils.findRenderedDOMComponentWithClass(
+      instance,
+      "configuration-map-toggle-value"
+    );
+
+    expect(content.innerHTML).toContain("primary");
```
Poltergeist: This could be it's own test case. Like `it shows initially the primary value`. And then add a second case for the secondary value. It is usually better to only have a single expect per test case. So if the expect fails the developer directly knows what failed without looking into the test.

ahoskins: Good suggestion, thanks, will change.

ahoskins:
@nLight is this the proper location in the codebase for a component like this? Or should this be put into a folder in `foundation-ui`?

nLight:
@ahoskins as we don't really have a strategy moving forward with components, I'd say it's ok to have it in `/components`

PR #2484, opened 2017-10-31
docs(translation): remove framework cleaner instructions for 1.10

sascala:
The Delete an SDK message in 1.10 tells users to use the framework cleaner script to clean up persistent state. This is no longer necessary in 1.10.

**Checklist**


bstavroulakis:
We need to remove the text from the code as well and not only the translation area because then we'll have something like this.

image

bstavroulakis:
I removed the parts where this was used in 1.10

PR #2486, opened 2017-10-31
fix(FormGroupContainer): only apply hover styles if onClick provided

ahoskins:
This is fixing a regression from 5e2d832 which caused hover on all `FormGroupContainers`. This PR changes it to only apply the outline and cursor pointer when an `onClick` prop is provided.

Currently, the only time an `onClick` prop is provided to `FormGroupContainer` is in the container pods when creating a service. See below the hover classes on the container, but not networking or volumes pods.
screen recording 2017-10-31 at 02 47 pm

@leemunroe can finally stop bugging me about this one now 💯 😄 

Closes DCOS-18317.

**Checklist**

### DIFF
--- a/src/js/components/form/FormGroupContainer.js
+++ b/src/js/components/form/FormGroupContainer.js
@@ -1,5 +1,6 @@
 import React from "react";
 import { Tooltip } from "reactjs-components";
+import classNames from "classnames";
 
 import Icon from "../Icon";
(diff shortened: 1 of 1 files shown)


leemunroe:
Beautiful 🌸

orlandohohmeier:
LGTM 💯 ✨

PR #2492, opened 2017-11-03
chore(TasksTable-cy): remove only to run all tests

nLight:
Remove overlooked `only` function to run all the integration tests

### DIFF
--- a/tests/pages/catalog/packages/package/PackageTab-cy.js
+++ b/tests/pages/catalog/packages/package/PackageTab-cy.js
@@ -132,13 +132,15 @@ describe("Package Detail Tab", function() {
         .should("to.have.length", 1);
     });
 
-    it("closes the framework config when cancel button clicked", function() {
-      cy.get(".modal .modal-header button").contains("Cancel").click();
+    it("goes to form when back button is clicked", function() {
+      cy.get(".modal .modal-header button").contains("Back").click();
 
-      cy.get(".page-body-content h1").should("contain", "marathon");
+      cy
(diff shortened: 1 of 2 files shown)


nLight:
Two tests are failing 😩

nLight:
@ahoskins would you please have a look if the failing test is a regression?
`PackageTab-cy.js` => `closes the framework config when cancel button clicked`.

Looks like there was a cancel button that would just close the form, now there's always a Back button. Maybe we could just delete that test.

The second one that is failing seems to be a flaky one

nLight:
Alfred test this flaky one please

ahoskins:
Oops, yes this is an old test that should fail. I will tweak and push to this branch.

tests/pages/catalog/packages/package/PackageTab-cy.js
```diff
@@ -132,13 +132,15 @@ describe("Package Detail Tab", function() {
...
 
-    it("closes the framework config when cancel button clicked", function() {
-      cy.get(".modal .modal-header button").contains("Cancel").click();
+    it("goes to form when back button is clicked", function() {
+      cy.get(".modal .modal-header button").contains("Back").click();
 
-      cy.get(".page-body-content h1").should("contain", "marathon");
+      cy
+        .get('.modal .menu-tabbed-container input[name="group"]')
+        .type(`{selectall}group-1`);
```
juliangieseke: This one is missing a `should`-call, we probably should (ha!) always have that check in our tests instead of relying on cypress failing somehow.

PR #2494, opened 2017-11-03
fix(Licensing): split mount-point into one for each row

juliangieseke:
This fixup splits up the mount point into two to prevent ReactUtil from creating
an additional wrapper - which breaks the styling.

Closes DCOS-19398

### DIFF
--- a/src/js/pages/system/OverviewDetailTab.js
+++ b/src/js/pages/system/OverviewDetailTab.js
@@ -145,11 +145,11 @@ class OverviewDetailTab extends mixin(StoreMixin) {
           </ConfigurationMapValue>
         </ConfigurationMapRow>
         <MountService.Mount
-          type="OverviewDetailTab:AdditionalGeneralDetails:Licensing:NodeCapacity"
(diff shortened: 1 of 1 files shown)


src/js/pages/system/OverviewDetailTab.js
```diff
@@ -145,8 +145,12 @@ class OverviewDetailTab extends mixin(StoreMixin) {
           </ConfigurationMapValue>
         </ConfigurationMapRow>
         <MountService.Mount
-          type="OverviewDetailTab:AdditionalGeneralDetails"
-          limit={2}
+          type="OverviewDetailTab:AdditionalGeneralDetails:Licensing:NodeCapacity"
```
ahoskins: We tend to name mounted private plugins very generically, without specifically mentioning the enterprise feature in the code. For example, see the naming of mounted components in [secrets private plugin](https://github.com/mesosphere/dcos-ui-plugins-private/blob/master/secrets/hooks.js#L61)

ahoskins: I pushed a more generic rename to this branch 😄

PR #2500, opened 2017-11-07
fix(PodGeneralConfigSection): fix onEditClick being undefined

ahoskins:
Add conditionals to only render the actions if onEditClick is not undefined. **Now there should be nowhere with `onEditClick.bind()` without first checking that it's not an undefined reference.**

**Checklist**


bstavroulakis:
Should we add the `if (onEditClick) {` check in the other files that use it as well (ServiceStorageConfigSection.js, ...)?

ahoskins:
The other files had `if` checks already. If you grep the codebase for `onEditClick.bind`, there should be nowhere without an `if` check. @bstavroulakis

PR #2501, opened 2017-11-07
fix(validate-tests): throw error and exit 1 when regex match

ahoskins:
This will ensure that ".only" for integration tests, or "fit" for unit tests don't make it into our codebase.

**Checklist**

### DIFF
--- a/scripts/validate-tests
+++ b/scripts/validate-tests
@@ -7,7 +7,7 @@ source ${SCRIPT_PATH}/utils/git
 
 header "Validating tests..."
 
-! egrep --quiet -R "fdescribe\\(|fit\(" ./src && \
-! egrep --quiet -R "context\.only\(|it\.only\(" ./tests
+! egrep -R "fdescribe\\(|fit\(" ./src || exit 1 && \
+! egrep -R "context\.only\(|it\.only\(" ./tests || exit 1
 
 exit 0


scripts/validate-tests
```diff
@@ -7,7 +7,7 @@ source ${SCRIPT_PATH}/utils/git
 
 header "Validating tests..."
 
-! egrep --quiet -R "fdescribe\\(|fit\(" ./src && \
-! egrep --quiet -R "context\.only\(|it\.only\(" ./tests
+! egrep -R "fdescribe\\(|fit\(" ./src || exit 1 && \
+! egrep -R "context\.only\(|it\.only\(" ./tests || exit 1
```
bstavroulakis: @orlandohohmeier Why was the --quiet needed here?

MatApple: I'm also pretty sure that with `egrep` we don't need to escape the `(` like we do with `grep`.

orlandohohmeier: @bstavroulakis I have no idea why we decided to suppress the output and I think it's a good idea to not run the command in quite mode — though that's not the reason for the script failure.

@MatApple `egrep` is the same as `grep -E` and the search-pattern is regular expression thus we need to escape the parentheses. 

Side Note: `egrep` is actually only there for historic reasons and we should switch to `grep -E`

scripts/validate-tests
```diff
@@ -7,7 +7,7 @@ source ${SCRIPT_PATH}/utils/git
...
 
 exit 0
```
orlandohohmeier: We could change this to `exit $?` or remove the exit all-together instead of adding explicit exit statements to the respective `egrep` calls as this is the reason why the script failed to fail builds that included debug directives.

orlandohohmeier:
Thanks a lot for taking care and hot-fixing the script! Let's merge this fix and then rewrite the test validation so that we can run the test in our pre-commit hooks as well.

MatApple:
@orlandohohmeier just for further correctness  - `grep` and `egrep` both use regular expression. `egrep` or `grep -E` uses "extended regular expression".

orlandohohmeier:
@MatApple yep, that’s what I was trying to say. 😆

PR #2503, opened 2017-11-08
Add host compare method in RouterUtil

bstavroulakis:
Add host compare method in RouteUtil. This is needed for the cluster linking task.

**Checklist**

### DIFF
--- a/src/js/utils/RouterUtil.js
+++ b/src/js/utils/RouterUtil.js
@@ -189,15 +189,34 @@ const RouterUtil = {
   },
 
   /**
-   * Compair two urls if they have the same host name
+   * Get URL full path. If it is a full path url return, if not attach
+   * the current local origin.
+   *
+   * @param {String} url The url
+   * @returns {String} - The url with the full path
+   */
+  getFullPath(url) {
+    if (!/http:\/\/|https:\/\//.test(url)) {
+      url = `${global.location.origin}${url.replace(/^(\/|)/g, "/")}`;
+    }
+
+    return url;
+  },
+
+  /**
+   * Compares two urls if they have the same host name
    *
    * @param {String} urlA The first url
    * @param {String} urlB The second url
    * @returns {Boolean} - True if they have the same host, false if they don't
    */
   hasSameHost(urlA, urlB) {
-    const parsedUrlA = Util.parseUrl(urlA);
-    const parsedUrlB = Util.parseUrl(urlB);
+    if (typeof urlA !== "string" || typeof urlB !== "string") {
+      return false;
+    }
+
+    const parsedUrlA = Util.parseUrl(this.getFullPath(urlA));
+    const parsedUrlB = Util.parseUrl(this.getFullPath(urlB));
 
     if (!parsedUrlA || !parsedUrlB) {
       return false;
--- a/src/js/utils/RouterUtil.js
+++ b/src/js/utils/RouterUtil.js
@@ -186,6 +186,24 @@ const RouterUtil = {
       const blob = new Blob([data], { type: "application/json" });
       global.navigator.msSaveOrOpenBlob(blob, filename);
     }
+  },
+
+  /**
+   * Compair two urls if they have the same host name
+   *
+   * @param {String} urlA The first url
+   * @param {String} urlB The second url
+   * @returns {Boolean} - True if they have the same host, false if they don't
+   */
+  hasSameHost(urlA, urlB) {
+    const parsedUrlA = Util.parseUrl(urlA);
+    const parsedUrlB = Util.parseUrl(urlB);
+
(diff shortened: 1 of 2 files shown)


src/styles/components/modal/link-list/styles.less
```diff
@@ -0,0 +1,47 @@
+.modal-link-list {
```
ahoskins: Why are these styles in `dcos-ui`, but the modal in private plugins?

bstavroulakis: Removed

src/js/utils/RouterUtil.js
```diff
@@ -186,6 +186,24 @@ const RouterUtil = {
...
+   *
+   * @param {String} urlA The first url
+   * @param {String} urlB The second url
+   * @returns {Boolean} - True if they have the same host, false if they don't
+   */
+  hasSameHost(urlA, urlB) {
+    const parsedUrlA = Util.parseUrl(urlA);
+    const parsedUrlB = Util.parseUrl(urlB);
+
+    if (!parsedUrlA || !parsedUrlB) {
```
juliangieseke: this one doesnt work with relative URLs 🤔 

I'd expect these tests to pass 
```
    it("returns true for same relative paths", function() {
      const hasSameHost = RouterUtil.hasSameHost(
        "/some/path",
        "/some/path"
      );
      expect(hasSameHost).toEqual(true);
    });
    it("returns false for different relative paths", function() {
      const hasSameHost = RouterUtil.hasSameHost(
        "/some/path",
        "/some/other/path"
      );
      expect(hasSameHost).toEqual(false);
    });
    it("returns true for same relative and absolute paths", function() {
      const hasSameHost = RouterUtil.hasSameHost(
        CURRENTHOST+"/some/path",
        "/some/path"
      );
      expect(hasSameHost).toEqual(true);
    });
    it("returns true for same relative and absolute paths", function() {
      const hasSameHost = RouterUtil.hasSameHost(
        CURRENTHOST+"/some/path",
        "/some/other/path"
      );
      expect(hasSameHost).toEqual(false);
    });
```
(you have to mock CURRENTHOST)

src/js/utils/RouterUtil.js
```diff
@@ -186,6 +186,24 @@ const RouterUtil = {
       const blob = new Blob([data], { type: "application/json" });
       global.navigator.msSaveOrOpenBlob(blob, filename);
     }
+  },
+
+  /**
+   * Compair two urls if they have the same host name
```
juliangieseke: Compares*

src/js/utils/RouterUtil.js
bstavroulakis: I see what you're saying for relative URLs. Since this compares the host, any relative URL should probably return true as they have the same host?

src/js/utils/RouterUtil.js
bstavroulakis: Fixed

src/js/utils/RouterUtil.js
bstavroulakis: Now, for relative url it will attach the local origin and compare the local host

src/js/utils/RouterUtil.js
```diff
@@ -186,6 +186,43 @@ const RouterUtil = {
...
+  },
+
+  /**
+   * Get URL full path. If it is a full path url return, if not attach
+   * the current local origin.
+   *
+   * @param {String} url The url
+   * @returns {String} - The url with the full path
+   */
+  getFullPath(url) {
```
ahoskins: Why is this method needed? As far as I can tell, this is already handled by `Util.parseUrl` and I think all the tests will still pass without this method.

In parseUrl:
```
    const aElement = document.createElement("a");
    aElement.href = url;
```

bstavroulakis: They don't pass because for the relative path the host is null. It adds the url to the href but it doesn't mean that it will add a host to it or a full path.

bstavroulakis:
Alfred test this

src/js/utils/RouterUtil.js
```diff
@@ -186,6 +186,28 @@ const RouterUtil = {
...
+  },
+
+  /**
+   * Compares two urls if they have the same host name
+   *
+   * @param {String} urlA The first url
+   * @param {String} urlB The second url
+   * @returns {Boolean} - True if they have the same host, false if they don't
+   */
+  hasSameHost(urlA, urlB) {
```
nLight: Shall we start thinking of a `UrlUtil`? I bet `RouterUtil` has been introduced to collect the `react-router` utils. Would be grate to extract all the Url related functions from other Utils and put them together. What do you think?

PR #2516, opened 2017-11-17
DCOS-19557: Extract placement constraints form partial

ahoskins:
Edit @nlight:

This is a pure extraction. We want to reuse this partial in the SDK form on the later stage. Please verify that the Service create form still works as expected.

### DIFF
--- a/src/js/components/PlacementConstraintsPartial.js
+++ b/src/js/components/PlacementConstraintsPartial.js
@@ -1,0 +1,236 @@
+import React, { Component } from "react";
+import { Tooltip } from "reactjs-components";
+
+import { findNestedPropertyInObject } from "#SRC/js/utils/Util";
+import AddButton from "#SRC/js/components/form/AddButton";
+import DeleteRowButton from "#SRC/js/components/form/DeleteRowButton";
+import FieldAutofocus from "#SRC/js/components/form/FieldAutofocus";
+import FieldError from "#SRC/js/components/form/FieldError";
+import FieldHelp from "#SRC/js/components/form/FieldHelp";
+import FieldInput from "#SRC/js/components/form/FieldInput";
+import FieldLabel from "#SRC/js/components/form/FieldLabel";
+import FieldSelect from "#SRC/js/components/form/FieldSelect";
+import FormGroup from "#SRC/js/components/form/FormGroup";
+import FormGroupHeading from "#SRC/js/components/form/FormGroupHeading";
+import FormGroupHeadingContent
+  from "#SRC/js/components/form/FormGroupHeadingContent";
+import FormRow from "#SRC/js/components/form/FormRow";
+import Icon from "#SRC/js/components/Icon";
+
+import OperatorTypes from "#PLUGINS/services/src/js/constants/OperatorTypes";
+import PlacementConstraintsUtil
+  from "#PLUGINS/services/src/js/utils/PlacementConstraintsUtil";
+
+function placementConstraintLabel(name, tooltipText, options = {}) {
+  const { isRequired = false, linkText = "More information" } = options;
+
+  const tooltipContent = (
+    <span>
+      {`${tooltipText} `}
+      <a
+        href="https://mesosphere.github.io/marathon/docs/constraints.html"
+        target="_blank"
+      >
+        {linkText}
+      </a>.
+    </span>
+  );
+
+  return (
+    <FieldLabel>
+      <FormGroupHeading required={isRequired}>
+        <FormGroupHeadingContent primary={true}>
+          {name}
+        </FormGroupHeadingContent>
+        <FormGroupHeadingContent>
+          <Tooltip
+            content={tooltipContent}
+            interactive={true}
+            maxWidth={300}
+            wrapText={true}
+          >
+            <Icon color="grey" id="circle-question" size="mini" />
+          </Tooltip>
+        </FormGroupHeadingContent>
+      </FormGroupHeading>
+    </FieldLabel>
+  );
+}
+
+export default class PlacementSection extends Component {
+  getOperatorTypes() {
+    return Object.keys(OperatorTypes).map((type, index) => {
+      return <option key={index} value={type}>{type}</option>;
+    });
+  }
+
+  getPlacementConstraintsFields(data = []) {
+    const constraintsErrors = findNestedPropertyInObject(
+      this.props.errors,
+      "constraints"
+    );
+    const hasOneRequiredValue = data.some(function(constraint) {
+      return PlacementConstraintsUtil.requiresValue(constraint.operator);
+    });
+    const hideValueColumn = data.every(function(constraint) {
+      return PlacementConstraintsUtil.requiresEmptyValue(constraint.operator);
+    });
+
+    return data.map((constraint, index) => {
+      let fieldLabel = null;
+      let operatorLabel = null;
+      let valueLabel = null;
+      const valueIsRequired = PlacementConstraintsUtil.requiresValue(
+        constraint.operator
+      );
+      const valueIsRequiredEmpty = PlacementConstraintsUtil.requiresEmptyValue(
+        constraint.operator
+      );
+      const fieldNameError = findNestedPropertyInObject(
+        constraintsErrors,
+        `${index}.fieldName`
+      );
+      const operatorError = findNestedPropertyInObject(
+        constraintsErrors,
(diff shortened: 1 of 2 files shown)


src/js/components/SchemaField.js
```diff
@@ -277,6 +334,10 @@ class SchemaField extends Component {
           return this.renderSelect(errorMessage, autofocus, this.props);
         }
 
+        if (schema.media && schema.media.type === "placement") {
```
ahoskins: Todo use the real media.types.

src/js/components/SchemaField.js
```diff
@@ -231,6 +233,61 @@ class SchemaField extends Component {
...
+      formData,
+      onChange,
+      onBlur,
+      onFocus
+    } = props;
+
+    // example formData: [[“@zone”,“GROUP_BY”,“3”],[“@zone”,“MAX_PER”,“2"],[“hostname “,”UNIQUE”]]”
+
+    // questions:
+    // 1. when parsing and reducing the formData, can we exactly re-use the constraints marathon JSON parser/reducer?
```
ahoskins: These questions are sort of blocking me and @bstavroulakis from progressing any further on this.

src/js/components/PlacementConstraintsWidget.js
```diff
@@ -0,0 +1,238 @@
+import React, { Component } from "react";
```
ahoskins: This file is direct copy-paste from `PlacementSection`. The intention is this widget will be used in `PlacementSection` as done above, and `SchemaField` as done below.

src/js/components/SchemaField.js
orlandohohmeier: We should be able to _ exactly_ re-use the marathon constraints JSON parser/reducer

src/js/components/SchemaField.js
```diff
@@ -231,6 +233,61 @@ class SchemaField extends Component {
...
+    // 2. do we want to render PlacementConstraintsWidget directly here, or build a higher component (ex: SpecialFormField)?
```
orlandohohmeier: The widget should be the adapter no need for an additional component.

src/js/components/SchemaField.js
```diff
@@ -231,6 +233,61 @@ class SchemaField extends Component {
...
+    // 3. call the onChange() from mozilla library (the local variable defined above) with the reduced value from 1)
+    // 4. the errors prop...figure out later
+    // 5. PlacementConstraintsFrameworkAdapter can hold all the logic above?
+
+    // replace this implementation (copied from CreateServiceFormModal) with our own
+    // will be also use Batch and Transaction or can we be less generic and manually push/splice the constraints?
+    const onAddItem = (event) => {
```
orlandohohmeier: We should move the add and remove item handler into the widget. The widget should implement the standard schema form  interface.

src/js/components/SchemaField.js
orlandohohmeier: How does this relate to field heading?

ahoskins: Talked offline

ahoskins:
Didn't make as much progress on this today as hoped. But the files in place, and I think general flow of how this will work. @orlandohohmeier @nLight  can hopefully take over and sort out the reducers / parsers, since they are familiar with how they work.

src/js/components/PlacementConstraintsPartial.js
```diff
@@ -0,0 +1,238 @@
...
+      "constraints"
+    );
+    const hasOneRequiredValue = data.some(function(constraint) {
+      return PlacementConstraintsUtil.requiresValue(constraint.operator);
+    });
+    const hideValueColumn = data.every(function(constraint) {
+      return PlacementConstraintsUtil.requiresEmptyValue(constraint.operator);
+    });
+
+    return data.map((constraint, index) => {
```
nLight: Please sync with @Poltergeist regarding this `index`. To implement the filtering we need, we have to stop relying on the map-index, as the `data` array could be pre-filtered thus, the native array index won't reflect the position of the constraint in the app definition JSON.

src/js/components/PlacementConstraintsFrameworkAdapter.js
```diff
@@ -0,0 +1,102 @@
...
+    super(props);
+
+    this.state = {
+      batch: new Batch(),
+      appConfig: this.getInitialAppConfig()
+    };
+  }
+
+  getInitialAppConfig() {
+    // parse
```
nLight: Do you think it's worth keeping this comment. It's pretty obvious, isn't it?

src/js/components/PlacementConstraintsFrameworkAdapter.js
```diff
@@ -0,0 +1,102 @@
...
+    const arrayified = JSON.parse(this.props.data);
+
+    // add a transaction for each in the array of given constraints
+    const batch = new Batch();
```
nLight: I bet we have to pass `arrayified` to the `Batch`

src/js/components/PlacementConstraintsFrameworkAdapter.js
```diff
@@ -0,0 +1,102 @@
...
+
+    return this.getAppConfig(batch);
+  }
+
+  getAppConfig(batch = this.state.batch, baseConfig = {}) {
+    const baseConfigCopy = deepCopy(baseConfig);
+    const newConfig = batch.reduce(JSONSingleContainerParser, baseConfigCopy);
```
nLight: `JSONSingleContainerParser` is just for sake of WIP, correct?

src/js/components/PlacementConstraintsFrameworkAdapter.js
```diff
@@ -0,0 +1,102 @@
...
+
+    return CreateServiceModalFormUtil.stripEmptyProperties(newConfig);
+  }
+
+  // direct copy paste
```
nLight: We might want to have a higher order component for the `Batch`. That would implement all the `onChange` mechanics. So that we could wrap the Partial with the component and avoid copy-pasting. What do you think @Poltergeist ?

src/js/components/PlacementConstraintsPartial.js
Poltergeist: Yes because of some constraints which should be displayed outside of the normal form we can't depend on the index of the array being right.

So as an example we do have the following example in the JSON.

```JS
let constraints = [ 
  ["@special", "IS", "SOMETHING"], 
  ["hostname" "UNIQUE"]
]
```
The special value we do not want to show meaning we would filter it before handing it off, But we really need to have the right index on the hostname.

So the structure I would like to propose would be:
```JS
let constraints = [
  {
    constraint: ["hostname", "UNIQUE"],
    index: 1
  }
]
```

src/js/components/PlacementConstraintsFrameworkAdapter.js
Poltergeist: Yes  that would be very helpful, I think something like `<BatchForm>`

src/js/components/PlacementConstraintsFrameworkAdapter.js
ahoskins: Changing this to `Constraints.JSONReducer`

src/js/components/PlacementConstraintsFrameworkAdapter.js
ahoskins: just WIP :)

src/js/components/PlacementConstraintsFrameworkAdapter.js
ahoskins: I think I'm going this correct now. I'm doing something very similar to `CreateServiceModalForm` getNewStateForJSON.

ahoskins:
I've rebased and pushed my latest changes. I would appreciate someone from HH @orlandohohmeier @nLight @Poltergeist to pick this up. I'm a confused if I'm doing this right, and this branch is currently not working right now.

juliangieseke:
Is this still WIP?

nLight:
nope

src/js/components/PlacementConstraintsPartial.js
nLight: Did we merge @Poltergeist PR?

juliangieseke: as discussed: Refactoring later, its mergable.

PR #2517, opened 2017-11-17
🚨 fix(OverviewDetailTab): parse JSON

nLight:
`@dcos/mesos-client#request` had a breaking change so that all requests have to be parsed.


orlandohohmeier:
LGTM! 🚢

PR #2519, opened 2017-11-21
feat(SchemaField): render yaml ace editor

ahoskins:
This renders an `AceEditor` when "application/x-yaml" is encountered in the JSON schema media type. We assume the YAML is base64 encoded.

**Testing**:
1. Checkout this branch
2. Turn on fixtures
3. Navigate to catalog, and click "Review and Run" on any package, then click "Edit Config" and scroll down to the YAML in the form.

**After doing the above three steps you will see this:**
screen shot 2017-11-22 at 2 06 19 pm

**If an invalid base64 encoded string is pasted into the JSON editor:**
screen shot 2017-11-22 at 2 07 12 pm

**But since we don't do YAML validation, there are many base64 encoded strings that are not valid YAML, but will still render:**
screen shot 2017-11-22 at 2 08 02 pm

**If the field is required, it will show the required field error just like other form inputs:**
screen shot 2017-11-22 at 2 00 43 pm

 

 

**Checklist**

### DIFF
--- a/src/styles/components/framework-configuration/styles.less
+++ b/src/styles/components/framework-configuration/styles.less
@@ -7,3 +7,13 @@
 .pre-install-notes.message.message-warning p {
   display: inline;
 }
+
+.framework-form-yaml-editor {
+  border: 1px solid @form-control-border-color;
+  border-radius: 4px;
+  width: 100%;
+
+  .ace_scroller.ace_scroll-left {
+    box-shadow: none;
+  }
+}
--- a/src/js/components/SchemaField.js
+++ b/src/js/components/SchemaField.js
@@ -3,6 +3,8 @@ import deepEqual from "deep-equal";
 import { Tooltip } from "reactjs-components";
 import DefaultSchemaField
   from "react-jsonschema-form/lib/components/fields/SchemaField";
+import AceEditor from "react-ace";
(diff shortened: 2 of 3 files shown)


src/styles/components/framework-configuration/styles.less
```diff
@@ -7,3 +7,14 @@
 .pre-install-notes.message.message-warning p {
   display: inline;
 }
+
+.framework-form-yaml-editor {
+  // todo use same border color as the inputs
```
Poltergeist: `@form-control-border-color` might be what you are looking for.

ahoskins: Nice, thanks 👍

ahoskins:
alfreddo test this

ahoskins:
After demo with the data services team and talking with @leemunroe, I've made some changes to this. 

1. Made it 100% width of it's container. This is more appropriate for editing bigger YAML files.
2. Added gutter which includes line numbers and folding.
3. 2 tab instead of 4 indent.
4. Font size normalized compared to the other font on the form. Other fonts are 15px, and this monospace font is 14px which makes it look same size as the other fonts.
5. Removed box shadow on scrolling, and the 80 character margin ruler thing.

PR #2521, opened 2017-11-23
fix(ServiceActions): adjust service edit actions back button test

orlandohohmeier:
Add a delay after changing the name input to ensure that the "type" operation is completed before the test hits the "Review & Run" button as cypress otherwise fails to properly change the name input. This change should increase the test reliability.


MatApple:
@orlandohohmeier I don't think timeout will help because Cypress run's every find/get on a loop and poll's till it finds the thing or times-out after a couple of seconds. So it has the timeout built in.

ahoskins:
These tests have been removed as of #2543, and they also don't fix this flakey test. I'm going to close this PR.

PR #2522, opened 2017-11-23
DCOS-19559: Adapt placement constraints widget to the SDK form

nLight:
✅   Depends on #2516 

1. Implement `BatchContainer` - a wrapper that takes care of the Batch even cycle
2. Implement an adapter for the `PlacementConstraintPartial`
3. Use the Adapter in the json-schema form 

Would be awesome to have a demo-universe to test this. @orlandohohmeier do you know how to put one together? We need a placement constraints with the media type `placement` and the json

Edit @nlight
This is what I did:
```
+---------------------------------------------------+          
|Adapter                    Batch               <------- String
|                              |                    |          
|  +---------------------------|----------------+   |          
|  |BatchForm                  |                |   |          
|  |                           ↓  Data |        |   |          
|  |   +-------------------------------|------+ |   |          
|  |   |Partial                        |      | |   |          
|  |   |                               |      | |   |          
|  |   |                               ↓      | |   |          
|  |   |                                      | |   |          
|  |   |                                      | |   |          
|  |   |                                      | |   |          
|  |   |                  |Events             | |   |          
|  |   +------------------|-------------------+ |   |          
|  |                      |                     |   |          
|  |                      ↓     |Batch          |   |          
|  +----------------------------|---------------+   |          
|                               ↓                 -----> String
+---------------------------------------------------+          

made with textik.com
```

### DIFF
--- a/src/js/components/BatchContainer.js
+++ b/src/js/components/BatchContainer.js
@@ -1,0 +1,63 @@
+import React, { Component } from "react";
+
+import Transaction from "#SRC/js/structs/Transaction";
+import TransactionTypes from "#SRC/js/constants/TransactionTypes";
+
+export default class BatchContainer extends Component {
+  constructor() {
+    super(...arguments);
+
+    this.handleChange = this.handleChange.bind(this);
+    this.handleAddItem = this.handleAddItem.bind(this);
+    this.handleRemoveItem = this.handleRemoveItem.bind(this);
+  }
+
+  handleChange(event) {
+    const fieldName = event.target.getAttribute("name");
+    if (!fieldName) {
+      return;
+    }
+
+    const value = event.target.type === "checkbox"
+      ? event.target.checked
+      : event.target.value;
+
+    const path = fieldName.split(".");
+    const batch = this.props.batch.add(new Transaction(path, value));
+
+    this.props.onChange(batch, fieldName);
+  }
+
+  handleAddItem(event) {
+    const { value, path } = event;
+
+    const batch = this.props.batch.add(
+      new Transaction(path.split("."), value, TransactionTypes.ADD_ITEM)
+    );
+
+    this.props.onChange(batch);
+  }
+
+  handleRemoveItem(event) {
+    const { value, path } = event;
+
+    const batch = this.props.batch.add(
+      new Transaction(path.split("."), value, TransactionTypes.REMOVE_ITEM)
+    );
+
+    this.props.onChange(batch);
+  }
+
+  render() {
+    const { children } = this.props;
+
+    return (
+      <div onChange={this.handleChange}>
+        {React.cloneElement(children, {
+          onAddItem: this.handleAddItem,
+          onRemoveItem: this.handleRemoveItem
+        })}
+      </div>
+    );
+  }
+}
--- a/src/js/components/PlacementConstraintsFrameworkAdapter.js
+++ b/src/js/components/PlacementConstraintsFrameworkAdapter.js
@@ -1,0 +1,94 @@
+import React, { Component } from "react";
+import Batch from "#SRC/js/structs/Batch";
+import PlacementConstraintsPartial
+  from "#SRC/js/components/PlacementConstraintsPartial";
+import BatchContainer from "#SRC/js/components/BatchContainer";
+import {
+  JSONReducer,
+  JSONParser
+} from "#PLUGINS/services/src/js/reducers/serviceForm/JSONReducers/Constraints";
+import { combineReducers } from "#SRC/js/utils/ReducerUtil";
+import {
+  FormReducer
+} from "#PLUGINS/services/src/js/reducers/serviceForm/FormReducers/Constraints";
+import CreateServiceModalFormUtil
+  from "#PLUGINS/services/src/js/utils/CreateServiceModalFormUtil";
+
+const jsonReducer = combineReducers({ constraints: JSONReducer });
+
+export default class PlacementConstraintsFrameworkAdapter extends Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      batch: this.generateBatchFromInput()
+    };
+
+    this.handleBatchChange = this.handleBatchChange.bind(this);
+  }
+
+  componentWillReceiveProps(nextProps) {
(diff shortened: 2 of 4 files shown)


src/js/components/PlacementConstraintsFrameworkAdapter.js
```diff
@@ -0,0 +1,89 @@
...
+    } else {
+      this.setState({ batch });
+    }
+  }
+
+  render() {
+    const { batch } = this.state;
+    const data = { constraints: batch.reduce(FormReducer) };
+
+    // TODO what to do about errors
```
Poltergeist: Did we create an issue for this?

nLight:
Alfred test this please

src/js/components/PlacementConstraintsFrameworkAdapter.js
nLight: Nothing to do at this stage. Removed 🙇

nLight: Please 👍 @Poltergeist

src/js/components/SchemaField.js
```diff
@@ -277,6 +296,10 @@ class SchemaField extends Component {
           return this.renderSelect(errorMessage, autofocus, this.props);
         }
 
+        if (schema.media && schema.media.type === "placement") {
```
orlandohohmeier: The `type` is wrong. Please change to "application/x-**constraints**+json" as documented in the respective JIRA. I'd also change this to a regular expression as we eventually want to support different variations (for details please see JIRA).

src/js/components/BatchContainer.js
```diff
@@ -0,0 +1,63 @@
...
+
+    this.props.onChange(batch);
+  }
+
+  render() {
+    const { children } = this.props;
+
+    return (
+      <div onChange={this.handleChange}>
+        {React.cloneElement(children, {
```
orlandohohmeier: 💭   Not blocking but we should check if there are other (better) ways than cloning to _connect_ two elements as we the _surrounding_  component might overwrite props directly passed to the children. /cc @dcos/frontend-foundation

src/js/components/PlacementConstraintsFrameworkAdapter.js
```diff
@@ -0,0 +1,88 @@
...
+  JSONParser
+} from "#PLUGINS/services/src/js/reducers/serviceForm/JSONReducers/Constraints";
+import { combineReducers } from "#SRC/js/utils/ReducerUtil";
+import {
+  FormReducer
+} from "#PLUGINS/services/src/js/reducers/serviceForm/FormReducers/Constraints";
+import CreateServiceModalFormUtil
+  from "#PLUGINS/services/src/js/utils/CreateServiceModalFormUtil";
+
+const jsonReducer = combineReducers({ constraints: JSONReducer });
```
ahoskins: Ah interesting, this is something I was missing when trying to make this work! I need to look more into what the point of this is.

src/js/components/SchemaField.js
ahoskins: Yes, this is my bad in being misleading about this. I used "placement" just as my temporary data. The Jira with the proper types is here: https://jira.mesosphere.com/browse/INFINITY-2620

src/js/components/PlacementConstraintsFrameworkAdapter.js
nLight: Yes, `combineReducers` not only combines them, but also injects the state into the resulting reducer, referring to your comment on `this` wasn't being available.

src/js/components/BatchContainer.js
nLight: We can use react context - it's like props, but if you need to go deeper -_-
But this is the technique we're using all over the place in the router. We just clone elements and mix props we want to pass down.

src/js/components/SchemaField.js
nLight: That was actually my fault not copying the linked issue 😅

src/js/components/BatchContainer.js
orlandohohmeier: I know that we're using this technique in many places and I think it's about time that we question the approach and think about a different solution.

nLight:
@ahoskins there's no straightforward way to address this in my opinion. We either go with a totally custom header here, or adjust the headers helpers in the form in general. For example I'd like to capitalize the title, I can do this with CSS, but that will hit all the headers/labels. Which is outside of scope of this PR. How do you think @orlandohohmeier?

src/js/components/BatchContainer.js
nLight: Even though pitfalls are pretty obvious, I'd say there's not much you can do. Overusing context is also considered a bad practice.

orlandohohmeier:
@nLight I'd address that in a followup PR. Please file a JIRA and pull into this sprint. @ahoskins can you please help estimate that one?

ahoskins:
Jira created: https://jira.mesosphere.com/browse/DCOS-19700

src/js/components/PlacementConstraintsFrameworkAdapter.js
```diff
@@ -0,0 +1,88 @@
...
+    const stateChanged = nextState.batch !== this.state.batch;
+
+    return propsChanged || stateChanged;
+  }
+
+  generateBatchFromInput(props = this.props) {
+    const { data } = props;
+
+    const parsedInput = CreateServiceModalFormUtil.stripEmptyProperties(
+      JSON.parse(data)
```
ahoskins: We should `try catch` this in the case when exception in parsing the JSON data. Then I think just return the previous batch, to prevent going into a weird state:

```
    let parsedInput;
    try {
      parsedInput = CreateServiceModalFormUtil.stripEmptyProperties(
        JSON.parse(data)
      );
    } catch (e) {
      return this.state.batch;
    }
```

To reproduce, just delete one of the brackets in the JSON editor "placement_constraints

nLight: Good one! Thank you!

nLight:
Addressed all the issues and rebased! @orlandohohmeier @ahoskins please have a look!

PR #2523, opened 2017-11-24
DCOS-19657: add advanced type to constraints input

Poltergeist:
This introduces the possibility to give constraints a type as a optional field which will be used to
prevent jumping to custom fields which will be added later.

To test this basically play around with the Placement section nothing should behave differently.

Closes DCOS-19657

### DIFF
--- a/plugins/services/src/js/reducers/serviceForm/FormReducers/Constraints.js
+++ b/plugins/services/src/js/reducers/serviceForm/FormReducers/Constraints.js
@@ -13,12 +13,21 @@ function processTransaction(state, { type, path, value }) {
   let newState = state.slice();
 
   if (type === ADD_ITEM) {
-    newState.push({ fieldName: null, operator: null, value: null });
+    let defaultValue = { fieldName: null, operator: null, value: null };
+
+    if (value != null && value.type != null) {
+      defaultValue = Object.assign({}, defaultValue, value);
+    }
+
+    newState.push(defaultValue);
   }
   if (type === REMOVE_ITEM) {
     newState = newState.filter((item, index) => {
       return index !== value;
     });
+  }
+  if (type === SET && name === "type") {
+    newState[index].type = value;
   }
   if (
     type === SET &&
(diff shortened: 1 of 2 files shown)


src/js/components/PlacementConstraintsPartial.js
```diff
@@ -223,7 +223,8 @@ export default class PlacementSection extends Component {
           <FormGroup className="column-12">
             <AddButton
               onClick={this.props.onAddItem.bind(this, {
-                path: "constraints"
+                path: "constraints",
+                value: { type: "advanced" }
```
orlandohohmeier: 🐷  I'd rename this to something more generic.

plugins/services/src/js/reducers/serviceForm/FormReducers/Constraints.js
```diff
@@ -13,13 +13,22 @@ function processTransaction(state, { type, path, value }) {
   let newState = state.slice();
 
   if (type === ADD_ITEM) {
-    newState.push({ fieldName: null, operator: null, value: null });
+    let defaultValue = { fieldName: null, operator: null, value: null };
+
+    if (value != null && value.type != null) {
```
ahoskins: How about `value.type === "default"` to be fully explicit in the linkage between the code below and this check?

ahoskins:
Alfreddo test this please

nLight:
`@zone` isn't jumping for pods

juliangieseke:
Is this still `hold merge`?

plugins/services/src/js/reducers/serviceForm/FormReducers/Constraints.js
Poltergeist: If we would go with `value.type === "default"` We would exclude all the other types. With the condition above we are making sure that value is not null and the type is set.

Poltergeist: The confusion here is that we do have a defaultValue and a default type which has nothing to do with each other. The defaultValue is the defaultValue for a constraint, the type describes the type in the future we might add types like hostname for example.

PR #2525, opened 2017-11-27
DCOS-19481 | fix(placement): swap placement form fields

juliangieseke:
This commit swaps the `operator` and `fieldname`
fields for better usability.

Closes DCOS-19481

### DIFF
--- a/src/js/components/PlacementConstraintsPartial.js
+++ b/src/js/components/PlacementConstraintsPartial.js
@@ -4,7 +4,6 @@ import { Tooltip } from "reactjs-components";
 import { findNestedPropertyInObject } from "#SRC/js/utils/Util";
 import AddButton from "#SRC/js/components/form/AddButton";
 import DeleteRowButton from "#SRC/js/components/form/DeleteRowButton";
-import FieldAutofocus from "#SRC/js/components/form/FieldAutofocus";
 import FieldError from "#SRC/js/components/form/FieldError";
 import FieldHelp from "#SRC/js/components/form/FieldHelp";
 import FieldInput from "#SRC/js/components/form/FieldInput";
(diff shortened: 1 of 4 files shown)


juliangieseke:
wait…

juliangieseke:
hm, I've removed the autofocus in this case because I find it confusing to autofocus the second input - now the tests are failing because there is no autofocused field anymore.

Should I adapt the tests or add autofocus again? I'd say adapting the tests.

juliangieseke:
alfred test this please (havent even touched that file 🤔 )

Poltergeist:
Hey @juliangieseke was there an offline discussion?

juliangieseke:
no, I have adapted the tests on my own - feel free to request changes if you want.

src/js/components/PlacementConstraintsPartial.js
```diff
@@ -155,8 +138,24 @@ export default class PlacementSection extends Component {
...
+            showError={Boolean(fieldNameError)}
+          >
+            {fieldLabel}
+            <FieldInput
+              name={`constraints.${index}.fieldName`}
+              type="text"
+              placeholder="hostname"
+              value={constraint.fieldName}
+            />
+            <FieldHelp>Ex: zone, region, hostname</FieldHelp>
```
ahoskins: I see you snuck in these help texts. 😉  Why didn't you also change the "Value" field help text to be like in the [prototypes](https://mesosphere.invisionapp.com/d/#/console/11374342/261629378/preview)?

tests/pages/services/ServiceFormModal-cy.js
```diff
@@ -728,11 +728,6 @@ describe("Service Form Modal", function() {
           .contains("Add Placement Constraint")
           .click();
 
-        cy
```
ahoskins: I think this is good removing this because the first field is now a select, which we don't need to focus. 👍

src/js/components/PlacementConstraintsPartial.js
juliangieseke: Mostly because I missed it 🤔 
See comment below for more details.

juliangieseke:
I have added the info field to value, but as there already was our optional info, it now has two texts in some cases. IMHO it is okay that way. 👍 
screen shot 2017-11-29 at 09 49 36

nLight:
I'd say if we go with help text we should integrate "This field is optional" text into the help text, so they don't stack on top of each other.

src/js/components/PlacementConstraintsPartial.js
nLight: Could we extract this as a PR? I feel like we're blocking a committed feature because of this being snuck in.

Poltergeist:
I agree with @nLight only one help text looks better.

juliangieseke:
removed the text 👍

nLight:
I'd say it's not a fix but a feature though 🙈

PR #2528, opened 2017-11-28
fix(FrameworkConfiguration): move legal t&c text when installing

ahoskins:
It is a legal requirement to show the terms and conditions notice on the same page as the action that deploys the service. This change does this by moving the T&C link to the new framework review screen. This will only show when deploying from catalog, not when editing the framework. This change also removes the T&C link from the package detail page.

**Most frameworks have preinstall notes, so we append the T&C to the end:**
image 2017-11-28 at 1 50 09 pm

**If it doesn't have preinstall notes, we just show the T&C (faked example, couldn't find a service without preinstall notes):**
image 2017-11-28 at 1 56 01 pm

**The T&C link is now removed from package detail:**
image 2017-11-28 at 2 19 17 pm

Closes DCOS-19359.

**Checklist**

### DIFF
--- a/src/js/components/FrameworkConfiguration.js
+++ b/src/js/components/FrameworkConfiguration.js
@@ -198,32 +198,53 @@ class FrameworkConfiguration extends Component {
     return actions;
   }
 
+  getTermsAndConditions() {
+    const { packageDetails } = this.props;
+
+    const termsUrl = packageDetails.isCertified()
+      ? "https://mesosphere.com/catalog-terms-conditions/#certified-services"
+      : "https://mesosphere.com/catalog-terms-conditions/#community-services";
+
+    return {
+      __html: `By running this service you agree to the <a href=${termsUrl} target='_blank' title='Terms and Conditions'>terms and conditions</a>.`
+    };
+  }
+
+  getWarningMessage() {
+    const { packageDetails } = this.props;
+
+    const preInstallNotes = packageDetails.getPreInstallNotes();
+    if (preInstallNotes) {
+      const message = StringUtil.parseMarkdown(preInstallNotes);
+      message.__html = `<strong>Preinstall Notes: </strong>${message.__html} ${this.getTermsAndConditions().__html}`;
+
+      return (
+        <div
+          dangerouslySetInnerHTML={message}
+          className="pre-install-notes message message-warning"
+        />
+      );
+    }
+
+    return (
+      <div
+        dangerouslySetInnerHTML={this.getTermsAndConditions()}
+        className="pre-install-notes message message-warning"
+      />
+    );
+  }
+
   getReviewScreen() {
-    const {
-      packageDetails,
-      deployErrors,
-      isInitialDeploy,
-      formData
-    } = this.props;
+    const { deployErrors, isInitialDeploy, formData } = this.props;
 
     const fileName = "config.json";
     const configString = JSON.stringify(formData, null, 2);
 
     const renderKeys = this.getHashMapRenderKeys(formData);
 
-    const preInstallNotes = packageDetails.getPreInstallNotes();
-    let preinstall = null;
-    if (preInstallNotes && isInitialDeploy) {
-      const preInstallNotesParsed = StringUtil.parseMarkdown(preInstallNotes);
-      preInstallNotesParsed.__html =
-        "<strong>Preinstall Notes: </strong>" + preInstallNotesParsed.__html;
-
-      preinstall = (
-        <div
-          dangerouslySetInnerHTML={preInstallNotesParsed}
-          className="pre-install-notes message message-warning"
-        />
-      );
+    let warningMessage = null;
+    if (isInitialDeploy) {
+      warningMessage = this.getWarningMessage();
     }
 
     let errorsAlert = null;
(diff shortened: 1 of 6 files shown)


ahoskins:
@leemunroe looping you in since you are creator of the Jira.

src/js/structs/UniversePackage.js
```diff
@@ -68,7 +68,13 @@ class UniversePackage extends Item {
   }
 
   getPreInstallNotes() {
-    return this.get("preInstallNotes");
+    // punctuate the end if not present
```
ahoskins: I'm adding this because there are some packages (for example [cassandra](https://github.com/mesosphere/universe/blob/version-3.x/repo/packages/C/cassandra/29/package.json#L19)) don't end the preinstall notes with a period. The proper fix would be for the package itself to not do this, but I think it's worth the patch in the UI code. If I don't add this, then when appending the T&C it lacks a period, which is not nice.

ahoskins:
Alfreddo test this please

src/js/structs/UniversePackage.js
weblancaster: It feels like this should be inside a util instead the struct

ahoskins: Yeah true, I can move into string util.

src/js/utils/StringUtil.js
```diff
@@ -84,6 +84,18 @@ const StringUtil = {
     return string;
   },
 
+  punctuate(string) {
```
nLight: Would be great to see a test or two for this util :)

src/js/utils/StringUtil.js
```diff
@@ -84,6 +84,18 @@ const StringUtil = {
...
+    if (typeof string !== "string") {
+      return null;
```
orlandohohmeier: 🎨  I'd return an empty string to not change the return type.

ahoskins:
@orlandohohmeier @nLight I've added tests to `punctuate` and change to return `""` in error case. Should be good to 🚢

src/js/utils/__tests__/StringUtil-test.js
```diff
@@ -206,6 +206,30 @@ describe("StringUtil", function() {
...
+        "butterfly is nice."
+      );
+    });
+
+    it("does not further punctuate a sentence with a period", function() {
+      expect(StringUtil.punctuate("butterfly is nice.")).toEqual(
+        "butterfly is nice."
+      );
+    });
+  });
```
juliangieseke: I'd love to see some edge cases here. 
`StringUtil.punctuate("");`
`StringUtil.punctuate(new Array());`
`StringUtil.punctuate(null);`
`StringUtil.punctuate(1);`

ahoskins:
@juliangieseke I've added a couple more test cases for weird inputs 👍

leemunroe:
@ahoskins FYI the link color in this message should inherit the color of the message text i.e. should be brown not purple. Similar how we do it for other messages and alerts e.g. this danger message

image

ahoskins:
@leemunroe Cool thanks, I didn't notice that. Will push a fix for this.

src/js/components/FrameworkConfiguration.js
```diff
@@ -198,34 +198,55 @@ class FrameworkConfiguration extends Component {
...
-        "<strong>Preinstall Notes: </strong>" + preInstallNotesParsed.__html;
+    if (preInstallNotes) {
+      const message = StringUtil.parseMarkdown(preInstallNotes);
+      message.__html = `<strong>Preinstall Notes: </strong>${message.__html} ${this.getTermsAndConditions().__html}`;
 
-      preinstall = (
+      return (
         <div
-          dangerouslySetInnerHTML={preInstallNotesParsed}
+          dangerouslySetInnerHTML={message}
```
juliangieseke: any chance to get rid of this?

src/js/components/FrameworkConfiguration.js
```diff
@@ -198,34 +198,55 @@ class FrameworkConfiguration extends Component {
...
           className="pre-install-notes message message-warning"
         />
       );
     }
 
+    return (
+      <div
+        dangerouslySetInnerHTML={this.getTermsAndConditions()}
```
juliangieseke: same here

src/js/components/FrameworkConfiguration.js
```diff
@@ -235,7 +256,7 @@ class FrameworkConfiguration extends Component {
       <div className="flex-item-grow-1">
         <div className="container container-wide">
           {errorsAlert}
-          {preinstall}
+          {warningMessage}
```
juliangieseke: why did you decide to change this? is it really a "warning"? wouldn't it be possible to split them into two? 
```
{preinstall}
{tandc}
```

src/js/components/FrameworkConfiguration.js
nLight: I think the answer is "no" since those messages can contain html we want to render.

ahoskins: Yeah, this is a pattern we follow [elsewhere](https://github.com/dcos/dcos-ui/blob/master/src/js/pages/catalog/PackageDetailTab.js#L429) after parsing markdown. It's ok - I don't feel too dangerous doing this 😉

src/js/components/FrameworkConfiguration.js
ahoskins: It must be as one because the content is part of the same paragraph (the preinstall and t&c are concat)

PR #2530, opened 2017-11-29
DCOS-19354: add IPv6 healthcheck support

Poltergeist:
This introduces IPv6 form support to healthchecks.

To test this you have to set up a http healthcheck on a service with a docker image, so container type needs to be DOCKER. 

image

**Checklist**

### DIFF
--- a/plugins/services/src/js/components/forms/HealthChecksFormSection.js
+++ b/plugins/services/src/js/components/forms/HealthChecksFormSection.js
@@ -20,6 +20,7 @@ import FormGroupHeadingContent
   from "#SRC/js/components/form/FormGroupHeadingContent";
 import FormRow from "#SRC/js/components/form/FormRow";
 import Icon from "#SRC/js/components/Icon";
+import { findNestedPropertyInObject } from "#SRC/js/utils/Util";
 
 import {
   MESOS_HTTP,
@@ -222,7 +223,7 @@ class HealthChecksFormSection extends Component {
 
     return (
       <FormRow>
-        <FormGroup className="column-12" showError={Boolean(errors.value)}>
+        <FormGroup className="column-6" showError={Boolean(errors.value)}>
           <FieldLabel>
             <FormGroupHeading>
               <FormGroupHeadingContent primary={true}>
@@ -249,6 +250,34 @@ class HealthChecksFormSection extends Component {
     return data.portDefinitions.map((port, index) => {
       return <option key={index} value={index}>{port.name || index}</option>;
     });
+  }
+
+  getIpProtocol(data) {
+    const { healthCheck, key, errors } = data;
+    const runtime = findNestedPropertyInObject(
+      this.props.data,
+      "container.type"
+    );
+    // IPv6 Healthchecks is currently only supported in DOCKER runtime.
+    if (runtime !== "DOCKER") {
+      return null;
+    }
+
+    return (
+      <FormGroup showError={false} className="column-12">
+        <FieldLabel>
+          <FieldInput
+            checked={healthCheck.ipProtocol === "IPv6"}
+            name={`healthChecks.${key}.ipProtocol`}
+            type="checkbox"
+            value="IPv6"
+          />
+          {"Make "}
+          <span className="truecase">IPv6</span>
+        </FieldLabel>
+        <FieldError>{errors.ipProtocol}</FieldError>
+      </FormGroup>
+    );
   }
 
   getHTTPFields(healthCheck, key) {
@@ -341,6 +370,9 @@ class HealthChecksFormSection extends Component {
           </FieldLabel>
           <FieldError>{errors.protocol}</FieldError>
         </FormGroup>
+      </FormRow>,
+      <FormRow key="IPv6">
+        {this.getIpProtocol({ healthCheck, key, errors })}
       </FormRow>
     ];
   }
--- a/plugins/services/src/js/reducers/serviceForm/JSONReducers/HealthChecks.js
+++ b/plugins/services/src/js/reducers/serviceForm/JSONReducers/HealthChecks.js
@@ -8,29 +8,52 @@ import {
   MESOS_HTTPS
 } from "../../../constants/HealthCheckProtocols";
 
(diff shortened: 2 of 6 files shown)


plugins/services/src/js/components/forms/HealthChecksFormSection.js
```diff
@@ -251,6 +252,36 @@ class HealthChecksFormSection extends Component {
...
+    const { healthCheck, key, errors } = data;
+    const image = findNestedPropertyInObject(
+      this.props.data,
+      "container.docker.image"
+    );
+    const runtime = findNestedPropertyInObject(
+      this.props.data,
+      "container.type"
+    );
+    if (image == null || runtime !== "DOCKER") {
```
orlandohohmeier: 🤔  Why is `image`  a requirement?

plugins/services/src/js/components/forms/HealthChecksFormSection.js
```diff
@@ -251,6 +252,36 @@ class HealthChecksFormSection extends Component {
...
+    }
+
+    return (
+      <FormGroup showError={false} className="column-6">
+        <FieldLabel>
+          <FieldInput
+            checked={healthCheck.ipProtocol === "IPv6"}
+            name={`healthChecks.${key}.ipProtocol`}
+            type="checkbox"
+            value="HTTPS"
```
nLight: `value="HTTPS"` 🤔

plugins/services/src/js/reducers/serviceForm/JSONReducers/HealthChecks.js
```diff
@@ -8,23 +8,38 @@ import {
...
-  const newItem = Util.omit(item, ["path", "command", "protocol", "https"]);
-
-  if (item.protocol != null) {
-    newItem.protocol = item.protocol;
-    if (item.protocol.toUpperCase() === COMMAND && item.command != null) {
-      newItem.command = {
-        value: item.command
-      };
-    }
+function getMapHealthChecks(docker) {
```
orlandohohmeier: 🎨   Is there any better way to pass `docker`? How about binding the function (e.g.,  `mapHealthChecks.bind(null, docker)`) or creating a support structure that includes the needed information?

nLight:
I can't get the checkbox to show, what am I doing wrong?

plugins/services/src/js/components/forms/HealthChecksFormSection.js
Poltergeist: hmm that slipped.

plugins/services/src/js/reducers/serviceForm/JSONReducers/HealthChecks.js
Poltergeist: Well how is binding the variable and in that way creating a new function a better way then currying the docker?

orlandohohmeier: I'd argue that changing the function signature to  `mapHealthChecks(docker, item)` and then partially applying `docker` is easier to reason about. It's also using generic operations instead of creating a very specific `getMapHealthChecks` function that returns a _bound_ version of `mapHealthChecks`. Anyway, as indicated this is a purely stylistic comment and by no means blocks this PR.

ahoskins:
In the PR description your screenshot looks different. I can't find any invision mockups for this, but just want to make sure the layout is correct 😄 

image 2017-11-30 at 11 27 59 am

leemunroe:
Vertically stacked makes more sense to me here. As per the first screenshot.

I don't have any background knowledge on this feature but wondering if the label should be more descriptive e.g. `Make IPv6` or `Enable IPv6 support`

Also I think the v is meant to be lowercase i.e. `IPv6`

Poltergeist:
After I have created the Pr I decided that it might look better to have the checkbox in line with the `HTTPS` I should stop doing this :P I will put the box below the HTTPS checkbox and go with `MAKE IPv6`. This adds the possibility to support IPv6 HTTP health checks for Docker environments. Basically Marathon added that feature and asked us if it will work with the UI given the work necessary to support this also as a option in the UI we decided we just add it. This is why there is no invision for this. This is currently very much limited and only will show up if you select a docker runtime.

juliangieseke:
Alfred test this please

plugins/services/src/js/components/forms/HealthChecksFormSection.js
```diff
@@ -251,6 +252,34 @@ class HealthChecksFormSection extends Component {
     });
   }
 
+  getIpProtocol(data) {
+    const { healthCheck, key, errors } = data;
+    const runtime = findNestedPropertyInObject(
```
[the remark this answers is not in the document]
Poltergeist: Could do but I rather don't because of readability. the function plus parameters are quite long. it is better to grasp if the condition is like `(runtime !== "DOCKER")` instead of `(findNestedPropertyInObject("container.docker.image", data) !=== "DOCKER")`

plugins/services/src/js/reducers/serviceForm/JSONReducers/HealthChecks.js
nLight: I'm ok with either of those, but how about renaming docker argument to be `runtime`? We're using it only once here `docker.type === "DOCKER"` to become `runtime === "DOCKER"`.

*Side note*: it's all that pesky functional programming. In OOP I'd just create two mappers and decided to use one or another upfront without obscure functions that return function :trollface:

Poltergeist: Well in OOP you would consider to have two classes which will do this and a factory which would return either a docker runtime mapper or a not docker runtime mapper based on the input. I guess it is far more code to maintain with a slim curry.

juliangieseke:
Alfred test this please

nLight:
After-merge thoughts: `"ipProtoctol": true` also switches on IPv6 checkbox, though I couldn't find a mention of *true* being a valid value in the MARATHON ticket

PR #2532, opened 2017-11-29
fix(SchemaField): fix z-index issue with YAML editor

ahoskins:
This resets the YAML gutter to `z-index: 0`, preventing it from rendering on top of the JSON editor when screen narrow and JSON is expanded.

Repro to find the bug:
1. Checkout latest master
2. Turn fixtures on
3. Open the catalog deploy framework form
3. Narrow the window and open JSON so it takes up fullscreen
4. The unpleasant bug is below:
screen shot 2017-11-29 at 1 04 17 pm

Thanks for @nLight for finding this issue. My bad for not testing this case on the original PR. 

**Checklist**

### DIFF
--- a/src/styles/components/framework-configuration/styles.less
+++ b/src/styles/components/framework-configuration/styles.less
@@ -16,4 +16,8 @@
   .ace_scroller.ace_scroll-left {
     box-shadow: none;
   }
+
+  .ace_gutter {
+    z-index: 0;
+  }
 }


ahoskins:
The [Ace library](https://github.com/ajaxorg/ace/blob/master/lib/ace/css/editor.css#L57) originally sets this gutter to z-index = 4. I'm not sure why they do this, but overriding it seems to work. I've tested this fix on our supported browsers, and no issues anywhere.

orlandohohmeier:
🚢

PR #2533, opened 2017-11-30
test(ServiceList-test): add jest snapshot

weblancaster:
More about [snapshots](https://facebook.github.io/jest/docs/en/snapshot-testing.html#content)

Replace "manual" component test to use snapshot instead.

**Before example**:
`expect(componetBlaInstance.props.className).toEqual("someClass")`

**Now**:
You match the expected render result.

**Checklist**

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -8143,6 +8143,11 @@
         }
       }
     },
+    "react-test-renderer": {
+      "version": "15.4.1",
+      "from": "react-test-renderer@15.4.1",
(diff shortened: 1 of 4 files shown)


juliangieseke:
We had some issues in other repositories with snapshot tests. Even if I really like them. I'm not completely sure if we should use them here while not able to use them in other ones 🤔

weblancaster:
@juliangieseke I believe it's fine to use them only where it makes sense, the rule of thumb I'm using is when I need to check a component layout/structure/content.

PR #2541, opened 2017-12-05
DCOS_OSS-1779: Update Container Classes

ashenden:
Later versions of CNVS use different class name conventions for the `.container` layout component. This PR resolves any display issues experienced as a result of this change. The impact of this change is most notable in the Service Create view.

Changes are limited to class name updates, reducing/removing styles, and CNVS npm version bump following some patches made in CNVS.


MatApple:
We need to bump reactjs-components before merging this.

package.json
```diff
@@ -32,7 +32,7 @@
     "browser-info": "0.4.0",
     "classnames": "2.2.3",
     "clipboard": "1.5.10",
-    "cnvs": "1.1.11",
+    "cnvs": "1.1.14",
```
MatApple: We need to run `npm run shrinkwrap` once we've also bumped reactjs-components.

ashenden: @MatApple when I run `npm run shrinkwrap` locally I get some errors and I'm reluctant to fix them for fear of breaking something.

MatApple:
I'll take care of fixing them. Our shrinkwrap has some quirks

package.json
MatApple: Fixed.

ashenden:
@MatApple from what I can see, there are no visual issues resulting from the reactJS version bump.  I worked in a couple commits relating to some patches made to the `Alert` component, but that was related to the CNVS bump not ReactJS.  Feel free to review and merge if you're good with it.

ashenden:
@MatApple fixed tests.  All tests now pass now.  Ok to merge when you are.

PR #2544, opened 2017-12-06
fix(HealthChecks): fix IPv6 support to set right values

Poltergeist:
This fixes the earlier introduced feature for IPv6 Support. The problem was that the field is
setting a string but is providing a boolean. By parsing the JSON we made it possible to have the
value `true` and `IPv6` as valid values for the checkbox to be ticked.

The solution to the problem was to change the name of the field to something else and set the value
based on the boolean. And have the parsed JSON set the boolean based on the input.

Because code explains better then a thousand words see this exert:
```JS
if (`healthChecks.${index}.ipProtocol` === joinedPath) {
  state[index].ipProtocolCheckbox = value === "IPv6";
  state[index].ipProtocol = value;
}
if (`healthChecks.${index}.ipProtocolCheckbox` ===
    joinedPath) {
  state[index].ipProtocolCheckbox = value;
  state[index].ipProtocol = value ===
    true ? "IPv6" : "IPv4";
}
```

- For testing create a new single container service.
- Add a Healthcheck
- Select the HTTP type
- Go to the JSON editor
- Add to the HealthCheck the `ipProtocol` field with arbitrary values

Expected behavior is that only the value "IPv6" should add the check mark to the check box. Please
do test `"IPv4"`, `true` and `false`.

### DIFF
--- a/plugins/services/src/js/components/forms/HealthChecksFormSection.js
+++ b/plugins/services/src/js/components/forms/HealthChecksFormSection.js
@@ -267,10 +267,9 @@ class HealthChecksFormSection extends Component {
       <FormGroup showError={false} className="column-12">
         <FieldLabel>
           <FieldInput
-            checked={healthCheck.ipProtocol === "IPv6"}
-            name={`healthChecks.${key}.ipProtocol`}
+            checked={healthCheck.ipProtocolCheckbox === true}
+            name={`healthChecks.${key}.ipProtocolCheckbox`}
             type="checkbox"
-            value="IPv6"
           />
           {"Make "}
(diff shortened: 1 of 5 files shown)


nLight:
`hold merge` because this fix introduces another issue. I was able to get `ipProtocolCheckbox: true` in JSON

bstavroulakis:
For me there seems to be an issue. Although the IPv6 is set in the JSON it is not mapping with the checkbox

image

nLight:
Sorry https://github.com/dcos/dcos-ui/pull/2549

PR #2545, opened 2017-12-06
fix(PlacementConstraintsSchemaField): escape/unescape output

nLight:
SDK uses handlebars engine to generate JSON definition, and handlebars would escape quotes in partial JSONs so that we cant just parse JSON, we should unescape it first. But since SDK expects it to be escaped we should escape the output back again.

The fastest way to do this was regexp.


nLight:
Alfred test this

PR #2547, opened 2017-12-06
DCOS-19658: Adjust system tests cluster config

nLight:
Turned out we have to  use `cent-os-7-dcos-prereqs` image that has all the good stuff and the `dcos-launch` link was outdated

Ask for credentials, then:
```
$ docker-compose build
$ docker-compose up -d
$ docker-compose exec toolchain system-tests/_scripts/launch-cluster.sh
```

It should create a cluster, don't forget to tear it down with:
```
$ docker-compose exec toolchain system-tests/_scripts/delete-cluster.sh
```

And shut down the toolchain
```
$ docker-compose down
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -36,7 +36,7 @@ RUN set -x \
   # Install cypress
   && cypress install --cypress-version ${CYPRESS_VERSION} \
   # Install dcos-launch
-  && curl 'https://downloads.dcos.io/dcos-test-utils/bin/linux/dcos-launch' > /usr/local/bin/dcos-launch \
+  && curl 'https://downloads.dcos.io/dcos-launch/bin/linux/dcos-launch' > /usr/local/bin/dcos-launch \
   && chmod +x /usr/local/bin/dcos-launch \
   # Ensure entrypoint is executable
   && chmod +x /usr/local/bin/dcos-ui-docker-entrypoint \
--- a/system-tests/_scripts/launch-cluster.sh
+++ b/system-tests/_scripts/launch-cluster.sh
@@ -20,7 +20,7 @@ installer_url: ${1:-$INSTALLER_URL}
 platform: aws
 provider: onprem
 aws_region: us-west-2
-os_name: cent-os-7
+os_name: cent-os-7-dcos-prereqs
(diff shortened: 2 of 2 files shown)


system-tests/driver-config/jenkins.sh
```diff
@@ -17,7 +17,7 @@ targets:
 
     env:
       PROXIED_CLUSTER_URL: http://127.0.0.1:4201
-      INSTALLER_URL: "${INSTALLER_URL}"
+      INSTALLER_URL: ${INSTALLER_URL}
```
orlandohohmeier: Those quotes shouldn't be a problem. You're sure that we've to remove them?

nLight: Was an issue in my local setup, would break and complain that yml was malformed:
```yml
---
...
INSTALLER_URL: ""http://{blablabla}""
```

juliangieseke: even `""` shouldnt be an issue 🤔 can we add the quotes again or also remove them from the next three lines? #ocd

nLight: I'll add them back but then if it stops working for me I'll be asking everyone to run the system-tests for me :trollface:

nLight:
Please retest @Poltergeist @juliangieseke

juliangieseke:
Alfred test this please

nLight:
`HOLD MERGE`: even though I can create a cluster there are issues preventing tests from running

nLight:
So... could anyone pick this up? Because double quotes get in the way for me. @juliangieseke?
`Could not parse a JSON or YAML output from the configuration response`

juliangieseke:
I assume you got this message after running it through `docker-compose`? Looks like our `docker-compose.yml` is bad.
https://github.com/docker/compose/issues/2854 => https://docs.docker.com/compose/compose-file/#args
I'll create a PR which fixes our `docker-compose.yml`. after that one is merged we should be able to merge this one as it is now. 👍

nLight:
@juliangieseke thank you for figuring that out! @Poltergeist would you please have a look?

PR #2549, opened 2017-12-06
fix(HealthChecks): use bool as internal representation

nLight:
This fixes the earlier introduced feature for IPv6 Support. The problem was that the field is
setting a string but is providing a boolean. By parsing the JSON we made it possible to have the
value true and IPv6 as valid values for the checkbox to be ticked.

Now it should be impossible to set `true` in JSON and have checkbox to be checked.

How to test: PLEASE give it a hard time. Check the checkbox, set path from JSON, go back and forth from editing JSON to editing form multiple times. Verify that there's no side effects

screen shot 2017-12-06 at 15 20 30

### DIFF
--- a/plugins/services/src/js/reducers/serviceForm/FormReducers/HealthChecks.js
+++ b/plugins/services/src/js/reducers/serviceForm/FormReducers/HealthChecks.js
@@ -38,7 +38,7 @@ module.exports = {
       if (joinedPath === "healthChecks") {
         switch (type) {
           case ADD_ITEM:
-            state.push(value || {});
+            state.push(Object.assign({}, value) || {});
             break;
           case REMOVE_ITEM:
             state = state.filter((item, index) => {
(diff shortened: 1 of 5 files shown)


plugins/services/src/js/reducers/serviceForm/FormReducers/HealthChecks.js
```diff
@@ -65,9 +65,7 @@ module.exports = {
           state[index].command = value;
         }
         if (`healthChecks.${index}.ipProtocol` === joinedPath) {
-          state[index].ipProtocol = value === true || value === "IPv6"
-            ? "IPv6"
-            : "IPv4";
+          state[index].ipProtocol = value;
```
orlandohohmeier: Could you please change `ipProtocol` to `IPv6` enable or something alike?

ahoskins:
alfred test this

orlandohohmeier:
LGTM 💯 ✨

PR #2550, opened 2017-12-07
DCOS-19832: Filter active frameworks only

nLight:
To test:
start confluent-kafka
kill confluent-kafka
start a new one (with the same name)
observe that you see it's tasks

without the fix you wouldn't see `kafka_broker` because MesosStateStore would pick tasks for the killed instance.

screen shot 2017-12-06 at 16 44 16


ahoskins:
alfreddo test this please

PR #2551, opened 2017-12-07
DCOS-19828: Fix mesos stream parsers 😅

nLight:
This PR fixes various bugs in the MesosStream. 

1. Fix FRAMEWORK_REMOVED message format
2. Fix AGENT_REMOVED message format
3. Checking only active frameworks when filtering

Testing :trollface: 

1. Easy: Install `beta-confluent-kafka`, wait until it's fully up and running, then delete it. Install  `beta-confluent-kafka` again. UI should show all the tasks etc _without_ reloading the browser.
2. Hard: You gotta remove an agent to see what happens. I may even have an idea how to do that. Sometimes our precious colleagues from the customer  support team know those tricks better.
3. Nightmare: I don't really know how to test this one. Since we're using this function only when we're fetching tasks that's been launched by Marathon and in theory we can't have killed root Marathon on the cluster I fixed that one *just in case*.

Closes DCOS-19828

### DIFF
--- a/src/js/stores/MesosStateStore.js
+++ b/src/js/stores/MesosStateStore.js
@@ -211,7 +211,7 @@ class MesosStateStore extends GetSetBaseStore {
     }
 
     const framework = frameworks.find(function(framework) {
-      return framework.name === serviceName;
+      return framework.active && framework.name === serviceName;
     });
 
     if (framework) {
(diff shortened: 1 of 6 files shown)


Poltergeist:
So login into the leader with
`dcos node ssh --master-proxy --leader` and then ssh'ing into a agent (find ip on the nodes table) and then do `sudo systemctl stop dcos-mesos-slave.service` to stop the agent. With the same command you can restart the agent `sudo systemctl start dcos-mesos-slave.service`

Some of the input came from @ahoskins from this PR: 
https://github.com/dcos/dcos-ui/pull/2346#pullrequestreview-53668042

Hope this helps.

PR #2552, opened 2017-12-07
refactor(ServiceStatusIcon): add tooltip to serviceStatusIcon

weblancaster:
Keep the UI consistent by adding tooltip to serviceStatusIcon displaying # of services running

Closes DCOS-19747

**How to test:**
The icons in service table (status column) should display tooltips
Same goes for service breadcrumb and panel section in dashboard page

**Checklist**

### DIFF
--- a/plugins/services/src/js/components/__tests__/__snapshots__/ServiceList-test.js.snap
+++ b/plugins/services/src/js/components/__tests__/__snapshots__/ServiceList-test.js.snap
@@ -16,11 +16,16 @@ exports[`ServiceList #getServices returns services that have a value of two elem
       </span>
       <div
         className="dashboard-health-list-health-label">
-        <svg
-          className="icon icon-light-grey icon-mini">
-          <use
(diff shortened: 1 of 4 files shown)


plugins/services/src/js/components/__tests__/__snapshots__/ServiceList-test.js.snap
```diff
@@ -16,11 +16,16 @@ exports[`ServiceList #getServices returns services that have a value of two elem
       </span>
```
weblancaster: this is the updated snapshot

ahoskins:
@weblancaster Isn't this a feature, not a refactor?

plugins/services/src/js/components/__tests__/__snapshots__/ServiceList-test.js.snap
ahoskins: social-lg

PR #2553, opened 2017-12-08
fix(SchemaField): avoid framework regex errors

ahoskins:
This is a sort of complicated situation, let me explain:

Marathon has a schema entry like this (notice there's **no** default value and the field is **not** required):
```
group: {
     regex: <must-be-two-alphanumeric-characters>,
     description: "service group"
}
```

Currently, when fetching this package configuration we were setting `""` as the form value for `group` because no default value was provided by the schema. This caused an error in the framework form like this:

screen shot 2017-12-08 at 11 25 28 am

Which makes sense, because `""` does not validate against the regex. It was a bad idea to set `""` as the fallback value when no default is provided and the field is not required.

**What this PR does:**
 This PR removes the `""` as the empty state for optional fields. Instead, we will set the field as `undefined`. This affects two places:

1. When initializing the data in DeployFrameworkConfigurationwhen, remove the `""` as the fallback value.
2. In onChange for text and number inputs, set the field as `undefined` when the field is optional. When it's required, still set as `""`, this will keep our validation working.

**Testing**:
1. Checkout this branch.
2. Search for marathon package in catalog.
3. When the form opens, the regex errors won't be there. Change the `group` value in the form and JSON. When the `group` is empty, it won't give an error.
 
Closes DCOS-19771

**Checklist**

### DIFF
--- a/src/js/components/SchemaField.js
+++ b/src/js/components/SchemaField.js
@@ -147,6 +147,16 @@ class SchemaField extends Component {
       onFocus
     } = props;
 
+    const handleChange = event => {
+      const value = event.target.value;
+
+      if (value === "" && !required) {
+        onChange(undefined);
+      } else {
+        onChange(value);
+      }
+    };
+
     let field = (
       <FieldInput
         id={name}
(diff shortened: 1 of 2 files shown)


src/js/components/SchemaField.js
```diff
@@ -193,7 +203,9 @@ class SchemaField extends Component {
     const handleChange = event => {
       const value = event.target.value;
 
-      if (value === "") {
+      if (value === "" && !required) {
```
weblancaster: we can re-use this condition in one method.

PR #2554, opened 2017-12-08
DCOS-19883: fix duplicate local string in region/ zone column

bstavroulakis:
Fix duplicate local string [(Local) (Local)] if region/ zone doesn't exist in region/ zone task columns.
If no zone/ region exists show "N/A"

Closes DCOS-19869

To test:
- create an open cluster
- create a service with a task
- go to tasks page and see "N/A" in the region/ zone column

**Checklist**


weblancaster:
Tested and looks good.

bstavroulakis:
Alfred test this

PR #2558, opened 2017-12-09
fix(docker-compose): fix yml to use mapping instead of list

juliangieseke:
_further reading:_
https://github.com/docker/compose/issues/2854
https://docs.docker.com/compose/compose-file/#args

---

this PR unblocks #2547 - rebase it on this one and it should run successfully when run through `docker-compose exec toolchain dcos-system-test-driver /dcos-ui/system-tests/driver-config/jenkins.sh`

---

HINT: you can prevent the cluster (and tests) from being started with adding `exit 1` in the first line of `system-tests/_scripts/launch-cluster.sh` - it will fail with `Process '../_scripts/launch-cluster.sh' exited with 1` which is okay and means the config is correctly parsed.

if you see the error @nLight mentioned in #2547 it didnt work.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -6,10 +6,10 @@ services:
     build:
       context: .
     environment:
-       - CLUSTER_URL="${CLUSTER_URL}"
-       - INSTALLER_URL="${INSTALLER_URL}"
-       - AWS_ACCESS_KEY_ID=${AWS_ACCESS_KEY_ID}
-       - AWS_SECRET_ACCESS_KEY=${AWS_SECRET_ACCESS_KEY}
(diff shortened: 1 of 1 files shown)


juliangieseke:
Alfred test this please

juliangieseke:
Alfred test this please

ahoskins:
@juliangieseke I'm a bit confused: in the description it says this PR unblocks #2547, but 2547 has merged without this merging first. Is this PR still needed?

juliangieseke:
Yes, we decided that this one isnt _blocking_ #2547, because it only prevents us from running the tests through `docker-compose` - but it is still a fix we should apply.

I simply forgot to update this PR after merging the other one

PR #2559, opened 2017-12-09
DCOS-19400: Refactor - local volumes

Poltergeist:
This PR is the first of a series of refactoring efforts.

This PR mostly just changes localVolumes logic to Volumes and removes the localVolumes header

PR #2560, opened 2017-12-09
DCOS-19401: refactor(externalVolumes): add external volumes logic to volumes

Poltergeist:
Based on #2559 (MERGED)

This PR introduces the externalVolumes logic to the now called volumes reducers.

This is the second of three PR's which will merge external and local volumes

### DIFF
--- a/plugins/services/src/js/components/forms/VolumesFormSection.js
+++ b/plugins/services/src/js/components/forms/VolumesFormSection.js
@@ -59,24 +59,7 @@ class VolumesFormSection extends Component {
 
     return (
       <FormRow>
-        <FormGroup className="column-3" showError={Boolean(sizeError)}>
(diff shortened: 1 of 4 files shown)


juliangieseke:
Alfred test this

ahoskins:
@Poltergeist *Note*: this issue is not happening on #2561 👏

Poltergeist:
Yes sorry forgot to note this here. As there are two ways to handle external volumes in this PR the form will not work as expected, this is also why the test does fail. But this was impossible to avoid by keeping the PR as small as possible. So as soon as you enter the JSON editor and do something chaos breaks loose. As the parser will parse a *volume* and a *externalVolume* from one volume.

Poltergeist:
Alfred retest this please

PR #2543, opened 2017-12-06
feat(EditServiceModal): only show framework edit if hook true

ahoskins:
This hook is added because the API is EE for now. The `cosmos/service/update` API is not yet migrated.

Closes DCOS-19784.

**Checklist**
- [ ] Did you add a JIRA issue in a commit message or as part of the branch name?
- [ ] Did you add new unit tests?
- [ ] Did you add new integration tests?
- [ ] If this is a regression, did you write a test to catch this in the future?

PR #2565, opened 2017-12-11
feat(EditFrameworkConfiguration): handle empty resolvedOptions

ahoskins:
Frameworks that are deployed prior to version 1.10 of DC/OS won't have configuration information saved from `cosmos/service/describe`. This PR handles this in the UI by displaying a warning, and populating the edit flow with default package configuration.

In doing this change, I also refactor this flow a bit by adding some methods to `ServiceUtil` and cleaning up passing props.

**Testing**:
1. Toggle enterprise = true in config.dev.json
2. Since we don't have access to old running services, and I don't want to change the fixture to be this case, manually paste this code into **line 44 of src/js/pages/EditFrameworkConfiguration**:
```
fullPackage.resolvedOptions = null;
```
4. Click the edit on any running framework
5. The form will show the default framework configuration values, and show the warning message.

Closes DCOS-19688

**Checklist**

### DIFF
--- a/plugins/services/src/js/pages/EditFrameworkConfiguration.js
+++ b/plugins/services/src/js/pages/EditFrameworkConfiguration.js
@@ -8,6 +8,7 @@ import UniversePackage from "#SRC/js/structs/UniversePackage";
 import FrameworkConfiguration from "#SRC/js/components/FrameworkConfiguration";
 import Loader from "#SRC/js/components/Loader";
 import RequestErrorMsg from "#SRC/js/components/RequestErrorMsg";
+import { getDefaultFormState } from "react-jsonschema-form/lib/utils";
 
 class EditFrameworkConfiguration extends mixin(StoreMixin) {
   constructor(props) {
@@ -18,7 +19,8 @@ class EditFrameworkConfiguration extends mixin(StoreMixin) {
       deployErrors: null,
       formData: null,
       formErrors: {},
-      hasError: false
+      hasError: false,
+      defaultConfigWarning: null
     };
 
     this.store_listeners = [
@@ -40,13 +42,18 @@ class EditFrameworkConfiguration extends mixin(StoreMixin) {
     const fullPackage = CosmosPackagesStore.getServiceDetails();
     const packageDetails = new UniversePackage(fullPackage.package);
 
-    // necessary to have review screen same order of tabs as the form
-    const formData = this.reorderResolvedOptions(
+    const schema = packageDetails.getConfig();
+    const formData = getDefaultFormState(
+      schema,
       fullPackage.resolvedOptions,
-      packageDetails.getConfig()
+      schema.definitions
     );
 
-    this.setState({ packageDetails, formData });
+    const defaultConfigWarning = fullPackage.resolvedOptions
+      ? null
+      : "This service was initially deployed to a previous version of DC/OS that did not store service configuration settings. The default package values were used to populate the configuration in this form. Carefully verify the default settings are correct, prior to deploying the new configuration.";
+
+    this.setState({ packageDetails, formData, defaultConfigWarning });
   }
 
   onCosmosPackagesStoreServiceDescriptionError() {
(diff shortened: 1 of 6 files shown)


ahoskins:
I didn't add tests for this because we temporarily [removed](https://github.com/dcos/dcos-ui/pull/2543/files) these cypress tests due to the feature being disabled in open (because API not migrated yet).

ahoskins:
Alfred test this please

src/js/utils/ServiceUtil.js
```diff
@@ -17,6 +19,39 @@ const ServiceUtil = {
     const labels = service.getLabels();
 
     return labels.DCOS_PACKAGE_NAME != null;
+  },
+
+  getFrameworkConfigurationFromSchema(config) {
```
MatApple: @ahoskins we need unit tests for these new methods :)

src/js/components/FrameworkConfiguration.js
```diff
@@ -279,32 +297,23 @@ class FrameworkConfiguration extends Component {
...
-          formData={formData}
-          formErrors={formErrors}
           focusField={focusField}
           activeTab={activeTab}
-          deployErrors={deployErrors}
-          onFormDataChange={onFormDataChange}
-          onFormErrorChange={onFormErrorChange}
           handleActiveTabChange={this.handleActiveTabChange}
           handleFocusFieldChange={this.handleFocusFieldChange}
+          {...Util.omit(this.props, [
```
MatApple: Let's please avoid this where possible - it's much more clear what is happening when it is explicit like before.

ahoskins:
@mesosphere-ci ok to test

ahoskins:
Jira task for adding these tests once this feature is re-enabled: https://jira.mesosphere.com/browse/DCOS-19930

ahoskins:
I've addressed the comments from @MatApple removing the `Util.omit()`. In looking more closely at the mozilla lib, I also noticed that I can directly call the *very robust* [getDefaultFormState](https://github.com/mozilla-services/react-jsonschema-form/blob/77704620c3c9b80e75ac296182814ef53c6f3831/src/utils.js#L167), instead of the form data init methods previously had in ServiceUtil. This `getDefaultFormState` is the same thing called when the form [updates](https://github.com/mozilla-services/react-jsonschema-form/blob/132a6cce1fc94c09308250a379c0220b94d84ec8/src/components/Form.js#L41) to reconcile the current data This is a lot nicer because it avoids us reimplementing this code.

ahoskins:
Holding merge till #2570 is merged because it re-adds the tests and the ability to edit frameworks.

ahoskins:
@orlandohohmeier or @Poltergeist: I want to add integration tests to `ServiceActions-cy` to test this warning message appearing. But I'm not sure how to do this because the `CosmosPackagesActions` fixtures are setup in a way to always return the same serviceDescribeFixture. In the integration tests, I don't really have the ability to modify these fixtures unless I do something like this:

```   global.actionTypes.CosmosPackagesActions.fetchServiceDescription.success.response.resolvedOptions = null
```

MatApple:
@ahoskins I ran out of time to thoroughly review this again but I see you've made adjustments so feel free to dismiss my review :)

ahoskins:
Alfred test this please

ahoskins:
I'm not sure how to add tests for this right now, but let's merge this before wednesday. @weblancaster @bstavroulakis @juliangieseke can I get two reviews please? 👍

plugins/services/src/js/pages/EditFrameworkConfiguration.js
```diff
@@ -40,13 +42,18 @@ class EditFrameworkConfiguration extends mixin(StoreMixin) {
...
-      packageDetails.getConfig()
+      schema.definitions
     );
 
-    this.setState({ packageDetails, formData });
+    const defaultConfigWarning = fullPackage.resolvedOptions
+      ? null
+      : "This service was initially deployed to a previous version of DC/OS that did not store service configuration settings. The default package values were used to populate the configuration in this form. Carefully verify the default settings are correct, prior to deploying the new configuration.";
+
+    this.setState({ packageDetails, formData, defaultConfigWarning });
```
orlandohohmeier: 🤔  Why do we need to store the entire message in the component state? Wouldn't a simple flag be enough?

ahoskins: I did this so the `FrameworkConfigurationForm` could stay more general. The `EditFrameworkConfiguration` handles the details of the cosmos API, so I thought it made sense for this component to set the message.

PR #2569, opened 2017-12-13
fix(Jenkinsfile): add 1Gb more memory

nLight:
Pipeline is failing with OOM


nLight:
Alfred test this

juliangieseke:
So is this needed / does it help?

PR #2574, opened 2017-12-15
refactor(PlacementConstraintsFields) Improve advanced constraints

weblancaster:
Improve advanced constraints with minor UX/design changes and refactor the code to
a new file that way making easier to understand what is going on.

Closes DCOS-19719

**Checklist**

### DIFF
--- a/plugins/services/src/js/constants/OperatorTypes.js
+++ b/plugins/services/src/js/constants/OperatorTypes.js
@@ -1,38 +1,55 @@
+const DEFAULT_HELP = "A string, integer or regex value.";
+const OPTIONAL_HELP = `${DEFAULT_HELP} This field is optional.`;
+
 const OperatorTypes = {
   UNIQUE: {
     requiresValue: false,
     requiresEmptyValue: true,
-    stringNumberValue: false
+    stringNumberValue: false,
+    tooltipContent: "The unique operator does not accept a value.",
+    helpContent: DEFAULT_HELP
   },
   CLUSTER: {
     requiresValue: true,
     requiresEmptyValue: false,
-    stringNumberValue: false
+    stringNumberValue: false,
+    tooltipContent: null,
+    helpContent: DEFAULT_HELP
   },
   GROUP_BY: {
     requiresValue: false,
     requiresEmptyValue: false,
-    stringNumberValue: true
+    stringNumberValue: true,
+    tooltipContent: null,
+    helpContent: OPTIONAL_HELP
   },
   IS: {
     requiresValue: true,
     requiresEmptyValue: false,
-    stringNumberValue: false
+    stringNumberValue: false,
+    tooltipContent: null,
+    helpContent: DEFAULT_HELP
   },
   LIKE: {
     requiresValue: true,
     requiresEmptyValue: false,
-    stringNumberValue: false
+    stringNumberValue: false,
+    tooltipContent: null,
+    helpContent: DEFAULT_HELP
   },
   UNLIKE: {
     requiresValue: true,
     requiresEmptyValue: false,
-    stringNumberValue: false
+    stringNumberValue: false,
+    tooltipContent: null,
+    helpContent: DEFAULT_HELP
   },
   MAX_PER: {
     requiresValue: true,
     requiresEmptyValue: false,
-    stringNumberValue: true
+    stringNumberValue: true,
+    tooltipContent: null,
+    helpContent: DEFAULT_HELP
   }
 };
 
--- a/src/js/components/PlacementConstraintsPartial.js
+++ b/src/js/components/PlacementConstraintsPartial.js
@@ -14,53 +14,30 @@ import FormGroupHeading from "#SRC/js/components/form/FormGroupHeading";
 import FormGroupHeadingContent
   from "#SRC/js/components/form/FormGroupHeadingContent";
 import FormRow from "#SRC/js/components/form/FormRow";
-import Icon from "#SRC/js/components/Icon";
 
 import OperatorTypes from "#PLUGINS/services/src/js/constants/OperatorTypes";
 import PlacementConstraintsUtil
   from "#PLUGINS/services/src/js/utils/PlacementConstraintsUtil";
 
-function placementConstraintLabel(name, tooltipText, options = {}) {
-  const { isRequired = false, linkText = "More information" } = options;
-
-  const tooltipContent = (
-    <span>
-      {`${tooltipText} `}
-      <a
-        href="https://mesosphere.github.io/marathon/docs/constraints.html"
-        target="_blank"
-      >
-        {linkText}
-      </a>.
-    </span>
-  );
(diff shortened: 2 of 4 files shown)


src/js/components/PlacementConstraintsPartial.js
```diff
@@ -102,26 +78,21 @@ export default class PlacementSection extends Component {
...
+        operatorLabel = this.getPlacementConstraintLabel("Operator");
+        valueLabel = this.getPlacementConstraintLabel("Value");
       }
 
+      const fieldValue = (
+        <FieldInput
+          name={`constraints.${index}.value`}
+          type="text"
+          value={constraint.value}
+          disabled={!!valueIsRequiredEmpty}
```
ahoskins: The type coercion is unnecessary because this value comes from `requiresEmptyValue()` which always returns boolean.

src/js/components/PlacementConstraintsPartial.js
```diff
@@ -136,8 +107,17 @@ export default class PlacementSection extends Component {
...
+              {Object.keys(OperatorTypes).map((type, index) => {
+                return (
+                  <option key={index} value={type}>
+                    {type}
+                  </option>
+                );
+              })}
             </FieldSelect>
+            <FieldHelp>
+              Specify where your app will run
```
ahoskins: Please add a period here, sir. Or get rid of the period on "A string, integer or regex value."

src/js/components/PlacementConstraintsPartial.js
weblancaster: Not always boolean, during my tests I remember where was `undefined` for that reason because of the `disabled` attribute I decided to keep it safe.

ahoskins: Okay, cool makes sense 👍

plugins/services/src/js/constants/OperatorTypes.js
```diff
@@ -1,38 +1,55 @@
+const DEFAULT_HELP = "A string, integer or regex value.";
+const OPTIONAL_HELP = `${DEFAULT_HELP} This field is optional.`;
```
juliangieseke: I'm not sure if we should keep these two texts here, you also have it as default below and we can calculate if a field ist optional or not 🤔 

I will add texts here as well in my PR and I already felt bad adding these, IMHO this file should be as clean/small as possible.

src/js/components/PlacementConstraintsPartial.js
```diff
@@ -152,27 +129,27 @@ export default class PlacementSection extends Component {
...
-            >
-              This field is optional
+            {valueIsRequiredEmpty
+              ? this.getToolTip(
+                  fieldValue,
+                  OperatorTypes[constraint.operator].tooltipContent
+                )
+              : fieldValue}
+            <FieldHelp>
+              {OperatorTypes[constraint.operator]
```
juliangieseke: related to my comment above, shouldnt it be possible to write something like this?
```
{OperatorTypes[constraint.operator].requiresValue ||
  OperatorTypes[constraint.operator].requiresValueEmpty 
  ? "A string, integer or regex value." 
  : "A string, integer or regex value. This field is optional."
}
```
maybe we can also include this part into `fieldValue`

plugins/services/src/js/constants/OperatorTypes.js
```diff
@@ -1,38 +1,55 @@
...
+
 const OperatorTypes = {
   UNIQUE: {
     requiresValue: false,
     requiresEmptyValue: true,
-    stringNumberValue: false
+    stringNumberValue: false,
+    tooltipContent: "The unique operator does not accept a value.",
+    helpContent: DEFAULT_HELP
```
juliangieseke: see above, I'd remove this key from the object.

plugins/services/src/js/constants/OperatorTypes.js
```diff
@@ -1,38 +1,55 @@
+const DEFAULT_HELP = "A string, integer or regex value.";
+const OPTIONAL_HELP = `${DEFAULT_HELP} This field is optional.`;
+
 const OperatorTypes = {
   UNIQUE: {
     requiresValue: false,
     requiresEmptyValue: true,
-    stringNumberValue: false
+    stringNumberValue: false,
+    tooltipContent: "The unique operator does not accept a value.",
```
juliangieseke: same thing here, is it possible to calculate this based on `requiresEmptyValue`?

src/js/components/PlacementConstraintsPartial.js
```diff
@@ -152,27 +129,27 @@ export default class PlacementSection extends Component {
...
-              value={constraint.value}
-            />
-            <FieldHelp
-              className={{ hidden: valueIsRequired || valueIsRequiredEmpty }}
-            >
-              This field is optional
+            {valueIsRequiredEmpty
+              ? this.getToolTip(
+                  fieldValue,
+                  OperatorTypes[constraint.operator].tooltipContent
```
juliangieseke: As I said above, maybe something like
`The ${constraint.operator} operator does not accept a value.`?

weblancaster:
@juliangieseke I understand where you coming from I had the same idea at first, but as I implemented more and more conditional checks in the component UI the worse felt, so having a constant file (in this case `OperatorTypes.js`) that way every operator is flexible and know it's own requirements instead of keep adding conditions in the UI component layer, does that makes sense?

One question that I asked that made me choose the constant over conditions in the component was. What if we need to add more operator specific rules/requirements? 

So having a constant knowing the rules instead the component having conditions and hard-coded rules felt better.

I'm open for more suggestions though.

tests/pages/catalog/packages/package/PackageTab-cy.js
```diff
@@ -226,7 +226,7 @@ describe("Package Detail Tab", function() {
             .should("not.exist");
         });
 
-        it('Should hide the "value" when "Unique" is selected in operator dropdown', function() {
+        it("Should disable the field value when Unique is selected in operator dropdown", function() {
```
bstavroulakis: Not important, since we've edited this, we could remove the should.

tests/pages/services/ServiceFormModal-cy.js
```diff
@@ -815,7 +815,7 @@ describe("Service Form Modal", function() {
             .should("not.exist");
         });
 
-        it('Should hide the "value" when "Unique" is selected in operator dropdown', function() {
+        it("Should disable the field value when Unique is selected in operator dropdown", function() {
```
bstavroulakis: Not important, since we've edited this, we could remove the should.

tests/pages/catalog/packages/package/PackageTab-cy.js
weblancaster: Agree, I just didn't remove because I would have to remove for all of the tests and didn't want to create more noise in the PR

PR #2575, opened 2017-12-18
DCOS-19140: DSL Tasks Filters Integration Tests

bstavroulakis:
Integration tests for DSL task filters.

This is dependent on https://github.com/dcos/dcos-ui/pull/2472

**Checklist**

### DIFF
--- a/tests/pages/services/TasksTable-cy.js
+++ b/tests/pages/services/TasksTable-cy.js
@@ -144,6 +144,47 @@ describe("Tasks Table", function() {
     });
   });
 
+  context("Task DSL filters", function() {
+    beforeEach(function() {
+      cy.configureCluster({
+        mesos: "1-service-with-executor-task"
+      });
+      cy.visitUrl({ url: "/services/detail/%2Fcassandra/tasks?_k=rh67gf" });
+    });
+
+    it("Has by default the is:active filter", function() {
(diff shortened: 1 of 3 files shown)


ahoskins:
Alfred test this please

tests/pages/services/TasksTable-cy.js
```diff
@@ -144,6 +144,47 @@ describe("Tasks Table", function() {
...
+      cy.get(".form-control.filter-input-text").clear();
+      cy.get(".form-control.filter-input-text").type("region:eu-central-1");
+      cy.get("table tr td.task-table-column-name").should("to.have.length", 1);
+      cy.get("td.task-table-column-region-address").contains("eu-central-1");
+    });
+
+    it("Filters tasks by zone", function() {
+      cy.get(".form-control.filter-input-text").clear();
+      cy.get(".form-control.filter-input-text").type("zone:eu-central-1b");
+      cy.get("table tr td.task-table-column-name").should("to.have.length", 1);
```
weblancaster: Can't we just use `task-table-column-name` here? I feel like this is over-specifying.

ps: same goes for the rest

ahoskins: Don't we love over-specifying. I've had to change some cypress tests a few weeks ago because they too unspecific and when content was added to the page they failed.

So I say keep it specific ✋

PR #2588, opened 2017-12-27
fix(Jenkinsfile): adjust timeout

orlandohohmeier:
Some of our jobs take rather long to complete which esp. in combination with retires leads to false negatives. This PR increase the timeout to reduce the number of false negatives due to timeouts. I thought about removing the timeout but decided that we should always investigate what happened if a pipeline runs for longer than 3 hours and thus kept the timeout. The timeout also saves us form rouge builds that run for ever, an issue other teams ran into.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -12,7 +12,7 @@ pipeline {
   }
 
   options {
-    timeout(time: 1, unit: 'HOURS')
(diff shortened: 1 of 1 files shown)


weblancaster:
We should look into why our tests are taking so long to finish.

PR #2589, opened 2017-12-27
Update Application Preloader Spinner Image

ashenden:
- Add 2dpi image fo Applicatino Preloader Spinner for retina-frienly monitors.


src/js/config/Config.js
```diff
@@ -5,7 +5,7 @@ import ConfigTest from "./Config.test.js";
 var Config = {
   analyticsKey: "51ybGTeFEFU1xo6u10XMDrr6kATFyRyh",
   acsAPIPrefix: "/acs/api/v1",
-  applicationRenderDelay: 1000,
+  applicationRenderDelay: 2000,
```
orlandohohmeier: ☝️  Loading the UI already takes quite some time as many of the assets and JavaScript resources are rather large, which is bad for the user experience as studies conducted by Google, Amazon, Facebook and the likes have shown. Adding any delay to the already long loading time is counterproductive and we should try remove the need for any loading screen instead of investing time to implement a nice screen and further  increasing the delay.

src/styles/components/application-preloader/styles.less
```diff
@@ -65,3 +65,16 @@
...
 }
+
+& when (@screen-resolution-2x-enabled) {
+
+  @media only screen and (min-resolution: 2dppx) {
+
+    & when (@application-preloader-enabled) {
+
+      .application-preloader-spinner {
+        background-image: url('../../../img/components/application-preloader/application-preloader-spinner@2x.png');
```
orlandohohmeier: 🤔  Why can't we use the svg instead of adding a 2dppx spinner image? Please keep in mind that people use the UI on there mobile phones (many of which feature retina screens) and this image unnecessarily increases the load time.

ashenden: We can't generate an angular gradient using web-friendly SVGs :/ Unless you know how? So, we're left with a PNG.  That said, I ran the images through ImageOptim and was able to shave off ~70% of the original file size.

src/js/config/Config.js
ashenden: I agree that we should aim for the fastest possible experience.  With that said, if we are going to display a preloader prior to rendering the login screen, having the logo flash for a split second, then disappear looks strange -- almost broken.  The render delay was added specifically to solve for this problem. Two seconds comes and goes in the blink of an eye, and it's just enough time for the preloader to justify it's existence as both a bridge and a branding mechanism.

If we can gaurantee the application will always loading in less than 2 seconds, than I'd sooner pull the preloader out entirely.  If we can't, then we need a preloader, and we should define a minimum display.

juliangieseke: It looks like something else is wrong on master, the first request for the spinner.png receives a 404 because its not properly included, the second insertion of the spinner happens after everything is loaded and would be ready to render. that being the reason for the spinner to be displayed too short, the supposed fix would make things even worse.

instead of increasing the delay, we need fo fix it on master, then the spinner should be displayed for ~2secs anyway, because it would be displayed while stuff is loading - not after the loading process is done. 👍

src/styles/components/application-preloader/styles.less
juliangieseke: it looks like there is a way to fake it: http://wiki.inkscape.org/wiki/index.php/Advanced_Gradients#Conical_gradient

IMHO we should try to insert it as SVG instead of a referencing an external image.

src/js/config/Config.js
orlandohohmeier: @ashenden I understand your argument and agree that having the logo flash does indeed look strange but would argue that a "preloader" shouldn't be used as a bridge or branding mechanism. A "preloader" is only meant to indicate that something (loading) is happening as we otherwise risk that people assume that nothing is happening. 

I'd only display a spinner and optimize the UI to reduce the load time to a few milliseconds. What do you think?

src/styles/components/application-preloader/styles.less
orlandohohmeier: @juliangieseke I think the file is small enough and might even perform better (rendering) than a _faked_ conical gradient.

juliangieseke: I was thinking about the extra request - so yes, use whatever is smaller in size and, if its the png, include id as base64 in the HTML. 👍

src/img/components/application-preloader/application-preloader-spinner.svg
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <svg width="100px" height="100px" viewBox="0 0 100 100" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
-    <!-- Generator: Sketch 48.1 (47250) - http://www.bohemiancoding.com/sketch -->
+    <!-- Generator: Sketch 48.2 (47327) - http://www.bohemiancoding.com/sketch -->
```
Poltergeist: Could we remove the unnecessary meta information sketch drops in those files on export?

src/js/config/Config.js
nLight: If showing a logo for a split second feels broken I'd opt in to remove the logo and show only the spinner. How'd you feel finding out that apple engineers have a 2 sec delay for no other reason but showing the logo in your iphone? :)

ashenden: Guys, let's not be dramatic here. You're all acting like a 2second delay is going to cause our customers to leave DC/OS.  This isn't a site that needs near instantaneous load time for it maintain customer retention and value.

FWIW, I'd be willing to bet that Apple bakes in a delay to it's iPhone and mac loading screens to ensure you see their logo for a couple seconds. 

I have a proposed compromise.  Let's lose the Mesosphere logo mark, and go with just the spinner -- limiting the delay to `1000ms`.  I do need one of you to look into why the image is error'ing and why it is taking so long to render in the DOM.

juliangieseke: I have fixed the 404 (using base64 inline code) and now the spinner is displayed for around 2.5 seconds (thats how long it takes to load our UI locally 🙀) - IMHO in this case your delay is okay (2s _overall_ loading time is okay, but _increasing_ our already long time with 2 _addidtional_ seconds was too much - and it tried to fix the wrong thing)

orlandohohmeier: @ashenden I agree that we probably don't need near-instantaneous load time to maintain customer retention and value, but we should, by all means, consider performance both during development as well as design. 

> Page load time and how fast your site feels is a large part of this user experience and should be weighed equally with the aesthetics of your site. 

/source  Designing for Performance by Lara Callender Hogan

My quick search for "Apple Logo Delay" seconds that, as it resulted in a long list people wondering whether their device is broken only because it took a bit longer to "load" than usual.

Showing only the spinner and limiting the delay to one second sound like a good compromise. Do you have the bandwidth to adjust this PR accordingly? Otherwise I'm more than happy to take care.

ashenden: if you could handle it, it would get done more quickly.

src/styles/components/application-preloader/styles.less
```diff
@@ -53,7 +53,8 @@
...
...ixgXpwgsjSvkkU7aMYfjMo3JRFiXEV/EZkeNeO9Iigx23APRtFf9CBfmGwnE+p+QhrYz4NioXZRXflM8IHvU5Kk+J+/7qcNsqr40om0etyj1lwvdRuShxFeM7Hh928ORp7gaXtp22fjcM7PcHp5BSVts2KhflvoT4IbL7v13pIG6q2+6L+w/l1X5Hs0lDV1tH5W5W+m5ECZsHUCqGhbcrF+URfior+jdndRArW+wuSMYIUjf37RMtVg6gXJAp/FLe6dV5XMWIW9tIdBuO340ofa1bkuPKmrVRuSgvjSATrQt6XJm3Oip3mvhbWQ/sVcGOCmJvu8QFyRpB8gG56XJHGbQ2IhdkDX+UuwG5DW1bBQlbG5ELMmkEebb3iyvYVoLWRuSCDBtB3sjLRgWxucPrgsTwV/m494vr2MK29bufXJCgEeR7QGIolk8I3eLwn/Lnf3tngeNQDMTQMjMzw/1PuOgvWbOMbvtd6R2g85QMNE0wNiHkQmzpxEL0Y3cL2bMQcVL3B33ImSg7KlohaxbixlAvZA4RJ9ByVPSXjJ2IgaOiFdJmIf7bweVcTnkEa0dF3y0eicNl/ITr0vdAuNISCxkHIX1HRT8GZiELR0UrpAARe0KcRyxlFYQ0HRX9VRA7QtyPWEgJIpiCI6OVMicZWx8JuowxypZY+RdEfde+ISEbJ/fLSO4bYqFfJe5JNoG6I6OVMoCINZjrV4n/lLgmVq649FJ6JAOo32DyvRyLIGTkyGil1EnGErgMFksZJTLAwofp9GXwIpEBxC+sWUoVIpiBS2GtlA7JmIOuVorvDBwlMsAMdz35Ixw+TkgGcOWlTvITFoIk3/L2pR2tTBMZRMdStA/VT0jGBPS0JbFXyhgimKHnXtqcMiQZYzDykVRt9dVnGUTHp1d0fUrzFSFDUHfC1yX74StCBkj45Wt54P7WtrAuywh0ca1E9hJFJNzqXR79IKMPesgvVXWZHETkwG1KwZdrvCaE6CL/lP4zCCQgD3JSIYJtrBFlgC7ogCZWVwFB+QsBBYKEKKToxdRfk0G0QQs8iqyiICiGAGYDOcCBL4ISKIJ3pchlCLayCgL+rhDQBA2iDmqgCipEGZQ+KQSrwo1lFYFXCMm/L8HHj0oIcPOPhBR+tiV5ApB/oIjAVklApBolUM65CgH3I73MdzrpM08AAAAASUVORK5CYII=');
```
orlandohohmeier: We should use the webpack url-loader instead of manually converting this one into a base64 string, to make sure that we always display the correct spinner and don't forget to update this one when we get new assets.

src/styles/components/application-preloader/styles.less
```diff
@@ -65,3 +66,17 @@
...
   }
 }
+
+& when (@screen-resolution-2x-enabled) {
+
+  @media only screen and (min-resolution: 2dppx) {
+
+    & when (@application-preloader-enabled) {
+
+      .application-preloader-spinner {
```
orlandohohmeier: We shouldn't include the 2x version as a base64 string as we're currently only generating one css file for all resolutions meaning we would always load both versions of the spinner even if only one is needed.

src/js/config/Config.js
orlandohohmeier: Alright, I'll look into it.

src/styles/components/application-preloader/styles.less
juliangieseke: removed

juliangieseke:
I had another look into it: The 404 happens because we have a script which fetches the `index.less` file and removes everything which isnt needed on the first pageload. since this script isnt compatible with our webpack setup, all referenced images are not valid once webpack has bundled our code. 

I (re)added the base64 version, also I removed the logo and reduced the delay back to one second.

nLight:
Alfred test this

src/styles/components/application-preloader/styles.less
juliangieseke: with our current seupt it is not (easily) possible to parse the CSS for our bootstrap-page, this is the reason why referencing images will caus a 404. including the images as base64 isnt the best approach, but it works for now. JIRA: https://jira.mesosphere.com/browse/DCOS_OSS-1564

src/styles/components/application-preloader/styles.less
```diff
@@ -53,7 +53,8 @@
 
   .application-preloader-spinner {
     animation: application-preloader-spinner 1s linear infinite;
-    background-image: url('../../../img/components/application-preloader/application-preloader-spinner.png');
```
DanielMSchmidt: Sad that we don't use postcss, there are nice ways to inline images on compile time. 🤷‍♂️

juliangieseke:
Alfred test this

DanielMSchmidt:
Alfred test this please :)

DanielMSchmidt:
Alfred test this please :)

bstavroulakis:
@ashenden This probably needs a rebase for the tests to pass.

bstavroulakis:
run integration tests

PR #2595, opened 2018-01-02
fix(volumes): fix any volume to external conversion

Poltergeist:
This makes sure that the right external fields are getting set after a JSON has been parsed.

To test create a service with a local volume (persistent, dss, host) and then edit something in the
JSON editor or edit a application with a local volume and try to change that volume to a external
volume. The JSON should now consist of the provider information.


orlandohohmeier:
LGTM! 🚢

PR #2596, opened 2018-01-02
refactor(launch-cluster): adjust scrip to improve compatibility

orlandohohmeier:
---
⚠️  _This PR includea changes that will be introduced with #2594._ 

---

Adjust the python call to output the "public_ip" to use `sys.stdout` instead of print to make the call compatible with both python2 as well as python3


nLight:
Alfred test this

PR #2599, opened 2018-01-03
fix(volumes): fix parser to ignore unknown volumes

Poltergeist:
This fixes the JSON parser for volumes to ignore volumes of type unknown.

To test use this JSON

```
{
  "id": "/",
  "instances": 1,
  "portDefinitions": [],
  "container": {
    "type": "DOCKER",
    "volumes": [
    {
      "hostPath": "s",
      "mode": "RW"
    },
    {
      "mode": "RW"
    }
    ]
  },
  "cpus": 0.1,
  "mem": 128,
  "requirePorts": false,
  "networks": [],
  "healthChecks": [],
  "fetch": [],
  "constraints": []
}
```

If you edit the JSON and then the form and then the JSON again, only one volume should exist in the form.

### DIFF
--- a/plugins/services/src/js/reducers/serviceForm/JSONReducers/Volumes.js
+++ b/plugins/services/src/js/reducers/serviceForm/JSONReducers/Volumes.js
@@ -179,8 +179,17 @@ module.exports = {
       return [];
     }
 
-    return state.container.volumes.reduce(function(memo, item, index) {
-      /**
+    return state.container.volumes
+      .filter(function(item) {
+        return (
+          item.persistent != null ||
+          item.external != null ||
+          item.hostPath != null ||
+          item.mode != null
+        );
+      })
+      .reduce(function(memo, item, index) {
+        /**
          * For the volumes we have a special case as all the volumes
          * are present in the `container.volumes` But in this parser we only
          * want to parse the local volumes. which means that we first filter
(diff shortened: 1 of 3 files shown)


plugins/services/src/js/reducers/serviceForm/JSONReducers/Volumes.js
```diff
@@ -179,8 +179,16 @@ module.exports = {
       return [];
     }
 
-    return state.container.volumes.reduce(function(memo, item, index) {
-      /**
+    return state.container.volumes
+      .filter(function(item) {
```
Poltergeist: This is the only thing changed in the file the rest is just indentation.

plugins/services/src/js/reducers/serviceForm/JSONReducers/Volumes.js
```diff
@@ -179,8 +179,16 @@ module.exports = {
...
+        return (
```
Fabs: It is not clear to me that this function excludes unknown volumes. Especially because I see 4 types of volumes, and 3 conditions. Persistent Volume is persistent. External Volume is external, so I am left to assume that both DC/OS Storage and Host Volume have a hostPath. 

Maybe a comment that explains why this is enough to define "known volume" would be helpful.

juliangieseke:
Alfred please test this

plugins/services/src/js/reducers/serviceForm/JSONReducers/Volumes.js
Poltergeist: So DCOS storage volumes are also persistent volumes. :-) 
Host volumes do not have specific field except hostPath

For a better overview you can look at https://github.com/mesosphere/marathon/blob/master/docs/docs/rest-api/public/api/v2/types/volumes.raml

orlandohohmeier:
LGTM! 💯

juliangieseke:
The flaky one again - Alfred test this please

weblancaster:
two ✅  plus alfred 💚  so merging this PR.

PR #2602, opened 2018-01-03
fix(package.json): use github dep for eslint-config-dcos

nLight:
The package eslint-config-dcos is not on npm. So we can't use plain version.
It went undetected because `npm install` relies on npm-shrinkwrap.json

Introduced https://github.com/dcos/dcos-ui/commit/d7e679f3df5a69621d71d9baac5f92d50e4b45e8#diff-b9cfc7f2cdf78a7f4b91a753d10865a2L88

### DIFF
--- a/package.json
+++ b/package.json
@@ -88,7 +88,7 @@
     "conventional-changelog-lint-config-angular": "0.4.1",
     "css-loader": "0.23.1",
     "eslint": "3.15.0",
-    "eslint-config-dcos": "0.2.0",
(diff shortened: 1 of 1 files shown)


nLight:
@juliangieseke eslint plugins must follow a naming convention https://eslint.org/docs/developer-guide/working-with-plugins so what we could have would be `@dcos/eslint-config-dcos` :) I had an idea of moving the repo in `dcos-labs`

juliangieseke:
`@dcos/eslint-config-dcos` is fine :)

PR #2562, opened 2017-12-11
DCOS-19658: Adjust system-tests discrepancies

nLight:
System tests have diverged, let's fix them 💪

PR #2604, opened 2018-01-03
fix(PodEnvironmentVariablesConfigSection): fix props

nLight:
This PR fixes Pod Environment variable config section.

1. Create a pod with an env var
2. Click Review and run
3. Without the fix you should get an exception

extracted from #2562


orlandohohmeier:
LGTM 💯 ✨

PR #2611, opened 2018-01-05
DCOS_OSS-2011: Add IPv6 to the networks table

juliangieseke:
this adds the functionality to display ipv6 subnets in
the networks overview table

networks now are able to have IPv4 and/or IPV6 subnets, so we need to display both. 

To test this you can turn on fixtures, I've added an IPv6 subnet to one of them.

### DIFF
--- a/src/js/pages/network/VirtualNetworksTable.js
+++ b/src/js/pages/network/VirtualNetworksTable.js
@@ -37,7 +37,15 @@ class VirtualNetworksTable extends React.Component {
       {
         className: getClassName,
         getValue(overlay) {
-          return overlay.getSubnet();
+          if (overlay.getSubnet() && overlay.getSubnet6()) {
+            return (
+              <span>
+                IPv4: {overlay.getSubnet()}<br />IPv6: {overlay.getSubnet6()}
+              </span>
+            );
+          }
+
+          return overlay.getSubnet() || overlay.getSubnet6();
         },
         headerClassName: getClassName,
         heading,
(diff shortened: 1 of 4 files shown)


juliangieseke:
There is a small design glitch when having IPv4 & IPv6, I opened a ticket against CNVS to get this fixed.
https://github.com/mesosphere/cnvs/issues/116

juliangieseke:
Alfred test this

src/js/pages/network/VirtualNetworksTable.js
```diff
@@ -37,7 +37,15 @@ class VirtualNetworksTable extends React.Component {
       {
         className: getClassName,
         getValue(overlay) {
-          return overlay.getSubnet();
+          if (overlay.getSubnet() && overlay.getSubnet6()) {
```
Fabs: I was wondering if there is a simple way to reduce this duplication. My idea is a little contrived because the API of the table expects a function, but I decided to share anyway to capture your opinion.

Having the function:
```js
const ipDescriptionForVersions = (property)  => {
  if (property.ipv4 && property.ipv6) {
    return (
      <span>
         IPv4: {property.ipv4}<br />IPv6: {property.ipv6}
       </span>
      );
  }
  return property.ipv4 || property.ipv6
}
```
Then it could be used like this:

```js
getValue(overlay){
  const subnet = {ipv4: overlay.getSubnet(), ipv6: ov overlay.getSubnet6()} 
  return ipDescriptionForProperty(subnet);
}

#and

getValue(overlay){
  const prefix = {ipv4: overlay.getPrefix(), ipv6: ov overlay.getSubnet6()}
  return ipDescriptionForProperty(prefix);
}
```

nLight: I'd keep the duplication 🙈

nLight:
Alfred test this

src/js/pages/network/VirtualNetworksTable.js
juliangieseke: I'm with Dima in this case, sometimes KISS is better then DRY. 

IMHO there only is a benefit if we are able to move such a helper into an util, but just moving code inside the file into a different function makes it harder to read and not always has a benefit (it has, if the original function gets too big - which isnt the case here).

Fabs: I see your point @nLight and @juliangieseke.

juliangieseke:
✅

PR #2614, opened 2018-01-08
DCOS-20051: fixes the infinite increase of CPU usage

Fabs:
Fixes [DCOS-20051](https://jira.mesosphere.com/browse/DCOS-20051) by ignoring completed tasks when asking for the Host resources grouped by Framework.

**Need help**
- There are two uses of `getHostResourcesByFramework`. IMHO with this function name, it should never return results including resources that are completed (TASK_FINISHED, TASK_FAILED), but I am not sure that the second usage at [MesosStateStore](https://github.com/dcos/dcos-ui/blob/8d619d24b32e06d8b5e734fb31d5aed1142436cf/src/js/stores/MesosStateStore.js#L90) relies on that behavior. No unit tests failed because of that anyway.

### DIFF
--- a/src/js/utils/__tests__/MesosStateUtil-test.js
+++ b/src/js/utils/__tests__/MesosStateUtil-test.js
@@ -307,4 +307,130 @@ describe("MesosStateUtil", function() {
       });
     });
   });
+
+  describe("#getHostResourcesByFramework", function() {
+    beforeEach(function() {
+      this.mesosState = {
+        frameworks: [
+          {
+            name: "marathon",
+            id: "marathon_1"
+          },
+          {
+            name: "spark",
+            id: "spark_1"
+          }
+        ],
+        tasks: [
+          {
+            name: "spark",
+            framework_id: "marathon_1",
+            id: "spark.1",
+            slave_id: "slave-uid",
+            resources: { cpus: 1, mem: 128, disk: 100 }
+          },
+          {
+            name: "spark",
+            framework_id: "marathon_1",
+            id: "spark.2",
+            slave_id: "slave-uid",
+            resources: { cpus: 1, mem: 128, disk: 100 }
+          },
+          {
+            name: "alpha",
+            framework_id: "marathon_2",
+            id: "alpha.2",
+            slave_id: "slave-uid",
+            resources: { cpus: 0.5, mem: 256, disk: 150 }
+          },
+          {
+            name: "nginx",
+            framework_id: "marathon_3",
+            id: "nginx.1",
+            slave_id: "slave-uid",
+            resources: { cpus: 1, mem: 128, disk: 100 }
+          }
+        ]
+      };
+    });
+
+    it("aggregates resources by framework", function() {
+      expect(
+        MesosStateUtil.getHostResourcesByFramework(this.mesosState)
+      ).toEqual({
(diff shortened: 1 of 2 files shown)


Fabs:
Alfred test this

src/js/utils/__tests__/MesosStateUtil-test.js
```diff
@@ -296,4 +296,130 @@ describe("MesosStateUtil", function() {
       });
     });
   });
+
+  describe("#getHostResourcesByFramework", function() {
+    beforeEach(function() {
+      this.mesosState = {
```
ahoskins: 🤔 Do we have an existing `/tests/fixtures` file that we use in place of this?

ahoskins:
@Fabs As far as I can tell, `MesosStateUtil.getHostResourcesByFramework()` is only called in one place, and that is the MesosStateStore (and then confusingly, the method in MesosStateStore has the same name). So there is only one path calling this code! 👍

Fabs:
@ahoskins I see what you mean by the functions now, one call the other and that is just one use. 

About the fact that it is still broken, you are very right, it Is. There are probably other types of tasks for each it needs to be filtered, because I have a new cluster with `spark` and `kubernetes` and no task restarting, but the CPU continues to go up and up. 

Thanks for catching that.

src/js/utils/__tests__/MesosStateUtil-test.js
Fabs: I copied this from the same file some tests up. I will get some help here in Hamburg today to understand if I can leverage the _fixtures here.

Fabs:
@ahoskins there was also a second cause for the bug:

```
    The function getHostResourcesByFramework was overriding the
    resources property from the MesosStore.get('lastMesosState')
    causing the whole ui to miscalculate the number of cpus, memory
    and disk in used.
````

I fixed it as well.

src/js/utils/__tests__/MesosStateUtil-test.js
Fabs: Looking at it again, I don't think we should use fixtures from Integration tests at unit testing.

There are fixtures at `MesosStateWithStory.js`, but they do not apply to my use case. I thought of two options here, both with merits.
1. Leave it like this - because we only need `state.tasks` and the work to create a scenario for state that would be fully complete (and reusable) it is too much for this
2. Create a new fixture `MesosStateMixedTasks.js` and add just the tasks part of the object to it, and use it the same way is used `MesosStateWithStory.js`.

Fabs:
tests pass locally so:

Alfred test this bitte

Fabs:
Alfred test this bitte

Fabs:
Alfred test this bitte

src/js/utils/__tests__/MesosStateUtil-test.js
ahoskins: Okay, I'm fine leaving it as is :+1: Thanks for clarifying and looking into it.

Fabs:
Alfred test this bitte

Fabs:
Alfred test this bitte

Fabs:
Alfred test this

Fabs:
Alfred test this bitte

PR #2615, opened 2018-01-09
DCOS-20156: Fix MesosStream memory leak

nLight:
The problem with the `mesos-client` has been resolved. Quick explanation: we're keeping all the messages as strings in Rxjs replay subject, we are getting those strings by slicing one big string which represents the stream. The way V8 optimizes sliced strings got in the way here: V8 just keeps the reference with the original string which in our situation means we kept copying that one gigantic string every time we'd get a message -> huge memory consumption.


orlandohohmeier:
LGTM! 🚢

PR #2616, opened 2018-01-09
DCOS-20094: space button 

DanielMSchmidt:
**Checklist**


DanielMSchmidt:
Need to remove the uppercase, it was only added due to miscommunication 😞

PR #2617, opened 2018-01-10
DCOS-14753: adds unanchored property to marathon errors

juliangieseke:
JIRA: https://jira.mesosphere.com/browse/DCOS-14753

---

This adds a new `unanchored` property to errors which enable the `ErrorMessageUtil` to remove the path from the displayed message.

This change will allow the ErrorAlert to de-duplicate these messages.

Example:
```
cmd: You must specify a command, an argument or a container
args: You must specify a command, an argument or a container
container.docker.image: You must specify a command, an argument or a container
```
becomes
```
You must specify a command, an argument or a container
```

### DIFF
--- a/src/js/utils/ErrorMessageUtil.js
+++ b/src/js/utils/ErrorMessageUtil.js
@@ -17,6 +17,9 @@ const ErrorMessageUtil = {
       return rule.match.exec(pathString);
     });
 
+    if (error.isUnanchored) {
+      return message;
+    }
     if (!rule && !pathString) {
       return message;
     }
--- a/plugins/services/src/js/validators/MarathonAppValidators.js
+++ b/plugins/services/src/js/validators/MarathonAppValidators.js
@@ -32,14 +32,27 @@ const MarathonAppValidators = {
     if (hasCmd && hasArgs) {
       const notBothMessage = "Please specify only one of `cmd` or `args`";
       const type = PROP_CONFLICT;
+      const isUnanchored = true;
       const variables = {
         feature1: "cmd",
         feature2: "args"
       };
 
       return [
-        { path: ["cmd"], message: notBothMessage, type, variables },
-        { path: ["args"], message: notBothMessage, type, variables }
+        {
+          path: ["cmd"],
+          message: notBothMessage,
+          type,
+          isUnanchored,
+          variables
+        },
+        {
+          path: ["args"],
+          message: notBothMessage,
+          type,
+          isUnanchored,
+          variables
+        }
       ];
     }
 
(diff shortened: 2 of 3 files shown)


src/js/utils/ErrorMessageUtil.js
```diff
@@ -17,6 +17,9 @@ const ErrorMessageUtil = {
       return rule.match.exec(pathString);
     });
 
+    if (error.unanchored) {
```
DanielMSchmidt: I think some name here is really misleading, either it's `getUnanchoredErrorMessage` or `unanchored`.
When I read this method I understand that `getUnanchoredErrorMessage` and I wouldn't understand why I would need to add a `unanchored` flag on the error itself. If there is a good explaination I would be fine as long as we add it to the jsdoc string, e.g.

```js
/**
* @param {Object} error - The error message
* @param {Boolean} error. unanchored - <why the flag is needed>
*/
```

juliangieseke: I'm with you that this is a bit confusing, but there is a Google Document which explains how these errors work, I've added the "unanchored" property there and explained what it is. 
https://docs.google.com/document/d/1kwdPxBo3UlRZvMxLLIUqSZqLlUJitEuIAWderUgWY1c/edit#heading=h.pw1xuuy946c9

DanielMSchmidt: Wow, that is super interesting. It makes me a bit sad that we can't put the link in the JSDoc because it would be super helpful. Any idea on how we might make this information more accessible?

juliangieseke: @orlandohohmeier mentioned a JIRA for refactoring this, we should keep this in mind.

juliangieseke:
Alfred test this

juliangieseke:
Alfred test this

juliangieseke:
Alfred test this

DanielMSchmidt:
Alfred test this please :)

nLight:
is adjusting errors in JSON Editor not a part of the issue?
screen shot 2018-01-15 at 12 19 57

juliangieseke:
Hm, since this is only eye candy, I'd say no, its not. 🤔 
IMHO we should leave the error message in the JSON editor as is, because there the path is even more important/related than in the Form.

juliangieseke:
image

Fabs:
Rebased that one.

Fabs:
Alfred run test por favor

Fabs:
Alfred test this por favor

src/js/utils/ErrorMessageUtil.js
nLight: Shall we call it `isUnanchored`? It's kind of a convention in Javascript calling boolean flags starting with `is<*>`

juliangieseke: changed 👍

PR #2618, opened 2018-01-10
DCOS-20094: help texts

DanielMSchmidt:
This PR changes the behavior of the help text like described in [DCOS-20094](https://jira.mesosphere.com/browse/DCOS-20094)

dcos-20094-2

**Checklist**

### DIFF
--- a/plugins/services/src/js/constants/OperatorTypes.js
+++ b/plugins/services/src/js/constants/OperatorTypes.js
@@ -1,5 +1,4 @@
-const DEFAULT_HELP = "A string, integer or regex value.";
-const OPTIONAL_HELP = `${DEFAULT_HELP} This field is optional.`;
+const OPTIONAL_HELP = "optional";
 
 const OperatorTypes = {
   UNIQUE: {
@@ -7,7 +6,6 @@ const OperatorTypes = {
     requiresEmptyValue: true,
     stringNumberValue: false,
     tooltipContent: "The unique operator does not accept a value.",
-    helpContent: DEFAULT_HELP,
     name: "Unique",
     description: "Run each app task on a unique attribute ID"
   },
@@ -16,7 +14,6 @@ const OperatorTypes = {
     requiresEmptyValue: false,
     stringNumberValue: false,
     tooltipContent: null,
-    helpContent: DEFAULT_HELP,
     name: "Cluster",
     description: "Run app tasks on nodes that share a certain attribute ID"
   },
@@ -34,7 +31,6 @@ const OperatorTypes = {
     requiresEmptyValue: false,
     stringNumberValue: false,
     tooltipContent: null,
-    helpContent: DEFAULT_HELP,
     name: "Is",
     description: "Run app tasks on nodes having attribute ID with a specific value"
   },
@@ -43,7 +39,6 @@ const OperatorTypes = {
     requiresEmptyValue: false,
     stringNumberValue: false,
     tooltipContent: null,
-    helpContent: DEFAULT_HELP,
     name: "Like",
     description: "Run app tasks on a particular set of attribute IDs"
   },
@@ -52,7 +47,6 @@ const OperatorTypes = {
     requiresEmptyValue: false,
     stringNumberValue: false,
     tooltipContent: null,
-    helpContent: DEFAULT_HELP,
     name: "Unlike",
     description: "Don't run app tasks on a particular set of attribute IDs"
   },
@@ -61,7 +55,6 @@ const OperatorTypes = {
     requiresEmptyValue: false,
     stringNumberValue: true,
     tooltipContent: null,
-    helpContent: DEFAULT_HELP,
     name: "Max Per",
     description: "Run max number of app tasks on each attribute ID"
   }
--- a/plugins/services/src/js/constants/OperatorTypes.js
+++ b/plugins/services/src/js/constants/OperatorTypes.js
@@ -1,5 +1,3 @@
-const OPTIONAL_HELP = "optional";
-
 const OperatorTypes = {
   UNIQUE: {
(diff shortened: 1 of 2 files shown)


plugins/services/src/js/constants/OperatorTypes.js
```diff
@@ -1,13 +1,11 @@
-const DEFAULT_HELP = "A string, integer or regex value.";
-const OPTIONAL_HELP = `${DEFAULT_HELP} This field is optional.`;
+const OPTIONAL_HELP = "optional";
```
juliangieseke: I think this should be "This field is optional" - my fault just typing "optional" when doing the mockup 👍

src/js/components/PlacementConstraintsPartial.js
```diff
@@ -18,6 +18,9 @@ import OperatorTypes from "#PLUGINS/services/src/js/constants/OperatorTypes";
 import PlacementConstraintsUtil
   from "#PLUGINS/services/src/js/utils/PlacementConstraintsUtil";
 
+// &nbsp; to space the table rows
+const NBSP = "\u00A0";
```
juliangieseke: 🤔

src/js/components/PlacementConstraintsPartial.js
```diff
@@ -119,10 +124,12 @@ export default class PlacementConstraintsPartial extends Component {
                 );
               })}
             </Select>
-            <FieldHelp>
-              Specify where your app will run.
-            </FieldHelp>
-            <FieldError>{operatorError}</FieldError>
+
+            {operatorError && <FieldError>{operatorError}</FieldError>}
+            {!operatorError &&
```
juliangieseke: Do we really only want to display an error OR the help message? I'd say when an error occured the help is more needed than ever. 🤔

plugins/services/src/js/constants/OperatorTypes.js
juliangieseke: Also, after your change, this is basically only a true/false flag, right? It should be possible to rewrite it that it uses `requiresValue` - which means we should be able to remove it from the file 🕺

plugins/services/src/js/constants/OperatorTypes.js
```diff
@@ -1,13 +1,11 @@
...
 
 const OperatorTypes = {
   UNIQUE: {
     requiresValue: false,
     requiresEmptyValue: true,
     stringNumberValue: false,
     tooltipContent: "The unique operator does not accept a value.",
-    helpContent: DEFAULT_HELP,
```
juliangieseke: If we keep it, instead of removing, I'd vote for `helpContent: null`. but see my other comments, please.

plugins/services/src/js/constants/OperatorTypes.js
DanielMSchmidt: You are right, let's make it a bool 👍

plugins/services/src/js/constants/OperatorTypes.js
DanielMSchmidt: Let's remove it and make the requiresValue our source of truth 👍

src/js/components/PlacementConstraintsPartial.js
DanielMSchmidt: &nbsp; gets escaped 😭

src/js/components/PlacementConstraintsPartial.js
DanielMSchmidt: Lily wrote 

>   In a case like that, you should remove the help text and only have the error text

in the [issue comments](https://jira.mesosphere.com/browse/DCOS-20094?focusedCommentId=133266&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-133266) that is the reason for this decision. I am not married to this decision, should we ping it back to Lily?

DanielMSchmidt:
TODO:

- [ ] check if the error message has enough information to resolve the issue

juliangieseke:
Yup, LGTM 👍

DanielMSchmidt:
@juliangieseke I think I addressed the issues you were concerned about, would <3 another review

src/js/components/PlacementConstraintsPartial.js
```diff
@@ -148,19 +155,20 @@ export default class PlacementConstraintsPartial extends Component {
...
+              <FieldError className={{ hidden: hideValueColumn }}>
+                {valueError}
+              </FieldError>}
+            {!valueError &&
+              <FieldHelp>
+                {isLastField ? "A string, integer or regex value. " : NBSP}
+                {typeSettings &&
+                  !typeSettings.requiresValue &&
+                  !typeSettings.requiresEmptyValue &&
+                  "This field is optional."}
```
juliangieseke: 👍

Fabs:
One little thing I spotted. When there is more than one message for the value field, they look a little inconsistent, since one is a sentence with proper punctuation and the other a word.

mesosphere_dc_os_enterprise

My eyes feel like scanning it as a sentence, and it took me a while to get it.

Fabs:
Also, are they supposed to appear only in the latest one? The screens on the ticket show optional everywhere.

image

DanielMSchmidt:
@Fabs Sorry the images were outdated, here is what it looks like currently:

bildschirmfoto 2018-01-11 um 14 46 13

As you can see the text is a complete sentence.
The "optional" phrase should only appear on the group by constraint, as this is the only one with an optional value 👍

DanielMSchmidt:
Alfred test this please 😏

DanielMSchmidt:
Alfred test this please 😏

DanielMSchmidt:
Alfred test this please 😏

PR #2623, opened 2018-01-11
DCOS-20244: Don't show (Local) on zones

DanielMSchmidt:
[Fixes DCOS-20244](https://jira.mesosphere.com/browse/DCOS-20244)

bildschirmfoto 2018-01-11 um 18 20 51
bildschirmfoto 2018-01-11 um 18 20 45

**Checklist**

### DIFF
--- a/plugins/services/src/js/utils/TaskUtil.js
+++ b/plugins/services/src/js/utils/TaskUtil.js
@@ -145,18 +145,7 @@ const TaskUtil = {
       return "N/A";
     }
 
-    const nodeZoneName = node.getZoneName();
-    const masterNode = CompositeState.getMasterNode();
-
-    if (
-      masterNode &&
-      nodeZoneName === masterNode.getZoneName() &&
-      nodeZoneName !== "N/A"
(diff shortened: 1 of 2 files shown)


plugins/services/src/js/utils/TaskUtil.js
```diff
@@ -146,15 +146,6 @@ const TaskUtil = {
     }
 
     const nodeZoneName = node.getZoneName();
```
bstavroulakis: Small detail, since we just return the zone name this could be `return node.getZoneName()`

DanielMSchmidt: I agree, I'll fix that 👍

DanielMSchmidt:
@juliangieseke @bstavroulakis Thank you for your review comment, I addressed them, could you take another look please :)

DanielMSchmidt:
Alfred test this please :)

Fabs:
You guys too fast :-). LGTM to mee too.

PR #2624, opened 2018-01-11
fix(PodStorageConfigSection): display DSS type and size

ahoskins:
Three fixes for the volumes pod review screen:
1. Show type as DSS
2. Show a size column
3. Fix the edit button (previously the ID was wrong)

**Before**:
screen shot 2018-01-11 at 3 34 08 pm

**After**:
screen shot 2018-01-11 at 3 32 19 pm

Closes DCOS-20258

**Checklist**

### DIFF
--- a/plugins/services/src/js/service-configuration/PodStorageConfigSection.js
+++ b/plugins/services/src/js/service-configuration/PodStorageConfigSection.js
@@ -28,7 +28,7 @@ class PodStorageConfigSection extends React.Component {
         prop: "type"
       },
       {
-        heading: "Size",
+        heading: "Size (MiB)",
         prop: "size"
       },
       {
@@ -66,16 +66,26 @@ class PodStorageConfigSection extends React.Component {
         type = VolumeConstants.type.host;
       }
       if (volume.persistent != null) {
-        type = VolumeConstants.type.localPersistent;
+        if (volume.persistent.profileName != null) {
+          type = VolumeConstants.type.dss;
+        } else {
+          type = VolumeConstants.type.localPersistent;
+        }
       }
       if (Object.keys(volume).length === 1 && volume.name != null) {
         type = VolumeConstants.type.ephemeral;
       }
 
+      let size;
+      if (volume.persistent != null && volume.persistent.size != null) {
+        size = volume.persistent.size;
+      }
+
       const volumeInfo = {
         type,
         volume: volume.name,
-        hostPath: volume.host
+        hostPath: volume.host,
+        size
       };
 
       // Fetch all mounts for this volume in the containers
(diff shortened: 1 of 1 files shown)


plugins/services/src/js/service-configuration/PodStorageConfigSection.js
```diff
@@ -27,7 +27,7 @@ class PodStorageConfigSection extends React.Component {
         prop: "type"
       },
       {
-        heading: "Size",
+        heading: "Size (MiB)",
```
ahoskins: It might not be true that the unit is MiB?

weblancaster: External volumes aren't MiB

juliangieseke: yup, we should append the unit to every row 👍

DanielMSchmidt:
Alfred test this please :)

ahoskins:
Holding till we re-enable the DSS plugin.

nLight:
run integration tests

nLight:
run system tests

plugins/services/src/js/service-configuration/PodStorageConfigSection.js
ahoskins: @Poltergeist are external volumes an option for pods? In the dropdown not appearing. If they aren't, then all others are MiB and we can keep it here.

ahoskins:
run system tests

plugins/services/src/js/service-configuration/PodStorageConfigSection.js
DanielMSchmidt: @Poltergeist @ahoskins Did you find an answer to this question?

ahoskins: Haha no I did not 🙃

juliangieseke: so is this wrong right now? then we should block this PR 🙈

juliangieseke: so is this wrong right now? then we should block this PR 🙈
feel free to dismiss my block once you verified this is correct.

Poltergeist: External volumes are not available for pods so far. This may change in the future. So we may keep this for now.

ahoskins:
I've confirmed with @Poltergeist: no external volumes on pods. Therefore I'm confident to merge this as-is! @juliangieseke 👍

DanielMSchmidt:
@ahoskins I get an `uglifyjs is deprecated - use uglify-js instead.` Error which prevents me from starting the server, can we maybe rebase this?

PR #2629, opened 2018-01-12
Update login view with revised logo

ashenden:
- Logo updated with revised images based on changes to logotype.


nLight:
Can't we do this massive gradient with CSS?

juliangieseke:
Alfred test this

juliangieseke:
Alfred test this

juliangieseke:
Alfred test this

DanielMSchmidt:
@ashenden Would this be close enough for you? https://jsfiddle.net/83p4jc61/
This would be a CSS gradient solution that would spare our users aprox. 1.5MB. I am not that could visually, but if you say both are close enough I could include this gradient to our login screen 👍 

bildschirmfoto 2018-01-17 um 11 16 36

ashenden:
@DanielMSchmidt nice work.  I wouldn't be against using the CSS gradient.  However, our background image isn't a pure linear gradient.  It's actually comprised of several heavily blurred shapes layered to create that composite background.

The @2x image comes in just shy of 1MB.  Are we really that concerned about the file size of an image like this?

DanielMSchmidt:
@ashenden Okay, yeah I miss the eye for that, I hope you can help me get better at this 👍  I think it is totally fine, I personally tend to get obsessed with optimizations like this, but I think we are going in the right direct there :) Just to make sure, you put the images through [imageoptim](https://imageoptim.com/mac) right? (Or are the optimized in our prod pipeline? I am not aware of that 🙈 )

ashenden:
@DanielMSchmidt no problem.  I'm glad to know you and others are looking out and pushing for these things.

AFAIK, we are not using ImageOptim in our build process.  I just manually ran the images through it on my laptop before committing.

PR #2630, opened 2018-01-14
DCOS-16480: Improvements on test documentation

Fabs:
This PR makes improvements in content for TESTING (Unit, Integration, and System).

That would be nice to also have feedback on how to handle the private system testing, and any other content you would like to have on this page that is not present as of now. 

I consider this the V1 of the improvements and I think it will need some hammering until it can be merged. 

NOTES:
- I moved TESTING.md inside CONTRIBUTING.md, the idea is that it makes more complicated to organize the content if it is spread in many files (TESTING, BEST_PRACTICES and etc) and that once we have enough organized and good quality content, we can break it down.
- When writing the session of system testing, I learned about the internal documentation on system-testing (https://docs.google.com/document/d/14Jl4072onW1gcroEUvcBKONKiJcRARBLP-qDsZ0LlJ0). I think somehow this work and my work are highly related and should be in the same place (although I understand the driver is still closed source).

TIPS:
- You can use GRIP to run this locally (https://github.com/joeyespo/grip)


CONTRIBUTING.md
```diff
@@ -210,6 +214,231 @@ it is helpful to use [npm link](https://docs.npmjs.com/cli/link).
...
+
+### Example of a Unit Test
+
+This test verifies that unit `decomposePodTaskId` when given the input string 
+"podname.instance-instancename.taskname" returns an object with the `podID`, 
+`instanceID` and `taskName respectively`. 
+
+```js
+  describe("#decomposePodTaskId", function() {
+    it("should correctly de-compose", function() {
```
weblancaster: We shouldn't use `should`

CONTRIBUTING.md
```diff
@@ -11,6 +11,10 @@
   - [Breaking Changes](#breaking-changes)
```
weblancaster: I rather have `TESTING.md` just because is a nice practice and I know I needed to go to that file to file what I needed and didn't need to scan a file with many lines.

So my two cents is having multiple files with smaller lines to read through.

Fabs: I am of that opinion too. I like to have the separation so it makes for easy to consume chunks. @orlandohohmeier pointed out though that the separation we have right now is a little confusing, and the links not really good. I think it is better to have a large text, that content that is intuitive to find.

What if I leave the TESTING.md file (because in this case, it is indeed a standard) and make it point to the CONTRIBUTING document?

CONTRIBUTING.md
Fabs: 🙈 . Damn late night copy and paste.

Fabs:
Alfred test this bitte

CONTRIBUTING.md
```diff
@@ -210,6 +214,231 @@ it is helpful to use [npm link](https://docs.npmjs.com/cli/link).
...
+
+While Integration Tests looks for the composition within the system, 
+[System Tests](#system-testing) test the integration with the real external 
+dependencies trying to as close as possible to the production use of a client 
+(in the case of DC/OS, against a real cluster). They are even harder to setup 
+than integration tests because you often have to create custom tooling to 
+operate the whole product. This contributes to them being slower and more often 
+prone to fail occasionally, for example, because of network issues like latency.
+
+## Unit Testing
```
DanielMSchmidt: What is the purpose fo this paragraph?

CONTRIBUTING.md
```diff
@@ -210,6 +214,231 @@ it is helpful to use [npm link](https://docs.npmjs.com/cli/link).
...
+```
+
+You can even pass parameters to the test engine (in this case jest), when you, 
+for instance, want to run a single spec, for example, `MesosStateUtil`:
+
+```sh
+npm run test -- --watch MesosStateUtil
+```
+
+### Example of a Unit Test
```
DanielMSchmidt: I have the feeling this might become stale after a while (e.g. we could also write this test with snapshot testing, so as a reader I might think we don't do that at all).

Wouldn't it be simpler to direct the reader directly to one of our test directories?

CONTRIBUTING.md
Fabs: Not quite sure which paragraph we are talking about.

Case 1: To ensure that individual...
You can navigate straight to Unit testing via a link, so for a reader that starts at this point of the document, jumping straight to the sum example looked to much, so I provided a small sentence to introduce the topic of unit testing.

Case 2: Think of a sum ...
This was there before, it is an example of unit testing, I left also because I build on this on the later topics (Integration tests).

CONTRIBUTING.md
Fabs: I think we should make an effort to keep this up to date when we introduce new techniques.

For snapshot testing, I think we should add a paragraph. 

I think adding the link to one of the test folders is a great idea (do you recommend any?). I don't think those two are exclusive (we can have both this content and the link)

CONTRIBUTING.md
DanielMSchmidt: Sorry for being unclear, I will try to improve 👍 

I meant the entire unit testing part, for me it is text that we kind of have to maintain and I think a link might be sufficient.

CONTRIBUTING.md
DanielMSchmidt: You don't need to add a snapshot paragraph, just wanted to address that I think this paragraph is not ultimatively useful and might go stale ^^

CONTRIBUTING.md
Poltergeist: I think it is not a bad idea to have a description about what we do understand by unit tests to set a baseline.

CONTRIBUTING.md
Poltergeist: If we want to keep the example let's rather have a generic example. e.g. let's take the sum example from above.

CONTRIBUTING.md
DanielMSchmidt: 🤔 Yeah, I haven't thought about it in terms of setting a baseline.

DanielMSchmidt:
Run integration tests :)

DanielMSchmidt:
Run integration tests

CONTRIBUTING.md
Fabs: I also like the setting the baseline idea. I am creating a ticket about documentation moving further, so we can keep improving this for us, and our community.

I think also if someone can provide a great tests documentation on the internet, for this basic topics, we could just link and be happy (like the betterspecs.org).

CONTRIBUTING.md
Fabs: I will put this improving to a generic example as an improvement.

Fabs:
Thanks everybody for the great feedback.

PR #2637, opened 2018-01-16
Emit MESOS_STATE_CHANGE events only having the full state

nLight:
Some components rely on the first `MESOS_STATE_CHANGE` event providing them with a full state. Previously `GET_MASTER` request would also trigger `MESOS_STATE_CHANGE` rendering those components broken as they'd get the first update with a partial state and unsubscribe from further updates.

With this fix we always ensure we're having full state before we start emitting `MESOS_STATE_CHANGE`

Closes DCOS-20308

## Testing

- Please refer to the ticket to test the original issue [DCOS-20308](https://jira.mesosphere.com/browse/DCOS-20308)
- Additionally please give it a hard time going through the Services page, going in every tab with and without reloading.

## What's going on

To achieve that we have to await until we have the first event from the stream plus response from the GET_MASTER request `waitStream`. 
- Once they arrived we start `eventTriggerStream` that merges the data stream and an interval stream. 
- Interval stream is there to support the legacy codebase that expects CHANGE events every two seconds. 
- It is possible that `eventTriggerStream` triggers more frequently then 2 `Config.getRefreshRate()` that is a good thing(*), yet we still want to debounce the whole thing.

(*) it's a good thing because we update the UI as events come without having to wait for another cycle.


src/js/stores/MesosStateStore.js
```diff
@@ -73,20 +67,34 @@ class MesosStateStore extends GetSetBaseStore {
...
+    const eventTriggerStream = dataStream.merge(
+      // Lots of things in DCOS UI rely on the MesosStateStore emitting
+      // MESOS_STATE_CHANGE events but since we're switching to the stream there will
+      // be no events coming. To avoid a bigger refactoring I decided to go with
+      // a fake emitter that starts emitting events once the initial state came through
+      // TODO: https://jira.mesosphere.com/browse/DCOS-18277
+      Rx.Observable.interval(Config.getRefreshRate())
+    );
+
+    waitStream
```
ahoskins: (new to rxjs and want to clarify if I understand this correctly):

Isn't there going to be duplicates when the original `getMasterRequest` stream emits? The `dataStream` merged this stream's events in, and the `waitStream` will also emit events when `getMasterRequest` emits. Then these two are eventually concat together, and `onStreamData` is subscribed. So when GET_MASTER updates, I would expect two events, but this isn't happening when I test in the console. What am I not understanding correctly?

ahoskins: **Update:** ahh I understand! I should have read the PR description more carefully - my understanding of the code now aligns with that! 😄

nLight:
Alfred test this

Fabs:
I am writing tests for that one right now.

src/js/stores/MesosStateStore.js
Fabs: There is a crude diagram we did, maybe it helps to understand a little better. 
img_5840

- M. Req = mesos request
- EVI = Event Stream (Observable.interval(...))
- BLA = 😃  is the waitStream.concat(eventTrigger)

Fabs: @nLight do we have a reference doc on how to draw streams? I found this on marble tests but not quite what we did (https://github.com/ReactiveX/RxJS/blob/master/doc/writing-marble-tests.md).

src/js/stores/MesosStateStore.js
```diff
@@ -73,20 +67,34 @@ class MesosStateStore extends GetSetBaseStore {
...
+      .concat(eventTriggerStream)
+      .debounceTime(1000)
```
orlandohohmeier: Please calculate the debounce time as `Config.getRefreshRate() * .5` and document the min/max frequency of events.

src/js/stores/MesosStateStore.js
```diff
@@ -1,5 +1,6 @@
 import PluginSDK from "PluginSDK";
 import { request } from "@dcos/mesos-client";
+import Rx from "rxjs";
```
orlandohohmeier: 🎨  I'd import `Observable` instead of `Rx` to reduce the amount of redundancy.

DanielMSchmidt: I like how rxjs has these split up imports <3 `import { Observable } from 'rxjs/Observable';`

Fabs:
Alfred test this bitte

Fabs:
@orlandohohmeier I fixed your points.

src/js/stores/MesosStateStore.js
ahoskins: @Fabs thanks for the diagram! I found this as a nice way to visualize the API: http://rxmarbles.com/#concat. Once I realized the difference between concat and merge, the code was pretty intuitive to understand.

Fabs: Nice for the reference @ahoskins. I think for web, `merge`, `switch`, `concat` are the most useful stream.

I also recommend this, to go deeper.  https://frontendmasters.com/courses/asynchronous-javascript/

Fabs:
when tests are ✅ , then its good to go. Removed the label.

Fabs:
Although I have to say I had this same 4 system tests to fail before, so there might be something there that breaks the system tests :/

ahoskins:
Alfredo test this please

orlandohohmeier:
@Fabs could please squash c401c76 and 311da69? Not blocking but we might also want to revisit the comments to iron out typos and remove redundant information.

DanielMSchmidt:
Alfred test this please :)

Fabs:
@orlandohohmeier. Shortened the comments and passed them on grammarly. Squashed the commits also.

src/js/stores/MesosStateStore.js
```diff
@@ -290,19 +305,23 @@ class MesosStateStore extends GetSetBaseStore {
...
-      );
-    }
+  onStreamData() {
+    this.emit(MESOS_STATE_CHANGE);
   }
 
   onStreamError(error) {
     this.emit(MESOS_STATE_REQUEST_ERROR, error);
+
+    // This was added in the past so the stream does not swallow errors. In its
```
orlandohohmeier: 🔡  This was added in the past so the stream does not swallow errors. In its
current for_m_, it causes a problem wit_h_ fixtures 1-pod and 1-empty-group and make ServiceFormModal and PodDetailpPage tests fail.

orlandohohmeier:
@Fabs thanks for addressing the comments. I've found another typo but that shouldn't block this PR.

src/js/stores/MesosStateStore.js
Fabs: Oh. I only checked the comments on the subscribe method(). Thanks for catching this.

Fabs: just fixed that now, that was actually much worse than the others :-)

Fabs:
@DanielMSchmidt and @orlandohohmeier. My latest push removed stale reviews. Could you take a look again? I just changed the comments on The MesosStateStore (latest commit).

package.json
```diff
@@ -68,6 +68,7 @@
     "reactjs-mixin": "0.0.2",
     "redux": "3.3.1",
     "reflect-metadata": "0.1.10",
+    "rxjs": "5.4.3",
```
Poltergeist: Since we are adding a dependency here shouldn't we have additions to the npm-shrinkwrap? currently I only see removals of `fsevents` which is unfortunate that it landed on master again but okay. 

Just want to make sure this is right.

DanielMSchmidt: I asked that here in person, rxjs was already a dependency of one of our dependencies 👍

nLight: It's there already: https://github.com/dcos/dcos-ui/blob/ac7bb50cab9ed517f89c2abe3558fb9ab4f6961c/npm-shrinkwrap.json#L8555

Poltergeist: good!

nLight:
Alfred test this

nLight:
Alfred test this

Poltergeist:
Alfred you can't test this. 👯 👯‍♂️ 🎉

Fabs:
Alfred test this

PR #2639, opened 2018-01-16
chore(cypress): do not upload video when tests pass

nLight:
`--config videoUploadOnPasses=false,numTestsKeptInMemory=0` CLI command doesn't seem to be working, let's try adding it to the config file. With the settings cypress won't compress/upload video when test are green saving precious minutes. `numTestsKeptInMemory` should also reduce the memory footprint


nLight:
Looks like `videoUploadOnPasses` is the new name, digging up the old one, for our version

nLight:
Ok, we don't have this option on our version, sorry 😓

PR #2650, opened 2018-01-18
fix(ServiceModals) error on delete

weblancaster:
Fix intermittent error when deleting a service and the page throws an error blocking
the user from navigating/using the application and forcing to refresh

Closes DCOS-19743 (https://jira.mesosphere.com/browse/DCOS-19743)

**Checklist**

### DIFF
--- a/plugins/services/src/js/components/modals/ServiceModals.js
+++ b/plugins/services/src/js/components/modals/ServiceModals.js
@@ -25,6 +25,10 @@ class ServiceModals extends React.Component {
       pendingActions
     } = this.props;
 
+    if (modalProps.service == null) {
+      return null;
+    }
+
     const key = ActionKeys.GROUP_CREATE;
     const { service } = modalProps;
 


MatApple:
alred retest this please

MatApple:
alfred retest this please

plugins/services/src/js/components/modals/ServiceModals.js
```diff
@@ -25,6 +25,10 @@ class ServiceModals extends React.Component {
       pendingActions
     } = this.props;
 
+    if (modalProps.service == null) {
```
MatApple: @weblancaster are you sure this is the correct place to fix? Seems odd fixing this issue by not rendering the Group modal

weblancaster: Makes sense, let me take a look

weblancaster:
I wasn't able to replicate @ahoskins screenshot, neither was him able a second time.

plugins/services/src/js/components/modals/ServiceModals.js
weblancaster: You were right @MatApple the issue goes deep in the rabbit hole.

DanielMSchmidt: I agree with both of you, but as this is a blocker, let's get it merged and find a proper fix next print when there is more time. I added an issue so that we don't forget 👍 https://jira.mesosphere.com/browse/DCOS-20391

Poltergeist:
run integration tests

Poltergeist:
run integration test

Poltergeist:
Since there are 2 approvals + green tests I will merge this

DanielMSchmidt:
Great, thank you @Poltergeist

PR #2653, opened 2018-01-19
[1.9] chore(cypress): enable new Jenkins setup

nLight:
This PR enables the upcoming change in Jenkins as cypress has switched from a global installation to a dependency.


nLight:
Alfred tests this

nLight:
Alfred test this

PR #2659, opened 2018-01-23
refactor(SchemaField): improve framework heading sizes

ahoskins:
The purpose of this is to make the framework forms more visually parsable. Changes:
1. Give each TitleField it sizing appropriate to it's location in the json schema. Previously, they were all rendered as `<h2>` for simplicity, but this made it hard to notice which were nested fields.
2. Bigger sizing on Placement Constraints widget label.
3. Capitalize each word in all field labels
4. Extra margin below Placement section to visually separate from rest of form

**Testing**:
1. Turn enterprise off
2. Open the catalog for beta-kafka. Notice the form differences. H1 and H2 and H3 for layers of config headings, and Placement Constraints label is proper size.

Thanks to @leemunroe for suggestions and assistance. 

**Before**:
screen shot 2018-01-22 at 4 31 58 pm
screen shot 2018-01-22 at 4 31 50 pm

**After**:
screen shot 2018-01-22 at 4 30 32 pm
screen shot 2018-01-22 at 4 30 20 pm

**Checklist**

### DIFF
--- a/src/js/components/FrameworkConfigurationForm.js
+++ b/src/js/components/FrameworkConfigurationForm.js
@@ -245,11 +245,26 @@ class FrameworkConfigurationForm extends Component {
     } = this.props;
 
     const TitleField = props => {
-      return (
-        <h2 className="flush-top short-bottom">
-          {StringUtil.capitalizeEveryWord(props.title)}
-        </h2>
-      );
+      const level = props.formContext.level;
+      if (level === 1) {
+        return (
+          <h1 className="flush-top short-bottom">
+            {StringUtil.capitalizeEveryWord(props.title)}
+          </h1>
+        );
+      } else if (level === 2) {
+        return (
+          <h2 className="short-bottom">
+            {StringUtil.capitalizeEveryWord(props.title)}
+          </h2>
+        );
+      } else {
+        return (
+          <h3 className="short-bottom">
+            {StringUtil.capitalizeEveryWord(props.title)}
+          </h3>
+        );
+      }
     };
 
     const jsonEditorPlaceholderClasses = classNames(
@@ -311,6 +326,7 @@ class FrameworkConfigurationForm extends Component {
                       validate={this.validate}
                       showErrorList={false}
                       transformErrors={this.transformErrors}
+                      formContext={{ level: -1 }}
                     >
                       <div />
                     </SchemaForm>
--- a/src/js/components/FrameworkConfigurationForm.js
+++ b/src/js/components/FrameworkConfigurationForm.js
@@ -254,19 +254,20 @@ class FrameworkConfigurationForm extends Component {
             {StringUtil.capitalizeEveryWord(props.title)}
           </h1>
         );
-      } else if (level === FrameworkConfigurationConstants.headingLevel.H2) {
+      }
+      if (level === FrameworkConfigurationConstants.headingLevel.H2) {
         return (
           <h2 className="short-bottom">
             {StringUtil.capitalizeEveryWord(props.title)}
           </h2>
         );
-      } else {
-        return (
-          <h3 className="short-bottom">
-            {StringUtil.capitalizeEveryWord(props.title)}
-          </h3>
-        );
       }
+
+      return (
+        <h3 className="short-bottom">
+          {StringUtil.capitalizeEveryWord(props.title)}
+        </h3>
+      );
     };
 
     const jsonEditorPlaceholderClasses = classNames(
--- a/src/js/components/FrameworkConfigurationForm.js
+++ b/src/js/components/FrameworkConfigurationForm.js
@@ -19,6 +19,8 @@ import StringUtil from "#SRC/js/utils/StringUtil";
 import PlacementConstraintsSchemaField
   from "#SRC/js/components/PlacementConstraintsSchemaField";
 import YamlEditorSchemaField from "#SRC/js/components/YamlEditorSchemaField";
+import FrameworkConfigurationConstants
+  from "#SRC/js/constants/FrameworkConfigurationConstants";
 
 MountService.MountService.registerComponent(
   PlacementConstraintsSchemaField,
@@ -246,13 +248,13 @@ class FrameworkConfigurationForm extends Component {
 
     const TitleField = props => {
       const level = props.formContext.level;
-      if (level === 1) {
+      if (level === FrameworkConfigurationConstants.headingLevel.H1) {
         return (
           <h1 className="flush-top short-bottom">
             {StringUtil.capitalizeEveryWord(props.title)}
           </h1>
         );
-      } else if (level === 2) {
+      } else if (level === FrameworkConfigurationConstants.headingLevel.H2) {
         return (
           <h2 className="short-bottom">
             {StringUtil.capitalizeEveryWord(props.title)}
@@ -326,7 +328,6 @@ class FrameworkConfigurationForm extends Component {
                       validate={this.validate}
                       showErrorList={false}
                       transformErrors={this.transformErrors}
-                      formContext={{ level: -1 }}
                     >
                       <div />
                     </SchemaForm>
--- a/src/js/components/PlacementConstraintsSchemaField.js
+++ b/src/js/components/PlacementConstraintsSchemaField.js
@@ -110,7 +110,7 @@ export default class PlacementConstraintsSchemaField extends Component {
   }
 
   render() {
-    const { errorMessage, fieldProps } = this.props;
+    const { label, fieldProps, errorMessage } = this.props;
     const { batch } = this.state;
 
     const appendToList = (memo, transaction) => memo.concat(transaction);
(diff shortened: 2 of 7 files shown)


src/js/components/FrameworkConfigurationForm.js
```diff
@@ -245,11 +245,26 @@ class FrameworkConfigurationForm extends Component {
     } = this.props;
 
     const TitleField = props => {
-      return (
-        <h2 className="flush-top short-bottom">
-          {StringUtil.capitalizeEveryWord(props.title)}
-        </h2>
-      );
+      const level = props.formContext.level;
```
weblancaster: @dcos/frontend The code below feels odd having to check to "random selected" numbers to set the headers size.

I debugged a bit to see if there was a better approach and then discussed with @ahoskins my discomfort about it.

I'm curious to know if anyone has a better idea we can do this.

src/js/components/SchemaField.js
```diff
@@ -317,10 +318,15 @@ class SchemaField extends Component {
...
 
   render() {
-    const { schema, errorSchema, uiSchema } = this.props;
+    const { schema, errorSchema, uiSchema, registry } = this.props;
 
     if (schema.type === "object") {
-      return <DefaultSchemaField {...this.props} />;
+      const nextRegistry = {
+        ...registry,
+        formContext: { level: registry.formContext.level + 1 }
```
weblancaster: Related to https://github.com/dcos/dcos-ui/pull/2659/files#r163395547

src/js/components/FrameworkConfigurationForm.js
ahoskins: A way we could **almost know** this is from the `idSchema`, which is passed into each schema value. IdSchema (as the name suggests) gives each node in the schema a unique string. For example "root_service_name" is `service.name` in the JSON schema. The issue is how id schema will split **multi-word keys** - it'll do "root_coordinator_nodes_name" for `coordinator_nodes.name`. So you can't just count the underscores to know the depth.

I don't understand why [idSchema](https://github.com/mozilla-services/react-jsonschema-form/blob/51d403ab448326183f58b5cc7c2c750c200f0242/src/utils.js#L622) would split by "_" rather than ".", because this would get rid of this ambiguity. 🙃 

I think the current solution is the simplest we can do, but I agree with @weblancaster that it's a bit odd.

MatApple: @ahoskins is there any context that is passed down? We should not use the `idSchema` as that is not reliable, nor a good practice :)

MatApple: Actually, I'm just noticing that there is `formContext`. @weblancaster can you point to where the randomness is so I can understand the oddity a bit better please?

weblancaster: @MatApple the `level` comes from `registry.formContext.level + 1` which I think is a bit random and odd when checking for `level === 1`.

What I meant by that the code here isn't telling us anything. 
A good example on the other hand, would be something like `serviceStatus === ServiceStatus.WARNING`

Does that make sense?

If the team agree this is the best solution I agree too, just feels like this is code smell.

MatApple: I see - you're suggesting a constant would be more appropriate here instead of an assigned integer and then the mapping of headings to integers should also be a constant. Sounds reasonable to me. I am also ok with the nesting counter as is `level + 1` since we're measuring recursion depth. How that is mapped to heading elements should be easily editable as a constant.

ahoskins: Agree. I will introduce constants here.

weblancaster: Thanks for the changes @ahoskins 

Small change but reads so much better now.

DanielMSchmidt:
On this branch, if I configure it to soak and open the config of a kafka service it hangs (only shows the black overlay and couple of hundred `_PlacementsValidators2.default.validateNoBatchError` errors in the console. I am not sure if it really has something to do with this branch, but on master it works fine 🤷‍♂️

orlandohohmeier:
@Fabs could you please look into the issue @DanielMSchmidt reported? Could be related to the adjustments you did.

ahoskins:
@DanielMSchmidt I'm pretty certain that is fixed by the work @Fabs did. Reason it happens on this branch is that it hasn't been rebased for a while. I'll do that now, then problems should all go away 🌈

ahoskins:
While trying to rebase this, I'm running into all sorts of errors in the snapshots. And not sure how we want to go about updating our snapshots: just `jest --updateSnapshot`, or another way? @weblancaster because I think you introduced snaps, and @Fabs because you wrote these particular snap files.

Fabs:
@ahoskins, I usually have jest running, and then when it gives me an error, I press `u`. Think your command does the same.

You need to look the snapshot file and see whether it makes sense, it starts with a description of the test case, and then the HTML. Like image snapshots, it's up to us to see whether it makes sense the change. Probably it does because you just touched that component. 

When you push the snapshots, I can also review them.

src/js/components/FrameworkConfigurationForm.js
```diff
@@ -245,11 +247,26 @@ class FrameworkConfigurationForm extends Component {
...
+      if (level === FrameworkConfigurationConstants.headingLevel.H1) {
+        return (
+          <h1 className="flush-top short-bottom">
+            {StringUtil.capitalizeEveryWord(props.title)}
+          </h1>
+        );
+      } else if (level === FrameworkConfigurationConstants.headingLevel.H2) {
```
bstavroulakis: Personal preference note.
Since we are returning we could maybe remove the else if and add the last statement as the default return.

bstavroulakis: Also, now that the tests are working we can probably remove the "pass" label.

ahoskins:
@Fabs I've rebased this on master so the issues @DanielMSchmidt mentioned should be gone. 

Also updated snapshot (I think I did it correctly).

src/js/components/FrameworkConfigurationForm.js
ahoskins: Flattened this, thanks

ahoskins:
Alfred test this please

src/js/components/__tests__/__snapshots__/PlacementConstraintsSchemaFields-test.js.snap
```diff
@@ -18,10 +18,23 @@ exports[`PlacementConstraintsSchemaField with invalid content displays form to e
 `;
 
 exports[`PlacementConstraintsSchemaField with valid content displays edit constraints area 1`] = `
```
Fabs: Looks good to me the snapshots.

ahoskins:
run integration tests

ahoskins:
alfred test this please

ahoskins:
alfred test this please

weblancaster:
@ahoskins as @Fabs said, as long the snapshots match the expected component output we should be fine.
If doesn't match the component output and it's failing that's a red flag.

ahoskins:
run integration tests

juliangieseke:
run system tests

src/styles/components/form-elements/form-group/styles.less
```diff
@@ -10,6 +10,10 @@
       }
     }
 
+    .extra-bottom {
```
weblancaster: I REALLY dislike we need to keep adding this kinda of css all over.

leemunroe: We shouldn't need to touch CSS for things like this. It should all be magic if the markup is accurate.

src/js/components/PlacementConstraintsSchemaField.js
```diff
@@ -131,11 +133,22 @@ export default class PlacementConstraintsSchemaField extends Component {
         batch.reduce(combineReducers({ constraints: JSONReducer }))
       )
     );
+    const isRequired = fieldProps.required ? "*" : "";
 
     return (
-      <div>
+      <div className="extra-bottom">
         <FieldLabel>
```
leemunroe: Is there a reason this is a field label? Field labels are for form components. But this is a heading.

src/styles/components/form-elements/form-group/styles.less
weblancaster: Thanks for commenting @leemunroe 

@ahoskins do you want to take another look on it and see if you can accomplish the same or close resul without this please?

src/js/components/PlacementConstraintsSchemaField.js
weblancaster: Good point!

src/styles/components/form-elements/form-group/styles.less
ahoskins: Yes I also really dislike this! I remember looking previously and not finding a way existing. Will look again.

ahoskins: The issue is the vertical margins come from the heading. And in this case the next field in the form is just a <label> and class "form-group-heading", which does not provide any top margin.

screen shot 2018-02-09 at 3 29 12 pm

ahoskins: Fixed this by using .pod from cnvs, and flush on all sides except the bottom. 👍

ahoskins:
@weblancaster Comments addressed.

DanielMSchmidt:
@ahoskins rebasing this against master as we had this problem with marathon changing the release path

ahoskins:
@DanielMSchmidt Thanks. Yes comments have been addressed. This PR is just an "improvement" PR, hence not blocker. When someone else gets the time, they can do one last review.

PR #2666, opened 2018-01-24
fix(ServiceTableUtil): fix status sorting

weblancaster:
Fix Service status column sorting

Closes DCOS-19732

**How to test:**
- Go to service page
- Start a couple services with different statuses (Deploying, Running, Recovering, etc)
- Click the status name/header column
- You will see the service page sort by the order below

_ps: be aware that sometimes they can have the same status but might have the warning icon which is expected and shouldn't change anything._

Order:
```
Recovering
Deploying
Deleting
Stopped
Running
NA
```

service-status-sorting

**Checklist**


weblancaster:
alfred test this

PR #2667, opened 2018-01-24
DCOS-20475: Fix residency validators

Poltergeist:
This fixes a legacy issue which was introduced into marathon a very long time ago. The issue was when you wanted to create a app with a persistent volume you needed to also define the residency setting, otherwise the volume would be present but not persistent ... unfortunately there was only one residency setting and changing it did not have any other purpose. With the latest changes Marathon is going to set residency if there is a persistent volume defined. So it is not mandatory to add it via the ui. 

With this PR we are removing the validators which were checking if the residency was set. And also we are removing the logic which was adding residency if a persistent volume was added. This has been done for single container and multi container services.

To test this take this JSON:
```JSON
{
  "id": "/foo",
  "instances": 1,
  "cpus": 0.1,
  "mem": 32,
  "cmd": "echo Running && sleep 1000",
  "container": {
    "type": "MESOS",
    "volumes": [
      {
        "containerPath": "data",
        "mode": "RW",
        "persistent": {
          "size": 32
        }
      }
    ]
  }
}
```

And bring it to the create service modal and put it in the JSON editor of the Single container form then hit Review and run and go back again. Check the JSON there should be no residency present but the JSON may look different.

Also bring the JSON to the JSON-configuration form (aka JSON only mode) there should be no validation errors and hit Review and run and go back the JSON should not be changed.

Extra points: If you deploy the app(single container service) and hit edit there should be a residency present in the JSON. 

For Pods (multi container service) Use the Form to add a local persistent volume and check if the residency is set in: `scheduling.residency`

Closes DCOS-20475

### DIFF
--- a/plugins/services/src/js/components/modals/CreateServiceModal.js
+++ b/plugins/services/src/js/components/modals/CreateServiceModal.js
@@ -96,7 +96,6 @@ const METHODS_TO_BIND = [
 const APP_VALIDATORS = [
   AppValidators.App,
   MarathonAppValidators.containsCmdArgsOrContainer,
-  MarathonAppValidators.complyWithResidencyRules,
   MarathonAppValidators.mustContainImageOnDocker,
   MarathonAppValidators.validateConstraints,
   MarathonAppValidators.validateLabels,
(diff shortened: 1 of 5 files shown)


Poltergeist:
run integration tests

Poltergeist:
run integration tests

PR #2671, opened 2018-01-24
DCOS-20465 - OSS Placement constraints format

Fabs:
This is an alternative approach to solve the form problem with placement constraints being on an invalid format. It was inspired by the way it is handled in volumes and health check.

inspiration

The current code of two branches combined achieves that:

handover-idea

Solution idea:

To make the parsers add an ERROR type transaction to the batch, that could be validated down on the line only by the component who needs it. Combined to handling incorrect complicate JSON the same way volumes and health check does. It is achieved in 4 steps.

1. Enhance TransactionTypes with ERROR type (on dcos-ui-plugins-private)
2. Enhance the parser to add those ERROR transactions to the batch when needed
3. Create a validator that detects the presence of ERROR transactions in the batch
4. Use that validator on the responsible component (placement constraints) that catches that and render an alternative component instead.

To become a full work solution it still needs:
- Unit tests for the new Transaction behavior and constraints parser
- Integration tests for when the configuration is wrong
- Refactor the code (it is just a big uglly block)
- Port the solution to other components who need it too (indicated by comments)
- Also make the solution available on batch updates, and other cases (now only on loading and changing the json)

Final design is a little bit different, but I believe more consistent with where the errors go
image

### DIFF
--- a/plugins/services/src/js/reducers/serviceForm/JSONReducers/Constraints.js
+++ b/plugins/services/src/js/reducers/serviceForm/JSONReducers/Constraints.js
@@ -1,4 +1,6 @@
+import { ERROR } from "#SRC/js/constants/TransactionTypes";
 import { findNestedPropertyInObject } from "#SRC/js/utils/Util";
+import Transaction from "#SRC/js/structs/Transaction";
 import { isEmpty } from "#SRC/js/utils/ValidatorUtil";
 import { JSONParser, JSONReducer } from "../common/Constraints";
 
@@ -19,12 +21,18 @@ module.exports = {
     const constraints = findNestedPropertyInObject(state, "constraints") || [];
 
     if (!Array.isArray(constraints)) {
-      return [];
+      return [new Transaction(["constraints"], "not-list", ERROR)];
     }
 
     return JSONParser(
       constraints.reduce(function(memo, constraint) {
+        if (constraint == null) {
+          return memo;
+        }
+
         if (!Array.isArray(constraint)) {
+          memo.push({ error: true });
+
           return memo;
         }
         const [fieldName, operator, value] = constraint;
--- a/plugins/services/src/js/reducers/serviceForm/common/Constraints.js
+++ b/plugins/services/src/js/reducers/serviceForm/common/Constraints.js
@@ -1,4 +1,9 @@
-import { ADD_ITEM, REMOVE_ITEM, SET } from "#SRC/js/constants/TransactionTypes";
+import {
+  ADD_ITEM,
+  REMOVE_ITEM,
+  SET,
+  ERROR
+} from "#SRC/js/constants/TransactionTypes";
 import Transaction from "#SRC/js/structs/Transaction";
 import { isEmpty } from "#SRC/js/utils/ValidatorUtil";
 
@@ -40,13 +45,44 @@ function processTransaction(state, { type, path, value }) {
 
 module.exports = {
   JSONParser(constraints) {
-    // Ignore non-array constraints
     if (!Array.isArray(constraints)) {
-      return [];
+      return [new Transaction(["constraints"], "not-list", ERROR)];
     }
 
     return constraints.reduce(function(memo, item, index) {
-      if (typeof item !== "object") {
+      if (Array.isArray(item)) {
+        memo.push(
+          new Transaction(
+            ["constraints", index, "value"],
+            "value-not-converted-to-object",
+            ERROR
+          )
+        );
+
+        return memo;
+      }
+
+      if (typeof item != "object") {
+        memo.push(
+          new Transaction(
+            ["constraints", index, "value"],
+            "value-not-object",
+            ERROR
+          )
+        );
+
+        return memo;
+      }
+
+      if (item.error) {
+        memo.push(
+          new Transaction(
+            ["constraints", index, "value"],
+            "value-is-malformed",
+            ERROR
+          )
+        );
+
         return memo;
       }
 
--- a/plugins/services/src/js/validators/PlacementsValidators.js
+++ b/plugins/services/src/js/validators/PlacementsValidators.js
@@ -1,5 +1,7 @@
 import ValidatorUtil from "#SRC/js/utils/ValidatorUtil";
 import { findNestedPropertyInObject } from "#SRC/js/utils/Util";
+import { ERROR } from "#SRC/js/constants/TransactionTypes";
+
 import PlacementConstraintsUtil from "../utils/PlacementConstraintsUtil";
 import { PROP_MISSING_ONE, SYNTAX_ERROR } from "../constants/ServiceErrorTypes";
 
@@ -138,6 +140,12 @@ const PlacementsValidators = {
 
       return errors;
     }, []);
+  },
+
+  validateNoBatchError(constraintTransactions) {
+    const isErrorTransaction = transaction => transaction.type === ERROR;
+
+    return !constraintTransactions.some(isErrorTransaction);
   }
 };
 
--- a/src/js/components/PlacementConstraintsSchemaField.js
+++ b/src/js/components/PlacementConstraintsSchemaField.js
@@ -20,7 +20,34 @@ import CreateServiceModalFormUtil
 import MarathonAppValidators
   from "#PLUGINS/services/src/js/validators/MarathonAppValidators";
 
+import PlacementValidators
+  from "#PLUGINS/services/src/js/validators/PlacementsValidators";
+
 const jsonReducer = combineReducers({ constraints: JSONReducer });
+
+const InvalidJsonField = props => (
+  <div>
+    <FieldLabel>
+      {props.label}
+    </FieldLabel>
+    <pre>
+      {JSON.stringify(props.json, props.replacer, props.space)}
+    </pre>
+    <FieldError>Unable to edit {props.fieldName}</FieldError>
+  </div>
+);
+
+InvalidJsonField.propTypes = {
+  fieldName: React.PropTypes.string.isRequired,
+  label: React.PropTypes.string.isRequired,
+  json: React.PropTypes.object.isRequired,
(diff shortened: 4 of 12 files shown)


ahoskins:
For anyone testing this locally: easiest way is to put 

```
formData.service.placement_constraint = "[\"hostname\", \"MAX_PER\", \"1\"]";
```

in line 58 of DeployFrameworkConfiguration.js, then deploy *beta-kafka**.

Fabs:
Now that all the code is here, it would also pay of to disable the EE plugin on config dev (to test it).
```js
       placement: {
          enabled: false
        }
```

Fabs:
run integration tests

Fabs:
@Poltergeist your opinion regarding those changes is *highly* appreciated :-D

src/js/components/PlacementConstraintsSchemaField.js
```diff
@@ -83,6 +110,20 @@ export default class PlacementConstraintsSchemaField extends Component {
...
     const { batch } = this.state;
 
+    const identity = (memo, transaction) => memo.concat(transaction);
+    const hasBatchErrors = !PlacementValidators.validateNoBatchError(
+      batch.reduce(identity, [])
+    );
+
+    if (hasBatchErrors) {
+      const fieldName = this.props.fieldProps.name;
+      const fieldValue = this.props.fieldProps.formData;
```
ahoskins: 💯

Fabs: I noticed that behavior too. I think when the parenthesis is not balanced this failure happens on an earlier stage and we receive an empty/null information that does not even come to the parser. 

I will see if I can narrow down this case quickly and understand why it happens.

Fabs:
run integration tests

src/js/components/PlacementConstraintsSchemaField.js
Fabs: I checked this by adding some debugging to the JSON parsers. They do not get triggered when the string is not JSON. That tells me that the step that verifies that is up in the flow, and hence I would have to implement extra logic there to catch this case.

I am creating a ticket to improve on this.
https://jira.mesosphere.com/browse/DCOS-20624

Fabs:
I am done here I believe (unless there are more changes). I have also a port of this solution to the EE component here: https://github.com/mesosphere/dcos-ui-plugins-private/pull/725

Fabs:
alfred test this please

Fabs:
Suite pass locally:
image

Fabs:
run integration tests

Fabs:
alfred test this please

src/js/components/PlacementConstraintsSchemaField.js
ahoskins: Thanks for creating ticket. I agree, not something for this PR 👍

Fabs:
run integration tests

Fabs:
alfred test this please

ahoskins:
run integration tests

plugins/services/src/js/reducers/serviceForm/common/Constraints.js
```diff
@@ -40,13 +45,44 @@ function processTransaction(state, { type, path, value }) {
...
+            ["constraints", index, "value"],
+            "value-not-converted-to-object",
+            ERROR
+          )
+        );
+
+        return memo;
+      }
+
+      if (typeof item != "object") {
```
orlandohohmeier: 🎨  Please always use `!==` or `===` except for the cases in which you really want to convert types.

plugins/services/src/js/reducers/serviceForm/JSONReducers/Constraints.js
```diff
@@ -19,12 +21,18 @@ module.exports = {
...
     }
 
     return JSONParser(
       constraints.reduce(function(memo, constraint) {
+        if (constraint == null) {
+          return memo;
+        }
+
         if (!Array.isArray(constraint)) {
+          memo.push({ error: true });
```
orlandohohmeier: 💡 I'd push an `Error` instead of an object with an error flag, as it would provide a more secure error condition (see https://github.com/dcos/dcos-ui/pull/2671/files#diff-5b04656bc39045c79cea0891f97950cbR77) and  allow us to pass a message.

plugins/services/src/js/reducers/serviceForm/common/Constraints.js
```diff
@@ -40,13 +45,44 @@ function processTransaction(state, { type, path, value }) {
...
+            ["constraints", index, "value"],
+            "value-not-object",
+            ERROR
+          )
+        );
+
+        return memo;
+      }
+
+      if (item.error) {
```
orlandohohmeier: screen shot 2018-01-29 at 11 26 07

🎨 Please improve this check to avoid false positives (e.g. `{error:''}`).

plugins/services/src/js/validators/PlacementsValidators.js
```diff
@@ -138,6 +140,12 @@ const PlacementsValidators = {
 
       return errors;
     }, []);
+  },
+
+  validateNoBatchError(constraintTransactions) {
```
orlandohohmeier: 🎨  Let's move this one to it's own util as it seems to be a rather general utility.

plugins/services/src/js/reducers/serviceForm/common/Constraints.js
Fabs: I tend to follow what "You don't know javascript" for this coercion style issues, on his book he says:

https://github.com/getify/You-Dont-Know-JS/blob/master/types%20%26%20grammar/ch4.md

More precisely.
```
Tip: Another place where coercion is guaranteed not to bite you is with the typeof operator. typeof is always going to return you one of seven strings (see Chapter 1), and none of them are the empty "" string. As such, there's no case where checking the type of some value is going to run afoul of implicit coercion. typeof x == "function" is 100% as safe and reliable as typeof x === "function". Literally, the spec says the algorithm will be identical in this situation. So, don't just blindly use === everywhere simply because that's what your code tools tell you to do, or (worst of all) because you've been told in some book to not think about it. You own the quality of your code.
```

I also went on because the linter did not complain. So not really sure this is an issue.

plugins/services/src/js/validators/PlacementsValidators.js
Fabs: Also don't agree much here, because we will make general something that is not at this moment. I believe if/when we make the batch error standard, then that could be refactored to its own utility. Right now it only servers for Placement Constraints.

plugins/services/src/js/reducers/serviceForm/JSONReducers/Constraints.js
Fabs: Makes a lot of sense yes.

src/js/components/PlacementConstraintsSchemaField.js
```diff
@@ -20,8 +20,35 @@ import CreateServiceModalFormUtil
 import MarathonAppValidators
   from "#PLUGINS/services/src/js/validators/MarathonAppValidators";
 
+import PlacementValidators
+  from "#PLUGINS/services/src/js/validators/PlacementsValidators";
+
 const jsonReducer = combineReducers({ constraints: JSONReducer });
 
+const InvalidJsonField = props => (
```
Fabs: Orlando said:
This looks like a good candidate for our component/pattern library, but we should rename the component to simply JsonField.

src/js/components/PlacementConstraintsSchemaField.js
```diff
@@ -83,6 +110,20 @@ export default class PlacementConstraintsSchemaField extends Component {
     const { label, errorMessage } = this.props;
     const { batch } = this.state;
 
+    const appendToList = (memo, transaction) => memo.concat(transaction);
```
Fabs: @orlandohohmeier said:

🎨 We might wanna add a some or every operator to the batch so that we can use a simple predicate instead of reducing the batch to an array and then iterating over it. /cc @Poltergeist @nLight

src/js/components/PlacementConstraintsSchemaField.js
Fabs: Done.

plugins/services/src/js/reducers/serviceForm/common/Constraints.js
Fabs: Done

plugins/services/src/js/reducers/serviceForm/common/Constraints.js
orlandohohmeier: Yeah, it's fine as is and you don't need to change it, but I'd prefer us to always use strict checks as these are the things that usually get lost during a refactor. P.S. Comments I flag with `🎨` are purely stylistic and it's up to you wether you want to adjust your implementation.

src/js/components/PlacementConstraintsSchemaField.js
Fabs: I could have reduced over to a boolean instead of a list. I actually had to use reduce here to implement map. Before we think about `some` and `every` (which are I believe more high-level abstractions), we should think of `filter`, but before filter, I think we actually need `map` (naturally we can also always reduce to achieve all of the above).

I also think the API of the Validator should not be `validate(batch)`, `but validate(list)`. Unless we see batch as a list subtype.

plugins/services/src/js/reducers/serviceForm/common/Constraints.js
Fabs: Alright. I will keep 🎨  in mind. I think that by justifying myself I am sharing my 🎨  opinion.

Fabs:
I merged this to soon 🔫 .

Here are the improvements. When we are happy there, I will also do EE. 
https://github.com/dcos/dcos-ui/pull/2682

PR #2672, opened 2018-01-24
DCOS-20485: Adjust implementation to improve performance

orlandohohmeier:
This PR adds a simple memoize utility and memoize the `cleanJobJSON`, `cleanServiceJSON` and `copyRawObject` utils to address performance issues due to very frequent invocations. It also adjusts the `flagSDKTask` utility to only flag tasks that aren't already flagged to further improve the performance. 

Closes DCOS-20485

### DIFF
--- a/plugins/services/src/js/structs/Framework.js
+++ b/plugins/services/src/js/structs/Framework.js
@@ -9,6 +9,21 @@ import Application from "./Application";
 import FrameworkSpec from "./FrameworkSpec";
 
 module.exports = class Framework extends Application {
+  constructor() {
+    super(...arguments);
+
+    // For performance reasons only one instance of the spec is created
+    // instead of creating a new instance every time a user calls `getSpec()`.
+    //
+    // State and other _useless_ information is removed to create a clean
+    // service spec.
+    //
+    // The variable is prefixed because `Item` will expose all the properties
+    // it gets as a properties of this object and we want to avoid any naming
+    // collisions.
+    this._spec = new FrameworkSpec(cleanServiceJSON(this.get()));
+  }
+
   getPackageName() {
     return this.getLabels().DCOS_PACKAGE_NAME;
   }
@@ -28,8 +43,11 @@ module.exports = class Framework extends Application {
     return ROUTE_ACCESS_PREFIX + (this.get("name") || "").replace(regexp, "");
   }
 
+  /**
+   * @override
+   */
   getSpec() {
-    return new FrameworkSpec(cleanServiceJSON(this.get()));
+    return this._spec;
   }
 
   getTasksSummary() {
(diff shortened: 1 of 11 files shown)


src/js/utils/StructUtil.js
```diff
@@ -9,7 +10,7 @@ const StructUtil = {
    *                      One of Object, Array, Date, String, Number, or Boolean.
    * @return {Object}     Cloned Object without Structs
    */
-  copyRawObject: function copyRawObject(obj) {
+  copyRawObject: memoize(function copyRawObject(obj) {
```
MatApple: I have to say I'm not sold on this as it negates the very point of this function to begin with and seems highly risky.

I expect a copy function to always return a new copy such that `input !== output`.

We could be swapping one problem for another here. Accidental mutations are very real

orlandohohmeier: I agree and I'm not happy with this solution either. I'll see if there is a way to either completely remove the need for this function or adjust the plugin system to not waste that many cycles to copy objects that are never really used.

orlandohohmeier:
TL;DR: @nLight and I will look into different ways to fix this as we agree that we don't need an hot fix and should rather look into a proper™ fix.

@MatApple @ahoskins I'm sorry that I didn't better explain the issues as well as my reasoning for running with this hot fix. These functions didn't change in the last month and it's no news that they're slow, but with the changes to the network and data layer (Mesos Stream) we're now seeing these functions causing crashes as the UI is now able to load an parse the state of rather large and busy clusters. 
Every time the store is update we're triggering a cascade of different updates that leads to the respective effects. I went with a simple memoize functions as the input data to these functions doesn't change as frequently as we trigger the updates (e.g. new Mesos but no new Marathon data).   In other words this "fix" relies on receiving  the same input not on garbage collecting the old objects. I went with my own naïve memoize as not all impl. use weak map which would prevent GC when used as a key.

src/js/utils/memoize.js
```diff
@@ -0,0 +1,28 @@
+import { isCallable } from "./ValidatorUtil";
```
nLight: ❤️

MatApple:
@orlandohohmeier @nLight instead of updating the state store so frequently with the stream events, could we have an adaptor that reduces those events in front of the store? That was the original implementation I worked with Tommy and Quinn on and it had not issues with large clusters and frequent events. The reason being that the state store (and the rest of our application) is tightly bound to the 2 second update interval. Doing this would also fix the gray bars in our charts. Putting a dam upstream would buy us time to fix the downstream bottlenecks.

orlandohohmeier:
@MatApple we're already limiting the number of events to 1s max which should be enough to not flood the system with events. Please keep in mind that with the changes to the network and data layer we're now able to handle super large clusters with a rather high number of running services and tasks which is the reason for the increased load.

orlandohohmeier:
@DanielMSchmidt Yeah, we definitely need tooling to measure performance and identify bottlenecks, but I like to stress that these weren't newly introduced issues, they only surfaced due to the UI being able to handle much larger clusters now.

DanielMSchmidt:
@orlandohohmeier I am aware of that, just want to make sure we can deliver a great experience also to our bigger customers in the future 👍

PR #2673, opened 2018-01-25
DCOS-20475: Fix residency validators

Poltergeist:
This fixes a legacy issue which was introduced into marathon a very long time ago. The issue was when you wanted to create a app with a persistent volume you needed to also define the residency setting, otherwise the volume would be present but not persistent ... unfortunately there was only one residency setting and changing it did not have any other purpose. With the latest changes Marathon is going to set residency if there is a persistent volume defined. So it is not mandatory to add it via the ui. 

With this PR we are removing the validators which were checking if the residency was set. And also we are removing the logic which was adding residency if a persistent volume was added. This has been done for single container and multi container services.

For Pods (multi container service) Use the Form to add a local persistent volume and check if the residency is set in: `scheduling.residency`

Closes DCOS-20475

### DIFF
--- a/plugins/services/src/js/reducers/JSONMultiContainerReducers.js
+++ b/plugins/services/src/js/reducers/JSONMultiContainerReducers.js
@@ -24,10 +24,7 @@ module.exports = {
   environment,
   scaling,
   labels,
-  scheduling(
-    state = { residency: {}, placement: { constraints: [] } },
-    transaction
-  ) {
+  scheduling(state = { placement: { constraints: [] } }, transaction) {
     const constraintsState = state != null && state.placement != null
       ? state.placement.constraints
       : [];
--- a/plugins/services/src/js/reducers/serviceForm/JSONReducers/MultiContainerResidency.js
+++ b/plugins/services/src/js/reducers/serviceForm/JSONReducers/MultiContainerResidency.js
@@ -1,7 +1,6 @@
-import { SET, ADD_ITEM, REMOVE_ITEM } from "#SRC/js/constants/TransactionTypes";
+import { SET } from "#SRC/js/constants/TransactionTypes";
 import { findNestedPropertyInObject } from "#SRC/js/utils/Util";
 import Transaction from "#SRC/js/structs/Transaction";
-import VolumeConstants from "../../../constants/VolumeConstants";
 
 module.exports = {
   JSONReducer(state, { type, path, value }) {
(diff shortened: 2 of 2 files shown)


plugins/services/src/js/reducers/JSONMultiContainerReducers.js
```diff
@@ -24,10 +24,7 @@ module.exports = {
   environment,
   scaling,
   labels,
-  scheduling(
-    state = { residency: {}, placement: { constraints: [] } },
```
DanielMSchmidt: Isn't this a problem in line 36 where we do residency.bind(this)?

DanielMSchmidt:
Alfred test this please :)

plugins/services/src/js/reducers/JSONMultiContainerReducers.js
Fabs: I think `residency` on line 36 comes from the import on line 24.

```js
import {
  JSONReducer as residency
} from "./serviceForm/JSONReducers/MultiContainerResidency";
```

weblancaster:
Alfred test this please

weblancaster:
If I understood correctly I shouldn't be able to create a service with persistent volume correct?

But I was able to create a service and Pods with persistent volume in both branches `master` and `ph/fix/DCOS-20475-remove-residency`

Let me know the test steps to reproduce the bug that way I can make sure the fix work

ahoskins:
@weblancaster Before is on CCM right now, After is on this branch. Notice that before the `scheduling.residency` field is in the JSON. This is for creating a pod service.

**before**:
screen shot 2018-01-29 at 10 18 08 am

**after**:
screen shot 2018-01-29 at 10 18 14 am

Poltergeist:
Hey @weblancaster so before it was also possible to create a pod with persistent volume, but the residency setting would always also be set. Fortunately this is not needed anymore. This is why we removed the coupling between persistent volumes and residency. This is removing complexity and a bug source. So functionality wise if you create a pod with persistent volume with this PR the only difference to before is that the residency is not set / will not be pushed to the server. Marathon will set it by default. By that the User will probably have no visible impact. So in both cases it will deploy in both cases residency will be set after deployment but only in the new case residency is not pushed to the server. @ahoskins pointed this also out.

@DanielMSchmidt To this case regarding the default value. This is only removing the default value for residency but we still will need to execute the residency reducer which is done in the function with the `bind(this)`. I hope this answers your question. So it is like @Fabs pointed out.

PR #2676, opened 2018-01-26
Update docs links to use new site

leemunroe:
When creating a service the links to docs were pointing to Marathon docs.

This PR updates those links to point to DC/OS docs and ensures they open in new tabs. This should be less confusing for the user.

cc @pavisandhu

PR #2678, opened 2018-01-27
feat(scripts): add basic link checker script

ahoskins:
Purpose of this script is to automate the process of detecting broken links to docs in our code.

**How it works:**
1. It loops through the file checked into git
2. It skips the files that are known files without links (png, gif, svg). I can probably add more to this blacklist.
3. It checks for direct hardcoded links in the files.
4. It then checks for usages of buildDocsURI.
5. It sends requests to these links, and prints the links to stdout that fail.

**Checklist**

### DIFF
--- a/scripts/are-links-broken
+++ b/scripts/are-links-broken
@@ -1,0 +1,42 @@
+#!/usr/bin/env bash
+
+# example usage: $ are-links-broken 1.11
+#  it will print the links that won't return 200 to stdout: <link> <http-status-code>
+
+docsVersion=$1
+
+files=$(git ls-files)
+for file in $files
+do
+  # skip files types we know won't contain links
+  matchesNonCodeFile=$(echo $file | grep -Eo "(\.gif|\.png|\.svg)")
+  if [ "${matchesNonCodeFile}" ]; then
+    continue
+  fi
+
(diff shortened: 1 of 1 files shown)


weblancaster:
@frontend-team

Today I spent couple mins monitoring the performance (CPU, MEM and DOM Nodes). that being said, now I believe we have multiple performance issues from different places.

Case 1: I noticed a spike in CPU from ~20/30% usage to over ~100% usage was coming from the `<Loader />` animation while navigating.
Test: Removed Loader and noticed a considerable performance improvement navigating and CPU drop not passing 100%.

Case 2: A known issue but got worse in the past month, when a table has MANY tasks and because of "real time" updates/changes the page becomes impossible to use and crashes at some point
Possible solution: give the user the possibility to request the latest data, or only request the data based on user choosing the timespan, or find the real root cause and fix it.

Case 3 (General): Considerably increase over time of DOM Nodes (from average 3k to over 35k), JS event listeners (from average 100 to over 300) and memory (interesting enough memory seems the least problematic average 400 MB)

Please don't take this as final word especially because I spent just a couple mins on this

ahoskins:
Closed moving to #2676

PR #2681, opened 2018-01-29
feat(EmptyStates): consolidate config value empty states

ahoskins:
We have changed the design pattern to use unicode 2014 character (&mdash;) to denote the empty state of value in tables. Frameworks already use this pattern because the forms were recently built. This PR changes pods and services to use the same pattern.

**Testing**:
1. Deploy a service. The review screen should should long dash for unconfigured values.
2. Deploy a pod. The review screen should should long dash for unconfigured values.
3. Deploy a framework. The review screen should should long dash for unconfigured values.
4. Look at the configuration tab of service, pod, and framework. They should all have long dash in this view also.

**Before**:
screen shot 2018-01-29 at 2 17 06 pm

**After**:
screen shot 2018-01-29 at 2 16 57 pm

**Checklist**

@leemunroe for visibility.

### DIFF
--- a/plugins/services/src/js/components/__tests__/ConfigurationMapTable-test.js
+++ b/plugins/services/src/js/components/__tests__/ConfigurationMapTable-test.js
@@ -4,6 +4,7 @@ const ReactDOM = require("react-dom");
 /* eslint-enable no-unused-vars */
 const TestUtils = require("react-addons-test-utils");
 const JestUtil = require("#SRC/js/utils/JestUtil");
+const EmptyStates = require("#SRC/js/constants/EmptyStates");
 const ConfigurationMapTable = require("../ConfigurationMapTable");
 
 describe("ConfigurationMapTable", function() {
@@ -132,7 +133,7 @@ describe("ConfigurationMapTable", function() {
     ).map(JestUtil.mapTextContent);
 
     expect(headerText).toEqual(["A", "B"]);
-    expect(cellText).toEqual(["1", "Not Configured"]);
+    expect(cellText).toEqual(["1", EmptyStates.CONFIG_VALUE]);
   });
 
   it("respects `placeholder` column property", function() {
--- a/plugins/services/src/js/components/ConfigurationMapBooleanValue.js
+++ b/plugins/services/src/js/components/ConfigurationMapBooleanValue.js
@@ -3,6 +3,7 @@ import React from "react";
 
 import ConfigurationMapValue from "#SRC/js/components/ConfigurationMapValue";
 import ValidatorUtil from "#SRC/js/utils/ValidatorUtil";
+import EmptyStates from "#SRC/js/constants/EmptyStates";
 
 /**
  * Render a boolean value as a <ConfigurationMapValue>, with it's values being
@@ -33,7 +34,7 @@ class ConfigurationMapBooleanValue extends React.Component {
 }
 
 ConfigurationMapBooleanValue.defaultProps = {
-  defaultValue: <em>Not Configured</em>,
+  defaultValue: <em>{EmptyStates.CONFIG_VALUE}</em>,
   options: {
     truthy: "Enabled",
     falsy: "Disabled"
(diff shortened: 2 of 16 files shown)


ahoskins:
Alfred run tests please

ahoskins:
Alfred test this please

ahoskins:
@orlandohohmeier Lee suggested we simply remove the rows from the config screen that previously had "Not Supported" (Container Image, Extended Runtime Priv, Force Pull on Launch). But after looking into doing this, it's not as simple as I thought, and I suspect there is a deeper reason why we show these rows on the config even when not set.

*So my question:* Why do we have the concept of "Not Supported" values on the review screen?

ahoskins:
run integration tests

plugins/services/src/js/components/__tests__/ConfigurationMapTable-test.js
```diff
@@ -132,7 +132,7 @@ describe("ConfigurationMapTable", function() {
     ).map(JestUtil.mapTextContent);
 
     expect(headerText).toEqual(["A", "B"]);
-    expect(cellText).toEqual(["1", "Not Configured"]);
+    expect(cellText).toEqual(["1", "\u2014"]);
```
weblancaster: why not importing `#SRC/js/constants/EmptyStates` and using here?

orlandohohmeier:
@ahoskins I don't think there was a deeper reason for showing "Not Supported" other than design asking us to do so. Maybe @Poltergeist know of any reason?

leemunroe:
@ahoskins @orlandohohmeier If I recall, when we were working on this there was a tight deadline and it was just easier to say `Not Supported` instead of having logic to hide the row. But I agree, we should ideally hide these rows. No need to show the user on the review screen or config tab if not supported.

plugins/services/src/js/components/__tests__/ConfigurationMapTable-test.js
ahoskins: Yes I should do that, good catch.

ahoskins:
@orlandohohmeier Okay - I've got a bit of confusion on this.  Currently on master branch we say "Not Configured" for some.  After this PR, these will for sure changing to the new empty state (&mdash;). But then for some other fields currently on master we say "Not Supported". And @leemunroe you're saying we should remove these rows? If we remove these rows why are we showing the other rows with empty state? What's the difference? It seems arbitrary to me.

In my opinion we should hold off modifying which rows do and do not show.

orlandohohmeier:
@ahoskins yeah, sounds like we need to clarify how we want to handle all these cases first, hope @leemunroe can help with this. I agree that we shouldn't change the scope of this PR and I'm sorry if my comment suggested to do so.

leemunroe:
Taking a deeper look. There's a mix of inconsistent patterns and errors in what we have today depending on what container runtime is selected. Here are some examples...

### Docker Engine
#### With all possible values configured in the `Service` tab
image
* Looks good
* I'm not sure why network mode is not configured here as it is [set to host](https://cl.ly/3s0W2Y2t2c3Y) by default in the UI (separate issue)

#### With some values left empty
image
* In this case `Disk` and `Command` should have `—` (which I believe your PR fixes)

### UCR
#### With all possible values configured in the `Service` tab
image
* As `Extended Runtime Priv.` is not supported by UCR we shouldn't list it here (@orlandohohmeier please confirm if this is true)

#### With some values left empty
image
* Looks like there are a few bugs here
* `Disk` in this case should be `—`
* `Container Image` we say `Not Supported` here but that is wrong AFAIK - it is supported therefore should have a `—`  (@orlandohohmeier please confirm if this is true)
* `Extended Runtime Priv.` is not supported by UCR therefore I would expect to not see it listed here
* `Force Pull on Launch` we say `Not Supported` here but that is wrong AFAIK - it is supported by UCR therefore should have a `—`  (@orlandohohmeier please confirm if this is true)
* `GPU` is missing from this list but it is supported by UCR therefore I would expect it to be here and have a `—`

ahoskins:
@leemunroe Thanks for the overview. A lot going on here. And looking at the code, and the inconsistencies you mentioned above, I think a proper refactor is needed. Agree @orlandohohmeier? This can be a task for a different PR.

ahoskins:
In this refactor it can be an opportunity to combine the config screen code for pods, services, and frameworks - getting rid of other inconsistencies like header sizes, spacing, edit buttons, etc. It's a big undertaking, but we should schedule this for soon.

nLight:
I accidentally rebased this one instead of the plugins 😅

ahoskins:
Alfred test this please

PR #2682, opened 2018-01-29
style(PlacementConstraintsSchemaField): fixes naming and error handling

Fabs:
Fixes the mistakes that were introduced by:
https://github.com/dcos/dcos-ui/pull/2671
- The Error type is more precise to detect than {error: true}
- Better naming

### DIFF
--- a/plugins/services/src/js/reducers/serviceForm/common/Constraints.js
+++ b/plugins/services/src/js/reducers/serviceForm/common/Constraints.js
@@ -74,7 +74,7 @@ module.exports = {
         return memo;
       }
 
-      if (item.error) {
+      if (item instanceof Error) {
         memo.push(
           new Transaction(
             ["constraints", index, "value"],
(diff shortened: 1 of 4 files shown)


Fabs:
This goes with: 
https://github.com/mesosphere/dcos-ui-plugins-private/pull/728/files

plugins/services/src/js/reducers/serviceForm/common/Constraints.js
```diff
@@ -74,7 +74,7 @@ module.exports = {
         return memo;
       }
 
-      if (item.error) {
+      if (item instanceof Error) {
         memo.push(
```
nLight: Does it make sense to push the actual Error message instead of `"value-is-malformed"` on the line 81?

Fabs: Could be, but then I would say that the error message should be "value-is-malformed". And I would split the consistency of messages across files. 

I would propose to use the same name on `new Error`, and the message who is sent, but not to expose error.msg

Fabs: Hum, changed my mind... will return item.message

Fabs:
if you feel like @nlight comments where addressed, please dismiss his review.

nLight:
L👀 king...

Fabs:
run system tests

Fabs:
run integration tests

PR #2683, opened 2018-01-29
DCOS-20672: fix(structs): add defaults to resources

nLight:
Default missing resource declaration to 0 to fix framework resource aggregation.

Spotted on soak cluster when there are few frameworks running in a group group would report 0 Disk even though frameworks inside the group report some disk usage.

Closes DCOS-20672

### DIFF
--- a/plugins/services/src/js/structs/ApplicationSpec.js
+++ b/plugins/services/src/js/structs/ApplicationSpec.js
@@ -80,10 +80,10 @@ module.exports = class ApplicationSpec extends ServiceSpec {
    */
   getResources() {
     return {
-      cpus: this.get("cpus"),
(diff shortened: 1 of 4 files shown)


orlandohohmeier:
@nLight  please open a PR to back-port the fix to DC/OS UI 1.10

juliangieseke:
Code LGTM, but UI isnt fixed?! After checking out your branch, anything else i need to do?!
image
image

nLight:
If a service doesn't use disk we show zero. The issue was that groups wouldn't aggregate disk usage and always show zero when there are frameworks running inside

orlandohohmeier:
LGTM! 🚢

PR #2685, opened 2018-01-29
[1.10] DCOS-20672: add defaults to resources

nLight:
Default missing resource declaration to 0 to fix framework resource aggregation.

Spotted on soak cluster when there are few frameworks running in a group group would report 0 Disk even though frameworks inside the group report some disk usage.

Closes DCOS-20675

### DIFF
--- a/plugins/services/src/js/structs/ApplicationSpec.js
+++ b/plugins/services/src/js/structs/ApplicationSpec.js
@@ -80,10 +80,10 @@ module.exports = class ApplicationSpec extends ServiceSpec {
    */
   getResources() {
     return {
-      cpus: this.get("cpus"),
-      mem: this.get("mem"),
-      gpus: this.get("gpus"),
-      disk: this.get("disk")
+      cpus: this.get("cpus") || 0,
+      mem: this.get("mem") || 0,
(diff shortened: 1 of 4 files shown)


nLight:
I don't see this happening on 1.10 though

nLight:
We aren't building Docker every time on Jenkins and we've removed cypress from the image 💥

nLight:
=========== PLEASE DON'T RERUN TESTS THEY WILL FAIL =================

MatApple:
@nLight I've asked Ben to point to a service in the 1.10 soak cluster that has this issue. He's currently in the car on his way to the office.

MatApple:
@nLight given the problem statement of the UI already reporting 0 when it should report > 0, adding default of 0 does not resolve the issue. After investigating, Marathon reports 0 and it shouldn't. Issue has been raised with Marathon on-call. https://mesosphere.slack.com/archives/C03UXUVH7/p1517106158000067

nLight:
In 1.11 I observed that groups would show 0 while having frameworks with >0 disk, the fix but applied to 1.11 fixed that issue.

PR #2686, opened 2018-01-29
DCOS_OSS-2186: Upgrade node to 8.9

DanielMSchmidt:
We tried to do as many small commits as possible to give enough review context.

After merging this we need to either move our test setup to the jenkins file or move our normal checks to a newer version of node


nLight:
⚔️  flicts!

DanielMSchmidt:
@nLight updated :)

juliangieseke:
foo run alfred tests bar

juliangieseke:
foo run alfred tests bar

juliangieseke:
run integration tests

package.json
```diff
@@ -72,6 +72,8 @@
     "tv4": "1.2.7"
   },
   "devDependencies": {
+    "node": "8.9.4",
```
weblancaster: 🎉 🎈

package.json
```diff
@@ -16,14 +16,14 @@
...
   "license": "Apache-2.0",
-  "engines": {
-    "node": "4.4.x",
-    "npm": "3.9.x"
-  },
   "browserslist": [
     "last 2 versions",
     "ie 11"
   ],
+  "engines": {
```
juliangieseke: can we remove this part?

DanielMSchmidt: I would like to leave it for a while. I know that we have it as a dependency, but as we as a team first need to switch to a node version that supports having these as dependencies and as we most likely will still need to switch to a different node version using `nvm` or something similar we should leave it like that.

juliangieseke: I'm not talking about removeing the check, just the duplicate part in the package.json.

DanielMSchmidt: Why duplicate?

nLight: There's no duplication as far as I can see.

DanielMSchmidt: The duplication was with the devDependencies, discussed offline that this could be a next PR

webpack/webpack.config.babel.js
```diff
@@ -46,10 +46,10 @@ new Promise(function(resolve, reject) {
   bootstrap.CSS = css;
 });
 
-function requireFromString(src, filename) {
+function requireFromString(src) {
   const Module = module.constructor;
   const sourceModule = new Module();
-  sourceModule._compile(src, filename);
+  sourceModule._compile(src, "");
```
Fabs: What does that do?

.nvmrc
```diff
@@ -1 +1 @@
-4.4
+8.9
```
Fabs: Line 22 of the README.md has wrong instructions about node and NPM. 

Node 4.4.x and NPM 3.9.x (and above) are **required**. We suggest using [nvm](https://github.com/creationix/nvm) or [n](https://github.com/tj/n) to keep multiple Node versions on your system.


PLUS but not blocking:
I think we also could add a little doc on upgrading from node 4.4 (running nvm install again, and npm install, or even other forms if you dont use nvm)

DanielMSchmidt: Good catch, got it updated 👍

DanielMSchmidt:
@nLight @juliangieseke Could you re-review 🙈

DanielMSchmidt:
@Fabs not now 😝

weblancaster:
Hey Daniel, let me know if I'm testing this correctly.

I checkout to your branch then ran `rm -rf node_modules && npm i && npm start`
But the output is an error.

screen shot 2018-03-06 at 8 38 03 am

DanielMSchmidt:
@weblancaster are you on the `8.9.4` node version? You can find out by running `node -v`. If not please do `nvm alias default 8.9.4`, open a new Tab, run `rm -rf node_modules && npm i && npm start` again

PR #2696, opened 2018-01-31
Disable chrome security to allow self signed https

nLight:
Cypress isn't able to open self signed HTTPS pages unless this option is disabled. https://github.com/cypress-io/cypress/issues/771

Blocks https://github.com/mesosphere/dcos-ui-plugins-private/pull/729


nLight:
Alfred test this

PR #2698, opened 2018-02-01
feat(FrameworkConfiguration): open to form when deploying

ahoskins:
Based on feedback from users, we want to open the service to the form when deploying. This also makes it the same as when editing a running framework.

screen recording 2018-02-01 at 03 56 pm

Closes DCOS-19301

**Checklist**

### DIFF
--- a/tests/pages/catalog/packages/package/PackageTab-cy.js
+++ b/tests/pages/catalog/packages/package/PackageTab-cy.js
@@ -114,6 +114,8 @@ describe("Package Detail Tab", function() {
         .get(".page-body-content .package-action-buttons button")
         .contains("Review & Run")
         .click();
+
+      cy.get(".modal .modal-header button").contains("Review & Run").click();
     });
 
     it("opens framework configuration when Review & Run clicked", function() {
--- a/plugins/services/src/js/pages/EditFrameworkConfiguration.js
+++ b/plugins/services/src/js/pages/EditFrameworkConfiguration.js
@@ -122,7 +122,6 @@ class EditFrameworkConfiguration extends mixin(StoreMixin) {
         handleRun={this.handleRun.bind(this)}
         onFormDataChange={this.onFormDataChange.bind(this)}
         onFormErrorChange={this.onFormErrorChange.bind(this)}
-        isInitialDeploy={false}
         packageDetails={packageDetails}
         deployErrors={deployErrors}
         formData={formData}
--- a/src/js/components/FrameworkConfiguration.js
+++ b/src/js/components/FrameworkConfiguration.js
@@ -39,7 +39,7 @@ class FrameworkConfiguration extends Component {
     const { activeTab, focusField } = this.getFirstTabAndField();
 
     this.state = {
-      reviewActive: props.isInitialDeploy,
+      reviewActive: false,
       activeTab,
       focusField,
       jsonEditorActive: false,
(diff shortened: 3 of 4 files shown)


ahoskins:
@leemunroe because he will be happy to see this 🎉

Fabs:
Let's say I wanna run Kafka, but then decide not. If I click cancel in the form, and confirm, I am sent back to the first page of the "starting kafka funnel". I believe I should get back to the catalog.

ahoskins:
@Fabs Agree, maybe not an ideal flow, but outside scope 👍

ahoskins:
I pushed to fix tests, should be good now!

ahoskins:
Alfred test this please

bstavroulakis:
@ahoskins Failing tests, please rebase to master to fix tests.

ahoskins:
run integration tests

ahoskins:
Alfred test this please

tests/pages/catalog/packages/package/PackageTab-cy.js
```diff
@@ -114,6 +114,8 @@ describe("Package Detail Tab", function() {
         .get(".page-body-content .package-action-buttons button")
         .contains("Review & Run")
         .click();
+
+      cy.get(".modal .modal-header button").contains("Review & Run").click();
```
juliangieseke: looks like this line breaks a few tests

juliangieseke: hm, one moment… this file is integration tests. forget it

Fabs:
@ahoskins do we have a ticket for the issue of the back button (clicking back after opening the launch, takes to edit instead of the catalog)? it is very confusing and I think it will soon become a UX issue. Also, although this is maybe out of scope, I think because this fix introduces that problem, we could check with @leemunroe to see if this is ok.

Fabs:
run system tests

Fabs:
run integration tests

DanielMSchmidt:
Run system tests

leemunroe:
@Fabs If I'm understanding correctly, I'm not sure if that proposal would make sense. If I hit cancel I expect the form (aka full page modal) to close and to see what I was looking at before I opened it. From a flow perspective, the user may be looking at Kafka, hit review and run, form opens, they realize they selected the wrong version of kafka, cancel, select different version, run again.

MatApple:
@ahoskins can you also please back port this to 1.11

ahoskins:
@backport-bot can you do this for me?

bstavroulakis:
@leemunroe Isn't is a bit strange for the user to press the same button twice? I would think so the "Review & Run" the first time is different than the "Review & Run" the second time?

Review & Run -> Review & Run -> Run Service

Fabs:
@leemunroe and @bstavroulakis. I think I might not have explained my idea properly. I agree with you @leemunroe. If I click Back/Cancel I should get back to the previous step I was in. *That is not happening anymore due to this PR*. If I change my mind while launching a service and click Back, I ended up going to the edit form, instead of back to the catalog.

DanielMSchmidt:
Hey @ahoskins we had this problem with the changed build directory for the marathon raml, that's why I am rebasing this for you (also to review it properly and get that high prio task merged 👍)

DanielMSchmidt:
run integration tests

PR #2699, opened 2018-02-01
DCOS_OSS-1564: Let's update webpack

DanielMSchmidt:
## ToDos

- [X] update most webpack plugins
- [X] `npm run build-assets`
- [X] `npm test`
- [x] `npm start`
- [X] `npm run testing`
- [X] Check out if build works
- [x] Get favicon working
- [x] Switch away from React in development version


CONTRIBUTING.md
```diff
@@ -153,10 +153,10 @@ If you want to add a new npm package to 'node_modules' you will need to `--save-
 2. Create a synced npm-shrinkwrap.json with devDependencies included by running
 
     ```
-    npm run shrinkwrap
+    npm run build-shrinkwrap
```
weblancaster: Why so verbose? `build`

Jenkinsfile
```diff
@@ -27,6 +27,15 @@ pipeline {
       }
     }
 
+    stage('Install Correct Node Version') {
+      steps {
+        ansiColor('xterm') {
+          sh 'npm install -g n'
```
weblancaster: Why are we using `n` to install "manually" the node version we want? I believe we should have the node version already available in the environment that it's running the jenkinsfile.

e.g: On Jenkins you can configure to run on docker environment `node:8.9.4-alpine`

And `n` hasn't been updated since `Oct 18, 2017` so I believe won't receive maintenance anymore

README.md
```diff
@@ -35,7 +35,7 @@ Node 4.4.x and NPM 3.9.x (and above) are **required**. We suggest using [nvm](ht
   ```sh
   npm run scaffold
   ```
-  This will create `webpack/proxy.dev.js` and `src/js/config/Config.dev.js`. These files aren't tracked by git and provide a place to override proxy and general dcos-ui configuration. Edit `webpack/proxy.dev.js` to point to the  address of the `dcos-vagrant` machine.
+  This will create `webpack.proxy.dev.js` and `src/js/config/Config.dev.js`. These files aren't tracked by git and provide a place to override proxy and general dcos-ui configuration. Edit `webpack.proxy.dev.js` to point to the  address of the `dcos-vagrant` machine.
```
weblancaster: I noticed that you removed the folder `webpack`, not sure what others think but I like that organization where we have something like `scripts`, `webpack` and etc.

webpack.config.js
```diff
@@ -0,0 +1,126 @@
...
+  ],
+  module: {
+    rules: [
+      {
+        test: /\.js$/,
+        exclude: /node_modules/,
+        use: {
+          loader: "babel-loader",
+          options: {
+            presets: requireAll([
```
weblancaster: Why not use `.babelrc`

webpack.config.js
```diff
@@ -0,0 +1,126 @@
...
+              "babel-preset-es2015",
+              "babel-preset-stage-3",
+              "babel-preset-react"
+            ])
+          }
+        }
+      },
+      // TODO: resource folders
+      // TODO: add image optimizaiton
```
weblancaster: Yes but to `webpack.prod.config.js` and not part of base that way we gain some local development build speed

webpack.proxy.template.js
```diff
@@ -22,5 +22,8 @@ module.exports = {
   //     req.url = req.url.replace(/^\/mesos/, '');
   //   }
   // },
-  "*": "http://m1.dcos"
+  "*": {
```
weblancaster: hoping we are going to keep the `webpack` folder to organize webpack modules instead having all in the root creating noise.

CONTRIBUTING.md
DanielMSchmidt: because it seemed like just shrinkwrap caused an infinite loop. It also gets executed when npm shrinkwrap is called, therefore calling itself over and over again

Jenkinsfile
DanielMSchmidt: @juliangieseke You had a reason for that right? Was it that we didn't want to change the dockerfile, yet?

README.md
DanielMSchmidt: Having the config in a sub-folder lead to a problem with the paths, by moving it root level we don't have that problem. We can move it back into a folder later, but first I did not want to deal with that complexity

webpack.config.js
DanielMSchmidt: 100% with you, but we had it like this before and I don't want to change everything in one single PR. I would suggest to move to a babelrc later on when we update babel 👍

webpack.config.js
DanielMSchmidt: I would like to have this sorted out by node environment or maybe using [webpacks new `target` feature](https://webpack.js.org/concepts/targets/) 🤔

DanielMSchmidt:
Thank you for taking the time to have a look at this early work @weblancaster 👍  I like your feedback a lot and I agree with you mostly in the goals 👍

CONTRIBUTING.md
weblancaster: Makes sense, thanks for the explanation

README.md
weblancaster: Usually path problems can be solved by `path.join(__dirname, 'path/to/somewhere')`. Would nice to spend more time in fixing the problem than removing the organization.

webpack.config.js
weblancaster: I mean there is a lot things being change here so adding `.babelrc` won't make big of a difference when reviewing anyways.

But at the same time I would not block the PR just because there is no `.babelrc`

webpack.config.js
weblancaster: Not sure if I understood your response @DanielMSchmidt 

What `targets` will help in `image optimizaiton` configuration?

README.md
bstavroulakis: I agree with Michael, this means that we haven't set our paths correctly to begin with and is a code smell.

Jenkinsfile
ahoskins: I'd +1 @weblancaster 's question about `n`. Not sure who active of project it is - especially since original creator TJ has his own company now. We already have a `.nvmrc`, so if we have to use something probably use nvm is better.

ahoskins: Also why do we need to use a node version manager at all. Can't we just `npm install -g node@8.9`?

package.json
```diff
@@ -101,51 +101,53 @@
     "eslint-plugin-react": "6.9.0",
     "esprima": "3.1.3",
     "extract-text-webpack-plugin": "1.0.1",
-    "file-loader": "0.8.5",
+    "file-loader": "1.1.6",
```
ahoskins: How did you arrive at using these versions? Is it the latest, or just a version that works with this new version of node?

webpack.config.js
bstavroulakis: Actually there is a .babelrc file, the problem is that we're loading .js files from the external plugins folder that don't contain the babel-react e.t.c. node module.

README.md
bstavroulakis: @DanielMSchmidt I'm working on grouping the config again under the webpack folder. Will commit here once done

CONTRIBUTING.md
DanielMSchmidt: 👍

Jenkinsfile
DanielMSchmidt: Yeah, I agree with y'all it should also already be in the docker container, therefore I'll remove it :)

README.md
DanielMSchmidt: @bstavroulakis Thanks for taking over 👍 
@weblancaster It wasn't that easy at least from my tries, having the plugins makes this a lot harder actually.

webpack.config.js
DanielMSchmidt: Yeah, we would need to load our own babelrc, parse it and stuff the output through the `requireAll` function (at least from my perspective)

webpack.config.js
DanielMSchmidt: Sorry @weblancaster I think I have got too much context in my head 🙈 

I don't want to have a `webpack.prod.config.js`. I find it hard to either merge things or duplicate a lot of code, both ways are error prone imho. I want to keep it in this file and do image optimization when the target is production and let it be if target is development. Does that make sense?

Jenkinsfile
juliangieseke: AFAIK npm 3.9 isnt capable of installing node versions - if it is, this would be way better 👍 
but this whole thing is only a dirty hack because we havent changed our Dockerfile yet, we will change this in future to be able to install the correct node version with `npm install` (which depends on #2686)

package.json
juliangieseke: AFAIK all of them are the latest version at the time we updated 👍

DanielMSchmidt: I took everything that was related to webpack, upgraded it to the latest version and wrote a webpack config from scratch

nLight:
I smoke tested the cluster, generally working, few css issue here and there.

Select component double arrow:
screen shot 2018-02-23 at 18 21 59

Charts don't work
screen shot 2018-02-23 at 18 22 38

Weird styling of the preformated areas
screen shot 2018-02-23 at 18 23 07

cluster name cutoff on a smaller screens
screen shot 2018-02-23 at 18 23 29

webpack/plugins/svg-compiler-plugin.js
```diff
@@ -1,8 +1,8 @@
-import fs from "fs";
-import glob from "glob";
-import path from "path";
-import SVGSprite from "svg-sprite";
-import vinyl from "vinyl";
+const fs = require("fs");
```
weblancaster: hey @DanielMSchmidt why are using `require` and not `import`?

webpack/webpack.production.js
```diff
@@ -23,10 +23,8 @@ module.exports = merge(common, {
   plugins: [
     new ModuleConcatenationPlugin(),
     new DefinePlugin({
-      "process.env": {
-        NODE_ENV: JSON.stringify("production"),
-        VERSION: packageInfo.version
-      }
+      "process.env.NODE_ENV": JSON.stringify("production"),
```
Fabs: A little weird

package.json
```diff
@@ -132,24 +132,27 @@
...
     "stylelint-config-dcos": "0.0.3",
     "stylelint-config-standard": "13.0.0",
     "svg-sprite": "1.3.1",
     "transform-loader": "0.2.3",
+    "uglifyjs-webpack-plugin": "1.2.0",
     "vinyl": "1.1.1",
-    "webpack": "1.14.0",
-    "webpack-dev-server": "1.14.1",
-    "webpack-notifier": "1.3.0"
+    "webpack": "3.10.0",
```
weblancaster: Why webpack v3 and not v4?

postcss.config.js
```diff
@@ -0,0 +1,14 @@
...
+  const shouldOptimize = env === "production" ? {} : false;
+
+  return {
+    parser: false,
+    plugins: {
+      autoprefixer: shouldOptimize,
+      "postcss-merge-rules": shouldOptimize,
+      "postcss-discard-duplicates": shouldOptimize,
+      "postcss-merge-longhand": shouldOptimize,
+      "postcss-discard-comments": shouldOptimize
```
weblancaster: I believe we can pass the configuration to minimize to webpack instead.
In the css-loader something like
```
options: {
         minimize: true
    }
```

package.json
DanielMSchmidt: Because 4 is so new that our dependencies haven't updated yet. Also most of the work was done already before 4 came out, so I prioritized ship it here.
I plan to revisit this in ~ 1 month. The API changes in webpack were only severe for the packages and not for the configuration, so upgrading should work smoothly once the ecosystem adopted

postcss.config.js
DanielMSchmidt: As far as I get [the docs](https://github.com/webpack-contrib/css-loader#minimize) you can pass in cssnano configuration there, but these things are postcss (which we need for autoprefixer). Feel free to do a follow-up PR to also add cssnano, would love to see the effects on the output

juliangieseke:
something went wrong when we tried to rebase this PR - work is continued in #2817

PR #2702, opened 2018-02-02
DCOS_OSS-1341: stay in group view when service/group deleted

bstavroulakis:
When deleting a group inside a group, instead of staying inside that group you're being thrown back to the root of the namespace.

Closes DCOS_OSS-1341

delete group

**Checklist**

### DIFF
--- a/plugins/services/src/js/structs/ServiceTree.js
+++ b/plugins/services/src/js/structs/ServiceTree.js
@@ -110,13 +110,16 @@ module.exports = class ServiceTree extends Tree {
 
     let parentFound = null;
 
-    this.list.forEach(child => {
+    for (let i = 0; i < this.list.length; i++) {
+      const child = this.list[i];
       if (child instanceof ServiceTree) {
         parentFound = child.getItemParent(id, this);
       } else if (child.getId() === id) {
         parentFound = this;
+
+        break;
       }
-    });
+    }
 
     return parentFound;
   }
--- a/plugins/services/src/js/structs/ServiceTree.js
+++ b/plugins/services/src/js/structs/ServiceTree.js
@@ -99,6 +99,24 @@ module.exports = class ServiceTree extends Tree {
     });
   }
 
+  getItemParent(id) {
+    if (!this.list || this.list.length === 0) {
+      return null;
+    }
+
+    let parentFound = null;
+
+    this.list.forEach(child => {
+      if (child instanceof ServiceTree) {
+        parentFound = child.getItemParent(id);
+      } else if (child.getId() === id) {
+        parentFound = this;
+      }
+    });
+
+    return parentFound;
+  }
+
   /**
    * @param {string} id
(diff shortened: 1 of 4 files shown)


plugins/services/src/js/structs/ServiceTree.js
```diff
@@ -99,6 +99,28 @@ module.exports = class ServiceTree extends Tree {
...
+    }
+
+    if (!this.list || this.list.length === 0) {
+      return null;
+    }
+
+    let parentFound = null;
+
+    this.list.forEach(child => {
+      if (child instanceof ServiceTree) {
```
Fabs: If we have nested groups here, I believe we can run in a bug where I have `parentFound = this` but it is the first element on the list, so I keep searching further, even when I already found the parent. This cause me to return null.

For instance, given this tree, the test `getItemParent("/test/foo")`, fails.
```js
      this.instance = new ServiceTree({
        id: "/",
        items: [
          {
            id: "/test",
            items: [
              {
                id: "/test/foo"
              },
              {
                id: "/test/bar"
              },
              {
                id: "/test/baz/boo",
                items: [
                  {
                    id: "/test/baz/boo/1"
                  }
                ]
              }
            ]
          },
          {
            id: "/alpha",
            cmd: "cmd"
          },
          {
            id: "/beta",
            cmd: "cmd",
            items: {
              DCOS_PACKAGE_FRAMEWORK_NAME: "beta"
            }
          }
        ]
      });
```

a solution would be: 
changing line 114 to: ` if (child instanceof ServiceTree && parentFound == null) {`

I tried changing 114 to return this, but it did not work :/

plugins/services/src/js/structs/ServiceTree.js
```diff
@@ -99,6 +99,28 @@ module.exports = class ServiceTree extends Tree {
...
+        parentFound = child.getItemParent(id, this);
+      } else if (child.getId() === id) {
+        parentFound = this;
```
Fabs: Instead of the fix on 114, I would also recommend to move away from forEach, and break/return on line 117.

It seems to me pointless to keep looking when we already found the item. I know it is premature optimization, but it is also more in line with a classic DFS.

bstavroulakis:
@Fabs That is true, yes I missed that. If you could please check now, I added the break to improve performance and fix the nested groups issue.

bstavroulakis:
run system tests

DanielMSchmidt:
run system tests

bstavroulakis:
The System Test that is failing " Services Applications creates an app with command health check" is a known broken system test. I'll fix this in a future PR. @Fabs if you could please re-check.

Fabs:
@bstavroulakis sorry it took me two days to look it again.

Fabs:
run system tests

juliangieseke:
run all tests

bstavroulakis:
run integration tests

PR #2703, opened 2018-02-02
DCOS-20331: Update DSL filter results in real time

bstavroulakis:
Update DSL filter results in real time.

image

**Checklist**
- [ ] Did you add a JIRA issue in a commit message or as part of the branch name?
- [ ] Did you add new unit tests?
- [ ] Did you add new integration tests?
- [ ] If this is a regression, did you write a test to catch this in the future?

PR #2704, opened 2018-02-02
DCOS-20332: DSL show dropdown on focus

bstavroulakis:
image

**Checklist**

### DIFF
--- a/src/js/components/DSLFilterField.js
+++ b/src/js/components/DSLFilterField.js
@@ -9,6 +9,7 @@ import DSLFormDropdownPanel from "./DSLFormDropdownPanel";
 const METHODS_TO_BIND = [
   "handleDismissClick",
   "handleDropdownClick",
+  "handleInputFocus",
   "handleIgnoreClick",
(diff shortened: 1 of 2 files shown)


bstavroulakis:
Alfred test this

bstavroulakis:
Run integration tests

DanielMSchmidt:
I like these changes a lot, it gives the user a way better experience. For me, this change leveled up the user interaction here, therefore I became a bit more critical: 
I think it is super neat that changes in the textbox automatically reflect in the dropdown, but on the same note having to press "apply" after changes in the dropdown feels a bit weird. Can you make a new story out of this or fix it if it is super quick? Maybe it is also planned like this, @leemunroe can you tell me more here? (You are the right person to ping for UX issues, right?) 

This gif shows the problem in detail:

filter-dropdown

bstavroulakis:
Yes, great idea @DanielMSchmidt there is a PR for that => https://github.com/dcos/dcos-ui/pull/2703

PR #2710, opened 2018-02-06
fix(JobUtil): need to preserve schedule id and concurrency policy

kensipe:
Closes COPS-2668

### DIFF
--- a/src/js/utils/JobUtil.js
+++ b/src/js/utils/JobUtil.js
@@ -78,9 +78,9 @@ const JobUtil = {
       Object.assign(spec.run, { docker });
     }
 
-    // default
-    var scheduleId = "default";
-    var schedulePolicy = "ALLOW";
+    // default values for id and policy
+    let scheduleId = "default";
+    let schedulePolicy = "ALLOW";
     // preserve id and concurrencyPolicy
     if (typeof spec.schedules != "undefined" && spec.schedules.length > 0) {
       scheduleId = spec.schedules[0].id;
--- a/src/js/utils/JobUtil.js
+++ b/src/js/utils/JobUtil.js
@@ -78,6 +78,15 @@ const JobUtil = {
       Object.assign(spec.run, { docker });
     }
 
+    // default
+    var scheduleId = "default";
+    var schedulePolicy = "ALLOW";
+    // preserve id and concurrencyPolicy
+    if (typeof spec.schedules != "undefined" && spec.schedules.length > 0) {
+      scheduleId = spec.schedules[0].id;
+      schedulePolicy = spec.schedules[0].concurrencyPolicy;
+    }
+
     // Reset schedules
     spec.schedules = [];
 
@@ -85,11 +94,11 @@ const JobUtil = {
     // defaults
     if (!schedule || schedule.runOnSchedule) {
       const {
-        id = "default",
+        id = scheduleId,
         enabled = true,
         cron,
         timezone,
-        concurrencyPolicy = "ALLOW",
+        concurrencyPolicy = schedulePolicy,
         startingDeadlineSeconds
       } = schedule || {};
 
--- a/src/js/utils/__tests__/JobUtil-test.js
+++ b/src/js/utils/__tests__/JobUtil-test.js
@@ -50,6 +50,49 @@ describe("JobUtil", function() {
       ]);
     });
 
+    it("job schedule maintains id and policy", function() {
+      const job = JobUtil.createJobFromFormModel({
+        general: { id: "test", cmd: "sleep 1000;" },
+        schedule: {
+          id: "fluffy",
+          cron: "* * * * *",
+          enabled: true,
+          concurrencyPolicy: "FORBID",
+          runOnSchedule: true
+        }
+      });
+
+      expect(job.getSchedules()).toEqual([
+        {
+          id: "fluffy",
+          cron: "* * * * *",
+          enabled: true,
+          concurrencyPolicy: "FORBID"
+        }
+      ]);
+    });
+
+    it("job schedule defaults concurrencyPolicy if not provided", function() {
+      const job = JobUtil.createJobFromFormModel({
+        general: { id: "test", cmd: "sleep 1000;" },
+        schedule: {
+          id: "fluffy",
+          cron: "* * * * *",
+          enabled: true,
+          runOnSchedule: true
+        }
+      });
+
+      expect(job.getSchedules()).toEqual([
+        {
+          id: "fluffy",
+          cron: "* * * * *",
+          enabled: true,
+          concurrencyPolicy: "ALLOW"
+        }
+      ]);
+    });
+
     it("removes schedule if deactivated", function() {
       const job = JobUtil.createJobFromFormModel({
         general: { id: "test", cmd: "sleep 1000;" },


kensipe:
Integration failure seems completely unrelated... is this flaky?

kensipe:
Run integration tests

kensipe:
run system tests

kensipe:
again... CI tests failure is completely unrelated

```
[cos-ui-oss-pipeline_PR-2710-LEF4AN63OUZTA2US7T4YWS2BDNEA22FHVVRUJZCGPOTM7GPCGQAA] Running shell script

+ dcos-system-test-driver -j1 -v ./system-tests/driver-config/jenkins.sh

Traceback (most recent call last):

  File "/usr/local/lib/python3.6/site-packages/pkg_resources/__init__.py", line 666, in _build_master

    ws.require(__requires__)

  File "/usr/local/lib/python3.6/site-packages/pkg_resources/__init__.py", line 984, in require

    needed = self.resolve(parse_requirements(requirements))

  File "/usr/local/lib/python3.6/site-packages/pkg_resources/__init__.py", line 875, in resolve

    raise VersionConflict(dist, req).with_context(dependent_req)

pkg_resources.ContextualVersionConflict: (six 1.11.0 (/usr/local/lib/python3.6/site-packages), Requirement.parse('six==1.10.0'), {'dcos-system-test-driver'})



During handling of the above exception, another exception occurred:



Traceback (most recent call last):

  File "/usr/local/bin/dcos-system-test-driver", line 6, in <module>

    from pkg_resources import load_entry_point

  File "/usr/local/lib/python3.6/site-packages/pkg_resources/__init__.py", line 3144, in <module>

    @_call_aside

  File "/usr/local/lib/python3.6/site-packages/pkg_resources/__init__.py", line 3128, in _call_aside

    f(*args, **kwargs)

  File "/usr/local/lib/python3.6/site-packages/pkg_resources/__init__.py", line 3157, in _initialize_master_working_set

    working_set = WorkingSet._build_master()

  File "/usr/local/lib/python3.6/site-packages/pkg_resources/__init__.py", line 668, in _build_master

    return cls._build_from_requirements(__requires__)

  File "/usr/local/lib/python3.6/site-packages/pkg_resources/__init__.py", line 681, in _build_from_requirements

    dists = ws.resolve(reqs, Environment())

  File "/usr/local/lib/python3.6/site-packages/pkg_resources/__init__.py", line 870, in resolve

    raise DistributionNotFound(req, requirers)

pkg_resources.DistributionNotFound: The 'six==1.10.0' distribution was not found and is required by dcos-system-test-driver

script returned exit code 1
```

juliangieseke:
run system tests

src/js/utils/JobUtil.js
```diff
@@ -78,18 +78,27 @@ const JobUtil = {
       Object.assign(spec.run, { docker });
     }
 
+    // default
+    var scheduleId = "default";
```
Fabs: we can use the `let` keyword here instead of var. Var scoping is more complicated to reason about due to hoisting.

src/js/utils/JobUtil.js
```diff
@@ -78,18 +78,27 @@ const JobUtil = {
...
+    var schedulePolicy = "ALLOW";
+    // preserve id and concurrencyPolicy
+    if (typeof spec.schedules != "undefined" && spec.schedules.length > 0) {
+      scheduleId = spec.schedules[0].id;
+      schedulePolicy = spec.schedules[0].concurrencyPolicy;
```
Fabs: I just wanna check the assumption here that the schedule[0] is present, then we will always have Id and Policy. Otherwise, this might cause a bug where we have an empty schedule policy and the default is ignored.

src/js/utils/JobUtil.js
```diff
@@ -78,18 +78,27 @@ const JobUtil = {
...
+
     // Reset schedules
     spec.schedules = [];
 
     // Only transfer schedule if checkbox is set, and create job with reasonable
     // defaults
     if (!schedule || schedule.runOnSchedule) {
       const {
-        id = "default",
+        id = scheduleId,
```
Fabs: The whole new behaviour here needs to be covered by unit tests on this file:

src/js/utils/__tests__/JobUtil-test.js

Fabs:
Maybe also @nLight or @Poltergeist could see whether this is aligned with past implementation. From my perspective, it solves the problem on the ticket and works as expected.

Fabs:
Also, I tried to manually change the json, and I run into the following error:

image

is it FORBID the right value? I looked here:
https://github.com/dcos/metronome/blob/97b8be1d94077282cf761002f7e74efc935bc511/api/src/main/resources/public/api/v0/schema/jobspec_v0.schema.json

Maybe it was fixed introduced recently and it is not yet on dcos/master?

src/js/utils/JobUtil.js
kensipe: if `schedules` is defined (it is an array) and it has at least 1 element... that element will be a schedule.   that schedule will always have an id (because it is required) and it should have a concurrencyPolicy.   I'll check to see if it is possibly for this to be undefined or empty.

kensipe:
@Fabs it is likely not possible to change on your cluster... but it is something that can change (for metronome 0.4.0) and the value should be maintained to whatever it is set to.

src/js/utils/JobUtil.js
kensipe: yeah... good catch... after some thinking on this... I know it is possible for this to be missing... I'll update.

kensipe: checked and verified... metronome automatically provides the ALLOW concurrencyPolicy if a policy isn't specified.  this value will always be defined here.

kensipe: although this tests out fine on metronome... it is possible (as I've tested) that the UI user switches to the JSON edit, removes it, switches to UI edit, then switches back to JSON... this will result in an undefined element :( .   Seems unlikely so you know it will happen.

kensipe: haha... will oddly the UI code already handles this in an ok way... the event that follows this code under these circumstances is `createJobSpecFromJob` which defaults the concurrencyPolicy if not provided.
although it is possible for this value to be undefined (in the context of the UI alone)...  it is handled well within the UI.   If sent to Metronome it is defaulted there.   We could add defensive code, but it would convolute the code in a way that isn't necessary... open to other opinions / preferences.

kensipe:
@Fabs I think we are ready to go!
all feedback incorporated... 2 new unit tests added:
1) test to make sure the schedule id isn't changed
2) test to make sure that a non defined concurrencyPolicy defaults correctly

kensipe:
run integration tests

kensipe:
run Alfred tests

ahoskins:
alfred test this please

kensipe:
seems like some flaky tests 👎 
test failure:  installs a community package.Universe installs a community package
has nothing to do with the work submitted

kensipe:
run integration tests

ahoskins:
@juliangieseke @nLight Are system tests suppose to be fixed?

nLight:
@ahoskins yes

nLight:
run system tests

nLight:
run integration tests

Fabs:
seems good now, the tests.
image

Poltergeist:
**Awesome** @kensipe Thanks for your contribution.

PR #2722, opened 2018-02-08
feat(serviceID): disable id field on service edit

Poltergeist:
Disable the id input field in the edit form to prevent the user to change a exisiting id.

Closes DCOS-20956

### DIFF
--- a/plugins/services/src/js/components/forms/GeneralServiceFormSection.js
+++ b/plugins/services/src/js/components/forms/GeneralServiceFormSection.js
@@ -322,6 +322,8 @@ class GeneralServiceFormSection extends Component {
       </span>
     );
 
+    const isEditPage = /\/edit\//g.test(global.location.hash);
+
     return (
       <div>
         <h1 className="flush-top short-bottom">
@@ -351,7 +353,12 @@ class GeneralServiceFormSection extends Component {
               </FormGroupHeading>
             </FieldLabel>
             <FieldAutofocus>
-              <FieldInput name="id" type="text" value={data.id} />
+              <FieldInput
+                name="id"
+                type="text"
+                value={data.id}
+                disabled={isEditPage}
+              />
             </FieldAutofocus>
             <FieldHelp>
               Give your service a unique name within the cluster, e.g. my-service.


DanielMSchmidt:
Alfred test this please :)

DanielMSchmidt:
😲 Alfred test this please tests all checks again? 😲

plugins/services/src/js/components/forms/GeneralServiceFormSection.js
```diff
@@ -351,7 +351,12 @@ class GeneralServiceFormSection extends Component {
               </FormGroupHeading>
             </FieldLabel>
             <FieldAutofocus>
-              <FieldInput name="id" type="text" value={data.id} />
+              <FieldInput
+                name="id"
+                type="text"
+                value={data.id}
+                disabled={/\/edit\//g.test(global.location.hash)}
```
DanielMSchmidt: I know this might be very nitty picky, but would you mind moving this in a variable that gives a bit more context? Something like 
```js 
const isEditPage = /\/edit\//g.test(global.location.hash);
```
maybe?

/cc @juliangieseke I know you have strong opinions against too many variables, would you still agree on this one or am I a prisoner of my own customs?

DanielMSchmidt:
Please dismiss my review if you don't feel the same way 👍

DanielMSchmidt:
Also @Poltergeist I tested this and I still can change the value in the JSON editor, is this a wanted behavior / out of scope for this ticket?

plugins/services/src/js/components/forms/GeneralServiceFormSection.js
juliangieseke: Well, putting this into a variable because it is hard to follow the inline regex would be a fix for a problem which should be fixed in a different way - IMHO a simple `global.location.hash.includes('edit')` would be sufficient and easier to read/understand 🤔 (assuming babel will polyfill it)

Poltergeist:
The JSON editor should always let the user do what they want it is considered to be the advanced mode for user which do not need the form.

Poltergeist:
Alfred you can't test this. 👯 👯‍♂️ 🎉

DanielMSchmidt:
run integration tests

DanielMSchmidt:
run integration tests

Poltergeist:
run Integration Tests

DanielMSchmidt:
@Poltergeist does this have a forward port? Just wanting to make sure we don't miss a spot

PR #2737, opened 2018-02-09
1.11: DCOS-21229: fix sorting in tables

DanielMSchmidt:
The sorting of the service task tables was broken, this PR aims to fix this. Namely fields that are not in the original JSON of the task type but fetched from other sources to be displayed are not being considered while sorting. As the "tie breaker" argument resorts the table anyways and we normally don't have nice test data for region and zone no one spotted the mistake earlier.

I added stubs for our integration tests for this part, the real change is much smaller than 500 lines.
I also did not refactor the table but let the leaky abstraction be as we plan to rewrite that part soon.

\cc @nLight pinging you so that you are aware of this problem for the next iteration 👍 

feb -12-2018 11-28-46

**Checklist**
- [X] Did you add new integration tests?

### DIFF
--- a/plugins/services/src/js/utils/TaskTableUtil.js
+++ b/plugins/services/src/js/utils/TaskTableUtil.js
@@ -3,6 +3,7 @@ import React from "react";
 /* eslint-enable no-unused-vars */
 import TableUtil from "#SRC/js/utils/TableUtil";
 import Util from "#SRC/js/utils/Util";
+import TaskUtil from "#PLUGINS/services/src/js/utils/TaskUtil";
 import TaskStatusSortingOrder from "../constants/TaskStatusSortingOrder";
 
 function getUpdatedTimestamp(model) {
@@ -18,6 +19,18 @@ const TaskTableUtil = {
 
       if (prop === "updated") {
         return getUpdatedTimestamp(item) || 0;
+      }
+
+      if (prop === "zone") {
+        return TaskUtil.getZoneName(item);
+      }
+
+      if (prop === "host") {
+        return TaskUtil.getHostName(item);
+      }
+
+      if (prop === "region") {
+        return TaskUtil.getRegionName(item);
       }
 
       if (prop === "status" && !hasGetter) {
--- a/plugins/services/src/js/containers/tasks/TaskTable.js
+++ b/plugins/services/src/js/containers/tasks/TaskTable.js
@@ -73,6 +73,7 @@ class TaskTable extends React.Component {
   getColumns() {
     var className = this.getClassName;
     var heading = ResourceTableUtil.renderHeading(TaskTableHeaderLabels);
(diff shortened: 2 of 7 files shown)


nLight:
run system tests

nLight:
run integration tests

plugins/services/src/js/utils/TaskTableUtil.js
```diff
@@ -3,6 +3,7 @@ import React from "react";
 /* eslint-enable no-unused-vars */
 import TableUtil from "#SRC/js/utils/TableUtil";
 import Util from "#SRC/js/utils/Util";
+import TaskUtil from "#PLUGINS/services/src/js/utils/TaskUtil";
```
juliangieseke: 😢

nLight: Hey @juliangieseke, what's wrong with that specific line?

juliangieseke: TaskUtil is broken, because it fetches data from the `CompositeState` - the more we use it, the severe the issue gets.

juliangieseke: Yeah, its not broken by its functionality, but from a design POV. Also its not a "Bug" which spreads in our codebase, but it is a (very) bad example which one could use as such - and copy it. thats why i said "more severe"

but in the end you are correct - we can refactor it all at once, thats why I approved it :)

DanielMSchmidt:
run integration tests

DanielMSchmidt:
@nLight the tests should be fixed now 👍

nLight:
Alfred test this

bstavroulakis:
run integration test

bstavroulakis:
run integration tests

DanielMSchmidt:
@ahoskins I think you are right, I will prepare a second PR to add these 👍 (don't want to block this blocker)

nLight:
run integration tests

plugins/services/src/js/utils/TaskTableUtil.js
```diff
@@ -20,6 +21,14 @@ const TaskTableUtil = {
         return getUpdatedTimestamp(item) || 0;
       }
 
+      if (prop === "zone") {
```
ahoskins: 🤔 I'm curious why we don't have a case for "region" also?

MatApple: Good catch

juliangieseke: blocking this PR because of andrews catch.

DanielMSchmidt:
@ahoskins @MatApple @juliangieseke thank you for the catch, I added a fix for that 👍

nLight:
run integration tests

nLight:
Service Create Form for external volume test has failed. Seems unrelated.

PR #2739, opened 2018-02-09
[backport 1.11] fix(find-broken-links): fix regex in script

ahoskins:
Previously was broken and it would stop matching after the port number. This fixes it. Now it simply matches http|https followed by an unlimited amount of (sane) valid URL characters.

Testing:
https://regex101.com/r/KwvvN5/1

**Checklist**


ahoskins:
run system tests

PR #2743, opened 2018-02-10
DCOS-21158: Add flush horizontal to secondary buttons in full screen form flows

leemunroe:
This PR adds `button-flush-horizontal` to secondary buttons in the full screen modals. Whenever there is a text button we should remove padding like this so alignment is flush with other grid elements. This can be seen when running a service or configuring a package (Back and Cancel buttons).

image

### DIFF
--- a/plugins/services/src/js/components/modals/CreateServiceModal.js
+++ b/plugins/services/src/js/components/modals/CreateServiceModal.js
@@ -861,7 +861,7 @@ class CreateServiceModal extends Component {
 
     return [
       {
-        className: "button-primary-link",
+        className: "button-primary-link button-flush-horizontal",
         clickHandler: this.handleGoBack,
         label
       }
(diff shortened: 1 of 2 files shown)


DanielMSchmidt:
Hey @leemunroe I rebased your PR because we had a problem with marathons documentation being expected in the wrong directory by our build script. Will review it now 👍

DanielMSchmidt:
@leemunroe For me secondary button means in this specific case back button, right? I tried to find a difference, but I couldn't. Can you give me any pointers on when I would see a difference (screen width, browser,...)?

leemunroe:
@DanielMSchmidt Yes. Usually it's "Back" or "Cancel" in modals. In some instances it is "Close". I can share some docs with you offline regarding button types.

bstavroulakis:
run integration tests

PR #2746, opened 2018-02-12
fix(raml-scripts): updates the download link for the raml files

juliangieseke:
looks like the marathon team has changed the folder structure.

to test this run `npm install` and watch for something saying "updating raml files" (looks like there was a `npm run update-raml` command, but it is gone :( )

### DIFF
--- a/scripts/pre-install
+++ b/scripts/pre-install
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 SCRIPT_PATH="$(dirname "$0")/$(dirname "$(readlink "$0")")"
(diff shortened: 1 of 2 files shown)


juliangieseke:
Alfred test this

nLight:
Alfred test this

juliangieseke:
i dont think that we want to backport this exact raml file (which marathon version is in 1.11)?

nLight:
Right now 1.11 points to `1.6.0-pre-218-g36ad9bd` I'm pretty sure we don't want to keep it at `pre`

PR #2751, opened 2018-02-12
Run integration tests file by file

DanielMSchmidt:
## Proof of Concept
This is a proof of concept, please be aware that there is no intention of getting this exact code to master.

**Question**: Will running the tests one by one help us switch back to a smaller jenkins node label? (Currently we use one which is way too big for what we do)

**Other Metrics**:

- How long does the test run take opposed to "normal"?
- Will the flackiness be positively effected by retrying individual files instead of everything?

**How to move forward if this is a success**:

My idea would be to discuss the findings in the team and identify what we need to change in our system test runner to have a similar experience

**How to move forward if this is a failure**:

If we can't get under the 1GB threshold we should discuss about different testing frameworks and see if we can somehow get a feeling for weather they improve the situation or not.

**Is there anything worse after taking this route?**

Yes, indeed. We have lost the overall video & the overall run xml for jenkins and we need to somehow come up with a solution on that and on our own (or see whats out there)

## Findings

**Answer**: Yes, 100% success rate in 4 tries on our old jenkins label. Jenkins is able to pick up the XML files and doesn't care if it's one or more (guessing from the readme), so this isn't a problem once we have adjusted the setup accordingly.

**Other Metrics**:

- Duration: 24 - 46 minutes depending on the amount of retriess (currently ~ 18-48 minutes)
- Flakiness: No more complete errors due to flaky tests

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -51,23 +51,7 @@ pipeline {
     stage('Unit Test') {
       steps {
         ansiColor('xterm') {
-          sh '''npm run test -- --coverage'''
-        }
-      }
-
-      post {
-        always {
-          junit 'jest/test-results/*.xml'
-          step([$class             : 'CoberturaPublisher',
-                autoUpdateHealth   : false,
-                autoUpdateStability: false,
-                coberturaReportFile: 'coverage/cobertura-coverage.xml',
-                failUnhealthy      : true,
-                failUnstable       : true,
-                maxNumberOfBuilds  : 0,
-                onlyStable         : false,
-                sourceEncoding     : 'ASCII',
-                zoomCoverageChart  : false])
+          sh '''npm run test -- --maxWorkers=2'''
         }
       }
     }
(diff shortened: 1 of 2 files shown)


nLight:
> Yes, indeed. We have lost the overall video & the overall run xml for jenkins and we need to somehow come up with a solution on that and on our own (or see whats out there)

We can ask Jenkins collecting `*.xml` and all the videos, I can show you how 👍

DanielMSchmidt:
As discussed in slack I would be ok with merging this as an interim solution 🙈 The integration tests itself can now be run on `mesos-med`, but jest not, therefore no downgrade here, yet

Jenkinsfile
```diff
@@ -51,23 +51,7 @@ pipeline {
...
-                autoUpdateHealth   : false,
-                autoUpdateStability: false,
-                coberturaReportFile: 'coverage/cobertura-coverage.xml',
-                failUnhealthy      : true,
-                failUnstable       : true,
-                maxNumberOfBuilds  : 0,
-                onlyStable         : false,
-                sourceEncoding     : 'ASCII',
-                zoomCoverageChart  : false])
+          sh '''npm run test -- --maxWorkers=2'''
```
DanielMSchmidt: I had to introduce this "hack" by @nLight to make sure we get the unit tests passing

DanielMSchmidt:
Oh, also the xml pickup issue I reported was a non-issue :)

PR #2754, opened 2018-02-12
fix(Endpoints): remove spaces from the comma delimited list of endpoints

bstavroulakis:
Remove spaces from the comma delimited list of endpoints.

This is a big problem for people that want to copy and paste to the terminal. They extra space that we add between the information (address, dns, vip e.t.c), throws errors in some cases when using them as raw input.

image

**Checklist**


MatApple:
@bstavroulakis do we need to do it for other fields too like address?

MatApple:
@bstavroulakis can you please back port this to 1.11?

PR #2755, opened 2018-02-13
Add fixture for multi region and zone pod

DanielMSchmidt:
**Discontinued**

It appears that we don't stub the nodes endpoint at all in the fixtures which means that we would need to add this stub here, too. This opens up a big rabbit hole and I don't feel ok with spending so much time in the fixtures right now. I will leave this here to remind us to be working on that somewhen in the future.

**What would need to happen?**

- [ ] stub away the nodes endpoint entirely
- [ ] add different regions and zones to the nodes
- [ ] in the marathon group tasks lists set the `agent: {id}` to the id of the nodes

### DIFF
--- a/tests/_fixtures/marathon-1-task/mesos-subscribe.js
+++ b/tests/_fixtures/marathon-1-task/mesos-subscribe.js
@@ -797,6 +797,96 @@ var response = JSON.stringify({
               disk: 0,
               mem: 0
             },
+            domain: {
+              fault_domain: {
+                region: {
+                  name: "eu-central-1"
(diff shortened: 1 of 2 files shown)


ahoskins:
@DanielMSchmidt I think it's better to track this kind of thing on a Jira ticket. Can checkout a branch with the Jira task ID in the name to link this work.

Especially since I'm sure there are many related tasks to improving our fixtures.

DanielMSchmidt:
As we will probably replace our stub server I will close this PR for now

PR #2760, opened 2018-02-14
Node duplication bug reproduction

Fabs:
Most of the discussions on how to reproduce this one is here:
https://jira.mesosphere.com/browse/DCOS-20507

image

### DIFF
--- a/plugins/nodes/src/js/pages/nodes/nodes-table/NodesTableContainer.js
+++ b/plugins/nodes/src/js/pages/nodes/nodes-table/NodesTableContainer.js
@@ -27,6 +27,15 @@ class NodesTableContainer extends mixin(StoreMixin, QueryParamsMixin) {
       },
       { name: "state", events: ["success"], suppressUpdate: false }
     ];
+
+    setInterval(this.messIt.bind(this), 1000);
+  }
+
+  messIt() {
+    const items = this.state.filteredNodes.list;
+    items.push(items[0]);
+    this.setState({filteredNodes: new NodesList({items: items})})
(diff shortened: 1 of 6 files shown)


Fabs:
image

Fabs:
https://cl.ly/0u1n0Z0D3P1j

juliangieseke:
some context? :) 
what exactly is the issue here? 
by pushing the first item, you basicly duplicate it. 
so the issue is that is is displayed twice?

plugins/nodes/src/js/pages/nodes/nodes-table/NodesTableContainer.js
```diff
@@ -27,6 +27,15 @@ class NodesTableContainer extends mixin(StoreMixin, QueryParamsMixin) {
       },
       { name: "state", events: ["success"], suppressUpdate: false }
     ];
+
+    setInterval(this.messIt.bind(this), 1000);
+  }
+
+  messIt() {
+    const items = this.state.filteredNodes.list;
```
weblancaster: Instead `this.state.filteredNodes.list` I would suggest `this.state.filteredNodes.list.slice(0)` that way avoiding shallow copy.

Fabs: I think this is intended, or at least it helps me to cause the bug I want to make happen. I will try to remove and see if it causes the bug.

Fabs:
On the second commit, this is a weird behavior, where it shows the Composite Store seeing two different truths at the same time.
image

PR #2688, opened 2018-01-30
feat(Pod): add DSL filter for pods

bstavroulakis:
Add DSL filters for pods.
They should look similar to the simple service tasks but keep the same functionality as the pods one has now.

Closes DCOS-20249

The following gif show the before and after:

pods

**Checklist**
- [ ] Did you add a JIRA issue in a commit message or as part of the branch name?
- [ ] Did you add new unit tests?
- [ ] Did you add new integration tests?
- [ ] If this is a regression, did you write a test to catch this in the future?

PR #2764, opened 2018-02-14
DCOS-21229: Sorting in pods table

DanielMSchmidt:
This PR aims to write integration tests for sorting pod instances and add the functionality which seems to be missing

**Checklist**

### DIFF
--- a/plugins/services/src/js/containers/pod-instances/PodInstancesTable.js
+++ b/plugins/services/src/js/containers/pod-instances/PodInstancesTable.js
@@ -11,6 +11,7 @@ import ExpandingTable from "#SRC/js/components/ExpandingTable";
 import Icon from "#SRC/js/components/Icon";
 import TimeAgo from "#SRC/js/components/TimeAgo";
 import Units from "#SRC/js/utils/Units";
+import TableUtil from "#SRC/js/utils/TableUtil";
 
 import Pod from "../../structs/Pod";
 import PodInstanceList from "../../structs/PodInstanceList";
@@ -138,6 +139,8 @@ class PodInstancesTable extends React.Component {
   }
 
   getColumns() {
+    const sortTieBreaker = "name";
+
     return [
       {
         className: this.getColumnClassName,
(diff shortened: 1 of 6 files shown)


plugins/services/src/js/containers/pod-instances/PodInstancesTable.js
```diff
@@ -138,6 +138,8 @@ class PodInstancesTable extends React.Component {
   }
 
   getColumns() {
+    // TODO: add sort function here
```
weblancaster: @DanielMSchmidt the sorting methods shouldn't be written in this method. You can follow what we have at `https://github.com/dcos/dcos-ui/blob/f0e9c4425538c107bebb8b06784ab914a8be07b9/plugins/services/src/js/containers/tasks/TaskTable.js#L77`

plugins/services/src/js/utils/InstanceUtil.js
```diff
@@ -38,13 +38,21 @@ const InstanceUtil = {
 
   getNode(task) {
     const nodesList = CompositeState.getNodesList();
-    const id = (task && task.agent && task.agent.id) || null;
+    let id = null;
+
+    if (task.agent && task.agent.id) {
+      id = task.agent.id;
+    } else if (task.agentId) {
```
weblancaster: no need for `else if`. Seems like we have preferred to use only `if` whenever possible.

ahoskins:
@DanielMSchmidt There is no attached Jira to this work. Please either update the branch to include the Jira or write in the description what this relates to.

plugins/services/src/js/utils/InstanceUtil.js
weblancaster: also, please take a look at @nLight comments here https://github.com/dcos/dcos-ui/pull/2688/files/a4354273ccf7f4f58d4818b8ee7483675243fbcc#diff-419e17a15d4143340f20a92b4aa4fe76

DanielMSchmidt:
@ahoskins Added it 👍

DanielMSchmidt:
Will do the backport once this is merged

PR #2765, opened 2018-02-14
DCOS 21173: Update status bar styles

leemunroe:
We have some old legacy styles from the dark UI in the status bars. This PR removes the dark line on status bars and removes the CSS that was making the bar smaller in headers. Making these more inline with our UI kit vision.

Before
image
image

After
image
image

Closes DCOS-21173


nLight:
run integration tests

PR #2770, opened 2018-02-15
Fix system tests for jenkinsfile

DanielMSchmidt:
I hope this will fix the fails on the jenkinsfiles
**Checklist**


nLight:
run integration tests

DanielMSchmidt:
Yay, it ran through 🎉

DanielMSchmidt:
As we don't rely on the Jenkinsfile Pipeline yet I would like to merge this soonish to test if it works reliably 👍

PR #2774, opened 2018-02-16
DCOS-21167: placement constraint show too early error

DanielMSchmidt:
My idea was to refactor the constraint fields into one component which tracks the dirty state (so if the value has been changed). That way I think I could hide the error if there has not been any editing yet. 

**Checklist**

### DIFF
--- a/src/js/components/PlacementConstraintsField.js
+++ b/src/js/components/PlacementConstraintsField.js
@@ -1,0 +1,184 @@
+import React, { Component } from "react";
+import PropTypes from "prop-types";
+import { Select, SelectOption } from "reactjs-components";
+
+import { findNestedPropertyInObject } from "#SRC/js/utils/Util";
+import DeleteRowButton from "#SRC/js/components/form/DeleteRowButton";
+import FieldError from "#SRC/js/components/form/FieldError";
+import FieldHelp from "#SRC/js/components/form/FieldHelp";
+import FieldInput from "#SRC/js/components/form/FieldInput";
+import FormGroup from "#SRC/js/components/form/FormGroup";
+import FormRow from "#SRC/js/components/form/FormRow";
+import FieldLabel from "#SRC/js/components/form/FieldLabel";
+import FormGroupHeading from "#SRC/js/components/form/FormGroupHeading";
+import FormGroupHeadingContent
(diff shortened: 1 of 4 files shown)


src/js/components/PlacementConstraintsField.js
```diff
@@ -0,0 +1,198 @@
...
+import FormGroupHeading from "#SRC/js/components/form/FormGroupHeading";
+import FormGroupHeadingContent
+  from "#SRC/js/components/form/FormGroupHeadingContent";
+
+import OperatorTypes from "#PLUGINS/services/src/js/constants/OperatorTypes";
+import PlacementConstraintsUtil
+  from "#PLUGINS/services/src/js/utils/PlacementConstraintsUtil";
+
+// &nbsp; to space the table rows
+const NBSP = "\u00A0";
```
weblancaster: I think there is a util that returns NBSP

src/js/components/PlacementConstraintsPartial.js
```diff
@@ -43,148 +22,30 @@ export default class PlacementConstraintsPartial extends Component {
...
 
   getPlacementConstraintsFields(data = []) {
-    const constraintsErrors = findNestedPropertyInObject(
-      this.props.errors,
-      "constraints"
-    );
     const hideValueColumn = data.every(function(constraint) {
       return PlacementConstraintsUtil.requiresEmptyValue(constraint.operator);
     });
+    debugger;
```
weblancaster: forgot `debugger` here

src/js/components/PlacementConstraintsPartial.js
```diff
@@ -43,148 +22,30 @@ export default class PlacementConstraintsPartial extends Component {
...
-            {!valueError &&
-              <FieldHelp>
-                {isLastField ? "A string, integer or regex value. " : NBSP}
-                {typeSettings &&
-                  !typeSettings.requiresValue &&
-                  !typeSettings.requiresEmptyValue &&
-                  "This field is optional."}
-              </FieldHelp>}
-          </FormGroup>
+      debugger;
```
weblancaster: And here.

DanielMSchmidt:
This is a duplicate of #2779

PR #2777, opened 2018-02-19
DCOS-21359: fix(ServicesContainer): checks and updates modalProps if needed

juliangieseke:
This PR adds a `getCleanModalProps` method to the `ServicesContainer`
which cleans and updates the `modalProps` object which is passed down to
the modals.
Before this change it was possible to have an invalid state in `modalProps`
like this: { service: undefined, id: "delete" }. This should be fixed now so
that `service` is always set, if `id` is. If anything is wrong the method
tries to correct the object as good as possible.

--- 

Reproduction of the Issue / Test this PR
Open DevTools & try this on Services "/" and in a Group:
1. create a Service/Pod/Package witl multiple instances (>10)
2. open delete modal for the Service/Pod/Package
3. copy the name (you will need it)
4. quickly delete it again (you have 2s)

=> if the modal says "cant delete" everything is good, if something breaks (watch console), this fix wasnt sufficient.

### DIFF
--- a/plugins/services/src/js/containers/services/ServicesContainer.js
+++ b/plugins/services/src/js/containers/services/ServicesContainer.js
@@ -432,20 +432,53 @@ class ServicesContainer extends React.Component {
     };
   }
 
-  getModals(service) {
-    const modalProps = Object.assign({}, this.state.modal);
+  /**
+   * This function validates the current modalProps and returns
+   * them validated/corrected
+   *
+   * This function updates the `service` information from DCOSStore -
+   * if possible - or deletes the modal `id` from the object - if
+   * necessary (effectivly closes the modal, when the respective
+   * service got deleted)
+   *
+   * Depending on the current state `modalProps` need to contain either:
+   *
+   * - service and id ({ service: Service, id: String })
+   *   when the modal `id` is open or should be opened.
+   *  when opening the modal, `modalProps.service` is not yet set,
+   *  so we have to set it.
+   *
+   * - only a service ({ service: Service })
+   *   when no modal is open or the current modal is to be closed
+   *   reson is, that even if no modal is open, they are rendered into
+   *   the dom and need a valid service
+   *
+   * @param {object} props - an object which contains id and a service ("modalProps")
+   * @param {ServiceTree} [service] - service information
+   * @returns {object} updated and cleaned up modal information (props)
+   */
+  getCorrectedModalProps(props, service) {
+    const modalProps = Object.assign({}, props);
 
-    // This is needed to refresh the state of the service from the store once the
-    // modal is loaded. In our delete group modal for example we need feedback
-    // of the service while it is being deleted
-    if (!modalProps.service) {
+    if (!modalProps.service && service) {
       modalProps.service = service;
-    } else {
+    }
+
+    if (
+      modalProps.service &&
+      DCOSStore.serviceTree.findItemById(modalProps.service.id)
+    ) {
       modalProps.service = DCOSStore.serviceTree.findItemById(
         modalProps.service.id
       );
+    } else if (modalProps.id) {
+      delete modalProps.id;
     }
 
+    return modalProps;
+  }
+
+  getModals(service) {
     return (
       <ServiceModals
         actions={this.getActions()}
@@ -453,7 +486,7 @@ class ServicesContainer extends React.Component {
         clearError={this.clearActionError}
         onClose={this.handleModalClose}
         pendingActions={this.state.pendingActions}
-        modalProps={modalProps}
+        modalProps={this.getCorrectedModalProps(this.state.modal, service)}
       />
     );
   }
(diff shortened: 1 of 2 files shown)


ahoskins:
Tests! 💥

juliangieseke:
Yup, rendering `null` wasnt a good idea 😅`<div />` should do it 🤔

DanielMSchmidt:
@juliangieseke Isn't this WIP as the tests are still missing?

juliangieseke:
true, also the tests are still failing :( not sure how to fix this, maybe this approach works?!

juliangieseke:
run integration tests

juliangieseke:
run system tests

juliangieseke:
🤞

plugins/services/src/js/containers/services/ServicesContainer.js
```diff
@@ -432,28 +432,54 @@ class ServicesContainer extends React.Component {
...
+   * @returns {object} updated and cleaned up modal information (props)
+   */
+  getUpdatedModalProps(stateService, service) {
+    const modalProps = Object.assign({}, stateService);
 
-    // This is needed to refresh the state of the service from the store once the
-    // modal is loaded. In our delete group modal for example we need feedback
-    // of the service while it is being deleted
-    if (!modalProps.service) {
+    // modalProps.service only is set, when a modal is already open, if we want to
```
ahoskins: This method is already pretty confusing, and I don't think having all these comments in the method body helps. I recommend putting all the comments above (and try to shorten a bit).

plugins/services/src/js/containers/services/ServicesContainer.js
```diff
@@ -432,28 +432,54 @@ class ServicesContainer extends React.Component {
...
+   * of the service while it is being deleted. In other places (after deleting
+   * an app) on the other hand, it is possible that the modal is still open,
+   * DCOSStore returns undefined but we still need the (old) information stored
+   * in modalProps.service - and want to close the modal.
+   *
+   * @param {object} stateService - an object which contains id and a service
+   * @param {ServiceTree} [service] - service information
+   * @returns {object} updated and cleaned up modal information (props)
+   */
+  getUpdatedModalProps(stateService, service) {
```
ahoskins: Maybe rename this to be more specific for what the method is doing?

plugins/services/src/js/containers/services/ServicesContainer.js
```diff
@@ -432,28 +432,54 @@ class ServicesContainer extends React.Component {
...
+    // modal id from the object - if necessary.
+    if (
+      modalProps.service &&
+      DCOSStore.serviceTree.findItemById(modalProps.service.id)
+    ) {
       modalProps.service = DCOSStore.serviceTree.findItemById(
         modalProps.service.id
       );
+    } else if (modalProps.id) {
+      delete modalProps.id;
```
ahoskins: This case looks to always return `{}`, so I think you should just `return {}` here.

juliangieseke: tried this before, but it breaks some modals :(

plugins/services/src/js/containers/services/ServicesContainer.js
juliangieseke: hm, any Idea how this should be named instead?

plugins/services/src/js/containers/services/ServicesContainer.js
juliangieseke: Hm, I will try to clarify them a little bit tomorrow. The issue is that this method basicly does two things: "open the modal" and "close the modal" - but not only one modal, but all of them (`id` describes the modal which should be open right now)

plugins/services/src/js/containers/services/ServicesContainer.js
juliangieseke: let me add a bit more context: when `id` is unset, this basicly says "close the modal", since the modal is still open in the next render-interval, the `service` has to be still valid, also unsetting `service` here would break a still-open-but-to-be-closed modal…

ahoskins: Ok makes sense.

plugins/services/src/js/containers/services/ServicesContainer.js
ahoskins: Something like `getModalWithPropsWithCorrectedService`.

juliangieseke:
found a new issue: when deleting the last service inside a group, it throws you back to the top level list 😢

nLight:
@juliangieseke it's a known issue. https://jira.mesosphere.com/browse/DCOS_OSS-2004

juliangieseke:
yup, seems to be fixed on master (was testing the 1.11 branch). phew…

juliangieseke:
run integration tests

juliangieseke:
run integration tests

PR #2780, opened 2018-02-20
DCOS-21337: Add linearBackoff retry to the Mesos stream

nLight:
We have a retry in the https://github.com/dcos-labs/mesos-client but it will catch only the Network errors and in a very stubborn fashion. This PR addresses all the parsing errors that can happen when processing messages and  consuming the stream.

`linearBackoff` ensures that we don't bombard the server with retries.

To enable easier testing I added `rxjs-marbles` library that provides some helpers.

To test this in the browser you could emulate the network split and monitor the Networking tab in the Inspector. You could also add some `throw` and `console.log` if that pleases you.

**Checklist**

Closes DCOS-21337

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
   },
   "dependencies": {
     "@dcos/http-service": "0.2.1",
-    "@dcos/mesos-client": "0.1.10",
+    "@dcos/mesos-client": "0.2.0",
     "babel-polyfill": "6.23.0",
     "browser-info": "0.4.0",
     "classnames": "2.2.3",
@@ -127,6 +127,7 @@
     "raml-validator-loader": "0.1.10",
     "react-hot-loader": "1.3.1",
     "react-test-renderer": "15.4.1",
+    "rxjs-marbles": "2.3.1",
     "source-map-loader": "0.1.5",
     "string-replace-webpack-plugin": "0.0.3",
     "style-loader": "0.13.1",
--- a/src/js/stores/MesosStateStore.js
+++ b/src/js/stores/MesosStateStore.js
@@ -15,10 +15,13 @@ import {
 } from "../constants/EventTypes";
 import MesosStateUtil from "../utils/MesosStateUtil";
 import pipe from "../utils/pipe";
+import { linearBackoff } from "../utils/rxjsUtils";
 import { MesosStreamType } from "../core/MesosStream";
 import container from "../container";
 import * as mesosStreamParsers from "./MesosStream/parsers";
 
+const MAX_RETRIES = 5;
+const RETRY_DELAY = 5000;
 const METHODS_TO_BIND = ["setState", "onStreamData", "onStreamError"];
 
 class MesosStateStore extends GetSetBaseStore {
@@ -70,7 +73,7 @@ class MesosStateStore extends GetSetBaseStore {
     const getMasterRequest = request(
       { type: "GET_MASTER" },
       "/mesos/api/v1?get_master"
-    ).retry(3);
+    ).retryWhen(linearBackoff(MAX_RETRIES, RETRY_DELAY));
 
     const parsers = pipe(...Object.values(mesosStreamParsers));
     const dataStream = mesosStream
@@ -102,6 +105,7 @@ class MesosStateStore extends GetSetBaseStore {
     this.stream = waitStream
       .concat(eventTriggerStream)
       .debounceTime(Config.getRefreshRate() * 0.5)
+      .retryWhen(linearBackoff(MAX_RETRIES, RETRY_DELAY))
       .subscribe(this.onStreamData, this.onStreamError);
   }
 
--- a/src/js/utils/rxjsUtils.js
+++ b/src/js/utils/rxjsUtils.js
@@ -1,0 +1,18 @@
+import { timer } from "rxjs/observable/timer";
+
+import "rxjs/add/operator/delayWhen";
+import "rxjs/add/operator/scan";
+
+/* eslint-disable import/prefer-default-export */
+export function linearBackoff(maxRetries = 5, delay = 1000, scheduler = null) {
+  return errors =>
+    errors
+      .scan((count, error) => {
+        if (count >= maxRetries) {
+          throw error;
+        }
+
+        return count + 1;
+      }, 0)
+      .delayWhen(val => timer(val * delay, null, scheduler));
+}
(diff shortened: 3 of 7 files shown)


src/js/utils/rxjsUtils.js
```diff
@@ -0,0 +1,19 @@
+import { timer } from "rxjs/observable/timer";
+
+import "rxjs/add/operator/delayWhen";
+import "rxjs/add/operator/retryWhen";
+import "rxjs/add/operator/scan";
+
+/* eslint-disable import/prefer-default-export */
+export function linearBackoff(maxRetries = 5, delay = 1000, scheduler = null) {
```
weblancaster: This is a really cool solution.

Have you looked at how AWS has figured out a "better way" to use exponential backoff?
https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/

src/js/utils/rxjsUtils.js
```diff
@@ -0,0 +1,19 @@
+import { timer } from "rxjs/observable/timer";
+
+import "rxjs/add/operator/delayWhen";
+import "rxjs/add/operator/retryWhen";
```
weblancaster: is `retryWhen` included here only for the test or is it not in use at all?

src/js/utils/rxjsUtils.js
nLight: I haven't, that's a good read BTW, thank you! I wonder if we need this kind of backoff in this case 🤔

orlandohohmeier: I very much like the idea of using jitter to spread the requests but don't think that we need this kind of backoff strategy in this case. AWS develop this technique to improve the DynamoDB *write* performance (note that writes are blocking) when a huge number of clients tries to do so at the same time. Using a linear or exponential backoff should be sufficient,  as we only read data and don't have a huge number of clients connecting at the same time.

src/js/utils/rxjsUtils.js
```diff
@@ -0,0 +1,18 @@
...
+  return errors =>
+    errors
+      .scan((count, error) => {
+        if (count >= maxRetries) {
+          throw error;
+        }
+
+        return count + 1;
+      }, 0)
+      .delayWhen(val => timer(val * delay, null, scheduler));
```
Fabs: I like it that you make it flexible by passing the scheduler. But since it looks like you are not using it for tests, nor there is a test who covers this behavior, I would rather not have the linearBackoff care about it.

Other option would be to test passing a custom scheduler.

src/js/stores/MesosStateStore.js
```diff
@@ -70,7 +73,7 @@ class MesosStateStore extends GetSetBaseStore {
     const getMasterRequest = request(
       { type: "GET_MASTER" },
       "/mesos/api/v1?get_master"
-    ).retry(3);
+    ).retryWhen(linearBackoff(MAX_RETRIES, RETRY_DELAY));
```
Fabs: So I am still asking myself, couldn't we have this rxjsUtils as part of the mesos client. I  think it would even make sense to move this getMasterRequest as a stream that we could also import from mesos client. Maybe it could also expose rxjs utils, and export the linearBackoff so we can use somewhere else too?

Fabs: Or maybe even we want to put on a new rxjsUtil packages, in this case, I believe we would merge here first and leave a todo to extract if/when it spreads.

src/js/utils/rxjsUtils.js
Fabs: Agree with @orlandohohmeier. @weblancaster this makes me think of network collision strategies like CSMA/CD (https://en.wikipedia.org/wiki/Carrier-sense_multiple_access_with_collision_detection)

package.json
```diff
@@ -30,7 +30,7 @@
   },
   "dependencies": {
     "@dcos/http-service": "0.2.1",
-    "@dcos/mesos-client": "0.1.10",
+    "@dcos/mesos-client": "0.2.0",
```
nLight: updated the PR to include the changes https://github.com/dcos-labs/mesos-client/pull/7

src/js/stores/MesosStateStore.js
nLight: The intention behind the `mesos-client` is to provide a pair of primitives one could build upon. Those are just `request` and `stream` as this is everything Mesos provides according to the [operator api documentation](mesos.apache.org/documentation/latest/operator-http-api/). It's a bare bones solution and because of that it is not opinionated. 

Down the road I'd like seeing the [suggested backoff strategy](http://mesos.apache.org/documentation/latest/operator-http-api/#heartbeat) inside the `mesos-client` as an option. But I don't really thing `rxjsUtils` with all the strategies belongs there.

Feel free to challenge me 👍

src/js/utils/rxjsUtils.js
nLight: I agree that it looks like a "just in case" passage, but I wanted to be a good RxJS citizen 👮

nLight:
run all tests

nLight:
Hey @orlandohohmeier I tested the code and it wouldn't retry errors it actually reconnected to the endpoint every time

src/js/utils/rxjsUtils.js
ahoskins: Careful @Fabs, you're giving me flashbacks to networking architecture courses I had to take in University... 😏

Fabs: Haha. Someone got the Knuth books in HH :D, and there is also a CLRS. Life is remembering 🐰

orlandohohmeier:
@nLight it properly reconnects if the get master requests fails but fails to reconnect if subscribe fails as you can see in the following two recordings:

*GET_MASTER Request*
_It seems to properly reconnect, but shouldn't retry that often 



*SUBSCRIBE Request*
_Please note that there is only one "blocked" request instead of 5._

nLight:
@orlandohohmeier I implemented the fix, thank you!

ahoskins:
Just surfacing this Jira here. Briefly talked with @leemunroe about it: https://jira.mesosphere.com/browse/DCOS-21371

nLight:
@ahoskins have you `npm install`ed? :)

orlandohohmeier:
@ahoskins I  run `npm install`
LGTM! 🚢

PR #2782, opened 2018-02-21
Fabs/1.11/fix/dcos 21361 nodes resource failure backports

Fabs:
Backports https://github.com/dcos/dcos-ui/pull/2781

### DIFF
--- a/src/js/structs/Node.js
+++ b/src/js/structs/Node.js
@@ -39,8 +39,13 @@ class Node extends Item {
   }
 
   getUsageStats(resource) {
-    const total = this.get("resources")[resource];
-    const value = this.get("used_resources")[resource];
+    const total = (this.get("resources") || {})[resource];
+    const value = (this.get("used_resources") || {})[resource];
+
+    if (total === undefined || value === undefined) {
+      return { percentage: 0, total: 0, value: 0 };
+    }
+
     const percentage = Math.round(100 * value / Math.max(1, total));
 
     return { percentage, total, value };
--- a/src/js/structs/__tests__/Node-test.js
+++ b/src/js/structs/__tests__/Node-test.js
@@ -108,18 +108,36 @@ describe("Node", function() {
   });
 
   describe("#getUsageStats", function() {
-    it("returns usage stats for given resource", function() {
-      const node = new Node({
-        resources: { cpus: 10 },
-        used_resources: { cpus: 5 }
-      });
-      const stats = {
-        percentage: 50,
-        total: 10,
-        value: 5
-      };
+    describe("with resources from mesos", function() {
+      it("returns usage stats for given resource", function() {
+        const node = new Node({
+          resources: { cpus: 10 },
+          used_resources: { cpus: 5 }
+        });
+        const stats = {
+          percentage: 50,
+          total: 10,
+          value: 5
+        };
 
-      expect(node.getUsageStats("cpus")).toEqual(stats);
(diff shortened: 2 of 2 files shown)


src/js/structs/Node.js
```diff
@@ -39,8 +39,13 @@ class Node extends Item {
   }
 
   getUsageStats(resource) {
-    const total = this.get("resources")[resource];
-    const value = this.get("used_resources")[resource];
+    const total = (this.get("resources") || {})[resource];
```
DanielMSchmidt: I wouldn't want to block this PR for this, but I think having a function like this to use is already a pattern for us:

```js
getResources() {
	return this.get("resources") || {};
}
```

src/js/structs/Node.js
```diff
@@ -39,8 +39,13 @@ class Node extends Item {
...
+    const value = (this.get("used_resources") || {})[resource];
+
+    if (total === undefined || value === undefined) {
```
DanielMSchmidt: Are we sure it can't be `null`? I think typeof checks might give the reader more context and us more security. Like 
```js 
typeof total === "number"
```

src/js/structs/__tests__/Node-test.js
```diff
@@ -108,18 +108,36 @@ describe("Node", function() {
...
-        total: 10,
-        value: 5
-      };
+    });
 
-      expect(node.getUsageStats("cpus")).toEqual(stats);
+    describe("without resources from mesos", function() {
+      it("returns empty resources and used_resources", function() {
+        const node = new Node({
+          resources: null,
```
DanielMSchmidt: I think this is really misleading. At first I thought, "oh I must have misread something, it seems to check for null, too", but this test doesn't really test this. I personally would love to have two "duplicate" tests, one for `null` and one for `undefined`

src/js/structs/Node.js
Fabs: You are right, I assumed that because it is an empty hash or a value, it can't never be null. But this would be better code.

src/js/structs/Node.js
Fabs: Yes. Orlando raised it in the original PR, my point was not adding something to the public API of a class that is not being used somewhere else. As in a like to carefully think what methods I want it to expose and not, and right now there is no requirement for that method.

Although for a struct it might not be as concerned as for a reacr component. We went without doing it on the original PR

src/js/structs/__tests__/Node-test.js
Fabs: I thought about having 4 tests, 2 for each field is null or undefined. Or maybe even just two, one for both undefined and both null.

You are right those tests do not cover all the code paths.

Fabs:
Thanks for the comments @DanielMSchmidt, definitely don't think addressing then here would be a good idea because of it being a backports, but I am glad that you challenged those points anyway, and made my brain work :D.

src/js/structs/Node.js
Fabs: Also I think I have read somewhere in my life that JS never puts null for you, you have to put null there yourself.

Fabs:
alfred run tests

Fabs:
The integration failure looks related (as in the visual component that uses the class changed on this pr), but the other build (against master) just passed so it could be a flake. I will rebuild.

image

Fabs:
run integration tests

ahoskins:
run integration tests

PR #2783, opened 2018-02-21
DCOS_OSS-2185: Upgrade jest

DanielMSchmidt:
### Impact:

The max MB used as a heap are 476MB 👍 This means smaller CI nodes for us

### ToDos

- [X] Update jest version and get jest to execute something
- [x] Jest has [an issue with `this.foo`](https://github.com/facebook/jest/issues/3970) in tests, a [jscodeshift](https://gist.github.com/DanielMSchmidt/9e1bc548ab5d8d1a3fadddc2d735f8c2) is work in progress
- [x] `RouterUtil-test` depends hard on the location, which [seems to be not settable anymore](https://github.com/facebook/jest/issues/890). I would like to change the API and use `window.location. ...` as a default value to make this easier testable
- [x] `ReducerUtil-test.js` fails to a "valid" reason, needs more investigation
- [x] `PackageDetailTab-test.js` fails due to transition from jests two argument expect to only one argument being supported (removing the second doesn't work, seems like it was the this or sth?)
- [x] `advanceTimersByTime` does not properly move the time forward in `MetronomeStore-test`
- [X] *We skip this kind of fails and address it in DCOS_OSS-2206/#2798* `Authenticated-test` fails due to `findAllInRenderedTree: instance must be a composite component`
- [x] try to change root path so I can cmd+click on failing tests to open them from the terminal
- [x] Read through changelog and find nice jest features we might enable by now
- [x] Measure memory consumption 

**Checklist**


plugins/services/src/js/pages/task-details/__tests__/TaskFileViewer-test.js
```diff
@@ -9,7 +9,7 @@ import TaskFileViewer from "../TaskFileViewer";
 
 let thisContainer, thisInstance;
 
-describe("TaskFileViewer", function() {
+describe.skip("TaskFileViewer", function() {
```
Fabs: I think we should expend 1hr here trying to have it not skippable.

Jenkinsfile
```diff
@@ -51,7 +51,23 @@ pipeline {
     stage('Unit Test') {
       steps {
         ansiColor('xterm') {
-          sh '''npm run test -- --maxWorkers=2'''
+          sh '''npm run test -- --maxWorkers=2 --coverage'''
```
Fabs: Could we move it to another PR so we make sure that this does not get too many resources, or that it really works independently?

jest/gen-config.js
```diff
@@ -56,7 +56,7 @@ var config = {
   moduleFileExtensions: ["js", "json", "es6"],
   modulePathIgnorePatterns: ["/tmp/", "/node_modules/", "/.module-cache/"],
   timers: "fake",
-  coverageReporters: ["json", "lcov", "cobertura", "text"],
+  coverageReporters: ["cobertura"],
```
Fabs: If you move the previous commit, you have to move this too.

plugins/services/src/js/pages/task-details/__tests__/TaskFileViewer-test.js
DanielMSchmidt: Did you get to do this yet?

Jenkinsfile
DanielMSchmidt: I openend this PR: https://github.com/dcos/dcos-ui/pull/2801

jest/gen-config.js
DanielMSchmidt: https://github.com/dcos/dcos-ui/pull/2801 is the PR for this now

DanielMSchmidt:
He @Fabs Could you review again? :)

plugins/services/src/js/pages/task-details/__tests__/TaskFileViewer-test.js
Fabs: Not yet, but will have some time later today. A Will do it or unblock it.

Fabs:
I ran:
- nvm install
- npm install
- npm run build-shrinkwrap (because I had changes there)

And I was left with some jest untracked files and changes to npm-shrink-wrap. 
image


https://gist.github.com/Fabs/e158d9b8031e7a3e16b5bec8f55a9eb3

Fabs:
I think I have to run with the respective private branch.
https://github.com/mesosphere/dcos-ui-plugins-private/pull/757

DanielMSchmidt:
@Fabs the untracked file is because you already had a `jest/config.json`. It is in this PR called `jest.config.json`, you need to delete the old one.

For the shrinkwrap, they can change with time, because dependencies of dependencies might change

Fabs:
Also I went to fix a commit and I got a hook to complain about many linter errors:
image

PR #2786, opened 2018-02-22
DCOS-21305: Respect DC/OS min required version when installing a package.

juliangieseke:
This disables the install button when the current dcos
version is incompatible

**Checklist**

### DIFF
--- a/tests/pages/catalog/packages/PackagesTab-cy.js
+++ b/tests/pages/catalog/packages/PackagesTab-cy.js
@@ -165,4 +165,24 @@ describe("Packages Tab", function() {
       });
     });
   });
+
+  context("respect DCOS version", function() {
+    beforeEach(function() {
+      cy
+        .route(/dcos-version/, {
+          version: "1.6",
+          "dcos-image-commit": "null",
+          "bootstrap-id": "null"
+        })
+        .visitUrl({ url: "/catalog", logIn: true });
+    });
+
+    it("cant install package because of DCOS version", function() {
+      cy.get(".panel").contains("arangodb").click();
+      cy
+        .get(".button.button-primary")
+        .contains("Review & Run")
+        .should("be.disabled");
+    });
+  });
 });
--- a/package.json
+++ b/package.json
@@ -69,6 +69,7 @@
     "redux": "3.3.1",
     "reflect-metadata": "0.1.10",
     "rxjs": "5.4.3",
+    "semver": "5.5.0",
     "tv4": "1.2.7"
   },
   "devDependencies": {
--- a/src/js/pages/catalog/PackageDetailTab.js
+++ b/src/js/pages/catalog/PackageDetailTab.js
@@ -9,18 +9,20 @@ import { StoreMixin } from "mesosphere-shared-reactjs";
 import { Dropdown, Tooltip, Modal } from "reactjs-components";
 
 import Icon from "#SRC/js/components/Icon";
-import Breadcrumb from "../../components/Breadcrumb";
-import BreadcrumbTextContent from "../../components/BreadcrumbTextContent";
-import CosmosPackagesStore from "../../stores/CosmosPackagesStore";
+import Breadcrumb from "#SRC/js/components/Breadcrumb";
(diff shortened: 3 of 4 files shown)


tests/pages/catalog/packages/PackagesTab-cy.js
```diff
@@ -165,4 +165,24 @@ describe("Packages Tab", function() {
       });
     });
   });
+
+  context("respect DCOS version", function() {
```
DanielMSchmidt: gif

src/js/utils/VersionUtil.js
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    if (!verA.prerelease && verB.prerelease) {
+      return 1;
+    }
+    if (verA.prerelease && !verB.prerelease) {
+      return -1;
+    }
+
+    // cant compare prereleases,
```
nLight: Since we can't really compare pre-releases, should we block unless they are the same? Like:
```js
if(verA.prerelease !== verB.prerelease) {
  // return whatever blocks the install button
}
```

src/js/utils/VersionUtil.js
```diff
@@ -0,0 +1,82 @@
...
+   *   mnetadata: string
+   * }
+   *
+   * This function mainly tries to return a valid semver-complaint object,
+   * it might truncate invalid input instead of throwing an error.
+   *
+   * @param {string} versionString
+   * @return {object} - an object with versioninformation
+   */
+  convertVersionStringToObject(versionString) {
```
nLight: We are that close from the OOP approach :) 
```js
class SemanticVersion {}

const verA = new SemanticVersion(versionStringA);
const verB = new SemanticVersion(versionStringB);

if(verA.notEqual(verB)) {
}
```
I miss ruby sometimes...

MatApple:
@juliangieseke curious if we explored existing libraries/utils for handling semver comparisons?

juliangieseke:
Hey Mat,

yes, there are a few tools/packages, but first I thought they didnt fit our needs:
1. semver-compare (does basicly the same I do right now, but it is more strict and our versions not always follow semver)
2. this function: https://gist.github.com/TheDistantSea/8021359
3. semver (from the npm guy & used by npm - I thought it was too much/heavy for the first attempt, but before I start writing a "proper class", I think we can use that one 🤔)

So I have implemented it with `semver` - thanks for challenging! 👍

weblancaster:
@juliangieseke is it planned to let the user go to `review & run` view and only then disable the button?

@leemunroe can you leave your feedback, please?

Other than that looks good.

screen shot 2018-02-27 at 4 24 53 pm

screen shot 2018-02-27 at 4 25 05 pm

leemunroe:
@weblancaster What are the (2) errors in your screenshot?

For the min version, the intent was to disable the `Review & Run` button on the package detail view with a tooltip `This version of Kafka requires DC/OS version 1.10 or higher, but you are running version 1.9`

weblancaster:
@leemunroe Ok thanks for the quick feedback (cc @juliangieseke)

The errors are not part of this PR and were something I did not fill in the form.

juliangieseke:
run integration tests

weblancaster:
Hey @juliangieseke I tested again today but looks like the comments @leemunroe and I left weren't addressed.

Thanks.

juliangieseke:
Hey Michael, can you clarify what exactly is wrong right now? Might be a misinterpretation on my side, but I think the PR does exactly what Lee asked for? 🤔

nLight:
I think the question was that we want to block initial button, not the one in the form
screen shot 2018-03-02 at 13 38 54

juliangieseke:
Yes, we want, but that is what happens if the version dont match 🤔 

I should have give some testing instructions, because being on current master there is no package which requires 1.13 😅

**Steps to test this:**
- Open `Metadatastore.js` and add `return "1.6"` before the existing return statement
- try to install (almost) any package, you should see the disables button with tooltip.

GIF: https://cl.ly/08103f2N1J10

juliangieseke:
@weblancaster ^ 🙇

PR #2788, opened 2018-02-22
[Backport 1.11] DCOS-21359: fix(ServicesContainer): checks and updates modalProps if needed

juliangieseke:
---

ℹ️Cherry-picked backport from #2777 - please review the other one 👍

---

This PR adds a `getCleanModalProps` method to the `ServicesContainer`
which cleans and updates the `modalProps` object which is passed down to
the modals.
Before this change it was possible to have an invalid state in `modalProps`
like this: { service: undefined, id: "delete" }. This should be fixed now so
that `service` is always set, if `id` is. If anything is wrong the method
tries to correct the object as good as possible.

Closes DCOS-21359


ahoskins:
Holding until the `master` fix goes in, because the review will happen over there.

juliangieseke:
Master PR got merged 👍 - Now this one 🙇

PR #2789, opened 2018-02-22
DCOS-21181: Update service filters

nLight:
This PR aligns Services filter with what we show in the Table. I removed Delayed and Waiting filters and added Deleting and Recovering filters. On top of that I fixed Catalog filter and adjusted positioning of the Pod filter.

Before | After
------------ | -------------
screen shot 2018-02-22 at 13 56 53 |screen shot 2018-02-22 at 13 59 55

**Checklist**

Closes DCOS-21181


nLight:
Legit integration fail, don't rerun

nLight:
@leemunroe please have a look.

leemunroe:
@nLight Nice. You even fixed the odd spacing issue with Pod 🌟

I think we're good to keep health for this PR but something we should keep an eye on or discuss. Wondering if we should remove it since we don't have a service level health representation. Or if we can assume healthy unless unhealthy.

I wasn't able to deploy a service with a volume and test it. Can you confirm the volume filter works? All others looked good.

nLight:
Hey @leemunroe, thanks for having a look! I tested the Volumes filter and it's working. There's also a test that ensures all filters are working.

PR #2790, opened 2018-02-22
DCOS-21266: Files UI breaks after navigating through the switch button

DanielMSchmidt:
I originally wanted to write integration tests, but this part was pretty hard to fixture away, so I wrote unit tests instead.

### Commits:

1. Refactor to have the function testable
2. Write tests reflecting current state of function
3. Fix the functionality

**Checklist**

### DIFF
--- a/plugins/services/src/js/pages/task-details/TaskFileViewer.js
+++ b/plugins/services/src/js/pages/task-details/TaskFileViewer.js
@@ -13,21 +13,6 @@ import SearchLog from "../../components/SearchLog";
 import TaskDirectory from "../../structs/TaskDirectory";
 import TaskDirectoryActions from "../../events/TaskDirectoryActions";
 
-export function getNewRoute(routePath, params, path) {
-  const hasFilePathParam = routePath.indexOf(":filePath") !== -1;
-  if (!hasFilePathParam && routePath.endsWith("/")) {
-    routePath += ":filePath";
-  }
-  if (!hasFilePathParam && !routePath.endsWith("/")) {
-    routePath += "/:filePath";
-  }
-
-  return formatPattern(
-    routePath,
-    Object.assign({}, params, { filePath: encodeURIComponent(path) })
-  );
-}
-
 export default class TaskFileViewer extends React.Component {
   constructor() {
     super(...arguments);
@@ -71,7 +56,12 @@ export default class TaskFileViewer extends React.Component {
     const { params, routes } = this.props;
     const routePath = RouterUtil.reconstructPathFromRoutes(routes);
 
-    this.context.router.push(getNewRoute(routePath, params, path));
+    this.context.router.push(
+      formatPattern(
+        RouterUtil.getCorrectedFileRoutePath(routePath),
+        Object.assign({}, params, { filePath: encodeURIComponent(path) })
+      )
+    );
   }
 
   getLogFiles() {
--- a/plugins/services/src/js/pages/task-details/TaskFileViewer.js
+++ b/plugins/services/src/js/pages/task-details/TaskFileViewer.js
@@ -47,27 +47,16 @@ class TaskFileViewer extends React.Component {
   }
 
   handleViewChange(currentFile) {
-    const { params, routes } = this.props;
     const path = currentFile.get("path");
     if (path === this.getSelectedFile().get("path")) {
       // File path didn't change, let's not try to update path
       return;
     }
 
-    let routePath = RouterUtil.reconstructPathFromRoutes(routes);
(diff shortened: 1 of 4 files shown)


DanielMSchmidt:
run integration tests

plugins/services/src/js/pages/task-details/TaskFileViewer.js
```diff
@@ -240,4 +229,23 @@ TaskFileViewer.propTypes = {
   task: PropTypes.object
 };
 
+TaskFileViewer.prototype.getNewRoute = function getNewRoute(
```
nLight: I'd suggest either extract the function to a util or writing a snapshot test. Let's please not sacrifice conventions to accommodate testing.

DanielMSchmidt: It was pretty hard to get an instance set up correctly, that's why I wanted to export a function.
Exporting also didn't work because of the `module.exports` thing so I went the easier route. I will investigate how to do this

nLight: What's the `module.exports` thing? We use es6 modules in some newer places in the codebase if that'd help you.

nLight: We could also pair to see how we could mock the setup

DanielMSchmidt: The test broke and I think I correlated some UI break which was a coincidence and associated that there is a special reason not to use imports here. Is there a reason we don't use imports all over the place?

nLight: We have some cyclic dependencies I think that was the reason we don't use es6 imports everywhere

DanielMSchmidt: I think mocking makes it worse in that area

plugins/services/src/js/pages/task-details/TaskFileViewer.js
```diff
@@ -13,7 +13,22 @@ import SearchLog from "../../components/SearchLog";
 import TaskDirectory from "../../structs/TaskDirectory";
 import TaskDirectoryActions from "../../events/TaskDirectoryActions";
 
-class TaskFileViewer extends React.Component {
+export function getNewRoute(routePath, params, path) {
+  const hasFilePathParam = routePath.indexOf(":filePath") !== -1;
+  if (!hasFilePathParam && routePath.endsWith("/")) {
```
juliangieseke: If you store `routePath.indexOf(":filePath") `, you also can store `routePath.endsWith("/")` 😀 (you know what I'd go for 😇)

plugins/services/src/js/pages/task-details/TaskFileViewer.js
```diff
@@ -13,7 +13,22 @@ import SearchLog from "../../components/SearchLog";
 import TaskDirectory from "../../structs/TaskDirectory";
 import TaskDirectoryActions from "../../events/TaskDirectoryActions";
 
-class TaskFileViewer extends React.Component {
+export function getNewRoute(routePath, params, path) {
```
juliangieseke: On this whole function: could we move this into something like `UrlUtil.normalizeFilePath` or `RouterUtil.buildFilePath`? `getNewRoute` sounds too generic, also I'd love to see some JSDoc comments for new functions we write.

plugins/services/src/js/pages/task-details/TaskFileViewer.js
DanielMSchmidt: I respectfully disagree. endsWith explains in detail what it does, whereas indexOf !== -1 is not very explaining

plugins/services/src/js/pages/task-details/TaskFileViewer.js
DanielMSchmidt: Is that a common thing for us to do JSDocs?

plugins/services/src/js/pages/task-details/TaskFileViewer.js
juliangieseke: Well, you can use `routePath.includes(":filePath")` :trollface:

plugins/services/src/js/pages/task-details/TaskFileViewer.js
juliangieseke: Some methods have, some have not… i dont see a bad thing having them but they can help a lot.

DanielMSchmidt:
@nLight Could you review this refactoring by @juliangieseke 
@juliangieseke Could you approve? I reviewed your refactoring and it looks fine to me

PR #2791, opened 2018-02-22
fix(ResourcesUtil): fix undefined or null object

weblancaster:
Fix issue when trying to convert undefined or null to object.

Closes DCOS-21367

**Checklist**

### DIFF
--- a/src/js/structs/Node.js
+++ b/src/js/structs/Node.js
@@ -83,6 +83,17 @@ class Node extends Item {
 
     return sum;
   }
+
+  getUsedResources() {
+    return (
+      this.get("used_resources") || {
+        cpus: 0,
+        mem: 0,
+        gpus: 0,
+        disk: 0
+      }
+    );
+  }
 }
 
 module.exports = Node;
--- a/src/js/utils/ResourcesUtil.js
+++ b/src/js/utils/ResourcesUtil.js
@@ -55,7 +55,8 @@ const ResourcesUtil = {
       }
     }
 
-    let resources = Object.keys(item.get("used_resources"));
+    const usedResources = item.getUsedResources();
+    let resources = Object.keys(usedResources);
 
     if (excludeList.length > 0) {
       resources = resources.filter(function(resource) {
--- a/plugins/services/src/js/structs/Framework.js
+++ b/plugins/services/src/js/structs/Framework.js
@@ -130,4 +130,15 @@ module.exports = class Framework extends Application {
         };
       }, frameworkResources);
   }
+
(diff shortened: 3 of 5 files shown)


src/js/utils/ResourcesUtil.js
```diff
@@ -55,7 +55,13 @@ const ResourcesUtil = {
       }
     }
 
-    let resources = Object.keys(item.get("used_resources"));
+    let usedResources = item.get("used_resources");
+
+    if (!usedResources) {
+      usedResources = {};
```
nLight: Hey, Michael!

In `structs/Node.js` we already guard against this case in one method like this 
```js
const value = (this.get("used_resources") || {})[resource];
``` 

Wouldn't it be better to create a getter method for `used_resources` that will always return an object use this getter method everywhere?

weblancaster: I agree @nLight, I will address this.

nLight:
run system tests

src/js/structs/Node.js
```diff
@@ -83,6 +83,10 @@ class Node extends Item {
 
     return sum;
   }
+
+  getResources() {
```
nLight: Let's please rename it to `getUsedResources` since we do have `this.get("resources")`

weblancaster: @nLight the reason I left as `getResource` and not `getUsedResources` is that in `structs/Framework.js` we already use `getResources` in which is used here https://github.com/dcos/dcos-ui/pull/2791/files#diff-def918eb22daa7712f7a5e81753e82ebR51

So to keep it consistent and easier to know what to expect I kept `getResources`, Now I wonder. Should I change this method to be the same as in `Framework.js` or introduce `getUsedResources` for framework and node then change line `https://github.com/dcos/dcos-ui/pull/2791/files#diff-def918eb22daa7712f7a5e81753e82ebR51` to get `getUsedResources`

Thanks for the feedback

MatApple: @weblancaster while I understand the path you chose to prevent further code changes, this is a good opportunity to address the name issue here and name it correctly.

weblancaster: Comments addressed

orlandohohmeier:
LGTM!

orlandohohmeier:
Next Steps: Let's change all case to use getUsedResource instead of accessing the item directly. We probably also want to introduce a `getResource` getter to align the usage.

nLight:
@orlandohohmeier is there a JIRA for that refactoring?

PR #2793, opened 2018-02-25
chore(reactjs-components): update package to 0.20.3

Fabs:
The new version of VirtualList on 0.20.3 does not cause
react to render infinitely when, while rendering the list
there is an exception.

This fixed behavior helps prevent a few bugs we had in the ui.

**Testing**

The new version of reacts components have changed on the virtual list. You can check some of the lists around the app.

More specifically, you can go on the struct/Node and make it throw an exception by adding an error in any of the methods that are used on the Nodes Table. There you should see there is an error in console, but the list does not render wrong content.

**Checklist**
- [x] Test it myself

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -2897,7 +2897,7 @@
     },
     "decamelize": {
       "version": "1.2.0",
-      "from": "decamelize@>=1.0.0 <2.0.0",
+      "from": "decamelize@>=1.1.2 <2.0.0",
       "resolved": "https://registry.npmjs.org/decamelize/-/decamelize-1.2.0.tgz"
     },
     "decompress": {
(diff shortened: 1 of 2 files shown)


nLight:
`renders proper review screen and JSON for an app with persistent volume.Services Applications renders proper review screen and JSON for an app with persistent volume` looks like a legit failure 🤔

DanielMSchmidt:
run integration tests

DanielMSchmidt:
@nLight Our Jenkinsfile is green, so can't be legit failures 🤔

Fabs:
It passed everything now.

I look at those failures, it goes like dcos-01 on a td, I have seen them in different places. Do we still catalog flake tests?

juliangieseke:
@Fabs yes: https://jira.mesosphere.com/browse/DCOS-20241

PR #2796, opened 2018-02-28
Improve Jenkins pipeline

bstavroulakis:
🎉 🎈 

**Description**:
This PR contains the long-awaited Jenkins release and bump process we've been working on for the past few weeks. To do this we've **added two stages** to the Jenkinsfile. Both of these stages will only run when the branch is master (this will prevent these stages from running on PRs).

1. "Upload Build" stage: uploading the built dcos-ui artifact to an S3 bucket.
2. "Update Github" stage: there are two paths here depending on the presence of a tag on the current commit. If there **is a tag**, then a new PR is opened on `dcos/dcos`, and the dcos-ui changelog between this tag and the previous tag is generated and formatted into the PR body. If there is **no tag**, then we will commit and push onto our [mesosphere:dcos-ui/latest](https://github.com/dcos/dcos/pull/2506) with the latest dcos-ui artifact.

More on the no tag case: The benefit of always updating the dcos-ui/latest PR is we will always have a cluster with our latest dcos-ui master. This will be really nice for the design team and other engineering teams to quickly see our changes.

More on the tag case: To be clear, this happens when there is a tag on the current commit (it checks this with `git tag -l --contains HEAD`). The new release process process will be much less cumbersome for our team. Now the whole release process will be triggered by pushing a release tag to dcos-ui. Jenkins will listen for new tag pushes (that match the release tag format ex: v1.11.0-rc.16). It will then start the Jenkins job, build, upload to S3, then generate the changelog, and open a PR for you against dcos/dcos.

**Testing**:
If you comprehended the above, it should make sense that there are two cases to test - with a tag, and without a tag.

Testing without a tag is easy. Go to Jenkins dashboard. Click on Frontend CI Jobs -> DCOS UI OSS Release Builder -> bill/feat/dcos-ui-cd branch -> click Build Now in the top left. This will trigger a job. We expect this job to update the dcos-ui/latest PR.

Testing with a tag is a bit more difficult. First, checkout this branch. Then push the tag `test-dcos-ui-cd2`. Then navigate toe Jenkins dashboard. Click on Frontend CI Jobs -> DCOS UI OSS Release Builder -> test-dcos-ui-cd2 -> click Build Now in the top left. This will trigger a job (the same job as before). But the different now is it will take the different update-github path. Now we expect this job to open a new PR against dcos/dcos. *Note*: if existing branch on mesosphere/dcos it will fail. Make sure to delete before testing this.

**Checklist**


Jenkinsfile.release
```diff
@@ -14,7 +14,7 @@ pipeline {
   }
 
   parameters {
-    booleanParam(defaultValue: false, description: 'Should the build be uploaded to S3?', name: 'SHOULD_UPLOAD_BUILD')
+    booleanParam(defaultValue: false, description: 'Should run the Authorization?', name: 'SHOULD_AUTHORIZE')
```
bstavroulakis: Now, every time a PR is merged into master we want to automatically create a build tar and bump the nightly DC/OS PR. So from "Should upload", we can change this to "Should Authorize". 

The Authorize requires someone to manually go into Jenkins and click on Proceed which will make the whole CI/CD process much less automatic and streamlined I think. But we can keep this if we want to enable it in the future or from Jenkins.

Jenkinsfile.release
```diff
@@ -84,12 +73,23 @@ pipeline {
...
           archiveArtifacts 'buildinfo.json'
         }
       }
     }
+
+    stage('Update DC/OS OSS Open PR'){
+      steps {
+        withCredentials([
+          usernamePassword(credentialsId: 'a7ac7f84-64ea-4483-8e66-bb204484e58f', passwordVariable: 'GIT_PASSWORD', usernameVariable: 'GIT_USER'),
+          string(credentialsId: '3f0dbb48-de33-431f-b91c-2366d2f0e1cf',variable: 'AWS_ACCESS_KEY_ID'),
```
DanielMSchmidt: We don't need this, right?

scripts/ci/update-open-pr
```diff
@@ -0,0 +1,15 @@
...
+git clone --depth 1 -b dcos-ui/latest https://$GIT_USER:$GIT_PASSWORD@github.com/mesosphere/dcos dcos-latest
+ 
+ls -l
+cp -fr buildinfo.json dcos-latest/packages/dcos-ui/buildinfo.json
+cd dcos-latest
+
+git add "packages/dcos-ui/buildinfo.json"
+git commit -m "Update DC/OS UI to latest build"
+
+git push origin dcos-ui/latest
```
DanielMSchmidt: I think it is a convention that shell files should end with a newline

Jenkinsfile.release
DanielMSchmidt: The reason why it asks for authorization currently is because you don't run it on the master branch. Once this is merged it will no longer ask for authorization as this jenkins job normally only runs on master.

DanielMSchmidt: The switch was thought as "should this be a real release that is used somewhere in production"

scripts/ci/update-open-pr
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+# Setup Git
+git config --global user.email "$GIT_USER@users.noreply.github.com"
+git config --global user.name "MesosphereCI Robot"
+git clone --depth 1 -b dcos-ui/latest https://$GIT_USER:$GIT_PASSWORD@github.com/mesosphere/dcos dcos-latest
```
DanielMSchmidt: Shellcheck [SC2086](https://github.com/koalaman/shellcheck/wiki/SC2086) tells me that it would be nice to have double quotes around the URL here

scripts/ci/update-open-pr
```diff
@@ -0,0 +1,15 @@
...
+ 
+ls -l
+cp -fr buildinfo.json dcos-latest/packages/dcos-ui/buildinfo.json
+cd dcos-latest
```
DanielMSchmidt: Shellcheck [SC2164](https://github.com/koalaman/shellcheck/wiki/SC2164) tells me that we should do something like 

```sh
cd dcos-latest || exit 1
```

to have the script fail if we can't cd into the directory

scripts/ci/update-open-pr
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
```
DanielMSchmidt: I personally love to have a `set -x` here. This let's the script fail once a process ends with a non-zero exit code. That way we have no silent failures.

scripts/ci/update-open-pr
```diff
@@ -0,0 +1,15 @@
...
+# Setup Git
+git config --global user.email "$GIT_USER@users.noreply.github.com"
+git config --global user.name "MesosphereCI Robot"
+git clone --depth 1 -b dcos-ui/latest https://$GIT_USER:$GIT_PASSWORD@github.com/mesosphere/dcos dcos-latest
+ 
+ls -l
```
DanielMSchmidt: I think this should not be in here

Jenkinsfile
```diff
@@ -22,12 +22,6 @@ pipeline {
   }
 
   stages {
-    stage('Authorization') {
```
DanielMSchmidt: We can not remove this, we need to make sure that no one can steal credentials by checking in a different Jenkinsfile

scripts/ci/update-github
```diff
@@ -0,0 +1,62 @@
...
+# Setup Git
+git config --global user.email "$GIT_USER@users.noreply.github.com"
+git config --global user.name "MesosphereCI Robot"
+git config credential.helper 'cache --timeout=300'
+git clone "https://$GIT_USER:$GIT_PASSWORD@github.com/mesosphere/dcos" dcos-nightly
+ls -l
+
+cd dcos-nightly || exit 1
+
+if [[ $TAG_NAME ]]
```
DanielMSchmidt: Have you tried if the Tag name works with our PR approach of merging PRs? I would have thought that if you have a commit which is tagged in a PR and merge it you get a merge commit as the new HEAD, but I am not fully aware if this is really true.

scripts/ci/update-github
```diff
@@ -0,0 +1,62 @@
...
+then
+  git checkout -- .
+  git remote add upstream "https://$GIT_USER:$GIT_PASSWORD@github.com/dcos/dcos"
+  git checkout master
+  git checkout -b dcos-ui/$TAG_NAME
+
+  cp -fr ../buildinfo.json packages/dcos-ui/buildinfo.json
+
+  git add "packages/dcos-ui/buildinfo.json"
+  git commit -m "chore(dcos-ui): update package"
```
DanielMSchmidt: (Optional) Can we add a commit body which states where the change is coming from and that it is automated?

scripts/ci/upload-build
```diff
@@ -2,35 +2,34 @@
...
-cd "$DIST_PATH"
+cd "dist"
 tar czf "../$RELEASE_NAME" .
 cd ..
+
 BUILD_SHA=$(shasum ${RELEASE_NAME} | cut -d " " -f1)
 ENCODED_RELEASE_NAME=$(echo ${RELEASE_NAME} | sed "s/\+/\%2B/g")
-DOWNLOAD_URL=$(echo "https://${AWS_BUCKET}/${BUCKET_PATH}/${ENCODED_RELEASE_NAME}")
-FILENAME='buildinfo.json'
+DOWNLOAD_URL=$(echo "https://${AWS_BUCKET}/dcos-ui/${ENCODED_RELEASE_NAME}")
```
DanielMSchmidt: (Optional) Why did you inline the `$BUCKET_PATH` 🤔

tags-delete
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
```
DanielMSchmidt: Why is this script needed?

scripts/ci/upload-build
```diff
@@ -2,35 +2,34 @@
...
 
-# build release filename
-FILENAME='dcos-ui'
-PKG_VERSION=$(python -c "import sys, json; \
-  pkg = json.load(open('$PACKAGE_PATH/package.json')); \
-  sys.stdout.write(str(pkg['version']));")
 PKG_BRANCH=$(echo ${BRANCH_NAME} | sed -E 's/([^/]*)$|./\1/g')
-RELEASE_NAME="${PKG_BRANCH}+${FILENAME}${BUILD_VARIANT}-v${PKG_VERSION}+${SHA_SLUG}.tar.gz"
+
+# ex. master+dcos-ui-v1_12_0-rc_1.tar.gz
```
nLight: ❤️

Jenkinsfile
```diff
@@ -132,6 +124,41 @@ pipeline {
...
+      post {
+        always {
+          archiveArtifacts 'buildinfo.json'
+        }
+      }
+    }
+
+    stage('Update Github'){
+      when {
+        expression { env.BRANCH_NAME == 'master' || env.BRANCH_NAME == 'bill/feat/dcos-ui-cd' }
```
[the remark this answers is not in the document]
ahoskins: Yes, I will remove this. Adding this makes it testable on this branch. We can re-add this line locally to test.

scripts/ci/generate-changelog
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+
```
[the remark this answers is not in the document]
ahoskins: Good idea - added set -ex at the top 👍

scripts/ci/update-github
```diff
@@ -0,0 +1,62 @@
...
+rm -rf dcos-nightly
+
+# Setup Git
+git config --global user.email "$GIT_USER@users.noreply.github.com"
+git config --global user.name "MesosphereCI Robot"
+git config credential.helper 'cache --timeout=300'
+git clone "https://$GIT_USER:$GIT_PASSWORD@github.com/mesosphere/dcos" dcos-nightly
+ls -l
+
+cd dcos-nightly || exit 1
```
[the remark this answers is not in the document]
ahoskins: Yes, good point

scripts/ci/update-github
```diff
@@ -0,0 +1,62 @@
+#!/bin/bash
+set -ex
```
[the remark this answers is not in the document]
ahoskins: It won't clutter the changelog because the changelog is not taking everything printed to console. It's just taking everything assigned to `MARKDOWN` variable.

scripts/ci/update-github
ahoskins: Yes, the merge commit would be the HEAD of master immediately following a merge. But this isn't an issue.

If you "build now" on Jenkins from a tracked tag, it will be the HEAD (from that tags perspective). Does that make sense? Think of tags like branches on Jenkins. When triggering a build from a tag it checks out the commit of that tag.

scripts/ci/update-github
ahoskins: @bstavroulakis thoughts on this?

tags-delete
ahoskins: Just a helper for deleting tags quickly locally. @bstavroulakis added it

scripts/ci/upload-build
ahoskins: Fixed.

scripts/ci/generate-changelog
```diff
@@ -0,0 +1,27 @@
...
+JIRA_PREFIX="DCOS|MARATHON"
+MARKDOWN=""
+
+for COMMIT in $COMMITS; do
+
+  COMMIT_BODY=$(git log -1 ${COMMIT})
+  COMMIT_BODY_ONELINE=$(echo $COMMIT_BODY)
+  COMMIT_COMMENT=$(git log --format=%B -n 1 ${COMMIT})
+
+  if [[ $COMMIT_BODY_ONELINE =~ Closes\ DCOS-[0-9]* ]]; then
```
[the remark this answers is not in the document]
ahoskins: DCOS|MARATHON|DCOS_OSS, right?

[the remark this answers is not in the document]
ahoskins: I will add them all in the JIRA_PREFIX variable, then use it 👍

scripts/ci/generate-changelog
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+
+GIT_TAGS=$(git tag -l --sort=-version:refname)
+
+TAGS=($GIT_TAGS)
+LATEST_TAG=${TAGS[0]}
+PREVIOUS_TAG=${TAGS[1]}
+COMMITS=$(git log $PREVIOUS_TAG..$LATEST_TAG --pretty=format:"%H")
+JIRA_PREFIX="DCOS|MARATHON"
```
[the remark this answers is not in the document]
ahoskins: Good catch, see below comment

tags-delete
DanielMSchmidt: okay 👍

Jenkinsfile
```diff
@@ -132,6 +124,41 @@ pipeline {
          junit 'results/results.xml'
        }
      }
-   }
+    }
+
+    stage('Upload Build') {
+      when {
+        expression { env.BRANCH_NAME == 'master' || env.BRANCH_NAME == 'bill/feat/dcos-ui-cd' }
```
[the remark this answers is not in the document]
ahoskins: Thanks, removed.

scripts/ci/update-github
DanielMSchmidt: You are right, tried it out 👍 at least locally

scripts/ci/update-github
ahoskins: Actually, I don't think need to do this. The commit author is `mesosphere-ci`, so we already know it's not a person committing, it's something automated.

screen shot 2018-03-13 at 2 03 41 pm

scripts/ci/update-github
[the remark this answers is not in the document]
ahoskins: For debugging in jenkins it's nice to have the -x flag!

Jenkinsfile
ahoskins: We discussed a bit on slack and need to investigate this more...

ahoskins:
We have a known issue when creating a new PR for the release. We need to serialize the newlines for the JSON (into `\\n` I think).

screen shot 2018-03-13 at 5 47 07 pm

Jenkinsfile
```diff
@@ -68,70 +68,105 @@ pipeline {
...
+    //      junit 'results/results.xml'
+    //    }
+    //  }
+    // }
+
+    // when env.BRANCH_NAME == 'master', we will update the mesosphere:dcos-ui/latest branch
+    // when env.BRANCH_NAME =~ /(tag-name-regex)/, we will open a release PR against dcos/dcos
+    stage('Upload Build') {
+      when {
+        expression { env.BRANCH_NAME == 'master' || env.BRANCH_NAME =~ /^v?(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(-(0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(\.(0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*)?(\+[0-9a-zA-Z-]+(\.[0-9a-zA-Z-]+)*)?/ || env.BRANCH_NAME == 'bill/feat/dcos-ui-cd' || env.BRANCH_NAME == 'test-dcos-ui-cd2' }
```
ahoskins: Note: the last two clauses in these expression will be removed before merging this PR. Just for testing for now.

Jenkinsfile
```diff
@@ -68,70 +68,105 @@ pipeline {
...
-       withCredentials([
-          [
-            $class: 'AmazonWebServicesCredentialsBinding',
-            credentialsId: 'f40eebe0-f9aa-4336-b460-b2c4d7876fde',
-            accessKeyVariable: 'AWS_ACCESS_KEY_ID',
-            secretKeyVariable: 'AWS_SECRET_ACCESS_KEY'
-          ]
+    stage('Update Github'){
+      when {
+        expression { env.BRANCH_NAME == 'master' || env.BRANCH_NAME =~ /^v?(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(-(0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(\.(0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*)?(\+[0-9a-zA-Z-]+(\.[0-9a-zA-Z-]+)*)?/ || env.BRANCH_NAME == 'bill/feat/dcos-ui-cd' || env.BRANCH_NAME == 'test-dcos-ui-cd2' }
```
ahoskins: Same with here.

Jenkinsfile
ahoskins: Also note: the reason there is a regex for branches, is because when triggered from a tag, Jenkins sets the BRANCH_NAME to the tag name.

Jenkinsfile
```diff
@@ -40,21 +40,21 @@ pipeline {
       }
     }
 
-    stage('Lint') {
```
DanielMSchmidt: I understand that you don't need linting and the like for the PR bumps, but I think they are very valuable for our PRs and I would like them to stay

Jenkinsfile
DanielMSchmidt: I think removing them will stale the reviews, right? 🤔

DanielMSchmidt: I don't understand the tag name regex, why is the tag name the branch name? I thought our release branch names are `release/1.10`, `release/1.11` and so on.

Jenkinsfile
DanielMSchmidt: Can we somehow store this Regex at one place? Maybe as a global environment variable or something similar?

Jenkinsfile
```diff
@@ -68,70 +68,105 @@ pipeline {
...
+        ]) {
+          sh "BRANCH_NAME=${env.BRANCH_NAME} ./scripts/ci/upload-build"
         }
       }
 
       post {
         always {
-          archiveArtifacts 'cypress/**/*'
-          junit 'cypress/results.xml'
+          archiveArtifacts 'buildinfo.json'
```
DanielMSchmidt: Do we still need the artifact if everything is done automatically? We don't need to remove it I just wondered

Jenkinsfile
DanielMSchmidt: Might be a bit esoteric, but couldn't I fork this project and run a PR from my master which then executes the upload build script? Maybe we can add a small follow-up Jira to mitigate the risk here or am I on the wrong track?

scripts/ci/update-github
DanielMSchmidt: I just want to prevent a situation where someone asks "Who the hell is comitting this?" and it is not clear where this originates from. But I am not super strong with this opinion, so yeah nevermind

scripts/ci/update-github
```diff
@@ -0,0 +1,62 @@
...
+  git checkout -- .
+  git remote add upstream "https://$GIT_USER:$GIT_PASSWORD@github.com/dcos/dcos"
+  git checkout master
+  git checkout -b dcos-ui/$TAG_NAME
+
+  cp -fr ../buildinfo.json packages/dcos-ui/buildinfo.json
+
+  git add "packages/dcos-ui/buildinfo.json"
+  git commit -m "chore(dcos-ui): update package"
+  git push --force origin dcos-ui/$TAG_NAME
```
juliangieseke: this being automated, do we really want to force push here? Shouldn't it fail if there is already is a branch with the same name?

scripts/ci/update-github
```diff
@@ -0,0 +1,62 @@
...
+cd dcos-nightly
+
+if [[ $TAG_NAME ]]
+then
+  git checkout -- .
+  git remote add upstream "https://$GIT_USER:$GIT_PASSWORD@github.com/dcos/dcos"
+  git checkout master
+  git checkout -b dcos-ui/$TAG_NAME
+
+  cp -fr ../buildinfo.json packages/dcos-ui/buildinfo.json
```
juliangieseke: what is `-r`? 🤔 I'm only aware of `-R` for directories.

scripts/ci/upload-build
juliangieseke: this isnt a valid semver version number :(

juliangieseke: this isnt a valid semver version number :(

scripts/ci/upload-build
```diff
@@ -2,35 +2,34 @@
...
 BRANCH_NAME=${BRANCH_NAME:-'master'}
-DIST_PATH=${DIST_PATH:-'dist'}
-PACKAGE_PATH=${PACKAGE_PATH:-'.'}
-SHA_SLUG=${SHA_SLUG:-$(git rev-parse --short HEAD)}
+PKG_VERSION=$(git tag -l --contains HEAD)
 
-# aws settings
-AWS_BUCKET='downloads.mesosphere.io'
-BUCKET_PATH='dcos-ui'
+if [[ $PKG_VERSION ]]
```
juliangieseke: what was the reason for changing this script, was there anything which wasnt possible with the old one?

juliangieseke: on this very variable: we want to have the correct version number represented in out package.json, when pushing a tag, we should also change it there. that said, this change shouldnt be needed.

juliangieseke: on this very variable: we want to have the correct version number represented in out package.json, when pushing a tag, we should also change it there. that said, this change shouldnt be needed. 🤔

scripts/ci/upload-build
ahoskins: This will be the tag name. So It will be valid! 💯

ahoskins: I'll update the comment to be accurate.

scripts/ci/update-github
```diff
@@ -0,0 +1,62 @@
+#!/bin/bash
+set -ex
+
+TAG_NAME=$(git tag -l --contains HEAD)
+
+rm -rf dcos-nightly
```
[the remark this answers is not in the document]
ahoskins: There is no harm at all using `-f` here. We want to make sure this directory does not exist before cloning below.

scripts/ci/update-github
```diff
@@ -0,0 +1,62 @@
...
+
+# Setup Git
+git config --global user.email "$GIT_USER@users.noreply.github.com"
+git config --global user.name "MesosphereCI Robot"
+git config credential.helper 'cache --timeout=300'
+git clone "https://$GIT_USER:$GIT_PASSWORD@github.com/mesosphere/dcos" dcos-nightly
+ls -l
```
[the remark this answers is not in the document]
ahoskins: I like this kind of thing for debugging! No harm imo.

scripts/ci/update-github
ahoskins: It recursively copies the directory. Same meaning as -R.

scripts/ci/update-github
ahoskins: Hm, I just think in general it's better to be forceful in this script. Imagine I wanted to update the position of a tag, I'd want this job to do that forcefully, and not fail because it already exists. 

But I don't feel strong about this, and I could imagine arguments for either.

scripts/ci/upload-build
ahoskins: This variable will be the tag name. This `if` block just prepends it with the dash. 

We seem to *never* update the package.json version 🙃. That's why we totally removed this from this script. The `tag` will always be here, because *tags* drive our release process.

ahoskins:
Thanks for all the reviews @juliangieseke @DanielMSchmidt @nLight. I've addressed all changes, and fixed the issue I mentioned with the PR description formatting.

I'm confident this is ready to go. Give it another pass-through, but I strongly encourage ✅  and 🚢 ... 😄

ahoskins:
Me and Bill were just looking over this, and realized we should just rebase and then amend the commit on the dcos-ui/latest branch. This keeps it fully in sync with latest changes on dcos, and keeps the number of commits low. There is no need for this history because we override the builds on S3 each time anyways.

ahoskins:
Link to successful dcos-ui/latest build for reference: https://jenkins.mesosphere.com/service/jenkins/job/frontend/job/dcos-ui-oss-release/job/bill%252Ffeat%252Fdcos-ui-cd/132/console

scripts/ci/generate-changelog
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+set -ex
+
+GIT_TAGS=$(git tag -l --sort=-version:refname)
```
DanielMSchmidt: I tried it locally by switching to an old tag and running this command from there and I got all tags as a result. If I understand this script correctly we will always generate the changelog between the latest two.

E.g. I do a tag 1.10.5 => I want a changelog for the commits between 1.10.4 and 1.10.5. This script will deliver me the changelog for 1.11.1 and 1.11.2, which is unexpected and a show stopper to me

scripts/ci/update-github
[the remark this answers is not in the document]
DanielMSchmidt: (Optional / Opinion) To me `-f` should never be the default.

scripts/ci/update-github
```diff
@@ -0,0 +1,61 @@
...
+cd dcos-nightly
+
+if [[ $TAG_NAME ]]
+then
+  git checkout -- .
+  git remote add upstream "https://$GIT_USER:$GIT_PASSWORD@github.com/dcos/dcos"
+  git checkout master
+  git checkout -b dcos-ui/$TAG_NAME
+
+  cp -fR ../buildinfo.json packages/dcos-ui/buildinfo.json
```
DanielMSchmidt: I think we don't need the `-R` here as it is just one file to copy (and it was a bit confusing to me, I went ahead to find out why `buildinfo.json` should be a directory

scripts/ci/update-github
```diff
@@ -0,0 +1,61 @@
...
+
+  git add "packages/dcos-ui/buildinfo.json"
+  git commit -m "chore(dcos-ui): update package"
+  git push --force origin dcos-ui/$TAG_NAME
```
DanielMSchmidt: I think force pushing on a tag name is harmful if someone else updated the branch in the meantime, but there is an easy fix for that: [force-with-lease](https://developer.atlassian.com/blog/2015/04/force-with-lease/).

We do a new commit, so I don't think we should even force it.

scripts/ci/update-github
```diff
@@ -0,0 +1,61 @@
...
+
+  curl -i -u "$GIT_USER:$GIT_PASSWORD" https://api.github.com
+  curl -H "Content-Type: application/json" -u "$GIT_USER:$GIT_PASSWORD" -X POST -d "${pullRequest}" https://api.github.com/repos/dcos/dcos/pulls
+else
+  git remote add upstream "https://$GIT_USER:$GIT_PASSWORD@github.com/dcos/dcos"
+  git checkout -- .
+  git checkout dcos-ui/latest
+  git pull --rebase -X theirs upstream master
+
+  cp -fr ../buildinfo.json packages/dcos-ui/buildinfo.json
```
DanielMSchmidt: again, I don't think `-r` should be here

scripts/ci/upload-build
```diff
@@ -2,35 +2,33 @@
...
-BRANCH_NAME=${BRANCH_NAME:-'master'}
-DIST_PATH=${DIST_PATH:-'dist'}
-PACKAGE_PATH=${PACKAGE_PATH:-'.'}
-SHA_SLUG=${SHA_SLUG:-$(git rev-parse --short HEAD)}
-
-# aws settings
 AWS_BUCKET='downloads.mesosphere.io'
 BUCKET_PATH='dcos-ui'
+BRANCH_NAME=${BRANCH_NAME:-'master'}
+PKG_VERSION=$(git tag -l --contains HEAD)
```
DanielMSchmidt: Neat 👍

scripts/ci/generate-changelog
ahoskins: Hey thank for point this out! We changed this quite a bit, and I tested locally and on Jenkins. Seems to be working well. Let me know if any questions or issues with it 👍

scripts/ci/update-github
ahoskins: I think being forceful is good because it gives more certainty to the script. We always want this branch to reflect what we want from this Jenkins job. And no real damage to be done - can always use reflog manually if something awful happens and we override something we shouldn't have.

scripts/ci/update-github
ahoskins: On the man page it says "In -R mode, cp will continue copying even if errors are detected." I say we keep this because it has been here for all testing, no issues, and I don't want to mess with little things at this point :)

Jenkinsfile
```diff
@@ -132,6 +130,43 @@ pipeline {
...
        }
      }
-   }
+    }
+
+    // when env.BRANCH_NAME == 'master', we will update the mesosphere:dcos-ui/latest branch
+    // when env.BRANCH_NAME =~ /(tag-name-regex)/, we will open a release PR against dcos/dcos
+    stage('Upload Build') {
+      when {
+        expression { env.BRANCH_NAME == 'master' || env.BRANCH_NAME =~ /^v?(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(-(0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(\.(0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*)?(\+[0-9a-zA-Z-]+(\.[0-9a-zA-Z-]+)*)?/ }
```
nLight: So in one case it's a branch name in other case it's a tag name, correct?

scripts/ci/update-github
DanielMSchmidt: Maybe I am too cautious here 🤷‍♂️

DanielMSchmidt: Force with lease also force pushes, it just reduces the risk to screw something up by a lot and it is to my knowledge a best practice (I got it in my force push alias, too 👍)

scripts/ci/update-github
DanielMSchmidt: Just because it was there for testing is a really weak argument for keeping this for me.
My arguments against it are:

- on ubuntu `-R` does not have the same meaning. If you run `docker run ubuntu cp --help` you get the information for the docker image we are using as base image. It states `-R, -r, --recursive copy directories recursively`
- `-R` is used to copy directories as mentioned above so having this flag here makes me think about why this `.json` thing is a folder. In UNIX it could very well be a folder, there is no reason why it couldn't, so I need to check somewhere if it really is a folder and what the contents is.
- We are already using `-f` to overwrite the file, so `-R` doesn't make it more error prove at all, even if it would "copy even if errors are detected"

scripts/ci/update-github
```diff
@@ -0,0 +1,61 @@
...
+
+  git add "packages/dcos-ui/buildinfo.json"
+  git commit --amend --no-edit
+  git push --force origin dcos-ui/latest
```
DanielMSchmidt: I think we should use `--force-with-lease` here too

scripts/ci/update-github
```diff
@@ -0,0 +1,61 @@
...
+# Setup Git
+git config --global user.email "$GIT_USER@users.noreply.github.com"
+git config --global user.name "MesosphereCI Robot"
+git config credential.helper 'cache --timeout=300'
+git clone "https://$GIT_USER:$GIT_PASSWORD@github.com/mesosphere/dcos" dcos-nightly
+cd dcos-nightly
+
+if [[ $TAG_NAME ]]
+then
+  git checkout -- .
```
DanielMSchmidt: This line and the next one are duplicated after the else part, I think we should [DRY](http://programmer.97things.oreilly.com/wiki/index.php/Don't_Repeat_Yourself) this up and move it before the if

scripts/ci/update-github
```diff
@@ -0,0 +1,61 @@
...
+
+  cp -fR ../buildinfo.json packages/dcos-ui/buildinfo.json
+
+  git add "packages/dcos-ui/buildinfo.json"
+  git commit -m "chore(dcos-ui): update package"
+  git push --force origin dcos-ui/$TAG_NAME
+
+  cd ../
+
+  CHANGELOG=$(TAG_NAME=$TAG_NAME ./scripts/ci/generate-changelog)
```
DanielMSchmidt: I think the generate-changelog script is not dependent on the folder it is called from, at least as far as I could tell, so let's spare the `cd ../` and `cd dcos-nightly` and make it `../scripts/ci/generate-changelog`. The reason why I'm mentioning this is that I often find it quite hard to keep track of where I am directory wise in long bash scripts. Removing these parts free my mind a lot.

scripts/ci/update-github
```diff
@@ -0,0 +1,61 @@
...
+  PR_BODY_NO_NEWLINES_NO_HEADING=$(echo "${PR_BODY_NO_NEWLINES//$'#'/}")
+
+  pullRequest="{
+    \"title\": \"chore(dcos-ui): update package\",
+    \"body\": \"${PR_BODY_NO_NEWLINES_NO_HEADING}\",
+    \"head\": \"mesosphere:dcos-ui/$TAG_NAME\",
+    \"base\": \"master\"
+  }"
+
+  curl -i -u "$GIT_USER:$GIT_PASSWORD" https://api.github.com
```
DanielMSchmidt: Is this extra curl needed or was it just for testing?

scripts/ci/generate-changelog
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+set -ex
+
+MAJOR_MINOR_MATCH=$(echo $TAG_NAME | grep -Eo "^v?(0|[1-9]\d*)\.(0|[1-9]\d*)")
```
DanielMSchmidt: 👍

scripts/ci/generate-changelog
```diff
@@ -0,0 +1,32 @@
...
+
+  COMMIT_BODY=$(git log -1 ${COMMIT})
+  COMMIT_BODY_ONELINE=$(echo $COMMIT_BODY)
+  COMMIT_COMMENT=$(git log --format=%B -n 1 ${COMMIT})
+
+  if [[ $COMMIT_BODY_ONELINE =~ Closes\ ${JIRA_PREFIX}-[0-9]* ]]; then
+    TASK_NUMBER=$(echo $COMMIT_BODY_ONELINE | grep -Eo "${JIRA_PREFIX}-[0-9]*")
+    TASK_DESCRIPTION=$(echo "${COMMIT_COMMENT}" | grep -Eo -m1 ".*:(.*)" | cut -d: -f2)
+
+    MARKDOWN+="[${TASK_NUMBER}](https://jira.mesosphere.com/browse/${TASK_NUMBER}): ${TASK_DESCRIPTION}"
```
DanielMSchmidt: Wow, this requires a lot of discipline for our commit messages ^^

tags-delete
DanielMSchmidt: Wait a second; I did not realize this is on the root level of our repository. 
If it were in some scripts folder I would not care that much, but I think we should remove it entirely. It adds no value to our code base, but it is a source of confusion. If you find it super helpful, please move it to a [gist](http://gist.github.com/) and link it in our documentation.

scripts/ci/generate-changelog
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
```
nLight: Looks fragile to me. Have you looked into https://github.com/conventional-changelog/conventional-changelog with a custom preset?

tags-delete
ahoskins: Ooh, I agree. I will just delete this. Thanks

scripts/ci/update-github
ahoskins: Ok nice, that makes sense. I've switched it to use --force-with-lease

scripts/ci/update-github
ahoskins: Ok you've convinced me 😄, I've switched to remove the "r"!

Jenkinsfile
ahoskins: Yeah. When the build is triggered from the tag Jenkins will set the env.BRANCH_NAME to the tag name.

scripts/ci/update-github
ahoskins: Switched to use it.

scripts/ci/update-github
ahoskins: 👍

scripts/ci/update-github
ahoskins: Ah yes. We must do it this way, because we need to be in the proper `.git` context. The dcos-nightly has it's own git context, and we don't want the changelog to use this context.

scripts/ci/update-github
ahoskins: Yes we can remove, thanks

scripts/ci/generate-changelog
ahoskins: I agree this looks fragile. But let me explain why we went with this approach.

Within our dream team we discussed using conventional-changelog to generate changelogs. There are a few issues:
- it generates the changelog in a format not the same as the dcos/dcos PR template. So it would require us to write another (fragile) scripting phase to convert whatever it generates to the [template format](https://github.com/dcos/dcos/blob/master/PULL_REQUEST_TEMPLATE.md).
- conventional-changelog causes more changes for how we do releases. It would increment the version in `package.json` (which we currently *never* change). Also I have lots of questions regarding how this would all play with EE, since it doesn't have a `package.json`.

Me and @bstavroulakis think we should should this as-is, then make a follow-up ticket to improve the changelog in the future.

scripts/ci/generate-changelog
ahoskins: It simply requires us to follow the commitlint format, which is already enforced as post-commit
 script 😸

ahoskins:
Me and @bstavroulakis addressed these changes today, and really want to merge this. Follow up ticket will be to re-visit the changelog, if our current custom scripting becomes a problem.

Thanks for all the reviews over the last few days @nLight @DanielMSchmidt @juliangieseke! 😄

Jenkinsfile
```diff
@@ -4,6 +4,8 @@
 
 def master_branches = ["master", ] as String[]
 
+def SEMVER_REGEX = /^v?(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(-(0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(\.(0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*)?(\+[0-9a-zA-Z-]+(\.[0-9a-zA-Z-]+)*)?/
```
DanielMSchmidt: Awesome 👍

Jenkinsfile
```diff
@@ -40,21 +42,21 @@ pipeline {
...
-    }
-
-    stage('Unit Test') {
-      steps {
-        ansiColor('xterm') {
-          sh '''npm run test -- --maxWorkers=2'''
-        }
-      }
-    }
+    // stage('Lint') {
```
DanielMSchmidt: 😢 I think I would like to keep the tests

Jenkinsfile
```diff
@@ -68,70 +70,106 @@ pipeline {
...
-     steps {
-       withCredentials([
-          [
-            $class: 'AmazonWebServicesCredentialsBinding',
-            credentialsId: 'f40eebe0-f9aa-4336-b460-b2c4d7876fde',
-            accessKeyVariable: 'AWS_ACCESS_KEY_ID',
-            secretKeyVariable: 'AWS_SECRET_ACCESS_KEY'
-          ]
+    stage('Update Github'){
+      // when {
```
DanielMSchmidt: Why is this commented out?

Jenkinsfile
```diff
@@ -68,70 +70,106 @@ pipeline {
...
+    //      archiveArtifacts 'results/**/*'
+    //      junit 'results/results.xml'
+    //    }
+    //  }
+    // }
+
+    // when env.BRANCH_NAME == 'master', we will update the mesosphere:dcos-ui/latest branch
+    // when env.BRANCH_NAME =~ /(tag-name-regex)/, we will open a release PR against dcos/dcos
+    stage('Upload Build') {
+      // when {
```
DanielMSchmidt: Why is this commented out?

scripts/ci/update-github
```diff
@@ -0,0 +1,59 @@
...
+git config --global user.email "$GIT_USER@users.noreply.github.com"
+git config --global user.name "MesosphereCI Robot"
+git config credential.helper 'cache --timeout=300'
+git clone "https://$GIT_USER:$GIT_PASSWORD@github.com/mesosphere/dcos" dcos-nightly
+cd dcos-nightly
+git checkout -- .
+
+if [[ $TAG_NAME ]]
+then
+  git remote add upstream "https://$GIT_USER:$GIT_PASSWORD@github.com/dcos/dcos"
```
DanielMSchmidt: We can move this also before the if statement, right?

scripts/ci/update-github
```diff
@@ -0,0 +1,59 @@
...
+else
+  git remote add upstream "https://$GIT_USER:$GIT_PASSWORD@github.com/dcos/dcos"
+  git checkout dcos-ui/latest
+  git pull --rebase -X theirs upstream master
+
+  cp -f ../buildinfo.json packages/dcos-ui/buildinfo.json
+
+  git add "packages/dcos-ui/buildinfo.json"
+  git commit --amend --no-edit
+  git push --force-with-lease origin dcos-ui/latest
```
DanielMSchmidt: Thank you for adding `--force-with-lease` 👍

scripts/ci/upload-build
juliangieseke: hm, true. but i'd love to see us finally updating the version and using it correctly :(

Jenkinsfile
```diff
@@ -132,6 +133,54 @@ pipeline {
...
+    }
+
+    // trigger the other job to update the upstream reference
+    stage ('Trigger Enterprise Update') {
+      when {
+        expression { env.BRANCH_NAME == 'master' || env.BRANCH_NAME =~ SEMVER_REGEX }
+      }
+
+      steps {
+        build job: "frontend/dcos-ui-ee-release/${env.BRANCH_NAME}", parameters: [[$class: 'StringParameterValue', name: 'BRANCH_NAME', value: env.BRANCH_NAME]]
```
DanielMSchmidt: nice 👍

scripts/ci/generate-changelog
DanielMSchmidt: @ahoskins sounds good, can you link the follow-up ticket here?

ahoskins:
Follow-up for conventional-changelog improvement in the future: https://jira.mesosphere.com/browse/DCOS-21740.

Jenkinsfile
```diff
@@ -132,6 +133,54 @@ pipeline {
...
+      when {
+        expression { env.BRANCH_NAME == 'master' || env.BRANCH_NAME =~ SEMVER_REGEX }
+      }
+
+      steps {
+        withCredentials([
+            string(credentialsId: '3f0dbb48-de33-431f-b91c-2366d2f0e1cf',variable: 'AWS_ACCESS_KEY_ID'),
+            string(credentialsId: 'f585ec9a-3c38-4f67-8bdb-79e5d4761937',variable: 'AWS_SECRET_ACCESS_KEY')
+        ]) {
+          sh "BRANCH_NAME=${env.BRANCH_NAME} ./scripts/ci/upload-build"
```
juliangieseke: this is `BRANCH_NAME=v1.12.0-rc.1` when called from a tag, isnt it? this will break the `upload-build` script (see my comment below) :(

scripts/ci/upload-build
```diff
@@ -2,35 +2,33 @@
...
-FILENAME='dcos-ui'
-PKG_VERSION=$(python -c "import sys, json; \
-  pkg = json.load(open('$PACKAGE_PATH/package.json')); \
-  sys.stdout.write(str(pkg['version']));")
 PKG_BRANCH=$(echo ${BRANCH_NAME} | sed -E 's/([^/]*)$|./\1/g')
-RELEASE_NAME="${PKG_BRANCH}+${FILENAME}${BUILD_VARIANT}-v${PKG_VERSION}+${SHA_SLUG}.tar.gz"
+
+# ex. master+dcos-ui-v1.12.0-rc.11.tar.gz
+# ex. master+dcos-ui-ee-v1.12.0-rc.11.tar.gz
+RELEASE_NAME="${PKG_BRANCH}+dcos-ui${PKG_VERSION}.tar.gz"
```
juliangieseke: if this script is called with a tag (e.g. `BRANCH_NAME=v1.12.0-rc.1`) this will result in a broken filename (e.g. `v1.12.0-rc.1+dcos-ui-v1.12.0-rc.1.tar.gz` instead of `master+dcos-ui-v1.12.0-rc.1.tar.gz`) :(

juliangieseke: The Idea behind these filenames was to be able to have different version numbers for different branches. Imagine we use semver and there is a DC/OS UI version `1.5.3` which is released for DC/OS `1.11`, so the release name will be `1.11+dcos-ui-v1.5.3-somesha.tar.gz`.

juliangieseke: _(hit shift+enter again…)_

now imagine that we branched 1.11 on version `1.5.3` and on `master` (1.12), we're already on version `1.5.8`, but we want to cut a new release for DC/OS 1.11, the next `semver` release would be: `1.11+dcos-ui-1.5.4-somesha.tar.gz`

now, we have two `1.5.4` releases, one for DC/OS 1.11 and a second one for DC/OS "Next" (1.12), both files have to have a different filename, which was archivieved by the old script. the bucket would contain the following releases:
`1.11+dcos-ui-v1.5.3-shomesha.tar.gz`
`1.11+dcos-ui-v1.5.4-shomesha.tar.gz`
`master+dcos-ui-v1.5.4-shomesha.tar.gz`
`…`
`master+dcos-ui-v1.5.8-shomesha.tar.gz`
notice to "duplicate" on 1.5.4.

we are aware that this isnt a goot solution, but its the situation we have to deal with and for us it was the best compromise dealing with the issue. we basicly tried divide our releases by DC/OS target branch and handle them independently from each other. 

And, at last, sorry for not communitcatin this to you guys, this was discussed between orlando and myself and we (I) simply forgot to tell you guys why we went this way - I assumed you guys dont touch this script because it wasnt part of "your" work :( Apologies!

juliangieseke: To move forward, we could simply remove the `PKG_BRANCH` part from the filename, since we are currently using the old version scheme filenames like `dcos-ui-v1.12.0-rc.1` should be okay…

scripts/ci/upload-build
```diff
@@ -2,35 +2,33 @@
...
-# aws settings
 AWS_BUCKET='downloads.mesosphere.io'
 BUCKET_PATH='dcos-ui'
+BRANCH_NAME=${BRANCH_NAME:-'master'}
+PKG_VERSION=$(git tag -l --contains HEAD)
+
+if [[ $PKG_VERSION ]]
+then
+  PKG_VERSION="-$PKG_VERSION"
+fi
```
juliangieseke: If this is called from master, `PKG_VERSION` is empty, isnt it? which filename does the `latest` release have? If I get this correct ( https://github.com/dcos/dcos/pull/2506/files ), currently it will lead to a filename without a version or the `latest` suffix 🤔 (`master+dcos-ui.tar.gz`) - we should make sure, that this doesnt happen - but please to not simply add a `else` part here, because we dont want the `latest` release to be default…

Jenkinsfile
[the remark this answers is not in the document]
juliangieseke: ah ok, it is true until we merge the respective PR in EE - IMHO the EE one should be a dependency then, because once we merged this one, we will break it when we merge the EE one… 🤔

scripts/ci/upload-build
DanielMSchmidt: @juliangieseke I thought we wanted to do this refactoring in one of the next tickets next sprint

juliangieseke: yes, but I wasnt asking to refactor everything right now?!
The filenames are broken and this has to be fixed. Just wanted to provide some background info :)

juliangieseke: Cleaned up everything and added short notice that it is background info & what i think would be a proper fix right now. 👍

DanielMSchmidt: Ah okay, thank you for explaining, I got you now 👍

Jenkinsfile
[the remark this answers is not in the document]
juliangieseke: on this one again: can we move this to a follow-up PR please? there is a ticket for this, I dont think it is covered by the current tickets.

https://jira.mesosphere.com/browse/DCOS-21554

scripts/ci/upload-build
juliangieseke: lets get this merged anf fix the naming stuff later, I've created a ticket in pulled it into the sprint 👍 https://jira.mesosphere.com/browse/DCOS-21766

juliangieseke:
changed my mind, we can change the EE part in two steps. first move everything to the `Jenkinsfile`, second change reference here, third delete `Jenkinsfile.release`

PR #2798, opened 2018-02-28
DCOS_OSS-2206: Enzyme

DanielMSchmidt:
Some tests could not be migrated while updating jest, so I would like to move us in the direction of enzyme, that makes also writing tests that render more easily


ahoskins:
@DanielMSchmidt how does this relate to the other Enzyme PR?

DanielMSchmidt:
@ahoskins this is the new version, forgot to close the old one

PR #2799, opened 2018-03-01
V1.10.5a

kensipe:
This is a cherry pick of changes needed to support Jobs against metronome 0.4.0 which includes FORBID concurrency policy.

gif/image

```
<!-- Suplemental codeblock -->
```

**Checklist**

### DIFF
--- a/src/js/utils/JobUtil.js
+++ b/src/js/utils/JobUtil.js
@@ -82,7 +82,7 @@ const JobUtil = {
     let scheduleId = "default";
     let schedulePolicy = "ALLOW";
     // preserve id and concurrencyPolicy
-    if (typeof spec.schedules != "undefined" && spec.schedules.length > 0) {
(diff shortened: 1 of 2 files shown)


src/js/utils/JobUtil.js
```diff
@@ -78,18 +78,27 @@ const JobUtil = {
       Object.assign(spec.run, { docker });
     }
 
+    // default values for id and policy
+    let scheduleId = "default";
+    let schedulePolicy = "ALLOW";
+    // preserve id and concurrencyPolicy
+    if (typeof spec.schedules != "undefined" && spec.schedules.length > 0) {
```
MatApple: @kensipe as discussed - let's adjust to `Array.isArray(spec.schedules)` instead of `typeof spec.schedules != "undefined"`.

MatApple:
@kensipe thank you for the contribution 💯 Team always!!

PR #2800, opened 2018-03-02
Enable service endpoints on UCR

nLight:
This PR enables Service Endpoints partials for the UCR runtime. Previously it wasn't possible to have a Service endpoints running on UCR.

Please test JSON to Form and Form to JSON. Launching and editing the app.

## Testing
Choose UCR as a runtime
screen shot 2018-03-02 at 12 45 57

Give it a hard time testing permutations
screen shot 2018-03-02 at 12 46 13

Closes DCOS_OSS-1550

**Checklist**

### DIFF
--- a/plugins/services/src/js/components/forms/NetworkingFormSection.js
+++ b/plugins/services/src/js/components/forms/NetworkingFormSection.js
@@ -31,11 +31,9 @@ import SingleContainerPortMappings
 import {
   FormReducer as networks
 } from "../../reducers/serviceForm/FormReducers/Networks";
-import ContainerConstants from "../../constants/ContainerConstants";
 import ServiceConfigUtil from "../../utils/ServiceConfigUtil";
 
 const { BRIDGE, HOST, CONTAINER } = Networking.type;
-const { MESOS } = ContainerConstants.type;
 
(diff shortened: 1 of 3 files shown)


nLight:
run integration tests

juliangieseke:
One thing where I'm not exactly sure if it is as expected: when entering a port in the json, nothing happens in the form, but the port is actually send to marathon 🤔 
screen shot 2018-03-05 at 18 04 15

weblancaster:
@juliangieseke The host port isn't selectable. But does raise the question why we are sending the port to Marathon.

nLight:
@juliangieseke does this PR introduce the issue or is is reproducible for Docker runtime on master? I reported a port related bug https://jira.mesosphere.com/browse/DCOS-21307 maybe this is somewhat related

juliangieseke:
well, there is no way changing this at all on master, not sure if it introduces it, but before there was no form element to edit this fields. Not sure if we should reset the `port` when the form is edited or not 🤔

nLight:
@juliangieseke I'm not sure what's gone wrong in your case. I created an app with UCR and dcos network. Then I edited the app and it didn't switch network to the host mode. If your bug regarding the host network is reproducible please create a jira.

PR #2801, opened 2018-03-02
Downgrade jenkins label and enable coverage reporting

DanielMSchmidt:
**Checklist**


DanielMSchmidt:
@Fabs What do you mean? 🙈

Fabs:
Code Coverage as a tool will instrument your code to get its stats. That means turning it on definitely means tests itself use more CPU and memory. Also there is the extra time to compile the final report. In that sense, I just wanted to raise a flag 🎏 to make sure we are not adding unnecessary stress on our tests.

DanielMSchmidt:
Now I get you, thank you for explaining this in detail. As a matter of fact our unit tests currently take around 5 minutes. Our integration and system tests take 25 minutes each. Even if the time consumption would be bigger I think we still wouldn't sense it in the normal work life. 

Also I think getting metrics over our coverage can help us understand if we are going in the right direction and how much we are improving, therefore they add a value, at least to me

PR #2761, opened 2018-02-14
DCOS-21122: fix(ServiceForm): fixes focus/change/blur behaviour

juliangieseke:
---

ℹ️ Backport: #2762 

---

This ~commit~ PR changes the onChange handler of the ServiceForm so that
it no longer implicitly sets state.editingFieldPath, instead a new
internal function `handleFormFocus` is introduced.

Closes DCOS-21122

PR #2802, opened 2018-03-02
[Backport 1.10] DCOS-19865 fix focus/change/blur behaviour

juliangieseke:
ℹ️ Backport of https://github.com/dcos/dcos-ui/pull/2761

---

This commit changes the onChange handler of the ServiceForm so that
it no longer implicitly sets state.editingFieldPath, instead a new
internal function `handleFormFocus` is introduced.

The old implementation was broken and also was one of the reasons
why our tests were flaky, in fact some tests had to be fixed
after changing the form - they were expecting the broken state.

Also the previous implementation caused the form to lose information
when a radio or checkbox inpout was changed last in some situations.

Firefox for Example doesnt focus a radio input and therefore
never fires a blur event.

### DIFF
--- a/package.json
+++ b/package.json
@@ -129,6 +129,7 @@
   "scripts": {
     "check-env": "./scripts/validate-engine-versions",
     "clean": "rm -rf dist",
+    "cypress": "cypress run",
     "shrinkwrap": "npm shrinkwrap --dev && ./node_modules/.bin/gulp fixShrinkwrap",
     "preinstall": "./scripts/pre-install",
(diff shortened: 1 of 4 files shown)


juliangieseke:
run integration tests

juliangieseke:
closing this PR for now, we need to find a better solution for older releases and also want to rethink if we want to `rebase`. this task will be picked up again in a later sprint.

juliangieseke:
closed wrong PR…

PR #2803, opened 2018-03-02
DCOS-21520: check node version on checkout

juliangieseke:
This adds a small notice after you changed the branch, this should come in handy once we introduced the node upgrade but still have to work on older releases 👍 

Example with correct Node version:
screen shot 2018-03-02 at 17 15 22
And wrong one:
screen shot 2018-03-02 at 17 20 18

### DIFF
--- a/scripts/post-checkout
+++ b/scripts/post-checkout
@@ -1,0 +1,30 @@
+#!/usr/bin/env bash
+
+SCRIPT_PATH="$(dirname "$0")/$(dirname "$(readlink "$0")")"
+
+# Import utils
+source ${SCRIPT_PATH}/utils/message
+
+# strong some numbers
+NPM_VERSION="$(npm -v)"
+MAJOR_NPM_VERSION="$(npm -v | cut -d. -f1)"
+
+NODE_VERSION="$(node -v)"
+
+PACKAGE_DEV_DEPENDENCY="$(node -p -e "require('./package.json').devDependencies.node")"
+PACKAGE_ENGINE_DEPENDENCY="$(node -p -e "require('./package.json').engines.node")"
+
+VALIDATION_EXIT_CODE=$(${SCRIPT_PATH}/validate-engine-versions > /dev/null; echo $?)
+
+# checking some stuff
+if [ "$PACKAGE_DEV_DEPENDENCY" != "undefined" ] && [ "$MAJOR_NPM_VERSION" -lt 5 ]; then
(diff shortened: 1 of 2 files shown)


scripts/pre-install
```diff
@@ -12,6 +12,6 @@ source ${SCRIPT_PATH}/utils/marathon
 # git hooks fail to install
 ${SCRIPT_PATH}/validate-engine-versions && \
   install_marathon_raml "$MARATHON_VERSION" && \
-  install_hooks "pre-commit post-rewrite post-commit"
+  install_hooks "pre-commit post-rewrite post-commit post-checkout"
```
DanielMSchmidt: Nice to know there is something like post-checkout :)

scripts/post-checkout
```diff
@@ -0,0 +1,22 @@
...
+    fi
+else
+    ${SCRIPT_PATH}/validate-engine-versions > /dev/null
+    if [ $? -ne 0 ]; then
+        warning "This branch requires node version v$(node -p -e "require('./package.json').engines.node"), but you're running ${NODE_VERSION}."
+        exit 1
+    fi
+fi
+
+info "Looks like your node version is fine."
```
DanielMSchmidt: I think bash scripts have the convention to end with a new line

scripts/post-checkout
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
+
+SCRIPT_PATH="$(dirname "$0")/$(dirname "$(readlink "$0")")"
+
+# Import utils
+source ${SCRIPT_PATH}/utils/message
+NODE_VERSION=$(node -v)
+
+if [ "$(node -p -e "require('./package.json').devDependencies.node")" != "undefined" ]; then
+    if [ "$(echo $NODE_VERSION | | cut -d. -f1)" -lt 6 ]; then
```
DanielMSchmidt: As far as I understand after looking at it for 10 minutes this checks for the major version. Could you move this into a variable? `$MAJOR_NODE_VERSION -lt 6` would be nicer to read IMHO

juliangieseke:
updated the code to contain more variables, it should be way easier to understand now 👍 

also I changed the check a bit, it now checks vor `npm -v` instead of `node -v` - reason is, that `npm` has to be up to date in order to support `node` as a dependency (also it has to be a decent node version, but that should be assured in the package.json, not here…

PR #2806, opened 2018-03-05
DCOS-21369: fix(TasksView): fix undefined service in TasksView

juliangieseke:
Small fix which assures that service is set before accessing its childnodes.

Closes DCOS-21369


juliangieseke:
run system tests

PR #2807, opened 2018-03-06
fix(commitlint): update the commit linting to new versions

ahoskins:
The old packages were deprecated. This uses the new packages, and changes the config
file to the new way.

**Checklist**

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,6 +72,8 @@
     "tv4": "1.2.7"
   },
   "devDependencies": {
+    "@commitlint/cli": "^6.1.3",
+    "@commitlint/config-conventional": "^6.1.3",
     "@dcos/eslint-config": "0.3.0",
     "autoprefixer": "6.3.6",
     "babel-cli": "6.24.1",
@@ -87,8 +89,6 @@
     "babel-runtime": "6.23.0",
     "colors": "1.1.2",
     "compression-webpack-plugin": "0.4.0",
-    "conventional-changelog-lint": "1.1.9",
-    "conventional-changelog-lint-config-angular": "0.4.1",
     "css-loader": "0.23.1",
     "cypress": "1.4.1",
     "eslint": "3.15.0",
@@ -157,7 +157,7 @@
     "eslint": "bash -c 'opts=$@ ; set -f; if [[ $# -lt 1 ]]; then opts=\"{src/js,plugins,packages}/**/*.js\"; fi ; eslint --quiet ${opts}' 0",
     "cypress": "cypress run",
     "stylelint": "bash -c 'opts=$@ ; set -f; if [[ $# -lt 1 ]]; then opts=\"{src/styles,plugins,packages}/**/*.less\"; fi ; stylelint ${opts}' 0",
-    "conventional-changelog-lint": "conventional-changelog-lint",
+    "commitlint": "commitlint",
     "lint-plugins": "PLUGINS=$npm_config_externalplugins; npm run eslint -- \"$PLUGINS/**/*.js\" && npm run stylelint -- \"$PLUGINS/**/*.less\"",
     "prebuild": "./scripts/validate-tests",
     "serve": "npm run build-assets && ./node_modules/.bin/gulp serve",
--- a/package.json
+++ b/package.json
@@ -72,8 +72,8 @@
     "tv4": "1.2.7"
   },
   "devDependencies": {
-    "@commitlint/cli": "^6.1.3",
-    "@commitlint/config-conventional": "^6.1.3",
+    "@commitlint/cli": "6.1.3",
+    "@commitlint/config-conventional": "6.1.3",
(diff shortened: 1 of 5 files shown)


juliangieseke:
```
FAIL  plugins/services/src/js/reducers/serviceForm/__tests__/Container-test.js
  ● Test suite failed to run

    Error
      uglifyjs is deprecated - use uglify-js instead.
```
:(

DanielMSchmidt:
@juliangieseke My PR based on this fails with the same error, there seems to be a real problem

juliangieseke:
something is weird with the shrinkwrap here 🤔

ahoskins:
I re-did the shrinkwrap and hopefully it fixes this.  Also @DanielMSchmidt, agree. Let's add more more rules for committing. I don't see any issue doing it here in this PR - no rush to get this merged.

ahoskins:
@DanielMSchmidt Just noticed: since I extend the "angular" commit config it already uses the scope-enum rule. For full list of rules that will be enforced, see [here](https://github.com/marionebl/commitlint/blob/master/docs/reference-rules.md#scope-enum). We may want to add more still.

ahoskins:
I don't know what's going on here with that uglifyjs thing. What I'm doing is (with node 4.4):

1. rm npm-shrinkwrap
2. rm -rf node_modules
3. npm i
4. npm run shrinkwrap
5. commiting the shrinkwrap

Anyone feel free to push with correction if I'm doing this wrong... 😏

weblancaster:
@ahoskins to fix the failing job.

`rm -rf npm-shrinkwrap.json && rm -rf node_modules` then modify your `package.json` new devDependencies to look like below.

```
"@commitlint/cli": "6.1.3",
"@commitlint/config-conventional": "6.1.3",
```

After all that run `npm i && npm run build-shrinkwrap`. To test worked run `npm run prebuild-assets` and then just update your PR with the changes.

Let me know if that works for you.

package.json
```diff
@@ -157,7 +157,7 @@
     "eslint": "bash -c 'opts=$@ ; set -f; if [[ $# -lt 1 ]]; then opts=\"{src/js,plugins,packages}/**/*.js\"; fi ; eslint --quiet ${opts}' 0",
     "cypress": "cypress run",
     "stylelint": "bash -c 'opts=$@ ; set -f; if [[ $# -lt 1 ]]; then opts=\"{src/styles,plugins,packages}/**/*.less\"; fi ; stylelint ${opts}' 0",
-    "conventional-changelog-lint": "conventional-changelog-lint",
+    "commitlint": "./node_modules/.bin/commitlint",
```
juliangieseke: (why) do we need this? feels like a workaround for something we should fix in a different way 🤔

ahoskins: Do you mean that in the `post-commit` we should just directly call call the commitlint binary, rather than add it to the package.json as a script?

juliangieseke: No, I was wondering why we were able to call the old script without `./node_modules/.bin/`

weblancaster: @juliangieseke 
This is not a workaround but a good nodejs practice to not target global module installs.

Note that we are also using for other npm commands, so nothing new.

juliangieseke: @weblancaster do you have any references for this? AFAIK npm prepends `./node_modules/.bin/` to your `$PATH` variable - if it is configured to do so. 

I remember reading something about absolute paths and that they should be avoided because npm might be set up differently on certain machines. 🤔 
so let me ask again: what is the reason you decided to write it like so? did you try to fix a conflict with a globally installed version?

juliangieseke:
```

weblancaster: @juliangieseke you just experienced one of the problems when pointing to global modules. As you pointed out your version is `5.2.8` but the expected version is `6.1.3`. Which `npm run commitlint -- --version` is correct.

ahoskins: Yes, agree with @weblancaster. We all want to be using the same version of packages, and being explicit about the path in `package.json` does this.

weblancaster: There are many references you can use

http://2ality.com/2016/01/locally-installed-npm-executables.html
https://firstdoit.com/no-need-for-globals-using-npm-dependencies-in-npm-scripts-3dfb478908

juliangieseke: Ok, looks like we were talking about different things. 🤔
My only question is, why you wrote `./node_modules/.bin/commitlint` instead of just `commitlint` in the package.json. I’m not saying we should use a globally installed version!

From your second link:
> But npm has another trick up its sleeve. When evaluating scripts, npm adds node_modules/.bin/ to the PATH environment variable. That means you can forego the folder prefix entirely!

weblancaster: Yes npm automatically looks at the local bin but let's keep it safer and strictly point to local especially because we are already using that way.

juliangieseke: As we pointed out in our last retro: "because we are already using that way" isnt a good reason to do something. I'd appreciate if we

PR #2808, opened 2018-03-06
Lint all commits in a PR

DanielMSchmidt:
Had 10 minutes before lunch and @juliangieseke wanted to have each commit tested for a PR

```sh
[cos-ui-oss-pipeline_PR-2808-BOEJZQ7G7EXOOUCJZXKSP3XVN25NMZK4BHFHRIP5AADZEGLKJDYA] Running shell script

+ ./scripts/ci/lint-all-commits

fatal: ambiguous argument 'master..HEAD': unknown revision or path not in the working tree.

Use '--' to separate paths from revisions, like this:

'git <command> [<revision>...] -- [<file>...]'

script returned exit code 128
```

**Checklist**

### DIFF
--- a/scripts/ci/lint-all-commits
+++ b/scripts/ci/lint-all-commits
@@ -1,0 +1,17 @@
+#!/bin/bash
+
+set -e
+
+GIT_BRANCH=${GIT_BRANCH#*/}
+
+SHAS=$(git log master..$(git rev-parse --abbrev-ref HEAD) --pretty=format:%h)
+
+for SHA in "${SHAS[@]}"
+do
+  echo "Linting $SHA"
+  MESSAGE=$(git log --format=%B -n 1 $SHA)
+
+  echo "$MESSAGE" | npm run commitlint
+done
+
+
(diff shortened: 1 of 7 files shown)


scripts/ci/lint-all-commits
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+set -e
+
+GIT_BRANCH="${env.BRANCH_NAME}"
```
DanielMSchmidt: @juliangieseke env.BRANCH_NAME is the jenkins naming, you would need to do `GIT_BRANCH=env.BRANCH_NAME ./scripts/ci/lint-all-commits` in the Jenkinsfile

ahoskins:
I'm a bit confused why this is needed. Don't we lint commits when we commit them, and that's good enough?

DanielMSchmidt:
@ahoskins I think @juliangieseke mentioned that only the last commit is looked at 🤔

juliangieseke:
yup, wanted to check if we can lint not only the last commit in a PR. but all of them (to be able to generate our changelogs more reliable)

@DanielMSchmidt are you okay with closing this for now and picking this up in a later stage? Related ticket is already created: https://jira.mesosphere.com/browse/DCOS-21600

PR #2814, opened 2018-03-07
DCOS_OSS-2206: Move to enzyme for rendering tests

DanielMSchmidt:
react-addons-test-utils is deprecated, we need to drop it in order to move to React 16.

**Checklist**


Fabs:
Did you write a magic conversion or you are doing it manually?

Have you considered moving only the broken tests to enzyme, maybe write some docs and reference use (or just share links) and move on?

I think we can as a team move the tests to enzyme, that helps us share knowledge and do it gradually. We can say from now on that if there is a new PR you cant use the old way.

DanielMSchmidt:
@Fabs Yeah, I would love to move step by step in the course of the next months, but I think that removing the test utils would provide a huge value (e.g. us being able to move to React 16 and also hopefully more rendering tests because it got easier).
If you feel like I am wasting my time, just tell me & let's discuss. This work can be merged and the transition could be moved forward at a later point in time, there is no problem with it.

DanielMSchmidt:
Oh and it is manual, magic doesn't work here as there are too many edge cases

package.json
```diff
@@ -128,6 +129,7 @@
     "prettier": "1.2.2",
     "purify-css": "1.2.5",
     "raml-validator-loader": "0.1.11",
+    "react-addons-test-utils": "15.6.2",
```
DanielMSchmidt: We need this as long as we are under React 15 because of `enzyme-adapter-react-15.4`

plugins/nodes/src/js/pages/__tests__/NodeDetailPage-test.js
```diff
@@ -1,123 +0,0 @@
-const JestUtil = require("#SRC/js/utils/JestUtil");
```
DanielMSchmidt: I found this test to be really hard to migrate and not really of value because the only check was `expect(node instanceof Node).toEqual(true);`

plugins/services/src/js/components/ServiceBreadcrumbs.js
```diff
@@ -92,6 +92,10 @@ class ServiceBreadcrumbs extends React.Component {
     const lastBreadcrumbNode = breadcrumbsNode.querySelector(
       ".breadcrumb:last-child"
     );
+
+    if (!lastBreadcrumbNode) {
```
DanielMSchmidt: I had to add this fix (that doesn't hurt but secures against errors on older browsers), otherwise the test would have been super hard to write

plugins/services/src/js/components/__tests__/LogView-test.js
```diff
@@ -52,58 +42,44 @@ describe("LogView", function() {
...
       expect(Array.isArray(res)).toEqual(true);
     });
 
     it("gets empty screen when log is empty", function() {
-      thisInstance.state.fullLog = "";
-      var res = thisInstance.getLog();
-      expect(TestUtils.isElementOfType(res, EmptyLogScreen)).toEqual(true);
-    });
-
-    it("calls getLog when log is empty", function() {
```
DanielMSchmidt: I removed these tests, because spying on `getLog` was very hard, but if you see the component, it just has a render method that always calls `getLog`

src/js/utils/__tests__/RouterUtil-test.js
```diff
@@ -68,27 +73,32 @@ describe("RouterUtil", function() {
       ]);
       const component = components[0].props.children;
 
-      expect(
-        ReactTestUtils.isElementOfType(component, ReactRouter.Redirect)
-      ).toBeTruthy();
+      try {
+        // Duck typing to check if it is a route
```
DanielMSchmidt: I know, this is not the nicest solution, but I found it easier and better then other similar soltions that would have removed ReactTestUtils

ahoskins:
@DanielMSchmidt What is the state of this? Tests are passing, and looks fine to me. Ready for final approvals?

DanielMSchmidt:
Yeah, it is just waiting for someone to review it. Therefore no label

Andrew Hoskins <notifications@github.com> schrieb am Fr., 16. März 2018,
23:54:

src/js/utils/__tests__/RouterUtil-test.js
nLight: Let's try another approach 🙇

packages/foundation-ui/src/mount/__tests__/Mount-test.js
```diff
@@ -25,147 +25,128 @@ describe("Mount", function() {
...
+      result.containsAllMatchingElements([<strong>foo</strong>, <em>bar</em>])
+    ).toBeTruthy();
   });
 
   it("renders null if no component is registered and no children defined", function() {
-    const renderer = TestUtils.createRenderer();
-    renderer.render(<Mount type="children-test" />);
+    const result = shallow(
+      <Mount type="children-test">
+        <Mount type="children-test" />
```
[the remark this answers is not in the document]
DanielMSchmidt: Strange 🤔  Might be an error, will rewrite it.

plugins/nodes/src/js/pages/__tests__/NodeDetailPage-test.js
DanielMSchmidt: Why do you think it provides value?

plugins/services/src/js/components/ServiceBreadcrumbs.js
DanielMSchmidt: enzyme threw an error that it couldn't calculate the value which is calculated here, because `lastBreadcrumbNode` was undefined. I searched for it and there was not really a good solution, at least I did not find it. I thought this check doesn't make our codebase worse, but maybe more robust, so I don't see any harm in it

src/js/utils/__tests__/RouterUtil-test.js
DanielMSchmidt: Let's pair on this, I have no idea how to do it differently

DanielMSchmidt:
@nLight addressed your issues, could you take another look?

PR #2815, opened 2018-03-08
refactor(githooks): move and improve post-checkout checks into validate-engines

juliangieseke:
the current version of this check complains about node version 9.x being incorrect (it needs 8.9.4). this fixes this issue and also moves the checks into our validate-engines script.

to test this, please checkout this branch with latest node version (should NOT complain), then downgrade node to v4 and try `npm start` (should complain)

### DIFF
--- a/package.json
+++ b/package.json
@@ -163,12 +163,13 @@
     "serve": "npm run build-assets && ./node_modules/.bin/gulp serve",
     "serve-notify": "export NOTIFY=true && npm run serve",
     "scaffold": "./node_modules/.bin/gulp ensureConfig && ./node_modules/.bin/gulp ensureDevProxy",
+    "prestart": "npm run check-env",
     "start": "NODE_ENV=development webpack-dev-server --config ./webpack/webpack.dev.babel.js --hot --progress --colors --content-base ./build --host 0.0.0.0 --port $npm_package_config_port",
     "test": "node jest/gen-config.js && NODE_PATH=node_modules jest --config=jest.config.json --no-cache",
     "test:watch": "npm test -- --watch",
     "testing": "NODE_ENV=testing webpack-dev-server --config ./webpack/webpack.dev.babel.js --progress --colors --content-base ./build --host 0.0.0.0 --port $npm_package_config_port",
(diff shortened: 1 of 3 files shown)


ahoskins:
@juliangieseke When using nvm, it does not have an issue with me using node 4.4! When reading the script itself, this makes sense. I think we have to modify the assertions a bit.

juliangieseke:
Yay! Even better, having `npm` as devDependency also installs the correct version locally! this means it _doesnt matter at all_ which version is installed on our machines. 🎉 

check this:
```
# Julian@Cookie:~/Code/dcos-ui/master/dcos-ui (jgieseke/improve-version-check *$%=)
$ node -v
v4.4.7
# Julian@Cookie:~/Code/dcos-ui/master/dcos-ui (jgieseke/improve-version-check *$%=)
$ npm -v
3.9.6
# Julian@Cookie:~/Code/dcos-ui/master/dcos-ui (jgieseke/improve-version-check *$%=)
$ npm run versions

> dcos-ui@1.1.14 versions /Users/Julian/Code/dcos-ui/master/dcos-ui
> echo npm $(npm --version) && echo node $(node --version)

npm 5.6.0
node v8.9.4
```

DanielMSchmidt:
@ahoskins I think when you use nvm you need to open a new tab to have the right version. Like this: 
node

ahoskins:
screen shot 2018-03-13 at 11 08 02 am

DanielMSchmidt:
Strange, might be my shell. It's fish

juliangieseke:
@ahoskins was right, it didnt complain. see my comment above, now it is even better :)
@ahoskins pls re-review 👍

ahoskins:
I re-installed node_modules, and then used node 4.4.7, and still no error...

ahoskins:
It's interesting what's happening here. So I switch my global node 4.4 successfully. Then if you notice, my PATH doesn't include node_modules/bin. But then, when invoking the script, it appends .node_modules/bin to my path, causing it to use the correct node version in the script. 

image 2018-03-14 at 4 37 54 pm

PR #2816, opened 2018-03-08
DCOS_OSS-2243: fix(taskdetails): remove misleading endpoints row from config table

juliangieseke:
This removes the endpoints Row from the Task Detail Table because it was misleading (Discussion over there: https://jira.mesosphere.com/browse/DCOS_OSS-1554 )

To test this create a service or pod, go to Services > <SERVICENAME> > <TASK> and check that Endpoints is not listed anymore.

A follow-up ticket will implement a new "IP Addresses" row which is what this should have been in the first place: https://jira.mesosphere.com/browse/DCOS_OSS-2244

### DIFF
--- a/plugins/services/src/js/pages/task-details/TaskDetailsTab.js
+++ b/plugins/services/src/js/pages/task-details/TaskDetailsTab.js
@@ -17,7 +17,7 @@ import Units from "#SRC/js/utils/Units";
 
 import MarathonTaskDetailsList from "../../components/MarathonTaskDetailsList";
 import TaskDirectoryStore from "../../stores/TaskDirectoryStore";
-import TaskEndpointsList from "../../components/TaskEndpointsList";
+// import TaskEndpointsList from "../../components/TaskEndpointsList";
 import TaskUtil from "../../utils/TaskUtil";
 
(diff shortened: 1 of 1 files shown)


plugins/services/src/js/pages/task-details/TaskDetailsTab.js
```diff
@@ -17,7 +17,7 @@ import Units from "#SRC/js/utils/Units";
 
 import MarathonTaskDetailsList from "../../components/MarathonTaskDetailsList";
 import TaskDirectoryStore from "../../stores/TaskDirectoryStore";
-import TaskEndpointsList from "../../components/TaskEndpointsList";
+// import TaskEndpointsList from "../../components/TaskEndpointsList";
```
weblancaster: why not just remove the code @juliangieseke? I'm afraid this will be here for a long time.

juliangieseke: (hopefully) not, this is subtask #1 of its parent ticket, subtask #2 is also in this sprint, but i have to sync with dima what exactly we want to do.

PR #2819, opened 2018-03-09
[Backport 1.10] - fix(labelsReducer): fix labels reducer to accept empty strings

Poltergeist:
This commit is fixing the labels reducer to support labels with an empty string value.

Closes DCOS-19693

### DIFF
--- a/plugins/services/src/js/components/modals/CreateServiceModalForm.js
+++ b/plugins/services/src/js/components/modals/CreateServiceModalForm.js
@@ -82,7 +82,20 @@ class CreateServiceModalForm extends Component {
     // Hint: When you add something to the state, make sure to update the
     //       shouldComponentUpdate function, since we are trying to reduce
     //       the number of updates as much as possible.
+    // In the Next line we are destructing the config to keep labels as it is and even keep labels with an empty value
+    const { labels = {}, ...serviceConfig } = ServiceUtil.getServiceJSON(
+      this.props.service
+    );
 
+    let newServiceConfig = {
+      labels,
+      ...CreateServiceModalFormUtil.stripEmptyProperties(serviceConfig)
+    };
+    if (Object.keys(labels).length === 0) {
+      newServiceConfig = CreateServiceModalFormUtil.stripEmptyProperties(
+        serviceConfig
+      );
+    }
     this.state = Object.assign(
       {
         appConfig: null,
@@ -95,9 +108,7 @@ class CreateServiceModalForm extends Component {
         jsonParser() {}
       },
       this.getNewStateForJSON(
-        CreateServiceModalFormUtil.stripEmptyProperties(
-          ServiceUtil.getServiceJSON(this.props.service)
-        ),
+        newServiceConfig,
         this.props.service instanceof PodSpec
       )
     );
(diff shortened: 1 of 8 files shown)


juliangieseke:
Alfred test this

Fabs:
cluster for 1.10 - 54.186.128.184

Fabs:
cluster for 1.9 - 34.217.26.84 (4h)

Fabs:
I have the same problem here:
https://github.com/dcos/dcos-ui/pull/2820#pullrequestreview-103379166

plugins/services/src/js/components/modals/CreateServiceModalForm.js
```diff
@@ -82,7 +82,22 @@ class CreateServiceModalForm extends Component {
     // Hint: When you add something to the state, make sure to update the
     //       shouldComponentUpdate function, since we are trying to reduce
     //       the number of updates as much as possible.
+    // In the Next line we are destructing the config to keep labels as it is and even keep labels with an empty value
+    let { serviceConfig } = ServiceUtil.getServiceJSON(this.props.service);
```
DanielMSchmidt: I don't understand what happens here (next 14 lines). 

You put the lables into a constant, remove? the labels from the config and add them again after stripping the empty properties, but only if there were labels?
This whole segment is super confusing to me, I need to lookup `omit` and `CreateServiceModalFormUtil.stripEmptyProperties`, understand why it's there twice.

It seems like you "just" want to have labels in the result of `CreateServiceModalFormUtil.stripEmptyProperties` if it wasn't empty and this change seems far to complicated for that, at least to me. 

Am I missing some context? Can you refactor this?

plugins/services/src/js/components/modals/CreateServiceModalForm.js
```diff
@@ -318,7 +331,18 @@ class CreateServiceModalForm extends Component {
       baseConfigCopy
     );
 
-    return CreateServiceModalFormUtil.stripEmptyProperties(newConfig);
+    const config = omit(newConfig, ["labels"]);
+    const { labels } = newConfig;
+
+    if (Object.keys(labels).length === 0) {
```
DanielMSchmidt: It seems like this could also be a function, given the config returning whatever version of the config is tried to be build here

DanielMSchmidt:
Oh, it's a backport, I should have said something in the first PR I guess 🙈

DanielMSchmidt:
run integration tests

juliangieseke:
whoever continues to work on this: please drop my commit, it didnt help…

DanielMSchmidt:
@juliangieseke removed it

DanielMSchmidt:
I tried to make the code more robust for nulls, but I think that did not help. It appears to me that some services just don't spin up properly, so it might be a valid cause

DanielMSchmidt:
The tests are legit failures, so I am setting the WIP flag here 🎏

weblancaster:
@DanielMSchmidt I don't understand why this is a WIP if we have build failures. When WIP I don't even look at the PR because it's not ready for review.

I thought WIP was only for missing work in the PR and not for legit failures due to bugs or such.

DanielMSchmidt:
It's WIP because there is still work to be done. We can not merge the PR like this, because the tests are broken, therefore it makes no sense that anyone looks at it right now

nLight:
@Poltergeist ping

Poltergeist:
There was more back-porting needed since the original fix used stage-3 features. which I now enabled. So there will be no legit failures anymore.

weblancaster:
This works fine, the only thing I have is that when I add a label but I don't any value to them and then deploy.

Note the configuration display the `labels` section.

screen shot 2018-04-09 at 3 52 18 pm

Poltergeist:
Let’s go!

PR #2820, opened 2018-03-12
[Backport 1.9] fix(labelsReducer): fix labels reducer to accept empty strings

Poltergeist:
This commit is fixing the labels reducer to support labels with an empty string value.

Closes DCOS-21467


plugins/services/src/js/components/modals/NewCreateServiceModalForm.js
```diff
@@ -83,7 +84,22 @@ class NewCreateServiceModalForm extends Component {
...
+    const { labels = {} } = serviceConfig;
 
+    serviceConfig = omit(serviceConfig, ["labels"]);
+
+    let newServiceConfig = Object.assign(
+      {},
+      labels,
+      CreateServiceModalFormUtil.stripEmptyProperties(serviceConfig)
+    );
+    if (Object.keys(labels).length === 0) {
```
jaxesn: Why is this if block necessary? What does line 93 produce when `labels = {}`?

Fabs:
My first question would be: This does not affect 1.11? just 1.10 and 1.09? I asking because I started testing it on master, and it works. I am now trying on release/1.09.

Also maybe we should indicate on the header of this PR that it is 1.09

plugins/services/src/js/components/modals/NewCreateServiceModalForm.js
Fabs: I think it would be awesome if we could add a test for this behavior, or bring the tests we have for 1.11 to here too. Also if the test is not present on master, I would recommend adding tests on master too (maybe even 1.11)

Fabs:
cluster for 1.9 - 34.217.26.84 (4h)

Fabs:
Also I get an error when I try this feature, I have release/1.9 as plugins and ph/fix/DCOS-21467-labels-1.9-backports.

I cant launch a service, I get the error: 
https://cl.ly/0C332t2a3s2F

Poltergeist:
We already merged this in 1.11 and 1.12 ... I had tests for this but they are not present in this PR since we added tests in 1.11 they are also not present in 1.10.

plugins/services/src/js/components/modals/NewCreateServiceModalForm.js
Poltergeist: Line 93 would produce:
```JSON
{labels: {}}
```
The better question is what line 99 produces if labels is `{}`
Because in that case the labels are not present as they are getting stripped.

Fabs:
Could we backport the tests also from 1.11 @Poltergeist ?

Poltergeist:
Sure we can but then I need to reestimate the issue, AS I also need to adjust all the other tests which are in that file. So that they pass for 1.9.

DanielMSchmidt:
I get this error when trying to go in the edit field. Doesn't happen on 1.9 branch.
bildschirmfoto 2018-04-06 um 15 54 42

juliangieseke:
@Fabs you're still blocking this, can you revisit please?

PR #2823, opened 2018-03-14
Low Prio: do not generate the jest config on every run

DanielMSchmidt:
It just bothered me and I gave it 5 minutes, seems to work just fine 👍 

**Checklist**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,7 +14,6 @@ tmp/
 *.orig
 .plugins/
 .external_plugins/
-jest.config.json
 jest/test-results
 webpack/proxy.dev.js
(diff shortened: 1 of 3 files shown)


juliangieseke:
@DanielMSchmidt not exactly sure what happened, but this PR also has the `http-server permission denied` issue from the webpack PR 🤔 Cant see anything obvious on master which could have caused this - but i noticed, that `http-server` isnt installed in the `Dockerfile` anymore? 🤔

DanielMSchmidt:
Seemed like a flaky issue, now it's green 🤔

PR #2824, opened 2018-03-14
DCOS_OSS-2244: Show task ip addresses

nLight:
This PR adds `IP Addresses` row to the Task Details Table.

## Testing

Please run several apps/pods with different networks: host, dcos, bridge. Verify that the tab shows proper IP Addresses for tasks. You can look up the addresses in the Marathon `groups` response.
For Apps it will be in: `apps[].tasks[].ipAddresses[]`
For Pods `pods[].instances[].networks[].addresses[]`

**NB!** Please note there's a  bug in the UI. UCR runtime doesn't actually support `dcos6`

## Look
screen shot 2018-03-14 at 19 36 24

**Checklist**


plugins/services/src/js/components/TaskEndpointsList.js
```diff
@@ -1,172 +0,0 @@
-import PropTypes from "prop-types";
```
DanielMSchmidt: Great that you removed all this old code 👍

plugins/services/src/js/pages/task-details/TaskIpAddressesRow.js
```diff
@@ -0,0 +1,43 @@
...
+import Application from "#PLUGINS/services/src/js/structs/Application";
+import ConfigurationMapLabel from "#SRC/js/components/ConfigurationMapLabel";
+import ConfigurationMapRow from "#SRC/js/components/ConfigurationMapRow";
+import ConfigurationMapValue from "#SRC/js/components/ConfigurationMapValue";
+import DCOSStore from "#SRC/js/stores/DCOSStore";
+import Pod from "#PLUGINS/services/src/js/structs/Pod";
+
+export default ({ task }) => {
+  if (task == null) {
+    return <noscript />;
```
DanielMSchmidt: can't we use null here?

plugins/services/src/js/pages/task-details/TaskIpAddressesRow.js
```diff
@@ -0,0 +1,43 @@
+import React from "react";
+
+import Application from "#PLUGINS/services/src/js/structs/Application";
+import ConfigurationMapLabel from "#SRC/js/components/ConfigurationMapLabel";
+import ConfigurationMapRow from "#SRC/js/components/ConfigurationMapRow";
+import ConfigurationMapValue from "#SRC/js/components/ConfigurationMapValue";
+import DCOSStore from "#SRC/js/stores/DCOSStore";
+import Pod from "#PLUGINS/services/src/js/structs/Pod";
+
+export default ({ task }) => {
```
DanielMSchmidt: As far as I know (from @juliangieseke I think) we only use arrow functions when they are necessary. I personally don't like that, I use them in other projects all the time, but I think we should have a convention here

plugins/services/src/js/pages/task-details/TaskIpAddressesRow.js
```diff
@@ -0,0 +1,43 @@
...
+  if (task == null) {
+    return <noscript />;
+  }
+
+  let ipAddresses = [];
```
DanielMSchmidt: I have seen this pattern often, is there a special reason why we don't do things like this in a  function? For me 

```js
function getIPAddressesForTask(service, task) {
  if (service instanceof Application) {
    const marathonTask = service.findTaskById(task.id) || { ipAddresses: [] };
 
    return marathonTask.ipAddresses.map(
      ipAddress => ipAddress.ipAddress
    );
  }
 
  if (service instanceof Pod) {
    const instance = service.findInstanceByTaskId(task.id);
 
    return instance ? instance.getIpAddresses() : [];
  }
  
  return [];
}
```

would be clearer, given the assumption that it can only be either an Application or a Pod. The current implementation leaves room for speculation there. For me the main advantage would be that we could export this function and test it seperately

plugins/services/src/js/pages/task-details/TaskIpAddressesRow.js
DanielMSchmidt: Can we use propTypes here or why does it make no sense?

plugins/services/src/js/pages/task-details/TaskIpAddressesRow.js
nLight: not in a stateless component

plugins/services/src/js/pages/task-details/TaskIpAddressesRow.js
nLight: ...and the convention was _not_ using them at all, and it was me who suggested that :trollface: I will fall back to a regular component. That will allow me to use `null` as a return value. Thank you for 👮

plugins/services/src/js/pages/task-details/TaskIpAddressesRow.js
DanielMSchmidt: I'll put this to my TIL list 👍

nLight:
@DanielMSchmidt I addressed your comments, please give it another look! :)

plugins/services/src/js/pages/task-details/TaskIpAddressesRow.js
```diff
@@ -0,0 +1,52 @@
...
+
+import Application from "#PLUGINS/services/src/js/structs/Application";
+import ConfigurationMapLabel from "#SRC/js/components/ConfigurationMapLabel";
+import ConfigurationMapRow from "#SRC/js/components/ConfigurationMapRow";
+import ConfigurationMapValue from "#SRC/js/components/ConfigurationMapValue";
+import DCOSStore from "#SRC/js/stores/DCOSStore";
+import Pod from "#PLUGINS/services/src/js/structs/Pod";
+
+export default class TaskIpAddressesRow extends React.Component {
+  getIPAddressesForTask(service, taskId) {
```
DanielMSchmidt: Also possible, just out of curiosity, why did you not use a function instead of a method?

nLight: It wasn't difficult to test it with snapshots and I decided keeping it in the class rather then extracting it.

DanielMSchmidt: okay 👍

plugins/services/src/js/utils/ServiceUtil.js
```diff
@@ -119,10 +119,10 @@ const ServiceUtil = {
   getServiceIDFromTaskID(taskID = "") {
     // Parse the task id (e.g. foo_bar.abc-123) to get the corresponding
     // service id parts (foo, bar)
-    const parts = taskID.match(/([^_]+)(?=[_.])/g);
+    const parts = taskID.split(".")[0].split("_");
```
jaxesn: nice cleanup

ahoskins:
^^ Having this issue where it's not showing in the table.

nLight:
@ahoskins Could you please post the full spec of the app. I can't replicate the issue you're facing.

DanielMSchmidt:
@ahoskins @nLight I got the same behaviour for services where no task has been in the Running state yet, but I thought that would be ok. I mean there are no IP addresses if there is no task running, right?

juliangieseke:
maybe we should hide the whole line of a task isnt running?

DanielMSchmidt:
@juliangieseke I think that having it sometimes there and sometimes not is not so nice for our users. If you feel strongly about it, let's make a follow-up ticket to define the empty state for this 👍

PR #2826, opened 2018-03-15
fix(scripts): make post-checkout executable

juliangieseke:
file permissions were wrong :(


DanielMSchmidt:
That warning was so hard to spot, sorry for not picking it up during review :/

PR #2829, opened 2018-03-15
DCOS_OSS-2190: Remove unused dependencies

DanielMSchmidt:
I could not find any traces of these dependencies, most of them were pretty old, so I removed them.

**Checklist**


jaxesn:
Probably too much for the context of this PR/ticket, but maybe worth looking into something like https://github.com/maxogden/dependency-check for this kind of stuff.

DanielMSchmidt:
@jaxesn I will give it a timeboxed try and see if I can get the same info out of it as my current solution 🤔

jaxesn:
Totally up to you. If it works out, might be something to add to the automated build process?

DanielMSchmidt:
@jaxesn gave it a few minutes, but I think our webpack aliases and ES6 / JSX Syntax makes it hard for the tool 🤷‍♂️

jaxesn:
makes sense. thanks for trying it!

DanielMSchmidt:
Thanks for suggesting 👍

weblancaster:
@DanielMSchmidt @jaxesn Keep in mind that we are planning to add https://renovateapp.com/ too.

juliangieseke:
@weblancaster yes, this was an effort to see if we can remove dependencies which are only used once and for very little tasks. was hoping to find a bit more tho 🤔 

rebased and resolved conflicts 👍

DanielMSchmidt:
@juliangieseke If you look into the ticket, there are a few libraries noted that we don't use that often, if you like you can add follow-up tickets for them ;)

PR #2830, opened 2018-03-15
DCOS-21440: keep VIP label

nLight:
This PR fixes the way the UI supports VIPs. 

UI used to be very opinionated about the way VIPs should be. It would require a VIP to be `VIP_<enpoint_index>` but that's not the way backend expects them. Backend allows anything that starts with `vip|VIP` to be a valid VIP label. Because of that we need to keep vip label that we're getting from JSON.

## Testing

Please create an app with a few Service Endpoints and override VIP_* to anything valid:
- vipBla
- VIPblop
- VIP_129312

Please check different network modes along with single and multi container.

NB! `Vip` `vIp` etc are invalid.

Closes DCOS-21440

**Checklist**

### DIFF
--- a/plugins/services/src/js/reducers/serviceForm/Container.js
+++ b/plugins/services/src/js/reducers/serviceForm/Container.js
@@ -151,7 +151,7 @@ const containerJSONReducer = combineReducers({
 
     // Convert portDefinitions to portMappings
     return this.portDefinitions.map((portDefinition, index) => {
-      const vipLabel = `VIP_${index}`;
+      const vipLabel = portDefinition.vipLabel || `VIP_${index}`;
       const vipPort = Number(portDefinition.vipPort) || null;
       const containerPort = Number(portDefinition.containerPort) || 0;
       const servicePort = parseInt(portDefinition.servicePort, 10) || null;
--- a/plugins/services/src/js/reducers/serviceForm/Container.js
+++ b/plugins/services/src/js/reducers/serviceForm/Container.js
@@ -151,7 +151,8 @@ const containerJSONReducer = combineReducers({
 
     // Convert portDefinitions to portMappings
     return this.portDefinitions.map((portDefinition, index) => {
-      const vipLabel = portDefinition.vipLabel || `VIP_${index}`;
+      const vipLabel =
+        portDefinition.vipLabel || VipLabelUtil.defaultVip(index);
       const vipPort = Number(portDefinition.vipPort) || null;
       const containerPort = Number(portDefinition.containerPort) || 0;
       const servicePort = parseInt(portDefinition.servicePort, 10) || null;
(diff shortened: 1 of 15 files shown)


plugins/services/src/js/reducers/serviceForm/PortMappings.js
```diff
@@ -119,19 +119,28 @@ module.exports = {
         });
       }
 
-      const vip = findNestedPropertyInObject(item, `labels.VIP_${index}`);
+      const vip = Object.entries(item.labels || {}).find(([key, _]) =>
```
DanielMSchmidt: I see this `.find(([key, _]) => key.match(/^(vip|VIP)/)` a third time and it is a bit cryptic to me, which is normal with regexes I guess. Can we move it into a util, I think there already is one for labels

jaxesn: any reason not to just use a case insensitive search? `key.match(/^vip/i)`

plugins/services/src/js/reducers/serviceForm/Container.js
```diff
@@ -151,7 +151,7 @@ const containerJSONReducer = combineReducers({
 
     // Convert portDefinitions to portMappings
     return this.portDefinitions.map((portDefinition, index) => {
-      const vipLabel = `VIP_${index}`;
+      const vipLabel = portDefinition.vipLabel || `VIP_${index}`;
```
jaxesn: Probably not super important, but the default label shows up a few places `VIP_${index}`, worth making that a constant/helper somewhere?

plugins/services/src/js/reducers/serviceForm/PortMappings.js
DanielMSchmidt: Yeah, I read somewhere that VIP and vip is okay, Vip or vIP not

jaxesn: makes sense, thanks

nLight: @DanielMSchmidt ✅

plugins/services/src/js/reducers/serviceForm/Container.js
nLight: It was a bug fix PR, I added that refactoring, please test thoroughly.

nLight:
@DanielMSchmidt please give it another shot!

nLight:
@jaxesn @DanielMSchmidt refactoring has been addressed, please retest.

nLight:
Hold merge until re-tested re-reviewed

juliangieseke:
hm, github hasnt dismissed the reviews? why didnt you manually dismiss them? 🤔

DanielMSchmidt:
@nLight re-reviewed it 👍

PR #2832, opened 2018-03-16
feat(mesos-stream): repeat the stream

nLight:
This PR adds `repeat()` operator that will re-connect to the stream in case Mesos closes the connection.

To test this I wrote a simple mock server and pointed MesosStream.js to the `http://localhost:3000`. Server yields messages with the minimum possible payload that RecordIO is able to parse. Without that fix UI would try connecting once. 

NB! In the Networking tab you will see two calls because one of them is OPTIONS

```js
const http = require('http')
const port = 3000

const requestHandler = (request, response) => {
  console.log("Connected")
  response.setHeader("Access-Control-Allow-Origin", "*")
  response.setHeader("Access-Control-Allow-Headers", "Origin, X-Requested-With, Content-Type, Accept")
  response.write("2\n{}")
  setTimeout(function() {
    response.write("2\n{}")
  }, 5000)
  setTimeout(function() {
    response.end("2\n{}")
  }, 10000)
}

const server = http.createServer(requestHandler)

server.listen(port, (err) => {
  if (err) {
    return console.log('something bad happened', err)
  }

  console.log(`server is listening on ${port}`)
})
```

Closes DCOS-21701

**Checklist**

### DIFF
--- a/src/js/core/MesosStream.js
+++ b/src/js/core/MesosStream.js
@@ -3,5 +3,6 @@ import { ReplaySubject } from "rxjs/ReplaySubject";
 
 export const MesosStreamType = Symbol("MesosStreamType");
 export default stream({ type: "SUBSCRIBE" }, "/mesos/api/v1?subscribe")
+  .repeat()
   .multicast(() => new ReplaySubject())
   .refCount();


src/js/core/MesosStream.js
```diff
@@ -3,5 +3,6 @@ import { ReplaySubject } from "rxjs/ReplaySubject";
 
 export const MesosStreamType = Symbol("MesosStreamType");
 export default stream({ type: "SUBSCRIBE" }, "/mesos/api/v1?subscribe")
+  .repeat()
```
DanielMSchmidt: We are not talking about [RxJS.Obersvable.repeat](https://github.com/Reactive-Extensions/RxJS/blob/master/src/core/perf/operators/repeat.js) here, right?

nLight: We're talking exactly about that :) When MesosStream closes the connection the observable completes and we need to repeat it

DanielMSchmidt: Okay, this makes sense, if no new data can be fetched because some service is down, just show the old state, do I get that right?

Does `stream` re-open / retry the HTTP Request or how does that part work?

nLight: It will just attempt to reconnect.
[As you can see](https://github.com/dcos-labs/http-service/blob/master/src/stream.js#L20) in the `http-client` if connection completes then observer completes which means it's done, no new data. This is where `repeat` kicks in, so it will eventually open a new connection.

I'd be happy guiding you through the `http-client` code

DanielMSchmidt: Oh, so repeat does not mean "send the same value as the last value" but "call the same observable again"? I understood the docs differently, but I am also new to RxJS

Fabs:
@nLight do you think it is worth sharing the mockserver and how to reproduce your test?

Fabs:
@nLight. Did the stub code was just added to the description, or it was always there? In any case, thanks (And let me grab some coffee :D)

PR #2835, opened 2018-03-19
fix(FrameworkConfigurationForm): improve schema errors

ahoskins:
This is the lowest hanging fruit on improving schema validation errors for Frameworks. This commit merely improves styling, and not the content of the error.

We have talked with cosmos and data services about this general issue, and in the future someone will work on a solution to improve the contents of this error message. Currently this disappointingly obscure message is delivered by the mozilla json-schema library. Data services would like to use these json-schema clauses like oneOf, allOf - but the challenge is mapping this onto a good UX for the UI and CLI. The improved messages in the future might come directly from a "validate" endpoint in cosmos. This would be ideal because then CLI can also get good errors.

Closes DCOS-19740

**It looks like this now**:

screen shot 2018-03-19 at 11 05 40 am
screen recording 2018-03-19 at 11 06 am

**Testing**:
1. Deploy kafka from catalog
2. Toggle "enabled" for SSL Authentication
3. See the error message appear
4. Toggling "enabled" for transport encryption (further down the form) will remove the error.

^The reason for the error is the [allOf json schema clause](https://spacetelescope.github.io/understanding-json-schema/reference/combining.html#allof).

**Checklist**

### DIFF
--- a/src/js/components/FrameworkConfigurationForm.js
+++ b/src/js/components/FrameworkConfigurationForm.js
@@ -8,6 +8,7 @@ import TabButton from "#SRC/js/components/TabButton";
 import TabButtonList from "#SRC/js/components/TabButtonList";
 import Tabs from "#SRC/js/components/Tabs";
 import Util from "#SRC/js/utils/Util";
+import ErrorsAlert from "#SRC/js/components/ErrorsAlert";
 import JSONEditor from "#SRC/js/components/JSONEditor";
 import FluidGeminiScrollbar from "#SRC/js/components/FluidGeminiScrollbar";
 import PageHeaderNavigationDropdown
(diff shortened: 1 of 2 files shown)


DanielMSchmidt:
Hey @ahoskins I think this is a huge approvement, great job 👍 
I think this error looks a bit strange with the `[object Object]` fragment, a bit like `JSON.stringify` was missing

bildschirmfoto 2018-03-21 um 13 40 12

ahoskins:
@DanielMSchmidt sadly this kind of error is just from the mozilla library we use to render json-schema forms. In the future, we will work on a solution improve this. Thanks for the reviews, guys. 👍

PR #2838, opened 2018-03-20
DCOS-21768: Improve Integration Test Script (prepares backport to 1.11)

juliangieseke:
Jenkins complained about `**` after its restart, this fixes the issue by replacing the globbing by `find`

also we moved `junit-merge` from the dockerfile to our package.json because it is part of the project - we want to make the `run-integration-tests` script our entrypoint also for local testing in a follow-up PR 👍 

---

I have decided to put a bit more effort into this to allow us to backport this ASAP to 1.11 👍 After we backported this we should be able to drasticly reduce `run integration tests` calls on 1.11 branch 🎉 

Also we are able to run single tests in an easy way with a command like this `npm run integration-tests -- Sidebar` 🎉

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -33,7 +33,7 @@ RUN set -x \
   /var/lib/dpkg/info/ca-certificates-java.postinst configure \
   # Install npm dependencies
   && cd /dcos-ui \
-  && npm install -g junit-merge@1.3.0 git://github.com/johntron/http-server.git#proxy-secure-flag \
+  && npm install -g git://github.com/johntron/http-server.git#proxy-secure-flag \
   # Install dcos-launch
   && curl 'https://downloads.dcos.io/dcos-launch/bin/linux/dcos-launch' > /usr/local/bin/dcos-launch \
   && chmod +x /usr/local/bin/dcos-launch \
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,7 +31,7 @@ RUN set -x \
   && rm -rf /var/lib/apt/lists/* \
   # Post-install java certificates
   /var/lib/dpkg/info/ca-certificates-java.postinst configure \
-  # Install npm dependencies
+  # Install npm dependencies (System Tests!)
   && cd /dcos-ui \
   && npm install -g git://github.com/johntron/http-server.git#proxy-secure-flag \
   # Install dcos-launch
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -5,18 +5,18 @@
   "requires": true,
   "dependencies": {
     "@babel/code-frame": {
-      "version": "7.0.0-beta.44",
-      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.0.0-beta.44.tgz",
-      "integrity": "sha512-cuAuTTIQ9RqcFRJ/Y8PvTh+paepNcaGxwQwjIDRWPXmzzyAeCO4KqS9ikMvq0MCbRk6GlYKwfzStrcP3/jSL8g==",
+      "version": "7.0.0-beta.42",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.0.0-beta.42.tgz",
+      "integrity": "sha512-L8i94FLSyaLQpRfDo/qqSm8Ndb44zMtXParXo0MebJICG1zoCCL4+GkzUOlB4BNTRSXXQdb3feam/qw7bKPipQ==",
       "dev": true,
       "requires": {
-        "@babel/highlight": "7.0.0-beta.44"
+        "@babel/highlight": "7.0.0-beta.42"
       }
     },
     "@babel/highlight": {
-      "version": "7.0.0-beta.44",
-      "resolved": "https://registry.npmjs.org/@babel/highlight/-/highlight-7.0.0-beta.44.tgz",
-      "integrity": "sha512-Il19yJvy7vMFm8AVAh6OZzaFoAd0hbkeMZiX3P5HGD+z7dyI7RzndHB0dg6Urh/VAFfHtpOIzDUSxmY6coyZWQ==",
+      "version": "7.0.0-beta.42",
+      "resolved": "https://registry.npmjs.org/@babel/highlight/-/highlight-7.0.0-beta.42.tgz",
+      "integrity": "sha512-X3Ur/A/lIbbP8W0pmwgqtDXIxhQmxPaiwY9SKP7kF9wvZfjZRwMvbJE92ozUhF3UDK3DCKaV7oGqmI1rP/zqWA==",
       "dev": true,
       "requires": {
         "chalk": "2.3.2",
@@ -89,7 +89,7 @@
           "dev": true,
           "requires": {
             "babel-runtime": "6.26.0",
-            "core-js": "2.5.4",
+            "core-js": "2.5.3",
             "regenerator-runtime": "0.10.5"
           }
         },
@@ -99,7 +99,7 @@
           "integrity": "sha1-llxwWGaOgrVde/4E/yM3vItWR/4=",
           "dev": true,
           "requires": {
-            "core-js": "2.5.4",
+            "core-js": "2.5.3",
             "regenerator-runtime": "0.11.1"
           },
           "dependencies": {
@@ -167,7 +167,7 @@
           "integrity": "sha1-llxwWGaOgrVde/4E/yM3vItWR/4=",
           "dev": true,
           "requires": {
-            "core-js": "2.5.4",
+            "core-js": "2.5.3",
             "regenerator-runtime": "0.11.1"
           }
         },
@@ -272,7 +272,7 @@
       "dev": true,
       "requires": {
         "conventional-changelog-angular": "1.6.6",
-        "conventional-commits-parser": "2.1.7"
+        "conventional-commits-parser": "2.1.6"
       }
     },
     "@commitlint/read": {
@@ -284,7 +284,7 @@
         "@commitlint/top-level": "6.1.3",
         "@marionebl/sander": "0.6.1",
         "babel-runtime": "6.23.0",
-        "git-raw-commits": "1.3.6"
+        "git-raw-commits": "1.3.5"
       }
     },
     "@commitlint/resolve-extends": {
@@ -307,7 +307,7 @@
           "integrity": "sha1-llxwWGaOgrVde/4E/yM3vItWR/4=",
           "dev": true,
           "requires": {
-            "core-js": "2.5.4",
+            "core-js": "2.5.3",
             "regenerator-runtime": "0.11.1"
           }
         },
@@ -361,7 +361,7 @@
     "@cypress/xvfb": {
       "version": "1.1.3",
       "resolved": "https://registry.npmjs.org/@cypress/xvfb/-/xvfb-1.1.3.tgz",
-      "integrity": "sha1-YpSn0f63UfEjAiSPIIn8U0xKy38=",
+      "integrity": "sha512-EfRzw+wgI0Zdb4ZlhSvjh3q7I+oenqEYPXvr7oH/2RnzQqGDrPr7IU1Pi2yzGwoXmkNUQbo6qvntnItvQj0F4Q==",
       "dev": true,
       "requires": {
         "lodash.once": "4.1.1"
@@ -419,7 +419,7 @@
     "@dcos/recordio": {
       "version": "0.1.6",
       "resolved": "https://registry.npmjs.org/@dcos/recordio/-/recordio-0.1.6.tgz",
-      "integrity": "sha1-wrwEXvPUjClifgCKKzJukO1v27Q=",
+      "integrity": "sha512-ej0oXW9T3Hpm25JY3ifuWalXGjASPQ3u6z4GokHwH5J5chziae8BkXtgisDsQFgz5C0CUG1wWjU+xGP9N8lD7w==",
       "requires": {
         "@dcos/copychars": "0.1.0"
       }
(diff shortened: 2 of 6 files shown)


scripts/ci/run-integration-tests
```diff
@@ -5,8 +5,8 @@
 # replace it with a test runner in the future, please see it more as a proof of
 # concept
 
-set -e
-shopt -s globstar # to allow ** in globs
+set -ex
+# shopt -s globstar # to allow ** in globs
```
DanielMSchmidt: Can we remove this please?

npm-shrinkwrap.json
```diff
@@ -7866,6 +7866,910 @@
       "integrity": "sha1-FQStJSMVjKpA20onh8sBQRmU6k8=",
       "dev": true
     },
+    "fsevents": {
```
DanielMSchmidt: fsevent should not be in here

juliangieseke: Jenkins knows that this is optional and not supported, it will be skipped 👍

juliangieseke:
```
npm WARN optional SKIPPING OPTIONAL DEPENDENCY: fsevents@1.1.3 (node_modules/fsevents):

npm WARN notsup SKIPPING OPTIONAL DEPENDENCY: Unsupported platform for fsevents@1.1.3: wanted {"os":"darwin","arch":"any"} (current: {"os":"linux","arch":"x64"})
```
https://jenkins.mesosphere.com/service/jenkins/blue/organizations/jenkins/frontend%2Fdcos-ui-oss-pipeline/detail/PR-2838/2/pipeline/46#step-51-log-149

DanielMSchmidt: Still it should not be added in this PR

npm-shrinkwrap.json
```diff
@@ -7740,132 +8072,1020 @@
...
+      "integrity": "sha1-4y/AMKLM7kSmtTcTCNpUvgs5fSc=",
+      "dev": true
+    },
+    "fs.realpath": {
+      "version": "1.0.0",
+      "resolved": "https://registry.npmjs.org/fs.realpath/-/fs.realpath-1.0.0.tgz",
+      "integrity": "sha1-FQStJSMVjKpA20onh8sBQRmU6k8=",
+      "dev": true
+    },
+    "fsevents": {
```
Fabs: fsevents still here, did we forgot to run the build-shrinkwrap? I know it is ignored now, but if we decided to add it I dont think that should be the PR to do it.

DanielMSchmidt:
@juliangieseke Could you fix the fsevents & the merge conflict in the npm-shrinkwrap?

scripts/ci/run-integration-tests
```diff
@@ -6,48 +6,59 @@
...
+  kill $SERVER_PID
+  echo 'Merging results'
+  junit-merge ./cypress/*/result.xml --out ./cypress/results.xml
+  echo 'Done'
+}
+
+# Starting http-server & setting up trap
+http-server -sp 4200 dist&
+SERVER_PID=$!
+trap "cleanup $SERVER_PID" EXIT
```
DanielMSchmidt: Neat 👍

juliangieseke: nice, you approved the old version but your review wasnt dismissed when I force-pushed :D

juliangieseke: now its dismissed 🤔 :(

Dockerfile
```diff
@@ -31,9 +31,9 @@ RUN set -x \
   && rm -rf /var/lib/apt/lists/* \
   # Post-install java certificates
   /var/lib/dpkg/info/ca-certificates-java.postinst configure \
-  # Install npm dependencies
+  # Install npm dependencies (System Tests!)
   && cd /dcos-ui \
-  && npm install -g junit-merge@1.3.0 git://github.com/johntron/http-server.git#proxy-secure-flag \
+  && npm install -g git://github.com/johntron/http-server.git#proxy-secure-flag \
```
DanielMSchmidt: Do we still need this if it's in the package.json?

juliangieseke:
hold merge because the changed file is accessed from external plugins, we need to copy the old one first :(

scripts/ci/run-integration-tests
```diff
@@ -6,48 +6,59 @@
 # concept
 
 set -e
-shopt -s globstar # to allow ** in globs
 
 PROJECT_ROOT="$( cd "$( dirname "${BASH_SOURCE[0]}" )/../../" && pwd )"
 RETRIES=3
 
-function executeCypress {
```
DanielMSchmidt: I think having a function here was more readable, but feel free to overrule me :)

Dockerfile
juliangieseke: Yeah, system tests rely on it :(

scripts/ci/run-integration-tests
juliangieseke: hm, yeah… i can revert this if you like 👍

juliangieseke:
this needs one more rebase & some clarification next week…

nLight:
What am I doing wrong?

screen shot 2018-04-03 at 14 00 46

juliangieseke:
not sure 🤔 will have a look

scripts/ci/run-integration-tests
```diff
@@ -5,49 +5,183 @@
...
+  local RUN=0
+  local EXIT_CODE=1
+  while [ $EXIT_CODE -ne 0 ] && [ $RUN -lt $RETRIES ]; do
+    RUN=$((RUN+1))
+    local OUTPUT_PATH="${PROJECT_ROOT}/${CYPRESS_FOLDER}/${RELATIVE_PATH}-${RUN}"
 
-    executeCypress "$1"
-    exit_code=$?
+    echo "===> Executing Cypress (${RUN}/${RETRIES})"
+    CYOUT="$(FILE="${PROJECT_ROOT}/tests/${RELATIVE_PATH}" RUN=${RUN} OUTPUT_PATH=${OUTPUT_PATH} executeCypress)"
```
DanielMSchmidt: I am not sure if I like this, because even If I would not supress output I would not know about the last output when my test hangs for some reason. I would rather drop the feature of silencing then having this drawback, but this is only my point of view

scripts/ci/run-integration-tests
```diff
@@ -5,49 +5,183 @@
...
   done
 
-  echo "==> Couldn't get a success in $RETRIES retries"
-  exit 1
+  if [ $EXIT_CODE -ne 0 ]; then
+    echo "===> Couldn't get a success in $RETRIES retries for $RELATIVE_PATH"
+  fi
+
+  if [ ${EXIT_CODE} -ne 0 ] && [ ${STOP_ON_FAIL} -eq 0 ]; then
+    echo "==> Continuing despite failures…"
```
DanielMSchmidt: Why do we want to use the special character "…" instead of "..."?

scripts/ci/run-integration-tests
```diff
@@ -5,49 +5,183 @@
...
+    return 0
+  fi
+
+  return $EXIT_CODE
+}
+
+# copies tests from plugins folder into project
+# this is necessary because Cypress doesnt allows external files to be tested…
+function setup_plugins {
```
DanielMSchmidt: Neat 👍

scripts/ci/run-integration-tests
```diff
@@ -5,49 +5,183 @@
 # replace it with a test runner in the future, please see it more as a proof of
 # concept
 
-set -e
```
DanielMSchmidt: We don't want to exit on error? I understand this for the test execution part, but if something else fails we want to error out, right?

scripts/ci/run-integration-tests
```diff
@@ -5,49 +5,183 @@
...
+  echo "==> Killing Proxy…"
+  kill $PROXY_PID
+  echo "===> done."
+}
+
+# sets up everything
+function setup {
+  echo ""
+  echo "=> Setup"
+
```
DanielMSchmidt: I think we should check if the dist folder is not empty to make it really solid

scripts/ci/run-integration-tests
```diff
@@ -5,49 +5,183 @@
...
+}
+
+# copies tests from plugins folder into project
+# this is necessary because Cypress doesnt allows external files to be tested…
+function setup_plugins {
+  if [ -n "${PLUGINS_PATH}" ]; then
+    echo "==> Setup external plugins…"
+    rm -rf ${PROJECT_ROOT}/${TESTS_FOLDER}/${PLUGINS_FOLDER}/
+    mkdir ${PROJECT_ROOT}/${TESTS_FOLDER}/${PLUGINS_FOLDER}/
+    rsync -aH ${PROJECT_ROOT}/${PLUGINS_PATH}/${TESTS_FOLDER}/ ${PROJECT_ROOT}/${TESTS_FOLDER}/${PLUGINS_FOLDER}/
```
DanielMSchmidt: I think this doesn't work, it assumes that the tests of a plugin know where they get copied to, which seems pretty fragile to me 🤔 Either we verify / document this assumption or we need to be more hacky with the path so that it works

scripts/ci/run-integration-tests
DanielMSchmidt: I am just unsure if older editors support that 🤔

scripts/ci/run-integration-tests
juliangieseke: sadly it doesnt output anything helpful to stdout :(

scripts/ci/run-integration-tests
DanielMSchmidt: For me the plus is 

- I can not forget to do a `|| exit 1` at the end of each statement
- Writing `|| exit 1` at the end of each statement feels repetitive
- If some command fails in an unexpected way (e.g. rsync not being installed on some machine) the script can go rouge without me noticing at first, because it just runs over it. Best case I need to search for the error line and fix it, worst case it screws up my system because some assumptions fail and weird stuff happens
- I almost never want to have a command in my script just fail randomly (except for cypress here). For me, it is the same as writing `try{}catch(e){}` around every statement of my JS code.

scripts/ci/run-integration-tests
DanielMSchmidt: The subfolder isn't the problem, see PM

scripts/ci/run-integration-tests
juliangieseke: ok, added 👍

nLight:
We could detect that there's no `/dist` folder and run build before running tests :)

juliangieseke:
hm, yes, we _could_ do it, but im not sure if i want to call `npm run build` there 🤔 
also I'm sure that I dont want to call `npm run build-assets` 😅 
thats the reason why i left it to the user 🤷‍♂️

juliangieseke:
have to change this slightly to fix plugins :(

juliangieseke:
had to change this PR a bit in order to make Plugins tests green, just putting them in a subfolder didnt work because too many things are kinda hardcoded in cypress (loading fixtures for example). the current solution is "okay-ish" but we have to be careful here and there, will update the plugins documentation tomorrow.

scripts/ci/run-integration-tests
```diff
@@ -43,8 +43,17 @@ function runTestFile {
...
+# Starting http-server & setting up trap
+http-server -sp 4200 dist&
+SERVER_PID=$!
+trap "kill $SERVER_PID" EXIT
+
+# pass in simple search strings for filenames here (find syntax)
+SEARCH=${1:-'*'}
+
+echo "Found these matching files ('./tests/${SEARCH}-cy.js')"
+find $PROJECT_ROOT/tests -type f -name ${SEARCH}"-cy.js" -print | tee /dev/tty | while read file; do runTestFile "$file"; done
```
DanielMSchmidt: What does `tee /dev/tty` do?

scripts/ci/run-integration-tests
```diff
@@ -5,55 +5,183 @@
...
+PLUGINS_PATH=${PLUGINS_PATH:-''}
+
 
+
+## Functions
+#####################################################################
+
+# executes Cypress
+#
+# from callee:
```
DanielMSchmidt: I like the comments 👍

scripts/ci/run-integration-tests
```diff
@@ -5,55 +5,183 @@
 # replace it with a test runner in the future, please see it more as a proof of
 # concept
 
-set -ex
+## Configuration
```
DanielMSchmidt: I like the way you wrote the script and the effort you put into it. I really hope we can someday move this in the JS world to test it

scripts/ci/run-integration-tests
```diff
@@ -140,8 +140,12 @@ function setup {
     echo "==> ERROR: Port ${PROXY_PORT} is already in use"
     exit 1
   fi
-  echo "==> Starting http-server…"
-  http-server -sp ${PROXY_PORT} dist&
+  if [ ! -d ${PROJECT_ROOT}/dist ]; then
+    echo "==> ERROR: no dist folder! forgot npm run build?"
+    exit 1
```
DanielMSchmidt: I like the decision to not run `npm build` from here. Yeah, could me more convenient, but it could also mask some setup errors we wouldn't spot then

scripts/ci/run-integration-tests
```diff
@@ -110,9 +109,18 @@ function setup_plugins {
 # removes copied plugins files…
 function teardown_plugins {
   if [ -n "${PLUGINS_PATH}" ]; then
-    echo "==> Removing plugins folder…"
-    rm -r "${PROJECT_ROOT}/${TESTS_FOLDER}/${PLUGINS_FOLDER}/"
-    echo "===> done."
+    if [ ${CLEANUP_TESTS} -eq 0 ]; then
+      echo "==> Cleaning up ./${TESTS_FOLDER} folder"
+      rm -r ${PROJECT_ROOT}/${TESTS_FOLDER}
+      git checkout ${PROJECT_ROOT}/${TESTS_FOLDER}
```
DanielMSchmidt: We rely on git anyways, I think what you really want to do here is a `git clean -dfx "${PROJECT_ROOT}/${TESTS_FOLDER}"` instead of rm and git checkout.

DanielMSchmidt:
@juliangieseke Sorry that the comments are in outdated files, I went through these changes commit be commit :)

scripts/ci/run-integration-tests
juliangieseke: its a hack to be able to pipe (stdout, one output of `tee`) AND have it on the current terminal (which is - file wise - `/dev/tty`, the file `tee` writes to)

scripts/ci/run-integration-tests
juliangieseke: good point, will change this if i touch the script again 👍

scripts/ci/run-integration-tests
DanielMSchmidt: Thank you for explaining :)

juliangieseke:
hm, looks like the `http-server` is dead already 🤔 not exactly sure why, maybe your shell kills all sub processes?

juliangieseke:
but it shouldnt be an issue and I dont want to pray for green tests again :(

DanielMSchmidt:
That really shows the sad state of our test infrastructure 😢

scripts/run-integration-tests
```diff
@@ -0,0 +1,202 @@
...
+  echo ""
+  echo "=> Setup"
+
+  ## Setup proxy
+  if [ -n "$(lsof -i :${PROXY_PORT})" ]; then
+    echo "==> ERROR: Port ${PROXY_PORT} is already in use"
+    exit 1
+  fi
+  if [ ! -d ${PROJECT_ROOT}/dist ]; then
+    echo "==> ERROR: no dist folder! forgot npm run build?"
```
orlandohohmeier: 🔤 `No`

orlandohohmeier:
LGTM! 🚢

DanielMSchmidt:
@juliangieseke git checkout problem again, we need to force push I guess?

juliangieseke:
done. 😢

juliangieseke:
@orlandohohmeier @DanielMSchmidt 
had to force push, but just changed the commit message, no code changed, can you please reapprove? 🙏

PR #2840, opened 2018-03-20
DCOS-21640: Wrap cosmos repository API with http-service

nLight:
This PR refactors Settings -> Package Repositories page APIs wrapping them with `http-service`. This is a stepping stone in the data-layer effort.

So listing/deleting/adding repository should just work. 

~Beware that verbose error messages will be broken! Therefore `hold merge`
There's an [issue](https://github.com/dcos-labs/http-service/issues/4) for that and a [JIRA](https://jira.mesosphere.com/browse/DCOS-21727)~

Closes DCOS-21640

**Checklist**

### DIFF
--- a/plugins/catalog/src/js/repositories/repositoriesStream.js
+++ b/plugins/catalog/src/js/repositories/repositoriesStream.js
@@ -1,0 +1,32 @@
+import { request } from "@dcos/http-service";
+import Config from "#SRC/js/config/Config";
+
+export const fetchRepositories = type =>
+  request(`${Config.rootUrl}${Config.cosmosAPIPrefix}/repository/list`, {
+    method: "POST",
+    body: JSON.stringify({ type }),
+    headers: {
+      Accept: "application/vnd.dcos.package.repository.list-response+json;charset=utf-8;version=v1",
+      "Content-Type": "application/vnd.dcos.package.repository.list-request+json;charset=utf-8;version=v1"
+    }
+  });
+
+export const addRepository = (name, uri, index) =>
+  request(`${Config.rootUrl}${Config.cosmosAPIPrefix}/repository/add`, {
+    method: "POST",
+    body: JSON.stringify({ name, uri, index }),
+    headers: {
+      Accept: "application/vnd.dcos.package.repository.add-response+json;charset=utf-8;version=v1",
+      "Content-Type": "application/vnd.dcos.package.repository.add-request+json;charset=utf-8;version=v1"
+    }
+  });
+
+export const deleteRepository = (name, uri) =>
+  request(`${Config.rootUrl}${Config.cosmosAPIPrefix}/repository/delete`, {
+    method: "POST",
+    body: JSON.stringify({ name, uri }),
+    headers: {
+      Accept: "application/vnd.dcos.package.repository.delete-response+json;charset=utf-8;version=v1",
+      "Content-Type": "application/vnd.dcos.package.repository.delete-request+json;charset=utf-8;version=v1"
+    }
+  });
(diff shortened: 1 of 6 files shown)


plugins/catalog/src/js/repositories/repositoriesStream.js
```diff
@@ -0,0 +1,32 @@
+import { request } from "@dcos/http-service";
+import Config from "#SRC/js/config/Config";
+
+export const fetchRepositories = type =>
+  request(`${Config.rootUrl}${Config.cosmosAPIPrefix}/repository/list`, {
+    method: "POST",
+    body: JSON.stringify({ type }),
```
jaxesn: What do you think about making the request function support object bodies and automatically stringify?

plugins/catalog/src/js/repositories/repositoriesStream.js
```diff
@@ -0,0 +1,32 @@
+import { request } from "@dcos/http-service";
```
jaxesn: I like the pattern of moving the request logic to own "libraries", is the `plugins/catalog` folder the place for this kind of thing going forward, or is there something about repositories specifically that dictates this?

plugins/catalog/src/js/repositories/repositoriesStream.js
nLight: The idea behind http-service was that it is rather dumb. Having `request` stringifying payloads is a bit more magic

jaxesn: totally fair. I was just thinking the most likely case is that we are passing a json body that needs to be stringified, having to do this as the consumer of the api every time seemed less than ideal.  I agree too much magic is no good, *if* we were going to do something, maybe adding a `json` property that tells request to stringify the object, would be a compromise, but not saying its necessary.

plugins/catalog/src/js/repositories/repositoriesStream.js
nLight: We have a plugin SDK that loads stuff from plugins. This particular code doesn't implement a plugin interface, but is in plugins directory so that it is separated once we get to modularization and actually start pulling plugins out of the repo.

src/js/events/CosmosPackagesActions.js
```diff
@@ -382,68 +384,34 @@ const CosmosPackagesActions = {
...
+      function success(response) {
         AppDispatcher.handleServerAction({
           type: REQUEST_COSMOS_REPOSITORY_ADD_SUCCESS,
           data: response,
           name,
           uri
         });
       },
-      error(xhr) {
+      function error({ xhr = {} }) {
```
Fabs: I still think that with us removing the error messages this PR should be against `features/data-layer` and not master. The reason being when it touches master we are removing functionality because we will lose the error messages that the http-service does not provide.

by the way, do we have a ticket for that?

nLight: https://jira.mesosphere.com/browse/DCOS-21727

nLight: Since this PR has a legit fail of integration tests I'm inclining following your advice here and change the base.

Fabs: 👍

nLight:
Uh-oh, someone needs to rebase the target branch

Fabs:
@nLight think we are good now.

nLight:
it's green \o/ ✅

juliangieseke:
cant we just merge this into master? 🤔

Fabs:
@juliangieseke. Yes. It backports the store to use the streams, and expose them so I can use it with graphql on the work I am doing now. So it is safe to merge because it is backward compatible.

nLight:
@juliangieseke that was the original idea and since I fixed the error messages it could go into master.

nLight:
@juliangieseke it's targeting master now

DanielMSchmidt:
Why do we merge this with 1 approval? 😕 I am fine because I just wanted to approve it, I am just wondering if we changed our policy there

nLight:
Uh-oh... I thought you've added one more 😓

PR #2841, opened 2018-03-20
DCOS-21752 DOCUMENTATION describes the mediator api

Fabs:
Presents an example of the mediator API and some documentation.

An implementation of the Settings > Package Repository page using this api is on the go here.

This documentation will be merged into `data-layer`. And once we are ok with all the pieces we will merge this integration in master

### DIFF
--- a/packages/data-service/README.md
+++ b/packages/data-service/README.md
@@ -1,0 +1,193 @@
+This package is to be extracted to dcos-ui/data-service.
+
+# API v1
+
+## Mediator Pattern
+
+The [mediator pattern](https://en.wikipedia.org/wiki/Mediator_pattern) is used here
+to decouple data handling from the view components by encapsulating the way they
+communicate. 
+
+With this pattern, a component will receive down all the pieces of data `(data down)` it needs via props,
+and send `events up` so the mediator can react to events. The key concept here is that both the 
+data provider and the view component have no idea how each other work.
+
+The caller of `componentFromStream` bellow is the mediator responsible for the exchange between
+the part of the application which handles data, and the UI components.
+
+```js
+  /*
+   * Receives a stream of React.Component and outputs a single component that 
+   * renders the most recent component emitted by the observer.
+  */
+  componentFromStream(component$: Observable<React.Component>): React.Component;
+```
+
+Note that the way we achieve the Mediator pattern is not in an OO fashion, but
+by abstracting all the communication between the data sources and the view components
+by using a stream.
+
+### Query Example
+
+This example uses a stream of `repository of packages` that are fetched via one or more
+APIs (abstracted here by a single stream). We use graphql to describe the query and streams
+to achieve the concrete implementation. We also use streams to pass event from the components
+in order to implement a search filter.
+
+This example is an extract from [DC/OS UI](https://github.com/dcos/dcos-ui/).
+
+```js
+import React from "react";
+import { Observable } from "rxjs";
+
+// Recompose helps handle data down and event up as streams
+import {
+  createEventHandler,
+  componentFromStream
+} from "data-service/recomposeRx";
+
+// Our in-house solution to resolve graphql queries using streams
+import { graphqlObservable } from "data-service/graphqlObservable";
+
+// tools to make easier to create schemas and queries
+import { makeExecutableSchema } from "graphql-tools";
+import gql from "graphql-tag";
+
+// Streams we get our data from
+// That could be already a more complex stream that actually merge
+// different api's endpoints, sockets with REST and etc.
+import { fetchRepositories } from "#PLUGINS/catalog/src/js/repositories/repositoriesStream";
+
+// The graphql schema and resolvers for those streams;
+// Normally you would import it from another file, but we inlined them here
+// for a more complete example.
+// import { typeDefs, resolvers } from "#SRC/js/model/repositoryModel";
+export const typeDefs = `
+  type PackageRepository {
+    id: ID!
+    name: String!
+    uri: String!
+  }
+  
+  type Query {
+    packageRepository(filter: String): [PackageRepository!]!
+  }
+
+  type Mutation {
+    addPackageRepository(uri: String!, name: String!): PackageRepository
+  }
+`;
+
+export const resolvers = {
+  Query: {
+    packageRepository: (parent, args, context) => {
+      const { filter } = args;
+
+      // Filter Logic Backwards compatible with the previous struct/RepositoryList
+      return context.query.combineLatest(filter, (result, filter) =>
+        Object.values(
+          new RepositoryList({ items: result.repositories })
+            .filterItemsByText(filter)
+            .getItems()
+        )
+      );
+    }
+  },
+};
+
+// UI components
+import RepositoriesTabUI from "#SRC/js/components/RepositoriesTabUI";
+import RepositoryList from "#SRC/js/structs/RepositoryList";
+
+// Using the data layer
+
+// 1. We first make a schema out of the resolvers and typeDefinitions
+const schema = makeExecutableSchema({
+  typeDefs,
+  resolvers
+});
+
+// 2. We declare which data our component needs in graphql (data down)
+const packageRepositoryQuery = gql`
+  query {
+    packageRepository(filter: $filter) {
+      name
+      uri
+    }
+  }
+`;
+
+// 3. We create an event handler as a stream to deal with search (events up)
+const { handler: onSearch, stream: searchEvent$ } = createEventHandler();
+
+// Almost always you need to provide an initial value for two reasons:
+// First. it should be defined what is the first value of search,
+// even when it is null, you should startWith(null)
+// Second. if you don't emit a first value, the graphqlStream that depend
+// on this, to filter will actually emit no result until they have the first
+// value for searchTerm.
+const searchTerm$ = searchEvent$.startWith("");
+
+// 4. We compose the result with the query and massage it down the pipe until
+// we get the props for our component.
+// The idea is from an Observable of data and events, compose a stream of React.Components
+const components$ = graphqlObservable(packageRepositoryQuery, schema, {
+  query: fetchRepositories(),
+  filter: searchTerm$
+})
+  .combineLatest(searchTerm$, (result, searchTerm) => {
+    // Backwards compatible with the previous struct/RepositoryList for packages
+    return {
+      packageRepository: new RepositoryList({
+        items: result.packageRepository
+      }),
+      searchTerm
+    };
+  })
+  // We map over the data and return a component to render
+  .map(data => {
+    return (
+      <RepositoriesTabUI
+        repositories={data.packageRepository}
+        searchTerm={data.searchTerm}
+        onSearch={onSearch}
+      />
+    );
+  })
+  // The first component is the loading
+  .startWith(<RepositoriesTabUI.Loading />)
+  // If anything goes wrong on the stream, we render an error component
+  // Note. that does not catch errors from within RepositoriesTabUI
+  .catch(err => Observable.of(<RepositoriesTabUI.Error err={err} />));
+
+// componentFromStream create a single component who render the latest emitted
+// component from the stream
+const RepositoriesTab = componentFromStream(() => components$);
+
+// the reason we don`t `export componentFromStream()` is so you can add those
+// router specific things
+RepositoriesTab.routeConfig = {
+  label: "Package Repositories",
+  matches: /^\/settings\/repositories/
+};
+
+// We end by exporting a single React component that will change according to the stream
+module.exports = RepositoriesTab;
+```
+
+Notice how easy is to provide error handling whenever an exception happens and how 
+we could start the stream of components providing a loading component.
+
+To pass information from the components to the graphql, we use another capability 
+from `recompose`, the `createEventHandler()`.
+
+```js
+/*
+ * Creates a stream that will emit a the value evetytime you call handler(value) 
+*/
+const { handler: onSearch, stream: searchEvent$ } = createEventHandler();
+```
+
+### Adding and removing Packages
+
+TBD with graphql mutations. 


packages/data-service/README.md
```diff
@@ -0,0 +1,79 @@
...
+// Imagine we manage a launchpad for interplanetary ships
+import { shuttles$, launchShuttle } from "shuttles/streams";
+import { shuttleSchema } from "shuttles/schema";
+import { Space, LaunchPad, LoadingUI, ErrorMessageUI } from "ui/components"
+import { renderComponents } from "data-service"
+
+// The data composition bellow is using only streams, 
+// IF we had graphql, it could look like this
+/*
+  const query = "
```
DanielMSchmidt: Can you maybe extract the graphql example into a second example, it's kinda hard to comprehend this way. For me everything here is new and I need to sort out what happens, so thinking how it could be with graphql on thr side is too much for me 😇

Fabs:
@DanielMSchmidt the milestone for the data-layer is a consistent API + reference implementation + documentation. Those things need to come together in pieces first, and then we can have something for master. 

One way to see it as the data-service being a package, there is no PR to updated the version of the package on master, until there is a release of the package. The feature branch is that v0.1 release in progress. We don't wanna have to repos at this point for velocity reasons.

packages/data-service/README.md
```diff
@@ -0,0 +1,79 @@
+This package is to be extracted to dcos-ui/data-service.
+
+#API v1
```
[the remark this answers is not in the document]
Fabs: Thanks for that, my editor did not complained about it :/

packages/data-service/README.md
Fabs: I agree that that makes reading linearly harder. Will think about something. Thanks.

packages/data-service/README.md
```diff
@@ -0,0 +1,79 @@
...
+
+```js
+// Imagine we manage a launchpad for interplanetary ships
+import { shuttles$, launchShuttle } from "shuttles/streams";
+import { shuttleSchema } from "shuttles/schema";
+import { Space, LaunchPad, LoadingUI, ErrorMessageUI } from "ui/components"
+import { renderComponents } from "data-service"
+
+// The data composition bellow is using only streams, 
+// IF we had graphql, it could look like this
```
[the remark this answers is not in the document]
Fabs: I actually put the big If on purpose, but you are right :D

packages/data-service/README.md
DanielMSchmidt: a second complete example would be cool, I think that is the easiest to comprehend

packages/data-service/README.md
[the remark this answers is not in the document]
DanielMSchmidt: okay, thats fine

packages/data-service/README.md
Fabs: Ok

Fabs:
@DanielMSchmidt. I corrected the mistakes and added the second example. I also changed the renderComponents to the recompose version we are using. Please feel free to review again and challenge if you think it is not good or if you are still confused.

packages/data-service/README.md
```diff
@@ -0,0 +1,113 @@
...
+  }
+`;
+
+// This would be a graphql mutation
+const launch = (name) => {
+  const launchDate = new Date();
+  graphql(launchMutation, shuttleSchema).mutate({ name, launchDate });
+}
+
+const components$ = content$.map((data) =>
```
DanielMSchmidt: How does the data get passed into the graphql query? I mean $location needs be specified for the query to be resolvable, right?

packages/data-service/README.md
```diff
@@ -0,0 +1,113 @@
...
+const launchMutation = `
+  mutation {
+    launchShip(name: $name, launchDate: $launchDate) {
+      name
+      launchDate 
+    }
+  }
+`;
+
+// This would be a graphql mutation
```
DanielMSchmidt: Does this comment make sense here?

packages/data-service/README.md
Fabs: At this point of the writing that is not clear. I still have to figure out the details on how we are going to use graphl which will be the rest of the work I am doing. Notice that the location is hardcoded on the non-graphql example anyway. 

To make this example whole, I can remove `$location`.

packages/data-service/README.md
Fabs: Not anymore, no.

packages/data-service/README.md
Fabs: I solved it in a different way, by adding an idea about how the parameters could be.

Fabs:
@DanielMSchmidt is there still any doubt, or any comment to address here?

packages/data-service/README.md
```diff
@@ -0,0 +1,112 @@
...
+const nextLaunches$ = shuttles$.filter((sht) => sht.location === "earth");
+const content$ = launched$.combineLatest(nextLaunches$,
+                          (launched, nextLaunches) => {launched, nextLaunches});
+
+//See also bellow about graphql mutations
+const launch = (name) => {
+  launchShuttle({name, launchDate: new Date()})
+}
+
+const components$ = content$.map((data) =>
```
juliangieseke: Is there any possibility to wrap this in a JSX-Component?

packages/data-service/README.md
```diff
@@ -0,0 +1,112 @@
...
+    launchShip(name: $name, launchDate: $launchDate) {
+      name
+      launchDate 
+    }
+  }
+`;
+
+const launch = (name) => {
+  const launchDate = new Date();
+  graphql(launchMutation, shuttleSchema).mutate({ name, launchDate });
```
juliangieseke: shouldnt even `.mutate` return something "to work with" in order to be able to notify the component about changed data from somewhere else? how to we match this to a stream we're listening to?

packages/data-service/README.md
```diff
@@ -0,0 +1,112 @@
...
+}
+
+const components$ = content$.map((data) => 
+                                <World>
+                                  <LaunchPad shuttles={data.nextLaunches} />
+                                  <Space shuttles={data.launched} />
+                                  <LaunchButton onLaunch={launch} />
+                                </World>)
+                            .startWith(<LoadingUI />)
```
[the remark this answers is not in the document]
Fabs: The pattern is to convert a stream of data into a stream of components. If there is one component handling loading and data, then the `startWith` move to the data part, and yields some information that the query started as first value `{isLoading: true}` for instance.

Same for the error, if there is a single component handling it, the catch move to the data part, and emits data when error happens `{isError: true, message: "blabla"`. 

I prefer the way I exemplified, but it is a question of choice.

packages/data-service/README.md
```diff
@@ -0,0 +1,112 @@
...
+}
+
+const components$ = content$.map((data) => 
+                                <World>
+                                  <LaunchPad shuttles={data.nextLaunches} />
+                                  <Space shuttles={data.launched} />
+                                  <LaunchButton onLaunch={launch} />
+                                </World>)
+                            .startWith(<LoadingUI />)                                
+                            .catch((err) => <ErrorMessageUI err>));
```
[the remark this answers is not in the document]
Fabs: see above :D

packages/data-service/README.md
```diff
@@ -0,0 +1,112 @@
...
+import { shuttles$, launchShuttle } from "shuttles/streams";
+import { Space, LaunchPad, LoadingUI, ErrorMessageUI } from "ui/components"
+import { renderComponents } from "data-service"
+
+
+// Merging this streams should be done via graphql in the future
+// See bellow for the equivalent graphql.
+const launched$ = shuttles$.filter((sht) => sht.location === "space");
+const nextLaunches$ = shuttles$.filter((sht) => sht.location === "earth");
+const content$ = launched$.combineLatest(nextLaunches$,
```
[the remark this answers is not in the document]
Fabs: I am replacing to a better example.

packages/data-service/README.md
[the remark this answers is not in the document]
Fabs: You are right. The example is just not very good.

packages/data-service/README.md
Fabs: I will stop providing an api for mutations as I have no idea how it is going to work for now.

Fabs:
@juliangieseke I had a session with julian and I made some comments on his very valid questions.

I provided also a better working example now :D. And at this point in time removed everything about mutations. We will have then once I work on the api (this week yet). But they are are scoped also in a different ticket.

Fabs:
Can you look again @juliangieseke and @DanielMSchmidt.

Fabs:
Also, this can't fail right? :D

juliangieseke:
Did you rebase this lately? Its an integration test which failed three times 🤷‍♂️

Fabs:
it is just a README file, CI is broken, I would be in favor to get a second approve and merge even with CI failing.

packages/data-service/README.md
```diff
@@ -0,0 +1,147 @@
...
+
+With this pattern, a component will receive down all the data `(data down)` it needs via props,
+and send `events up` so the mediator can perform mutations (represented here as a simple callback). 
+The key concept here is that both the data provider and the view component have no idea how each other work.
+Only the caller of `renderComponents` understands how it works.  
+
+```js
+  /*
+   * Receives a stream of React.Component and outputs a single components that 
+   * renders the most recevent event from the Observable.
```
nLight: 🔤 the most recent

packages/data-service/README.md
```diff
@@ -0,0 +1,147 @@
...
+
+```js
+import React from "react";
+import { Observable } from "rxjs";
+
+// Recompose helps handle data down and event up as streams
+import {
+  createEventHandler,
+  componentFromStream
+} from "#PACKAGES/data-service/recomposeRx";
```
nLight: should be possible just `import * from "data-service/recomposeRx";` We had this alias in webpack so that `./packages` act as `node_modules`

packages/data-service/README.md
```diff
@@ -0,0 +1,147 @@
...
+
+// tools to make easier to create schemas and queries
+import { makeExecutableSchema } from "graphql-tools";
+import gql from "graphql-tag";
+
+// Streams we get our data from
+import { fetchRepositories } from "#PLUGINS/catalog/src/js/repositories/repositoriesStream";
+
+// The graphql schema and resolvers for those streams;
+import { typeDefs, resolvers } from "#SRC/js/model/repositoryModel";
```
nLight: I know it's an example but this belong to `"#PLUGINS/catalog..."`

packages/data-service/README.md
```diff
@@ -0,0 +1,147 @@
...
+const searchTerm$ = searchEvent$.startWith("");
+
+// 4. We compose the result with the query, and massage it down the pipe until
+// we get the props for our component.
+// The idea is from an Observable of data and events, compose a stream o React.Components
+const components$ = graphqlObservable(packageRepositoryQuery, schema, {
+  query: fetchRepositories(),
+  filter: searchTerm$
+})
+  .combineLatest(searchTerm$, (result, searchTerm) => {
```
nLight: Why do we need passing `searchTerm$` twice?

packages/data-service/README.md
```diff
@@ -0,0 +1,147 @@
...
+    return (
+      <RepositoriesTabUI
+        repositories={data.packageRepository}
+        searchTerm={data.searchTerm}
+        onSearch={onSearch}
+      />
+    );
+  })
+  // The first component is the loading
+  .startWith(<RepositoriesTabUI.Loading />)
```
nLight: Name is confusing 🤔

packages/data-service/README.md
```diff
@@ -0,0 +1,147 @@
...
+  // If anything goes wrong, we render an error component
+  .catch(err => Observable.of(<RepositoriesTabUI.Error err={err} />));
```
juliangieseke: again, short info what is catched here? (not errors from `<RepositoriesTabUI>` but stuff from the stream)

packages/data-service/README.md
```diff
@@ -0,0 +1,147 @@
...
+      <RepositoriesTabUI
+        repositories={data.packageRepository}
+        searchTerm={data.searchTerm}
+        onSearch={onSearch}
+      />
+    );
+  })
+  // The first component is the loading
+  .startWith(<RepositoriesTabUI.Loading />)
+  // If anything goes wrong, we render an error component
```
juliangieseke: see note above 👍

packages/data-service/README.md
```diff
@@ -0,0 +1,147 @@
...
+```js
+/*
+ * Creates a stream that will emit a the value evetytime you call handler(value) 
+*/
+const { handler: onSearch, stream: searchEvent$ } = createEventHandler();
+```
+
+### Mutations
+
+TBD
```
juliangieseke: Ah ok 👍

packages/data-service/README.md
juliangieseke: I actually like being explicit here 🤔

packages/data-service/README.md
Fabs: ✅

packages/data-service/README.md
Fabs: #PACKAGES would be dcos-ui specific, and this package is going to be called data-service in the future, and move to dcos-labs, so actually "#PACKAGES" only make change if you know dcos-ui.

packages/data-service/README.md
Fabs: You are right. On the actual implementation, I moved that on Friday. think that does not have any impact on the example.

packages/data-service/README.md
Fabs: The first is passed as an argument to the query so it is used as a filter by graphql. You are filtering the PackageRepository mode, which has the fileds (name, priority and uri). But you also need to show the search term on the UI. Since the searchTerm is not a coming from the "data", but from the stream on `createEventHandler`, you have to combine it again.

The semantic is, combine the data stream with the events from my components and pass it down as props.

packages/data-service/README.md
```diff
@@ -0,0 +1,147 @@
...
+// The graphql schema and resolvers for those streams;
+import { typeDefs, resolvers } from "#SRC/js/model/repositoryModel";
+
+// UI components
+import RepositoriesTabUI from "#SRC/js/components/RepositoriesTabUI";
+import RepositoryList from "#SRC/js/structs/RepositoryList";
+
+// Using the data layer
+
+// 1. We first make an schema out of the resolvers and typeDefinitions
```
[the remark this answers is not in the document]
Fabs: Agree.

packages/data-service/README.md
```diff
@@ -0,0 +1,147 @@
...
+const schema = makeExecutableSchema({
+  typeDefs,
+  resolvers
+});
+
+// 2. We declare which data our component needs in graphql (data down)
```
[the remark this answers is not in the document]
Fabs: It is assumed you know graphql and rxjs. Otherwise I would be written a bunch of stuff the internet does much better. I added the schema and the resolvers as asked on the previous comment, that should be enough to understand given those assumptions.

packages/data-service/README.md
```diff
@@ -0,0 +1,147 @@
...
+    packageRepository(filter: $filter) {
+      name
+      uri
+    }
+  }
+`;
+
+// 3. We create an event handler as a stream to deal with search (events up)
+const { handler: onSearch, stream: searchEvent$ } = createEventHandler();
+const searchTerm$ = searchEvent$.startWith("");
```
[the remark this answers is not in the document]
Fabs: done.

packages/data-service/README.md
```diff
@@ -0,0 +1,147 @@
+This package is to be extracted to dcos-ui/data-service.
+
+# API v1
+
+## Mediator Pattern
+
+The [mediator pattern](https://en.wikipedia.org/wiki/Mediator_pattern) is used here
```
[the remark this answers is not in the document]
Fabs: I believe the pattern is an idea, not the implementation. It can manifest in many ways. We use the idea to come out with our solution.

packages/data-service/README.md
```diff
@@ -0,0 +1,147 @@
...
+  }
+`;
+
+// 3. We create an event handler as a stream to deal with search (events up)
+const { handler: onSearch, stream: searchEvent$ } = createEventHandler();
+const searchTerm$ = searchEvent$.startWith("");
+
+// 4. We compose the result with the query, and massage it down the pipe until
+// we get the props for our component.
+// The idea is from an Observable of data and events, compose a stream o React.Components
```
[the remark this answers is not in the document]
Fabs: thanks

packages/data-service/README.md
```diff
@@ -0,0 +1,147 @@
...
+const components$ = graphqlObservable(packageRepositoryQuery, schema, {
+  query: fetchRepositories(),
+  filter: searchTerm$
```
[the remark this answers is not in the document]
Fabs: Yes. The  data is built lively, so its arguments are also streams.

packages/data-service/README.md
```diff
@@ -0,0 +1,147 @@
...
+
+// Our in-house solution to resolve graphql queries using streams
+import { graphqlObservable } from "#PACKAGES/data-service/graphqlObservable";
+
+// tools to make easier to create schemas and queries
+import { makeExecutableSchema } from "graphql-tools";
+import gql from "graphql-tag";
+
+// Streams we get our data from
+import { fetchRepositories } from "#PLUGINS/catalog/src/js/repositories/repositoriesStream";
```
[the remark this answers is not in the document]
Fabs: ok. Won't mention marathon or mesos, but mention that it is not an ednpoint.

packages/data-service/README.md
Fabs: ok

nLight:
I wonder why tests didn't even run

Fabs:
@nLight that is actually weird :/

I think on any case that this PR should not be prevented to merge because of tests, since it is just a text file and can't possibly brake tests. 

Maybe @juliangieseke, @DanielMSchmidt, @bstavroulakis or @ahoskins are in a better position to investigate that.

packages/data-service/README.md
```diff
@@ -0,0 +1,147 @@
...
+# API v1
+
+## Mediator Pattern
+
+The [mediator pattern](https://en.wikipedia.org/wiki/Mediator_pattern) is used here
+to decouple data handling from the view components by encapsulating the way they
+communicate. 
+
+With this pattern, a component will receive down all the data `(data down)` it needs via props,
+and send `events up` so the mediator can perform mutations (represented here as a simple callback).
```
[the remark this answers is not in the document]
juliangieseke: ?

packages/data-service/README.md
```diff
@@ -0,0 +1,147 @@
...
+## Mediator Pattern
+
+The [mediator pattern](https://en.wikipedia.org/wiki/Mediator_pattern) is used here
+to decouple data handling from the view components by encapsulating the way they
+communicate. 
+
+With this pattern, a component will receive down all the data `(data down)` it needs via props,
+and send `events up` so the mediator can perform mutations (represented here as a simple callback). 
+The key concept here is that both the data provider and the view component have no idea how each other work.
+Only the caller of `renderComponents` understands how it works.
```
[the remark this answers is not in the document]
juliangieseke: ?

packages/data-service/README.md
```diff
@@ -0,0 +1,147 @@
...
+communicate. 
+
+With this pattern, a component will receive down all the data `(data down)` it needs via props,
+and send `events up` so the mediator can perform mutations (represented here as a simple callback). 
+The key concept here is that both the data provider and the view component have no idea how each other work.
+Only the caller of `renderComponents` understands how it works.  
+
+```js
+  /*
+   * Receives a stream of React.Component and outputs a single components that
```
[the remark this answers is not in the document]
juliangieseke: ?

packages/data-service/README.md
```diff
@@ -0,0 +1,147 @@
...
+
+```js
+  /*
+   * Receives a stream of React.Component and outputs a single components that 
+   * renders the most recevent event from the Observable.
+  */
+  componentFromStream(component$: Observable<React.Component>): React.Component;
+```
+
+Here is an example of how it could be used to render a page for `space shuttles`
```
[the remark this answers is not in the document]
juliangieseke: ?

packages/data-service/README.md
```diff
@@ -0,0 +1,147 @@
...
+that shows which shuttles are on space, and which are ready to be launched.
+
+Note that the way we achieve the Mediator pattern is not in an OO fashion, but
```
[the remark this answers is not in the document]
juliangieseke: ?

packages/data-service/README.md
```diff
@@ -0,0 +1,147 @@
...
+by abstracting all the communication between the data sources and the view components
+by using a stream.
+
+## Query Example
+
+To exemplify how you would use the Mediator pattern to get data from a graphql
```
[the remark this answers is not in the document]
juliangieseke: ?

packages/data-service/README.md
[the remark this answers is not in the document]
Fabs: That is weird, I will look into it again. I probably have a copy on Grammarly.

Fabs:
Sorry @juliangieseke, now we have the latest, latest.

juliangieseke:
@Fabs can you force push something again? This weird issue is a Jenkins bug which happens "sometimes" (according to their Bugtracker…)

Fabs:
@juliangieseke. I caused it to trigger again :D.

PR #2842, opened 2018-03-20
DCOS-21638 use the mediator api on repository settings

Fabs:
This implements the mediator pattern for the Settings page using recompose + rxjs. In the end we had to implement 0 lines of code on the data-service 🎆 .

To test this just play with Settings > Package Repositories, add, remove and filter packages.

Come to me if you need valid repos to test, or just remove first and then you can add again.

- [x] adequate with the API here #2841 

**Checklist**

### DIFF
--- a/src/js/components/RepositoriesTabUI.js
+++ b/src/js/components/RepositoriesTabUI.js
@@ -1,0 +1,116 @@
+import { Link } from "react-router";
+import React from "react";
+
+import AddRepositoryFormModal
+  from "#SRC/js/components/modals/AddRepositoryFormModal";
+import Breadcrumb from "#SRC/js/components/Breadcrumb";
+import BreadcrumbTextContent from "#SRC/js/components/BreadcrumbTextContent";
+import FilterBar from "#SRC/js/components/FilterBar";
+import FilterInputText from "#SRC/js/components/FilterInputText";
+import Loader from "#SRC/js/components/Loader";
+import Page from "#SRC/js/components/Page";
+import RepositoriesTable from "#SRC/js/components/RepositoriesTable";
+import RequestErrorMsg from "#SRC/js/components/RequestErrorMsg";
+
+const RepositoriesBreadcrumbs = addButton => {
+  const crumbs = [
+    <Breadcrumb key={-1} title="Repositories">
+      <BreadcrumbTextContent>
+        <Link to="/settings/repositories">Package Repositories</Link>
+      </BreadcrumbTextContent>
+    </Breadcrumb>
+  ];
+
+  return (
+    <Page.Header.Breadcrumbs
+      iconID="settings"
+      breadcrumbs={crumbs}
+      addButton={addButton}
+    />
+  );
+};
+
+const METHODS_TO_BIND = ["handleCloseAddRepository", "handleOpenAddRepository"];
+
+export default class RepositoriesTabUI extends React.Component {
+  constructor() {
+    super();
+
+    this.state = {
+      addRepositoryModalOpen: false
+    };
+
+    METHODS_TO_BIND.forEach(method => {
+      this[method] = this[method].bind(this);
+    });
+  }
+
+  handleCloseAddRepository() {
+    this.setState({ addRepositoryModalOpen: false });
+  }
+
+  handleOpenAddRepository() {
+    this.setState({ addRepositoryModalOpen: true });
+  }
+
+  getContent() {
+    const { addRepositoryModalOpen } = this.state;
+
+    const { repositories, searchTerm, onSearch } = this.props;
+
+    return (
+      <div>
+        <FilterBar>
+          <FilterInputText
+            className="flush-bottom"
+            placeholder="Search"
+            searchString={searchTerm}
+            handleFilterChange={onSearch}
+          />
+        </FilterBar>
+        <RepositoriesTable repositories={repositories} filter={searchTerm} />
+        <AddRepositoryFormModal
+          numberOfRepositories={repositories.getItems().length}
+          open={addRepositoryModalOpen}
+          onClose={this.handleCloseAddRepository}
+        />
+      </div>
+    );
+  }
+
+  render() {
+    return (
+      <RepositoriesTabUI.Page addButton={this.handleOpenAddRepository}>
+        {this.getContent()}
+      </RepositoriesTabUI.Page>
+    );
+  }
+}
+
+RepositoriesTabUI.Loading = () => {
+  return (
+    <RepositoriesTabUI.Page>
+      <Loader />
+    </RepositoriesTabUI.Page>
+  );
+};
+
+RepositoriesTabUI.Error = () => {
+  return (
+    <RepositoriesTabUI.Page>
+      <RequestErrorMsg />
+    </RepositoriesTabUI.Page>
+  );
+};
+
+RepositoriesTabUI.Page = ({ addButton, children }) => {
+  return (
+    <Page>
+      <Page.Header
+        addButton={addButton}
+        breadcrumbs={<RepositoriesBreadcrumbs />}
+      />
+      {children}
+    </Page>
+  );
+};
--- a/src/js/pages/system/RepositoriesTab.js
+++ b/src/js/pages/system/RepositoriesTab.js
@@ -1,158 +1,31 @@
-import mixin from "reactjs-mixin";
-import { Link } from "react-router";
-/* eslint-disable no-unused-vars */
 import React from "react";
-/* eslint-enable no-unused-vars */
-import { StoreMixin } from "mesosphere-shared-reactjs";
+import { componentFromStream } from "recompose";
 
-import AddRepositoryFormModal
-  from "../../components/modals/AddRepositoryFormModal";
-import Breadcrumb from "../../components/Breadcrumb";
-import BreadcrumbTextContent from "../../components/BreadcrumbTextContent";
-import CosmosPackagesStore from "../../stores/CosmosPackagesStore";
-import FilterBar from "../../components/FilterBar";
-import FilterInputText from "../../components/FilterInputText";
-import Loader from "../../components/Loader";
-import Page from "../../components/Page";
-import RepositoriesTable from "../../components/RepositoriesTable";
-import RequestErrorMsg from "../../components/RequestErrorMsg";
+import RepositoriesTabUI from "#SRC/js/components/RepositoriesTabUI";
 
-const RepositoriesBreadcrumbs = addButton => {
-  const crumbs = [
-    <Breadcrumb key={-1} title="Repositories">
-      <BreadcrumbTextContent>
-        <Link to="/settings/repositories">Package Repositories</Link>
-      </BreadcrumbTextContent>
-    </Breadcrumb>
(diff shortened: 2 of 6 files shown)


packages/data-service/README.md
```diff
@@ -0,0 +1,79 @@
+This package is to be extracted to dcos-ui/data-service.
```
Fabs: There is a PR for discussions on this file:
https://github.com/dcos/dcos-ui/pull/2841

DanielMSchmidt: But why is it also in this PR 🤔

src/js/streams/CosmosPackagesStream.js
```diff
@@ -0,0 +1,24 @@
...
+import CosmosPackagesStore from "#SRC/js/stores/CosmosPackagesStore";
+
+const cosmosPackageSource = Observable.timer(1000, 2000)
+  .map(() => CosmosPackagesStore.fetchRepositories())
+  .map(() => CosmosPackagesStore.getRepositories())
+  .distinctUntilChanged();
+
+const searchPackages = new Subject();
+const searchTerms = searchPackages
+  .merge(Observable.of(""))
```
DanielMSchmidt: What does `Observable.of("")` mean?

src/js/streams/CosmosPackagesStream.js
```diff
@@ -0,0 +1,24 @@
+// to be replaced with https://github.com/dcos/dcos-ui/pull/2840
+import { Observable, Subject } from "rxjs";
+
+import CosmosPackagesStore from "#SRC/js/stores/CosmosPackagesStore";
+
+const cosmosPackageSource = Observable.timer(1000, 2000)
+  .map(() => CosmosPackagesStore.fetchRepositories())
```
DanielMSchmidt: Couldn't this be `.map(CosmosPackagesStore.fetchRepositories)`?

packages/data-service/README.md
Fabs: I rebased from the same place to get all the web pack and other niceties. I was betting that by the time this one goes in, the other one would also have been in

src/js/streams/CosmosPackagesStream.js
Fabs: `Observable.of("")` creates an observable of a single value, emits it, and completes.

The `.merge(Observable.of(""))` could be more clearly written as startWith(""). It is a way to give a first value for a stream. 

Will fix that.

packages/data-service/README.md
Fabs: I put this on hold merge until the other one is merged.

src/js/components/RepositoriesTabUI.js
```diff
@@ -0,0 +1,123 @@
...
+
+RepositoriesTabUI.Error = () => {
+  return (
+    <RepositoriesTabUI.Page>
+      <RequestErrorMsg />
+    </RepositoriesTabUI.Page>
+  );
+};
+
+RepositoriesTabUI.Page = ({ addButton, children }) => {
```
jaxesn: Is this used?

src/js/streams/CosmosPackagesStream.js
Fabs: Unfortunately not. Because the fetchRepositories accepts a type parameter, and if I do that, I am passing the number from the timer as this parameter, but that is not correct. It is actually not gonna work because of that.

I only pass function references to map if my function is of arity 1 and is actually using the value coming from the map function `(item) => f(item)`. Otherwise, the semantics is not equivalent anymore.

src/js/components/RepositoriesTabUI.js
Fabs: on line 107: <RepositoriesTabUI.Page>

Fabs: and line 99 too.

Fabs:
@jaxesn. At first, I planned to implement `componentFromStream` myself. That I would type the observables so you would pass an Observable<React.Component>. Also, the handler functions could also have better-defined types.

Now that I got everything from recomposing, and had to implement nothing, I might check the @ts/recompose and see if adds values for us. 

I think types would be really helpful to implement the dcos-labs/data-service package that is going to be extracted from packages/data-service folder, but as for now, I have implemented nothing 💃

packages/data-service/README.md
Fabs: Actually, I will also rebase to not have that commit.

src/js/components/RepositoriesTabUI.js
jaxesn: doh. Is there any way to share this in the main render method?

DanielMSchmidt:
Oh okay, so we already had tests for that piece, I wasn't aware of that. Will review tomorrow 👍

Fabs:
Thanks, @DanielMSchmidt. I started the efforts on TypeScript tonight.

src/js/components/RepositoriesTabUI.js
Fabs: Is the problem that the component is part of the public api of RepositoresTabUI? I could do

```
const = RepositoriesTabPage = ({ addButton, children }) => {
  return (
    <Page>
      <Page.Header
        addButton={addButton}
        breadcrumbs={<RepositoriesBreadcrumbs />}
      />
      {children}
    </Page>
  );
};
```

And this way do not export the page, but just use it internally. Does that address your issue?

DanielMSchmidt:
@Fabs please start also some efforts in the direction of sleeping :trollface:

src/js/components/RepositoriesTabUI.js
jaxesn: I dont really have an issue, I was just noticing it looked very similar to the render method on line 81.  I think this is just because its small and mostly the Page/Header boilerplate. I wouldn't bother changing anything.

Fabs:
@DanielMSchmidt. I have many clones :D

src/js/components/RepositoriesTabUI.js
Fabs: Right.

Fabs:
@DanielMSchmidt I tried to use enzyme yesterday to test this. I ran into an interesting problem @orlandohohmeier keep reminds us about. I added this ticket next sprint so we can address testing the integration between the React UI components and the streams.
https://jira.mesosphere.com/browse/DCOS-21792

DanielMSchmidt:
@Fabs Ok, seems fine, let's pair on it next time 👍

Fabs:
@DanielMSchmidt. Thanks for for the pairing offer. if it is ok can you approve?

Just to add some more info. The ability to test the mediator properly depends on us being able to inject data dependencies here. Which is something that @orlandohohmeier and @nLight are looking at with inversify.js and etc. That is why It was hard to test (and not because of enzyme).

Regarding TS I think we are lucky we don't have it here, because it will still take a while to get it merged. (webbbpack -> ts -> tslint).

DanielMSchmidt:
@Fabs I think we need to make sure our review cycles are faster. Taking them as a reason to not add type checking just doesn't feel right to me. Therefore I think I should not approve this PR, nor reject it :) Waiting for others to air their opinion :)

Fabs:
@DanielMSchmidt typescript is a nice to have for the data layer, not a must-have. I decided to add it because it looked like a quick win and it would be beneficial to the entire team. I would see it more like as a choice I would have to make something better, and less as an obligation to deliver it on typescript. Especially given the goals that this project has to achieve to enable MT.

The types we will have to describe in grpahql schemas, is the real must have for the Data Layer.

DanielMSchmidt:
@Fabs Agreed, I didn't saw this as testing the mediator but more as the first step and I didn't want us to just start with JS when we could use TS instead. I see your point with fast iteration cycles, I just hope we can really leverage TS here in the long term

Fabs:
@DanielMSchmidt. I think we are aligned on this wish. I would love if the data-service (the package we will extract the data-layer implementation too) is entirely written in .ts.

Thanks for challenging and for the help :D.

Fabs:
Apparently the fail is legit, will look into it.

src/js/components/RepositoriesTabUI.js
```diff
@@ -0,0 +1,123 @@
...
+          }}
+          breadcrumbs={<RepositoriesBreadcrumbs />}
+        />
+        {this.getContent()}
+      </Page>
+    );
+  }
+}
+
+RepositoriesTabUI.Loading = () => {
```
juliangieseke: I dont like this approach and AFAIK we have a eslint rule which should prevent us from creating multiple components in one file 🤔 
What exactly do you try to archive here? is it possible to create a kind genereric `<Loading>` Component?

src/js/components/RepositoriesTabUI.js
```diff
@@ -0,0 +1,123 @@
...
+  return (
+    <RepositoriesTabUI.Page>
+      <Loader />
+    </RepositoriesTabUI.Page>
+  );
+};
+
+RepositoriesTabUI.Error = () => {
```
juliangieseke: same as above

src/js/components/RepositoriesTabUI.js
juliangieseke: and again :)

src/js/components/RepositoriesTabUI.js
```diff
@@ -0,0 +1,123 @@
...
+          open={addRepositoryModalOpen}
+          onClose={this.handleCloseAddRepository}
+        />
+      </div>
+    );
+  }
+
+  render() {
+    return (
+      <Page>
```
juliangieseke: This should be `RepositoriesTabUI.Page` :) (no, it shouldn't! see comments below)

src/js/pages/system/RepositoriesTab.js
```diff
@@ -1,158 +1,31 @@
...
-      </div>
+      <RepositoriesTabUI
+        repositories={data.repositories}
+        searchTerm={data.searchTerm}
+        onSearch={onSearch}
+      />
     );
-  }
+  })
+  .startWith(<RepositoriesTabUI.Loading />)
```
juliangieseke: 🙅‍♂️

src/js/pages/system/RepositoriesTab.js
```diff
@@ -1,158 +1,31 @@
...
+  .catch(_err => <RepositoriesTabUI.Error />);
```
juliangieseke: 🙅‍♂️

src/js/components/RepositoriesTabUI.js
juliangieseke: not sure if we want to stick to one component per file?!

Fabs: That is the same pattern for `Page`, `Page.Header`. It is still one component per file `as per lint rule`.

src/js/components/RepositoriesTabUI.js
Fabs: it should, bad rebase

Fabs:
This PR provided a great discussion. We have a second version already that supplements this one, so this PR is no longer necessary:

https://github.com/dcos/dcos-ui/pull/2842

PR #2851, opened 2018-03-22
fix(MesosStateStore): adjust stream data handling

orlandohohmeier:
Use `Promise.resolve().then()` to trigger the stream subscription asynchronously on the same macrotask (by using a microtask). Events emitted by the strean, trigger `onStreamData` which is used to emit an event (using EventEmitter) called `MESOS_STATE_CHANGE`. This change will mitigate issues in which errors in view components lead to stale data because the error propagation broke the stream.

N.B. `EventEmitter` is synchronous meaning that every delay in an event handler effects the `emit` execution time and errors propagate up to the `emit` call which sometimes leads to undesireable behavior.

Closes DCOS-21784

**how to test?**
You want to test this against an enterprise cluster that does not have regions. An example of that cluster is the `nuc` cluster sitting on @fabs desk (sorry SF). On master, if you go to the Networking page on the said cluster, you will get an error. Once that error happens, the mesos stream is broken and if you go to Nodes > Grid, you won't be able to see any grid data. If you then switch to this branch, the data should come normally, event with the error happening.  I recommend you test other pages that rely on data coming from the mesos stream. 

Closes DCOS-21784

### DIFF
--- a/src/js/stores/MesosStateStore.js
+++ b/src/js/stores/MesosStateStore.js
@@ -56,7 +56,7 @@ class MesosStateStore extends GetSetBaseStore {
     this.on(eventName, callback);
 
     if (this.shouldSubscribe()) {
-      this.subscribe();
+      Promise.resolve().then(() => this.subscribe());
     }
   }
 
@@ -99,7 +99,8 @@ class MesosStateStore extends GetSetBaseStore {
     // refresh limits to the UI. They are:
     //
     // MOST once every (Config.getRefreshRate() * 0.5) ms. due to debounceTime.
-    // LEAST once every tick of Config.getRefreshRate() ms in Observable.interval
+    // LEAST once every tick of Config.getRefreshRate() ms in
+    // Observable.interval
     //
     // TODO: https://jira.mesosphere.com/browse/DCOS-18277
     this.stream = waitStream


nLight:
run tests

nLight:
run all tests

juliangieseke:
🏆 longest branchname ever?! 😆

src/js/stores/MesosStateStore.js
```diff
@@ -99,14 +99,15 @@ class MesosStateStore extends GetSetBaseStore {
...
+    // LEAST once every tick of Config.getRefreshRate() ms in
+    // Observable.interval
     //
     // TODO: https://jira.mesosphere.com/browse/DCOS-18277
     this.stream = waitStream
       .concat(eventTriggerStream)
       .debounceTime(Config.getRefreshRate() * 0.5)
       .retryWhen(linearBackoff(MAX_RETRIES, RETRY_DELAY))
-      .subscribe(this.onStreamData, this.onStreamError);
+      .subscribe(Promise.resolve().then(this.onStreamData), this.onStreamError);
```
juliangieseke: any reason you chose `Promise` over `setTimeout`? Might be personal taste, but I'd argue `setTimeout` is adds a bit less complexity here. specially because you dont `catch`.

nLight: Would it make more sense wrapping the whole `subscribe` method in `addChangeListener`. Those integration tests bug me, not reproducible locally though.

Fabs:
@juliangieseke 
image

Fabs:
run integration tests

src/js/stores/MesosStateStore.js
Fabs: found this on why the promise approach:
https://stackoverflow.com/questions/25915634/difference-between-microtask-and-macrotask-within-an-event-loop-context

orlandohohmeier: @juliangieseke I went with `Promise.resolve().then()` as it schedules microtasks which are executed right after the script is done but before the next tasks are started. Using `setTimeout` would schedule a new task which is not necessary in this case. 

I recommend the following post for details on tasks and microtasks: https://jakearchibald.com/2015/tasks-microtasks-queues-and-schedules/ 

@nLight What do you mean with "wrapping the whole subscribe method in addChangeListener"?

nLight: Wouldn't this do the same trick and stop exception propagation?

```js
  addChangeListener(eventName, callback) {
    this.on(eventName, callback);

    if (this.shouldSubscribe()) {
        Promise.resolve().then(() => this.subscribe());
    }
```

Fabs: Regarding micro/macro tasks and observer scheduler, we settled that in this case it does not really matter for us, but I would like to share a little bit of what I learned about schedulers.

Contrary to rxjs 4, rxjs 5 default scheduler depends on the operator.
https://github.com/ReactiveX/rxjs/blob/master/doc/scheduler.md

The default scheduler for `from` rxjs 5 is synchronous `Scheduler.queue`:
https://github.com/Reactive-Extensions/RxJS/blob/master/src/modular/observable/from.js
https://jsbin.com/keriluy/edit?js,console

For `Interval` on the other hand, it is `Scheduler.async`:
https://github.com/Reactive-Extensions/RxJS/blob/master/src/modular/observable/interval.js

Fabs:
I will fix the tests and continue this monday.

src/js/stores/MesosStateStore.js
Fabs: The `.subscribe(Promise.resolve().then....)` breaks the nodes page (my guess because that page stops listening). `Promise.resolve().then(this.subscribe)` (by @nLight), contrary to what I said, fix the issue without causing the nodes page to break. 

I am switching to that fix.

PR #2817, opened 2018-03-08
DCOS_OSS-1564: Let's update webpack (2nd try)

juliangieseke:
Something went wront when we tried to rebase the old PR (#2699), this was the latest backup and work is continued here 👍 

please refer to the old PR for more information.

PR #2852, opened 2018-03-22
DCOS-18691 typescript support

Fabs:
~Hold merge *only if* #2817 is not merged. Otherwise please remove the label, change the base to master and merge :D~

To test, checkout and add src/js/banana.tsx
```js
import { Observable } from "rxjs";

const logger = (s: string) => {
  const obs: Observable<string> = Observable.of(s).merge(Observable.of("ya"));
  obs.subscribe(console.log);
};

console.log("with bananas");

export interface IBadgeProps {
  appearance: string;
  children: React.ReactNode | string;
}

class Badge extends React.PureComponent<IBadgeProps, {}> {
  public static defaultProps: Partial<IBadgeProps> = {
    appearance: "default"
  };

  public render() {
    const { appearance, children } = this.props;

    return <span className={appearance}>Aaaaaaah!{children}</span>;
  }
}

export { logger, Badge };
```
Then on some page, you can import the `logger` or the `Badge` shamefully stolen from dcos-labs/ui-kit 🙈 , thanks @Poltergeist and @weblancaster).

you can also run:
`npx tsc` or `npx tsc:watch`
 
I added types for react to test that typescript found my types.

NEED HELP WITH
- [ ] get the peer dependencies of the tslint (might mean updating prettier)
- [ ] warnings/errors on console
- [ ] react types might be wrong, but we dont have the exact version available


tsconfig.json
```diff
@@ -0,0 +1,75 @@
+{
```
weblancaster: Hey @Fabs please cleanup removing the comments.

Fabs: We decided to left the comments to help people understand what the settings mean. I have no strong opinion here. Why would you clean up?

DanielMSchmidt: Would also like to have the comments, for readability reasons

package.json
```diff
@@ -177,6 +181,8 @@
     "test": "NODE_PATH=node_modules jest --config=jest.config.js --no-cache",
     "test:watch": "npm test -- --watch",
     "testing": "NODE_ENV=testing webpack-dev-server --config ./webpack/webpack.dev.js --progress --colors --content-base ./build --host 0.0.0.0 --port $npm_package_config_port",
+    "tsc": "NODE_ENV=development $(npm bin)/tsc",
```
DanielMSchmidt: I think we don't need that, right?

package.json
```diff
@@ -177,6 +181,8 @@
...
+    "tsc:watch": "NODE_ENV=development $(npm bin)/tsc -w --noEmit",
```
DanielMSchmidt: Why do you need the npm bin subshell?

webpack/webpack.config.js
```diff
@@ -73,19 +75,26 @@ module.exports = {
...
         exclude: /node_modules/,
-        use: {
-          loader: "babel-loader",
-          options: {
-            presets: requireAll([
-              "babel-preset-es2015",
-              "babel-preset-stage-3",
-              "babel-preset-react"
-            ])
+        use: [
```
DanielMSchmidt: The change here is unnecessay, It can be an object instead of an array of object. I don't think it has a different meaning, but it's still something that just could stay the same

webpack/webpack.config.js
```diff
@@ -11,11 +11,13 @@ function requireAll(array) {
 
 module.exports = {
   entry: "./src/js/index.js",
+  devtool: "inline-source-map",
   output: {
     filename: "[name].js",
     path: path.resolve(__dirname, "../dist")
   },
   resolve: {
+    extensions: [".ts", ".tsx", ".js"],
```
DanielMSchmidt: Do we want to use the `tsx` suffix? We don't use `.jsx`, so I would opt-in for a no here

webpack/webpack.config.js
```diff
@@ -11,11 +11,13 @@ function requireAll(array) {
 
 module.exports = {
   entry: "./src/js/index.js",
+  devtool: "inline-source-map",
```
DanielMSchmidt: I think you can remove that, the source maps are set on the extending webpack configs (prod and dev)

package.json
Fabs: That one maybe not, but the watch is actually useful. I think for consistency if we have `tsc:watch` we should have `tsc`.

package.json
Fabs: I find it more elegant than to directly access ./node_modules/bin. Would you do it that way?

webpack/webpack.config.js
Fabs: you are right, that should go away.

webpack/webpack.config.js
Fabs: Could not make the `tsc` understand jsx syntax on a `ts` file. So if we don't support `tsx`  that means we do not support using types on a react component.

That is also the same way it is used on ui-kit.

webpack/webpack.config.js
Fabs: Will try to fix it.

package.json
DanielMSchmidt: For me it is a little source of confusion, I would think I need to run tsc now besides every npm start

package.json
DanielMSchmidt: You don't need that at all, npm adds ./node_modules/.bin to your path so you can simply write tsc

package.json
DanielMSchmidt: If you want to use it locally with the watch, you could just do `npx tsc -w --noEmit` 👍

webpack/webpack.config.js
DanielMSchmidt: Let's see if we can teach that somehow to `tsc`. I really dislike the idea of having ts and tsx files. Let's timebox it to an hour and if we don't succeed I am fine with that

webpack/webpack.config.js
DanielMSchmidt: Removing it should be sufficient 👍

webpack/webpack.config.js
Fabs: Actually, my timebox for typescript is already over. If you have you or someone else have some time to do it, that would be awesome. But I will be picking it up once I have graphql in the data layer.

tsconfig.json
weblancaster: I disagree with the choice of leaving the comments especially because we can and should always use the official documentation.

otherwise, why don't we do that for all config files? `.dockerignore` `.babelrc` `.editorconfig` `.eslintignore` `.eslintrc` `.stylelintrc` `Dockerfile` `jenkinsfile` `cypress.json`

That being said, I'm ok if we decide to leave the comments.

webpack/webpack.config.js
weblancaster: @DanielMSchmidt TS is strict about the file type and if I'm not wrong it only accepts ts or tsx.

I also suggest using ts/tsx to TS files and js/jsx for JS files because even though they are both JS, TS is a language (e.g even GH support TS/TSX as a different language than JS).

Fabs: I would recommend us to bring this one to sync for discussion next week. I am putting the topic on the sync for the case we cant solve until there. Maybe we should also research if it is even possible to have ts file only (so far I could not).

tsconfig.json
Fabs: I am inclined to think like @weblancaster (siding with consistency). I started with no comments, I read the docs. But when I ran `tsc --init` it had comments and other team members proposed leaving then. My original intention was to remove, but I caved in 🙈

webpack/webpack.config.js
Fabs: done. It looks like the source maps still work

package.json
```diff
@@ -89,6 +92,7 @@
     "babel-runtime": "6.23.0",
     "colors": "1.1.2",
     "compression-webpack-plugin": "1.1.6",
+    "console": "0.7.0",
```
Fabs: ts-loader needs this one, for some reason.

webpack/webpack.config.js
```diff
@@ -16,6 +16,7 @@ module.exports = {
     path: path.resolve(__dirname, "../dist")
   },
   resolve: {
+    extensions: [".ts", ".tsx", ".js"],
```
Fabs: We still want to discuss the extension here.

DanielMSchmidt:
@Fabs I tried to use the [`appendtsxsuffixto-regexp-default`](https://github.com/TypeStrong/ts-loader#appendtsxsuffixto-regexp-default) option in the ts-loader, but I failed. Although I am a bit disappointed that this seems to be impossible, let's stick with the TS convention of having `.tsx`

Fabs:
@DanielMSchmidt ok. I will look into it again next sprint then, but I al agree too merge it as is.

tsconfig.json
Fabs: @DanielMSchmidt, in the end I choose to remove the comments. Are you ok with that decision?

DanielMSchmidt:
I get these errors only on this branch, there seem to be some errors still
bildschirmfoto 2018-03-28 um 10 52 08

Fabs:
@DanielMSchmidt about the warnings, maybe they appeared after I removed the `devtool: 'inline-source-map'` (see here: https://webpack.js.org/guides/typescript/). I notice that issue now too.

juliangieseke:
rebased 👍 
@weblancaster maybe you can re-review 🙇 
@DanielMSchmidt is on vacation, not sure if he will unblock this before tuesday :(

juliangieseke:
we have `devtool` in out config and even changing it to `inline-source-map` doesnt fix it :(
https://github.com/dcos/dcos-ui/blob/d72689df49c3b9a6c0676c971849345f271fcf43/webpack/webpack.dev.js#L16

Fabs:
Also add tslint from: https://github.com/dcos-labs/tslint-config

Fabs:
I added tslint again. Did not fix the errors though.  Would love help.

Fabs:
That has the potential to bite us: 
npm WARN tslint-plugin-prettier@1.3.0 requires a peer of prettier@^1.4.0 but none is installed. You must install peer dependencies yourself.

we have a much older version of prettier, and when I tried to update it gave me trouble.

Fabs:
I would say this is now ready for prime time.

weblancaster:
@Fabs the warnings are gone but I'm seeing an error still when trying to run the local build.

screen shot 2018-04-04 at 11 07 43 am

screen shot 2018-04-04 at 11 07 56 am

juliangieseke:
you need to fix you `proxy.dev.js`, see `proxy.template.js` for reference. 👍

PR #2853, opened 2018-03-22
fix(ServiceStorageConfigSection): adjust getDefinition default

orlandohohmeier:
Adjust the `ServiceStorageConfigSection.getDefinition` default to align the return a type with the superclass.

Closes DCOS-20949


DanielMSchmidt:
@orlandohohmeier would you mind rebasing this? It appears to me that this is based of a 4 week old master (so pre node 8, etc).

PR #2854, opened 2018-03-22
DCOS-18693 add support to tslint

Fabs:
You can use the same code here #2852  to test for linting error. It is on hold merge because it depends on #2852


tslint.json
```diff
@@ -0,0 +1,74 @@
+{
```
jaxesn: Did you build this to roughly match our eslint config rules?

weblancaster: Hey @jaxesn the `tslint.json` here was taken from the work done in the UI Kit repo (https://github.com/dcos-labs/ui-kit). 

But to answer your question, yes I used our eslint as a reference point for the rules that tslint support.

A lot of those rules aren't supported by tslint or are a bit different so for that reason, we have to use https://github.com/buzinas/tslint-eslint-rules but even though some rules aren't present which I believe should be fine.

weblancaster:
@Fabs since now we are also going to use the `tslint` on DCOS UI.. what do you think of us follow the same example of `eslint-config` (https://github.com/dcos-labs/eslint-config) and created a shared `tslint-config` repo that can be used in the UI Kit and DCOS-UI?

Fabs:
@weblancaster I love the idea to use the same tslint (eslint even) everywhere for consistency. Can we make that step 2 though? And first have it in as is? I copied your settings for tslint completely so we will start with the same.

Also for the typescript.config, it would be cool to have the same as well. Take a look on the one we pick for dcos-ui to see what you think. It is more strict, but also prevent a wider set of errors that happen with nulls and tihngs you can forget.

tslint.json
DanielMSchmidt: @nLight Would it make sense to have this config in a similar way as our eslint config, I mean as a seperate repository? Not for now, just as a future improvement

tslint.json
```diff
@@ -0,0 +1,74 @@
...
+    "no-unused-variable": [
+      true,
+      "check-parameters",
+      {
+        "ignore-pattern": "^_"
+      }
+    ]
+  },
+  "no-use-before-declare": false,
+  "quotemark": [false],
```
DanielMSchmidt: The [eslint-plugin-test-names](https://github.com/DanielMSchmidt/eslint-plugin-test-names) rule is missing, I would really love to have it added. Currently we have no tests written in TS, but that should change in the future

tslint.json
Fabs: Isn't that the same thing as @weblancaster proposed?

tslint.json
Fabs: The reason we introduced TS quickly is to allow us to move forward with a better-written data layer. Having a lint rule to verify tests do not contain "should" is a great thing, but does not really help with that goal. 

I would propose either to be added together with the first test written in .ts, or to be done when we unify the two tslint.config files into a single repo.

We should move forward here without that rule.

weblancaster: @DanielMSchmidt please take a look at my review for more info on tslint rules. https://github.com/dcos/dcos-ui/pull/2854#discussion_r176583067

DanielMSchmidt: I think a part of introducing typescript needs to be having the same coding standards enforced in typescript and in javascript. I know the config is based on the one in the ui-kit, it was my bad that I did not spot it there. The only alternative for me would be to have a new JIRA ticket for having a docs-labs/tslint-config repo, so that we are on the same page here. Adding them on the fly when we need them is IMHO not the best approach when it comes to linting (I know we won't come in this situation, but enforcing new standards at a later point of a project really is a pain, [see this ever stale PR](https://github.com/wix/detox/pull/517))

Fabs: As long as it is not blocking us delivering great things, I agree we could take a slower but higher quality approach. I would give +1 to that, and try to get this ticket in the next sprint. I am ok waiting for this to get typescript in.

DanielMSchmidt: I wouldn't call 10 LOC to be copied over blocking, TBH.
For me we can clean up afterwards, but I want to make sure we have a ticket for that in the next sprint

tslint.json
DanielMSchmidt: @Fabs Yeah, I did not see his comment

tslint.json
weblancaster: @DanielMSchmidt we are aware of that and that's the goal, to have a single source of truth for the tslint config.

**One thing that we need to keep in mind** is that tslint and eslint don't have the same rules and that's a fact nothing we can do about it besides trying to collaborate to tslint project adding the missing rules.

weblancaster:
@DanielMSchmidt @Fabs please take a look at the https://github.com/dcos-labs/tslint-config for a shared tslint-config Jira https://jira.mesosphere.com/browse/DCOS_OSS-2321.

tslint.json
DanielMSchmidt: Oh I understood https://github.com/buzinas/tslint-eslint-rules like it could "polyfill" the same interface, but I looked it up and it doesn't work that way. If we start writing tests in TS (what I would like us to do) I will just write a tslint plugin for the same thing :)

DanielMSchmidt:
bildschirmfoto 2018-03-29 um 10 09 42

Fabs:
I think we should close this in favor of the work @weblancaster did on tslint. https://github.com/dcos-labs/tslint-config

@nLight I think needs to give us rights on npm.

PR #2856, opened 2018-03-23
POC: Switch Marathon Store from polling to EventBus

nLight:
# 🚨README ! Not to be merged 🚨

This is a POC demonstrating that we could switch to Marathon Event Bus as a signaling stream and reduce the polling.

I vaguely split events by type to call certain endpoints. We should be able to optimize this even further by buffering events and making a call at most `updateInterval`.

Also keep in mind that we should keep the polling since Edge doesn't support SSE.

Closes DCOS-21383

### DIFF
--- a/plugins/services/src/js/stores/MarathonStore.js
+++ b/plugins/services/src/js/stores/MarathonStore.js
@@ -1,4 +1,5 @@
 import PluginSDK from "PluginSDK";
+import Rx from "rxjs";
 
 import { SERVER_ACTION } from "#SRC/js/constants/ActionTypes";
 import AppDispatcher from "#SRC/js/events/AppDispatcher";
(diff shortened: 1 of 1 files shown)


plugins/services/src/js/stores/MarathonStore.js
```diff
@@ -86,33 +87,102 @@ import Framework from "../structs/Framework";
...
-    global.clearInterval(requestInterval);
-    requestInterval = null;
-  }
+    source.addEventListener("open", function(event) {
+      observer.next(event);
+    });
+
+    const onMessage = event => observer.next(event);
+
+    source.addEventListener("pod_created_event", onMessage);
```
DanielMSchmidt: I would like to see something like 

```js
const marathonEvents = ["pod_updated_event",...,"framework_message_event"];
marathonEvents.forEach(event => source.addEventListener(event, onMessage));
source.addEventListener("error", function(event) {
   observer.error({
    target: event.target,
    event
   });
});
```

plugins/services/src/js/stores/MarathonStore.js
```diff
@@ -86,33 +87,102 @@ import Framework from "../structs/Framework";
...
+    source.addEventListener("error", function(event) {
+      observer.error({
+        target: event.target,
+        event
+      });
+    });
+
+    // Closing the source when there are no subscribers left
+    return function teardown() {
+      console.info("teardown");
```
DanielMSchmidt: I really think we should log sth like this, do we have some sort of logger which is not stipped out in production and has some kind of scoping (like "MarathonStore#teardown")

plugins/services/src/js/stores/MarathonStore.js
```diff
@@ -86,33 +87,102 @@ import Framework from "../structs/Framework";
...
+let eventSource$ = null;
 let shouldEmbedLastUnusedOffers = false;
 
-function startPolling() {
-  if (requestInterval == null) {
-    poll();
-    requestInterval = global.setInterval(poll, Config.getRefreshRate());
-  }
-}
+function marathonEventBus(url, options = { withCredentials: true }) {
```
DanielMSchmidt: For me this needs to be a different file with a different set of tests

plugins/services/src/js/stores/MarathonStore.js
```diff
@@ -86,33 +87,102 @@ import Framework from "../structs/Framework";
...
 }
 
-function poll() {
-  const options = {};
+function startPolling() {
+  if (eventSource$ == null) {
+    eventSource$ = marathonEventBus(
+      `${Config.rootUrl}${Config.marathonAPIPrefix}/events?plan-format=light`
+    );
+    eventSource$.subscribe(
```
DanielMSchmidt: I would love to have sth like 

```js
const deploymentEvents$ = eventSource$.filter(event => event.type.startsWith("deployment_"))
```

but that might be just because I am new to rxjs and don't see the advantage with your approach :)

plugins/services/src/js/stores/MarathonStore.js
DanielMSchmidt: sure, but "feedback wanted"

plugins/services/src/js/stores/MarathonStore.js
Fabs: I like that idea, and I think we should have it as a facility for dcos-ui. I can think many places I would like to use that.

plugins/services/src/js/stores/MarathonStore.js
```diff
@@ -86,33 +87,102 @@ import Framework from "../structs/Framework";
...
+        if (shouldEmbedLastUnusedOffers) {
+          options.params = "?embed=lastUnusedOffers";
+        }
 
-  if (shouldEmbedLastUnusedOffers) {
-    options.params = "?embed=lastUnusedOffers";
+        if (event.type.startsWith("deployment_")) {
+          MarathonActions.fetchDeployments();
+          MarathonActions.fetchQueue(options);
+        } else if (event.type.startsWith("group_change_")) {
```
Fabs: I don't think the match event name by string would be my preferred way to group events. I would like to be able to have groups of events based on a listing.

Like:
RUNNING_DEPLOYMENTS -> [deployment_success, deployment_failed, deployment_step_success, deployment_step_failure]

HEALTH -> [failed_health_check_event, health_status_changed_event]

that could even be:
DEPLOYMENT -> ALL_EVENTS.filter((event) => event.type.startsWith("deployment_"))

Or even maybe be able to open channels with groups of events:
```js
marathon$.filter(filterEventsByGroup(RUNNING_DFEPLOYMENTS)).subscribe()
```

plugins/services/src/js/stores/MarathonStore.js
```diff
@@ -86,33 +87,102 @@ import Framework from "../structs/Framework";
...
+          MarathonActions.fetchGroups();
+        } else {
+          MarathonActions.fetchDeployments();
```
Fabs: Building up on the previous idea, I would build something like.

```js
const deploymentHandlers = () => {
  MarathonActions.fetchDeployments();  
  MarathonActions.fetchQueue(options);
  MarathonActions.fetchGroups();
}

filter(filterEventsByGroup(DEPLOYMENT)).subscribe(deploymentHandlers)
```

plugins/services/src/js/stores/MarathonStore.js
```diff
@@ -86,33 +87,102 @@ import Framework from "../structs/Framework";
...
+        console.error(error);
+      }
+    );
   }
+}
 
-  MarathonActions.fetchGroups();
-  MarathonActions.fetchQueue(options);
-  MarathonActions.fetchDeployments();
+function stopPolling() {
```
Fabs: there is no way marathon can support websockets?

nLight: There's a way, I'm taking to @wavesoft. It's just a matter of priorities. start/stop polling here it's just the interface I didn't want breaking as this research goes.

nLight:
Thanks for the input everyone! We will use it in the implementation.
Research is done and proved the viability of the idea.
Closing the PR, but keeping the branch :)

PR #2865, opened 2018-03-29
fix(ci): run ee job always (also on PRs)

juliangieseke:
we do want to test our EE run on every PR we have, npot just on Master (because if we do it on master only, we could break something in a PR (not catched) and then master is red)


Jenkinsfile
```diff
@@ -174,12 +174,8 @@ pipeline {
 
     // trigger the other job to update the upstream reference
     stage ('Trigger Enterprise Update') {
-      when {
-        expression { env.BRANCH_NAME == 'master' || env.BRANCH_NAME =~ SEMVER_REGEX }
-      }
-
       steps {
-        build job: "frontend/dcos-ui-ee-release/${env.BRANCH_NAME}", parameters: [[$class: 'StringParameterValue', name: 'BRANCH_NAME', value: env.BRANCH_NAME]]
+        build job: "frontend/dcos-ui-ee-pipeline/${env.BRANCH_NAME}", parameters: [[$class: 'StringParameterValue', name: 'BRANCH_NAME', value: env.BRANCH_NAME]]
```
Fabs: I think we need to run it all the time because we don't wanna break plugins unintentionally.

nLight:
We should consider skipping EE part in case we need to ship the code that should go in both EE and OSS at the same time

juliangieseke:
This might slow us down a bit, but it should never block us. 
First we merge OSS part, once this is merged, we can merge EE part. 
OSS never relies on EE - or am I wrong?

juliangieseke:
Its not that easy (sure it isnt :():
> No item named frontend/dcos-ui-ee-pipeline/PR-2865 found

we need to figure out how EE can run on a PR 🤔 
or maybe we should remove this step for now and try again 😐

juliangieseke:
I'm dropping this, current solution is good enough. 👍 more details in the JIRA Ticket.

PR #2868, opened 2018-04-03
Spike Mock Server

DanielMSchmidt:
We need to find out if mock server would really work for us and how hard it would be for us to configure it in a way that is useful for us. Therefore we do a spike in this PR to see if it shall be our solution


ahoskins:
What is a spike?

DanielMSchmidt:
Something similar to a prototype, but it doesn't necessarily need to be a code change

PR #2871, opened 2018-04-04
DCOS-21798: render empty modal if the service is not loaded

DanielMSchmidt:
## How to test

1. Add a new service with this configuration:

```
{
  "id": "/do-not-delete-me",
  "version": "2018-03-22T11:28:46.456Z",
  "containers": [
    {
      "name": "container-1",
      "resources": {
        "cpus": 0.1,
        "mem": 128,
        "disk": 0
      },
      "image": {
        "kind": "DOCKER",
        "id": "nginx"
      },
      "endpoints": [
        {
          "name": "http",
          "containerPort": 80,
          "hostPort": 0,
          "protocol": [
            "tcp"
          ]
        }
      ],
      "volumeMounts": [
        {
          "name": "home",
          "mountPath": "/etc/log"
        }
      ]
    }
  ],
  "volumes": [
    {
      "name": "home",
      "persistent": {
        "type": "root",
        "size": 10,
        "constraints": []
      }
    }
  ],
  "networks": [
    {
      "name": "dcos",
      "mode": "container"
    }
  ],
  "scaling": {
    "instances": 1,
    "kind": "fixed"
  },
  "scheduling": {
    "placement": {
      "constraints": []
    }
  },
  "executorResources": {
    "cpus": 0.1,
    "mem": 32,
    "disk": 10
  },
  "fetch": []
}
```

2. Visit the edit page of this service
3. Reload the edit page of this service and see there is no error

## Context
We accessed getLabel on the service without checking if the service is
even there. Therefore reloading the edit page caused an error.

By checking for a service and rendering an empty modal otherwise
we make sure that we don't try to render before we have the data.

Rendering the create form in these cases lead to a mutation of the data,
which in return lead to the edit form never being rendered.
Rendering null if there is no service lead to a dark flickering for the
time before the service loads. An empty modal gives a nicer UX.

**Checklist**

### DIFF
--- a/plugins/services/src/js/components/modals/EditServiceModal.js
+++ b/plugins/services/src/js/components/modals/EditServiceModal.js
@@ -5,12 +5,17 @@ import EditFrameworkConfiguration
 import CreateServiceModal
   from "#PLUGINS/services/src/js/components/modals/CreateServiceModal";
 import DCOSStore from "#SRC/js/stores/DCOSStore";
+import FullScreenModal from "#SRC/js/components/modals/FullScreenModal";
 
 class EditServiceModal extends Component {
   render() {
     const { id = "/" } = this.props.params;
     const serviceID = decodeURIComponent(id);
     const service = DCOSStore.serviceTree.findItemById(serviceID);
+
+    if (!service) {
+      return <FullScreenModal open={true} />;
+    }
 
     if (
       service.getLabels().DCOS_PACKAGE_DEFINITION != null ||
--- a/plugins/services/src/js/components/modals/__tests__/EditServiceModal-test.js
+++ b/plugins/services/src/js/components/modals/__tests__/EditServiceModal-test.js
@@ -1,0 +1,56 @@
+import React from "react";
+import { shallow } from "enzyme";
+
+jest.mock("#SRC/js/stores/DCOSStore");
+
+const DCOSStore = require("#SRC/js/stores/DCOSStore");
+const EditServiceModal = require("../EditServiceModal").default;
+
+describe("EditServiceModal", function() {
+  it("renders an empty modal if no service is found", function() {
+    DCOSStore.serviceTree = {
+      findItemById() {
+        return null;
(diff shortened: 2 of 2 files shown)


plugins/services/src/js/components/modals/EditServiceModal.js
```diff
@@ -5,13 +5,18 @@ import EditFrameworkConfiguration
...
+import FullScreenModal from "#SRC/js/components/modals/FullScreenModal";
 
 class EditServiceModal extends Component {
   render() {
     const { id = "/" } = this.props.params;
     const serviceID = decodeURIComponent(id);
     const service = DCOSStore.serviceTree.findItemById(serviceID);
 
+    if (!service) {
+      return <FullScreenModal open={true} />;
```
nLight: Hm... now if I go to `http://localhost:4200/#/services/detail/%2Fi-do-not-exist/edit/` it will just render a blank page but I'd rather see an Error message that the service couldn't be found.

nLight: How bad it would be implementing the message? I'm ok merging 1.11 as is but in 1.12 I'd rather invest in a better user experience.

DanielMSchmidt: Oh you are right, in that case we have a problem.
I wonder how we can do this without having the error message flash up every time before the endpoint returns the data 🤔 

I would argue that this is quite a scope change from "let's not have this error anymore",would you be ok with me creating a new Jira and pinging design for some UX ideas here?

nLight: I'm pretty sure the endpoint returns 404 at this point we could make the use of.
But ok, let's create a JIRA

DanielMSchmidt: Created this Jira: https://jira.mesosphere.com/browse/DCOS_OSS-2355

PR #2873, opened 2018-04-04
DCOS-20260: Include id search in task text filter

bstavroulakis:
Include ID search in task text filter.

The neat new search does not work against the id field of the task table.

https://cl.ly/1c0l3H3x3236

**Checklist**

### DIFF
--- a/plugins/services/src/js/filters/TaskNameTextFilter.js
+++ b/plugins/services/src/js/filters/TaskNameTextFilter.js
@@ -23,7 +23,10 @@ class TaskNameTextFilter extends DSLFilter {
    */
   filterApply(resultSet, filterType, filterArguments) {
     return resultSet.filterItems(task => {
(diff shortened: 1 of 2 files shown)


plugins/services/src/js/filters/TaskNameTextFilter.js
```diff
@@ -23,7 +23,10 @@ class TaskNameTextFilter extends DSLFilter {
    */
   filterApply(resultSet, filterType, filterArguments) {
     return resultSet.filterItems(task => {
-      return task.name.indexOf(filterArguments.text) !== -1;
+      return (
+        (task.id && task.id.indexOf(filterArguments.text) !== -1) ||
```
juliangieseke: not blocking, but I'd love us using `includes` instead of `indexOf` in these cases 👍  (same below with `task.name.indexOf`)

PR #2877, opened 2018-04-05
feat(FrameworkConfiguration): show name and version in config tab

ahoskins:
The cosmos `config.json` doesn't have the name and version for frameworks.
This change adds the name and version to the top of this page, so users can easily view
this information without clicking the package to edit.

image 2018-04-06 at 2 56 26 pm

Closes DCOS-19913

**Checklist**

### DIFF
--- a/plugins/services/src/js/containers/framework-configuration/FrameworkConfigurationContainer.js
+++ b/plugins/services/src/js/containers/framework-configuration/FrameworkConfigurationContainer.js
@@ -9,6 +9,7 @@ import FrameworkConfigurationReviewScreen
   from "#SRC/js/components/FrameworkConfigurationReviewScreen";
 import Loader from "#SRC/js/components/Loader";
 import CosmosPackagesStore from "#SRC/js/stores/CosmosPackagesStore";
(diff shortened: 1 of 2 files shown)


ahoskins:
Holding merge till @leemunroe confirms this is fine design for now.

leemunroe:
If we're going to add it here we should make sure it is consistent with the app view design. Currently this is what we have.
image


So I would expect this (without the dropdown caret since it rolling back is not possible in the UI).
image

Brownie points if you make the headings consistent too (`<h2>` >> `<h1>`)

ahoskins:
Thanks for the help @leemunroe. I've updated to be consistent with your suggestion. 👍

DanielMSchmidt:
@ahoskins we can remove hold merge now, right?

ahoskins:
@DanielMSchmidt Yes, I've removed hold-merge. Thanks for review.

juliangieseke:
Jenkins cant checkout again :( @ahoskins to fix this issue, you need to force push something in order to change the HEAD sha on your branch.

PR #2880, opened 2018-04-06
DCOS-21931: GPU dashboard

bstavroulakis:
Add GPU summary in the dashboard.

**Checklist**


src/js/utils/Units.js
```diff
@@ -4,7 +4,7 @@ const Units = {
   formatResource(resource, value) {
     value = Maths.round(value, 2);
 
-    if (resource !== "cpus") {
+    if (resource !== "cpus" && resource !== "gpus") {
```
weblancaster: If I'm not mistaken this should be an or `||` instead of and `&&`

bstavroulakis: In that case if the resource is "cpus" and not "gpus" it will run, which we don't want to.

weblancaster: ok thanks.

DanielMSchmidt:
I retriggered the CI job, it seems like cypress just froze

bstavroulakis:
@DanielMSchmidt @weblancaster If you can approve again, I updated the integration test that was complaining.

DanielMSchmidt:
@bstavroulakis No, but I can review again :trollface:

PR #2884, opened 2018-04-09
DCOS_OSS-2355: show error if service was not found

DanielMSchmidt:
To test this you can visit the edit page of a service and change the name in the URL you should see an error page

**Checklist**

### DIFF
--- a/plugins/services/src/js/components/modals/EditServiceModal.js
+++ b/plugins/services/src/js/components/modals/EditServiceModal.js
@@ -6,15 +6,31 @@ import CreateServiceModal
   from "#PLUGINS/services/src/js/components/modals/CreateServiceModal";
 import DCOSStore from "#SRC/js/stores/DCOSStore";
 import FullScreenModal from "#SRC/js/components/modals/FullScreenModal";
+import RequestErrorMsg from "#SRC/js/components/RequestErrorMsg";
+import Page from "#SRC/js/components/Page";
 
 class EditServiceModal extends Component {
   render() {
     const { id = "/" } = this.props.params;
     const serviceID = decodeURIComponent(id);
+    const serviceLoaded = DCOSStore.serviceDataReceived;
     const service = DCOSStore.serviceTree.findItemById(serviceID);
 
-    if (!service) {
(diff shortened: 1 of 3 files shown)


plugins/services/src/js/components/modals/EditServiceModal.js
```diff
@@ -6,17 +6,33 @@ import CreateServiceModal
...
+    const serviceLoaded = DCOSStore.serviceDataReceived;
     const service = DCOSStore.serviceTree.findItemById(serviceID);
 
-    if (!service) {
+    // Loading, showing an empty modal instead
+    if (!serviceLoaded) {
       return <FullScreenModal open={true} />;
     }
 
+    // Service not found
```
weblancaster: I believe the ideal way is to add a route to the `routes/service.js` with a new not found path/route at the top level and not adding conditions to low level components.

```
{
    type: Route,
    path: "*",
    component: NotFoundPage
},
```

_ps: we actually have a `NotFoundPage` component._

juliangieseke:
Ping @DanielMSchmidt

DanielMSchmidt:
@juliangieseke Just rediscovered this 🙈 fix is pushed, please review :)

juliangieseke:
@DanielMSchmidt do you think the thing mentioned by fabs is an easy win? if so we could try to implement it in this PR, otherwise I'd also suggest a follow up ticket 👍

DanielMSchmidt:
@juliangieseke @Fabs I created an issue as there was no easy win here if I don't want to contradict @weblancaster request for an own route: https://jira.mesosphere.com/browse/DCOS_OSS-2370

DanielMSchmidt:
@weblancaster Can you review this, please 🙏

plugins/services/src/js/components/modals/EditServiceModal.js
nLight: I don't think it's that easy. For that we would have to move the logic whether a service is there or not into the react router. While even though we don't have a service the route is still matchable.

juliangieseke:
Ping @weblancaster

PR #2888, opened 2018-04-09
fix(MesosStateStore): adjust stream data handling

orlandohohmeier:
---
⚠️  _This branch includes changes that will be introduced with #2908. For a list of changes please see: https://github.com/dcos/dcos-ui/compare/orlandohohmeier/master/fix/DCOS-21928-adjust-dsl-input-change-handling...orlandohohmeier/master/fix/DCOS-21784-adjust-mesos-stream-data-handling_ 

---

Use `Promise.resolve().then()` to schedule a micro tasks and asynchronously  trigger `onStreamData` which is used to emit the `MESOS_STATE_CHANGE`. This change will mitigate issues in which errors in view components lead to stale data as the error propergation broke the stream.

N.B. `EventEmitter` is synchronous meaning that every delay in an event handler effects the `emit` execution time and errors propagate up to the `emit` call which sometimes leads to undesireable behavior.

## Testing

Throw an error in one of the `MESOS_STATE_CHANGE` handler – this error should have no effect on the stream.

```js
    // ServicesContainer (L:154)
    MesosStateStore.addChangeListener(MESOS_STATE_CHANGE, function() {
      throw new Error("Test");
    });
 ```

Closes DCOS-21784

### DIFF
--- a/src/js/stores/MesosStateStore.js
+++ b/src/js/stores/MesosStateStore.js
@@ -107,7 +107,10 @@ class MesosStateStore extends GetSetBaseStore {
       .concat(eventTriggerStream)
       .sampleTime(Config.getRefreshRate() * 0.5)
       .retryWhen(linearBackoff(MAX_RETRIES, RETRY_DELAY))
-      .subscribe(this.onStreamData, this.onStreamError);
+      .subscribe(
+        () => Promise.resolve().then(this.onStreamData),
+        this.onStreamError
+      );
   }
 
   getLastMesosState() {


DanielMSchmidt:
I restarted the tests so we can get this merged 👍 Seemed to be the flaky nodes page again

nLight:
Those 3 Nodes Page tests over and over again

ahoskins:
I tried to understand why this is needed, and not sure. Can someone clarify what this is changing?

nLight:
Ok, Node tests are broken. @orlandohohmeier do we disable them or try hard fixing? There's that feeling that our code could be broken actually :D

Fabs:
@ahoskins @bstavroulakis. This solves and error handling problem. On the `subscribe` method of the mesos stream, we pass a function as paramenter to the `onNext` argument that uses the EventStream to send messages to stores and the rest of the system, when we have something new on the mesos stream. The issue is that **EventStream is synchronous** so when there is an error by someone listening to the events, this error propagates to the onNext callback, and breaks the stream. To solve that, we wrap the callback to onNext with a promise, so when the promise fails, it does not bubble the error up to the stream. The breakage of the stream was responsible for all sort of stale data errors in the past, and missing tasks, and etc.

One discussion that we had to brush on while doing this is the concept of microtasks and microtasks on the scheduling callback mechanism of javascript in the browser. For that I recomment this article:
https://jakearchibald.com/2015/tasks-microtasks-queues-and-schedules/

juliangieseke:
@orlandohohmeier the failing tests (NodesPage) are pretty stable - I think in this PR they are legit ones. There is not one single passed run in this PR…

juliangieseke:
trying a rebase…

orlandohohmeier:
@juliangieseke the test fail due to the way we handle state in the respective filter component which when updated at the _wrong_ time results in the respective filtering issues and test failures. As the changes in the PR worsen the problem as they slightly change the time. I'm currently working on a fix to unblock this PR and improve our overall test reliability.

orlandohohmeier:
@DanielMSchmidt I'm happy to add a label but would refrain from moving the JIRA back to "In Progress". The respective test failures are tracked in a separate JIRA and not related to the changes in this PR – the changes introduced with this PR only surface the issue.

DanielMSchmidt:
@orlandohohmeier Maybe I have the wrong definition of "hold merge" then, doesn't it mean it makes no sense to review it right now, because there needs to be additional effort done first?

orlandohohmeier:
@DanielMSchmidt I agree with you definition but it doesn't imply that this PR is still in progress – only that there are there needs to be additional effort done first.

DanielMSchmidt:
@orlandohohmeier For me "additional effort done first" contrasts with my definition of in review ("ready to be reviewed right now"). Let's discuss this with the whole team somewhen else, I think this can spark some opportunities for optimizations

juliangieseke:
@DanielMSchmidt @orlandohohmeier I think the ticket for this PR should be moved to 'blocked' and linked to the one which fixes the issue - _that_ one should be in progress_ 👍

DanielMSchmidt:
@orlandohohmeier @juliangieseke Added it to the sync, let's cut it here :)

juliangieseke:
:trollface:

PR #2890, opened 2018-04-10
fix(MesosStateStore): adjust stream event sampling

orlandohohmeier:
Use `sampleTime` instead of `debounceTime` to reduce the frequency of events. Debounce emits values from the source Observable only after a particular time span has passed *without* another source emission – which in case the source update frequency is higher than the `dueTime` would mean that we don't receive any updates. The `sampleTime` operator emits the last emitted value from the source Observable within time intervals.

Closes DCOS_OSS-2340

### DIFF
--- a/src/js/stores/MesosStateStore.js
+++ b/src/js/stores/MesosStateStore.js
@@ -98,13 +98,14 @@ class MesosStateStore extends GetSetBaseStore {
     // Since we introduced the fake event above, we have to guarantee certain
     // refresh limits to the UI. They are:
     //
-    // MOST once every (Config.getRefreshRate() * 0.5) ms. due to debounceTime.
-    // LEAST once every tick of Config.getRefreshRate() ms in Observable.interval
+    // MOST once every (Config.getRefreshRate() * 0.5) ms. due to sampleTime.
+    // LEAST once every tick of Config.getRefreshRate() ms in
+    // Observable.interval
     //
     // TODO: https://jira.mesosphere.com/browse/DCOS-18277
     this.stream = waitStream
       .concat(eventTriggerStream)
-      .debounceTime(Config.getRefreshRate() * 0.5)
+      .sampleTime(Config.getRefreshRate() * 0.5)
       .retryWhen(linearBackoff(MAX_RETRIES, RETRY_DELAY))
       .subscribe(this.onStreamData, this.onStreamError);
   }


DanielMSchmidt:
@orlandohohmeie Thank you for the explanation. Is there any way I can test this except opening the nodes detail page?

orlandohohmeier:
@DanielMSchmidt you could mock the Mesos stream using the following super simple server which replays recorded messages at a given rate.

```js
const http = require('http');
const port = 3000;
const fs = require('fs');
const recordio = fs.readFileSync('recorded_stream', 'utf8')
  .split('\n')
  .filter(function (record) {
    return record !== "";
  }).map(function (record) {
    return record.length + "\n" + record;
  });

const requestHandler = (request, response) => {
  console.log("Client connected")

  response.setHeader("Access-Control-Allow-Origin", "*")
  response.setHeader("Access-Control-Allow-Headers", "Origin, X-Requested-With, Content-Type, Accept")

  if (request.method === "OPTIONS") {
    response.end("");
    return;
  }

  recordio.map(function (record, index) {
    setTimeout(function () {
      console.log(recordio.length, index)
      response.write(record);
    }, 15000 * index);
  });
}

const server = http.createServer(requestHandler)

server.listen(port, (err) => {
  if (err) {
    return console.log('Something went wrong. ', err)
  }

  console.log(`Server is listening on ${port}.`)
})
```

nLight:
Pipeline haven't even started. @juliangieseke how can I re-trigger that one, please?

juliangieseke:
force push 😢 
https://jenkins.mesosphere.com/service/jenkins/job/frontend/job/dcos-ui-oss-pipeline/job/PR-2890/4/console
this is the equivalent issue to "loosing a PR" in the old Jenkins setup, thankfully we can fix it with force-pushing something.

juliangieseke:
best explanation i was able to find in the interwebs: https://github.com/kubernetes/kubernetes/issues/27462#issuecomment-226388630

nLight:
Jenkins is an open source project :trollface:

juliangieseke:
I have removed the second source from our pipeline job (though i did this already 🤔 ) lets see if it happens again

nLight:
@orlandohohmeier could you please force push to force the Jenkins build? Thanks

juliangieseke:
force pushed again - its running now…

PR #2892, opened 2018-04-10
fix: remove webpack-dashboard

weblancaster:
The reason behind removing webpack-dashboard is because
the log view is small, the log view scroll behaves oddly,
the package info displayed isn't needed for most cases,
Using webpack-dashboard also keeps throwing errors

screen shot 2018-04-10 at 11 02 26 am

**Checklist**


weblancaster:
@juliangieseke let's open another PR with the fix only otherwise this PR will become out the scope. Thanks.

PR #2893, opened 2018-04-11
DCOS-21932: GPU services

bstavroulakis:
Add GPU field in services table

**Checklist**

### DIFF
--- a/plugins/services/src/js/containers/services/__tests__/ServicesTable-test.js
+++ b/plugins/services/src/js/containers/services/__tests__/ServicesTable-test.js
@@ -16,6 +16,7 @@ describe("ServicesTable", function() {
   const healthyService = new Application({
     healthChecks: [{ path: "", protocol: "HTTP" }],
     cpus: 1,
+    gpus: 1,
     instances: 1,
     mem: 2048,
     disk: 0,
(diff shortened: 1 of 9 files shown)


plugins/services/src/js/utils/ServiceTableUtil.js
```diff
@@ -64,6 +64,17 @@ function cpusCompareFunction(a, b) {
   return numberCompareFunction(a.getResources().cpus, b.getResources().cpus);
 }
 
+/**
+ * Compare service gpus
+ * @param {Service|ServiceTree} a
```
DanielMSchmidt: Fun fact, we had an eslint rule in one of my last companies for having a certain minimum length of variables (3 it was). This resulted in comparison variables being named `aaa` or `bbb` :trollface:

plugins/services/src/js/containers/services/__tests__/ServicesTable-test.js
```diff
@@ -95,6 +96,15 @@ describe("ServicesTable", function() {
       expect(ReactDOM.findDOMNode(cpusCell).textContent).toEqual("1");
     });
 
+    it("renders resources/stats gpus property", function() {
+      var gpusCell = ReactDOM.render(
```
DanielMSchmidt: It seems like I overlooked this part while trying to move to enzyme, would you mind moving these tests to use enzyme? A no would also be fine, I would file a jira issue then

weblancaster: I would like to see this moved to enzyme too

nLight:
@juliangieseke Jenkins it's not red but PR is red, do you have an idea why this could happen? 🤔

juliangieseke:
Jenkins is reporting unstable builds as fails to github (see my explanation in slack yesterday night)

juliangieseke:
> Nodes Page -- Nodes grid -- Filters nodes grid -- doesnt display any nodes

That one again…

nLight:
I must have missed the failed test glancing over the test results, sorry. Re-triggered

juliangieseke:
In Jenkins UI its Yellow = Unstable = Flaky :)

PR #2898, opened 2018-04-13
DCOS-22110: CI refactoring Day1 patch

juliangieseke:
I _knew_ it 🤷‍♂️ 

Not exactly sure what causes Cypress to hang sometimes, but it could be related to IO. This PR tries to Limit IO a bit.

If this doesnt help the next Idea is … no idea 😅

(first I thought its hanging every time, but this assumption seems to be false)


DanielMSchmidt:
@juliangieseke where is the associated JIRA ticket? Could you link it here?

juliangieseke:
added the ticket 👍

DanielMSchmidt:
Thank you @juliangieseke <3

PR #2899, opened 2018-04-13
DCOS-22058: change package repositories ADD to the new data-layer patterns

Fabs:
~Depends on #2870, and then we can switch base to `feature/data-layer`. This was a split of that one to make it simpler.~

Moves add repositories away from mixings and flux stores by using the new data-layer patterns (streams and mediators). This is also done in a preparation to use graphql.

TESTING
Just open the Settings > Repositories and try to add, remove and filter packages.
 
AUTOMATED TESTING
No new tests were added since this is a refactor and rely on the current tests.

 **Checklist**
- [x] Explain how to test it

### DIFF
--- a/plugins/catalog/src/js/repositories/RepositoriesAdd.js
+++ b/plugins/catalog/src/js/repositories/RepositoriesAdd.js
@@ -1,0 +1,59 @@
+/* eslint-disable no-unused-vars */
+import React from "react";
+/* eslint-enable no-unused-vars */
+
+import { componentFromStream } from "data-service";
+
+import { Subject } from "rxjs/Subject";
+import "rxjs/add/operator/combineLatest";
+import "rxjs/add/operator/do";
+import "rxjs/observable/empty";
+
+import { Observable } from "rxjs/Observable";
+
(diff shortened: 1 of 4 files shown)


nLight:
I'm afraid you need to rebase the whole chain to make the NodesPage test go away.

plugins/catalog/src/js/repositories/RepositoriesAdd.js
```diff
@@ -0,0 +1,60 @@
...
+// Imported from the Cosmos Store
+const getErrorMessage = (response = {}) => {
+  if (typeof response === "string") {
+    return response;
+  }
+
+  return response.description || response.message || "An error has occurred.";
+};
+
+const addRepositoryEven$ = new Subject();
```
ahoskins: should this say `addRepositoryEvent$`? Not sure what "even" means.

Fabs: you are right, going to fix that typo.

Fabs:
@nlight. Done.

PR #2900, opened 2018-04-14
feat(taskstable): add gpu to task table

bstavroulakis:
Add GPU to task table.

Closes DCOS-21933

To test properly first run a cluster with GPU support. And then create a task that uses GPU.

https://github.com/dcos/dcos-launch/blob/master/dcos_launch/sample_configs/aws-terraform-with-gpus.yaml#L18

**Checklist**

### DIFF
--- a/src/js/stores/MesosStateStore.js
+++ b/src/js/stores/MesosStateStore.js
@@ -81,6 +81,7 @@ class MesosStateStore extends GetSetBaseStore {
       .distinctUntilChanged()
       .map(message => parsers(this.getLastMesosState(), JSON.parse(message)))
       .do(state => {
+        // The tasks received here have not GPU
         CompositeState.addState(state);
         this.setState(state);
       });
--- a/src/js/stores/MesosStateStore.js
+++ b/src/js/stores/MesosStateStore.js
@@ -81,7 +81,6 @@ class MesosStateStore extends GetSetBaseStore {
       .distinctUntilChanged()
       .map(message => parsers(this.getLastMesosState(), JSON.parse(message)))
       .do(state => {
-        // The tasks received here have not GPU
         CompositeState.addState(state);
         this.setState(state);
       });
--- a/plugins/services/src/js/constants/TaskTableHeaderLabels.js
+++ b/plugins/services/src/js/constants/TaskTableHeaderLabels.js
@@ -1,5 +1,6 @@
 var TaskTableHeaderLabels = {
   cpus: "CPU",
+  gpus: "GPU",
   disk: "Disk",
   host: "Host",
   zone: "Zone",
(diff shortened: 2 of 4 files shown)


src/js/stores/MesosStateStore.js
```diff
@@ -81,6 +81,7 @@ class MesosStateStore extends GetSetBaseStore {
       .distinctUntilChanged()
       .map(message => parsers(this.getLastMesosState(), JSON.parse(message)))
       .do(state => {
+        // The tasks received here have not GPU
```
bstavroulakis: @nLight For some reason the tasks coming form Mesos don't have GPU, but only CPU and Mem. Is this something we can fix?

nLight: From the UI point of view if Mesos provides the info it should be there. Can we make sure that mesos provides the info? Unless it does we need an urgent ticket for the Mesos team 💃

ahoskins:
@bstavroulakis What's the status of this? We need to clarify something with mesos or this good?

bstavroulakis:
I'll probably have to create a Mesos ticket

image

src/js/stores/MesosStateStore.js
nLight: Could you please remove the comment?

DanielMSchmidt:
@bstavroulakis You set up a cluster with GPU support, right? Could you share how to do this so that I can review this?

nLight:
Locally those tests pass 🤔

DanielMSchmidt:
@nLight In CI now, too 😀

nLight:
And right align the value please.

nLight:
Also this. Even on full screen MBP 15" I can't see the column if the Sidebar is open. I think we should tweak those columns once again.

screen shot 2018-04-26 at 13 55 59

bstavroulakis:
@nLight Looks like a commit was removed that broke the CSS and the null check. That is fixed now.

bstavroulakis:
@leemunroe What do you believe we should do with Dima's concern? In my screen it looks ok.
@nLight Can you please check with the CSS updates? Now the GPU has a solid width and it will always be visible

image

bstavroulakis:
tasksresponsive

bstavroulakis:
@leemunroe @nLight ^If we are OK with this UI then we are merge this PR and then discuss about the ID column again. 

It will be a long conversation again and I don't want it to keep this PR open forever.

bstavroulakis:
Task columns widths 

taskcolumnsizes

DanielMSchmidt:
@bstavroulakis @leemunroe Can we move the responsiveness discussion into a JIRA ticket to keep track of it and link it here or doesn't that make any sense?

juliangieseke:
Ping @nLight

PR #2902, opened 2018-04-16
Add text filter in nodes

bstavroulakis:
Add text filter in nodes page.

Closes DCOS-22127

**Checklist**

### DIFF
--- a/plugins/nodes/src/js/filters/NodesTextFilter.js
+++ b/plugins/nodes/src/js/filters/NodesTextFilter.js
@@ -1,0 +1,31 @@
+import DSLFilterTypes from "#SRC/js/constants/DSLFilterTypes";
+import DSLFilter from "#SRC/js/structs/DSLFilter";
+
+/**
+ * This filter handles the `text` attributes against service's `getName` value
+ */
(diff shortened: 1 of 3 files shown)


bstavroulakis:
It is that way to be extendable in the future if needed. By free text we want to include/remove fields such as title, id e.t.c. If it was NameTextSearch, it would be always tied to the name.

plugins/nodes/src/js/filters/NodesTextFilter.js
```diff
@@ -0,0 +1,33 @@
...
+  }
+
+  /**
+   * Keep only nodes whose name has text
+   *
+   * @override
+   */
+  filterApply(resultset, filterType, filterArguments) {
+    return resultset.filterItems(node => {
+      const hostName = node.getHostName();
```
ahoskins: We only want to filter by the `hostName`? Would be cool to have it do a weighted filter of all columns.

bstavroulakis: This could happen in the future if needed. Let us add the filter for now and extend when needed.

PR #2911, opened 2018-04-20
fix(DSLInputField): adjust change event handling

orlandohohmeier:
---
↩️  _This PR back-ports a fix to release/1.11 introduced with #2908._

---

> Remove the change event debouncing to resolve issues in which we lost the filter expression data.
> 
> The component was designed to be stateless meaning that the hosting component is responsible for updating the expression property. Debouncing the event potentially leads to data loss as the hosting component might overwrite the expression before the change event was fired.
> 
> Closes DCOS-21928


ahoskins:
run integration tests

ahoskins:
run integration tests

PR #2905, opened 2018-04-17
DCOS-22174: Fix CI Pull Request Creation

juliangieseke:
PR creation failed due to a bad multiline bash command… :( this should fix that issue and also automatically adds a comment to set the created PR to "ready for review".
Also I added some minor fixes I stumbled upon while working in the files…

PR #2912, opened 2018-04-21
DCOS-21926: Fix jenkins (plugins)

juliangieseke:
⚠️ This PR depends on #2905 

---

some changes to unblock & integrate plugins setup - more info over there (see referencing PR)

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -186,4 +186,33 @@ pipeline {
       }
     }
   }
+
+  post {
+    failure {
+      withCredentials([
+        string(credentialsId: '8b793652-f26a-422f-a9ba-0d1e47eb9d89', variable: 'SLACK_TOKEN')
+      ]) {
+        slackSend (
+          channel: '#frontend-dev',
+          color: 'danger',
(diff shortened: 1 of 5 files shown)


Jenkinsfile
```diff
@@ -174,5 +174,45 @@ pipeline {
...
+          release_branches.contains(BRANCH_NAME) && params.CREATE_RELEASE == false
+        }
+      }
+      steps {
+        build job: "frontend/dcos-ui-ee-pipeline/" + env.BRANCH_NAME.replaceAll("/", "%2F"), wait: false, propagate: false
+      }
+    }
+  }
+
+  post {
```
DanielMSchmidt: I think this should really have been a seperate ticket. I am okay with merging this right now, but please open a separate ticket for these bigger changes. This has nothing to do with "wrong paths in some scripts", nor does triggering the enterprise pipeline. I don't want to stay in the way of progress and I know how much work went into this solution, so please let's not get into this situation the next time.

juliangieseke:
we're going to rethink the whole 'automatic release' thing, DC/OS is going to change the way we have to do our bumps… 

I'll create a new PR wrapping up the CI optimizations without release stuff, from that point I'll start over migrating the ci/cd stuff to the new DC/OS way…

PR #2913, opened 2018-04-22
DCOS_OSS-2059 Update Config Headings

lilysamimi:
Problem: Configuration headings for SDK and Marathon apps are inconsistent. 

1. "Services" heading with underline

Marathon
screen shot 2018-04-20 at 2 41 46 pm

SDK
image

2. "Configuration" label on review screen 

Marathon
screen shot 2018-04-22 at 10 49 09 am

SDK
image

Solution:

1. Made Services heading the same 
2. Got rid of the Configuration title on SDK confirmation page 

**Checklist**

### DIFF
--- a/src/js/components/HashMapDisplay.js
+++ b/src/js/components/HashMapDisplay.js
@@ -47,7 +47,7 @@ class HashMapDisplay extends React.Component {
       ) {
         // Increase the heading level for each nested description list, making
         // ensuring we don't surpass heading level 6.
-        const nextHeadingLevel = Math.min(headingLevel, 6);
+        const nextHeadingLevel = Math.min(headingLevel + 1, 6);
 
         return (
           <HashMapDisplay
--- a/src/js/components/HashMapDisplay.js
+++ b/src/js/components/HashMapDisplay.js
@@ -47,7 +47,7 @@ class HashMapDisplay extends React.Component {
       ) {
         // Increase the heading level for each nested description list, making
         // ensuring we don't surpass heading level 6.
-        const nextHeadingLevel = Math.min(headingLevel + 1, 6);
+        const nextHeadingLevel = Math.min(headingLevel, 6);
 
         return (
           <HashMapDisplay
--- a/src/js/components/FrameworkConfiguration.js
+++ b/src/js/components/FrameworkConfiguration.js
@@ -255,7 +255,6 @@ class FrameworkConfiguration extends Component {
           {defaultConfigWarningMessage}
           <FrameworkConfigurationReviewScreen
             frameworkData={formData}
-            title={"Configuration"}
             onEditClick={this.handleEditConfigurationButtonClick}
           />
         </div>
--- a/src/js/components/FrameworkConfigurationReviewScreen.js
+++ b/src/js/components/FrameworkConfigurationReviewScreen.js
@@ -150,6 +150,7 @@ class FrameworkConfigurationReviewScreen extends React.Component {
         </div>
         <HashMapDisplay
           hash={frameworkData}
+          headingLevel={0}
           renderKeys={this.getHashMapRenderKeys(frameworkData)}
           headlineClassName={"text-capitalize"}
           emptyValue={EmptyStates.CONFIG_VALUE}


src/js/components/HashMapDisplay.js
```diff
@@ -47,7 +47,7 @@ class HashMapDisplay extends React.Component {
       ) {
         // Increase the heading level for each nested description list, making
         // ensuring we don't surpass heading level 6.
-        const nextHeadingLevel = Math.min(headingLevel + 1, 6);
+        const nextHeadingLevel = Math.min(headingLevel, 6);
```
juliangieseke: just from reading the code here it sounds like we want to increase the level… if something fundamental has changed we should try to refactor this part in a way, that the variable names are still correct.

leemunroe: @lilysamimi @juliangieseke My bad this didn't work as expected (I suggested this change).

The goal here is to ensure the first heading is a `<h1>` so it aligns with the other configuration views. Due to the default heading level being set to `1` and the `nextHeadingLevel` always plussing 1, the headings where starting at `<h2>`.

@juliangieseke You might have a better idea how to fix this, but I believe if we add the `+ 1` back and change line 109 to `headingLevel: 0,` this will achieve what we're looking for.

image

juliangieseke: @lilysamimi @leemunroe no, if we change the default value to 0, it could break something else. also we most probably dont want `0` as a valid value, because there is no `<h0>` element. for this reason it is limited by line 115: `headingLevel: PropTypes.oneOf([1, 2, 3, 4, 5, 6]),`. 

the `HashMapDisplay` Component is fine and there is not change needed, you need to find the place where it is called and reduce the level there. I'm not super familiar with the forms and configuration screens, but i doubt its a one-line-change to fix this issue :(

nLight: I'd second @juliangieseke it's rather `ServiceGeneralConfigSection.js` should be refactored to use `HashMapDisplay`, or, as a shortcut, we could replace `h1` in `ServiceGeneralConfigSection` and adjust css class names.

leemunroe: I'm not following. If we want the package configuration headings to be `H1 > H2 > H3` instead of today where they are `H2 > H3 > H4` how would updating `ServiceGeneralConfigSection` help us do that?

leemunroe: We could pass `headingLevel={0}` from `FrameworkConfigurationReviewScreen.js` which also achieves what we want and doesn't touch HashMapDisplay.

```
<HashMapDisplay
  hash={frameworkData}
  headingLevel={0}
  renderKeys={this.getHashMapRenderKeys(frameworkData)}
  headlineClassName={"text-capitalize"}
  emptyValue={EmptyStates.CONFIG_VALUE}
/>
```

image

juliangieseke: After looking into the code I dont think allowing something like `headingLevel=0` will solve the issue, its more like a hacky workaround. :(

A better solution is to change `FrameworkConfigurationReviewScreen#render` so that it doesnt pass down `frameworkData` to one HashMapDisplay, but all first-level children to its own `HashMapDisplay` instance. This is most probably nothing you want do, so I'd suggest talking to @orlandohohmeier to prioritize this. 🤔

ahoskins: image 2018-04-27 at 2 25 08 pm

lilysamimi: Sounds good @juliangieseke! I'll let you guys change FrameworkConfigurationReviewScreen#render. In the meantime, here's the workaround that gets the hierarchy that we want so it's consistent the app config. I've copy and pasted the screenshots above in the first comment.

ahoskins: We definitely have to increment `headingLevel` here, because the above issue screenshot here is the current state, which is not good.

lilysamimi:
Sounds good @juliangieseke! I'll let you guys change FrameworkConfigurationReviewScreen#render. In the meantime, here's the workaround that gets the hierarchy that we want so it's consistent the app config. I've copy and pasted the screenshots above in the first comment.

juliangieseke:
I'm fine with this if we create a JIRA to fix this properly and linking it in the code (add a comment in `FrameworkConfigurationReviewScreen.js @ Line 153`). 
@lilysamimi can you do this please? 
@weblancaster @ahoskins would that be okay for you?

PR #2915, opened 2018-04-23
 DCOS_OSS-2397: shows filter default on task page loading

Fabs:
Not having new zones or regions prevented the state
of the component to be updated causing the default filter
to load with an empty value.

**Checklist**


Fabs:
@nLight Thanks :D. Hope to churn more of those from time to time :). Thanks for providing the hardware.

juliangieseke:
do we want to have an integration test for this? :)

nLight:
@juliangieseke for me having unit is enough

PR #2919, opened 2018-04-24
fix(MesosStream): delay reconnections

orlandohohmeier:
Delay reconnection attempts to reduce load on Mesos and improve the UI performance. Without this adjustment, the UI will attempt to re-connect immediately which results in a rather high number of connection attempts which leads to a high CPU load if Mesos is unable to establish the stream (XHR).

Closes DCOS-22303


juliangieseke:
integration tests are green 🎉

nLight:
@juliangieseke the same error I saw on Bill's PR, but locally it doesn't happen

juliangieseke:
yup, looks like its an issue on master - we're on it.

PR #2925, opened 2018-04-24
fix(package): fix compression version

bstavroulakis:
The latest version of compression-webpack-plugin doesn't support webpack 1.x

I have specified the last version that supports webpack 1.x

**Checklist**


bstavroulakis:
Chose the least invasive change for the 1.9 back-port fix. Not sure why it was done this way in the first place.

PR #2926, opened 2018-04-25
DCOS-22327 feat: add support to mutations

Fabs:
Our homemade graphql rxjs implementation only supported queries, we have added now also the support for mutations. 

To test it:
> npm test -- --watch graphql query

and then play with the provided test.

NOTES
There is a follow up PR addressing issues regarding graphql spec compliance. That PR will also bring up many tests from the graphql-js lib.

**Checklist**

### DIFF
--- a/packages/data-service/src/__tests__/graphqlObservable-test.js
+++ b/packages/data-service/src/__tests__/graphqlObservable-test.js
@@ -49,10 +49,16 @@ const mockResolvers = {
   },
   Mutation: {
     createShuttle: (parent, args, ctx) => {
-      return ctx.mutation;
+      return ctx.mutation.map(() => ({
+        name: args.name
+      }));
     },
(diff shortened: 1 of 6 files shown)


packages/data-service/src/__tests__/graphqlObservable-test.js
```diff
@@ -218,4 +231,89 @@ describe("graphqlObservable", function() {
...
+          }
+        }
+      `;
+
+      const fakeRequest = { name: "RocketShip" };
+      const commandContext = Observable.of(fakeRequest);
+
+      const result = graphqlObservable(mutation, schema, {
+        mutation: commandContext,
+        name: "RocketShip"
```
nLight: why do we need to put it twice?

Fabs: Done!

PR #2927, opened 2018-04-25
DCOS-22332 Graphql spec compliance for field values and aliases

Fabs:
This PR ports a group of tests form graphql-js wrapping them into Observable so they are compatible with our implementation. We believe that by making sure the reference implementation tests pass on our implementation is a great first step.

We also refactored part of the code to be easier to understand.

HOW DO I REVIEW?

My suggestion would be to run checkout and run `npm test -- --watch graphql star`. You can write more test cases or modify the current if you can think of a case where it would break. You can also come up with queries by playing on any query explorer (like the github one on notes), and trying to map then to the starwars example.

NOTES:
When testing manually, we are validating our queries structure here: https://developer.github.com/v4/explorer/

The graphql specs:
https://facebook.github.io/graphql/October2016/#sec-Language.Query-Document

**Checklist**

### DIFF
--- a/packages/data-service/src/__tests__/reference/starWarsData.js
+++ b/packages/data-service/src/__tests__/reference/starWarsData.js
@@ -95,7 +95,7 @@ const droidData = {
  */
 function getCharacter(id) {
   // Returning a promise just to illustrate GraphQL.js's support.
-  return humanData[id] || droidData[id];
+  return Promise.resolve(humanData[id] || droidData[id]);
 }
 
 /*
--- a/packages/data-service/src/__tests__/reference/starWarsData.js
+++ b/packages/data-service/src/__tests__/reference/starWarsData.js
@@ -1,0 +1,134 @@
+// https://github.com/graphql/graphql-js/blob/master/src/__tests__/starWarsData.js
+/**
+ * Copyright (c) 2015-present, Facebook, Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow strict
+ */
+
+/**
+ * This defines a basic set of data for our Star Wars Schema.
+ *
+ * This data is hard coded for the sake of the demo, but you could imagine
+ * fetching this data from a backend service rather than from hardcoded
+ * JSON objects in a more complex demo.
+ */
+
+const luke = {
+  type: "Human",
+  id: "1000",
+  name: "Luke Skywalker",
+  friends: ["1002", "1003", "2000", "2001"],
+  appearsIn: [4, 5, 6],
+  homePlanet: "Tatooine"
+};
+
+const vader = {
+  type: "Human",
+  id: "1001",
+  name: "Darth Vader",
+  friends: ["1004"],
+  appearsIn: [4, 5, 6],
+  homePlanet: "Tatooine"
+};
+
+const han = {
+  type: "Human",
+  id: "1002",
+  name: "Han Solo",
+  friends: ["1000", "1003", "2001"],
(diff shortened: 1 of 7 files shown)


packages/data-service/src/MINIROADMAP
```diff
@@ -0,0 +1,5 @@
+// "graphqlObservable error: kind ${definition.kind} is not supported"
```
DanielMSchmidt: This should not be commited, right?

packages/data-service/src/__tests__/reference/starWarsData.js
```diff
@@ -0,0 +1,133 @@
+// https://github.com/graphql/graphql-js/blob/master/src/__tests__/starWarsData.js
+/**
```
DanielMSchmidt: Nice that you could port them over

packages/data-service/src/graphqlObservable.js
```diff
@@ -44,21 +45,53 @@ function resolveStep(typeMap, definition, context, parent) {
...
+    const resolvedObservable = resolver.resolve(
       parent,
       args,
       context,
       null // that would be the info
     );
 
+    if (!resolvedObservable) {
+      return Observable.throw(
+        new Error("graphqlObservable error: empty resolution")
```
DanielMSchmidt: I wonder if a custom error with that prefix might be nicer here 🤔

packages/data-service/src/graphqlObservable.js
```diff
@@ -44,21 +45,53 @@ function resolveStep(typeMap, definition, context, parent) {
...
+
     return resolvedObservable.map(emittedResults => {
+      if (!emittedResults) {
+        return Observable.throw(
+          new Error("graphqlObservable error: result not emitted")
+        );
+      }
+
+      // Make the concat map appear
+      // const resultStream = !!emittedResults.map ? emittedResults : Observable.of(resultStream);
```
DanielMSchmidt: Forgotten?

ahoskins:
@Fabs Can you clarify why we are having a PR like this adding star wars related stuff? 😏

Fabs:
sure @ahoskins. This PR is work in progress, so I have not yet cared writing a good description.

There is a reference implementation to graphql, graphql-js, and on this implementation, they have a bunch of tests that use this star wars fake data. Here, we are copying their tests and adapting to our implementation to get to a good test coverage, with little effort.

packages/data-service/src/graphqlObservable.js
```diff
@@ -2,100 +2,149 @@ import { Observable } from "rxjs/Observable";
 import "rxjs/add/observable/throw";
 import "rxjs/add/observable/of";
 import "rxjs/add/operator/combineLatest";
+import "rxjs/add/operator/do";
 import "rxjs/add/operator/map";
+import "rxjs/add/operator/concatMap";
 
-// TODO: refactor this once it has a more comprehensive implementation
-// of the graphql api
+// TODO: add types with typescript
```
nLight: Is there JIRA for that?

packages/data-service/src/graphqlObservable.js
```diff
@@ -2,100 +2,149 @@ import { Observable } from "rxjs/Observable";
...
+  }
+
+  if (!(resolvedObservable instanceof Observable)) {
+    return throwObservable("resolver does not return an observable");
   }
 
-  return doc.definitions.reduce((acc, definition) => {
-    const resolvedObservable = resolveStep(
-      schema._typeMap,
+  return resolvedObservable.concatMap(emitted => {
```
nLight: wonder if there's a convention we could pick from Rx code to name `emitted` better.

PR #2930, opened 2018-04-26
DCOS-22349: add gpu in nodes table

bstavroulakis:
Add GPU in nodes table.

Closes DCOS-22349

**Checklist**

### DIFF
--- a/plugins/nodes/src/js/components/NodesTable.js
+++ b/plugins/nodes/src/js/components/NodesTable.js
@@ -16,7 +16,8 @@ import UnitHealthUtil from "#SRC/js/utils/UnitHealthUtil";
 const COLOR_CLASSNAMES = {
   cpus: "color-1",
   mem: "color-2",
-  disk: "color-3"
+  disk: "color-3",
+  gpus: "color-1"
 };
 
 class NodesTable extends PureComponent {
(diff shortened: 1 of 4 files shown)


DanielMSchmidt:
@juliangieseke someone did 🤷‍♂️

bstavroulakis:
@juliangieseke @DanielMSchmidt I added the ticks. Sorry about that.

plugins/nodes/src/js/components/NodesTable.js
```diff
@@ -16,7 +16,8 @@ import UnitHealthUtil from "#SRC/js/utils/UnitHealthUtil";
 const COLOR_CLASSNAMES = {
   cpus: "color-1",
   mem: "color-2",
-  disk: "color-3"
+  disk: "color-3",
+  gpus: "color-1"
 };
```
leemunroe: @bstavroulakis I believe the intent here was to align the colors so each resource is the same color and they would be green/yellow/red depending on the percentage. Looking at this code I'm guessing that is not the case.

cc @lilysamimi

bstavroulakis:
@ahoskins I follow the same pattern as we already had. Made no assumptions for style changes

image

plugins/nodes/src/js/components/NodesTable.js
bstavroulakis: Hm, this seems not to be the case, the code assigns each resource with a fixed color value

lilysamimi:
Head up @bstavroulakis I'm waiting on further clarification from @nLight and @orlandohohmeier on this.

DanielMSchmidt:
I think the spacing is not right yet, but let's fix that forward maybe ^^ 

gpu

DanielMSchmidt:
I think the spacing is not right yet, but let's fix that forward maybe ^^

DanielMSchmidt:
Added hold merge due to @lilysamimi comment

PR #2931, opened 2018-04-27
DCOS-22338: cleanup Jenkinsfile, remove release parts

juliangieseke:
This PR cleans up our `Jenkinsfile` and reverts most of the release related stuff.

p.s. sorry for this being one big commit, otherwise I'd have to cherrypick from over 50 commits… 😶


Jenkinsfile
```diff
@@ -1,119 +1,75 @@
 #!/usr/bin/env groovy
 
-@Library('sec_ci_libs@v2-latest') _
+@Library("sec_ci_libs@v2-latest") _
```
DanielMSchmidt: It's not blocking for me, but could we have these style changes in a seperate commit next time? They are confusing to review

Jenkinsfile
```diff
@@ -146,32 +102,48 @@ pipeline {
...
-
-          sh "CREATE_RELEASE=1 PUSH_RELEASE=1 ./scripts/ci/bump-version"
+        build job: "frontend/dcos-ui-ee-pipeline/" + env.BRANCH_NAME.replaceAll("/", "%2F")
+      }
+    }
+  }
 
-          sh "./scripts/ci/release-version"
-        }
+  post {
```
DanielMSchmidt: These changes are not really in the same scope as "Removing release parts". Can we please add them in a separate commit / PR / Ticket and not scope creep here? Don't get me wrong, I like the change in general, but we are not showing our real performance if we do stuff like this "just on the side". Sorry for being picky here 🙈

juliangieseke: This PR is about wrapping up my work done in the `feature/ci` branch. not _only_ about removing release stuff. so no scope creep here 👍

Jenkinsfile
juliangieseke: You're right, but please consider my comment in the PR description, I created a new branch from master and copied over the Jenkinsfile - then deleted the release related parts.

Jenkinsfile
DanielMSchmidt: Okay, I see the slack send in the Jira body

Jenkinsfile
DanielMSchmidt: I think workflow issues shouldn't reflect on the commit history when it comes to a PR. But that might be my very own point of view on this. IMHO typing is not the bottlneck in this work here, so there could have been other approaches. Just pointing this out 😇

Jenkinsfile
juliangieseke: yeah, because I added it :trollface:

DanielMSchmidt: Is this how we work now? 😕

juliangieseke: I forgot it in the first place, you've been picky, I've added it… I guess thats a "yes"?

DanielMSchmidt: Detail and scope are very different things, we should make clear we don't mix them here

juliangieseke: Thats what I'm talking about. The *scope* of this ticket is to wrap up the work done on `feature/ci` - while creating this ticket, I forgot to add a *detail* and might have written a bad description, but there was no scope creep.

DanielMSchmidt: Okay, now I got you 👍 So the main problem was the ambiguity of the ticket itself, because "cleaning up" is very vague, got it 👍 Let's make sure we make clearer tickets next time, I will have an extra focused view on that next sprints :)

juliangieseke: I'm more then happy about any help I can get with cleaning up our old Epics and moving them under https://jira.mesosphere.com/browse/PRODUCT-1000 👍 just ping me, if you want to help.

PR #2935, opened 2018-04-29
DCOS-22443: guard host.framework_ids before accessing it

juliangieseke:
This should fix our NodePage flaky test… There was a Stacktrace all the time btw :(
https://jenkins.mesosphere.com/service/jenkins/blue/organizations/jenkins/frontend%2Fdcos-ui-oss-pipeline/detail/PR-2930/5/tests/


juliangieseke:
If there’s a util for doing this we should use it - feel free to block this PR

PR #2943, opened 2018-05-04
Fix configuration ui for frameworks that are not cosmos apps (master) (do not merge)

justinrlee:
See [COPS-3203](https://jira.mesosphere.com/browse/COPS-3203)

If you have a Marathon app that is a framework (has `DCOS_PACKAGE_FRAMEWORK_NAME` to pick up child tasks) but is not a Cosmos app (doesn't have all other accompanying `DCOS_PACKAGE_X` labels), if you go to the DC/OS UI and try to view/edit configuration, you'll start getting 500s.

This PR makes the detection of a Cosmos app more explicit; rather than just checking for DCOS_PACKAGE_FRAMEWORK_NAME, it also checks for `DCOS_PACKAGE_NAME` and `DCOS_PACKAGE_VERSION`

**Checklist**
- [?] Did you add new unit tests?
- [?] Did you add new integration tests?

Modified existing test, added a new test.  Not sure if it's an integration or unit test, but it's similar format to the existing tests and should help with this in the future.

### DIFF
--- a/plugins/services/src/js/structs/__tests__/ServiceTree-test.js
+++ b/plugins/services/src/js/structs/__tests__/ServiceTree-test.js
@@ -24,7 +24,9 @@ describe("ServiceTree", function() {
           {
             id: "/group/beta",
             labels: {
-              DCOS_PACKAGE_FRAMEWORK_NAME: "beta"
+              DCOS_PACKAGE_FRAMEWORK_NAME: "beta",
+              DCOS_PACKAGE_NAME: "beta",
+              DCOS_PACKAGE_VERSION: "beta"
             }
           },
           {
@@ -126,7 +128,9 @@ describe("ServiceTree", function() {
           {
             id: "/group/beta",
             labels: {
-              DCOS_PACKAGE_FRAMEWORK_NAME: "beta"
+              DCOS_PACKAGE_FRAMEWORK_NAME: "beta",
+              DCOS_PACKAGE_NAME: "beta",
+              DCOS_PACKAGE_VERSION: "beta"
             }
           },
           {
--- a/plugins/services/src/js/utils/ServiceValidatorUtil.js
+++ b/plugins/services/src/js/utils/ServiceValidatorUtil.js
@@ -72,7 +72,7 @@ const ServiceValidatorUtil = {
     // Check the DCOS_PACKAGE_FRAMEWORK_NAME label to determine if the item
     // should be converted to an Application or Framework instance.
     // It is possible to DCOS_PACKAGE_FRAMEWORK_NAME without being a Cosmos package,
-    // for example, with Marathon-on-Marathon EE, which is intentionally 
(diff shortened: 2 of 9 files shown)


mesosphere-mergebot-dev:
```
@mesosphere-mergebot ship-it  
@mesosphere-mergebot test  
```

ahoskins:
Hey @justinrlee thanks for the PR! I'm surprised this issue is still happening. About a month ago I changed somewhere else to check for [different labels](https://github.com/dcos/dcos-ui/commit/13ab6be91e08282b1e55a7d84b90456fa8960636#diff-bac56c72345501a3f2fe561e3a07f496). I thought this fixed the "edit" view for frameworks.

justinrlee:
Good point @ahoskins.  I was initially thinking there might be packages that don't have the right metadata, but I guess the cosmos api wouldn't be able to handle it with what I've provided either - this PR was an initial move.  But I may spend some time fixing it to use the two labels you've indicated.

Not sure if you wanna discuss this here or in the JIRA, but I basically see four types of apps:

1. Standard Marathon apps (not Cosmos, not Frameworks) - should behave as regular apps (config/edit displays the marathon json directly)
2. Marathon apps that are also frameworks, but not from Cosmos (are not packages) - should behave as regular apps, but also present the list of child framework tasks
3. Cosmos apps (from Universe packages) that are not frameworks - should behave as Cosmos apps (config/edit displays the config.json, parsed out)
4. Cosmos apps (from Universe packages) that are frameworks - should behave as Cosmos apps, and should additionally present child framework tasks.

I believe (but am not 100% sure) that the UI as currently built handles 1, 2, and 4 properly, but it does not cover case 2.  Case 2 is admittedly probably not super common, but there are two situations where we'll see it:
* Marathon on Marathon Enterprise (which is not available in Universe)
* Older packages that were installed prior to the various Cosmos labels started getting generated (packages installed on 1.9 and below) - this has been problematic with at least one of my customers.

Based on what you mentioned, I'm thinking we should operate as follows;
* Detect Cosmos vs. Marathon apps via the two indicated labels (`DCOS_PACKAGE_DEFINITION` and `DCOS_PACKAGE_DEFINITION`)
* Detect Mesos framework vs. non-Mesos-framework via `DCOS_PACKAGE_FRAMEWORK_NAME` (ideally, we'd have an independent label here that is specifically just `FRAMEWORK_NAME` or something else, but that would require rework in at least 2-3 other areas.
* The two detections above should be handled separately.

I'll have to take another look through the code to see if we can separate out the detection mechanisms.

This is my first time fiddling with the UI, so open to any feedback.

nLight:
For now I'm labeling the PRs as WIP. Please keep us updated @justinrlee and thanks for taking care!

justinrlee:
Gonna close this in favor of https://jira.mesosphere.com/browse/DCOS-38447, which will need a better engineering design.

PR #2946, opened 2018-05-08
DCOS-28735: Adjust Jest setup to work with Typescript

nLight:
While working on https://jira.mesosphere.com/browse/DCOS-21930 I spend days trying to fix our Jest setup. 

First, I had to enable Typescript processing. When that was done it turned out that Typescript doesn't like when we use `setMock`, so I removed all usages that don't make sense and replaced other with `mock` which is a recommended way. 

`unMockStores` uses `setMock` under the hood and is going to be removed. First, it doesn't in fact unmock. It actually _sets_ a mock and mocks stores with an empty object. `¯\_(ツ)_/¯ `

**Dependencies**
- Please merge this first: https://github.com/mesosphere/dcos-ui-plugins-private/pull/794

**Checklist**

Closes DCOS-28735


mesosphere-mergebot-dev:
```
@mesosphere-mergebot ship-it  
@mesosphere-mergebot test  
```

jest/preprocessor.js
```diff
@@ -21,8 +23,16 @@ module.exports = {
       if (isJISON) {
         src = new jison.Generator(src).generate();
       }
+      if (filename.endsWith(".ts") || filename.endsWith(".tsx")) {
```
DanielMSchmidt: Is there anything that holds us back from using a separate transform statement? [It should be possible like this](https://github.com/mesosphere/mockserver/blob/master/package.json#L58).

For me this is a lot cleaner and I would like to move away from this file as it has some more complexity

DanielMSchmidt: If it's about our babel file not being picked up, they have docs on that: https://github.com/kulshekhar/ts-jest#using-a-custom-babel-config

jest/setupEnv.js
```diff
@@ -1,4 +1,5 @@
 require("babel-polyfill");
+require("reflect-metadata");
```
DanielMSchmidt: This is a fragment of the plugins POC, right?

nLight: It actually should be there already since we're using inversify.

jest/preprocessor.js
nLight: As discussed: I decided to extract jest config refactoring from this piece of work.

PR #2949, opened 2018-05-15
fix(NestedServiceLinks): group link

weblancaster:
Fix group link when searching/filtering by name
on services.

Closes DCOS-28721

**Before**
filter-group

**After**
screen recording 2018-05-15 at 11 12 am

**Checklist**
- [ ] Did you add a JIRA issue in a commit message or as part of the branch name?
- [ ] Did you add new unit tests?
- [ ] Did you add new integration tests?
- [ ] If this is a regression, did you write a test to catch this in the future?

PR #2950, opened 2018-05-15
[Backport 1.11] fix(NestedServiceLinks): group link

weblancaster:
Fix group link when searching/filtering by name
on services.

Closes DCOS-28721

**Before**
filter-group

**After**
screen recording 2018-05-15 at 11 12 am

**Checklist**

### DIFF
--- a/src/js/components/NestedServiceLinks.js
+++ b/src/js/components/NestedServiceLinks.js
@@ -6,13 +6,13 @@ import React from "react";
 import Icon from "./Icon";
 
 class NestedServiceLinks extends React.Component {
-  getMinorLink(label, params, key, minorLinkClasses, minorLinkAnchorClasses) {
+  getMinorLink(label, id, key, minorLinkClasses, minorLinkAnchorClasses) {
     return (
       <div key={key} className="table-cell-value">
         <div className={minorLinkClasses}>
           <Link
             className={minorLinkAnchorClasses}
-            to={`/services/detail/${params.id}`}
+            to={`/services/overview/${id}`}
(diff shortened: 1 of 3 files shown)


src/js/components/NestedServiceLinks.js
```diff
@@ -46,11 +46,8 @@ class NestedServiceLinks extends React.Component {
       minorLinkAnchorClassName
     );
     let nestedGroups = this.getServicePathParts();
-    let popCount = 1;
+    const popCount = 1;
```
ahoskins: This is a confusing variable name, and used only once, so maybe just inline the `1`.

src/js/components/NestedServiceLinks.js
```diff
@@ -175,7 +160,6 @@ const classPropType = PropTypes.oneOfType([
 
 NestedServiceLinks.propTypes = {
```
ahoskins: You forgot to add `serviceLink` to propTypes here.

juliangieseke:
hm, this should be a subtask of #2949 - shouldnt it? 🤔Also adding WIP label.

src/js/components/NestedServiceLinks.js
weblancaster: agree.

src/js/components/NestedServiceLinks.js
weblancaster: good catch

weblancaster:
@DanielMSchmidt Sorry but I didn't get what you mean.

PR #2952, opened 2018-05-16
DCOS-35757: remove 2 seconds timeout

nLight:
Everything appears to be working except our dashboard looks interesting :)

screen shot 2018-05-16 at 11 48 39

## Testing
Connect to a i.e. soak 1.10 and switch network to Slow 3G. Without the fix UI won't be able to ever load state.json

**Checklist**
- [x] Did you add a JIRA issue in a commit message or as part of the branch name?
- [ ] Did you add new unit tests?
- [ ] Did you add new integration tests?
- [ ] If this is a regression, did you write a test to catch this in the future?

Closes DCOS-35757

PR #2953, opened 2018-05-16
feat(mesos-state-store): throttle state fetching

nLight:
This PR replaces timer based polling with a throttled sequential polling.
The UI will poll Mesos state endpoint after the previous call has finished but max once in `Config.getRefreshRate()` seconds.

Along with #2952 it will enable the UI to function on bigger clusters and slower connections.

...and it fixes the dashboard :)

screen shot 2018-05-16 at 12 38 37

**Checklist**

Closes DCOS-35759


mesosphere-mergebot-dev:
```
@mesosphere-mergebot ship-it  
@mesosphere-mergebot test  
```

juliangieseke:
<IMHO> the base for this should be `release/1.10` instead of the other PR (and it should be rebased to that base). Its fine if this is red on CI until the other one is merged and for review we can rebase locally. </IMHO>

nLight:
@juliangieseke it doesn't really make much sense without the first one.

juliangieseke:
I know, you can add a notice( '⚠️Depends on #2952' something similar)… we had the situation where such a PR got merged into the first pr-branch after that one got merged into master already… just trying to avoid this in future 👍

PR #2954, opened 2018-05-16
DCOS-37423: improve performance of composite store nodes list getter

DanielMSchmidt:
**Checklist**

**How to review**

- Check that we did not break the tasks list by accident (or anything else using the composite state store (e.g. Nodes View))
- Add `console.count` to the getter and see the amount of calls before and after the change

### DIFF
--- a/src/js/structs/CompositeState.js
+++ b/src/js/structs/CompositeState.js
@@ -83,7 +83,10 @@ class CompositeState {
       return Object.assign({}, oldDatum, matchedNode);
     });
 
-    this.data.slaves = newData;
+    this.data = {
+      ...this.data,
+      slaves: newData
+    };
   }
 
   addState(data) {
@@ -100,8 +103,21 @@ class CompositeState {
     });
   }
 
+  set data(newData) {
+    this._data = newData;
+    this._nodesList = null;
+  }
+
+  get data() {
+    return this._data;
+  }
+
   getNodesList() {
-    return new NodesList({ items: this.data.slaves });
+    if (this._nodesList === null) {
+      this._nodesList = new NodesList({ items: this.data.slaves });
+    }
+
+    return this._nodesList;
   }
 }
 


mesosphere-mergebot-dev:
```
@mesosphere-mergebot ship-it  
@mesosphere-mergebot test  
```

src/js/structs/CompositeState.js
```diff
@@ -47,6 +47,7 @@ const mergeObjectsById = function(newData, data = []) {
 class CompositeState {
   constructor(data = {}) {
     this.data = data;
+    this.updateNodesList();
```
juliangieseke: hm. this could be archived via `setter` instead of manually calling it everytime its changed… 🤔(should be less error prone)

DanielMSchmidt:
@orlandohohmeier @juliangieseke Could you take another look. I think it is a bit messier, but I understand that we won't forget about adding some of the update calls in the future.

src/js/structs/CompositeState.js
```diff
@@ -100,8 +102,20 @@ class CompositeState {
...
+  set data(newData) {
+    this._data = newData;
+    this.updateNodesList();
+  }
+  get data() {
+    return this._data;
+  }
+
+  updateNodesList() {
+    this.nodesList = new NodesList({ items: this.data.slaves });
```
orlandohohmeier: Let's just reset the node list and the create the list in the respective getter to defer the creation as illustrated below:

```js

set data(newData) {
 this._data = newData;
 this._nodesList = null;
}

getNodesList() {
 if(this._nodesList === null){
     this._nodesList = new NodesList({ items: this.data.slaves });
  }

return this._nodesList;
}
```

DanielMSchmidt:
@orlandohohmeier Yeah, deferring the computation seems to make sense, took your code now 👍

PR #2961, opened 2018-05-16
DCOS-37436: Adjust mesos state store

orlandohohmeier:
Introduce a `schedulerTasksMap`  to improve lookups when assigning the scheduler task field. This change drastically improves the performance as `assignSchedulerTaskField`  was using the `some` operator to check whether there is a matching scheduler task resulting in n*m iterations.

Closes DCOS-37436

**Checklist**

### DIFF
--- a/src/js/stores/MesosStateStore.js
+++ b/src/js/stores/MesosStateStore.js
@@ -44,15 +44,15 @@ function stopPolling() {
 /**
  * Assigns a property to task if it is a scheduler task.
  * @param  {Object} task
- * @param  {Array} schedulerTasks Array of scheduler task
+ * @param  {Object} schedulerTasksMap Map of scheduler task
  * @return {Object} task
  */
-function assignSchedulerTaskField(task, schedulerTasks) {
-  if (schedulerTasks.some(({ id }) => task.id === id)) {
-    return Object.assign({}, task, { schedulerTask: true });
+function assignSchedulerTaskField(task, schedulerTasksMap) {
+  if (schedulerTasksMap[task.id] == null) {
+    return task;
   }
 
-  return task;
+  return { ...task, schedulerTask: true };
 }
 
 /**
@@ -231,7 +231,7 @@ class MesosStateStore extends GetSetBaseStore {
     const frameworks = this.get("lastMesosState").frameworks || [];
     const memberTasks = {};
 
-    const schedulerTasks = this.getSchedulerTasks();
+    const schedulerTasksMap = this.getSchedulerTasksMap();
 
     frameworks.forEach(framework => {
       framework.tasks.forEach(function(task) {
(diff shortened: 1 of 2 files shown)


mesosphere-mergebot-dev:
```
@mesosphere-mergebot ship-it  
@mesosphere-mergebot test  
```

src/js/stores/MesosStateStore.js
```diff
@@ -44,15 +44,15 @@ function stopPolling() {
...
  * @param  {Object} task
- * @param  {Array} schedulerTasks Array of scheduler task
+ * @param  {Object} schedulerTasksMap Map of scheduler task
  * @return {Object} task
  */
-function assignSchedulerTaskField(task, schedulerTasks) {
-  if (schedulerTasks.some(({ id }) => task.id === id)) {
-    return Object.assign({}, task, { schedulerTask: true });
+function assignSchedulerTaskField(task, schedulerTasksMap) {
+  if (schedulerTasksMap[task.id] == null) {
```
DanielMSchmidt: Double equals always make me a bit nervous, I think I personally would prefer to do `!schedulerTasksMap[task.id]`, but it's okay for me like this 🤷‍♂️

src/js/stores/MesosStateStore.js
```diff
@@ -283,6 +283,13 @@ class MesosStateStore extends GetSetBaseStore {
     });
   }
 
+  getSchedulerTasksMap() {
+    return this.getSchedulerTasks().reduce(
+      (acc, { id, ...rest }) => ({ [id]: rest, ...acc }),
```
DanielMSchmidt: Is there a special reason why we don't want the task id to be part of the task in the schedulerTasksMap?

DanielMSchmidt:
@nLight Wanted to give @orlandohohmeier the chance to react my comments, so let's keep it this way for now

src/js/stores/MesosStateStore.js
orlandohohmeier: There's no special reason...  I didn't add the `id` as it's an `id` to task data map and thus the `id` would redundant, but I'm happy to add the `id` if you think that this adds value.

DanielMSchmidt: Right, you already have the id when calling it, makes sense to me 👍

src/js/stores/MesosStateStore.js
orlandohohmeier: It doesn't really matter in this case as we only need to check for `undefined`. I went with `value == null` to relax the condition also catch cases in which some one `null`fied the value. Please note that `value == null` checks for `null` and `undefined` and `!value` for all falsey values.

orlandohohmeier:
@jaxesn I've tested with about 4k tasks and 10 schedulers resulting in 40k iterations, just to set the scheduler task flag.

PR #2962, opened 2018-05-17
DCOS-37556: prevent unnecessary updates in ServiceBreadcrumb

Poltergeist:
<details><summary><strong>fix: prevent service breadcrumb to update without changes</strong></summary><hr/>

This adds a `shouldComponentUpdate` function to the `ServiceBreadcrumb` component. This ensures the
component is only updated if any of the props has changed.

To test you can use the following service config:

```JSON
{
  "id": "/volumes",
  "containers": [
    {
      "name": "container-1",
      "resources": {
        "cpus": 0.1,
        "mem": 128,
        "disk": 0
      },
      "image": {
        "kind": "DOCKER",
        "id": "alpine"
      },
      "exec": {
        "command": {
          "shell": "sleep 1000;"
        }
      },
      "volumeMounts": [
        {
          "name": "home",
          "mountPath": "/var/home"
        },
        {
          "name": "nothome",
          "mountPath": "/var/nothome"
        }
      ]
    }
  ],
  "volumes": [
    {
      "name": "home",
      "persistent": {
        "type": "root",
        "size": 1,
        "constraints": []
      }
    },
    {
      "name": "nothome",
      "persistent": {
        "type": "root",
        "size": 1,
        "constraints": []
      }
    }
  ],
  "networks": [
    {
      "mode": "host"
    }
  ],
  "scaling": {
    "instances": 1,
    "kind": "fixed"
  },
  "scheduling": {
    "placement": {
      "constraints": []
    }
  },
  "executorResources": {
    "cpus": 0.1,
    "mem": 32,
    "disk": 10
  },
  "fetch": []
}
```

This is going to create a multi container application with multiple volumes. Then you should
navigate to the service detail page.
Check if you can navigate to the task detail page and back.
Check if you can visit the volume detail page.
Check if you can switch between the two volume detail pages.
  - Copy the URL of the first volume Detail page,
  - Visit the second one, go to the address bar paste the URL and hit enter.
  - After this you may use the browser history to go back and forth.

Close DCOS-22453
</details>

---
Commit:
6922ed07c4a4a5044263227ace2bf3673b2cb25c

Testwise from https://github.com/dcos/dcos-ui/pull/2962/files#diff-90b25207bf60e5708a7ba1ffdec09510R86 to the bottom is interesting this is new code the rest is moved. I created a separate `describe` block since the old one had a before each So I thought it made sense to create a separate block.

**Checklist**

### DIFF
--- a/plugins/services/src/js/components/ServiceBreadcrumbs.js
+++ b/plugins/services/src/js/components/ServiceBreadcrumbs.js
@@ -1,21 +1,20 @@
-import ReactDOM from "react-dom";
-import PropTypes from "prop-types";
-import React from "react";
-import { Link } from "react-router";
-
-import StringUtil from "#SRC/js/utils/StringUtil";
-import DCOSStore from "#SRC/js/stores/DCOSStore";
 import Breadcrumb from "#SRC/js/components/Breadcrumb";
 import BreadcrumbSupplementalContent
   from "#SRC/js/components/BreadcrumbSupplementalContent";
 import BreadcrumbTextContent from "#SRC/js/components/BreadcrumbTextContent";
 import PageHeaderBreadcrumbs from "#SRC/js/components/PageHeaderBreadcrumbs";
+import DCOSStore from "#SRC/js/stores/DCOSStore";
+import StringUtil from "#SRC/js/utils/StringUtil";
 import Util from "#SRC/js/utils/Util";
-
+import deepEqual from "deep-equal";
+import PropTypes from "prop-types";
+import React from "react";
+import ReactDOM from "react-dom";
+import { Link } from "react-router";
+import ServiceTree from "../structs/ServiceTree";
 import ServiceStatusProgressBar from "./ServiceStatusProgressBar";
 import ServiceStatusWarningWithDebugInformation
   from "./ServiceStatusWarningWithDebugInstruction";
-import ServiceTree from "../structs/ServiceTree";
 
 // The breadcrumb's margin is hardcoded to avoid calling #getComputedStyle.
 const BREADCRUMB_CONTENT_MARGIN = 7;
@@ -43,6 +42,22 @@ class ServiceBreadcrumbs extends React.Component {
   componentDidMount() {
     this.checkBreadcrumbOverflow();
     global.addEventListener("resize", this.handleViewportResize);
+  }
+
+  shouldComponentUpdate(nextProps) {
+    const hasServiceIDChanged = this.props.serviceID !== nextProps.serviceID;
+    const hasTaskIDChanged = this.props.taskID !== nextProps.taskID;
+    const hasTaskNameChanged = this.props.taskName !== nextProps.taskName;
+    const hasExtraNotChanged =
+      this.props.extra === nextProps.extra ||
+      deepEqual(this.props.extra, nextProps.extra);
+
+    return (
+      hasServiceIDChanged ||
+      hasTaskIDChanged ||
+      hasTaskNameChanged ||
+      !hasExtraNotChanged
+    );
   }
 
   componentDidUpdate() {
--- a/plugins/services/src/js/components/__tests__/ServiceBreadcrumbs-test.js
+++ b/plugins/services/src/js/components/__tests__/ServiceBreadcrumbs-test.js
@@ -1,5 +1,5 @@
+import { mount, shallow } from "enzyme";
 import React from "react";
-import { mount } from "enzyme";
 
 jest.mock("#SRC/js/stores/DCOSStore");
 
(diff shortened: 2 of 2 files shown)


plugins/services/src/js/components/ServiceBreadcrumbs.js
```diff
@@ -45,6 +44,22 @@ class ServiceBreadcrumbs extends React.Component {
     global.addEventListener("resize", this.handleViewportResize);
   }
 
+  shouldComponentUpdate(nextProps) {
+    const hasServiceIDChanged = this.props.serviceID !== nextProps.serviceID;
+    const hasTaskIDChanged = this.props.taskID !== nextProps.taskID;
+    const hasTaskNameChanged = this.props.taskName !== nextProps.taskName;
+    const hasExtraNotChanged =
```
DanielMSchmidt: Why is this the only time where it is not changed instead of changed?

plugins/services/src/js/components/ServiceBreadcrumbs.js
```diff
@@ -45,6 +44,22 @@ class ServiceBreadcrumbs extends React.Component {
...
+      this.props.extra === nextProps.extra ||
+      deepEqual(this.props.extra, nextProps.extra);
```
DanielMSchmidt: Why the equality check before the deepEqual?

plugins/services/src/js/components/__tests__/ServiceBreadcrumbs-test.js
```diff
@@ -10,69 +10,154 @@ const PageHeaderBreadcrumbs = require("#SRC/js/components/PageHeaderBreadcrumbs"
...
-
-      const links = instance.find(Link);
-      expect(links.length).toEqual(4);
-      // Icon
-      expect(links.at(0).prop("to")).toEqual("/services");
+    it(`is not updating on same serviceID`, function() {
+      const wrapper = shallow(
+        <ServiceBreadcrumbs {...{ serviceID: "/same" }} />
+      );
+      const spy = jest.spyOn(wrapper.instance(), "componentDidUpdate");
```
DanielMSchmidt: Could we name the spy in a more meaningful way? `componentDidUpdateSpy` maybe?

plugins/services/src/js/components/ServiceBreadcrumbs.js
Poltergeist: Good question. The extra prop is an array of nodes. But they are only set on certain occasions so in most cases, extra will be undefined. This is caught by the first comparison. If checking for changes the comparison would be a lot more complex that is why I picked the not changed instead. 

Just to give more meaning to that the comparison would start with something like that:
```JS
(this.props.extra === undefined && nextProps.extra !== undefined) ||
(this.props.extra !== undefined && nextProps.extra === undefined) ||
!deepEqual(this.props.extra, nextProps.extra)
```

And I would also need to make sure to check for null. So I choose to rather have one Not Changed instead of a Changed. But also I could wrap the condition into brackets so it would look like this:
```JS
const hasExtraChanged = !( 
  his.props.extra === nextProps.extra ||
  deepEqual(this.props.extra, nextProps.extra)
);
```
But my Opinion is that this more complex.

plugins/services/src/js/components/ServiceBreadcrumbs.js
Poltergeist: As deepEqual is not a lightweight function I thought we would rather not call that function in obvious cases. 

Unfortunately, we have to call it since nodes are never equal.

Also please consider that this function is called every 2 seconds. And extra is undefined in most cases.

plugins/services/src/js/components/__tests__/ServiceBreadcrumbs-test.js
Poltergeist: Sure will do, sorry for being so lazy.

plugins/services/src/js/components/ServiceBreadcrumbs.js
DanielMSchmidt: I think this looks readable:

```js
const hasExtraChanged = this.props.extra !== nextProps.extra || !deepEqual(this.props.extra, nextProps.extra);
```

plugins/services/src/js/components/ServiceBreadcrumbs.js
DanielMSchmidt: Okay, got you there 👍

plugins/services/src/js/components/__tests__/ServiceBreadcrumbs-test.js
DanielMSchmidt: That wasn't what I tried to express 🙈

Poltergeist: But that is why I named it that way. I was too lazy to pick a proper name.

plugins/services/src/js/components/ServiceBreadcrumbs.js
DanielMSchmidt: Offline discussed: The first part is truthy if it's a different object which is completely equal

plugins/services/src/js/components/ServiceBreadcrumbs.js
orlandohohmeier: @Poltergeist could you please file a JIRA to adjust deepEqual so that we don't need to  first check whether the two options are the same here?

orlandohohmeier:
@user please open a PR to back-port the adjustment to DC/OS UI 1.10 and 1.11

Poltergeist:
Yes @orlandohohmeier I have this on my plate I just wanted to get the change once merged.

plugins/services/src/js/components/ServiceBreadcrumbs.js
DanielMSchmidt: Maybe we can do that on master, but let's stay with `deep-equal` on the legacy branches 👍

juliangieseke: +1 Michaels suggestion

Fabs: I like the idea too, I just would defer that for a later time, given that we want quickly ship these performance improvements.

juliangieseke: After offline discussion with philipp: replacing deep-equal wit isEqual should be its own task. AFAIK philipp is checking right now if it is significantly faster 👍

Poltergeist: I looked into the performance, and yes lo-dash isEqual is seemed to be a lot faster. But my Opinion on this is that introducing a new module is not better solving the issue. And it is a scope creep. We have the deep-equal module already installed and use it in multiple places in the code base. So I only adopted that pattern. I created an issue for the next sprint to replace deep-equal with lodash.isEqual [DCOS-37562](https://jira.mesosphere.com/browse/DCOS-37562) I hope that this approach is resolving this issue @orlandohohmeier @weblancaster

PR #2968, opened 2018-05-17
[master] DCOS-37440: schedulerTasksMap with lookups

DanielMSchmidt:
Introduce a `schedulerTasksMap`  to improve lookups when assigning the scheduler task field. This change drastically improves the performance as
`assignSchedulerTaskField`  was using the `some` operator to check whether there is a matching scheduler task resulting in n*m iterations.

Closes DCOS-37437

**Checklist**


nLight:
We may not even need the map... :trollface:

nLight:
https://github.com/dcos/dcos-ui/pull/2975

PR #2970, opened 2018-05-18
DCOS-37461: Run assignSchedulerTaskField once

nLight:
As we don't _need_ to run `assignSchedulerTaskField` over and over again it makes sense to run it once when data changes.

Then we have a chicken and egg problem because `getSchedulerTasksMap` needs `frameworks` and  `assignSchedulerTaskField` needs the map. To resolve this I had to make other functions pure and pass `frameworks` as an argument rather then `getting` it.

**Checklist**

Closes DCOS-37461


mesosphere-mergebot-dev:
```
@mesosphere-mergebot ship-it  
@mesosphere-mergebot test  
```

PR #2973, opened 2018-05-18
DCOS-37475: Cache Service Tree

DanielMSchmidt:
**Checklist**

### DIFF
--- a/src/js/stores/DCOSStore.js
+++ b/src/js/stores/DCOSStore.js
@@ -204,6 +204,7 @@ class DCOSStore extends EventEmitter {
       );
     });
 
+    this.clearServiceTreeCache();
     this.emit(DCOS_CHANGE);
   }
 
@@ -237,6 +238,7 @@ class DCOSStore extends EventEmitter {
     // Populate deployments with services data immediately
     this.onMarathonDeploymentsChange();
 
+    this.clearServiceTreeCache();
     this.emit(DCOS_CHANGE);
   }
 
@@ -280,6 +282,7 @@ class DCOSStore extends EventEmitter {
       }
     });
 
+    this.clearServiceTreeCache();
     this.emit(DCOS_CHANGE);
   }
 
@@ -295,6 +298,7 @@ class DCOSStore extends EventEmitter {
 
     currentVersions.set(versionID, version);
 
+    this.clearServiceTreeCache();
     this.emit(DCOS_CHANGE);
   }
 
@@ -308,6 +312,7 @@ class DCOSStore extends EventEmitter {
     }
 
     versions.set(serviceID, nextVersions);
+    this.clearServiceTreeCache();
     this.emit(DCOS_CHANGE);
   }
 
@@ -318,6 +323,7 @@ class DCOSStore extends EventEmitter {
     }
 
     this.data.mesos = states;
+    this.clearServiceTreeCache();
     this.emit(DCOS_CHANGE);
   }
 
@@ -328,6 +334,7 @@ class DCOSStore extends EventEmitter {
     metronome.jobTree = MetronomeStore.jobTree;
     metronome.dataReceived = true;
 
+    this.clearServiceTreeCache();
     this.emit(DCOS_CHANGE);
   }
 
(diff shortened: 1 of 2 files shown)


mesosphere-mergebot-dev:
```
@mesosphere-mergebot ship-it  
@mesosphere-mergebot test  
```

src/js/stores/DCOSStore.js
```diff
@@ -441,7 +457,11 @@ class DCOSStore extends EventEmitter {
   }
 
   get taskLookupTable() {
-    return this.buildFlatServiceTree(this.serviceTree);
+    if (!this._flatServiceTree) {
```
nLight: we don't invalidate `this._flatServiceTree`. Does that mean that after we've invalidated `_flatServiceTree` we still going to be returning stalled version of `this._flatServiceTree`?

DanielMSchmidt: @nLight We do whenever service tree is changed. See the get serviceTree if statement

jaxesn: Is there a scenario where you could have called one of the above methods that nulls out the serviceTree, but havent called `getServiceTree`, which nulls out flatServiceTree, before calling `getTaskLookupTable`? Or is it guaranteed that getting the lookup table will only happen after calling getservicetree?

DanielMSchmidt: @nLight @jaxesn I understood your problem and fixed it 😇

src/js/stores/DCOSStore.js
```diff
@@ -204,6 +204,7 @@ class DCOSStore extends EventEmitter {
       );
     });
 
+    this.serviceTree = null;
```
weblancaster: Why not have a method that set `serviceTree` to `null`?

DanielMSchmidt:
@weblancaster what would you like to have unit tested? There is no additional functionality, so I wouldn't know which meaningful test I can add, without having a super tight coupling with the implementation. Do you have an example of what you would like to see?

src/js/stores/DCOSStore.js
juliangieseke: We already have an `if event is DCOS_CHANGE` in `this.emit` - wouldn't that be a better place for this reset? 🤔

Fabs: I agree here. IMHO it would be better to have a cache cleaning method like `clearServiceTreeCache()`

juliangieseke:
and on testing: I'd love to see a test for `this.emit` :D

src/js/stores/DCOSStore.js
Fabs: I think to make this simpler to reason, we could invalidate both serviceTree and flatServiceTree together on a  method like `clearServiceTree`. Then they are both rebuilt lazily when needed (when the getter is called, we build one, or both.

src/js/stores/DCOSStore.js
Fabs: The goal here is to lazily evaluate the tree so if no one calls `get serviceTree` we don't compute the tree at all. At the same time, if any new data comes, we need to clear the cached tree by setting it to null. This improves on the previous behavior because we would always recompute the tree when we have new data.

Fabs:
@DanielMSchmidt. I am just guessing, but the tests that @weblancaster might mean (and the ones I feel could exist too) are the ones who verify the cache behavior. By for instance getting the tree, getting the table, making a data change, getting both again and see if they did not got stale.

Fabs:
@juliangieseke I would not like to see that put on an event. First, because events could be async, and introducing a race condition will break the cache. On async, clear the cache, get the tree and the lookup table can happen in any order, and if the cache clear happens in between you get bugs.

Also, the cache is already complex as itself, I would not couple that with events.

juliangieseke:
@Fabs totally true and FULL ACK - BUT check this out: :( https://github.com/dcos/dcos-ui/blob/e8fd34cd96f3a3d0e4863f635ccc363f75d6d0e9/src/js/stores/DCOSStore.js#L211
/cc @DanielMSchmidt

DanielMSchmidt:
@Fabs @juliangieseke I added some more tests, please take a look

src/js/stores/__tests__/DCOSStore-test.js
```diff
@@ -650,4 +650,41 @@ describe("DCOSStore", function() {
       expect(Object.keys(flatTree).length).toEqual(1);
     });
   });
+
+  describe("taskLookupTable", function() {
```
Fabs: Nice. I have not thought it that way but I see the value on those tests.

PR #2976, opened 2018-05-22
DCOS-37560: Remove track.js

nLight:
This PR is a part of our effort to replace track.js It should be already removed from segment.com so that the code's just throwing an error now.

The PR is removing the code that throws an error.

**Testing**
Without the PR you should see a console.log error: `configure of undefined` with the PR no error.

**Checklist**

Closes DCOS-37560


juliangieseke:
is there a way to test this? 🤔

PR #2980, opened 2018-05-22
[1.10] DCOS-37472: use isSchedulerTask instead of schedulerTask

DanielMSchmidt:
Closes DCOS-37472

**Checklist**


src/js/stores/MesosStateStore.js
```diff
@@ -358,18 +305,11 @@ class MesosStateStore extends GetSetBaseStore {
...
-
     lastMesosState.frameworks = lastMesosState.frameworks.reduce(
       (acc, framework) => {
-        const tasks = (framework.tasks || [])
-          .map(task => assignSchedulerTaskField(task, schedulerTasksMap));
-        const completed_tasks = (framework.completed_tasks || [])
-          .map(task => assignSchedulerTaskField(task, schedulerTasksMap));
-        const unreachable_tasks = (framework.unreachable_tasks || [])
-          .map(task => assignSchedulerTaskField(task, schedulerTasksMap));
+        const tasks = framework.tasks || [];
```
nLight: We can remove this `reduce` entirely as its sole purpose was to ~pass butter~ `assignSchedulerTaskField `

plugins/services/src/js/containers/tasks/TasksView.js
```diff
@@ -147,6 +147,10 @@ class TasksView extends mixin(SaveStateMixin) {
     );
   }
 
+  isSchedulerTask(task) {
+    return task.labels.some(({ key }) => key === "DCOS_PACKAGE_FRAMEWORK_NAME");
```
nLight: So those are plan objects and not `Task` structs? How do you think is it better to put in into `TaskUtil`?

src/js/stores/MesosStateStore.js
nLight: Though I still wonder if that'd be faster to assign the flag once rather than calculating the value in the `TaskView`

jaxesn: It is a whole lot less code...almost suspiciously less... One advantage this new approach might have over the old, if we aren't always rendering all tasks, maybe only completed or only unreachable, we wouldn't be looping over all of them for no reason. That said, depending on how often we are re-rendering, that might be negated...

plugins/services/src/js/containers/tasks/TasksView.js
weblancaster: Agree with @nLight although I would not block.

PR #2981, opened 2018-05-22
DCOS-37559: remove track.js

nLight:
This PR is a part of our effort to replace track.js It should be already removed from segment.com so that the code's just throwing an error now.

The PR is removing the code that throws an error.

**Testing**
Without the PR you should see a console.log error: `configure of undefined` with the PR no error.

**Checklist**

Closes DCOS-37559

### DIFF
--- a/plugins/tracking/hooks.js
+++ b/plugins/tracking/hooks.js
@@ -9,15 +9,11 @@ let SDK = require('./SDK').getSDK();
 let {
   AuthStore,
   Config,
-  DOMUtils,
-  EventTypes,
-  MetadataStore
+  DOMUtils
 } = SDK.get([
   'AuthStore',
   'Config',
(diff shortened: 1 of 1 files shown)


nLight:
Alfred please test this

nLight:
@DanielMSchmidt I think that's because of the `routes` action. We don't have it anywhere in the file, but trying to bind it anyway. Not related to my work, though would you like me addressing the issue in the same PR?

DanielMSchmidt:
@nLight No, I just have not enough experience with 1.8 to understand what I was doing wrong 😇

nLight:
Alfred please test this

nLight:
@DanielMSchmidt that was an actual `cherry-pick` issue, I addressed it

PR #2987, opened 2018-05-24
[1.11] DCOS-37476: Cache Service Tree

DanielMSchmidt:
**Checklist**

### DIFF
--- a/src/js/stores/DCOSStore.js
+++ b/src/js/stores/DCOSStore.js
@@ -204,6 +204,7 @@ class DCOSStore extends EventEmitter {
       );
     });
 
+    this.clearServiceTreeCache();
     this.emit(DCOS_CHANGE);
   }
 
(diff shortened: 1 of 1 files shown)


src/js/stores/DCOSStore.js
```diff
@@ -402,6 +409,23 @@ class DCOSStore extends EventEmitter {
...
+    if (!this._serviceTree) {
+      this.serviceTree = this.buildServiceTree();
+    }
+
+    return this._serviceTree;
+  }
+
+  set serviceTree(newValue) {
+    this._serviceTree = newValue;
+    this._flatServiceTree = null;
```
nLight: Same as on master. `this._flatServiceTree` is not being used. Maybe we should just delete it

DanielMSchmidt:
@nLight @Fabs I removed the reference for `_flatServiceTree` now 👍

PR #2989, opened 2018-05-25
[master] DCOS-37480: Defer service spec creation

DanielMSchmidt:
Adjust the service structs to defer the service spec creation to avoid unnecessary instantiations.

Closes DCOS-37480

**Checklist**


nLight:
Dunno if it actually helps. On the Services screen `structs/Pod#getSpec` is being called every render. Though the PR works and good to go

Fabs:
@nLight 🤔 isn't the benefit because we don't call the constructor so often as we call the getter, and hence we save the operation?

PR #2992, opened 2018-05-25
[1.10] DCOS-22145: Backport of "File UI breaks using the switch button"

DanielMSchmidt:
Backport of https://github.com/dcos/dcos-ui/pull/2790

## How to review

1. Launch a Service with a Task
2. View a file (e.g. Stdout)
3. Switch to other files, the UI should not break then

**Checklist**

### DIFF
--- a/plugins/services/src/js/pages/task-details/TaskFileViewer.js
+++ b/plugins/services/src/js/pages/task-details/TaskFileViewer.js
@@ -12,7 +12,22 @@ import SearchLog from "../../components/SearchLog";
 import TaskDirectory from "../../structs/TaskDirectory";
 import TaskDirectoryActions from "../../events/TaskDirectoryActions";
 
-class TaskFileViewer extends React.Component {
+export function getNewRoute(routePath, params, path) {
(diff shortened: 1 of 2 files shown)


plugins/services/src/js/pages/task-details/TaskFileViewer.js
```diff
@@ -12,7 +12,22 @@ import SearchLog from "../../components/SearchLog";
 import TaskDirectory from "../../structs/TaskDirectory";
 import TaskDirectoryActions from "../../events/TaskDirectoryActions";
 
-class TaskFileViewer extends React.Component {
+export function getNewRoute(routePath, params, path) {
+  const hasFilePathParam = routePath.indexOf(":filePath") !== -1;
```
weblancaster: Not a blocker but why not `includes(":filePath")` instead `indexOf(":filePath") !== -1`?

DanielMSchmidt: Just because I forgot it in the original implementation. This is just a backport 🙈 but next time I will use includes

PR #2996, opened 2018-05-31
DCOS-37619: Refactor jobs detail page

Poltergeist:
<details><summary><strong>refactor(jobs): remove state from job detail page</strong></summary><hr/>

This introduces a refactoring to remove the state from the job detail page and move it into a
designated container component. The container component does keep all the state except the tabs
state.
</details>

---
Commit:
d433ce72f67b623af4b7b888f4c12433386812c3
<details><summary><strong>refactor(jobs): pull jobs detail container out into a seperate file</strong></summary><hr/>

This extracts the JobDetailContainer into a seperate file.
</details>

---
Commit:
802cc5f10b9c4d9121666e8441993a1aa2f068c8

This pulls out most of the state and store logic to a container component for the  component.


Fabs:
I think by splitting the ticket like this (with I like a lot) we also need to split some subtasks on JIRA since this PR will not close DCOS-37619.

src/js/routes/jobs.js
```diff
@@ -1,22 +1,23 @@
...
   from "../../../plugins/services/src/js/pages/task-details/TaskFileViewer";
+import TaskFilesTab
+  from "../../../plugins/services/src/js/pages/task-details/TaskFilesTab";
 import TaskLogsContainer
   from "../../../plugins/services/src/js/pages/task-details/TaskLogsContainer";
+import JobsPage from "../pages/JobsPage";
+/* eslint-enable no-unused-vars */
+import {
+  JobDetailPageContainer as JobDetailPage
+} from "../pages/jobs/JobDetailPageContaiener";
```
bstavroulakis: oops, I think this should be JobDetailPageContainer.js

Fabs: I think that is the correct import, but that the as is a little confusing for me too. I have no strong opinion here so I am approving, but maybe it makes more sense to just not use JobDetailPage and have JobDetailPageContainer everywhere on this file.

Fabs: Ohh, the typo ;)

nLight:
Ready for review? JIRA is not

PR #2998, opened 2018-05-31
DCOS-37767 Simplifies the reference implementation to make it easier to use the data-layer

Fabs:
This is a "small" refactor to remove the use of context, and to better name the client for the reference implementation so it makes easier for people to use the reference implementation from the data layer. 

It was motivated by my office hours with bill where I struggled to explain why we did this way, and also from orlando feedback that this is super complicated.


Fabs:
Looks like a broke something. Will checkt it more carefully.

PR #2999, opened 2018-06-01
DCOS-37612: Jobs Menu Data Layer - Run Job

bstavroulakis:
This PR has 2 goals:

1) The groundwork to detach the menu from the JobsDetails page.
2) Add the Run Job with the new Data Layer pattern.

**Checklist**

### DIFF
--- a/plugins/jobs/src/js/JobsMenu.js
+++ b/plugins/jobs/src/js/JobsMenu.js
@@ -1,0 +1,36 @@
+/* eslint-disable no-unused-vars */
+import React from "react";
+/* eslint-enable no-unused-vars */
+
+import { componentFromStream } from "data-service";
+import "rxjs/add/operator/combineLatest";
+
+import Page from "#SRC/js/components/Page";
+import JobsBreadcrumbs from "./components/JobsBreadcrumbs";
+import { jobsRunNowAction, jobsRunNow$ } from "./JobsRunNow";
+
+const getActions = function(props) {
+  const [edit, ...rest] = props.actions;
+  const newActions = [];
+
+  newActions.push(edit);
+
+  newActions.push(jobsRunNowAction(props.job.getId()));
+
+  return newActions.concat(rest);
+};
+
+const JobsMenu = componentFromStream(props$ => {
+  return props$.combineLatest(jobsRunNow$.startWith({}), props => {
+    return (
+      <Page.Header
+        actions={getActions(props)}
+        breadcrumbs={<JobsBreadcrumbs tree={props.tree} item={props.job} />}
+        tabs={props.tabs}
+        isPageHeader={props.isPageHeader}
+      />
+    );
+  });
+});
+
+export default JobsMenu;
--- a/plugins/jobs/src/js/JobsRunNow.js
+++ b/plugins/jobs/src/js/JobsRunNow.js
@@ -1,0 +1,36 @@
+/* eslint-disable no-unused-vars */
+import React from "react";
+/* eslint-enable no-unused-vars */
+
+import { graphqlObservable } from "data-service";
+import gql from "graphql-tag";
+import "rxjs/add/operator/concatMap";
+
+import { Subject } from "rxjs/Subject";
+import defaultSchema from "./data/JobModel";
+
+const runNowJobMutation = gql`
+  mutation {
+    runJob(id: $id) {
+      id
+    }
+  }
+`;
+
+const runNowJobGraphql = id => {
+  return graphqlObservable(runNowJobMutation, defaultSchema, id);
+};
+
+export const runNowEvent$ = new Subject();
+export const jobsRunNow$ = runNowEvent$.switchMap(id => {
+  return runNowJobGraphql(id);
+});
+
+export const jobsRunNowAction = id => {
+  return {
+    label: "Run Now",
+    onItemSelect() {
+      runNowEvent$.next({ id });
+    }
+  };
+};
--- a/plugins/jobs/src/js/__tests__/JobsRunNow-test.js
+++ b/plugins/jobs/src/js/__tests__/JobsRunNow-test.js
@@ -1,0 +1,35 @@
+import { Observable } from "rxjs";
+import { marbles } from "rxjs-marbles/jest";
+import { resolvers } from "../data/JobModel";
+import * as jobsRunNow from "../JobsRunNow";
+
+describe("Jobs Run Now Action", function() {
+  describe("#runAction", function() {
+    it("runNowEvent call is made with onItemSelect", function(done) {
+      const jobsRunNowAction = jobsRunNow.jobsRunNowAction("test");
+      jobsRunNow.runNowEvent$.subscribe(job => {
+        expect(job.id).toBe("test");
+        done();
+      });
+      jobsRunNowAction.onItemSelect();
+    });
+
+    it(
+      "returns the jobId",
+      marbles(m => {
+        m.bind();
+        const result$ = resolvers({
+          fetchJobDetail: () => Observable.of({ id: "foo" }),
+          pollingInterval: m.time("-|"),
+          runJob: () => Observable.of({ jobId: "foo" })
+        }).Mutation.runJob({}, { id: "foo" });
+
+        m.expect(result$.take(1)).toBeObservable(
+          m.cold("(x|)", {
+            x: { jobId: "foo" }
+          })
+        );
+      })
+    );
+  });
+});
--- a/src/js/pages/jobs/JobDetailPage.js
+++ b/src/js/pages/jobs/JobDetailPage.js
@@ -9,7 +9,8 @@ import mixin from "reactjs-mixin";
 
 import JobFormModalContainer from "#PLUGINS/jobs/src/js/JobFormModalContainer";
 import TaskStates from "#PLUGINS/services/src/js/constants/TaskStates";
-import JobsBreadcrumbs from "#PLUGINS/jobs/src/js/components/JobsBreadcrumbs";
+import JobsMenu from "#PLUGINS/jobs/src/js/JobsMenu";
+
 import Icon from "../../components/Icon";
 import Page from "../../components/Page";
 import TimeAgo from "../../components/TimeAgo";
@@ -187,11 +188,6 @@ class JobDetailPage extends mixin(TabsMixin) {
       onItemSelect: this.props.handleEditButtonClick
     });
 
-    actions.push({
-      label: "Run Now",
-      onItemSelect: this.props.handleRunNowButtonClick
-    });
-
     if (schedule != null && schedule.enabled) {
       actions.push({
         label: "Disable Schedule",
@@ -246,10 +242,12 @@ class JobDetailPage extends mixin(TabsMixin) {
 
     return (
       <Page>
-        <Page.Header
+        <JobsMenu
           actions={this.getActions()}
-          breadcrumbs={<JobsBreadcrumbs tree={jobTree} item={job} />}
           tabs={this.getTabs()}
+          job={job}
+          tree={jobTree}
+          isPageHeader={true}
         />
         {this.tabs_getTabView(job)}
         <JobFormModalContainer
--- a/packages/data-service/index.js
+++ b/packages/data-service/index.js
@@ -1,0 +1,2 @@
+export { graphqlObservable } from "./src/graphqlObservable";
+export { createEventHandler, componentFromStream } from "./src/recomposeRx";
--- a/packages/data-service/src/index.js
+++ b/packages/data-service/src/index.js
@@ -1,2 +1,0 @@
-export { graphqlObservable } from "./graphqlObservable";
-export { createEventHandler, componentFromStream } from "./recomposeRx";
--- a/plugins/jobs/src/js/__tests__/JobsMenu-test.js
+++ b/plugins/jobs/src/js/__tests__/JobsMenu-test.js
@@ -1,0 +1,32 @@
+/* eslint-disable no-unused-vars */
+import React from "react";
+/* eslint-enable no-unused-vars */
+
+import Job from "#SRC/js/structs/Job";
+import JobTree from "#SRC/js/structs/JobTree";
+
+import Page from "#SRC/js/components/Page";
+import JobsMenu from "../JobsMenu";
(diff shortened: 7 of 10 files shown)


Fabs:
@bstavroulakis I am also rebasing it so it is aligned with master.

src/js/pages/jobs/JobDetailPage.js
```diff
@@ -329,10 +331,7 @@ class JobDetailPage extends mixin(StoreMixin, TabsMixin) {
       onItemSelect: this.handleEditButtonClick
     });
 
-    actions.push({
```
weblancaster: because the store method isn't used anymore, you should be able to remove `handleRunNowButtonClick` from `JobDetailPage`

nLight:
Ready for review? Jira is in progress

plugins/jobs/src/js/JobsMenuClient.js
```diff
@@ -0,0 +1,31 @@
...
+import { makeExecutableSchema } from "graphql-tools/dist/index";
+import { runJob } from "#SRC/js/events/MetronomeClient";
+
+export const typeDefs = `
+  type Job {
+      id: String,
+      jobId: String,
+      status: String,
+      createdAt: String
+  }
```
Fabs: If we are only using the jobId right now, that would be the only thing I would put on this type.

plugins/jobs/src/js/JobsRunNow.js
```diff
@@ -0,0 +1,40 @@
...
+import React from "react";
+/* eslint-enable no-unused-vars */
+
+import { graphqlObservable } from "data-service";
+import gql from "graphql-tag";
+
+import { Subject } from "rxjs/Subject";
+import "rxjs/add/operator/combineLatest";
+import "rxjs/add/operator/do";
+import "rxjs/observable/empty";
```
Fabs: I would only leave here the rxjs imports we use (which probably means removing all the operator/observable imports).

plugins/jobs/src/js/JobsRunNow.js
```diff
@@ -0,0 +1,40 @@
...
+});
+
+export const jobsRunNowAction = id => {
+  return {
+    label: "Run Now",
+    onItemSelect() {
+      jobsRunNow$.next({ jobId: id });
+    }
+  };
+};
```
Fabs: Here we could test that running the callback (onItemSelect()) really triggers the graphql call.

plugins/jobs/src/js/JobsMenu.js
```diff
@@ -0,0 +1,40 @@
...
+  const newActions = [];
+
+  newActions.push(edit);
+
+  newActions.push(jobsRunNowAction(props.job.getId()));
+
+  return newActions.concat(rest);
+};
+
+const JobsMenu = componentFromStream(props$ => {
```
Fabs: What about snapshot testing this one? We can have the cases where it has the schedule or not.

plugins/jobs/src/js/JobsRunNow.js
bstavroulakis: When I run the method from the test, for some reason the .next doesn't make the graphql call. Will need to research this more.

plugins/jobs/src/js/JobsMenu.js
bstavroulakis: Also, in the test with the snapshot the componentFromStream doesn't initialize correctly. Will need some investigation.

plugins/jobs/src/js/JobsMenu.js
```diff
@@ -0,0 +1,40 @@
+/* eslint-disable no-unused-vars */
+import React from "react";
+/* eslint-enable no-unused-vars */
+
+import { componentFromStream } from "data-service";
+
+import "rxjs/add/operator/combineLatest";
```
weblancaster: this line and the 2 below below aren't in use so they can be removed

plugins/jobs/src/js/JobsMenu.js
```diff
@@ -0,0 +1,36 @@
...
+
+  newActions.push(edit);
+
+  newActions.push(jobsRunNowAction(props.job.getId()));
+
+  return newActions.concat(rest);
+};
+
+const JobsMenu = componentFromStream(props$ => {
+  return props$.combineLatest(jobsRunNow$.startWith({}), props => {
```
DanielMSchmidt: Why do we need `jobRunNow$`? I can't see that we are utilizing any of its observables output

plugins/jobs/src/js/__tests__/JobsRunNow-test.js
```diff
@@ -0,0 +1,23 @@
...
+describe("Jobs Run Now Action", function() {
+  it("runNowJobGraphql call if made with onItemSelect", function(done) {
+    const runNowJobGraphql = jest.fn().mockReturnValue(
+      new Observable(() => {
+        expect(runNowJobGraphql).toBeCalled();
+        done();
+      })
+    );
+
+    const jobsRunNowAction = jobsRunNow.jobsRunNowAction(
```
DanielMSchmidt: Do I get that right that we have no test which verifies that the GraphQL schema goes together with what we expect the client to deliver?

plugins/jobs/src/js/JobsRunNow.js
```diff
@@ -0,0 +1,38 @@
...
+export const jobsRunNow$ = runNowEvent$.switchMap(jobId => {
+  return runNowJobGraphql(jobId);
+});
+
+export const jobsRunNowAction = (id, externalRunNowJobGraphql) => {
+  return {
+    label: "Run Now",
+    onItemSelect() {
+      if (externalRunNowJobGraphql) {
+        runNowJobGraphql = externalRunNowJobGraphql;
```
DanielMSchmidt: This seems like a super hacky solution for testing to me. I would rather like to have the respective functions hoisted to achieve something like this.

plugins/jobs/src/js/JobsRunNow.js
```diff
@@ -0,0 +1,38 @@
...
+
+import { graphqlObservable } from "data-service";
+import gql from "graphql-tag";
+
+import { Subject } from "rxjs/Subject";
+import { defaultSchema } from "./JobsMenuClient";
+
+const runNowJobMutation = gql`
+  mutation {
+    runJob(jobId: $jobId) {
```
DanielMSchmidt: How does GraphQL know what $jobId is? You only pass a variable to the graphqlObservable, right?

plugins/jobs/src/js/__tests__/JobsRunNow-test.js
```diff
@@ -0,0 +1,23 @@
...
+      "test",
+      runNowJobGraphql
+    );
+
+    jobsRunNow.jobsRunNow$.forEach(() => {});
```
DanielMSchmidt: What's the purpose of this?

webpack/webpack.dev.js
```diff
@@ -13,7 +13,7 @@ module.exports = merge(common, {
     index: "./src/js/index.js",
     vendor: Object.keys(dependencies)
   },
-  devtool: "cheap-module-eval-source-map",
+  devtool: "source-map",
```
DanielMSchmidt: This seems unrelated, could you pull this into a seperate PR and open a ticket for why this is needed?

plugins/jobs/src/js/JobsMenuClient.js
```diff
@@ -0,0 +1,28 @@
...
+import { makeExecutableSchema } from "graphql-tools/dist/index";
+import { runJob } from "#SRC/js/events/MetronomeClient";
+
+export const typeDefs = `
+  type Job {
+      jobId: String
+  }
+
+  type Mutation {
+    runJob(jobId: String!): Job!
```
DanielMSchmidt: The return type is `JobRun` and not Job. For a job the id would be just id. As I am currently working on typing this stuff out and transforming it into something with the shape of a struct I think a renaming here would be suficcient. Otherwise we would do double work if you would adapt the schema here and did the data transformation

bstavroulakis: I changed the type to JobRun and the ids to id.

webpack/webpack.dev.js
bstavroulakis: Oops, I do this to make debug work in VSCode. Had an auto-ignore script that probably changed now that the webpack config changed and this slipped. No need for us to change this, cheap eval is faster for dev.

plugins/jobs/src/js/__tests__/JobsRunNow-test.js
bstavroulakis: Basically you have to listen to the observable for it to actually start as it is lazy. Changed this to subscribe.

plugins/jobs/src/js/JobsRunNow.js
bstavroulakis: Changed to id: $id. I'm not 100% how does this work, but if I put id: String or something else it doesn't map correctly with the defaultSchema. Probably @Fabs can help us here.

plugins/jobs/src/js/JobsRunNow.js
bstavroulakis: Yes, was frustrated and wanted to make this work. Definitely not optimal, I changed this.

plugins/jobs/src/js/__tests__/JobsRunNow-test.js
bstavroulakis: Yeah, no test for that. Would we want a test for this?

npm-shrinkwrap.json
```diff
@@ -1150,12 +1150,6 @@
       "integrity": "sha512-e74sM9W/4qqWB6D4TWV9FQk0WoHtX1X4FJpbjxucMSVJHtFjbQOH3H6yp+xno4br0AKG0wz/kPtaN599GUOvAg==",
       "dev": true
     },
-    "@types/jest": {
```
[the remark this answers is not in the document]
bstavroulakis: Hmm, I just did ```rm -rf node_modules``` and ran ```npm install``` again. This is odd.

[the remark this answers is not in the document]
bstavroulakis: I will revert to the old one.

plugins/jobs/src/js/JobsMenu.js
bstavroulakis: When I remove this and just do ```props$.subscribe(props => {})``` I get ```cannot use props$ undefined```. I guess we do this to extend with an observable that always exist. Maybe @Fabs can clear this out a bit.

plugins/jobs/src/js/JobsRunNow.js
```diff
@@ -0,0 +1,36 @@
...
+    }
+  }
+`;
+
+const runNowJobGraphql = id => {
+  return graphqlObservable(runNowJobMutation, defaultSchema, id);
+};
+
+export const runNowEvent$ = new Subject();
+export const jobsRunNow$ = runNowEvent$.switchMap(id => {
```
bstavroulakis: @DanielMSchmidt Over here we use switchmap with the runNowEvent Subject to actually listen to this observable but also have the ability to trigger it as well.

plugins/jobs/src/js/JobsMenu.js
Fabs: `combineLatest` will not emit unless both the streams it is combining emits at least once. So if you combinLatest the props with jobsRunNow, the menu will never appear. You need to give jobsRunNow an initial value.

plugins/jobs/src/js/JobsMenuClient.js
```diff
@@ -0,0 +1,32 @@
+import "rxjs/add/operator/combineLatest";
+import "rxjs/add/operator/map";
+
+import { makeExecutableSchema } from "graphql-tools/dist/index";
+import { runJob } from "#SRC/js/events/MetronomeClient";
+
+export const typeDefs = `
```
Fabs: Now that Daniel has this schema:
https://github.com/dcos/dcos-ui/pull/3006

I suggest we make this queries against that schema, and move our resolvers and mutations to that file (I believe it is a good idea to rebase from that branch).

Also the naming of this file is a little of. The files with extension client are usually network to data layer adapters. The files with graphql resolvers and type definitions are supposed to be Models. And we are aiming to have an unified model for jobs.

plugins/jobs/src/js/JobsMenu.js
DanielMSchmidt: I get that, but `jobRunNow$` is called with `.next({id})` and this id property is not used here as far as I can see, so why do we need the value of the stream?

plugins/jobs/src/js/JobsMenuClient.js
Fabs: Moving that I also think we should test the resolvers for
- returning a stream
- the stream has the transformations I want (which in this case might mean just that the runJob is being called with the id of the args I passed.

plugins/jobs/src/js/JobsMenu.js
Fabs: You need to pass the id to the Subject, to trigger the runNow of the particular id. 

jobRunNow$ is a stream of all the results of the executions of the graphqlMutation to run a job. To issue the command you have to onNext to the Subject related to that, so the subject will switchMap into the request.

bstavroulakis:
Thanks @DanielMSchmidt for the review, fixed conflicts.

src/js/pages/jobs/JobDetailPage.js
```diff
@@ -245,10 +242,12 @@ class JobDetailPage extends mixin(TabsMixin) {
 
     return (
       <Page>
-        <Page.Header
+        <JobsMenu
```
nLight: Please consider keeping `Page.Header` looks like the way @orlandohohmeier suggested in the thread https://mesosphere.slack.com/archives/C03UXUVH7/p1529066379000323 could work. With that we won't need to modify the TemplateUtil

DanielMSchmidt:
@bstavroulakis This can run against master now <3 sorry for all the conflicts 😞

Fabs:
Closing that in favor of #3023, will incorporate the comments here, there.

PR #3002, opened 2018-06-01
DCOS-37791: Connect to MesosStream on all pages

nLight:
On a bigger clusters fetching state can take a while. With this fix we connect to the stream as soon as we can so that when a user opens a page that requires the data it's already there (or half way there at least)

## Testing

Remember that the UI doesn't connect to the stream on the Dashboard - now it will. Please keep in mind that even in the current version once we connect to the stream we never unsubscribe.

## Checklist

Closes DCOS-37791

### DIFF
--- a/src/js/pages/Index.js
+++ b/src/js/pages/Index.js
@@ -9,6 +9,7 @@ import ConfigStore from "../stores/ConfigStore";
 import EventTypes from "../constants/EventTypes";
 import InternalStorageMixin from "../mixins/InternalStorageMixin";
 import MetadataStore from "../stores/MetadataStore";
+import MesosStateStore from "../stores/MesosStateStore";
 import Modals from "../components/Modals";
 import RequestErrorMsg from "../components/RequestErrorMsg";
 import ServerErrorModal from "../components/ServerErrorModal";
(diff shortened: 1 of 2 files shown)


nLight:
Legit Integration test failure

src/js/pages/Index.js
```diff
@@ -40,6 +41,7 @@ var Index = React.createClass({
   componentWillMount() {
     MetadataStore.init();
     SidebarStore.init();
+    MesosStateStore.addChangeListener(EventTypes.MESOS_STATE_CHANGE, () => {});
```
Fabs: I know we never unsubscribe, but shouldn't we at least remove the listener when the component unmounts? My reason being if for any reason someone/something starts mounting unmounting the component, I think we can run on a memory leak where those callbacks only get added to the list of listeners (assuming here it is a list of it).

nLight:
Hey, @Fabs I addressed the commen!

tests/smoke-cy.js
```diff
@@ -50,7 +50,7 @@ describe("DC/OS UI [00j]", function() {
     });
 
     it("displays one row on the table [00t]", function() {
-      cy.get("table tbody tr").should("to.have.length", 3).contains("dcos-01");
+      cy.get("table tbody tr").should("to.have.length", 6).contains("dcos-01");
```
jaxesn: Why does this change the number of rows?

nLight: Good question. In fact even without my change there were more than 3 of them. The difference I observe that without my change `/dcos-history-service/` call would error out in Cypress, with the change it doesn't. Looks like it is something that can make or break our test.

src/js/pages/Index.js
nLight: 👉 poke

tests/smoke-cy.js
jaxesn: makes sense

PR #3004, opened 2018-06-05
DCOS-37750: Refactor JobsTab component

juliangieseke:
I started moving files to `plugins/jobs`, please feel free to point me to your PRs if they got conflicts. I'm fine fixing them for you 👍 

---

This splits up the old JobsTab component into a set of 'dumb' UI components (in JobsTab.js) and a container component (JobsTabContainer.js).

This is the first step of multiple steps to refactor the Jobs Overview Page.

---

**how to review?**
check that everything works as before (compare with daily cluster) and that tests are green (CI). other than that is looking at the code, check if everything is feasable and as expected 👍


src/js/pages/jobs/JobsTab.js
```diff
@@ -76,185 +127,56 @@ class JobsTab extends mixin(StoreMixin) {
...
-            breadcrumbs={<JobsBreadcrumbs tree={DCOSStore.jobTree} />}
-          />
-          <Loader />
-        </Page>
-      );
+    if (!isLoading) {
+      return <JobsTabLoading root={root} />;
     }
 
+    // render list view
```
DanielMSchmidt: I think these comments don't provide any value. They explain what is written in the next line without further context or any more information and I can't really see why we need them

juliangieseke: You're right, but there were already two comments (above & below), so my ocd insisted to also add this one 😅

DanielMSchmidt: I meant all of them actually

src/js/pages/jobs/JobsTab.js
```diff
@@ -76,185 +127,56 @@ class JobsTab extends mixin(StoreMixin) {
...
-    if (!DCOSStore.jobDataReceived) {
-      return (
-        <Page>
-          <Page.Header
-            breadcrumbs={<JobsBreadcrumbs tree={DCOSStore.jobTree} />}
-          />
-          <Loader />
-        </Page>
-      );
+    if (!isLoading) {
```
nLight: We could move this into Container. Since it is in charge of the loading state. So that the container either shows `LoadingScreen` or `JobsTab` what do you think? And no need passing a boolean flag down

src/js/pages/jobs/JobsTabContainer.js
```diff
@@ -0,0 +1,150 @@
...
+      DCOSStore.jobTree.findItem(function(item) {
+        return item instanceof JobTree && item.id === id;
+      }) || DCOSStore.jobTree;
+    const isLoading = DCOSStore.jobDataReceived;
+    const root = DCOSStore.jobTree;
+    const filteredJobs = this.getFilteredJobs(item);
+    const searchString = this.state.searchString;
+    const handleFilterChange = this.handleFilterChange;
+    const resetFilter = this.resetFilter;
+    const hasFilterApplied = Object.keys(
```
nLight: How would it look like if we passed down the data and left the decision up to `JobsTab`? `JobsTab` still would be dumb but wouldn't leek its logic to the parent component.

src/js/pages/jobs/JobsTab.js
```diff
@@ -13,61 +7,118 @@ import FilterHeadline from "../../components/FilterHeadline";
...
+    <Page>
+      <Page.Header
+        addButton={{
+          label: "Create a job",
+          onItemSelect: handleOpenJobFormModal
+        }}
+        breadcrumbs={<JobsBreadcrumbs tree={root} item={item} />}
+      />
+      <div className="flex-grow">
+        <ConditionalFilterHeadline
```
nLight: We could simply have lift up the `hasFilterApplied` logic to the `JobsTabList` and keep `FilterHeadline` plain no logic dumb

juliangieseke:
I already did some of your suggestions as part of the next PR, will move them into this to make the refactoring complete here 👍

nLight:
flicts

juliangieseke:
(update incoming…)

plugins/jobs/src/js/JobsTabContainer.js
```diff
@@ -0,0 +1,155 @@
...
+    const resetFilter = this.resetFilter;
+    const hasFilterApplied = Object.keys(DEFAULT_FILTER_OPTIONS).some(
+      filterKey => {
+        return (
+          this.state[filterKey] != null && this.state[filterKey].length > 0
+        );
+      }
+    );
+
+    return isLoading ? (
```
nLight: you sure you get a component here?

juliangieseke: yup, fixed the issue i had 👍

plugins/jobs/src/js/JobsTabContainer.js
```diff
@@ -0,0 +1,156 @@
...
+};
+
+class JobsTabContainer extends mixin(StoreMixin) {
+  constructor() {
+    super(...arguments);
+
+    this.state = { ...DEFAULT_FILTER_OPTIONS };
+
+    this.store_listeners = [
+      { name: "dcos", events: ["change"], suppressUpdate: false },
```
Fabs: Are we getting rid of the parts that we know are not needed, like this DcosStore dependency?

plugins/jobs/src/js/JobsTabContainer.js
```diff
@@ -0,0 +1,156 @@
...
+
+    this.context.router.push({ pathname, query });
+  }
+
+  resetFilter() {
+    var state = Object.assign({}, this.state, DEFAULT_FILTER_OPTIONS);
+    this.setState(state, this.resetFilterQueryParams);
+  }
+
+  updateFromProps(props) {
```
Fabs: Kill it! Here too right?

plugins/jobs/src/js/JobsTabContainer.js
Fabs: 🎨  I think the if isLoading return <...> is easier to read.

plugins/jobs/src/js/JobsTabContainer.js
```diff
@@ -0,0 +1,156 @@
...
+      }
+    ];
+
+    METHODS_TO_BIND.forEach(method => {
+      this[method] = this[method].bind(this);
+    });
+  }
+
+  componentDidMount() {
+    super.componentDidMount();
```
Fabs: Those two I think were not needed too.

plugins/jobs/src/js/JobsTabContainer.js
juliangieseke: Yeah, this will be part of the next PR 👍🏻

plugins/jobs/src/js/JobsTabContainer.js
juliangieseke: I can change this, but then I'll add https://eslint.org/docs/rules/no-ternary to our eslint rules… It annoys me that this is something we allow but i get a comment _everytime_ I use it…

Fabs: Hum, I would not pull extreme to the other side. In my opinion on style, ternaries are good for short clauses, I try not to use them if it spans more than a single line. I would not be in favor of the death of all ternaries.

DanielMSchmidt: +1 on @Fabs point of view

plugins/jobs/src/js/components/JobsBreadcrumbs.d.ts
```diff
@@ -0,0 +1,10 @@
+import { Component } from "react";
+import JobTree from "#SRC/js/structs/JobTree";
+import Job from "#SRC/js/structs/Job";
+
+export default class Loader extends Component {
+  props: {
+    tree: JobTree
+    item: JobTree | Job
```
juliangieseke: this is weird 🤔

juliangieseke:
Added some of the types @nLight did in his PR, but the failing test is weird 🤔 :(

plugins/jobs/src/js/JobsTabContainer.js
juliangieseke: this _is_ a short clause 🤔

juliangieseke:
@nLight looks like this is depending on your stuff now, no Idea whats wrong with that test 😅
Also you're still blocking this, so would be awesome if you could revisit 👍

plugins/jobs/src/js/JobsTabContainer.js
Fabs: I dont think this is a short clause. Short is a single line, with probably the predicate being a variable or a very simple condition.

```js
return isLoading ? jobsTabLoading : JobsTab;
```
That I would be fine with. But I won`t block it because of that.

plugins/jobs/src/js/components/JobsBreadcrumbs.d.ts
Fabs: I think too.

tslint.json
```diff
@@ -1,5 +1,7 @@
 {
-    "extends": [
-        "@dcos/tslint-config"
-    ]
+  "extends": ["@dcos/tslint-config"],
+  "rules": {
+    "interface-name": [false],
+    "no-submodule-imports": [false]
```
Fabs: Can we have a small note to why `no-submodule-imports` on this PR description?

plugins/jobs/src/js/JobsTabContainer.js
DanielMSchmidt: I'm fine with this ternary.

plugins/jobs/src/js/components/JobsBreadcrumbs.js
```diff
@@ -4,15 +4,15 @@ import PropTypes from "prop-types";
 import React from "react";
 import { Tooltip } from "reactjs-components";
 
+// everything that is references as #ALIAS/something has to be refactored once our DI system is in place
```
DanielMSchmidt: What is the purpose of this comment? I get the message, but this information seems not relevant to me in this file.

tslint.json
DanielMSchmidt: If you change this, please also remove  all comments like `// TODO: remove this disable with https://jira.mesosphere.com/browse/DCOS_OSS-3579` and all the `// tslint:disable:no-submodule-imports` and `// tslint:disable-next-line:no-submodule-imports`

juliangieseke: > `Submodule import paths from this package are disallowed; import from the root instead`  when importing via `#ALIAS/`

https://mesosphere.slack.com/archives/C03UXUVH7/p1528733271000724

plugins/jobs/src/js/components/JobsBreadcrumbs.d.ts
```diff
@@ -0,0 +1,10 @@
+import { Component } from "react";
+import JobTree from "#SRC/js/structs/JobTree";
+import Job from "#SRC/js/structs/Job";
+
+export default class Loader extends Component {
```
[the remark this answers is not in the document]
juliangieseke: Because I f*ed up :D Good catch 👍

tslint.json
juliangieseke: I think I did (at least in my code) 🤔 can check again 👍

plugins/jobs/src/js/components/JobsBreadcrumbs.js
juliangieseke: We want to change these once we are able to, the idea is that whoever stumbles upon these imports know immediately, that they are not best practice but legacy compatibility.

DanielMSchmidt: I don't know if this really has the effect you aim for. I would rather remove them, otherwise you have these stray comments that appear when you plan for something but the plan changes. You see them a lot in older parts of our code base

tslint.json
DanielMSchmidt: In your code yes, but I am talking about the entire codebase

plugins/jobs/src/js/JobsTabContainer.js
juliangieseke: Ok, short for me is `return (clause) ? A : B;`. I dont really care if `A = true` or `A = <Component with={} a={} lot={} of={} props={} />`.
Long for me would be something like `return (clauseA) ? (clauseAB) ? AB1 : AB2 : (clauseAC) ? AC1 : AC2;` 

whatever, please lets stop commenting this on every single PR if we dont add the rule 🙏

Fabs: I also receive comments about style from time to time, that I disagree with. Nonetheless, I think we should not shy about sharing our perceptions on style.

The ternary on this piece of code does not match my bar for `short`, I would also not be in favor of banning ternaries for life.

I approved the PR nonetheless ✅

tslint.json
Fabs: Thanks.

plugins/jobs/src/js/JobsTabContainer.js
juliangieseke: I might be overreacting here, but I received comments like yours almost _everytime_ I used ternaries - if it is such a bad thing, we should ban it, if not, please stop complaining (the team as a whole, not you in particular). nvm - thanks for approving 🙇

plugins/jobs/src/js/components/JobSearchFilter.tsx
```diff
@@ -0,0 +1,25 @@
+/* eslint-disable no-unused-vars */
```
nLight: Created JIRA to remove those comments https://jira.mesosphere.com/browse/DCOS-38200

plugins/jobs/src/js/components/JobsBreadcrumbs.d.ts
```diff
@@ -0,0 +1,10 @@
+import { Component } from "react";
+import JobTree from "#SRC/js/structs/JobTree";
+import Job from "#SRC/js/structs/Job";
+
+export default class JobsBreadcrumbs extends Component {
+  props: {
```
nLight: I thought we were going with interfaces for props :)

juliangieseke: yeah, missed that… I'd like to fix it in a second PR, this one is green and I'd like to have it merged asap 👍

plugins/jobs/src/js/JobsTabContainer.js
orlandohohmeier: I'd argue that the usage of a ternary here makes it hard to follow the flow and would say that an early exit would probably be easier to follow, but agree with the general sentiment that we shouldn't manually enforce styling.

The problem here is that there is no eslint rule for this case only; thus we either need to write one or would have to ban ternaries in general. Prohibiting ternaries altogether would bloat cases like the following and is therefore not desirable. 

```jsx
 <span>
        {predicate ? a : b}
 </span>
```
Thus we should implement a rule for cases like this one and/or accept that there are always cases in which we will discuss and manually enforce style as not everything that is theoretically "allowed" is also desirable.

PR #3006, opened 2018-06-11
DCOS-38118: Add a job model

DanielMSchmidt:
Goal of this PR is to provide a Typescript and GraphQL definition for the jobs endpoint and to implement the resolvers for it

**Checklist**

### DIFF
--- a/plugins/jobs/src/js/data/JobModel.ts
+++ b/plugins/jobs/src/js/data/JobModel.ts
@@ -1,0 +1,293 @@
+import { makeExecutableSchema, IResolvers } from "graphql-tools";
+import { Observable } from "rxjs";
+
+export interface Query {
+  jobs: JobConnection | null;
+  job: Job | null;
+}
+
+export interface GeneralArgs {
+  [key: string]: any;
+}
+
+export interface JobsQueryArgs {
+  filter?: string | null;
+  sortBy?: SortOption | null;
+  sortDirection?: SortDirection | null;
+}
+
+export interface JobQueryArgs {
+  id: string;
+  filter?: string | null;
+  sortBy?: SortOption | null;
+  sortDirection?: SortDirection | null;
+}
+
+export type SortOption = "ID" | "STATUS" | "LAST_RUN";
+
+export type SortDirection = "ASC" | "DESC";
+
+export interface JobConnection {
+  filteredCount: number;
+  totalCount: number;
+  nodes: Job[];
+}
+
+export interface Job {
+  activeRuns: JobRunConnection;
+  command: string;
+  cpus: number;
+  description: string;
+  disk: number;
+  docker: Docker | null;
+  id: string;
+  jobRuns: JobRunConnection;
+  json: string;
+  labels: Label[];
+  lastRunsSummary: JobHistorySummary;
+  lastRunStatus: JobRunStatusSummary;
+  mem: number;
+  name: string;
+  schedules: ScheduleConnection;
+  scheduleStatus: JobStatus;
+}
+
+export interface JobRunConnection {
+  longestRunningActiveRun: JobRun | null;
+  nodes: JobRun[];
+}
+
+export interface JobRun {
+  dateCreated: number;
+  dateFinished: number;
+  jobID: string;
+  status: JobRunStatusSummary;
+  tasks: JobTaskConnection;
+}
+
+export interface JobRunStatusSummary {
+  status: JobRunStatus | null;
+  time: number | null;
+}
+
+export type JobRunStatus = "FAILED" | "NOT_AVAILABLE" | "SUCCESS";
+
+export interface JobTaskConnection {
+  longestRunningTask: JobTask;
+  nodes: JobTask[];
+}
+
+export interface JobTask {
+  dateCompleted: number;
+  dateStarted: number;
+  status: JobTaskStatus;
+  taskId: string;
+}
+
+export type JobTaskStatus =
+  | "TASK_CREATED"
+  | "TASK_DROPPED"
+  | "TASK_ERROR"
+  | "TASK_FAILED"
+  | "TASK_FINISHED"
+  | "TASK_GONE"
+  | "TASK_GONE_BY_OPERATOR"
+  | "TASK_KILLED"
+  | "TASK_KILLING"
+  | "TASK_LOST"
+  | "TASK_RUNNING"
+  | "TASK_STAGING"
+  | "TASK_STARTED"
+  | "TASK_STARTING"
+  | "TASK_UNKNOWN"
+  | "TASK_UNREACHABLE";
+
+export interface Docker {
+  forcePullImage: boolean;
+  image: string;
+}
+
+export interface Label {
+  key: string;
+  value: string;
+}
+
+export interface JobHistorySummary {
+  failureCount: number;
+  lastFailureAt: string | null;
+  lastSuccessAt: string | null;
+  successCount: number;
+}
+
+export interface ScheduleConnection {
+  nodes: Schedule[];
+}
+
+export interface Schedule {
+  cron: string;
+  enabled: boolean;
+  id: string;
+  startingDeadlineSeconds: number;
+  timezone: string;
+}
+
+export type JobStatus =
+  | "ACTIVE"
+  | "COMPLETED"
+  | "FAILED"
+  | "INITIAL"
+  | "RUNNING"
+  | "SCHEDULED"
+  | "STARTING"
+  | "UNSCHEDULED";
+
+export const typeDefs = `
+type Job {
+  activeRuns: JobRunConnection!
+  command: String!
+  cpus: Float!
+  description: String!
+  disk: Float!
+  docker: Docker
+  id: ID!
+  jobRuns: JobRunConnection!
+  json: String!
+  labels: [Label]!
+  lastRunsSummary: JobHistorySummary!
+  lastRunStatus: JobRunStatusSummary!
+  mem: Int!
+  name: String!
+  schedules: ScheduleConnection!
+  scheduleStatus: JobStatus!
+}
+type JobRunConnection {
+  longestRunningActiveRun: JobRun
+  nodes: [JobRun]!
+}
+type JobRun {
+  dateCreated: Int!
+  dateFinished: Int!
+  jobID: String!
+  status: JobRunStatusSummary!
+  tasks: JobTaskConnection!
+}
+type JobTaskConnection {
+  longestRunningTask: JobTask!
+  nodes: [JobTask]!
+}
+type JobTask {
+  dateCompleted: Int!
+  dateStarted: Int!
+  status: JobTaskStatus!
+  taskId: String!
+}
+type JobHistorySummary {
+  failureCount: Int!
+  lastFailureAt: String
+  lastSuccessAt: String
+  successCount: Int!
+}
+type Docker {
+  forcePullImage: Boolean!
+  image: String!
+}
+type Label {
+  key: String!
+  value: String!
+}
+type JobRunStatusSummary {
+  status: JobRunStatus
+  time: Int
+}
+type ScheduleConnection {
+  nodes: [Schedule]!
+}
+type Schedule {
+  cron: String!
+  enabled: Boolean!
+  id: ID!
+  startingDeadlineSeconds: Int!
+  timezone: String!
+}
+enum JobTaskStatus {
+  TASK_CREATED
+  TASK_DROPPED
+  TASK_ERROR
+  TASK_FAILED
+  TASK_FINISHED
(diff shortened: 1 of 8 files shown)


plugins/jobs/src/js/data/JobModel.ts
```diff
@@ -0,0 +1,293 @@
...
+}
+
+export interface JobHistorySummary {
+  failureCount: number;
+  lastFailureAt: string | null;
+  lastSuccessAt: string | null;
+  successCount: number;
+}
+
+export interface ScheduleConnection {
```
Fabs: How do you get away not putting "I" on all interfaces here. I think this is a good example on why the pattern to start interfaces with the letter "I" should die.

src/js/constants/JobStates.ts
```diff
@@ -2,7 +2,16 @@
  * Sort order is ordered by most important (lowest number, top of list)
  * to least important (largest number, bottom of list)
  */
-const JobStates = {
+
+export interface IJobStates {
```
Fabs: Back to the discussion about the "I". it looks a little mixed to me that we use some places and others not.

tslint.json
```diff
@@ -1,5 +1,6 @@
 {
-    "extends": [
-        "@dcos/tslint-config"
-    ]
+  "extends": ["@dcos/tslint-config"],
+  "rules": {
+    "interface-name": [false]
```
Fabs: Ah, I see, we disabled the "I". Can we then be consistent and not use it anywhere?

plugins/jobs/src/js/data/JobModel.ts
```diff
@@ -0,0 +1,293 @@
...
+  | "ACTIVE"
+  | "COMPLETED"
+  | "FAILED"
+  | "INITIAL"
+  | "RUNNING"
+  | "SCHEDULED"
+  | "STARTING"
+  | "UNSCHEDULED";
+
+export const typeDefs = `
```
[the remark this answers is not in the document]
DanielMSchmidt: Yes, we have, the solutions are already planned out, but the tickets are just not in the sprint. We went for a one-time generation of the types as this was the most pragmatic choice.

plugins/jobs/src/js/data/JobModel.ts
```diff
@@ -0,0 +1,293 @@
...
+  job(
+    id: ID!
+    filter: String
+    sortBy: SortOption
+    sortDirection: SortDirection
+  ): Job
+}
+`;
+
+// Pass implementation of MetronomeClient to the resolvers for easier testability
```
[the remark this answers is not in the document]
DanielMSchmidt: Can you explain the problem to me, I don't really see it with the current implementation. That being said I am not married to it, I can also switch it out and mock the respective requests 🤷‍♂️

plugins/jobs/src/js/data/JobModel.ts
DanielMSchmidt: Disabled it in the tslint config. We discussed it in the sync and we moved the item a couple of times by saying we still need to find some strong cases for it. Had a short discussion with @orlandohohmeier on that topic and we didn't want to introduce so many types that we would need to rename later on, possibly causing confusion.

tslint.json
DanielMSchmidt: We could, but I think this should be the scope of a Jira for changing it in the global config. If you insist I will add it nonetheless to get this one moved on faster.

DanielMSchmidt:
@Fabs on the feature branch topic: I don't really see a point. We are not merging something broken here and we are currently not using it, so I don't see a downside in having this on master.

plugins/jobs/src/js/data/JobModel.ts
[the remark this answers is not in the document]
Fabs: Cool, I will love to see that.

plugins/jobs/src/js/data/JobModel.ts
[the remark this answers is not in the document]
Fabs: I think the current implementation should just use MetronomeClient directly. It is simpler and it makes easy to add the final solution we want later.

This is not going to stay like this. There will be a way to inject MetronomeClient using Dependency Injection. So the problem with this implementation is that it is trying to solve a problem on a suboptimal way by allowing the resolvers to be configured by passing the client to them, on a scenario where we already know we wanna solve that with the Extension Kid.

Saying more on that topic, if we were not planning to use the IOC container, I would propose we do it like Prisma does, passing the MetronomeClient to the resolvers via the context. That is also the way we currently do on the reference implementation (see repositoryModel.js). That said, that context solution is also suboptimal in the light of the Extension Kid.

plugins/jobs/src/js/data/JobModel.ts
```diff
@@ -0,0 +1,293 @@
...
+
+export type SortDirection = "ASC" | "DESC";
+
+export interface JobConnection {
+  filteredCount: number;
+  totalCount: number;
+  nodes: Job[];
+}
+
+export interface Job {
```
nLight: I'd rather see these declarations split in several files: https://github.com/dcos/dcos-ui/pull/3007/files
I think your declarations are much more precise though

tslint.json
Fabs: I agree. I think the only thing that is really needed here is to now steamroll the previous team decision. I am in favor of this being here, and the global one happened in another ticket.

plugins/jobs/src/js/data/JobModel.ts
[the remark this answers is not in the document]
nLight: @Fabs it's gonna take a while for Extension Kid to be fully there, so don't count on it. Some required tasks are in the next sprint.

plugins/jobs/src/js/data/__tests__/JobModel-test.ts
```diff
@@ -0,0 +1,27 @@
+describe("JobModel Resolver", () => {
+  describe("jobs", () => {
+    it("polls the endpoint");
+    it("waits for running requests to finish before sending new ones");
+
+    // TODO: write this part in greater detail
```
nLight: Jira?

plugins/jobs/src/js/data/JobModel.ts
DanielMSchmidt: The idea behgínd having it here is that a) they are generated from the GraphQL schema and b) that we can use them (import) for any of the other files. What would speak for moving them away from their source?

src/js/constants/JobStatus.ts
```diff
@@ -2,7 +2,15 @@
  * Sort order is ordered by most important (lowest number, top of list)
  * to least important (largest number, bottom of list)
  */
-const JobStatus = {
+
+export interface JobStatus {
+  [state: string]: {
```
nLight: I didn't know you can do that!

plugins/jobs/src/js/data/__tests__/JobModel-test.ts
DanielMSchmidt: I added these afterwards, shall I add them? https://jira.mesosphere.com/browse/DCOS-38126
https://jira.mesosphere.com/browse/DCOS-38125

src/js/events/MetronomeClient.ts
```diff
@@ -2,33 +2,107 @@ import { request } from "@dcos/http-service";
...
+  description: string;
+  labels: object;
+  run: {
+    cpus: number;
+    mem: number;
+    disk: number;
+    cmd: string;
+    env: {};
+    placement: {
+      constraints: any[];
```
nLight: I bet it's gonna be `string[]`, no?

src/js/events/MetronomeClient.ts
```diff
@@ -2,33 +2,107 @@ import { request } from "@dcos/http-service";
...
+
+export interface ScheduleData {
+  id: string;
+}
 const defaultHeaders = {
   "Content-Type": "application/json; charset=utf-8"
 };
 
-export function createJob(data: IJobData) {
+export function createJob(data: JobData) {
```
nLight: shall we specify return type?

src/js/events/MetronomeClient.ts
DanielMSchmidt: I don't know. We originally wanted to get the types from the RAML definitions, but that ticket is not in the sprint. If you see the commit, the primary idea here was to just move in some types and improve them while working on the respective resolvers.

src/js/events/MetronomeClient.ts
DanielMSchmidt: Yes, whenever we use these functions I would say (https://github.com/dcos/dcos-ui/pull/3006#discussion_r194408693)

plugins/jobs/src/js/data/JobModel.ts
nLight: Aren't we going to have duplications that way?

plugins/jobs/src/js/data/JobModel.ts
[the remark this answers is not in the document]
Fabs: @nLight I ack what you said and I think this is not a problem. For this PR, I think we should not try to do any magic here, just use the MetronomeClient directly and then this week start discussing the timeline to use the Extension Kid. If we can't use the Extension Kid immediately, that is ok.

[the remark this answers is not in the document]
DanielMSchmidt: I think you might misunderstand my intentions here. The schema is the default export and we won't expose a way to configure the MetronomeClient or any other implementation there, this should be the work of Extension Kid, I agree.

We use the named export for the resolvers only for easier testing and in this context, I would like to be able to pass in the functions the resolver needs. I think this is nicer than mocking the entire endpoint for each test and being dependant on path changes and stuff like that.

plugins/jobs/src/js/data/JobModel.ts
DanielMSchmidt: Why should we have duplicates this way?

src/js/constants/JobStates.ts
Poltergeist: I guess there is a good reason for that, but we should be consistent so either do it or not.

src/js/constants/JobStatus.ts
Poltergeist: yes, you can :-)

plugins/jobs/src/js/data/JobModel.ts
nLight: Correct me if I'm wrong. We have js files `Job.js` `JobRun.js` etc that we have to write `.d.ts` for. Plus we have these declarations that are the same? If they aren't than it is way too confusing and we have to figure out a way making them less ambiguous.

juliangieseke: this `Job` is something different to the `Job` struct… eventually these will replace the struct entirely.

juliangieseke: Also: Keep in mind that this file isnt in use yet, once it is, `Job` struct will be gone (i know, you're writing all the definitions right now, but thats part of the downside of parallelization :()

nLight: OK this is really confusing. But let's see how it turns out

src/js/events/MetronomeClient.ts
Poltergeist: There is no RAML for Metronome, right? but in that case, it should be `string[]`.
Could we add a Jira to replace the any with something more meaningful in general?

DanielMSchmidt: @Poltergeist There were some RAML definitions when I looked. From my point of view these are part of implementing the respective resolvers, so in this case writing the JobDetail resolver. If we need the data we type it. These were already in place, so I decided to add them. If that's something blocking I am fine with removing them from this PR

DanielMSchmidt:
@juliangieseke I disagree, the commit is related as we will need to use these constants in the upcoming resolver PRs. I don't get why I should drop perfectly fine work that improves our code base.

plugins/jobs/src/js/data/JobModel.ts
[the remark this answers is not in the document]
Fabs: I agree that is a better design for testing. I am actually going to try that in another PR I a having problems with.

I just did not want to be understood as the design for the resolvers, because we wanna use extensions.

src/js/constants/JobStates.ts
Fabs: This should not be called IJobStates after our conversation on dropping the I.

DanielMSchmidt: Fixed that 👍

DanielMSchmidt:
@Fabs @juliangieseke @nLight @Poltergeist I addressed your comments, could you take another look?

PR #3007, opened 2018-06-11
DCOS-38119: Typescript type definitions

nLight:
With these definitions my stuff compiles and works \o/ Feel free reviewing it / copy-pasting in your branches.

**Checklist**

Closes DCOS-38119


juliangieseke:
Thanks a lot! IMHO we can merge this as is (almost, commit message needs to be fixed)

src/js/components/ToggleButton.d.ts
```diff
@@ -0,0 +1,11 @@
...
+
+interface IToggleButtonProps {
+  checked: boolean;
+  children: any;
+  onChange: () => void;
+  checkboxClassName: string[] | object | string;
+  className: string[] | object | string;
+}
+
+export default class ToggleButton extends Component<IToggleButtonProps, {}> {}
```
juliangieseke: AFAIK you can say `Component<PropsInterface>` and dont need to pass a state type if there is no state - and if there is state, I think we should add it 🤷‍♂️

src/js/components/modals/ModalHeading.d.ts
```diff
@@ -0,0 +1,11 @@
+import { ReactElement } from "react";
+
+type HeadingLevel = 1 | 2 | 3 | 4 | 5 | 6;
```
juliangieseke: this translates to string, correct? (in js component)

src/typings/react-ace.d.ts
```diff
@@ -0,0 +1,15 @@
+import { Component, ReactElement } from "react";
+
+export default class Ace extends Component {
+  props: {
```
juliangieseke: After fiddling around with some types: I like the `Component<Props>` approach way more 👍

src/js/components/modals/ModalHeading.d.ts
nLight: According to propTypes those should be numbers. Then they translate to strings

nLight:
Everything should be fixed now @Fabs @juliangieseke

nLight:
@juliangieseke sorry :D I added ts configs

juliangieseke:
…now you need to change the commit message :)

nLight:
@juliangieseke doh! Fixed

src/js/structs/Item.d.ts
```diff
@@ -0,0 +1,5 @@
+export default class Item {
```
Fabs: This starts with I :trollface:

nLight: it's not an interface (touché) :D

nLight:
@juliangieseke @Fabs github made me push force

PR #3010, opened 2018-06-12
DCOS-38125: Add Job Model resolver

DanielMSchmidt:
**How to review**

Look through the code and see if the tests are fine 👍 

**Checklist**


plugins/jobs/src/js/data/JobModel.ts
```diff
@@ -281,13 +378,33 @@ export const resolvers = (): IResolvers => ({
     ): Observable<JobConnection | null> {
       return Observable.of(null);
     },
-    job(_obj = {}, _args: GeneralArgs, _context = {}): Observable<Job | null> {
-      return Observable.of(null);
+    job(_obj = {}, args: GeneralArgs, _context = {}): Observable<Job | null> {
```
Fabs: Nice. I think I can see the direction here now. 

I outlined down here an approach that I think could achieve the same goals and could have saved us a couple days of work. I do that because I still think we can save time for the jobs resolver if we do that (see comment bellow).

```js
return responses$
  .map(response => new Job(response))
  .map(job => toGraphqlType(job))
```

```js
function toGraphqlJob(job) {
   return {
         id: job.getId(),
         description: job.description,
         ///....
         activeRuns: job.getActiveRuns()
      };
}
```

Fabs: For the jobs resolver it would be something like:

```
return responses$
  .map(responseList => responseList.map((jobData) => new Job(jobData))
  .map(jobs => new JobList({ items: jobs }))
  .map(jobsList => jobsList.filter(...).sort(...).whatever(..))
  .map(finalJobsList => jobstList.getItems())
  .map(jobs => jobs.map((job) => toGraphqlType(job))
```

Fabs: I dont wanna revive Friday discussion. I just wanted to provide that piece of example of how I think we could solve it simple. On the case we don't see equal, I am commited to drive what you have already decided.

Fabs: @orlandohohmeier, @juliangieseke @DanielMSchmidt

plugins/jobs/src/js/data/JobModel.ts
```diff
@@ -281,13 +378,33 @@ export const resolvers = (): IResolvers => ({
...
+      // TODO: We should not need to do the casting here
+      const { id } = args as JobQueryArgs;
```
Fabs: I put this comment a little bit down here but its for the `_obj`.

I believe the name we should use here is `_parent`. I know appolo names it obj. but the definition they gave is even: 

```
The object that contains the result returned from the resolver on the parent field, or, in the case of a top-level Query field, the rootValue passed from the server configuration. This argument enables the nested nature of GraphQL queries.
```

juliangieseke:
VSCode is happy with `JobModel.ts` file ❤️

plugins/jobs/src/js/data/JobModel.ts
```diff
@@ -271,8 +253,301 @@ type Query {
...
+  JobRunConnection: {
+    longestRunningActiveRun(
+      runs: Array<MetronomeClient.ActiveJobRun | JobHistoryRunWithStatus>
+    ): JobRun | null {
+      if (!runs.length) {
+        return null;
+      }
+
+      const sortedRuns = runs.sort((a, b) => {
+        if (a.createdAt == null && b.createdAt == null) {
```
Fabs: Should we avoid the type coercion here? (===) or we are also guarding against undefined?

plugins/jobs/src/js/data/JobModel.ts
```diff
@@ -281,13 +556,21 @@ export const resolvers = (): IResolvers => ({
...
+
+      return responses$.map(response => typeResolvers.Job(response));
     }
   }
 });
 
 export default makeExecutableSchema({
   typeDefs,
-  resolvers: resolvers()
+  resolvers: resolvers({
```
Fabs: I fix that on my PR, so I would say it is ok, but otherwise:

```
  resolvers: resolvers({
    fetchJobDetail: MetronomeClient.fetchJobDetail,
    pollingInterval: Config.getRefreshRate()
  })
```

plugins/jobs/src/js/data/JobModel.ts
```diff
@@ -271,8 +253,301 @@ type Query {
...
+
+function isActiveRun(arg: any): arg is MetronomeClient.ActiveJobRun {
+  return arg.jobId !== undefined;
+}
+
+function isHistoryRun(arg: any): arg is MetronomeClient.JobHistoryRun {
+  return arg.id !== undefined;
+}
+
+const typeResolvers = {
```
Fabs: Moving forward I kind think we want to separate those type resolvers per their own model files. So we don't have a high churning 1k lines code. 

I think the graphql part of the data layer needs a lot of improvements in making our life easier.

plugins/jobs/src/js/data/__tests__/JobModel-test.js
```diff
@@ -0,0 +1,1379 @@
...
+                  }
+                ]
+              }
+            })
+          );
+        })
+      );
+    });
+  });
+});
```
Fabs: This is what I would call througly tested. Kudos :D

plugins/jobs/src/js/data/JobModel.ts
```diff
@@ -271,8 +253,301 @@ type Query {
...
+  },
+  JobRun: {
+    jobID(
+      run: MetronomeClient.ActiveJobRun | MetronomeClient.JobHistoryRun
+    ): string {
+      if (isActiveRun(run)) {
+        return run.jobId;
+      }
+
+      if (isHistoryRun(run)) {
```
Fabs: Maybe here instead of the type guard just an else. If your run is guaranteed to be one of the two types, the else will infer right that it is a history Run, and accept the id. You will then not need the return "";

(I guess that works)

plugins/jobs/src/js/data/JobModel.ts
```diff
@@ -271,8 +253,301 @@ type Query {
...
+    }
+  },
+  JobTaskConnection: {
+    longestRunningTask(tasks: JobTask[]): JobTask | null {
+      if (!tasks.length) {
+        return null;
+      }
+
+      const sortedTasks = [...tasks].sort((a, b) => {
+        if (a.dateStarted == null && b.dateStarted == null) {
```
Fabs: 🎨  This could be much simpler written as:

```js
return (a.dateStarted || 0) - (b.dateStarted || 0 );
```

plugins/jobs/src/js/data/JobModel.ts
Fabs: Also here if we find the base value that would make `DateUtil.strToMs(baseValue) === 0`, we could write:

```js
return DateUtil.strToMs(a.createdAt || baseValue) - DateUtil.strToMs(b.createdAt || baseValue));
```

plugins/jobs/src/js/data/JobModel.ts
```diff
@@ -27,6 +31,8 @@ export type SortOption = "ID" | "STATUS" | "LAST_RUN";
 
 export type SortDirection = "ASC" | "DESC";
 
+export type JobRunStatus = "FAILED" | "NOT_AVAILABLE" | "SUCCESS";
```
Fabs: The expected states, and the states on JobStates are "Failed", "N/A" and "Success". If we decide to change it here, we have to fix it everywhere, otherwise, it will be just a huge source of confusion.

plugins/jobs/src/js/data/JobModel.ts
DanielMSchmidt: Yeah, I would like to do a second PR somewhen soon to move the type resolvers into the normal resolvers file. For the file size, I think we should separate at least the graphql and the TS definitions derived from that, this would help a lot already. If it's still bad, maybe let's split it up by type?

plugins/jobs/src/js/data/JobModel.ts
DanielMSchmidt: Makes sense 👍

plugins/jobs/src/js/data/JobModel.ts
```diff
@@ -271,8 +253,301 @@ type Query {
...
+      activeRuns: typeResolvers.JobRunConnection(response.activeRuns),
+      schedules: fieldResolvers.Job.schedules(response)
+    };
+  },
+  JobRun(run: MetronomeClient.ActiveJobRun | JobHistoryRunWithStatus): JobRun {
+    return {
+      jobID: fieldResolvers.JobRun.jobID(run),
+      dateCreated: DateUtil.strToMs(run.createdAt),
+      dateFinished: fieldResolvers.JobRun.dateFinished(run),
+      status: run.status, // TODO: derive from where it comes from before passing to this function
```
Fabs: What is this todo abou?

juliangieseke: not that I'm aware of - maybe copied from the old implementation?

DanielMSchmidt: might be 🤷‍♂️

DanielMSchmidt: @Fabs is going to fix it on his branch

juliangieseke:
Ok, can we then merge this? /cc @fabs (you're blocking)

Fabs:
@nLight I agree that needs a refactor so certain things fall into more well-bounded classes, and then get stitched together. This work though prevents everybody from moving further, and as much as I feel you that there is a lot going on here, we have to find a way to move further, or we will not deliver the DL this sprint.

nLight:
@Fabs I understand the reason why you would like to have it merged. It blocks everyone, a lot of hours have been put into it, it's an interim step you would like getting it done until vacation. Yet those aren't good enough reasons. 
I'd like to take my time to try to at least understand what it does. In this shape I can't see this as an improvement over what we've had. On top of that I'm afraid that as we merge it there's a _slight_ chance of that code not being refactored. Then we're going to have fun times on-boarding new people

Fabs:
@nLight. Sure, as history goes you have been more right than me in accessing when something has the potential of being forgotten in the future. I find what you say all valid. My number one ticket proposal for the DL is to refactor that code so we can use this as a reference for the future. I share all the same concerns you share regarding this code, and I also think in this current shape it is hard to see the benefits.

I disagree though that the fact that it is blocking does not give a different approach to how we merge here, because the final goal should be to move it forward properly, but faster.

orlandohohmeier:
@nLight I agree that we shouldn't rush this as we need to make sure that we ship good code and a PR with this many changes takes some time to review properly, but then please don't jump to conclusions until you've reviewed properly. I don't see how this code is in any way worse than what we have and how this will impact on-boarding – only because you already know how structs, stores, actions, mixins, and utils are mangled together doesn't mean that newcomers do thus this change might not make a difference for them.

orlandohohmeier:
@Fabs Let's please not rush this PR, there are other ways to unblock the rest of the work.

jest.config.js
```diff
@@ -54,6 +54,11 @@ module.exports = {
   testRegex: "/__tests__/.*\\-test\\.(es6|js|ts)$",
   moduleFileExtensions: ["js", "json", "es6", "ts", "tsx"],
   modulePathIgnorePatterns: ["/tmp/", "/node_modules/", "/.module-cache/"],
+  moduleNameMapper: {
+    "#SRC/([^\\.]*)$": "<rootDir>/src/$1",
```
nLight: that should be on master already...

plugins/jobs/src/js/data/JobModel.ts
```diff
@@ -271,8 +254,280 @@ type Query {
...
+interface ResolverArgs {
+  fetchJobDetail: (id: string) => Observable<MetronomeClient.JobDetailResponse>;
+  pollingInterval: number;
+}
+
+function isActiveRun(arg: any): arg is MetronomeClient.ActiveJobRun {
+  return arg.jobId !== undefined;
+}
+
+const typeResolvers = {
```
nLight: I would very much like seeing those as separate modules that export functions / consts

plugins/jobs/src/js/data/JobModel.ts
```diff
@@ -271,8 +254,280 @@ type Query {
...
+};
+
+interface JobHistoryRunWithStatus extends MetronomeClient.JobHistoryRun {
+  status: MetronomeClient.JobStatus;
+}
+
+// We currently don't support field resolvers directly, so we need to
+// call these manually.
+const fieldResolvers = {
+  Job: {
```
nLight: esp. this nested name spacing could benefit from being extracted to a module

plugins/jobs/src/js/data/JobModel.ts
```diff
@@ -271,8 +254,280 @@ type Query {
...
+    };
+  },
+  JobRunConnection(
+    runs: Array<MetronomeClient.ActiveJobRun | JobHistoryRunWithStatus>
+  ): JobRunConnection {
+    return {
+      longestRunningActiveRun: fieldResolvers.JobRunConnection.longestRunningActiveRun(
+        runs
+      ),
+      nodes: runs.map(run => typeResolvers.JobRun(run))
```
nLight:
```js
import {JobRunResolver} from "./type_resolvers";

nodes: runs.map(JobRunResolver)
```

plugins/jobs/src/js/data/JobModel.ts
```diff
@@ -271,8 +254,280 @@ type Query {
...
+      if (scheduleLength === 0 && activeRunsLength === 0) {
+        return "UNSCHEDULED";
+      }
+
+      return "COMPLETED";
+    }
+  },
+  JobRun: {
+    jobID(
+      run: MetronomeClient.ActiveJobRun | MetronomeClient.JobHistoryRun
```
nLight: Would it reduce verbosity if we'd splat imported MetronomeClient? Does it clash with anything?
```js
run: ActiveJobRun | JobHistoryRun
```

plugins/jobs/src/js/data/JobModel.ts
```diff
@@ -281,13 +536,21 @@ export const resolvers = (): IResolvers => ({
     ): Observable<JobConnection | null> {
       return Observable.of(null);
     },
-    job(_obj = {}, _args: GeneralArgs, _context = {}): Observable<Job | null> {
-      return Observable.of(null);
+    job(_obj = {}, args: GeneralArgs, _context = {}): Observable<Job | null> {
+      // TODO: Write typeguard here
+      const { id } = args as JobQueryArgs;
```
nLight: Why do we declare `args` to be `GeneralArgs` and then coerce it?

src/js/structs/JobRunList.d.ts
```diff
@@ -1,6 +1,14 @@
+import { JobStatus } from "#SRC/js/events/MetronomeClient";
 import List from "./List";
 import JobRun from "./JobRun";
 
+interface JobRunItem {
+  status: JobStatus;
+}
+interface JobRunSchedule {}
```
nLight: Seems like we don't use the interface and it is empty

jest.config.js
juliangieseke: yeah, this PR is outdated 🤷‍♂️

plugins/jobs/src/js/data/JobModel.ts
DanielMSchmidt: Because we follow the GraphQL "contract" / type for resolvers. This types the args as an object with key and values as strings, because it

DanielMSchmidt: 's the way GraphQL sees them. I solved this better in my last commit, with a typeguard. That way we can just error if the resolver was called without the information needed.

plugins/jobs/src/js/types/Job.ts
```diff
@@ -24,16 +27,13 @@ import {
   JobStatusSchema
 } from "#PLUGINS/jobs/src/js/types/JobStatus";
 import { JobHistoryRun } from "#PLUGINS/jobs/src/js/types/JobHistoryRun";
```
nLight: Why have you decided to go with `#PLUGINS` alias even for the plugin itself?

juliangieseke: because vscode auto importer does it like that :trollface:

juliangieseke: I think we want to improve our tooling that we are able to write something like `jobsplugin/types/JobHistoryRun` like rxjs does

plugins/jobs/src/js/types/Job.ts
```diff
@@ -0,0 +1,200 @@
...
+} from "#PLUGINS/jobs/src/js/types/JobStatus";
+import { JobHistoryRun } from "#PLUGINS/jobs/src/js/types/JobHistoryRun";
+import {
+  JobDockerTypeResolver,
+  JobDocker,
+  JobDockerSchema
+} from "#PLUGINS/jobs/src/js/types/JobDocker";
+import { JobLabel, LabelSchema } from "#PLUGINS/jobs/src/js/types/JobLabel";
+import { cleanJobJSON } from "#SRC/js/utils/CleanJSONUtil";
+export interface Job {
```
Poltergeist: Could we add one empty line here to differentiate between import and export?

plugins/jobs/src/js/types/Job.ts
```diff
@@ -0,0 +1,200 @@
...
+  mem: Int!
+  name: String!
+  schedules: ScheduleConnection!
+  scheduleStatus: JobStatus!
+}
+`;
+
+export function JobTypeResolver(response: MetronomeJobDetailResponse): Job {
+  return {
+    id: JobFieldResolvers.id(response),
```
Poltergeist: Why do we need so many function calls?

plugins/jobs/src/js/types/JobTask.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export const JobTaskFieldResolvers = {
+  dateStarted(task: MetronomeJobRunTask): number {
+    return DateUtil.strToMs(task.createdAt);
+  },
+  dateCompleted(task: MetronomeJobRunTask): number | null {
+    return task.finishedAt ? DateUtil.strToMs(task.finishedAt) : null;
+  },
+  status(task: MetronomeJobRunTask): JobTaskStatus {
+    // TODO: this should use JobTaskStatusTypeResolver…
```
nLight: ...but it doesn't. Is it ok?

juliangieseke: JobTaskStatus is an enum (primitive) -> there is no type resolver 😅

nLight:
Legit build crash

plugins/jobs/src/js/types/Job.ts
Poltergeist: Okay got it, but why are the functions part of an object? I am scared that we at some point remove a resolver but we forget to remove the function for the resolver from the object and then we carry unused code around.

juliangieseke: @DanielMSchmidt :trollface:

juliangieseke:
@DanielMSchmidt can you check the fail? its related to labels, looks like our assumption was wrong 🤔 

```
ERROR in /jenkins/workspace/cos-ui-oss-pipeline_PR-3010-DXMPJI4A6DAJYLJ7ZVXJ76B745NT735AXR5TA2I4NN2S63VUFHRQ/src/js/events/__tests__/MetronomeClient-test.ts

[tsl] ERROR in /jenkins/workspace/cos-ui-oss-pipeline_PR-3010-DXMPJI4A6DAJYLJ7ZVXJ76B745NT735AXR5TA2I4NN2S63VUFHRQ/src/js/events/__tests__/MetronomeClient-test.ts(54,9)
      TS2322: Type '{ id: string; description: string; labels: {}; run: { cpus: number; mem: number; disk: number; cm...' is not assignable to type 'JobDetailResponse'.
  Types of property 'labels' are incompatible.
    Type '{}' is not assignable to type 'JobLabels'.
      Property 'key' is missing in type '{}'.
```

plugins/jobs/src/js/types/Job.ts
DanielMSchmidt: @Poltergeist 

a) Code is always subject to change, why is this here different then somewhere else?
b) These will be inlined later to a place where they are right beside their schema. Literally 10 lines away. The will also not be called by hand but by graphql. If three of us overlook this while coding and reviewing we deserve to have stray code around. 

I honestly think you have a point here in general and we should file a jira to research tooling to find dead code, but I think this is unrelated to this PR

plugins/jobs/src/js/types/Job.ts
DanielMSchmidt: Done

plugins/jobs/src/js/types/JobTask.ts
DanielMSchmidt: Removed

DanielMSchmidt:
@nLight @juliangieseke @Poltergeist Build worked now (at least on my machine), could you review it again?

plugins/jobs/src/js/types/Job.ts
Poltergeist: 👍 That is why this is not blocking 👍

nLight:
🤞

PR #3011, opened 2018-06-12
DCOS-38114 : fix(tables/styles): fix FF collapsed table td 

weblancaster:
**Issue:**
The label is collapsed on Firefox, breaking the ui.

**Fix:**
Removes the `max-width: 0;` which was causing the ui to be `width 0`.

**Result:**
FF and Chrome tested.

**Test:**
Please test on Chrome and FF to make sure is wor

**Before**
screen shot 2018-06-11 at 3 02 21 pm

**After**
screen shot 2018-06-12 at 10 56 32 am

**Checklist**

Closes DCOS-38114

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@ build/
 npm-debug.log
 node_modules/
 src/js/config/Config.dev.js
+src/js/config/Config.dev.ts
 serve
 tmp/
 *.orig
--- a/src/styles/content/tables/styles.less
+++ b/src/styles/content/tables/styles.less
@@ -103,9 +103,7 @@
 
     th,
     td {
-      // This property is necessary for table cells to respect the text-overflow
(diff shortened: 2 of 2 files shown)


nLight:
It's in review but in the future sprint 🤔

DanielMSchmidt:
@nLight I moved it into the sprint to reflect the work being done on this.

.gitignore
```diff
@@ -9,6 +9,7 @@ build/
 npm-debug.log
 node_modules/
 src/js/config/Config.dev.js
+src/js/config/Config.dev.ts
```
DanielMSchmidt: Why do we need this? Can't we just remove the file when we switch branches?

nLight: Could be annoying when switching between 1.11 and master in the same dir. I'd say we merge it and bump the release until the deadline. As it is not a blocker is it?

weblancaster: Correct, not a blocker

PR #3013, opened 2018-06-13
DCOS-38126 - list of jobs resolver

Fabs:
This PR is about the jobs resolver who allows us to fetch a list of jobs. For this PR to go, we still need some more work to implement `jobListToGraphql` from the work on #3010. 

In order to make it compliant to the Job schema, we needed to do an N + 1 query. We have to fix that soon since it is far from prod ready. https://jira.mesosphere.com/browse/DCOS-38201

Testing this is tricky since it is not connected to any view yet. But you can trust my world on it 😄 or play with the tests to convince yourself.

**Checklist**


plugins/jobs/src/js/data/JobModel.ts
```diff
@@ -289,5 +382,9 @@ export const resolvers = (): IResolvers => ({
 
 export default makeExecutableSchema({
   typeDefs,
-  resolvers: resolvers()
+  resolvers: resolvers({
+    fetchJobs: MetronomeClient.fetchJobs,
+    fetchJobDetail: MetronomeClient.fetchJobDetail,
+    pollingInterval: 2000
```
Fabs: What is the constant that goes here again?

plugins/jobs/src/js/data/JobModel.ts
```diff
@@ -264,22 +267,112 @@ type Query {
...
+  filter: string | null
+): MetronomeClient.JobResponse[] {
+  if (filter === null) {
+    return jobs;
+  }
+
+  return jobs.filter(({ id }) => id.indexOf(filter) !== -1);
+}
+
+function jobListToGraphql(response: MetronomeClient.JobDetailResponse): Job {
```
Fabs: replace that with jobListToGraphql

plugins/jobs/src/js/data/JobModel.ts
DanielMSchmidt: `Config.getRefreshRate()`

Fabs: Thanks.

PR #3014, opened 2018-06-14
DCOS-38126 - list of jobs resolver

Fabs:
This PR is about the jobs resolver who allows us to fetch a list of jobs. 

In order to make it compliant to the Job schema, we needed to do an N + 1 query. We have to fix that soon since it is far from prod ready. https://jira.mesosphere.com/browse/DCOS-38201

Testing this is tricky since it is not connected to any view yet. But you can trust my world on it 😄 or play with the tests to convince yourself.

**Checklist**

### DIFF
--- a/plugins/jobs/src/js/data/JobModel.ts
+++ b/plugins/jobs/src/js/data/JobModel.ts
@@ -122,10 +122,10 @@ function sortJobByStatus(a: Job, b: Job): number {
 }
 
 function sortJobByLastRun(a: Job, b: Job): number {
-  return (
-    JobStatus[a.lastRunStatus.status].sortOrder -
-    JobStatus[b.lastRunStatus.status].sortOrder
-  );
+  const statusA = a.lastRunStatus ? a.lastRunStatus.status : "N/A";
+  const statusB = b.lastRunStatus ? b.lastRunStatus.status : "N/A";
+
+  return JobStatus[statusA].sortOrder - JobStatus[statusB].sortOrder;
 }
 
 function filterJobsById(
@@ -176,21 +176,15 @@ export const resolvers = ({
       );
       const count$ = jobsInNamespace$.map(jobs => jobs.length);
 
-      return (
-        jobsInNamespace$
-          .map(jobs => filterJobsById(jobs, filter))
-          //TODO We need to remove the N + 1 query
-          // https://jira.mesosphere.com/browse/DCOS-38201
-          .map(jobs => jobs.map(job => fetchJobDetail(job.id)))
-          .switchMap(observable$ => Observable.combineLatest(observable$))
-          .map(requests => requests.map(JobTypeResolver))
-          .map(jobs => sortJobs(jobs, sortBy, sortDirection))
-          .combineLatest(count$, (jobs, totalCount) => ({
-            filteredCount: jobs.length,
-            totalCount,
-            nodes: jobs
-          }))
-      );
+      return jobsInNamespace$
+        .map(jobs => filterJobsById(jobs, filter))
+        .map(jobs => jobs.map(JobTypeResolver))
+        .map(jobs => sortJobs(jobs, sortBy, sortDirection))
+        .combineLatest(count$, (jobs, totalCount) => ({
+          filteredCount: jobs.length,
+          totalCount,
+          nodes: jobs
+        }));
     },
     job(_obj = {}, args: GeneralArgs, _context = {}): Observable<Job | null> {
       if (!isJobQueryArg(args)) {
--- a/plugins/jobs/src/js/data/JobModel.ts
+++ b/plugins/jobs/src/js/data/JobModel.ts
@@ -91,13 +91,13 @@ function sortJobs(
 
     switch (sortBy) {
       case "ID":
-        result = sortJobById(a, b);
+        result = compareJobById(a, b);
         break;
       case "STATUS":
-        result = sortJobByStatus(a, b);
+        result = compareJobByStatus(a, b);
         break;
       case "LAST_RUN":
-        result = sortJobByLastRun(a, b);
+        result = compareJobByLastRun(a, b);
         break;
       default:
         result = 0;
@@ -110,18 +110,18 @@ function sortJobs(
   return jobs;
 }
 
-function sortJobById(a: Job, b: Job): number {
+function compareJobById(a: Job, b: Job): number {
   return a.id.localeCompare(b.id);
 }
 
-function sortJobByStatus(a: Job, b: Job): number {
+function compareJobByStatus(a: Job, b: Job): number {
   return (
     JobStates[a.scheduleStatus].sortOrder -
     JobStates[b.scheduleStatus].sortOrder
   );
 }
 
-function sortJobByLastRun(a: Job, b: Job): number {
+function compareJobByLastRun(a: Job, b: Job): number {
   const statusA = a.lastRunStatus ? a.lastRunStatus.status : "N/A";
   const statusB = b.lastRunStatus ? b.lastRunStatus.status : "N/A";
 
--- a/plugins/jobs/src/js/data/JobModel.ts
+++ b/plugins/jobs/src/js/data/JobModel.ts
@@ -1,11 +1,15 @@
 import { makeExecutableSchema, IResolvers } from "graphql-tools";
 import { Observable } from "rxjs/Observable";
 import {
+  fetchJobs,
   fetchJobDetail,
+  JobResponse as MetronomeJobResponse,
   JobDetailResponse as MetronomeJobDetailResponse
 } from "#SRC/js/events/MetronomeClient";
 
 import Config from "#SRC/js/config/Config";
+import JobStates from "#PLUGINS/jobs/src/js/constants/JobStates";
+import JobStatus from "#PLUGINS/jobs/src/js/constants/JobStatus";
 import {
   JobConnection,
   JobConnectionSchema
@@ -22,6 +26,7 @@ export interface Query {
 }
 
 export interface ResolverArgs {
+  fetchJobs: () => Observable<MetronomeJobResponse[]>;
   fetchJobDetail: (id: string) => Observable<MetronomeJobDetailResponse>;
   pollingInterval: number;
 }
@@ -32,8 +37,9 @@ export interface GeneralArgs {
 
 export interface JobsQueryArgs {
   filter?: string | null;
-  sortBy?: SortOption | null;
-  sortDirection?: SortDirection | null;
+  namespace?: string | null;
+  sortBy?: SortOption;
+  sortDirection?: SortDirection;
 }
 
 export interface JobQueryArgs {
@@ -60,14 +66,12 @@ export const typeDefs = `
   type Query {
     jobs(
       filter: String
+      namespace: String
       sortBy: SortOption
       sortDirection: SortDirection
     ): JobConnection
     job(
       id: ID!
-      filter: String
-      sortBy: SortOption
-      sortDirection: SortDirection
     ): Job
   }
   `;
@@ -76,17 +80,117 @@ function isJobQueryArg(arg: any): arg is JobQueryArgs {
   return arg.id !== undefined;
 }
 
+// Sort and filtering
+function sortJobs(
+  jobs: Job[],
+  sortBy: SortOption = "ID",
+  sortDirection: SortDirection = "ASC"
+): Job[] {
+  jobs.sort((a, b) => {
+    let result;
+
+    switch (sortBy) {
+      case "ID":
+        result = sortJobById(a, b);
+        break;
+      case "STATUS":
+        result = sortJobByStatus(a, b);
+        break;
+      case "LAST_RUN":
+        result = sortJobByLastRun(a, b);
+        break;
+      default:
+        result = 0;
+    }
+
+    const direction = sortDirection === "ASC" ? 1 : -1;
+    return result * direction;
+  });
+
+  return jobs;
+}
+
+function sortJobById(a: Job, b: Job): number {
+  return a.id.localeCompare(b.id);
+}
+
+function sortJobByStatus(a: Job, b: Job): number {
+  return (
+    JobStates[a.scheduleStatus].sortOrder -
+    JobStates[b.scheduleStatus].sortOrder
+  );
+}
+
+function sortJobByLastRun(a: Job, b: Job): number {
+  return (
+    JobStatus[a.lastRunStatus.status].sortOrder -
+    JobStatus[b.lastRunStatus.status].sortOrder
+  );
+}
+
+function filterJobsById(
+  jobs: MetronomeJobResponse[],
+  filter: string | null
+): MetronomeJobResponse[] {
+  if (filter === null) {
+    return jobs;
+  }
+
+  return jobs.filter(({ id }) => id.indexOf(filter) !== -1);
+}
+
+function filterJobsByNamespace(
+  jobs: MetronomeJobResponse[],
+  namespace: string | null
+): MetronomeJobResponse[] {
+  if (namespace === null) {
+    return jobs;
+  }
+
+  return jobs.filter(({ id }) => id.startsWith(namespace));
+}
+
 export const resolvers = ({
+  fetchJobs,
   fetchJobDetail,
   pollingInterval
 }: ResolverArgs): IResolvers => ({
   Query: {
     jobs(
-      _obj = {},
-      _args: GeneralArgs = {},
+      _parent = {},
+      args: GeneralArgs = {},
       _context = {}
     ): Observable<JobConnection | null> {
-      return Observable.of(null);
+      const {
+        sortBy = "ID",
+        sortDirection = "ASC",
+        filter = null,
+        namespace = null
+      } = args as JobsQueryArgs;
+
+      const pollingInterval$ = Observable.timer(0, pollingInterval);
+      const responses$ = pollingInterval$.exhaustMap(fetchJobs);
+
+      const jobsInNamespace$ = responses$.map(jobs =>
+        filterJobsByNamespace(jobs, namespace)
+      );
+      const count$ = jobsInNamespace$.map(jobs => jobs.length);
+
+      return (
+        jobsInNamespace$
+          .map(jobs => filterJobsById(jobs, filter))
+          //TODO We need to remove the N + 1 query
+          // https://jira.mesosphere.com/browse/DCOS-38201
+          .map(jobs => jobs.map(job => fetchJobDetail(job.id)))
+          .switchMap(observable$ => Observable.combineLatest(observable$))
+          .map(requests => requests.map(JobTypeResolver))
+          .map(jobs => sortJobs(jobs, sortBy, sortDirection))
+          .combineLatest(count$, (jobs, totalCount) => ({
+            filteredCount: jobs.length,
+            totalCount,
+            nodes: jobs
+          }))
+      );
     },
     job(_obj = {}, args: GeneralArgs, _context = {}): Observable<Job | null> {
       if (!isJobQueryArg(args)) {
@@ -106,6 +210,7 @@ export const resolvers = ({
 export default makeExecutableSchema({
   typeDefs,
   resolvers: resolvers({
+    fetchJobs,
     fetchJobDetail,
     pollingInterval: Config.getRefreshRate()
   })
--- a/plugins/jobs/src/js/types/Job.ts
+++ b/plugins/jobs/src/js/types/Job.ts
@@ -1,6 +1,9 @@
 import {
   JobDetailResponse as MetronomeJobDetailResponse,
-  JobStatus as MetronomeJobStatus
+  GenericJobResponse as MetronmeGenericJobResponse,
+  JobResponse as MetronomeJobResponse,
+  JobStatus as MetronomeJobStatus,
+  isDetailResponse as isMetronomeJobDetailResponse
 } from "#SRC/js/events/MetronomeClient";
 import {
   JobRunConnection,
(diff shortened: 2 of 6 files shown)


plugins/jobs/src/js/data/JobModel.ts
```diff
@@ -524,17 +527,98 @@ const fieldResolvers = {
...
-      return Observable.of(null);
+    ): Observable<Job[] | null> {
+      const {
+        sortBy = "ID",
+        sortDirection = "ASC",
+        filter = null
+      } = args as JobsQueryArgs;
+
+      const pollingInterval$ = Observable.timer(0, pollingInterval);
+      const responses$ = pollingInterval$.switchMap(fetchJobs);
```
juliangieseke: Here we used `exhaustMap` in our first implementation 🤔

plugins/jobs/src/js/data/JobModel.ts
```diff
@@ -524,17 +527,98 @@ const fieldResolvers = {
...
+      return (
+        responses$
+          .map(jobs => filterJobs(jobs, filter))
+          // N+1 request madness
+          .map(jobs => jobs.map(job => fetchJobDetail(job.id)))
+          .switchMap(obs => Observable.combineLatest(obs))
+          .map(requests => requests.map(typeResolvers.Job))
+          .map(jobs =>
+            //TODO Why typescript can't figure out that these value is always
+            // defined?
```
juliangieseke: because `sortBy` and `sortDirection` aren't typed, this should work 🤔 
```
+      const {
+        sortBy: SortOption = "ID",
+        sortDirection: SortDirection = "ASC",
+        filter: string | null = null
+      } = args;
```

plugins/jobs/src/js/data/JobModel.ts
```diff
@@ -524,17 +527,98 @@ const fieldResolvers = {
...
+
+      const pollingInterval$ = Observable.timer(0, pollingInterval);
+      const responses$ = pollingInterval$.switchMap(fetchJobs);
+
+      return (
+        responses$
+          .map(jobs => filterJobs(jobs, filter))
+          // N+1 request madness
+          .map(jobs => jobs.map(job => fetchJobDetail(job.id)))
+          .switchMap(obs => Observable.combineLatest(obs))
```
juliangieseke: what is this doing? 🤔

plugins/jobs/src/js/data/JobModel.ts
Fabs: We used, and then we had that discussion in the Curiosity room with Orlando. switchMap is also not the best solution here. We have to do the pooling ticket Daniel added to the sprint, and that you requested on the http-service.

plugins/jobs/src/js/data/JobModel.ts
```diff
@@ -524,17 +527,98 @@ const fieldResolvers = {
...
+        filter = null
+      } = args as JobsQueryArgs;
+
+      const pollingInterval$ = Observable.timer(0, pollingInterval);
+      const responses$ = pollingInterval$.switchMap(fetchJobs);
+
+      return (
+        responses$
+          .map(jobs => filterJobs(jobs, filter))
+          // N+1 request madness
```
DanielMSchmidt: I would rather like to see a comment pointing to a JIRA ticket that is about removing the n+1 query

plugins/jobs/src/js/data/JobModel.ts
Fabs: You are right. I solved that this way before, and then I rebased and lost that.

DanielMSchmidt: We have the `JobsQueryArgs`, they are generated from GraphQL and well typed. You would need to write a [user defined type guard](https://github.com/basarat/typescript-book/blob/master/docs/types/typeGuard.md#user-defined-type-guards) to "cast" it into the right type. I would prefer the generated code over the handwritten TBH, I don't know how you feel about this @juliangieseke

plugins/jobs/src/js/data/JobModel.ts
```diff
@@ -23,8 +25,8 @@ export interface JobsQueryArgs {
 
 export interface JobQueryArgs {
   id: string;
-  filter?: string | null;
-  sortBy?: SortOption | null;
+  filter?: string;
```
DanielMSchmidt: Why did you remove the null?

plugins/jobs/src/js/data/JobModel.ts
juliangieseke: generally yes, but it looks like its already messed up, not sure if we are able to re-generate 😢

plugins/jobs/src/js/data/JobModel.ts
Fabs: There is a fiddle here where you can play with the code:
https://jsfiddle.net/fabsn/x5zaf129/23/

The main idea is, you have the first Observable<Jobs[]> (an observable that emits jobs). So you have to map in a request for every job. That is what line 613 makes `Jobs[] => Observable<JobsWithDetail>[]`. But a list of Observables (and not an observable of lists) is not very interesting because the list data structure is boring :). So we use Observable.combineLatest to convert `Observable<JobsWithDetail>[] => Observable<JobsWithDetail[]>`, this is line 614 map part. 

At this step you Have an Observable<Observable<JobsWithDetail[]>>, and you need switchMap (or other strategy) to remove the nesting. This is the switch part of line 614.

plugins/jobs/src/js/data/JobModel.ts
Fabs: Sure.

plugins/jobs/src/js/data/JobModel.ts
Fabs: I think that was also an attempt to fix the same problem we discussed above.

But on another pass, I don't think the null should be necessary. I think by saying the parameter value is optional, you pass it or not (I know that is just covering the undefined). I see no point in assigning null to the value if for the filter, or for the sorting, the value of null has no specific meaning.

DanielMSchmidt: yeah, I think that's fine with me 🤷‍♂️

plugins/jobs/src/js/data/JobModel.ts
juliangieseke: Hm, the discussion was about another delay after an already delayed response finished, not about canceling (or not) an already running request 🤔

Fabs: Yes. But both exhaust and switch have their advantages. Switch will keep dropping and never complete, but exhaust also would have older data for a while until it gets stuck.

We need the polling feature on next sprint.

Fabs:
@DanielMSchmidt @juliangieseke please also do not forget this fix:
+      status: run.status, // TODO: derive from where it comes from before passing to this function

plugins/jobs/src/js/data/JobModel.ts
DanielMSchmidt: Agreed with @Fabs this is not the right place to solve it :)

juliangieseke: My comment wasnt meant to block this PR… 🤷‍♂️

plugins/jobs/src/js/data/JobModel.ts
Fabs: I actually was not able to type on destruction, I made the item non nullable.

plugins/jobs/src/js/data/JobModel.ts
```diff
@@ -524,17 +527,99 @@ const fieldResolvers = {
...
+    }
+
+    const direction = sortDirection === "ASC" ? 1 : -1;
+    return result * direction;
+  });
+
+  return jobs;
+}
+
+function sortJobById(a: Job, b: Job): number {
```
juliangieseke: 🤚 this _compares_ jobs => `compareJobById`?

plugins/jobs/src/js/data/JobModel.ts
```diff
@@ -524,17 +527,99 @@ const fieldResolvers = {
...
+  return a.id.localeCompare(b.id);
+}
+
+function sortJobByStatus(a: Job, b: Job): number {
```
juliangieseke: 🤚 same as above

plugins/jobs/src/js/data/JobModel.ts
```diff
@@ -524,17 +527,99 @@ const fieldResolvers = {
...
+  return (
+    JobStates[a.scheduleStatus].sortOrder -
+    JobStates[b.scheduleStatus].sortOrder
+  );
+}
+
+function sortJobByLastRun(a: Job, b: Job): number {
```
juliangieseke: 🤚 same as above

plugins/jobs/src/js/data/JobModel.ts
```diff
@@ -524,17 +527,99 @@ const fieldResolvers = {
...
+      _parent = {},
+      args: GeneralArgs = {},
       _context = {}
-    ): Observable<JobConnection | null> {
-      return Observable.of(null);
+    ): Observable<Job[] | null> {
+      const {
+        sortBy = "ID",
+        sortDirection = "ASC",
+        filter = null
```
juliangieseke: 🤚 filter is never `null`, default should be "" (at least from the old implementation)

plugins/jobs/src/js/data/JobModel.ts
```diff
@@ -524,17 +527,99 @@ const fieldResolvers = {
...
+        responses$
+          .map(jobs => filterJobs(jobs, filter))
+          //TODO We need to remove the N + 1 query
+          // https://jira.mesosphere.com/browse/DCOS-38201
+          .map(jobs => jobs.map(job => fetchJobDetail(job.id)))
+          .switchMap(obs => Observable.combineLatest(obs))
+          .map(requests => requests.map(typeResolvers.Job))
+          .map(jobs =>
+            //TODO Why typescript can't figure out that these value is always
+            // defined?
```
juliangieseke: 🎨 if this is fixed, can we remove the comment?

juliangieseke:
@Fabs your issue is because the files have moved, did you rebase or something?

plugins/jobs/src/js/data/JobModel.ts
Fabs: I changed that because it helps me solve the `blabla || null`

plugins/jobs/src/js/data/JobModel.ts
Fabs: Are you ok to the switchMap then?

plugins/jobs/src/js/data/JobModel.ts
```diff
@@ -524,17 +527,99 @@ const fieldResolvers = {
...
   Query: {
     jobs(
-      _obj = {},
-      _args: GeneralArgs = {},
+      _parent = {},
+      args: GeneralArgs = {},
       _context = {}
-    ): Observable<JobConnection | null> {
-      return Observable.of(null);
+    ): Observable<Job[] | null> {
```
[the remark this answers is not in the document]
Fabs: Hum. I see that, we need to add the countings. I would say ✋. That was an oversight on my side and needs fixing.

plugins/jobs/src/js/data/JobModel.ts
```diff
@@ -17,14 +19,14 @@ export interface GeneralArgs {
 
 export interface JobsQueryArgs {
   filter?: string | null;
-  sortBy?: SortOption | null;
-  sortDirection?: SortDirection | null;
+  sortBy?: SortOption;
```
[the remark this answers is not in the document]
Fabs: I agree with your point, but from my idea of how filters and sort work, you could specify no filter and no sorting. I would even like that because sort is expensive. 

In any case, I would not break compatibility, and would them do the types as you propose.

plugins/jobs/src/js/data/JobModel.ts
```diff
@@ -4,6 +4,8 @@ import * as MetronomeClient from "#SRC/js/events/MetronomeClient";
 import JobRunList from "#SRC/js/structs/JobRunList";
 import DateUtil from "#SRC/js/utils/DateUtil";
 import { cleanJobJSON } from "#SRC/js/utils/CleanJSONUtil";
+import JobStates from "#SRC/js/constants/JobStates";
+import JobStatus from "#SRC/js/constants/JobStatus";
```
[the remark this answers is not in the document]
Fabs: ah, ok. That is what broke it them. Next 🤠 dev should fix it :D

plugins/jobs/src/js/data/JobModel.ts
Fabs: You are right, this is the compare function and not the sort function. My bad to blind copy and paste.

plugins/jobs/src/js/data/JobModel.ts
Fabs: Agree

plugins/jobs/src/js/data/JobModel.ts
Fabs: yes :D

plugins/jobs/src/js/data/JobModel.ts
Fabs: Should, I believe I lost this one on rebase this morning. But you are right.

plugins/jobs/src/js/data/JobModel.ts
Fabs: I removed that. I rebased it.

plugins/jobs/src/js/data/JobModel.ts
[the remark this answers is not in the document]
juliangieseke: We can guard filtering by checking against empty string

plugins/jobs/src/js/types/Job.ts
```diff
@@ -84,25 +84,25 @@ type Job {
...
-    json: fieldResolvers.json(response),
-    labels: fieldResolvers.labels(response),
-    lastRunStatus: fieldResolvers.lastRunStatus(response),
-    name: fieldResolvers.name(response),
-    jobRuns: fieldResolvers.jobRuns(response),
-    lastRunsSummary: fieldResolvers.lastRunsSummary(response),
-    scheduleStatus: fieldResolvers.scheduleStatus(response),
-    activeRuns: fieldResolvers.activeRuns(response),
-    schedules: fieldResolvers.schedules(response)
+    id: JobFieldResolvers.id(response),
```
juliangieseke: gnah, this needs to be in the other PR 😩

juliangieseke: it is 🤔

plugins/jobs/src/js/data/JobModel.ts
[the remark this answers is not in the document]
DanielMSchmidt: I disagree, it is perfectly fine to have it undefined or null here, why would we want to force the user of the datalayer to always filter or sort?

juliangieseke:
recovered fabs commits here, lets try again on monday…

juliangieseke:
rebased again, this time it should be fine, but there is one failing test we need to investigate… ( I think @DanielMSchmidt can fix them pretty fast - he fixed something similar in the previous PR)

plugins/jobs/src/js/data/JobModel.ts
juliangieseke: As discussed offline with orlando: 
```
Rx.Observable
  .concat(Rx.Observable.from([1]))
  .delay(1000)
  .repeat(-1)
```

plugins/jobs/src/js/data/JobModel.ts
```diff
@@ -76,17 +80,119 @@ function isJobQueryArg(arg: any): arg is JobQueryArgs {
...
+        sortDirection = "ASC",
+        filter = null,
+        namespace = null
+      } = args as JobsQueryArgs;
+
+      const pollingInterval$ = Observable.timer(0, pollingInterval);
+      const responses$ = pollingInterval$.exhaustMap(fetchJobs);
+
+      const jobsInNamespace$ = responses$.map(jobs =>
+        filterJobsByNamespace(jobs, namespace)
```
nLight: we could have created a curried function here :trollface: 
```js
const jobsInNamespace$ = responses$.map(getNamespaceJobsFilter(namespace))
```

plugins/jobs/src/js/data/JobModel.ts
```diff
@@ -76,17 +80,119 @@ function isJobQueryArg(arg: any): arg is JobQueryArgs {
...
+      );
+      const count$ = jobsInNamespace$.map(jobs => jobs.length);
+
+      return (
+        jobsInNamespace$
+          .map(jobs => filterJobsById(jobs, filter))
+          //TODO We need to remove the N + 1 query
+          // https://jira.mesosphere.com/browse/DCOS-38201
+          .map(jobs => jobs.map(job => fetchJobDetail(job.id)))
+          .switchMap(obs => Observable.combineLatest(obs))
```
nLight: I'm sorry but when I see `jobs` and `obs` close to each other I can't stop thinking there's a `j` missing. Let's not abbreviate variables 💪 plus if argument is an Observable we may wanna consider adding `$` at the end

plugins/jobs/src/js/data/JobModel.ts
```diff
@@ -76,17 +80,119 @@ function isJobQueryArg(arg: any): arg is JobQueryArgs {
...
+        filterJobsByNamespace(jobs, namespace)
+      );
+      const count$ = jobsInNamespace$.map(jobs => jobs.length);
+
+      return (
+        jobsInNamespace$
+          .map(jobs => filterJobsById(jobs, filter))
+          //TODO We need to remove the N + 1 query
+          // https://jira.mesosphere.com/browse/DCOS-38201
+          .map(jobs => jobs.map(job => fetchJobDetail(job.id)))
```
nLight: Why do we even need N+1 here? Can't we just have a separate stream for the `JobDetailPageContainer`?

plugins/jobs/src/js/data/JobModel.ts
DanielMSchmidt: I would argue that this is unnecessarily complicated to read, but if you like 🤷‍♂️

plugins/jobs/src/js/data/JobModel.ts
DanielMSchmidt: We made two decisions in the past:

1) We want to have one single GraphQL schema
2) We start with a lightweight GraphQL implementation that does not support field resolvers by now

Currently we need to resolve the type `Job` in the index response as part of the `JobsConnection` (Decision 1). This means we need to give the `Job` type in it's entireness to the GraphQL implementation (this would be lazy instead if we had field resolvers) (Decision 2). Because the `Job` type needs more information then the index API endpoint delivers we now need to fetch the detail endpoint to not break the contract of GraphQL.

plugins/jobs/src/js/data/JobModel.ts
nLight: no-no, that was more a comment to troll @Fabs, please ignore

Fabs: I actually started with the partially applied function, then I realized the elegance might have a negative impact on comprehension.

plugins/jobs/src/js/data/JobModel.ts
```diff
@@ -176,21 +176,15 @@ export const resolvers = ({
...
-          .switchMap(observable$ => Observable.combineLatest(observable$))
-          .map(requests => requests.map(JobTypeResolver))
-          .map(jobs => sortJobs(jobs, sortBy, sortDirection))
-          .combineLatest(count$, (jobs, totalCount) => ({
-            filteredCount: jobs.length,
-            totalCount,
-            nodes: jobs
-          }))
-      );
+      return jobsInNamespace$
```
juliangieseke: since there is no n+1 madness anymore, we should be able to simply do 
```
responses$
  .map(jobs => JobConnectionTypeResolver(jobs))
  .map(jobConnection => sortJobs(jobConnection, sortBy, sortDirection);
```
which then contains all the needed information, so no mapping over all jobs (here) and `combineLatest` necessary anymore.

plugins/jobs/src/js/types/Job.ts
```diff
@@ -64,26 +66,28 @@ ${JobDockerSchema}
...
   mem: Int!
   name: String!
   schedules: ScheduleConnection!
   scheduleStatus: JobStatus!
 }
 `;
 
-export function JobTypeResolver(response: MetronomeJobDetailResponse): Job {
+export function JobTypeResolver(
+  response: MetronomeJobDetailResponse | MetronomeJobResponse
```
juliangieseke: 😭

plugins/jobs/src/js/types/Job.ts
```diff
@@ -103,72 +107,102 @@ export function JobTypeResolver(response: MetronomeJobDetailResponse): Job {
...
+          return "SCHEDULED";
+        }
       }
     }
 
     if (
-      scheduleConnection.nodes.length === 0 &&
-      JobRunConnectionTypeResolver(job.activeRuns).nodes.length === 0
+      scheduleConnection.nodes.length === 0 // &&
+      // TODO: clairify this if
```
juliangieseke: ?

plugins/jobs/src/js/types/Job.ts
```diff
@@ -103,73 +107,99 @@ export function JobTypeResolver(response: MetronomeJobDetailResponse): Job {
...
+  labels(job: MetronomeJobDetailResponse | MetronomeJobResponse): JobLabel[] {
     return Object.entries(job.labels).map(([key, value]) => ({ key, value }));
   },
-  lastRunsSummary(job: MetronomeJobDetailResponse): JobHistorySummary {
-    return JobHistorySummaryTypeResolver(job.history);
+  lastRunsSummary(
+    job: MetronomeJobDetailResponse | MetronomeJobResponse
+  ): JobHistorySummary | null {
+    return isMetronomeJobDetailResponse(job)
+      ? JobHistorySummaryTypeResolver(job.history)
```
juliangieseke: this needs a small fixup, its `job.history` but `jobs[].historySummary` - content is almost the same tho

nLight: let's fix this

DanielMSchmidt:
@juliangieseke @nLight This is ready again, could you take another look. Please have in mind that we can change this file, later on, let's not feature creep 🙈

src/js/events/MetronomeClient.ts
```diff
@@ -5,43 +5,15 @@ import { Observable } from "rxjs/Observable";
...
-  id: string;
-  description: string;
   labels: LabelResponse;
   run: {
     cpus: number;
     mem: number;
     disk: number;
     cmd: string;
-    env: {};
+    env: object;
```
nLight: 👍

DanielMSchmidt: 👍

PR #3015, opened 2018-06-14
DCOS-37767 Simplifies the reference implementation to make it easier to use the data-layer

Fabs:
This is a "small" refactor to remove the use of context, and to better name the client for the reference implementation so it makes easier for people to use the reference implementation from the data layer. 

It was motivated by my office hours with bill where I struggled to explain why we did this way, and also from orlando feedback that this is super complicated. 

*To test it*.

Check out the code, build against a cluster and then test all the operations of package repository (lis them, filter them, remove them, add them).


plugins/catalog/src/js/repositories/RepositoriesList.js
```diff
@@ -1,101 +1,50 @@
-import React from "react";
```
Fabs: This file makes no sense, I am fixing that.

plugins/catalog/src/js/repositories/RepositoriesList.js
```diff
@@ -8,16 +8,11 @@ import "rxjs/add/operator/switchMap";
 import { componentFromStream, graphqlObservable } from "data-service";
 
 // tools to make easier to create schemas and queries
-import { makeExecutableSchema } from "graphql-tools";
+// import { makeExecutableSchema } from "graphql-tools";
```
nLight: Do we still want to keep the comment?

plugins/catalog/src/js/repositories/data/repositoriesModel.js
```diff
@@ -4,6 +4,13 @@ import "rxjs/add/operator/map";
 import RepositoryList from "#SRC/js/structs/RepositoryList";
 import { makeExecutableSchema } from "graphql-tools/dist/index";
 
+// Streams we get our data from
+import {
+  liveFetchRepositories,
+  addRepository,
+  deleteRepository
+} from "#PLUGINS/catalog/src/js/repositories/data/PackageRepositoryClient";
```
nLight: VSCode does that for me sometimes. So I wonder if it's your IDE or you've deliberately chosen to use the `#PLUGINS` alias?

plugins/catalog/src/js/repositories/RepositoriesList.js
Fabs: I would say yes, it is commented because in this case we got the schema built on line 18. But I think this comment goes well with line 12 where we explain how it is built. But if I move them closer, when you uncomment the linter yells at you because of the import order. 

It is not ideal, but I would leave it like this.

plugins/catalog/src/js/repositories/data/repositoriesModel.js
Fabs: No, I did not. I think for imports inside a plugin, it makes no sense to use the plugins path. I think it was a global switch and replace or the editor. 

I appreciate you did not block because of this one. I will move further like this, this time because I know there will be opportunities to correct that in the future as we move forward, and because it is not a huge deal.

PR #3016, opened 2018-06-14
DCOS-37751 & DCOS-37752: Intro JobsOverview Mediator

juliangieseke:
This implements the mediator for jobs overview.

~there is not much to test because this mediator isn in use yet 🤷‍♂️~ decided to add both subtasks in this PR. please review per commit.


plugins/jobs/src/js/JobsOverview.tsx
```diff
@@ -0,0 +1,116 @@
...
+        id
+        lastRunsSummary
+        lastRunStatus
+        scheduleStatus
+      }
+    }
+  }
+`;
+
+const DEFAULT_FILTERS: JobsQueryArgs = {
```
juliangieseke: todo: figure out how to reset the stream instead of setting its value back to some defaults

plugins/jobs/src/js/JobsOverview.tsx
```diff
@@ -0,0 +1,116 @@
...
+  filter: "",
+  sortBy: "ID",
+  sortDirection: "ASC"
+};
+const filter$ = new BehaviorSubject<JobsQueryArgs>(DEFAULT_FILTERS);
+
+const jobs$ = filter$.debounceTime(250).switchMap((filters: JobsQueryArgs) => {
+  const { filter = "", sortBy, sortDirection } = filters;
```
juliangieseke: todo: here i have the same issue fabs had, I want to get rid of the default here

plugins/jobs/src/js/JobsOverview.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+  return graphqlObservable(
+    jobsOverviewQuery(filter, sortBy, sortDirection),
+    JobModel,
+    {}
+  ).map((jobs: JobConnection) => ({ filters, jobs }));
```
juliangieseke: todo: verify that this is correct

plugins/jobs/src/js/JobsOverview.tsx
```diff
@@ -0,0 +1,116 @@
...
+});
+
+const JobsOverview = componentFromStream((prop$: any) => {
+  return prop$
+    .combineLatest(jobs$)
+    .distinctUntilChanged()
+    .map((
+      filters: JobsQueryArgs,
+      props: JobsOverviewProps, // TODO pick up changes from props (props.params.id)
```
juliangieseke: i think this is possible before entering this map… 🤔

plugins/jobs/src/js/JobsOverview.tsx
```diff
@@ -0,0 +1,116 @@
...
+          filteredJobs={filteredJobs}
+          searchString={searchString}
+          handleFilterChange={handleFilterChange}
+          resetFilter={resetFilter}
+          hasFilterApplied={hasFilterApplied}
+        />
+      );
+
+      // TODO UI refactoring part 2
+      // return (
```
juliangieseke: this is how it should be 😅

plugins/jobs/src/js/JobsOverview.tsx
```diff
@@ -0,0 +1,116 @@
...
+      //   <JobsTabDummy
+      //     jobs={jobs}
+      //     filters={filters}
+      //     onChange={(filter, sortBy, sortDirection) =>
+      //       filter$.next({ filter, sortBy, sortDirection })
+      //     }
+      //   />
+      // );
+    })
+    .startWith(<JobsTabLoading />);
```
juliangieseke: todo: add `.catch()`

plugins/jobs/src/js/__tests__/JobsList-test.ts
```diff
@@ -0,0 +1,3 @@
+describe("JobsList", () => {
+  it("tests");
```
juliangieseke: todo: not sure if I need to add some tests here 🤔 theoretically everything should be there…

src/js/routes/jobs.js
```diff
@@ -3,7 +3,7 @@ import React from "react";
 /* eslint-enable no-unused-vars */
 import { IndexRoute, Redirect, Route } from "react-router";
 
-import JobsTabContainer from "#PLUGINS/jobs/src/js/JobsTabContainer";
```
DanielMSchmidt: This is the line we would need to change back to not have it active, right?

plugins/jobs/src/js/JobsOverview.tsx
```diff
@@ -10,98 +8,127 @@ import { graphqlObservable, componentFromStream } from "data-service";
...
-        } else {
-          item = root;
+        const modal = (
+          <JobFormModalContainer
+            open={isModalOpen}
+            onClose={handleCloseModal}
+          />
+        );
+
+        if (jobs.totalCount > 0) {
```
DanielMSchmidt: I would like to have it the other way around and use the `if` for the exceptional case: that data is missing.

plugins/jobs/src/js/JobsOverviewRouteProvider.tsx
```diff
@@ -0,0 +1,46 @@
...
+      pathname,
+      query
+    });
+  }
+
+  render() {
+    return (
+      <JobsOverview
+        namespace={this.props.params.id ? this.props.params.id.split(".") : []}
+        initialFilter={this.props.location.query.searchString || ""}
```
DanielMSchmidt: This should be `this.context.location`, shouldn't it?

plugins/jobs/src/js/JobsOverviewRouteProvider.tsx
```diff
@@ -0,0 +1,46 @@
...
+import { routerShape, LocationDescriptor } from "react-router";
+import JobsOverview from "#PLUGINS/jobs/src/js/JobsOverview";
+
+interface JobsOverviewRouteProviderProps {
+  params: { id: string };
+  location: LocationDescriptor;
+  children: React.ReactElement<any>;
+}
+
+export default class JobsOverviewRouteProvider extends React.Component<
```
DanielMSchmidt: In general (this is not blocking) I would like to see this route provider more as a general thing. It could expose the router and the location to the child components. This way it would be reusable and not entangled with the JobsOverviewRouteProvider. But we can do this as a separate refactoring.

plugins/jobs/src/js/JobsOverview.tsx
juliangieseke: I agree. Will change if I touch the file again - just copy pastes old code here 🤷‍♂️

plugins/jobs/src/js/JobsOverviewRouteProvider.tsx
juliangieseke: agree, but we might want to double check if this is something we _want_ to have a a general thing.

plugins/jobs/src/js/JobsOverviewRouteProvider.tsx
juliangieseke: nope, location gets injected as prop by router (or whatever component it does) - accessing it via context results in a deprecation warning

src/js/routes/jobs.js
juliangieseke: yes - and restore all the old files (I'd go with `#SRC/js/…` in this case)

juliangieseke:
I'll close this and will create several smaller new PRs. For more information see https://jira.mesosphere.com/browse/DCOS-37610

PR #3019, opened 2018-06-15
DCOS-38126: List of Jobs resolver - Take 2

DanielMSchmidt:
I failed when trying to rebase, this is my attempt.

**TODOs**

The only thing left is that we would need to "merge" both test files, so take the `describe("jobs", () => {...});` from `plugins/jobs/src/js/data/JobModel-test-from-fabs-copy.js` and merge it into `plugins/jobs/src/js/data/JobModel-test.js`.

After that the tests will be red, because Fabs assumption was that we will return an array of jobs. We will return a JobRunConnection instead (see implementation), so we would need to map the output of the result$ in these tests like `result$.map(result => result.nodes)` to make them green.

Once they are green we would need to test `totalCount` to deliver us the number of all jobs despite being filtered and `filteredCount` to return the number after filtering.

**Checklist**

### DIFF
--- a/plugins/jobs/src/js/data/JobModel-test-from-fabs-copy.js
+++ b/plugins/jobs/src/js/data/JobModel-test-from-fabs-copy.js
@@ -1,0 +1,1663 @@
+import { Observable } from "rxjs";
+import { marbles } from "rxjs-marbles/jest";
+import { resolvers } from "../JobModel";
+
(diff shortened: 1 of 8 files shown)


juliangieseke:
please lets not work on this branch, I have restored fabs commits on his PR 👍 #3014

PR #3020, opened 2018-06-19
Jobs Menu Experiment - Use array instead of componentsToStream

bstavroulakis:
This is for you fellows you review if this approach is better than the previous one where we stream with the component.

With this approach we may just export and array and when you click on to do something we can just run the streams from the event.

This is linked to https://github.com/dcos/dcos-ui/pull/2999. If we prefer it this way I can merge it and continue from there.

**Checklist**


plugins/jobs/src/js/JobsMenu.js
```diff
@@ -1,36 +1,9 @@
...
-import "rxjs/add/operator/combineLatest";
-
-import Page from "#SRC/js/components/Page";
-import JobsBreadcrumbs from "./components/JobsBreadcrumbs";
-import { jobsRunNowAction, jobsRunNow$ } from "./JobsRunNow";
-
-const getActions = function(props) {
-  const [edit, ...rest] = props.actions;
+export default function(jobId) {
   const newActions = [];
```
bstavroulakis: This will gradually include all of the menu items

plugins/jobs/src/js/JobsRunNow.js
```diff
@@ -21,16 +19,11 @@ const runNowJobGraphql = id => {
...
-  return runNowJobGraphql(id);
-});
-
-export const jobsRunNowAction = id => {
+export default function(id) {
   return {
     label: "Run Now",
     onItemSelect() {
-      runNowEvent$.next({ id });
+      runNowJobGraphql({ id }).subscribe();
```
bstavroulakis: Don't really like that I have to add the subscribe to run it. But the previous way where is create a subject and then merge the 2 to run .next() looks like a bit hacky as well.

plugins/jobs/src/js/data/JobModel.ts
```diff
@@ -73,7 +73,7 @@ export const typeDefs = `
     ): Job
   }
   type Mutation {
-    runJob(id: String!): JobRun!
+    runJob(id: String!): Job!
```
bstavroulakis: This can change back

src/js/pages/jobs/JobDetailPage.js
```diff
@@ -242,12 +245,10 @@ class JobDetailPage extends mixin(TabsMixin) {
 
     return (
       <Page>
-        <JobsMenu
+        <Page.Header
```
bstavroulakis: This is the juice of the change. To keep the Page.Header and not create a new component with the new data layer. Please tell me if you like this way better.

src/js/pages/jobs/JobDetailPage.js
```diff
@@ -242,12 +245,10 @@ class JobDetailPage extends mixin(TabsMixin) {
...
           actions={this.getActions()}
```
bstavroulakis: this.getActions() will eventually be menuGetActions(job)

plugins/jobs/src/js/JobsRunNow.js
Fabs: Calling subscribe, means you have to do:
```js
const subscription = stream$.subscribe();

//somehwere later.
subscription.unsubscribe();
```
Otherwise, you may be causing a memory leak.

The merging of the subject is actually the correct way to go. That is actually a pattern you might hope to find on any use of the subject as a signal/event passed up. You can look at `createEventHandler` from recompose (_https://github.com/acdlite/recompose/blob/master/docs/API.md#createeventhandler). The subject merged/combined/filter with other operators, is the way to treat the event (stream), and the calling of next is the handler (the function to pass the signal up).

plugins/jobs/src/js/data/JobModel.ts
```diff
@@ -113,6 +113,7 @@ export const resolvers = ({
       _args: GeneralArgs,
       _context = {}
     ): Observable<Job | null> {
+      debugger;
```
Fabs: this should go ;)

src/js/pages/jobs/JobDetailPage.js
Fabs: I agree with that, when we started I have not seem the problem it would cause to Page.Header. I agree we should not touch it, I like the passing an array solution.

nLight:
i like the solution better. Thanks for taking the time putting it together 💪

PR #3023, opened 2018-06-20
DCOS-37612 - Run Jobs Menu Item

Fabs:
This is the replacement of #2999 and #3020.

This PR has 2 goals:

- The groundwork to detach the menu from the JobsDetails page.
- Add the Run Job with the new Data Layer pattern.

Checklist

**Checklist**
- [X] Did you add new integration tests?

### DIFF
--- a/plugins/jobs/src/js/__tests__/jobsRunNow-test.js
+++ b/plugins/jobs/src/js/__tests__/jobsRunNow-test.js
@@ -1,0 +1,23 @@
+import { Observable } from "rxjs";
+import "rxjs/add/observable/of";
+import { graphqlObservable } from "data-service";
+import jobsRunNow from "../jobsRunNow";
+
+jest.mock("data-service", () => ({
+  graphqlObservable: jest.fn()
+}));
+
+describe("JobsRunNow", function() {
+  describe("#jobsRunNow", function() {
+    it("onItemSelect triggers a graphql mutation", function() {
+      graphqlObservable.mockReturnValue(Observable.of("response"));
+      jobsRunNow("aJobId").onItemSelect();
+
+      expect(graphqlObservable).toBeCalledWith(
+        expect.anything(),
+        expect.anything(),
+        { jobId: "aJobId" }
+      );
+    });
+  });
+});
--- a/plugins/jobs/src/js/data/JobModel.ts
+++ b/plugins/jobs/src/js/data/JobModel.ts
@@ -3,6 +3,7 @@ import { Observable } from "rxjs/Observable";
 import {
   fetchJobs,
   fetchJobDetail,
+  runJob,
   JobResponse as MetronomeJobResponse,
   JobDetailResponse as MetronomeJobDetailResponse
 } from "#SRC/js/events/MetronomeClient";
@@ -20,6 +21,8 @@ import {
   JobSchema
 } from "#PLUGINS/jobs/src/js/types/Job";
 
+import { JobLink, JobLinkSchema } from "#PLUGINS/jobs/src/js/types/JobLink";
+
 export interface Query {
   jobs: JobConnection | null;
   job: Job | null;
@@ -29,6 +32,7 @@ export interface ResolverArgs {
   fetchJobs: () => Observable<MetronomeJobResponse[]>;
   fetchJobDetail: (id: string) => Observable<MetronomeJobDetailResponse>;
   pollingInterval: number;
+  runJob: (id: string) => Observable<JobLink>;
 }
 
 export interface GeneralArgs {
@@ -52,6 +56,7 @@ export type SortDirection = "ASC" | "DESC";
 
 export const typeDefs = `
   ${JobSchema}
+  ${JobLinkSchema}
   ${JobConnectionSchema}
 
   enum SortOption {
@@ -73,6 +78,10 @@ export const typeDefs = `
     job(
       id: ID!
     ): Job
+  }
+  
+  type Mutation {
+    runJob(id: String!): JobLink!
   }
   `;
 
@@ -153,7 +162,8 @@ function filterJobsByNamespace(
 export const resolvers = ({
   fetchJobs,
   fetchJobDetail,
-  pollingInterval
+  pollingInterval,
+  runJob
 }: ResolverArgs): IResolvers => ({
   Query: {
     jobs(
@@ -186,7 +196,11 @@ export const resolvers = ({
           nodes: jobs
         }));
     },
-    job(_obj = {}, args: GeneralArgs, _context = {}): Observable<Job | null> {
+    job(
+      _parent = {},
+      args: GeneralArgs,
+      _context = {}
+    ): Observable<Job | null> {
       if (!isJobQueryArg(args)) {
         return Observable.throw("The job resolver expects an id as argument");
       }
@@ -198,6 +212,15 @@ export const resolvers = ({
 
       return responses$.map(response => JobTypeResolver(response));
     }
+  },
+  Mutation: {
+    runJob(
+      _parent = {},
+      args: GeneralArgs,
+      _context = {}
+    ): Observable<JobLink> {
+      return runJob(args.id).map(({ jobId }) => ({ jobId }));
+    }
   }
 });
 
@@ -206,6 +229,7 @@ export default makeExecutableSchema({
   resolvers: resolvers({
     fetchJobs,
     fetchJobDetail,
-    pollingInterval: Config.getRefreshRate()
+    pollingInterval: Config.getRefreshRate(),
+    runJob
   })
 });
--- a/plugins/jobs/src/js/jobsRunNow.js
+++ b/plugins/jobs/src/js/jobsRunNow.js
@@ -1,0 +1,27 @@
+import "rxjs/add/operator/take";
+
+import { graphqlObservable } from "data-service";
+import gql from "graphql-tag";
+
+import defaultSchema from "./data/JobModel";
+
+const runNowJobMutation = gql`
+  mutation {
+    runJob(id: $jobId) {
+      jobId
+    }
+  }
+`;
+
+export default jobId => {
+  return {
+    label: "Run Now",
+    onItemSelect() {
+      // take(1) makes sure the observable is going to complete after finishing
+      // the request, so we don't have to care about unsubscribing.
+      graphqlObservable(runNowJobMutation, defaultSchema, { jobId })
+        .take(1)
+        .subscribe();
+    }
+  };
+};
--- a/plugins/jobs/src/js/types/JobLink.ts
+++ b/plugins/jobs/src/js/types/JobLink.ts
@@ -1,0 +1,9 @@
+export const JobLinkSchema = `
+type JobLink {
+  jobId: String!
+}
+`;
+
+export interface JobLink {
+  jobId: string;
+}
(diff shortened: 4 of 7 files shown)


plugins/jobs/src/js/JobsRunNow.js
```diff
@@ -0,0 +1,34 @@
...
+  return graphqlObservable(runNowJobMutation, defaultSchema, { jobId });
+};
+
+export default function(jobId) {
+  return {
+    label: "Run Now",
+    onItemSelect() {
+      // take(1) makes sure the observable is going to complete after finishing
+      // the request, so we don't have to care about unsubscribing.
+      runNowJobGraphql(jobId)
```
Fabs: @bstavroulakis it looks you were right here, I could not do it with just the observable, it would not trigger the mutation. Which in my rxjs book (and @orlandohohmeier ) is weird :D

plugins/jobs/src/js/JobsMenu.js
```diff
@@ -0,0 +1,9 @@
+import jobsRunNowAction from "./JobsRunNow";
+
+export default function(jobId) {
+  const newActions = [];
```
DanielMSchmidt: Why don't we return the array right now?

plugins/jobs/src/js/data/JobModel.ts
```diff
@@ -29,6 +30,7 @@ export interface ResolverArgs {
   fetchJobs: () => Observable<MetronomeJobResponse[]>;
   fetchJobDetail: (id: string) => Observable<MetronomeJobDetailResponse>;
   pollingInterval: number;
+  runJob: (id: string) => Observable<Job | null>;
```
DanielMSchmidt: why is this an observable of null?

DanielMSchmidt:
@Fabs your first commit seems to be "broken" 🤔

plugins/jobs/src/js/JobsRunNow.js
```diff
@@ -0,0 +1,34 @@
+/* eslint-disable no-unused-vars */
+import React from "react";
+/* eslint-enable no-unused-vars */
+
+import "rxjs/add/operator/switchMap";
```
DanielMSchmidt: We don't need a switchMap here as far as I can see

packages/data-service/index.js
```diff
@@ -0,0 +1,2 @@
+export { graphqlObservable } from "./src/graphqlObservable";
```
DanielMSchmidt: Can you pull this move of data-service into a seperate commit and explain why you need to do this?

Fabs: I think that fixes an importing problem bill was having. I will try to figure out, but I don`t really know.

plugins/jobs/src/js/JobsRunNow.js
Fabs: Yeap. It is reminiscent before I did the the take(1). Thanks.

Fabs:
I have those failed system tests:

image

plugins/jobs/src/js/data/JobModel.ts
```diff
@@ -186,7 +194,11 @@ export const resolvers = ({
           nodes: jobs
         }));
     },
-    job(_obj = {}, args: GeneralArgs, _context = {}): Observable<Job | null> {
+    job(
+      _parent = {},
+      args: GeneralArgs,
+      _context = {}
+    ): Observable<Job | null> {
```
juliangieseke: you added `!` in gql, so you can remove the `null` here :)

plugins/jobs/src/js/types/Job.ts
```diff
@@ -57,6 +57,10 @@ export interface Job {
   scheduleStatus: JobStatus;
 }
 
+export interface JobLink {
```
juliangieseke: This should live in its own `JobLink.ts` Typefile.

plugins/jobs/src/js/types/Job.ts
```diff
@@ -84,6 +88,10 @@ type Job {
   schedules: ScheduleConnection!
   scheduleStatus: JobStatus!
 }
+
+type JobLink {
```
juliangieseke: same here.

src/js/pages/jobs/JobDetailPage.js
```diff
@@ -10,6 +10,8 @@ import mixin from "reactjs-mixin";
 import JobFormModalContainer from "#PLUGINS/jobs/src/js/JobFormModalContainer";
 import TaskStates from "#PLUGINS/services/src/js/constants/TaskStates";
 import JobsBreadcrumbs from "#PLUGINS/jobs/src/js/components/JobsBreadcrumbs";
+import menuGetActions from "#PLUGINS/jobs/src/js/JobsMenu";
```
juliangieseke: if using default export, please dont rename it from Filename :(

src/js/pages/jobs/JobDetailPage.js
```diff
@@ -180,17 +182,14 @@ class JobDetailPage extends mixin(TabsMixin) {
...
     actions.push({
       label: "Edit",
       onItemSelect: this.props.handleEditButtonClick
     });
 
-    actions.push({
-      label: "Run Now",
-      onItemSelect: this.props.handleRunNowButtonClick
-    });
+    actions = actions.concat(menuGetActions(job.getId()));
```
juliangieseke: even if I like immutability a lot, only changing it for one line (but not above), isnt a good idea IMHO.

Also I dont really understand why you are extracting this into a new file, but most probably you'll add the otheres into that file?

plugins/jobs/src/js/data/JobModel.ts
```diff
@@ -16,6 +17,7 @@ import {
 } from "#PLUGINS/jobs/src/js/types/JobConnection";
 import {
   Job,
+  JobLink,
```
juliangieseke: This should live in its own `JobLink.ts` File 👍

src/js/pages/jobs/JobDetailPage.js
Fabs: You are right, let that one slip. They should all be the immutable or not.

plugins/jobs/src/js/data/JobModel.ts
Fabs: Oh, I was not supposed to touch that type. Nice catch. I only wanted to change the _obj to _parent.

src/js/pages/jobs/JobDetailPage.js
Fabs: Do you know why it is a good practice not to rename it?

src/js/pages/jobs/JobDetailPage.js
Fabs: By the way, it needs to be concat here, because menuGetActions returns a list. I will make everything concat.

src/js/pages/jobs/JobDetailPage.js
juliangieseke: It is what we do in the codebase, using default and calling it the same as the file 🤷‍♂️  Also there is the Rule that one should only declare one component per file.

Fabs: I am just curious about the why. I will rename anyway.

juliangieseke: OCD 😅

src/js/pages/jobs/JobDetailPage.js
juliangieseke: `menuGetActions(job.getId()).map(item => action.push(item))` :trollface:

Fabs:
Sorry, @juliangieseke. For that, there is an easy fix. I won't dismiss your comments anymore.

Fabs:
The snapshot tests for jest I wrote are not supported on CI. I will find a way to deal with that....

plugins/jobs/src/js/jobsRunNow.js
```diff
@@ -0,0 +1,35 @@
+/* eslint-disable no-unused-vars */
```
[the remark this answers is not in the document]
Fabs: Nope.

plugins/jobs/src/js/jobsRunNow.js
```diff
@@ -0,0 +1,35 @@
...
+  return graphqlObservable(runNowJobMutation, defaultSchema, { jobId });
+}
+
+export default jobId => {
+  return {
+    label: "Run Now",
+    onItemSelect() {
+      // take(1) makes sure the observable is going to complete after finishing
+      // the request, so we don't have to care about unsubscribing.
+      jobsRunNowGraphql(jobId)
```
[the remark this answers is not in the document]
Fabs: You mean the import of the take? I don't really get this one.

src/js/pages/jobs/JobDetailPage.js
```diff
@@ -180,37 +182,42 @@ class JobDetailPage extends mixin(TabsMixin) {
...
     const [schedule] = job.getSchedules();
 
-    const actions = [];
+    let actions = [];
 
-    actions.push({
-      label: "Edit",
-      onItemSelect: this.props.handleEditButtonClick
-    });
+    actions = actions.concat([
```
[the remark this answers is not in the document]
Fabs: None of this really matter now. In the end, jobMenu will return an array with all the items on the menu, and all this code is going to be gone. Some of us do not like mixing push with concat.

This code is going to be removed during https://jira.mesosphere.com/browse/DCOS-37616.

I will add that comment.

plugins/jobs/src/js/jobsRunNow.js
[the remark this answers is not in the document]
nLight: I mean at least `take`, but maybe `subscribe` as well. Does it make sense?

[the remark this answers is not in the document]
Fabs: I would find a little weird to have this import in this scope function instead of the top level. I know that there might be some build size benefits, but I would not go there.

Also for the subscribe, there is no import for it. It comes when you import Observable, and I believe you don't ever have to import it.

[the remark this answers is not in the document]
nLight: I don't mean import at all:
```js
function jobsRunNowGraphql(jobId) {
  return graphqlObservable(runNowJobMutation, defaultSchema, { jobId }).take(1).subscribe();
}
```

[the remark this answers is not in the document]
Fabs: Now I get you then.

plugins/jobs/src/js/types/JobLink.ts
```diff
@@ -0,0 +1,9 @@
+export const JobLinkSchema = `
+type JobLink {
```
DanielMSchmidt: We need job links as we only want to return the jobId, right?

plugins/jobs/src/js/__tests__/jobsRunNow-test.js
```diff
@@ -0,0 +1,26 @@
+import { Observable } from "rxjs";
+import "rxjs/add/observable/of";
+
+const mockGraphqlObservable = jest.fn();
+jest.mock("data-service", () => ({
+  graphqlObservable: mockGraphqlObservable
+}));
+
+const jobsRunNow = require("../jobsRunNow");
```
DanielMSchmidt: Can't we use an import here, too?

plugins/jobs/src/js/types/JobLink.ts
Fabs: I find this better than just say it returns a string. Maybe there is a type alias on graphql. But I would rather prefer the more verbose type now. 

If there is a way to alias a type, I can revisit that on the ticket I will focus on graphql.

plugins/jobs/src/js/__tests__/jobsRunNow-test.js
Fabs: yes

plugins/jobs/src/js/types/JobLink.ts
DanielMSchmidt: Yeah, me too, just wanted to be sure I understand what is going on there

PR #3026, opened 2018-06-21
DCOS-38399: Improve mesos backoff strategy

orlandohohmeier:
Adjust the backoff strategy to allow _unlimitted_ retries, reduce the start delay and limit the delay to 5 seconds.

The adjusted strategy should make the Mesos data stream more resilient to connection issues and basically falls back to polling for changes in cases where chunked streaming doesn't work.

This change also improves the perceived performance as delaying the first retry longer than 500ms results in users thinking that nothing is happening at all.

Closes DCOS-38399

### DIFF
--- a/src/js/stores/MesosStateStore.js
+++ b/src/js/stores/MesosStateStore.js
@@ -21,7 +21,8 @@ import container from "../container";
 import * as mesosStreamParsers from "./MesosStream/parsers";
 
 const MAX_RETRIES = 5;
-const RETRY_DELAY = 5000;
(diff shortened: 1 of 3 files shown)


nLight:
@juliangieseke if you plot two graphs then logarithmic back of is just a smoother version of linear backoff with cap. I'm curious where would the value add be in your opinion?

juliangieseke:
I think its more of an academic thing, I like it more to have one algorithm producing a graph then having an "if below cap, use algorithm, otherwise use cap"-thing 🤷‍♂️

PR #3027, opened 2018-06-25
DCOS-38479: Implement new breadcrumb based on job objects

DanielMSchmidt:
This is work extracted from #3016 to ship small chunks. It is needed to unblock the job overview PR, not as part of this work, but preparing this.

**Checklist**

### DIFF
--- a/plugins/jobs/src/js/components/Breadcrumbs.js
+++ b/plugins/jobs/src/js/components/Breadcrumbs.js
@@ -1,0 +1,59 @@
+import { Link } from "react-router";
+import PropTypes from "prop-types";
+import React from "react";
+
+import Breadcrumb from "#SRC/js/components/Breadcrumb";
+import BreadcrumbTextContent from "#SRC/js/components/BreadcrumbTextContent";
+import PageHeaderBreadcrumbs from "#SRC/js/components/PageHeaderBreadcrumbs";
+
+export default function Breadcrumbs({ item, children, states }) {
+  function getBreadcrumb(item, id = "", name = "") {
+    const isDetailPage = [...item.path, item.name].join(".") === id;
+    const link = isDetailPage ? `/jobs/detail/${id}` : `/jobs/overview/${id}`;
+
+    return (
+      <Breadcrumb key={id} title="Jobs">
+        <BreadcrumbTextContent>
+          <Link to={link}>{name === "" ? "Jobs" : name}</Link>
+        </BreadcrumbTextContent>
+
+        {isDetailPage ? states : null}
+      </Breadcrumb>
+    );
+  }
+
+  function getBreadcrumbList(item) {
+    if (item == null) {
+      return [];
+    }
+
+    const pathSegments = [...item.path, item.name];
+    const segments = pathSegments.map((_, index) =>
+      pathSegments.slice(0, index + 1)
+    );
+
+    return segments.map(segment =>
+      getBreadcrumb(item, segment.join("."), segment[segment.length - 1])
+    );
+  }
+
+  let breadcrumbs = [];
+
+  if (item) {
+    breadcrumbs = [].concat(
+      getBreadcrumb(item),
+      getBreadcrumbList(item),
+      React.Children.toArray(children)
+    );
+  }
+
+  return <PageHeaderBreadcrumbs iconID="jobs" breadcrumbs={breadcrumbs} />;
+}
+
+Breadcrumbs.propTypes = {
+  states: PropTypes.element,
+  item: PropTypes.shape({
+    name: PropTypes.string.isRequired,
+    path: PropTypes.arrayOf(PropTypes.string).isRequired
+  }).isRequired
+};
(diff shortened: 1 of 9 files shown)


plugins/jobs/src/js/components/Breadcrumbs.js
```diff
@@ -0,0 +1,120 @@
...
+    <BreadcrumbSupplementalContent>
+      <div className="service-page-header-status muted">
+        ({TaskStates[status].displayName})
+      </div>
+    </BreadcrumbSupplementalContent>
+  );
+}
+
+function ItemSchedule({ item: { schedules } }) {
+  if (!schedules || !schedules.nodes.length || !schedules.nodes[0].enabled) {
```
nLight: Could be rewritten using `findNestedPropertyInObject` as well, though in this place it's not critical at all.

plugins/jobs/src/js/components/Breadcrumbs.js
```diff
@@ -0,0 +1,120 @@
...
+function getBreadcrumb(item, id = "", name = "") {
+  const isDetailPage = [...item.path, item.id].join(".") === id;
+  const link = isDetailPage ? `/jobs/detail/${id}` : `/jobs/overview/${id}`;
+
+  return (
+    <Breadcrumb key={id} title="Jobs">
+      <BreadcrumbTextContent>
+        <Link to={link}>{name === "" ? "Jobs" : name}</Link>
+      </BreadcrumbTextContent>
+      {isDetailPage ? <ItemSchedule item={item} /> : null}
```
nLight: We could do those in one take, of course that would require adding `key` prop, which in this case could be as straightforward as `key="schedule"` and `key="status"`

plugins/jobs/src/js/components/Breadcrumbs.js
```diff
@@ -0,0 +1,120 @@
...
+    </Breadcrumb>
+  );
+}
+
+function getBreadcrumbList(item) {
+  if (item == null) {
+    return [];
+  }
+
+  const pathSegments = [...item.path, item.name];
```
nLight: Wonder why here we go with a `name` while at line 56 https://github.com/dcos/dcos-ui/pull/3027/files#diff-2dbad4fb56742926b95061ce75f631b7R56 we go with `id`?

plugins/jobs/src/js/components/Breadcrumbs.js
```diff
@@ -0,0 +1,120 @@
...
+import BreadcrumbSupplementalContent from "#SRC/js/components/BreadcrumbSupplementalContent";
+import BreadcrumbTextContent from "#SRC/js/components/BreadcrumbTextContent";
+import Icon from "#SRC/js/components/Icon";
+import PageHeaderBreadcrumbs from "#SRC/js/components/PageHeaderBreadcrumbs";
+
+function ItemStatus({ item: { jobRuns } }) {
+  if (!jobRuns) {
+    return null;
+  }
+  const { longestRunningActiveRun } = jobRuns;
```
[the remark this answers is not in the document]
DanielMSchmidt: Didn't knew about them, will do 👍

plugins/jobs/src/js/components/Breadcrumbs.js
DanielMSchmidt: I think following @juliangieseke suggestion I will let them be passed in as render props or children

plugins/jobs/src/js/components/Breadcrumbs.js
DanielMSchmidt: argh, that should be name as well in the other place 🙈

DanielMSchmidt:
@juliangieseke @nLight updated accordingly 👍

plugins/jobs/src/js/components/Breadcrumbs.js
```diff
@@ -0,0 +1,59 @@
...
+    const isDetailPage = [...item.path, item.name].join(".") === id;
+    const link = isDetailPage ? `/jobs/detail/${id}` : `/jobs/overview/${id}`;
+
+    return (
+      <Breadcrumb key={id} title="Jobs">
+        <BreadcrumbTextContent>
+          <Link to={link}>{name === "" ? "Jobs" : name}</Link>
+        </BreadcrumbTextContent>
+
+        {isDetailPage ? renderStates(item) : null}
```
nLight: I have mixed feelings about that one. So this is going to be a very special component for Jobs Breadcrumbs?

nLight: to clarify, having the function `renderStates` feels uncomfortable

DanielMSchmidt: Please propose an alternative. I need a way to not handle that data within this component to a) be able to also handle the job overview page and b) comply with @juliangieseke feedback

nLight: My suggestion would be a separate component for the Overview page.

DanielMSchmidt: Okay, so we would basically use the status and schedule component from within here and do a copy without them for the overview? I feel like we should discuss this at least with @juliangieseke because the result would be very very similar to what I had before coming to work yesterday, so I would like to double down to the decision to not waste any more time.

DanielMSchmidt:
@nLight @juliangieseke I updated this PR and added a JIRA

PR #3029, opened 2018-06-25
DCOS-38453: update jobmodel to fit overview mediator needs

juliangieseke:
This PR extracts the work from #3016 which is related to the JobModel. It prepares the introduction of the Overview Mediator, which will be one of the next things extracted from the old PR.

**How to Review/Test**
The JobModel isnt in use yet, in order to test this PR run `npm test -- JobModel`

### DIFF
--- a/plugins/jobs/src/js/data/JobModel.ts
+++ b/plugins/jobs/src/js/data/JobModel.ts
@@ -8,11 +8,11 @@ import {
 } from "#SRC/js/events/MetronomeClient";
 
 import Config from "#SRC/js/config/Config";
-import JobStates from "#PLUGINS/jobs/src/js/constants/JobStates";
-import JobStatus from "#PLUGINS/jobs/src/js/constants/JobStatus";
 import {
   JobConnection,
-  JobConnectionSchema
+  JobConnectionSchema,
+  JobConnectionTypeResolver,
+  JobsQueryArgs
 } from "#PLUGINS/jobs/src/js/types/JobConnection";
 import {
   Job,
@@ -35,20 +35,9 @@ export interface GeneralArgs {
   [key: string]: any;
 }
 
-export interface JobsQueryArgs {
-  filter?: string | null;
-  namespace?: string | null;
-  sortBy?: SortOption;
-  sortDirection?: SortDirection;
-}
-
 export interface JobQueryArgs {
   id: string;
 }
-
-export type SortOption = "ID" | "STATUS" | "LAST_RUN";
-
-export type SortDirection = "ASC" | "DESC";
 
 export const typeDefs = `
   ${JobSchema}
@@ -66,10 +55,10 @@ export const typeDefs = `
   type Query {
     jobs(
       filter: String
-      namespace: String
+      path: String
       sortBy: SortOption
       sortDirection: SortDirection
-    ): JobConnection
+    ): JobConnection!
     job(
       id: ID!
     ): Job
@@ -80,74 +69,8 @@ function isJobQueryArg(arg: any): arg is JobQueryArgs {
   return arg.id !== undefined;
 }
 
-// Sort and filtering
-function sortJobs(
-  jobs: Job[],
-  sortBy: SortOption = "ID",
-  sortDirection: SortDirection = "ASC"
-): Job[] {
-  jobs.sort((a, b) => {
-    let result;
-
-    switch (sortBy) {
-      case "ID":
-        result = compareJobById(a, b);
-        break;
-      case "STATUS":
-        result = compareJobByStatus(a, b);
-        break;
-      case "LAST_RUN":
-        result = compareJobByLastRun(a, b);
-        break;
-      default:
-        result = 0;
-    }
-
-    const direction = sortDirection === "ASC" ? 1 : -1;
-    return result * direction;
-  });
-
-  return jobs;
-}
-
-function compareJobById(a: Job, b: Job): number {
-  return a.id.localeCompare(b.id);
-}
-
-function compareJobByStatus(a: Job, b: Job): number {
-  return (
-    JobStates[a.scheduleStatus].sortOrder -
-    JobStates[b.scheduleStatus].sortOrder
-  );
-}
-
-function compareJobByLastRun(a: Job, b: Job): number {
-  const statusA = a.lastRunStatus ? a.lastRunStatus.status : "N/A";
-  const statusB = b.lastRunStatus ? b.lastRunStatus.status : "N/A";
-
-  return JobStatus[statusA].sortOrder - JobStatus[statusB].sortOrder;
-}
-
-function filterJobsById(
-  jobs: MetronomeJobResponse[],
-  filter: string | null
-): MetronomeJobResponse[] {
-  if (filter === null) {
-    return jobs;
-  }
-
-  return jobs.filter(({ id }) => id.indexOf(filter) !== -1);
-}
-
-function filterJobsByNamespace(
-  jobs: MetronomeJobResponse[],
-  namespace: string | null
-): MetronomeJobResponse[] {
-  if (namespace === null) {
-    return jobs;
-  }
-
-  return jobs.filter(({ id }) => id.startsWith(namespace));
+function isJobsQueryArg(arg: GeneralArgs): arg is JobsQueryArgs {
+  return (arg as JobsQueryArgs).path !== undefined;
 }
 
 export const resolvers = ({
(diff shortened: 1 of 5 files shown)


plugins/jobs/src/js/data/__tests__/JobModel-test.js
```diff
@@ -587,6 +589,42 @@ describe("JobModel Resolver", () => {
       })
     );
 
+    it(
+      "returns the path (job has path)",
```
DanielMSchmidt: 🎨I would like us to write more meaningful statements in general, sth like "a job with a path returns path segments"

plugins/jobs/src/js/types/JobConnection.ts
```diff
@@ -1,19 +1,107 @@
...
+      ? namespacedResponse.filter(({ id }) => id.includes(filter))
+      : namespacedResponse;
+
+  const resolvedJobs = filteredResponse.map(JobTypeResolver);
+
+  const sortedJobs: Job[] = resolvedJobs.sort((a, b) =>
+    compareJobs(a, b, sortBy, sortDirection)
+  );
+
+  // I decided not to go with fieldresolvers here, because the logic is a bit more complex and we need to find a better solution first
```
DanielMSchmidt: Can you explain why you choose to not use field resolvers? I understand that the sorting and filtering prevents you from using them, but we could do this work within a function and reuse it. Or we could make the field resolvers work on the filtered responses.

I would really like us to stick in the pattern of field resolvers, but if you think that is more than 30 minutes of work, let's sync on it

juliangieseke: I agree in general, but in this particular case I decided to keep it all in one because of the potential performance impact of calling the whole thing twice (for filteredCount & nodes). also we already know that we need to redo this… 🤷‍♂️

DanielMSchmidt: Then let's have a ticket for refactoring this to field resolvers

plugins/jobs/src/js/data/JobModel.ts
```diff
@@ -80,74 +69,8 @@ function isJobQueryArg(arg: any): arg is JobQueryArgs {
...
-function filterJobsByNamespace(
-  jobs: MetronomeJobResponse[],
-  namespace: string | null
-): MetronomeJobResponse[] {
-  if (namespace === null) {
-    return jobs;
-  }
-
-  return jobs.filter(({ id }) => id.startsWith(namespace));
+function isJobsQueryArg(arg: any): arg is JobsQueryArgs {
```
nLight: I'd say `arg` isn't `any`, shall we use `object` instead? Or do we really expect getting `any` here?

DanielMSchmidt: For these user-defined type guards you need to use any AFAIK

nLight: From the documentation 🤷‍♂️ 
```js
function isFish(pet: Fish | Bird): pet is Fish {
    return (<Fish>pet).swim !== undefined;
}
```

plugins/jobs/src/js/data/__tests__/JobModel-test.js
juliangieseke: Your example conflicts with our convention to write descriptions like this `it("does something…`. Do you have a better idea to name the test?

DanielMSchmidt: I would say "returns the path when the job has one", but it's not blocking for me

plugins/jobs/src/js/data/JobModel.ts
DanielMSchmidt: ah okay, good to know 👍

DanielMSchmidt:
Alfred test this please :)

juliangieseke:
@DanielMSchmidt @nLight addressed all the comments, please revisit 🙇

plugins/jobs/src/js/data/JobModel.ts
```diff
@@ -160,31 +83,14 @@ export const resolvers = ({
...
-
-      const pollingInterval$ = Observable.timer(0, pollingInterval);
-      const responses$ = pollingInterval$.exhaustMap(fetchJobs);
-
-      const jobsInNamespace$ = responses$.map(jobs =>
-        filterJobsByNamespace(jobs, namespace)
-      );
-      const count$ = jobsInNamespace$.map(jobs => jobs.length);
+    ): Observable<JobConnection> {
+      if (!isJobsQueryArg(args)) {
```
DanielMSchmidt: I think this is a bit strange as the check only "by accident" that the path should be part of the argument. I would either like the error message to read something like "you need to pass in a JobsQuerArg" or have a test like `if (!args.path)`.

I want to prevent us from having the error check and message drift from each other.

plugins/jobs/src/js/data/__tests__/JobModel-test.js
```diff
@@ -139,7 +139,7 @@ describe("JobModel Resolver", () => {
   describe("jobs", () => {
     const oneJobsResponse = (
       m,
-      filter = "",
+      filter = undefined,
```
DanielMSchmidt: isn't the default default value undefined?

plugins/jobs/src/js/data/JobModel.ts
juliangieseke: updated Errormessage below 👍

plugins/jobs/src/js/data/JobModel.ts
```diff
@@ -160,35 +83,22 @@ export const resolvers = ({
...
-          nodes: jobs
-        }));
+      return Observable.timer(0, pollingInterval)
+        .exhaustMap(fetchJobs)
+        .map(response => JobConnectionTypeResolver(response, args));
     },
     job(_obj = {}, args: GeneralArgs, _context = {}): Observable<Job | null> {
       if (!isJobQueryArg(args)) {
-        return Observable.throw("The job resolver expects an id as argument");
+        return Observable.throw(
```
DanielMSchmidt: Thank you for also fixing this error message 👍

plugins/jobs/src/js/types/JobConnection.ts
```diff
@@ -1,19 +1,109 @@
...
+  path: [String]!
   filteredCount: Int!
   totalCount: Int!
   nodes: [Job]!
 }
 `;
 
-export function JobConnectionTypeResolver(response: any): any {
-  return response;
+export interface JobsQueryArgs {
```
Fabs: should JobQueryArgs be its on type?

plugins/jobs/src/js/data/JobModel.ts
```diff
@@ -69,85 +58,19 @@ export const typeDefs = `
       namespace: String
```
juliangieseke: missed this line…

juliangieseke:
@DanielMSchmidt found onemore typo, canyou please revisit? 🙇

plugins/jobs/src/js/types/JobConnection.ts
Fabs: I see your confusion, it was because of my bad wording.

I mean, the type JobsQueryArgs is found on the JobConnection.ts. I would try to look for it on JobsQueryArgs.ts (or d.ts).

Is there a reason for the type not to have its own file?

plugins/jobs/src/js/data/__tests__/JobModel-test.js
```diff
@@ -246,7 +246,7 @@ describe("JobModel Resolver", () => {
           marbles(m => {
             m.bind();
 
-            const result$ = oneJobsResponse(m, null, "STATUS", "ASC").map(
+            const result$ = oneJobsResponse(m, undefined, "STATUS", "ASC").map(
```
Fabs: I am just curious on why would we prefer undefined here, instead of null.

juliangieseke:
CI is green, @Fabs can you please revisit? 👍 (I think you're already on it? 🤔)

plugins/jobs/src/js/types/JobConnection.ts
```diff
@@ -1,19 +1,109 @@
...
+export type SortDirection = "ASC" | "DESC";
+
+export function JobConnectionTypeResolver(
+  response: MetronomeJobResponse[],
+  args: JobsQueryArgs
+): JobConnection {
+  const { sortBy = "ID", sortDirection = "ASC", filter = "", path = [] } = args;
+
+  const namespacedResponse: MetronomeJobResponse[] =
+    path.length > 0
```
Fabs: No joke, but I found the ternary rule I would propose adding:
https://eslint.org/docs/rules/multiline-ternary
`never`

Fabs: Discussion for that is on sync. https://jira.mesosphere.com/browse/DCOS-38625

plugins/jobs/src/js/data/__tests__/JobModel-test.js
juliangieseke: I explained my point of view here: https://github.com/dcos/dcos-ui/pull/3030#discussion_r198786706
`undefined` = what the word says, "nothing" or "not passed" (as a parameter)
`null` = not _yet_ set, but will be set soon and is not supposed to return to value `null` after that.

in addition to that, dealing with `undefined` is way easier when it comes to defaults, because  `null` will not cause the default to be used (see changes in `JobConnection` TypeResolver)

plugins/jobs/src/js/types/JobConnection.ts
juliangieseke: Ah, fair point actually. And now as you say it, I remember that we wanted to change the foldername in the initial PR (but forgot to do so) 😕

Some History: When we created these files, Daniel wanted the folder `types` not named like that because of this exact reason. These files should define whatever type we use in GraphQL and contain all the necessary information related to it. I think we need to revisit and clarify this 🤔 @DanielMSchmidt, what is your opinion on this today?

That said, I dont think it is in scope of this PR.

plugins/jobs/src/js/types/JobConnection.ts
juliangieseke: Ähem. https://github.com/dcos/dcos-ui/pull/3004#discussion_r194372319

plugins/jobs/src/js/types/JobConnection.ts
DanielMSchmidt: Let's do it in a cleanup JIRA

PR #3028, opened 2018-06-25
DCOS-37618 Edit and Schedule Toggle

Fabs:
**SCOPE**

Refactor the menu in the shape of DCOS-37612.

**HOW TO TEST**

Checkout, run against the cluster and test that all menu items are the same.
Check this new format for reviews, see if you need more info.

TEMPLATE MVP: https://gist.github.com/Fabs/1e107028a5578a706061d41e29523060

**WHAT WAS TAKEN OUT OF SCOPE**

I pointed out the tickets as TODO

- Integrate with the graphql model
https://jira.mesosphere.com/browse/DCOS-37617

- The Delete menu with the modal
https://jira.mesosphere.com/browse/DCOS-38492

PR #3030, opened 2018-06-25
DCOS-38202: jobs modal container mediator

nLight:
This PR introduces `JobCreateEditFormModal` "mediator" instead of `JobFormModalContainer`. Some parts of the container went into the modal component, other parts went into the mediator. Everything should be working. This is a dorp-in replacement.


plugins/jobs/src/js/JobCreateEditFormModal.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+type JobFormModalInput = [
+  JobCreateEditFormModalProps,
+  IsJsonMode,
+  ErrorMessage
+];
+
+const JobCreateEditFormModal = componentFromStream<JobCreateEditFormModalProps>(
+  props$ => {
+    const isJsonMode$ = new BehaviorSubject<IsJsonMode>(false);
```
juliangieseke: I just went with `boolean` here 🤷‍♂️

plugins/jobs/src/js/JobCreateEditFormModal.tsx
```diff
@@ -0,0 +1,103 @@
...
+    const errorMessage$ = new BehaviorSubject<ErrorMessage>(null);
+
+    function handleInputModeToggle() {
+      isJsonMode$.next(!isJsonMode$.value);
```
juliangieseke: Hmmm, is this WIP, or does this actually work?

plugins/jobs/src/js/JobCreateEditFormModal.tsx
```diff
@@ -0,0 +1,103 @@
...
+      } else {
+        MetronomeStore.updateJob(job && job.getId(), jobSpec);
+      }
+    }
+    function handleErrorMessageChange(errorMessage: ErrorMessage) {
+      errorMessage$.next(errorMessage);
+    }
+    function metronomeSuccessHandler() {
+      console.log("close"); // tslint:disable-line
+      // props.onClose(); // Props aren't there yet to use ¯\_(ツ)_/¯
```
juliangieseke: check dis out: https://github.com/dcos/dcos-ui/pull/3016/files#diff-8d85ca1c176390d33bb055916f3dc3e8R54 (credits to @DanielMSchmidt )

plugins/jobs/src/js/JobCreateEditFormModal.tsx
```diff
@@ -0,0 +1,103 @@
...
+    }
+    MetronomeStore.addListener(
```
nLight: Any idea where to `removeListener`s?

plugins/jobs/src/js/JobCreateEditFormModal.tsx
```diff
@@ -0,0 +1,103 @@
...
+      METRONOME_JOB_CREATE_SUCCESS,
+      metronomeSuccessHandler
+    );
+    MetronomeStore.addListener(
+      METRONOME_JOB_UPDATE_SUCCESS,
+      metronomeSuccessHandler
+    );
+
+    const state = [isJsonMode$, errorMessage$];
```
juliangieseke: I dont see the reason for this array - either go with a stream here or drop this completely please :D ~(you can check out my PR for some reference how I did it - please dont see it as "best practice", but I'd love to have feedback - even if its closed.)~ looks like i decided to remove my `state$` entirely 😅

plugins/jobs/src/js/JobCreateEditFormModal.tsx
juliangieseke: maybe props$.complete? 🤔

plugins/jobs/src/js/JobCreateEditFormModal.tsx
```diff
@@ -0,0 +1,103 @@
...
+    return combineLatest<JobFormModalInput>([props$, ...state])
```
juliangieseke: interesting, I had to go with `Subscribable<React.ReactNode>` 🤔 but the way i declared types is a bit different

plugins/jobs/src/js/JobCreateEditFormModal.tsx
```diff
@@ -0,0 +1,103 @@
...
+      .map(data => {
+        const [props, isJsonMode, errorMessage] = data;
```
juliangieseke: i like `.map(([filter, isModalOpen, jobs]) => {` more - but thats personal taste. Are there any benefits I might have missed in your approach?

nLight: that could be a leftover. As you can see above I tried to type the interface. At some point it couldn't understand what I wanted from it but that version worked.

plugins/jobs/src/js/JobCreateEditFormModal.tsx
nLight: Oh no...

plugins/jobs/src/js/JobCreateEditFormModal.tsx
juliangieseke: maybe ts wasnt able to resolve the spread 🤷‍♂️

plugins/jobs/src/js/JobCreateEditFormModal.tsx
nLight: It _does_ work :D though it doesn't close the modal on success. Plus I still don't like it.

juliangieseke: sure, for this you have to call `props.onClose` - but maybe `props$.value.onClose` works? 🤔

nLight:
Legit cypress errors

plugins/jobs/src/js/components/JobFormModal.tsx
```diff
@@ -102,39 +145,229 @@ export default class JobFormModal extends React.Component<
...
+      this.resetState();
+    }
+  }
+
+  handleFormChange(event: { model: object }) {
+    if (!event.model) {
+      return;
+    }
+
+    this.props.handleErrorMessageChange(null);
```
DanielMSchmidt: why is null passed here?

Fabs:
@nLight is it possible we can have more information on how to test this PR, and what should we not expect to be there? Something like on this pr: https://github.com/dcos/dcos-ui/pull/3028, but not necessary identical.

plugins/jobs/src/js/components/JobFormModal.tsx
```diff
@@ -102,39 +145,229 @@ export default class JobFormModal extends React.Component<
...
+    } catch (e) {
+      this.props.handleErrorMessageChange({
+        message: "Invalid JSON syntax",
+        details: [{ path: "/", errors: [e.toString()] }]
+      });
+
+      return null;
+    }
+
+    const job = new Job(jobDefinition);
```
DanielMSchmidt: Do we plan to move away from the job struct?

nLight:
@Fabs as I stated above that is a drop in replacement. You just test the PR no setup needed. Everything should be there

plugins/jobs/src/js/components/JobFormModal.tsx
```diff
@@ -175,10 +408,10 @@ export default class JobFormModal extends React.Component<
           />
         ) : (
           <JobForm
-            defaultTab={defaultTab}
-            onTabChange={this.handleTabChange}
-            onChange={handleFormChange}
-            getTriggerSubmit={handleTriggerSubmit}
+            defaultTab={""}
+            // onTabChange={handleTabChange}
```
DanielMSchmidt: Why is this commented out?

plugins/jobs/src/js/components/JobFormModal.tsx
```diff
@@ -102,39 +145,229 @@ export default class JobFormModal extends React.Component<
...
+  handleSubmit() {
+    const { isEdit } = this.props;
+    const job = this.createJobFromEditorContents();
+    if (!job) {
+      return;
+    }
+
+    const jobSpec = JobUtil.createJobSpecFromJob(job);
+
+    this.props.handleSubmit(isEdit, jobSpec as JobData, job.getId());
```
DanielMSchmidt: Can we use a type check instead of a type cast here?

plugins/jobs/src/js/components/JobFormModal.tsx
nLight: Not in this PR I'd say

plugins/jobs/src/js/components/JobFormModal.tsx
nLight: To reset error message

DanielMSchmidt: ah, I see 👍

plugins/jobs/src/js/components/JobFormModal.tsx
DanielMSchmidt: Seems reasonable 👍

plugins/jobs/src/js/JobCreateEditFormModal.tsx
```diff
@@ -0,0 +1,57 @@
...
+  onClose: () => void;
+}
+
+type JobFormModalInput = [JobCreateEditFormModalProps, ErrorMessage];
+
+const JobCreateEditFormModal = componentFromStream<JobCreateEditFormModalProps>(
+  props$ => {
+    const errorMessage$ = new BehaviorSubject<ErrorMessage>(null);
+
+    function handleErrorMessageChange(errorMessage: ErrorMessage) {
```
nLight: What I would like to think about is how we could use recompose `createEvenHandler` so that we operate solely with streams to get rid of these custom handlers

plugins/jobs/src/js/JobCreateEditFormModal.tsx
```diff
@@ -0,0 +1,57 @@
...
+      errorMessage$.next(errorMessage);
+    }
+
+    function handleSubmit(isEdit: boolean, jobSpec: JobData, jobId: string) {
+      const request$ = isEdit ? updateJob(jobId, jobSpec) : createJob(jobSpec);
+
+      combineLatest(props$, request$)
+        .take(1) // unsubscribe after the first one to break ∞ react update loop
```
nLight: There's also `first()` we could use. We need this because `props.onClose()` changes state of the parent component which pushes new props down the `props$` stream ...which triggers `props.onClose()` :)

plugins/jobs/src/js/components/JobFormModal.tsx
```diff
@@ -102,39 +145,251 @@ export default class JobFormModal extends React.Component<
...
 
     this.state = {
-      defaultTab: ""
+      jobFormModel: null,
+      jobJsonString: null,
+      isJsonMode: false
     };
 
-    this.handleTabChange = this.handleTabChange.bind(this);
+    this.triggerFormSubmit = undefined;
```
juliangieseke: 🎨 I'd prefer `null` here. 

If I get it correct, this method is _not yet_ set, but there is no legit case where it can be set to `undefined` _once it is set_. In these cases, I'd use `null`. 
`undefined` IMHO is better for params that can be skipped or something that also can be (re)set to  "nothing" during code execution. 🤔 

That said, this usage of `null` comes from languages which only have `null`, but no additional `undefined` 😅Also it is more a gut feeling without me knowing if there is any convention/best practice in Javascript

plugins/jobs/src/js/components/JobFormModal.tsx
```diff
@@ -102,39 +145,251 @@ export default class JobFormModal extends React.Component<
...
+    this.handleSubmit = this.handleSubmit.bind(this);
+    this.handleJSONEditorChange = this.handleJSONEditorChange.bind(this);
+    this.handleFormChange = this.handleFormChange.bind(this);
+    this.handleTriggerSubmit = this.handleTriggerSubmit.bind(this);
+    this.handleInputModeToggle = this.handleInputModeToggle.bind(this);
+  }
+
+  componentWillReceiveProps(nextProps: JobFormModalProps) {
+    if (nextProps.isOpen !== this.props.isOpen) {
+      this.resetState();
```
juliangieseke: 🎨 not sure if this is legacy code, but can we rename this method (`resetState`) to `updateJobFromProps` or similar?

juliangieseke: or simply move the code here, looks like this the only place where it is called…

plugins/jobs/src/js/components/JobFormModal.tsx
```diff
@@ -102,39 +145,251 @@ export default class JobFormModal extends React.Component<
...
+    if (!event.model) {
+      return;
+    }
+
+    this.props.handleErrorMessageChange(null);
+  }
+
+  handleSubmit() {
+    const { isEdit } = this.props;
+    const job = this.createJobFromEditorContents();
```
juliangieseke: 🎨 here we create a job from form/json…

plugins/jobs/src/js/components/JobFormModal.tsx
```diff
@@ -102,39 +145,251 @@ export default class JobFormModal extends React.Component<
...
+    if (!job) {
+      return;
+    }
+
+    const jobSpec = JobUtil.createJobSpecFromJob(job);
```
juliangieseke: 🎨 …only to re-generate json?

plugins/jobs/src/js/components/JobFormModal.tsx
```diff
@@ -102,39 +145,251 @@ export default class JobFormModal extends React.Component<
...
+    }
+
+    this.props.handleErrorMessageChange(null);
+  }
+
+  handleSubmit() {
+    const { isEdit } = this.props;
+    const job = this.createJobFromEditorContents();
+    if (!job) {
+      return;
```
juliangieseke: 🎨 …to check if it is "truthy"…

plugins/jobs/src/js/components/JobFormModal.tsx
```diff
@@ -102,39 +145,251 @@ export default class JobFormModal extends React.Component<
...
+    }
+
+    const jobSpec = JobUtil.createJobSpecFromJob(job);
+
+    this.props.handleSubmit(isEdit, jobSpec, job.getId());
+  }
+
+  handleInputModeToggle() {
+    const job = this.createJobFromEditorContents(true) || this.props.job;
```
juliangieseke: 🎨 this seems to be the same as above…

plugins/jobs/src/js/components/JobFormModal.tsx
```diff
@@ -102,39 +145,251 @@ export default class JobFormModal extends React.Component<
...
+    this.setState({ jobJsonString }, () =>
+      this.props.handleErrorMessageChange(null)
+    );
+  }
+
+  handleTriggerSubmit(submitFunction: () => DataTriple) {
+    this.triggerFormSubmit = submitFunction;
+  }
+
+  resetState() {
```
juliangieseke: 🎨 as stated above, I think we can get rid of this…

plugins/jobs/src/js/components/JobFormModal.tsx
```diff
@@ -102,39 +145,251 @@ export default class JobFormModal extends React.Component<
...
+      jobFormModel: JobUtil.createFormModelFromSchema(JobSchema, job),
+      jobJsonString: JSON.stringify(
+        cleanJobJSON(JobUtil.createJobSpecFromJob(job as Job)),
+        null,
+        2
+      )
+    });
+  }
+
+  createJobFromEditorContents(keepValidationErrors = false) {
```
juliangieseke: 🎨 _Uff_…

plugins/jobs/src/js/JobCreateEditFormModal.tsx
nLight: I did and discussed with @Fabs. Decision was not to proceed with it at this stage.

plugins/jobs/src/js/JobCreateEditFormModal.tsx
nLight: in the following PR

plugins/jobs/src/js/JobCreateEditFormModal.tsx
```diff
@@ -0,0 +1,57 @@
+import * as React from "react";
+
+import { BehaviorSubject } from "rxjs/BehaviorSubject";
+import { combineLatest } from "rxjs/observable/combineLatest";
+
+import { componentFromStream } from "data-service";
+
+import { createJob, JobData, updateJob } from "#SRC/js/events/MetronomeClient";
+import Job from "#SRC/js/structs/Job";
```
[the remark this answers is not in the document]
nLight: Scoped out to a sub task that we may even consider to drop

plugins/jobs/src/js/components/JobFormModal.tsx
nLight: I appreciate your effort going through this piece of legacy, though I'd say we're keeping it for now.

nLight:
@DanielMSchmidt interesting, I can't replicate it. Would you mind sharing the exact steps, video?

nLight:
Though I realized why we need the `defaultTab`! Going to fix it

DanielMSchmidt:
@nLight Could not reproduce it again, strange 🤔

juliangieseke:
there was an issue with labels in the PR before, but that might be unrelated and fixed somewhere else

nLight:
@Fabs @juliangieseke @DanielMSchmidt folks, I fixed the default tab issue (switching to json would reset tab when you switch back). Please merge :)

PR #3031, opened 2018-06-25
Jobs Breadcrumbs for Overview

juliangieseke:
I dont think this is the best approach possible, but it works for details & overview and is relatively simple 👍

@DanielMSchmidt If overview isnt in scope for your PR, I'll have to adapt the Breadcrumb anyway, so I though I just give it a shot… 🤷‍♂️  Feel free to use this or close if you dont want to mess with that right now 👍


plugins/jobs/src/js/components/Breadcrumbs.js
```diff
@@ -6,54 +6,60 @@ import Breadcrumb from "#SRC/js/components/Breadcrumb";
...
-
+export default function Breadcrumbs({ path, name, children, extra }) {
+  /**
+   * @param  {string} link
+   * @param  {string[]} pathname path including name (if given)
+   * @param  {JSX.Element} [extra=null] extra JSX Components to render
+   *
+   * @return {JSX.Element} Breadcrumb Part Component
+   */
+  function buildJSX(link, pathname, extra = null) {
```
DanielMSchmidt: I think the function name has little to no meaning as does the extra.

the variable name pathname reads like a string, can we make it paths to be clear that it means an array?

plugins/jobs/src/js/components/Breadcrumbs.js
```diff
@@ -6,54 +6,60 @@ import Breadcrumb from "#SRC/js/components/Breadcrumb";
...
       </Breadcrumb>
     );
   }
 
-  function getBreadcrumbList(item) {
-    if (item == null) {
-      return [];
-    }
+  // we're starting from the back, so first we have children
+  let breadcrumbs = React.Children.toArray(children);
```
DanielMSchmidt: If you have t explain this, why don't we work the other way around?

plugins/jobs/src/js/components/Breadcrumbs.js
juliangieseke: re fname: agree, do you have a better idea?

re pathname: I went with pathname because it contains path + name in some situations. 🤔 also its typed array, so it should be obvious 🤔 do you have a better idea?

plugins/jobs/src/js/components/Breadcrumbs.js
juliangieseke: I dont _have to_ :trollface: - started from the end because otherwise I would have to declare an empty array manually only to concat it with another empty array later…

juliangieseke:
Closing this for now, will create a JIRA and update this PR later to merge it into master 👍

PR #3024, opened 2018-06-20
DCOS-37615 & DCOS-37619: Job Detail Page

DanielMSchmidt:
**How to review**
Navigate directly to a Jobs Detail page and see if everything still behaves the same

**Current Flaws**
The jobs page is broken this is why we have a **hold merge** label here. I would still like to receive feedback and I would say it is still ready for approval, but just not ready to be merged.

**Design Decisions**
We decided to not go with the the Jobs Tree anymore, which meant we removed it from the breadcrumbs, basically resulting in a tiny rewrite. 

**Checklist**
- [X] Did you add a JIRA issue in a commit message or as part of the branch name?
- [ ] Did you add new unit tests?
- [ ] Did you add new integration tests?
- [ ] If this is a regression, did you write a test to catch this in the future?

PR #3032, opened 2018-06-26
DCOS-37615:  add JobDetail page with mediator pattern

DanielMSchmidt:
## How to review

```diff
-import { JobDetailPageContainer } from "../pages/jobs/JobDetailPageContainer";
+import JobDetailPageContainer from "#PLUGINS/jobs/src/js/JobDetailPageContainer";
```

After that check out the Job Detail Page, it should work except for the missing pieces described below.

## Missing pieces

- No menu items except edit
- No breadcrumb at all

This is the follow up on https://github.com/dcos/dcos-ui/pull/3024

**Checklist**

### DIFF
--- a/plugins/jobs/src/js/pages/JobConfiguration.js
+++ b/plugins/jobs/src/js/pages/JobConfiguration.js
@@ -9,129 +9,105 @@ import ConfigurationMapSection from "#SRC/js/components/ConfigurationMapSection"
 import ConfigurationMapValue from "#SRC/js/components/ConfigurationMapValue";
 import HashMapDisplay from "#SRC/js/components/HashMapDisplay";
 
-export default class JobConfiguration extends React.Component {
-  static getSchedule(job) {
-    const lastSchedule = job.getSchedules()[0];
+const GeneralSection = ({ id, description, cpus, mem, disk, command }) => (
+  <ConfigurationMapSection>
+    <ConfigurationMapHeading>General</ConfigurationMapHeading>
+    <ConfigurationMapRow>
+      <ConfigurationMapLabel>ID</ConfigurationMapLabel>
+      <ConfigurationMapValue>{id}</ConfigurationMapValue>
+    </ConfigurationMapRow>
+    <ConfigurationMapRow>
+      <ConfigurationMapLabel>Description</ConfigurationMapLabel>
+      <ConfigurationMapValue>{description}</ConfigurationMapValue>
+    </ConfigurationMapRow>
+    <ConfigurationMapRow>
+      <ConfigurationMapLabel>CPUs</ConfigurationMapLabel>
+      <ConfigurationMapValue>{cpus}</ConfigurationMapValue>
+    </ConfigurationMapRow>
+    <ConfigurationMapRow>
+      <ConfigurationMapLabel>Memory (MiB)</ConfigurationMapLabel>
+      <ConfigurationMapValue>{mem}</ConfigurationMapValue>
+    </ConfigurationMapRow>
+    <ConfigurationMapRow>
+      <ConfigurationMapLabel>Disk Space (Mib)</ConfigurationMapLabel>
+      <ConfigurationMapValue>{disk}</ConfigurationMapValue>
+    </ConfigurationMapRow>
+    <ConfigurationMapRow>
+      <ConfigurationMapLabel>Command</ConfigurationMapLabel>
+      <ConfigurationMapValue>
+        <pre className="flush transparent wrap">{command}</pre>
+      </ConfigurationMapValue>
+    </ConfigurationMapRow>
+  </ConfigurationMapSection>
+);
 
-    if (!!lastSchedule && !!lastSchedule.cron) {
-      return lastSchedule.cron;
-    }
+const ScheduleSection = ({
+  id,
+  enabled,
+  cron,
+  timezone,
+  startingDeadlineSeconds
+}) => (
+  <ConfigurationMapSection>
+    <ConfigurationMapHeading>Schedule</ConfigurationMapHeading>
+    <ConfigurationMapRow>
+      <ConfigurationMapLabel>ID</ConfigurationMapLabel>
+      <ConfigurationMapValue>{id}</ConfigurationMapValue>
+    </ConfigurationMapRow>
+    <ConfigurationMapRow>
+      <ConfigurationMapLabel>Enabled</ConfigurationMapLabel>
+      <ConfigurationMapValue>{enabled}</ConfigurationMapValue>
+    </ConfigurationMapRow>
+    <ConfigurationMapRow>
+      <ConfigurationMapLabel>CRON Schedule</ConfigurationMapLabel>
+      <ConfigurationMapValue>{cron}</ConfigurationMapValue>
+    </ConfigurationMapRow>
+    <ConfigurationMapRow>
+      <ConfigurationMapLabel>Time Zone</ConfigurationMapLabel>
+      <ConfigurationMapValue>{timezone}</ConfigurationMapValue>
+    </ConfigurationMapRow>
+    <ConfigurationMapRow>
+      <ConfigurationMapLabel>Starting Deadline</ConfigurationMapLabel>
+      <ConfigurationMapValue>{startingDeadlineSeconds}</ConfigurationMapValue>
+    </ConfigurationMapRow>
+  </ConfigurationMapSection>
+);
 
-    return "No schedule available.";
-  }
+const DockerContainerSection = ({ image }) => (
+  <ConfigurationMapSection>
+    <ConfigurationMapHeading>Docker Container</ConfigurationMapHeading>
+    <ConfigurationMapRow>
+      <ConfigurationMapLabel>Image</ConfigurationMapLabel>
+      <ConfigurationMapValue>{image}</ConfigurationMapValue>
+    </ConfigurationMapRow>
+  </ConfigurationMapSection>
+);
 
-  getGeneralSection(job) {
-    return (
-      <ConfigurationMapSection>
-        <ConfigurationMapHeading>General</ConfigurationMapHeading>
-        <ConfigurationMapRow>
-          <ConfigurationMapLabel>ID</ConfigurationMapLabel>
-          <ConfigurationMapValue>{job.id}</ConfigurationMapValue>
-        </ConfigurationMapRow>
-        <ConfigurationMapRow>
-          <ConfigurationMapLabel>Description</ConfigurationMapLabel>
-          <ConfigurationMapValue>{job.description}</ConfigurationMapValue>
-        </ConfigurationMapRow>
-        <ConfigurationMapRow>
-          <ConfigurationMapLabel>CPUs</ConfigurationMapLabel>
-          <ConfigurationMapValue>{job.cpus}</ConfigurationMapValue>
-        </ConfigurationMapRow>
-        <ConfigurationMapRow>
-          <ConfigurationMapLabel>Memory (MiB)</ConfigurationMapLabel>
-          <ConfigurationMapValue>{job.mem}</ConfigurationMapValue>
-        </ConfigurationMapRow>
-        <ConfigurationMapRow>
-          <ConfigurationMapLabel>Disk Space (Mib)</ConfigurationMapLabel>
-          <ConfigurationMapValue>{job.disk}</ConfigurationMapValue>
-        </ConfigurationMapRow>
-        <ConfigurationMapRow>
-          <ConfigurationMapLabel>Command</ConfigurationMapLabel>
-          <ConfigurationMapValue>
-            <pre className="flush transparent wrap">{job.command}</pre>
-          </ConfigurationMapValue>
-        </ConfigurationMapRow>
-      </ConfigurationMapSection>
-    );
-  }
+const LabelsSection = ({ labels }) => (
+  <HashMapDisplay
+    hash={labels.reduce(
+      (memo, { key, value }) => ({ ...memo, [key]: value }),
+      {}
+    )}
+    headline="Labels"
+  />
+);
 
-  getScheduleSection(job) {
-    const [schedule] = job.schedules.nodes;
-    if (schedule == null) {
-      return null;
-    }
+export default function JobConfiguration({ job }) {
+  const [schedule] = job.schedules.nodes;
 
-    return (
-      <ConfigurationMapSection>
-        <ConfigurationMapHeading>Schedule</ConfigurationMapHeading>
-        <ConfigurationMapRow>
-          <ConfigurationMapLabel>ID</ConfigurationMapLabel>
-          <ConfigurationMapValue>{schedule.id}</ConfigurationMapValue>
-        </ConfigurationMapRow>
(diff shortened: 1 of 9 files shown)


plugins/jobs/src/js/pages/JobDetailPage.js
```diff
@@ -0,0 +1,195 @@
...
+  }
+
+  renderRunHistoryTabView(job) {
+    return <JobRunHistoryTable job={job} />;
+  }
+
+  getActions() {
+    const actions = [];
+
+    actions.push({
```
nLight: Are we going to add more actions depending on any logic? Curious why don't we just return an array directly here?

plugins/jobs/src/js/pages/JobDetailPageContainer.js
```diff
@@ -0,0 +1,112 @@
...
+    .map(({ data: { job } }) => job);
+
+  return Observable.combineLatest([
+    job$,
+    props$,
+    jobActionDialog$,
+    disabledDialog$
+  ])
+    .map(([job, props, jobActionDialog, disabledDialog]) => {
+      const enhancedProps = {
```
nLight: I'd go with throwing everything into JSX, not a big deal tho.
```js
<JobDetailPage {...props}
  ...
  closeDialog={closeDialog}
  disableDialog={disableDialog}
 />
```

plugins/jobs/src/js/pages/JobDetailPage.js
DanielMSchmidt: Yes, indeed. But that shouldn't hinder us for now, I will refactor this 👍

Fabs: Please do not touch or care about this code. I already took care of it while doing the menus.

plugins/jobs/src/js/pages/JobDetailPageContainer.js
```diff
@@ -0,0 +1,110 @@
...
+    disabledDialog$
+  ])
+    .map(([job, props, jobActionDialog, disabledDialog]) => (
+      <JobDetailPage
+        {...props}
+        job={job}
+        jobActionDialog={jobActionDialog}
+        disabledDialog={disabledDialog}
+        errorMsg={null}
+        handleRunNowButtonClick={() => {}}
```
Fabs: I really don't think it is a good idea we dropped the menu here, and all the callbacks. I can't test the dialog anymore (which is a responsibility of this component). 

I really think this should work with the store in a further ticket, and not just wait for the menu to get done and be integrated later. It also break a bunch of contracts that I am still relying on (which is to some of those handlers that needed to be passed on).

https://github.com/dcos/dcos-ui/pull/3023

I would be ok approving that one, but I would love to see a ticket that integrates and makes it work with 3023, before all the rest of the menu is done.

plugins/jobs/src/js/pages/JobDetailPage.js
```diff
@@ -0,0 +1,193 @@
...
+
+  renderConfigurationTabView(job) {
+    return <JobConfiguration job={job} />;
+  }
+
+  renderRunHistoryTabView(job) {
+    return <JobRunHistoryTable job={job} />;
+  }
+
+  getActions() {
```
Fabs: I have an impression that this here is breaking a lot of contracts that the menu will rely on. Do you think in your plans to move further you could integrate this work with:

https://github.com/dcos/dcos-ui/pull/3023

I would not like to wait for the full menu to be done in order for this to work.

plugins/jobs/src/js/pages/JobDetailPageContainer.js
```diff
@@ -0,0 +1,110 @@
...
+const ErrorScreen = function() {
+  return (
+    <Page>
+      <Page.Header breadcrumbs={[]} />
+      <RequestErrorMsg />
+    </Page>
+  );
+};
+
+const disabledDialog$ = new BehaviorSubject(null);
```
Fabs: I can't really check that work right, because the dialog is no longer on the menu. 

I think this is another reason to have at least all the itens in the menu working, even if they do nothing.

plugins/jobs/src/js/pages/JobDetailPageContainer.js
DanielMSchmidt: For me, that's what https://jira.mesosphere.com/browse/DCOS-38483 is about.
I limited the scope to no menu items except the edit form, because the dialog was not working previously and I wanted to tackle this explicitly. I am not expecting the actions to work without any more things to be done on this component.

plugins/jobs/src/js/pages/JobDetailPage.js
DanielMSchmidt: This will be done in https://jira.mesosphere.com/browse/DCOS-38483 but as explicitly stated it's not part of this ticket. I needed to cut the scope somehow.

plugins/jobs/src/js/pages/JobDetailPageContainer.js
DanielMSchmidt: You are right, these parts would be tangled in at a later PR, but I can remove it if you like.

DanielMSchmidt:
I answered the comments, therefore I moved my Jira into Review again (slack is down, otherwise I would have written it there 😆 )

nLight:
@DanielMSchmidt oh, that makes sense!

plugins/jobs/src/js/pages/JobConfiguration.js
```diff
@@ -0,0 +1,139 @@
...
+import ConfigurationMap from "#SRC/js/components/ConfigurationMap";
+import ConfigurationMapHeading from "#SRC/js/components/ConfigurationMapHeading";
+import ConfigurationMapLabel from "#SRC/js/components/ConfigurationMapLabel";
+import ConfigurationMapRow from "#SRC/js/components/ConfigurationMapRow";
+import ConfigurationMapSection from "#SRC/js/components/ConfigurationMapSection";
+import ConfigurationMapValue from "#SRC/js/components/ConfigurationMapValue";
+import HashMapDisplay from "#SRC/js/components/HashMapDisplay";
+
+export default class JobConfiguration extends React.Component {
+  static getSchedule(job) {
```
juliangieseke: ❓Is this legacy code? Also, where is this used? I'm asking because this should be handled by `job.schedules.something`…

plugins/jobs/src/js/pages/JobConfiguration.js
```diff
@@ -0,0 +1,139 @@
...
+            <pre className="flush transparent wrap">{job.command}</pre>
+          </ConfigurationMapValue>
+        </ConfigurationMapRow>
+      </ConfigurationMapSection>
+    );
+  }
+
+  getScheduleSection(job) {
+    const [schedule] = job.schedules.nodes;
+    if (schedule == null) {
```
juliangieseke: 🎨 I'd prefer `return (schedule == null) ? null : (<JSX … >)` :trollface:

plugins/jobs/src/js/pages/JobConfiguration.js
```diff
@@ -0,0 +1,139 @@
...
+    );
+  }
+
+  render() {
+    const { job } = this.props;
+
+    return (
+      <div className="container">
+        <ConfigurationMap>
+          {this.getGeneralSection(job)}
```
juliangieseke: 🎨 I think this is legacy code, but personally I dont like these `this.getSomething()` things… why not just create a new `<GeneralSection />` component?

plugins/jobs/src/js/pages/JobDetailPage.js
```diff
@@ -0,0 +1,193 @@
...
+import Page from "#SRC/js/components/Page";
+import TimeAgo from "#SRC/js/components/TimeAgo";
+import TabsMixin from "#SRC/js/mixins/TabsMixin";
+import StringUtil from "#SRC/js/utils/StringUtil";
+import Job from "#SRC/js/structs/Job";
+import JobConfiguration from "./JobConfiguration";
+import { DIALOGS } from "./JobDetailPageContainer";
+import JobRunHistoryTable from "./JobRunHistoryTable";
+
+class JobDetailPage extends mixin(TabsMixin) {
```
juliangieseke: 🎨 Is it possible to relatively easy 🔥that mixin?

plugins/jobs/src/js/pages/JobDetailPage.js
```diff
@@ -0,0 +1,193 @@
...
+  getPrettySchedule(job) {
+    const schedules = job.schedules;
+    if (schedules == null || schedules.length === 0) {
+      return null;
+    }
+
+    const schedule = schedules[0];
+    if (schedule.enabled) {
+      return prettycron.toString(schedule.cron);
+    }
```
juliangieseke: final return missing? 🤔 (if it is impossible to get here, that is no obvious)

plugins/jobs/src/js/pages/JobDetailPage.js
```diff
@@ -0,0 +1,193 @@
...
+  }
+
+  getSubTitle(job) {
+    const nodes = [];
+    const scheduleText = this.getPrettySchedule(job);
+    let longestRunningTask = null;
+    const longestRunningActiveRun = job
```
juliangieseke: This is still working on the struct? 🤔

plugins/jobs/src/js/pages/JobDetailPage.js
```diff
@@ -0,0 +1,193 @@
...
+          job={new Job(JSON.parse(job.json))}
+          open={this.props.jobActionDialog === DIALOGS.EDIT}
+          onClose={this.props.closeDialog}
+        />
+        {this.getDestroyConfirmDialog()}
+      </Page>
+    );
+  }
+}
+
```
juliangieseke: 🎨 tried to figure out which jobs is passed in here, but no proptypes :(

plugins/jobs/src/js/pages/JobDetailPage.js
```diff
@@ -0,0 +1,193 @@
...
+    const nodes = [];
+    const scheduleText = this.getPrettySchedule(job);
+    let longestRunningTask = null;
+    const longestRunningActiveRun = job
+      .getActiveRuns()
+      .getLongestRunningActiveRun();
+
+    if (longestRunningActiveRun != null) {
+      longestRunningTask = longestRunningActiveRun
+        .getTasks()
```
juliangieseke: same as above 🤔

plugins/jobs/src/js/pages/JobDetailPage.js
```diff
@@ -0,0 +1,193 @@
...
+
+import JobFormModalContainer from "#PLUGINS/jobs/src/js/JobFormModalContainer";
+import TaskStates from "#PLUGINS/services/src/js/constants/TaskStates";
+
+import Icon from "#SRC/js/components/Icon";
+import Page from "#SRC/js/components/Page";
+import TimeAgo from "#SRC/js/components/TimeAgo";
+import TabsMixin from "#SRC/js/mixins/TabsMixin";
+import StringUtil from "#SRC/js/utils/StringUtil";
+import Job from "#SRC/js/structs/Job";
```
juliangieseke: Mediator is passing in a new job type, but this is still using the old struct? I'm confused 🤔

plugins/jobs/src/js/pages/JobDetailPageContainer.js
```diff
@@ -0,0 +1,110 @@
...
+  return Observable.combineLatest([
+    job$,
+    props$,
+    jobActionDialog$,
+    disabledDialog$
+  ])
+    .map(([job, props, jobActionDialog, disabledDialog]) => (
+      <JobDetailPage
+        {...props}
+        job={job}
```
juliangieseke: this is the new job, correct? I cant find a place where this is converted into old struct, but UI code is working on the struct… can you clarify please? 😕

plugins/jobs/src/js/pages/JobDetailPageContainer.js
```diff
@@ -0,0 +1,110 @@
+import React from "react";
+import { componentFromStream, graphqlObservable } from "data-service";
+import { Observable, BehaviorSubject } from "rxjs";
+import gql from "graphql-tag";
```
juliangieseke: so this file is the mediator? why is it in `pages` folder? 😕

plugins/jobs/src/js/pages/JobConfiguration.js
DanielMSchmidt: I strongly disagree here

plugins/jobs/src/js/pages/JobConfiguration.js
DanielMSchmidt: Yes, I copied too much there

plugins/jobs/src/js/pages/JobConfiguration.js
DanielMSchmidt: done 👍

plugins/jobs/src/js/pages/JobDetailPage.js
DanielMSchmidt: yes, the mediator is using the graphql resolved json for the most part, except for the `JobFormModalContainer`, which is part of DCOS-37611 AFAIK. It uses the json property on the graphql output to pass a valid version over to the container (it doesn't need the runs and things like that that are filtered out by the json representation)

plugins/jobs/src/js/pages/JobDetailPage.js
DanielMSchmidt: I don't know and this is really out of scope for this PR, IMHO

plugins/jobs/src/js/pages/JobDetailPage.js
DanielMSchmidt: Yeah, final return seems to be missing, I only moved that code over, but I can refactor it to use an early return instead 👍

plugins/jobs/src/js/pages/JobDetailPage.js
DanielMSchmidt: The function was not in use, I removed it.

plugins/jobs/src/js/pages/JobDetailPageContainer.js
DanielMSchmidt: Because it's a page container. As far as I have seen the page containers are always right beside the pages. Where would you have searched for it?

plugins/jobs/src/js/pages/JobDetailPageContainer.js
DanielMSchmidt: Yes it is, the place where it's converted is the modal

plugins/jobs/src/js/pages/JobDetailPageContainer.js
juliangieseke: repository mediators are in `js` folder (one up), Dimas and my own Mediator are there 🤔

plugins/jobs/src/js/pages/JobDetailPageContainer.js
juliangieseke: looks like you addressed the legacy parts 👍

PR #3033, opened 2018-06-27
DCOS-37751: Add new JobOverview meditator which uses data layer

juliangieseke:
This is part 2 of the new overview page, to play around with this, you can navigate to http://localhost:4200/#/jobs/overview2

in order to test this, apply this patch to `jobs.js`:
```diff
@@ -3,7 +3,7 @@ import React from "react";
 /* eslint-enable no-unused-vars */
 import { IndexRoute, Redirect, Route } from "react-router";

-import JobsTabContainer from "#PLUGINS/jobs/src/js/JobsTabContainer";
+import JobsOverview from "#PLUGINS/jobs/src/js/JobsOverview";
 import TaskDetailsTab from "#PLUGINS/services/src/js/pages/task-details/TaskDetailsTab";
 import TaskFileBrowser from "#PLUGINS/services/src/js/pages/task-details/TaskFileBrowser";
 import TaskFileViewer from "#PLUGINS/services/src/js/pages/task-details/TaskFileViewer";
@@ -28,12 +28,12 @@ const jobsRoutes = [
     children: [
       {
         type: Route,
-        component: JobsTabContainer,
+        component: JobsOverview,
         path: "overview",
         children: [
           {
             type: Route,
-            path: ":id"
+            path: ":path"
           }
         ]
       },
There is a very basic UI to verify that the Mediator works as expected, also there are some `console.log` commands to help verifying/debugging it. All of this will be removed and replaced with the proper UI in DCOS-37752 👍

### DIFF
--- a/plugins/jobs/src/js/JobsOverview.tsx
+++ b/plugins/jobs/src/js/JobsOverview.tsx
@@ -1,0 +1,205 @@
+import * as React from "react";
+import Router, {
+  InjectedRouter,
+  LocationDescriptor,
+  Link,
+  withRouter
+} from "react-router";
+
+// tslint:disable-next-line:no-submodule-imports
+import { BehaviorSubject } from "rxjs/BehaviorSubject";
+// tslint:disable-next-line:no-submodule-imports
+import "rxjs/add/observable/empty";
+
+import { graphqlObservable, componentFromStream } from "data-service";
+import gql from "graphql-tag";
+
+// import JobsTabLoading from "./components/JobsTabLoading";
+// import JobsTabError from "./components/JobsTabError";
+// import JobsTabList from "./components/JobsTabList";
+// import JobsTabEmpty from "./components/JobsTabEmpty";
+
+import JobModel from "./data/JobModel";
+import { combineLatest } from "rxjs/observable/combineLatest";
+import { Observable, Subscribable } from "rxjs/Observable";
+import { JobConnection } from "#PLUGINS/jobs/src/js/types/JobConnection";
+import { Job } from "#PLUGINS/jobs/src/js/types/Job";
+
+interface JobsOverviewProps {
+  path: string[];
+  router: InjectedRouter;
+  location: LocationDescriptor;
+  params: Router.Params;
+}
+
+/**
+ * This Component is only to enable us to manually check the mediator,
+ * everything should work as expected - but looks kinda rough :)
+ *
+ * It will removed and replaced by refactored UI components in the next PR.
+ *
+ * Features are:
+ * - working input field
+ * - working links for namespaces
+ *
+ * Known limitations:
+ * - Links to job detail are disabled
+ *
+ * @param props object
+ */
+function DummyComponent(props: any): JSX.Element {
+  function updateFilter(event: React.FormEvent<HTMLInputElement>) {
+    props.handleFilterChange(event.currentTarget.value);
+  }
+
+  // tslint:disable-next-line:no-console
+  console.log("rendering with props", props, props.pathname);
+
+  return (
+    <div style={{ background: "white", width: "100%" }}>
+      <p>Path</p>
+      <p>
+        {props.data &&
+          props.data.path.reduce(
+            (acc: any[], val: string, idx: number) => {
+              const link = (
+                <Link
+                  to={
+                    "/jobs/overview/" +
+                    props.data.path.slice(0, idx + 1).join(".")
+                  }
+                >
+                  {val}
+                </Link>
+              );
+              return acc.concat([" / ", link]);
+            },
+            [
+              <Link key="Jobs" to={"/jobs/overview/"}>
+                Jobs
+              </Link>
+            ]
+          )}
+      </p>
+      <p>Filter</p>
+      <input value={props.filter} onChange={updateFilter} />
+      <p>Jobs</p>
+      {props.data &&
+        props.data.nodes.map((job: Job) => {
+          const next = job.path
+            .concat([job.name])
+            .reduce<string[]>(
+              (acc, val, idx) =>
+                props.data.path[idx] === val ? acc : acc.concat([val]),
+              []
+            );
+          return next.length > 1 ? (
+            <p key={job.id}>
+              <Link
+                to={`/jobs/overview/${props.data.path
+                  .concat([next.slice(0, 1)])
+                  .join(".")}`}
+              >{`link: ${next.slice(0, 1)}`}</Link>
+            </p>
+          ) : (
+            <p key={job.id}>${next} (detail)</p>
+          );
+        })}
+      <p>Children</p>
+      <div>{props.children}</div>
+    </div>
+  );
+}
+
+const JobsOverview = withRouter(
+  componentFromStream<JobsOverviewProps>(
+    (
+      props$: Subscribable<JobsOverviewProps>
+    ): Subscribable<React.ReactNode> => {
+      const jobsOverviewQuery: any = gql`
+        query {
+          jobs(
+            path: $path
+            filter: $filter
+            sortBy: $sortBy
+            sortDirection: $sortDirection
+          ) {
+            path
+            filteredCount
+            totalCount
+            nodes
+          }
+        }
+      `;
+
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: plugins/jobs/src/js/JobsOverview.tsx [hunk 1]
```diff
@@ -0,0 +1,153 @@
...
+      <p>Jobs</p>
+      {props.data &&
+        props.data.nodes.map((job: Job) => <p key={job.id}>{job.name}</p>)}
+      <p>Children</p>
+      <div>{props.children}</div>
+    </div>
+  );
+}
+
+const JobsOverview = withRouter(
```
juliangieseke: I found this which does what our self-written component does, but navigating back in browser history doesnt work currently :(

juliangieseke: Looks like this is only broken for "index" (`/jobs/overview2`) route 🤔

plugins/jobs/src/js/JobsOverview.tsx
```diff
@@ -0,0 +1,165 @@
...
+  componentFromStream<JobsOverviewProps>(
+    (
+      props$: Subscribable<JobsOverviewProps>
+    ): Subscribable<React.ReactNode> => {
+      const jobsOverviewQuery: any = gql`
```
nLight: Curious why have you went with graphql here?

plugins/jobs/src/js/JobsOverview.tsx
```diff
@@ -0,0 +1,165 @@
...
+            totalCount
+            nodes
+          }
+        }
+      `;
+
+      const filter$ = new BehaviorSubject<string>("");
+      const path$ = new BehaviorSubject<string[]>([]);
+
+      (props$ as Observable<JobsOverviewProps>)
```
nLight: it should actually work as you push props 🤔 no `componentDidMount` should be needed

plugins/jobs/src/js/JobsOverview.tsx
juliangieseke: Our understanding when we started working on this whole thing was that we want to get rid of the Structs _now_. This very query (together with the one from Detail Mediator) was the reason why we introduced `JobModel` 🤷‍♂️

plugins/jobs/src/js/JobsOverview.tsx
juliangieseke: it does work when you're navigating between namespaces, but it brakes once you hit "overview root" :(

Example:

imagine there is a job with the full id: `foo.bar.baz.jobname` and you click trough the namespaces, when you navigate back in browser history, `baz -> bar` works, `bar -> foo` also works, but `foo -> /` breaks :( I'll investigate this after I'm done going through our PRs.

plugins/jobs/src/js/JobsOverview.tsx
nLight: Yeah, but as I understand this query has nothing to do with a struct it's just to pass the routing params, no?

plugins/jobs/src/js/JobsOverview.tsx
nLight: So the router doesn't push props to the `JobsOverview` on the `overview2` route because there's nothing to push. It does push `:path` partial when you hit its child route as it should

nLight: Because of that `props$` never get anything when you go back to `overview2`

plugins/jobs/src/js/JobsOverview.tsx
juliangieseke: Well, it has nothing to do with old Job Struct, because it replaces it 😅Not sure if I got your question 100%, so let me try to explain it in a bit more detail.

---

When the query is submitted, the `jobs` resolver returns a response with type `JobConnection`, the requested `nodes` field below contains an array of `Job`, which is the new type we introduced with `JobModel` to replace our Structs. 

`JobConnection` Type: https://github.com/dcos/dcos-ui/blob/6880237bdf904d68bb6f554cc7861d01073e2020/plugins/jobs/src/js/types/JobConnection.ts#L6
New `Job` Type used in `JobConnection.nodes`: https://github.com/dcos/dcos-ui/blob/6880237bdf904d68bb6f554cc7861d01073e2020/plugins/jobs/src/js/types/Job.ts#L41 
=> If you check out field resolvers n both files, they contain all the logic Job Struct had before.

--- 

Also filtering logic moved from UI component into `JobModel`, which is the reason why we "just" pass down the routing params 😅
https://github.com/dcos/dcos-ui/blob/6880237bdf904d68bb6f554cc7861d01073e2020/plugins/jobs/src/js/types/JobConnection.ts#L39
https://github.com/dcos/dcos-ui/blob/6880237bdf904d68bb6f554cc7861d01073e2020/plugins/jobs/src/js/types/JobConnection.ts#L44

---

that said, yes, the query is "just" passing down some params to the resolver 😇
If this didnt answer your initial question/concern, can you please try to clarify it?

plugins/jobs/src/js/JobsOverview.tsx
juliangieseke: hm ok, so I need to change the route in `jobs.js` file to also match "with not path"

plugins/jobs/src/js/JobsOverview.tsx
nLight: oh, now I see. My initial understanding was that we put `router.push` into graphQL for no real reason

juliangieseke: Oh, no 😅

plugins/jobs/src/js/JobsOverview.tsx
juliangieseke: fixed 👍

src/js/routes/jobs.js
```diff
@@ -37,6 +38,17 @@ const jobsRoutes = [
           }
         ]
       },
+      {
+        type: Route,
+        component: JobsOverview,
+        path: "overview2",
```
DanielMSchmidt: Let's not introduce a second route here, but let's only change it for development and testing until we think it is completely ready.

juliangieseke:
@DanielMSchmidt @nLight comments addressed, its ready for review again 👍

src/js/routes/jobs.js
```diff
@@ -3,7 +3,7 @@ import React from "react";
 /* eslint-enable no-unused-vars */
 import { IndexRoute, Redirect, Route } from "react-router";
 
-import JobsTabContainer from "#PLUGINS/jobs/src/js/JobsTabContainer";
+import JobsOverview from "#PLUGINS/jobs/src/js/JobsTabContainer";
```
DanielMSchmidt: Why the renaming? This change seems to be unncessary

juliangieseke: slipped :( the new component is named like this.

DanielMSchmidt: please change, so I can take a final look and approve

nLight:
is it how it is supposed to look like?
screen shot 2018-07-02 at 14 28 16

juliangieseke:
@nLight exactly like this - UI components will be updated in next PR. The code isnt in use yet. 

@DanielMSchmidt I'm on it.

juliangieseke:
@DanielMSchmidt @nLight pushed a fix for 'failureCount' issue, please revisit 🙇

juliangieseke:
Ping @nLight - regarding your UI question, #3046 is the follow-up PR I was talking about 👍

plugins/jobs/src/js/JobsOverview.tsx
```diff
@@ -0,0 +1,205 @@
+import * as React from "react";
+import Router, {
+  InjectedRouter,
+  LocationDescriptor,
+  Link,
+  withRouter
+} from "react-router";
+
+// tslint:disable-next-line:no-submodule-imports
```
nLight: Please remove in the folloup

juliangieseke: will do 👍

PR #3034, opened 2018-06-28
DCOS-38475: Use graphQL mutations to create and update jobs

nLight:
This PR switches jobs form modal create/update actions to graphQL mutations.
Ho special testing setup is needed.

⚠️ Based on #3030 

**Checklist**


plugins/jobs/src/js/JobCreateEditFormModal.tsx
```diff
@@ -27,11 +45,14 @@ const JobCreateEditFormModal = componentFromStream<JobCreateEditFormModalProps>(
...
+    function handleSubmit(
+      isEdit: boolean,
+      data: { jobId: string; jobSpec: JobData }
+    ) {
+      const mutation = isEdit ? updateJobMutation : createJobMutation;
 
-      combineLatest(props$, request$)
-        .take(1) // unsubscribe after the first one to break ∞ react update loop
+      combineLatest(props$, graphqlObservable(mutation, defaultSchema, data))
+        .first() // unsubscribe after the first one to break ∞ react update loop
```
Fabs: This comment is more like if you missed this discussion of first here: https://github.com/dcos/dcos-ui/pull/3028

(unfortunately, I lack the skill to get the link of the discussion :/)

src/js/events/MetronomeClient.ts
```diff
@@ -142,7 +142,7 @@ const defaultHeaders = {
   "Content-Type": "application/json; charset=utf-8"
 };
 
-export function createJob(data: JobData) {
+export function createJob(data: JobDetailResponse): Observable<any> {
```
DanielMSchmidt: Can we please type the response as we need it right now? I postponed writing the types until we need them, but now we do need them, so it would be cool to have types here.

plugins/jobs/src/js/data/JobModel.ts
```diff
@@ -130,6 +133,20 @@ export const resolvers = ({
       _context = {}
     ): Observable<JobLink> {
       return runJob(args.id).map(({ jobId }) => ({ jobId }));
+    },
+    createJob(
+      _parent = {},
+      args: GeneralArgs,
+      _context = {}
+    ): Observable<JobLink> {
+      return createJob(args.data).map(({ jobId }) => ({ jobId }));
```
Fabs: I got this comment on my PR from @juliangieseke. where it would be a good idea to check for the existence of the args.

I think createJob already says the type is mandatory, but since I am not sure our implementation enforces it (could be it tests query validity), you might want to protect from data not existing. 

One other idea would be to check if you can issue the graphqlObservable of a mutation without the data on args, and see if our engine complains.

plugins/jobs/src/js/data/JobModel.ts
```diff
@@ -130,6 +133,20 @@ export const resolvers = ({
...
+    },
+    updateJob(
+      _parent = {},
+      args: GeneralArgs,
+      _context = {}
+    ): Observable<JobLink> {
+      return updateJob(args.id, args.data).map(({ jobId }) => ({ jobId }));
```
Fabs: on the JobModel-test.js there are some tests for the runJob, you could inspire on them to write some tests for those resolvers.

nLight:
@Fabs @juliangieseke addressed all the comments and CI is green \o/

PR #3035, opened 2018-06-28
DCOS-38481: Add breadcrumb to Job Detail Page

DanielMSchmidt:
## How to review

```diff
-import { JobDetailPageContainer } from "../pages/jobs/JobDetailPageContainer";
+import JobDetailPageContainer from "#PLUGINS/jobs/src/js/pages/JobDetailPageContainer";
```

After that check out the Job Detail Page and see if the Breadcrumb works

This is the follow up on https://github.com/dcos/dcos-ui/pull/3032 and it depends on https://github.com/dcos/dcos-ui/pull/3029

**Checklist**

### DIFF
--- a/plugins/jobs/src/js/pages/JobDetailPage.js
+++ b/plugins/jobs/src/js/pages/JobDetailPage.js
@@ -9,15 +9,20 @@ import mixin from "reactjs-mixin";
 import Page from "#SRC/js/components/Page";
 import TabsMixin from "#SRC/js/mixins/TabsMixin";
 import Job from "#SRC/js/structs/Job";
+import Util from "#SRC/js/utils/Util";
 
 import JobFormModalContainer from "../JobFormModalContainer";
 import JobConfiguration from "./JobConfiguration";
 import { DIALOGS } from "./JobDetailPageContainer";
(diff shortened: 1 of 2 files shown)


plugins/jobs/src/js/pages/JobDetailPage.js
```diff
@@ -45,6 +50,27 @@ class JobDetailPage extends mixin(TabsMixin) {
     return <JobRunHistoryTable job={job} />;
   }
 
+  renderBreadcrumbStates(item) {
```
Fabs: 🎨 I would usually prefer to create a `JobDetailPage.Breadcrumbs` "pure subcomponent" on the top namespace. I find it a better pattern so the components are detached and you could in the future refactor them together. Also i prefere that because you don't have to make the this binding and etc.

```js
JobDetailPage.Breadcrumbs = ({item}) => {
   return ...
}

class JobDetailPage extends mixin(TabsMixin) {
....
```

Fabs:
I merged this against the branch and not master. I will open another PR.

Fabs:
Hey @Poltergeist, I merged this one against the wrong branch. Can you please look at this one also to approve it? https://github.com/dcos/dcos-ui/pull/3038

PR #3039, opened 2018-06-29
DCOS-38491: update breadcrumb to meet overview requirements

juliangieseke:
Breadcrumb was already updated to meet requirements for detail page, but we decided to not do all of the required changes for overview in the last PR. This PR now changes the breadcrumbs interface to completely meet the requirements for Overview pages.

**How to test?**
Build should be green (it has unit tests) and when viewing detail page, everything should still work. Overview UI will come in the next PR. 👍


plugins/jobs/src/js/components/Breadcrumbs.js
```diff
@@ -6,57 +6,59 @@ import Breadcrumb from "#SRC/js/components/Breadcrumb";
...
 
-export default function Breadcrumbs({ item, children, states }) {
-  function getBreadcrumb(item, id = "", name = "") {
-    const isDetailPage = [...item.path, item.name].join(".") === id;
-    const link = isDetailPage ? `/jobs/detail/${id}` : `/jobs/overview/${id}`;
-
+export default function Breadcrumbs({ path, name, children, extra }) {
+  /**
+   * @param  {string} link
+   * @param  {string[]} pathname path including name (if given)
```
Fabs: I think this doc should say that pathname should always contain the name, because line 21 does the slice(-1) implying that the last item on the path list will always have this special meaning.

🎨 I would also call it pathWithName, since pathname sounds to me as "the name of the path". But this is more like my opinion.

plugins/jobs/src/js/components/Breadcrumbs.js
```diff
@@ -6,57 +6,59 @@ import Breadcrumb from "#SRC/js/components/Breadcrumb";
...
 Breadcrumbs.propTypes = {
-  states: PropTypes.oneOfType([
-    PropTypes.element,
-    PropTypes.arrayOf(PropTypes.element)
-  ]),
-  item: PropTypes.shape({
-    name: PropTypes.string.isRequired,
-    path: PropTypes.arrayOf(PropTypes.string).isRequired
-  }).isRequired
+  extra: PropTypes.node,
```
Fabs: 🎨 I would try to think of something with a more specific meaning than "extra".  I was not able to understand the difference of the Breadcrumbs children and the extra. At least I would make clear that those extra are nodes, extraChildren, extraContent....

plugins/jobs/src/js/components/Breadcrumbs.js
```diff
@@ -6,57 +6,59 @@ import Breadcrumb from "#SRC/js/components/Breadcrumb";
...
+  // we're starting from the back, so first we have children
+  let breadcrumbs = React.Children.toArray(children);
 
-    return segments.map(segment =>
-      getBreadcrumb(item, segment.join("."), segment[segment.length - 1])
-    );
+  // if we got a name, thats a detail link
+  if (name !== undefined && name !== "") {
+    breadcrumbs = [
+      buildJSX(
```
nLight: then here you would just go 
```js
<PartialBreadcrumb 
  link={["/jobs/detail/"].concat(path, [name]).join(".")}
  pathname={path.concat([name])}
  extra={extra} />
```

plugins/jobs/src/js/components/Breadcrumbs.js
```diff
@@ -6,57 +6,59 @@ import Breadcrumb from "#SRC/js/components/Breadcrumb";
...
-
+export default function Breadcrumbs({ path, name, children, extra }) {
+  /**
+   * @param  {string} link
+   * @param  {string[]} pathname path including name (if given)
+   * @param  {JSX.Element} [extra=null] extra JSX Components to render
+   *
+   * @return {JSX.Element} Breadcrumb Part Component
+   */
+  function buildJSX(link, pathname, extra = null) {
```
nLight: We could actually pull the function on the upper level as a component itself. That would allow us to give it more concise name like `Breadcrumb` or `Crumb` or `Partial` :) etc. `buildJSX` name is too abstract. Why building JSX, why not just using JSX? Plus why keeping a function inside a function when we could just extract it?

nLight: I wouldn't expect something called `pathname` being an array. `pathParts`?

plugins/jobs/src/js/components/Breadcrumbs.js
juliangieseke: I struggled a bit naming this param, because it does _not_ always include the name :/

Imagine there is a job with id `foo.bar.jobname`:
1. when you're on overview path `/`: `pathname` will be `[]`
2. when you're on overview path  `/foo`: `pathname` will be `['foo','bar']`
3. when you're on overview path `/foo.bar`: `pathname` will be `['foo','bar']`
4. when you're on detail page for that job: `pathname` will be `['foo','bar','jobname']`

Now imagine there is a second job with id `nonamespace`, when you're on detail page for _that_ job, `pathname` will be `['nonamespace']` :/

You see, it can contein either only parts of the path, the whole pth, path + name or only the name :confused:

nLight:
Could you please clarify what the specific needs of the Overview page are?

plugins/jobs/src/js/components/Breadcrumbs.js
juliangieseke: Second thing I struggled with 😅

Currently we use this to display additional information (Schedule and Status, which I find it already hard to find one name for) but in theory, this can be anything 🤷‍♂️

plugins/jobs/src/js/components/Breadcrumbs.js
juliangieseke: re function vs component: fair point 👍 

re pathname: see my explanation above, `pathParts` would be okay-ish, but also not completely spot-on 🤔

juliangieseke:
@nLight I started explaining this a bit here: https://github.com/dcos/dcos-ui/pull/3039#discussion_r199163995

The initial reason for this refactoring was, that overview has no job to pass down to the breadcrumb, because it displays multiple of them (or none, if you want to see it like that :D). When I asked Daniel about this he already extracted the `jobInfo` Part into extra components (@DanielMSchmidt thanks a lot again!) so the change I had to make was relatively small.

Also it needs to be able to call the breadcrumb without any path (or an empty one) on overview index / empty & error pages.

plugins/jobs/src/js/components/Breadcrumbs.js
```diff
@@ -6,57 +6,66 @@ import Breadcrumb from "#SRC/js/components/Breadcrumb";
...
-    );
-  }
-
-  function getBreadcrumbList(item) {
-    if (item == null) {
-      return [];
-    }
+/**
+ * @param  {string} link
+ * @param  {string[]} pathParts path including name (if given)
```
DanielMSchmidt: Can we change this comment? I initially thought that pathParts are optional, whereas only the name is optional.

Can we phrase it like: "path of names, empty means overview page"

plugins/jobs/src/js/components/Breadcrumbs.js
```diff
@@ -6,57 +6,66 @@ import Breadcrumb from "#SRC/js/components/Breadcrumb";
...
+  href: PropTypes.string.isRequired,
+  pathParts: PropTypes.arrayOf(PropTypes.string).isRequired
+};
 
-    const pathSegments = [...item.path, item.name];
-    const segments = pathSegments.map((_, index) =>
-      pathSegments.slice(0, index + 1)
-    );
+export default function Breadcrumbs({ path, name, children, jobInfo }) {
+  // we're starting from the back, so first we have children
```
DanielMSchmidt: That we have this comment here shows that the direction is unnatural. Would it really be that much harder to change the direction?

plugins/jobs/src/js/components/Breadcrumbs.js
```diff
@@ -6,57 +6,66 @@ import Breadcrumb from "#SRC/js/components/Breadcrumb";
...
+  let breadcrumbs = React.Children.toArray(children);
 
-    return segments.map(segment =>
-      getBreadcrumb(item, segment.join("."), segment[segment.length - 1])
-    );
+  // if we got a name, thats a detail link
+  if (name !== undefined) {
+    breadcrumbs = [
+      <BreadcrumbPart
+        href={["/jobs/detail/"].concat(path, [name]).join(".")}
```
DanielMSchmidt: 🎨 Can we move this into a separate testable / typeable function? This is too much logic inside a jsx assignment imho

plugins/jobs/src/js/components/Breadcrumbs.js
```diff
@@ -6,57 +6,66 @@ import Breadcrumb from "#SRC/js/components/Breadcrumb";
...
 Breadcrumbs.propTypes = {
-  states: PropTypes.oneOfType([
-    PropTypes.element,
-    PropTypes.arrayOf(PropTypes.element)
-  ]),
-  item: PropTypes.shape({
-    name: PropTypes.string.isRequired,
-    path: PropTypes.arrayOf(PropTypes.string).isRequired
-  }).isRequired
+  jobInfo: PropTypes.node,
```
DanielMSchmidt: I can't see states here anymore, how would they be handled now?

DanielMSchmidt:
@juliangieseke the breadcrumb you have changed should not be in use currently so your review instructions are missing something like this:

```diff
-import { JobDetailPageContainer } from "../pages/jobs/JobDetailPageContainer";
+import JobDetailPageContainer from "#PLUGINS/jobs/src/js/pages/JobDetailPageContainer";
```

juliangieseke:
@DanielMSchmidt @nLight addressed all comments, please revisit 🙇

nLight:
It does not work, sorry.
jobs-bread

Also it may need a rebase. `JobDetailPage` still relies on `JobFormModalContainer`

juliangieseke:
@nLight fixed the issue 👍 it was actually on the master 😭 also sneaked in a small fix for breadcrumbs. snapshot tests didnt catch malformed url :/ /cc @DanielMSchmidt

plugins/jobs/src/js/pages/JobDetailPage.js
```diff
@@ -108,23 +108,21 @@ class JobDetailPage extends mixin(TabsMixin) {
...
-
     return (
       <Page>
         <Page.Header
           actions={this.getActions()}
           breadcrumbs={
             <Breadcrumbs
-              states={this.renderBreadcrumbStates(breadcrumbJob)}
-              item={breadcrumbJob}
+              jobPath={job.path}
```
nLight: How about we pass `job` as the only prop and extract `this.renderBreadcrumbStates(job)` as an internal component into `plugins/jobs/src/js/components/Breadcrumbs.js` because it's a pure function already and uses no state of the `JobDetailPage` class.

That way we would have component composition and separation of responsibility as it doesn't feel like `<JobDetailPage />` component is really responsible for constructing the `<JobInfo />`

So that will become:
```js
<Breadcrumbs job={job} />
```

Or if you decide it's better to still keep `jobInfo`
```js
<Breadcrumbs job={job} jobInfo={<JobInfo job={job} />} />
```

plugins/jobs/src/js/components/Breadcrumbs.js
```diff
@@ -6,57 +6,51 @@ import Breadcrumb from "#SRC/js/components/Breadcrumb";
...
+    const partId = jobPath.concat([jobName]).join(".");
+    breadcrumbParts = breadcrumbParts.concat([
+      <Breadcrumb key={jobName} title="Jobs">
         <BreadcrumbTextContent>
-          <Link to={link}>{name === "" ? "Jobs" : name}</Link>
+          <Link to={"/jobs/detail/" + partId}>{jobName}</Link>
         </BreadcrumbTextContent>
-
-        {isDetailPage ? states : null}
+        {jobInfo}
```
nLight: If we could have `jobInfo` not as a prop but as a component that could become 

```js
<JobInfo job={job} />
```

plugins/jobs/src/js/pages/JobDetailPage.js
juliangieseke: we cant pass `job` because overview, error and loading pages doesnt have one. 😕

also we explicitly wanted to remove that extra logic from the breadcrumb component and allow whatever component uses it to pass in jsx which shall be rendered in the `infofield`.

reason is, and thats IMHO, it is bad practice to pass in such a huge thing and do "magic" on it instead of just pass in what is needed and to as less logic as possible in the component.

plugins/jobs/src/js/components/Breadcrumbs.js
juliangieseke: no, that is exactly what we just got rid of… 😭 (see comment above)

plugins/jobs/src/js/pages/JobDetailPage.js
juliangieseke: we could replace `renderBreadcrumbStates` with a `<JobInfo>` component, but that isnt in scope for this PR.

PR #3040, opened 2018-06-29
DCOS-38641 stop action

Fabs:
**SCOPE**

- Extract the JobStopAction into the mediator pattern
- Minor changes to existing JobStopModal component

**HOW TO TEST**

Scenario 1: A job

Go to the Job History page, run a job, and then selected it on the list (via the checkbox), a Stop button appears, you can stop the job there.

- Not using graphql yet.

### DIFF
--- a/plugins/jobs/src/js/JobStopRun.js
+++ b/plugins/jobs/src/js/JobStopRun.js
@@ -1,0 +1,80 @@
+import * as React from "react";
+import { componentFromStream } from "data-service";
+import { stopJobRun } from "#SRC/js/events/MetronomeClient";
+import PropTypes from "prop-types";
+
+import { Subject } from "rxjs/Subject";
+import "rxjs/add/operator/map";
+import "rxjs/add/operator/switchMap";
+import "rxjs/add/operator/do";
+import "rxjs/add/operator/combineLatest";
+import "rxjs/add/operator/startWith";
+import "rxjs/add/operator/catch";
+
+import JobStopRunModal from "./components/JobStopRunModal";
+
+function executeStopRequest({ jobID, jobRun, onSuccess }) {
+  return stopJobRun(jobID, jobRun)
+    .mapTo({ done: true })
+    .do(_ => onSuccess())
+    .startWith({ done: false });
+}
+
+function stopEventHandler() {
+  const stopSubject$ = new Subject();
+  const stop$ = stopSubject$.switchMap(executeStopRequest).catch(error => {
+    return stop$.startWith({
+      errorMsg: error.response.message,
+      done: true
+    });
+  });
+
+  return {
+    stop$,
+    stopHandler: (jobID, jobRun, onSuccess) => {
+      stopSubject$.next({ jobID, jobRun, onSuccess });
+    }
+  };
+}
+
+export function stopSingleJobRun(stopHandler, jobId, jobRuns, onSuccess) {
+  // TODO DCOS-8763 introduce support for multiple job run ID
+  if (jobRuns.length === 1) {
+    stopHandler(jobId, jobRuns[0], onSuccess);
+  }
+}
+
+export const JobStopRun = componentFromStream(props$ => {
+  const { stop$, stopHandler } = stopEventHandler();
+  const stopRequestState$ = stop$.startWith({ done: null });
+
+  return props$
+    .combineLatest(stopRequestState$, (props, { done }) => {
+      return { ...props, done };
+    })
+    .map(({ jobID, jobRuns, onClose, onSuccess, open, done }) => {
+      function onSuccessHandler() {
+        stopSingleJobRun(stopHandler, jobID, jobRuns, onSuccess);
+      }
+
+      return (
+        <JobStopRunModal
+          jobID={jobID}
+          onClose={onClose}
+          onSuccess={onSuccessHandler}
+          open={open}
+          disabled={done === false}
+          selectedItems={jobRuns}
+        />
+      );
+    });
+});
+
+JobStopRun.propTypes = {
+  jobID: PropTypes.string.isRequired,
+  onClose: PropTypes.func.isRequired,
+  onSuccess: PropTypes.func,
+  open: PropTypes.bool.isRequired,
+  disabled: PropTypes.bool.isRequired,
+  selectedItems: PropTypes.array.isRequired
+};
--- a/plugins/jobs/src/js/components/JobStopRunModal.js
+++ b/plugins/jobs/src/js/components/JobStopRunModal.js
@@ -1,55 +1,9 @@
 import { Confirm } from "reactjs-components";
-import mixin from "reactjs-mixin";
 import PropTypes from "prop-types";
 import React from "react";
-import { StoreMixin } from "mesosphere-shared-reactjs";
+import ModalHeading from "#SRC/js/components/modals/ModalHeading";
 
-import MetronomeStore from "../../stores/MetronomeStore";
-import ModalHeading from "../modals/ModalHeading";
-
-const METHODS_TO_BIND = ["handleButtonConfirm"];
-
-class JobStopRunModal extends mixin(StoreMixin) {
-  constructor() {
-    super(...arguments);
-
-    this.store_listeners = [
-      {
-        name: "metronome",
-        events: ["jobStopRunSuccess", "jobStopRunError"],
-        suppressUpdate: true
-      }
-    ];
-
-    this.state = {
-      pendingRequest: false
-    };
-
-    METHODS_TO_BIND.forEach(method => {
-      this[method] = this[method].bind(this);
-    });
-  }
-
-  handleButtonConfirm() {
-    const { selectedItems, jobID } = this.props;
-    // TODO DCOS-8763 introduce support for multiple job run IDs
-    if (selectedItems.length === 1) {
-      MetronomeStore.stopJobRun(jobID, selectedItems[0]);
-    }
-
-    this.setState({ pendingRequest: true });
-  }
-
-  onMetronomeStoreJobStopRunSuccess() {
-    this.setState({ pendingRequest: false });
-    this.props.onClose();
-    this.props.onSuccess();
-  }
-
-  onMetronomeStoreJobStopRunError() {
-    this.props.onClose();
-  }
-
+class JobStopRunModal extends React.Component {
   getContentHeader(selectedItems, selectedItemsLength) {
     let headerContent = ` ${selectedItemsLength} Job Runs`;
     if (selectedItemsLength === 1) {
@@ -98,7 +52,7 @@ class JobStopRunModal extends mixin(StoreMixin) {
     return (
       <Confirm
         closeByBackdropClick={true}
-        disabled={this.state.pendingRequest}
+        disabled={this.props.disabled}
         header={this.getContentHeader(selectedItems, selectedItemsLength)}
         open={open}
         onClose={onClose}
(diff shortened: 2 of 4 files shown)


plugins/jobs/src/js/JobStopRun.js
```diff
@@ -0,0 +1,69 @@
...
+const JobStopRun = componentFromStream(props$ => {
+  const { stop$, stopHandler } = stopOperation();
+  const stopEmit$ = stop$.startWith({ done: null });
+
+  return props$
+    .combineLatest(stopEmit$, (props, stopOperation) => {
+      return { ...props, ...stopOperation };
+    })
+    .map(({ jobID, jobRuns, onClose, onSuccess, open, done }) => {
+      const disabled = done === false;
```
nLight: I'd go with `isDisabled` as a well known pattern in JS for booleans.

plugins/jobs/src/js/JobStopRun.js
```diff
@@ -0,0 +1,69 @@
...
+
+      const onSuccessEvent = () => {
```
nLight: `onSuccessHandler`?

nLight: We have no-bind/no-arrow functions in JSX rule as it causes performance degradation. Extracting function may confuse auto-checks, though it doesn't fix the issue :) I'd strongly considered not creating ad-hoc functions here.

plugins/jobs/src/js/JobStopRun.js
```diff
@@ -0,0 +1,69 @@
...
+    stop$,
+    stopHandler: (jobID, jobRun, onSuccess) => {
+      stopSubject$.next({ jobID, jobRun, onSuccess });
+    }
+  };
+}
+
+const JobStopRun = componentFromStream(props$ => {
+  const { stop$, stopHandler } = stopOperation();
+  const stopEmit$ = stop$.startWith({ done: null });
```
nLight: Why do we default `executeStop` to `{done: false}` just to redefine it here to `{done: null}`?

nLight: If we could generalize `startWith` we wouldn't need coming up we a name like `stopEmit$` if we can't I'd be looking for a better name.

plugins/jobs/src/js/components/JobStopRunModal.js
```diff
@@ -0,0 +1,86 @@
...
+      </div>
+    );
+  }
+
+  render() {
+    const { onClose, open, selectedItems } = this.props;
+    let rightButtonText = "Stop Job Run";
+    const selectedItemsLength = selectedItems.length;
+
+    if (selectedItems.length > 1) {
```
nLight: I would actually go with ternary here. Ternary sends a clear message that the value could be either or. In this case default value and this logic are even broken in space

plugins/jobs/src/js/components/JobStopRunModal.js
```diff
@@ -0,0 +1,86 @@
...
+      <div className="text-align-center">
+        {this.getConfirmTextBody(selectedItems, selectedItemsLength)}
+      </div>
+    );
+  }
+
+  render() {
+    const { onClose, open, selectedItems } = this.props;
+    let rightButtonText = "Stop Job Run";
+    const selectedItemsLength = selectedItems.length;
```
nLight: We assign `selectedItems.length` and then use it again 1 line after.

plugins/jobs/src/js/JobStopRun.js
Fabs: There is a semantic difference from null and false. `null` means I never made a request, `false` means I am in the middle of one.  Because we combine latest we need to emit once, so the component renders. 

Notice that `executeStop` is inside `switchMap` so it is not going to be triggered at loading. So what actually happens is:

We load the component with `done: null`, when we trigger a request done alternate between `false` and `true`, staying at `true` most of the time when a request is not happening.

plugins/jobs/src/js/components/JobStopRunModal.js
Fabs: Can look into that. I am trying to champion a non-multiline-ternary rule. So if I can express it here in a single line. Sure.

plugins/jobs/src/js/components/JobStopRunModal.js
Fabs: There is a lot of things I don't like on this component, but it a plain UI component from which I took of all the data logic to do the data layer refactoring. I don't plan to rewrite this.

plugins/jobs/src/js/components/JobStopRunModal.js
nLight: Any specific reason for championing except of a taste?

plugins/jobs/src/js/JobStopRun.js
Fabs: Thanks, that is actually important. Keep an eye on it.

plugins/jobs/src/js/components/JobStopRunModal.js
Fabs: There is a larger discussion here: https://jira.mesosphere.com/browse/DCOS-38625.

I personally find readability to be the major issue as the markers for conditional are `?` and `:`, over multiline it gets hard to see there is a condition.

plugins/jobs/src/js/JobStopRun.js
Fabs: Yeap.

Fabs:
@nLight I addressed the comments regarding the new code I wrote.

plugins/jobs/src/js/components/JobStopRunModal.js
Fabs: Daniel made a comment here https://github.com/dcos/dcos-ui/pull/3043/files, that made me inline some code that made me fix that issue on #3043.

plugins/jobs/src/js/components/JobStopRunModal.js
```diff
@@ -0,0 +1,86 @@
...
+JobStopRunModal.propTypes = {
+  jobID: PropTypes.string.isRequired,
+  onClose: PropTypes.func.isRequired,
+  onSuccess: PropTypes.func,
+  open: PropTypes.bool.isRequired,
+  disabled: PropTypes.bool.isRequired,
+  selectedItems: PropTypes.array.isRequired
+};
+
+module.exports = JobStopRunModal;
```
DanielMSchmidt: 🎨 Can we make this a default export inline with the class declaration?

plugins/jobs/src/js/JobStopRun.js
```diff
@@ -0,0 +1,70 @@
...
+  }
+}
+
+export const JobStopRun = componentFromStream(props$ => {
+  const { stop$, stopHandler } = stopOperation();
+  const stopEmit$ = stop$.startWith({ done: null });
+
+  return props$
+    .combineLatest(stopEmit$, (props, stopOperation) => {
+      return { ...props, ...stopOperation };
```
juliangieseke: this is _really_ hard to reason about, what is the content of `props` and especially `stopOperation` here? 😢 please add some kind of types for both somewhere or dont use spread operator.

plugins/jobs/src/js/JobStopRun.js
```diff
@@ -0,0 +1,70 @@
...
+import "rxjs/add/operator/map";
+import "rxjs/add/operator/switchMap";
+import "rxjs/add/operator/do";
+import "rxjs/add/operator/combineLatest";
+import "rxjs/add/operator/startWith";
+import "rxjs/add/operator/catch";
+
+import JobStopRunModal from "./components/JobStopRunModal";
+
+function executeStop({ jobID, jobRun, onSuccess }) {
```
juliangieseke: can you please add JSDoc comments to all of these functions? I cant really follow the code below without types :(

juliangieseke:
looks like #3043 is overhauling most of this code anyway, so to prevent rebase-hell I'd say it is good enough if you add the types (and maybe change the ternary) in #3043 - no need to fix it here. what about you @nLight?

Fabs:
I would prefer actually fixing that here, and leave to fix graphql stuff on #3043.

plugins/jobs/src/js/JobStopRun.js
Fabs: I gave it better function names, and I remove splat operations where I believe it is confusion.

plugins/jobs/src/js/JobStopRun.js
Fabs: Added prop types, remove the splat of the stop stream, renamed also to make it more clear. I still splat props, but now there is no ambiguity on which stream the data comes from.

Thanks for raising that up, please let me know if your life got better.

plugins/jobs/src/js/components/JobStopRunModal.js
```diff
@@ -0,0 +1,86 @@
...
+
+    return (
+      <ModalHeading key="confirmHeader">
+        {`Are you sure you want to stop ${headerContent}?`}
+      </ModalHeading>
+    );
+  }
+
+  getConfirmTextBody(selectedItems, selectedItemsLength) {
+    let bodyText;
```
nLight: I still think it's more concise and easier to read and follow :D prettier uses prefix operator notation so `?` and `:` are very prominent.
```js
const bodyText =
  selectedItemsLength === 1
    ? `the job run with id ${selectedItems[0]}`
    : "the selected job runs";
```

Look ma no mutation! :D

plugins/jobs/src/js/components/JobStopRunModal.js
```diff
@@ -0,0 +1,86 @@
...
+    return <span key="confirmText">You are about to stop {bodyText}.</span>;
+  }
+
+  getModalContents() {
+    const { selectedItems } = this.props;
+    const selectedItemsLength = selectedItems.length;
+
+    return (
+      <div className="text-align-center">
+        {this.getConfirmTextBody(selectedItems, selectedItemsLength)}
```
nLight: It appears to me that this function needs only 1 argument: `selectedItems`

DanielMSchmidt:
@Fabs @juliangieseke is the "changes requested" up to date? I am unsure if this is ready again for review or not 🙈

Fabs:
@DanielMSchmidt, I will ping you when its done, should be a couple more minutes.

DanielMSchmidt:
@Fabs Just move the ticket back into "In Review" ;)

plugins/jobs/src/js/JobStopRun.js
```diff
@@ -0,0 +1,80 @@
...
+import "rxjs/add/operator/do";
+import "rxjs/add/operator/combineLatest";
+import "rxjs/add/operator/startWith";
+import "rxjs/add/operator/catch";
+
+import JobStopRunModal from "./components/JobStopRunModal";
+
+function executeStopRequest({ jobID, jobRun, onSuccess }) {
+  return stopJobRun(jobID, jobRun)
+    .map(_ => ({ done: true }))
```
nLight: We could use `mapTo` here http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html#instance-method-mapTo and spare a function

Fabs:
@DanielMSchmidt , @nLight  and @juliangieseke. That is ready for review again.

plugins/jobs/src/js/components/JobStopRunModal.js
Fabs: I agree, but I am still not willing to start improving this UI component out of the scope of refactoring the Store out of it.

plugins/jobs/src/js/JobStopRun.js
Fabs: Nice :D.

plugins/jobs/src/js/components/JobStopRunModal.js
Fabs: That is also true, but also some presentational logic that I am not willing to touch.

PR #3042, opened 2018-07-01
DCOS-38492 delete job menu refactor

Fabs:
**SCOPE**

- Extract the JobDeleteModal into its own component
- Use the mediator pattern to delete jobs

**HOW TO TEST**

Scenario 1: Simple Job
Create a Job, and remove it. Try with some 3G slow to see if the modal states go fine

Scenario 2: Running Job
If the job is running, the modal should ask for extra confirmation.

Compare all these scenarios with a running cluster.

- Not using graphql yet.


plugins/jobs/src/js/pages/JobDetailPage.js
```diff
@@ -11,7 +11,7 @@ import TabsMixin from "#SRC/js/mixins/TabsMixin";
 import Job from "#SRC/js/structs/Job";
 import Util from "#SRC/js/utils/Util";
 
-import JobFormModalContainer from "../JobFormModalContainer";
+import JobCreateEditFormModal from "../JobCreateEditFormModal";
```
Fabs: Fixed that, I think two things got merged that were incompatible.

plugins/jobs/src/js/JobDelete.js
```diff
@@ -0,0 +1,70 @@
...
+  };
+}
+
+const JobDelete = componentFromStream(prop$ => {
+  const { delete$, deleteHandler } = deleteOperation();
+  const deleteEmit$ = delete$.startWith({ done: null });
+
+  return prop$
+    .combineLatest(deleteEmit$, (props, deleteOp) => {
+      return { ...props, ...deleteOp };
```
juliangieseke: basically the same as in #3040 😢(hard to reason about due to missing types and spread Operator)

plugins/jobs/src/js/JobDelete.js
```diff
@@ -0,0 +1,70 @@
...
+import "rxjs/add/operator/startWith";
+import "rxjs/add/operator/catch";
+
+import JobDeleteModal from "./components/JobDeleteModal";
+
+function errorIsTaskCurrentRunning(errorMsg) {
+  return /stopCurrentJobRuns=true/.test(errorMsg);
+}
+
+function executeDelete({ jobId, stopCurrentJobRuns, onSuccess, errorMsg }) {
```
juliangieseke: can you please add JSDoc comments describing the types to all of these functions?

plugins/jobs/src/js/components/JobDeleteModal.js
```diff
@@ -0,0 +1,63 @@
...
+  return stopCurrentJobRuns ? stopCurrentRunsLabel : defaultLabel;
+}
+
+const ModalMessage = ({ jobId, stopCurrentJobRuns }) => {
+  const defaultMessage = `Are you sure you want to ${
+    UserActions.DELETE
+  } ${jobId}? This action is irreversible.`;
+
+  const stopCurrentRunsMessage = `Couldn't ${
+    UserActions.DELETE
```
juliangieseke: was this prettier? 😱

juliangieseke:
looks like #3043 is overhauling most of this code anyway, so to prevent rebase-hell I'd say it is good enough if you add the types  in #3043 - no need to fix it here. (but there is a conflict :( )

plugins/jobs/src/js/components/JobDeleteModal.js
Fabs: Yes.

Fabs:
@juliangieseke I really like you trying to help that move forward. but I would rather fix this PR, and only fix graphql stuff on the #3043

nLight:
Flicts!

plugins/jobs/src/js/components/JobDeleteModal.js
nLight: Prettier would be fine with a multiline string. Dunno if our component would be fine with it though 🤔 
```js
const stopCurrentRunsMessage = `Couldn't ${UserActions.DELETE} ${jobId} as there
 are currently active job runs. Do you want to stop all runs and
 ${UserActions.DELETE} the job?`;
```

plugins/jobs/src/js/JobDelete.js
```diff
@@ -0,0 +1,70 @@
...
+  return deleteJob(jobId, stopCurrentJobRuns)
+    .map(_ => ({ done: true, stopCurrentJobRuns, errorMsg }))
```
nLight: We could go with `mapTo` for consistency

plugins/jobs/src/js/JobDelete.js
```diff
@@ -0,0 +1,70 @@
...
+    .do(_ => onSuccess())
```
nLight: 🤔 wonder if `.do(onSuccess)` would do

plugins/jobs/src/js/JobDelete.js
```diff
@@ -0,0 +1,70 @@
...
+    .startWith({ done: false, stopCurrentJobRuns, errorMsg });
+}
+
+function deleteOperation() {
+  const deleteSubject$ = new Subject();
+  const delete$ = deleteSubject$.switchMap(executeDelete).catch(error => {
+    return delete$.startWith({
+      errorMsg: error.response.message,
```
nLight: Are we sure `error.response.message` is always there?

plugins/jobs/src/js/JobDelete.js
```diff
@@ -0,0 +1,70 @@
...
+      done: true
+    });
+  });
+
+  return {
+    delete$,
+    deleteHandler: (jobId, stopCurrentJobRuns, onSuccess, errorMsg) => {
```
nLight: Looks like `errorMsg` is a pattern of ours, though I'd be willing to start rewriting those with `errorMessage`

plugins/jobs/src/js/JobDelete.js
nLight: When I see return to be the only expression I do
```js
  return prop$.combineLatest(deleteEmit$, (props, deleteOp) => ({
    ...props,
    ...deleteOp
  }));
```

plugins/jobs/src/js/JobDelete.js
```diff
@@ -0,0 +1,70 @@
...
+import "rxjs/add/operator/map";
+import "rxjs/add/operator/switchMap";
+import "rxjs/add/operator/do";
+import "rxjs/add/operator/combineLatest";
+import "rxjs/add/operator/startWith";
+import "rxjs/add/operator/catch";
+
+import JobDeleteModal from "./components/JobDeleteModal";
+
+function errorIsTaskCurrentRunning(errorMsg) {
```
nLight: Still missing the point of `errorIsTaskCurrentRunning` https://github.com/dcos/dcos-ui/pull/3043#discussion_r199782377

plugins/jobs/src/js/JobDelete.js
nLight: That is the only question that needs an action ☝️

plugins/jobs/src/js/JobDelete.js
Fabs: It will change the semantic and pass the value emited from the stream to onSucess. I don't want it to happen.

plugins/jobs/src/js/JobDelete.js
Fabs: yes :D

plugins/jobs/src/js/JobDelete.js
Fabs: It is already fixed there.

plugins/jobs/src/js/JobDelete.js
Fabs: done

plugins/jobs/src/js/JobDelete.js
Fabs: Yes. This comes from request, but I can guard against that.

plugins/jobs/src/js/JobDelete.js
Fabs: I had that impulse too, but did not. I can do the full name here.

plugins/jobs/src/js/JobDelete.js
Fabs: I usually avoid that pattern because soon enough it grows to too lines and then I have to change to the full form.

plugins/jobs/src/js/components/JobDeleteModal.js
Fabs: When I run prettier, it reverts it back. Might be something with my editor.

Fabs:
@DanielMSchmidt, @nLight  and @juliangieseke. Ready again.

Fabs:
I was sure that this had two reviews, but I guess @DanielMSchmidt one's got dismissed.

plugins/jobs/src/js/JobDelete.js
```diff
@@ -0,0 +1,92 @@
...
+import "rxjs/add/operator/switchMap";
+import "rxjs/add/operator/do";
+import "rxjs/add/operator/combineLatest";
+import "rxjs/add/operator/startWith";
+import "rxjs/add/operator/catch";
+
+import JobDeleteModal from "./components/JobDeleteModal";
+
+function isTaskCurrentRunning(errorMessage) {
+  return /stopCurrentJobRuns=true/.test(errorMessage);
```
juliangieseke: 🎨 why did you went with a regex for such a simple match? I think it is _way_ faster to just use `String.includes` 🤔

juliangieseke: ah, I did comment on it :D

PR #3043, opened 2018-07-01
DCOS-37617 Move menu actions to Graphql

Fabs:
⚠DEPENDS: ~#3040~, ~#3042~

Ignore the merge commits, they will be dropped once we have the dependency PRs in.

**SCOPE**
- Move the delete action to graphql
- Move the stop action to graphql
- Move the toggle schedule action to graphql

**HOW TO TEST**

Basically, go over all the menu scenarios. Stop, Start, Schedule, Remove, Edit. Think a bit of the corner cases too.

Nothing!

### DIFF
--- a/plugins/jobs/src/js/JobDelete.js
+++ b/plugins/jobs/src/js/JobDelete.js
@@ -1,7 +1,6 @@
 import * as React from "react";
-import { componentFromStream } from "data-service";
+import { componentFromStream, graphqlObservable } from "data-service";
 import PropTypes from "prop-types";
-import { deleteJob } from "#SRC/js/events/MetronomeClient";
 import { Subject } from "rxjs/Subject";
 import "rxjs/add/operator/map";
 import "rxjs/add/operator/switchMap";
@@ -10,19 +9,29 @@ import "rxjs/add/operator/combineLatest";
 import "rxjs/add/operator/startWith";
 import "rxjs/add/operator/catch";
 
+import gql from "graphql-tag";
+
 import JobDeleteModal from "./components/JobDeleteModal";
+import defaultSchema from "./data/JobModel";
 
-function isTaskCurrentRunning(errorMessage) {
-  return /stopCurrentJobRuns=true/.test(errorMessage);
-}
+const deleteJobMutation = gql`
+  mutation {
+    deleteJob(id: $jobId, stopCurrentJobRuns: $stopCurrentJobRuns) {
+      jobId
+    }
+  }
+`;
 
-function executeDeleteRequest({
+function executeDeleteMutation({
   jobId,
   stopCurrentJobRuns,
   onSuccess,
   errorMessage
 }) {
-  return deleteJob(jobId, stopCurrentJobRuns)
+  return graphqlObservable(deleteJobMutation, defaultSchema, {
+    jobId,
+    stopCurrentJobRuns
+  })
     .mapTo({ done: true, stopCurrentJobRuns, errorMessage })
     .do(_ => onSuccess())
     .startWith({ done: false, stopCurrentJobRuns, errorMessage });
@@ -31,7 +40,7 @@ function executeDeleteRequest({
 function deleteEventHandler() {
   const deleteSubject$ = new Subject();
   const delete$ = deleteSubject$
-    .switchMap(executeDeleteRequest)
+    .switchMap(executeDeleteMutation)
     .catch(error => {
       return delete$.startWith({
         errorMessage: error && error.response && error.response.message,
@@ -52,6 +61,10 @@ function deleteEventHandler() {
   };
 }
 
+function isTaskCurrentlyRunning(errorMessage) {
+  return (errorMessage || "").includes("stopCurrentJobRuns=true");
+}
+
 const JobDelete = componentFromStream(prop$ => {
   const { delete$, deleteHandler } = deleteEventHandler();
   const deleteWithInitialValue$ = delete$.startWith({ done: null });
@@ -61,7 +74,7 @@ const JobDelete = componentFromStream(prop$ => {
       return { ...props, done, errorMessage };
     })
     .map(({ open, jobId, onClose, onSuccess, errorMessage, done }) => {
-      const stopCurrentJobRuns = isTaskCurrentRunning(errorMessage);
+      const stopCurrentJobRuns = isTaskCurrentlyRunning(errorMessage);
 
       function onSuccessEvent() {
         deleteHandler(jobId, stopCurrentJobRuns, onSuccess, errorMessage);
@@ -85,7 +98,6 @@ JobDelete.propTypes = {
   onClose: PropTypes.func.isRequired,
   onSuccess: PropTypes.func,
   open: PropTypes.bool.isRequired,
-  disabled: PropTypes.bool.isRequired,
   selectedItems: PropTypes.array.isRequired
 };
 
--- a/plugins/jobs/src/js/JobStopRun.js
+++ b/plugins/jobs/src/js/JobStopRun.js
@@ -1,8 +1,7 @@
 import * as React from "react";
-import { componentFromStream } from "data-service";
-import { stopJobRun } from "#SRC/js/events/MetronomeClient";
 import PropTypes from "prop-types";
 
+import { componentFromStream, graphqlObservable } from "data-service";
 import { Subject } from "rxjs/Subject";
 import "rxjs/add/operator/map";
 import "rxjs/add/operator/switchMap";
@@ -10,11 +9,25 @@ import "rxjs/add/operator/do";
 import "rxjs/add/operator/combineLatest";
 import "rxjs/add/operator/startWith";
 import "rxjs/add/operator/catch";
+import gql from "graphql-tag";
 
 import JobStopRunModal from "./components/JobStopRunModal";
 
-function executeStopRequest({ jobID, jobRun, onSuccess }) {
-  return stopJobRun(jobID, jobRun)
+import defaultSchema from "./data/JobModel";
+
+const stopJobRunMutation = gql`
+  mutation {
+    stopJobRun(id: $jobId, jobRunId: $jobRunId) {
+      jobId
+    }
+  }
+`;
+
+function executeStopJobRunMutation({ jobId, jobRunId, onSuccess }) {
+  return graphqlObservable(stopJobRunMutation, defaultSchema, {
+    jobId,
+    jobRunId
+  })
     .mapTo({ done: true })
     .do(_ => onSuccess())
     .startWith({ done: false });
@@ -22,17 +35,19 @@ function executeStopRequest({ jobID, jobRun, onSuccess }) {
 
 function stopEventHandler() {
   const stopSubject$ = new Subject();
-  const stop$ = stopSubject$.switchMap(executeStopRequest).catch(error => {
-    return stop$.startWith({
-      errorMsg: error.response.message,
-      done: true
+  const stop$ = stopSubject$
+    .switchMap(executeStopJobRunMutation)
+    .catch(error => {
+      return stop$.startWith({
+        errorMessage: error && error.response && error.response.message,
+        done: true
+      });
     });
-  });
 
   return {
     stop$,
-    stopHandler: (jobID, jobRun, onSuccess) => {
-      stopSubject$.next({ jobID, jobRun, onSuccess });
+    stopHandler: (jobId, jobRunId, onSuccess) => {
+      stopSubject$.next({ jobId, jobRunId, onSuccess });
     }
   };
 }
--- a/plugins/jobs/src/js/components/JobStopRunModal.js
+++ b/plugins/jobs/src/js/components/JobStopRunModal.js
@@ -29,30 +29,19 @@ class JobStopRunModal extends React.Component {
     return <span key="confirmText">You are about to stop {bodyText}.</span>;
   }
 
-  getModalContents() {
-    const { selectedItems } = this.props;
-    const selectedItemsLength = selectedItems.length;
-
-    return (
-      <div className="text-align-center">
-        {this.getConfirmTextBody(selectedItems, selectedItemsLength)}
-      </div>
-    );
-  }
-
   render() {
-    const { onClose, open, selectedItems } = this.props;
-    let rightButtonText = "Stop Job Run";
+    const { onClose, open, selectedItems, disabled } = this.props;
     const selectedItemsLength = selectedItems.length;
+    let rightButtonText = "Stop Job Run";
 
-    if (selectedItems.length > 1) {
+    if (selectedItemsLength > 1) {
       rightButtonText = "Stop Job Runs";
     }
 
     return (
       <Confirm
         closeByBackdropClick={true}
-        disabled={this.props.disabled}
+        disabled={disabled}
         header={this.getContentHeader(selectedItems, selectedItemsLength)}
         open={open}
         onClose={onClose}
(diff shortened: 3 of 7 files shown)


plugins/jobs/src/js/JobDelete.js
```diff
@@ -0,0 +1,90 @@
...
+import "rxjs/add/operator/combineLatest";
+import "rxjs/add/operator/startWith";
+import "rxjs/add/operator/catch";
+
+import gql from "graphql-tag";
+
+import JobDeleteModal from "./components/JobDeleteModal";
+import defaultSchema from "./data/JobModel";
+
+const deleteJob = gql`
```
DanielMSchmidt: 🎨 might be only me, but I find the naming confusing. I went ahead and looked up the definition of `graphqlObservable` because I thought a function would be passed to it

plugins/jobs/src/js/JobDelete.js
```diff
@@ -0,0 +1,90 @@
...
+
+  return {
+    delete$,
+    deleteHandler: (jobId, stopCurrentJobRuns, onSuccess, errorMsg) => {
+      deleteSubject$.next({ jobId, stopCurrentJobRuns, onSuccess, errorMsg });
+    }
+  };
+}
+
+function errorIsTaskCurrentRunning(errorMsg) {
```
DanielMSchmidt: I like the verbosity of this funcion name a lot :)

plugins/jobs/src/js/JobDelete.js
```diff
@@ -0,0 +1,90 @@
...
+  return /stopCurrentJobRuns=true/.test(errorMsg);
+}
+
+const JobDelete = componentFromStream(prop$ => {
+  const { delete$, deleteHandler } = deleteOperation();
```
DanielMSchmidt: We would create a new subject every time the props change, does this really make sense?
(I see this in multiple occasions but choose to only comment it here to discuss it more easily)

plugins/jobs/src/js/JobDelete.js
```diff
@@ -0,0 +1,90 @@
...
+  const deleteEmit$ = delete$.startWith({ done: null });
+
+  return prop$
+    .combineLatest(deleteEmit$, (props, deleteOp) => {
+      return { ...props, ...deleteOp };
+    })
+    .map(({ open, jobId, onClose, onSuccess, errorMsg, done }) => {
+      const isDisabled = done === false;
+      const isOpen = open;
```
DanielMSchmidt: This variable could easily be removed :)

plugins/jobs/src/js/JobDelete.js
```diff
@@ -0,0 +1,90 @@
...
+const JobDelete = componentFromStream(prop$ => {
+  const { delete$, deleteHandler } = deleteOperation();
+  const deleteEmit$ = delete$.startWith({ done: null });
+
+  return prop$
+    .combineLatest(deleteEmit$, (props, deleteOp) => {
+      return { ...props, ...deleteOp };
+    })
+    .map(({ open, jobId, onClose, onSuccess, errorMsg, done }) => {
+      const isDisabled = done === false;
```
DanielMSchmidt: What would you think of inlining this? My reasoning is that if you see this at the top of this function you need to keep it in mind and see if anything reacts to it, just to see that it is simply passed through

plugins/jobs/src/js/JobDelete.js
```diff
@@ -0,0 +1,90 @@
...
+      const isOpen = open;
+
+      const stopCurrentJobRuns = errorIsTaskCurrentRunning(errorMsg);
+
+      const onSuccessEvent = () => {
+        deleteHandler(jobId, stopCurrentJobRuns, onSuccess, errorMsg);
+      };
+
+      const onCloseEvent = () => {
+        onClose();
```
DanielMSchmidt: Can't we simply pass onClose into the modal?

plugins/jobs/src/js/JobStopRun.js
```diff
@@ -0,0 +1,81 @@
...
+
+  return props$
+    .combineLatest(stopEmit$, (props, stopOperation) => {
+      return { ...props, ...stopOperation };
+    })
+    .map(({ jobID, jobRuns, onClose, onSuccess, open, done }) => {
+      const disabled = done === false;
+
+      const onSuccessEvent = () => {
+        if (jobRuns.length === 1) {
```
DanielMSchmidt: This is really confusing to me, why do we check for `jobRuns`?

plugins/jobs/src/js/JobStopRun.js
```diff
@@ -0,0 +1,81 @@
...
+
+  return {
+    stop$,
+    stopHandler: (jobId, jobRunId, onSuccess) => {
+      stopSubject$.next({ jobId, jobRunId, onSuccess });
+    }
+  };
+}
+
+const JobStopRun = componentFromStream(props$ => {
```
DanielMSchmidt: Can we please add some basic tests for this? I would at least like to test the `if (jobRuns.length === 1) {` line below, see if a mutation goes through and if we can deal with an API failure gracefully

plugins/jobs/src/js/components/JobDeleteModal.js
```diff
@@ -0,0 +1,77 @@
...
+  };
+}
+
+function modalContent(jobId, stopCurrentJobRuns) {
+  const contentBuilder = stopCurrentJobRuns ? stopJobsCotent : defaultContent;
+
+  return contentBuilder(jobId);
+}
+
+function modalContentHtml(message) {
```
DanielMSchmidt: Let's make this a component on its own 👍

plugins/jobs/src/js/components/JobDeleteModal.js
```diff
@@ -0,0 +1,77 @@
...
+
+  const defaultLabel = `${StringUtil.capitalize(UserActions.DELETE)} Job`;
+
+  return {
+    label: defaultLabel,
+    message: modalContentHtml(defaultMessage)
+  };
+}
+
+function stopJobsCotent(jobId) {
```
DanielMSchmidt:
```diff
- function stopJobsCotent(jobId) {
+ function stopJobsContent(jobId) {
```

plugins/jobs/src/js/components/JobDeleteModal.js
```diff
@@ -0,0 +1,77 @@
...
+    UserActions.DELETE
+  } the job?`;
+
+  return {
+    label: stopCurrentRunsLabel,
+    message: modalContentHtml(stopCurrentRunsMessage)
+  };
+}
+
+function modalContent(jobId, stopCurrentJobRuns) {
```
DanielMSchmidt: This is one of the rare cases where I think that having these put into functions makes it more unreadable and harder to understand. Can we either do an if on the call-side or get rid of this builder pattern?

plugins/jobs/src/js/components/JobStopRunModal.js
```diff
@@ -0,0 +1,86 @@
...
+        onClose={onClose}
+        leftButtonText="Cancel"
+        leftButtonCallback={onClose}
+        leftButtonClassName="button button-primary-link"
+        rightButtonText={rightButtonText}
+        rightButtonClassName="button button-danger"
+        rightButtonCallback={this.props.onSuccess}
+        showHeader={true}
+      >
+        {this.getModalContents()}
```
DanielMSchmidt: 🎨 we could also just inline this

plugins/jobs/src/js/data/JobModel.ts
```diff
@@ -141,7 +156,31 @@ export const resolvers = ({
       args: GeneralArgs,
       _context = {}
     ): Observable<JobLink> {
-      return runJob(args.id).map(({ jobId }) => ({ jobId }));
+      if (args.id) {
```
DanielMSchmidt: What would you think about using a special interface here and using a user-defined type check?

plugins/jobs/src/js/data/JobModel.ts
```diff
@@ -141,7 +156,31 @@ export const resolvers = ({
...
+        return runJob(args.id).map(({ jobId }) => ({ jobId }));
+      }
+
+      return Observable.throw({
```
DanielMSchmidt: I think we need to talk about what we want to throw as a team. I in the other resolvers we just throw strings and I would like to see it concise, but I think having an object with additional info is also nice 🤔

plugins/jobs/src/js/JobDelete.js
Fabs: I can verify that. But I think that is not the case. componentFromStream will be called only once to generate the component. I believe this function will be called on construction or mounting. 

Notice that this is outside of the `prop$.combine.map..` which means that it does not get called on every property refresh.

plugins/jobs/src/js/JobDelete.js
Fabs: Sure, Thanks.

plugins/jobs/src/js/JobDelete.js
Fabs: Yeap.

plugins/jobs/src/js/JobStopRun.js
Fabs: This is a legacy implementation, I will restore its comment.

You have a list where you can select many jobs, but the list only allows you to select one (because the implementation is not complete), so this bit test that your expectations are met, and you can only stop one job at a time. (although the nice Ux would be allowing it to do many).

plugins/jobs/src/js/JobStopRun.js
Fabs: I agree.

plugins/jobs/src/js/JobDelete.js
DanielMSchmidt: Oh okay, please verify. If it is like you say then it's totally fine

plugins/jobs/src/js/data/JobModel.ts
Fabs: I agree, and I would love to discuss that here. But changing it I believe would be out of the scope since I am following the API all the other resolvers use, and I think we should fix them all together. 

I think it is maybe a topic for the patterns we wanna have with the data layer.

plugins/jobs/src/js/components/JobDeleteModal.js
Fabs: I agree with you here and the HTML function. I did the `quick extraction ™`

plugins/jobs/src/js/components/JobStopRunModal.js
Fabs: Yes, now that is short enough that we can.

plugins/jobs/src/js/data/JobModel.ts
Fabs: I like that idea. But I think it is adding one more level of complexity to something that should not exist. Graphql should check the existence and type of my args, I should not have to add all those guards to the resolver.

I think here I would also make "the same mistake" everywhere and push forward to have our graphql implementation to care about it.

plugins/jobs/src/js/JobDelete.js
Fabs: fair point.

plugins/jobs/src/js/data/JobModel.ts
DanielMSchmidt: I think we missed that already and we are unaligned already: https://github.com/dcos/dcos-ui/blob/8429c574a2a99c7704127ee4029a90f69302df55/plugins/jobs/src/js/data/JobModel.ts#L99

plugins/jobs/src/js/data/JobModel.ts
DanielMSchmidt: Can you explain what you mean with "our graphql implementation to care about it"

Fabs: For Runtime:
The query this resolver resolves states: `runJob(id: String!)`. Which means that the grapqhl implementation should throw an error if `id` is not present. So at runtime, I expect not to have to write an if.

For type checking:
As for the typescript types, I understand and like the elegance of the user-defined type check like `isJobQueryArg()`, but on this particular case I believe it just hides a cast away from me.  

What I really want to right is:
`args: RunJobArgs`

but then IResolvers do not get happy. That I think is the type who needs a tweek to be able to support me to say `args: RunJobArgs`.

So overall I think the `if` here is also the simple solution to those both problems combined.

plugins/jobs/src/js/JobDelete.js
nLight: It's hard to understand what's going on from the name. Will it error out? Will it detect error? Shall it be `errorIfTaskCurrentRunning` or `errorIfCurrentTaskIsRunning`?

plugins/jobs/src/js/data/JobModel.ts
Fabs: I have no real preference between the `String` and the `response.mesage.String`. if you have a preference I would use it.

plugins/jobs/src/js/data/JobModel.ts
Fabs: does that argument, solve this issue?

plugins/jobs/src/js/JobDelete.js
Fabs: What about: isTaskCurrentRunning?

plugins/jobs/src/js/JobDelete.js
Fabs: It is only called once.

Fabs:
@nLight and @DanielMSchmidt . I fixed it, but I am not setting it to review again until the other dependencies are in.

plugins/jobs/src/js/JobDelete.js
nLight: Is `TaskCurrent` an entity? Why then not `CurrentTask`?

plugins/jobs/src/js/JobsOverview.tsx
```diff
@@ -158,7 +158,7 @@ const JobsOverview = withRouter(
           return graphqlObservable(jobsOverviewQuery, JobModel, {
             path,
             filter
-          }) as Observable<{ data: { jobs: JobConnection } }>;
```
Fabs: Should vanish once we get this fix on another PR to master.

nLight:
Flicts!

nLight:
All deps are in, please resolve the flicts!

plugins/jobs/src/js/JobDelete.js
```diff
@@ -0,0 +1,105 @@
...
+        jobId,
+        stopCurrentJobRuns,
+        onSuccess,
+        errorMessage
+      });
+    }
+  };
+}
+
+function isTaskCurrentRunning(errorMessage) {
```
nLight: I hate to get back to this but Task preceding Current still doesn't make any sense to me. Am I missing something?

Fabs: I think what is wrong here is that it should be written "currently". Does that work for you? `isTaskCurrentlyRunning`

nLight: that would make total sense 🙇

plugins/jobs/src/js/JobDelete.js
```diff
@@ -0,0 +1,105 @@
...
+  return /stopCurrentJobRuns=true/.test(errorMessage);
```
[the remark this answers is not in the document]
juliangieseke: haha, was going to comment, but thoght its not important enough :D

Fabs:
That PR is dead, CI wont build it anymore :/

Fabs:
I moved to a new PR this seems not to build.

PR #3044, opened 2018-07-02
DCOS-38709 change linearBackoff interface

juliangieseke:
follow-up of #3041

### DIFF
--- a/src/js/utils/__tests__/rxjsUtils-test.js
+++ b/src/js/utils/__tests__/rxjsUtils-test.js
@@ -1,4 +1,5 @@
 import "rxjs/add/operator/retryWhen";
+import "rxjs/add/operator/take";
 import { marbles } from "rxjs-marbles/jest";
 
 import { linearBackoff } from "../rxjsUtils";
@@ -19,9 +20,22 @@ describe("linearBackoff", function() {
 
       // In test env we don't want to wait for the real wall clock
       // so we encode time intervals with a special helper `m.time`
-      const result = source.retryWhen(linearBackoff(2, m.time("--|")));
+      const result = source.retryWhen(linearBackoff(m.time("--|"), 2));
 
       m.expect(result).toBeObservable(expected);
+    })
+  );
+
+  it(
+    "retries infinitey when no maxRetries given",
+    marbles(function(m) {
+      m.bind();
+
+      const source = m.cold("1--2#");
+      const expected = m.cold("1--2--1--2----1--2------1--(2|)");
+
+      const result = source.retryWhen(linearBackoff(m.time("--|")));
+      m.expect(result.take(8)).toBeObservable(expected);
     })
   );
 
@@ -33,7 +47,7 @@ describe("linearBackoff", function() {
       const source = m.cold("1--2#");
       const expected = m.cold("1--2--1--2----1--2------1--2#");
 
-      const result = source.retryWhen(linearBackoff(3, m.time("--|")));
+      const result = source.retryWhen(linearBackoff(m.time("--|"), 3));
       m.expect(result).toBeObservable(expected);
     })
   );
@@ -47,7 +61,7 @@ describe("linearBackoff", function() {
       const expected = m.cold("1--2--1--2----1--2----1--2----1--2#");
 
       const result = source.retryWhen(
-        linearBackoff(4, m.time("--|"), m.time("----|"))
+        linearBackoff(m.time("--|"), 4, m.time("----|"))
       );
       m.expect(result).toBeObservable(expected);
     })
--- a/src/js/utils/rxjsUtils.js
+++ b/src/js/utils/rxjsUtils.js
@@ -4,22 +4,30 @@ import "rxjs/add/operator/delayWhen";
 import "rxjs/add/operator/scan";
 
 /* eslint-disable import/prefer-default-export */
-export function linearBackoff(
-  maxRetries = 5,
-  delay = 1000,
-  maxDelay = Number.MAX_SAFE_INTEGER,
-  scheduler = null
-) {
+/**
+ * linearBackoff callback to use with e.g. delayWhen
+ *
+ * @param  {number} delay
+ * @param  {number} [maxRetries=-1]
+ * @param  {number} [maxDelay=Number.MAX_SAFE_INTEGER]
+ * @param  {Scheduler} [scheduler=null]
+ * @return {Observable} notifier obervable
+ */
+export function linearBackoff(delay, maxRetries = -1, maxDelay, scheduler) {
   return errors =>
     errors
       .scan((count, error) => {
-        if (count >= maxRetries) {
+        if (maxRetries >= 0 && count >= maxRetries) {
           throw error;
         }
 
         return count + 1;
       }, 0)
       .delayWhen(val =>
-        timer(Math.min(val * delay, maxDelay), null, scheduler)
+        timer(
+          Math.min(val * delay, maxDelay || Number.MAX_SAFE_INTEGER),
+          undefined,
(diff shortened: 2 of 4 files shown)


src/js/utils/__tests__/rxjsUtils-test.js
```diff
@@ -19,12 +19,29 @@ describe("linearBackoff", function() {
...
+      m.bind();
+
+      const source = m.cold("1--2#");
+      // const expected = m.cold("1--2--1--2----1--2------1--2");
+      // const timer = m.cold("1--2--1--2----1--2------1--2|");
+
+      // const result = source.retryWhen(linearBackoff(m.time("--|"), -1));
+      // m.expect(result.take(8)).toBeObservable(expected);
+      expect(() =>
+        source.retryWhen(linearBackoff(m.time("--|"), -1))
```
juliangieseke: not sure if this works as exptected, the commented out part above didnt (`take` is undefined on marbles `ColdObservable`) - will check if I can implement this without marbels.

src/js/utils/rxjsUtils.js
```diff
@@ -2,18 +2,29 @@ import { timer } from "rxjs/observable/timer";
...
+ * @param  {number} maxRetries
+ * @param  {number} [maxDelay=Number.MAX_SAFE_INTEGER]
+ * @param  {Scheduler} [scheduler=null]
+ * @return {Observable} notifier obervable
+ */
 export function linearBackoff(
-  maxRetries = 5,
-  delay = 1000,
+  delay,
+  maxRetries,
```
juliangieseke: I made this argument required, which is the same behavior as in `retry()`

Fabs: I think retry is actually defaulting to -1.

```
export function retry<T>(count: number = -1): MonoTypeOperatorFunction<T> {
  return (source: Observable<T>) => source.lift(new RetryOperator(count, source));
}
```
https://github.com/ReactiveX/rxjs/blob/53b51f35689278e05e400506dbc7388a6df5168e/src/internal/operators/retry.ts

src/js/utils/rxjsUtils.js
```diff
@@ -2,18 +2,29 @@ import { timer } from "rxjs/observable/timer";
...
   maxDelay = Number.MAX_SAFE_INTEGER,
   scheduler = null
 ) {
   return errors =>
     errors
       .scan((count, error) => {
-        if (count >= maxRetries) {
+        if (maxRetries >= 0 && count >= maxRetries) {
```
Fabs: I would rather not increase count at all on this case by `return 0` everytime maxRetries < 0.

src/js/utils/rxjsUtils.js
juliangieseke: narf… was just looking at the documentation, not in the code…

src/js/utils/rxjsUtils.js
juliangieseke: but then there is no increase in backoff time? 🤔

Fabs: Yes, you are absolutely right. 

If you touch that again, can you change the name of `val`  to count on `delayWhen`. That would have helped me not make this mistake :D

src/js/utils/rxjsUtils.js
```diff
@@ -2,18 +2,29 @@ import { timer } from "rxjs/observable/timer";
 
 import "rxjs/add/operator/delayWhen";
 import "rxjs/add/operator/scan";
+// import { Observable } from "rxjs/Observable";
```
nLight: 👮

src/js/utils/rxjsUtils.js
```diff
@@ -2,18 +2,29 @@ import { timer } from "rxjs/observable/timer";
...
+// import { Scheduler } from "rxjs/Scheduler";
 
 /* eslint-disable import/prefer-default-export */
+/**
+ * linearBackoff callback to use with e.g. delayWhen
+ *
+ * @param  {number} delay
+ * @param  {number} maxRetries
```
nLight: please specify default value

src/js/utils/__tests__/rxjsUtils-test.js
```diff
@@ -19,12 +20,25 @@ describe("linearBackoff", function() {
...
     })
   );
 
+  it(
+    "retries infinitey when no maxRetries given",
+    marbles(function(m) {
+      m.bind();
+
+      const source = m.cold("1--2#");
+      const expected = m.cold("1--2--1--2----1--2------1--(2|)");
```
nLight: but that is 3 retries, not 8, no?

juliangieseke: 8 events, not retries 👍

juliangieseke:
@Fabs @nLight comments are addressed, please revisit 🙇

src/js/utils/rxjsUtils.ts
```diff
@@ -0,0 +1,30 @@
...
+        if (maxRetries >= 0 && count >= maxRetries) {
+          throw error;
+        }
+
+        return count + 1;
+      }, 0)
+      .delayWhen(val =>
+        timer(
+          Math.min(val * delay, maxDelay || Number.MAX_SAFE_INTEGER),
+          undefined,
```
nLight: unusual seeing `undefined` to be passed to skip the default  value 🤷‍♂️

src/js/utils/rxjsUtils.ts
```diff
@@ -0,0 +1,30 @@
...
+import "rxjs/add/operator/delayWhen";
+import "rxjs/add/operator/scan";
+import { Scheduler } from "rxjs/Scheduler";
+import { Observable } from "rxjs/Observable";
+
+export function linearBackoff(
+  delay: number,
+  maxRetries: number = -1,
+  maxDelay?: number,
+  scheduler?: Scheduler
```
nLight: Do we want to have `IScheduler` here instead? 🤔

src/js/utils/rxjsUtils.ts
nLight: In RxJs they use `private period: number = -1;` as default which could be a better value as it is more like a pattern https://github.com/ReactiveX/rxjs/blob/b5ed59d8626376d16ac09aa98ac9c00835340fe3/src/observable/TimerObservable.ts#L82

src/js/utils/__tests__/rxjsUtils-test.js
nLight: Sorry I still don't understand the math behind

src/js/utils/rxjsUtils.ts
nLight: It will be perfectly fine, though underlying implementation expects `IScheduler`. Having `Scheduler` is just more strict where we want it to be more permissive.

src/js/utils/rxjsUtils.ts
nLight: ok!

src/js/utils/rxjsUtils.ts
juliangieseke: Ok, changed the type 👍

juliangieseke:
@nLight I hope I fixed all of your comments - looks like github UI had a hiccup here, can you please revisit?

src/js/utils/__tests__/rxjsUtils-test.js
nLight: Oh true! I'm sorry :)

juliangieseke:
Yay \o/
@DanielMSchmidt I have forcepushed (`Scheduler` -> `IScheduler`), can you revisit please? 🙇

PR #3037, opened 2018-06-28
Add Docker Parameter Support

kensipe:
https://jira.mesosphere.com/browse/DCOS_OSS-3687
Support for Docker Parameters

gif/image

```
<!-- Suplemental codeblock -->
```

**Checklist**
- [ ] Did you add a JIRA issue in a commit message or as part of the branch name?
- [ ] Did you add new unit tests?
- [ ] Did you add new integration tests?
- [ ] If this is a regression, did you write a test to catch this in the future?

PR #3047, opened 2018-07-03
[1.9] feat: support docker parameters

weblancaster:
Support docker parameters.
This is based on @kensipe PR https://github.com/dcos/dcos-ui/pull/3037

*Testing:*
To really test this you need a special cluster. Maybe someone from the Marathon team will be able to help. In general we should verify if the shape of the JSON the form produces is in line with JIRA and switching from JSON and the Form back and forth doesn't mess up JSON.

Closes DCOS_OSS-3687

**Checklist**

### DIFF
--- a/src/js/schemas/JobSchema.js
+++ b/src/js/schemas/JobSchema.js
@@ -6,6 +6,7 @@ import Docker from "./job-schema/Docker";
 import General from "./job-schema/General";
 import Labels from "./job-schema/Labels";
 import Schedule from "./job-schema/Schedule";
+import Parameters from "./job-schema/Parameters";
 
 const JobSchema = {
   type: "object",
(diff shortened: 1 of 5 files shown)


Fabs:
@weblancaster, @nLight. I know this needs to move faster, but I think since it is even hard to get a cluster to test this, we should have some integration or even better system level testing to verify this feature.

weblancaster:
@Fabs I would leave the system test for the refactoring of this feat.

juliangieseke:
adding hold merge label, see slack #frontend for more info 🤔

weblancaster:
@juliangieseke maybe we should close, what you think?

nLight:
@weblancaster wait for it, I think we are merging it

Fabs:
I was given a cluster by @kensipe. The UI is not losing parameters anymore, and the form seems to work for me.

PR #3048, opened 2018-07-04
DCOS-38339: Address Memory Issues

orlandohohmeier:
Update the mesos-client to 0.2.3 as this version great reduces the memory footprint and adjust `MesosStream` to stop multicasting the stream using the replay subject. This change allows the browser to garbage collect dependent/sliced strings created while processing the Mesos response.

_Please note that this is a temporary adjustment as it would result in multiple open connections if consumed by more than one component. We're going to revert this change once the `http-service`  to emit event objects instead of the response as well as changed the `mesos-client` to output parsed JSON._

**Testing**

* Checkout the code and verify that the number of open mesos subscribe connections is correct and the updates are flowing through. 
* Make memory snapshots in Firefox and Chrome and observer the Safari memory footprint while looking at a task page.

Closes DCOS-38339

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -52,9 +52,9 @@
       "resolved": "https://registry.npmjs.org/@dcos/connections/-/connections-0.1.0.tgz"
     },
     "@dcos/copychars": {
-      "version": "0.1.0",
-      "from": "@dcos/copychars@0.1.0",
-      "resolved": "https://registry.npmjs.org/@dcos/copychars/-/copychars-0.1.0.tgz"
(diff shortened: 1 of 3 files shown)


DanielMSchmidt:
@orlandohohmeier test fails are legit, please take a look :)

nLight:
Going to check if those marathon fails 
screen shot 2018-07-05 at 14 12 39

cause Region and zone to be N/A
screen shot 2018-07-05 at 14 12 58

nLight:
And it's green 💚 @Fabs @DanielMSchmidt

PR #3052, opened 2018-07-06
DCOS-37617: Jobs / refactor: move logic from the mediator to graphql

Fabs:
Ressurect #3043 because CI wont build it anymore

### DIFF
--- a/plugins/jobs/src/js/data/JobModel.ts
+++ b/plugins/jobs/src/js/data/JobModel.ts
@@ -5,7 +5,10 @@ import {
   fetchJobs,
   fetchJobDetail,
   runJob,
+  updateSchedule,
   updateJob,
+  deleteJob,
+  stopJobRun,
   JobResponse as MetronomeJobResponse,
   JobDetailResponse as MetronomeJobDetailResponse
 } from "#SRC/js/events/MetronomeClient";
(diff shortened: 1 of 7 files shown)


Fabs:
@juliangieseke. You are not able to stop more than one job by design. I did not dig deep to find the reason, but I believe it is due the way the api works, so if you like to stop 10 jobs, you would be issuing 10 request. Anyway, that was always the case before the Data-Layer.

Also, on #3043 there is a long argument on why I did not do the typeguards.

juliangieseke:
I did not try to stop multiple jobs, just one… 🤔 

re Type Guards: I disagree on both of your arguments, it doesnt add a level of complexity - the opposite is true, it _reduces_ complexity (but increases _abstraction_). Also I dont think that `if (args.id)` is any better then `if (isSometype(args))`. 🤔

plugins/jobs/src/js/data/JobModel.ts
```diff
@@ -141,7 +156,31 @@ export const resolvers = ({
       args: GeneralArgs,
       _context = {}
     ): Observable<JobLink> {
-      return runJob(args.id).map(({ jobId }) => ({ jobId }));
+      if (args.id) {
+        return runJob(args.id).map(({ jobId }) => ({ jobId }));
```
nLight: Generally I'd prefer an early return under test condition and not the main logic under a condition.

nLight:
Confirmed, it's impossible to stop a job, the button does nothing

nLight:
Works like a charm on master tho

Fabs:
@juliangieseke. Ok will take a look on the stop not working.

Fabs:
@nLight You are right, I reversed the other of the test to error in the oposite condition.

@juliangieseke left the if's instead of the type guards as we discussed.

Fabs:
@nLight and @juliangieseke also the stop modal is working now.

Fabs:
@juliangieseke there is a very old ticket linked to that, related to support stoping multiple jobs. And I also think that the fix here should come on the new jobs and queues. So I am hesitant to propose using radios as the solution here.

PR #3053, opened 2018-07-06
DCOS-38939: Branding / Add Empty Header Bar

juliangieseke:
**DESCRIPTION**

This PR prepares UI for the HeaderBar (1. commit) and adds the bar itself (2. commit)

---

**HOW TO REVIEW**

`npm install` because ui-kit was updated, then make sure everything works as expected. I changed styles for sidebar and content, so please verify that sidebar collapses, expands, sticks and unsticks correctly. also check different browsers,  windows sizes and some modals please :)

---

**KNOWN ISSUES**

Well, Headerbar only contains foobar text… 

---

**HOW TO ENABLE**

In `src/js/components/HeaderBar.tsx`:
```diff
 export default function HeaderBar() {
-  //remove this to activate component
-  if (arguments) {
-    return null;
-  }
-
   return (
```

### DIFF
--- a/packages/ui-kit-stage/HeaderBar.tsx
+++ b/packages/ui-kit-stage/HeaderBar.tsx
@@ -1,0 +1,23 @@
+/**
+ * Component: HeaderBar
+ * Added: 2018-07-08
+ * JIRA: https://jira.mesosphere.com/browse/DCOS-39074
+ */
+import * as React from "react";
+import { coreColors } from "@dcos/ui-kit/dist/packages/shared/styles/color";
+
+const headerBarStyles: object = {
+  backgroundColor: coreColors().purpleDarken4,
+  height: "32px",
+  width: "100%",
+  flexShrink: 1,
+  flexGrow: 0,
+  flexBasis: "auto",
+  zIndex: 999
+};
+
+export class HeaderBar extends React.PureComponent {
+  render() {
+    return <div style={headerBarStyles}>{this.props.children}</div>;
+  }
+}
--- a/src/js/components/HeaderBar.tsx
+++ b/src/js/components/HeaderBar.tsx
@@ -1,0 +1,18 @@
+import * as React from "react";
+import { HeaderBar as UIHeaderBar } from "ui-kit-stage/HeaderBar";
+
+export default function HeaderBar() {
+  // remove this to activate component
+  if (arguments) {
+    return null;
+  }
+
+  return (
+    <UIHeaderBar>
+      <span>toggle component here</span>
+      <span>logo component here</span>
+      <span>user-menu component here</span>
+      <span>cluster-menu component here</span>
+    </UIHeaderBar>
+  );
+}
--- a/src/js/pages/Index.js
+++ b/src/js/pages/Index.js
@@ -13,6 +13,7 @@ import MesosStateStore from "../stores/MesosStateStore";
 import Modals from "../components/Modals";
 import RequestErrorMsg from "../components/RequestErrorMsg";
 import ServerErrorModal from "../components/ServerErrorModal";
+import HeaderBar from "../components/HeaderBar";
 import Sidebar from "../components/Sidebar";
 import SidebarActions from "../events/SidebarActions";
 import SidebarStore from "../stores/SidebarStore";
@@ -151,6 +152,7 @@ var Index = React.createClass({
 
     return (
       <div className={classSet}>
+        <HeaderBar />
         <div className="application-wrapper-inner">
           {this.getScreenOverlays(showErrorScreen)}
           <Sidebar location={this.props.location} />
--- a/src/js/pages/Index.js
+++ b/src/js/pages/Index.js
@@ -151,14 +151,16 @@ var Index = React.createClass({
 
     return (
       <div className={classSet}>
-        {this.getScreenOverlays(showErrorScreen)}
-        <Sidebar location={this.props.location} />
-        {this.props.children}
-        <Modals
-          showErrorModal={this.state.showErrorModal}
-          modalErrorMsg={this.state.modalErrorMsg}
-        />
-        <ServerErrorModal />
+        <div className="application-wrapper-inner">
+          {this.getScreenOverlays(showErrorScreen)}
+          <Sidebar location={this.props.location} />
+          {this.props.children}
+          <Modals
+            showErrorModal={this.state.showErrorModal}
+            modalErrorMsg={this.state.modalErrorMsg}
+          />
+          <ServerErrorModal />
+        </div>
       </div>
     );
   }
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -14,19 +14,19 @@
       }
     },
     "@babel/helper-module-imports": {
-      "version": "7.0.0-beta.32",
-      "resolved": "https://registry.npmjs.org/@babel/helper-module-imports/-/helper-module-imports-7.0.0-beta.32.tgz",
-      "integrity": "sha512-9jxfqCBrwCIa0p5ZIy1sakzKKm8x8tn0C52qpPr0M0WJ/k9gpD4ilS/mTV2v0tgmw4agjYdUXQ8slq51/5oOzQ==",
+      "version": "7.0.0-beta.51",
+      "resolved": "https://registry.npmjs.org/@babel/helper-module-imports/-/helper-module-imports-7.0.0-beta.51.tgz",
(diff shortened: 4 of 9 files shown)


DanielMSchmidt:
@juliangieseke Snapshot tests are failing, please take a look

src/js/components/HeaderBar.tsx
```diff
@@ -0,0 +1,18 @@
+import * as React from "react";
+import { HeaderBar as UIHeaderBar } from "#PLUGINS/ui-kit-candidates/HeaderBar";
```
DanielMSchmidt: I am unsure about the ui-kit-candidates approach here. I have these problems:

- For me, it communicates that all other components will never be moved to the ui-kit
- We now have two positions to look for ui components. We had the same at my last company and it leed to a "starvation" of the right place, because every component was "just" moved to the easier to integrate place with a light polish.

plugins/ui-kit-candidates/HeaderBar.tsx
```diff
@@ -0,0 +1,19 @@
...
+  flexShrink: 1,
+  flexGrow: 0,
+  flexBasis: "auto",
+  zIndex: 999
+};
+
+export class HeaderBar extends React.PureComponent {
+  render() {
+    const { children } = this.props;
+    return <div style={headerBarStyles}>{children}</div>;
```
DanielMSchmidt: 🎨 this is one of the rare cases where I like `this.props.children` better 😆

juliangieseke:
@DanielMSchmidt  yup, ui kit updated changes classnames - pushing a fix as soon as linting passes 👍

plugins/ui-kit-candidates/HeaderBar.tsx
```diff
@@ -0,0 +1,19 @@
...
+import { coreColors } from "@dcos/ui-kit/dist/packages/shared/styles/color";
+
+const headerBarStyles: object = {
+  backgroundColor: coreColors().purpleDarken4,
+  height: "32px",
+  width: "100%",
+  flexShrink: 1,
+  flexGrow: 0,
+  flexBasis: "auto",
+  zIndex: 999
```
DanielMSchmidt: Would it make sense to put this into a constant, so we have named levels?

src/js/pages/Index.js
```diff
@@ -150,15 +151,32 @@ var Index = React.createClass({
...
-        <Sidebar location={this.props.location} />
-        {this.props.children}
-        <Modals
-          showErrorModal={this.state.showErrorModal}
-          modalErrorMsg={this.state.modalErrorMsg}
-        />
-        <ServerErrorModal />
+      <div
+        className={classSet}
+        style={{
```
DanielMSchmidt: Why do we need to use inline styles here?

plugins/ui-kit-candidates/HeaderBar.tsx
juliangieseke: We could introduce something like that in UI Kit, yeah. (not in scope for this PR I'd say)

plugins/ui-kit-candidates/HeaderBar.tsx
juliangieseke: agree 😆 there was more code before which I managed to strip out - that last refactoring slipped. I'll change it if I touch the file again.

src/js/components/HeaderBar.tsx
juliangieseke: you are correct and I discussed it with philipp, we need to be careful with this directory and have to maintain it properly. the Idea was that we move everything that sits in there within a short period of time to ui-kit (or not). 

Do you want me to change something or create issues regarding the move?

src/js/pages/Index.js
juliangieseke: because I had no idea how to deal with CNVS. styling there is such a mess and as we want to move to emotion with colocated styles anyway, I felt that having it as inline styles isnt _that_ bad. 🤷‍♂️ 

is that something you want me to change? or just a question?

src/js/components/HeaderBar.tsx
DanielMSchmidt: Can you explain to me why we can not add it to the `ui-kit`?

plugins/ui-kit-candidates/HeaderBar.tsx
DanielMSchmidt: Hmm, having UI Kit here is a problem. Ok nevermind

src/js/pages/Index.js
nLight: if it's a candidate for the ui-kit, shall we extract the styles the ui-kit way?

nLight:
What about the background color? I thought it's gonna be 💜 not black.

src/js/pages/Index.js
DanielMSchmidt: Isn't this the index page? Why is it a ui-kit candidate?

src/js/components/HeaderBar.tsx
juliangieseke: @nLight oh - gotcha! I'll fix that.
@DanielMSchmidt we can, and we should, but I wanted to move forward and optimize later. (same thing as in the tables PR - I'll create a Ticket)

plugins/ui-kit-candidates/HeaderBar.tsx
juliangieseke: I dont think it is a problem actually, UI Kit can efine levels for certain types of components.

juliangieseke:
After discussion with @DanielMSchmidt and @Fabs , I changed `ui-kit-candidates` to `ui-kit-stage` and created a ticket moving the component to UI Kit 👍 (Also I added a small Readme to the folder)
Also addressed the comments /cc @nLight

nLight:
@juliangieseke there was no `npm install` step described 🤷‍♂️

src/js/pages/Index.js
DanielMSchmidt: I would really like the styles being moved to CSS

src/js/components/Sidebar.js
```diff
@@ -351,7 +351,10 @@ class Sidebar extends React.Component {
         >
           {overlay}
         </CSSTransitionGroup>
-        <div className="sidebar flex flex-direction-top-to-bottom">
+        <div
+          className="sidebar flex flex-direction-top-to-bottom"
+          style={{ position: "relative" }}
```
nLight: Can't we just add `position: relative` to the `sidebar` class?

src/js/pages/Index.js
nLight: Please extract inline styles to `.application-wrapper`.

src/js/pages/Index.js
```diff
@@ -150,15 +151,32 @@ var Index = React.createClass({
...
+          display: "flex",
+          flexFlow: "column",
+          height: "100%"
+        }}
+      >
+        <HeaderBar />
+        <div
+          style={{
```
nLight: 👮 no inline styles

packages/ui-kit-stage/HeaderBar.tsx
```diff
@@ -0,0 +1,23 @@
+/**
+ * Component: HeaderBar
+ * Added: 2018-07-08
+ * JIRA: https://jira.mesosphere.com/browse/DCOS-39074
+ */
+import * as React from "react";
+import { coreColors } from "@dcos/ui-kit/dist/packages/shared/styles/color";
+
+const headerBarStyles: object = {
```
nLight: if it's a UI Kit candidate why not using emotion here? Example: https://github.com/dcos-labs/ui-kit/blob/master/packages/badge/components/badge.tsx

juliangieseke: IMHO that would be part of the ui-kit issue, there is no emotion in dcos-ui yet 🤷‍♂️ If I finalize the component here, then there is no need for this staging folder thing… the Idea was to move on fast and optimize later instead of blocking this whole PR until that component is finalized 😕

Fabs:
I took a swing at running it, but I will wait the next round of changes so I check the code.

juliangieseke:
@nLight @DanielMSchmidt moved styles into css files 👍

juliangieseke:
@Fabs ping :)

packages/ui-kit-stage/HeaderBar.tsx
weblancaster: @juliangieseke @nLight I would not suggest adding this in the UI Kit. This case is more structural than functional/composable.

it's just returning a div with styles. what should be happening is the components inside this div are part of the UI Kit, not the div wrapper part of the UI Kit.

PR #3055, opened 2018-07-06
DCOS-21723: use metadata instead of calculation

nLight:
**TL;DR** This PR removes all the resources aggregation logic and replaces it with the `used_resources` for Frameworks.

**Tell me more**
Looks like we've mistakenly assumed that allocated resources for Frameworks should be aggregated from their tasks as with Marathon Services. Turned out that there's a special metadata on the Framework protobuf that has `used_resources` information.

**Checklist**

Closes DCOS-21723

### DIFF
--- a/plugins/services/src/js/containers/services/ServicesContainer.js
+++ b/plugins/services/src/js/containers/services/ServicesContainer.js
@@ -2,7 +2,7 @@ import PropTypes from "prop-types";
 import React from "react";
 import { routerShape } from "react-router";
 
-import { DCOS_CHANGE, MESOS_STATE_CHANGE } from "#SRC/js/constants/EventTypes";
+import { DCOS_CHANGE } from "#SRC/js/constants/EventTypes";
 import { reconstructPathFromRoutes } from "#SRC/js/utils/RouterUtil";
 import AppDispatcher from "#SRC/js/events/AppDispatcher";
 import ContainerUtil from "#SRC/js/utils/ContainerUtil";
@@ -11,7 +11,6 @@ import DSLExpression from "#SRC/js/structs/DSLExpression";
 import DSLFilterList from "#SRC/js/structs/DSLFilterList";
 import Icon from "#SRC/js/components/Icon";
 import Loader from "#SRC/js/components/Loader";
-import MesosStateStore from "#SRC/js/stores/MesosStateStore";
 import Page from "#SRC/js/components/Page";
 import RequestErrorMsg from "#SRC/js/components/RequestErrorMsg";
 import {
@@ -143,10 +142,6 @@ class ServicesContainer extends React.Component {
 
   componentDidMount() {
     DCOSStore.addChangeListener(DCOS_CHANGE, this.onStoreChange);
-
-    // Don't block the whole screen if Mesos state is not there
-    // Mesos state is needed to aggregate frameworks resources correctly
-    MesosStateStore.addChangeListener(MESOS_STATE_CHANGE, function() {});
 
     // Listen for server actions so we can update state immediately
     // on the completion of an API request.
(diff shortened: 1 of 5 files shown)


nLight:
Anecdotally I've fixed the issue once already https://github.com/dcos/dcos-ui/commit/49b2308bcabec8e79b0ecced4c38e5118018970b :D

nLight:
Checking if the solution is applicable for 1.11, looks like it is not

plugins/services/src/js/containers/services/ServicesContainer.js
```diff
@@ -144,10 +143,6 @@ class ServicesContainer extends React.Component {
   componentDidMount() {
     DCOSStore.addChangeListener(DCOS_CHANGE, this.onStoreChange);
 
-    // Don't block the whole screen if Mesos state is not there
```
jaxesn: This seems unrelated?

plugins/services/src/js/structs/Framework.js
```diff
@@ -93,45 +93,10 @@ module.exports = class Framework extends Application {
...
-        };
-      }, frameworkResources);
-  }
-
-  getUsedResources() {
+    // There's an unfortunate naming issue in Mesos.
+    // used_resources is actually allocated resources
+    // the name can't be changed to keep backward compatibility.
+    // This is why `getResources` returns `used_resources`
     return (
```
jaxesn: Well thats easier...

src/js/stores/DCOSStore.js
```diff
@@ -440,7 +440,7 @@ class DCOSStore extends EventEmitter {
       .getServiceList()
       .reduceItems(function(memo, framework) {
         if (framework instanceof Item) {
-          memo[framework.get("name")] = framework.get();
+          memo[`/${framework.get("name")}`] = framework.get();
```
jaxesn: Is this just a cleanup, or was the `/` missing and required somewhere else?

nLight: Have a look at the change down below. `/` was required to make them equal

plugins/services/src/js/containers/services/ServicesContainer.js
nLight: I looked back to the commit where I added that `getResources` method back then. There I added this event listener. It was required since we haven't had stream and tended to use `MesosStateStore` to calculate resources as use can see from the method I deleted

nLight:
@fabs that's funny. Those aren't frameworks are they? You can always compare with `/mesos` UI

Fabs:
@nLight I think they are the simplest framework, the hello-world. I deployed them from the catalog, and they have 3 tasks, a hello-world (which I suppose is the framework) and one `hello` and one `world`.  I left them running in the SF cluster.

Fabs:
Hum, as a starter, mesos know all tasks:
image

Fabs:
Yeah, your values match what mesos know.

image

Fabs:
@orlandohohmeier @nLight I think this will need a communication to our users that will after update just see that the values they are used to see changes. I think that should be very well explained in CHANGELOG.

nLight:
Making sense of data: 
services table show allocated resources.
Task detail page shows used resources
allocated resources are the resources that Mesos gave to the framework. Framework can use less resources that has been allocated. In this scenario 2 of 3 tasks use 25 Mb disk which sums up to 50 meg. They also use less RAM which is ok because Task

nLight:
We could add a tooltip to the Services Table indicating that resources are allocated resources

Fabs:
@nLight hum. I think that could help. But I also think it is something that needs more direct comunication. like:

BREAKING CHANGES - The ui change this way to count resources, to this way to count resources, but only on this particular area. Here is the blogs.dcos.happyusers/how-the-dcos-ui-count-resources. 

And then have this amazing text explaining our cluster operators what changed.

nLight:
I mean, that would be rather short amazing embarrassing text 😆 Like we used to sum all the tasks resources now we just show what Mesos gives us :D

nLight:
https://jira.mesosphere.com/browse/DCOS-39095

PR #3057, opened 2018-07-09
Rewrite pull request template

DanielMSchmidt:
This is our draft to a new PR template, please take a look 👍 

Please also do not merge after 2 approvals so everybody can take a look.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,19 +1,19 @@
-<!--- Thank you for your contribution! Please provide enough information for others to best review your code. -->
+## Testing
 
-<!-- Prefer **small pull requests**. These are much easier to review and more likely to get merged. Make sure the PR does only one thing, otherwise please split it. -->
+<!--
+What is needed to test the changes? e.g. specific cluster, service definitions
+How can one see the result of your work? e.g. configurations, URLs
+-->
 
-<!-- Description of motivation for making this change, what does it solve and steps needed to see change. -->
+## Trade-offs
 
-![gif/image]()
+<!--
+Are you aware of any weak spots? e.g. performance, functionality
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,19 +1,19 @@
-<!--- Thank you for your contribution! Please provide enough information for others to best review your code. -->
+## Testing
```
Fabs: I am wondering how do I know what does this PR accomplishes? 

One could say that this is in the ticket, but often times the ticket states the approval criteria, but sometimes not. I think we might want something here that asserts that the acceptance criteria is somewhere, or to have the acceptance criteria on the PR

DanielMSchmidt: We discussed that and came to the conclusion that there are two sides of defining what you accomplish in this PR. One is the **what** has changed in the UI and this would go into the testing section, e.g. try to use the jobs overview page and see if the filtering works or look into the performance profiler and verify that there is no more memory leak. 
The other is **why** it has changed and for this you should take a look into the Jira issue. If it is not in a shape where you would feel someone outside may understand the context of the issue you could either bring it in this shape or if that is somehow not possible you can extend this template. It is just a starting point, not an extensive list of everything allowed to be written here :)

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,19 +1,19 @@
...
 
-<!-- Prefer **small pull requests**. These are much easier to review and more likely to get merged. Make sure the PR does only one thing, otherwise please split it. -->
+<!--
+What is needed to test the changes? e.g. specicfic cluster, service definitions
```
DanielMSchmidt: TYPO: specific

PR #3059, opened 2018-07-10
DCOS-39106: Show allocated resources for frameworks

nLight:
Backporting #3055 

If you testing with soak keep in mind that some data job may "blink" switching from all zeros to a value. Please verify with mesos UI that is an expected behavior.

Closes DCOS-39106

### DIFF
--- a/plugins/services/src/js/structs/Framework.js
+++ b/plugins/services/src/js/structs/Framework.js
@@ -93,41 +93,27 @@ module.exports = class Framework extends Application {
   }
 
   getResources() {
-    // TODO: Circular reference workaround DCOS_OSS-783
-    const MesosStateStore = require("#SRC/js/stores/MesosStateStore");
-
-    const tasks = MesosStateStore.getTasksByService(this) || [];
-
-    const instances = this.getInstancesCount();
-    const {
-      cpus = 0,
-      mem = 0,
-      gpus = 0,
-      disk = 0
-    } = this.getSpec().getResources();
-
-    const frameworkResources = {
-      cpus: cpus * instances,
-      mem: mem * instances,
-      gpus: gpus * instances,
-      disk: disk * instances
+    // There's an unfortunate naming issue in Mesos.
+    // used_resources is actually allocated resources
+    // the name can't be changed to keep backward compatibility.
+    // This is why `getResources` returns `used_resources`
+    const allocatedFrameworkResources = this.get("used_resources") || {
+      cpus: 0,
+      mem: 0,
+      gpus: 0,
+      disk: 0
     };
 
-    // Aggregate all the child tasks resources
-    // resources of child frameworks won't be aggregated
-    return tasks
-      .filter(function(task) {
-        return task.state === "TASK_RUNNING" && !task.isStartedByMarathon;
-      })
-      .reduce(function(memo, task) {
-        const { cpus = 0, mem = 0, gpus = 0, disk = 0 } = task.resources;
+    // Framework doesn't know how many resources its scheduler consumes.
+    // Scheduler is launched by Marathon not the Framework itself.
+    // So we should get this information separately from the Marathon spec
+    const schedulerResources = this.getSpec().getResources();
 
-        return {
-          cpus: memo.cpus + cpus,
-          mem: memo.mem + mem,
-          gpus: memo.gpus + gpus,
-          disk: memo.disk + disk
-        };
-      }, frameworkResources);
+    return {
+      cpus: allocatedFrameworkResources.cpus + schedulerResources.cpus,
+      mem: allocatedFrameworkResources.mem + schedulerResources.mem,
+      gpus: allocatedFrameworkResources.gpus + schedulerResources.gpus,
+      disk: allocatedFrameworkResources.disk + schedulerResources.disk
+    };
   }
 };
--- a/plugins/services/src/js/containers/services/ServicesContainer.js
+++ b/plugins/services/src/js/containers/services/ServicesContainer.js
@@ -2,7 +2,7 @@ import PropTypes from "prop-types";
 import React from "react";
 import { routerShape } from "react-router";
 
-import { DCOS_CHANGE, MESOS_STATE_CHANGE } from "#SRC/js/constants/EventTypes";
+import { DCOS_CHANGE } from "#SRC/js/constants/EventTypes";
 import { reconstructPathFromRoutes } from "#SRC/js/utils/RouterUtil";
 import AppDispatcher from "#SRC/js/events/AppDispatcher";
 import ContainerUtil from "#SRC/js/utils/ContainerUtil";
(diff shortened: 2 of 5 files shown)


plugins/services/src/js/structs/Framework.js
```diff
@@ -93,41 +93,13 @@ module.exports = class Framework extends Application {
...
-    // resources of child frameworks won't be aggregated
-    return tasks
-      .filter(function(task) {
-        return task.state === "TASK_RUNNING" && !task.isStartedByMarathon;
-      })
-      .reduce(function(memo, task) {
-        const { cpus = 0, mem = 0, gpus = 0, disk = 0 } = task.resources;
-
-        return {
-          cpus: memo.cpus + cpus,
```
DanielMSchmidt: Why exactly can we remove all this computation?

nLight: Mesos does it for us and in a proper way. Back then, when I was writing this function, I didn't know that mesos gives `used_resources` info on the framework level. Since we aggregate resources for Marathon it was the way I fixed the issue. Now this aggregation is not needed anymore.

DanielMSchmidt: Thank you for explaining this so thoroughly

Fabs:
I keep looking at the data, and hitting those WTF moments as me being an operator do not understand what is going on.

I deployed hello-world again for this cluster.

1) 
This is how the detailed view reports resources for each of the 3 tasks in order (1 hello-word framework, 1 world, 1 hello).

image

So the tasks report *allocated*

2)
Messos shows me something updating over time. For instance, the task world.
https://drive.google.com/file/d/1Vdik6v8xHpwkcsIVDbK-ssTbLHrSCQxR/view

3)
Mesos also tells me that the values we are showing, are the Allocated values for framework (and not used).

image

It looks to me that the source of confusion is between Offered, Allocated, and Used.

On Mesos UI
- A task, shows me Allocated and Used (real time).
- A Framework, shows me  Offered (all zeroes in my case) and shows what we show as Used, as Allocated. And it looks like it ignores its own launch task

nLight:
@Fabs why are you keep looking into the TaskTable?

nLight:
changes has been made to the _ServicesTable_ what you see in the mesos side bar is what we show in the ServicesTable

Fabs:
@nLight. I am looking on the Services Table, but the values there need to make sense with what I see on the TasksTable too. The values on ServicesTable align with what Mesos show on Framework UI, but I still can't make sense of the overall picture.

Fabs:
For instance, a question I would have is. Why the sum of the individual resources on the TaskTable, does not match what I see on the ServicesTable? is it supposed to work by design? If yes, how I as a user makes sense of that?

nLight:
@Fabs the scope of the issue to fix services table. Tasks can and do use generally _less_resources that has been allocated

nLight:
@Fabs Please go ahead and create a JIRA issue to add hints/tooltips etc across the UI to help operators making sense of those numbers.

Fabs:
@nLight thanks for trying to explain. I don't think the tooltip is the issue I am trying to address. I can't verify myself that the numbers shown are correct because I can't make sense of them. I think that given the information I have now, I am unable to review this PR. The core of my inability to verify lies in two observations.

1) I don't understand how Mesos account for resources, and hence I can't tell we are displaying the correct value for our users.
2) As regarding the values the Services page show, I find odd they don't match the Tables View, and I can't explain why they are different.

nLight:
@Fabs then I want to ask why have you approved the original PR and invoke the SHIP IT rule ;)

Fabs:
This time I just looked a little more carefully and have more questions. To my knowledge at that time the fix was ok. 

As regards to SHIP IT, note that I am still looking at it trying to address (1). Also have not blocked this PR because of my inability to help you.

nLight:
@Fabs I'd suggest asking Core on call or someone from the support team to help you making sense of how Mesos works and then sharing this knowledge with the team

nLight:
@juliangieseke @Fabs Mesos UI say cassandra uses 0.6
screen shot 2018-07-11 at 12 10 32

Fabs:
That is a nice idea @nLight, and I will find a way to figure that out. I believe though as the person implementing that fix, you should be the one able to provide an argument on why this fix works this way, and why this is the correct sum. So I am not taking on that responsibility in order to deliver this fix.

nLight:
@Fabs I'm trying to do so and it looks like I'm failing that's why I'm referring to a 3rd common believable person to resolve this

juliangieseke:
ServiceTable: (0.6 CPU)
screen shot 2018-07-11 at 12 12 46

TasksTable (0.5 + 1 = 1.5 CPU)
screen shot 2018-07-11 at 12 12 48

nLight:
screen shot 2018-07-11 at 12 14 32
@juliangieseke I wonder why Mesos UI shows 1 CPU task as completed

juliangieseke:
🤷‍♂️  - also they have the same inconsistency in their UI… nevertheless I'd say we should not merge this fix until we can show consistent data to the customer 😞

nLight:
@juliangieseke that would mean we should revert 1.11

nLight:
Plus it looks like we can't be consistent across the board. As DCOS runs on Marathon and Cassandra has been launched on Marathon so that TasksTable shows both: Cassandra that has been launched by marathon and all Tasks that Cassandra has launched. 

Bu services table shows frameworks, it doesn't account for the Marathon. I don't really think there's an easy way to resolve this

nLight:
So the 1 CPU part that Cassandra uses as a framework has been subtracted from Marathons resources. Cassandra itself has only 0.6 allocated to itself

nLight:
Does it makes sense to you @Fabs @juliangieseke

nLight:
DCOS UI can't be fully consistent with Mesos UI. Because in DCOS UI we're mixing two worlds Marathon which is a framework on Mesos itself and Mesos. 

To be consistent with Mesos UI we should remove Framework tasks that has ben launched by Marathon from the TasksTable view. Which doesn't make sense because then we're missing the Marathon part of the pie.

Adding resources consumed by Framework executors to the resources allocated for them to the ServicesTable doesn't make sense either because those are different things.

The confusion grows from this duality (Marathon resources / Mesos Resources), though as I hope everyone on the same page now, I say we merge this and then see what we can do to make those numbers as prominent as possible.

If it is still unclear why we have more tasks on DCOS UI than in Mesos UI thus more resources let's have a call and I'll try explaining again @Fabs @juliangieseke @DanielMSchmidt

juliangieseke:
I'm not talking about being consistent with Mesos UI, its about the two screenshots I shared above. 
I dont care about Mesos UI at all - but inside DC/OS UI, we should be consistent. 😞

juliangieseke:
> Adding resources consumed by Framework executors to the resources allocated for them to the ServicesTable doesn't make sense either because those are different things.
So _ServiceTable_ displays only the Framework itself, not its executors? _TasksTable_ shows executors but they are totally unrelated to its framework? This is weird 🤔

nLight:
> So ServiceTable displays only the Framework itself, not its executors
Yes, because executor is launched by Marathon

> TasksTable shows executors, but they are totally unrelated to their framework?
Executors that have been launched by marathon + Tasks that have been launched by executor that has been launched by Marathon 😆 

It actually makes sense unless you try calculating the numbers. We could split TasksTable view into two tables. Explicitly saying this is executor for you to kill via Marathon it and those are Tasks that you can't actually control via Marathon.

juliangieseke:
^ Like it, if its not possible yet, we have our task. 👍  We should check back with product/lee if the current state is ok to merge.

nLight:
Hold merge until we fix master and then we will have to decide:
- if it is a breaking change so that we don't whip it to 1.11
- it is not really a breaking change so that we want to backport it

PR #3061, opened 2018-07-11
DCOS-38943: Add Logo to Header Bar

tnguyen0:
Add the images and responsive styling for the Mesosphere logo on the header bar.
Show just the logo if the viewport is small. Otherwise, show the full logo with the label
"Mesosphere DC/OS".

Closes DCOS-38943

Browser
screen shot 2018-07-11 at 3 31 54 pm

Mobile (excuse bad crop)
screen shot 2018-07-11 at 3 32 31 pm

**Checklist**

### DIFF
--- a/src/js/components/HeaderBar.tsx
+++ b/src/js/components/HeaderBar.tsx
@@ -10,7 +10,9 @@ export default function HeaderBar() {
   return (
     <UIHeaderBar>
       <span>toggle component here</span>
-      <span>logo component here</span>
+      <div className="header-bar-logo-wrapper">
+        <div className="header-bar-logo" />
+      </div>
       <span>user-menu component here</span>
       <span>cluster-menu component here</span>
     </UIHeaderBar>
--- a/src/styles/components/header-bar/styles.less
+++ b/src/styles/components/header-bar/styles.less
@@ -1,0 +1,27 @@
+& when (@header-bar-enabled) {
+
+  .header-bar-logo-wrapper {
+    display: inline-block;
+    height: 16px;
+    margin-left: 10px;
+    margin-top: 10px;
+    overflow: hidden;
+    width: @full-logo-width;
+
+    .header-bar-logo {
+      background-image: url(@product-logo-src);
+      background-repeat: no-repeat;
+      background-size: contain;
+      display: inherit;
+      height: inherit;
+      width: @full-logo-width;
+    }
+
+    & when (@layout-screen-small-enabled) {
+
+      @media (max-width: @layout-screen-small-min-width) {
+        width: @short-logo-width;
+      }
+    }
+  }
+}
--- a/src/styles/components/header-bar/variables.less
+++ b/src/styles/components/header-bar/variables.less
@@ -1,0 +1,13 @@
+@header-bar-enabled: true;
+
+/*
+ * Product Logo
+ */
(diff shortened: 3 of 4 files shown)


tnguyen0:
@leemunroe can you get it a quick look?

juliangieseke:
On that Feature branch: We decided not to do feature branches anymore, instead we disable WIP stuff before we commit.

leemunroe:
Nice work @tnguyen0. This looks on track. Here's the final state we're targeting. Assuming color, spacing, hamburger is a separate task/PR?

image
image

leemunroe:
@tnguyen0 For the logo, couple of comments. 

1. Is there a reason we didn't use the logo that already exists in the product? We use one on the login. I think this could be reused so we're not loading more assets.

2. Looking at your CSS, couldn't we use the same logo graphic for both desktop and responsive views? Since we're using it like a background sprite the "Mesosphere DC/OS" part would just be hidden given the width.

3. I have a feeling these logo graphics are not suited for retina displays.

juliangieseke:
@leemunroe regarding the colors and other parts: they will be added in parallel issues and there is one “janitor issue” at the end. Your points about the images are valid tho 🤔

tnguyen0:
Thanks @leemunroe for your feedback. Regarding your points:

1. The existing logo wasn't the correct size. I had originally resized it using CSS, which worked fine. But then I thought it would be better to just add a new asset so there is no coupling.

2. I thought about doing this, but again opted to add a new asset.

But If we want to avoid adding new assets when possible, then I can make the changes.

3. How can I check this? They are essentially same graphics for login screen. So would the same point apply there too?

leemunroe:
@tnguyen0 I think if you were to reuse that graphic, use CSS to resize, that would solve 1, 2 & 3 issues above. By scaling it down that solves the retina issue. By reusing it then we're not introducing two new assets and easier to maintain.

tnguyen0:
@juliangieseke @leemunroe done

src/js/components/HeaderBar.tsx
```diff
@@ -1,4 +1,5 @@
 import * as React from "react";
+
```
juliangieseke: 🤔

src/js/components/HeaderBar.tsx
```diff
@@ -9,10 +10,12 @@ export default function HeaderBar() {
 
   return (
     <UIHeaderBar>
-      <span>toggle component here</span>
-      <span>logo component here</span>
-      <span>user-menu component here</span>
-      <span>cluster-menu component here</span>
+      {/* <span>toggle component here</span> */}
```
juliangieseke: please do not touch these lines to prevent conflicts 👍

juliangieseke:
gnah, forgot the summary: Yeah! LGTM 👍  - But two very small things which are important because we do want to work in parallel and need to be careful which lines we actually change.

src/js/components/HeaderBar.tsx
tnguyen0: Indeed

src/js/components/HeaderBar.tsx
tnguyen0: Good point

tnguyen0:
@DanielMSchmidt @juliangieseke comments addressed

PR #3064, opened 2018-07-11
DCOS-38822: NodesTable / CPU Columns

juliangieseke:
## Testing

Change this line in `NodesTableContainer.js`

```diff
-import NodesTable from "../../../components/NodesTable";
+import NodesTable from "../../../components/NodesTable-new";
```

After that please take a look at the nodes table and see if the health column looks right and sorts correctly

## Trade-offs

- CPU percentage and resource bar are two different columns, once we have finished DCOS-39147, we can merge them and hide the bar on smaller screens

## Dependencies

None

### DIFF
--- a/plugins/nodes/src/js/columns/NodesTableCPUColumn.tsx
+++ b/plugins/nodes/src/js/columns/NodesTableCPUColumn.tsx
@@ -7,12 +7,9 @@ import { IWidthArgs as WidthArgs } from "#PLUGINS/nodes/src/js/types/IWidthArgs"
 import { SortDirection } from "plugins/nodes/src/js/types/SortDirection";
 
 export function cpuRenderer(data: Node): React.ReactNode {
-  // TODO: DCOS-38822
-  return <span>{data.get("used_resources").cpus.toString()}</span>;
+  return `${data.getUsageStats("cpus").percentage}%`;
 }
 export function cpuSorter(data: Node[], sortDirection: SortDirection): Node[] {
-  // TODO: DCOS-38822
-  // current implementation is a rough idea, not sure if it is the best one…
   const sortedData = data.sort((a, b) =>
     compareValues(
       a.get("used_resources").cpus.toString(),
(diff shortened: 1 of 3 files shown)


plugins/nodes/src/js/columns/NodesTableCPUColumn.tsx
```diff
@@ -7,8 +7,7 @@ import { IWidthArgs as WidthArgs } from "#PLUGINS/nodes/src/js/types/IWidthArgs"
 import { SortDirection } from "plugins/nodes/src/js/types/SortDirection";
 
 export function cpuRenderer(data: Node): React.ReactNode {
-  // TODO: DCOS-38822
-  return <span>{data.get("used_resources").cpus.toString()}</span>;
+  return `${data.getUsageStats("cpus").percentage}%`;
```
DanielMSchmidt: Funny how template strings lead to us writing `${}` instead of `+` 😆

juliangieseke:
sidenote: would be perfectly fine to block this PR because of its missing description, I simply forgot adding one 😟

plugins/nodes/src/js/columns/NodesTableCPUColumn.tsx
```diff
@@ -7,8 +7,7 @@ import { IWidthArgs as WidthArgs } from "#PLUGINS/nodes/src/js/types/IWidthArgs"
...
 }
 export function cpuSorter(data: Node[], sortDirection: SortDirection): Node[] {
   // TODO: DCOS-38822
```
DanielMSchmidt: There is a comment with this ticket number missing and the sorting is inverse of what the old implementation provides

juliangieseke: hm, you're right, not the question is: was the old one intentional or was it a bug? 🤔

DanielMSchmidt: No, unintentional

DanielMSchmidt: It seems like github does not allow me to comment on the right side, but only here

juliangieseke: after slack discussion: we keep it as is 👍

juliangieseke:
removed todo comment 👍  @nLight @DanielMSchmidt please revisit 🙇

PR #3065, opened 2018-07-11
DCOS-21723: Include scheduler resources 

nLight:
After lengthy discussions with many parties we've decided to include scheduler resources into report. To make data more sensible from the stand point of *DCOS*. 

So now on the Services view for frameworks we show allocated resources + scheduler resources:
screen shot 2018-07-11 at 16 15 49

Which is now aligned with the way we display Framework Tasks:
screen shot 2018-07-11 at 16 15 59

💁‍♂️  DCOS is not Mesos. (Duh..)

In DCOS we have a concept of root Marathon that we don't really reflect on the DCOS UI.

This root Marathon runs all the pods, services as well as scheduler tasks for frameworks.

We display Pods and Services on the Services view on the top level, like the `blip` task on the screenshot, wheres in Mesos they go under the root Marathon.

We display frameworks on the top level as well as they would go under Marathon, `cassandra` and `hello-world`, but not as a _task_ but as a separate entity a Framework. Though on Mesos those frameworks go _not_ under the root Marathon, they go _along_ with the root Marathon. 

Here where the complication starts.

As we don't display scheduler task but rather a Framework on the Services view, we move scheduler task to the Tasks view, as it would be launched by the framework itself.

Cassandra task you see on the Tasks page is a scheduler that is started by the root Marathon. This is why when we request allocated resources for Cassandra from Mesos we don't get scheduler's resource. We should add scheduler's resources from the Marathon's spec.

But wait, `CPU: 1 + 0.5 != 1.6` still doesn't look right! Well, what can I say? _I just sum_ what Mesos and Marathon give me. From my conversations with the core team the extra `0.1` is used by the executor.

We even have this info, but can we really subtract it or render it at some place of the Tasks view? I don't know. Is it the part of the ticket? Definitely not.

screen shot 2018-07-12 at 12 22 48

### DIFF
--- a/plugins/services/src/js/structs/Framework.js
+++ b/plugins/services/src/js/structs/Framework.js
@@ -97,13 +97,23 @@ module.exports = class Framework extends Application {
     // used_resources is actually allocated resources
     // the name can't be changed to keep backward compatibility.
     // This is why `getResources` returns `used_resources`
-    return (
-      this.get("used_resources") || {
-        cpus: 0,
-        mem: 0,
-        gpus: 0,
-        disk: 0
-      }
-    );
+    const allocatedFrameworkResources = this.get("used_resources") || {
+      cpus: 0,
+      mem: 0,
+      gpus: 0,
+      disk: 0
+    };
+
+    // Framework doesn't know how many resources its scheduler consumes.
+    // Scheduler is launched by Marathon not the Framework itself.
+    // So we should get this information separately from the Marathon spec
+    const schedulerResources = this.getSpec().getResources();
+
+    return {
+      cpus: allocatedFrameworkResources.cpus + schedulerResources.cpus,
+      mem: allocatedFrameworkResources.mem + schedulerResources.mem,
+      gpus: allocatedFrameworkResources.gpus + schedulerResources.gpus,
+      disk: allocatedFrameworkResources.disk + schedulerResources.disk
+    };
   }
 };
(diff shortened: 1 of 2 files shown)


plugins/services/src/js/structs/Framework.js
```diff
@@ -97,13 +97,35 @@ module.exports = class Framework extends Application {
...
-    );
+    const allocatedFrameworkResources = this.get("used_resources") || {
+      cpus: 0,
+      mem: 0,
+      gpus: 0,
+      disk: 0
+    };
+
+    // TODO: Circular reference workaround DCOS_OSS-783
+    const MesosStateStore = require("#SRC/js/stores/MesosStateStore");
```
orlandohohmeier: 🤔 Wonder whether we can read the scheduler resource form the spec using `this.getSpec().getResources();` instead of using the Mesos state store.

nLight: We can explore the possibility

juliangieseke: IMHO: create JIRA and fix later 👍 (I'll approve once the JIRA is there)

nLight: I think it makes sense exploring it on the spot. We have 8 SP planned and the performance win can be big.

DanielMSchmidt: @nLight Do I get it right that you are then still working on this? Could we move the Jira into Progress then?

nLight: There and back again 💪

nLight: @orlandohohmeier you were right, it looks like we can just use the spec, at least numbers don't change 🤷‍♂️

nLight:
@juliangieseke do you mean explained on the Task view?

PR #3066, opened 2018-07-12
DCOS-38827: add health column

DanielMSchmidt:
## Testing

Change this line in `NodesTableContainer.js`

```diff
-import NodesTable from "../../../components/NodesTable";
+import NodesTable from "../../../components/NodesTable-new";
```

After that please take a look at the nodes table and see if the health column looks right and sorts correctly

## Trade-offs

- I decided not to move the sorting utils to typescript, because for me this would be out of scope
- I choose to use table driven tests because the test cases looked very similar to me

## Dependencies

None

### DIFF
--- a/plugins/nodes/src/js/columns/NodesTableHealthColumn.tsx
+++ b/plugins/nodes/src/js/columns/NodesTableHealthColumn.tsx
@@ -1,5 +1,5 @@
 import * as React from "react";
-import { compareValues } from "#PLUGINS/nodes/src/js/utils/compareValues";
+import UnitHealthUtil from "#SRC/js/utils/UnitHealthUtil";
 import Node from "#SRC/js/structs/Node";
 // TODO: DCOS-39079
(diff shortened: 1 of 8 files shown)


plugins/nodes/src/js/columns/NodesTableHealthColumn.tsx
```diff
@@ -1,31 +1,24 @@
...
-      a.getHealth().title.toLowerCase(),
-      b.getHealth().title.toLowerCase(),
-      a.getHostName().toLowerCase(),
-      b.getHostName().toLowerCase()
-    )
-  );
+  const sortedData = data.sort(UnitHealthUtil.getHealthSortFunction);
   return sortDirection === "ASC" ? sortedData : sortedData.reverse();
 }
+
```
juliangieseke: 🤔

DanielMSchmidt: I like it better that way 🤷‍♂️

PR #3068, opened 2018-07-12
DCOS-38819: new table column name

GeorgiSTodorov:
refactor(NodesTable): add old functionality

Add old column name functionality to new refactored table

Closes DCOS-38819

**Checklist**

### DIFF
--- a/plugins/nodes/src/js/columns/NodesTableHostnameColumn.tsx
+++ b/plugins/nodes/src/js/columns/NodesTableHostnameColumn.tsx
@@ -5,16 +5,43 @@ import Node from "#SRC/js/structs/Node";
 // import { IWidthArgs as WidthArgs } from "@dcos/ui-kit/packages/table/components/Column";
 import { IWidthArgs as WidthArgs } from "#PLUGINS/nodes/src/js/types/IWidthArgs";
 import { SortDirection } from "plugins/nodes/src/js/types/SortDirection";
+import { Link } from "react-router";
+import { Tooltip } from "reactjs-components";
+import Icon from "#SRC/js/components/Icon";
 
 export function hostnameRenderer(data: Node): React.ReactNode {
-  // TODO: DCOS-38819
-  return <span>{data.getHostName().toString()}</span>;
+  const nodeID = data.get("id");
+  let headline = data.get("hostname");
+
+  if (!data.isActive()) {
+    return (
+      <Link className="table-cell-link-primary" to={`/nodes/${nodeID}`}>
+        <span title={headline}>
+          <Tooltip anchor="start" content="Connection to node lost">
+            <Icon
+              className="icon-alert icon-margin-right"
+              color="neutral"
+              id="yield"
+              size="mini"
+            />
+            {headline}
+          </Tooltip>
+        </span>
+      </Link>
+    );
+  }
+
+  return (
+    <Link className="table-cell-link-primary" to={`/nodes/${nodeID}`}>
+      <span title={headline}>{headline}</span>
+    </Link>
+  );
 }
+
 export function hostnameSorter(
   data: Node[],
   sortDirection: SortDirection
 ): Node[] {
-  // TODO: DCOS-38819
   // current implementation is a rough idea, not sure if it is the best one…
   const sortedData = data.sort((a, b) =>
     compareValues(
@@ -26,7 +53,7 @@ export function hostnameSorter(
   );
   return sortDirection === "ASC" ? sortedData : sortedData.reverse();
 }
+
 export function hostnameSizer(args: WidthArgs): number {
-  // TODO: DCOS-38819
-  return Math.max(100, args.width / args.totalColumns);
+  return Math.max(150, args.width / args.totalColumns);
 }
--- a/src/js/components/Icon.d.ts
+++ b/src/js/components/Icon.d.ts
@@ -1,0 +1,11 @@
+import * as React from "react";
+
+interface IconProps {
+  className?: string[] | object | string;
+  color?: string;
+  family?: string;
+  id: string;
+  size?: string;
+}
+
+export default class Icon extends React.Component<IconProps> {}
--- a/src/typings/reactjs-components.d.ts
+++ b/src/typings/reactjs-components.d.ts
@@ -47,4 +47,23 @@ declare module "reactjs-components" {
 
   export class Modal extends Component<ModalProps, {}> {}
   export class Dropdown extends Component<DropdownProps, {}> {}
+
+  interface TooltipProps {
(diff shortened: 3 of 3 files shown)


weblancaster:
@GeorgiSTodorov Looks good, I would just remove the `npm-shrinkwrap.json` fromt the commit/PR.

GeorgiSTodorov:
@weblancaster I reverted `npm-shrinkwrap.json`.

juliangieseke:
@jaxesn the bot is for our CI 😆 No idea how to enable it tho 🤔 @nLight ?

Fabs:
I get this [object:object] Native tooltip on chrome. Otherwise it looks fine. 

https://cl.ly/3P1Q3E2F1j1h

plugins/nodes/src/js/columns/NodesTableHostnameColumn.tsx
```diff
@@ -5,16 +5,39 @@ import Node from "#SRC/js/structs/Node";
...
 import { IWidthArgs as WidthArgs } from "#PLUGINS/nodes/src/js/types/IWidthArgs";
 import { SortDirection } from "plugins/nodes/src/js/types/SortDirection";
+import { Link } from "react-router";
+import { Tooltip } from "reactjs-components";
+import Icon from "#SRC/js/components/Icon";
 
 export function hostnameRenderer(data: Node): React.ReactNode {
-  // TODO: DCOS-38819
-  return <span>{data.getHostName().toString()}</span>;
+  let headline = data.get("hostname");
```
Fabs: I find a little confusing to follow the `headline` variable, especially the `headline = "... {headline}..."` bit.

what do you think of separating the warning tooltip logic away from the cell?
```js
interface NodeProps {
  node: Node
}

const HostName:React.StatelessComponent<NodeProps> = ({node}) => {
  const hostname = node.get("hostname");
  
  if (node.isActive()) {
    return <span title={hostname}>{hostname}</span>;
  }

  const hostnameWithTooltip = (
    <Tooltip anchor="start" content="Connection to node lost">
      <Icon
        className="icon-alert icon-margin-right"
        color="neutral"
        id="yield"
        size="mini"
      />
      {hostname}
    </Tooltip>
  );

  return <span title={hostname}>{hostnameWithTooltip}</span>;
};

export function hostnameRenderer(data: Node): React.ReactNode {
  const nodeID = data.get("id");

  return (
    <Link className="table-cell-link-primary" to={`/nodes/${nodeID}`}>
      <HostName node={data} />
    </Link>
  );
}
```

Fabs:
Also think there is a layout issue when the screen is small.

image

GeorgiSTodorov:
@Fabs I figured out the object object issue, I will discuss the other two comments with Julian.

plugins/nodes/src/js/columns/NodesTableHostnameColumn.tsx
juliangieseke: This code is going to be replaced in DCOS-39159, I'd vote not to invest more time than necessary here.

Also I dont agree with your solution Fabs, if `cellRenderer` gets too big, there is an issue with the abstraction and it shouldnt be fixed in that particular renderer - also I agree with you, that this renderer is too big 😅

IMHO the proper solution would be to split up the column into two (hostname & warningIcon).

plugins/nodes/src/js/columns/NodesTableHostnameColumn.tsx
```diff
@@ -26,7 +49,7 @@ export function hostnameSorter(
   );
   return sortDirection === "ASC" ? sortedData : sortedData.reverse();
 }
+
 export function hostnameSizer(args: WidthArgs): number {
-  // TODO: DCOS-38819
   return Math.max(100, args.width / args.totalColumns);
```
juliangieseke: I think we need to increase min-width a bit to resolve the line breaking issue

juliangieseke:
@Fabs forgot to ping you above ^

plugins/nodes/src/js/columns/NodesTableHostnameColumn.tsx
Fabs: can we then just avoid?
```js
let headline = ...
headline = "... {headline}..."
```

juliangieseke: > This code is going to be replaced in DCOS-39159, I'd vote not to invest more time than necessary here.

You block it, Georgi changes it 🤷‍♂️

GeorgiSTodorov:
I fixed the "object object" and the layout, now it looks good even on narrow windows. I decided to leave the tooltip logic as it is, at least for now.

plugins/nodes/src/js/columns/NodesTableHostnameColumn.tsx
Fabs: I would like to see the variable naming fixed. Especially because we introduced now that we do:

```js
let headline = data.get("hostname");
headline = "... {headline}..."
...
...
<span title={data.get("hostname")}>{headline}</span>
```

Don't you think that if we had those names right on the beginning, we could have avoided the "object object" bug? 

(I am dismissing my block on any case, just would like that point to be considered)

juliangieseke: I agree, yes, but this is legacy code Georgi just copy & pasted… 🤷‍♂️

tnguyen0: @juliangieseke DCOS-39159 references the health/status column. Do you mean DCOS-39160? If so, this ticket is meant to clean up/refactor all the columns that were added? Shouldn't we just do it right in this ticket?

juliangieseke:
I‘ll clarify this with Lily 👍🏻

DanielMSchmidt:
@GeorgiSTodorov There are merge conflicts, could you take a look?

GeorgiSTodorov:
It should be ok now.

plugins/nodes/src/js/columns/NodesTableHostnameColumn.tsx
DanielMSchmidt: As we decided that we need to own our copy and pasted legacy code I think we need to make this better. Let's get rid of the headline variable :)

GeorgiSTodorov: @DanielMSchmidt I have updated the code, please share your opinion.

juliangieseke:
I have explicitly added the removal of the legacy code to https://jira.mesosphere.com/browse/DCOS-39159 - thanks for approving 🙇

PR #3069, opened 2018-07-13
DCOS-38826: add region column

DanielMSchmidt:
## Testing

```diff
-import NodesTable from "../../../components/NodesTable";
+import NodesTable from "../../../components/NodesTable-new";
```

Please take a look at the nodes table and see if the regions column behaves the same as before (except for the sort direction, which got fixed)

## Trade-offs

- We need to use a different version of the renderer because we need to
pass in an extra parameter coming from the props. I decided to bind a partial application to the scope of the nodes table instead of loosening up the entire API, I am open to change here if you think there is a nicer solution or if there are more examples of props being needed there.

## Dependencies

All is ready 👍

### DIFF
--- a/plugins/nodes/src/js/columns/NodesTableRegionColumn.tsx
+++ b/plugins/nodes/src/js/columns/NodesTableRegionColumn.tsx
@@ -6,16 +6,22 @@ import Node from "#SRC/js/structs/Node";
 import { IWidthArgs as WidthArgs } from "#PLUGINS/nodes/src/js/types/IWidthArgs";
 import { SortDirection } from "plugins/nodes/src/js/types/SortDirection";
 
-export function regionRenderer(data: Node): React.ReactNode {
-  // TODO: DCOS-38826
-  return <span>{data.getRegionName().toString()}</span>;
+export function regionRenderer(
+  masterRegion: string,
+  data: Node
+): React.ReactNode {
+  const regionName = data.getRegionName();
+  return (
+    <span title={regionName}>
+      {regionName}
+      {masterRegion === regionName ? " (Local)" : null}
+    </span>
+  );
 }
 export function regionSorter(
   data: Node[],
   sortDirection: SortDirection
 ): Node[] {
-  // TODO: DCOS-38826
-  // current implementation is a rough idea, not sure if it is the best one…
   const sortedData = data.sort((a, b) =>
     compareValues(
       a.getRegionName().toLowerCase(),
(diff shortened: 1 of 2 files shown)


plugins/nodes/src/js/components/NodesTable-new.tsx
```diff
@@ -78,6 +78,8 @@ export default class NodesTable extends React.Component<
   NodesTableProps,
   NodesTableState
 > {
+  private regionRenderer: (data: Node) => React.ReactNode;
```
juliangieseke: Uhm, this makes me sad 😢 

Cant we find another way injecting the necessary data? I'd rather go with `data.map(node => injectMasterRegion(node, this.props.masterRegion))` then messing around with the table already… 🙇 

```
function injectMasterRegion(node: Node, masterRegion string) {
node.__thisisbadbuttheotherooptionisevenworse.masterRegion = masterRegion;
return node;
}
```

plugins/nodes/src/js/columns/NodesTableRegionColumn.tsx
```diff
@@ -27,6 +33,5 @@ export function regionSorter(
   return sortDirection === "ASC" ? sortedData : sortedData.reverse();
 }
 export function regionSizer(args: WidthArgs): number {
-  // TODO: DCOS-38826
-  return Math.max(100, args.width / args.totalColumns);
+  return Math.max(170, args.width / args.totalColumns);
```
Fabs: Does the 170 magic number have some meaning?

plugins/nodes/src/js/components/NodesTable-new.tsx
juliangieseke: can you please create a JIRA (and add a comment to the code) which finds a proper solution for this redirect?

Fabs: What about enhancing `data` so we can call `data.getMaterRegion` or even `data.isInMasterRegion`. 

I think the table API has a invariant I would like to keep: every cell gets a `data: Node` which contains all the data it needs to render. So the solution to maintain this invariant would be to pass as part of the data object that missing information IMHO.

What do you think?

DanielMSchmidt: @Fabs I thought about this, but it would mean that we need to pass extra data to the constructor of the nodes at every place or that we would need to access the deprecated composite state store from within the nodes struct. 

I would like to see if we need other workarounds like this for the other tables. If we do, we might need to find a fix for the table, in general, to allow us to pass in some additional general information. If we don't we should make sure that we refactor this away into the data layer once we apply it here.

plugins/nodes/src/js/columns/NodesTableRegionColumn.tsx
juliangieseke: `min-width`

Fabs: Can we name it that then?

PR #3072, opened 2018-07-13
DCOS-38941: Implement cluster header dropdown 

DanielMSchmidt:
## Testing

Please change the following lines in `src/js/components/HeaderBar.tsx`

```diff
export default function HeaderBar() {
  // remove this to activate component
- if (arguments) {
-  return null;
- }
```

## Trade-offs

- I choose to refactor the dropdown instead of moving and changing it as that allows me to do a lot of small refactoring commits moving into the right direction and I think these are easier to review. Please give me feedback on this assumption 🙏 

## Dependencies

None

### DIFF
--- a/src/js/components/ClusterDropdown.d.ts
+++ b/src/js/components/ClusterDropdown.d.ts
@@ -1,0 +1,9 @@
+import { ReactNode, Component } from "react";
+
+interface Props {}
+
+interface State {
+  isTextCopied: boolean;
+}
+
+export default class ClusterDropdown extends Component<Props, State> {}
--- a/src/js/components/ClusterDropdown.js
+++ b/src/js/components/ClusterDropdown.js
@@ -1,0 +1,181 @@
+import * as React from "react";
+import mixin from "reactjs-mixin";
+import { Dropdown } from "reactjs-components";
+import { StoreMixin } from "mesosphere-shared-reactjs";
+import { MountService } from "foundation-ui";
+import { routerShape } from "react-router";
+
+import MesosSummaryStore from "#SRC/js/stores/MesosSummaryStore";
+import MetadataStore from "#SRC/js/stores/MetadataStore";
+import SidebarActions from "#SRC/js/events/SidebarActions";
+import ClipboardTrigger from "#SRC/js/components/ClipboardTrigger";
+
+function ClusterDropdownTrigger({ onTrigger, children }) {
+  return (
+    <span className="header-bar-dropdown-trigger" onClick={onTrigger}>
+      {children}
+    </span>
+  );
+}
+const METHODS_TO_BIND = ["handleItemSelect", "handleTextCopy"];
+export default class ClusterDropdown extends mixin(StoreMixin) {
+  constructor() {
+    super(...arguments);
+
+    this.state = {
+      isTextCopied: false
+    };
+
+    this.store_listeners = [
+      {
+        name: "metadata",
+        events: ["success"],
+        listenAlways: false
+      },
+      {
+        name: "summary",
+        events: ["success"],
+        listenAlways: false
+      }
+    ];
+
+    METHODS_TO_BIND.forEach(method => {
+      this[method] = this[method].bind(this);
+    });
+  }
+
+  getClusterName() {
+    const states = MesosSummaryStore.get("states");
+    let clusterName = null;
+
+    if (states) {
+      const lastState = states.lastSuccessful();
+
+      if (lastState) {
+        clusterName = lastState.getClusterName();
+      }
+    }
+
+    return clusterName;
+  }
+
+  getPublicIP() {
+    const metadata = MetadataStore.get("metadata");
+
+    if (
+      typeof metadata !== "object" ||
+      metadata.PUBLIC_IPV4 == null ||
+      metadata.PUBLIC_IPV4.length === 0
+    ) {
+      return null;
+    }
+
+    return metadata.PUBLIC_IPV4;
+  }
+
+  handleTextCopy() {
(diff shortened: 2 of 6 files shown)


src/js/components/ClusterDropdown.d.ts
```diff
@@ -0,0 +1,9 @@
+import { ReactNode, Component } from "react";
+
+interface Props {}
+
+interface State {
```
juliangieseke: 🎨 I think we used to `ComponentNameProps` and should stickt to one pattern… But thats nothing blocking here

src/js/components/ClusterDropdown.js
```diff
@@ -0,0 +1,181 @@
...
+
+function ClusterDropdownTrigger({ onTrigger, children }) {
+  return (
+    <span className="header-bar-dropdown-trigger" onClick={onTrigger}>
+      {children}
+    </span>
+  );
+}
+const METHODS_TO_BIND = ["handleItemSelect", "handleTextCopy"];
+export default class ClusterDropdown extends mixin(StoreMixin) {
```
juliangieseke: 🎨 if you're refactoring anyway, why not getting rid of the StoreMixing? :trollface:

src/js/components/ClusterDropdown.js
```diff
@@ -0,0 +1,181 @@
...
+  }
+
+  getClusterName() {
+    const states = MesosSummaryStore.get("states");
+    let clusterName = null;
+
+    if (states) {
+      const lastState = states.lastSuccessful();
+
+      if (lastState) {
```
juliangieseke: :art: legacy code? so many ifs :(

juliangieseke:
I'm almost sure I'm doing something wrong (CI is green) 🤔 
```
Failed to compile.

./src/js/components/HeaderBar.tsx
[tsl] ERROR in /Users/Julian/Code/dcos-ui/master/dcos-ui/src/js/components/HeaderBar.tsx(18,7)
      TS2605: JSX element type 'ClusterDropdown' is not a constructor function for JSX elements.
  Property 'setState' is missing in type 'ClusterDropdown'.
```

DanielMSchmidt:
@juliangieseke Verified that it is not happening for me. I did a `git clean -dfx` and started the server afterwards and everything worked fine

src/js/components/ClusterDropdown.js
```diff
@@ -0,0 +1,181 @@
+import * as React from "react";
+import mixin from "reactjs-mixin";
+import { Dropdown } from "reactjs-components";
+import { StoreMixin } from "mesosphere-shared-reactjs";
```
weblancaster: I thought we were not supposed to be using mixins anymore. Can you check please @DanielMSchmidt

src/js/components/ClusterDropdown.js
weblancaster: Agree!

src/js/components/ClusterDropdown.js
```diff
@@ -0,0 +1,181 @@
...
+      }
+    ];
+
+    METHODS_TO_BIND.forEach(method => {
+      this[method] = this[method].bind(this);
+    });
+  }
+
+  getClusterName() {
+    const states = MesosSummaryStore.get("states");
```
weblancaster: Is there a `struct` we should move this to?

src/js/components/ClusterDropdown.js
DanielMSchmidt: That is one of the things I meant in the sync, it is just copied over logic and I don't really think removing the store mixin here and moving everything away from the stores, in general, is part of this ticket 🙈

src/js/components/ClusterDropdown.d.ts
DanielMSchmidt: Can you explain why? I didn't understand the pattern in the first place

DanielMSchmidt:
@weblancaster One of the refactorings was easy because there is already a struct, the other one would be adding a new struct for the `MetadataStore`. I think this is not the direction we want to go, therefore I just pushed one refactoring commit.

@tnguyen0 @juliangieseke Please take another look :)

src/js/components/ClusterDropdown.d.ts
juliangieseke: I think its the same as having `JobTypeResolver` vs `TypeResolver` and `hostnameRenderer` vs `renderer` 🤷‍♂️ 

Personally, I like the more verbose thing more - but if there is a reason to use the shorter ones (e.g. for parsing) - I'm happy change my opinion 👍

src/js/components/ClusterDropdown.js
nLight: As discussed offline. Let's try just replacing the mixing with empty subscriptions. The component doesn't have any loading state etc. it just uses the data without prior checking whether it has been loaded and unsubscribes after the first response.

src/js/components/ClusterDropdown.js
```diff
@@ -0,0 +1,172 @@
...
+
+    return states ? states.getClusterName() : null;
```
juliangieseke: so this can return `null` if the clustername wasnt fetched yet, correct? 
dont rellay like it TBH :( (and this is your code, you added `getClusterName` to SummaryList :D 

Maybe some generic `Cluster Options` text would be better 🤔 

...wasnt able to "break" it while testing tho -> not blocking

DanielMSchmidt: I understand your concerns and I think we should make it a lot better when moving to the data layer. I also don't see it as blocking, therefore I am merging 😇

PR #3073, opened 2018-07-14
DCOS-38940: Add User Account Dropdown on Header Bar

tnguyen0:
Creates AccountDropdown as the menu list and AccountDropdownTrigger as the
dropdown display. Adds accompanying typescript files. Adds truncation and
responsive sizing.

Closes DCOS-38940

## Testing
* Checkout `origin/tommy/feature/DCOS-38940-user-dropdown-header-bar`
* Enable the header bar by commenting out the following lines in `HeaderBar.tsx`
```
  // remove this to activate component
  if (arguments) {
    return null;
  }
```
* Comment out `<span>toggle component here</span>` in `HeaderBar.tsx`

## Result

header-bar-dropdown

## Dependencies

@DanielMSchmidt I added a placeholder for the cluster dropdown, you should be able to drop what you do in there.

### DIFF
--- a/packages/ui-kit-stage/HeaderBar.tsx
+++ b/packages/ui-kit-stage/HeaderBar.tsx
@@ -6,6 +6,7 @@
 import * as React from "react";
 import { coreColors } from "@dcos/ui-kit/dist/packages/shared/styles/color";
 
+// TODO: DCOS-39074
 const headerBarStyles: object = {
   backgroundColor: coreColors().purpleDarken4,
   height: "32px",
--- a/src/js/components/AccountDropdownTrigger.js
+++ b/src/js/components/AccountDropdownTrigger.js
@@ -1,0 +1,43 @@
+import mixin from "reactjs-mixin";
+import PropTypes from "prop-types";
+import React from "react";
+import { StoreMixin } from "mesosphere-shared-reactjs";
+
+class AccountDropdownTrigger extends mixin(StoreMixin) {
+  constructor() {
+    super(...arguments);
+
+    this.store_listeners = [
+      {
+        name: "metadata",
+        events: ["success"],
+        listenAlways: false
+      }
+    ];
+  }
+
+  componentDidUpdate() {
+    if (this.props.onUpdate) {
+      this.props.onUpdate();
+    }
+  }
+
+  render() {
+    const { content, onTrigger } = this.props;
+
+    // TODO: DCOS-38944
+    return (
+      <a className="header-bar-dropdown" onClick={onTrigger}>
+        <span>{content}</span>
+      </a>
+    );
+  }
+}
+
+AccountDropdownTrigger.propTypes = {
+  content: PropTypes.string.isRequired,
+  onUpdate: PropTypes.func,
+  onTrigger: PropTypes.func
+};
+
+module.exports = AccountDropdownTrigger;
--- a/src/js/components/HeaderBar.tsx
+++ b/src/js/components/HeaderBar.tsx
@@ -1,6 +1,8 @@
 import * as React from "react";
+
 import { HeaderBar as UIHeaderBar } from "ui-kit-stage/HeaderBar";
 import ClusterDropdown from "./ClusterDropdown";
+import AccountDropdown from "./AccountDropdown";
 
 export default function HeaderBar() {
   // remove this to activate component
@@ -14,7 +16,7 @@ export default function HeaderBar() {
       <div className="header-bar-logo-wrapper">
         <div className="header-bar-logo" />
       </div>
-      <span>user-menu component here</span>
+      <AccountDropdown />
       <ClusterDropdown />
     </UIHeaderBar>
   );
--- a/src/js/stores/AuthStore.js
+++ b/src/js/stores/AuthStore.js
@@ -96,6 +96,20 @@ class AuthStore extends GetSetBaseStore {
     }
   }
 
+  getUserLabel() {
+    const user = this.getUser();
+
+    if (!user) {
+      return null;
+    }
+
+    if (!user.is_remote) {
+      return user.description;
+    }
+
+    return user.uid;
+  }
+
   processLoginSuccess() {
     Hooks.doAction("userLoginSuccess");
     this.emit(AUTH_USER_LOGIN_CHANGED);
--- a/src/js/stores/BaseStore.d.ts
+++ b/src/js/stores/BaseStore.d.ts
@@ -1,0 +1,6 @@
+import { EventEmitter } from "events";
+
+export default class BaseStore extends EventEmitter {
+  addChangeListener: (eventName: string, callback: () => void) => void;
+  removeChangeListener: (eventName: string, callback: () => void) => void;
+}
--- a/src/js/components/AccountDropdown.d.ts
+++ b/src/js/components/AccountDropdown.d.ts
@@ -1,0 +1,9 @@
+import { Component } from "react";
+
+interface AccountDropdownProps {
+  userName?: null | string;
+  menuItems?: object;
+  willAnchorRight?: boolean;
+}
+
+export default class AccountDropdown extends Component<AccountDropdownProps> {}
--- a/src/js/components/AccountDropdown.js
+++ b/src/js/components/AccountDropdown.js
@@ -1,0 +1,56 @@
+import { Dropdown } from "reactjs-components";
+import React from "react";
+
+import AuthStore from "#SRC/js/stores/AuthStore";
+import AccountDropdownTrigger from "./AccountDropdownTrigger";
+
+class AccountDropdown extends React.Component {
+  constructor() {
+    super(...arguments);
+    this.userLabel = AuthStore.getUserLabel();
+  }
+
+  getMenuItems() {
+    const menuItems = [
+      {
+        className: "dropdown-menu-section-header",
+        html: <label>{this.userLabel}</label>,
+        id: "header-support",
+        selectable: false
+      },
+      {
+        html: "Sign Out",
+        id: "sign-out",
+        onClick: AuthStore.logout
+      }
+    ];
+
+    return menuItems;
+  }
+
+  getTrigger() {
+    return <AccountDropdownTrigger content={this.userLabel} />;
+  }
+
+  handleItemSelection(item) {
+    if (item.onClick) {
+      item.onClick();
+    }
+  }
+
+  render() {
+    return (
+      <Dropdown
+        trigger={this.getTrigger()}
+        dropdownMenuClassName="user-account-dropdown-menu dropdown-menu header-bar-dropdown-menu"
+        dropdownMenuListClassName="user-account-dropdown-list dropdown-menu-list"
+        items={this.getMenuItems()}
+        onItemSelection={this.handleItemSelection}
+        persistentID="dropdown-trigger"
+        transition={true}
+      />
+    );
+  }
+}
+
+module.exports = AccountDropdown;
--- a/src/js/stores/AuthStore.d.ts
+++ b/src/js/stores/AuthStore.d.ts
@@ -1,0 +1,13 @@
+import GetSetBaseStore from "./GetSetBaseStore";
+
+export default class AuthStore extends GetSetBaseStore {
+  constructor();
+  static login: () => void;
+  static logout(): void;
+  static isLoggedIn: () => boolean;
+  static getUser: () => null | object;
+  static getUserLabel: () => null | string;
+  processLoginSuccess: () => void;
+  processLogoutSuccess: () => void;
+  static storeID: () => string;
+}
--- a/src/js/stores/GetSetBaseStore.d.ts
+++ b/src/js/stores/GetSetBaseStore.d.ts
@@ -1,0 +1,6 @@
+import BaseStore from "./BaseStore";
+
+export default class GetSetBaseStore extends BaseStore {
+  get: (key: string) => null | object;
+  set: (data: object) => void;
+}
--- a/src/styles/components/header-bar/styles.less
+++ b/src/styles/components/header-bar/styles.less
@@ -3,8 +3,7 @@
   .header-bar-logo-wrapper {
     display: inline-block;
     height: 16px;
-    margin-left: 10px;
-    margin-top: 10px;
+    margin-top: 5px;
     overflow: hidden;
     width: @full-logo-width;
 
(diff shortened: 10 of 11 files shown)


tnguyen0:
@leemunroe can you give it a look?

nLight:
screen shot 2018-07-16 at 12 10 12

packages/ui-kit-stage/HeaderBar.tsx
```diff
@@ -13,7 +13,12 @@ const headerBarStyles: object = {
   flexShrink: 1,
   flexGrow: 0,
   flexBasis: "auto",
-  zIndex: 999
+  zIndex: 999,
+  display: "flex",
+  paddingLeft: "20px",
```
juliangieseke: this should be handled by the children IMHO 🤔

src/js/components/HeaderBar.tsx
```diff
@@ -1,20 +1,67 @@
 import * as React from "react";
+import AuthStore from "#SRC/js/stores/AuthStore";
+
 import { HeaderBar as UIHeaderBar } from "ui-kit-stage/HeaderBar";
+import AccountDropdown from "./AccountDropdown";
+
+function getUserAccountDropdown() {
```
juliangieseke: 🎨 I'd love to see this moved to its own component, `HeaderBar` should really only be the bar itself 🤔  (applies to all of the added code here)

src/js/stores/AuthStore.js
```diff
@@ -96,6 +96,20 @@ class AuthStore extends GetSetBaseStore {
...
+
+    let userLabel = null;
+
+    if (user && !user.is_remote) {
+      userLabel = user.description;
+    } else if (user && user.is_remote) {
+      userLabel = user.uid;
+    }
+
+    return userLabel;
```
juliangieseke: 🎨  FYI: we had a lot of discussions about code style lately discussing about the need of `let`, `if .. else` and ternaries.

That said, I'd love to have an early exit for `user == undefined` and then just `return user.is_remote ? user.uid : user.description;`

src/js/components/HeaderBar.tsx
```diff
@@ -1,20 +1,67 @@
...
+
   return (
     <UIHeaderBar>
       <span>toggle component here</span>
       <div className="header-bar-logo-wrapper">
         <div className="header-bar-logo" />
       </div>
-      <span>user-menu component here</span>
-      <span>cluster-menu component here</span>
+      <div className="header-bar-dropdown-wrapper">
```
juliangieseke: This PR is only about the user menu, would love to see the code not touching the cluster menu to prevent conflicts with @DanielMSchmidt :/ (@DanielMSchmidt does this work for your code?)

juliangieseke:
@nLight ^ ellipsis ticket: https://jira.mesosphere.com/browse/DCOS-38944

Fabs:
The previous label for `Logout` was `Sign Out`. Should we keep it?

image

src/js/components/AccountDropdownTrigger.js
```diff
@@ -0,0 +1,49 @@
...
+    if (this.props.onUpdate) {
+      this.props.onUpdate();
+    }
+  }
+
+  render() {
+    const { primaryContent } = this.props;
+
+    return (
+      <a className="header-bar-dropdown" onClick={this.props.onTrigger}>
```
Fabs: 🎨 I would also disconstruct onTrigger on line 26.

src/js/components/AccountDropdownTrigger.js
```diff
@@ -0,0 +1,49 @@
...
+    );
+  }
+}
+
+AccountDropdownTrigger.defaultProps = {
+  showCaret: false,
+  onTrigger() {}
+};
+
+AccountDropdownTrigger.propTypes = {
```
Fabs: Are we going to use those props later? right now it does not seem that they matter.

Should we also add here to this list `primaryContent` and `onTrigger`?

src/js/components/HeaderBar.tsx
```diff
@@ -1,20 +1,67 @@
 import * as React from "react";
+import AuthStore from "#SRC/js/stores/AuthStore";
+
 import { HeaderBar as UIHeaderBar } from "ui-kit-stage/HeaderBar";
+import AccountDropdown from "./AccountDropdown";
+
+function getUserAccountDropdown() {
+  const menuItems = [
+    {
+      html: "Logout",
```
Fabs: Should we keep the `Sign out` we had before this change?

src/js/components/HeaderBar.tsx
```diff
@@ -1,20 +1,67 @@
...
+      id: "logout",
+      onClick: AuthStore.logout
+    }
+  ];
+
+  var userLabel = AuthStore.getUserLabel();
```
Fabs: Can we use `const` instead of `var` since we re not mutating userLabel.

src/js/components/HeaderBar.tsx
```diff
@@ -1,20 +1,67 @@
...
+  );
+}
 
 export default function HeaderBar() {
   // remove this to activate component
   if (arguments) {
     return null;
   }
 
+  var userAccountDropdown = getUserAccountDropdown();
```
Fabs: Can we use `const` instead of `var` since we are not mutating those two values?

src/js/components/HeaderBar.tsx
```diff
@@ -1,20 +1,67 @@
...
+    />
+  );
+}
+
+function getClusterDropdown() {
+  const menuItems = [
+    {
+      html: "Install CLI",
+      id: "install-cli",
+      onClick: AuthStore.logout
```
DanielMSchmidt: This seems odd...

src/js/components/HeaderBar.tsx
```diff
@@ -1,20 +1,67 @@
...
+    }
+  ];
+
+  const userLabel = "frontend-us-west-2-bit1j110bd";
```
DanielMSchmidt: This seems to be forgotten 🤔

packages/ui-kit-stage/HeaderBar.tsx
tnguyen0: Can you clarify? Are you talking about just the padding?

src/js/components/AccountDropdownTrigger.js
tnguyen0: :thumbsup:

src/js/components/AccountDropdownTrigger.js
tnguyen0: 👍

src/js/components/HeaderBar.tsx
tnguyen0: What do you mean exactly? We are removing the sign out from the sidebar.

src/js/components/HeaderBar.tsx
tnguyen0: yep, good catch

src/js/components/HeaderBar.tsx
tnguyen0: Yeah, my intention was it to be a placeholder for you to fill in.

src/js/components/HeaderBar.tsx
tnguyen0: Again, placeholder.

src/js/components/HeaderBar.tsx
Fabs: I mean the menu item, had a label called `Sign out` that we changed to the word `Logout`. I believe we should still call it `Sign out`.

src/js/components/HeaderBar.tsx
tnguyen0: 👍

src/js/stores/AuthStore.js
tnguyen0: Good call

src/js/components/HeaderBar.tsx
tnguyen0: @leemunroe what are your thoughts on this?

src/js/components/HeaderBar.tsx
tnguyen0: Yes, I can remove the cluster dropdown part of this PR. I added because I needed to see how the user dropdown would look/perform adjacent to another dropdown (and so you guys could check as well).

tnguyen0:
Thanks for all the feedback team!

@nLight I can't reproduce. I tried on Safari/Chrome/Firefox. Does the user/cluster name NOT shrink for anyone else?

src/js/components/HeaderBar.tsx
tnguyen0: `HeaderBar.tsx` is using `UIHeaderBar`, which is just the bar itself.

src/js/components/HeaderBar.tsx
leemunroe: Ah lets keep `Sign Out`. Oversight on my part. Not deliberate.

src/js/components/HeaderBar.tsx
juliangieseke: I know, but I think we will drop the `UiHeaderBar` component…

weblancaster:
@Fabs @nLight @juliangieseke it's working here.

screen recording 2018-07-16 at 03 31 pm

packages/ui-kit-stage/HeaderBar.tsx
weblancaster: @judithpatudith I believe this should be in the parent instead of the children that way creating this wrapper where whenever we have a new child the component will be positioned at the same place we expect.

If we add to the child whenever we replace/ or add elements to the edges we will need to re-apply the padding.

weblancaster: I asked @tnguyen0 to try using the cnvs grid system.

src/js/components/HeaderBar.tsx
tnguyen0: What do you mean drop? I thought we are just moving it into the right folder in dcos-ui. https://jira.mesosphere.com/browse/DCOS-39074

src/js/stores/AuthStore.js
```diff
@@ -96,6 +96,20 @@ class AuthStore extends GetSetBaseStore {
     }
   }
 
+  getUserLabel() {
+    const user = this.getUser();
+
```
weblancaster: This is nit-picking so no need to change.. but in some cases when declaring a variable there is no need to break to a new line.

tnguyen0: :angry:

packages/ui-kit-stage/HeaderBar.tsx
tnguyen0: Took a look at cnvs, didn't appear applicable to me.

tnguyen0:
@nLight @juliangieseke @Fabs @DanielMSchmidt 

Can you guys give another look?

Removed truncating parts and placeholders, it can now be found on the branch `tommy/feature/DCOS-38940-user-cluster-dropdown-ellipsis-test`

juliangieseke:
Hm, username in dropdown is still missing, or am I missing something and we decided not to do it? 🤔 
screen shot 2018-07-17 at 13 44 06

packages/ui-kit-stage/HeaderBar.tsx
juliangieseke: After looking at this again, I think it has to be changed… on my machine, the logo is centered instead of left-aligned. 
screen shot 2018-07-17 at 13 45 59

tnguyen0: Refer to testing steps. `<span>toggle component here</span>` needs to be commented out.

src/js/components/HeaderBar.tsx
tnguyen0: Fair enough. We can do all this for DCOS-39074.

packages/ui-kit-stage/HeaderBar.tsx
tnguyen0: Yeah, the fix would be to put the toggle component in a wrapper with the logo. This can be fixed in the UI issue and truncation ticket.

tnguyen0:
Thanks Lee. These UI issues will be accounted for in DCOS-38944.

src/js/components/HeaderBar.tsx
```diff
@@ -13,8 +15,7 @@ export default function HeaderBar() {
       <div className="header-bar-logo-wrapper">
         <div className="header-bar-logo" />
       </div>
-      <span>user-menu component here</span>
-      <span>cluster-menu component here</span>
```
DanielMSchmidt: you removed my placeholder 😡 

Just kidding, no need for it anymore ;)

src/js/stores/BaseStore.d.ts
```diff
@@ -0,0 +1,6 @@
+import { EventEmitter } from "events";
+
+export default class BaseStore extends EventEmitter {
+  addChangeListener: (eventName: string, callback: void) => void;
```
DanielMSchmidt: I think the type `void` is wrong here (at least judging from the [docs](https://www.typescriptlang.org/docs/handbook/basic-types.html#void). I think you want to express something like 

```diff
-addChangeListener: (eventName: string, callback: void) => void;
+addChangeListener: (eventName: string, callback: () => void) => void;
```

packages/ui-kit-stage/HeaderBar.tsx
juliangieseke: just dont do `justifyContent: "space-between"` :D 

Can we please remove the styles from Headerbar and fix that in DCOS-38944? 🙇

src/js/components/AccountDropdownTrigger.js
```diff
@@ -0,0 +1,43 @@
+import mixin from "reactjs-mixin";
+import PropTypes from "prop-types";
+import React from "react";
+import { StoreMixin } from "mesosphere-shared-reactjs";
+
+class AccountDropdownTrigger extends mixin(StoreMixin) {
```
juliangieseke: :art: I think @DanielMSchmidt removed the store from his PR, shall we also remove it here? 🤔 (only applicable if he did, not blocking)

src/js/components/HeaderBar.tsx
juliangieseke: actually - since you touch this PR anyway, can you please revert this change? :) 
reason is, that it will cause a conflict in daniels PR once this one is merged… (the import also might cause a conflict tho 🤔 )

absolutely **not** blocking

tnguyen0:
@juliangieseke @weblancaster @DanielMSchmidt should be good to go

src/js/components/AccountDropdownTrigger.js
tnguyen0: remove, he did not

PR #3076, opened 2018-07-16
DCOS_OSS-3801: 1.11 / Support Docker Parameters in Jobs

nLight:
Support docker parameters in Metronome jobs.
This is based on @kensipe PR https://github.com/dcos/dcos-ui/pull/3037

Closes DCOS_OSS-3801

## Testing

Test job JSON that UI shouldn't change when you switch back and forward between JSON view and Form view.
```js
{
  "description": "Example Application",
  "id": "docker-param",
  "run": {
    "cmd": "sleep inf",
    "cpus": 0.2,
    "mem": 32,
    "docker": {
      "image": "ubuntu",
      "parameters": [
        {
          "key": "cap-drop",
          "value": "ALL"
        },
        {
          "key": "cap-add",
          "value": "SYSLOG"
        }
      ]
    }
  }
}
```

### DIFF
--- a/src/js/schemas/job-schema/Docker.js
+++ b/src/js/schemas/job-schema/Docker.js
@@ -1,7 +1,3 @@
-/* eslint-disable no-unused-vars */
-import React from "react";
-/* eslint-enable no-unused-vars */
-
 const General = {
   title: "Docker Container",
   description: "Configure your job settings",
@@ -13,6 +9,13 @@ const General = {
       type: "string",
       getter(job) {
         return job.getDocker().image;
+      }
+    },
+    privileged: {
+      type: "boolean",
+      description: "Run this docker image in privileged mode",
+      getter(job) {
+        return job.getDocker().privileged;
       }
     }
   },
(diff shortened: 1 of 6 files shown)


nLight:
Looks like Metronome support isn't there in 1.11 yet thus system test failure. @kensipe would you please clarify which versions of DCOS gained the feature. Right now it looks like we have a regression from 1.9 to later versions.

src/js/schemas/job-schema/Docker.js
```diff
@@ -14,6 +10,13 @@ const General = {
       getter(job) {
         return job.getDocker().image;
       }
+    },
+    privileged: {
```
tnguyen0: Nit: title: "Privileged" for capitalize

src/js/schemas/job-schema/Parameters.js
```diff
@@ -0,0 +1,29 @@
+const Parameters = {
```
tnguyen0: Nit: Maybe rename file to DockerParameters

PR #3082, opened 2018-07-18
DCOS-39112: fix comparison in nodes table

DanielMSchmidt:
## Testing

```diff
-import NodesTable from "../../../components/NodesTable";
+import NodesTable from "../../../components/NodesTable-new";
```

Please take a look at the nodes table and see if all the sortings look correct (especially the ones with numbers, like mem, cpu, and disk.

## Trade-offs

- I decided to go for a bigger solution in this sorting problem as I saw multiple sorting issues in the past and having one place to look at when it comes to sorting seems like a good idea to me.
- I decided to write design the API of the sorting package in a very composable way, which means there might be a bit of extra typing for some use cases, just because the sort is not smart and tries not to guess.
- I decided to allow an unlimited amount of cascading tie breakers, because the implementation was not harder and I thought there might be benefits in hard to sort columns down the road
- I decided to name all exports which do export comparators exactly like this as the beginning of an effort to unify this and have a uniform interface for each column.
- I decided to not let the default sort of the package be the same as the one of `Array.prototype.sort` because that would add some "magic" to the API, IMHO

## Dependencies

None

### DIFF
--- a/packages/@dcos/sorting/LICENSE
+++ b/packages/@dcos/sorting/LICENSE
@@ -1,0 +1,201 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
+      control with that entity. For the purposes of this definition,
+      "control" means (i) the power, direct or indirect, to cause the
+      direction or management of such entity, whether by contract or
+      otherwise, or (ii) ownership of fifty percent (50%) or more of the
+      outstanding shares, or (iii) beneficial ownership of such entity.
+
+      "You" (or "Your") shall mean an individual or Legal Entity
+      exercising permissions granted by this License.
+
+      "Source" form shall mean the preferred form for making modifications,
+      including but not limited to software source code, documentation
+      source, and configuration files.
+
+      "Object" form shall mean any form resulting from mechanical
+      transformation or translation of a Source form, including but
+      not limited to compiled object code, generated documentation,
+      and conversions to other media types.
+
+      "Work" shall mean the work of authorship, whether in Source or
+      Object form, made available under the License, as indicated by a
+      copyright notice that is included in or attached to the work
+      (an example is provided in the Appendix below).
+
+      "Derivative Works" shall mean any work, whether in Source or Object
+      form, that is based on (or derived from) the Work and for which the
+      editorial revisions, annotations, elaborations, or other modifications
+      represent, as a whole, an original work of authorship. For the purposes
+      of this License, Derivative Works shall not include works that remain
+      separable from, or merely link (or bind by name) to the interfaces of,
+      the Work and Derivative Works thereof.
+
+      "Contribution" shall mean any work of authorship, including
+      the original version of the Work and any modifications or additions
+      to that Work or Derivative Works thereof, that is intentionally
+      submitted to Licensor for inclusion in the Work by the copyright owner
+      or by an individual or Legal Entity authorized to submit on behalf of
+      the copyright owner. For the purposes of this definition, "submitted"
+      means any form of electronic, verbal, or written communication sent
+      to the Licensor or its representatives, including but not limited to
+      communication on electronic mailing lists, source code control systems,
+      and issue tracking systems that are managed by, or on behalf of, the
+      Licensor for the purpose of discussing and improving the Work, but
+      excluding communication that is conspicuously marked or otherwise
+      designated in writing by the copyright owner as "Not a Contribution."
+
+      "Contributor" shall mean Licensor and any individual or Legal Entity
+      on behalf of whom a Contribution has been received by Licensor and
+      subsequently incorporated within the Work.
+
+   2. Grant of Copyright License. Subject to the terms and conditions of
+      this License, each Contributor hereby grants to You a perpetual,
+      worldwide, non-exclusive, no-charge, royalty-free, irrevocable
+      copyright license to reproduce, prepare Derivative Works of,
+      publicly display, publicly perform, sublicense, and distribute the
+      Work and such Derivative Works in Source or Object form.
+
+   3. Grant of Patent License. Subject to the terms and conditions of
+      this License, each Contributor hereby grants to You a perpetual,
+      worldwide, non-exclusive, no-charge, royalty-free, irrevocable
+      (except as stated in this section) patent license to make, have made,
+      use, offer to sell, sell, import, and otherwise transfer the Work,
+      where such license applies only to those patent claims licensable
+      by such Contributor that are necessarily infringed by their
+      Contribution(s) alone or by combination of their Contribution(s)
+      with the Work to which such Contribution(s) was submitted. If You
+      institute patent litigation against any entity (including a
+      cross-claim or counterclaim in a lawsuit) alleging that the Work
+      or a Contribution incorporated within the Work constitutes direct
+      or contributory patent infringement, then any patent licenses
+      granted to You under this License for that Work shall terminate
+      as of the date such litigation is filed.
+
+   4. Redistribution. You may reproduce and distribute copies of the
+      Work or Derivative Works thereof in any medium, with or without
+      modifications, and in Source or Object form, provided that You
+      meet the following conditions:
+
+      (a) You must give any other recipients of the Work or
+          Derivative Works a copy of this License; and
+
+      (b) You must cause any modified files to carry prominent notices
+          stating that You changed the files; and
+
+      (c) You must retain, in the Source form of any Derivative Works
+          that You distribute, all copyright, patent, trademark, and
+          attribution notices from the Source form of the Work,
+          excluding those notices that do not pertain to any part of
+          the Derivative Works; and
+
+      (d) If the Work includes a "NOTICE" text file as part of its
+          distribution, then any Derivative Works that You distribute must
+          include a readable copy of the attribution notices contained
+          within such NOTICE file, excluding those notices that do not
+          pertain to any part of the Derivative Works, in at least one
+          of the following places: within a NOTICE text file distributed
+          as part of the Derivative Works; within the Source form or
+          documentation, if provided along with the Derivative Works; or,
+          within a display generated by the Derivative Works, if and
+          wherever such third-party notices normally appear. The contents
+          of the NOTICE file are for informational purposes only and
+          do not modify the License. You may add Your own attribution
+          notices within Derivative Works that You distribute, alongside
+          or as an addendum to the NOTICE text from the Work, provided
+          that such additional attribution notices cannot be construed
+          as modifying the License.
+
+      You may add Your own copyright statement to Your modifications and
+      may provide additional or different license terms and conditions
+      for use, reproduction, or distribution of Your modifications, or
+      for any such Derivative Works as a whole, provided Your use,
+      reproduction, and distribution of the Work otherwise complies with
+      the conditions stated in this License.
+
+   5. Submission of Contributions. Unless You explicitly state otherwise,
+      any Contribution intentionally submitted for inclusion in the Work
+      by You to the Licensor shall be under the terms and conditions of
+      this License, without any additional terms or conditions.
+      Notwithstanding the above, nothing herein shall supersede or modify
+      the terms of any separate license agreement you may have executed
+      with Licensor regarding such Contributions.
+
+   6. Trademarks. This License does not grant permission to use the trade
+      names, trademarks, service marks, or product names of the Licensor,
+      except as required for reasonable and customary use in describing the
+      origin of the Work and reproducing the content of the NOTICE file.
+
+   7. Disclaimer of Warranty. Unless required by applicable law or
+      agreed to in writing, Licensor provides the Work (and each
+      Contributor provides its Contributions) on an "AS IS" BASIS,
+      WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+      implied, including, without limitation, any warranties or conditions
+      of TITLE, NON-INFRINGEMENT, MERCHANTABILITY, or FITNESS FOR A
+      PARTICULAR PURPOSE. You are solely responsible for determining the
+      appropriateness of using or redistributing the Work and assume any
+      risks associated with Your exercise of permissions under this License.
+
+   8. Limitation of Liability. In no event and under no legal theory,
+      whether in tort (including negligence), contract, or otherwise,
+      unless required by applicable law (such as deliberate and grossly
+      negligent acts) or agreed to in writing, shall any Contributor be
+      liable to You for damages, including any direct, indirect, special,
+      incidental, or consequential damages of any character arising as a
+      result of this License or out of the use or inability to use the
+      Work (including but not limited to damages for loss of goodwill,
+      work stoppage, computer failure or malfunction, or any and all
+      other commercial damages or losses), even if such Contributor
+      has been advised of the possibility of such damages.
+
+   9. Accepting Warranty or Additional Liability. While redistributing
+      the Work or Derivative Works thereof, You may choose to offer,
+      and charge a fee for, acceptance of support, warranty, indemnity,
+      or other liability obligations and/or rights consistent with this
+      License. However, in accepting such obligations, You may act only
+      on Your own behalf and on Your sole responsibility, not on behalf
+      of any other Contributor, and only if You agree to indemnify,
+      defend, and hold each Contributor harmless for any liability
+      incurred by, or claims asserted against, such Contributor by reason
+      of your accepting any such warranty or additional liability.
+
+   END OF TERMS AND CONDITIONS
+
+   APPENDIX: How to apply the Apache License to your work.
+
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "{}"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright {yyyy} {name of copyright owner}
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
--- a/packages/@dcos/sorting/README.md
+++ b/packages/@dcos/sorting/README.md
@@ -1,0 +1,59 @@
+# DC/OS Sorting
+
+This project provides elementary sorting building blocks, which we use in the DC/OS UI.
+
+## Example
+
+```ts
+import {
+  sort,
+  reverseComparator,
+  compareNumber,
+  compareString
+} from "@dcos/sorting";
+
+function sortMyTable(data: Node[], isReversed) {
+  function getHostname(node: Node): string {
+    if (!node.host) {
+      return null;
+    }
+
+    return node.host.name;
+  }
+
+  function getCpuUsage(node: Node): number {
+    return node.cpuUsage;
+  }
+
+  return sort<Node>(
+    [compareString(getHostname), compareNumber(getCpuUsage)].map(
+      comparator => (isReversed ? reverseComparator(comparator) : comparator)
+    ),
+    data
+  );
+}
+```
+
+## Combinators
+
+### `sort(comparatorsArray)`
+
+Sorts by the first comparator. If it is neutral (returns 0), sorts by the second. If it is neutral, ...
+
+### `reverseComparator(comparatorFunction)`
+
+Takes a comparator and returns a new comparator that does the opposite
+
+## Comparators
+
+### `compareString(getter)`
+
+Returns a function that uses the getter to get the values to be compared as string. It expects the values to be retrieved to be of type string, otherwise unexpected behaviour may occur.
+
+### `compareNumber(getter)`
+
+Returns a function that uses the getter to get the values to be compared as number. It expects the values to be retrieved to be of type number, otherwise unexpected behaviour may occur.
+
+## When to add a new comparator
+
+If you have a comparator which you want to use in DC/OS UI please make sure to add it here if it resembles a fundamental type. Examples for this would be number, string, health status, IP. Non-examples would be a node, a task or a service.
--- a/packages/@dcos/sorting/__tests__/compare-as-number-test.ts
+++ b/packages/@dcos/sorting/__tests__/compare-as-number-test.ts
@@ -4,7 +4,7 @@ const idLense = (x: any) => x;
 
 describe("compareNumber", () => {
   it("compares number", () => {
-    expect(compareNumber(idLense)(5, 8)).toBe(ComparisonResult.SMALLER_THAN);
+    expect(compareNumber(idLense)(5, 8)).toBe(ComparisonResult.LESS_THAN);
     expect(compareNumber(idLense)(8, 5)).toBe(ComparisonResult.GREATER_THAN);
     expect(compareNumber(idLense)(5, 5)).toBe(ComparisonResult.EQUAL);
   });
--- a/packages/@dcos/sorting/__tests__/compare-as-number-test.ts
+++ b/packages/@dcos/sorting/__tests__/compare-as-number-test.ts
@@ -1,0 +1,20 @@
+import { compareNumber, ComparisonResult } from "../";
+
+const idLense = (x: any) => x;
+
+describe("compareNumber", () => {
+  it("compares number", () => {
+    expect(compareNumber(idLense)(5, 8)).toBe(ComparisonResult.SMALLER_THAN);
+    expect(compareNumber(idLense)(8, 5)).toBe(ComparisonResult.GREATER_THAN);
+    expect(compareNumber(idLense)(5, 5)).toBe(ComparisonResult.EQUAL);
+  });
+
+  it("executes the lense before comparing", () => {
+    const mockLense = jest.fn();
+    const comparator = compareNumber(mockLense);
+    expect(mockLense).not.toHaveBeenCalled();
+
+    comparator(10, 100);
+    expect(mockLense).toHaveBeenCalledTimes(2);
+  });
+});
--- a/packages/@dcos/sorting/__tests__/compare-as-string-test.ts
+++ b/packages/@dcos/sorting/__tests__/compare-as-string-test.ts
@@ -4,9 +4,7 @@ const idLense = (x: string) => x;
 
 describe("comparString", () => {
   it("compares strings", () => {
-    expect(compareString(idLense)("a", "b")).toBe(
-      ComparisonResult.SMALLER_THAN
-    );
+    expect(compareString(idLense)("a", "b")).toBe(ComparisonResult.LESS_THAN);
     expect(compareString(idLense)("b", "a")).toBe(
       ComparisonResult.GREATER_THAN
     );
--- a/packages/@dcos/sorting/__tests__/compare-as-string-test.ts
+++ b/packages/@dcos/sorting/__tests__/compare-as-string-test.ts
@@ -1,0 +1,24 @@
+import { compareString, ComparisonResult } from "../";
+
+const idLense = (x: string) => x;
+
+describe("comparString", () => {
+  it("compares strings", () => {
+    expect(compareString(idLense)("a", "b")).toBe(
+      ComparisonResult.SMALLER_THAN
+    );
+    expect(compareString(idLense)("b", "a")).toBe(
+      ComparisonResult.GREATER_THAN
+    );
+    expect(compareString(idLense)("b", "b")).toBe(ComparisonResult.EQUAL);
+  });
+
+  it("executes the lense before comparing", () => {
+    const mockLense = jest.fn().mockReturnValue("42");
+    const comparator = compareString(mockLense);
+    expect(mockLense).not.toHaveBeenCalled();
+
+    comparator("10", "100");
+    expect(mockLense).toHaveBeenCalledTimes(2);
+  });
+});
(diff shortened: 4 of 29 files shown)


packages/@dcos/sorting/LICENSE
```diff
@@ -0,0 +1,201 @@
+                                 Apache License
```
Fabs: Have we considered the benefits of it being an npm package vs the extra work that is to maintain that? I like the idea of it to belong to its on package, not sure if I really like it being its own package.

packages/@dcos/sorting/__tests__/compare-as-number-test.ts
```diff
@@ -0,0 +1,34 @@
+import { compareAsNumber, ComparisonResult } from "../";
+
+const idLense = (x: any) => x;
+
+describe("compareAsNumber", () => {
+  it("returns 0 if any on the values is null", () => {
```
Fabs: What about making `null` and `undefined` be 0 or Inf while sorting numbers?

The current solution does not seem like it works for me.
```js
[1,null,2, null, 2, 1].sort((a,b) => a == null || b == null ? 0 : a - b)
> [ 1, null, 2, null, 1, 2 ]
```

*short hypothesis*
The reason would be null being equivalent to any other value they never move.

*long hypothesis*

making `1 > null be 0`, and `1 < null also be 0` breaks antisymmetry `if a ≤ b and b ≤ a, then a = b` which makes our comparators not being a Partial Order [0].

Failing to be a partial order means that `null` will be equal to every value, making null never sort. 

[0] - https://en.wikipedia.org/wiki/Partially_ordered_set

packages/@dcos/sorting/index.ts
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  return a < b ? ComparisonResult.SMALLER : ComparisonResult.BIGGER;
+}
+
+function compareNumericValues(a: any, b: any): ComparisonResult {
+  const aNumber = parseInt(a, 10);
+  const bNumber = parseInt(b, 10);
+
+  if (isNaN(aNumber) || isNaN(bNumber) || aNumber === bNumber) {
```
Fabs: I think we should have tests mixing NaN, null, and undefined in the list.

packages/@dcos/sorting/index.ts
```diff
@@ -0,0 +1,61 @@
...
+        return result;
+      }
+    }
+
+    return 0;
+  });
+}
+
+export function reverseComparator<T>(comparator: Comparator<T>): Comparator<T> {
+  return (a, b) => {
```
Fabs: Why not just ?
return (a,b) => -1 * comparator(a,b);

packages/@dcos/sorting/__tests__/compare-as-string-test.ts
```diff
@@ -0,0 +1,30 @@
+import { compareAsString, ComparisonResult } from "../";
+
+const idLense = (x: any) => x;
+
+describe("compareAsString", () => {
+  it("returns 0 if any on the values is null", () => {
```
Fabs: Same issue with strings I guess. 

Why not null being "" here?

packages/@dcos/sorting/__tests__/compare-as-string-test.ts
```diff
@@ -0,0 +1,30 @@
...
+    expect(compareAsString(idLense)("n", null)).toBe(ComparisonResult.EQUAL);
+  });
+
+  it("compares strings", () => {
+    expect(compareAsString(idLense)("a", "b")).toBe(ComparisonResult.SMALLER);
+    expect(compareAsString(idLense)("b", "a")).toBe(ComparisonResult.BIGGER);
+    expect(compareAsString(idLense)("b", "b")).toBe(ComparisonResult.EQUAL);
+  });
+
+  it("converts unknown type to strings before comparing", () => {
```
Fabs: I think a more meaningful test could be:
`"9" > "100" is bigger` because than it contradicts common knowledge

packages/@dcos/sorting/__tests__/sort-test.ts
```diff
@@ -0,0 +1,40 @@
+import { sort } from "../";
+
+describe("sort", () => {
+  it("does not mutate the input array", () => {
```
Fabs: 💘

packages/@dcos/sorting/index.ts
```diff
@@ -0,0 +1,61 @@
...
+  BIGGER = 1,
+  EQUAL = 0,
+  SMALLER = -1
+}
+export type Comparator<T> = (a: T, b: T) => ComparisonResult;
+export type Lense<T> = (data: T) => any;
+
+export function sort<T>(comps: Array<Comparator<T>>, data: T[]): T[] {
+  return [...data].sort((a, b) => {
+    for (const comperator of comps) {
```
Fabs: ✍️ "comparator"

packages/@dcos/sorting/index.ts
```diff
@@ -0,0 +1,61 @@
...
+      const result = comperator(a, b);
+
+      if (result !== 0) {
+        return result;
+      }
+    }
+
+    return 0;
```
Fabs: Why not default to the javascript default sort behavior? (which I believe is the UNICODE codepoint value of T converted to string).

packages/@dcos/sorting/index.ts
```diff
@@ -0,0 +1,61 @@
+export enum ComparisonResult {
+  BIGGER = 1,
+  EQUAL = 0,
+  SMALLER = -1
+}
+export type Comparator<T> = (a: T, b: T) => ComparisonResult;
+export type Lense<T> = (data: T) => any;
```
Fabs: I really like the incorporation of lenses to zoom in the particular attribute to be sorted.

packages/@dcos/sorting/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+Returns a function that uses the attributeLense to get the values to be compared as strings
+
+### `compareAsNumber(attributeLense)`
+
+Returns a function that uses the attributeLense to get the values to be compared as number
+
+## When to add a new comparator
+
+If you have a comparator which you want to use in DC/OS UI please make sure to add it here if it resembles a fundamental type. Examples for this would be number, string, health status, IP. Non-examples would be a node, a task or a service.
```
Fabs: Have you considered drawing this line a little lower? Like base types being only those from JS, and IP, Health and everything domain specific on DCOS being at the same place as Task, Service and etc...

Fabs:
Overall I would say great work :D.

plugins/nodes/src/js/columns/NodesTableCPUColumn.tsx
```diff
@@ -1,24 +1,34 @@
...
+
 export function cpuSorter(data: Node[], sortDirection: SortDirection): Node[] {
-  const sortedData = data.sort((a, b) =>
-    compareValues(
-      a.get("used_resources").cpus.toString(),
-      b.get("used_resources").cpus.toString(),
-      a.getHostName().toLowerCase(),
-      b.getHostName().toLowerCase()
-    )
+  return sort(
```
Fabs: Here were are tieing sort and compare together, in a way that the cpuComparator done here is not reusable in other parts of the system. Have you considered moving the lenses and the comparator closer to the type, or at least out of the column?

packages/@dcos/sorting/__tests__/compare-as-number-test.ts
Fabs: I said something wrong. We actually need a Total Order for sorting (https://en.wikipedia.org/wiki/Total_order). Total Order also needs to respect `asymmetry`.

packages/@dcos/sorting/LICENSE
DanielMSchmidt: Yes, I think having it as an extra pain will be more pain, but the right kind of pain. I hope that if we aggressively split out packages we will be forced to improve our tooling so that spinning up a new package will become pretty trivial. I also think someone else could use this sorting pretty well 🤷‍♂️

packages/@dcos/sorting/__tests__/compare-as-number-test.ts
DanielMSchmidt: I don't think it breaks antisymmetry, to be honest. the value zero expresses equality, so it's more like everything is equal to null, which I think is okay.

packages/@dcos/sorting/index.ts
DanielMSchmidt: 👍

packages/@dcos/sorting/__tests__/compare-as-string-test.ts
DanielMSchmidt: Had a hard time coming up with tests, but I like this one ;)

packages/@dcos/sorting/index.ts
DanielMSchmidt: My reasoning was if I sort by nothing I expect no change.

packages/@dcos/sorting/index.ts
DanielMSchmidt: Had a discussion with @juliangieseke about this and we did not want to spread `-0` values into our system as we don't know if this might have side-effects. I know, there is no `-0` mathematically, but in JS there is.

packages/@dcos/sorting/index.ts
DanielMSchmidt: With tests you mean unit tests and not checks, right?

plugins/nodes/src/js/columns/NodesTableCPUColumn.tsx
DanielMSchmidt: I decided to not move it somewhere else as it handles the sorting for this column. I don't want to have a global "this sorts by CPU" functionality, because I think it is prone to be extended to 🍝 code. It's also not so much code that I really see a point in drying it up yet, but feel free to convince me :)

packages/@dcos/sorting/index.ts
Fabs: I see. I think `-0` would not be an issue as:
`-0` === '`0`. -0 also makes sense mathematically, it is just rarely written. 

But I  understand not wanting to introduce the confusion so 👍

packages/@dcos/sorting/index.ts
Fabs: I think my expectation would actually be that I can call `sort([1,4,5])` without any comparator. but maybe that does not make sense on this library.

packages/@dcos/sorting/index.ts
Fabs: More as some funny things about -0. https://en.wikipedia.org/wiki/Signed_zero

packages/@dcos/sorting/index.ts
juliangieseke: 🔡  _Lens_ 😅

packages/@dcos/sorting/index.ts
```diff
@@ -0,0 +1,61 @@
...
+
+export function sort<T>(comps: Array<Comparator<T>>, data: T[]): T[] {
```
juliangieseke: 🎨 I like `comparators` more than `comps` here

packages/@dcos/sorting/LICENSE
DanielMSchmidt: `sortBy` is a great tool for sorting an array of objects by a property, but I think it doesn't quite fit our use case here. We wanted to have tiebreakers in our sorting and we also wanted to support sorting for different data types which are not only JS data types. For me, it would be a replacement for `Array.prototype.sort`, but it might be that I misunderstand the API

packages/@dcos/sorting/README.md
```diff
@@ -0,0 +1,59 @@
+# DC/OS Sorting
+
+This project provides elementar sorting building blocks, which we use in the DC/OS UI.
```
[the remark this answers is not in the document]
DanielMSchmidt: 👍

packages/@dcos/sorting/README.md
DanielMSchmidt: @Fabs As we are moving away from structs I don't really see a point in adding the comparators there. What would be the benefit? 

@nLight I don't really understand, can you explain this a bit more? Do you think we should add a section of when to add something for DC/OS UI in general?

nLight: @DanielMSchmidt How a developers would know they should add a comparator in a sorting lib if they don't know about its existence? I mean we probably should add a line about the lib into https://github.com/dcos/dcos-ui/blob/master/CONTRIBUTING.md

packages/@dcos/sorting/__tests__/compare-as-number-test.ts
DanielMSchmidt: Maybe I should check with product what should be the case. the null behavior is what we currently have, but that, of course, does not mean we need to keep it :)

I will follow up on this and inform you ;)

packages/@dcos/sorting/__tests__/compare-as-string-test.ts
DanielMSchmidt: > Why not null being "" here?

If you mean `"null"` I would say that is a perfectly valid string

packages/@dcos/sorting/index.ts
DanielMSchmidt: 👍

packages/@dcos/sorting/README.md
DanielMSchmidt: Ah, get you now 👍 will drop in a line

packages/@dcos/sorting/__tests__/compare-as-string-test.ts
Fabs: Nope, I mean the same problem for the numbers. If you say null is the same as any value then sort does not work as expected `also for the other values on the list`. 

I think whatever solves the other comment, solves that one by analogy.

packages/@dcos/sorting/__tests__/compare-as-number-test.ts
Fabs: ok.

plugins/nodes/src/js/columns/NodesTableCPUColumn.tsx
Fabs: I would be ok with a private function on this module.

packages/@dcos/sorting/__tests__/compare-as-string-test.ts
DanielMSchmidt: @juliangieseke and I had some discussions around this and after this discussion I think we should just not support sorting by null at all. Let's make sure the lens always returns the right type and we should be good to go.

plugins/nodes/src/js/columns/NodesTableCPUColumn.tsx
DanielMSchmidt: I can't follow you. Do you mean typescript private or something like `_sortByHealth`? What is the problem with making this public?

Fabs:
```ts
sort(cpuComparator(sortDirection), data)

function cpuComparator(sortDirection): Comparator<Node> {
  return directionAwareComparators(
      [compareAsNumber(getCpuUsage), compareAsString(getHostName)],
      sortDirection
    );
}
```

My proposal is to separate comparison, from sorting into its own functions. What do you think of this code? You can leave it here, but you can also move somewhere else if the need to compare emerges as part of the type somewhere else.

DanielMSchmidt: I don't really understand why this is significantly better to you or why it matters to you in general.

packages/@dcos/sorting/__tests__/compare-as-string-test.ts
Fabs: So if I understood you correctly, the lens should look at null and give it a meaning inside that type. Which I decide. Could be 0 for Numbers and "" for Strings If I so choose. Is that what you mean?

Fabs: Also, will it be any runtime guarantees?

plugins/nodes/src/js/columns/NodesTableCPUColumn.tsx
Fabs: Separate comparison from sort, which are for me different concerns. I would go as to say that running the sort could even be in the table, and not in the column. 

If you think on the Strategy pattern terms, this function couple Strategy and Algorithm tightly so Strategy is not reusable.

But on any case, I don't really think it is important, so if you still disagree lets just keep it this way.

packages/@dcos/sorting/README.md
Fabs: @DanielMSchmidt 👍

plugins/nodes/src/js/columns/NodesTableCPUColumn.tsx
DanielMSchmidt: I see your point with having the sort on the table, I would be completely fine with just exporting a set of comparators and that's it :)

Can you explain to me why it couples strategy and algorithm? We have the sorting algorithms split out, the only "algorithm" we still have is for extracting data from the node struct as far as I understand it.

Thank you for explaining it again :)

packages/@dcos/sorting/index.ts
Fabs: What about this point?

plugins/nodes/src/js/components/NodesTable-new.tsx
```diff
@@ -95,28 +97,30 @@ export default class NodesTable extends React.Component<
...
+        return memComparators;
       case "disk":
-        return diskSorter;
+        return diskComparators;
       case "gpu":
-        return gpuSorter;
+        return gpuComparators;
       default:
-        return (data, _sortDirection) => data;
+        return [equalComparator];
```
Fabs: Could it be a potential source of confusion that whenever you have a single comparator you have to wrap it on a list? 

Can we now, or on future work, provide an API that can detect a single function and wrap it on the array for you?

packages/@dcos/sorting/index.ts
DanielMSchmidt: Good point, I would like to note this down as a trade-off.

packages/@dcos/sorting/README.md
```diff
@@ -0,0 +1,59 @@
+# DC/OS Sorting
+
+This project provides elementary sorting building blocks, which we use in the DC/OS UI.
+
+## Example
+
+```ts
+import {
```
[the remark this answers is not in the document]
DanielMSchmidt: I updated it. It won't deal with null, we expect the getters to deliver us a number to have a strong contract here.

packages/@dcos/sorting/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+Sorts by the first comparator. If it is neutral (returns 0), sorts by the second. If it is neutral, ...
+
+### `reverseComparator(comparatorFunction, shouldBeReversed)`
+
+Takes a comparator and returns a new comparator that does the opposite
+
+## Comparators
+
+### `compareAsString(attributeLense)`
```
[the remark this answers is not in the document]
DanielMSchmidt: Will do, thanks for catching 👍

packages/@dcos/sorting/index.ts
```diff
@@ -0,0 +1,65 @@
...
+    } else {
+      return (-1 * previousResult) as ComparisonResult;
+    }
+  };
+}
+
+function compareStringValues(a: string, b: string): ComparisonResult {
+  // Does not always return -1 or 1: https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare#Description
+  const result = a.localeCompare(b);
+  return result > 0
```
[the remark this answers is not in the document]
DanielMSchmidt: I don't think so.

plugins/nodes/src/js/components/NodesTable-new.tsx
DanielMSchmidt: I originally wanted to use a spread to do the trick, but I found the API to be kinda convoluted then. Would you prefer something like this instead?

```js
sort(comparatorA, comparatorB, comparatorC, data) // Option A
sort(comparatorA, comparatorB, comparatorC)(data) // Option B
```

On second thought, in this case here we could also write an empty array, would  you like this better?

packages/@dcos/sorting/index.ts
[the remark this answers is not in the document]
Fabs: I think Orlando mentioned it in one of the syncs...

plugins/nodes/src/js/components/NodesTable-new.tsx
Fabs: I think my ideas was more like, you cam pass a Function or a list of Function and the API does what you would expect. Just to solve for the single function case. 

For many comparators, I am fine with that being a list.

plugins/nodes/src/js/columns/NodesTableCPUColumn.tsx
Fabs: Sorry, missed that one. Sort, Min, Max are all algorithms in the sense they process a collection and return some meaningful information. But compareBy is the strategy they use to achieve their result. They usually have an implicit strategy supported, because they are used all the time, but all those 3 algorithms are equally configurable by passing them a comparison strategy.

```js
sort(["a","b","c"], byDictionaryOrder)
sort(["a","b","c"], byUnicodeOrder)
max(["a","b","c"], byDictionaryOrder)
//...
```


In that sense,  comparison is like a class, a trait. And every type can specify one or more ways to implement this trait, but it has nothing to do with sorting per se, although sort uses it

DanielMSchmidt: Ah, now I get you. Would you be fine with having them in a seperate package? I would have thought of "sorting" as "everything that you need to sort your data", but that might be too big :) We could do `@dcos/sort` and `@dcos/comparators` or maybe put the sort part into a Util or something the like. I would like to have them co-located a bit more, but I would be fine with this change

plugins/nodes/src/js/components/NodesTable-new.tsx
DanielMSchmidt: I don't feel really good about this. You trade `[]` on the call side with

Fabs: But don't you think as a package maintainer it would be great to make the best api for whoever uses your lib? I really feel like caring about creating a list of one makes the dev-user live harder.

What about having an union type? Would that give you the type guarantees you want `type comparator = Comparator<T> | Array<Comparator<T>>);`. We could combine that with a user-defined-type-guard to get the behavior of  `makeArrayIfItIsntAlready` bad at all. 

You could maybe do all of this by just introducing a nother function who offers this api, and smartly call the underlying sort with the Array<Comparator<T>>.

packages/@dcos/sorting/index.ts
Fabs: Alright.

packages/@dcos/sorting/index.ts
```diff
@@ -0,0 +1,65 @@
+export enum ComparisonResult {
+  BIGGER = 1,
```
nLight: Shall we stick with a well known terms like `LESS_THAN` and `GREATER_THAN`?  Or even `LT` `GT` https://en.wikipedia.org/wiki/Less-than_sign

packages/@dcos/sorting/index.ts
```diff
@@ -0,0 +1,65 @@
...
+  };
+}
+
+function compareStringValues(a: string, b: string): ComparisonResult {
+  // Does not always return -1 or 1: https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare#Description
+  const result = a.localeCompare(b);
+  return result > 0
+    ? ComparisonResult.BIGGER
+    : result < 0
+      ? ComparisonResult.SMALLER
```
nLight: I thought we were banning nested ternaries. https://github.com/dcos-labs/eslint-config/blob/master/index.js#L22 Please change

packages/@dcos/sorting/README.md
```diff
@@ -0,0 +1,59 @@
...
+    return node.host.name;
+  }
+
+  function getCpuUsage(node: Node): number {
+    return node.cpuUsage;
+  }
+
+  return sort<Node>(
+    [compareString(getHostname), compareNumber(getCpuUsage)].map(
+      comparator => reverseComparator(comparator, isReversed)
```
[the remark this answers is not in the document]
DanielMSchmidt: 👍 Thank you for catching :)

packages/@dcos/sorting/index.ts
```diff
@@ -0,0 +1,67 @@
+export enum ComparisonResult {
+  GREATER_THAN = 1,
+  EQUAL = 0,
+  SMALLER_THAN = -1
```
nLight: Let's stick to `LESS_THAN`

DanielMSchmidt:
@nLight I fixed that, sorry for the mistake 🙈

DanielMSchmidt:
@nLight Could you take another look? I just rebased from master :)

packages/@dcos/sorting/index.ts
```diff
@@ -0,0 +1,67 @@
...
+  EQUAL = 0,
+  LESS_THAN = -1
+}
+export type Comparator<T> = (a: T, b: T) => ComparisonResult;
+export type Getter<I, T> = (data: I) => T;
+type compareWithGetter<T> = <I>(getter: Getter<I, T>) => Comparator<I>;
+
+export function sort<T>(comparators: Array<Comparator<T>>, data: T[]): T[] {
+  return [...data].sort((a, b) => {
+    for (const comparator of comparators) {
```
juliangieseke: not blocking, but please consider thinking about for a second: 
I'd love to get rid of this for loop:

```
[...data].sort((a,b) => sortByComparators(a, b, comparators))
```

```
function sortByComparators<T>(a: T, b: T, comparators: Array<Comparator<T>>): ComparisonResult {
  if(comparators.length === 0) return 0;
  return (comparators[0](a, b) === 0) ? 0 : sortByComparators(a, b, comparators.slice(1));
}
```

plugins/nodes/src/js/components/NodesTable-new.tsx
juliangieseke: I second @Fabs here, having the possibility to call `sort(comparator, data)` could be beneficial🤔 

assuming my change above, we could do something like this:

```
export function sort<T>(comparators: Array<Comparator<T>> | Comparator<T>, data: T[]): T[] {
  return (Array.isArray(comparators)) 
    ? [...data].sort((a,b) => sortByComparators(a, b, comparators)) 
    : [...data].sort(comparator);
}
```
we also could export two functions if you want to keep the interface strict: `sortByComparator` and `sortByComparators`:
```
export function sortByComparator<T>(comparator: Comparator<T>, data: T[]): T[] {
  return [...data].sort(comparator);
}
export function sortByComparators<T>(comparators: Array<Comparator<T>>, data: T[]): T[] {
  return [...data].sort((a,b) => sortByComparators(a, b, comparators))
}
```
(again, these are assuming my change from above)

plugins/nodes/src/js/components/NodesTable-new.tsx
```diff
@@ -1,46 +1,50 @@
...
 import { SortableColumnHeader } from "ui-kit-stage/SortableColumnHeader";
-import { SortDirection } from "../types/SortDirection";
+import {
+  SortDirection,
+  directionAwareComparators
+} from "../types/SortDirection";
 
 import {
-  hostnameSorter,
+  comparators as hostnameComparators,
```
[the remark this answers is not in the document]
DanielMSchmidt: Sorry, I did this refactoring too early, thinking about the next step already. Will try to be more mindful next time 👍

plugins/nodes/src/js/components/NodesTable-new.tsx
DanielMSchmidt: > But don't you think as a package maintainer it would be great to make the best api for whoever uses your lib?

DanielMSchmidt: > But don't you think as a package maintainer it would be great to make the best api for whoever uses your lib? 

Yes I do think so, but the best API IMHO is not the one with the most indirection, but with the clearest and most concise API. I am aware that my opinion here is pretty dogmatic, but I don't see why cases like this should be handled here. If you argue I would like to be able to call it with only a comparator then we could also argue we want it to be able to handle null or undefined, because I don't want to check if my variable exists first. 

I chose this trade-off because I wanted to have the simplest API possible, which does not need a lot of documentation to understand it. Just give it a list of comparators and that's it.

packages/@dcos/sorting/index.ts
DanielMSchmidt: Why do you want to get rid of this for loop?

packages/@dcos/sorting/LICENSE
weblancaster: I would rather see us using something like Lodash than having to maintain this in the long run.

packages/@dcos/sorting/index.ts
juliangieseke: Loops are error prone, adding complexity and are hard to reason about - even the simple ones. Also being able to think of a loop as recursive function calls makes it a lot easier to write complete, secure and generally "good" code. 

there are two general best practices violated in this particular loop:
1. never break a loop (instead write your condition so that it stops at the correct time)
2. never have if clauses (*) inside loops (again, write your loop condition appropriately)

* early exists might be okay in some cases (i know that this team likes them - personally I dont)

DanielMSchmidt: I honestly think that in this specific case adding a new function and having a recursion to traverse a list makes it harder to reason about the code. 

The for loop is pretty clear, because the if is an early exit. 

But I don't care enough, I'll just rewrite it, whatever 🤷‍♂️

juliangieseke: > The for loop is pretty clear, because the if is an early exit.

well, this is the reason why I dont like early exits.
```
function this_is_ok(bar) {
  if(!bar) return 1;
}
function this_is_not(bar) {
  const baz = do_weird_shit_with(bar);
  if(!baz) return 1;
}
```

> But I don't care enough, I'll just rewrite it, whatever 🤷‍♂️
Awesome…

DanielMSchmidt: I still see no problem with this example, but as I said, going to change it.

DanielMSchmidt:
@Fabs @juliangieseke Funny thing about the "should be callable with array or with a single comparator" discussion we had. I was kinda asleep while writing the last code, there is currently no use case for a single comparator 🙈

DanielMSchmidt:
@Fabs @juliangieseke @nLight @weblancaster I am closing this PR and using a third party library instead, sorry for wasting your time.

plugins/nodes/src/js/pages/nodes/nodes-table/NodesTableContainer.js
```diff
@@ -6,7 +6,7 @@ import CompositeState from "#SRC/js/structs/CompositeState";
 import QueryParamsMixin from "#SRC/js/mixins/QueryParamsMixin";
 import NodesList from "#SRC/js/structs/NodesList";
 
-import NodesTable from "../../../components/NodesTable";
+import NodesTable from "../../../components/NodesTable-new";
```
GeorgiSTodorov: Are we going to start using the new table once this PR is merged?

PR #3087, opened 2018-07-20
DCOS-23825: Nodes / remove graph from nodes page (and move filter)

juliangieseke:
## ℹ️ Description

This PR removes the big graph from the Nodes Page 🎉 

- 🔖 JIRA: [DCOS-23825](https://jira.mesosphere.com/browse/DCOS-23825)
- ~⚠️Please dont merge before Lily had a chance to see it 👍~ Lily Approved ✅  

## 🚧Trade-Offs

We dont have the components and new designs for that new filter dropdown thingy _yet_, so I went ahead and simply moved the old component. Looks good on Desktop, but there is a small glitch on smaller screens (which was there before already for the list/grid switch - see first half of gif)
screen recording 2018-07-20 at 10 19 am
(Gif runs in 2x speed for some reason 🤷‍♂️ )

/cc @lilysamimi 

## 💥Testing

1. Get some load on your cluster
2. Open Nodes Page and verify that Graph is removed
3. Switch between Grid & List to verify that the Filter is only displayed for Grid view
4. Change Filter in Grid view to verify that it changes content of Grid circles


juliangieseke:
Yes, but I'm pretty confident that even that bullet point might be ignored… having this without an emoji buried in depenendiec section it would be guaranteed, that someone misses it and hits "merge"

juliangieseke:
can we please keep this discussion in the other PR?

DanielMSchmidt:
Sure, sorry

plugins/nodes/src/js/pages/nodes-overview/HostsPageContent.js
```diff
@@ -1,14 +1,16 @@
...
 import React from "react";
+import classNames from "classnames";
 
 import DSLExpression from "#SRC/js/structs/DSLExpression";
 import DSLFilterList from "#SRC/js/structs/DSLFilterList";
 import DSLFilterField from "#SRC/js/components/DSLFilterField";
 import FilterBar from "#SRC/js/components/FilterBar";
 import FilterHeadline from "#SRC/js/components/FilterHeadline";
 import NodesList from "#SRC/js/structs/NodesList";
-import ResourceBarChart from "#SRC/js/components/charts/ResourceBarChart";
```
DanielMSchmidt: It seems like this is unused now, can we remove the file please?

plugins/nodes/src/js/pages/nodes-overview/HostsPageContent.js
```diff
@@ -154,6 +156,33 @@ class HostsPageContent extends React.Component {
...
+    if (!this.props.location.pathname.includes("grid")) {
+      // we have to return an empty div here (instead of null),
+      // because FilterBar has fixed rightAlignLastNChildren={2}
+      return <div />;
+    }
+
+    const resourceColors = ResourcesUtil.getResourceColors();
+    const resourceLabels = ResourcesUtil.getResourceLabels();
+    const buttons = ResourcesUtil.getDefaultResources().map(resource => {
+      const classSet = classNames("button button-outline", {
```
DanielMSchmidt: 🎨Funny that I ask you that right now, but can we maybe inline the classSet and use the classnames package for the entire string concatenation there? It seems odd to me to also use template strings here.

plugins/nodes/src/js/pages/nodes-overview/HostsPageContent.js
```diff
@@ -154,6 +156,33 @@ class HostsPageContent extends React.Component {
       </div>
     );
   }
+  getGridViewResourceSwitch() {
+    if (!this.props.location.pathname.includes("grid")) {
+      // we have to return an empty div here (instead of null),
+      // because FilterBar has fixed rightAlignLastNChildren={2}
```
DanielMSchmidt: Would it be possible to extract the `this.props.location.pathname.includes("grid")` into a function so that we don't have to use a fixed value?

plugins/nodes/src/js/pages/nodes-overview/HostsPageContent.js
juliangieseke: I think the way this package is used in out codebase it is "correct" this way. having all the default classes as string and only adding optional ones from the object… 🤔

juliangieseke:
@DanielMSchmidt please revisit 🙇 
@lilysamimi 😇

plugins/nodes/src/js/pages/nodes-overview/HostsPageContent.js
DanielMSchmidt: Yeah, the package usage is correct, but below you do `className={`${classSet} path-color-${resourceColors[resource]}`}`. You could put the `path-color-${resourceColors[resource]}` also into classNames, that woul be easier

plugins/nodes/src/js/pages/nodes-overview/HostsPageContent.js
```diff
@@ -164,36 +194,22 @@ class HostsPageContent extends React.Component {
       handleFilterChange,
       hosts,
       isFiltering,
-      nodeCount,
-      onResourceSelectionChange,
```
tnguyen0: For the props you removed here, why is it not removed from `HostsPageContent.propTypes` and the `<HostsPageContent />`, do we intend on using it later?

juliangieseke: forgot to remove them, will fix 👍

juliangieseke:
(finally) addressed your comment @tnguyen0 - please revisit 🙇 
/cc @DanielMSchmidt

plugins/nodes/src/js/pages/nodes-overview/HostsPageContent.js
```diff
@@ -154,6 +156,34 @@ class HostsPageContent extends React.Component {
...
+    const buttons = ResourcesUtil.getDefaultResources().map(resource => {
+      const classSet = classNames("button button-outline", {
+        active: this.props.selectedResource === resource
+      });
+
+      return (
+        <button
+          key={resource}
+          className={`${classSet} path-color-${resourceColors[resource]}`}
+          onClick={this.props.onResourceSelectionChange.bind(null, resource)}
```
tnguyen0: We are still using `onResourceSelectionChange` here, can you add it back.

nLight:
screen shot 2018-07-30 at 15 10 25
Dunno if those gaps while filtering caused by your change

nLight:
No, they not

plugins/nodes/src/js/pages/nodes-overview/HostsPageContent.js
```diff
@@ -154,6 +156,34 @@ class HostsPageContent extends React.Component {
       </div>
     );
   }
+  isGridView() {
+    return this.props.location.pathname.includes("grid");
```
nLight: We start scattering the logic. How would you feel passing `isGridActive` as a prop down from `NodesOverview`?

juliangieseke: hm, I dont think it is a good idea. in #3117 we will update this so that NodesOverview holds Agents and Masters Tabs… I dont think it is a good idea to have this in `NodesOverview` then.

plugins/nodes/src/js/pages/nodes-overview/HostsPageContent.js
juliangieseke: true. 🤷‍♂️  (legacy code, just copy pasted…)

PR #3091, opened 2018-07-23
DCOS-39331: Add isPublic to node api

DanielMSchmidt:
## Testing

- See the tests that we wrote and verify that they handle each edge case
- Read through [the docs](https://docs.mesosphere.com/1.11/overview/architecture/node-types/#public-agent-nodes) and see if we are misusing the API

## Trade-offs

- We chose to not use the "public_agent" role, because of what we found [here](https://docs.mesosphere.com/1.11/overview/architecture/node-types/#public-agent-nodes)
> The Mesos agents on public agent nodes also have the public_ip:true agent attribute to assist in their discovery.
It seemed to us like this is more precise

## Dependencies

None

### DIFF
--- a/src/js/structs/Node.d.ts
+++ b/src/js/structs/Node.d.ts
@@ -13,4 +13,5 @@ export default class Node extends Item {
   getOutput: () => any;
   sumTaskTypesByState: (state: any) => any;
   getUsedResources: () => any;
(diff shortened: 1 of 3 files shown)


GeorgiSTodorov:
Could `attributes.public_ip` be null or undefined?

PR #3093, opened 2018-07-23
DCOS-38482: Use new Breadcrumb in the JobTaskDetailPage

DanielMSchmidt:
## Testing

- Checkout the task detail page and see if the breadcrumb still behaves the same (easiest with some long running job like nginx)

## Trade-offs

None

## Dependencies

None


plugins/jobs/src/js/pages/JobTaskDetailPage.js
```diff
@@ -2,16 +2,13 @@ import PropTypes from "prop-types";
 import React from "react";
 import { Link } from "react-router";
 
-import DCOSStore from "#SRC/js/stores/DCOSStore";
+import MesosStateStore from "#SRC/js/stores/MesosStateStore";
+import Page from "#SRC/js/components/Page";
+import Breadcrumb from "#SRC/js/components/Breadcrumb";
```
nLight: Sorry my dumb question. Is this `Breadcrumb` the new one? It is in `#SRC/js/components` as I can see which hints that it is a generic one, not the Jobs specific one

DanielMSchmidt: No worries, we have `Breadcrumb` which is a general styling only Breadcrumb element and we also use the `Breadcrumbs` here, which is the new Breadcrumb bar basically, with all the linking logic.

nLight: Oh, I overlooked that bottom import!

PR #3089, opened 2018-07-20
DCOS-39265: Use Cell from UI Kit for the content of the tables

GeorgiSTodorov:
The Cell Component from @dcos/ui-kit was added to all columns in the new (refactored) table.
Closes DCOS-39265

## Testing

run npm install

In `NodesTableContainer.js`:
```diff
-import NodesTable from "../../../components/NodesTable";
+import NodesTable from "../../../components/NodesTable-new";
```

## Trade-offs

I have decided to add a `<span>` element inside the Cell for some files that didn't have it (for example NodesTableGPUColumn.tsx). Some other files already had such a `<span>` and I wanted to be consistent.

## Dependencies
None

PR #3094, opened 2018-07-23
DCOS-39262: Activate new NodesTable

DanielMSchmidt:
## Testing

- Take a look at the nodes page and see if the nodes table still behaves the same

## Trade-offs

- I added a bit more magic to the NodesTable to settle with a less verbose solution as it might be easier to read, understand and extend. I did it a separate commit and I would be open to remove it if you don't like the change
- @juliangieseke pointed me towards moving the bar and the column it refers to together, but that lead to code like this:

```ts
export function renderer(data: Node): React.ReactNode {
  return [
    <Cell style={{float: 'left'}}>
      <ProgressBar
        data={[
          { value: data.getUsageStats("cpus").percentage, className: "color-1" }
        ]}
        total={100}
      />
    </Cell>,
    <TextCell>
      <span>{data.getUsageStats("cpus").percentage}%</span>
    </TextCell>
  ];
}
```
Please note the usage of inline styles here. Given that the benefit is not so big I would rather not do it or have a seperate issue preparing something like this in the UI-Kit maybe? My understanding here was that we should not need to interfer with the styling of the ui-kit components

## Dependencies

- DCOS-39112: https://github.com/dcos/dcos-ui/pull/3082 => ✅ 
- DCOS-39112: https://github.com/dcos/dcos-ui/pull/3089 => ✅


DanielMSchmidt:
Hold merge as there are still open dependencies

plugins/nodes/src/js/columns/NodesTableCPUBarColumn.tsx
```diff
@@ -6,7 +6,7 @@ import Node from "#SRC/js/structs/Node";
 import { IWidthArgs as WidthArgs } from "#PLUGINS/nodes/src/js/types/IWidthArgs";
 import ProgressBar from "#SRC/js/components/ProgressBar";
 
-export function cpubarRenderer(data: Node): React.ReactNode {
+export function renderer(data: Node): React.ReactNode {
```
juliangieseke: can you please integrate all the `*bar` columns into 😅

DanielMSchmidt:
@juliangieseke I will work on the bar refactoring once https://github.com/dcos/dcos-ui/pull/3089 is merged, before that it doesn't seem to make sense in terms of merge conflicts

DanielMSchmidt:
@juliangieseke  PR is merged, will work on this now 👍

plugins/nodes/src/js/components/NodesTable.tsx
```diff
@@ -0,0 +1,201 @@
...
+import * as NodesTableCPUColumn from "../columns/NodesTableCPUColumn";
+import * as NodesTableMemBarColumn from "../columns/NodesTableMemBarColumn";
+import * as NodesTableMemColumn from "../columns/NodesTableMemColumn";
+import * as NodesTableDiskBarColumn from "../columns/NodesTableDiskBarColumn";
+import * as NodesTableDiskColumn from "../columns/NodesTableDiskColumn";
+import * as NodesTableGPUBarColumn from "../columns/NodesTableGPUBarColumn";
+import * as NodesTableGPUColumn from "../columns/NodesTableGPUColumn";
+
+interface NodesTableColumn {
+  displayName: string;
```
juliangieseke: this is not a string (will become important below). Also please consider renaming it, e.g.: `headerContent?: React.ReactNode;`

plugins/nodes/src/js/components/NodesTable.tsx
```diff
@@ -0,0 +1,201 @@
...
+  sortKey?: string;
```
juliangieseke: using my code below, this needs to be split up into `id: string` and `sortable: boolean`

plugins/nodes/src/js/components/NodesTable.tsx
```diff
@@ -0,0 +1,201 @@
...
+  }
+
+  render() {
+    const { data, sortColumn, sortDirection } = this.state;
+
+    return (
+      <Table data={data.slice()}>
+        {columns.map(
+          column =>
+            column.sortKey ? (
```
juliangieseke: Hm, I think here's something wrong 🤔

Thinking of this table as an example for others, we cant assume that `sortKey` and "visible Header" are tied together. It might be the case in this particular example (and it isnt anymore, after we integrated the `*bar` columns) - but there might be different tables somewhere - also there's a bug in this refactoring :/ 

That said, I think this whole code block should look like this (sorry for just presenting a solution here, had to think about it for a while…):
```
{columns.map(column => {
  return (
    <Column
      key={column.id} // this should be mandatory and _used as sortKey_ if sortable
      header={getColumnHeader(column, this.handleSortClick, sortColumn === column.id ? sortDirection : null)} // this will have a bit ore logic now…
      cellRenderer={column.component.renderer}
      width={column.component.sizer}
    />
  );
})}
```
`getColumnHeader` will look something like this (this should/could be some kind of a UI Kit helper…)
```
function getColumnHeader(
  column: NodesTableColumn,
  handleSortClick: (columnName: string) => void,
  sortDirection: SortDirection | null
): React.ReactNode {
  if (column.displayName !== undefined && column.sortable === true)
    return (
      <SortableColumnHeader
        columnContent={column.displayName}
        sortHandler={handleSortClick.bind(null, column.id)}
        sortDirection={sortDirection}
      />
    );

  return <HeaderCell>{column.displayName}</HeaderCell>;
}
``

Comment (juliangieseke) [reply plugins/nodes/src/js/components/NodesTable.tsx]: actually, there is an even better version of `getColumnHEader` where the `< SortableColumnHeader />` only returns the _contents_ of `<HeaderCell />`. I think that would be even better 👍

File: plugins/nodes/src/js/components/NodesTable.tsx [hunk 5]
```diff
@@ -0,0 +1,201 @@
...
+    displayName: "Region"
+  },
+  { sortKey: "zone", component: NodesTableZoneColumn, displayName: "Zone" },
+  {
+    sortKey: "health",
+    component: NodesTableHealthColumn,
+    displayName: "Health"
+  },
+  { sortKey: "tasks", component: NodesTableTasksColumn, displayName: "Tasks" },
+  { component: NodesTableCPUBarColumn, displayName: "CPUBar" },
```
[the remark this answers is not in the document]
DanielMSchmidt: @juliangieseke Sorry, was still working on it and just finished noting the trade-off down, please see the PR description.

lilysamimi:
@tnguyen0 Good find. Yes, those two things you mentioned are expected.

tnguyen0:
Would the needed changes above be blocking? @lilysamimi @juliangieseke @DanielMSchmidt 

Can easily add the cursor pointer, not about the hover show caret.

juliangieseke:
@tnguyen0 @lilysamimi https://jira.mesosphere.com/browse/DCOS-39862

tests/pages/nodes/NodesPage-cy.js
```diff
@@ -6,7 +6,7 @@ describe("Nodes Page", function() {
         nodeHealth: true
       });
       cy.visitUrl({ url: "/nodes" });
-      cy.get("tbody tr td:first-child").as("hostnames");
+      cy.get(".BottomLeftGrid_ScrollWrapper").as("hostnames");
```
nLight: What other selector could we use? The idea of those tests is to operate with human readable selectors. It wasn't perfect, but a programmer could say we're looking for the first cell in a table. `.BottomLeftGrid_ScrollWrapper` doesn't tell me anything.

tests/pages/nodes/node/HealthTab-cy.js
```diff
@@ -9,8 +9,14 @@ describe("Node Health Tab [0fa]", function() {
   context("Navigate to tab [0fb]", function() {
     it("navigates to health tab [0fc]", function() {
       cy.visitUrl({ url: "/nodes", identify: true, fakeAnalytics: true });
-      cy.get("tr a").eq(0).click({ force: true });
-      cy.get(".menu-tabbed-item").contains("Health").click();
+      cy.get(
+        ".BottomLeftGrid_ScrollWrapper .ReactVirtualized__Grid__innerScrollContainer a"
```
nLight: same

tests/pages/nodes/node/NodeDetail-cy.js
```diff
@@ -10,8 +10,9 @@ describe("Nodes Detail Page", function() {
     it("navigates to node detail page", function() {
       cy.visitUrl({ url: "/nodes", identify: true, fakeAnalytics: true });
       var nodeName;
-      cy
-        .get("tr a")
+      cy.get(
+        ".BottomLeftGrid_ScrollWrapper .ReactVirtualized__Grid__innerScrollContainer a"
```
nLight: 😩

tests/pages/nodes/NodesPage-cy.js
DanielMSchmidt: I could use `.table-cell-link-primary` but this is not really more verbose I would argue and it also might become an easier target to change.

The problem is that our HTML is not semantic anymore, we just use a bunch of divs and spans now plus every class coming from the ui-kit is a) randomly generated and b) changes with every release of the UI-kit

nLight: Please create a JIRA to find a solution.

PR #3090, opened 2018-07-20
refactor(sidebar): use only toggleSidebar state

weblancaster:
This PR refactors the sidebar to use only one state `isVisible` and `toggleSidebar` removing the need for `isDock` (dock/undock is not needed anymore 🎉 )

- Refactor various parts of the code to only use toggleSidebar to open/close Sidebar and removes the now not in use code.
- Move overlay to index to display in mobile screen when sidebar is open.
- Replace`isDock` to `isVisible` in UserSettingsStore.

ps: I have worked with @leemunroe to make sure I'm following his requirements.

## Testing

1. `npm start`
2. resize the screen and note the sidebar open/close depending on the screen size
3. click on the sidebar icon (when visible)

## Result

screen recording 2018-07-20 at 04 07 pm

PR #3097, opened 2018-07-23
feat(PageHeader): add SidebarToggle to header

weblancaster:
This PR adds the sidebar toggle based on the work in https://github.com/dcos/dcos-ui/pull/3090 but be aware the design will be added by @tnguyen0 in another PR, this is only an implementation of the button.

## Testing

1. `npm start`
2. Go to `HeaderBar.tsx` comment out line 12 - 14.
3. Click on the header hamburger button and note the sidebar open/close

screen recording 2018-07-24 at 05 19 pm

### DIFF
--- a/plugins/nodes/src/js/columns/__tests__/__snapshots__/NodesTableHealthColumn-test.js.snap
+++ b/plugins/nodes/src/js/columns/__tests__/__snapshots__/NodesTableHealthColumn-test.js.snap
@@ -2,7 +2,7 @@
 
 exports[`NodesTableHealthColumn renderer renders with health status 'Healthy' 1`] = `
 <div
-  className="css-19m5nce"
+  className="css-4tf7uh"
 >
   <span
     className="text-success"
@@ -14,7 +14,7 @@ exports[`NodesTableHealthColumn renderer renders with health status 'Healthy' 1`
 
 exports[`NodesTableHealthColumn renderer renders with health status 'NA' 1`] = `
 <div
-  className="css-19m5nce"
+  className="css-4tf7uh"
 >
   <span
     className="text-mute"
@@ -26,7 +26,7 @@ exports[`NodesTableHealthColumn renderer renders with health status 'NA' 1`] = `
 
 exports[`NodesTableHealthColumn renderer renders with health status 'Unhealthy' 1`] = `
 <div
-  className="css-19m5nce"
+  className="css-4tf7uh"
 >
   <span
     className="text-danger"
@@ -38,7 +38,7 @@ exports[`NodesTableHealthColumn renderer renders with health status 'Unhealthy' 
 
 exports[`NodesTableHealthColumn renderer renders with health status 'Warn' 1`] = `
 <div
-  className="css-19m5nce"
+  className="css-4tf7uh"
 >
   <span
     className="text-warning"
--- a/plugins/nodes/src/js/columns/__tests__/__snapshots__/NodesTableHealthColumn-test.js.snap
+++ b/plugins/nodes/src/js/columns/__tests__/__snapshots__/NodesTableHealthColumn-test.js.snap
@@ -2,7 +2,7 @@
 
 exports[`NodesTableHealthColumn renderer renders with health status 'Healthy' 1`] = `
 <div
-  className="css-4tf7uh"
+  className="css-19m5nce"
 >
   <span
     className="text-success"
@@ -14,7 +14,7 @@ exports[`NodesTableHealthColumn renderer renders with health status 'Healthy' 1`
 
 exports[`NodesTableHealthColumn renderer renders with health status 'NA' 1`] = `
 <div
-  className="css-4tf7uh"
+  className="css-19m5nce"
 >
   <span
     className="text-mute"
@@ -26,7 +26,7 @@ exports[`NodesTableHealthColumn renderer renders with health status 'NA' 1`] = `
 
 exports[`NodesTableHealthColumn renderer renders with health status 'Unhealthy' 1`] = `
 <div
-  className="css-4tf7uh"
+  className="css-19m5nce"
 >
   <span
     className="text-danger"
@@ -38,7 +38,7 @@ exports[`NodesTableHealthColumn renderer renders with health status 'Unhealthy' 
 
 exports[`NodesTableHealthColumn renderer renders with health status 'Warn' 1`] = `
 <div
-  className="css-4tf7uh"
+  className="css-19m5nce"
 >
   <span
     className="text-warning"
--- a/src/js/components/Sidebar.js
+++ b/src/js/components/Sidebar.js
@@ -12,7 +12,8 @@ import PrimarySidebarLink from "../components/PrimarySidebarLink";
 import ScrollbarUtil from "../utils/ScrollbarUtil";
 import SidebarActions from "../events/SidebarActions";
 import SidebarHeader from "./SidebarHeader";
-import { isMobile } from "../utils/Util";
+import { getCurrentViewport } from "../utils/ViewportUtil";
+import * as viewport from "../constants/Viewports";
 
 const {
   NavigationService,
@@ -124,7 +125,7 @@ class Sidebar extends React.Component {
   }
 
   handleSubmenuItemClick() {
-    if (isMobile()) {
+    if (getCurrentViewport() === viewport.MOBILE) {
       SidebarActions.close();
     }
   }
--- a/src/js/components/Sidebar.js
+++ b/src/js/components/Sidebar.js
@@ -12,6 +12,7 @@ import PrimarySidebarLink from "../components/PrimarySidebarLink";
 import ScrollbarUtil from "../utils/ScrollbarUtil";
 import SidebarActions from "../events/SidebarActions";
 import SidebarHeader from "./SidebarHeader";
+import { isMobile } from "../utils/Util";
 
 const {
   NavigationService,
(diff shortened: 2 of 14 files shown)


DanielMSchmidt:
@weblancaster I changed base to master as #3090 got merged, now there are conflicts. Hope this was okay to you 😇

DanielMSchmidt:
Wouldn't @tnguyen0 styling PR be a dependency then?

tests/components/Sidebar-cy.js
```diff
@@ -142,31 +142,30 @@ describe("Sidebar", function() {
       });
     });
 
-    it("renders within the viewport on initial load", function() {
-      cy.get(".sidebar").then(function($sidebar) {
-        const sidebarRect = $sidebar.get(0).getBoundingClientRect();
+    it("Close sidebar when sidebarToggle button clicked", function() {
```
tnguyen0: Nit: lowercase titles and "closes"

tests/components/Sidebar-cy.js
```diff
@@ -142,31 +142,30 @@ describe("Sidebar", function() {
...
-        });
-      });
+      // reset viewport back
+      cy.viewport("macbook-15");
     });
 
-    it("offsets the page content so that they are sitting directly adjacent to one another without overlapping", function() {
-      cy.get(".sidebar").then(function($sidebar) {
-        const sidebarRect = $sidebar.get(0).getBoundingClientRect();
+    it("Display overlay when sidebar is open on mobile/tablet", function() {
```
tnguyen0: Can you group the mobile tests under same context

tests/components/Sidebar-cy.js
tnguyen0: Can we also have a test for open sidebar to test the toggle

tests/components/Sidebar-cy.js
tnguyen0: Can we test toggle for mobile too

tests/components/Sidebar-cy.js
weblancaster: I rather have then running separately (minimum test) that way easier to catch when there is a bug/failure.

tests/components/Sidebar-cy.js
weblancaster: I rather have then running separately (minimum test) that way easier to catch when there is a bug/failure.

weblancaster: mobile doesn't change the environment.. the toggle is the same so I believe there is no need to test the button again on mobile.

src/js/components/Sidebar.js
```diff
@@ -123,7 +124,9 @@ class Sidebar extends React.Component {
   }
 
   handleSubmenuItemClick() {
-    SidebarActions.close();
+    if (isMobile()) {
```
weblancaster: We only want to close the sidebar on mobile.

src/js/components/SidebarToggle.d.ts
```diff
@@ -0,0 +1,5 @@
+import { Component } from "react";
+
+interface SidebarToggleProps {}
```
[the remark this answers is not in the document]
weblancaster: for some reason isn't complaining.

src/js/pages/Index.js
```diff
@@ -146,19 +145,20 @@ var Index = React.createClass({
 
   handleWindowResize(event) {
     const windowWidth = event.target.innerWidth;
-    const { isVisible } = this.internalStorage_get();
 
-    if (windowWidth <= MOBILE_THRESHOLD && isVisible) {
-      SidebarActions.close();
+    if (isViewportChangingTo(previousWidth, windowWidth) === viewport.DESKTOP) {
```
weblancaster: I spent some time talking to @Fabs about how we could do this in a good way state isn't everywhere.

the idea was to build something to be re-used.

src/js/utils/Util.js
```diff
@@ -249,6 +250,30 @@ const Util = {
...
+
+    // Transition from Desktop to Mobile
+    if (
+      windowWidth < viewport.MOBILE_THRESHOLD &&
+      previousWidth > viewport.MOBILE_THRESHOLD
+    ) {
+      return viewport.MOBILE;
+    }
+
+    return false;
```
juliangieseke: When it is indented to be reusable and part of a util, this function either should return the current viewport (`getCurrentView`) or boolean (`viewHasChanged`). But not both.
The way it is implemented now is kinda opinionated for the current usecase instead of having helpers you can use to check common things. 🤔

src/js/utils/Util.js
```diff
@@ -249,6 +250,30 @@ const Util = {
       search: aElement.search,
       username: aElement.username
     };
+  },
+
+  isMobile() {
```
juliangieseke: Shall we move these to `ViewportUtil` or similar? (Both/all new functions)

src/js/constants/Viewports.js
```diff
@@ -0,0 +1,7 @@
+// We use this value in the css
+// when targeting tablet or mobile
+export const MOBILE_THRESHOLD = 994;
+
+export const DESKTOP = "DESKTOP";
```
juliangieseke: Was there a reason not to chose enum here? Wonder which one fits better 🤔

weblancaster: When @Fabs and I paired we like how declarative this option is.

src/js/pages/Index.js
```diff
@@ -18,21 +17,22 @@ import HeaderBar from "../components/HeaderBar";
...
   return {
     isVisible: SidebarStore.get("isVisible")
   };
 }
 
-// We use this value in the css
-// when targeting tablet or mobile
-const MOBILE_THRESHOLD = 994;
+// By default users use desktop
+let previousWidth = null;
```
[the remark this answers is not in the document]
weblancaster: The reason I went with local variable instead state is so we don't trigger the lifecycle which is not needed.

src/js/utils/Util.js
weblancaster: Sounds good.

src/js/utils/Util.js
Fabs: I think there is a point here that we are returning different types depending on the logic, which I think makes it harder to understand this function. I would try to return the same type everywhere.

src/js/pages/Index.js
```diff
@@ -144,21 +138,34 @@ var Index = React.createClass({
...
-    const { isVisible } = this.internalStorage_get();
+  handleWindowResize() {
+    const currentWindowWidth = window.innerWidth;
 
-    if (windowWidth <= MOBILE_THRESHOLD && isVisible) {
-      SidebarActions.close();
+    if (
+      !hasViewportChanged(this.state.previousWindowWidth, currentWindowWidth)
+    ) {
+      this.setState({
```
juliangieseke: there is no need to update this here AFAIK 🤔

src/js/pages/Index.js
```diff
@@ -144,21 +138,34 @@ var Index = React.createClass({
...
-    if (windowWidth > MOBILE_THRESHOLD && !isVisible) {
+    if (getCurrentViewport() === viewport.DESKTOP) {
       SidebarActions.open();
     }
+
+    if (getCurrentViewport() === viewport.MOBILE) {
+      SidebarActions.close();
+    }
+
+    this.setState({
```
juliangieseke: this can be moved into the if clauses 🤔

src/js/pages/Index.js
weblancaster: We do need to register the state every time there is a window change.

src/js/pages/Index.js
weblancaster: Not sure why I would do that if I will need to update every time there is a change.

src/js/pages/Index.js
```diff
@@ -60,17 +58,14 @@ var Index = React.createClass({
...
-    this.internalStorage_set(state);
   },
 
   componentDidMount() {
     SidebarStore.addChangeListener(
       EventTypes.SIDEBAR_CHANGE,
       this.onSideBarChange
     );
-    global.addEventListener("resize", this.handleWindowResize);
+    global.addEventListener("resize", this.handleWindowResize.bind(this));
```
Poltergeist: I am getting an error here.
image

weblancaster: This is an warning not a error BUT I removed the bind anyways because I'm removing the listener so no need to bind.

weblancaster:
@Poltergeist correct, that's expected behavior.

src/js/pages/Index.js
juliangieseke: yeah, you're right… 😅

DanielMSchmidt:
@weblancaster Would be cool if you could add the lines I need to change to see the headerbar in the PR next time ;)

plugins/nodes/src/js/columns/__tests__/__snapshots__/NodesTableHealthColumn-test.js.snap
```diff
@@ -2,7 +2,7 @@
 
 exports[`NodesTableHealthColumn renderer renders with health status 'Healthy' 1`] = `
 <div
-  className="css-19m5nce"
+  className="css-4tf7uh"
```
DanielMSchmidt: Did you change anything at the nodes column? Or is this cause by the UI-KIt update?

weblancaster: This is a snapshot update, whenever there is a build the `css-randomString` changes and the snapshots need to be updated

Poltergeist: yes could be because of the ui-kit update.

tnguyen0:
@weblancaster conflicts need to be resolved

PR #3101, opened 2018-07-24
Close DCOS-38595: fix framework configuration form

Poltergeist:
This fixes FrameworkConfigurationForm to gracefully handle an invalid
`formData` and provide an error message.

image

Close DCOS-38595

## Testing

Please have a look at the issue it has all the information to set up the cluster with the breaking state.

## Trade-offs

The trade-off is that currently in case of an invalid solution you can't disable the json-editor in the edit form.

## Dependencies

This change has no dependencies.


nLight:
Since it has been discovered in 1.11 please backport it

PR #3102, opened 2018-07-25
1.10 DCOS-39661: Check instance type instead of labels

nLight:
After fixing #3092 @orlandohohmeier and I decided to propagate the check forward so that we don't regress in the code base in the way we're checking if a service is a Framework or not.

---

Service has already been wrapped into a Framework struct at this point. No need in re-checking its type via labels.

Closes DCOS-39661

## Testing

Deploy a Framework from the Catalog, check to see if Debug tab still works.


DanielMSchmidt:
@nLight Do we want to merge this without system tests?

PR #3106, opened 2018-07-25
fix(serviceForm): fix environment variable reducer

Poltergeist:
This fixes the environment variables reducer to support empty values.

image

Close DCOS-38433

## Testing

Try to create a service with the following config:

```JSON
{
  "env": {
    "key": ""
  }
}
```

I worked together with @tnguyen0 on this issue.

### DIFF
--- a/plugins/services/src/js/components/modals/CreateServiceModalForm.js
+++ b/plugins/services/src/js/components/modals/CreateServiceModalForm.js
@@ -72,6 +72,34 @@ const CONSTANTLY_UNMUTED_ERRORS = [
   /^volumes\.[0-9]+\./
 ];
 
+function cleanConfig(config) {
+  const { labels = {}, env = {}, environment = {}, ...serviceConfig } = config;
+
+  let newServiceConfig = CreateServiceModalFormUtil.stripEmptyProperties(
+    serviceConfig
+  );
+  if (Object.keys(labels).length !== 0) {
+    newServiceConfig = {
+      labels,
+      ...newServiceConfig
+    };
+  }
+  if (Object.keys(env).length !== 0) {
+    newServiceConfig = {
+      env,
+      ...newServiceConfig
+    };
+  }
+  if (Object.keys(environment).length !== 0) {
+    newServiceConfig = {
+      environment,
+      ...newServiceConfig
+    };
+  }
+
+  return newServiceConfig;
+}
+
 class CreateServiceModalForm extends Component {
   constructor() {
     super(...arguments);
@@ -80,19 +108,9 @@ class CreateServiceModalForm extends Component {
     //       shouldComponentUpdate function, since we are trying to reduce
     //       the number of updates as much as possible.
     // In the Next line we are destructing the config to keep labels as it is and even keep labels with an empty value
-    const { labels = {}, ...serviceConfig } = ServiceUtil.getServiceJSON(
-      this.props.service
+    const newServiceConfig = cleanConfig(
+      ServiceUtil.getServiceJSON(this.props.service)
     );
-
-    let newServiceConfig = {
-      labels,
-      ...CreateServiceModalFormUtil.stripEmptyProperties(serviceConfig)
-    };
-    if (Object.keys(labels).length === 0) {
-      newServiceConfig = CreateServiceModalFormUtil.stripEmptyProperties(
-        serviceConfig
-      );
-    }
     this.state = Object.assign(
       {
         appConfig: null,
@@ -327,16 +345,7 @@ class CreateServiceModalForm extends Component {
     );
 
     // In the Next line we are destructing the config to keep labels as it is and even keep labels with an empty value
-    const { labels, ...config } = newConfig;
-
-    if (Object.keys(labels).length === 0) {
-      return CreateServiceModalFormUtil.stripEmptyProperties(config);
-    }
-
-    return {
-      labels,
-      ...CreateServiceModalFormUtil.stripEmptyProperties(config)
-    };
+    return cleanConfig(newConfig);
   }
 
   getErrors() {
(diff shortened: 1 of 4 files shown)


plugins/services/src/js/reducers/serviceForm/JSONReducers/EnvironmentVariables.js
```diff
@@ -65,7 +65,7 @@ module.exports = {
 
     return Object.keys(state.env)
       .filter(function(key) {
-        return typeof state.env[key] === "string";
+        return typeof state.env[key] === "string" || state.env[key] == null;
```
tnguyen0: supppper-nit: null check first 🙈

plugins/services/src/js/components/modals/CreateServiceModalForm.js
```diff
@@ -80,18 +80,33 @@ class CreateServiceModalForm extends Component {
     //       shouldComponentUpdate function, since we are trying to reduce
     //       the number of updates as much as possible.
     // In the Next line we are destructing the config to keep labels as it is and even keep labels with an empty value
-    const { labels = {}, ...serviceConfig } = ServiceUtil.getServiceJSON(
-      this.props.service
+    const {
+      labels = {},
+      env = {},
+      environment = {},
```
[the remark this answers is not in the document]
Poltergeist: I would rather keep it at environment otherwise I would need to do 

```JS
environment: podEnv,
...conf
```
Later in the code.

plugins/services/src/js/components/modals/CreateServiceModalForm.js
```diff
@@ -80,18 +80,33 @@ class CreateServiceModalForm extends Component {
...
-
-    let newServiceConfig = {
-      labels,
-      ...CreateServiceModalFormUtil.stripEmptyProperties(serviceConfig)
-    };
-    if (Object.keys(labels).length === 0) {
-      newServiceConfig = CreateServiceModalFormUtil.stripEmptyProperties(
-        serviceConfig
-      );
+    if (Object.keys(labels).length !== 0) {
```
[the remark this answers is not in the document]
Poltergeist: I'll have a look. I am in fear but I think we could pull it out.

nLight:
I have a strange feeling that we're fixing the bug over and over again...

plugins/services/src/js/components/modals/CreateServiceModalForm.js
```diff
@@ -72,6 +72,34 @@ const CONSTANTLY_UNMUTED_ERRORS = [
   /^volumes\.[0-9]+\./
 ];
 
+function cleanConfig(config) {
+  const { labels = {}, env = {}, environment = {}, ...serviceConfig } = config;
+
+  let newServiceConfig = CreateServiceModalFormUtil.stripEmptyProperties(
```
DanielMSchmidt: Would it be feasible to add a param for exceptions to the `CreateServiceModalFormUtil. stripEmptyProperties` method? I would expect something like this could work an be a bit easier on this side of the table

```js
stripEmptyProperties(object, blacklist = []) {
    if (typeof object !== "object" || object === null) {
      return object;
    }

    // Pick base object according to type
    let baseObject = {};
    if (Array.isArray(object)) {
      baseObject = [];
    }

    return Object.keys(object).reduce(function(memo, key) {
      if (
        (!ValidatorUtil.isEmpty(object[key]) && !Number.isNaN(object[key])) ||
        Array.isArray(object[key])
      ) {
        // Apply the strip function recursively and keep only non-empty values
        const value = CreateServiceModalFormUtil.stripEmptyProperties(
          object[key]
        );
        if (
          !blacklist.includes(key) ||
          (!ValidatorUtil.isEmpty(value) && !Number.isNaN(value)) ||
          Array.isArray(value)
        ) {
          memo[key] = value;
        }
      }

      return memo;
    }, baseObject);
  }
};
```

plugins/services/src/js/reducers/serviceForm/JSONReducers/EnvironmentVariables.js
```diff
@@ -65,7 +65,7 @@ module.exports = {
 
     return Object.keys(state.env)
       .filter(function(key) {
-        return typeof state.env[key] === "string";
+        return state.env[key] == null || typeof state.env[key] === "string";
```
nLight: We removed `null` why checking it here?

plugins/services/src/js/components/modals/CreateServiceModalForm.js
nLight: I wonder if we want to keep it. What is there left for cleaning?

plugins/services/src/js/reducers/serviceForm/JSONReducers/EnvironmentVariables.js
Poltergeist: This is the parser I just wanted to make sure we are catching 
```JSON
{
  "envvar": null
}
```

and
```JSON
{
  "envvar": ""
}
```

And we should not remove null ... did I miss something?

plugins/services/src/js/components/modals/CreateServiceModalForm.js
Poltergeist: `portDefinitions` comes to my mind. `containers` in Pods but rather unlikely. `fetch`, `constraints`. There are a couple of array values which we want to clean.

Poltergeist: That is actually a neat idea. We already kind of established a bad habit and we use a bad pattern. I'll look into this.

nLight: I wonder because every time we're trying to touch JSON it breaks someone's process

plugins/services/src/js/reducers/serviceForm/JSONReducers/EnvironmentVariables.js
nLight: We used to set `value: null` as a default, but you are changing this. Wonder if there are other places we could get `null`

PR #3107, opened 2018-07-26
DCOS-39112: Fix comparison of numeric values in Nodes Table

DanielMSchmidt:
## Testing

```diff
-import NodesTable from "../../../components/NodesTable";
+import NodesTable from "../../../components/NodesTable-new";
```

Please take a look at the nodes table and see if all the sortings look correct (especially the ones with numbers, like mem, cpu, and disk. I also fixed that the carret should now be shown while sorting every column (except for the bars which are not sortable

## Trade-offs

- I decided to use an external library which provided the tiebreaker feature we want instead of writing something my own, because the contract it provided looked clear and it seems to solve the problem. Given the experience with my last try on designing an API for this problem I also wanted to cut discussions about the API and expected or unexpected behavior
- I introduced a comparators array in most columns, because I like to see the core of the sorting at one glance
- I did not refactor to DRY everything up, because I don't want to have anything disputable in this PR as it blocks quite a lot of efforts.

## Dependencies

None

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,6 +33,7 @@
     "@dcos/mesos-client": "0.2.3",
     "@dcos/tslint-config": "0.1.0",
     "@dcos/ui-kit": "1.3.0",
+    "array-sort": "1.0.0",
     "babel-polyfill": "6.23.0",
     "browser-info": "0.4.0",
     "classnames": "2.2.3",
--- a/plugins/nodes/src/js/columns/NodesTableDiskColumn.tsx
+++ b/plugins/nodes/src/js/columns/NodesTableDiskColumn.tsx
@@ -35,7 +35,7 @@ export function diskSorter(data: Node[], sortDirection: SortDirection): Node[] {
   const reverse = sortDirection !== "ASC";
   return sort(data, comparators, { reverse });
 }
-export function diskSizer(args: WidthArgs): number {
+export function diskSizer(_args: WidthArgs): number {
   // TODO: DCOS-39147
-  return Math.min(60, Math.max(60, args.width / args.totalColumns));
+  return 70;
 }
--- a/plugins/nodes/src/js/columns/NodesTableDiskColumn.tsx
+++ b/plugins/nodes/src/js/columns/NodesTableDiskColumn.tsx
@@ -1,5 +1,5 @@
 import * as React from "react";
-import { compareValues } from "#PLUGINS/nodes/src/js/utils/compareValues";
+import sort from "array-sort";
 import Node from "#SRC/js/structs/Node";
 // TODO: DCOS-39079
 // import { IWidthArgs as WidthArgs } from "@dcos/ui-kit/packages/table/components/Column";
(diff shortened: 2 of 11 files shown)


nLight:
Build is failing

ERROR in ./node_modules/array-sort/index.js

Module not found: Error: Can't resolve 'default-compare' in '/jenkins/workspace/cos-ui-oss-pipeline_PR-3107-LK5LQ2HR5MDRELO24E5FNVFWXFQKZQSEKZVLTZL4AWB7MMZKSYRQ/node_modules/array-sort'

 @ ./node_modules/array-sort/index.js 10:21-47

DanielMSchmidt:
@nLight Fixed that 👍

package.json
```diff
@@ -33,6 +33,7 @@
     "@dcos/mesos-client": "0.2.3",
     "@dcos/tslint-config": "0.1.0",
     "@dcos/ui-kit": "1.3.0",
+    "array-sort": "1.0.0",
```
nLight: 👍

plugins/nodes/src/js/columns/NodesTableGPUColumn.tsx
```diff
@@ -1,30 +1,40 @@
...
     </TextCell>
   );
 }
 
+function compareNodesByGpuUsage(a: Node, b: Node): number {
+  return getGpuUsage(a) - getGpuUsage(b);
+}
+
+function compareNodesByHostname(a: Node, b: Node): number {
+  return a
```
nLight: Wonder if we need a `compareNodesByHostname ` Util function 🤔

DanielMSchmidt: Thought about it, but I wanted to keep the scope of this as little as possible. If you add this as a utility you open up the discussion of why do we have it for hostnames and not the others.

plugins/nodes/src/js/columns/NodesTableDiskColumn.tsx
```diff
@@ -1,33 +1,41 @@
...
-      a.getHostName().toLowerCase(),
-      b.getHostName().toLowerCase()
-    )
-  );
-  return sortDirection === "ASC" ? sortedData : sortedData.reverse();
+  const reverse = sortDirection !== "ASC";
+  return sort(data, comparators, { reverse });
 }
-export function diskSizer(args: WidthArgs): number {
+export function diskSizer(_args: WidthArgs): number {
```
GeorgiSTodorov: This will conflict with another currently open PR - https://github.com/dcos/dcos-ui/pull/3110/files#diff-1e1caaa9ce248bceffeda8f7749a5de1R33

plugins/nodes/src/js/columns/NodesTableMemColumn.tsx
```diff
@@ -1,34 +1,42 @@
...
-  return sortDirection === "ASC" ? sortedData : sortedData.reverse();
+  const reverse = sortDirection !== "ASC";
+  return sort(data, comparators, { reverse });
 }
 
-export function memSizer(args: WidthArgs): number {
+export function memSizer(_args: WidthArgs): number {
   // TODO: DCOS-39147
-  return Math.min(60, Math.max(60, args.width / args.totalColumns));
+  return 70;
```
GeorgiSTodorov: This too.

DanielMSchmidt:
@GeorgiSTodorov The sizing lead to the carret not being shown, therefore I thought it might make sense to fix this bug right on the spot

GeorgiSTodorov:
Your change makes sense, because the other PR was only posted today, while your PR is from yesterday, but still the one who gets merged second will have to solve these conflicts.

DanielMSchmidt:
That's how it goes 🤷‍♂️ I can take a look at the other one and see if the bug exists there, too

GeorgiSTodorov:
I am just testing the other one and you are right, it exists there as well. I will post a comment on the other PR.

PR #3111, opened 2018-07-27
Improve PR template

DanielMSchmidt:
## Testing

Please take a look and decide if you think this will improve the process of us creating PRs and discussion in a better way.

## Trade-offs

We did a lot of discussion of all sort of trade-offs but I would like to get feedback without any of these thoughts going into your decision making	

## Dependencies

None

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,19 +1,18 @@
-## Testing
+## 📝 Summary
 
 <!--
-What is needed to test the changes? e.g. specific cluster, service definitions
-How can one see the result of your work? e.g. configurations, URLs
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,19 +1,18 @@
-## Testing
+## 📝 Summary
 
 <!--
```
tnguyen0: I'm not sure the summary title is needed? When you make a PR from a branch, it auto puts the commit message at the beginning. If there is a summary title, I'd have to copy and move it under the Summary title every time. I sort of feel the summary is already implied. But keep the instructions regarding the summary.

juliangieseke:
After talking to daniel again, we decided to wait a few more weeks before changing

PR #3114, opened 2018-07-29
DCOS-22340: semantic release

juliangieseke:
## Testing

No chance to test from this PR… but semantic release worked locally on my machine (https://github.com/dcos/dcos-ui/releases/tag/master%2Bv2.0.1) and I tested some dry runs on CI (https://jenkins.mesosphere.com/service/jenkins/blue/organizations/jenkins/frontend%2Fdcos-ui-oss-pipeline/detail/jg%2FDCOS-22340-semantic-release/3/pipeline/48).

## Trade-offs

- I decided to add a `retry(3)` to the system-tests to make our pipeline a bit more robust… 🤷‍♂️ 
- also, because CI is the whole idea of semantic release, I decided to do a quick update to the way we upload binaries for dcos - we already made the filename unique before, but we've overriden the version number…
   => from now on, we can use them for our releases 💃 (I'll update our wiki once this is stable)

## Dependencies

EE part should be merged (almost) at the same time.

### DIFF
--- a/.releaserc
+++ b/.releaserc
@@ -1,0 +1,7 @@
+{
+  "branch": "master",
+  "tagFormat": "master+v${version}",
+  "verifyConditions": ["@semantic-release/github"],
+  "prepare": [],
+  "publish": ["@semantic-release/github"]
+}
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "dcos-ui",
-  "version": "2.0.0",
+  "version": "0.0.0-dev+semantic-release",
   "description": "User Interface of the DCOS",
   "main": "js/index.js",
   "keywords": [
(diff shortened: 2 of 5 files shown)


.releaserc
```diff
@@ -0,0 +1,7 @@
+{
+  "branch": "master",
+  "tagFormat": "master+v${version}",
+  "verifyConditions": ["@semantic-release/github"],
+  "prepare": [],
+  "publish": ["@semantic-release/github"]
```
nLight: Is that the way `semantic-release` inherits configuration?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "dcos-ui",
-  "version": "2.0.0",
+  "version": "0.0.0-dev+semantic-release",
```
nLight: > # release version (after switching to semantic-release, we cant use package.json anymore)

I wonder if it's gonna stay `0.0.0`

PR #3115, opened 2018-07-30
refactor(ci): change timeout behaviour

juliangieseke:
Cypress kept dying and with retry mechanisms in place there is no need anymore for running tests in parallel.

## Testing

verify that jenkins isnt running stages in parallel anymore 😢

## Trade-offs

Ci will take longer now… but IMHO thats better than having killed runs 🤷‍♂️

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -18,7 +18,7 @@ pipeline {
   }
 
   options {
-    timeout(time: 2, unit: "HOURS")
+    timeout(time: 5, unit: "HOURS")
     disableConcurrentBuilds()
   }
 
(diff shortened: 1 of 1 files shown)


DanielMSchmidt:
@juliangieseke Would you mind creating a JIRA for this and pulling it into the sprint? I would love to have it for some more additional context so that we know when we might enable them back in the future.

Jenkinsfile
```diff
@@ -18,7 +18,7 @@ pipeline {
   }
 
   options {
-    timeout(time: 2, unit: "HOURS")
+    timeout(time: 5, unit: "HOURS")
```
weblancaster: 😮

juliangieseke:
5 hours because this is only "stage two fallback" stages have their own timeouts 👍

PR #3117, opened 2018-07-30
DCOS-39832 elected leader tab

Fabs:
## Testing

Launch a multi-master cluster. Look at the data it shows, go on the control panel for your cluster (AWS) and kill the machine that is, running.

Go to: http://localhost:4200/#/nodes/masters

## Trade-offs

**GET_MASTER quickfix** 
The state store only updates the MesosState once it
parsed the entire state, which means that we wait seconds
on a large cluster, before we are able to show a piece
of information, we get almost instantly (GET_MASTER).

**GET_MASTER parser elected and started**
I added code to the parser to add elected and started time to the master_info object. Not quite sure I like adding those there but do not have any better idea. I considered another structure like master_events with both. I also do not like to add them flatten to the mesos state we keep

**Toggle elected and started time**
We dont have this behaviour as a reusable component, so I will create a ticket (ticket here) to extract this component to the UI-kit, we can then use it later.

## Dependencies

Depends on #3118 for tabs

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@ build/
 npm-debug.log
 node_modules/
 src/js/config/Config.dev.ts
+src/js/config/Config.dev.js
 serve
 tmp/
 *.orig
--- a/plugins/nodes/src/js/components/LeaderGrid.js
+++ b/plugins/nodes/src/js/components/LeaderGrid.js
@@ -1,0 +1,87 @@
+import React from "react";
+import { FormattedMessage } from "react-intl";
+import moment from "moment";
+
+import ConfigurationMap from "#SRC/js/components/ConfigurationMap";
+import ConfigurationMapHeading from "#SRC/js/components/ConfigurationMapHeading";
+import ConfigurationMapLabel from "#SRC/js/components/ConfigurationMapLabel";
+import ConfigurationMapRow from "#SRC/js/components/ConfigurationMapRow";
+import ConfigurationMapSection from "#SRC/js/components/ConfigurationMapSection";
+import ConfigurationMapValue from "#SRC/js/components/ConfigurationMapValue";
+import Loader from "#SRC/js/components/Loader";
+
+function timeFromNow(time) {
+  if (!time) {
+    return null;
+  }
+
+  return moment(time * 1000).fromNow();
+}
+
+const Loading = () => <Loader size="small" type="ballBeat" />;
+
+const ConfigurationRow = ({ keyValue, title, value }) => {
+  const loadedValue = value || <Loading />;
+
+  return (
+    <ConfigurationMapRow key={keyValue}>
+      <ConfigurationMapLabel>{title}</ConfigurationMapLabel>
+      <ConfigurationMapValue>{loadedValue}</ConfigurationMapValue>
+    </ConfigurationMapRow>
+  );
+};
+
+export default class LeaderGrid extends React.Component {
+  getMesosDetails() {
+    const {
+      hostPort,
+      version,
+      electedTime,
+      startTime,
+      region
+    } = this.props.master;
+
+    return (
+      <ConfigurationMapSection>
+        <ConfigurationRow
+          keyValue="leader"
+          title="IP and Port"
+          value={hostPort}
+        />
+
+        <ConfigurationRow keyValue="region" title="Region" value={region} />
+
+        <ConfigurationRow
+          keyValue="version"
+          title={<FormattedMessage id="COMMON.VERSION" />}
+          value={version}
+        />
+
+        <ConfigurationRow
+          keyValue="version"
+          title={<FormattedMessage id="COMMON.STARTED" />}
+          value={timeFromNow(startTime)}
+        />
+
+        <ConfigurationRow
+          keyValue="elected"
+          title={<FormattedMessage id="COMMON.ELECTED" />}
+          value={timeFromNow(electedTime)}
+        />
+      </ConfigurationMapSection>
+    );
+  }
+
+  render() {
+    return (
+      <div className="container">
+        <ConfigurationMap>
+          <ConfigurationMapHeading className="flush-top">
+            Leader
+          </ConfigurationMapHeading>
+          {this.getMesosDetails()}
+        </ConfigurationMap>
+      </div>
+    );
+  }
+}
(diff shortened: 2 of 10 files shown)


plugins/nodes/src/js/pages/NodesAgents.js
```diff
@@ -0,0 +1,322 @@
+import classNames from "classnames";
```
Fabs: @juliangieseke should this be part of my PR? I guess not

plugins/nodes/src/js/data/MesosMasters.js
```diff
@@ -0,0 +1,83 @@
...
+  return Object.keys(obj).length === 0;
+}
+
+function hostPort(address) {
+  return `${address.hostname}:${address.port}`;
+}
+
+// ask how to fix better
+function supportsRegion(master) {
+  return (
```
Fabs: Could that be done better?

plugins/nodes/src/js/data/MesosMasters.js
```diff
@@ -0,0 +1,83 @@
...
+
+function isEmptyObject(obj) {
+  return Object.keys(obj).length === 0;
+}
+
+function hostPort(address) {
+  return `${address.hostname}:${address.port}`;
+}
+
+// ask how to fix better
```
Fabs: remove that.

plugins/nodes/src/js/data/MesosMasters.js
```diff
@@ -0,0 +1,83 @@
...
+import "rxjs/add/operator/do";
+import "rxjs/add/operator/map";
+import "rxjs/add/operator/retry";
+import "rxjs/add/operator/filter";
+
+import MesosStateStore from "#SRC/js/stores/MesosStateStore";
+
+import LeaderGrid from "../components/LeaderGrid";
+
+function isEmptyObject(obj) {
```
Fabs: Probabbly we have this util.

src/js/stores/MesosStateStore.js
```diff
@@ -74,7 +79,14 @@ class MesosStateStore extends GetSetBaseStore {
     const getMasterRequest = request(
       { type: "GET_MASTER" },
       "/mesos/api/v1?get_master"
-    ).retryWhen(linearBackoff(RETRY_DELAY, MAX_RETRIES));
+    )
```
Fabs: @orlandohohmeier and @nLight guess we need to talk about this one

plugins/nodes/src/js/components/LeaderGrid.js
```diff
@@ -0,0 +1,87 @@
...
+
+function timeFromNow(time) {
+  if (!time) {
+    return null;
+  }
+
+  return moment(time * 1000).fromNow();
+}
+
+const Loading = () => <Loader size="small" type="ballBeat" />;
```
juliangieseke: Unnecessary

.gitignore
```diff
@@ -9,6 +9,7 @@ build/
 npm-debug.log
 node_modules/
 src/js/config/Config.dev.ts
+src/js/config/Config.dev.js
```
juliangieseke: Unnecessary

plugins/nodes/src/js/components/LeaderGrid.js
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  render() {
+    return (
+      <div className="container">
+        <ConfigurationMap>
+          <ConfigurationMapHeading className="flush-top">
+            Leader
+          </ConfigurationMapHeading>
+          {this.getMesosDetails()}
```
juliangieseke: IMHO there is no need extracting this into a function… there is no additional logic 🤔

plugins/nodes/src/js/data/MesosMasters.js
```diff
@@ -0,0 +1,91 @@
...
+  return (
+    master.domain &&
+    master.domain.fault_domain &&
+    master.domain.fault_domain.region &&
+    master.domain.fault_domain.region.name
+  );
+}
+
+export function getRegion(master) {
+  return supportsRegion(master)
```
juliangieseke: 😍

plugins/nodes/src/js/data/MesosMasters.js
```diff
@@ -0,0 +1,91 @@
...
+    ? master.domain.fault_domain.region.name
+    : "N/A";
+}
+
+const STORE_POLL_INTERVAL = 2000;
```
juliangieseke: Please use Config.pollinginterval

plugins/nodes/src/js/data/MesosMasters.js
```diff
@@ -0,0 +1,91 @@
...
+    hostPort: undefined,
+    version: undefined,
+    electedTime: undefined,
+    startTime: undefined,
+    region: undefined
+  };
+}
+
+export function mesosMasterInfo(
+  masterDataSource,
```
juliangieseke: Typescript would have told me this is a function - can you add JSDoc comments? 😇

.gitignore
Fabs: I added because more than once I ended up committing that file by accident. Also I got external contributors to come to ask me about what to do with this file. On my setup, I usually use older clusters in the same branch, and I have to have both files.

Do you think this is enough reason to keep it?

plugins/nodes/src/js/data/MesosMasters.js
Fabs: This variable does not have the same meaning of the Config.pollinginterval, although it has the same value. I can move it to config, but I explicitly do not wanna use the same value.

plugins/nodes/src/js/data/MesosMasters.js
Fabs: Can you teach me? I do not know JSDoc.

.gitignore
weblancaster: doesn't hurt in having it added.

plugins/nodes/src/js/components/LeaderGrid.js
Fabs: I like your style comment, if there are something people ask me to change, I will go also and fix your comments. Otherwise I will roll with it.

Fabs:
This build is broken, but it is building here #3127

juliangieseke:
continued in #3127

PR #3119, opened 2018-07-30
feat(ClusterDropdown): authenticated cluster header label

weblancaster:
> Be aware this PR does not contain the styles for the header (@tnguyen0 has been working on it).

Commit adds authenticated cluster dropdown label mechanism
to display the correct info when metadata available to the user
or to display the hardcoded string.

Closes DCOS-39679

## Testing

1. `npm start`
2. Go to `HeaderBar.tsx` comment out line 12 - 14.
3. Check out corresponding branch on privates plugin

**Case 1:**
User has admin permissions (access to metadata): user should see the cluster name and IP on the header.
jul-31-2018 16-41-41

**Case 2**:
User doesn't have access to metadata = Cluster dropdown on Header displays `cluster` string only.
jul-31-2018 16-40-16

**How to create a user without metadata access**
jul-31-2018 17-55-40


plugins/oauth/components/AuthenticatedClusterDropdown.js
```diff
@@ -0,0 +1,14 @@
+import ClusterDropdown from "#SRC/js/components/ClusterDropdown";
+
+class AuthenticatedClusterDropdown extends ClusterDropdown {
```
juliangieseke: is it okay to extend react components? AFAIK react doesnt really support this pattern 🤔

src/js/components/HeaderBar.tsx
```diff
@@ -2,7 +2,7 @@ import * as React from "react";
 
 import { Link } from "react-router";
 import { HeaderBar as UIHeaderBar } from "ui-kit-stage/HeaderBar";
-import ClusterDropdown from "./ClusterDropdown";
+import ClusterHeader from "./ClusterHeader";
```
juliangieseke: why the rename? 🤔

juliangieseke:
^ and it is blocking EE PR…

nLight:
Let's please _not rush_ the merges even if they are blocking something when they are NOT of a quality.

plugins/oauth/hooks.js
```diff
@@ -14,7 +14,7 @@ import CookieUtils from "#SRC/js/utils/CookieUtils";
 import RouterUtil from "#SRC/js/utils/RouterUtil";
 import UsersPage from "#SRC/js/pages/system/UsersPage";
 
-import AuthenticatedUserAccountDropdown from "./components/AuthenticatedUserAccountDropdown";
+import AuthenticatedClusterDropdown from "./components/AuthenticatedClusterDropdown";
```
tnguyen0: The changes in these file are therefore not needed.

weblancaster: Discussed offline.. we need the file.

PR #3118, opened 2018-07-30
DCOS-39153: add tabs to NodesPage

juliangieseke:
Closes DCOS-39153

## Testing

navigate to http://localhost:4200/#/nodes/agents to see tabs. Agents page should work as before. masters tab is filled by #3117

PR #3120, opened 2018-07-31
DCOS-39712: Nodes Masters Tab

juliangieseke:
This Feature branch introduces Agents and Masters tabs. In its first iteration it will only contain a list of Details for the current leader.

Resolves  http://jira.mesosphere.com/browse/DCOS-39712

## Pull Requests in this Branch
- https://github.com/dcos/dcos-ui/pull/3118
- https://github.com/dcos/dcos-ui/pull/3127
- https://github.com/dcos/dcos-ui/pull/3156

## Testing

Before merging this branch: Navigate through the application, make sure all links pointing to nodes are still working and that agents and masters tabs are also working as expected. 

## Trade-offs

Due to lack of APIs, we decided not to implement a table yet, but go with a quick first implementation that only displays some data regarding the fcurrent leader. more info in JIRA.

**Update: 2018-08-14**: We'd like to merge this as-is and later add #3156 to master directly, reason is that there are some things I'd like to be discussed between @Fabs and @nLight and this branch already is in good shape, only thing thats missing is the "non leader list". everything else should be good to merge 👍 
/cc @orlandohohmeier 

## Dependencies

None

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@ build/
 npm-debug.log
 node_modules/
 src/js/config/Config.dev.ts
+src/js/config/Config.dev.js
 serve
 tmp/
 *.orig
--- a/plugins/nodes/src/js/components/LeaderGrid.js
+++ b/plugins/nodes/src/js/components/LeaderGrid.js
@@ -58,7 +58,7 @@ export default function LeaderGrid({ master }) {
           />
 
           <ConfigurationRow
-            keyValue="version"
+            keyValue="started"
             title={<FormattedMessage id="COMMON.STARTED" />}
             value={timeFromNow(master.startTime)}
           />
--- a/plugins/nodes/src/js/components/LeaderGrid.js
+++ b/plugins/nodes/src/js/components/LeaderGrid.js
@@ -31,57 +31,45 @@ const ConfigurationRow = ({ keyValue, title, value }) => {
   );
 };
 
-export default class LeaderGrid extends React.Component {
-  getMesosDetails() {
-    const {
-      hostPort,
-      version,
-      electedTime,
-      startTime,
-      region
-    } = this.props.master;
+export default function LeaderGrid({ master }) {
+  return (
+    <div className="container">
+      <ConfigurationMap>
+        <ConfigurationMapHeading className="flush-top">
+          Leader
+        </ConfigurationMapHeading>
+        <ConfigurationMapSection>
+          <ConfigurationRow
+            keyValue="leader"
+            title="IP and Port"
+            value={master.hostPort}
+          />
 
-    return (
-      <ConfigurationMapSection>
-        <ConfigurationRow
-          keyValue="leader"
-          title="IP and Port"
-          value={hostPort}
-        />
+          <ConfigurationRow
+            keyValue="region"
+            title="Region"
+            value={master.region}
+          />
 
-        <ConfigurationRow keyValue="region" title="Region" value={region} />
+          <ConfigurationRow
+            keyValue="version"
+            title={<FormattedMessage id="COMMON.VERSION" />}
+            value={master.version}
+          />
 
-        <ConfigurationRow
-          keyValue="version"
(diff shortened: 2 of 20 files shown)


Jenkinsfile
```diff
@@ -2,7 +2,7 @@
 
 @Library("sec_ci_libs@v2-latest") _
 
-def master_branches = ["master", ] as String[]
```
nLight: We don't want to add `feature/DCOS-39712-masters-tab` to master branches on master, do we?

juliangieseke: we can remove it on this branch, but then we have to approve every single CI run until it got merged…
The idea was to open a cleanup PR directly after this has landed 🤔

nLight: Why will we need approving PRs? Following Prs will go against master, not this branch as I understand.

juliangieseke: not PR, Jenkins CI run (security library…)

nLight: As I understand we have it here because otherwise every other PR against this `feature/DCOS-39712-masters-tab` branch will be considered unsecure. But now we are about to merge the feature branch itself, so there will be no PRs against the feature branch, correct?

juliangieseke: no, its here because every run of `continuous-integration/jenkins/branch` would be considered insecure, PRs arent affected by this line.

nLight: Ok and we want to run branch CI? How every other normal PR works then?

juliangieseke: check this (giant) if block here: https://github.com/mesosphere/sec-ci-libs/blob/master/vars/user_is_authorized.groovy#L77

depending on the given env vars the script checks for different things, as stated in your slack-update-pr, `CHANGE_BRANCH` and `CHANGE_AUTHOR` are set for PRs, so the lib is able to check the author. on a "branch run" both are null which means that the lib has no chance to check anything - so it asks us for permission… this list acts a whitelist (and while typing this, i realize that it is a bad idea to have a non-protected branch in this whitelist :/ -> will try to figure out a solution ASAP)

juliangieseke: @orlandohohmeier ^ fyi :/

nLight: Ok, so we need this to be whitelisted because we want to run `continuous-integration/jenkins/branch` for feature branches, correct? My point is as we are going to merge the feature branch itself into master, wouldn't it be enough just having the `continuous-integration/jenkins/pr-merge` to pass?

juliangieseke: hehe, fair point 😅once we're going to "use" the PR itself, there is no need to the branch CI to run… only thing it that jenkins will annoy us in our channel a few times 🤔

nLight: It's just that I wanted to ✅ and merge the branch if we update the whitelist. It's good to go without non-leading masters in my opinion.

juliangieseke:
rebased, removed jenkins commit and also disabled jenkins for feature branches for now - will try to find a better solution

juliangieseke:
hmmm… that didnt work as expected… 1sec 🚒

juliangieseke:
fixed 🤞

Fabs:
We reversed behavior on the arrows. Is that indented?

Before:
image

After:
image

Fabs:
When filtering. Information on CPU/MEM and etc goes to 0, before it shows. This behaviour was not like that in the past (just flagging, not a blocker)
https://cl.ly/001G470k2z46

Fabs:
I love ❤️ that the filter responds immediately now.

juliangieseke:
re reversed sorting: yes, we discovered the bug in the old implementaion while implementing new table component and decided to fix it instead of reimplementing the old (broken) way 👍

juliangieseke:
moved back to WIP because fabs PR is ready to review 👍 
=> https://github.com/dcos/dcos-ui/pull/3156

juliangieseke:
So, I think this is done & Ready for review 👍 
@Fabs any objections? can you verify? 🙇

PR #3124, opened 2018-07-31
DCOS-39160: implement columns for nodes table

DanielMSchmidt:
## Testing

See the nodes table, compare it to the mockup linked in the ticket & check if IP sorting works as expected.

## Trade-offs

- The status column is not there, so it's not a part of this anymore
- I used an external package for parsing the IP, because I don't see any sense in reimplementing it

## Dependencies

We need to check with @jaxesn if this is a new feature or just a polishing that can still make it. Therefore hold merge for now.


plugins/nodes/src/js/components/NodesTable.tsx
```diff
@@ -1,17 +1,13 @@
 import * as React from "react";
-import { Table, Column } from "@dcos/ui-kit";
+import { Table, Column } from "@dcos/ui-kit/dist/packages";
 
 import NodesList from "#SRC/js/structs/NodesList";
 import Node from "#SRC/js/structs/Node";
 
 import { SortableColumnHeader } from "ui-kit-stage/SortableColumnHeader";
-import { SortDirection } from "../types/SortDirection";
+import { SortDirection } from "#PLUGINS/nodes/src/js/types/SortDirection";
```
nLight: why?

plugins/nodes/src/js/columns/NodesTableIpColumn.tsx
```diff
@@ -7,10 +7,11 @@ import { Link } from "react-router";
 import { Tooltip } from "reactjs-components";
 import Icon from "#SRC/js/components/Icon";
 import { TextCell } from "@dcos/ui-kit";
+import ipToInt from "ip-to-int";
```
nLight: It doesn't support ipv6 and there's no related tradeoff. Please verify that ipv6 is not a requirement.

plugins/nodes/src/js/components/NodesTable.tsx
DanielMSchmidt: Plugin did shit -.- Removed the plugin

plugins/nodes/src/js/columns/NodesTableIpColumn.tsx
DanielMSchmidt: Would you rather drop the improvement? I checked with customer support, there are customers on ipv6, so they would fall back to the string sorting.

nLight:
flicts

nLight:
More flicts!

plugins/nodes/src/js/components/NodesTable.tsx
nLight: please push the changes

DanielMSchmidt:
We should be fine to merge this and go ahead 👍

PR #3128, opened 2018-08-01
DCOS-39901: NodesTable height of container should fill remaining space

DanielMSchmidt:
The NodesTable only showed two rows as this container was too small

Closes DCOS-39901

## Testing

Run this branch against soak and see if the nodes table has the entire height of the leftover space of the screen and also if there are not two scroll bars

## Trade-offs

- We had to add a display: flex to one of the main gemini scrollbar components. If there is a different solution I am not aware of, please tell.

## Dependencies

None

### DIFF
--- a/plugins/nodes/src/js/components/NodesTable.tsx
+++ b/plugins/nodes/src/js/components/NodesTable.tsx
@@ -188,151 +188,153 @@ export default class NodesTable extends React.Component<
     const { data, sortColumn, sortDirection } = this.state;
 
     return (
-      <Table data={data.slice()}>
-        <Column
-          header={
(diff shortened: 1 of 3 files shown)


src/styles/components/scrollbar/styles.less
```diff
@@ -67,6 +67,7 @@
     position: relative;
 
     .gm-scroll-view {
+      display: flex;
```
juliangieseke: this is a pretty big change, are we sure that is doesnt break something else? 😨

DanielMSchmidt: You can never be sure, but I tried my best to verify that nothing broke

PR #3129, opened 2018-08-01
ci(timeouts): change timeout behaviours

juliangieseke:
Jenkins isnt able to retry a timed out step, moved timeout to bash...

## Testing

on this PR it just have to be green - on master we need to wait for a failed & restarted build :( 

## Trade-offs

this introduces a maximum runtime of 20 minutes per test file… 🤷‍♂️


Dockerfile
```diff
@@ -25,7 +25,7 @@ RUN set -x \
   # Install cypress dependencies & JRE (required by Jenkins)
   && echo 'deb http://ftp.debian.org/debian jessie-backports main' >> /etc/apt/sources.list \
   && apt-get update \
-  && apt-get install -y xvfb libgtk2.0-0 libnotify-dev libgconf-2-4 libnss3 libxss1 lsof \
+  && apt-get install -y xvfb libgtk2.0-0 libnotify-dev libgconf-2-4 libnss3 libxss1 lsof coreutils \
```
DanielMSchmidt: why is this needed? Could you split it up to an extra commit so we can, later on, identify why it was added?

scripts/run-integration-tests
```diff
@@ -50,7 +50,7 @@ PLUGINS_PATH=${PLUGINS_PATH:-''}
 # FILE: file to test
 # OUTPUT_PATH: Directory where results are saved (e.g. "some/path/to/Testfile-cy.js-1")
 function executeCypress {
-  npm run --silent cypress -- \
+  timeout 20m npm run --silent cypress -- \
```
DanielMSchmidt: This kills the run after 20 minutes?

Dockerfile
Fabs: Coreutils is added so we can use the timeout function.

scripts/run-integration-tests
Fabs: Yes: https://www.gnu.org/software/coreutils/manual/html_node/timeout-invocation.html

Fabs:
Could we have a ticket, so this will be part of the board where we pick thinks to review and so we also make it accountable on the sprint?

PR #3133, opened 2018-08-02
ci(sr): add debug flag to investigate missing pr comments

juliangieseke:
## Testing

it only happens on jenkins, no way to test it 🤷‍♂️


nLight:
Do we actually want to merge it? If so let's please not use obscure `sr` as the scope

juliangieseke:
@nLight updated commit message 👍 /cc @weblancaster

juliangieseke:
looks like they already found a bug which could lead to the issues we have, https://github.com/semantic-release/github/pull/100 should fix it 👍 

=> lets wait for the PR to land and be released, update sr in our repo and see if it works.

PR #3135, opened 2018-08-06
[1.11] DCOS-39877: remove flickering filter bug

DanielMSchmidt:
## Testing

- Check the tasks page for a service and verify that the fuzzy search loses it's content if "apply" is not pressed directly
- Check this branch out and verify that the content of the fuzzy search is directly applied

### Before
old

### After

new

## Trade-offs

- There might have been a way to fix the bug without applying the fuzzy search directly to the search, but this was inconsistent with the buttons already (they apply directly) I decided to fix forward here and change the behavior (from my point of view) to the benefit of the customer.
- Decided for integration tests as there were no tests at all for this area. I thought system tests might not be necessary as it's more of a UI-only problem.

## Dependencies

- [x] Would like to have a small nod from @AmrAbdelrazik, @lilysamimi or @leemunroe as we have a little behaviour change here

### DIFF
--- a/plugins/services/src/js/components/dsl/FuzzyTextDSLSection.js
+++ b/plugins/services/src/js/components/dsl/FuzzyTextDSLSection.js
@@ -11,6 +11,10 @@ import FieldInput from "#SRC/js/components/form/FieldInput";
 import FieldLabel from "#SRC/js/components/form/FieldLabel";
 import FormGroup from "#SRC/js/components/form/FormGroup";
 
+import DSLFilterTypes from "#SRC/js/constants/DSLFilterTypes";
+import { FilterNode } from "#SRC/js/structs/DSLASTNodes";
+import DSLUpdateUtil from "#SRC/js/utils/DSLUpdateUtil";
+
 const EXPRESSION_PARTS = {
   text: DSLExpressionPart.fuzzy
 };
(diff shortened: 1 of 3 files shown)


leemunroe:
@DanielMSchmidt Looks solid 👍🚢

DanielMSchmidt:
Thank you for taking the time @leemunroe and @lilysamimi

lilysamimi:
@tnguyen0 @DanielMSchmidt wouldn't change the name of the button. Ideally, we want to get rid of the button and just have the data change in real-time as filters are selected.

tests/pages/services/ServiceDetail-cy.js
```diff
@@ -89,6 +89,91 @@ describe("Service Detail Page", function() {
         cy.hash().should("match", /services\/detail\/%2Fsleep\/volumes.*/);
       });
     });
+
+    context("Filter Tasks", function() {
+      const DEFAULT_ROWS = 3; // Headline plus invisible rows
+      beforeEach(function() {
+        cy.configureCluster({
```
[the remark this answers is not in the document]
DanielMSchmidt: Will change this :+1:

DanielMSchmidt:
Updated it @tnguyen0 :)

DanielMSchmidt:
Yeah, I should have created a sub-ticket 🙈 Originally the other PR was only about forward porting the integration tests, at that point that still kinda made sense

src/js/components/DSLFormDropdownPanel.js
```diff
@@ -56,6 +56,7 @@ class DSLFormDropdownPanel extends React.Component {
    */
   handleChange(expression) {
     this.setState({ expression });
+    this.props.onChange(expression);
```
nLight: Wonder if we should we go with 
```js
this.setState({ expression }, () => this.props.onChange(expression));
```

plugins/services/src/js/components/dsl/FuzzyTextDSLSection.js
```diff
@@ -40,13 +44,37 @@ class FuzzyTextDSLSection extends React.Component {
   }
 
   handleTextChange(event) {
+    const { onChange, expression } = this.props;
     const { target } = event;
     event.stopPropagation();
+
+    const value = target.value;
```
nLight: Looks like we could just go with `const { target: { value } } = event;`

src/js/components/DSLFormDropdownPanel.js
DanielMSchmidt: This is the exact same code we have in master that fixed it, do you want me to change it in both or in none of them?

nLight: I expected this to be a cherry-pick, yes. I just wonder if it's ok to have those values out of sync because `setState` is an async function. Sometimes it leads to interesting bugs. So I'm just asking whether we should change it or we totally sure values being out of sync is not an issue.

DanielMSchmidt: I haven't experienced any problems in that regard, but of course you can never be 100% sure. I can fix it and update it on master, too. This time I am going to create a new subticket

plugins/services/src/js/components/dsl/FuzzyTextDSLSection.js
Poltergeist: Correct.

DanielMSchmidt:
@tnguyen0 Could you take another look?

PR #3142, opened 2018-08-06
DCOS-39972: fix spacing between node dials and filter

tnguyen0:
* Fixes the spacing between node dials and filter row when the size of the screen is changed
* Fixes the header bar shrinking issue
* Adds a dot for "Other" in the side list
* Adds scrolling for the nodes grid view

Closes DCOS-39972

## Testing

* To see a large number of nodes, set the cluster to a soak cluster
* Verify that the aforementioned points are fixed/added

### DIFF
--- a/plugins/nodes/src/js/components/NodesGridView.js
+++ b/plugins/nodes/src/js/components/NodesGridView.js
@@ -87,7 +87,7 @@ var NodesGridView = React.createClass({
     // Add 'Others' node to the list
     if (activeServiceIds.length > MAX_SERVICES_TO_SHOW) {
       var classNameOther =
-        "service-legend-color service-color-" + OTHER_SERVICES_COLOR;
+        `dot service-color-${OTHER_SERVICES_COLOR}`;
       items.push(
         <li key="other">
           <span className={classNameOther} />
--- a/src/styles/components/filter-bar/styles.less
+++ b/src/styles/components/filter-bar/styles.less
@@ -2,6 +2,7 @@
 
   .filter-bar {
     display: flex;
+    flex-shrink: 0;
     flex-wrap: wrap-reverse;
     width: 100%;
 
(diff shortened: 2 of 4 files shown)


plugins/nodes/src/js/components/NodesGridView.js
```diff
@@ -87,7 +87,7 @@ var NodesGridView = React.createClass({
     // Add 'Others' node to the list
     if (activeServiceIds.length > MAX_SERVICES_TO_SHOW) {
       var classNameOther =
-        "service-legend-color service-color-" + OTHER_SERVICES_COLOR;
+        "dot service-color-" + OTHER_SERVICES_COLOR;
```
weblancaster: I would use template strings ``dot service-color-${OTHER_SERVICES_COLOR}`` no gain though.

src/styles/components/filter-bar/styles.less
```diff
@@ -2,6 +2,7 @@
 
   .filter-bar {
     display: flex;
+    flex-shrink: 0;
```
weblancaster: Can't this just be removed? I removed and worked fine.

src/styles/components/header-bar/styles.less
```diff
@@ -5,7 +5,7 @@
     display: flex;
     flex-basis: auto;
     flex-grow: 0;
-    flex-shrink: 1;
+    flex-shrink: 0;
```
weblancaster: same as https://github.com/dcos/dcos-ui/pull/3142/files#r208385484

weblancaster:
Although it works I believe there is something else happening and this fix is a workaround but not really the root cause.

You can notice in the screenshot below that the scrollbar is in the nodesTable grid/ container and not in the page.

screen shot 2018-08-07 at 2 33 15 pm

DanielMSchmidt:
Cypress died, retriggered build

PR #3143, opened 2018-08-06
fix(NodesTable): add consistency to link

weblancaster:
> I started to build a new component in the UI Kit called `Link` to wrap the `Link` component from React Router but because it's out of scope I returned to the original fix.

Fix link in NodesTable to add consistency
across tables.

Closes DCOS-39959

## Testing
1. `npm start`
2. Go to Nodes page and note the fix in the column name

**Before**
screen shot 2018-08-06 at 3 38 23 pm

**After**
screen shot 2018-08-06 at 3 34 46 pm

### DIFF
--- a/plugins/nodes/src/js/components/NodesTable.tsx
+++ b/plugins/nodes/src/js/components/NodesTable.tsx
@@ -188,7 +188,7 @@ export default class NodesTable extends React.Component<
     const { data, sortColumn, sortDirection } = this.state;
 
     return (
-      <div style={{ flexGrow: 1 }}>
+      <div className="table-wrapper">
         <Table data={data.slice()}>
           <Column
             header={
--- a/src/styles/content/tables/styles.less
+++ b/src/styles/content/tables/styles.less
@@ -1,3 +1,13 @@
+.table-wrapper {
+  flex-grow: 1;
+
+  .table-cell-link-primary {
+    color: @grey-dark;
+    font-weight: 600;
+    text-decoration: none;
+  }
+}
+
 & when (@table-enabled) {
 
   .table {


src/styles/content/tables/styles.less
```diff
@@ -1,3 +1,14 @@
+@greyDark: #1b2029;
+
+.table-wrapper {
+
+  .table-cell-link-primary {
+    color: @greyDark;
```
tnguyen0: You can just use `@grey-dark` here.

plugins/nodes/src/js/components/NodesTable.tsx
```diff
@@ -188,7 +188,7 @@ export default class NodesTable extends React.Component<
     const { data, sortColumn, sortDirection } = this.state;
 
     return (
-      <div style={{ flexGrow: 1 }}>
+      <div className="table-wrapper" style={{ flexGrow: 1 }}>
```
tnguyen0: Perhaps put the inline-styling as part of the `table-wrapper` class?

weblancaster: Agree.

src/styles/content/tables/styles.less
weblancaster: true.

plugins/nodes/src/js/components/NodesTable.tsx
tnguyen0: What do you mean by not accessible. It works when tried.

weblancaster: nvm, I forgot it's `grey-dark` instead of `greyDark`

DanielMSchmidt:
Retriggered build, it crashed on docker build

PR #3148, opened 2018-08-08
feat(README): improve dcos ui  instructions

weblancaster:
As all of us do not have the cluster setup
running locally, I'm moving this step to last and optional
to not cause confusion nor developers waste time setting up

This is based on @mperrotti feedback.


README.md
```diff
@@ -22,33 +22,34 @@ You can choose from two ways of developing this repository.
...
   npm i
   ```
 
-4. Add developer files for config overrides:
+3. Add developer files for config overrides:
 
   ```sh
   npm run scaffold
   ```
   This will create `webpack/proxy.dev.js` and `src/js/config/Config.dev.ts`. These files aren't tracked by git and provide a place to override proxy and general dcos-ui configuration. Edit `webpack/proxy.dev.js` to point to the  address of the `dcos-vagrant` machine.
```
juliangieseke: you removed the vagrant step but still referring to it here

DanielMSchmidt:
Might be a dumb question, but as we are touching on this: Would it make sense to write usage instructions for dcos-e2e? Afaik you can launch a docker cluster easily with it.

README.md
weblancaster: Agree. I updated the text to say cluster instead vagrant

weblancaster:
@juliangieseke can you take a look at this again please? thank you.

GeorgiSTodorov:
I like this, but I think leaving the `installing a local cluster` as an optional first (and not last) step is worth considering.

juliangieseke:
hm, not sure if having it as the first step only adds confusion - but daniel is correct, we should consider adding multiple _external_ installation techniques to this readme 🤔

PR #3151, opened 2018-08-09
DCOS-39896: add region information getter to struct

juliangieseke:
## Testing

This PR only adds the getter to our struct - `npm test`

## Trade-offs

I decided to sort the regions by taskcount and generate & store the array on first read

### DIFF
--- a/plugins/services/src/js/structs/Service.js
+++ b/plugins/services/src/js/structs/Service.js
@@ -7,6 +7,10 @@ import ServiceSpec from "./ServiceSpec";
 import VolumeList from "./VolumeList";
 
 module.exports = class Service extends Item {
+  constructor() {
+    super(...arguments);
+    this._regions = undefined;
+  }
   getId() {
     return this.get("id") || "";
   }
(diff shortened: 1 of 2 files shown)


plugins/services/src/js/structs/Service.js
```diff
@@ -54,6 +58,27 @@ module.exports = class Service extends Item {
...
+    if (!this._regions) {
+      const regionCounts = (this.get("tasks") || []).reduce(
+        (regions, { region }) => {
+          if (region) {
+            regions[region] = regions[region] ? regions[region] + 1 : 1;
+          }
+
+          return regions;
+        },
+        []
```
jaxesn: clearly it works, but wouldn't this be more appropriate as an object vs array?

plugins/services/src/js/structs/__tests__/Service-test.js
```diff
@@ -32,6 +32,23 @@ describe("Service", function() {
     });
   });
 
+  describe("#getRegions", function() {
```
jaxesn: might be worth adding a test for the case where the service has no tasks

plugins/services/src/js/structs/Service.js
juliangieseke: Oops, it was meant to be an object actually. Good catch!

PR #3154, opened 2018-08-09
DCOS-39895: Region Information in Services Table

juliangieseke:
## Feature Branch

This is a feature branch and code already was reviewed, this PR has to be checked for overall functionality

## Testing

when this is done there are region information on services table

## Screenshot
screen shot 2018-08-24 at 12 48 02
(only region column in this PR)

### DIFF
--- a/plugins/services/src/js/containers/services/ServicesTable.js
+++ b/plugins/services/src/js/containers/services/ServicesTable.js
@@ -15,6 +15,7 @@ import ResourceTableUtil from "#SRC/js/utils/ResourceTableUtil";
 import TableUtil from "#SRC/js/utils/TableUtil";
 import Units from "#SRC/js/utils/Units";
 import { isSDKService } from "#SRC/js/utils/ServiceUtil";
+import CompositeState from "#SRC/js/structs/CompositeState";
 import ServiceStatusProgressBar from "../../components/ServiceStatusProgressBar";
 import Pod from "../../structs/Pod";
 import Service from "../../structs/Service";
@@ -46,6 +47,7 @@ const columnClasses = {
   name: "service-table-column-name",
   status: "service-table-column-status",
   version: "service-table-column-version",
+  regions: "service-table-column-regions",
   instances: "service-table-column-instances",
   cpus: "service-table-column-cpus",
   mem: "service-table-column-mem",
@@ -60,6 +62,7 @@ const METHODS_TO_BIND = [
   "handleActionDisabledModalOpen",
   "handleActionDisabledModalClose",
   "renderHeadline",
+  "renderRegions",
   "renderStats",
   "renderStatus",
   "renderServiceActions"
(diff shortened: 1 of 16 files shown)


plugins/services/src/js/containers/services/ServicesTable.js
```diff
@@ -242,6 +245,30 @@ class ServicesTable extends React.Component {
...
+  renderRegions(prop, service) {
+    const localRegion = CompositeState.getMasterNode().getRegionName();
+    let regions = service.getRegions();
+
+    regions = regions.map(
+      region => (region === localRegion ? region + " (Local)" : region)
+    );
+
+    if (regions.length === 0) {
+      regions.push("N/A");
```
weblancaster: We are starting to have this `N/A` all over the code base, I would rather have a stateless component that renders that. I also record that at some point last year we wanted to replace `N/A` to `-` and I believe that we created a helper or a component that returns `-`

plugins/services/src/js/structs/Service.js
```diff
@@ -7,6 +7,10 @@ import ServiceSpec from "./ServiceSpec";
 import VolumeList from "./VolumeList";
 
 module.exports = class Service extends Item {
+  constructor() {
+    super(...arguments);
+    this._regions = undefined;
```
weblancaster: why `undefined` and not `null`?

plugins/services/src/js/structs/__tests__/PodInstance-test.js
```diff
@@ -52,6 +52,32 @@ describe("PodInstance", function() {
     });
   });
 
+  describe("#getAgentRegion", function() {
+    it("returns the correct value", function() {
```
weblancaster: Can we have a more descriptive title for all tests, please? e.g https://github.com/dcos/dcos-ui/pull/3154/files#diff-898ccecc5fbc3897c2529139b461bcccR78

plugins/services/src/js/containers/services/ServicesTable.js
mperrotti: There was also a recent case where the design team preferred having a blank cell. Should we talk to design about how we decide when to use `N/A` vs `-` vs nothing?

weblancaster: @mperrotti good point. I remember that case too now which was for services version.

juliangieseke: Yeah, I like the Idea of having small `<Empty />` and `<NotAvailable />` components to keep things aligned. For the empty case we also switch between `null`, `<div />` and `<span />` 😕

However I'd say it is not in scope for this feature branch to create these components, we can clean up the code later - @weblancaster @mperrotti can one of you create Epics and facilitate them? 🙇

plugins/services/src/js/structs/Service.js
juliangieseke: why `null` and not `undefined`? :trollface: 

scnr 😅  - havent thought about it when writing the code 🤷‍♂️ I'm not aware that we use one of them consistently thorough our codebase. Do we want to do so?

juliangieseke: thinking about it now, an empty array would have been better to be typesafe 🤔 will change 👍

juliangieseke:
rephrased test description and rebased branch. /cc @GeorgiSTodorov 🙇

PR #3160, opened 2018-08-13
chore(eslint): raise all warnings instead of fixing them

nLight:
We discovered that it was possible to commit code with linting errors. I figured that `--fix` would try fixing everything instead of raising errors.

### DIFF
--- a/package.json
+++ b/package.json
@@ -186,7 +186,7 @@
     "prebuild": "./scripts/validate-tests",
     "build": "npm run lint && ([ -z \"${npm_config_externalplugins}\" ] || npm run lint-plugins) && npm test && npm run build-assets && npm run validate-build",
     "lint": "npm run stylelint && npm run eslint && npm run tslint",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -186,7 +186,7 @@
     "prebuild": "./scripts/validate-tests",
     "build": "npm run lint && ([ -z \"${npm_config_externalplugins}\" ] || npm run lint-plugins) && npm test && npm run build-assets && npm run validate-build",
     "lint": "npm run stylelint && npm run eslint && npm run tslint",
-    "eslint": "bash -c 'opts=$@ ; set -f; if [[ $# -lt 1 ]]; then opts=\"{src/js,plugins,packages}/**/*.js\"; fi ; eslint --fix --quiet ${opts}' 0",
+    "eslint": "bash -c 'opts=$@ ; set -f; if [[ $# -lt 1 ]]; then opts=\"{src/js,plugins,packages}/**/*.js\"; fi ; eslint --quiet ${opts}' 0",
```
juliangieseke: what does `--quiet` do in this case? suppressing warnings? :trollface:

nLight: We don't care about warnings, but all the linting errors propagated when I tested.

PR #3163, opened 2018-08-14
DCOS-39822: add system tests for Nodes Table

DanielMSchmidt:
## Testing

- Verify that system tests test what they should
- Run system tests a couple of times to see if the tests themselves don't introduce any flakiness.

## Trade-offs

- Wanted to test if a spawned task is spawned on the right node, but writing an adequate selector for this was next to impossible without making it brittle via some `div:nth-child(4) > div:nth-child(2)` alike selectors.
- Could not test the detail page (wanted to copy the integration test for that), somehow cypress internal proxy seems to fail me here with this long lasting XHR request
	- Tried `cy.server.whitelist`
	- Searched for global "ignore" flags or disabling this altogether
- Wanted to check if the right nodes are shown rather then the the right amount, but passing this info down to cypress was harder then anticipated. I was thinking about giving it a second try via [`cy.exec`](https://docs.cypress.io/api/commands/exec.html#Arguments) but the certainty we gain was not worth the time spent IMHO

## Dependencies

None


nLight:
>  long lasting XHR request

I hope #3131 will resolve this

nLight:
What was the exact challenge passing the nodes down?

DanielMSchmidt:
> What was the exact challenge passing the nodes down?

wouldn't get passed down no matter what I did (process.env, cypress.env). After spending like 30 minutes I decided that it doesn't provide so much value

nLight:
How did you try to do that?

nLight:
ping @DanielMSchmidt

nLight:
WIP: because I'd like having this PR improved. Glad to help

DanielMSchmidt:
Sorry for the inactivity @nLight, my last comment doesn't seem to have been sent.

I got the nodes via the CLI as a string of ips, that went great. I passed this string down to the system tests via the cypress -e flag. From there on it was removed. At this point I spent a lot of time on other things in this PR, so I rethought if the test is useful. The only thing I am really testing is if the CLI and UI show the same nodes, so if one shows them faulty due to a change and the other doesn't we catch it here. For me this was not worth the effort. 

Is this the test you would like to see more completely? What do you want to secure against?

nLight:
System tests are red and looks like a legit fail:
```
✘ Nodes shows all nodes as healthy
```

system-tests/nodes/test-nodes.js
```diff
@@ -0,0 +1,21 @@
...
+  ".BottomLeftGrid_ScrollWrapper .ReactVirtualized__Grid__innerScrollContainer a";
+
+describe("Nodes", function() {
+  afterEach(() => {
+    cy.window().then(win => {
+      win.location.href = "about:blank";
+    });
+  });
+
+  it("shows all nodes as healthy", () => {
```
nLight: `shows correct amount of nodes`?

Dockerfile
```diff
@@ -44,7 +44,8 @@ RUN set -x \
   && ln -sf /bin/bash /bin/sh \
   # Fix system tests as long as upstream dependency has errors
   && pip install 'six==1.10.0' \
-  && pip install 'python-dateutil==2.6.0'
+  && pip install 'python-dateutil==2.6.0' \
+  && pip install 'PyYAML==3.12'
```
nLight: Do we still need this?

system-tests/nodes/test-nodes.js
```diff
@@ -0,0 +1,21 @@
...
+    cy.visitUrl("/nodes");
+    cy.get(hostnameSelector).should("have.length", 2);
+  });
+
+  it("shows all nodes as healthy", () => {
+    cy.get(
+      ".ReactVirtualized__Grid__innerScrollContainer > div:contains('Healthy')"
```
nLight: Why not extracting this as a const as well? Shall we write some cypress helpers for the new table?

DanielMSchmidt: Would rather make the new table a bit more approachable with a few extra classes in the future.

Dockerfile
DanielMSchmidt: 🤷‍♂️ Will give it a try without

DanielMSchmidt: Did build without it this time

DanielMSchmidt: But system test driver did not execute:

```
dcos-system-test-driver -j1 -v /dcos-ui-plugins/system-tests/driver-config/cluster-ee.sh -t nodes
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/pkg_resources/__init__.py", line 666, in _build_master
    ws.require(__requires__)
  File "/usr/local/lib/python3.6/site-packages/pkg_resources/__init__.py", line 984, in require
    needed = self.resolve(parse_requirements(requirements))
  File "/usr/local/lib/python3.6/site-packages/pkg_resources/__init__.py", line 875, in resolve
    raise VersionConflict(dist, req).with_context(dependent_req)
pkg_resources.ContextualVersionConflict: (PyYAML 3.13 (/usr/local/lib/python3.6/site-packages), Requirement.parse('PyYAML==3.12'), {'dcos-system-test-driver'})

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/bin/dcos-system-test-driver", line 6, in <module>
    from pkg_resources import load_entry_point
  File "/usr/local/lib/python3.6/site-packages/pkg_resources/__init__.py", line 3144, in <module>
    @_call_aside
  File "/usr/local/lib/python3.6/site-packages/pkg_resources/__init__.py", line 3128, in _call_aside
    f(*args, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/pkg_resources/__init__.py", line 3157, in _initialize_master_working_set
    working_set = WorkingSet._build_master()
  File "/usr/local/lib/python3.6/site-packages/pkg_resources/__init__.py", line 668, in _build_master
    return cls._build_from_requirements(__requires__)
  File "/usr/local/lib/python3.6/site-packages/pkg_resources/__init__.py", line 681, in _build_from_requirements
    dists = ws.resolve(reqs, Environment())
  File "/usr/local/lib/python3.6/site-packages/pkg_resources/__init__.py", line 870, in resolve
    raise DistributionNotFound(req, requirers)
pkg_resources.DistributionNotFound: The 'PyYAML==3.12' distribution was not found and is required by dcos-system-test-driver
```

DanielMSchmidt:
For me it's unclear what would still need to be done. Please move it back into progress and let me know what I can do better 🙇

nLight:
@DanielMSchmidt the test is still failing

nLight:
screen shot 2018-08-22 at 13 43 37

PR #3165, opened 2018-08-14
chore(ui-kit): upgrade ui-kit to 1.8.0

Poltergeist:
Close DCOS-39675, DCOS-39674, DCOS-39677


nLight:
Usually bumping versions is a `chore`

PR #3167, opened 2018-08-14
DCOS-39897: Add Region Column to Services Table

GeorgiSTodorov:
⚠️ this PR now also includes DCOS-39898 - it completely finishes the feature 👍 

---

## Testing
Go to `Services` Tab and make sure that the new `Region` column is displayed and it works correctly.

Run `npm run test /plugins/services/src/js/containers/services/__tests__` to vertify that the new tests that I added run successfully.

## Trade-offs
I've updated the tests as well as the snapshots.

The current behavior is that if the regions array is empty, the cell will display "N/A", if there is more than one region they will be shown separated with commas, for example "Region 1, Region 2" and if a region is the same as the local region, it will be shown as "Region (Local)"

For the width of the column for different screen sizes, check out `src/styles/components/service-table/styles.less`, the values I used are slightly larger than the values for the width of the status column in accordance with the jira mock-up:
image

I also named the column "Region" and not "Regions" in accordance with the mock-up, although I think that "Regions" can be considered as well, since it is possible to have more than one region.

## Dependencies
None.

### Before
screenshot from 2018-08-14 15-54-25

### After
screenshot from 2018-08-14 15-47-13

### DIFF
--- a/plugins/services/src/js/containers/services/ServicesTable.js
+++ b/plugins/services/src/js/containers/services/ServicesTable.js
@@ -15,6 +15,7 @@ import ResourceTableUtil from "#SRC/js/utils/ResourceTableUtil";
 import TableUtil from "#SRC/js/utils/TableUtil";
 import Units from "#SRC/js/utils/Units";
 import { isSDKService } from "#SRC/js/utils/ServiceUtil";
+import CompositeState from "#SRC/js/structs/CompositeState";
 import ServiceStatusProgressBar from "../../components/ServiceStatusProgressBar";
 import Pod from "../../structs/Pod";
 import Service from "../../structs/Service";
@@ -44,6 +45,7 @@ const columnClasses = {
   name: "service-table-column-name",
   status: "service-table-column-status",
   version: "service-table-column-version",
+  regions: "service-table-column-regions",
   instances: "service-table-column-instances",
   cpus: "service-table-column-cpus",
   mem: "service-table-column-mem",
@@ -58,6 +60,7 @@ const METHODS_TO_BIND = [
   "handleActionDisabledModalOpen",
   "handleActionDisabledModalClose",
   "renderHeadline",
+  "renderRegions",
   "renderStats",
   "renderStatus",
   "renderServiceActions"
@@ -238,6 +241,21 @@ class ServicesTable extends React.Component {
         </span>
       </div>
     );
+  }
+
+  renderRegions(prop, service) {
+    const localRegion = CompositeState.getMasterNode().getRegionName();
+    let regions = service.getRegions();
+
+    regions = regions.map(
+      region => (region === localRegion ? region + " (Local)" : region)
+    );
+
+    if (regions.length === 0) {
+      regions.push("N/A");
+    }
+
+    return <span title={regions.join(", ")}>{regions.join(", ")}</span>;
   }
 
   renderServiceActions(prop, service) {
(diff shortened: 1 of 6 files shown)


plugins/services/src/js/containers/services/ServicesTable.js
```diff
@@ -43,6 +44,7 @@ const StatusMapping = {
 const columnClasses = {
   name: "service-table-column-name",
   status: "service-table-column-status",
+  region: "service-table-column-region",
```
juliangieseke: I'd go with `regions` 🤔

plugins/services/src/js/containers/services/ServicesTable.js
```diff
@@ -239,6 +242,21 @@ class ServicesTable extends React.Component {
...
 
+  renderRegions(prop, service) {
+    const localRegion = CompositeState.getMasterNode().getRegionName();
+    let regions = service.getRegions();
+
+    regions = regions.map(
+      region => (region === localRegion ? region + " (Local)" : region)
+    );
+
+    if (!regions[0]) {
```
juliangieseke: how about 
```
if(regions.length === 0) { 
  regions.push("N/A")
}
```

src/styles/components/service-table/styles.less
```diff
@@ -37,6 +37,11 @@
         width: 115px;
       }
 
+      &-region {
+        max-width: 100%;
+        width: 140px;
```
mperrotti: Would be nice if we could automatically handle column layout by relying on padding and document flow. Then we could avoid the need for hard-coding the width of each column in pixels every time we change the table.

This feedback isn't specific for this PR, but something to think about as we build out our [Table component](https://github.com/dcos-labs/ui-kit/tree/master/packages/table) and use it on more tables in DCOS.

GeorgiSTodorov:
@juliangieseke and I decided to use an alternative to `<Tooltip>`. Right now it looks like this:

screenshot from 2018-08-16 17-08-54

Poltergeist:
The Integration tests fail, and it seems that this is legit. Could it be that the table breaks if no region is available?

GeorgiSTodorov:
@Poltergeist It should, if the regions array is empty, just `N/A` is displayed.

screenshot from 2018-08-17 10-43-38

nLight:
@GeorgiSTodorov Legit integration tests fails? Fail second time in a row.

GeorgiSTodorov:
@nLight Julian and I are looking into this.

juliangieseke:
Tests should be fixed 👍

tnguyen0:
I would like to be able to test this with services deployed on multiple regions. Anyone know how to set this up?

GeorgiSTodorov:
@tnguyen0 I wondered the same and tried to find a way to do it. After talking with @juliangieseke I found out it's not possible for now. He suggested to create an integration test to test this - https://jira.mesosphere.com/browse/DCOS-40645

tnguyen0:
Hmm, ok. I asked because I was wondering how `CompositeState.getMasterNode().getRegionName()` would encompass the different regions. I imagined it would be something like finding the node which the service is on and returning the region for that node.

juliangieseke:
@tnguyen0 Not exactly if this was you question, but `CompositeState.getMasterNode` returns a Node - which always has only one region (`masterRegion` In Georgis code). 
A Service can have multiple Tasks which can (theoretically) run on different nodes in differen regions - `service.getRegions()` therefore returns an array with regions from the tasks.

juliangieseke:
@mperrotti Georgi and I had some issues getting the tooltip/table styles correct… could you have a look please? 🙇

plugins/services/src/js/containers/services/ServicesTable.js
```diff
@@ -240,6 +243,21 @@ class ServicesTable extends React.Component {
...
+
+    regions = regions.map(
+      region => (region === localRegion ? region + " (Local)" : region)
+    );
+
+    if (regions.length === 0) {
+      regions.push("N/A");
+    }
+
+    return <span title={regions.join(", ")}>{regions.join(", ")}</span>;
```
juliangieseke: @mperrotti if possible, this should be `<Tooltip>` instead of `<span>`, but we failed getting ellipsis and line breaks working correctly :(

GeorgiSTodorov: screenshot from 2018-08-16 16-22-58

PR #3168, opened 2018-08-14
Add branch name to the slack message

nLight:
Problem: I can'e operate PR numbers, though our branch names are rather informative. This PR is an attempt to make `#frontend-ci-status` more helpful for me and maybe for everyone else.

## Testing

I'll re-run the PR couple of times to kick in flaky tests and see if slack message improves.

## Trade-offs

I decided not investing any time in a research whether we have PR title as an env var. I believe we should though. So I wen with a branch name for starters.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -138,7 +138,7 @@ pipeline {
         slackSend (
           channel: "#frontend-ci-status",
           color: "danger",
-          message: "FAILED: Job '${env.JOB_NAME} [${env.BUILD_NUMBER}]' (${env.RUN_DISPLAY_URL})",
+          message: "FAILED\nBranch: ${env.CHANGE_BRANCH}\nJob: <${env.RUN_DISPLAY_URL}|${env.JOB_NAME} [${env.BUILD_NUMBER}]>",
           teamDomain: "mesosphere",
           token: "${env.SLACK_TOKEN}",
         )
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -151,7 +151,7 @@ pipeline {
         slackSend (
           channel: "#frontend-ci-status",
           color: "warning",
-          message: "UNSTABLE: Job '${env.JOB_NAME} [${env.BUILD_NUMBER}]' (${env.RUN_DISPLAY_URL})",
+          message: "UNSTABLE: ${env.BRANCH_NAME} Job '${env.JOB_NAME} [${env.BUILD_NUMBER}]' (${env.RUN_DISPLAY_URL})",
```
juliangieseke: when triggered from a PR, `BRANCH_NAME` is `PR-1234` :(
But `CHANGE_BRANCH` could work (it is empty when triggered from a branch)

juliangieseke:
There is no easy way to get the pr title in jenkins - but using the branch name is a very good idea!

Jenkinsfile
nLight: sure, let's try CHANGE_BRANCH

nLight:
Wonder if adding that plugin real quick will make the difference https://github.com/jenkinsci/pipeline-github-plugin

juliangieseke:
from all I heard and read about it, the one thing you want to avoid in jenkins are plugins… you will have a tough time debugging them :(

juliangieseke:
why wip again? 🤔

nLight:
@juliangieseke wanna make sure it works. So far I've only seen previous changes resulted in not-so-useful change: `FAILED: PR-3168`

nLight:
The build is going to fail, let's see

nLight:
@juliangieseke I updated some styling as well. Do you think we can just merge it or shall we give it a few runs to prove that is works?

juliangieseke:
I'd say we should see it working once and then merge & fix forward (if necessary)

juliangieseke:
also, it is in its own channel anyway, correct? maybe we also could add `success` messages? 🤔 
https://jenkins.io/doc/book/pipeline/syntax/#post

juliangieseke:
```
UNSTABLE
Branch: dima/jenkins-pipeline-echo-branchname
Job: frontend/dcos-ui-oss-pipeline/PR-3168 [6]
```

Master will look like this:
```
UNSTABLE
Branch: 
Job: frontend/dcos-ui-oss-pipeline/master [366]
```

PR #3169, opened 2018-08-15
DCOS-20989: use CosmosPackagesStores icons for Services Page

tnguyen0:
* Uses the existing CosmosPackagesStores to retrieve the specific package from the availablePackages.
* Fetch its icons and display on the Services Page.

Closes DCOS-20989

## Testing
* Deploy a package from the catalog
* Check that the icon appears for the framework in the services table
* Go to the framework details page and check that its icon is in the breadcrumbs

### Fetching Icons
* Decided to fetch all the icons for all packages at once instead of having to make a request per service to get the icons

### Making Requests
* Opted to use the existing mechanisms for fetching the available packages
* Opted for the simpler approach to update the packages when the services page is rendered instead of using the data layer

### DIFF
--- a/plugins/services/src/js/pages/ServicesPage.js
+++ b/plugins/services/src/js/pages/ServicesPage.js
@@ -6,6 +6,8 @@ import Icon from "#SRC/js/components/Icon";
 import RouterUtil from "#SRC/js/utils/RouterUtil";
 import TabsMixin from "#SRC/js/mixins/TabsMixin";
 
+import CosmosPackagesStore from "#SRC/js/stores/CosmosPackagesStore";
+
 var ServicesPage = React.createClass({
   contextTypes: {
     router: routerShape
@@ -37,6 +39,10 @@ var ServicesPage = React.createClass({
       "/services/overview": "Services"
     };
     this.updateCurrentTab();
+  },
+
+  componentDidMount() {
+    CosmosPackagesStore.fetchAvailablePackages();
   },
 
   componentDidUpdate() {
--- a/plugins/services/src/js/structs/Framework.js
+++ b/plugins/services/src/js/structs/Framework.js
@@ -5,6 +5,8 @@ import {
   ROUTE_ACCESS_PREFIX,
   FRAMEWORK_ID_VALID_CHARACTERS
 } from "../constants/FrameworkConstants";
+import FrameworkUtil from "../utils/FrameworkUtil";
+
 import Application from "./Application";
 import FrameworkSpec from "./FrameworkSpec";
 
@@ -16,6 +18,13 @@ module.exports = class Framework extends Application {
     // it gets as a properties of this object and we want to avoid any naming
     // collisions.
     this._spec = null;
+  }
+
+  /**
+   * @override
+   */
+  getImages() {
+    return FrameworkUtil.getServiceImages(this.get("images"));
   }
 
   getPackageName() {
--- a/src/js/stores/__tests__/CosmosPackagesStore-test.js
+++ b/src/js/stores/__tests__/CosmosPackagesStore-test.js
@@ -69,13 +69,25 @@ describe("CosmosPackagesStore", function() {
       it("stores availablePackages when event is dispatched", function() {
         AppDispatcher.handleServerAction({
           type: ActionTypes.REQUEST_COSMOS_PACKAGES_SEARCH_SUCCESS,
-          data: [{ gid: "foo", bar: "baz" }],
+          data: { packages: [{ gid: "foo", bar: "baz" }] },
           query: "foo"
         });
 
         var availablePackages = CosmosPackagesStore.getAvailablePackages().getItems();
         expect(availablePackages[0].get("gid")).toEqual("foo");
         expect(availablePackages[0].get("bar")).toEqual("baz");
+      });
+
+      it("stores packageImages when event is dispatched", function() {
+        AppDispatcher.handleServerAction({
+          type: ActionTypes.REQUEST_COSMOS_PACKAGES_SEARCH_SUCCESS,
+          data: { images: { gid: "foo", bar: "baz" } },
+          query: "foo"
+        });
+
+        var packageImages = CosmosPackagesStore.getPackageImages();
+        expect(packageImages["gid"]).toEqual("foo");
+        expect(packageImages["bar"]).toEqual("baz");
       });
 
       it("dispatches the correct event upon success", function() {
@@ -86,7 +98,7 @@ describe("CosmosPackagesStore", function() {
         );
         AppDispatcher.handleServerAction({
           type: ActionTypes.REQUEST_COSMOS_PACKAGES_SEARCH_SUCCESS,
-          data: [{ gid: "foo", bar: "baz" }],
+          data: { packages: [{ gid: "foo", bar: "baz" }] },
           query: "foo"
         });
 
--- a/plugins/services/src/js/stores/MarathonStore.js
+++ b/plugins/services/src/js/stores/MarathonStore.js
@@ -4,6 +4,7 @@ import { SERVER_ACTION } from "#SRC/js/constants/ActionTypes";
 import AppDispatcher from "#SRC/js/events/AppDispatcher";
 import CompositeState from "#SRC/js/structs/CompositeState";
 import Config from "#SRC/js/config/Config";
+import CosmosPackagesStore from "#SRC/js/stores/CosmosPackagesStore";
 import GetSetBaseStore from "#SRC/js/stores/GetSetBaseStore";
 import VisibilityStore from "#SRC/js/stores/VisibilityStore";
 
@@ -85,6 +86,9 @@ import {
 import Framework from "../structs/Framework";
 import ServiceImages from "../constants/ServiceImages";
 import ServiceTree from "../structs/ServiceTree";
+
+import FrameworkUtil from "../utils/FrameworkUtil";
+import ServiceValidatorUtil from "../utils/ServiceValidatorUtil";
 
 let requestInterval = null;
 let shouldEmbedLastUnusedOffers = false;
@@ -226,6 +230,7 @@ class MarathonStore extends GetSetBaseStore {
           this.emit(MARATHON_SERVICE_RESTART_SUCCESS);
           break;
         case REQUEST_MARATHON_GROUPS_SUCCESS:
+          this.injectGroupsWithPackageImages(action.data);
           this.processMarathonGroups(action.data);
           break;
         case REQUEST_MARATHON_DEPLOYMENTS_SUCCESS:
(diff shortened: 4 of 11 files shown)


plugins/services/src/js/structs/Framework.js
```diff
@@ -18,6 +23,23 @@ module.exports = class Framework extends Application {
     this._spec = null;
   }
 
+  /**
+   * @override
+   */
+  getImages() {
```
juliangieseke: I fear that this is too expensive to be called every 2 seconds, we should memoize it 🤔

plugins/services/src/js/structs/Framework.js
```diff
@@ -18,6 +23,23 @@ module.exports = class Framework extends Application {
...
+    const parsedImages = {};
+    const pkg = CosmosPackagesStore.getAvailablePackage(this.getPackageName());
```
juliangieseke: is there any other way than this? I know we did this before, but structs directly accessing stores is one of the issues we face when working with our codebase right now :( 

Maybe we can try to move this to the place where the struct is generated and inject the images there? 🤔 Without digging too deep I think `ServiceTree.js#constructor` is the "best" place for this - but first we need to reintegrate `ServiceUtil#createServiceFromResponse` back into the `ServiceTree` class… 😢

I'd vote for this refactoring 👍 
@nLight @orlandohohmeier opinions?

plugins/services/src/js/pages/ServicesPage.js
```diff
@@ -43,6 +45,10 @@ var ServicesPage = React.createClass({
     this.updateCurrentTab();
   },
 
+  componentDidMount() {
+    CosmosPackagesStore.fetchAvailablePackages();
```
juliangieseke: i think instead of fetching the packages once it would be better to subscribe to changes 🤔 (no idea if `CosmosPackageStore` is built that way tho… 🤷‍♂️ )

plugins/services/src/js/structs/Framework.js
nLight: Why not asking the whole team for opinions? :)

juliangieseke: @dcos/frontend-dev ^

plugins/services/src/js/structs/Framework.js
DanielMSchmidt: +1 for memorization

plugins/services/src/js/pages/ServicesPage.js
Poltergeist: Packages will not change that often. This solution is not ideal but I guess it is okay.

plugins/services/src/js/structs/Framework.js
Poltergeist: I second Daniel here.

juliangieseke: I thought that was related to the same _commit_, not _PR_. 🤔

juliangieseke: Also, this isnt a bug, its new code 🤔

plugins/services/src/js/utils/ServiceUtil.js
```diff
@@ -44,13 +47,30 @@ const getFindPropertiesRecursive = function(service, item) {
   }, {});
 };
 
+const getFrameworkImages = function(packageName) {
```
tnguyen0: I tried to put this in `FrameworkUtil`, but for whatever reason unit tests for `UniversePackagesList` fail when I import `CosmosPackagesStore` into `FrameworkUtil`.

plugins/services/src/js/pages/ServicesPage.js
tnguyen0: It was a middle ground that was decided on versus fetching it once or subscribing to changes. Open to suggestions.

tnguyen0:
Thank you @Poltergeist @nLight @juliangieseke @DanielMSchmidt for the feedback.

Made the changes that were suggested. Would love some more feedback :).

I pivoted from using the data layer to solve this problem, AFAIK there is not really a benefit here. I'm open to revisiting it again if someone feels strongly about it.

plugins/services/src/js/structs/Framework.js
DanielMSchmidt: Okay, might be a misunderstanding on my side. I understood it as in not in the same PR/ticket.
It's new code fixing a bug. When we now go to change something in the structs I fear that we introduce new bugs with this refactoring.

plugins/services/src/js/pages/ServicesPage.js
weblancaster: Because packages don't change over time I rather see us fetching only when we need it (which is not often) and caching it.

tnguyen0:
@Fabs would love your input.

plugins/services/src/js/utils/ServiceUtil.js
Fabs: > TL;DR there is a better solution (IMHO) using the data layer that does not violate our architecture.

I am with Julian, `structs` should not have dependencies on how the data is fetched, and `utils` should be pure functions from the external perspective. That is an invariant that I believe helps us have a maintainable system, and I would like to keep it that way. 

I tried to provide a solution to having the cosmos dependency, by thinking what if I pass the result of `CosmosPackagesStore.getPackageImages()` as imagesDatabase to`getFrameworkImage(packageName, imageDatabase)`. But the `getFramework` method is called by `ServiceUtil.createServiceFromResponse`, which is then called by the ServiceTree `constructor`. There I stopped because I have no idea on how to tackle that simply. It seems to me that the solution is to make the DCOSStore aware of changes on package icons, it also looks to me that other places might build the tree. Also, the logic to build the tree is optimized via cache so we have to consider that the icons do not invalidate the cache unnecessarily..... That is altogether not simple.

This dependency of Cosmos on the util is for me a show stopper. I would gladly be overriden by the team if they think that is ok. I would also love if @nLight or @orlandohohmeier just jumped in with a simpler solution.

On the case, nothing simpler comes or the team thinks this dependency is fine (which I don't). I would gladly help you either way you chose to go: finish the direction you started or shift gears and still achieve that using the data layer.

*The data layer solution for me* is to create a mediator component who fetches the image just once using a ReplaySubject to cache the image data. We can use the CosmosStore to save time, but ideally, we should make the request ourselves using streams. The mediator component would be provided right before passing the service to the table, if it can find a cached image for the service, it passes it down, if It cannot it returns an empty image or placeholder. That part of the Service struct could require some thinking, but we at least avoid dealing with the tree.

src/js/stores/__tests__/CosmosPackagesStore-test.js
```diff
@@ -69,7 +69,7 @@ describe("CosmosPackagesStore", function() {
       it("stores availablePackages when event is dispatched", function() {
```
Fabs: I also think we should provide an integration test.

plugins/services/src/js/utils/ServiceUtil.js
```diff
@@ -44,13 +47,30 @@ const getFindPropertiesRecursive = function(service, item) {
...
+
 const ServiceUtil = {
   createServiceFromResponse(data) {
     if (ServiceValidatorUtil.isPodResponse(data)) {
       return new Pod(data);
     }
 
     if (ServiceValidatorUtil.isFrameworkResponse(data)) {
+      const packageName = data.labels.DCOS_PACKAGE_NAME;
+      data.images = getFrameworkImages(packageName);
```
Fabs: What happens in the system when the image object is empty?

plugins/services/src/js/utils/ServiceUtil.js
Fabs: On an aside. To my own dismay, I looked into all the Utils and there is a single use of a Store in a util which I wrote when fixing the performance of the ServicePage. Is totally not needed as the callers of the `mergeTaskData`could pass DCOSStore.taskLookupTable.

plugins/services/src/js/structs/Framework.js
weblancaster: You are right @Fabs the reason we decided to go this path is that is more performant, meaning not calling every 2 secs.

plugins/services/src/js/utils/ServiceUtil.js
tnguyen0: Thank you for your detailed feedback @Fabs!

I feel we should definitely go for the the better solution, instead of the ship-it that violates our architecture. Given that this ticket is a blocker for 1.12, I am not sure how much time it would take for me to implement the data layer solution. I wonder if we should just ship this solution and create a tech debt ticket to be addressed later.

@dcos/frontend-dev what does everyone think the course of action should be here?

src/js/stores/__tests__/CosmosPackagesStore-test.js
tnguyen0: 👍

plugins/services/src/js/utils/ServiceUtil.js
tnguyen0: The images are passed to the `FrameworkUtil.getServiceImages()` method, so will return default image if object is empty.

plugins/services/src/js/utils/ServiceUtil.js
Fabs: What does happen if we do not ship it on 1.12? Will our users start seeing broken links everywhere?

Fabs:
@tnguyen0. Is it possible that you provide a test case where on master I would not be able to see the icons, but on your PR the icons appear correctly?

plugins/services/src/js/utils/ServiceUtil.js
tnguyen0: Yes, the label is removed for 1.12. I was informed we have roughly 6 weeks until then. Given we have time, I think we should go for the best solution.

tnguyen0:
@Fabs, you could remove the use of the `DCOS_PACKAGE_METADATA` label.

tnguyen0:
After speaking with @juliangieseke, we can instead accessing the `CosmosPackagesStore` in `MarathomStore#processMarathonGroups`.

plugins/services/src/js/utils/ServiceUtil.js
tnguyen0: After speaking with @juliangieseke, we can instead accessing the `CosmosPackagesStore` in `MarathomStore#processMarathonGroups`.

juliangieseke: It would be the only place not in a `util` or `struct` - not sure what issues we'll face changing the store and which places I havent seen tho… Ideas anyone?

tnguyen0:
@weblancaster @Fabs @Poltergeist @juliangieseke can you please take a look? For the integration test, look at the specific commit for it.

PR #3173, opened 2018-08-15
DCOS-40523: Enable empty env vars

orlandohohmeier:
---
↩️  _This PR back-ports a fix to release/1.0 introduced with https://github.com/dcos/dcos-ui/pull/3106._

---

This fixes the environment variables reducer to support empty values.

## Testing

Try to create a service with the following config:

```json
{
  "env": {
    "key": ""
  }
}
```

### DIFF
--- a/plugins/services/src/js/components/modals/CreateServiceModalForm.js
+++ b/plugins/services/src/js/components/modals/CreateServiceModalForm.js
@@ -78,6 +78,34 @@ const CONSTANTLY_UNMUTED_ERRORS = [
   /^volumes\.[0-9]+\./
 ];
 
+function cleanConfig(config) {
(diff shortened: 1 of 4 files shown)


juliangieseke:
I think this is still buggy 😢

after opening the given config for editing, our form ignores the "wrong" variable `key`
screen recording 2018-08-15 at 09 43 pm

and changing something in our form overrides old states (also happens with `key` from above):
screen recording 2018-08-15 at 09 46 pm

jaxesn:
I saw the backport PR, but was @juliangieseke's concern above addressed or at least deemed a non-issue?

juliangieseke:
We tried to reproduce on daniels machine but werent able to do so - looks like my local installation was outdated

PR #3175, opened 2018-08-15
DCOS-40526: add region information getter to pod struct

juliangieseke:
Closes DCOS-40526

## Testing

ui isnt implemented yet - `npm run test` it is.

## Trade-offs

I decided to just duplicate the code instead of extracting something into a util. I dont think that extracting it would add much value.

Also, this is a fixup commit for the last one, we're on a feature branch here which I'll have to rebase anyway - it _should_ work. I'm aware that this messes up our JIRAs a bit because there is no commit closing DCOS-40526 - lets see 🤷‍♂️

### DIFF
--- a/plugins/services/src/js/structs/Pod.js
+++ b/plugins/services/src/js/structs/Pod.js
@@ -208,4 +208,26 @@ module.exports = class Pod extends Service {
   getVolumesData() {
     return new VolumeList({ items: this.get("volumeData") || [] });
   }
+
+  /**
+   * @override
+   */
+  getRegions() {
+    if (!this._regions) {
+      const regionCounts = this.getInstanceList()
+        .getItems()
+        .reduce((regions, instance) => {
+          const region = instance.getAgentRegion();
+          if (region !== "") {
+            regions[region] = regions[region] ? regions[region] + 1 : 1;
+          }
+
+          return regions;
+        }, {});
+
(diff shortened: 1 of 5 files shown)


plugins/services/src/js/structs/Pod.js
```diff
@@ -208,4 +208,28 @@ module.exports = class Pod extends Service {
...
+          const region = instance.getAgentRegion();
+          if (region !== "") {
+            regions[region] = regions[region] ? regions[region] + 1 : 1;
+          }
+
+          return regions;
+        }, {});
+
+      this._regions = Object.keys(regionCounts).sort(
+        (a, b) => regionCounts[b] - regionCounts[a]
```
nLight: wonder why not natural sort? Couldn't find any explanation in the jira

juliangieseke: yeah, was my decision - I thought its a good idea to sort by tasks running on that instance 🤷‍♂️

nLight: Couldn't it be confusing for a user as we don't show the actual count?

juliangieseke: The actual column is added here: https://github.com/dcos/dcos-ui/pull/3167
I'll change the implementation to use natural sort - have to rebase the feature branch anyway 🤷‍♂️

PR #3177, opened 2018-08-15
WIP: Webpack 4

nLight:
I wanted to get the feeling where our build is slow. Webpack 4 has a profiler plugin. Therefore I timeboxed some slack time and updated webpack. It compiles the app and gives me the profiling bundle. My Chrome crashes when I'm trying to load it though.

Almost working, not faster then current version we're using.

Plan: I monitor webpack repo to see when they fix some performance issues. I'm also going to be using this setup to try getting the profiler working.


nLight:
@weblancaster but then we should switch to Puppeteer, right?

weblancaster:
@nLight not necessarily, we can use both but that's why I mentioned a separated job just for performance which doesn't block the official CI

brandonc:
I think we should close this in favor of #3310. I was able to profile with a different package: https://github.com/dcos/dcos-ui/pull/3307/commits/8771fe2c05e3cdca6ee63efef6f3e601ea8f9080

PR #3178, opened 2018-08-15
DCOS-39945: Re-add colors to NodesTable progress bars

mperrotti:
## Testing
- Go to the Nodes page
- See the new colors in the progress bars

If any bars are at 0%, manually add the bar markup:
```
<span class="bar color-{color id}" style="width: 25%;"></span>
```

I checked for other uses of `ProgressBar`, and they will not be affected by the style changes in `src/styles/components/status-bar/styles.less`

## Trade-offs
None that I can think of

## Dependencies
Nothing because the ticket for DCOS-39944 has already merged


mperrotti:
Thanks for adding the screenshot @GeorgiSTodorov!

PR #3181, opened 2018-08-16
DCOS-21183: no duplicate names when adding containers to service

mperrotti:
## Testing
After navigating to the Services tab:
1. Tap the "plus" in the upper-right corner to add a new service
2. Select "Multi-container (Pod)"
3. Add two new containers, so you have "container-1", "container-2", and "container-3"
4. Delete "container-2"
5. Add another new container; it should be named "container-4". Before these changes, it would have a duplicate name, "container-3"

## Trade-offs
No

## Dependencies
Nothing

## Screenshots
**Before**
containernamedupe_before

**After**
containernamedupe_after

### DIFF
--- a/src/js/utils/ContainerUtil.js
+++ b/src/js/utils/ContainerUtil.js
@@ -18,6 +18,19 @@ const ContainerUtil = {
    */
   adjustPendingActions(pendingActions, actionType, isPending) {
     return Object.assign({}, pendingActions, { [actionType]: isPending });
+  },
+
+  getNewContainerName(containerLength, newState) {
+    const name = `container-${containerLength + 1}`;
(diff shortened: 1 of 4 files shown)


src/js/utils/ContainerUtil.js
```diff
@@ -18,6 +18,19 @@ const ContainerUtil = {
    */
   adjustPendingActions(pendingActions, actionType, isPending) {
     return Object.assign({}, pendingActions, { [actionType]: isPending });
+  },
+
+  getNewContainerName(containerLength, newState) {
```
mperrotti: `ContainerUtil.js` seemed like an appropriate spot for this, but I'm not sure. Can anybody recommend somewhere better?

PR #3182, opened 2018-08-16
DCOS-40506: sort Nodes table by health

mperrotti:
## Testing
Go to the Nodes tab and check that it's sorted by "Health"

## Trade-offs
None

## Dependencies
None


Poltergeist:
@jaxesn Yes the order of states can be seen here: https://github.com/dcos/dcos-ui/blob/master/src/js/constants/UnitHealthStatus.js

PR #3188, opened 2018-08-17
    DCOS-40495  Nodes Table Region Column shows "N/A (Local)"

GeorgiSTodorov:
Remove the (Local) addition when the local region is "N/A"

jira: https://jira.mesosphere.com/browse/DCOS-40495

## Testing
In `Nodes` tab, in `Region` column, if there is no local region, the name should be shown just as `N/A` instead of `N/A (Local)`.

Alternatively, to test it, try replacing the function `regionRenderer` in `/plugins/nodes/src/js/columns/NodesTableRegionColumn.tsx` the following way:
```
export function regionRenderer(
  masterRegion: string,
  data: Node
): React.ReactNode {
  data;
  masterRegion = "N/A";
  const regionString = "N/A";
  const regionName =
  regionString +
    (masterRegion === regionString && masterRegion !== "N/A"
      ? " (Local)"
      : "");

  return (
    <TextCell>
      <span title={regionName}>{regionName}</span>
    </TextCell>
  );
}
```

## Trade-offs
None that I am aware of.

## Dependencies
None.

### Before
screenshot from 2018-08-17 15-17-05

### After
screenshot from 2018-08-17 15-47-48

### DIFF
--- a/plugins/nodes/src/js/columns/NodesTableRegionColumn.tsx
+++ b/plugins/nodes/src/js/columns/NodesTableRegionColumn.tsx
@@ -11,7 +11,9 @@ export function regionRenderer(
 ): React.ReactNode {
   const regionName =
     data.getRegionName() +
-    (masterRegion === data.getRegionName() ? " (Local)" : "");
+    (masterRegion === data.getRegionName() && masterRegion !== "N/A"
+      ? " (Local)"
+      : "");
 
   return (
     <TextCell>


plugins/nodes/src/js/columns/NodesTableRegionColumn.tsx
```diff
@@ -11,7 +11,9 @@ export function regionRenderer(
 ): React.ReactNode {
   const regionName =
     data.getRegionName() +
-    (masterRegion === data.getRegionName() ? " (Local)" : "");
+    (masterRegion === data.getRegionName() && masterRegion !== "N/A"
```
weblancaster: This fixes it but why not fix the root cause in `structs/Node.js` line `23`? in theory the components should be receiving the type of data it expects and just display it.

I believe we don't want to have `"N/A"` ever displayed. Is that correct @leemunroe

leemunroe: For OSS we decided to go with `N/A` from what I recall. I don't think you would ever see is on a EE cluster. So the scenarios are:

* If EE and local then `aws/us-west-2 (Local)`
* If EE and not local then `aws/us-west-3`
* If OSS then `N/A`

GeorgiSTodorov: We discussed the same thing here for the services table - https://jira.mesosphere.com/browse/DCOS-39898?focusedCommentId=169987&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-169987

I think the component is receiving the type of data it expects and is working as intended. It's just a matter of not displaying `Local` when there is no region.

Fabs: @leemunroe and @weblancaster what should an EE cluster display when it is on Premises, or on some infrastructure other than AWS that does not support regions/zones?

I run an EE cluster on my desk that does not have region/zone, and I ran into bugs in the past because we assumed EE always have regions.

Said that, I think @GeorgiSTodorov has a proper fix to this error.

tnguyen0: @GeorgiSTodorov, bit of a nit here:

What would having a `Node#hasRegion` method and then it would be:
`data.hasRegion() && masterRegion === data.getRegionName()`

juliangieseke:
ping @weblancaster

plugins/nodes/src/js/columns/NodesTableRegionColumn.tsx
GeorgiSTodorov: It seems redundant to me to have two methods in Node where of them has the same information as the other plus more.

We can add a local variable:
```
hasRegion = masterRegion !== "N/A"
```
but I think it's readable enough in the current version.

PR #3190, opened 2018-08-17
DCOS-20273: Detect that field value is disabled when unique selected

natmegs:
Fix integration test for service form modal > create layout > placement > add placement to detect that field value is disabled when unique selected.

## Testing

Run integration test for ServiceFormModal-cy.js and ensure that  **Service: Placement > Add Placement Constraint > Should disable the field "Value" when Unique is selected in operator dropdown** test passes

## Trade-offs

Not sure if prettier/linting rules have changed since this file was last worked on, but the pre-commit linting checks were failing which is why the diff shows a huge number of lines changed. If there's another way to get around this please let me know. The only non-formatting changes here are lines 821-832. https://github.com/dcos/dcos-ui/pull/3190/files#diff-0796ef68862d3b5d45e239ff56ee5be7R821

### DIFF
--- a/tests/pages/services/ServiceFormModal-cy.js
+++ b/tests/pages/services/ServiceFormModal-cy.js
@@ -5,23 +5,25 @@ describe("Service Form Modal", function() {
 
   context("Create", function() {
     function openServiceModal() {
-      cy.get(".page-header-actions button").first().click();
+      cy.get(".page-header-actions button")
+        .first()
+        .click();
     }
 
     function clickRunService() {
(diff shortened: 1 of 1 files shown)


jaxesn:
one way that might have helped, made two commits. The first with just the lint change then the second with the actual code changes.

natmegs:
@jaxesn Good call. I will keep that in mind for the other tests that need work! And if anyone's having trouble finding the actual change, let me know and I'll redo the commits.

PR #3191, opened 2018-08-19
DCOS_OSS-3992: getUsedResources not applicable to Service

brandonc:
Hello, friends. I decided to get DC/OS running this weekend in anticipation of starting work on Sept. 10 and found an issue. When loading the dashboard and no nodes are present, the fallback call to `Service#getUsedResources` fails because the method name is `getResources` there (and in `Framework`) but the `Node` type was written using the legacy name https://github.com/dcos/dcos-ui/commit/ff0673d4c601d5a8230602be36f84a3345c38687

There appear to be no remaining calls to getUsedResources

Closes DCOS_OSS-3992 🙌

screen shot 2018-08-19 at 5 04 20 pm

## Testing

Visit dashboard with no connected nodes

## Trade-offs

The name getUsedResources better reflects the API property name but this comment https://github.com/dcos/dcos-ui/blob/master/plugins/services/src/js/structs/Framework.js#L103-L106 makes it appear that it is not preferred.

## Dependencies

n/a


Fabs:
Thanks a lot, @brandonc. I am looking forward to welcome you Sept 10 💃 

That is pretty weird that we let pass something like that. Would you be willing to add a test case that would fail before you applied the fix? If you cannot that is ok, we would take care of that :D.

brandonc:
I found an issue with a test module that I can add a fix to. Just need a little more time.

brandonc:
Something to consider: Is ServicesList a valid dependency of `CompositeState`? It seems strange that a core component would depend on a plugin component. I don't know enough yet to propose something better, but in general do plugin dependencies go in both directions?

nLight:
Hey @brandonc! Looking forward having you in the team! :) we have quite some questions like the one you've asked. Though, having to think pragmatically, we sometimes decide to spend our cycles to ship some value instead of doing a refactoring. Would be happy having such discussions with you! :)

brandonc:
✌️✌️Feels good looking forward to that

PR #3192, opened 2018-08-20
DCOS-20689: introduce Sentry bug tracking SaaS

nLight:
Track errors with [Sentry](https://sentry.io)

Closes DCOS-20689

## Testing

- Enable `tracking` plugin in your Config.dev.ts
- Add an exception to any place in code
- Reach this place on the UI
- Verify that the error is reflected and all tags (environment, variant, version) are set.

## Trade-offs

We track two versions: DCOS and DCOS-UI. DCOS-UI version will be set as a `release` in Sentry where dcos version as a tag. I did this because we are going to be shipping UI as a package thus having the same UI release running on different DCOS versions. Plus I'm going to connect Sentry and Github to provide better dev experience.

### DIFF
--- a/plugins/tracking/hooks.js
+++ b/plugins/tracking/hooks.js
@@ -4,6 +4,9 @@ import React from "react";
 import AuthStore from "#SRC/js/stores/AuthStore";
 import Config from "#SRC/js/config/Config";
 import DOMUtils from "#SRC/js/utils/DOMUtils";
+import EventTypes from "#SRC/js/constants/EventTypes";
+import MetadataStore from "#SRC/js/stores/MetadataStore";
+import Util from "#SRC/js/utils/Util";
 
 import { ANALYTICS_LOAD_TIMEOUT } from "./constants/PluginConstants";
 import Actions from "./actions/Actions";
@@ -22,7 +25,12 @@ module.exports = {
     "userAddPolicy"
   ],
 
-  actions: ["userLoginSuccess", "userLogoutSuccess", "routes"],
+  actions: [
+    "pluginsConfigured",
+    "userLoginSuccess",
+    "userLogoutSuccess",
+    "routes"
+  ],
 
   initialize(configuration) {
     this.configuration = configuration;
@@ -57,6 +65,37 @@ module.exports = {
     promiseArray.push(promise);
 
     return promiseArray;
+  },
+
+  pluginsConfigured() {
+    // Ensure analytics is actually ready, because in #pluginsLoadedCheck we
+    // may skip the check so that we don't completely block the applicaiton
+    const metadata =
+      Util.findNestedPropertyInObject(this, "configuration.metadata") || {};
+
+    global.analytics.ready(() => {
+      const setContext = () => {
+        if (!global.Raven) {
+          return;
+        }
+
+        global.Raven.setTagsContext({
+          ...metadata,
+          environment: process.env.NODE_ENV,
+          dcosVersion: MetadataStore.version,
+          dcosVariant: MetadataStore.variant
+        });
+      };
+
+      if (!MetadataStore.version) {
+        MetadataStore.addChangeListener(
+          EventTypes.DCOS_METADATA_CHANGE,
+          setContext
+        );
+      } else {
+        setContext();
+      }
+    });
   },
 
   userLoginSuccess() {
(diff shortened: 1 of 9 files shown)


plugins/tracking/hooks.js
```diff
@@ -59,6 +67,37 @@ module.exports = {
     return promiseArray;
   },
 
+  pluginsConfigured() {
+    // Ensure analytics is actually ready, because in #pluginsLoadedCheck we
+    // may skip the check so that we don't completely block the applicaiton
+    const metadata =
+      Util.findNestedPropertyInObject(this, "configuration.metadata") || {};
+
+    global.analytics.ready(() => {
```
jaxesn: I think i've asked this before, but where does `global.analytics` come from?

nLight: It comes from segment.io that we're using to inject all other services.

nLight:
WIP: turned out because of the current `semantic-release` setup we don't have release version in `package.json` our build/release process has to be adjusted.

juliangieseke:
I did a local test which was able to replace the placeholder and echo the upload command.

---

updated `dist/index.html`:
```
<script type="text/javascript"> window.DCOS_UI_VERSION = "master+v2.14.0"; </script>
```
**=> is the release name correct? yes / no**

---

`aws s3 cp` upload cmd:
```
aws s3 cp …/release.tar.gz s3://…/master+dcos-ui-v2.14.0.tar.gz
```
**=> is the filename correct? yes / no**

---

Once CI is green we should be able to merge - after we merged, we should keep an eye on the build to verify that everything actually works as expected 🤷‍♂️  (if not, its not _that_ bad - worst case is that our daily FE cluster is broken - we cant break anything _really_ important)

scripts/ci/upload-release
```diff
@@ -32,8 +32,8 @@ PKG_TARGET=$(python -c "import sys, json; \
   dcosrc = json.load(open('$PROJECT_ROOT/.dcosrc')); \
   sys.stdout.write(str(dcosrc['base']));")
 
-# release version (after switching to semantic-release, we cant use package.json anymore)
-PKG_VERSION="$(git tag -l --sort=-version:refname | grep ${PKG_TARGET} | head -1 | cut -d'+' -f2)"
+# next release version (passed in by semantic-release)
+PKG_VERSION="${1}"
```
juliangieseke: version is now injected by semanric-release ->  https://github.com/dcos/dcos-ui/pull/3192/files#diff-b78e9a225940bc2471e85d5d79220450R9

scripts/ci/upload-release
```diff
@@ -47,6 +47,14 @@ DCOS_BRANCH="dcos-ui/${PKG_TARGET}+dcos-ui-latest"
 ## Create Assets & Generate Upload Data
 #####################################################################
 
+
+# first, replace placeholder in dist bundle
+sed -i "" "s/0.0.0-dev+semantic-release/${PKG_TARGET}+${PKG_VERSION}/" ${PROJECT_ROOT}/dist/index.html
```
juliangieseke: here we inject the new version into `dist/index.html`

scripts/ci/upload-release
```diff
@@ -47,6 +47,19 @@ DCOS_BRANCH="dcos-ui/${PKG_TARGET}+dcos-ui-latest"
 ## Create Assets & Generate Upload Data
 #####################################################################
 
+VERSION_PLACEHOLDER="0.0.0-dev+semantic-release"
```
nLight: @DanielMSchmidt added

juliangieseke: we could read it from `package.json` as we do with `PKG_TARGET` and `.dcosrc` above :trollface:

plugins/tracking/hooks.js
```diff
@@ -59,6 +67,37 @@ module.exports = {
...
+      const setContext = () => {
+        if (!global.Raven) {
+          return;
+        }
+
+        global.Raven.setTagsContext({
+          ...metadata,
+          environment: process.env.NODE_ENV,
+          dcosVersion: MetadataStore.version,
```
weblancaster: @nLight on line `90` you are checking for `!MetadataStore.version` but not here, so I wonder if you are ok in having a falsey `dcosVersion`?

PR #3193, opened 2018-08-20
chore: run system tests in parallel

nLight:
Testing if we can enable parallel system tests


nLight:
We can't enable || tests until we fix system test driver

PR #3199, opened 2018-08-21
DCOS-37650: Change default request error message

nLight:
This PR changes the default misleading error message. 

Closes DCOS-37650

## Testing

Easiest way I've found is to load the UI, kill the Webpack server and navigate to Jobs.

## Trade-offs

`Cannot connect with the server` implies TCP level issue, but we use the message even in case of permission errors.

More generic `There was a problem with network request` is a bandaid until we start detecting error type and showing proper messages.

## Dependencies

There's going to be plugins PR that should go right after.

## Screenshots

screen shot 2018-08-21 at 15 57 01


nLight:
@lilysamimi would love your input on the message

lilysamimi:
@nLight, what can the user do to remediate the issue? Where can they go to learn more? Or is the only option to contact us through slack and/or email? 

Also, would be great if we could show a red banner for this message as it is an error but not sure if that's within the scope of this ticket.

nLight:
@lilysamimi the scope of the ticket is just wording. 

This is the message we will be showing in case of _any_ network error. Including network is down, resource not existing, permissions etc. There's pretty much nothing a user can do about that until she knows what's the issue. To address that and further improve the US we have a follow up issue https://jira.mesosphere.com/browse/DCOS-12713

src/js/components/RequestErrorMsg.js
```diff
@@ -46,7 +46,7 @@ class RequestErrorMsg extends React.Component {
 
 RequestErrorMsg.defaultProps = {
   columnClasses: {},
-  header: "Cannot Connect With The Server",
+  header: "There was a problem with network request",
```
juliangieseke: Not sure how important this is, but I'd avoid the word "Problem". 
What about something like "DC/OS UI has difficulties loading the requested information, please try again"? 🤔

jaxesn:
Regarding @lilysamimi's banner comment, if we do not have a component/css for a banner, lets get that added on the UI kit roadmap. Either way, lets add a ticket to maybe go around and standardize our error/success styling.  cc @mperrotti

src/js/components/RequestErrorMsg.js
lilysamimi: How about: "DC/OS UI cannot retrieve the requested information at this moment."

nLight: I like it!

nLight:
I updated the message.

PR #3201, opened 2018-08-21
 DCOS-40691: in Services table, region column to use Tooltip component

mperrotti:
In the first pass at adding the Regions table, we used a span element with a title attribute to show a native tooltip in order to side-step text wrapping issues we were seeing when using the Tooltip component

Closes: DCOS-40691

## Testing
- Hard-code an array of region names in the "renderRegions" method in `/plugins/services/src/js/containers/services/ServicesTable.js`
- Go to the Services view
- Hover the text in the cells of the "Region" column

The tooltip should be hovering in the center of the cell, and the tooltip content text should be wrapping, not  overflowing the tooltip bubble or being chopped off

## Trade-offs
Nothing I can think of

## Dependencies
None

## Screenshots
**Before** (in `feature/DCOS-39895-services-region-info`):
tooltipconsistency_before

**After:**
tooltipconsistency_after

### DIFF
--- a/plugins/services/src/js/containers/services/ServicesTable.js
+++ b/plugins/services/src/js/containers/services/ServicesTable.js
@@ -255,7 +255,16 @@ class ServicesTable extends React.Component {
       regions.push("N/A");
     }
 
-    return <span title={regions.join(", ")}>{regions.join(", ")}</span>;
+    return (
+      <Tooltip
+        elementTag="span"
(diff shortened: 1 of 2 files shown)


tnguyen0:
It's not in the middle for me, and also doesn't seem to place nice on screen resize.
screen recording 2018-08-21 at 03 49 pm

mperrotti:
Thanks @GeorgiSTodorov, made the change

PR #3202, opened 2018-08-22
DCOS-22446: remove ignore override

DanielMSchmidt:
## Testing

- Run `npm test` and see if it works

## Trade-offs

- For me the execution time did not drop by a lot, but removing unnecessary config lead me to still do the PR

## Dependencies

None


nLight:
120 vs 140 without the change

PR #3197, opened 2018-08-20
DCOS-40638: Backport form control overflow fix for Firefox

natmegs:
Backport changes that fixed the form control icon overflow issues in Firefox for 1.11. This PR takes the approved changes from [#3186](https://github.com/dcos/dcos-ui/pull/3186) and adds them to the 1.11 branch.

Closes DCOS-40638

## Testing

In Firefox, go to the Components tab and start typing in the filter input. The icons to the left and right of the text should remain aligned. Try also in Chrome to make sure the behaviour remains as expected there as well. 

## Trade-offs

I noticed that the cancel (x) icon to the right of the filter text was aligned noticeably lower than in the screenshots for [#3186](https://github.com/dcos/dcos-ui/pull/3186). `font-size: 0` was added to align the icon within its container. I'm keen for feedback on this, and whether the new alignment is correct (maybe we do not need it at all).

## Comparison

Firefox before: 
firefox-before

Firefox with icons fixed but without right icon alignment: 
firefox-misaligned

Firefox with icons fixed and right icon aligned:
firefox-aligned

Chrome without right icon aligned: 
chrome-before

Chrome with right icon aligned: 
chrome-aligned

PR #3204, opened 2018-08-23
DCOS-40687: [Firefox] Form control icons overflow issue (for other components)

GeorgiSTodorov:
Add min-width to the input to fix overflowing icons in Firefox.

Closes DCOS-40687

## Testing
Verify that the icons are fixed for all input bars (see screenshots). If possible, verify that the icons look good in other browsers.

## Trade-offs
~~I added the `font-size: 0` attribute that @natmegs proposed in https://github.com/dcos/dcos-ui/pull/3197~~
Edit: I added `{display: block}}` to the icons like Mike suggested.

## Dependencies
None

## Screenshots
These are all the places that I found:

### Jobs > (Create a job if needed)
screenshot from 2018-08-23 11-24-22

### Nodes > Node > Health
screenshot from 2018-08-23 11-24-54

### Services > Service > Logs
screenshot from 2018-08-23 11-27-00

### Catalog
screenshot from 2018-08-23 11-28-46

### Networking > Networks
screenshot from 2018-08-23 11-29-33

### Secrets
screenshot from 2018-08-23 11-46-57

### Components
screenshot from 2018-08-23 11-47-35

### Settings > Package Repositories
screenshot from 2018-08-23 11-47-51

### Settings > Identity Providers
screenshot from 2018-08-23 12-15-36

### Organization > Users
screenshot from 2018-08-23 11-50-03

### Organization > Groups
screenshot from 2018-08-23 11-50-12

### Organization > Service Accounts
screenshot from 2018-08-23 11-50-23


natmegs:
Really minor, but on the Nodes page (Agents tab) the filter shows two icons to the right of the text and in this case, those two icons are slightly misaligned. I think it is because the font-size: 0 is being applied only to the last-child icon. It's negligible in Chrome but more noticeable in Firefox. @mperrotti what do you think?
nodes-filter

mperrotti:
@natmegs - the vertical centering is getting ruined by the `line-height` those spans inherit, which is probably why `font-size: 0` works. Using font-size to fix layout feels like something we might regret later.

I recommend we make icons that are children of `.form-control-group-add-on` to be `display: block;` so they don't factor line-height into their height calculation. Currently, they're `display: inline-block;`
```
.form-control-group-add-on .icon { display: block; }
```

---

Alternatively, if there's some reason we _must_ have those elements be `display: block`, you could force the `.form-control-group-add-on` span to fill it's container and vertically align it's contents. The quickest way to do that, would be something like this:
```
.form-control-group-add-on {
    display: flex;
    align-items: center;
    align-self: stretch;
}
```

If neither of these work, I'm sure we could come up with something else :)

natmegs:
LGTM

PR #3206, opened 2018-08-24
DCOS-40519: Add Intercom plugin

tnguyen0:
Adds intercom plugin.

Closes DCOS-40519

## Testing
* Add the bits from https://github.com/dcos/dcos-ui/pull/3206/files#diff-33ee3a011e0fc7f6816d74252110fedd to your `Config.dev.ts` and make sure `useUIConfigFixtures` set to `true`
* Test on Open and EE cluster
* Intercom widget appears when logged in
* Intercom widget disappears when logged out
* Send a message on intercom
* Check on intercom everything was updated (Inbox -> Messages)

### DIFF
--- a/plugins/intercom/SDK.js
+++ b/plugins/intercom/SDK.js
@@ -1,0 +1,10 @@
+let SDK;
+
+module.exports = {
+  getSDK() {
+    return SDK;
+  },
+  setSDK(pluginSDK) {
+    SDK = pluginSDK;
+  }
+};
--- a/plugins/intercom/__tests__/IntercomPlugin-test.js
+++ b/plugins/intercom/__tests__/IntercomPlugin-test.js
@@ -1,0 +1,40 @@
+jest.setMock("react-router", {
+  hashHistory: {
+    location: { pathname: "/foo" },
+    listen() {}
+  }
+});
+
+const mockAuthStore = {
+  getUser: jest.fn(),
+  isLoggedIn: jest.fn()
+};
+
+jest.mock("#SRC/js/stores/AuthStore", function() {
+  return mockAuthStore;
+});
+
+const PluginTestUtils = require("PluginTestUtils");
+const DOMUtils = require("#SRC/js/utils/DOMUtils");
+
+const SDK = PluginTestUtils.getSDK("intercom", { enabled: true });
+require("../SDK").setSDK(SDK);
+
+const IntercomHooks = require("../hooks");
+
+describe("IntercomHooks", function() {
+  describe("Listeners", function() {
+    beforeEach(function() {
+      DOMUtils.appendScript = jasmine.createSpy();
+    });
+
+    describe("#pluginsConfigured", function() {
+      it("appends scripts to the document head if plugin enabled", function() {
+        global.analytics = { ready() {} };
+        IntercomHooks.initialize({ appId: "id-123" });
+        SDK.Hooks.doAction("pluginsConfigured");
+        expect(DOMUtils.appendScript.calls.count()).toEqual(1);
+      });
+    });
+  });
+});
--- a/plugins/intercom/hooks.js
+++ b/plugins/intercom/hooks.js
@@ -1,0 +1,94 @@
+import { Hooks } from "PluginSDK";
+
+import DOMUtils from "#SRC/js/utils/DOMUtils";
+import AuthStore from "#SRC/js/stores/AuthStore";
+
+import { INTERCOM_CHANGE } from "./constants/EventTypes";
+import IntercomStore from "./stores/IntercomStore";
+
+const SDK = require("./SDK").getSDK();
+
+const INTERCOM_LOAD_TIMEOUT = 500;
+
+module.exports = {
+  filters: ["pluginsLoadedCheck"],
+  actions: ["pluginsConfigured", "userLoginSuccess", "userLogoutSuccess"],
+
+  initialize(configuration) {
+    this.configuration = configuration;
+
+    this.filters.forEach(filter => {
+      SDK.Hooks.addFilter(filter, this[filter].bind(this));
+    });
+    this.actions.forEach(action => {
+      SDK.Hooks.addAction(action, this[action].bind(this));
+    });
+
+    this.bootIntercom = this.bootIntercom.bind(this);
+    this.onIntercomStoreUpdate = this.onIntercomStoreUpdate.bind(this);
+
+    const intercomScript = `
+      var APP_ID = "${this.configuration.appId}";
+      (function(){var w=window;var ic=w.Intercom;if(typeof ic==="function"){ic('reattach_activator');ic('update',intercomSettings);}else{var d=document;var i=function(){i.c(arguments)};i.q=[];i.c=function(args){i.q.push(args)};w.Intercom=i;function l(){var s=d.createElement('script');s.type='text/javascript';s.async=true;s.src='https://widget.intercom.io/widget/' + APP_ID;var x=d.getElementsByTagName('script')[0];x.parentNode.insertBefore(s,x);}if(w.attachEvent){w.attachEvent('onload',l);}else{w.addEventListener('load',l,false);}}})()
+    `;
+
+    DOMUtils.appendScript(global.document.head, intercomScript);
+  },
+
+  pluginsLoadedCheck(promiseArray) {
+    const promise = new Promise(resolve => {
+      if (global.Intercom) {
+        resolve();
+      }
+
+      setTimeout(resolve, INTERCOM_LOAD_TIMEOUT);
+    });
+
+    promiseArray.push(promise);
+
+    return promiseArray;
+  },
+
+  pluginsConfigured() {
+    if (AuthStore.isLoggedIn()) {
+      this.bootIntercom();
+    }
+  },
+
+  userLoginSuccess() {
+    this.bootIntercom();
+  },
+
+  userLogoutSuccess() {
+    IntercomStore.removeChangeListener(
+      INTERCOM_CHANGE,
+      this.onIntercomStoreUpdate
+    );
+
+    global.Intercom("shutdown");
+
+    Hooks.doAction("intercomShutdown");
+  },
+
+  onIntercomStoreUpdate() {
+    global.Intercom("update", IntercomStore.attributes);
+  },
+
+  bootIntercom() {
+    const user = AuthStore.getUser();
+
+    global.Intercom("boot", {
+      app_id: this.configuration.appId,
+      email: user.email,
+      name: user.name,
+      user_id: user.uid
+    });
+
+    IntercomStore.addChangeListener(
+      INTERCOM_CHANGE,
+      this.onIntercomStoreUpdate
+    );
+
+    Hooks.doAction("intercomBoot");
+  }
+};
--- a/plugins/intercom/index.js
+++ b/plugins/intercom/index.js
@@ -1,0 +1,9 @@
+import SDK from "./SDK";
+
+module.exports = function(PluginSDK) {
+  SDK.setSDK(PluginSDK);
+
+  const PluginHooks = require("./hooks");
+
+  PluginHooks.initialize(PluginSDK.config);
+};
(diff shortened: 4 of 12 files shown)


src/js/config/Config.template.ts
```diff
@@ -11,6 +11,10 @@ export default {
         banner: {
           enabled: false
         },
+        intercom: {
+          enabled: true,
+          appId: "wn4z9z0y"
```
DanielMSchmidt: 💥 Can we please use something like "<appId from intercom>" here?

plugins/intercom/index.js
```diff
@@ -0,0 +1,9 @@
+import SDK from "./SDK";
+
+module.exports = function(PluginSDK) {
```
DanielMSchmidt: What problem does this solve?

src/js/stores/IntercomStore.js
```diff
@@ -0,0 +1,41 @@
+import PluginSDK from "PluginSDK";
+
+import GetSetBaseStore from "./GetSetBaseStore";
+import { INTERCOM_CHANGE } from "../constants/EventTypes";
+
+class IntercomStore extends GetSetBaseStore {
+  constructor() {
+    super(...arguments);
+
+    this.getSet_data = {
```
DanielMSchmidt: This seems unused, why do we need it?

plugins/intercom/hooks.js
```diff
@@ -0,0 +1,82 @@
...
+        email: user.email,
+        name: user.name,
+        userId: user.uid
+      };
+
+      Object.keys(attributes).forEach(attribute => {
+        IntercomStore.addAttribute(attribute, attributes[attribute]);
+      });
+
+      window.Intercom("boot", IntercomStore.attributes);
```
DanielMSchmidt: So worst case we boot three times, right? Doesn't this overmount the entire thing, so if the change handlers are triggered later on after the user already wrote things it would be just replaced? I wonder if [update](https://developers.intercom.com/installing-intercom/docs/intercom-javascript#section-intercomupdate) might be a better use-case

plugins/intercom/hooks.js
```diff
@@ -0,0 +1,82 @@
...
+    this.actions.forEach(action => {
+      SDK.Hooks.addAction(action, this[action].bind(this));
+    });
+
+    this.bootIntercom = this.bootIntercom.bind(this);
+
+    const intercomScript = [
+      `var APP_ID = "${this.configuration.appId}";`,
+      `(function(){var w=window;var ic=w.Intercom;if(typeof ic==="function"){ic('reattach_activator');ic('update',intercomSettings);}else{var d=document;var i=function(){i.c(arguments)};i.q=[];i.c=function(args){i.q.push(args)};w.Intercom=i;function l(){var s=d.createElement('script');s.type='text/javascript';s.async=true;s.src='https://widget.intercom.io/widget/' + APP_ID;var x=d.getElementsByTagName('script')[0];x.parentNode.insertBefore(s,x);}if(w.attachEvent){w.attachEvent('onload',l);}else{w.addEventListener('load',l,false);}}})()`
+    ].join("");
```
DanielMSchmidt: 🎨 This is already an array of multi-line strings, couldn't we just use one big multi-line string?

plugins/intercom/hooks.js
```diff
@@ -0,0 +1,82 @@
...
+
+    DOMUtils.appendScript(global.document.head, intercomScript);
```
DanielMSchmidt: If I get it right this includes a script that loads a bigger script from some backend service? Is it really save to assume that `window.Intercom` is always there? What happens if we block the endpoint or throttle it? I would like is to have a bit more caution here.

plugins/intercom/SDK.js
```diff
@@ -0,0 +1,10 @@
+let SDK;
+
+module.exports = {
```
DanielMSchmidt: Why do we need this? What does it solve?

src/js/config/Config.template.ts
nLight: I went with the code because we already have `analyticsKey: "39uhSEOoRHMw6cMR6st9tYXDbAL3JSaP"` in the same file and that would ease the friction testing intercom for every newcomer. It's a public key after all.

plugins/intercom/index.js
nLight: This is the common interface of the current plugin system. This is how we ensure we use the very same instance of `SDK` for all plugins.

plugins/intercom/hooks.js
```diff
@@ -0,0 +1,82 @@
...
+      });
+
+      window.Intercom("boot", IntercomStore.attributes);
+    };
+
+    if (!MetadataStore.version) {
+      MetadataStore.addChangeListener(EventTypes.DCOS_METADATA_CHANGE, boot);
+    }
+
+    if (!MetadataStore.clusterId) {
```
nLight: If `version` is there then `clusterId` should also be there. To be on a safe side we could merge two checks into one so that we don't add too many listeners

src/js/stores/IntercomStore.js
nLight: Please look inside `GetSetBaseStore` this is the way to define internal scope `attributes` so that you could do `this.get("attributes")`

src/js/stores/IntercomStore.js
```diff
@@ -0,0 +1,41 @@
...
+      },
+      unmountWhen() {
+        return true;
+      },
+      listenAlways: true
+    });
+  }
+
+  addAttribute(key, value) {
+    const attributes = this.get("attributes");
```
nLight: Here I'd use internal `set` method instead:
```js
this.set("attributes", { [key]: value })
```

plugins/intercom/hooks.js
```diff
@@ -0,0 +1,82 @@
...
+        "dcos-image-commit": MetadataStore.imageCommit,
+        dcosVersion: MetadataStore.version,
+        dcosVariant: MetadataStore.variant,
+        email: user.email,
+        name: user.name,
+        userId: user.uid
+      };
+
+      Object.keys(attributes).forEach(attribute => {
+        IntercomStore.addAttribute(attribute, attributes[attribute]);
```
nLight: Can't we just `IntercomStore.set("attributes", attributes)`?

GeorgiSTodorov:
`Check on intercom everything was updated (Inbox -> Messages)`
How to check that?

nLight:
@GeorgiSTodorov it's basically one of us either @tnguyen0 or me should do. Not everyone has access since we have limited number of seats

GeorgiSTodorov:
@nLight I send a "hi" message to test it, was it received? And do I need to do anything else to test https://github.com/mesosphere/dcos-ui-plugins-private/pull/826 or is testing this current PR is enough?

plugins/intercom/index.js
DanielMSchmidt: How does it ensure that?

DanielMSchmidt:
@GeorgiSTodorov Yeah, got an email with a hi message 👍

plugins/intercom/hooks.js
nLight: oh, those are different things!

nLight: Though I wonder if the `IntercomStore` should listen to every other store and emit its own events and the plugin should be subscribed to `IntercomStore` only? Can we actually do a data layer real quick here? @Fabs

plugins/intercom/hooks.js
```diff
@@ -0,0 +1,101 @@
...
+
+  userLogoutSuccess() {
+    global.Intercom("shutdown");
+  },
+
+  bootIntercom() {
+    const boot = () => {
+      const user = AuthStore.getUser();
+
+      const attributes = {
```
Fabs: I am not sure if this data is being properly passed to intercom. I started a chat with myself, but was unable to see that information on Intercom under

plugins/intercom/hooks.js
Fabs: That might be related with the fact that I do not see the user information on bootIntercom(), on the Intercom Dashboard

plugins/intercom/hooks.js
Fabs: I am not sure If I understand the use case of the DL here. Can you elaborate?

I see an opportunity to use a stream to simplify that code (according to my subjective perception of simple). The pattern I see is that we call boot() when a lot of things happen. There is also the async load of the JS code, and even a timeout. We could make a stream that has a source for all these events and transforms into the call to boot, also triggering the Hooks.doAction.

nLight: Use DL to pump data into the plugin instead of writing another store

plugins/intercom/stores/IntercomStore.js
```diff
@@ -0,0 +1,99 @@
...
+} from "#SRC/js/constants/EventTypes";
+import MetadataStore from "#SRC/js/stores/MetadataStore";
+
+import { INTERCOM_CHANGE } from "../constants/EventTypes";
+
+class IntercomStore extends GetSetBaseStore {
+  constructor() {
+    super(...arguments);
+
+    this.isListening = false;
```
weblancaster: @nLight why we had to introduce this state? I would expect us to know from the lifecycle when there is the need to add or remove listeners without having to use conditional statements.

plugins/intercom/stores/IntercomStore.js
```diff
@@ -0,0 +1,99 @@
...
+        return true;
+      },
+      listenAlways: true
+    });
+  }
+
+  removeChangeListener(eventName, callback) {
+    this.removeListener(eventName, callback);
+
+    if (this.isListening) {
```
weblancaster: Related to https://github.com/dcos/dcos-ui/pull/3206/commits/734dec5d478cdd03a5bc1fb5b65750598ee3c0fd#r213110433

plugins/intercom/stores/IntercomStore.js
```diff
@@ -0,0 +1,99 @@
...
+        METADATA_CHANGE,
+        this.onMetadataChange
+      );
+    }
+  }
+
+  addChangeListener(eventName, callback) {
+    this.on(eventName, callback);
+
+    if (!this.isListening) {
```
weblancaster: Related to https://github.com/dcos/dcos-ui/pull/3206/commits/734dec5d478cdd03a5bc1fb5b65750598ee3c0fd#r213110433

plugins/intercom/stores/IntercomStore.js
```diff
@@ -0,0 +1,99 @@
+import PluginSDK from "PluginSDK";
```
weblancaster: should we add tests for `IntercomStore.js`

tnguyen0:
@DanielMSchmidt @nLight @Fabs @weblancaster @philipnrmn @GeorgiSTodorov 

Ready to go I think, let's get it merged!

plugins/intercom/stores/IntercomStore.js
weblancaster: Haha I gave him some direction while writing it.

plugins/intercom/__tests__/IntercomPlugin-test.js
```diff
@@ -0,0 +1,39 @@
...
+
+const IntercomHooks = require("../hooks");
+
+describe("IntercomHooks", function() {
+  describe("Listeners", function() {
+    beforeEach(function() {
+      DOMUtils.appendScript = jasmine.createSpy();
+    });
+
+    describe("#pluginsConfigured", function() {
```
weblancaster: not a big deal but I wonder why so many nested `describe`?

plugins/intercom/hooks.js
```diff
@@ -0,0 +1,94 @@
...
+    DOMUtils.appendScript(global.document.head, intercomScript);
+  },
+
+  pluginsLoadedCheck(promiseArray) {
+    const promise = new Promise(resolve => {
+      if (global.Intercom) {
+        resolve();
+      }
+
+      setTimeout(resolve, INTERCOM_LOAD_TIMEOUT);
```
weblancaster: Aren't we assuming that regardless of what happens we will always call "succeed"/`resolve`? is `global.Intercom` expected to be`true` based on `INTERCOM_LOAD_TIMEOUT` only.

natmegs: I was curious about this as well

PR #3209, opened 2018-08-24
Updated DC/OS CLI installation link to latest.

ArmandGrillet:
The new DC/OS CLI is designed to work for any version of DC/OS >= 1.10.
We can thus safely use the latest release for the current and upcoming
versions of DC/OS.

### DIFF
--- a/src/js/components/modals/CliInstallModal.js
+++ b/src/js/components/modals/CliInstallModal.js
@@ -65,14 +65,9 @@ class CliInstallModal extends React.Component {
     }
     const clusterUrl = `${protocol}://${hostname}${port}`;
     const { selectedOS } = this.state;
-    let version = MetadataStore.parsedVersion;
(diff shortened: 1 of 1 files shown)


jaxesn:
Is there value in having a specific version link tho.  As we update, older clusters would still point to their respective version of the documentation?

nLight:
@jaxesn I think CLI is version agnostic nowadays. Is that correct @AmrAbdelrazik? Why do we want to always point to the latest?

jaxesn:
Oh yeah, sorry I was thinking docs for some reason, not cli 🤦‍♂️

ArmandGrillet:
The new CLI is now version agnostic (works with any DC/OS version as long as it's>= 1.10). There will be no release for 1.12 thus the current link is currently broken on `master`.

ArmandGrillet:
@juliangieseke Thanks for the help, the commit message should now be properly formatted.

PR #3210, opened 2018-08-24
DCOS-40018: upgrade ui-kit to 1.10.1

mperrotti:
## Testing
1. Use soak cluster https://soak112s.testing.mesosphe.re
2. Navigate to Nodes view
3. Scroll up and down very fast

The row alignment should not shifting

## Trade-offs
Not that I can think of

## Dependencies
Nothing

## Screenshots
See https://github.com/dcos-labs/ui-kit/pull/212 for screenshots and more information

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
     "@dcos/http-service": "0.3.0",
     "@dcos/mesos-client": "0.2.4",
     "@dcos/tslint-config": "0.1.0",
-    "@dcos/ui-kit": "1.9.0",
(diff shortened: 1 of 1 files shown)


weblancaster:
@jaxesn I agree, I think we can benefit from using range in this case.

mperrotti:
I made this change and screwed up the rebase locally, so I'm just going to merge as-is and make sure I do `@dcos/ui-kit` as a range going forward

PR #3217, opened 2018-08-28
ci: use local dcos-launch bin

orlandohohmeier:
Adjust Dockerfile to use local docs-launch instead of installing it from
a remote location. This changes is a hot fix to workaround issues with
the latest dcos-launch release which broke out system tests.

## Testing

With this change our CI should be able to spin up a cluster for out system test again.

## Trade-offs

We need to revert this change once we've agreed on a release and integration process with the respective team to always integrate with latest release when possible.


nLight:
I see system tests are running again 💪

orlandohohmeier:
@juliangieseke @nLight could you please re-review?

PR #3218, opened 2018-08-28
DCOS-40048: add toggle to date information on leader tab

juliangieseke:
JIRA: [DCOS-40048](https://jira.mesosphere.com/browse/DCOS-40048)

## Testing

Navigate to `/nodes/masters` and click on the text stating `X Something ago` - it should toggle to a Date. Verify that that Date is formatted correctly.

## Trade-offs

* When using the `<ToggleContent />` component there is no dotted line, is this line supposed to be added in dcos-ui or should it be part of the `<ToggleContent />` component? 🤔 
* I decided to change the diplayed format of `DateUtil.msToDateStr` to our desired format, the function is only used on Nodes Details page and since we're on Master Detail here, I think it is good thing to have them aligned. 

## Screenshots

screen recording 2018-08-28 at 04 15 pm

### DIFF
--- a/plugins/nodes/src/js/components/LeaderGrid.js
+++ b/plugins/nodes/src/js/components/LeaderGrid.js
@@ -1,6 +1,6 @@
 import React from "react";
 import { FormattedMessage } from "react-intl";
-import moment from "moment";
+import { ToggleContent } from "@dcos/ui-kit";
 
 import ConfigurationMap from "#SRC/js/components/ConfigurationMap";
 import ConfigurationMapHeading from "#SRC/js/components/ConfigurationMapHeading";
@@ -9,14 +9,7 @@ import ConfigurationMapRow from "#SRC/js/components/ConfigurationMapRow";
 import ConfigurationMapSection from "#SRC/js/components/ConfigurationMapSection";
 import ConfigurationMapValue from "#SRC/js/components/ConfigurationMapValue";
 import Loader from "#SRC/js/components/Loader";
-
-function timeFromNow(time) {
-  if (!time) {
-    return null;
-  }
-
-  return moment(time * 1000).fromNow();
-}
+import { msToRelativeTime, msToDateStr } from "#SRC/js/utils/DateUtil";
 
 const Loading = () => <Loader size="small" type="ballBeat" />;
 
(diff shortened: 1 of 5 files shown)


plugins/nodes/src/js/components/LeaderGrid.js
```diff
@@ -18,6 +19,14 @@ function timeFromNow(time) {
   return moment(time * 1000).fromNow();
 }
 
+function formattedDate(time) {
+  if (!time) {
+    return null;
+  }
+
+  return moment(time * 1000).format("MMMM Do YYYY, h:mm:ss a");
```
Fabs: The original design had no seconds. Do we really think it is valuable to show seconds?

Fabs: Actually, I think the format compatible with what Licensing does would be:
MMMM Do, YYYY h:mm a

plugins/nodes/src/js/components/LeaderGrid.js
```diff
@@ -18,6 +19,14 @@ function timeFromNow(time) {
   return moment(time * 1000).fromNow();
 }
 
+function formattedDate(time) {
```
Fabs: What about refactoring both time functions to a time util. I think in the future, it will help a lot to deal consistently with time across DC/OS UI.

weblancaster: Agree.

juliangieseke: there actually already is `DateUtil.msToDateStr`, but its output looks weird: `
08-28-2018 at 9:17am` 🤔

natmegs: Looks like the only place that existing date util function is being used is in the Nodes Details tab. Maybe it would make sense to have the dates in the same format in both places? Also it looks like there is a `DateUtil.msToRelativeTime` that maybe could replace `timeFromNow` here?

juliangieseke:
@Fabs @natmegs @mperrotti @weblancaster comments addressed, please revisit 🙇

DanielMSchmidt:
> When using the <ToggleContent /> component there is no dotted line, is this line supposed to be added in dcos-ui or should it be part of the <ToggleContent /> component? 

I think this is not a ui-kit concern as the component seems pretty generic

DanielMSchmidt:
I found this bug:

bug => maybe we should solve it in the UI-kit, shall I create a ticket?

juliangieseke:
after offline chat with @DanielMSchmidt and @orlandohohmeier : it would be great to have a `TextToggle` component that can be used inline and also provides a consistent visualisation. meaning, the dotted line should be added in UI Kit. we also had the idea to show a tooltip with the not-currently-displayed value.

juliangieseke:
regarding the bug - yes, that also needs to be fixed in UI Kit

Fabs:
I think if we do not add the dotted line bellow the name, it is hard for the user to discover he can click. what do you think?

Also I have this selection behavior that I think should no be present. It makes a text selection if you click outside the text to cause a toggle (to the  right of it, in the white area, close to the text)
https://cl.ly/acb9c4e15dbe

natmegs:
LGTM (if we are handling dotted line in ui kit)

mperrotti:
My fix is in ui-kit version 1.10.2

lilysamimi:
Agree with you @Fabs that without the dotted line treatment, this feature is not noticeable to the user. Don't think we should ship without the dotted line treatment.

weblancaster:
@lilysamimi @Fabs this is not part of this PR and we should not block it.

@mperrotti merged and released the fix, we just need to update the UI Kit package in DCOS UI

juliangieseke:
UI Kit update PR: https://github.com/dcos/dcos-ui/pull/3220 🎉
@Fabs ping 🙇

PR #3219, opened 2018-08-29
chore: double quote paths so that it allows spaces

nLight:
I've messed up my new laptop setup and now I have a space in my home directory name 😩 
Because of that I realized that our scripts don't support that. So I double quoted everything that `shellcheck` asked me to double quote and what made sense to me.

On top of that I followed its advice and added `:?` to `rm -rf ${VAR}` so it become `rm -rf ${VAR:?}` so that if `$VAR` for whatever reason empty we don't kill `/`

## Testing

Try running those scripts from a path that has a space in it.

## Trade-offs

There are some more stuff `shellcheck` complains about. I decided keeping the scope rather small.

### DIFF
--- a/scripts/ci/upload-release
+++ b/scripts/ci/upload-release
@@ -25,7 +25,7 @@ set -x
 SCRIPT_PATH="$(cd $(dirname "$0")/$(dirname "$(readlink "$0")") && pwd)"
 
 # project root for this file
-PROJECT_ROOT="$( cd "$( echo ${SCRIPT_PATH} | sed s+/scripts/ci++)" && pwd )"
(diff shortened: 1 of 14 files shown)


nLight:
@juliangieseke yes, I did that and it has created a new user instead. Now I'm recovering my configs and stuff

nLight:
@orlandohohmeier puts a smiley but no ✅ :)

orlandohohmeier:
LGTM! 🚢

orlandohohmeier:
@nLight forgot to submit my review... 🤦‍♂️

PR #3221, opened 2018-08-30
DCOS-40780: Fix filter on networks table

natmegs:
Fix filter function to allow filtering of networks table. Previously, you could only type one character into the filter and the rest wouldn't appear, and the table contents weren't filtering as expected.

Closes DCOS-40780

## Testing

Go to networks table and try filtering based on Name and IP Subnet. Ensure that you try a search string that matches the subnet in both entries (if you're testing with the regular shared cluster).

## Trade-offs

None that are obvious to me.


Fabs:
The test is probably failing because master was red due to the problem fixed on #3224.

A rebase should do it :D

natmegs:
Rebased and CI is green if you guys don't mind taking one more look! @Fabs @Poltergeist @DanielMSchmidt

PR #3222, opened 2018-08-31
DCOS-40685: remove text truncation for load balanced address

tnguyen0:
Removes truncation and wraps content for "Load Balanced Address" column under the Networking section on the review service screen.

Closes DCOS-40685

## Testing

* Open Create Service Form and paste the following in JSON Editor:
```
{
  "id": "/truncation-test",
  "instances": 1,
  "container": {
    "portMappings": [
      {
        "containerPort": 80,
        "hostPort": 0,
        "labels": {
          "VIP_0": "/truncation-test:10101"
        },
        "protocol": "tcp",
        "name": "test"
      }
    ],
    "type": "DOCKER",
    "volumes": [],
    "docker": {
      "image": "nginx"
    }
  },
  "cpus": 0.1,
  "mem": 128,
  "requirePorts": false,
  "networks": [
    {
      "mode": "container/bridge"
    }
  ],
  "healthChecks": [],
  "fetch": [],
  "constraints": []
}
```
* Click "Review & Run"
* "Load Balanced Address" column content is not truncated

## Trade-offs
* When the header is truncated, the caret does not show on the column, I wasn't sure what the best way to fix this is

## Dependencies
None

### Before
screen shot 2018-08-30 at 5 19 02 pm

### After
screen shot 2018-08-30 at 5 17 52 pm

### DIFF
--- a/src/styles/content/tables/styles.less
+++ b/src/styles/content/tables/styles.less
@@ -111,6 +111,10 @@
       }
     }
 
+    td.wrap-content {
+      white-space: normal;
+    }
+
     th,
     td {
       // This property is necessary for table cells to respect the text-overflow
--- a/src/styles/content/tables/styles.less
+++ b/src/styles/content/tables/styles.less
@@ -12,13 +12,6 @@
 
   .table {
 
-    &.table-header-nowrap {
-
-      th {
-        white-space: nowrap;
-      }
-    }
-
     &.table-hide-header {
 
       thead {
--- a/plugins/services/src/js/components/DeclinedOffersTable.js
+++ b/plugins/services/src/js/components/DeclinedOffersTable.js
@@ -336,7 +336,7 @@ class DeclinedOffersTable extends React.Component {
     return (
       <MountService.Mount type="DeclinedOffersTable">
         <Table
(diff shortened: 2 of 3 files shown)


src/styles/content/tables/styles.less
```diff
@@ -12,13 +12,6 @@
 
   .table {
 
-    &.table-header-nowrap {
-
-      th {
-        white-space: nowrap;
```
juliangieseke: I bet this breaks something somewhere else… 😞 
Also, it is about `th`, is it necessary for this PR? 🤔

tnguyen0: @juliangieseke it is redundant because it is nowrap by default on th, so I decided to remove it. I made sure to check all instances were the class was used and I removed it there as well.

Not exactly necessary, but I figured I would remove some unnecessary code 🤷‍♂️

juliangieseke: Ah ok, I think it is a good idea then to keep bugfix commits separate from refactor commits. having the bugfix only fixing the bug makes it easier to cherry-pick and blame later on if necessary 🙇

PR #3229, opened 2018-08-31
DCOS_OSS-526: fix(package repository settings): remove url validation on add

Poltergeist:
This removes the regex validation of the url field and lets rather the backend decide if a url is
valid or not.

Close DCOS_OSS-526

For Review, you may look at 65a30cfa341e771d216a1a1037504455df8c5be1 this is the commit which if filtered of the code style changes.

 ## Testing

To Test this:
- go to `Settings > Package Repository`
- Click the `+`
- Add a random Repository Name
- Add a url like `http://0.0.0.1:8080`
- Add a priority `1`
- Click `Add`

Now you should be presented with the error from the server instead of the error on the input field.

 ## Trade-offs

A Trade off is that we are showing the error from the Server which can be unexpected long for example with the url `http://0.0.0.0:8080` it will show some long message which is not really helpful.

 ## Screenshots

*Before*:
image

*After*: 
image

### DIFF
--- a/plugins/catalog/src/js/repositories/components/AddRepositoryFormModal.js
+++ b/plugins/catalog/src/js/repositories/components/AddRepositoryFormModal.js
@@ -44,10 +44,12 @@ class AddRepositoryFormModal extends React.Component {
         name: "uri",
         placeholder: "URL",
         required: true,
-        validationErrorText: "Must be a valid url with http:// or https://",
+        showError: false,
         showLabel: false,
         writeType: "input",
-        validation: /^https?:\/\/.+\..+$/,
+        validation() {
+          return true;
+        },
         value: ""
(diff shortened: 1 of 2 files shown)


Poltergeist:
Failing tests are legit.

nLight:
The error message for the original stated case from the ticket is not very helpful.
The issue was that when trying to add an URL that is not a valid repo it should say something like `http://soandso.com` is not a valid Universe repo. Shall we forward that to the Catalog team?

screenshot 2018-09-03 at 10 28 06

Poltergeist:
I guess so, we need to ask the server if the repository is reachable. As you mentioned the message is not as helpful we should ask them to give us a better message.

PR #3231, opened 2018-08-31
DCOS-40781: Provide empty message for catalog search

natmegs:
Provide a message indicating when no results were found when searching Catalog page.

Closes DCOS-40781

## Testing

Go to Catalog page and try searching using strings that will match as well as strings that won't. Ensure regular filtering happens as expected and ensure non-matching strings show the empty message. Ensure 'view all' clears the filter as expected.

## Trade-offs

Seems a little convoluted to handle the search results from within `getCommunityPackagesGrid`. Maybe this could eventually be refactored into its own function.

## Screenshots

Before: 
before

After:
after

### DIFF
--- a/src/js/pages/catalog/PackagesTab.js
+++ b/src/js/pages/catalog/PackagesTab.js
@@ -228,7 +228,9 @@ class PackagesTab extends mixin(StoreMixin) {
         return (
           <div className="clearfix">
             {noResults}
-            (<a onClick={this.clearInput}>view all</a>)
+            (<a className="clickable" onClick={this.clearInput}>
+              view all
+            </a>)
           </div>
         );
       }
--- a/src/js/pages/catalog/PackagesTab.js
+++ b/src/js/pages/catalog/PackagesTab.js
@@ -62,7 +62,7 @@ const PackagesEmptyState = () => {
   );
 };
 
-const METHODS_TO_BIND = ["handleSearchStringChange"];
+const METHODS_TO_BIND = ["handleSearchStringChange", "clearInput"];
 
 const shouldRenderCatalogOption = Hooks.applyFilter(
   "hasCapability",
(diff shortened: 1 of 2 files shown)


src/js/pages/catalog/PackagesTab.js
```diff
@@ -208,16 +213,27 @@ class PackagesTab extends mixin(StoreMixin) {
...
-      const packagesWord = StringUtil.pluralize("service", foundPackagesLength);
+      if (foundPackagesLength < 1) {
+        const noResults = `No results were found for your search: "${
+          this.state.searchString
+        }" `;
+
+        return (
+          <div className="clearfix">
+            {noResults}
+            (<a onClick={this.clearInput}>view all</a>)
```
DanielMSchmidt: This should be `cursor: pointer`

src/js/pages/catalog/PackagesTab.js
```diff
@@ -208,16 +213,27 @@ class PackagesTab extends mixin(StoreMixin) {
...
     const titleClasses = classNames("flush-top", {
       short: !isSearchActive,
       tall: isSearchActive
     });
 
     if (isSearchActive) {
       const foundPackagesLength = packages.getItems().length;
-      const packagesWord = StringUtil.pluralize("service", foundPackagesLength);
+      if (foundPackagesLength < 1) {
+        const noResults = `No results were found for your search: "${
```
DanielMSchmidt: I personally see no benefit in this being not inlined, I am just curious why you decided to pull this out in a variable :)

src/js/pages/catalog/PackagesTab.js
Poltergeist: You may use `className="clickable"` for convenience

src/js/pages/catalog/PackagesTab.js
natmegs: That was an existing variable - it is used as well [here](https://github.com/dcos/dcos-ui/blob/00d40ed81619b90ddbaaae9b6abadec5bb91a78a/src/js/pages/catalog/PackagesTab.js#L236). Should I inline in both places?

DanielMSchmidt: no, I was just wondering :D

natmegs:
Good call, @tnguyen0. I've added an integration test (can look at last commit to isolate it from prettier fixes).

PR #3234, opened 2018-09-03
DCOS-20222: add error for environment values without keys

DanielMSchmidt:
## Testing

- Go to the pods / service form
- Go to the environment section
- Create a label without a key
- See that it is shown as an error

## Trade-offs

None that I am aware of

## Dependencies

None

## Screenshots

bildschirmfoto 2018-09-03 um 12 03 57

### DIFF
--- a/plugins/services/src/js/components/forms/EnvironmentFormSection.js
+++ b/plugins/services/src/js/components/forms/EnvironmentFormSection.js
@@ -49,10 +49,15 @@ class EnvironmentFormSection extends Component {
             </FieldLabel>
           );
         }
+        const isValueWithoutKey = !env.key && env.value;
 
         return (
           <FormRow key={key}>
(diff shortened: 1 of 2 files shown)


DanielMSchmidt:
@Poltergeist Fixed that 👍

natmegs:
I'm guessing this is out of scope for this ticket, but it looks like this is handled differently for environment variables, labels, and secrets. Labels without a key don't get added to JSON (no error), secrets with only Variable Name get assigned a random (?) secret value in JSON. Is this intentional?
environment

DanielMSchmidt:
@natmegs Yes, but you are right. I will create two small issues for that cause 👍

DanielMSchmidt:
@nLight @Poltergeist Could you have another look?

PR #3236, opened 2018-09-04
DCOS-40635: add tests for empty nodes page

DanielMSchmidt:
## Testing

🤷‍♂️ I could not reproduce the bug, so I only wrote an integration test for now

## Dependencies

- [ ] Get @GeorgiSTodorov Feedback on how to reproduce


plugins/nodes/src/js/components/NodesGridView.js
```diff
@@ -121,18 +121,20 @@ var NodesGridView = React.createClass({
     );
   },
 
-  render() {
+  shouldRenderLoadingScreen() {
```
nLight: ❤️

nLight:
I think we good to go here. Putting a fix and a test in place when it's hard to repro an issue should be the way to go.

PR #3247, opened 2018-09-06
DCOS-21202: MS Edge sidebar glitch

mperrotti:
In Microsoft Edge, the sidebar of Framework Configuration screens would flash, "glitch-out", or disappear entirely when the user opened the JSON editor and clicked inside with their cursor.
Edge was having rendering performance issues with the `position: fixed` style of the sidebar

Closes DCOS-21202

## Testing
Ping me for an invite to Browserstack
Open any Edge browser
Add a new framework from the "Catalog" section
After you hit "Review and Run" and you're landed on the "Edit Configuration" modal, toggle open the JSON editor
Click inside of the JSON editor

You can also get to the "Edit Configuration" screen by editing an existing framework running in the "Services" tab

After you click into the JSON editor, there should be no content flashing on the screen

## Trade-offs
TL;DR: This fix required some "clever" workarounds with CSS layout that feel a little hack-y

This bug was originally caused by Edge's poor handling of `position: fixed`. To achieve the same effect without `position: fixed` I had to make the right side of the screen a scrolling div. In order for the form content to scroll no matter where on the right half of the screen your cursor was, I had to use CSS grid layout for centering the content instead of relying on the `auto` margins. When the content is centered using `auto` left and right margins, if the cursor was in the margin or too far to the right of the screen, the form content would not scroll.
This layout technique feels a little hack-y, especially since it's to support a very specific use case in a browser with so few users.

## Dependencies
None

## Screenshots
Before:
screen recording 2018-09-06 at 11 06 am

After:
screen recording 2018-09-06 at 11 10 am

### DIFF
--- a/src/styles/components/menu-tabbed/styles.less
+++ b/src/styles/components/menu-tabbed/styles.less
@@ -11,12 +11,33 @@
   }
 
   .menu-tabbed-container-fixed {
+    display: grid;
+    grid-auto-rows: 100%;
+
+    /* Reason: fr actually _is_ a unit */
+    /* stylelint-disable unit-no-unknown */
+    grid-template-columns: repeat(7, 1fr);
+    /* stylelint-enable unit-no-unknown*/
+    height: 100%;
+    width: 100%;
 
     .menu-tabbed-vertical {
+      grid-column-end: 3;
       height: 100%;
-      position: fixed;
-      top: 135px;
+      justify-self: right;
     }
+  }
+
+  .menu-tabbed-view-container-fixed-scrollWrapper {
+    grid-column-end: 8;
+    grid-column-start: 1;
+    overflow: scroll;
+    width: 100%;
+  }
+
+  .menu-tabbed-view-container-fixed-scrollWrapper-content {
+    height: 100%;
+    max-width: @container-width-screen-jumbo - @menu-tabbed-vertical-width;
   }
 
   .menu-tabbed {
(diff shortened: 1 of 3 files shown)


src/styles/components/menu-tabbed/styles.less
```diff
@@ -11,14 +11,35 @@
   }
 
   .menu-tabbed-container-fixed {
+    display: grid;
+    grid-auto-rows: 100%;
+
+    /* Reason: fr actually _is_ a unit */
+    /* stylelint-disable unit-no-unknown */
+    grid-template-columns: repeat(7, 1fr);
```
mperrotti: We should probably update our Stylelint config to recognize `fr` units

mperrotti:
After giving it some thought overnight, I'm not sure this fix is worth it because:
- The original implementation with `position: fixed;` is cleaner
- It's difficult to reproduce the bug in Edge 17, which has by far more users than 15 or 16
- It's a very mild inconvenience to a very small subset of users

jaxesn:
This bug is def not a High, being that IE has such low usage and its not even "as bad" in the latest version of IE. I'd probably skip this and close the bug as wont fix.

nLight:
Since it's there already we could merge it. Do you suggest we just drop it not to waste cycles on review? @jaxesn

nLight:
Ok, we've decided to drop it.

PR #3248, opened 2018-09-06
DCOS-41121: version cell tooltip alignment

mperrotti:
Overflowing text was causing the tooltip to misalign

Closes DCOS-41121

## Testing
Add filler text in the `Tooltip` returned by the `renderVersion()` method to cause an ellipsis
Go to Services tab
Hover the version cell

## Trade-offs
No

## Dependencies
No

## Screenshots
Before:
versioncellalign-before

After:
versioncellalign-after

### DIFF
--- a/plugins/services/src/js/containers/services/ServicesTable.js
+++ b/plugins/services/src/js/containers/services/ServicesTable.js
@@ -429,7 +429,15 @@ class ServicesTable extends React.Component {
         ? FrameworkUtil.extractBaseTechVersion(rawVersion)
         : "";
 
-    return <Tooltip content={rawVersion}>{displayVersion}</Tooltip>;
+    return (
+      <Tooltip
+        content={rawVersion}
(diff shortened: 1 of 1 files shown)


plugins/services/src/js/containers/services/ServicesTable.js
```diff
@@ -429,7 +429,15 @@ class ServicesTable extends React.Component {
         ? FrameworkUtil.extractBaseTechVersion(rawVersion)
         : "";
 
-    return <Tooltip content={rawVersion}>{displayVersion}</Tooltip>;
+    return (
```
jaxesn: This would be something we would need to do anywhere we use tooltips within the new table? Is there maybe a more general solution? not a blocker for this PR, just a thought.

mperrotti: A more general solution would be good, I just have to think about what that might look like. A more general solution would probably happen in ui-kit's Table styles

PR #3249, opened 2018-09-07
[1.10] DCOS_OSS-4034: show VIP fields in host mode

tnguyen0:
## Testing

Backport for 1.10. See: https://github.com/dcos/dcos-ui/pull/3233

## Trade-offs

* In 1.10, the load balanced port is the same as the container port when network type is set to container (Virtual Network: dcos in the UI)
* To add load balanced port for network type "Host", it did not make sense to me to set the load balanced port to be the same as host port, therefore I added the load balanced port field.


nLight:
run integration tests

nLight:
That one is tricky. We effectively backport a feature. But then ability to set VIP for Pod is also sounds more like a feature to me. So I'd say it's fine.

PR #3250, opened 2018-09-07
DCOS-41500: show JSON parsing errors as top level errors

DanielMSchmidt:
## Testing

Go to the Service form and open the JSON editor. Create invalid JSON. Admire the error message being shown on the left side and the review button not doing anything on click

## Trade-offs

- Decided against system tests, as this is a UI only issue IMHO

## Dependencies

None

## Screenshots

bildschirmfoto 2018-09-07 um 15 03 02

### DIFF
--- a/plugins/services/src/js/components/modals/CreateServiceModalForm.js
+++ b/plugins/services/src/js/components/modals/CreateServiceModalForm.js
@@ -53,6 +53,7 @@ const METHODS_TO_BIND = [
   "handleFormChange",
   "handleJSONChange",
   "handleJSONPropertyChange",
+  "handleJSONErrorStateChange",
   "handleRemoveItem",
   "handleClickItem"
 ];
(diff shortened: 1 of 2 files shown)


Fabs:
Do you think it makes sense to hide or disable/hard the form? I created a syntax error, then I went to edit the form, and it still performs changes on the JSON. I wonder if it could lead to a problem.

Also, there is this ticket for next sprint I guess (https://jira.mesosphere.com/browse/DCOS-39660) this is a fallback for the editor, for when the schema does not match, which causes our reducers tp break. I wonder if that ticket and this ticket do not overlap.

DanielMSchmidt:
@Fabs I don't think I understand, could you create a ticket for that?

Fabs:
@DanielMSchmidt I don't think it needs a ticket. My ask is for you to take a look on DCOS-16032 and DCOS-39660 and, with the knowledge you have about your ticket, and see if you have any comments about the way our team should proceed on those tickets.

It looks to me that this work, and the work that those tickets need are related.

PR #3262, opened 2018-09-11
DCOS-39800: include executor in getHostResourcesByFramework

tnguyen0:
The resources shown on the nodes table and nodes grid view were not in sync. The difference is accounted for by including the resources used by the executor. This was not included in the calculation for the nodes grid view.

Closes DCOS-39800

## Testing

* Deploy Kafka from the Catalog
* Click on Nodes Tab
* Check between the List/Grid view _(The resources (CPU/Mem/Disk) match between the two views)_

## Trade-offs

* First I looked to NOT use the tasks in computing the framework resource usage per slave
* From looking at the data returned in `lastMesosState` and `CompositeState`, I could see no good way to relate frameworks with slaves
* I decided to simply just include executors along with tasks in `getHostResourcesByFramework`

## Dependencies

None

### DIFF
--- a/src/js/utils/MesosStateUtil.js
+++ b/src/js/utils/MesosStateUtil.js
@@ -55,6 +55,22 @@ const MesosStateUtil = {
   },
 
   /**
+   * @param {Array.<Object>} resourceList - Verbose resource information
+   * @returns {Object} An object of only the resource values
+   */
+  extractExecutorResources(resourceList) {
+    const resources = {};
+
+    resourceList.forEach(resource => {
+      if (resource.type === "SCALAR") {
+        resources[resource.name] = resource.scalar.value;
+      }
+    });
+
+    return resources;
+  },
+
+  /**
    * Returns resource usage of non completed tasks grouped by Host and Framework
    *
    * @param  {Object} state A document of mesos state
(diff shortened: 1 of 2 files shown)


src/js/utils/MesosStateUtil.js
```diff
@@ -54,6 +54,41 @@ const MesosStateUtil = {
       });
   },
 
+  /**
+   * @param {Array.<Object>} resourceList - Verbose resource information
+   * @returns {Object} An object of only the resource values
+   */
+  extractExecutorResources(resourceList) {
```
juliangieseke: not entirely sure, but this should live in one of the mesosStream parsers I think 🤔- @nLight ?

src/js/utils/MesosStateUtil.js
```diff
@@ -54,6 +54,41 @@ const MesosStateUtil = {
...
+
+  /**
+   * @param {Array} tasks
+   * @param {Array} executors
+   * @returns {Array} Tasks and executors concatenated
+   */
+  concatTasksExecutors(tasks, executors) {
+    const tasksExecutors = [].concat(tasks);
+
+    executors.forEach(executor =>
```
juliangieseke: 🎨 Instead of using `forEach` and `push` here, you also could do `executors.map` to get a new array 🤔

src/js/utils/MesosStateUtil.js
```diff
@@ -63,32 +98,39 @@ const MesosStateUtil = {
...
-        }
-
-        return memo;
-      }, {});
+    const tasks = (state.tasks || []).filter(
+      task => !COMPLETED_TASK_STATES.includes(task.state)
+    );
+    const executors = state.executors || [];
+
+    return this.concatTasksExecutors(tasks, executors).reduce(function(
```
juliangieseke: 🎨 as stated above: `return tasks.concat(executors.map(…)).reduce(…)`

src/js/utils/MesosStateUtil.js
tnguyen0: I took a look, but I don't see a util being added there.

src/js/utils/MesosStateUtil.js
tnguyen0: Much better! 💯

tnguyen0:
@juliangieseke comments addressed

tnguyen0:
The `NodesTableContainer` uses `CompositeState` which contains an aggregate of the resources used per node.

For the nodes grid view, we want to know the resources used per framework per node, hence this aggregate does not suffice. We instead compute this aggregation ourselves, but it was missing the executor resources, hence the mismatch.

GeorgiSTodorov:
I just checked again on the frontend cluster and it's correct now. No idea why it was incorrect earlier.

PR #3263, opened 2018-09-11
DCOS-8601: Ensure package detail links open in new tab

natmegs:
Ensure links on package detail page rendered from markdown open in new tab.

Closes DCOS-8601.

## Testing

- Go to catalog page
- Click on spark
- Ensure link in Description section opens in new tab
- Back to catalog page
- Click on kubernetes
- Ensure link in pre install section opens in new tab

## Trade-offs

Some links were opening in new tabs already and some were not - the ones that were not were being rendered from markdown. This makes wrapping the individual links in a component difficult since they are not separated from other text. We are currently using a package called `marked` to parse the markdown in the StringUtil `parseMarkdown` method. I've modified the `marked` renderer to add target="_blank" to links. This affects links in the Description, Pre Install notes, and Post Install notes sections.

For a discussion about modifying `marked` links, see [this PR](https://github.com/markedjs/marked/pull/659) from the `marked` project.

The StringUtil method is also used in FrameworkConfiguration. I've made the assumption that we want the behaviour to be consistent in both locations.

### DIFF
--- a/src/js/utils/StringUtil.js
+++ b/src/js/utils/StringUtil.js
@@ -3,6 +3,20 @@ import marked from "marked";
 import React from "react";
 /* eslint-enable no-unused-vars */
 
+const markdownRenderer = {
+  rendererReady: false,
+  prepareMarkdownRenderer() {
+    const renderer = new marked.Renderer();
+    renderer.link = function() {
+      const out = marked.Renderer.prototype.link.apply(this, arguments);
+
+      return out.replace(/^<a/, '<a target="_blank"');
+    };
+    marked.setOptions({ renderer });
+    this.rendererReady = true;
+  }
+};
+
 const StringUtil = {
   arrayToJoinedString(array = [], separator = ", ") {
     if (Array.isArray(array)) {
@@ -189,6 +203,10 @@ const StringUtil = {
   parseMarkdown(text) {
     if (!text) {
       return null;
+    }
+
+    if (!markdownRenderer.rendererReady) {
+      markdownRenderer.prepareMarkdownRenderer();
     }
 
     const __html = marked(
--- a/tests/pages/catalog/packages/package/PackageTab-cy.js
+++ b/tests/pages/catalog/packages/package/PackageTab-cy.js
@@ -1,11 +1,9 @@
 describe("Package Detail Tab", function() {
   beforeEach(function() {
-    cy
-      .configureCluster({
-        mesos: "1-task-healthy",
-        universePackages: true
-      })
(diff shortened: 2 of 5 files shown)


src/js/utils/StringUtil.js
```diff
@@ -191,6 +191,14 @@ const StringUtil = {
       return null;
     }
 
+    const renderer = new marked.Renderer();
```
DanielMSchmidt: Would you mind adding some test cases for this? maybe something that could break it like

```md
[foo](http://a.com/<a-mean-uri)
```

natmegs: What would be the ideal behaviour in that case?

natmegs:
@DanielMSchmidt Added some tests for StringUtil#parseMarkdown if you don't mind taking a look! (@GeorgiSTodorov too)

src/js/utils/StringUtil.js
tnguyen0: Nit: I wonder if the changes here and `marked.setOptions` need to done every time or can it be done once on import.

tests/pages/catalog/packages/package/PackageTab-cy.js
```diff
@@ -44,53 +41,57 @@ describe("Package Detail Tab", function() {
         );
     });
 
+    it("contains _blank target for links in description", function() {
+      cy.get(".pod.pod-shorter.flush-top.flush-right.flush-left")
```
tnguyen0: Nit: Do you think we can use get on something with more meaning? I'm not sure if `contains("Description")` would work.

src/js/utils/StringUtil.js
```diff
@@ -191,6 +191,14 @@ const StringUtil = {
...
+    renderer.link = function() {
+      var out = marked.Renderer.prototype.link.apply(this, arguments);
```
brandonc: :neckbeard: could be `const`

DanielMSchmidt:
Leaving this unmerged to give @natmegs a chance to react to @brandonc and @tnguyen0 comment :) Would be fine with merging it like this 🤷‍♂️

natmegs:
@tnguyen0 @brandonc @DanielMSchmidt lots of great feedback on this one, thanks! I've changed the integration test to look for an `h2` containing "Description" to check the description links, and I've added a `markdownRenderer` object to the `StringUtil` to wrap the `marked` functionality. I think it would be great to have this in a separate MarkdownUtil, but I wonder if it would be worth doing that refactor in a separate ticket? Once this one is approved I will add these changes to the backport PRs as well

PR #3267, opened 2018-09-11
Add link to data-layer core document

Fabs:
Just a link to the google docs for the data layer so it is not lost in space.

### DIFF
--- a/packages/data-service/README.md
+++ b/packages/data-service/README.md
@@ -1,5 +1,9 @@
 This package is to be extracted to dcos-ui/data-service.
 
+# Design Document
+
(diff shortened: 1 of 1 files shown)


packages/data-service/README.md
```diff
@@ -1,5 +1,9 @@
 This package is to be extracted to dcos-ui/data-service.
 
+# Design Document
+
+There is a [document who outlines](https://docs.google.com/document/d/1S5754HMrUu0WCTUPGFdZ7NM_w6otn0clb1bLm75wE_U) the decisions that lead to this implementation.
```
tnguyen0: > ✏️ There is a document that outlines the decisions that led to this implementation. 😇

Fabs: Haha, that is embarrassing. I usually do those things with Grammarly, this time I used emacs, and no spellchecking. Thanks tommy.

PR #3268, opened 2018-09-11
DCOS-40861: right-align "Tasks" column header

mperrotti:
Use the updated ui-kit components to right-align the Tasks column header

Closes: DCOS-40861

## Testing
Navigate to the Nodes page
The text of the "Tasks" header should be right-aligned with the text of the other column cells

## Trade-offs
None

## Dependencies
Nothing

## Screenshots
Before:
screen shot 2018-09-11 at 10 05 01 am

After:
screen shot 2018-09-11 at 9 59 11 am


natmegs:
Looks like snapshots [here](https://github.com/dcos/dcos-ui/blob/45d570e01538c884bd615e19282d8fb9d7abcb50/plugins/nodes/src/js/columns/__tests__/__snapshots__/NodesTableHealthColumn-test.js.snap#L5) just need to have `className="css-19m5nce"` replaced with `className="css-1l7ir3d"`

GeorgiSTodorov:
I agree with Natalie and Tommy.

brandonc:
I added a sync discussion about the lockfile being out of date with package.json. I think we should omit it from this PR

PR #3274, opened 2018-09-12
DCOS-41769: Clipboard icon background fix / React warning

brandonc:
1. We were often passing in port values as numbers, which were triggering PropType warnings
2. I added a style fix for something that may or may be intentional. @lilysamimi can you confirm that this was an unintentional background color or perhaps a padding issue?

## Testing

1. Add nginx service
2. Visit service detail, Endpoints tab
3. No warning, no floating background

## Trade-offs

I could have alternatively loosened the proptype requirements for EndpointClipboardTrigger to accept numbers/strings.

## Screenshots

Warnings visible here

screen shot 2018-09-12 at 5 19 34 pm

Look closely and you can see the floating shadow/background

1.12
screen shot 2018-09-12 at 4 37 22 pm

1.11
screen shot 2018-09-12 at 5 02 24 pm

And here is the after: (tooltip not changed)

screen shot 2018-09-12 at 4 35 28 pm

### DIFF
--- a/plugins/services/src/js/containers/service-connection/ServiceConnectionEndpointList.js
+++ b/plugins/services/src/js/containers/service-connection/ServiceConnectionEndpointList.js
@@ -61,7 +61,9 @@ class ServiceConnectionEndpointList extends React.Component {
 
     if (hostPortValue) {
       return (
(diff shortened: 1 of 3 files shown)


brandonc:
Closing in favor of non-fork PR #3283

PR #3287, opened 2018-09-13
DCOS-22126: Don't validate forms until submit

brandonc:
## Background

Form controllers that are affected by this change are FramworkConfiguration and CreateServiceModal. They now need to trigger validation after submitting and not at all times.

Unfortunately, these two forms have different lifecycles with respect to real time validation:

- [x] CreateServiceModal performs validation during the render phase with calls to `getAllErrors` that validate the serviceSpec state. This same method is used to short-circuit the button handler.

- [x] FrameworkConfiguration uses a standard `react-jsonschema-form` and depends on `liveValidation={true}`. The existence of form errors that are derived whenever the form changes (when live validation is on) determines whether the Review & Run button is enabled.  This form is not wired up for submission and therefore react-jsonschema-form does not ever validate when liveValidation is turned off. The approach allows the modal's owned Review & Run button to trigger the form submit externally, as illustrated here https://jsfiddle.net/spacebaboon/g5a1re63/

## Testing

CreateServiceModal:

1. Create a single container service
2. clear Service ID
3. observe no validation errors
4. click Review & Run
5. observe validation errors

FrameworkConfiguration:

1. Create a cassandra service
2. clear name
3. observe no validation errors
4. click Review & Run
5. observe validation errors

## Trade-offs

This change introduces some damage and workarounds in FrameworkConfigurationForm due to the limitations of react-jsonschema-form

## Observations

1. On FrameworkConfiguration, the JSONEditor does not display errors until Review & Run is clicked. This is probably not desirable.

2. On CreateServiceModal, we don't currently display the number of form validation issues as a badge next to each tab.

## Dependencies

None

### DIFF
--- a/plugins/services/src/js/components/modals/CreateServiceModal.js
+++ b/plugins/services/src/js/components/modals/CreateServiceModal.js
@@ -370,7 +370,17 @@ class CreateServiceModal extends Component {
   }
 
   handleServiceChange(newService) {
-    this.setState({ serviceSpec: newService, hasChangesApplied: true });
+    // If there were previous error messages visible it's better to revalidate
+    // on each change going forward
+    const formErrors = this.state.submitFailed
+      ? this.validateServiceSpec(newService)
+      : [];
+
+    this.setState({
+      serviceSpec: newService,
+      hasChangesApplied: true,
+      formErrors
+    });
   }
 
   handleServiceErrorsChange(errors) {
@@ -451,15 +461,17 @@ class CreateServiceModal extends Component {
   }
 
   handleServiceReview() {
-    const errors = this.getFormErrors();
-    if (errors.filter(error => !error.isPermissive).length === 0) {
+    if (!this.anyCriticalFormErrors()) {
       this.setState({
         apiErrors: [],
+        formErrors: [],
         serviceReviewActive: true
       });
     } else {
       this.setState({
-        showAllErrors: true
+        showAllErrors: true,
+        submitFailed: true,
+        formErrors: this.validateServiceSpec(this.state.serviceSpec)
       });
     }
   }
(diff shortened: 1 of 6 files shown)


GeorgiSTodorov:
For me cassandra still displays error as soon as the name is deleted.

plugins/services/src/js/components/modals/CreateServiceModal.js
```diff
@@ -370,7 +370,19 @@ class CreateServiceModal extends Component {
   }
 
   handleServiceChange(newService) {
-    this.setState({ serviceSpec: newService, hasChangesApplied: true });
+    let formErrors = [];
```
DanielMSchmidt: 🎨 Would love to see this as a ternary rather than using `let`

src/js/components/FrameworkConfigurationForm.js
```diff
@@ -198,16 +199,58 @@ export default class FrameworkConfigurationForm extends Component {
...
+      // When liveValidate is true, errorSchema is available for each change
+      this.probeErrorsSchemaForm();
+    }
+  }
+
+  handleFormError() {
+    this.probeErrorsSchemaForm();
+  }
+
+  /* Unconventional code notes
```
DanielMSchmidt: 🎨 I would personally search this kind of message in the git commits rather then here as it might be prone to becoming stale

brandonc: OK, good to know. I will squash this out. It is already a commit message of 
db14226.

brandonc:
@GeorgiSTodorov that worries me that I did not find all the forms... That was not my experience when installing from the catalog. Were you editing the config in some other way?

screen shot 2018-09-14 at 11 00 14 am

GeorgiSTodorov:
@brandonc, no.

peek 2018-09-17 10-16

brandonc:
Feedback addressed-- @GeorgiSTodorov sorry I can't seem to find the issue can we do a zoom tomorrow after sync?

GeorgiSTodorov:
@brandonc We have another meeting after sync today. If it works for everyone else then it might be a problem on my end, so no worries.

leemunroe:
@brandonc Nice work! Looks good on my end.

PR #3291, opened 2018-09-14
DCOS-41922: fix framework form to work with all jsons

Poltergeist:
This fixes an issue where the user in some cases with a invalid json was not able to use the form
but rather was limited to the json editor.

Closes DCOS-41922

## Testing

For testing this, install a package from the catalog which does contain fields which are not present in the JSON Schema.

For convenience this is a configuration for marathon-lb:
```JSON
{
    "service": {
        "secretName": "mlb-secret",
        "principal": "mlb-principal",
        "user": "nobody"
    },
    "marathon-lb": {
          "container-syslogd": true,
          "sysctl-params": "net.ipv4.tcp_tw_reuse=1 net.ipv4.tcp_fin_timeout=30 net.ipv4.tcp_max_syn_backlog=10240 net.ipv4.tcp_max_tw_buckets=400000 net.core.somaxconn=10000",
          "haproxy_global_default_options": "redispatch,http-server-close,dontlognull,httplog,dontlog-normal"
    }
}
```

put it into `mlb-options.json` and then use the cli: `dcos package install --options=mlb-options.json marathon-lb --yes` 

This will install a service. Go to the detail package of the service and click edit. You should see the form and be able to change values. Your custom setting or `service` in the configuration from above should persist. Also, the form should not break.

## Trade-offs

In fact, this is reducing trade-offs which we had before.

## Screenshots

image


DanielMSchmidt:
@Poltergeist It seems like the system tests are failing for a reason, please take a look

nLight:
@Poltergeist too many integration tests fail hints to a legit problem

src/js/components/FrameworkConfigurationForm.js
```diff
@@ -163,6 +148,9 @@ export default class FrameworkConfigurationForm extends Component {
   }
 
   writeErrors(formData, schema, isRequired, errors) {
+    if (schema != null) {
```
Poltergeist: This should be `schema == null`

PR #3293, opened 2018-09-17
DCOS-19300: make the JSON editor wider

mperrotti:
We get complaints that the width of the JSON editor makes it hard to use, especially for catalog packages

Closes DCOS-19300

This is a good start, but I've filed a [follow-up issue](https://jira.mesosphere.com/browse/DCOS-42072) to make the editor resizable, per @leemunroe's suggestion

## Testing
1. Go to Catalog and select a framework
2. Tap "Review & Run"
3. Toggle the JSON editor using the toggle switch in the modal header

The JSON editor should be a comfortable width to work with, and the form on the left should still be scannable

Note: The JSON editor also appears when adding a new service (go to Services tab and tap the "+" icon in the upper right). I checked it in both places before submitting this PR

## Trade-offs
None

## Dependencies
None

## Screenshots
Before:
screen shot 2018-09-17 at 1 19 22 pm

After:
screen shot 2018-09-17 at 1 18 44 pm

### DIFF
--- a/tests/pages/services/ServiceFormModal-cy.js
+++ b/tests/pages/services/ServiceFormModal-cy.js
@@ -111,18 +111,15 @@ describe("Service Form Modal", function() {
           openServiceForm();
 
           // Switch to JSON
-          cy
-            .get(".modal-full-screen-actions label")
+          cy.get(".modal-full-screen-actions label")
             .contains("JSON Editor")
             .click();
 
-          cy
-            .get(".ace_text-input")
+          cy.get(".ace_text-input")
             .focus()
             .type("{selectall}{backspace}", { force: true });
 
-          cy
-            .get(".message")
+          cy.get(".message")
             .contains("Unexpected end of JSON input")
             .should("be.visible");
         });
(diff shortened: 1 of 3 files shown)


leemunroe:
@mperrotti Big improvement!

On testing there are some edge cases where form elements aren't as responsive as we probably expected e.g. if you configure Elastic there is a YAML input and form labels that bleed into the JSON editor.

image
image

Looking at the markup, it is really unclean and lacking some grid classes that I'd expect to help responsiveness e.g. there are form-groups and fieldsets inside other form-groups and fieldsets

image

Your call whether you want to fix it here or create a followup issue. Not a blocker IMO.

leemunroe:
@mperrotti Playing around with it, removing the 2x `<fieldset>`s fixed it visually. Not sure if they are needed for something else.

image
image

nLight:
Heads up! It's not only for the frameworks form. Services/Pods Form is also affected. 
As the issue mentioned by @leemunroe is reproducible on master and should be fixed separately I decided not to block the PR. @mperrotti please followup on the mentioned issue.

nLight:
Valid test failure `AssertionError: expected 720 to equal 500` https://github.com/dcos/dcos-ui/blob/master/tests/pages/services/ServiceFormModal-cy.js#L541

tests/pages/services/ServiceFormModal-cy.js
```diff
@@ -538,7 +531,9 @@ describe("Service Form Modal", function() {
         cy.get(".modal-full-screen-side-panel.is-visible").then(function(
           $jsonEditor
         ) {
-          expect($jsonEditor.width()).to.equal(500);
+          expect($jsonEditor.width()).to.equal(
+            Cypress.config("viewportWidth") / 2
```
mperrotti: This is the only real change in this file. The rest is from my text editor running Prettier on save

PR #3294, opened 2018-09-18
DCOS-42056: [1.12] Invalid labels have no error and are simply ignored

GeorgiSTodorov:
Added a check to see if a label with value and no key is created.

Closes https://jira.mesosphere.com/browse/DCOS-42056

## Testing
Click on Services Tab
Click on Single Container
Click on Environment
Click on `Add Label`

## Trade-offs
I added a separate commit with a css change that fixes the jumping : .
I also created a separate branch for some minor Environment Variables changes to fix console errors like this one:
screenshot from 2018-09-18 12-34-51

## Dependencies
None

### Before
peek 2018-09-18 10-09

### After
peek 2018-09-18 10-14

### After with css change
peek 2018-09-18 10-17


src/styles/components/form-elements/styles.less
```diff
@@ -62,6 +62,14 @@
     bottom: 0.8rem;
     left: 100%;
     position: absolute;
+
+    &.error {
+      top: 40%;
```
nLight: https://github.com/dcos/dcos-ui/pull/3299/files#r218788984

GeorgiSTodorov:
@brandonc I don't think that preventing the form from being submitted is in the scope of this issue and adding a value with no key doesn't really break the service.

I do agree about the refactoring, we should consider it.

PR #3298, opened 2018-09-19
DCOS-42035: Invalid Environment variables have no error and are simply ignored (1.11 backport)

GeorgiSTodorov:
backport for https://github.com/dcos/dcos-ui/pull/3234 with some minor fixes

Half of https://jira.mesosphere.com/browse/DCOS-42035

## Testing
1. Switch to branch 1.11 and set up node to use version 8.9.4 and npm to use version 5.6.0
I recommend using a new copy of the dcos-ui repository.
Setting up node is easy: just use a node version manager (I used https://github.com/creationix/nvm) and run (if you are using nvm) `nvm install 8.9.4` in the directory. Setting up npm should happen automatically in this case, but if it doesn't, it is a bit trickier - navigate to the nvm directory, for example `cd ~/.nvm/versions/node/v8.9.4/lib` and run `npm install npm@5.6.0`.
Then run `npm install`.
2. Test the changes
Click on Services Tab
Click on Single Container
Click on Environment
Click on `Add Environment Variable`

## Trade-offs
None

## Dependencies
None

### Before
peek 2018-09-18 16-02

### After
peek 2018-09-18 16-03

### After with css change
peek 2018-09-19 12-12

### DIFF
--- a/plugins/services/src/js/components/forms/EnvironmentFormSection.js
+++ b/plugins/services/src/js/components/forms/EnvironmentFormSection.js
@@ -55,17 +55,26 @@ class EnvironmentFormSection extends Component {
           );
         }
 
+        const isValueWithoutKey = Boolean(!env.key && env.value);
+
         return (
           <FormRow key={key}>
-            <FormGroup className="column-6" required={false}>
+            <FormGroup
+              className="column-6"
+              required={false}
+              showError={isValueWithoutKey}
+            >
               {keyLabel}
               <FieldAutofocus>
                 <FieldInput
                   name={`env.${key}.key`}
                   type="text"
-                  value={env.key}
+                  value={env.key || ""}
                 />
               </FieldAutofocus>
+              <FieldError>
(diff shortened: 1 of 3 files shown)


Fabs:
Hey @GeorgiSTodorov, have you checked this error message with UX? 
@lilysamimi, can you check the message? 

Why not something like: "The key value cannot be empty" as an error message?

GeorgiSTodorov:
@Fabs It's the message from https://github.com/dcos/dcos-ui/pull/3234, this is just a backport.

lilysamimi:
I like @Fabs suggestion!

GeorgiSTodorov:
@lilysamimi Okay, since the code for 1.12 (and for 1.10 as well) has already been merged, I will create a new jira for this. Do you think it should show the same message for labels and secrets as well?

GeorgiSTodorov:
@lilysamimi I think that the phrase "key value" is a bit confusing. What about "The key cannot be empty."?

PR #3300, opened 2018-09-19
DCOS-42055: [1.11] Invalid labels have no error and are simply ignored

GeorgiSTodorov:
Show error when a label has a value but no key

Closes https://jira.mesosphere.com/browse/DCOS-42055

## Testing
1. Switch to branch 1.11 and set up node to use version 8.9.4 and npm to use version 5.6.0
I recommend using a new copy of the dcos-ui repository.
Setting up node is easy: just use a node version manager (I used https://github.com/creationix/nvm) and run (if you are using nvm) `nvm install 8.9.4` in the directory. Setting up npm should happen automatically in this case, but if it doesn't, it is a bit trickier - navigate to the nvm directory, for example `cd ~/.nvm/versions/node/v8.9.4/lib` and run `npm install npm@5.6.0`.
2. Test the changes
Click on Services Tab
Click on Single Container
Click on Environment
Click on `Add Label`

## Trade-offs
None

## Dependencies
None

### Before
peek 2018-09-19 16-20

### After
peek 2018-09-19 16-21

### DIFF
--- a/plugins/services/src/js/components/forms/EnvironmentFormSection.js
+++ b/plugins/services/src/js/components/forms/EnvironmentFormSection.js
@@ -121,17 +121,20 @@ class EnvironmentFormSection extends Component {
         );
       }
 
+      const isValueWithoutKey = Boolean(!label.key && label.value);
+
       return (
         <FormRow key={key}>
-          <FormGroup className="column-6">
+          <FormGroup className="column-6" showError={isValueWithoutKey}>
             {keyLabel}
             <FieldAutofocus>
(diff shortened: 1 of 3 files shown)


GeorgiSTodorov:
@Fabs This is not a backport, but I included a similar message to the one in https://github.com/dcos/dcos-ui/pull/3234 for consistency

@brandonc thanks again for your input, I don't think that preventing the form from submission is needed here, but if we as a team decide that it is, we can add a new jira for that.

PR #1315, opened 2016-11-01
Fixes integration tests

rcorral:

PR #3302, opened 2018-09-19
Discussion: Is this vendor.js bundle useful?

brandonc:
Why are we bundling and loading all the package dependencies separately? It seems to me that we only need to load what we import through index.js. Our app size may be 2.5MB too large?

## Screenshots

Before (Production build)
screen shot 2018-09-19 at 12 13 22 pm

Before (Bundle Analysis)
screen shot 2018-09-19 at 12 18 00 pm

After
screen shot 2018-09-19 at 12 09 00 pm

### DIFF
--- a/webpack/webpack.dev.js
+++ b/webpack/webpack.dev.js
@@ -1,12 +1,7 @@
 const merge = require("webpack-merge");
 
-const packageInfo = require("../package");
 const common = require("./webpack.config.js");
 const SpeedMeasurePlugin = require("speed-measure-webpack-plugin");
-
-const dependencies = Object.assign({}, packageInfo.dependencies);
-delete dependencies["canvas-ui"];
-delete dependencies["cnvs"];
 
 const maybeProfile = process.env.PROFILE_WEBPACK
   ? new SpeedMeasurePlugin().wrap
@@ -16,8 +11,7 @@ module.exports = maybeProfile(
   merge(common, {
     mode: "development",
     entry: {
-      index: "./src/js/index.js",
-      vendor: Object.keys(dependencies)
+      index: "./src/js/index.js"
     },
     devtool: "cheap-module-eval-source-map"
   })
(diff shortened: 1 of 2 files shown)


nLight:
I wonder if that went broken during the webpack update. We could test previous releases to make sure. I was living under impression that vendor.js helps with incremental builds :)

nLight:
That's when it was introduced: https://github.com/dcos/dcos-ui/pull/1315/commits/131746ab7aed970c12ee1b95d0c20cb1cc31f264

It looks like we're missing `CommonsChunkPlugin` from the configuration. Could that be the reason why we duplicate stuff in vendor?

brandonc:
Adding CommonsChunkPlugin was actually my first approach. It reduces the size of both vendor.js and index.js, and works great. We get similar package size advantages but add two requests compared to removing it altogether.

Trouble is, I don't think there is an advantage to having a separate vendor.js and using it makes future code splitting less straightforward. I don't think having a separate vendor.js increases webpack rebuild performance.

brandonc:
I'll look into it

brandonc:
@nLight OK I looked into it and I think we're ready to make a decision.

Today in master branch the DOMContentLoaded event starts between 1s ~ 1.3s after load. There is a lot of time spent scripting with so much duplicated code.

Using splitChunks (webpack 4 replacement for common chunks) with an aggressive chunking configuration resulted in roughly the same improvement as this PR. The DOMContentLoaded event happened around the 650ms mark in both cases. The reason why we didn't see any improvement in parallelizing is because we don't have any async split points in our code-- all our pages depend on both entries being loaded.

You can see here that all chunks are evaluated serially anyway:

screen shot 2018-10-11 at 2 37 26 pm

My recommendation is to merge this version because it gives more flexibility with how we want to do code splitting in the future. For example, we could work to code split along major components and not have to worry about upsetting an existing config using something like this https://github.com/jamiebuilds/react-loadable

nLight:
Sold

nLight:
Thanks for taking time and making research

PR #3304, opened 2018-09-19
DCOS_OSS-1961: don't swallow container fields added in JSON editor

mperrotti:
Before, 'name' was the only part of the updated value being pushed into newState

Closes DCOS_OSS-1961

## Testing
1. Go to "Services"
2. Add a new service by clicking "+" icon or "Run a Service"
3. Click "Multi/container (Pod)"
4. Toggle the JSON editor open
5. Replace the first object in the `containers` array with:
```
    {
      "name": "container-1",
      "resources": {
        "cpus": 0.1,
        "mem": 128
      },
      "lifecycle": {
        "killGracePeriodSeconds": 600
      },
      "image": {
        "id": "nginx",
        "kind": "DOCKER"
      }
    }
```
6. Focus any field in the form
7. Add a Service ID
8. Tap "Review & Run"
9. Tap "Review & Run" on the confirmation screen
10. When you land back on the services screen, click the service you just added
11. Click the "Edit" button
12. Toggle the JSON editor open

The JSON editor should still contain `"lifecycle": { "killGracePeriodSeconds": 600 }`

## Trade-offs
Not that I know of

## Dependencies
None

## Screenshots
Before:
https://cl.ly/706236dc4395

After:
https://cl.ly/2f569868456c

### DIFF
--- a/plugins/services/src/js/reducers/serviceForm/JSONReducers/Containers.js
+++ b/plugins/services/src/js/reducers/serviceForm/JSONReducers/Containers.js
@@ -392,7 +392,9 @@ module.exports = {
             newState
           );
 
-          newState.push(Object.assign({}, DEFAULT_POD_CONTAINER, { name }));
+          newState.push(
+            Object.assign({}, DEFAULT_POD_CONTAINER, { name }, value)
+          );
           this.cache.push({});
           this.endpoints.push([]);
           break;
--- a/plugins/services/src/js/reducers/serviceForm/JSONReducers/__tests__/Containers-test.js
+++ b/plugins/services/src/js/reducers/serviceForm/JSONReducers/__tests__/Containers-test.js
@@ -239,6 +239,24 @@ describe("Containers", function() {
       });
     });
 
+    describe("container with arbitrary value", function() {
+      it("contains a container with arbitrary field", function() {
+        let batch = new Batch();
+
+        batch = batch.add(
+          new Transaction(["containers"], { someKey: "value" }, ADD_ITEM)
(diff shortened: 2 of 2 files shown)


mperrotti:
This was a 3-point ticket, so this fix feels way too simplistic. Am I missing something?

mperrotti:
@brandonc - it works the same with args in either order, but the order you suggested is more clear, so I changed it

brandonc:
No, nothing was not working. I just thought it was a logical problem since value always contained a name key

plugins/services/src/js/reducers/serviceForm/JSONReducers/Containers.js
```diff
@@ -392,7 +392,9 @@ module.exports = {
             newState
           );
 
-          newState.push(Object.assign({}, DEFAULT_POD_CONTAINER, { name }));
+          newState.push(
+            Object.assign({}, DEFAULT_POD_CONTAINER, value, { name })
```
Poltergeist: Actually, we want to overwrite the name with the name in the value if there is one. Although the parser will generate a transaction for the name anyway. But I would prefer it to be right. 
```JS
Object.assign({}, DEFAULT_POD_CONTAINER, { name }, value)
```

brandonc: 🤦‍♂️

Poltergeist:
> Before, 'name' was the only part of the updated value being pushed into newState
And that name was generated. :-D

PR #3305, opened 2018-09-20
DCOS-42230: Clarify resource totals on pod instances table

natmegs:
Clarify to users what the breakdown of the resources is on the pod instances table.

Added tooltips to the resource cells in the PodInstancesTable. If it is an expandable row (has child containers), it gives a breakdown of the resource summary as "Containers: X, Executor: Y". If is a non-expandable row (child container), it indicates how much of the allocated resource is being used as "Using X/Y".

Closes DCOS-42230.

## Testing

- Run a service from JSON config which can be found in [JIRA ticket comment](https://jira.mesosphere.com/browse/DCOS-17436?focusedCommentId=174971&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-174971). 
- Go to services, click on the service you just started, hover over the resource cells on the parent (expandable) row. They should give breakdown of resources being used by containers and executor resources. 
- Expand the instance row. Hover again over the resource cells in the rows. They should indicate what the container is using compared to what's allocated. (Using this config, when the task is complete, the used resources should be 0).

## Trade-offs

Many. Some background:

Historical instances and pod instances are merged together in the `PodInstancesContainer` (via the `PodUtil` `mergeHistoricalInstanceList`), and the `PodInstancesTable` receives this merged instance list eventually as a prop. If a task in a container has completed, it will be in both the pod instance list and historical instance list. When the containers are merged together, if there are two containers with the same containerId, only one is kept in the merged list. The container from the pod instance list has NO resources listed since it is not using resources. The matching container from the historical instance list HAS resources listed to represent the available (not used) resources for that container. So when we keep only one (the match from the historical instance list), there is a resources property on that container.

This is inconsistent with how non-completed tasks are represented, since the resources property on these objects represents the resources in use (not resources available). This discrepancy makes it difficult to determine when the resources property is indicating resources that are available or resources that are in use by a container.

**1st Approach:**

To alleviate the problem, I added an `activeResources` property to the containers to indicate which resources are in use. If there are NO historical instances (no completed tasks), the `activeResources` will simply be a copy of the regular resources property. If there ARE historical instances, when they are merged, we look at the pod instance match and if there is no resources property on this object, we add an empty `activeResources` object to the container to indicate that there are no resources being used.

**2nd Approach:**

Added `isHistoricalInstance` property to historical resources in the `mergeHistoricalInstanceList` method of `PodUtil`. Then do not use resources from these historical instances in calculations for currently active resources.

**Summary:** having access to the `activeResources` or `isHistoricalInstance` property depends currently on calling `mergeHistoricalInstanceList` (which always happens now in `PodInstancesContainer` but is not enforced at the `PodInstancesTable` component). Adding `activeResources` or `isHistoricalInstance` changes the "shape" of the pod instances list.

~~Also, as of right now a bunch of unit tests are broken~~

### DIFF
--- a/plugins/services/src/js/utils/PodUtil.js
+++ b/plugins/services/src/js/utils/PodUtil.js
@@ -90,14 +90,37 @@ var PodUtil = {
 
       // Filter combined container list to remove potential duplicates
       const containerIds = new Map();
-      combinedContainers = combinedContainers.filter(function({ containerId }) {
-        if (containerId != null && !containerIds.has(containerId)) {
-          containerIds.set(containerId);
+      const nullIdContainers = [];
+      combinedContainers = combinedContainers.filter(function(container) {
+        if (
+          container.containerId != null &&
+          !containerIds.has(container.containerId)
+        ) {
+          containerIds.set(container.containerId);
 
           return true;
         }
 
+        if (container.containerId == null) {
+          nullIdContainers.push(container);
+        }
+
         return false;
+      });
+
+      nullIdContainers.forEach(function(container) {
+        const match = combinedContainers.find(
+          matching => matching.name === container.name
+        );
+        if (match) {
+          match.activeResources = container.resources || {};
+        }
+      });
+
+      combinedContainers.forEach(function(container) {
+        if (!container.activeResources) {
+          container.activeResources = { ...container.resources };
+        }
       });
 
       podInstance.containers = combinedContainers;
@@ -105,6 +128,14 @@ var PodUtil = {
       return memo;
     },
     podInstancesMap);
+
+    if (!historicalInstances.length) {
+      Object.values(combinedInstanceMap).forEach(function(instance) {
+        instance.containers.forEach(function(container) {
+          container.activeResources = { ...container.resources };
+        });
+      });
+    }
 
     // Re-compose PodInstances from plain objects
     const instances = Object.values(combinedInstanceMap).map(function(
--- a/plugins/services/src/js/utils/PodUtil.js
+++ b/plugins/services/src/js/utils/PodUtil.js
@@ -85,12 +85,13 @@ var PodUtil = {
 
       let combinedContainers = [].concat(
         podInstance.containers,
-        historicalInstance.containers
+        historicalInstance.containers.map(container =>
+          Object.assign({}, container, { isHistoricalInstance: true })
+        )
       );
 
       // Filter combined container list to remove potential duplicates
       const containerIds = new Map();
-      const nullIdContainers = [];
       combinedContainers = combinedContainers.filter(function(container) {
         if (
           container.containerId != null &&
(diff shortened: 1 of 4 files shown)


GeorgiSTodorov:
Should the sum always add up? 
screenshot from 2018-09-20 11-53-55

What about non-active tasks?
screenshot from 2018-09-20 11-54-23

I also took a look at the failing tests and it looks like the needed values can't be found for some reason and they receive empty arrays (I am not sure why). I was able to fix one failing test by adding `activeResources: {},` in https://github.com/dcos/dcos-ui/blob/c36795c5bacd564c4761dc7153b87b3e97109b71/plugins/services/src/js/utils/__tests__/PodUtil-test.js#L129

natmegs:
@GeorgiSTodorov The totals will not always be the sum of the child resources - it will be a sum of the running tasks plus the executor resources. Interesting that it says "Containers: 0" in the first screenshot. What was the configuration for that pod?

GeorgiSTodorov:
I used the JSON from the jira, but at the time of the screenshot it was still deploying I think.

plugins/services/src/js/utils/PodUtil.js
```diff
@@ -90,22 +90,53 @@ var PodUtil = {
...
+          container.containerId != null &&
+          !containerIds.has(container.containerId)
+        ) {
+          containerIds.set(container.containerId);
 
           return true;
         }
 
+        if (container.containerId == null) {
+          nullIdContainers.push(container);
```
DanielMSchmidt: 🎨 This seems to be sorting a list into two lists, I would rather like to see side-effect free filters with just a negative predicate if it's really distinct or another predicate.

plugins/services/src/js/utils/PodUtil.js
```diff
@@ -90,22 +90,53 @@ var PodUtil = {
...
+      nullIdContainers.forEach(function(container) {
+        const match = combinedContainers.find(
+          matching => matching.name === container.name
+        );
+        if (match) {
+          match.activeResources = container.resources || {};
+        }
+      });
+
+      combinedContainers.forEach(function(container) {
```
DanielMSchmidt: 🎨 would love to see a `.map` here instead of a side-effect

plugins/services/src/js/utils/PodUtil.js
```diff
@@ -90,22 +90,53 @@ var PodUtil = {
...
+
         return false;
       });
 
+      nullIdContainers.forEach(function(container) {
+        const match = combinedContainers.find(
+          matching => matching.name === container.name
+        );
+        if (match) {
+          match.activeResources = container.resources || {};
```
DanielMSchmidt: 🎨 Would love to see a map here instead of a mutation

plugins/services/src/js/utils/PodUtil.js
```diff
@@ -90,22 +90,53 @@ var PodUtil = {
...
+
       podInstance.containers = combinedContainers;
 
       return memo;
     },
     podInstancesMap);
 
+    if (!historicalInstances.length) {
+      Object.values(combinedInstanceMap).forEach(function(instance) {
+        instance.containers.forEach(function(container) {
```
DanielMSchmidt: 🎨 I would love to see a map here instead of the mutation

tnguyen0:
@leemunroe @lilysamimi

leemunroe:
Are we able to fix the alignment on these columns in this PR while we're touching this code? The third/last row is misaligned for health, cpu, mem

image

natmegs:
Tried a more efficient approach that eliminated a bunch of loops/mutations @DanielMSchmidt (as separate commits for comparison purposes). @leemunroe I'll add a commit to address the spacing to this PR

PR #3307, opened 2018-09-20
Try this branch: Vastly improve webpack build times

brandonc:
## Overview

Uses various tricks to _greatly_ decrease webpack build time. On my 6 core i7 machine, initial `npm start` is about 50% faster, but successive `npm start`s are about 80% faster. Javascript hot reloading continues to work, with "watch" rebuilds taking about 4-5s

1. Turn off ExtractTextPlugin, image-webpack-plugin in local development. This was previously attempted but not actually happening

2. Parallelize our most intensive workers, babel-loader and ts-loader by pipelining files through thread-loader. I'm most interested in making sure this change doesn't break any ts syntax checking.

3. Cache the output of babel-loader (built into the package) and css loaders so that unchanged files don't need to be rebuilt. This will generated a ~16MB node_modules/.cache folder.

## Testing

Try branching off this branch for your next issue instead of master? Not ready to merge this until I make sure it doesn't mess up any workflows.

😎 

screen shot 2018-09-20 at 1 27 59 pm

PR #3310, opened 2018-09-21
Webpack 4 ✨

brandonc:
I had some time last night to upgrade to Webpack 4. It's slightly faster, but doesn't offer any big advantages that I know of right now. If we migrate to babel 7, we can drop ts-loader since babel 7 can transpile typescript now. That could potentially knock 20s off our build time

## Testing

`npm start` or `npm run build-assets`

## Trade-offs

I don't know of any risks.

## Dependencies

This branch is based on #3307


nLight:
Awesome! Last time I tried that there was some major uglify issues I had to manually patch https://github.com/dcos/dcos-ui/pull/3177

Poltergeist:
Can we solve the conflicts in the config files? Otherwise I like this direction.

brandonc:
I think I just need to reopen it after rebasing #3307

PR #3313, opened 2018-09-24
DCOS-42317: skip non-string fields in Universe response

nLight:
The PR adds a type check for the fields skipping all non-string fields.

Relevant slack discussion linked in the ticket: https://jira.mesosphere.com/browse/DCOS-42317

Closes DCOS-42317

## Testing

1. Try installing `grafana` from the Catalog without the patch.
2. See UI went broken
3. With the patch you should be able using the UI normally

## Trade-offs

We do extraction in prod even though we don't need it there.

### DIFF
--- a/plugins/services/src/js/utils/FrameworkUtil.js
+++ b/plugins/services/src/js/utils/FrameworkUtil.js
@@ -51,6 +51,10 @@ const FrameworkUtil = {
     const images = Object.assign({}, packageImages);
 
     Object.keys(images).forEach(image => {
+      if (typeof images[image] !== "string") {
+        return;
+      }
+
       const queryString = images[image].substring(
         images[image].lastIndexOf("?")
       );
--- a/plugins/services/src/js/utils/__tests__/FrameworkUtil-test.js
+++ b/plugins/services/src/js/utils/__tests__/FrameworkUtil-test.js
@@ -122,4 +122,52 @@ describe("FrameworkUtil", function() {
       expect(FrameworkUtil.extractBaseTechVersion()).toEqual("N/A");
     });
   });
+
+  describe("#extractImageUrls", () => {
+    it("extracts icon urls", () => {
+      const input = {
+        "icon-large":
+          "http://localhost/package/resource?url=https://downloads.mesosphere.com/assets/universe/000/grafana-icon-large.png",
+        "icon-medium":
+          "http://localhost/package/resource?url=https://downloads.mesosphere.com/assets/universe/000/grafana-icon-medium.png",
+        "icon-small":
+          "http://localhost/package/resource?url=https://downloads.mesosphere.com/assets/universe/000/grafana-icon-small.png"
+      };
+
+      expect(FrameworkUtil.extractImageUrls(input)).toEqual({
+        "icon-large":
+          "https://downloads.mesosphere.com/assets/universe/000/grafana-icon-large.png",
+        "icon-medium":
+          "https://downloads.mesosphere.com/assets/universe/000/grafana-icon-medium.png",
+        "icon-small":
+          "https://downloads.mesosphere.com/assets/universe/000/grafana-icon-small.png"
+      });
+    });
+
+    it("supports partial response", () => {
+      const input = {
+        "icon-small":
+          "http://localhost/package/resource?url=https://downloads.mesosphere.com/assets/universe/000/grafana-icon-small.png"
+      };
+
+      expect(FrameworkUtil.extractImageUrls(input)).toEqual({
+        "icon-small":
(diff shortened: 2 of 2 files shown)


mperrotti:
Code looks good, and I tested it to make sure:

Before: https://cl.ly/e3211d495a28

After: https://cl.ly/c7266852578e

plugins/services/src/js/utils/FrameworkUtil.js
```diff
@@ -47,17 +47,20 @@ const FrameworkUtil = {
...
-      const queryString = images[image].substring(
-        images[image].lastIndexOf("?")
-      );
-      images[image] = RouterUtil.getQueryStringInUrl(queryString).url;
-    });
+    return variants
+      .filter(variant => packageImages[variant] != null)
+      .reduce((acc, variant) => {
+        const queryString = packageImages[variant].substring(
+          packageImages[variant].lastIndexOf("?")
```
brandonc: Is it possible these urls won't have a querystring with a "url" key? How defensive should we be here?

nLight: Good point, we actually don't extract them on the catalog page, just call through the proxy.

nLight:
Let me double check if we even need extracting those

nLight:
Hey folks! 

@brandonc @mperrotti @natmegs @GeorgiSTodorov Please have a look into the second commit.

Turned out we don't need the transformation at all. The URLs Cosmos provides are sufficient. It's just that they don't work with our dev server. Looks like Cosmos takes the host name from the request, adds `https` to it and sends it back. In our case on DEV environment we are getting `https://localhost:4200/package/resource/...`. But webpack runs on http and switching it to https doesn't help either. So without this trick we wouldn't see icons on the dev environment. 

So our call to make: what would we willing to tolerate: env dependent code or missing icons on dev.

@tnguyen0 please correct me if I'm wrong on the `extractImageUrls` purpose.

GeorgiSTodorov:
Is there a precedent? Do we have environment dependent code somewhere else?

nLight:
@GeorgiSTodorov mostly for logs

GeorgiSTodorov:
I am in favour of doing environment dependent code to escape issues with this in the future and for better testing, but I am interested in hearing other opinions as well.

tnguyen0:
@nLight you are not wrong.

GeorgiSTodorov:
I am leaving this unmerged in case someone is against this change and wants to say something.

plugins/services/src/js/stores/MarathonStore.js
```diff
@@ -471,9 +471,14 @@ class MarathonStore extends GetSetBaseStore {
...
-          CosmosPackagesStore.getPackageImages()[item.labels.DCOS_PACKAGE_NAME]
-        );
+        const images = CosmosPackagesStore.getPackageImages()[
+          item.labels.DCOS_PACKAGE_NAME
+        ];
+
+        item["images"] =
+          process.env.NODE_ENV === "production"
+            ? images
+            : FrameworkUtil.extractImageUrls(images);
```
brandonc: The development version filters out any images keys that aren't `["icon-small", "icon-medium", "icon-large"] but in production it does not.

To resolve, you could do one of these:

- move the NODE_ENV condition to the reducer
- change the util method to transform these 3 keys and leave the other keys
- move the variant filter to here

plugins/services/src/js/utils/FrameworkUtil.js
```diff
@@ -47,17 +47,20 @@ const FrameworkUtil = {
    * @return {Object}               containing urls from the extracted query string
    * if all is not available in given object.
    */
-  extractImageUrls(packageImages) {
-    const images = Object.assign({}, packageImages);
+  extractImageUrls(packageImages = {}) {
```
brandonc: Should this method be moved to a place that highlights the fact it should only be used in development? May be worth noting that MarathonStore is the only caller.

Or perhaps it's better to tolerate the transformation in production.

plugins/services/src/js/stores/MarathonStore.js
nLight: Fair point. I think I will go with just fixing the bug as discussed on the latest Sync.

As for the webpack proxy I don't see how this will solve the `https` issue. We run webpack dev server on `http` thus all `https` requests to localhost will fail. I tried `https: true` option then we gonna have this self-signed certificate warning plus some API requests fail with that option being on.

GeorgiSTodorov:
"Simplicity is the ultimate sophistication."

nLight:
ping @brandonc

PR #3318, opened 2018-09-24
DCOS-42385: chore(npm): Replace npm-shrinkwrap with package-lock.json

brandonc:
## Tradeoffs

It turns out there was very little reason to switch. Both npm-shrinkwrap and package-lock.json are kept up to date when running `npm install --save`

However, this change may make things slightly simpler if the change to fsevents is no longer needed... but at the expense of being different in 1.10, 1.11. We may not want to merge this.

Closes DCOS-42385


GeorgiSTodorov:
I also think that there is little sense in switching. Should we still host a decision meeting, @brandonc ?

PR #3319, opened 2018-09-25
DCOS-42348: Show error page when the log file won't load

DanielMSchmidt:
## Testing

Look at a task and go to Log. Now open the network tab and search for "files/read". Block that URL. Reload. See an error message instead of infinite loading.

## Trade-offs

- Decided to use the very same error page for both errors as the origin of this issue is unclear in both occasions.
- Decided to directly show the error, might lead to blinking if it's first not there and then there, but this is really an upstream bug (if it even exists) and I am willing to do this tradeoff

## Dependencies

None

## Screenshots

Not so interesting here 🤷‍♂️

### DIFF
--- a/plugins/services/src/js/components/MesosLogContainer.js
+++ b/plugins/services/src/js/components/MesosLogContainer.js
@@ -15,7 +15,9 @@ const METHODS_TO_BIND = [
   "handleGoToWorkingDirectory",
   "handleFetchPreviousLog",
   "onMesosLogStoreError",
-  "onMesosLogStoreSuccess"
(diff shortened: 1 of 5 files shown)


nLight:
That `release/1.11-` got you in trouble man :D

GeorgiSTodorov:
screenshot from 2018-09-25 13-35-24
This is the message that I got, is it correct?

nLight:
As it goes for 1.11 I think that was our default message

nLight:
Sorry, hit `Close and comment` 😅

PR #3320, opened 2018-09-25
DCOS-42347: Show error page when the log file won't load

DanielMSchmidt:
## Testing

Look at a task and go to Log. Now open the network tab and search for "files/read". Block that URL. Reload. See an error message instead of infinite loading.

## Trade-offs

- Decided to use the very same error page for both errors as the origin of this issue is unclear in both occasions.
- Decided to directly show the error, might lead to blinking if it's first not there and then there, but this is really an upstream bug (if it even exists) and I am willing to do this tradeoff

## Dependencies

None

## Screenshots

Not so interesting here 🤷‍♂️

### DIFF
--- a/plugins/services/src/js/components/MesosLogContainer.js
+++ b/plugins/services/src/js/components/MesosLogContainer.js
@@ -14,7 +14,9 @@ const METHODS_TO_BIND = [
   "handleGoToWorkingDirectory",
   "handleFetchPreviousLog",
   "onMesosLogStoreError",
-  "onMesosLogStoreSuccess"
+  "onMesosLogStoreSuccess",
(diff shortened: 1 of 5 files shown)


nLight:
run integration tests

natmegs:
Pushed some changes to the TasksTable integration test to make them pass (locally at least). Only notable change was that the error message says "Cannot connect with the server" instead of "Cannot retrieve the requested information." Maybe the default error message has changed between 1.10 and 1.12?

natmegs:
run system tests

nLight:
@natmegs system tests for 1.10 are not required and broken :(

PR #3322, opened 2018-09-25
DCOS-42395: Fix error message for labels

GeorgiSTodorov:
Update error message when a environment label has a value but no key

Closes https://jira.mesosphere.com/browse/DCOS-42395

Relevant discussion: https://github.com/dcos/dcos-ui/pull/3298#issuecomment-422834336

## Testing
Click on Services Tab
Click on Single Container
Click on Environment
Click on `Add Label`
Verify that there is an error message and it says "The key cannot be empty."

## Trade-offs
None

## Dependencies
None

### Before
screenshot from 2018-09-25 15-10-30

### After
screenshot from 2018-09-25 15-07-44

### DIFF
--- a/plugins/services/src/js/components/forms/EnvironmentFormSection.js
+++ b/plugins/services/src/js/components/forms/EnvironmentFormSection.js
@@ -137,7 +137,7 @@ class EnvironmentFormSection extends Component {
                 value={label.key || ""}
               />
             </FieldAutofocus>
-            <FieldError>A label needs to contain at least a key.</FieldError>
+            <FieldError>The key cannot be empty.</FieldError>
             <span className="emphasis form-colon">:</span>
           </FormGroup>
           <FormGroup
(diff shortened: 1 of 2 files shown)


natmegs:
If I add a label with no key, I can still go to Review & Run. Is this error supposed to block the user from moving on to running the service?

brandonc:
@natmegs previous discussion https://github.com/dcos/dcos-ui/pull/3294#issuecomment-423072164

GeorgiSTodorov:
@natmegs , @brandonc and I had a discussion about that here: https://github.com/dcos/dcos-ui/pull/3294#pullrequestreview-157104349 The original jira did not mention preventing the user from submitting the form and I don't think that it makes much sense to, because a label with no key isn't added to the JSON so it isn't invalid, but if we as a team decide that it should, I wouldn't mind.

PR #3309, opened 2018-09-20
DCOS-21232: Better text filtering

brandonc:
https://jira.mesosphere.com/browse/DCOS-21232

StringUtil#filterByString now sorts matches by relevance score. This change affects node health checks and universe packages. Here's how scoring is calculated:

- Search phrase are tokenized and each term is evaluated individually, with multiple scores summed together. So if you type "foo bar baz" a package named "foo baz" would appear before "foo" or "bar"
- Exact term matches are scored very very high
- Otherwise, longer terms appearing closer to the beginning of the string are scored higher than shorter terms closer to the end
- Additionally, universe packages are re-ranked to show certified packages first.

The "string" being searched is the concatenated values of "filterProperties"... For universe packages, that would be the name, description, and tags.

I got some ideas from fuse.js and Sifterjs, but ultimately chose an algorithm that felt best.

## Testing

1. Visit catalog
2. Observe exact matches like "beta-kubernetes" rank before more proximate packages like "kubernetes"
3. Observe search phrase "kaf" has more relevant ranking

1. Visit node details health tab
2. Observe that text filtering still works and ranks according to the rules outlined above

## Trade-offs

It is plausible that future callers of StringUtil#filterByString may not want reordering to take place.

## Screenshots

Before:

screen shot 2018-09-20 at 4 36 54 pm

After:

screen shot 2018-09-20 at 4 34 25 pm

PR #3324, opened 2018-09-25
[1.11] DCOS-44243 Better text filtering

brandonc:
1.11 Backport of #3309

I had added a ts module and had to convert it to js for the backport. Is this normal? Seems like an unfortunate fork.


brandonc:
There's not much choice beyond adding typescript to 1.11. I can revert this later if the conversion is a problem.

PR #3328, opened 2018-10-01
 DCOS-40640: reuse Cypress cache

nLight:
With this PR we will be re-using cached Cypress installation that should reduce the build time.

Closes DCOS-40640

## Testing

CI should pass. 
Go inside the build and make sure we're not installing cypress for every set of tests.

This is what should not be present in the logs
screenshot 2018-10-01 at 18 17 48

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -32,7 +32,6 @@ pipeline {
     stage("Build") {
       steps {
         sh "npm --unsafe-perm install"
-        sh "npx cypress install"
         sh "npm run build"
       }
     }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -32,6 +32,7 @@ pipeline {
     stage("Build") {
       steps {
         sh "npm --unsafe-perm install"
+        sh "npx cypress install"
         sh "npm run build"
       }
     }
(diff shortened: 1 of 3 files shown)


Jenkinsfile
```diff
@@ -32,6 +32,7 @@ pipeline {
     stage("Build") {
       steps {
         sh "npm --unsafe-perm install"
```
brandonc: ❔I thought `cypress install` was already happening postinstall?

system-tests/_scripts/sandbox-setup.sh
```diff
@@ -4,4 +4,6 @@ SCRIPT=`realpath $0`
 SCRIPT_PATH=`dirname $SCRIPT`
 CYPRESS=`realpath "${SCRIPT_PATH}/../../node_modules/cypress/bin/cypress"`
 
-ln -s $CYPRESS "$TMPDIR/cypress"
\ No newline at end of file
+cp -R "/root/.cache" "$TMPDIR/../"
```
brandonc: ❔ What manages `/root/.cache`? Jenkins? Does this script work in local dev environment?

GeorgiSTodorov:
@nLight I am leaving this unmerged in case you want to address Brandon's comments.

Jenkinsfile
nLight: Fair, let's remove it. Thanks for pointing this!

system-tests/_scripts/sandbox-setup.sh
nLight: The script is supposed to be ran in docker even in local dev environment. So the user is `root`.

DanielMSchmidt: I would normally say let's just make it configurable, but the sandbox setup within system tests is pretty specific to being run on docker

PR #3329, opened 2018-10-02
lingui integration + a codemod for extracting translatable text

brandonc:
This is my best first attempt at building a codemod to wrap translatable JSX phrases in a lingui Trans macro.

You can see the unadulterated results of the codemod in the plugins-private branch here https://github.com/mesosphere/dcos-ui-plugins-private/compare/brandonc/DCOS-41831-l10n-wrapper?expand=1

## How it works

Look at every JSXElement, and if it meets the following criteria, wrap its children in a `<Trans>` component and add an import for lingui/macro to that file.

1. Must have children, and the children must not be a single expression container (like `<p>{this.props.children}</p>`
2. The children must all be literals, {} expressions, or whitelisted elements (PHRASING_CONTENT) containing only literals and {} expressions
3. The element must not already be wrapped in a higher level &lt;Trans&gt; component

## Examples

✅ From `src/js/schemas/job-schema/Schedule.js`
Illustrates algorithm laziness
screen shot 2018-10-01 at 10 20 09 pm

🚫 From `src/js/pages/NotFoundPage.js`
Illustrates an unfamiliar (but innocuous) component that fragments a phrase
screen shot 2018-10-01 at 10 22 41 pm

🚫 From `src/js/components/Sidebar.js`
Illustrates false positive
screen shot 2018-10-01 at 10 22 13 pm

## Testing

1. Install jscodeshift globally `npm install -g jscodeshift`
2. `jscodeshift --ignore-pattern "__tests__" -t scripts/codemod/wrap_jsx_translatables_in_macro.js src/`

## Trade-offs

The macro is used instead of our own L10n component because it is crucial to the `extract` CLI, which automates the creation of json files for the purposes of translation and analysis.

On the other hand, using the live macro requires that we complete the integration before merging.

## Notes

In React 16, a string is a valid react element, and lingui is set up to try and render those without markup by default. In React 15, an invariant violation is thrown with the default lingui config because a string is not a valid React element. So I had to add a config to `<I18nProvider>` in index.js like `defaultRender="span"`

Adding this markup and `<Trans>` for every translatable label may wreak havoc on our unit test assertions as you can see in the changes to TasksChart-test.js


nLight:
How many false positives does it give?

GeorgiSTodorov:
Another project that I worked on had all text messages and resources in constants. This will introduce some boilerplate, but make future translations easier. Has this idea been discussed here?

brandonc:
@GeorgiSTodorov Without the strings being in JSX I don't think there's a way to ensure they need to be in the translation file. They just look like any other string literal. We will have to add manual translation to those files. There are also many modifications this codemod made to plugins-private in the linked branch.

brandonc:
@nLight 8 in dcos-ui and 7 in plugins-private

GeorgiSTodorov:
I just noticed that there is a jira for this: https://jira.mesosphere.com/browse/DCOS-42518

nLight:
@brandonc let's move forward with the solution and correct all the edge cases manually

PR #3330, opened 2018-10-02
DCOS-41250: implement basic soak tests for DC/OS UI

DanielMSchmidt:
## Testing

That's pretty hard, you can run the script against a cluster by setting the environment variables (according to the readme, therefore I won't give you a command). 

## Trade-offs

- Decided to use puppeteer (which has a high level of abstraction) to move fast. We might need to move to sth that supports multiple browsers later, but first starting to make experiences with soaking in general is a good first step.
- Decided to manually log in, partly because I think our test should fail if the login fails, also because I think adding some python utils to handle token generation won't help wiht maintainability

## Dependencies

This PR needs to be merged before we can deploy it to soak.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -29,65 +29,76 @@ pipeline {
       }
     }
 
-    stage("Build") {
-      steps {
-        sh "npm --unsafe-perm install"
-        sh "npm run build"
-      }
-    }
+    // stage("Build") {
+    //   steps {
+    //     sh "npm --unsafe-perm install"
+    //     sh "npm run build"
+    //   }
+    // }
 
-    stage("Tests") {
-      parallel {
-        stage("Integration Test") {
-          steps {
-            sh "npm run integration-tests"
-          }
+    // stage("Tests") {
+    //   parallel {
+    //     stage("Integration Test") {
+    //       steps {
+    //         sh "npm run integration-tests"
+    //       }
 
-          post {
-            always {
-              archiveArtifacts "cypress/**/*"
-              junit "cypress/results.xml"
-            }
-          }
-        }
+    //       post {
+    //         always {
+    //           archiveArtifacts "cypress/**/*"
+    //           junit "cypress/results.xml"
+    //         }
+    //       }
+    //     }
 
-        stage("System Test") {
-          steps {
-            withCredentials([
-              [
-                $class: "AmazonWebServicesCredentialsBinding",
-                credentialsId: "f40eebe0-f9aa-4336-b460-b2c4d7876fde",
-                accessKeyVariable: "AWS_ACCESS_KEY_ID",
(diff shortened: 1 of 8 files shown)


Jenkinsfile
```diff
@@ -92,6 +92,22 @@ pipeline {
...
+    stage("Soak Docker Release") {
+      agent {
+        label "mesos-ubuntu"
+      }
+      steps {
+        withCredentials([
+          string(credentialsId: '7bdd2775-2911-41ba-918f-59c8ae52326d', variable: 'DOCKER_HUB_USERNAME'),
+          string(credentialsId: '42f2e3fb-3f4f-47b2-a128-10ac6d0f6825', variable: 'DOCKER_HUB_PASSWORD')
+        ]) {
+          sh "docker --help"
```
nLight: we don't really need this, do we?

soak-tests/soak112s/deploy.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+
+set -ex
+cd "${0%/*}" # to the folder containing this script.
+
+IMAGE_IDENTIFIER='mesosphere/dcos-ui-soak'
```
nLight: Shall we add label to distinguish between soak versions? I.e. 'mesosphere/dcos-ui-soak:112s'?

nLight:
Just crossed my mind. Having it in the OSS repo would mean we don't soak EE plugins, thus it won't be able to login, right?

soak-tests/soak112s/deploy.sh
DanielMSchmidt: Good point, will do

Jenkinsfile
DanielMSchmidt: We do as long as I debug why we can not push to dockerhub :D

DanielMSchmidt:
@nLight you are right, we should move this 🤔 I'll make it work and move it then

nLight:
Even with the special Jenkins node it doesn't work?

DanielMSchmidt:
@nLight Nope, couldn't get it running on any node so far. @juliangieseke mentioned that Docker in Docker could also referr to being able to use a different dockerfile on these agents and not promise a docker deamon to be available. We briefly thought about installing docker on the mac mini and calling it a day, but this seems like a non-ideal solution.

nLight:
Shall we move the code into EE repo and scope out the CI part? 
So that before another SOAK test we would just manually build the image? (< Should be documented)

nLight:
I bet we could just extract this step and setup a separate job that would run in `docker-in-docker` image. But I'd say let's scope that out

DanielMSchmidt:
@nLight I tried out 4 different jenkins labels with docker in docker and neither of them worked. I can move it over and we scope it out, seems reasonable to me.

PR #3332, opened 2018-10-04
DCOS-43055: Wrap text resources in app (src/js/pages)

TattdCodeMonkey:
## Testing

These changes should result in no visible changes to the UI. There may be some additional `<span>` tags rendered from using the `<Trans>` macro. But we would also be able to remove that once we upgrade to React 16 as string's will become valid return results from a component.

## Trade-offs

There was at least one place where the current code could be simpler and allow easier translation (see comment), but I left the current code in place to not increase the scope of these changes.

### DIFF
--- a/src/js/pages/catalog/PackagesTab.js
+++ b/src/js/pages/catalog/PackagesTab.js
@@ -1,3 +1,4 @@
+import { Trans } from "@lingui/macro";
 import classNames from "classnames";
 import mixin from "reactjs-mixin";
 import { Hooks } from "PluginSDK";
@@ -30,7 +31,9 @@ const PackagesBreadcrumbs = () => {
   const crumbs = [
     <Breadcrumb key={0} title="Catalog">
       <BreadcrumbTextContent>
-        <Link to="/catalog/packages">Catalog</Link>
+        <Link to="/catalog/packages">
+          <Trans render="span">Catalog</Trans>
+        </Link>
       </BreadcrumbTextContent>
     </Breadcrumb>
   ];
@@ -41,8 +44,10 @@ const PackagesBreadcrumbs = () => {
 const PackagesEmptyState = () => {
   return (
     <AlertPanel>
-      <AlertPanelHeader>No package repositories</AlertPanelHeader>
-      <p className="tall">
+      <AlertPanelHeader>
+        <Trans render="span">No package repositories</Trans>
+      </AlertPanelHeader>
+      <Trans render="p" className="tall">
         You need at least one package repository with some packages to be able
         to install packages. For more{" "}
         <a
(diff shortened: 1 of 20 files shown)


src/js/pages/catalog/PackagesTab.js
```diff
@@ -207,10 +216,10 @@ class PackagesTab extends mixin(StoreMixin) {
...
 
     let subtitle = (
-      <p className="tall flush-top">
+      <Trans render="p" className="tall flush-top">
         Community packages are unverified and unreviewed content from the
         community.
-      </p>
+      </Trans>
     );
     let title = "Community";
```
TattdCodeMonkey: We could have Community grid translated similarly to the Certified grid above this, if we separated the rendering of the search results to its own function instead of using  `getCommunityPackagesGrid()` to render community packages and search results. I didn't do the refactor now since it would of increased scope of this change.

src/js/pages/catalog/PackagesTab.js
```diff
@@ -226,12 +235,12 @@ class PackagesTab extends mixin(StoreMixin) {
         }" `;
 
         return (
-          <div className="clearfix">
+          <Trans render="div" className="clearfix">
             {noResults}
```
TattdCodeMonkey: Should we moved the text in this variable on line 233 to inside the this competent instead? it would make translating easier. 
instead of `{noResults} (<0>view all</0>)` the key would be 
`No results were found for your search: "{searchString}" (<0>view all</0>)`

GeorgiSTodorov:
I figured out the problem. Here: https://github.com/dcos/dcos-ui/pull/3332/files#diff-ef04d5bf3205041917c17a93e5389d86R317 the integration test expects the span to be disabled.

GeorgiSTodorov:
The easiest solution I found so far was to update https://github.com/dcos/dcos-ui/blob/master/tests/pages/catalog/packages/PackagesTab-cy.js#L176 to be either
```
      cy.get(".button.button-primary")
        .should("be.disabled");
```
or
```
      cy.get(".button.button-primary")
        .contains("Review & Run")
        .parent()
        .should("be.disabled");
```

GeorgiSTodorov:
And another thing I noticed while starting a service:
screenshot from 2018-10-05 11-41-27

GeorgiSTodorov:
The "Service is being installed" looks better now, let's see the tests.

TattdCodeMonkey:
@GeorgiSTodorov I ended up removing the `Trans` macro from there for now. I moved the whole message into a variable so we can address it later. But we have another issue to address wrapping Modal's with `I18nProvider` which is likely why that was not being evaluated correctly.

GeorgiSTodorov:
Hey, how did you add the messages to `messages.json`?

TattdCodeMonkey:
@GeorgiSTodorov you have to run `npx lingui extract`

PR #3333, opened 2018-10-04
DCOS-43050: add POC for wrapping const text resources

natmegs:
Add POC for wrapping const text resources outside of React components. Covers 2 cases:
- Constants marked for translation but not immediately translated. These are used in conjunction with `<Trans>` component.
- Strings that need to be pre-translated (not just marked for translation)

Closes DCOS-43050.

## Testing

- Run `npx lingui extract` and ensure that both text wrapped in `i18nMark()` from `RepositoriesTableHeaderLabels.js` and text wrapped in `i18n._(t``)` from `FrameworkConfiguration.js` has been added to `locale/en/messages.json`
- Run app 
- Ensure table in Settings > Package Repositories displays as expected
- Go to Catalog > cassandra > Review & Run > Cancel > ensure that `Cancel` and `Discard` button text on confirmation popup is visible as expected

## Trade-offs

Constants must be marked for translation (`i18nMark`) and then wrapped where they are used. It can be cumbersome to track down where constants have been passed through to be used. Pre-translated strings rely on `i18n` prop, which comes from using HOC `withI18n`.

### DIFF
--- a/locale/en/messages.json
+++ b/locale/en/messages.json
@@ -26,6 +26,15 @@
       ]
     ]
   },
+  "Cancel": {
+    "translation": "",
+    "origin": [
+      [
+        "src/js/components/FrameworkConfiguration.js",
+        345
+      ]
+    ]
+  },
   "Catalog": {
     "translation": "",
     "origin": [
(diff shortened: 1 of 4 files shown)


natmegs:
@brandonc Updated! Added example with the pretranslated strings in `FrameworkConfiguration`

GeorgiSTodorov:
Looks good. Are we going to merge this?

brandonc:
I think we should commit the `lingui extract` with these changes before merging

PR #3334, opened 2018-10-04
chore(lingui): Reconfig to include/exclude more paths

brandonc:
I initially neglected the plugins/ folder and added some new exclusions to slightly speed up the process of extraction. @natmegs


GeorgiSTodorov:
Should we merge this into 1.12?

brandonc:
@GeorgiSTodorov thankfully, 1.12 was cut before before I merged the lingui integration

PR #3341, opened 2018-10-05
DCOS-41415: Add universe publishing step to CI

DanielMSchmidt:
## Testing

- Verify that the dcos-ui-universe S3 bucket contains bundle (e.g. by doing a setup according to the readme and you should see dcos-ui in the catalog without the universe being active (removed as a repo e.g.))
- Verify that the CI builds and uploads to S3 by visiting this [url](https://s3.console.aws.amazon.com/s3/buckets/downloads.mesosphere.io/dcos-ui-update-service/?region=us-west-2&tab=overview) 

## Trade-offs

- I use the DC/OS commons tooling as I think rewriting an own solution might lead to fragmentation and we are cut out from tooling improvements.
- I use the default branch of DC/OS commons, this might lead to breaks, but it will help us keep up with the universe developments

## Dependencies

None


GeorgiSTodorov:
I could use a bit more detailed testing instructions. Thank you.

nLight:
As we going to be publishing EE version only does it make sense to move the code into plugins repo? And squashing the `WIP - comment pipeline out` commit please.

GeorgiSTodorov:
Looks good and the first step works, but I can't verify the second one, because I don't have an AWS account.

DanielMSchmidt:
@nLight Why don't we have both? https://github.com/mesosphere/dcos-ui-plugins-private/pull/851

nLight:
So that in Universe we have only one package `dcos-ui` why do you think having them separated is beneficial? We sowieso end up shiping ee ui openly.

nLight:
Would you please describe a bit how it works along with EE repo? I don't see EE package template there nor do I see EE asset in the template in this repository. Are we going to have two separate universe packages? `dcos-ui` `dcos-ui-ee`? If yes, how is this reflected in proposed structure?

DanielMSchmidt:
We have two different builds for the entire tests, one with the plugins and one without, therefore we can just upload the very same build. The packages will have the very same name, you can decide which repository to add and thereby choose between the ee and oss version

nLight:
@DanielMSchmidt which one of those will we be publishing to the Universe? Or both?

DanielMSchmidt:
@nLight I don't think this decision is in the scope of the ticket, I understood "For stable version we're going to manually create a PR against Universe." as a completely manual step, but I am sure we could also automate or semi-automate it on this basis

nLight:
ok, fair enough! Thanks for taking time to explain your position!

DanielMSchmidt:
Sorry that it fell through the cracks on friday 😇

PR #3342, opened 2018-10-05
DCOS_OSS-4222: Show cosmos errors

nLight:
Properly handle Cosmos errors.

Closes: DCOS_OSS-4222

## Testing

1. Install something from the Catalog i.e. Kafka
2. Try editing the service
3. It should open the Editing form
4. Block `/cosmos/service/describe` URL in Chrome inspector
5. Refresh the Form
6. It now should show an error

Without the fix error would not appear. UI would keep showing loading bubble.

## Trade-offs

I added some markup to fix the styling of the message. Otherwise it would be black on black and left-aligned.
## Screenshots
screenshot 2018-10-05 at 15 59 49


nLight:
@DanielMSchmidt there's one :)

PR #3352, opened 2018-10-09
DCOS-43087: Wrap text content from /plugins folder using Trans macro

natmegs:
Use `Trans` macro from `@lingui/macro` to wrap text content in `/plugins` folder. All plugin components included except `/services/components` which will be added to this PR or in a separate PR by @TattdCodeMonkey. 

Adds configuration to enable use of macros in typescript files.

## Testing

There should be no visible difference when using any of the functionality that comes from the `/plugins` folder. The purpose of this PR is to localize text content that can be wrapped in the `Trans` macro from `@lingui/macro`.

### DIFF
--- a/locale/en/messages.json
+++ b/locale/en/messages.json
@@ -1513,6 +1513,15 @@
       ]
     ]
   },
+  "Other": {
+    "translation": "",
+    "origin": [
+      [
+        "plugins/nodes/src/js/components/NodesGridView.js",
+        94
+      ]
+    ]
+  },
   "Package Repositories": {
     "translation": "",
     "origin": [
--- a/locale/en/messages.json
+++ b/locale/en/messages.json
@@ -1374,6 +1374,24 @@
       ]
     ]
   },
+  "Last Failure:": {
+    "translation": "",
+    "origin": [
+      [
+        "plugins/jobs/src/js/components/JobsOverviewTable.js",
+        260
+      ]
+    ]
+  },
+  "Last Success:": {
+    "translation": "",
+    "origin": [
+      [
+        "plugins/jobs/src/js/components/JobsOverviewTable.js",
+        249
+      ]
+    ]
+  },
   "Package Repositories": {
     "translation": "",
     "origin": [
--- a/locale/en/messages.json
+++ b/locale/en/messages.json
@@ -1513,6 +1513,15 @@
       ]
     ]
   },
+  "Non-Leaders": {
+    "translation": "",
+    "origin": [
+      [
+        "plugins/nodes/src/js/components/NonLeadersGrid.js",
+        70
+      ]
+    ]
+  },
   "Other": {
     "translation": "",
     "origin": [
@@ -2057,6 +2066,15 @@
       [
         "plugins/jobs/src/js/pages/JobConfiguration.js",
         96
+      ]
+    ]
+  },
+  "There are no more known masters in this cluster.": {
+    "translation": "",
+    "origin": [
+      [
+        "plugins/nodes/src/js/components/NonLeadersGrid.js",
+        40
       ]
     ]
   },
--- a/locale/en/messages.json
+++ b/locale/en/messages.json
@@ -1365,6 +1365,100 @@
       ]
     ]
   },
+  "CPUs": {
+    "translation": "",
+    "origin": [
+      [
+        "plugins/jobs/src/js/pages/JobConfiguration.js",
+        32
+      ]
+    ]
+  },
+  "CRON Schedule": {
+    "translation": "",
+    "origin": [
+      [
+        "plugins/jobs/src/js/pages/JobConfiguration.js",
+        84
+      ]
+    ]
+  },
+  "Command": {
+    "translation": "",
+    "origin": [
+      [
+        "plugins/jobs/src/js/pages/JobConfiguration.js",
+        50
+      ]
+    ]
+  },
+  "Description": {
+    "translation": "",
+    "origin": [
+      [
+        "plugins/jobs/src/js/pages/JobConfiguration.js",
+        26
+      ]
+    ]
+  },
+  "Disk Space (Mib)": {
+    "translation": "",
+    "origin": [
+      [
+        "plugins/jobs/src/js/pages/JobConfiguration.js",
+        44
+      ]
+    ]
+  },
+  "Docker Container": {
+    "translation": "",
+    "origin": [
+      [
+        "plugins/jobs/src/js/pages/JobConfiguration.js",
+        106
+      ]
+    ]
+  },
+  "Enabled": {
+    "translation": "",
+    "origin": [
+      [
+        "plugins/jobs/src/js/pages/JobConfiguration.js",
+        78
+      ]
+    ]
+  },
+  "General": {
+    "translation": "",
+    "origin": [
+      [
+        "plugins/jobs/src/js/pages/JobConfiguration.js",
+        16
+      ]
+    ]
+  },
+  "ID": {
+    "translation": "",
+    "origin": [
+      [
+        "plugins/jobs/src/js/pages/JobConfiguration.js",
+        20
+      ],
+      [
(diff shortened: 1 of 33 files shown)


locale/en/messages.json
```diff
@@ -17,6 +35,15 @@
       ]
     ]
   },
+  "Are you sure you want to stop {headerContent}?": {
```
GeorgiSTodorov: Is having {headerContent} here intended?

GeorgiSTodorov: Nevermind, I tested it and it works fine.

plugins/catalog/src/js/repositories/components/RepositoriesDeleteConfirm.js
```diff
@@ -11,11 +12,11 @@ const RepositoriesDeleteConfirmMessage = ({ repository, error }) => {
...
   return (
     <div>
-      <p>
-        {`Repository (${label}) will be ${UserActions.DELETED}
-          from ${Config.productName}. You will not be able to install any
-          packages belonging to that repository anymore.`}
-      </p>
+      <Trans render="p">
+        Repository ({label}) will be {UserActions.DELETED}
+        from {Config.productName}. You will not be able to install any packages
```
TattdCodeMonkey: Does this render with a space between packages and belonging? I'm seeing mixed results in testing.

plugins/catalog/src/js/repositories/components/RepositoriesPage.js
```diff
@@ -9,7 +10,9 @@ const RepositoriesBreadcrumbs = addButton => {
   const crumbs = [
     <Breadcrumb key={-1} title="Repositories">
       <BreadcrumbTextContent>
-        <Link to="/settings/repositories">Package Repositories</Link>
+        <Link to="/settings/repositories">
+          <Trans>Package Repositories</Trans>
```
TattdCodeMonkey: Should this have a `render="span"`?

plugins/jobs/src/js/components/JobDeleteModal.js
```diff
@@ -26,9 +27,9 @@ const ModalMessage = ({ jobId, stopCurrentJobRuns }) => {
 
   return (
     <div>
-      <h2 className="text-danger text-align-center flush-top">
+      <Trans render="h2" className="text-danger text-align-center flush-top">
         {StringUtil.capitalize(UserActions.DELETE)} Job
```
TattdCodeMonkey: if `UserActions.DELETE` is a constant what is the value of the variable here? I know it was like this already but for translation including the string literal or moving the `StringUtil.capitalize(UserActions.DELETE)` to a const to be translated with `t` or `i18nmark` _could_ be better for translating.

natmegs: I think we would handle that in the ticket associated with non-jsx/constant resources?

plugins/catalog/src/js/repositories/components/RepositoriesPage.js
natmegs: It seems that the only time I've had to add `render="span"` is when there are unit tests associated because they fail & complain about Invariant Violation

brandonc: You can also avoid the additional span here using:
`<Trans render={<Link to="/settings/repositories" />}>Package Repositories</Trans>`

plugins/jobs/src/js/components/Breadcrumbs.js
```diff
@@ -10,7 +11,9 @@ export default function Breadcrumbs({ jobPath, jobName, jobInfo, children }) {
   let breadcrumbParts = [
     <Breadcrumb key={"Jobs"} title="Jobs">
       <BreadcrumbTextContent>
-        <Link to={"/jobs/overview/"}>{"Jobs"}</Link>
+        <Link to={"/jobs/overview/"}>
+          <Trans render="span">Jobs</Trans>
```
brandonc: See above comment. I've been avoiding the span render in similar BreadcrumbTextContent components.

plugins/jobs/src/js/components/JobsOverviewTable.js
```diff
@@ -245,7 +246,9 @@ export default class JobsOverviewTable extends React.Component {
     if (lastSuccessAt != null) {
       nodes.push(
         <p className="flush-bottom" key="tooltip-success-at">
-          <span className="text-success">Last Success: </span>
+          <Trans render="span" className="text-success">
+            Last Success:
```
brandonc: ❓ You may need to add a `{" "}` here to get the exact same markup

plugins/jobs/src/js/components/JobsOverviewTable.js
```diff
@@ -254,7 +257,9 @@ export default class JobsOverviewTable extends React.Component {
     if (lastFailureAt != null) {
       nodes.push(
         <p className="flush-bottom" key="tooltip-failure-at">
-          <span className="text-danger">Last Failure: </span>
+          <Trans render="span" className="text-danger">
+            Last Failure:
```
brandonc: ❓ You may need to add a `{" "}` here to get the exact same markup

plugins/nodes/src/js/components/NodeBreadcrumbs.js
```diff
@@ -14,7 +15,7 @@ const NodeBreadcrumbs = ({ nodeID, taskID, taskName, unitID }) => {
     <Breadcrumb key={-1} title="Nodes">
       <BreadcrumbTextContent>
         <Link to="/nodes" key={-1}>
-          Nodes
+          <Trans render="span">Nodes</Trans>
```
brandonc: `render={Link}`

plugins/catalog/src/js/repositories/components/RepositoriesDeleteConfirm.js
```diff
@@ -11,11 +12,11 @@ const RepositoriesDeleteConfirmMessage = ({ repository, error }) => {
 
   return (
     <div>
-      <p>
-        {`Repository (${label}) will be ${UserActions.DELETED}
-          from ${Config.productName}. You will not be able to install any
-          packages belonging to that repository anymore.`}
-      </p>
+      <Trans render="p">
+        Repository ({label}) will be {UserActions.DELETED}
```
brandonc: I think you should include "deleted" as a literal here so it can be translated

plugins/jobs/src/js/components/JobDeleteModal.js
brandonc: I disagree. I think the translatable string we want in the messages.json file includes the word "Delete" and shouldn't be translated separately.

plugins/oauth/components/LoginPage.js
```diff
@@ -104,23 +105,25 @@ class LoginPage extends mixin(StoreMixin) {
         </div>
         <Modal
           onClose={this.handleModalClose}
-          open={this.state.showClusterError}
+          open={true}
```
nLight: Forgotten piece?

natmegs: Oh my - you are totally right. Good catch!!

natmegs:
Rebased and
@nLight Comment addressed (thank you!)
@brandonc @TattdCodeMonkey your reviews got dismissed with the above change + rebase

plugins/jobs/src/js/components/JobStopRunModal.js
```diff
@@ -26,7 +29,11 @@ class JobStopRunModal extends React.Component {
       bodyText = "the selected job runs";
     }
 
-    return <span key="confirmText">You are about to stop {bodyText}.</span>;
+    return (
+      <span key="confirmText">
+        <Trans render="span">You are about to stop {bodyText}.</Trans>
```
TattdCodeMonkey: instead of using `bodyText` here should we have two seperate returns so the extracted messages are more complete sentences and easier to translate?

PR #3357, opened 2018-10-10
DCOS-41626: Remove black screen transition

natmegs:
Remove black screen transition when loading form after clicking Review & Run from catalog. Used same loader from Catalog and other pages.

Closes DCOS-41626.

## Testing

Go to Catalog > Any package > click Review & Run. There should be no partial black screen when loading form.

## Trade-offs

None I'm aware of.

## Screenshots

Before:
before

After:
after

Catalog loader (for comparison):
catalogloader


GeorgiSTodorov:
Note for reviewers: This is a Chrome-only bug, so if you are usually using Firefox (like I am), please check in Chrome.

Question: Are we going to add this fix to master as well?

nLight:
@GeorgiSTodorov we will

nLight:
Shall we fix that one as well? When trying editing a catalog package there's this transition again.

oct-11-2018 11-43-46

PR #3359, opened 2018-10-12
DCOS-43371: mark strings in src/js/constants files for translation

natmegs:
Mark strings in src/js/constants files for translation. Note: most of these files do not contain user-facing strings and therefore do not need to be marked/translated.

## Testing

Should be no visual changes.... but to be safe:

**To test `UnitSummaries`:**
Go to Nodes > click a node > Health > click any entry > see if Summary renders correctly

**To test `BulkOptions`:**
Note: the oss version of this feature is replaced with the plugins-private version, so in src/js/plugin-bridge/Loader.js replace `externalPluginList` as follows:

```
let pluginsList;
let externalPluginsList = {};

// Try loading the list of plugins.
try {
  pluginsList = require("#PLUGINS/index.js");
} catch (err) {
  // No plugins
  pluginsList = {};
}

// Try loading the list of plugins.
// try {
//   externalPluginsList = require("#EXTERNAL_PLUGINS/index.js");
// } catch (err) {
//   externalPluginsList = {};
// }
```

- run app and go to Organization > Users
- click checkbox beside a user > Delete button should appear beside filter input
- click Delete button > modal text should appear as expected

## Trade-offs

Composing strings of multiple different constants = headache for localization and for translators as sentences have to be chopped up & grammatical context is lost

### DIFF
--- a/src/js/components/UnitsHealthNodeDetail.js
+++ b/src/js/components/UnitsHealthNodeDetail.js
@@ -3,7 +3,6 @@ import mixin from "reactjs-mixin";
 import React from "react";
 /* eslint-enable no-unused-vars */
 import { StoreMixin } from "mesosphere-shared-reactjs";
-import { Trans } from "@lingui/macro";
 
 import Loader from "./Loader";
 import RequestErrorMsg from "./RequestErrorMsg";
@@ -85,11 +84,6 @@ class UnitsHealthNodeDetail extends mixin(StoreMixin) {
     const unitSummary = UnitSummaries[unit.get("id")] || {};
     const unitDocsURL =
       unitSummary.getDocumentationURI && unitSummary.getDocumentationURI();
-    const joinedUnitSummary = unitSummary.summary
-      .map((sumPiece, i) => (
-        <Trans key={sumPiece + i} render="span" id={sumPiece} />
-      ))
-      .reduce((prev, curr) => [...prev, " ", curr], []);
 
     return (
       <UnitsHealthNodeDetailPanel
@@ -98,7 +92,7 @@ class UnitsHealthNodeDetail extends mixin(StoreMixin) {
         docsURL={unitDocsURL}
         hostIP={node.get("host_ip")}
         output={node.getOutput()}
-        summary={joinedUnitSummary}
+        summary={unitSummary.summary}
       />
     );
   }
--- a/src/js/config/Config.ts
+++ b/src/js/config/Config.ts
@@ -1,5 +1,5 @@
 /* eslint no-redeclare: 0 */
-import { i18nMark } from "@lingui/react";
+
 import ConfigDev from "./Config.dev";
 import ConfigTest from "./Config.test";
 
@@ -56,7 +56,7 @@ let Config: IConfiguration = {
   fullProductName: "Mesosphere DC/OS",
   marathonAPIPrefix: "/service/marathon/v2",
   metronomeAPI: "/service/metronome",
-  productName: i18nMark("Mesosphere DC/OS"),
+  productName: "Mesosphere DC/OS",
   productHomepageURI: "https://dcos.io",
   setInactiveAfter: 30000,
   testHistoryInterval: 10000,
--- a/src/js/constants/UnitSummaries.js
+++ b/src/js/constants/UnitSummaries.js
@@ -1,210 +1,218 @@
-import { i18nMark } from "@lingui/react";
+/* eslint-disable no-unused-vars */
+import React from "react";
+/* eslint-enable no-unused-vars */
+
+import { Trans } from "@lingui/macro";
 import Config from "../config/Config";
 import MetadataStore from "../stores/MetadataStore";
 
 const UnitSummaries = {
   "dcos-marathon.service": {
-    summary: [
-      i18nMark(`The`),
-      Config.productName,
-      i18nMark("Marathon instance starts and monitors"),
-      Config.productName,
-      i18nMark("applications and services.")
-    ],
+    summary: (
+      <Trans>
+        The {Config.productName} Marathon instance starts and monitors{" "}
+        {Config.productName} applications and services.
+      </Trans>
+    ),
     getDocumentationURI() {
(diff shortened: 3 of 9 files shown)


GeorgiSTodorov:
Looks good. One small question - why is `summary` sometimes an array and sometimes not?
See https://github.com/dcos/dcos-ui/pull/3359/files#diff-c50de96b4362575af1cf8885a89a6bcfR19 and https://github.com/dcos/dcos-ui/pull/3359/files#diff-c50de96b4362575af1cf8885a89a6bcfR27

natmegs:
@GeorgiSTodorov You can pass either an array of translated message components (when entire message can't be translated at once like in this case), or you can pass a string message.

src/js/constants/DefaultErrorMessages.js
```diff
@@ -24,8 +26,8 @@ const DefaultErrorMessages = [
   //
   // NOTE: Every time you are about to write a description for the error
   //       message make sure that it sounds natural *both* on it's own
-  //       (ex. 'Must be one of {{value}}') AND when prefixed with the
-  //       expression 'The service' (ex. 'The service must be one of {{value}}')
+  //       (ex. 'Must be one of ||value||') AND when prefixed with the
```
nLight: so `i18nMark` will replace `||` with `{{` ?

natmegs: We are replacing these values with variables manually in `ErrorMessageUtil`(using regex to match the `{{ }}` pattern). Lingui doesn't like this syntax in its message extraction/translation catalog so I changed it to `||` instead to get around the issue

src/js/config/Config.ts
```diff
@@ -56,7 +56,7 @@ let Config: IConfiguration = {
   fullProductName: "Mesosphere DC/OS",
   marathonAPIPrefix: "/service/marathon/v2",
   metronomeAPI: "/service/metronome",
-  productName: "Mesosphere DC/OS",
+  productName: i18nMark("Mesosphere DC/OS"),
```
mperrotti: Can this even be translated? 🤔

nLight: I guess we make no exceptions :)

src/js/components/UnitsHealthNodeDetail.js
```diff
@@ -84,6 +85,11 @@ class UnitsHealthNodeDetail extends mixin(StoreMixin) {
     const unitSummary = UnitSummaries[unit.get("id")] || {};
     const unitDocsURL =
       unitSummary.getDocumentationURI && unitSummary.getDocumentationURI();
+    const joinedUnitSummary = unitSummary.summary
```
nLight: `TypeError: unitSummary.summary is undefined, can't access property "map" of it` to repro please check `DC/OS Checks Timer (Healthy)`

GeorgiSTodorov: The same thing happens in https://github.com/dcos/dcos-ui/pull/3367 .

TattdCodeMonkey: #3367 is based on this branch to get the changes to `ErrorMessageUtil` variable format. I will rebase changes from this branch as they are made.

src/js/constants/UnitSummaries.js
```diff
@@ -1,156 +1,210 @@
...
+        "Periodically writes /etc/resolv.conf so that only currently active Mesos masters with working Mesos DNS are in it."
+      )
+    ],
     getDocumentationURI() {
       return MetadataStore.buildDocsURI("/installing/ent/");
     }
   },
   "dcos-mesos-slave-public": {
-    summary: "Runs a publicly accessible Mesos agent on the node.",
+    summary: [i18nMark("Runs a publicly accessible Mesos agent on the node.")],
```
brandonc: Looking back at this file, I wonder if it would make a better translation to macro them all with tokens for productName inside, and just make summary a React node. We'd have to change a couple prop types to support node instead of string, but it would yield a less fragmented translation

natmegs:
Really helpful feedback, appreciate it! Using the `Trans` macro in constants files really simplified things a lot and eliminated some issues. Let me know what you think. Outstanding debate: `{{` vs `||`

src/js/constants/UnitSummaries.js
```diff
@@ -1,210 +1,218 @@
...
   "dcos-keepalived.service": {
-    summary: [
-      i18nMark(
-        "Runs keepalived to make a VRRP load balancer that can be used to access the masters."
-      )
-    ],
+    summary: (
+      <Trans>
+        Runs keepalived to make a VRRP load balancer that can be used to access
+        the masters."
```
brandonc: ☝️

GeorgiSTodorov:
Looks good, what are the advantages and disadvantages of `{{` and `||`?

TattdCodeMonkey:
@GeorgiSTodorov `lingui` uses `{` internally to replace variables, so it errors when we try to mark strings that include them. so we needed to change the format we use for variable templates in error messages. `||` is just an arbitrary selection of a different character that would hopefully not be needed in an error message or variable name and it doesn't conflict with `lingui`

src/js/constants/DefaultErrorMessages.js
```diff
@@ -35,127 +37,127 @@ const DefaultErrorMessages = [
   {
     path: /.*/,
     type: "ENUM",
-    message: "Must be one of {{values}}"
+    message: i18nMark("Must be one of ||values||")
```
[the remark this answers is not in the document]
TattdCodeMonkey: what do you mean by "our audience"? Anything inside the template is replaced before being rendered. It is either filled with the variable value, or replaced with an empty string if the variable can't be found. 

The main audience we will need to worry about is communicating to translators that `||values||` is replaced at run time and they need to be included in the translation.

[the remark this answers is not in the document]
brandonc: doh sorry. I reviewed this before I saw the token replacer!

PR #3361, opened 2018-10-12
DCOS-43529: [1.13] Debug deployment time is not correct

GeorgiSTodorov:
Show time since last configuration when a service cannot deploy.

Closes https://jira.mesosphere.com/browse/DCOS-43529
Parent jira: https://jira.mesosphere.com/browse/DCOS-40674

## Testing
1. Start a single-container service.
2. Wait a few minutes.
3. Edit it so that it can't start (for example by setting a huge memory).
4. Wait 5 more minutes.
5. Verify that the warning message in the Debug tab of the service calculates time since last configuration update, not since the service was created.
6. Verify that the tooltip for this service in the services table is also using the correct time.
7. Test the same for multi-container service.

## Trade-offs
1. I noticed that for multi-container services, the configuration displayed is being updated every 2 seconds.
peek 2018-10-12 11-55

Not sure if this is intended. I still used the time it was last configured for the message. @leemunroe you created the jira, can you tell me if my assumption is correct?

2. Scale or restart actions don't reset the timer. Again, not sure if this is intended.
screenshot from 2018-10-12 12-45-30

## Dependencies
None.

### Before
screenshot from 2018-10-12 11-57-29

### After
screenshot from 2018-10-12 12-29-24

### DIFF
--- a/plugins/services/src/js/components/ServiceStatusIcon.js
+++ b/plugins/services/src/js/components/ServiceStatusIcon.js
@@ -35,9 +35,16 @@ const getTooltipContent = (service, content) => {
 class ServiceStatusIcon extends Component {
   getDeclinedOffersWarning(service) {
     if (DeclinedOffersUtil.displayDeclinedOffersWarning(service)) {
-      const timeWaiting =
-        Date.now() -
-        DateUtil.strToMs(DeclinedOffersUtil.getTimeWaiting(service.getQueue()));
+      // We are getting the last config change time in a different way for services and pods.
+      let lastChange = service.getQueue().since;
+
+      if (service.versionInfo) {
+        lastChange = service.versionInfo.lastConfigChangeAt;
+      } else if (service.spec) {
+        lastChange = service.spec.version;
+      }
+
+      const timeWaiting = Date.now() - DateUtil.strToMs(lastChange);
 
       return this.getTooltip(
         `DC/OS has been waiting for resources and is unable to complete this deployment for ${DateUtil.getDuration(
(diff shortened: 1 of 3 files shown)


leemunroe:
@GeorgiSTodorov Nice work.

> I noticed that for multi-container services, the configuration displayed is being updated every 2 seconds.

This seems like a bug. We should check with Marathon team to see what is going on here for pods. Maybe @nLight knows.

GeorgiSTodorov:
Looking into failing test.

GeorgiSTodorov:
```
/jenkins/workspace/_dcos-ui-ee-pipeline_PR-854-TKXLFOPQK4GKXQKOWVWLPI4JNS3NVSIWOVFBQGGPFOSC7OUXL2JA/dcos-ui-ee/dcos-ui-plugins-private/secrets/components/SecretActionsModal.js

  87:1  error  Expected indentation of 2 spaces but found 0  indent
```

nLight:
This i a rather tricky one. There still could be something in Marathon's queue that it couldn't schedule. We are operating of this information when showing the message. It's been reported to be rather confusing, yes.

Now imagine a scenario: you tried to deploy this huge service for a while so that Marathon ramped up the backoff timeout. Then you change the config and redeploy. Deploy went through, but we still show the indication that there's a stale deployment because the older one is still in the queue.

Related Jira https://jira.mesosphere.com/browse/DCOS-43443

I'm afraid we can't change one without changing another and that would be a bigger change. @leemunroe we may want to have a tab for showing the queue in an explicit way with the button to delete a certain queue item.

I'm saying that this experience should be designed, since resolving one confusion can add more confusion in another place.

GeorgiSTodorov:
@nLight thank you for commenting, let's see what design team has to say.

juliangieseke:
@nLight @GeorgiSTodorov is this blocked by the linked JIRA? If so we should close this PR and put the JIRA to blocked 😞(We're not going to implement the necessary feature this or next RI…)

GeorgiSTodorov:
I think that it is not blocked. It is a bug fix, that we may not want to merge, because it potentially breaks something else. I guess we are waiting for @leemunroe to say if we want to close this jira.

PR #3362, opened 2018-10-12
DCOS-43372 i18nmark services plugin constants

TattdCodeMonkey:
Using `i18nMark` to mark strings used in the `plugins/services/src/js/constants` for translation. Also updating usages of the constants to be translated using either the `Trans` macro or `i18n` instance object.

## Testing

There _should_ be no visual changes.

## Trade-offs

Instead of wrapping the `TaskTable` component in the `withI18n()` higher-order component directly, I wrapped it's parent `TaskView` in `withI18n()` and passed the `i18n` instance object to `TaskTable` as a prop. This was to not disrupt the existing `TaskTable` unit tests that are testing function on the `TaskTable` component. By wrapping `TaskTable` with `withI18n()` the functions were no longer easily accessible in the unit tests.  Another option would of been to wrap `TaskTable` in `withI18n()` for the default export but also export the unwrapped component. And yet another option would be to refactor the functions we want to unit test to a utils file instead of trying to access them on the component directly. 

The `ServiceErrorMessages` file was marked in this PR, but translations of error messages will be done in another PR by refactoring functions in the `ErrorMessageUtil` file.

### DIFF
--- a/src/js/components/PlacementConstraintsPartial.js
+++ b/src/js/components/PlacementConstraintsPartial.js
@@ -1,4 +1,5 @@
-import { Trans } from "@lingui/macro";
+import { Trans, t } from "@lingui/macro";
+import { withI18n } from "@lingui/react";
 import React, { Component } from "react";
 import { Tooltip, Select, SelectOption } from "reactjs-components";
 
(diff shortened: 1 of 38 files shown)


src/js/components/PlacementConstraintsPartial.js
```diff
@@ -211,7 +216,7 @@ export default class PlacementConstraintsPartial extends Component {
 
     return (
       <div>
-        {this.getPlacementConstraintsFields(data.constraints)}
+        {this.getPlacementConstraintsFields(data.constraints, i18n)}
```
brandonc: Alternatively you could pull i18n from props inside getPlacementConstraintsFields

PR #3363, opened 2018-10-12
DCOS-43373: Mark and translate jobs plugin constants

brandonc:
## Testing

There should be no visual differences or markup changes after this PR is merged.

## Dependencies

Once the dust settles we need to run `npm run lingui-extract-with-plugins` and commit the message catalog.

## Notes

I was able to remove some unused job methods that depended on some unused job status constants


nLight:
Unused code seems to be ok to remove

PR #3370, opened 2018-10-17
Fix logo quality and alignment

leemunroe:
This PR addresses the following visual issues:

1. Replaces the logo graphic with the 2x logo so it isn't blurry on retina screens.
2. Moves the logo down 2px so it is better aligned with the text in the header.
3. Adds some padding to the sidebar for better alignment as per DCOS-40327

Before
image
image

After
image
image


leemunroe:
@jaxesn I think you mean _wayyy_ better

jaxesn:
oh right

PR #3373, opened 2018-10-24
DCOS-43820: Mark and translate tab labels

brandonc:
Noticed this translation opportunity while working on adding translation to enterprise plugins.

## Testing

There should be no visual differences or markup changes as a result of merging this PR

### DIFF
--- a/src/js/mixins/TabsMixin.js
+++ b/src/js/mixins/TabsMixin.js
@@ -1,3 +1,4 @@
+import { Trans } from "@lingui/macro";
 import classNames from "classnames/dedupe";
 import { Link, formatPattern } from "react-router";
 /* eslint-disable no-unused-vars */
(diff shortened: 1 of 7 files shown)


nLight:
Do we need to update messages.json?

brandonc:
@nLight that file conflicts so much that we just want to extract one time after merging all the competing L10n PR's. Everything works without it because the message ids equal the english fallback

src/js/mixins/TabsMixin.js
```diff
@@ -74,9 +75,11 @@ const TabsMixin = {
         onClick={this.tabs_handleTabClick.bind(this, tab)}
         {...attributes}
       >
-        <span className="menu-tabbed-item-label-text">
-          {this.tabs_tabs[tab]}
-        </span>
+        <Trans
+          render="span"
+          className="menu-tabbed-item-label-text"
+          id={this.tabs_tabs[tab]}
```
nLight: We're translating all labels independently. Do we need this `Trans`

brandonc: `i18nMark` doesn't translate, it just let's the extractor know about a string that needs translation. You then plug any of those into a Trans component to perform the actual translation.

nLight: Oh! ok, thanks for explaining this

PR #3376, opened 2018-10-24
[master] general ci fixes

juliangieseke:

### DIFF
--- a/scripts/ci/upload-release
+++ b/scripts/ci/upload-release
@@ -65,7 +65,7 @@ RELEASE_NAME="${RELEASE_NAME}" \
 
 # Upload source maps to sentry.io
 SENTRY_AUTH_TOKEN="${SENTRY_AUTH_TOKEN}" \
-  RELEASE_NAME="${RELEASE_NAME}" \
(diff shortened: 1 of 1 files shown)


scripts/ci/upload-release
```diff
@@ -65,7 +65,7 @@ RELEASE_NAME="${RELEASE_NAME}" \
 
 # Upload source maps to sentry.io
 SENTRY_AUTH_TOKEN="${SENTRY_AUTH_TOKEN}" \
-  RELEASE_NAME="${RELEASE_NAME}" \
+  RELEASE_NAME="${PKG_TARGET}+${PKG_VERSION}" \
```
nLight: why was it important?

juliangieseke: check sentry UI, we uploaded sourcemaps for `master+dcos-ui+v1.23.4` but reported to `master+v1.23.4`… 🤷‍♂️

juliangieseke: `RELEASE_NAME` is only used as filename, not as release version string (we might want to get rid of that different name at all 🤔 )

PR #3379, opened 2018-10-25
[1.12] fix(ci): update cli version for system tests

juliangieseke:
since we released 1.12, lets use the correct CLI version 👍


nLight:
Looks like it didn't fix the issue

PR #3380, opened 2018-10-25
DCOS-42974: Update http-server

nLight:
We've been using `http-server` from a _branch_ of a _fork_. Here we're switching to _our_ fork. Where I went ahead and merged the branch to master plus I addressed the issue with proxy crashing on errors.

Proxy cashing crashes CI when cluster is not yet available when we start calling it.

## Testing

CI should pass.
To test locally you may want to `npm i`
and `npx http-server -P <valid_but_non_existing_URL>` and then visit the Url the http-server will give you. Without the change the server would crash. With the change it will just log an error.


nLight:
Gotta backport this if it's 🍏

PR #3386, opened 2018-10-30
chore(L10n): Extract messages that include plugins-private

brandonc:
Just the messages file that are included in https://github.com/mesosphere/dcos-ui-plugins-private/pull/863. This may be necessary for that PR to pass system tests.

### DIFF
--- a/locale/en/messages.json
+++ b/locale/en/messages.json
@@ -87,6 +87,42 @@
       ]
     ]
   },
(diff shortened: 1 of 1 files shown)


juliangieseke:
Just to keep it in mind: If we happen to need this more often, lets adjust jenkins to use OSS branch on EE pipeline

PR #3391, opened 2018-10-31
DCOS-43897: fix table cell collapse

mperrotti:
Removed min-width table-cell truncation hack because most browsers will obey the min-width: 0 style, causing the layout to break.

Tested in Firefox, Edge, Safari, and Chrome. Also tested to ensure text truncation still works.

## Testing
**Test in Firefox or Edge**
0. Check out `mp/fix/DCOS-43897-ff-table-cell` from dcos-ui-plugins-private
1. Create a new service
2. Go to the "Environment" tab
3. Click "+ Add Environment Variable" and enter a key and a value
4. Click "+ Add " and enter a key and a value
5. Click "+ Add " and enter a key and a value
6. Go to "Networking" tab
7. Click "+ Add Service Endpoint"
8. Go to "Health Checks" tab
9. Click "+ Add Health Check" and add enter dummy info
10. Click "Review & Run"

No columns in the table should be collapsed

## Trade-offs
This solution is a little bit of a table layout hack

## Dependencies
None

## Screenshots
Before:
screen shot 2018-10-31 at 9 42 28 am

After:
screen shot 2018-10-31 at 9 43 21 am


mperrotti:
`width: 0` is a hack, but it's intentional. I just forgot to remove the `.table-layout-fixed` class from that table. The other tables don't seem to have a problem with those styles, so I didn't remove it there.

I tested in FF, Safari, Chrome, and Edge. Here's a screenshot from Chrome:
tablenocollapse ong

GeorgiSTodorov:
Tested both unit and integration tests, both ran successfully.

PR #3392, opened 2018-11-01
Add GraphiQL for development

DanielMSchmidt:
## Testing

- check that the GraphiQL side menu button is not reachable on a production build
- check that the GraphiQL route is not reachable on a production build

## Trade-offs

- We use observables as a fetcher interface, so the fetch only stops if you press the button manually, but you also get new data as it comes through, which is nice (see video below)

## Dependencies

There is a corresponding PR

## Screenshots

graphiql

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5570,6 +5570,22 @@
       "integrity": "sha1-DQcLTQQ6W+ozovGkDi7bPZpMz3c=",
       "dev": true
     },
+    "codemirror": {
+      "version": "5.48.4",
+      "resolved": "https://registry.npmjs.org/codemirror/-/codemirror-5.48.4.tgz",
+      "integrity": "sha512-pUhZXDQ6qXSpWdwlgAwHEkd4imA0kf83hINmUEzJpmG80T/XLtDDEzZo8f6PQLuRCcUQhmzqqIo3ZPTRaWByRA==",
+      "dev": true
+    },
+    "codemirror-graphql": {
+      "version": "0.8.3",
+      "resolved": "https://registry.npmjs.org/codemirror-graphql/-/codemirror-graphql-0.8.3.tgz",
+      "integrity": "sha512-ZipSnPXFKDMThfvfTKTAt1dQmuGctVNann8hTZg6017+vwOcGpIqCuQIZLRDw/Y3zZfCyydRARHgbSydSCXpow==",
+      "dev": true,
+      "requires": {
+        "graphql-language-service-interface": "^1.3.2",
+        "graphql-language-service-parser": "^1.2.2"
+      }
+    },
     "collection-visit": {
       "version": "1.0.0",
       "resolved": "https://registry.npmjs.org/collection-visit/-/collection-visit-1.0.0.tgz",
@@ -6145,6 +6161,15 @@
       "integrity": "sha1-Z29us8OZl8LuGsOpJP1hJHSPV40=",
       "dev": true
     },
+    "copy-to-clipboard": {
+      "version": "3.2.0",
+      "resolved": "https://registry.npmjs.org/copy-to-clipboard/-/copy-to-clipboard-3.2.0.tgz",
+      "integrity": "sha512-eOZERzvCmxS8HWzugj4Uxl8OJxa7T2k1Gi0X5qavwydHIfuSHq2dTD09LOg/XyGq4Zpb5IsR/2OJ5lbOegz78w==",
+      "dev": true,
+      "requires": {
+        "toggle-selection": "^1.0.6"
+      }
+    },
     "core-js": {
       "version": "1.2.7",
       "resolved": "https://registry.npmjs.org/core-js/-/core-js-1.2.7.tgz",
@@ -6278,6 +6303,30 @@
       "requires": {
         "fbjs": "^0.8.0",
         "gud": "^1.0.0"
+      }
+    },
+    "cross-fetch": {
+      "version": "2.2.2",
+      "resolved": "https://registry.npmjs.org/cross-fetch/-/cross-fetch-2.2.2.tgz",
+      "integrity": "sha1-pH/09/xxLauo9qaVoRyUhEDUVyM=",
+      "dev": true,
+      "requires": {
+        "node-fetch": "2.1.2",
+        "whatwg-fetch": "2.0.4"
+      },
+      "dependencies": {
+        "node-fetch": {
+          "version": "2.1.2",
+          "resolved": "https://registry.npmjs.org/node-fetch/-/node-fetch-2.1.2.tgz",
+          "integrity": "sha1-q4hOjn5X44qUR1POxwb3iNF2i7U=",
+          "dev": true
+        },
+        "whatwg-fetch": {
+          "version": "2.0.4",
+          "resolved": "https://registry.npmjs.org/whatwg-fetch/-/whatwg-fetch-2.0.4.tgz",
+          "integrity": "sha512-dcQ1GWpOD/eEQ97k66aiEVpNnapVj90/+R+SXTPYGHpYBBypfKJEQjLrvMZ7YXbKm21gXd4NcuxUTjiv1YtLng==",
+          "dev": true
+        }
       }
     },
     "cross-spawn": {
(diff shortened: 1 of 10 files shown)


src/js/config/Config.ts
```diff
@@ -37,6 +37,7 @@ interface IConfiguration {
   useUIConfigFixtures?: boolean;
   unitHealthAPIPrefix: string;
   useFixtures?: boolean;
+  isDevelopment?: boolean;
```
pierrebeitz: i'm not sure whether graphiQL justifies the introduction of a `isDevelopment`-flag in the `Config`. i would be fine with using an env-var or other stuff that indicates that this was a quick decision. adding the notion of a `development`-version feels a little odd to me.

src/js/routes/development.js
```diff
@@ -0,0 +1,50 @@
+import React from "react";
```
pierrebeitz: is it hard to make this file TS?

webpack/webpack.config.js
```diff
@@ -112,6 +112,7 @@ module.exports = {
           }
         ]
       },
+      { test: /\.flow$/, loader: "ignore-loader" },
```
pierrebeitz: ## 🔴 

why? 🤷‍♂️ 

i'd love to read about why this must be here, as i feel like we have to deeply understand everything in our webpack-config to be able to leverage opportunities of optimisation.

webpack/webpack.config.js
```diff
@@ -168,6 +169,33 @@ module.exports = {
           }
         ]
       },
+      {
+        test: /\.css$/,
```
pierrebeitz: ## 🔴 

same as above. why do we need a fully fledged css-config just for graphiQL?

why do we need a `postcss`-config to include what looks like regular css?

src/js/config/Config.ts
DanielMSchmidt: Hmm, I thought of this to be way cleaner than a environment variable usage, because this is the central place we store configuration for the frontend in

src/js/config/Config.template.ts
```diff
@@ -4,6 +4,7 @@ export default {
   analyticsKey: "39uhSEOoRHMw6cMR6st9tYXDbAL3JSaP",
   rootUrl: "",
   historyServer: "",
+  isDevelopment: true,
```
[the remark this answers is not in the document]
DanielMSchmidt: This could also be solved by lazy loading this route, right?

src/js/config/Config.ts
pierrebeitz: you suggest to change our usages of `process.env.NODE_ENV` throughout the code then?

i see where you're coming from, but looking at this change in isolation it appears to introduce yet another concept of figuring out whether we are in development-mode (just to determine whether graphiQL should be visible). i'd be also fine with something along the lines of `showGraphiQL`, because it's much more specific and clear.

src/js/config/Config.template.ts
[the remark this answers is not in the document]
pierrebeitz: # 👍

webpack/webpack.config.js
DanielMSchmidt: The GraphiQL package loads some flow files which add external type definitions to the flow type checker used in that project. They are imported, but they mean nothing to us in our setup, but because the GraphiQL project is not compiled down to a single file or at least to plain javascript it brings these requires to `.flow` files. They can not be parsed by webpack, and ultimatively they don't need to, so I added an ignore loader that just lets them go to the void

webpack/webpack.config.js
DanielMSchmidt: Similar reason as above

webpack/webpack.config.js
pierrebeitz: i don't like the burden they force on us.


> but because the GraphiQL project is not compiled down to a single file or at least to plain javascript

have you seen their integration-examples? isn't that a compiled js-file we could use as well? https://github.com/graphql/graphiql/blob/master/example/index.html#L40-L41

src/js/routes/index.js
```diff
@@ -14,6 +16,7 @@ import services from "../../../plugins/services/src/js/routes/services";
 import settings from "./settings";
 import systemOverview from "./system-overview";
 import catalog from "./catalog";
+import development from "./development";
```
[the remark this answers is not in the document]
Poltergeist: We could use require instead within the condition?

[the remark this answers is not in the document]
DanielMSchmidt: I think we fixed this in newer commits, let me check

webpack/webpack.config.js
Poltergeist: Agree with @pierrebeitz I would love if we would not need to add something we don't need. This is unnecessary and error prone if someone cleans it up because they think it is a legacy thing which we don't need anymore.

src/js/routes/index.js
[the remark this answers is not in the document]
DanielMSchmidt: I checked and with the current solution we don't include anything GraphiQL realted into the bundle, so the problem is fixed

webpack/webpack.config.js
DanielMSchmidt: As discussed offline these examples don't improve the situation for us

DanielMSchmidt: @Poltergeist I don't think not having them is possible without working on the upstream GraphiQL package

src/js/routes/development.js
```diff
@@ -0,0 +1,36 @@
...
+    </div>
+  );
+}
+
+GraphiQLWrapper.routeConfig = {
+  label: "GraphiQL",
+  matches: /^\/development\/graphiql/
+};
+
+export default function developmentFactory(
```
pierrebeitz: this does not look like a factory to me. 🤔

i'd even prefer an anonymous function here over `developmentFactory`.

src/js/routes/index.js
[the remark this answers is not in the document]
pierrebeitz: you're right in that we don't have a big problem here having lazy-loaded graphiQL. 

i still think it would preferable to not ship anything development-related at all in production. can we come up with a clean solution for that here, so we want to reuse it in the future?

src/typings/graphiql.d.ts
```diff
@@ -0,0 +1,21 @@
...
+  import { Observable } from "rxjs";
+
+  interface FetchConfig {
+    query: string;
+    variables?: object;
+  }
+
+  type Fetcher = (
+    config: FetchConfig
+  ) => Promise<unknown> | Observable<unknown> | unknown;
```
pierrebeitz: is that something we chose or is that decided by `graphiQL`? 

if the latter, isn't this rather more like the following type? 

```typescript
type Fetcher<T> = (
  config: FetchConfig
) => Promise<T> | Observable<T>;
```

(i did not dig too much here, just want to make sure this has been a fully informed decision)

webpack/webpack.config.js
pierrebeitz: TL;DR

let's try waiting for https://github.com/graphql/graphiql/issues/793 a bit. we should be able to get rid of this then.

-----------------------

it boils down to graphiQL using a version of `codemirror-graphql` that "accidentally" includes `graphql-language-service-interface` which in turn tried to include flow-files into our bundle. 

`graphql-language-service-interface` says "This is currently in technical preview. We welcome your feedback and suggestions.", so this is kind of expected. i feel JS-fatigue here :)

webpack/webpack.config.js
```diff
@@ -135,7 +136,7 @@ module.exports = {
         ]
       },
       {
-        test: /\.less$/,
+        test: /\.(less|css)$/,
```
pierrebeitz: that's handy!

src/js/routes/development.js
```diff
@@ -0,0 +1,36 @@
...
+);
+
+function GraphiQLLoader() {
+  return <div />;
+}
+
+function GraphiQLWrapper() {
+  return (
+    <div style={{ width: "100%", backgroundColor: "#FFF" }}>
+      <Suspense fallback={GraphiQLLoader}>
```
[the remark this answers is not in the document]
DanielMSchmidt: I understood [the docs](https://reactjs.org/docs/code-splitting.html#suspense) in a way that indicates to me that we are required to use it, is that a false assumption I am making?

> If the module containing the OtherComponent is not yet loaded by the time MyComponent renders, we **must** show some fallback content while we’re waiting for it to load - such as a loading indicator

package-lock.json
```diff
@@ -10002,17 +10046,20 @@
...
-          "dev": true
+          "dev": true,
+          "optional": true
         },
         "console-control-strings": {
           "version": "1.1.0",
           "bundled": true,
-          "dev": true
+          "dev": true,
+          "optional": true
```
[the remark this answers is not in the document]
DanielMSchmidt: I thought npm now automagically uses the version stated in the package.json, no matter what I have installed 🤔

packages/@extension-kid/data-layer/dataLayer.ts
```diff
@@ -73,6 +73,10 @@ export default class DataLayer {
     return makeExecutableSchema({ typeDefs, resolvers });
   }
 
+  getExecutableSchema$(): Observable<GraphQLSchema> {
```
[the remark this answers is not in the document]
DanielMSchmidt: There is the concept of a schema, which is a thing that describes your API, but can't actually do anything and there is also the concept of adding the layer of functionality to it calling it an executable schema. The method above is called `getExecutableSchema`, so I would rather make this renaming in a different PR. I think schema makes more sense here, because there are only executable schemas in our DC/OS UI world

src/typings/graphiql.d.ts
DanielMSchmidt: You are right, I misread the docs

src/js/routes/development.js
DanielMSchmidt: We would need to add types to the entire foundation ui package, so I cut the scope here

src/js/routes/development.js
DanielMSchmidt: 👍

package-lock.json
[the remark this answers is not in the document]
pierrebeitz: `npx npm i` yields a different result.

src/js/routes/development.js
pierrebeitz: i must be missing something here. i meant to ask about converting `src/js/routes/development.js` to TS. i don't understand how `foundation-ui` is an issue yet. (also we can choose to `// @ts-ignore` some imports for now, if that's what your concerns are about - it's still better than not typing the file at all IMHO)

package-lock.json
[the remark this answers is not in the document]
DanielMSchmidt: huh, interesting, thank you 👍

pierrebeitz:
i pulled in a newer version. we don't need to bend our webpack anymore!

also i added a `.default` after the inline-require, so the frontend boots up again in dev...

i'd be fine with merging this as is although i think we should introduce a  development-menu-thingy at the very bottom as the firetruck draws way too much attention. 

also it would be nice to have some documentation. you're fluent in graphql - i have no idea what to write into graphQL even to have any benefit during development.

PR #3394, opened 2018-11-05
Try to speed up our CI a little

DanielMSchmidt:
## Testing

Seeing CI pass is enough testing for this change

## Trade-offs

- Could have used 3. Party `npm ci` alternative, but the speedup is similar + we would get a new dependency
- Thought about trying to run everything on a different host as npm install is now half the time, but that would be a bigger change I would rather do at a later point in time.

### Get node modules
```
npm:        69.00 real        96.28 user        69.38 sys
npm ci:     36.47 real        28.50 user        19.65 sys
```

### Enabled parallelization

Before: 1h 10min in best case
After: 40min in best case


Jenkinsfile
```diff
@@ -31,13 +31,19 @@ pipeline {
 
     stage("Build") {
       steps {
-        sh "npm --unsafe-perm install"
-        sh "npm run build"
```
nLight: We still need to keep linting on the CI + switching to `build-assets` will omit `prebuild` stage.

```
"build-assets": "npm run clean && lingui compile && NODE_ENV=production webpack --colors --config ./webpack/webpack.production.js",
./dist/index.html {} \\;",
"build": "npm run lint && ([ -z \"${npm_config_externalplugins}\" ] || npm run lint-plugins) && npm test && npm run build-assets && npm run validate-build",
```

DanielMSchmidt: I see, but if I'd move prebuild to pretest this should be fine, shouldn't it?

juliangieseke: Yup, thats better anyway

DanielMSchmidt:
@juliangieseke @nLight I think I got your feedback now

Jenkinsfile
nLight: I don't see the comment being resolved

DanielMSchmidt: See the package.json

nLight: We still skipping  `validate-build` step. That could mean once we merge master is broken. 
Shall we move it to `postbuild-assets?`

nLight:
```

[cos-ui-oss-pipeline_PR-3394-YYHNNIFCFYR6D4RABSJKSQVXLXH3C26TBIRECMDDC3DJCTUO6MYA] Running shell script

+ npm --unsafe-perm ci

npm ERR! cipm can only install packages when your package.json and package-lock.json or npm-shrinkwrap.json are in sync. Please update your lock file with `npm install` before continuing.

npm ERR! 

npm ERR! 

npm ERR! Invalid: lock file's @dcos/ui-kit@1.16.1 does not satisfy @dcos/ui-kit@1.17.1

npm ERR! 



npm ERR! A complete log of this run can be found in:

npm ERR!     /root/.npm/_logs/2018-11-23T12_56_19_594Z-debug.log

script returned exit code 1
```

DanielMSchmidt:
This was an old try to speed up CI that was not really helpful

nLight:
I believe using `npm ci` _is_ helpful, the surrounding stuff is out of scope :)

DanielMSchmidt:
It only helps with like 3 minutes and it was done without Jira issue, so let's try to do it properly instead 👍

PR #3396, opened 2018-11-05
chore(L10n): Add fake chinese translation and utility

brandonc:
## Testing

Switch to Chinese locale under UI Settings

## Screenshots

screen shot 2018-11-05 at 11 47 27 am

## Observations

The primary nav needs to be localized but the secondary nav seems to already be done!


GeorgiSTodorov:
Are we going to merge this?

brandonc:
@natmegs should we merge this or just keep it in case we want to test something

nLight:
If we decide merging it, may I suggest using the `□` symbol instead of fake words? Or something neutral.

PR #3397, opened 2018-11-05
DCOS-43383: Localize sidebar navigation

brandonc:
## Testing

No changes to rendering or markup should be detected

## Trade-offs

I had to remove a small capability in NavigationService (adding navigation labels that are React elements) that was not being used in order to keep translation simple

### DIFF
--- a/locale/en/messages.json
+++ b/locale/en/messages.json
@@ -1648,6 +1648,10 @@
         4
       ],
       [
+        "src/js/pages/CatalogPage.js",
+        13
+      ],
+      [
         "src/js/pages/catalog/PackageDetailTab.js",
         36
       ],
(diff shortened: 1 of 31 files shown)


locale/en/messages.json
```diff
@@ -2927,11 +2943,15 @@
     ]
   },
   "Dashboard": {
-    "translation": "",
+    "translation": "Super Dash",
```
natmegs: ❓

src/js/components/Sidebar.js
```diff
@@ -163,7 +163,11 @@ class Sidebar extends React.Component {
       }
 
       if (group.category !== "root") {
-        heading = <h3 className="sidebar-section-header">{group.category}</h3>;
+        heading = (
+          <Trans render="h3" className="sidebar-section-header">
```
natmegs: For this to do the translation I think we need to give `{group.category}` to `Trans` as `id` prop

locale/en/messages.json
brandonc: OOps this was my test. Ill rebase it out. Thanks!

brandonc:
@natmegs have another go

PR #3398, opened 2018-11-08
DCOS-41419: Check available updates for installed UI packages

GeorgiSTodorov:
If a newer version of dcos-ui is available, inform the user.

Closes https://jira.mesosphere.com/browse/DCOS-41419

## Testing
Change `CHECK_DELAY` in `UpdateStream.ts` (on line 11) to be a smaller number, like 1000 or 2000;
Add `compare.subscribe(newVersion => console.log(newVersion));` at the end of `UpdateStream.ts`.
Open the UI and check the console for a message informing of a newer version of DCOS UI.

## Trade-offs
Some TODOs remain, updating them is in the scope of another ticket.

## Dependencies
None.

## Screenshots
Not applicable.


src/js/core/UpdateStream.js
```diff
@@ -0,0 +1,81 @@
...
+      action: "list-versions",
+      actionType: "response",
+      entity: "package",
+      version: "v1"
+    })
+  },
+  body: JSON.stringify({ includePackageVersions: true, packageName: "dcos-ui" })
+}).retry(4);
+
+export const compare = Observable.interval(CHECK_DELAY).switchMap(() =>
```
brandonc: I don't think this will ever run until the app has been open for 24 hours. According to the code docs (http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html#static-method-interval) the first emission is not sent immedately.

src/js/core/UpdateStream.js
```diff
@@ -0,0 +1,81 @@
...
+  Observable.combineLatest(
+    savedVersion,
+    fetchedVersion,
+    (savedVersion, newVersionObject) => {
+      const newVersion = Object.keys(newVersionObject.response.results)[0];
+      if (newVersion > savedVersion) {
```
brandonc: You may need a semantic version comparator, similar to the version sorting implementation that we use for sorting services by version. Reason being, the string "2.24.0" is greater than "2.100.0", but the semantic version is greater

brandonc: Should you be comparing to the `window.DCOS_UI_VERSION` ?

src/js/core/UpdateStream.js
```diff
@@ -0,0 +1,81 @@
...
+function showNotification(newVersion) {
+  // TODO: show notification
+  console.log(
+    "A new version of DCOS UI is available. Install " +
+      newVersion +
+      ". Dismiss."
+  );
+}
+
+setVersionInLocalStorage("2.24.0"); // TODO: Check if a version exists
```
brandonc: This should contain the dismissed version numbers, no?

It might also be helpful to store the last time you checked for a new version, so you don't potentially trigger the notification every time the app loads.

GeorgiSTodorov: I am not sure what you mean with dismissed version numbers. I think it should contain the last version that is installed. The current "2.24.0" is just a stub, the jira ticket says setting it is in the scope of another ticket.

nLight: Local storage will contain versions that the UI has notified a user about and the version the user has _dismissed_. We don't do manually `setVersionInLocalStorage`. Current version comes from `window.DCOS_UI_VERSION` as @brandonc pointed it the prev comment.

GeorgiSTodorov: @nLight Thank you, I understand now.

GeorgiSTodorov:
Thank you for the review, @brandonc .I addressed most of the feedback, I still need to write test and use `window.DCOS_UI_VERSION` as the current version.

GeorgiSTodorov:
I addressed the other feedback, @brandonc what kind of tests would you like to see?

src/js/core/UpdateStream.js
```diff
@@ -0,0 +1,100 @@
...
+        return newVersion;
+      }
+
+      return false;
+    }
+  )
+);
+
+compare.subscribe(value => {
+  if (value) {
```
nLight: If we design the stream like `timer -> checkNewVersion -> filterDismissedVersions` then `subscribe` will fire only when there's a new version that has not yet been dismissed. So no `if (value)` needed.

src/js/core/UpdateStream.js
```diff
@@ -0,0 +1,100 @@
...
+  body: JSON.stringify({ includePackageVersions: true, packageName: "dcos-ui" })
+}).retry(4);
+
+export const compare = Observable.timer(0, CHECK_DELAY).switchMap(() =>
+  Observable.combineLatest(
+    dismissedVersion,
+    fetchedVersion,
+    (dismissedVersion, newVersionObject) => {
+      if (
+        new Date().getTime() <
```
nLight: Why do we need this conditional?

src/js/core/UpdateStream.js
```diff
@@ -0,0 +1,100 @@
...
+function getContentType({ action, actionType, entity, version }) {
+  return `application/vnd.dcos.${entity}.${action}-${actionType}+json;charset=utf-8;version=${version}`;
+}
+
+function showNotification(newVersion) {
+  // TODO: show notification
+  // When dismissing, save dismissed version in local storage
+  const currentVersion = window.DCOS_UI_VERSION;
+  console.log(
+    "A new version of DCOS UI is available. Upgrade " +
```
nLight: You can always use template strings 
```js
`A new version of DCOS UI is available. Upgrade ${currentVersion} to ${newVersion}. Dismiss.`
```

src/js/core/UpdateStream.js
```diff
@@ -0,0 +1,100 @@
...
+  return value;
+}
+
+function setInLocalStorage(key, value) {
+  const savedStates = UserSettingsStore.getKey(SAVED_STATE_KEY) || {};
+  savedStates[DCOS_UI_VERSION] = {
+    dismissedVersion: getFromLocalStorage("dismissedVersion") || null,
+    lastTimeCheck: getFromLocalStorage("lastTimeCheck") || null
+  };
+  savedStates[DCOS_UI_VERSION][key] = value;
```
nLight: Why do you think it's beneficial to store dismissed version based of `DCOS_UI_VERSION`?

src/js/core/UpdateStream.js
GeorgiSTodorov: It's something @brandonc suggested and I liked the idea.

src/js/core/UpdateStream.js
```diff
@@ -0,0 +1,100 @@
+import { Observable } from "rxjs/Observable";
+import UserSettingsStore from "#SRC/js/stores/UserSettingsStore";
+import { request } from "@dcos/http-service";
+import Util from "#SRC/js/utils/Util";
+import { SAVED_STATE_KEY } from "#SRC/js/constants/UserSettings";
+import compareVersions from "compare-versions";
```
nLight: nitpick please put the `package.json` imports first and `#SRC` imports afterwards.

src/js/core/UpdateStream.js
```diff
@@ -0,0 +1,100 @@
...
+      if (compareVersions(newVersion, dismissedVersion) === 1) {
+        return newVersion;
+      }
+
+      return false;
+    }
+  )
+);
+
+compare.subscribe(value => {
```
nLight: I understand this is of debugging? Please don't subscribe to it right away in the production version.

src/js/core/UpdateStream.js
```diff
@@ -0,0 +1,100 @@
...
+    "A new version of DCOS UI is available. Upgrade " +
+      currentVersion +
+      " to " +
+      newVersion +
+      ". Dismiss."
+  );
+}
+
+setInLocalStorage("dismissedVersion", "2.24.2"); // stub
+const dismissedVersion = Observable.of(
```
nLight: We can use `startsWith` for default values http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html#instance-method-startWith

src/js/core/UpdateStream.js
```diff
@@ -0,0 +1,100 @@
...
+      entity: "package",
+      version: "v1"
+    })
+  },
+  body: JSON.stringify({ includePackageVersions: true, packageName: "dcos-ui" })
+}).retry(4);
+
+export const compare = Observable.timer(0, CHECK_DELAY).switchMap(() =>
+  Observable.combineLatest(
+    dismissedVersion,
```
nLight: I wonder if `dismissedVersion` will be re-checked every time the timer fires

src/js/core/UpdateStream.js
nLight: I think by declaring it like that we effectively pin the version that was in the localStore on browser load and never re-check it. 

Why not making `getFromLocalStorage` observable with `Observable.create`?

src/js/core/UpdateStream.js
nLight: So we don't re-check if a user refreshes the browser?

GeorgiSTodorov: No, because I think otherwise the 24-hour timer would be sort of useless, I don't think anyone will keep the UI 24 hours without restarting/refreshing. Do you think that we should?

nLight: Sentry suggests that some folks keep the UI open over weekends

nLight: Regardless I'd suggest rewriting this check from being imperative to an appropriate rxjs operator like `filter`

src/js/core/UpdateStream.js
GeorgiSTodorov: When should we subscribe in production?

nLight: That's a good question. Let me try to explain how I see this.

Ideally the whole update addition should be a plugin. Then I'd subscribe on the plugin bootstrap. I also can see us having "Disable updates" feature that would mean we never subscribe. 

Although that fantasy doesn't answer your question, I hope it hints why not subscribe right beside the declaration.

Working with what we have today, I'd say `pages/Index.js`? 

I would also like to consider wrapping an old-school plugin around the UI part. Like `plugins/banner` as a bare bone example. Leveraging the plugin we could also give an opportunity for a cluster admin to opt out from using the service entirely.

src/js/core/UpdateStream.js
```diff
@@ -0,0 +1,100 @@
...
+    fetchedVersion,
+    (dismissedVersion, newVersionObject) => {
+      if (
+        new Date().getTime() <
+        getFromLocalStorage("lastTimeCheck") + CHECK_DELAY
+      ) {
+        return;
```
brandonc: Is there a behavioral difference between return undefined and return false (below)?

src/js/core/UpdateStream.js
brandonc: This is the global shadowing/confusing I was talking about ;)

src/js/core/UpdateStream.js
GeorgiSTodorov: Good point, there shouldn't be.

nLight:
I'm super sorry not bringing this up before but would you please rewrite the stream in Typescript. Thanks! 🙇

GeorgiSTodorov:
Started working on tests and did typescript migration, although I have a small problem with `UserSettingsStore.getKey` and `.setKey` not getting recognized by typescript (but still working) that I will look into tomorrow.

mperrotti:
Could you also implement `/api/v1/update/<version>` and `/api/v1/reset`?

mperrotti:
Nevermind that last comment. I think just linking to `/catalog/packages/{packageName}/deploy?version={newVersion}` will do what I need 👍

src/js/pages/Index.js
```diff
@@ -72,6 +77,9 @@ var Index = React.createClass({
 
     ConfigStore.fetchCCID();
     ConfigStore.addChangeListener(EventTypes.CONFIG_ERROR, this.onConfigError);
+    compare.subscribe(values => {
+      showNotification(getVersionFromVersionObject(values[1]));
```
nLight: shall we `map` filtered result so that we could 
```js
compare.subscribe(newVersion => showNotification(newVersion))
```
?

src/js/pages/Index.js
```diff
@@ -20,6 +20,11 @@ import SidebarStore from "../stores/SidebarStore";
 import UserLanguageStore from "../stores/UserLanguageStore";
 import { hasViewportChanged, getCurrentViewport } from "../utils/ViewportUtil";
 import * as viewport from "../constants/Viewports";
+import {
+  compare,
+  getVersionFromVersionObject,
+  showNotification
```
nLight: This function should live somewhere else. Not in the `UpdateStream`.

GeorgiSTodorov: You are right, Mike and I even talked about this yesterday.

GeorgiSTodorov: I will leave it there for now, until @mperrotti and I decide the best place to have it.

src/js/pages/Index.js
GeorgiSTodorov: Good point, done.

GeorgiSTodorov:
System tests in `test-universe.js` are broken, looking into them.

GeorgiSTodorov:
Let's see now.

src/js/constants/UserSettings.d.ts
```diff
@@ -0,0 +1 @@
+export var SAVED_STATE_KEY: "savedStates";
```
DanielMSchmidt: 🎨: could be const
```suggestion
export const SAVED_STATE_KEY: "savedStates";
```

src/js/core/UpdateStream.ts
```diff
@@ -0,0 +1,145 @@
+import { Observable } from "rxjs/Observable";
+import compareVersions from "compare-versions";
+import { request, RequestResponse } from "@dcos/http-service";
+import { BehaviorSubject } from "rxjs";
+
+import UserSettingsStore from "#SRC/js/stores/UserSettingsStore";
+import Util from "#SRC/js/utils/Util";
+import { SAVED_STATE_KEY } from "#SRC/js/constants/UserSettings";
+
+export const UpdateStreamType: symbol = Symbol("UpdateStreamType");
```
DanielMSchmidt: Nice use of symbols here 👍

src/js/core/UpdateStream.ts
```diff
@@ -0,0 +1,145 @@
...
+const CHECK_DELAY: number = 24 * 60 * 60 * 1000; // Once every 24 hours
```
DanielMSchmidt:
```suggestion
const CHECK_DELAY = 24 * 60 * 60 * 1000; // Once every 24 hours
```

src/js/core/UpdateStream.ts
```diff
@@ -0,0 +1,145 @@
...
+interface ContentTypeObject {
+  action: string;
+  actionType: string;
+  entity: string;
+  version: string;
+}
+
+interface DCOSUserSettings {
+  dcosUIVersion: DCOSUIVersion;
+  [key: string]: DCOSUIVersion;
```
DanielMSchmidt: What is this `[key: string]` part about?

src/js/core/UpdateStream.ts
```diff
@@ -0,0 +1,145 @@
...
+  [key: string]: string | number;
+}
+
+export function showNotification(newVersion: string): void {
+  // TODO: Move inside UpdateServiceBanner
+  alert(newVersion);
+}
+
+function getVersionFromVersionObject(versionObject: any): string {
+  return Object.keys(versionObject.response.results)[0];
```
DanielMSchmidt: I think we have TS misconfigured, this should scream so much at you for accessing unknowns. What should this return if the versionsObject does not have a response or the response has no results or the results is an empty array? Some of those will return no string but a null, others will throw an error, which means that the type said to be returned here is wrong.

src/js/core/UpdateStream.ts
```diff
@@ -0,0 +1,145 @@
...
+}
+
+function getFromLocalStorage(key: string): any {
```
DanielMSchmidt: Do we really have no idea what is returned here?

src/js/core/UpdateStream.ts
```diff
@@ -0,0 +1,145 @@
...
+    ? getFromLocalStorage(DISMISSED_VERSION)
+    : "0"
+);
+const localStorageCheckedTime: BehaviorSubject<number> = new BehaviorSubject(
+  getFromLocalStorage(LAST_TIME_CHECK)
+    ? getFromLocalStorage(LAST_TIME_CHECK)
+    : 0
+);
+
+setInLocalStorage(DISMISSED_VERSION, "2.24.2"); // stub
```
DanielMSchmidt: Stub seems to be left over, I guess this is meant to be removed

src/js/core/UpdateStream.ts
```diff
@@ -0,0 +1,145 @@
...
+}: ContentTypeObject): string {
+  return `application/vnd.dcos.${entity}.${action}-${actionType}+json;charset=utf-8;version=${version}`;
+}
+
+const localStorageDismissedVersion: BehaviorSubject<
+  string
+> = new BehaviorSubject(
+  getFromLocalStorage(DISMISSED_VERSION)
+    ? getFromLocalStorage(DISMISSED_VERSION)
+    : "0"
```
DanielMSchmidt: Shouldn't this be sth like null instead? A string zero seems to be strange to identify as much as "nothing" (from my understanding)

src/js/core/UpdateStream.ts
```diff
@@ -0,0 +1,145 @@
...
+
+interface DCOSUIVersion {
+  dismissedVersion: string;
+  lastTimeCheck: number;
+  [key: string]: string | number;
+}
+
+export function showNotification(newVersion: string): void {
+  // TODO: Move inside UpdateServiceBanner
+  alert(newVersion);
```
DanielMSchmidt: This alert seems to be wrong here

src/js/core/UpdateStream.ts
```diff
@@ -0,0 +1,145 @@
...
+}
+
+interface DCOSUIVersion {
+  dismissedVersion: string;
+  lastTimeCheck: number;
+  [key: string]: string | number;
+}
+
+export function showNotification(newVersion: string): void {
+  // TODO: Move inside UpdateServiceBanner
```
DanielMSchmidt: ❌ Isn't this part of this issue?

src/js/core/UpdateStream.ts
```diff
@@ -0,0 +1,145 @@
...
+  {
+    method: "POST",
+    headers: {
+      "Content-Type": getContentType({
+        action: "list-versions",
+        actionType: "request",
+        entity: "package",
+        version: "v1"
+      }),
+      Accept: getContentType({
```
DanielMSchmidt: The `getContentType` name is strange if we get it for the `Accept` key

src/js/core/UpdateStream.ts
GeorgiSTodorov: We do, but if I write `string | number` for example, then the `localStorageDismissedVersion` definition complains that
```
[ts]
Type 'BehaviorSubject<string | number>' is not assignable to type 'BehaviorSubject<string>'.
  Types of property 'value' are incompatible.
    Type 'string | number' is not assignable to type 'string'.
      Type 'number' is not assignable to type 'string'. [2322]
```
.

src/js/core/UpdateStream.ts
DanielMSchmidt: So we can't really merge this PR until mikes PR is done? We should note is as a dependency

src/js/core/UpdateStream.ts
DanielMSchmidt: Okay, gotcha, having the alert there and potentially merging it to master makes me worry though

src/js/core/UpdateStream.ts
GeorgiSTodorov: It's the other way around. Mike uses this PR to make his stuff work.

src/js/core/UpdateStream.ts
GeorgiSTodorov: It used to be `console.log`, but typescript doesn't allow those.

src/js/core/UpdateStream.ts
GeorgiSTodorov: I don't understand, can you explain?

src/js/core/UpdateStream.ts
GeorgiSTodorov: I searched for the error and I thought it was complaining about the key.

src/js/core/UpdateStream.ts
GeorgiSTodorov: I added `interface VersionObject`, but then I got
screenshot from 2018-11-20 14-47-18
.
And when I changed `RequestResponse`,
screenshot from 2018-11-20 14-50-05
. Observables with types are weird.

src/js/core/UpdateStream.ts
DanielMSchmidt: Not necessary, but it would be removing unnecessary information, making it more concise

GeorgiSTodorov: Then we have to remove the type of all other `const`s like `DCOS_UI_VERSION_SETTING` as well.

src/js/core/UpdateStream.ts
DanielMSchmidt: The problem is that you are doing `savedStates[DCOS_UI_VERSION_SETTING] =  ...`. As some mutating method could have changed `DCOS_UI_VERSION_SETTING` typescript tries to tell you that there could be every key accessed. If you would change the line to `savedStates.dcosUIVersion` there would not be an error

src/js/core/UpdateStream.ts
DanielMSchmidt: Typescript or tslint?

src/js/core/UpdateStream.ts
```diff
@@ -0,0 +1,145 @@
...
+  dismissedVersion: Observable<string>,
+  newFetchedVersion: Observable<RequestResponse<{}>>
+): Observable<string> {
+  return Observable.timer(0, delay)
+    .switchMap(() =>
+      Observable.combineLatest(dismissedVersion, newFetchedVersion)
+    )
+    .filter(([dismissed, nextVersion]) => {
+      const newVersion = getVersionFromVersionObject(nextVersion);
+      setInLocalStorage(LAST_TIME_CHECK, new Date().getTime());
```
DanielMSchmidt: Having a side effect in a filter seems strange, it should rather be in a `.do` block

src/js/core/UpdateStream.ts
DanielMSchmidt: So let me summarize what typescript tells you:

You told typescript that the request always returns an object on which the response key is exactly an empty object (by using `Observable<RequestResponse<{}>>`) and now you try to access on this empty object a result key, which is not there for the empty object, therefore the error.

If you set the `Observable<RequestResponse<{}>>` to something closer to the truth, maybe `Observable<RequestResponse<CosmosListResponse>>` and you give the `CosmosListResponse` a key attribute that is structured the same way as the HTTP response is then you should be better of. Keep in mind that if you don't know if a key is really always present you should add a ? `key?: number` or a null as union type `key: number | null`, to ensure that the type assertions are always true. After that, you need to alter the code in the map to be clearer with which types are always there and which ones are not.

If you find it confusing to work with typescript in observables I can recommend to use full functions with type signatures for every map, that way you can be sure of which type the outcome is.

src/js/core/UpdateStream.ts
DanielMSchmidt: Because the BehaviorSubject where you use this function only expects strings and no numbers. Either have this function return string or have the BehaviourSubject handle string | number.

In general, using the fewest amount, but the most specific types in a union type is a good rule of thumb

src/js/core/UpdateStream.ts
GeorgiSTodorov: Yes, but I wanted to make use of the constant.

src/js/core/UpdateStream.ts
GeorgiSTodorov: I don't like the solution of making both `BehaviorSubject`s `string | number`, but I don't see any other way, so I guess it's okay.

src/js/core/UpdateStream.ts
DanielMSchmidt: You can do a `// tslint:disable-next-line`

DanielMSchmidt: and therefore ignore the param being missing. or if it's unused just remove the function alltogether

src/js/core/UpdateStream.ts
GeorgiSTodorov: It makes it easier to change `"dcosUIVersion"`.

src/js/core/UpdateStream.ts
GeorgiSTodorov: It can still be reviewed separately, it just won't have a visible impact in the UI.

src/js/core/UpdateStream.ts
GeorgiSTodorov: The function is useful for testing this PR and I think Mike also made use of it.

src/js/pages/Index.js
```diff
@@ -72,6 +73,7 @@ var Index = React.createClass({
 
     ConfigStore.fetchCCID();
     ConfigStore.addChangeListener(EventTypes.CONFIG_ERROR, this.onConfigError);
+    compare.subscribe(newVersion => console.log(newVersion));
```
natmegs: Maybe we could add a TODO comment here to clean up the log once this is handled. Also, we need to assign this line to a `Subscription` object so that we can `unsubscribe` somewhere (could be in `componentWillUnmount`)

GeorgiSTodorov: Good points, the `console.log` was just for testing and unsubscribing should be possible.

src/js/core/UpdateStream.ts
```diff
@@ -0,0 +1,158 @@
...
+  entity: string;
+  version: string;
+}
+
+interface DCOSUserSettings {
+  dcosUIVersion: DCOSUIVersion;
+}
+
+interface DCOSUIVersion {
+  dismissedVersion: string | number | null;
```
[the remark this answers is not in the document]
GeorgiSTodorov: Because `getFromLocalStorage` need to return a string or a number, since I use it for both the version and the last time check. I added a potential string type change here - https://github.com/dcos/dcos-ui/pull/3398/files/b3ddf954ad5b7083f5b0d7a79652ad5b4632de89#diff-f53b1374c8ef3fe10aefc772e8eef6ccR146 and I couldn't break it while testing it.

GeorgiSTodorov:
I decided to remove the `subscribe` from `pages/Index.js`, because it was breaking integration tests for some reason and it will probably be used in a different way in the final version, judging by @mperrotti 's work in progress branch - https://github.com/dcos/dcos-ui/compare/gt/feat/DCOS-41419-check-dcos-ui-update...mp/feat/DCOS-41418-ui-upgrade-banner?expand=1 . I also updated the testing instructions to reflect this.

src/js/core/UpdateStream.ts
```diff
@@ -0,0 +1,158 @@
...
+  newFetchedVersion: Observable<VersionObject>
+): Observable<string | number | null> {
+  return Observable.timer(0, delay)
+    .switchMap(() =>
+      Observable.combineLatest(dismissedVersion, newFetchedVersion)
+    )
+    .filter(([dismissed, nextVersion]) => {
+      const newVersion = getVersionFromVersionObject(nextVersion);
+
+      return compareVersions(newVersion, String(dismissed)) === 1;
```
natmegs: If the dismissed version is `null` (which is currently allowed by our types), this errors with `Uncaught Error: Invalid argument not valid semver ('null' received)`. I think we either need to enforce that this is not null in the typing or do something like `return dismissed ? compareVersions(newVersion, String(dismissed)) === 1 : true;`

natmegs: Also even if the dismissed version is a string but `not valid semver` this will error

GeorgiSTodorov: You are right. When I wrote it like that I was thinking that `String(null)` is `"null"`, but I didn't consider that `comapreVersions` expect a specific string.

GeorgiSTodorov:
@natmegs , thank you for the feedback. I decided to address this issue with a try and catch block.

PR #3407, opened 2018-11-13
feat(jobs): adding support for UCR engine

wavesoft:
This commit introduces support for the job syntax from metronome where the user
can switch between UCR and Docker runtime engines

Closes DCOS_OSS-4436

## Testing

* This commit adds the required tests to acommodate changes to `Job` struct

## Trade-offs

* This still uses the old schema form that should be really removed.
* The `Docker Parameters` should be disabled when the user enables UCR
* The `Privileged` checkbox has to be disabled when the user enables UCR

## Dependencies

* The respective metronome PRs must be merged
  1. https://github.com/dcos/metronome/pull/292
  2. https://github.com/dcos/metronome/pull/291
  3. https://github.com/dcos/metronome/pull/290

## Screenshots

ucr mov


kensipe:
this build / test fail.. 
I don't think this is the experience we will want long term...   
thanks for the screenshot gif...   What happens if there are defined docker parameters and you click the UCR checkbox?

meichstedt:
Wondering … why is this targeting `release/1.12`? The feature in Metronome will be added for DC/OS 1.13.

nLight:
@meichstedt our master was broken, I suggested to target release/1.12 for the convo starter PR. I saw your line up for changes. We absolutely have to refactor the form otherwise it won't handle those updates really well.

nLight:
We ended up refactoring the form. Thanks for the effort!

PR #3410, opened 2018-11-14
Revert "DCOS-43779: Move webpack output to assets directory, except index.html"

nLight:
Reverts dcos/dcos-ui#3405

this breaks webpack server

PR #3412, opened 2018-11-15
DCOS-43779: Move webpack output to assets directory, except index.html

nLight:
Reverts dcos/dcos-ui#3410

When testing let's test both dev and prod setup + make sure your browser cache is busted.


GeorgiSTodorov:
No icons when running `npm start`.

screenshot from 2018-11-16 11-31-43

brandonc:
The [hash] bundle filename doesn't appear to work in webpack-dev-server when used in combination with some of this webpack config. This gives Chrome an excuse to load previously cached bundles named "index.js"

I modified webpack config to only output a nested folder structure in production builds. By default, everything else is still flat. This made "sprites.svg" resolution different. I need to be able to read "publicPath" config at runtime in order to avoid the condition in index.js

The missing icons are due to the fact that we route those image requests through the cluster (but only on navigation, not on initial page load?!), which creates a mixed content http/https problem in localhost, so I did not try to solve that. Shouldn't be a blocking issue for this change

PR #3413, opened 2018-11-15
DCOS-42063: add support for field resolvers to graphql observable

DanielMSchmidt:
## Testing

See if the tests are handling every important scenario you see

## Trade-offs

- Decided against passing down the info object, as we currently don't need it and I see no point in implementing it.

### DIFF
--- a/packages/data-service/src/__tests__/graphqlObservable-test.js
+++ b/packages/data-service/src/__tests__/graphqlObservable-test.js
@@ -16,15 +16,15 @@ const typeDefs = `
   type Shuttle {
     name: String!
   }
-  
+
   type Query {
     launched(name: String): [Shuttle!]!
   }
-  
+
   type Mutation {
     createShuttle(name: String): Shuttle!
     createShuttleList(name: String): [Shuttle!]!
-  }  
+  }
 `;
 
 const mockResolvers = {
@@ -66,6 +66,106 @@ const mockResolvers = {
 const schema = makeExecutableSchema({
   typeDefs,
   resolvers: mockResolvers
+});
+
+const fieldResolverSchema = makeExecutableSchema({
+  typeDefs: `
+    type Plain {
+      noFieldResolver: String!
+      fieldResolver: String!
+      giveMeTheParentFieldResolver: String!
+      giveMeTheArgsFieldResolver(arg: String!): String!
+      giveMeTheContextFieldResolver: String!
+    }
+
+    type ObjectValue {
+      value: String!
+    }
+
+    type Item {
+      nodeFieldResolver: ObjectValue!
+      giveMeTheParentFieldResolver: ObjectValue!
+      giveMeTheArgsFieldResolver(arg: String!): ObjectValue!
+      giveMeTheContextFieldResolver: ObjectValue!
+    }
+
+    type Nested {
+      firstFieldResolver: Nesting!
+    }
+
+    type Nesting {
+      noFieldResolverValue: String!
+      secondFieldResolver: String!
+    }
+
+    type Query {
+      plain: Plain!
+      item: Item!
+      nested: Nested!
+    }
+  `,
+  resolvers: {
+    Plain: {
+      fieldResolver() {
+        return Observable.of("I am a field resolver");
+      },
+      giveMeTheParentFieldResolver(parent) {
+        return Observable.of(JSON.stringify(parent));
+      },
+      giveMeTheArgsFieldResolver(_parent, args) {
+        return Observable.of(JSON.stringify(args));
+      },
+      giveMeTheContextFieldResolver(_parent, _args, context) {
+        return Observable.of(context.newValue);
+      }
+    },
+    Item: {
+      nodeFieldResolver() {
+        return Observable.of({ value: "I am a node field resolver" });
+      },
+      giveMeTheParentFieldResolver(parent) {
+        return Observable.of({ value: JSON.stringify(parent) });
+      },
+      giveMeTheArgsFieldResolver(_parent, args) {
+        return Observable.of({ value: JSON.stringify(args) });
+      },
+      giveMeTheContextFieldResolver(_parent, _args, context) {
+        return Observable.of({ value: context.newValue });
+      }
+    },
+    Nested: {
+      firstFieldResolver(_parent, _args, ctx) {
+        ctx.contextValue = " resolvers are great";
+
+        return Observable.of({ noFieldResolverValue: "nested" });
+      }
+    },
+    Nesting: {
+      secondFieldResolver({ noFieldResolverValue }, _, { contextValue }) {
+        return Observable.of(
+          noFieldResolverValue.toLocaleUpperCase() + contextValue
+        );
+      }
+    },
+    Query: {
+      plain(_parent, _args, ctx) {
+        ctx.newValue = "ContextValue";
+
+        return Observable.of({
+          noFieldResolver: "Yes"
+        });
+      },
+      item(_parent, _args, ctx) {
+        ctx.newValue = "NodeContextValue";
+
+        return Observable.of({ thisIsANodeFieldResolver: "Yes" });
+      },
+
+      nested() {
+        return Observable.of({});
+      }
+    }
+  }
 });
 
 // jest helper who binds the marbles for you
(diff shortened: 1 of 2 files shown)


packages/data-service/src/__tests__/graphqlObservable-test.js
```diff
@@ -192,6 +292,238 @@ describe("graphqlObservable", function() {
...
+          `;
+          const expected = m.cold("(a|)", {
+            a: { data: { plain: { noFieldResolver: "Yes" } } }
+          });
+          const result = graphqlObservable(query, fieldResolverSchema, {});
+          m.expect(result.take(1)).toBeObservable(expected);
+        });
+
+        itMarbles("if defined it executes the field resolver", function(m) {
+          const query = gql`
```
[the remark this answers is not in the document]
DanielMSchmidt: I was unaware of this problem, we have it in general, but having field resolvers makes it more prominent. We combine all data collected on every layer with a `combineLatest`, which only emits once every item emitted. So from what I understand, we would either need to find something else or concat the result of the resolver with an observable that returns null.

Either way, the implication is that we can no longer assume any level of the graphql query to never be null (which is fine I would say).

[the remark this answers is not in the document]
nLight: @DanielMSchmidt you are right. As I understood from the Apollo docs they would bubble up `null` to the first nullable type. I wonder if we could declare all deferred fields as nullable, would it make sense? Did I just say something completely outlandish? :)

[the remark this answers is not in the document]
DanielMSchmidt: Yes, that makes sense

packages/data-service/src/graphqlObservable.ts
```diff
@@ -0,0 +1,286 @@
...
+type SchemaNode = SelectionNode | DefinitionNode;
+
+function isOperationDefinition(node: any): node is OperationNode {
+  return node.kind === "OperationDefinition";
+}
+function isFieldNode(node: SchemaNode): node is FieldNode {
+  return node.kind === "Field";
+}
+
+function isDefinitionNode(node: SchemaNode): node is DefinitionNode {
```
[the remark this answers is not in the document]
DanielMSchmidt: needed to update graphql for that, but works almost fine (we use OPERATION_DEFINITION as entry point, so our isDefinitionNode needs to be differently defined

packages/data-service/src/graphqlObservable.ts
```diff
@@ -0,0 +1,286 @@
...
+
+interface TypeMap {
+  [key: string]: GraphQLType;
+}
+
+interface Schema extends GraphQLSchema {
+  _typeMap: TypeMap;
+}
+
+interface OperationNode {
```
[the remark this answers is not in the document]
DanielMSchmidt: We don't really need subscriptions and don't deal with them in all other places, so this is out of scope I would say. Also it is to be discussed if adding sth that conveys two levels of "real-timeiness" is a good idea for us.

DanielMSchmidt:
We scoped out the issue that Dima pointed out here: https://github.com/dcos/dcos-ui/pull/3413#discussion_r234402659. This is the issue: https://jira.mesosphere.com/browse/DCOS-45517

package.json
```diff
@@ -45,7 +45,7 @@
     "d3": "3.5.16",
     "dcos-dygraphs": "1.1.0-beta.3",
     "flux": "2.1.1",
-    "graphql": "0.13.2",
+    "graphql": "14.0.2",
```
nLight: ❤️

packages/data-service/src/graphqlObservable.ts
```diff
@@ -0,0 +1,286 @@
...
+        // and we pass in the wrong values as we don't need this feature currently
+        // @ts-ignore
+        return getNamedType(fieldType.resolveType(instance));
+      } else {
+        return fieldType;
+      }
+    }
+
+    // No field could be found, this seems to be an error
+    // No composition type so the parent is a scalar. What the fuck are we even doing here?
```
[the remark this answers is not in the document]
nLight: ping :)

packages/data-service/src/graphqlObservable.ts
```diff
@@ -0,0 +1,286 @@
...
+      } else {
+        resolvedObservable = Observable.of(parent[field.name]);
+      }
+
+      if (!resolvedObservable) {
+        return throwObservable("resolver returns empty value");
+      }
+
+      if (!(resolvedObservable instanceof Observable)) {
+        return throwObservable("resolver does not return an observable");
```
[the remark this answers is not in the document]
nLight: Would you please capture our discussion in a jira? Since it's been that way in the previous implementation let's extract the feature request.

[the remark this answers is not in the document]
DanielMSchmidt: I mean, it's done now, right? What would be in the JIRA?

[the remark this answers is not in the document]
nLight: Allowing resolvers to return scalar values. Did I miss anything?

packages/data-service/src/graphqlObservable.ts
```diff
@@ -0,0 +1,317 @@
...
+        return getNamedType(fieldType.resolveType(instance));
+      } else {
+        return fieldType;
+      }
+    }
+
+    return null;
+  }
+
+  return resolve(doc.definitions[0], context, null, null).map((data: T) => ({
```
brandonc: I'd like to see this expression moved to line 80 to help readability. Function hoisting allows it to appear before `function resolve`

packages/data-service/src/graphqlObservable.ts
```diff
@@ -0,0 +1,317 @@
...
+    definition: SchemaNode,
+    context: object,
+    parent: any,
+    type: GraphQLType | null
+  ): Observable<any> {
+    if (isDefinitionNode(definition)) {
+      return throwObservable("Definition types should not be present here");
+    }
+
+    if (definition.kind === Kind.FRAGMENT_SPREAD) {
```
brandonc: I appreciate all the new assertions sprinkled in here

packages/data-service/src/graphqlObservable.ts
```diff
@@ -0,0 +1,317 @@
...
+};
+
+function resolveField(
+  field: GraphQLField<any, any, { [argName: string]: any }>,
+  definition: FieldNode,
+  context: object,
+  parent: any
+) {
+  if (!isFieldWithResolver(field)) {
+    return Observable.of(parent[field.name]);
```
brandonc: I wonder if it would simplify things slightly to provide this as a default field resolver `parent => Observable.of(parent[field.name])`

so instead of defining and checking `isFieldWithResolver(field)` you could define `getFieldResolver(field)` which returns `field.resolver || parent => Observable.of(parent[field.name])`

packages/data-service/src/graphqlObservable.ts
```diff
@@ -0,0 +1,317 @@
...
+        context,
+        parent
+      );
+
+      if (!resolvedObservable) {
+        return throwObservable("resolver returned an empty value");
+      }
+
+      if (!(resolvedObservable instanceof Observable)) {
+        return throwObservable("resolver does not return an observable");
```
brandonc: I appreciate all the new assertions sprinkled in here. But these two checks (111, 115) are unreachable as written because resolveField always returns an Observable

packages/data-service/src/graphqlObservable.ts
DanielMSchmidt: I think we are going to have a similar problem with polymorphisms, I would like to see which version is clearer in the end. I hope this is okay for you 😇

DanielMSchmidt:
Thank you @GeorgiSTodorov

DanielMSchmidt:
@brandonc @natmegs @nLight @GeorgiSTodorov You can take another look now, I implemented some of the review comments 👍

packages/data-service/src/graphqlObservable.ts
```diff
@@ -0,0 +1,309 @@
...
+  const types = schema._typeMap;
+
+  return resolve(doc.definitions[0], context, null, null).map((data: T) => ({
+    data
+  }));
+
+  function resolve(
+    definition: SchemaNode,
+    context: object,
+    parent: any,
```
nLight: Is it possible to be a bit more specific here?

packages/data-service/src/graphqlObservable.ts
```diff
@@ -0,0 +1,309 @@
...
+// We don't treat OperationDefinitions as Definitions but as entry points for our execution
+function isDefinitionNode(node: SchemaNode): node is DefinitionNode {
+  return (
+    node.kind === Kind.FRAGMENT_DEFINITION ||
+    isTypeSystemDefinitionNode(node) ||
+    isTypeSystemExtensionNode(node)
+  );
+}
+
+interface FieldWithResolver extends GraphQLField<any, any> {
```
nLight: Can we pin point some of those `any`?

packages/data-service/src/graphqlObservable.ts
DanielMSchmidt: Unfortunately, as far as I understand TS and the problem I would say no.

This can in an abstract case be any object, any array, or any primitive value.
Once we update typescript we could and should move it to the newly created `unknown` type, but for now any is all there is.

To be more precise and work with generics I think we would need to have the schema at compile time somehow assembled into TS, but also then there are a lot of questions to be answered.

packages/data-service/src/graphqlObservable.ts
```diff
@@ -0,0 +1,309 @@
...
+        }
+
+        const nextType = getResultType(type, definition, emitted);
+        return resolveResult(definition, context, emitted, nextType);
+      });
+    }
+
+    // It is no operationDefinitionand no fieldNode, so it seems like an error
+    // TODO: throw better error here
+    return throwObservable("Input was strange");
```
nLight: Can we just say `Input doesn't look like operationDefinitionand nor fieldNode`?

DanielMSchmidt: yes, I completely overlooked this message when going through the code again, thanks for pointing it out 👍

DanielMSchmidt: done 👍

PR #3415, opened 2018-11-22
[1.12] DCOS-45494: fix truncated columns on DeclinedOffersTable

natmegs:
[COPS ticket](https://jira.mesosphere.com/browse/COPS-4160) reported that the Host IP addresses were being truncated, but the Constraint header as well as the Received content was also being truncated. Fix Host, Constraint, and Received columns which were being truncated with content cut off. Added a tooltip to the Host column.

Closes DCOS-45494.

## Testing

- Use a 1.12 soak cluster and go to Services 
- Choose a pod that's deploying 
- Debug tab 
- Look at table under Recent Resource Offers > Details and check that above mentioned columns are not truncated

## Trade-offs

None that I'm aware of

## Dependencies

None

## Screenshots

Before
screen shot 2018-11-22 at 1 29 49 pm

After
screen shot 2018-11-22 at 1 25 08 pm

### DIFF
--- a/plugins/services/src/js/components/DeclinedOffersTable.js
+++ b/plugins/services/src/js/components/DeclinedOffersTable.js
@@ -157,15 +157,16 @@ class DeclinedOffersTable extends React.Component {
   getColGroup() {
     return (
       <colgroup>
+        <col style={{ width: "135px" }} />
+        <col style={{ width: "9%" }} />
+        <col style={{ width: "15%" }} />
         <col />
-        <col style={{ width: "10%" }} />
-        <col style={{ width: "10%" }} />
-        <col className="hidden-large-up" style={{ width: "30%" }} />
-        <col className="hidden-medium-down" style={{ width: "10%" }} />
-        <col className="hidden-medium-down" style={{ width: "10%" }} />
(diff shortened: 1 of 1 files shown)


natmegs:
@nLight it was a trivial addition so I added the tooltip on the Host column (also I didn't see your second comment until after I added it haha)

DanielMSchmidt:
@nLight I think this would not be scope creep as the problem is basically not being able to see the whole hostname. Making the column wider helps with the most common use case of ipv4 addresses, but as these hostnames can be set (very rarely by some clients) in an arbitrary way making sure we can see them anyways is crucial to a long-term fix 👍

nLight:
It still can be squished, at least in Firefox. I wonder if we could make the column width static so that it fits `255.255.255.255` 

screenshot 2018-11-26 at 12 41 07

natmegs:
@nLight I added a static width to that column which ensures that `255.255.255.255` is always visible (for me) on chrome and firefox. Can quickly test by replacing https://github.com/dcos/dcos-ui/pull/3415/files#diff-dcc68da93de9ef7f8b03785364d96186R190 with `255.255.255.255`

PR #3416, opened 2018-11-23
DCOS-45393: Add GPU to Nodes Detail Page

GeorgiSTodorov:
Added GPU information to the node detail tab table.

Closes DCOS-45393

## Testing
Go to Nodes tab.
Click a node.
Open the Details tab.
Verify that the GPU row is there.

## Trade-offs
Unless I am missing something, I think this should be a 1-story point ticket.

## Dependencies
None.

### Before
screenshot from 2018-11-23 11-46-45

### After
screenshot from 2018-11-23 11-47-05

### DIFF
--- a/plugins/nodes/src/js/pages/nodes/NodeDetailTab.js
+++ b/plugins/nodes/src/js/pages/nodes/NodeDetailTab.js
@@ -144,6 +144,14 @@ class NodeDetailTab extends PureComponent {
                 {Units.formatResource("cpus", resources.cpus)}
               </ConfigurationMapValue>
             </ConfigurationMapRow>
+            <ConfigurationMapRow>
+              <ConfigurationMapLabel>
+                <Trans render="span">GPUs</Trans>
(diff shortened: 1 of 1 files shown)


nLight:
@GeorgiSTodorov how many SP would then a typo change be? :)

GeorgiSTodorov:
@nLight , the limit function of a typo is approaching zero.

nLight:
@GeorgiSTodorov I'd actually use fixing typo as a reference for 1sp bumping everything else up. That's ok to have 13 or 21sp issues if that's team's scale

GeorgiSTodorov:
Yeah, I was half-joking, but if this is 2, then for example https://jira.mesosphere.com/browse/DCOS-41419 should be higher than 3 I think.

nLight:
@GeorgiSTodorov that is the message I'm trying to convey on the grooming sessions

GeorgiSTodorov:
@nLight , okay, I agree with you.

PR #3417, opened 2018-11-23
chore: add beefier fixtures

nLight:
## Testing

1. run UI as usual
2. Login into the UI (it won't work with fixtures enabled)
3. in your `Config.dev.ts` set `useFixtures: true`

## Trade-offs

To be able overriding `request` function I imported the whole library and then destructed `{request}` from it. We could put `@dcos/mesos-client` to the IoC container instead, but I think that would be an overkill. We should come up with a better fixture strategy anyway. [Mockserver](https://github.com/mesosphere/mockserver) is a possible solution


GeorgiSTodorov:
Testing is setting `useFixtures: true` and seeing if the UI works?

nLight:
@GeorgiSTodorov basically yes. Verifying that you see thousands of nodes/services

GeorgiSTodorov:
I don't think it works as it should for me.
I see only 2 nodes, 2 services, and once I log out I get a 404 error.
screenshot from 2018-11-30 13-41-27

screenshot from 2018-11-30 13-41-41

screenshot from 2018-11-30 13-41-55

src/js/stores/MesosStateStore.js
```diff
@@ -304,4 +306,27 @@ class MesosStateStore extends GetSetBaseStore {
...
+  const getMasterFixture = require("../../../tests/_fixtures/v1/get_master.json");
+  const subscribeFixture = require("../../../tests/_fixtures/v1/subscribe.js");
+
+  const { MesosStreamType } = require("../core/MesosStream");
+
+  const oldRequest = request;
+
+  request = function(options, url) {
+    if (url === "/mesos/api/v1?get_master") {
+      console.log("Stub /mesos/api/v1?get_master");
```
DanielMSchmidt: Is this a wanted console.log?

nLight: it's a fixture mode only, we can totally remove the log

PR #3419, opened 2018-11-26
DCOS-38853: replace pure-render-mixin with PureComponent

nLight:
as the title suggests. I rewrote the NodesGridView in TS since it was the only component I would have to implement `shouldComponentUpdate` the other way around.

Closes DCOS-38853

## Testing

You may consider smoke testing related pages

## Trade-offs

Not entirely sure with the type definition I wrote for the `NodesGridDials`. `tslint` is fine but vscode tells me I should have a top level `declare`.


plugins/nodes/src/js/components/NodesGridView.tsx
```diff
@@ -1,49 +1,52 @@
...
+interface NodesGridViewProps {
+  hasLoadingError?: boolean;
+  hiddenServices?: [];
+  hosts: NodesList;
+  receivedEmptyMesosState?: boolean;
+  receivedNodeHealthResponse?: boolean;
+  resourcesByFramework: ResourcesByFramework;
+  selectedResource: string;
+  serviceColors: ServiceColors;
+  services: any[];
```
brandonc: Still new to typescript. Could this be object[]? To at least constrain it that much.

plugins/nodes/src/js/components/NodesGridDials.d.ts
```diff
@@ -0,0 +1,12 @@
+import { ClassicComponentClass } from "react";
+
+interface NodesGridDialsProps {
+  hosts: any[];
+  resourcesByFramework: object;
+  selectedResource: string;
+  serviceColors: { [index: string]: number };
+}
+
+const NodesGridDials: ClassicComponentClass<NodesGridDialsProps>;
```
natmegs: The error here is just typescript complaining because it requires `declare` for any top level non-interface declarations in .d.ts files. Error goes away with `declare const NodesGridDials: ClassicComponentClass<NodesGridDialsProps>;`

plugins/nodes/src/js/components/NodesGridView.tsx
```diff
@@ -1,49 +1,52 @@
...
-  defaultProps: {
+const MAX_SERVICES_TO_SHOW = 32;
+const OTHER_SERVICES_COLOR = 32;
+
+interface ServiceColors {
+  [index: string]: number;
+}
+
+interface ResourcesByFramework {
+  [index: string]: object;
```
natmegs: It would be cool if we could really take advantage of types here and define what each resource object would look like but that's starting to smell like scope creep....

GeorgiSTodorov:
Not merging this in case @nLight wants to change anything based on the feedback.

nLight:
@brandonc @natmegs thank you very much for the feedback, please give me a harder time 🙇
@GeorgiSTodorov thanks for not rushing it in 💪

nLight:
I introduced a small bug so integration tests were failing. Fixed 🙇

PR #3422, opened 2018-11-27
chore(messages.json): update translation messages

natmegs:
Update translation catalogs to accompany LDAP anonymous bind feature. (See [DCOS-43815](https://jira.mesosphere.com/browse/DCOS-43815))

## Testing

There should be no visual changes as a result of this PR.

## Trade-offs

N/A.

## Dependencies

This should be merged with mesosphere/dcos-ui-plugins-private#882

### DIFF
--- a/locale/zh/messages.json
+++ b/locale/zh/messages.json
@@ -244,7 +244,7 @@
     "origin": [
       [
         "src/js/components/CosmosErrorMessage.js",
-        38
+        39
       ]
     ]
   },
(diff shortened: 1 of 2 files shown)


locale/zh/messages.json
```diff
@@ -12618,6 +12640,15 @@
       ]
     ]
   },
+  "{runningInstances, plural, one {# instance running out of {instancesTotal}} other {# instances running out of {instancesTotal}}}": {
```
brandonc: Does this belong? Was it another forgotten extraction?

natmegs: Unless I'm missing something I think this goes with https://github.com/dcos/dcos-ui/blob/90853ef9285377a235d11eb428888a47d79b44de/plugins/services/src/js/components/ServiceStatusProgressBar.js#L22-L27

PR #3423, opened 2018-11-30
DCOS-41719: localize dates

natmegs:
Localize dates using DateFormat macro. Does not include relative dates.

Closes DCOS-41719.

## Testing

There should be no visual changes as a result of this PR.

## Trade-offs

Lingui `DateFormat` macro uses `Intl.DateTimeFormat` instead of `Date` functions like `toLocaleString` (not so much a tradeoff as just something to be aware of).

## Dependencies

None

## Screenshots

N/A

### DIFF
--- a/plugins/nodes/src/js/components/LeaderGrid.js
+++ b/plugins/nodes/src/js/components/LeaderGrid.js
@@ -9,7 +9,7 @@ import ConfigurationMapRow from "#SRC/js/components/ConfigurationMapRow";
 import ConfigurationMapSection from "#SRC/js/components/ConfigurationMapSection";
 import ConfigurationMapValue from "#SRC/js/components/ConfigurationMapValue";
 import Loader from "#SRC/js/components/Loader";
-import { msToRelativeTime } from "#SRC/js/utils/DateUtil";
+import DateUtil from "#SRC/js/utils/DateUtil";
 
 const Loading = () => <Loader size="small" type="ballBeat" />;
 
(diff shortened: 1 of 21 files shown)


plugins/services/src/js/containers/pod-instances/__tests__/PodInstancesTable-test.js
```diff
@@ -403,10 +425,25 @@ describe("PodInstancesTable", function() {
             return el.text().trim();
           });
 
+        const options = {
```
nLight: Shall we extract those options? Maybe creating a wrapper component that would have them as a default?

plugins/nodes/src/js/components/LeaderGrid.js
```diff
@@ -9,7 +9,7 @@ import ConfigurationMapRow from "#SRC/js/components/ConfigurationMapRow";
 import ConfigurationMapSection from "#SRC/js/components/ConfigurationMapSection";
 import ConfigurationMapValue from "#SRC/js/components/ConfigurationMapValue";
 import Loader from "#SRC/js/components/Loader";
-import { msToRelativeTime, msToDateStr } from "#SRC/js/utils/DateUtil";
```
nLight: Shall we remove `msToDateStr` from `DateUtil` since we removed all the usages?

DanielMSchmidt: That would be so cool 😻

natmegs:
Made a few changes to address feedback and I've removed the changes to the translation catalogs for now. They need to be done last so I will add once feedback has been finalized.

GeorgiSTodorov:
I get this error
screenshot from 2018-12-03 11-48-11
and it is fixed if I change `import DateUtil from "#SRC/js/utils/DateUtil";` to `import DateUtil from "#SRC/js/utils/DateUtil.ts";`. Does anyone else have this problem?

nLight:
@GeorgiSTodorov doesn't happen to me, hm..

nLight:
@GeorgiSTodorov did you test with EE PR checked out?

GeorgiSTodorov:
@nLight , good point, it works now, even without using the other PR. Not sure what broke it before.

PR #3424, opened 2018-11-30
DCOS-21600: show other implementation for linting

DanielMSchmidt:
This is the intended usage, which is AFAIK impossible due to our jenkins clone setup


DanielMSchmidt:
@nLight don't you mean `"... ${env.CHANGE_BRANCH}"`?

nLight:
no_idea_what_iam_doing.jpeg :D I copied that approach from below lines

DanielMSchmidt:
If you see the error ` ambiguous argument 'master..HEAD': unknown revision or path not in the working tree.
`

It points out that `master` is an unknown revision. the value `master` is passed correctly.

nLight:
did you try the CHANGE_BRANCH?

DanielMSchmidt:
I understood CHANGE_BRANCH would be for this PR `danielmschmidt/DCOS-21600-show-other-implementation`, which is not what I would want. I could use it in the `--to` flag to be clearer though. But then I have two branches suffering from the same "what do you mean with master" error

nLight:
`ambiguous argument 'danielmschmidt/DCOS-21600-show-other-implementation..HEAD'` interesting
so what part is ambiguous I wonder? That should be just _one_ last commit, right?

DanielMSchmidt:
So what I did was I followed the exact same steps for setting up the repo that jenkins does. In the end I did `git log master`, which failed like this, `git log danielmschmidt/DCOS-21600`, which failed like this and `git log HEAD` which worked if I recall correctly. My gist is that somehow `master` and I think even `origin/master` is ambigious

nLight:
I wonder where does it point to?

DanielMSchmidt:
I wonder how to debug. I often feel good with git, but these rare occasions let me question all my knowledge and I feel like I don't know anything for certain anymore :D

PR #3435, opened 2018-12-06
Auto update typescript with Renovate

DanielMSchmidt:
## Testing

There is none, tbh. I would say if you have any doubts that our CI will catch all problems arising from this auto update please raise them.

## Trade-offs

- I did the manual update and decided against simply pushing it and instead use renovate. It's going to be two PRs, but in return we stay on the latest possible version.


GeorgiSTodorov:
CI should catch all possible breaking changes, right?

GeorgiSTodorov:
We currently use `2.9.2` and the latest version is `3.2.2`, we might have some breaking changes: https://github.com/Microsoft/TypeScript/wiki/Breaking-Changes .

DanielMSchmidt:
Yes, CI should catch them. I checked the update and it worked fine 👍

PR #3436, opened 2018-12-06
DCOS-45801: handle cosmos errors

nLight:
Cherry-picking #3431 

Just checking @natmegs I _should_ recompile the translations as I'm touching the code, right? Basically I should be recompiling them always?

### DIFF
--- a/locale/en/messages.json
+++ b/locale/en/messages.json
@@ -13,7 +13,7 @@
     "origin": [
       [
         "src/js/components/FilterHeadline.js",
-        69
+        67
       ]
     ]
   },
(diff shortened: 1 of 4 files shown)


natmegs:
Yes for any PR `npm run lingui-extract-with-plugins` needs to be run and any changes to `messages.json` (en and zh) should be committed. Even if text content is not changing line numbers can :)

locale/en/messages.json
```diff
@@ -109,42 +109,6 @@
       ]
     ]
   },
-  "1 Backend": {
```
DanielMSchmidt: That's a lot of deletion..

DanielMSchmidt: We should find a better solution for storing these, it seems to add quite a lot of accidental complexity

nLight: I run `lingui-extract-with-plugins` and this is what I get with no errors... @natmegs any idea? 🙏

nLight: Oh, gotcha... some adjustments to the library should be made. Very helpful wiki page!

PR #3440, opened 2018-12-07
DCOS-44594: COPS-3759 affecting dcos-ui: When no nodes are in the cluster the message should have different wording

GeorgiSTodorov:
Changed the message when there are no nodes to be more accurate.

Closes https://jira.mesosphere.com/browse/DCOS-44594

## Testing
1. In `plugins/nodes/src/js/pages/NodesAgents.js` change line 314 to be `if (true) {`
2. Click on Nodes tab.
3. Verify that the message shown is changed.

## Trade-offs
Also changed "a" to "are".

## Dependencies
None.

### Before
screenshot from 2018-12-07 10-51-54

### After
screenshot from 2018-12-07 10-51-18


nLight:
@GeorgiSTodorov we def. need updating the translation files.

GeorgiSTodorov:
@nLight , thanks, will do.

GeorgiSTodorov:
Updated translation message catalogs.

GeorgiSTodorov:
@natmegs , do you know what is the policy when we change a message? My file conflict is because the new message doesn't have a translation.

GeorgiSTodorov:
Updated.

GeorgiSTodorov:
Updated message catalog format, please take another look when you have time.

PR #3441, opened 2018-12-07
DCOS-46053: Metronome client to stop sending Content-Type header with GET requests

nLight:
As this behaviour breaks proxying requests. More background in the issue.

## Testing

1. Verify that `fetchJobs` and `fetchJobDetail` stopped sending Content-Type on respective pages.

## Trade-offs

Fixed a typo in a sep commit 🤷‍♂️


jaxesn:
Probably still worth doing, but the blocker status can come down.  They found some stale frameworks on the original issue and determined that was the real problem.

https://jira.mesosphere.com/browse/COPS-4216

DanielMSchmidt:
`ServiceAction` flaked, restarting CI

PR #3449, opened 2018-12-07
DCOS-20690: Nodes overview page filter by service is wrong

GeorgiSTodorov:
Change the label to be more accurate

Closes https://jira.mesosphere.com/browse/DCOS-20690

## Testing
Go to Nodes tab and verify that the dropdown has the new label.

## Trade-offs
https://github.com/dcos/dcos-ui/compare/gt/fix/DCOS-20690-nodes-filter-by-service?expand=1#diff-758096c192079a23867a134b9aa82082L4296 needs to be translated, not sure what the procedure is.

https://github.com/dcos/dcos-ui/compare/gt/fix/DCOS-20690-nodes-filter-by-service?expand=1#diff-fcd9383c392695ef2467eb64d8a78981R54 isn't marked for translation, but I think it should be.

## Dependencies
None.

### Before
screenshot from 2018-12-07 17-19-52

### After
screenshot from 2018-12-07 17-19-16


locale/zh/messages.json
```diff
@@ -4292,11 +4292,11 @@
       ]
     ]
   },
-  "Filter by Service": {
+  "Filter by Framework": {
     "translation": "按服务筛选",
```
jaxesn: We should follow up with our vendor on this and see how to handle stuff like this going forward.  Like if we send this back to them to translate new strings, would they notice we changed the english and update the chinese? Or should we remove the Chinese version to force it to be updated? 

@nLight can you see about getting a meeting or just an email contact with the vendor so we can ask?

locale/zh/messages.json
```diff
@@ -4292,11 +4292,11 @@
       ]
     ]
   },
-  "Filter by Service": {
+  "Filter by Framework": {
```
natmegs: If these were changed manually they will need to be reverted and updated using `npm run lingui-extract-with-plugins`. In this case, the entire "Filter by Service" translation should be removed and a new entry with "Filter by Framework" added with an empty translation, so that there is no confusion with the old translation.

mperrotti:
Is there a reason we can't mark this string for translation? https://github.com/dcos/dcos-ui/compare/gt/fix/DCOS-20690-nodes-filter-by-service?expand=1#diff-fcd9383c392695ef2467eb64d8a78981R54

GeorgiSTodorov:
@mperrotti , good catch, we probably forgot it. Will add it.

GeorgiSTodorov:
Updated.

GeorgiSTodorov:
Updated message catalog format, please take another look when you have time.

PR #3451, opened 2018-12-07
DCOS_OSS-2065: accept floating points in FrameworkConfigurationForm

mperrotti:
## Testing
2. Go to the "Nodes" tab

## Trade-offs

## Dependencies
None

## Screenshots
Before:
floatingpoint-broken

After:
floatingpoint-fixed


nLight:
I bet this needs to be done https://wiki.mesosphere.com/display/ENGINEERING/Extract+and+update+translation+messages

locale/zh/messages.json
```diff
@@ -12331,6 +12340,15 @@
       ]
     ]
   },
+  "seconds": {
+    "translation": "",
+    "origin": [
+      [
+        "plugins/services/src/js/containers/service-debug/TaskStatsTable.js",
```
nLight: 🤔 rebase?

mperrotti:
@nLight - I tried updating langpacks, but got no changes 🤔

nLight:
Oh, I know why! That file doesn't have _any_ translations does it?

mperrotti:
No, it appears to have translations...
Here are two:
https://github.com/dcos/dcos-ui/blob/867bf2f6f1471c4ec00938e508d5ff553d9b4338/src/js/components/FrameworkConfigurationForm.js#L103

https://github.com/dcos/dcos-ui/blob/867bf2f6f1471c4ec00938e508d5ff553d9b4338/src/js/components/FrameworkConfigurationForm.js#L338

GeorgiSTodorov:
After thinking about this and https://jira.mesosphere.com/browse/DCOS-45866 I think that adding some form of validation if not in this PR then at least in another jira will be a very good idea.

mperrotti:
@GeorgiSTodorov - we already have our own validation:
screen shot 2018-12-14 at 3 35 31 pm

`no-validate` is just telling the browser not to do it's own validation

And the value that gets sent is formatted with a period - you can see this by changing the number (while using a comma separated decimal) and looking at the `value` attribute of the `<input />`

AmrAbdelrazik:
@mperrotti  and @GeorgiSTodorov ,  to @jaxesn 's point, I thought we inherit the validation from the jsonschema, why do we have to do anything the form code for this specific field?

GeorgiSTodorov:
@AmrAbdelrazik , I am not that familiar with jsonschema, but I think it was HTML5 validation causing this.

AmrAbdelrazik:
Please ping @nLight for more information, all form fields in DC/OS catalog are auto generated from a JSON schema object that specifies the field and the required validation. I want to understand why this exact field is behaving differently from that behavior.

PR #3452, opened 2018-12-07
DCOS-42113: no duplicate carat dropdown on select inputs

mperrotti:
## Testing
Using a 1.12 cluster, check out `release/1.12`
Find a <select> input (for example, there's one on Services>Add Service>Single Container>Placement tab)

The select input should only have 1 down chevron, it should not also show the browser-native arrow/s

## Trade-offs
None

## Dependencies
None

## Screenshots
Before:
screen shot 2018-12-07 at 2 50 17 pm

After:
screen shot 2018-12-07 at 2 50 25 pm

### DIFF
--- a/postcss.config.js
+++ b/postcss.config.js
@@ -4,7 +4,7 @@ module.exports = ({ env }) => {
   return {
     parser: false,
     plugins: {
-      autoprefixer: shouldOptimize,
+      autoprefixer: true,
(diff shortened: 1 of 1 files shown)


mperrotti:
@jaxesn - yes. I asked Dima if it's worth going forward with the ticket, and he recommended we do it

PR #3453, opened 2018-12-07
[1.12] DCOS-46088: get status for all deployments

natmegs:
Ensure status is retrieved from deployments, even when getStatus method is not available.

Closes DCOS-46088.

## Testing

Not sure if this will hold up as deployments change over time, but for me I tested using the 1.12 soak cluster.
- Go to Services
- Click on "X Deployments" in top right corner
- List of deployments in modal should match X

Compare this to same set of steps on soak cluster (not using this branch, just soak url). For me, it was 3 deployments but only 2 would show up in the modal (+ errors in dev console).

## Trade-offs

Code trade-offs, none that I can see. But it is weird that the format of the deployment items is not consistent. This issue came up because most of the items had all the regular "Service" properties, but there was one that only had 
```js
{
deployment: {...},
isStale: true,
serviceId: "/spark/spark-history-kerberos"
}
```
which is why the `getStatus` method was causing errors. Also id stored as property `serviceId` instead of `id` as expected. Is this typical of any "stale" service/application?

## Dependencies

None

## Screenshots

Before:
screen shot 2018-12-07 at 11 51 22 am

After:
screen shot 2018-12-07 at 11 50 17 am

### DIFF
--- a/plugins/services/src/js/components/DeploymentsModal.js
+++ b/plugins/services/src/js/components/DeploymentsModal.js
@@ -437,10 +437,11 @@ class DeploymentsModal extends mixin(StoreMixin) {
       {});
     }
 
-    let statusText = item.getStatus();
+    let statusText = !item.isStale ? item.getStatus() : null;
+    const itemId = item.isStale ? item.serviceID : item.id;
 
-    if (currentActions[item.id] != null) {
-      statusText = currentActions[item.id];
+    if (currentActions[itemId] != null) {
+      statusText = currentActions[itemId];
     }
 
     return statusText;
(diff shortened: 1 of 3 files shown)


natmegs:
@jaxesn Agree that it is strange. I'm wondering if this is typical for "stale" deployments/services? I'm not sure how to get a service into a stale state to replicate, and I can't find any other examples in other soak clusters

AmrAbdelrazik:
Can we reach out to the team who owns the API (likely the Marathon/Orchestration team) and confirm with them.

natmegs:
Ok, this makes a lot more sense now. Stale services do not have `getStatus` because of how we compile the data in `getChildTableData`, so I've adjusted accordingly and added a couple integration tests.

nLight:
@jaxesn you might want to reconsider your change request :D Given that CI is green

jaxesn:
Merge before ci gets you!

PR #3454, opened 2018-12-07
DCOS-44502: DC/OS UI crashes after attempting to edit config of marathon-lb

TattdCodeMonkey:
When attempting to `writeErrors` we check that the `formData` is an object,
but if so we assume the schema has properties. But if the `formData` is an
Array then its an object and the schema does not have any properties.

Added a check that the schema has properties before iterating over
`formData` keys. To resolve this issue.

## Testing

- See [COPS-3907](https://jira.mesosphere.com/browse/COPS-3907)

## Trade-offs

This was the simplest fix. We could also check the `schema.type === "object"`

## Dependencies

None

## Screenshots

None


GeorgiSTodorov:
It works, but wouldn't it make sense to check here: https://github.com/dcos/dcos-ui/pull/3454/files#diff-e9f792f5f3db4a89c3a6c506697ad36eR150  -  `if (schema == null || !schema.properties)`?

GeorgiSTodorov:
We agreed to only require one review for backports, but I guess we made it to only require one review for merging to older versions (some of which aren't backports, like this one for example).

TattdCodeMonkey:
@GeorgiSTodorov  if we added the check to `if (schema == null)` then the code to write errors on line `166` would never get called.

TattdCodeMonkey:
@nLight 👍, tests added

TattdCodeMonkey:
@DanielMSchmidt _maybe_, I'll give it a try and see what the LOE is on the migration.

DanielMSchmidt:
@TattdCodeMonkey Fair enough ;)

PR #3456, opened 2018-12-10
DCOS-46127: mark flaky tests as success

DanielMSchmidt:
## Testing

- Open a PR with a failing integration test and see if it still makes the build red

## Trade-offs

- We won't have any information known to Jenkins about flakiness, except for the archived reports themselves, but flaky tests runs will be green, allowing us to have some air to breathe.


DanielMSchmidt:
`Couldn't get a success in 3 retries for pages/services/EditFrameworkConfiguration-cy.js`

PR #3457, opened 2018-12-10
DCOS_OSS-4502: service config dropdown content overflows

mperrotti:
## Testing
1. In `plugins/services/src/js/containers/service-configuration/ServiceConfiguration.js`, on line 143, change the `<DateFormat />` component to just be a span with a long string
2. Open a service from the "Services" tab
3. Go to the "Configuration" tab
4. Click the dropdown at the top of that screen

The long string you replaced `<DateFormat />` with should not overflow the container - it should appear truncated

## Trade-offs
None

## Dependencies
None

## Screenshots
Before:
screen shot 2018-12-10 at 3 42 01 pm

After:
screen shot 2018-12-10 at 3 45 04 pm

Note: I tested this in Chrome, FF, Safari, and Edge


GeorgiSTodorov:
@nLight , there are no changed lines in this PR, there is no need to update them (just in case I ran the command, but no changes were made).

mperrotti:
Good catch - thanks @nLight !

AmrAbdelrazik:
Why are we truncating, why can’t we juat extend the button to the right, and how is this useful now to the user after truncation

GeorgiSTodorov:
@AmrAbdelrazik , I am not a designer, but I am pretty sure that having a version that long is very rare and the full version can still be seen in the button that opens the dropdown (see Dima's screenshot for an example).

GeorgiSTodorov:
@nLight , please take another look.

PR #3464, opened 2018-12-11
ci: bump timeouts

nLight:

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -18,7 +18,7 @@ pipeline {
   }
 
   options {
-    timeout(time: 3, unit: "HOURS")
+    timeout(time: 4, unit: "HOURS")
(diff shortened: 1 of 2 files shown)


cypress.json
```diff
@@ -1,5 +1,5 @@
 {
-  "defaultCommandTimeout": 6000,
+  "defaultCommandTimeout": 60000,
```
nLight: it's 10x more for system tests and they're fairly stable...

Jenkinsfile
```diff
@@ -18,7 +18,7 @@ pipeline {
   }
 
   options {
-    timeout(time: 3, unit: "HOURS")
+    timeout(time: 4, unit: "HOURS")
```
nLight: just in case

PR #3465, opened 2018-12-11
CI: remove parallelity to increase stability

DanielMSchmidt:


DanielMSchmidt:
@nLight Integration tests are green on first try (this was started before any other improvement was merged). Do you think this might be worthwhile or is the price too high?

nLight:
Let's keep this one open for now. Eventually we would like to merge all the improvements. Waiting 1h more and not retrying is better than waiting less and retrying 50 times :D

nLight:
@GeorgiSTodorov I'd like to keep this ace as a last resort option. It prolongs the build time.

PR #3467, opened 2018-12-11
DCOS-46169: get status for all deployments

natmegs:
Ensure status is retrieved from deployments, even when getStatus method is not available. Changes ported from #3453. 

Closes DCOS-46169.

## Testing

I haven't found a way to reliably get a service into a stale state so the easiest way to test this is to ensure the integration tests pass.

## Trade-offs

See #3453.

## Dependencies

N/A


GeorgiSTodorov:
@nLight is right, I forgot about that.

natmegs:
@nLight Ooops, good catch :flushed: updated!

DanielMSchmidt:
@natmegs It seems like CI is red because of a failing test in deployment modal, could you take a look?

nLight:
It looks that `DeploymentsModal-cy.js` test is being flakey, would you please verify if we could proactively unflake it? Thanks

DanielMSchmidt:
How do you come to the conclusion it is flaky and not really broken @nLight?

natmegs:
@DanielMSchmidt @nLight Based on looking at the videos recorded in CI it looks like the tests always fail at the open modal step of the before each hook (but it happens irregularly - not always on the same test within the suite). I'll spend a bit of time to see if there's an easy change here otherwise I'll document in a JIRA.

PR #3468, opened 2018-12-11
[1.11] DCOS-46171: get status for all deployments

natmegs:
Ensure status is retrieved from deployments, even when getStatus method is not available. Changes ported from #3453.

Closes DCOS-46171.

## Testing

I haven't found a way to reliably get a service into a stale state so the easiest way to test this is to ensure the integration tests pass.

## Trade-offs

See #3453.

## Dependencies

N/A


GeorgiSTodorov:
Re-triggering tests again.

PR #3469, opened 2018-12-12
DCOS-46160: Change the format of lingui message catalog to po gettext

GeorgiSTodorov:
Change lingui message catalog format to .po.
This is the default in lingui 3 and will make changes easier.

Closes https://jira.mesosphere.com/browse/DCOS-46160

## Testing
Once we clarify the questions I have, the Chinese translation should be working I think.

## Trade-offs
Not sure if we should be deleting `messages.json` or keeping it.
Not sure how we can migrate the translations from `messages.json` to `messages.po`

## Dependencies
None.

## Screenshots
None.


nLight:
1. Delete them
2. https://lingui.js.org/ref/cli.html#cmdoption-extract-convert-from

GeorgiSTodorov:
@nLight , they are deleted, so I just have to keep them that way. And thanks for the link.

nLight:
Please also remove the metadata with line numbers/files. We've agreed to do this but it is not super explicit in the ticket description

nLight:
What's wrong with convert-from? We assumed that there's a config option for metadata removal.

GeorgiSTodorov:
Even when using `convert-from` no translations are copied (I tried everything I could think of, from lingui to po, from lingui to minimal, from minimal to lingui, different options, etc.). And for the metadata lingui has the "minimal" format, which looks like
```
{
   "MessageID": "Translated Message"
}
```
but if we use it, we won't use po.

nLight:
@GeorgiSTodorov I wonder if there's another tool/online converter

nLight:
I think it's a bug in lingui :) or rather a contribution opportunity shall we say. They pass the argument but then don't use it 🤷‍♂️ https://github.com/lingui/js-lingui/blob/master/packages/cli/src/lingui-extract.js

GeorgiSTodorov:
This is interesting, because judging from this issue: https://github.com/lingui/js-lingui/issues/140 it worked before at least from minimal to lingui.

GeorgiSTodorov:
I experimented with it, but couldn't fix it. It really receives it, but doesn't use it.
Judging from https://github.com/lingui/js-lingui/issues/334 "* [x]  #347 fix `--convert-from` option for converting formats", it should be fixed in version 3.

brandonc:
I merged #3473

PR #3472, opened 2018-12-13
fix: update translations

juliangieseke:


locale/en/messages.json
```diff
@@ -892,15 +892,6 @@
       ]
     ]
   },
-  "Anonymous Bind": {
```
natmegs: Was your plugins-private repo pulled with the latest changes when you extracted? The bind translations should be included (they are from LDAP feature which has been merged)

package.json
```diff
@@ -34,7 +34,6 @@
     "@dcos/mesos-client": "0.2.5",
     "@dcos/tslint-config": "0.1.0",
     "@dcos/ui-kit": "1.17.1",
-    "@lingui/react": "2.7.0",
```
natmegs: I believe this needs to be included in dependencies rather than devDependencies

locale/en/messages.json
```diff
@@ -893,7 +893,6 @@
     ]
   },
   "Anonymous Bind": {
-    "translation": "",
```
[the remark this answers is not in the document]
natmegs: hm interesting, I haven't seen this happen before. I think it should be harmless but all the other messages in this file have this `"translation": ""` line. Maybe this is new in the updated lingui patch version

PR #3473, opened 2018-12-13
WIP: chore(messages): convert messages to minimal format from lingui

jaxesn:
lingui doesnt properly convert between these formats, even though it has
a convert-from parameter. I manually edited the lingui-cli code to properly
read the existing messages files using the proper formatter and then write
them with the new format(minimal).


GeorgiSTodorov:
Can you please show the changes you made to the code to enable it to work?

jaxesn:
In ./node_modules/@lingui/cli/lingui-extract.js

Add at Line 63:

```
  var format = config.format;
  if(options.prevFormat){
    config.format = options.prevFormat
  }
```

Add at line 121:
```
  if(format !== config.format){
    config.format = format;
    catalog = (0, _catalog2.default)(config);
  }
```

This is on the latest version Julian updated to yesterday.

GeorgiSTodorov:
@jaxesn , cool, thanks.

natmegs:
I prefer the `minimal` format, but I remember hearing (from @nLight I believe) that translators & translation management tools generally require `po` format?

nLight:
@natmegs they should be good with minimal as well :)

GeorgiSTodorov:
Fixed file conflicts so that this is review-able and merge-able. Also, for some reason I could not get this working for the po format.

jaxesn:
I would vote to switch to minimal, at least for the time being.  Its a theory, but removing the line numbers may help with some of the merge issues we have had in the last few weeks.  We do not have an example as you described yet, for that to happen an internal team or a client will have to point it out to us since the translators aren't looking at the app for context.  Even if one such case does arise, we can handle it by not using the english text for that specific key, as @nLight suggested.

GeorgiSTodorov:
@juliangieseke , I think this can be overcome in some way.

GeorgiSTodorov:
After thinking about it some more, while there are ways to make the same English text translated in a different way, I think this highlights a design problem.

GeorgiSTodorov:
Resolved conflicts again, I think that every time we merge something in master we break it here.

PR #3477, opened 2018-12-17
DCOS-46278: Chinese Translation: Make sure all mentions of IP, CPU, MEM, DISK, GPU and ID are expressed in both English and Chinese

GeorgiSTodorov:
Change abbreviations to be more accurate

Closes https://jira.mesosphere.com/browse/DCOS-46278

## Testing
~~(To be updated)~~
Switch to Chinese and check that "CPU" is translated as "CPU 分配" (meaning CPU Allocation) everywhere (for example Nodes tab, Services tab, etc).
Check that Memory and Mem are translated as "记忆".
Check that Disk is translated as "磁盘".
Check that CPU and GPU are NOT translated.

## Trade-offs
A lot.
~~1. Since the full scope of the jira is not clear to me, for now I am only changing CPU to CPU Allocation in Chinese. @AmrAbdelrazik , I would appreciate more information on what needs to be done since the jira description is not up-to-date anymore because of our slack conversation.~~
2. In the Services Table, the CPU Column header gets truncated (Mem as well, we should rethink our truncation in one place it hides the last Chinese symbol)
screenshot from 2018-12-17 14-13-04
3. In the task details tab, some columns are not translated.
screenshot from 2018-12-17 14-15-39
Check `dcos-ui/plugins/services/src/js/pages/task-details/TaskDetailsTab.js` line 69. When I wrapped that code in `Trans`, I got
screenshot from 2018-12-17 14-02-11
and I am not sure that this `{0}` should be in the message catalog.
4. Not sure if we should and can translate framework configuration variables.
screenshot from 2018-12-17 14-25-21

## Dependencies
None.

## Screenshots
screenshot from 2018-12-18 13-42-26

screenshot from 2018-12-18 13-42-57

screenshot from 2018-12-18 13-43-15

screenshot from 2018-12-18 13-43-45


GeorgiSTodorov:
Since I got more information on what needs to be done, I've updated the testing instructions and screenshots.
Related to the point 3 of the trade-offs, I added translation marks to two files https://github.com/dcos/dcos-ui/pull/3477/files#diff-eae7a6a90ebb28dbe29adbae1140a1af and https://github.com/dcos/dcos-ui/pull/3477/files#diff-def918eb22daa7712f7a5e81753e82eb , but the text remains in English (values that already have translations in the message catalog). I tried with and without `module.exports = withI18n()(Component)`. In `General.js`, adding `withI18n()` makes it impossible to untoggle the JSON view when creating a job.
@natmegs , do you perhaps know what else needs to be done to have translations in these two files?
screenshot from 2018-12-18 14-00-53

screenshot from 2018-12-18 14-01-04

GeorgiSTodorov:
@DanielMSchmidt , I know, I added translation marks, but apparently it needs something else (see my previous comment).

src/js/schemas/job-schema/General.js
```diff
@@ -1,30 +1,30 @@
...
+import { i18nMark } from "@lingui/react";
 import JobValidatorUtil from "../../utils/JobValidatorUtil";
 import ValidatorUtil from "../../utils/ValidatorUtil";
 import MesosConstants from "../../../../plugins/services/src/js/constants/MesosConstants";
 import JobResources from "../../constants/JobResources";
 
 const General = {
-  title: "General",
-  description: "Configure your job settings",
+  title: i18nMark("General"),
```
natmegs: `i18nMark` does not actually translate anything - it just adds the marked text to the translation catalog. To actually translate the text, it has to be wrapped in `<Trans id={title} />` wherever it's used (where `title` would be whatever the marked variable is called where it's used)

GeorgiSTodorov: Thank you.

GeorgiSTodorov:
Fixed file conflicts and the translations for one of the two pages.

src/js/schemas/job-schema/General.js
```diff
@@ -81,7 +82,7 @@ const General = {
               })
             ) {
               definition.showError =
-                "Mem must be a number and at least " +
+                i18nMark("Mem must be a number and at least ") +
                 MesosConstants.MIN_MEM +
                 " MiB";
```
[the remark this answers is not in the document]
GeorgiSTodorov: This is not necessarily about needing MiB translated, but about keeping the context. Sentence structure might be different in Chinese.

src/js/schemas/job-schema/General.js
```diff
@@ -1,30 +1,30 @@
...
 import JobValidatorUtil from "../../utils/JobValidatorUtil";
 import ValidatorUtil from "../../utils/ValidatorUtil";
 import MesosConstants from "../../../../plugins/services/src/js/constants/MesosConstants";
 import JobResources from "../../constants/JobResources";
 
 const General = {
-  title: "General",
-  description: "Configure your job settings",
+  title: i18nMark("General"),
+  description: i18nMark("Configure your job settings"),
```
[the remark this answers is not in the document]
GeorgiSTodorov: Yeah, I couldn't figure it out yesterday, today I will try again.

src/js/schemas/job-schema/General.js
```diff
@@ -57,9 +57,10 @@ const General = {
                 min: MesosConstants.MIN_CPUS
               })
             ) {
-              definition.showError =
+              definition.showError = i18nMark(
                 "CPUs must be a number at least equal to " +
```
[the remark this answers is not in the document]
GeorgiSTodorov: I will do it that way, but some context might be lost. Maybe in Chinese the correct sentence structure would be "Must be a number at least equal to ... for the CPUs" for example.

GeorgiSTodorov:
In `createFormModelFromSchema` in `dcos-ui/src/js/utils/JobUtil.js` we have the text for the jobs modal.
I can do something like
```const properties = schema.properties.general.properties.resources.properties;
if (typeof(properties.cpus.title) === "string") properties.cpus.title = <Trans render="label" id={properties.cpus.title} />;```
which is ugly, but works.
In general, we should create a function that checks all instances of "title" in this nested object and replaces them, but I don't think we can do that today.
Another option is to find where it is rendered, but even then, there is no guarantee that it will work.

**Comment by GeorgiSTodorov** [comment]
I made some progress with my idea.
In `dcos-ui/src/js/utils/JobUtil.js`
we define
```
function customFilter(object, result){
  if(object === null) return;
  if(object.hasOwnProperty('title')) {
      result.push(object);

      if (object.title && typeof(object.title) === "string") object.title = <Trans render="span" id={object.title} />;
  }

  for(var i=0; i<Object.keys(object).length; i++){
      if(typeof object[Object.keys(object)[i]] == "object"){
          customFilter(object[Object.keys(object)[i]], result);
      }
  }
}
```
and in `createFormModelFromSchema` we call `customFilter(schema.properties, []);`
There are some things left to fix and I need to do the same thing for the descriptions, but it's something that might be useful.
screenshot from 2018-12-20 18-09-45

File: locale/zh/messages.json [hunk 5]
```diff
@@ -131,10 +130,10 @@
   "CA Certificate": "CA 证书",
   "CA certificate chain": "CA 证书链",
   "CLI Only Package": "CLI Only Package（只 CLI 包）",
-  "CPU": "CPU",
+  "CPU": "CPU 分配",
   "CPU Allocation": "CPU 分配",
-  "CPU/MEM/DSK": "CPU/MEM/DSK",
-  "CPUs": "CPU",
+  "CPU/MEM/DSK": "CPU 分配/记忆/DSK",
```
nLight: Shall we translate DSK as 磁盘 as well?

natmegs: 👍 added

GeorgiSTodorov:
Follow-up PR for translating the Jobs modal: https://github.com/dcos/dcos-ui/pull/3492 .

PR #3479, opened 2018-12-17
DCOS_OSS-4579: [1.12] accept floating points in FrameworkConfigurationForm

mperrotti:
## Testing
2. Go to the "Nodes" tab

## Trade-offs

## Dependencies
None

## Screenshots
Before:
floatingpoint-broken

After:
floatingpoint-fixed


GeorgiSTodorov:
Probably not related to this PR, but the validation looks a bit weird.
peek 2018-12-17 17-32

mperrotti:
@GeorgiSTodorov,  we don't - feel free to create one :)

GeorgiSTodorov:
https://jira.mesosphere.com/browse/DCOS-46311

PR #3481, opened 2018-12-17
DCOS_OSS-4582: [1.11]  accept floating points in FrameworkConfigurationForm

mperrotti:
## Testing
2. Go to the "Nodes" tab

## Trade-offs

## Dependencies
None

## Screenshots
Before:
floatingpoint-broken

After:
floatingpoint-fixed

### DIFF
--- a/src/js/components/FrameworkConfigurationForm.js
+++ b/src/js/components/FrameworkConfigurationForm.js
@@ -315,6 +315,10 @@ export default class FrameworkConfigurationForm extends Component {
                       validate={this.validate}
                       showErrorList={false}
                       transformErrors={this.transformErrors}
+                      noHtml5Validate={true}
+                      ref={form => {
+                        this.schemaForm = form;
+                      }}
                     >
                       <div />
                     </SchemaForm>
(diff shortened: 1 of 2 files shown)


mperrotti:
@GeorgiSTodorov - can you send me a screen capture? I can't reproduce

intlnumberinput

GeorgiSTodorov:
@mperrotti , I started recording a gif, but I realized I have no way of showing that I am pressing "," and not delete.

mperrotti:
@GeorgiSTodorov That's ok, I understand you're pressing comma. I mostly just want to see if you're doing something different from me.

Is it only happening in 1.11?

GeorgiSTodorov:
Yes, I have only see it happen in 1.11.

brandonc:
Can confirm. As soon as I press "," in firefox, the entire input is deleted.

mperrotti:
@GeorgiSTodorov  - what language are you testing that 1.11 issue in? I've been using French, and I can't reproduce the issue you're seeing

i18nnumberinput

PR #3482, opened 2018-12-17
DCOS_OSS-4597: [1.12] service config dropdown content overflows

mperrotti:
## Testing
1. In `plugins/services/src/js/containers/service-configuration/ServiceConfiguration.js`, on line 143, change the `<DateFormat />` component to just be a span with a long string
2. Open a service from the "Services" tab
3. Go to the "Configuration" tab
4. Click the dropdown at the top of that screen

The long string you replaced `<DateFormat />` with should not overflow the container - it should appear truncated

## Trade-offs
None

## Dependencies
None

## Screenshots
Before:
screen shot 2018-12-10 at 3 42 01 pm

After:
screen shot 2018-12-10 at 3 45 04 pm

Note: I tested this in Chrome, FF, Safari, and Edge

### DIFF
--- a/plugins/services/src/js/containers/service-configuration/ServiceConfiguration.js
+++ b/plugins/services/src/js/containers/service-configuration/ServiceConfiguration.js
@@ -136,8 +136,10 @@ class ServiceConfiguration extends mixin(StoreMixin) {
         let itemCaption = localeVersion;
         if (version === service.getVersion()) {
           itemCaption = (
-            <span className="badge-container">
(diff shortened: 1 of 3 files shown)


GeorgiSTodorov:
For me, the text is already truncated in 1.12 even without this change.
screenshot from 2018-12-18 14-45-12

screenshot from 2018-12-18 14-51-14

GeorgiSTodorov:
Ah, it was already merged and I didn't update my page. Nevermind.

PR #3490, opened 2018-12-21
DCOS_OSS-4580: [1.10] accept floating points

mperrotti:
## Testing
2. Go to the "Nodes" tab

**Note: this requires an update from reactjs-components: https://github.com/mesosphere/reactjs-components/pull/492**

## Trade-offs

## Dependencies
reactjs-components release from this PR: https://github.com/mesosphere/reactjs-components/pull/492

## Screenshots
Before:
floatingpoint-broken1 10

After:
floatingpoint-fixed1 10

### DIFF
--- a/src/js/components/SchemaForm.js
+++ b/src/js/components/SchemaForm.js
@@ -419,6 +419,7 @@ class SchemaForm extends mixin(StoreMixin, InternalStorageMixin) {
           getTriggerSubmit={this.getTriggerTabFormSubmit}
           onChange={this.handleFormChange}
           onTabClick={this.handleTabClick}
+          noValidate={this.props.noValidate}
         />
       </div>
     );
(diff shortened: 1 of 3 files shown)


DanielMSchmidt:
@mperrotti We need a reactjs-components release and an updated version here

juliangieseke:
if this needs a reactjs-components bump, why is it approved then? 🤔

DanielMSchmidt:
Because it works with the bump 😇

juliangieseke:
so bump was done before? can we merge then?

mperrotti:
We can't. I'm having trouble publishing a reactjs-components release to NPM

DanielMSchmidt:
@mperrotti Could you resolve the publish trouble?

juliangieseke:
turned out that the new `reactjs-component` release depends on CNVS 1.1.18, but 1.10 is still on 1.1.9 - closing this PR because trying to backport this will be too much effort.

PR #3491, opened 2018-12-22
[WIP] DCOS-46498: fix mem, disk, ip in chinese message catalog

natmegs:
Fix Mem, Disk, IP translations in chinese message catalog.

TODO: 
- [ ] Clarify with @AmrAbdelrazik whether we should replace instances of "Memory Allocation" with just "Memory"

Closes DCOS-46498.

## Testing

Look at Dashboard (and other places where Mem/Memory, IP, Disk are shown):
- Memory in chinese (记忆) should be MEM 记忆
- Disk should be Disk 磁盘
- IP should be IP 主机

## Trade-offs

None

## Dependencies

None


GeorgiSTodorov:
I think we can close this.

PR #3496, opened 2019-01-04
[COPS] DCOS_OSS-4633: [1.13] dcos6 network is shown as available for UCR in the UI.

GeorgiSTodorov:
Make the virtual network dcos6 unavailable
when UCR is chosen as a container.

Closes https://jira.mesosphere.com/browse/DCOS_OSS-4633

## Testing
1. Go to Services tab
2. Create a new Single Container service
3. On the Service tab, click "More Settings" at the bottom and select "Universal Container Runtime (UCR)" for Container Runtime.
4. Go to Networking type, open the Network Type dropdown and verify that "dcos6" is not shown.

## Trade-offs
Somewhat quick and dirty solution. I would be fine if we decide that we don't want it and if this doesn't fix it, maybe it can at least point us in the right direction.

Edit: Also, should the same thing be done for pods?

## Dependencies
None.

## Screenshots
peek 2019-01-04 14-14


plugins/services/src/js/components/forms/NetworkingFormSection.js
```diff
@@ -556,7 +561,7 @@ class NetworkingFormSection extends mixin(StoreMixin) {
       })
       .getItems()
       .map((virtualNetwork, index) => {
-        return (
+        return mesosContainer && virtualNetwork.text === "dcos6" ? null : (
```
DanielMSchmidt: Could you do this with a `filterItems` instead? I think it would be the clearer and more functional approach

```diff
+ .filterItems(item => !mesosContainer || virtualNetwork.text !== "dcos6")
  .getItems()
  .map((virtualNetwork, index) => {
-    return mesosContainer && virtualNetwork.text === "dcos6" ? null : (
+    return (

**Comment by GeorgiSTodorov** [comment]
Updated with @DanielMSchmidt 's suggestion.

File: plugins/services/src/js/components/forms/NetworkingFormSection.js [hunk 2]
```diff
@@ -545,6 +545,11 @@ class NetworkingFormSection extends mixin(StoreMixin) {
   }
 
   getVirtualNetworks() {
+    const mesosContainer =
```
nLight: Could use `import { findNestedPropertyInObject } from "#SRC/js/utils/Util";`

GeorgiSTodorov:
Thanks, @nLight .

nLight:
Would you please adjust the solution according to Sergei's comment?
https://jira.mesosphere.com/browse/DCOS_OSS-2260?focusedCommentId=206481&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-206481

TL;DR we should check not on `dcos6` name but if `subnet6` key is present. 
Also would you please add checking on `enabled` as suggested in https://jira.mesosphere.com/browse/DCOS-47272 ? Thank you!

nLight:
I'm escalating this PR from no-pressure to cops

GeorgiSTodorov:
@nLight , updated with your suggestions.

plugins/services/src/js/components/forms/NetworkingFormSection.js
```diff
@@ -545,15 +545,24 @@ class NetworkingFormSection extends mixin(StoreMixin) {
...
         return {
+          enabled: overlay.info.enabled,
+          subnet6: overlay.getSubnet6(),
           text: name,
           value: `${CONTAINER}.${name}`
         };
       })
+      .filterItems(
+        virtualNetwork =>
+          !mesosContainer || !virtualNetwork.subnet6 || !virtualNetwork.enabled
```
nLight: `filter` should return `true` to keep elements, shouldn't it? 

I might not being clear enough, sorry.

We want to remove `enabled: false` items regardless of the network type and container type. Plus we want to remove `subnet6` networks for `mesosContainer`.

We also want to fix pods `MultiContainerNetworkingFormSection#getVirtualNetworks` pods _always_ use UCR so we should always remove `subnet6` networks and of course remove `enabled: false` as well.

GeorgiSTodorov:
@nLight , updated.

plugins/services/src/js/components/forms/NetworkingFormSection.js
```diff
@@ -545,15 +545,25 @@ class NetworkingFormSection extends mixin(StoreMixin) {
...
+          enabled: overlay.info.enabled,
+          subnet6: overlay.getSubnet6(),
           text: name,
           value: `${CONTAINER}.${name}`
         };
       })
+      .filterItems(
+        virtualNetwork =>
+          virtualNetwork.enabled &&
+          (mesosContainer ? !virtualNetwork.subnet6 : true)
```
nLight: We could add a small in-line function `isAvailableOnRuntime` just to clarify the intent:
```js
const isAvailableOnRuntime = (mesosContainer, {subnet6}) => !mesosContainer || !virtualNetwork.subnet6
```

How do you think? I'm afraid that complex logical expressions are hard to read.

GeorgiSTodorov: Done, I also added a comment.

GeorgiSTodorov:
Testing integration tests locally.

GeorgiSTodorov:
I updated tests, @nLight , please re-review.

tests/_fixtures/mesos/overlay-master.json
```diff
@@ -277,12 +277,26 @@
...
       },
       {
         "name": "dcos-2",
         "prefix": 24,
-        "subnet": "192.168.128.0\/17"
+        "subnet": "192.168.128.0\/17",
+        "enabled": true
+      },
+      {
+        "name": "dcos-3",
```
nLight: Let's use proper ipv6 network fixture 💪 
```
            {
                "enabled": true,
                "name": "dcos6",
                "prefix6": 80,
                "subnet6": "fd01:b::/64"
            }
```

nLight: name is not important here, but other fields are

GeorgiSTodorov:
Now the Networks tests are failing.

GeorgiSTodorov:
Updated tests, let's see now.
@nLight , @natmegs , please re-review.

GeorgiSTodorov:
@juliangieseke , I tested locally, but I am still not sure if all tests are passing.

PR #3499, opened 2019-01-04
DCOS-43909 SDK Plans Client

TattdCodeMonkey:
Created the client for querying a service's plans from the API and
associated types. This will be used by a data-layer GraphQL connection that will be written after this. That is why I placed it in the data folder for the services plugin.

## Testing

There are unit tests to run.

Otherwise this client will be used by a data-layer connection that will be written in another ticket.

## Trade-offs

I wasn't 100% sure where these files should live. I actually moved them 3 times while developing this. If you think there is a better place for this to live I am open to suggestions.

### DIFF
--- a/plugins/services/src/js/data/ServicePlansClient.ts
+++ b/plugins/services/src/js/data/ServicePlansClient.ts
@@ -1,0 +1,48 @@
+import { request, RequestResponse } from "@dcos/http-service";
+import { Observable } from "rxjs/Observable";
+
+export type ServicePlanStatus =
+  | "ERROR"
+  | "WAITING"
+  | "PENDING"
+  | "PREPARED"
+  | "STARTING"
+  | "STARTED"
+  | "COMPLETE"
+  | "IN_PROGRESS";
+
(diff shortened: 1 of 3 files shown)


juliangieseke:
@TattdCodeMonkey One small thing: you forgot to add the issue number to your commit (e.g. `Closes DCOS-43909`) - because of that our automated bump script will not pick it up and not add the JIRA to  the BUMP PR - can you please keep track of the issue or add it to the message? 😇

DanielMSchmidt:
Adding hold merge to give @TattdCodeMonkey the chance to add the Jira number 👍

TattdCodeMonkey:
@juliangieseke @DanielMSchmidt  I updated the commit message and re-pushed. Please 👀 & ✅ 😃

@GeorgiSTodorov The status come from the API, so we just need to ensure we support all of the possible values. [See here](https://github.com/mesosphere/dcos-commons/blob/4910aeb/sdk/scheduler/src/main/java/com/mesosphere/sdk/scheduler/plan/Status.java)

PR #3500, opened 2019-01-07
DCOS-44818: add script that detects .only statements

DanielMSchmidt:
## Testing

You can add a bash file like this on the top-level and call it:

```bash
#!/bin/bash

set -e
source "./scripts/utils/test"

echo "$(find_files_with_debug_directives './tests')"
```

It should exit with status code 1 if there is no file in ./tests to contain `.only`. If you change that it should display each file name in a new line

## Trade-offs

- Wrote it in bash as we wanted to reuse the code that is already there, so that we can rewrite it later together, rather then having two versions, one for bash and one for JS

### DIFF
--- a/scripts/utils/test
+++ b/scripts/utils/test
@@ -16,3 +16,12 @@ find_debug_directives(){
 
  echo -e "${matches}"
 }
+
+find_files_with_debug_directives(){
+  if [[ -z $* ]]
+  then
+   exit 0
+  fi
+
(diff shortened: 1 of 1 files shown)


DanielMSchmidt:
@nLight I rebased 🙇

DanielMSchmidt:
@nLight I fixed a bug @GeorgiSTodorov pointed out on linux, could you reapprove?

PR #3501, opened 2019-01-07
DCOS-41707: [1.12] DCOS UI Shows foreign tasks for a service if they match the name

GeorgiSTodorov:
Closes https://jira.mesosphere.com/browse/DCOS-41707

## Testing
1. Run a pod with the following configuration (taken from jira):
```
{
  "id": "/nginx-busybox",
  "version": "2018-09-17T16:36:00.220Z",
  "containers": [
    {
      "name": "nginx",
      "resources": {
        "cpus": 0.001,
        "mem": 32,
        "disk": 0
      },
      "image": {
        "kind": "DOCKER",
        "id": "nginx"
      }
    },
    {
      "name": "container-2",
      "resources": {
        "cpus": 0.001,
        "mem": 32,
        "disk": 0
      },
      "image": {
        "kind": "DOCKER",
        "id": "alpine"
      },
      "exec": {
        "command": {
          "shell": "sleep 100000"
        }
      }
    }
  ],
  "networks": [
    {
      "name": "dcos",
      "mode": "container"
    }
  ],
  "scaling": {
    "instances": 1,
    "kind": "fixed"
  },
  "scheduling": {
    "placement": {
      "constraints": []
    }
  },
  "executorResources": {
    "cpus": 0.1,
    "mem": 32,
    "disk": 10
  },
  "volumes": [],
  "fetch": []
}
```
2. Run `nginx` from the catalog
3. Verify that the task for `nginx-busybox` does not appear in the tasks table for `nginx`.
4. Verify that other catalog packages show and work correctly

## Trade-offs
~This is the best solution that I could think of. I tested some framework tasks, and they all had `discovery` and `discovery.name`.~

Added a new solution. It should work. An alternative that @zen-dog suggested is to check in Marathon (we currently use the Mesos store). I checked and we do have this information in the Marathon store, but I think that changing the store logic gives room for errors and since we are moving away from store in order to use the data layer, I don't think it is worth it.

## Dependencies
None.

### Before
screenshot from 2019-01-07 14-57-32

### After
screenshot from 2019-01-09 15-28-33

### DIFF
--- a/src/js/stores/MesosStateStore.js
+++ b/src/js/stores/MesosStateStore.js
@@ -255,8 +255,16 @@ class MesosStateStore extends GetSetBaseStore {
       }
     }
 
+    // Check which tasks belong to this service
     return tasks
-      .filter(task => task.isStartedByMarathon && task.name === mesosTaskName)
+      .filter(
+        task =>
+          task.isStartedByMarathon &&
+          task.name === mesosTaskName &&
+          (service.get("tasks") && service.findTaskById
+            ? service.findTaskById(task.id)
+            : true)
+      )
       .concat(serviceTasks)
       .map(task => MesosStateUtil.flagSDKTask(task, service));
   }
--- a/src/js/stores/__tests__/MesosStateStore-test.js
+++ b/src/js/stores/__tests__/MesosStateStore-test.js
@@ -158,7 +158,10 @@ describe("MesosStateStore", function() {
       var tasks = MesosStateStore.getTasksByService(
         new Framework({
           id: "/spark",
-          labels: { DCOS_PACKAGE_FRAMEWORK_NAME: "spark" }
+          labels: { DCOS_PACKAGE_FRAMEWORK_NAME: "spark" },
+          findTaskById: () => {
+            return true;
+          }
         })
       );
       expect(tasks).toEqual([
@@ -176,7 +179,12 @@ describe("MesosStateStore", function() {
 
     it("returns matching application tasks", function() {
       var tasks = MesosStateStore.getTasksByService(
-        new Application({ id: "/alpha" })
+        new Application({
+          id: "/alpha",
+          findTaskById: () => {
+            return true;
+          }
+        })
       );
       expect(tasks).toEqual([
         {
@@ -233,6 +241,9 @@ describe("MesosStateStore", function() {
           labels: {
             DCOS_COMMONS_API_VERSION: 1,
             DCOS_PACKAGE_FRAMEWORK_NAME: "spark"
+          },
+          findTaskById: () => {
+            return true;
           }
         })
       );


src/js/stores/MesosStateStore.js
```diff
@@ -256,7 +256,13 @@ class MesosStateStore extends GetSetBaseStore {
     }
 
     return tasks
-      .filter(task => task.isStartedByMarathon && task.name === mesosTaskName)
+      .filter(
+        task =>
+          task.isStartedByMarathon &&
+          task.name === mesosTaskName &&
+          task.discovery &&
```
nLight: According to protobufs `discovery` is optional https://github.com/apache/mesos/blob/40210dbcfb218d6ef3f647b047afbdda24133f4d/include/mesos/mesos.proto#L2184

GeorgiSTodorov: So what is the best way to distinguish service task from a framework task?

nLight: That's the core of the issue I'm afraid. I don't think I have a definitive answer.

src/js/stores/MesosStateStore.js
```diff
@@ -256,7 +256,13 @@ class MesosStateStore extends GetSetBaseStore {
     }
 
     return tasks
-      .filter(task => task.isStartedByMarathon && task.name === mesosTaskName)
+      .filter(
+        task =>
+          task.isStartedByMarathon &&
+          task.name === mesosTaskName &&
```
nLight: In our case tasks names do match. Name is `nginx` for both framework and the pod instance. Though Task IDs are different. `nginx-busybox.instance-92534c81-128a-11e9-aae4-ba5756c7ca71.nginx` I wonder if we could leverage this.

nLight: https://github.com/dcos/dcos-ui/pull/3501/commits/2f8699e2cce84f5831b2aca83ffc362844e5fc23#diff-22eec13077e445e982cd1a9232667028R233
```javascript
const mesosTaskName = service.getMesosId();
```
Fun fact. We call `MesosId` and assign the result to `taskName`. But if you look inside `getMesosId` you'd be surprised that this is UI only hoodoo 🤷‍♂️

nLight: I'd suggest we try telling _pods_ tasks from other tasks instead. 

The trouble is that if you have a pod with a container named like another task we have the issue. But you can' really have i.e. framework called `nginx` and single container app called `nginx`

GeorgiSTodorov: Yeah, I noticed the different task IDs, but I thought that the discovery.name was a cleaner solution, because I didn't know it was optional. I will try thinking of something else. Thank you for your help.

GeorgiSTodorov: I noticed that only pods tend to have `executor_id`, but it's marked as optional in `mesos.proto`, so I think that frameworks might have `executor_id` as well.

GeorgiSTodorov: The task id for the nginx framework looks like
```
task_id: Object { value: "nginx.b21e9bc9-1337-11e9-91f2-5e1fefcf97fb" }
```
The task id for a pod with a container image named "nginx" is:
```
task_id: Object { value: "nginx.instance-96331628-1337-11e9-91f2-5e1fefcf97fb.nginx" }
```
and the task id for a pod with a container image containing a dot is
```
task_id: Object { value: "nginx.instance-1.instance-1181ba75-1340-11e9-91f2-5e1fefcf97fb.nginx" }
```
So we can't just take the first part before the dot, because that might be misleading. I noticed that only the pods had a task_id ending with the container name, so I try to use that.

GeorgiSTodorov:
Looking into failing tests.

zen-dog:
> An alternative that @zen-dog suggested is to check in Marathon (we currently use the Mesos store). I checked and we do have this information in the Marathon store, but I think that changing the store logic gives room for errors and since we are moving away from store in order to use the data layer, I don't think it is worth it.

So the question is where does the data layer get the data from? I don't think just looking at Mesos tasks names is enough to associate them with a specific service. The edge case you discovered is just one of many - another example would be: 
- One can have multiple Marathon frameworks running (MoM)
- A service name is only unique **within** one MoM
- Nothing prevents 2 MoMs from starting a service called `/nginx`

GeorgiSTodorov:
@zen-dog , you are right, but fixing this in all cases would require more serious planning and commitment than we have for this ticket. I will talk to the team managers in order to give this problem the attention it deserves. Thank you for your help.

GeorgiSTodorov:
Tests should be fixed now. In `ServiceDetail-cy`, `shows configuration of a service` kept failing locally, but it was failing on the `release/1.12` branch as well, so let's see CI.

GeorgiSTodorov:
After talking with @nLight , he suggested to stop working on this and re-evaluate.

PR #3502, opened 2019-01-07
DCOS-46808: Rerun flaky tests on CI

DanielMSchmidt:
## Testing

1. Create a fresh branch based off of this
2. Edit one or multiple integration test files on that branch to have a `it.only` or `describe.only` and push these changes
3. Follow "Debugging flaky integration tests" in Contributing.md
4. See a successful run if you picked a non-flaky test case, otherwise it should be a failing run

## Trade-offs

- Stayed with everything written in bash, it would have been too much to move both files to JS and having one in JS would mean code duplication. As this script is not always in use I did not want to risk it.

## Dependencies

~This is based of #3500, so let's get that one merged first.~


DanielMSchmidt:
@juliangieseke Do you happen to know a JIRA epic to the flaky tests that we have?

GeorgiSTodorov:
@DanielMSchmidt , you mean this one - https://jira.mesosphere.com/browse/DCOS-20241 ?

DanielMSchmidt:
Yes that's the right one

DanielMSchmidt:
As you can see `pr-head` fails because there are no tests to run. Which is nice, so you know you forgot to commit a `.only` 🤷‍♂️ 

In this PR you can see it running: https://github.com/dcos/dcos-ui/pull/3504

nLight:
I believe if we rebase it some commits will go away?

DanielMSchmidt:
Yeah, that will trigger a run @nLight You could run it with parameters again to see it fail without any onlys

GeorgiSTodorov:
I am not very familiar with scripts, how does `rerun-integration-tests.sh` decide if a test is flaky? Does it only need one successful and one unsuccessful run? And if yes, can we stop the remaining runs if we have one successful and one unsuccessful run?

DanielMSchmidt:
@GeorgiSTodorov Thanks for asking. I sense two questions here, let me get to them individually:

> how does `rerun-integration-tests.sh` decide if a test is flaky?

It doesn't. It simply runs the tests the given time and merges the resulting junit reports to one big one. This report is then used by Jenkins to determine if the tests are flaky (it checks for the same test being run with different result).

> Does it only need one successful and one unsuccessful run? 

This depends on the question we want to answer. For "is it falky" that would be sufficient. I would like to give us the opportunity to answer "how flaky is it?" by looking at the result. Let's say we have a test that fails 50% of time time, so 50 out of 100 runs then it would be a huge improvement to get it to 1 out of 100, so 1% of flake. It might be sufficient for the moment because we have a lot of 10% flakes at the moment we should tackle sooner.

GeorgiSTodorov:
@DanielMSchmidt , thank you for the detailed answer.

PR #3503, opened 2019-01-07
DCOS-43910 SDK Service Plan data-layer

TattdCodeMonkey:
Introducing data-layer for the services plugin. We are adding this to
support querying for a SDK service's plans. These will be displayed in
the UI. To support querying other service information through the
data-layer the root query object is the `Service` even though currently
this only supports a name and the plans.

## Testing

For now just run the unit tests, but component coming soon!

## Trade-offs

A graphql query that includes `plans` will result in `n+1` requests to the API. A service should have a limited number of plans and 🤞 never more then 100, but this is something we should be aware of when building service queries. We should be mindful of the performance implications of the `graphqlobservalble` polling with `n+1 `queries as well.

### DIFF
--- a/plugins/services/src/js/data/__tests__/index-test.ts
+++ b/plugins/services/src/js/data/__tests__/index-test.ts
@@ -16,9 +16,8 @@ import {
   ServicePlanResponse
 } from "#PLUGINS/services/src/js/data/ServicePlansClient";
 
-function makeFakePlanResponse(name: string): ServicePlanResponse {
+function makeFakePlanResponse(): ServicePlanResponse {
   return {
-    name,
     strategy: "serial",
     status: "IN_PROGRESS",
     errors: [],
@@ -50,12 +49,12 @@ function makeResolverConfig(
       m.cold("(j|)", {
         j: { code: 200, message: "ok", response: plansResponse }
       }),
-    fetchServicePlanDetail: (_serviceName: string, planName: string) =>
+    fetchServicePlanDetail: (_serviceName: string, _planName: string) =>
       m.cold("(j|)", {
         j: {
           code: 200,
           message: "ok",
-          response: makeFakePlanResponse(planName)
+          response: makeFakePlanResponse()
         }
       }),
     pollingInterval: m.time("--|")
@@ -140,12 +139,12 @@ describe("Service data-layer", () => {
               response: ["plan-01"]
             }
           });
-        const fetchServicePlanDetail = (_service: string, planName: string) =>
+        const fetchServicePlanDetail = (_service: string, _planName: string) =>
           m.cold("--j|", {
             j: {
               code: 200,
               message: "ok",
-              response: makeFakePlanResponse(planName)
+              response: makeFakePlanResponse()
             }
           });
         const serviceSchema = makeExecutableSchema({
@@ -240,11 +239,16 @@ describe("Service data-layer", () => {
       marbles(m => {
         m.bind();
 
-        const resolversConfig = makeResolverConfig(m);
-        resolversConfig.fetchServicePlans = (_serviceName: string) =>
-          m.cold("(j|)", {
-            j: { code: 500, message: "Internal Server Error", response: {} }
-          });
+        const plansResult$ = m.cold("(j|)", {
+          j: { code: 500, message: "Internal Server Error", response: {} }
+        });
+        mockRequest.mockReturnValueOnce(plansResult$);
+
+        const resolversConfig = {
+          fetchServicePlans: fetchPlans,
+          fetchServicePlanDetail: fetchPlanDetails,
+          pollingInterval: m.time("--|")
+        };
 
         const serviceSchema = makeExecutableSchema({
           typeDefs: schemas,
@@ -266,7 +270,7 @@ describe("Service data-layer", () => {
 
         const expected$ = m.cold("#", undefined, {
           message:
-            "Service plans API request failed: 500 Internal Server Error:{}",
+            "Service Plans API request failed: 500 Internal Server Error:{}",
           name: "Error"
         });
 
@@ -279,14 +283,20 @@ describe("Service data-layer", () => {
       marbles(m => {
         m.bind();
 
-        const resolversConfig = makeResolverConfig(m);
-        resolversConfig.fetchServicePlanDetail = (
-          _serviceName: string,
-          _planName: string
-        ) =>
-          m.cold("(j|)", {
-            j: { code: 500, message: "Internal Server Error", response: {} }
-          });
+        const plansResult$ = m.cold("(j|)", {
+          j: { code: 200, message: "ok", response: ["plan-01"] }
+        });
+        const planDetailResult$ = m.cold("(j|)", {
+          j: { code: 500, message: "Internal Server Error", response: {} }
+        });
+        const results = [plansResult$, planDetailResult$];
+        mockRequest.mockImplementation(() => results.shift());
+
+        const resolversConfig = {
+          fetchServicePlans: fetchPlans,
+          fetchServicePlanDetail: fetchPlanDetails,
+          pollingInterval: m.time("--|")
+        };
 
         const serviceSchema = makeExecutableSchema({
           typeDefs: schemas,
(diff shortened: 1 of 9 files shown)


plugins/services/src/js/structs/ServicePlan.ts
```diff
@@ -0,0 +1,36 @@
+import { ServicePlanResponse } from "#PLUGINS/services/src/js/data/ServicePlansClient";
```
Poltergeist: Can we please put the types into a types directory since the structs represent something else?

TattdCodeMonkey:
Sure, I’ll move them.

plugins/services/src/js/data/ServicePlansModel.ts
```diff
@@ -0,0 +1,123 @@
...
+  serviceName: string;
+}
+
+export interface ServicePlanDetailQueryArgs {
+  serviceName: string;
+  planName: string;
+}
+
+export const typeDefs = `
+${ServicePlanStatusSchema}
```
DanielMSchmidt: I wonder if it would make sense to move these into separate files and use the 

```graphql
extend ServicePlan {
   steps: ...
}
```

Syntax together with passing them into `makeExecutableSchma` as arrays. 

This would allow us to have a nice place for the field resolvers if there were some to come, plus we can unit test these. Might be overengineered, I don't know 🙈

plugins/services/src/js/data/ServicePlansModel.ts
```diff
@@ -0,0 +1,123 @@
...
+${ServicePlanStepSchema}
+${ServicePlanPhaseSchema}
+${ServicePlanSchema}
+
+type Query {
+  servicePlans(serviceName: String!): [String!]!
```
DanielMSchmidt: While this is technically correct, we kinda waste the advantages of graphql here by treating it as rest. In an ideal world we had already the service type and as a consumer could do something like this:


```
{
  query {
    service(serviceName: "MyService") {
       plans {
          name
          phases {
             name
          }
       }
    }
  }
}
```

as we don't have services it sounds to me like we should rather do sth like this:


```
{
  query {
    servicePlans(serviceName: "MyService") {
       name
       phases {
          name
       }
    }
  }
}
```

GeorgiSTodorov:
Looks good for now. I created a small graph to better understand the relationship between the new files. Sharing it, because it might be useful to someone else as well.
Also, are we going to be translating these `Observable.throw`s?
dependency

TattdCodeMonkey:
@GeorgiSTodorov what do you mean by 
>Also, are we going to be translating these `Observable.throw`s ?

Should I implement those in another way? I was following the `JobModel` example which may not be the best in that case.

Also your graph is out of date 😬 `ServicePlansModel` was removed once I got `index.ts` functional and tested.

I also rebased and force-pushed. This is now ready for review and I _shouldn't_ be force pushing again :)

plugins/services/src/js/data/index.ts
```diff
@@ -0,0 +1,164 @@
...
+
+        return plan$;
+      } else {
+        // Otherwise query for all the service's plans, then pull details for each
+        // plan. Finally combine all of the detail's query observables.
+        const plans$ = pollingInterval$
+          .switchMap(() =>
+            fetchServicePlans(parent.name)
+              .map(
+                (reqResp: RequestResponse<string[]>): string[] => {
```
DanielMSchmidt: Why is there a need to cast these types?

plugins/services/src/js/data/index.ts
```diff
@@ -0,0 +1,164 @@
...
+                  handleAPIError(reqResp, "Service plans");
+                  return reqResp.response;
+                }
+              )
+              .map((plans: string[]) => {
+                return plans.map(name => ({ name }));
+              })
+          )
+          .switchMap(plans => {
+            const planDetials = plans.map(plan => {
```
DanielMSchmidt:
```suggestion
            const planDetails = plans.map(plan => {
```

plugins/services/src/js/data/index.ts
```diff
@@ -0,0 +1,164 @@
...
+              return fetchServicePlanDetail(parent.name, plan.name).map(
+                (reqResp: RequestResponse<ServicePlanResponse>) => {
+                  handleAPIError(reqResp, "Service plan detail");
+                  return ServicePlanDetailResolver(reqResp.response);
```
DanielMSchmidt: This pattern is not needed anymore as we have field resolvers now 🙈

plugins/services/src/js/types/Service.ts
```diff
@@ -0,0 +1,13 @@
...
+
+export interface Service {
+  name: string;
+  plans: ServicePlan[];
+}
+
+export const ServiceSchema = `
+type Service {
+  name: String!
+  plans(name: String = ""): [ServicePlan]!
```
DanielMSchmidt: I don't think our implementation honors the default value for an argument, partly because I have never heard about that graphql feature, nice work 👍

plugins/services/src/js/data/index.ts
Poltergeist: `planDetials` should be `planDetails` typo in detail :-)

plugins/services/src/js/data/__tests__/index-test.ts
```diff
@@ -0,0 +1,632 @@
...
+import { schemas, resolvers } from "../index";
+import {
+  fetchPlanDetails,
+  fetchPlans,
+  ServicePlanResponse
+} from "#PLUGINS/services/src/js/data/ServicePlansClient";
+
+function makeFakePlanResponse(name: string): ServicePlanResponse {
+  return {
+    name,
```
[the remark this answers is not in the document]
TattdCodeMonkey: 👍 fix this in my latest commit. Good catch.

TattdCodeMonkey:
@GeorgiSTodorov cood call, right now we're just using the default `RequestErrorMsg` in the UI. 
@Poltergeist what do you think?

Poltergeist:
@TattdCodeMonkey @GeorgiSTodorov We should definitely consider that I think we actually need to properly propagate the error through the streams but I think that is something we should do as a stretch goal.

plugins/services/src/js/data/__tests__/index-test.ts
```diff
@@ -0,0 +1,632 @@
...
+          graphqlObservable(query, serviceSchema, {}).take(1),
+          graphqlObservable(query, serviceSchema, {}).take(1)
+        ).subscribe(jest.fn(), jest.fn(), () => {
+          expect(resolverConfig.fetchServicePlans).toHaveBeenCalledTimes(1);
+        });
+      })
+    );
+
+    it(
+      "throws an error if plan detail API returns non-200",
```
GeorgiSTodorov: This should say neither "200", nor "202". In your code you are checking if the code is greater than 300.

plugins/services/src/js/data/index.ts
```diff
@@ -0,0 +1,164 @@
+import { makeExecutableSchema } from "graphql-tools";
+import { ServicePlanStatusSchema } from "#PLUGINS/services/src/js/types/ServicePlanStatus";
+import { ServicePlanStepSchema } from "#PLUGINS/services/src/js/types/ServicePlanStep";
+import { ServicePlanPhaseSchema } from "#PLUGINS/services/src/js/types/ServicePlanPhase";
+import {
+  ServicePlanDetailResolver,
+  ServicePlanSchema
+} from "#PLUGINS/services/src/js/types/ServicePlan";
+import { Service, ServiceSchema } from "#PLUGINS/services/src/js/types/Service";
+import { Observable } from "rxjs";
```
GeorgiSTodorov: I think that this import needs to be at the top.

plugins/services/src/js/data/index.ts
```diff
@@ -0,0 +1,164 @@
...
+export interface PlansQueryArgs {
+  name: string;
+}
+
+function isPlansQueryArgs(args: GeneralArgs): args is PlansQueryArgs {
+  return (args as PlansQueryArgs).name !== undefined;
+}
+
+function handleAPIError(reqResp: RequestResponse<any>, message: string = "") {
+  if (reqResp.code !== 200) {
```
GeorgiSTodorov: Should we do the same check as in "ServicePlansClient" here - https://github.com/dcos/dcos-ui/pull/3503/files#diff-709253ad12d4f5ba0075cd28d643b130R41 ?

PR #3505, opened 2019-01-09
chore(package): update webpack-dev-server

juliangieseke:
## Testing

check that `npm start` is still working with and without plugins 👍


nLight:
why?

juliangieseke:
the page might be only visible for admins 🤔 replaced the link with description

mperrotti:
@juliangieseke - what do you mean when you say "with and without plugins"?

GeorgiSTodorov:
@mperrotti , I think with and without the private plugins repository.

PR #3530, opened 2019-01-14
DCOS-47250: send metrics to datadog for each test case

DanielMSchmidt:
Closes 

## Testing

- See metrics being recorded by demo run into DataDog (via OneLogin).
- [build where we pushed to data dog](https://jenkins.mesosphere.com/service/jenkins/blue/organizations/jenkins/frontend%2Fdcos-ui-oss-pipeline/detail/PR-3530/9/pipeline/67)

## Trade-offs

- decided to use a node based tooling so we could script it better once we move the shell script to js


GeorgiSTodorov:
I have no access to DataDog in OneLogin.

DanielMSchmidt:
The issue is also not in review yet ;) I'm unsure how we can get you access, @juliangieseke could you help with that?

GeorgiSTodorov:
Github labels could help in that case. :D

GeorgiSTodorov:
Because in practice it often isn't. :D
Even if it was, we have other useful labels like.

DanielMSchmidt:
Maybe we should pick this discussion up in the sync

PR #3533, opened 2019-01-16
DCOS-47175: Update Table component in nodes page

GeorgiSTodorov:
Use the new version of the ui-kit to enable the nodes table
to auto resize columns based on their content width.

Closes https://jira.mesosphere.com/browse/DCOS-47175

~Currently blocked by https://github.com/dcos-labs/ui-kit/pull/276~

## Testing
1. Go to Nodes tab
2. Verify that the table doesn't jump as the data is loaded
3. Verify that the table works and the columns are as wide as the content

## Trade-offs
~I don't think that the new table looks better. I considered adding some minimal width to the columns, but then we would have the old table in almost every case (as the health column for example will never be wider than the old value of 120px). @lsamimi , @mperrotti , what do you think?~

I added a loader if the data array is empty, but some of the values in the table are slower to load, which causes it to move.

Also fixed some unrelated failing snapshots and some other minor things.

## Dependencies
~https://github.com/dcos-labs/ui-kit/pull/276~

### Before
screenshot from 2019-01-16 11-01-56

### After (old)
screenshot from 2019-01-16 11-04-40

### After (new)
screenshot from 2019-01-17 09-51-21

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -85,9 +85,9 @@
       }
     },
     "@babel/types": {
-      "version": "7.1.6",
-      "resolved": "https://registry.npmjs.org/@babel/types/-/types-7.1.6.tgz",
-      "integrity": "sha512-DMiUzlY9DSjVsOylJssxLHSgj6tWM9PRFJOGW/RaOglVOK9nzTxoOMfTfRQXGUCUQ/HmlG2efwC+XqUEJ5ay4w==",
+      "version": "7.3.0",
+      "resolved": "https://registry.npmjs.org/@babel/types/-/types-7.3.0.tgz",
+      "integrity": "sha512-QkFPw68QqWU1/RVPyBe8SO7lXbPfjtqAxRYQKpFpaB8yMq7X2qAqfwK5LKoQufEkSmO5NQ70O6Kc3Afk03RwXw==",
       "requires": {
         "esutils": "^2.0.2",
         "lodash": "^4.17.10",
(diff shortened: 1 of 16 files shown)


mperrotti:
We might still want the last "spacing" column to be used to take up the remaining space on the page. @lsamimi - thoughts?

lsamimi:
@mperrotti Yes good idea! Though I do wonder how would this work when there are individual row action buttons on the right. No way to make this table dynamic to expand and contract with screen size?

GeorgiSTodorov:
@lsamimi , what buttons will there be on the right?

lsamimi:
@mperrotti In some tables, like the Services table, the three dot button that would have a dropdown of actions.

GeorgiSTodorov:
@lsamimi , if I undestand you correctly, the table is already dynamic as the empty column takes as much space as available.
peek 2019-01-17 09-53

This doesn't work if we want to have a column on the right of the empty column, but we will figure something out.

mperrotti:
@GeorgiSTodorov - we could always make the column that contains the icon fill the remaining width and align it's contents to the right

natmegs:
Not a blocker - but it seems like there's a lot of unused space on the right side of the table compared to the layout of other tables. Is this by design? I noticed that in other tables in the UI, the columns/content are spread to take up the entire table width
screen shot 2019-01-17 at 2 15 38 pm
vs
screen shot 2019-01-17 at 2 15 47 pm

GeorgiSTodorov:
@natmegs , we are using the new feature that @mperrotti added which enables each column to take as much space as needed and this is the first table that uses this new feature.

nLight:
rebase needed

GeorgiSTodorov:
Rebased. @mperrotti and @nLight please re-review.

package-lock.json
```diff
@@ -1001,3599 +1001,233 @@
...
+        "react-delegate-component": "1.0.0",
         "react-emotion": "9.2.12",
+        "react-toggled": "1.2.7",
         "react-transition-group": "2.5.0",
         "react-virtualized": "9.20.1",
-        "semantic-release": "^15.9.6"
+        "semantic-release": "15.9.6"
       },
       "dependencies": {
-        "@octokit/rest": {
```
nLight: Could be because of https://github.com/dcos-labs/ui-kit/commit/b687444e4a5230d829e9e9c51bcfb49d1e627b62#diff-b9cfc7f2cdf78a7f4b91a753d10865a2

GeorgiSTodorov: Could be. I tested on master and the same thing happened - a lot of lines in package-lock were deleted.

GeorgiSTodorov:
Fixed `package-lock` file conflict. @nLight , @natmegs , please re-review.

PR #3534, opened 2019-01-17
DCOS-47339: [1.12 backport] Update Table component in nodes page

GeorgiSTodorov:
## Testing
1. Go to Nodes tab
2. Verify that the table doesn't jump as the data is loaded
3. Verify that the table works and the columns are as wide as the content

## Trade-offs
I added a loader if the data array is empty, but some of the values in the table are slower to load, which causes it to move.

Also fixed some unrelated failing snapshots and some other minor things.

## Dependencies
None.

### Before
screenshot from 2019-01-17 10-37-58

### After
screenshot from 2019-01-17 10-42-28


plugins/nodes/src/js/components/NodesTable.tsx
```diff
@@ -181,6 +152,10 @@ export default class NodesTable extends React.Component<
   render() {
     const { data, sortColumn, sortDirection } = this.state;
 
+    if (data.length === 0) {
+      return <Loader />;
```
nLight: What if there's no nodes? Will it be stuck with the loader?

GeorgiSTodorov: Yes, you are right. I will think of something.

GeorgiSTodorov: Actually, no. If there are no nodes, it will display the no nodes message. 
screenshot from 2019-01-17 15-57-00

To test, change this line - https://github.com/dcos/dcos-ui/blob/91c0d1df7d22dbdf450f09c141cde1c17237d7bf/plugins/nodes/src/js/pages/NodesAgents.js#L305 .

PR #3532, opened 2019-01-15
DCOS-43911 SDK Plans v1

TattdCodeMonkey:
## Testing

1. Install a SDK package e.g. `cassandra`|`confluent-kafka` etc.
2. Visit the `services/detail` page for the installed service
3. Ensure the `Plans` tabs exists
4. Visit the `Plans` tab
5. Ensure there is a dropdown of plans
6. Plan details should be displayed in table below the dropdown
7. Change plan selected in dropdown
8. Plan details in table should update.

## Trade-offs

N/A

## Dependencies

None

## Screenshots

image
image
image

PR #3543, opened 2019-01-19
refactor(sdkplans): service.name -> service.id

TattdCodeMonkey:
Renaming the usage of Service.name in the SDK plans client and
data-layer. This is actually the Service.id and using the correct names
for things matter :)


GeorgiSTodorov:
Could this PR be part of https://github.com/dcos/dcos-ui/pull/3532 ?

plugins/services/src/js/data/index.ts
```diff
@@ -55,7 +55,7 @@ export const resolvers = ({
 }: ResolverArgs) => ({
   Service: {
     plans(parent: any, args: GeneralArgs = {}) {
-      if (!parent.name) {
+      if (!parent.id) {
         return Observable.throw(
           "Service name must be available to resolve plans"
```
GeorgiSTodorov: Should this say "Service ID"?

PR #3547, opened 2019-01-21
chore(CI): add check for .only on CI

DanielMSchmidt:
Closes DCOS-47525

## Testing

- Open a PR based on this one
- Add a ‘it.only‘ in a test case
- The CI should fail

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -45,8 +45,7 @@ pipeline {
       }
     }
 
-
-    stage("Lint Commits") {
+    stage("Verify Commits") {
       when {
         expression {
           !master_branches.contains(BRANCH_NAME)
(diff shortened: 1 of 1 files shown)


GeorgiSTodorov:
This is how it should be, right?
screenshot from 2019-01-22 11-13-45

DanielMSchmidt:
Wow, great catch @GeorgiSTodorov 

It seems like we had already build what we wanted to build in this JIRA :D 
Having an overview over big systems is hard I guess 🤷‍♂️

GeorgiSTodorov:
So the error was not because of this branch?

DanielMSchmidt:
Nope, this branch should have failed before.

PR #3554, opened 2019-01-24
Fix integration tests

nLight:
As commit messages suggest:

1. Mute terrible flakes
2. Record video only on 3rd re-run
3. Helper function that is not being used atm
4. Helper flags to run tests in the Headed mode
5. Fixed some `skipped` tests to reserve skip for flakes

### DIFF
--- a/scripts/utils/integration-tests
+++ b/scripts/utils/integration-tests
@@ -52,7 +52,7 @@ PLUGINS_PATH=${PLUGINS_PATH:-''}
 function executeCypress {
   npm run --silent cypress -- \
     --spec "${FILE}" \
-    --config "screenshotsFolder=${OUTPUT_PATH}/screenshots,videosFolder=${OUTPUT_PATH}/videos" \
+    --config "screenshotsFolder=${OUTPUT_PATH}/screenshots,video=${RECORD_VIDEO:-'true'},videosFolder=${OUTPUT_PATH}/videos" \
     --reporter junit \
     --reporter-options "mochaFile=${OUTPUT_PATH}/result.xml"
 }
@@ -69,9 +69,11 @@ function runTestFile {
   while [ $EXIT_CODE -ne 0 ] && [ $RUN -lt $RETRIES ]; do
     RUN=$((RUN+1))
     local OUTPUT_PATH="${PROJECT_ROOT}/${CYPRESS_FOLDER}/${RELATIVE_PATH}-${RUN}"
+    # Record video only on the 3rd run, videos are heavy
+    local RECORD_VIDEO=$([ "$RUN" == 3 ] && echo "true" || echo "false")
 
     echo "===> Executing Cypress (${RUN}/${RETRIES})"
-    CYOUT="$(FILE="${PROJECT_ROOT}/tests/${RELATIVE_PATH}" RUN=${RUN} OUTPUT_PATH=${OUTPUT_PATH} executeCypress)"
+    CYOUT="$(FILE="${PROJECT_ROOT}/tests/${RELATIVE_PATH}" RUN=${RUN} OUTPUT_PATH=${OUTPUT_PATH} RECORD_VIDEO=${RECORD_VIDEO} executeCypress)"
     EXIT_CODE=$?
 
     if [ $EXIT_CODE -eq 0 ]; then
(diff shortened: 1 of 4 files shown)


scripts/utils/integration-tests
```diff
@@ -51,8 +53,10 @@ PLUGINS_PATH=${PLUGINS_PATH:-''}
 # OUTPUT_PATH: Directory where results are saved (e.g. "some/path/to/Testfile-cy.js-1")
 function executeCypress {
   npm run --silent cypress -- \
+    ${HEADED:+--headed} \
+    ${NO_EXIT:+--no-exit} \
     --spec "${FILE}" \
-    --config "screenshotsFolder=${OUTPUT_PATH}/screenshots,videosFolder=${OUTPUT_PATH}/videos" \
+    --config "screenshotsFolder=${OUTPUT_PATH}/screenshots,video=${RECORD_VIDEO:-'true'},videosFolder=${OUTPUT_PATH}/videos" \
```
juliangieseke: 🎨  I'd not use defaults here, but only in the beginning of file/function

scripts/utils/integration-tests
```diff
@@ -199,6 +205,18 @@ function report {
 
 # finds the files and runs the tests
 function run_tests {
+  run_tests_one_by_one
```
juliangieseke: 🤔

scripts/utils/integration-tests
```diff
@@ -51,8 +53,10 @@ PLUGINS_PATH=${PLUGINS_PATH:-''}
 # OUTPUT_PATH: Directory where results are saved (e.g. "some/path/to/Testfile-cy.js-1")
 function executeCypress {
   npm run --silent cypress -- \
+    ${HEADED:+--headed} \
```
juliangieseke: 🎨 not strictly necessary, but I'd always quote strings.

scripts/utils/integration-tests
```diff
@@ -199,6 +205,18 @@ function report {
 
 # finds the files and runs the tests
 function run_tests {
+  run_tests_one_by_one
+}
+
+function run_all_tests {
```
juliangieseke: this isnt used, was it a test to see if all tests at once are faster?is it the case? if not used, should it be removed?

scripts/utils/integration-tests
nLight: @juliangieseke do you maybe know how to pass those flags in a better way? Currently one can write `run integration tests Spec dima julian` and still have both `--headed` and `--no-exit` :D

nLight: would `run integration tests Spec -- --headed --no-exit` work 🤔

nLight: https://jonalmeida.com/posts/2013/05/26/different-ways-to-implement-flags-in-bash/

juliangieseke: dealing with arguments in bash is a common issue (say: complicated) and one of the reasons a lot of scripts rely on ENV_VARS… various implementations of `getopt(s)` are buggy in some ways and you still have to write a lot of gluecode :/

scripts/utils/integration-tests
nLight: I'd like to keep it. I know it will be faster  :)

scripts/utils/integration-tests
```diff
@@ -39,7 +47,14 @@ TESTS_FOLDER="tests"
 
 PLUGINS_PATH=${PLUGINS_PATH:-''}
 
-
+while getopts 's:hnr:' flag; do
```
nLight: @juliangieseke I would still like giving it a try. Internet suggests `getopt` is buggy but `getopts` should be solid. More over it will be used only on local machines, CI runs with defaults

scripts/utils/integration-tests
nLight: There will be a switch :)

nLight:
@juliangieseke @DanielMSchmidt first build was green in 1h I'm inclined to merge the PR unless there are bigger concerns around the CLI I proposed.

https://jenkins.mesosphere.com/service/jenkins/blue/organizations/jenkins/frontend%2Fdcos-ui-oss-pipeline/detail/PR-3554/1/pipeline/67/

juliangieseke:
its ci… lets fix forward…

scripts/utils/integration-tests
juliangieseke: IIRC `getopts` on BSD had some issues, but they might be resolved by now 🤔 also its less flexible as `getopt`. But it should be good enough here 👍

PR #3564, opened 2019-01-25
Fix Modal issue

nLight:
This PR updates `reactjs-components` to address `ref` issue we were having in our Modals. That wasn't only causing integration flake but also has been reported via sentry as happening in production.

## Testing

Smoke test modals and other reactjs-components


GeorgiSTodorov:
Re-triggered system tests.

nLight:
Ok, the update breaks jobs system tests will have a look

PR #3566, opened 2019-01-29
DCOS-47800: [1.13] Typo: "Jobss" on jobs page search

GeorgiSTodorov:
Make the Job string to be pluralized only once

Closes https://jira.mesosphere.com/browse/DCOS-47800

## Testing
In Jobs tab search something when two or more jobs exist and verify that "Job" isn't pluralized twice.

## Trade-offs
When I updated the catalog messages, two lines unrelated to this PR appeared, we probably forgot them before.

When we have more than one job, an "s" is added to the end of the word, even in Chinese. We will fix this when we do the pluralization.

## Dependencies
None.

### Before
screenshot from 2019-01-29 13-16-21

### After
screenshot from 2019-01-29 13-19-10


plugins/jobs/src/js/components/JobsOverviewList.tsx
```diff
@@ -62,7 +62,7 @@ class JobsOverviewList extends React.Component<
     const headline = filter ? (
       <FilterHeadline
         onReset={this.resetFilter}
-        name={i18n._(t`Jobs`)}
+        name={i18n._(t`Job`)}
```
juliangieseke: does this play nicely together with plurals? I expect chinese translation to be something like `作业s` 🤔

GeorgiSTodorov: We do have a TODO - https://github.com/dcos/dcos-ui/pull/3566/files/f819aebd71b8ab9db99617a294c7fa0ee52a5cd7#diff-d8210cdb69c20fa16971122caf1cc169R59 .

juliangieseke: Ok 👍

GeorgiSTodorov:
Updated Trade-offs section.

PR #3570, opened 2019-01-31
[1.12] Backport unflakes

nLight:
Making 1.12 CI more stable just in case.


GeorgiSTodorov:
Could the new reactjs-components version break something?

nLight:
@GeorgiSTodorov potentially yes. Looking at the commits those were mostly fixes and dep upgrades

PR #3574, opened 2019-02-01
[spike] DCOS-20182: Error Validation in Delete Modal

GeorgiSTodorov:
Show an error to the user when something goes wrong
with deleting the service when we don't have details.

Closes https://jira.mesosphere.com/browse/DCOS-20182

## Testing
0. In `dcos-ui/plugins/services/src/js/events/MarathonActions.js` in the `deleteService` method above the `RequestUtil`, add something like `url = "123"`, in order to show an error.
Similarity, in `dcos/dcos-ui/src/js/events/CosmosPackagesActions.js` in the `uninstallPackage` method, hard-code packageName to something like "123".
1. Start a service.
2. Try to delete it.
3. Verify that an error is shown instead of nothing happening.
4. Close the modal and verify that a toast notification is shown.
5. Try the same for pods.
6. Try the same for catalog packages.
7. (Optional) Add some fake translations in the Chinese message catalog and verify that they work.

Thanks @TattdCodeMonkey for explaining to me the way toast works.

## Trade-offs
I am not sure how to create a real error, so I am mocking one and in this case it looks like an error won't display because no details are passed. So I decided to create a default error message, which could have some room for improvement. @mperrotti , what do you think?

The toast capitalizes the service name.

## Dependencies
None.

## Screenshots
Peek 2019-07-19 17-18

Peek 2019-07-19 17-18-2


plugins/services/src/js/components/modals/ServiceDestroyModal.js
```diff
@@ -122,7 +168,9 @@ class ServiceDestroyModal extends React.PureComponent {
     }
 
     return (
-      <h4 className="text-align-center text-danger flush-top">{errorMsg}</h4>
+      <Trans render="h4" className="text-align-center text-danger flush-bottom">
+        {errorMsg}
```
TattdCodeMonkey: this probably won't work. you can only use `Trans` like this for static text, for variables you need to mark them and then used the marked text as the `id` for the the `Trans` component. if we can't mark the text because it comes from the server or is built, then it can't be used inside a `Trans` component.

GeorgiSTodorov: I just checked and it works (changed language to Chinese and set "123" as translation).
Screenshot from 2019-03-29 16-38-07

TattdCodeMonkey: It doesn't break, but it also doesn't function as you would expect. using a `Trans` with just a variable inside extract's a message of `"{errorMsg}"`. That doesn't do anything and is confusing.

GeorgiSTodorov: Okay, I will remove the translation for now. It is out of scope anyway.

plugins/services/src/js/containers/services/ServicesContainer.js
```diff
@@ -350,7 +350,7 @@ class ServicesContainer extends React.Component {
         this.unsetPendingAction(ActionKeys.SERVICE_DELETE);
         break;
       case REQUEST_COSMOS_PACKAGE_UNINSTALL_ERROR:
-        this.unsetPendingAction(ActionKeys.SERVICE_EDIT, action.data);
```
GeorgiSTodorov: I think this was an error.

GeorgiSTodorov:
Thanks, @mperrotti ! I am not merging this to get one more opinion.

PR #3576, opened 2019-02-01
test: fix xit xdescribe

nLight:
Un-xit'ing some tests and see what happens :)


nLight:
Tests are green, verified locally. Deleted test file was obsolete as this functionality of the Tracking plugin has been removed long time ago.

PR #3577, opened 2019-02-04
DCOS-47925: remove manual trigger for rerun script

DanielMSchmidt:
## Testing

- Check if this branch has a job with `head` that was triggered automatically without a manual check and ran successfully without rerunning the tests
- Check if this branch has a job with `head` that was triggered automatically without a manual check and ran successfully with rerunning a subset of tests
- Check if a `.only` in an integration test can be commited without a warning, renders the `merge` test red, but triggers the `branch` test to run reruns

## Trade-offs

- Decided that ~10min of work for not running is okay, compared to maintaining an extra check before the tests run. As always, feel free to challenge me on this 😇

### DIFF
--- a/scripts/ci/rerun-integration-tests.sh
+++ b/scripts/ci/rerun-integration-tests.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
-
 set -e
+
 # configurable by environment variable
 RERUNS=${RERUNS:-100}
 
@@ -13,9 +13,13 @@ PROJECT_ROOT="$( cd "$( echo "${SCRIPT_PATH}" | sed s+/scripts/ci++)" && pwd )"
 source "$SCRIPT_PATH/../utils/test"
 source "$SCRIPT_PATH/../utils/integration-tests"
 
+# This aborts if no files are found
+if [ -z $(find_files_with_debug_directives "$PROJECT_ROOT/$TESTS_FOLDER") ]; then
+  echo "No files found to run, aborting"
+  exit 0
+fi
 
 # Get all files with .only in it
-# This aborts if no files are found
 FILES_TO_RUN="$(find_files_with_debug_directives "$PROJECT_ROOT/$TESTS_FOLDER")"
 
 setup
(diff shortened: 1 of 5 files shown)


scripts/ci/rerun-integration-tests.sh
```diff
@@ -12,10 +11,16 @@ PROJECT_ROOT="$( cd "$( echo "${SCRIPT_PATH}" | sed s+/scripts/ci++)" && pwd )"
 
 source "$SCRIPT_PATH/../utils/test"
 source "$SCRIPT_PATH/../utils/integration-tests"
-
+set +e
```
juliangieseke: IIRC you can catch exit code from `find_files_with_debug_directives` by directly executing if condition:
```suggestion
if [ find_files_with_debug_directives "$PROJECT_ROOT/$TESTS_FOLDER" ]; then
  echo "No files found to run, aborting"
  exit 0
fi
```

scripts/ci/rerun-integration-tests.sh
```diff
@@ -1,6 +1,5 @@
 #!/bin/bash
 
-set -e
```
juliangieseke: not sure if I like 🤔
wasnt it you telling me `set -e` is basicly mandatory? 😅

scripts/ci/rerun-integration-tests.sh
juliangieseke: also, as `set -e` is removed at the top of this file, this switch isnt necessary anymore, isnt it? 🤔

scripts/ci/rerun-integration-tests.sh
DanielMSchmidt: I re-set it, later on before the test runs start

scripts/ci/rerun-integration-tests.sh
DanielMSchmidt: I see, I inlined the subshell and that also worked without the `+e`, so I'm using that now

PR #3578, opened 2019-02-04
DCOS-47964 feat: add cosmos client & data-layer

TattdCodeMonkey:
Added data layer and cosmos client to get a package's available
versions.

Closes DCOS-47964

## Testing

For now run tests. The ui that uses this data-layer will be implemented as a part of another ticket.

## Trade-offs

None

## Dependencies
N/A


src/js/data/CosmosClient.ts
```diff
@@ -0,0 +1,28 @@
+import { request, RequestResponse } from "@dcos/http-service";
```
nLight: Let's wrap  the client as an npm package and move it to `packages/cosmos-client` 💪 with all package.json etc.

src/js/data/CosmosClient.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export interface PackageVersionsResults {
+  [key: string]: string;
+}
+
+export interface PackageVersionsResponse {
+  results: PackageVersionsResults;
+}
+
+export function fetchPackageVersions(
```
nLight: We could keep naming consistency `listVersions`

src/js/data/index.ts
```diff
@@ -0,0 +1,96 @@
+import { Observable, of, throwError, timer } from "rxjs";
```
nLight: Having the [reactive-graphql](https://github.com/mesosphere/reactive-graphql) part of the client embedded in DCOS UI would be perfectly fine. 
We could have a dedicated space for it, like `data/cosmos/index.ts` because `data/index.ts` would be the entry point, right?

src/js/data/index.ts
```diff
@@ -0,0 +1,96 @@
...
+
+export interface Query {
+  package: Package | null;
+}
+
+export default makeExecutableSchema({
+  typeDefs: schemas,
+  resolvers: resolvers({
+    fetchPackageVersions,
+    pollingInterval: Config.getRefreshRate()
```
nLight: Why would we be willing to poll the endpoint every 2 seconds?

src/js/data/CosmosClient.ts
juliangieseke: is something we want to do for all new clients? ref: #3579

src/js/data/CosmosClient.ts
```diff
@@ -0,0 +1,28 @@
...
+  packageName: string
+): Observable<RequestResponse<PackageVersionsResponse>> {
+  return request("/package/list-versions", {
```
juliangieseke: talking about consistency, it seems like some clients are handling error, some not. 🤔 

https://github.com/dcos/dcos-ui/pull/3579/files#diff-71851b32b10f6c1d63d2962b2b6f6e20R15

nLight: That's a good point, we don't have any particular guid or even a good example of a client. So I would say we're in the process of establishing those good practices. Having said that I believe handling errors is a good practice.

TattdCodeMonkey: So for SDK Plans we did retries, caught and threw errors from the client. I've been torn on if its better to do that in the client on in the data-layer. I do agree we should be consistent, but I'm torn on the better place to place that logic.

src/js/data/index.ts
TattdCodeMonkey: I know @DanielMSchmidt has been working on extension kid for data-layer, but for now what would be the best way to compose different parts of the data-layer? 

For the implementation would we make 2 `graphQLObservable`s, one for the cosmos query and another for the ui-update-service? I could in many ways it would make sense for that to be one query/`graphQLObservable` instead of two. But if there isn't a great way to combine these yet then would 2 be the correct approach?

src/js/data/index.ts
TattdCodeMonkey: It wouldn't be. I initially didn't have that query using the timer at all because I wasn't sure we _should_ poll packageVersions. But then added it back for consistency with other queries. 

How should we decided where its right to return a polling observable vs not, and what the polling interval should be?

TattdCodeMonkey:
@DanielMSchmidt so the UI is going to have to be a plugin, since we need different behavior between OSS & Enterprise. But the data-layer piece should be shared. But it could be a package? 

Pre [this comment](https://github.com/dcos/dcos-ui/pull/3578#discussion_r253877887) do you have any insights on the best way to compose disparate data-layer pieces, or if that is something we should even do now with Extension Kid on the horizon?

src/js/data/index.ts
nLight: That's a good question. While we need to answer it in general in this particular case we want to get this data once, as we visit the settings page. Otherwise I'd argue that we should rather put the polling into the rxjs client implementing a `poll` in addition to `request` and `stream`. Am I missing something?

TattdCodeMonkey: no, that sounds good to me. I'll remove the `timer` all together which I prefer in this case.

src/js/data/index.ts
DanielMSchmidt: > but for now what would be the best way to compose different parts of the data-layer

The way you would combine those is by passing the type definitions and the schema as an array to the `makeExecutableSchema` call. E.g. You could have a schema describing Nodes and one describing Networks that adds a field to the nodes by passing both into `makeExecutableSchema` with a `{ Nodes: { network(parent, args) {...} }}` field resolver and having a schema like this:

Node Schema:
```graphql
type Node {
  name: String!
}
```

Network Schema:
```graphql
type Network {
   ip: String!
   ipv6: Boolean!
}

extend type Node {
  network: Network!
}
```

This way you can compose smaller schemas into a single bigger one, which is the underlying mechanism we are going to use when we go with the extension-kid.


If I understand your use-case correctly you would like to expose a root level query for searching a single package, which is a good choice IMHO. And now you would like to have the data layer answer the question if there is any update and to which version, am I right?

I could see three ways to solve this:

1) You could have a new root level query `UICanBeUpdated: Boolean!` which uses the logic you provide in your cosmos API to determine if an update can be done. And you probably would do an `updateUi(Version: String!): Boolean!` mutation that performs the update
2) You could use a "nested DataLayer" approach and use the data from the cosmos API in the query and mutation. I'm unsure of the performance impact of this pattern, but I have thought about using it in general for a while now 🤔 
3) You could write one mutation `updateUi(dryRun: Boolean!, version: String!): String!` that has a dry-run functionality, not actually doing the change, but just answering the question to which version it would update. I have read about this pattern in the past and the author seemed to be convinced of it, but I never tried it out

packages/cosmos-client/src/__tests__/CosmosClient-test.ts
```diff
@@ -0,0 +1,67 @@
...
+import { take } from "rxjs/operators";
+
+import { CosmosClient } from "../CosmosClient";
+
+describe("CosmosClient", () => {
+  let client: CosmosClient;
+  beforeEach(() => {
+    jest.clearAllMocks();
+
+    client = new CosmosClient("/", "package");
```
pierrebeitz: Hi @TattdCodeMonkey,

[The docs](https://docs.mesosphere.com/1.10/deploying-services/package-api/#routes) seem to say that cosmos has 3 `apiPrefixes` (`/service`, `/package`, `/capabilities`).

Doesn't that mean that if we wanted to ask for capabilities, we currently had to instantiate a `new CosmosClient("/", "capabilities")` to then implement `listCapabilities` if we wanted to stick to [the pattern introduced here](https://github.com/dcos/dcos-ui/pull/3578/files#diff-4559883914d0dedebfd6b7d5d8a392e2R25)?
 
So if we wanted to implement `listCapabilities`, wouldn't we either be forced to not use the instance's `apiPrefix` or to make sure it's `/capabilities` just before we kick off the request?

src/js/types/Package.ts
```diff
@@ -0,0 +1,13 @@
+import { PackageVersion } from "#SRC/js/types/PackageVersion";
```
nLight: why not having them in `data/cosmos`?

packages/cosmos-client/src/__tests__/CosmosClient-test.ts
TattdCodeMonkey: Thats a good point. I was following the example from [here](https://github.com/dcos/dcos-ui/blob/master/src/js/events/CosmosPackagesActions.js#L300) but agree we should just include the prefix in the function's so that the client could support the different prefixes,

src/js/types/Package.ts
TattdCodeMonkey: That would work for me. I was following the example in the jobs plugin and we did the same thing for SDKPlans in the services plugin.

nLight: That's true and both jobs and services are separate plugins with a separate scope. Here we are in dcos-ui context, not a plugin. I wonder if we should be more specific when not in a plugin context.

TattdCodeMonkey: Roger, moved.

src/js/data/CosmosClient.ts
TattdCodeMonkey: I ended up moving the error handling to the client since it would be the best place for parsing the error message from the response.

src/js/data/cosmos/index.ts
```diff
@@ -0,0 +1,79 @@
...
+import Config from "#SRC/js/config/Config";
+import { Package, PackageSchema } from "#SRC/js/data/cosmos/Package";
+import { PackageVersionSchema } from "#SRC/js/data/cosmos/PackageVersion";
+
+export interface ResolverArgs {
+  cosmosClient: CosmosClient;
+}
+
+export interface GeneralArgs {
+  [key: string]: any;
```
natmegs: Just curious, why this instead of just `object` type?

src/js/data/cosmos/index.ts
```diff
@@ -0,0 +1,79 @@
...
+
+export const resolvers = ({ cosmosClient }: ResolverArgs) => ({
+  Package: {
+    versions(parent: { name: string }) {
+      if (!parent.name) {
+        return throwError("Package name must be available to resolve versions");
+      }
+
+      return of({}).pipe(
+        switchMap(() => cosmosClient.listPackageVersions(parent.name)),
```
natmegs: Seems indicative of something wrong that this can't start with `cosmosClient.listPackageVersions(parent.name).pipe(....`, still stumped by this

packages/cosmos-client/index.ts
```diff
@@ -0,0 +1,13 @@
...
+  [key: string]: string;
+}
+
+export interface PackageVersionsResponse {
+  results: ListVersionsResults;
+}
+
+module.exports = {
+  CosmosClient
+};
```
pierrebeitz: Why not `export { CosmosClient };`?

Asking since this is the first occurrence of my editor tripping up on typings. With the above everything works fine.

TattdCodeMonkey:
I’ll have to try that. I originally had

export * from ‘./src/CosmosClient’;

And it was failing at run-time. That’s why I moved the interfaces to this
file. I’m not sure why it wasn’t working.

packages/cosmos-client/index.ts
TattdCodeMonkey: I’ll have to try that. I originally had 
export * from ‘./src/CosmosClient’; 

And it was failing at run-time. That’s why I moved the interfaces to this file. I’m not sure why it wasn’t working.

The module.exports is basically me defaulting back to something I knew would work.

src/js/data/cosmos/index.ts
TattdCodeMonkey: Me too, I moved it around but this was they way the retry would function as desired.

src/js/data/cosmos/index.ts
TattdCodeMonkey: At least for me `object` will fail tslint in my editor when I try to access any key.

pierrebeitz: Would a union of all "specificArgs" do? So maybe `export type GeneralArgs = Partial< PackageQueryArgs >;`?

packages/cosmos-client/index.ts
pierrebeitz: Sound totally reasonable!

I toyed around with the code a little and found [this](https://github.com/dcos/dcos-ui/pull/3587/commits/ee5bdeca52f5bab3ded69e25e523d9bad2c6b830) to be working - at runtime and for my editors type checking.

src/js/data/cosmos/index.ts
```diff
@@ -0,0 +1,70 @@
...
+  package(name: String!): Package
+}
+`;
+export const schemas: string[] = [
+  PackageVersionSchema,
+  PackageSchema,
+  baseSchema
+];
+
+export interface Query {
```
pierrebeitz: Where's that used?

packages/cosmos-client/package.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "cosmos-client",
+  "main": "index.ts",
```
nLight: That wouldn't really work outside of our repo, but that's ok :)
we should be rather using `index.js` as `main` entry point. Providing types like `"types": "index.d.ts"`

packages/cosmos-client/src/CosmosClient.ts
```diff
@@ -0,0 +1,43 @@
...
+          packageName
+        }),
+        headers: {
+          Accept: buildRequestHeader(
+            "list-versions",
+            "response",
+            "package",
+            "v1"
+          ),
+
```
nLight: 🙈

packages/cosmos-client/src/CosmosClient.ts
```diff
@@ -0,0 +1,43 @@
...
+          "Content-type": buildRequestHeader(
+            "list-versions",
+            "request",
+            "package",
+            "v1"
+          )
+        }
+      }
+    ).pipe(
+      map((reqResp: RequestResponse<PackageVersionsResponse>) => {
```
nLight: Let's embrace `partition` as the pattern for error handling. Semantically `map` is supposed to be doing mapping, not conditional errors. 

https://github.com/dcos/dcos-ui/pull/3579/files#diff-23616281663aac4135b3d9a84fb60f84R15

packages/cosmos-client/src/utils.ts
```diff
@@ -0,0 +1,29 @@
...
+export function buildRequestHeader(
+  action: string,
+  actionType: string,
+  entity: string,
+  version: string
+) {
+  return `application/vnd.dcos.${entity}.${action}-${actionType}+json;charset=utf-8;version=${version}`;
+}
+
+export function getErrorMessage(reqResp: RequestResponse<any>): string {
```
nLight: Let's use the full form `reqResp` -> `requestResponse` or even destruct the argument right there into `{ response, message }`.

nLight: `RequestResponse<any>` can we be more specific and avoid `any` here?

src/js/data/cosmos/index.ts
```diff
@@ -0,0 +1,70 @@
...
+import { PackageVersionSchema } from "#SRC/js/data/cosmos/PackageVersion";
+
+export type GeneralArgs = Partial<PackageQueryArgs>;
+
+export interface PackageQueryArgs {
+  name: string;
+}
+
+function isPackageQueryArgs(args: GeneralArgs): args is PackageQueryArgs {
+  return args.name !== undefined;
```
nLight: is it ok for name to be `null`?

src/js/data/cosmos/index.ts
```diff
@@ -0,0 +1,70 @@
...
+import { RequestResponse } from "@dcos/http-service";
+import { CosmosClient, PackageVersionsResponse } from "cosmos-client";
+
+import { Package, PackageSchema } from "#SRC/js/data/cosmos/Package";
+import { PackageVersionSchema } from "#SRC/js/data/cosmos/PackageVersion";
+
+export type GeneralArgs = Partial<PackageQueryArgs>;
+
+export interface PackageQueryArgs {
+  name: string;
```
nLight: non-nullable 👮 see below

packages/cosmos-client/src/CosmosClient.ts
Poltergeist: Let's not embrace partition since partition is intended to be deprecated at some point reactivex/rxjs#3797

Lets rather solve it with filters: 

```js
const [success, error] = stream.pipe(map(source => 
  [
    source.pipe(filter(({code})=> code <= 200)), 
    source.pipe(filter(({code}) => code > 200))
  ]
))

```

This is basically what partition is doing and it is cleaner and can be used for multiple cases and not only binary cases.

src/js/data/cosmos/index.ts
Poltergeist: `args.name != null` should be the better check or not?

packages/cosmos-client/src/CosmosClient.ts
nLight: That's fair! Thanks for the research!

src/js/data/cosmos/index.ts
pierrebeitz: actually it's typed to be a string, so checking whether it's actually a string seems even more reasonable to me.

packages/cosmos-client/src/utils.ts
pierrebeitz: We currently seem to have a working solution, that parses out error messages for all kinds of responses - those being rightfully represented by the `any` IMHO.

Modelling types in order to not have an any here actually would have quite some impact, given that we

1. do not seem have a [properly modelled](https://package.elm-lang.org/packages/elm/http/latest/Http#Error) errors for `RequestResponse` yet.
2. seem to have no documentation about the errors that might be returned by the Cosmos-API.
3. seem to not have an established pattern for how we want to model error-streams with RxJS yet, if i groked the discussions regarding `partition` and alike correctly.

Being an advocate of elm, where everything is typed in a sound way, I'd still propose to just go with the any for now.

packages/cosmos-client/package.json
pierrebeitz: we'll go with "but that's ok" for now.

packages/cosmos-client/src/CosmosClient.ts
pierrebeitz: I hope you meant that there's a newline.

packages/cosmos-client/src/CosmosClient.ts
pierrebeitz: Looks like a foundation/architecture-thingy to me.

Let's agree on a convention and then have this adapted afterwards.

nLight: Let's please go with @Poltergeist suggestion merging two filtered streams

src/js/data/index.ts
pierrebeitz: Does this (still) yield some action item? 

If so, could someone summarise the comments above to something that's digestible for me?
If not, please resolve!

packages/cosmos-client/src/CosmosClient.ts
TattdCodeMonkey: Should we return two streams from the client and then merge them in the data-layer?

Filtering and merging in the client is a lot of code for what is accomplished with 4 lines in a map. I implemented in the ui-service client but immediately reverted bc it seemed wrong.

pierrebeitz: we don't misuse `map` anymore but instead go with `tap`.

packages/cosmos-client/src/utils.ts
pierrebeitz: Addressed the first comment. If there are no objections we'll leave the any here for now.

PR #3581, opened 2019-02-05
DCOS-47965 UI-Update Service client & data-layer

TattdCodeMonkey:
Adding client and data-layer for making queries and requests to the
dcos-ui-update-service.

## Testing

Run unit tests, this will be used and testable by an upcoming feature.

### DIFF
--- a/packages/dcos-ui-update-client/src/DCOSUIUpdateClient.ts
+++ b/packages/dcos-ui-update-client/src/DCOSUIUpdateClient.ts
@@ -1,0 +1,35 @@
+import { Observable } from "rxjs";
+import { tap } from "rxjs/operators";
+import { request, RequestResponse } from "@dcos/http-service";
+
+import { handleBadResponses } from "./utils";
+
+import { UIVersionResponse } from "../";
+
+export const DCOSUIUpdateClient = {
+  fetchVersion: (
+    rootUrl: string = "/"
+  ): Observable<RequestResponse<UIVersionResponse>> => {
+    return request<UIVersionResponse>(
+      `${rootUrl}dcos-ui-update-service/api/v1/version/`
+    ).pipe(tap(handleBadResponses));
+  },
+  updateUIVersion: (
+    newPackageVersion: string,
+    rootUrl: string = "/"
+  ): Observable<RequestResponse<string>> => {
+    return request<string>(
+      `${rootUrl}dcos-ui-update-service/api/v1/update/${newPackageVersion}/`,
+      {
+        method: "POST"
+      }
+    ).pipe(tap(handleBadResponses));
+  },
+  resetUIVersion: (
+    rootUrl: string = "/"
+  ): Observable<RequestResponse<string>> => {
+    return request<string>(`${rootUrl}dcos-ui-update-service/api/v1/reset/`, {
+      method: "DELETE"
+    }).pipe(tap(handleBadResponses));
+  }
+};
--- a/packages/dcos-ui-update-client/src/utils.ts
+++ b/packages/dcos-ui-update-client/src/utils.ts
@@ -1,0 +1,18 @@
+import { RequestResponse } from "@dcos/http-service";
+
+export function getErrorMessage(reqResp: RequestResponse<any>): string {
+  const { response, message } = reqResp;
+  if (typeof response === "string") {
+    return response;
+  }
+  if (message) {
+    return message;
+  }
+  return "An error occurred.";
+}
+
+export function handleBadResponses(requestResponse: RequestResponse<any>) {
+  if (requestResponse.code >= 300) {
+    throw new Error(getErrorMessage(requestResponse));
+  }
+}
--- a/src/js/data/ui-update/index.ts
+++ b/src/js/data/ui-update/index.ts
@@ -1,0 +1,132 @@
+import { Observable, of, throwError } from "rxjs";
+import { map, publishReplay, refCount, retry, switchMap } from "rxjs/operators";
+import { RequestResponse } from "@dcos/http-service";
+import { makeExecutableSchema } from "graphql-tools";
+import { DCOSUIUpdateClient, UIVersionResponse } from "dcos-ui-update-client";
+
+import {
+  UIMetadata,
+  UIMetadataSchema
+} from "#SRC/js/data/ui-update/UIMetadata";
+import Config from "#SRC/js/config/Config";
+
+type PossibleMutationArgs = Partial<UIUpdateArgs>;
+
+interface QueryContext {
+  [key: string]: any;
+  versionStream?: Observable<RequestResponse<UIVersionResponse>>;
+}
+
+export interface UIUpdateArgs {
+  newVersion: string;
+}
+
+function isUIUpdateArgs(args: PossibleMutationArgs): args is UIUpdateArgs {
+  return typeof args.newVersion === "string";
+}
+
+export const resolvers = {
+  UIMetadata: {
+    clientBuild() {
+      return of(window.DCOS_UI_VERSION);
+    },
+    packageVersion(_parent = {}, _args = {}, context: QueryContext) {
+      if (!context.versionStream) {
+        return throwError(
+          "Version observable not available on query context object"
+        );
+      }
+      const version$ = context.versionStream;
+      return version$.pipe(
+        map(
+          ({ response }: RequestResponse<UIVersionResponse>) =>
+            response.packageVersion
+        )
+      );
+    },
+    packageVersionIsDefault(_parent = {}, _args = {}, context: QueryContext) {
+      if (!context.versionStream) {
+        return throwError(
+          "Version observable not available on query context object"
+        );
+      }
+      const version$ = context.versionStream;
+      return version$.pipe(
+        map(
+          ({ response }: RequestResponse<UIVersionResponse>) => response.default
+        )
+      );
+    },
+    serverBuild(_parent = {}, _args = {}, context: QueryContext) {
+      if (!context.versionStream) {
+        return throwError(
+          "Version observable not available on query context object"
+        );
+      }
+      const version$ = context.versionStream;
+      return version$.pipe(
+        map(
+          ({ response }: RequestResponse<UIVersionResponse>) =>
+            response.buildVersion
+        )
+      );
+    }
+  },
+  Query: {
+    ui(_parent = {}, _args = {}, context: QueryContext = {}) {
+      context.versionStream = of({}).pipe(
+        switchMap(() => DCOSUIUpdateClient.fetchVersion(Config.rootUrl)),
+        retry(2),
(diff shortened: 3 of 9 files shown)


src/js/data/ui-update/index.ts
```diff
@@ -0,0 +1,131 @@
...
+          map(
+            ({ response }: RequestResponse<UIVersionResponse>) =>
+              response.default
+          )
+        );
+      }
+    },
+    Query: {
+      ui(_parent = {}, _args = {}, context: QueryContext = {}) {
+        context._versionStream = of({}).pipe(
```
DanielMSchmidt: You could put this stream also just on the top of the function instead of here, so that you don't need to use the context.


```js
export function resolvers({ uiUpdateClient }: ResolverArgs) {
  const version$ = of({})....

  return {
}
```

src/js/data/ui-update/index.ts
```diff
@@ -0,0 +1,131 @@
...
+  resetDCOSUI: String!
+}
+`;
+
+const schemas = [UIVersionSchema, baseSchema];
+
+export function makeSchema() {
+  return makeExecutableSchema({
+    typeDefs: schemas,
+    resolvers: resolvers({
```
DanielMSchmidt: Normally this would be the point where we'd add a polling interval, does this make sense here too?

src/js/data/ui-update/index.ts
```diff
@@ -0,0 +1,131 @@
...
+            )
+          );
+      }
+    }
+  };
+}
+
+const baseSchema = `
+type Query {
+  ui: UIVersion!
```
DanielMSchmidt: Nitpick, but isn't it more of a set of metadata for the ui that we are modelling here?

src/js/data/ui-update/index.ts
TattdCodeMonkey: It is, we didn’t want to poll with any of these resolvers so I left it out for now.

src/js/data/ui-update/index.ts
TattdCodeMonkey: I like `UIMetadata` as a name here much more. 👍 thank you, naming things is hard.

src/js/data/ui-update/index.ts
TattdCodeMonkey: Not currently since no resolvers are polling.

DanielMSchmidt: 👍

src/js/data/ui-update/index.ts
DanielMSchmidt: 👍 awesome :)

src/js/data/ui-update/index.ts
DanielMSchmidt: I think we did not talk about the same thing here.

In this field resolver you create a obervable, which is subscribed to later on by different other field resolvers. While this is a great pattern I think we should try to use the context as little as possible, at best only to pass meta-data for the request from one level to the other. The observable you have here could also be held outside of any resolvers and subscribed to by a normal object reference, which seems to me way less error prone (e.g. eslint will cry if you rename that reference, but not if you mask it through an object). It's no blocker for me, I just wanted to be clear on what I meant before moving on 😇

TattdCodeMonkey: @DanielMSchmidt I don't think we can. Since we are not polling creating it at the top of the function would result in the API only being called a single time.

I put it here so that the stream is created when you make a query that includes `ui`, but is still shared between the `ui` field resolvers. That way we make 1 new request per query, if any of the field resolver's need to call the ui-service API. Other wise the stream is not executed and the API request is not made. At least based on the unit tests and manually testing that is the desired behavior.

DanielMSchmidt: Ah, damn, now I understand what you mean. I didn't make the connection between your words and what you meant with polling there :D please carry on

packages/dcos-ui-update-client/src/DCOSUIUpdateClient.ts
```diff
@@ -0,0 +1,57 @@
...
+    );
+  },
+  resetUIVersion: (): Observable<RequestResponse<string>> => {
+    return request(
+      `${Config.rootUrl || "/"}dcos-ui-update-service/api/v1/reset/`,
+      {
+        method: "DELETE"
+      }
+    ).pipe(
+      map((reqResp: RequestResponse<any>) => {
```
nLight: Let's stick to `partition` and ditch abbreviations as mentioned in pervious PR 💪

packages/dcos-ui-update-client/src/DCOSUIUpdateClient.ts
```diff
@@ -0,0 +1,57 @@
...
+    newPackageVersion: string
+  ): Observable<RequestResponse<string>> => {
+    return request(
+      `${Config.rootUrl ||
+        "/"}dcos-ui-update-service/api/v1/update/${newPackageVersion}/`,
+      {
+        method: "POST"
+      }
+    ).pipe(
+      map((reqResp: RequestResponse<any>) => {
```
nLight: `partition`

packages/dcos-ui-update-client/src/DCOSUIUpdateClient.ts
```diff
@@ -0,0 +1,57 @@
...
+
+import { UIVersionResponse } from "../";
+
+export const DCOSUIUpdateClient = {
+  fetchVersion: (): Observable<RequestResponse<UIVersionResponse>> => {
+    const resp$ = request(
+      `${Config.rootUrl || "/"}dcos-ui-update-service/api/v1/version/`
+    );
+    return resp$.pipe(
+      map((reqResp: RequestResponse<any>) => {
```
nLight: and here: `partition` and naming

packages/dcos-ui-update-client/src/DCOSUIUpdateClient.ts
```diff
@@ -0,0 +1,57 @@
...
+
+import Config from "#SRC/js/config/Config";
+import { getErrorMessage } from "./utils";
+
+import { UIVersionResponse } from "../";
+
+export const DCOSUIUpdateClient = {
+  fetchVersion: (): Observable<RequestResponse<UIVersionResponse>> => {
+    const resp$ = request(
+      `${Config.rootUrl || "/"}dcos-ui-update-service/api/v1/version/`
```
nLight: I'm sorry having overlooked this. Let's try to avoid importing `#SRC` in packages. We shall instead make it an optional argument. Please note that in my `config.ts` I have `rootUrl: ""` which will result in `dcos-ui-update-service/api/v1/version/` without the leading slash.

packages/dcos-ui-update-client/src/DCOSUIUpdateClient.ts
nLight: `<any>`

packages/dcos-ui-update-client/src/utils.ts
```diff
@@ -0,0 +1,12 @@
+import { RequestResponse } from "@dcos/http-service";
+
+export function getErrorMessage(reqResp: RequestResponse<any>): string {
```
nLight: `<any>`

nLight: also looks like we need a shared utils package or maybe adding that function to `http-service`

src/js/data/ui-update/index.ts
```diff
@@ -0,0 +1,137 @@
...
+  [key: string]: any;
+  _versionStream?: Observable<RequestResponse<UIVersionResponse>>;
+}
+
+export interface UIUpdateArgs {
+  newVersion: string;
+}
+
+function isUIUpdateArgs(args: GeneralArgs): args is UIUpdateArgs {
+  return (args as UIUpdateArgs).newVersion !== undefined;
```
nLight: why strict check for `undefined`?

packages/dcos-ui-update-client/src/DCOSUIUpdateClient.ts
TattdCodeMonkey: Is `Config.rootUrl` still used and useful? I added it after seeing it in the current [CosmosPackagesActions](https://github.com/dcos/dcos-ui/blob/master/src/js/events/CosmosPackagesActions.js#L62) because it seemed like it would be good to be consistent.

nLight: We could keep the prefix just making it an argument instead of environment related global

packages/dcos-ui-update-client/src/utils.ts
TattdCodeMonkey: This one is `any` on purpose. If different services return errors in the same way then having a shared util would make sense. I ended up putting these in the client specifically so that we could have service specific logic for building error messages from responses. 

Cosmos may return error messages in a different way then the `ui-update-service` etc. so this code could and likely would be slightly different based on which client it is.

packages/dcos-ui-update-client/src/DCOSUIUpdateClient.ts
Poltergeist: In which cases do we actually need the `rootUrl`? Does it even make sense to use it?

nLight: @Poltergeist you are longer on the team :bad_poker_face: we could do git archeology in `mesosphere/dcos-ui`

src/js/data/ui-update/index.ts
```diff
@@ -0,0 +1,137 @@
...
+        map(
+          ({ response }: RequestResponse<UIVersionResponse>) =>
+            response.buildVersion
+        )
+      );
+    }
+  },
+  Query: {
+    ui(_parent = {}, _args = {}, context: QueryContext = {}) {
+      context._versionStream = of({}).pipe(
```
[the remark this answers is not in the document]
TattdCodeMonkey: I'll remove the `_`, that's a #polyglotproblems. In python you prefix "private" properties/methods with `_` to denote them as "private". `versionStream` is intended to only be used by field resolvers and not set when making a query or mutation.

src/js/data/ui-update/index.ts
```diff
@@ -0,0 +1,137 @@
...
+}
+
+function isUIUpdateArgs(args: GeneralArgs): args is UIUpdateArgs {
+  return (args as UIUpdateArgs).newVersion !== undefined;
+}
+
+export const resolvers = {
+  UIMetadata: {
+    clientBuild() {
+      // @ts-ignore
```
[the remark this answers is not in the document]
TattdCodeMonkey: Turns out I can removed this one, while the editor yelled at me for `DCOS_UI_VERSION` not existing on `window` our tslint config didn't mind.

PR #3586, opened 2019-02-08
DCOS-44841: add network node resolver

Poltergeist:
This adds the nodes resolver using the nodes network client to generate nodes. This is a workaround
since the nodes resolver does not actually return nodes yet.

CLOSES DCOS-44841

## Testing

Check the tests for plausibility 

## Trade-offs

N/A

## Dependencies

This is depending on #3579 

## Screenshots

N/A

### DIFF
--- a/plugins/nodes/src/js/data/NodesNetworkClient.ts
+++ b/plugins/nodes/src/js/data/NodesNetworkClient.ts
@@ -1,25 +1,43 @@
-import { Observable, merge } from "rxjs";
 import { request, RequestResponse } from "@dcos/http-service";
+import { merge, Observable } from "rxjs";
 import { partition, tap } from "rxjs/operators";
 
 export interface NodeNetwork {
-  updated: string;
+  updated: string | Date;
   public_ips: string[];
   private_ip: string;
   hostname: string;
 }
 
-export type NodesNetwork = NodeNetwork[];
+export interface NodeNetworkResponse extends NodeNetwork {
+  updated: string;
+}
 
-export function fetchNodesNetwork(): Observable<RequestResponse<NodesNetwork>> {
+export type NodesNetwork = NodeNetwork[];
+export type NodesNetworkResponse = NodeNetworkResponse[];
+
+export const NodeNetworkSchema = `
+scalar Date
+
+type NodeNetwork {
+  updated: Date
+  public_ips: [String]
+  private_ip: String
+  hostname: String
+}
+`;
+
+export function fetchNodesNetwork(): Observable<
+  RequestResponse<NodesNetworkResponse>
+> {
   const [success, error] = partition(
-    (response: RequestResponse<NodesNetwork>) => response.code < 300
+    (response: RequestResponse<NodesNetworkResponse>) => response.code < 300
   )(request("/net/v1/nodes"));
 
   return merge(
     success,
     error.pipe(
-      tap((response: RequestResponse<NodesNetwork>) => {
+      tap((response: RequestResponse<NodesNetworkResponse>) => {
         const responseMessage =
           response.response && typeof response.response === "object"
             ? JSON.stringify(response.response)
--- a/plugins/nodes/src/js/data/NodesResolver.ts
+++ b/plugins/nodes/src/js/data/NodesResolver.ts
@@ -1,0 +1,124 @@
+import { RequestResponse } from "@dcos/http-service";
+import { makeExecutableSchema } from "graphql-tools";
+import { Observable, of, throwError, timer } from "rxjs";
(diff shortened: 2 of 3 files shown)


GeorgiSTodorov:
Build is failing, because
```
[cos-ui-oss-pipeline_PR-3586-PJ6NYP23X7UCWVN36KKJETLOVGNHJNHK6L5NBKTJNGQX3HTRJYMQ] Running shell script

+ npm run commitlint -- --from ph/feature/DCOS-44840-network-nodes-client


> dcos-ui@0.0.0-dev commitlint /jenkins/workspace/cos-ui-oss-pipeline_PR-3586-PJ6NYP23X7UCWVN36KKJETLOVGNHJNHK6L5NBKTJNGQX3HTRJYMQ

> commitlint "--from" "ph/feature/DCOS-44840-network-nodes-client"


/jenkins/workspace/cos-ui-oss-pipeline_PR-3586-PJ6NYP23X7UCWVN36KKJETLOVGNHJNHK6L5NBKTJNGQX3HTRJYMQ/node_modules/@commitlint/cli/lib/cli.js:70

		throw err;

		^


Error: fatal: ambiguous argument 'ph/feature/DCOS-44840-network-nodes-client..HEAD': unknown revision or path not in the working tree.

Use '--' to separate paths from revisions, like this:

'git <command> [<revision>...] -- [<file>...]'
```

GeorgiSTodorov:
File conflict.

Poltergeist:
I just rebase 👍

GeorgiSTodorov:
Do we not need to write tests like https://github.com/dcos/dcos-ui/pull/3589/files ?
With `fakeShedulers` and `of`.

Poltergeist:
I took another approach to the problem which @DanielMSchmidt pointed out to me here: https://daniel-s.de/2019/01/testing-rxjs-side-effects/ So I took the `async` | `await` approach. if I should use the `fakeSchedulars` please let me know.

plugins/nodes/src/js/data/NodesNetworkClient.ts
```diff
@@ -11,6 +11,15 @@ export interface NodeNetwork {
 
 export type NodesNetwork = NodeNetwork[];
 
+export const NodeNetworkSchema = `
+type NodeNetwork {
+  updated: String
```
GeorgiSTodorov: Isn't this a Date?

plugins/nodes/src/js/data/NodesResolver.ts
```diff
@@ -0,0 +1,119 @@
...
+export interface NodeQueryArgs {
+  privateIP: string;
+}
+
+export interface NodesQueryArgs {
+  privateIPs: string[];
+}
+
+function isNodeQueryArgs(args: GeneralArgs): args is NodeQueryArgs {
+  return (args as NodeQueryArgs).privateIP !== undefined;
```
GeorgiSTodorov: Should we also check if it is different than empty string?

plugins/nodes/src/js/data/NodesNetworkClient.ts
Poltergeist: It is an ISO datetime string.

GeorgiSTodorov: Do we check somewhere if it's the correct format and not just any string?

plugins/nodes/src/js/data/NodesResolver.ts
Poltergeist: We are checking here if it is defined since the resolver would fail if it isn't defined. With an empty string it would still work but not find any matching nodes which is fine, as long as it doesn't fail.

plugins/nodes/src/js/data/NodesResolver.ts
```diff
@@ -0,0 +1,119 @@
...
+}
+
+function isNodesQueryArgs(args: GeneralArgs): args is NodesQueryArgs {
+  return (args as NodesQueryArgs).privateIPs.length > 0;
```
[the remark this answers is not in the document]
nLight: Do we additionally check anywhere that field is defined? Why is it important to have at least 1 private ip instead of just checking `Array.isArray`?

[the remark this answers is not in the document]
Poltergeist: If you do not provide at least 1 private ip the resolver would not return any node. Although I can change to `Array.isArray`.

plugins/nodes/src/js/data/NodesResolver.ts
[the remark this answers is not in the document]
Poltergeist: Will change.

PR #3588, opened 2019-02-11
add instructions for setting up private-plugins

pierrebeitz:
As i stumbled upon setting up the private-plugins, here's a suggestion to update the README.md.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,17 +39,33 @@ Node 8.9.x and NPM 5.6.x (and above) are **required**. We suggest using [nvm](ht
   ```sh
   npm run scaffold
   ```
-  This will create `webpack/proxy.dev.js` and `src/js/config/Config.dev.ts`. These files aren't tracked by git and provide a place to override proxy and general dcos-ui configuration. Edit `webpack/proxy.dev.js` to point to the cluster address.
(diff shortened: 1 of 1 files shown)


GeorgiSTodorov:
I don't understand this change, what problems did you experience?

pierrebeitz:
@GeorgiSTodorov 

> I don't understand this change, what problems did you experience?

1. I skipped [this line](https://github.com/dcos/dcos-ui/pull/3588/files#diff-04c6e90faac2675aa89e2176d2eec7d8L42) after `...aren't tracked by git...` as i thought no more important information will follow. So i put the quite important "Edit `webpack/proxy.dev.js` to point to the cluster address." On a separate line. You may very well blame my way of reading readmes, but i guess i'm not the only trying to filter out as much information as quickly as possible - so this change should not be a problem.

2. I followed all points of the "DC/OS Installation Instructions" and then was greeted by a broken system that i could not log into because i did not do the stuff mentioned in "Using External Plugins" as it did not occur in the steps. So i put it there. Again you may blame me not reading the whole document at once. Again this might help future me in setting things up again.

3. Due to hardcoded stuff, the `ui-plugins-private` must be at a very specific path. I added that information.

4. There's a handy configuration for when you want to go enterprise (Config.template.js). I added a reference to it, as else you just have the default configuration with no idea on what to do to get EE running as this file also is not mentioned in the `ui-plugins-private`-README.

PR #3589, opened 2019-02-11
Fix RxJS tests: ensure all assertions are called

natmegs:
Ensure all test assertions are being called correctly, even when relying on an async observable subscription.

This is an issue in unit tests using `rxjs-marbles` with non-marble tests that rely on `subscribe` calls.

## Testing

Unit tests pass.

## Trade-offs

N/A

## Dependencies

N/A

## Screenshots

N/A

### DIFF
--- a/plugins/jobs/src/js/data/__tests__/JobModel-test.js
+++ b/plugins/jobs/src/js/data/__tests__/JobModel-test.js
@@ -1,6 +1,6 @@
 import { concat, of } from "rxjs";
 import { take, map } from "rxjs/operators";
-import { marbles } from "rxjs-marbles/jest";
+import { marbles, fakeSchedulers } from "rxjs-marbles/jest";
 import { resolvers } from "../JobModel";
 
 const defaultJobDetailData = {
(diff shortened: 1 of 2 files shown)


plugins/jobs/src/js/data/__tests__/JobModel-test.js
```diff
@@ -234,9 +234,9 @@ describe("JobModel Resolver", () => {
             jobsQuery({}, { path: [] }).pipe(take(1)),
             jobsQuery({}, { path: [] }).pipe(take(1)),
             jobsQuery({}, { path: [] }).pipe(take(1))
-          ).subscribe(jest.fn(), jest.fn(), () => {
-            expect(fetchJobs).toHaveBeenCalledTimes(1);
-          });
+          ).subscribe();
+          advance(20);
```
nLight: that's cool!

nLight: @DanielMSchmidt something we could try using in the data-layer extension plugin.

natmegs:
@TattdCodeMonkey Good catch, not sure how I missed that 😳 Added another commit for that test but I think it uncovered an actual legit test failure (I `.skip`ed them in the commit). I do not see a mechanism for sharing observable subscriptions in the code... @nLight do you know if I'm missing something?

TattdCodeMonkey:
Oh yes, I solved this when testing the new ui service data layer. Let’s sync up tomorrow and we can fix this.

natmegs:
@TattdCodeMonkey I removed the faulty tests as discussed

PR #3590, opened 2019-02-11
DCOS-44842: add public ip component

Poltergeist:
## Testing

Visit the nodes page this should display the public ip of the public available nodes.

## Trade-offs

The cell is refetching the public ip if a cell added to the viewport since only the new component is bound to the resolver. Alternative could be to move the resolver logic up above the table and merge all the nodes information with the available network information.

## Dependencies

#3586 

## Screenshots

image

### DIFF
--- a/plugins/nodes/src/js/columns/NodesTablePublicIPColumn.tsx
+++ b/plugins/nodes/src/js/columns/NodesTablePublicIPColumn.tsx
@@ -1,61 +1,25 @@
-import * as React from "react";
 import { TextCell } from "@dcos/ui-kit";
-
-import { default as schema } from "../data/NodesResolver";
-import { NodeNetwork } from "../data/NodesNetworkClient";
-import graphqlObservable from "reactive-graphql";
-import gql from "graphql-tag";
-import { map, switchMap, startWith } from "rxjs/operators";
-import { componentFromStream } from "@dcos/data-service";
-import { Observable } from "rxjs";
+import * as React from "react";
+import { Tooltip } from "reactjs-components";
 import { Trans } from "@lingui/react";
 
-const getGraphQl = (privateIP: string) =>
-  graphqlObservable<{ node: { network: NodeNetwork } }>(
-    gql`
-      query {
-        node(privateIP: $privateIP) {
-          network {
-            public_ips
-          }
-        }
-      }
-    `,
-    schema,
-    { privateIP }
-  ).pipe(map(response => response.data));
+import Node from "src/js/structs/Node";
 
-const Column = componentFromStream<{ privateIP: string }>(
-  (props$): Observable<React.ReactNode> => {
-    return (props$ as Observable<{ privateIP: string }>).pipe(
-      switchMap(({ privateIP }) => getGraphQl(privateIP)),
-      map(({ node }: { node: { network: { public_ips: string[] } } }) => {
-        if (
-          node.network == null ||
-          node.network.public_ips == null ||
-          node.network.public_ips.length === 0
-        ) {
-          return (
-            <TextCell>
-              <Trans>N/A</Trans>
-            </TextCell>
-          );
-        }
-        return <TextCell>{node.network.public_ips.join(", ")}</TextCell>;
-      }),
-      startWith(
-        <TextCell>
-          <Trans>Loading</Trans>
-        </TextCell>
-      )
+export default function(item: Node) {
+  const publicIps = item.getPublicIps();
+  if (publicIps.length === 0) {
+    return (
+      <TextCell>
+        <Trans>N/A</Trans>
+      </TextCell>
     );
   }
-);
-
-export function publicIPCellWidth(): number {
-  return 125;
-}
-
-export default function({ hostname }: { hostname: string }) {
-  return <Column privateIP={hostname} />;
+  if (publicIps.length === 1) {
+    return <TextCell>{publicIps[0]}</TextCell>;
+  }
+  return (
+    <TextCell>
+      <Tooltip content={publicIps.join(", ")}>{publicIps[0]}</Tooltip>
+    </TextCell>
+  );
 }
--- a/plugins/nodes/src/js/columns/NodesTablePublicIPColumn.tsx
+++ b/plugins/nodes/src/js/columns/NodesTablePublicIPColumn.tsx
@@ -1,0 +1,61 @@
+import * as React from "react";
+import { TextCell } from "@dcos/ui-kit";
+
+import { default as schema } from "../data/NodesResolver";
+import { NodeNetwork } from "../data/NodesNetworkClient";
+import graphqlObservable from "reactive-graphql";
+import gql from "graphql-tag";
+import { map, switchMap, startWith } from "rxjs/operators";
+import { componentFromStream } from "@dcos/data-service";
+import { Observable } from "rxjs";
+import { Trans } from "@lingui/react";
+
+const getGraphQl = (privateIP: string) =>
+  graphqlObservable<{ node: { network: NodeNetwork } }>(
+    gql`
+      query {
+        node(privateIP: $privateIP) {
+          network {
+            public_ips
+          }
+        }
+      }
+    `,
+    schema,
+    { privateIP }
+  ).pipe(map(response => response.data));
+
+const Column = componentFromStream<{ privateIP: string }>(
+  (props$): Observable<React.ReactNode> => {
+    return (props$ as Observable<{ privateIP: string }>).pipe(
+      switchMap(({ privateIP }) => getGraphQl(privateIP)),
+      map(({ node }: { node: { network: { public_ips: string[] } } }) => {
+        if (
+          node.network == null ||
+          node.network.public_ips == null ||
+          node.network.public_ips.length === 0
+        ) {
+          return (
+            <TextCell>
+              <Trans>N/A</Trans>
+            </TextCell>
+          );
+        }
+        return <TextCell>{node.network.public_ips.join(", ")}</TextCell>;
+      }),
+      startWith(
+        <TextCell>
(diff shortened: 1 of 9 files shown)


AmrAbdelrazik:
@Poltergeist @juliangieseke and @nLight 
1- Can we display all public IPs for all agents, not just Public agents (if the data exists)
2- If the data does not exist I can we display N/A (not empty field)

nLight:
Looks like we're going to need a click-to-copy-cell component :D

plugins/nodes/src/js/columns/NodesTablePublicIPColumn.tsx
```diff
@@ -0,0 +1,53 @@
...
+      switchMap(({ privateIP }) => getGraphQl(privateIP)),
+      map(({ node }: { node: { network: { public_ips: string[] } } }) => {
+        if (node.network == null) {
+          return null;
+        }
+        return <TextCell>{node.network.public_ips.join(", ")}</TextCell>;
+      }),
+      startWith(
+        <TextCell>
+          <Trans render="span">Loading</Trans>
```
nLight: ❤️

Poltergeist:
The points @AmrAbdelrazik is pointing out we discussed offline yesterday. 
We are exactly doing what point 1 is suggesting. But we are not doing what was suggest in point 2 we are going to add it. to make this screen less confusing.

plugins/nodes/src/js/columns/NodesTablePublicIPColumn.tsx
```diff
@@ -0,0 +1,57 @@
+import * as React from "react";
+import { TextCell } from "@dcos/ui-kit";
+
+import { default as schema } from "../data/NodesResolver";
```
GeorgiSTodorov: The local imports should be below the non-local ones.

plugins/nodes/src/js/columns/NodesTablePublicIPColumn.tsx
```diff
@@ -0,0 +1,57 @@
...
+      startWith(
+        <TextCell>
+          <Trans render="span">Loading</Trans>
+        </TextCell>
+      )
+    );
+  }
+);
+
+export function publicIPCellWidth(): number {
```
GeorgiSTodorov: I don't think we should be passing hard-coded width anymore, the new nodes table can adjust width based on content.

plugins/nodes/src/js/components/NodesTable.tsx
```diff
@@ -182,6 +191,16 @@ export default class NodesTable extends React.Component<
             cellRenderer={healthRenderer}
           />
 
+          <Column
+            header={
+              <HeaderCell>
```
GeorgiSTodorov: Why are we making this non-sortable?

plugins/nodes/src/js/columns/NodesTablePublicIPColumn.tsx
nLight: I wonder what has happened to our linter? We do have this as a rule, don't we?

plugins/nodes/src/js/columns/NodesTablePublicIPColumn.tsx
TattdCodeMonkey: FYI since we're on React@16 now we dont *have* to `render="span"` if we don't want to. Strings are considered valid components in v16 🎉

plugins/nodes/src/js/columns/NodesTablePublicIPColumn.tsx
```diff
@@ -0,0 +1,57 @@
...
+  (props$): Observable<React.ReactNode> => {
+    return (props$ as Observable<{ privateIP: string }>).pipe(
+      switchMap(({ privateIP }) => getGraphQl(privateIP)),
+      map(({ node }: { node: { network: { public_ips: string[] } } }) => {
+        if (
+          node.network == null ||
+          node.network.public_ips == null ||
+          node.network.public_ips.length === 0
+        ) {
+          return <TextCell>N/A</TextCell>;
```
TattdCodeMonkey: Should `N/A` be wrapped in `<Trans>`?

plugins/nodes/src/js/components/NodesTable.tsx
AmrAbdelrazik: Yeah, this should be sortable please

plugins/nodes/src/js/columns/NodesTablePublicIPColumn.tsx
Poltergeist: I had to pass in the width hard coded since the content is changing in the width as soon as the cell gets into the viewport.

plugins/nodes/src/js/components/NodesTable.tsx
Poltergeist: Oh sorry i was under the Impression that we should take the route which would take the least effort. Making this sortable means that we need to take another approach to this problem. Right now we can’t make it sortable since we only now the ip‘s of the current visible nodes. Also the nodes information is not getting merged with the nodes network information. Should we implement this sortable also if it means more effort since we need to merge the nodes information at a higher level?

plugins/nodes/src/js/columns/NodesTablePublicIPColumn.tsx
Poltergeist: You are absolutely right. I didn’t got into the habit yet to use trans around copy will change.

GeorgiSTodorov:
@Poltergeist , file conflicts.

plugins/nodes/src/js/components/NodesTable.tsx
DanielMSchmidt: I disagree here and in retrospect I understand where we had a miscomminucation, sorry for not clearing that up earlier. I believe that the current solution is not good enough, we should not compromise on a product level.

So what we currently do is that every Cell asks for their data, so the overall architecture of this table "knowing about everything in advance" is violated. This leads to the sorting not working properly and also to the Autosizer jamming up.

What we also could do without changing and complicating the data layer implementations and what I thought was the plan to begin with is that we request all data beforehand. As we don't want to scope creep and move the entire table to the data layer I would suggest an approach like this:


```js
const queryDataLayerForNodeAndReturnPublicIp = (hostname) => {};
// Put in the same props as you would in the NodesTable
export default const NodesTableWrapper = componentFromStream(props$ => {
  const data$ = props$.map(props => props.data);
  const dataWithPublicIp$ = props$.map(data => {
     const streamArray = data.map(item => {
       const publicIp$ = of(item).map(item => 
           queryDataLayerForNodeAndReturnPublicIp(item.hostname);
       );

       return item$.combineLatest(publicIp$, (item, publicIp) => 
          ({...item, publicIp })
       );
     });

     return of(42).combineLatest(...streamArray, (_ignore, ...items => items);

  return data$.combineLatest(dataWithPublicIp$, (props, data) => ({...props, data})).map(props => <NodesTable {...props} />)
});
```

I know this example is a bit lengthy, but I think we should take that route or another that leads to us shipping a feature without compromise

DanielMSchmidt: To be clear, I would like to block this for that reason, but I'm OOO for two weeks now

Poltergeist: I did my best to add the example of @DanielMSchmidt  which was unfortunately not as easy as laid out. I choose to select the lowest level possible so that I only have to interact with Node structs and not Nodeslist.

nLight:
@Poltergeist conflicts!

plugins/nodes/src/js/columns/NodesTablePublicIPColumn.tsx
```diff
@@ -0,0 +1,19 @@
+import { TextCell } from "@dcos/ui-kit";
+import * as React from "react";
+import { Tooltip } from "reactjs-components";
+import Node from "src/js/structs/Node";
+
+export default function(item: Node) {
+  const publicIps = item.getPublicIps();
+  if (publicIps.length === 0) {
+    return null;
```
GeorgiSTodorov: Why did we remove N/A?

plugins/nodes/src/js/columns/NodesTablePublicIPColumn.tsx
```diff
@@ -0,0 +1,19 @@
...
+  }
+  if (publicIps.length === 1) {
+    return <TextCell>{publicIps.join()}</TextCell>;
```
GeorgiSTodorov: I think `join`-ing a single item array is a bit confusing, I would prefer `publicIps[0]`.

plugins/nodes/src/js/components/NodesTable.tsx
```diff
@@ -159,124 +166,135 @@ export default class NodesTable extends React.Component<
...
+                  sortHandler={this.handleSortClick.bind(null, "health")}
+                  sortDirection={sortColumn === "health" ? sortDirection : null}
+                />
+              }
+              cellRenderer={healthRenderer}
+            />
+
+            <Column
+              header={
+                <HeaderCell>
```
GeorgiSTodorov: I still think this should be sortable.

plugins/nodes/src/js/columns/NodesTablePublicIPColumn.tsx
Poltergeist: true, will change.

plugins/nodes/src/js/components/NodesTable.tsx
Poltergeist: It is unfortunately not sortable since we only wrap the table and add the information only there but the sorting happens earlier we will address this at a later point.

Poltergeist:
Closed in favor for #3613

PR #3591, opened 2019-02-12
DCOS-48298: findDomNode does not work with this

DanielMSchmidt:
## Testing

- See that the Charts on the dashboard still work while resizing

### DIFF
--- a/src/js/components/charts/Chart.js
+++ b/src/js/components/charts/Chart.js
@@ -1,7 +1,6 @@
 import PropTypes from "prop-types";
 import React from "react";
 import createReactClass from "create-react-class";
-import ReactDOM from "react-dom";
 import { StoreMixin } from "mesosphere-shared-reactjs";
 
 import DOMUtils from "../../utils/DOMUtils";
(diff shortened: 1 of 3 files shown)


src/js/components/charts/Chart.js
```diff
@@ -96,7 +97,11 @@ var Chart = createReactClass({
   render() {
     // at the moment, 'chart' is used to inject the chart color palette.
     // we should reclaim it as the rightful className of <Chart />
-    return <div className="chart-chart">{this.getChildren()}</div>;
+    return (
+      <div ref="chart" className="chart-chart">
```
nLight: string based refs are deprecated. Shall we switch to `React.createRef()`?

DanielMSchmidt: I was under the impression `React.createRef()` only works in a class scenario, which we can't go to due to the mixins, am I mistaken? Could you point me to articles / docs for that?

nLight: then we should use `callback` syntax instead `ref={el => (this.chartRef = el)}`

Poltergeist: Which will fail at the linter since lambdas are not allowed in props we should create a function on the object which we call.

nLight: that could also be an option

DanielMSchmidt: I think this is an exception for the linting rule

DanielMSchmidt:
@nLight ref function works just fine, finally had the time to get to it

PR #3593, opened 2019-02-13
Fix react@16 ref related bugs in reactjs-components

nLight:
## Testing

Smoke testing UI and monitoring console if there're any errors.


GeorgiSTodorov:
We do have some errors in console. I don't think the idea here is to remove all console errors, we should just not get any new ones, right?

DanielMSchmidt:
@GeorgiSTodorov Right, thanks for clairifying. We should maybe somewhen try to remove the amount of errors we have by default 🙈

PR #3598, opened 2019-02-14
DCOS-48262 Manual update/reset UX

TattdCodeMonkey:
Adding the `ui-update` plugin that will be used to display the currently
installed ui version, as well as interaction with the new
dcos-ui-update-service. A user will be able to rollback updates as well
as install new versions of the ui if a newer version exists in cosmos.

Closes DCOS-48262

## Testing

Testing this PR requires an OSS cluster that is running the latest `dcos-ui-update-service`. You can deploy a cluster from `DCOS PR#4345` or if that has merged into `dcos` `master` a cluster from master.

1. Add the following package repository to your cluster:
`https://universe-converter.mesosphere.com/transform?url=https://s3-us-west-2.amazonaws.com/rnorris-test-files/dcos-ui-update-testing.json` with priority 0.
This stub universe points to the `dcos-ui` nightly build with a package version of `2.500.0` to allow testing a "new" package.
2. Navigate to the UI Settings page `/#/settings/ui-settings`
3. Wait for versions to load
4. Confirm you see "Available Version 2.500.0"
5. Click "Start Update"
6. Wait for update to complete
7. Confirm Available update row is not shown when update completes
8 Confirm "Refresh to load (X.X.X)" <- this build is the actual build # not the fake 2.500.0 if the package & build versions match these numbers would match
9. Refresh the Page
10. Confirm new installed version is displayed
11. Confirm "Available Version" is not displayed

After testing an update you can test a rollback
1.  Navigate to the UI Settings page `/#/settings/ui-settings`
2. Wait for versions to load
3. Note the Installed Version #
4. Click Rollback
5. Wait for rollback to complete
6. Confirm "Refresh to load (X.X.X)" is shown when rollback is compelte
7. Refresh the page
8. Confirm Installed Version is different
9. Confirm "Available Version" is displayed again

## Dependencies

✅ [DCOS 4345](https://github.com/dcos/dcos/pull/4345) ~needs to land before this can merged. You can deploy a cluster against that PR to test though.~

## Screenshots

image
image
image
image
image


plugins/ui-update/utils/index.ts
```diff
@@ -0,0 +1,228 @@
...
+  message: string;
+  name: string;
+}
+interface UIUpdateResult {
+  data: { updateDCOSUI: string };
+}
+interface UIRollbackResult {
+  data: { resetDCOSUI: string | null };
+}
+const FALLBACK_UI_METADATA: UIMetadata = {
```
TattdCodeMonkey: It strikes me that this may not be the best name or location for this `const` now. It made sense when I added for the error case in `queryUIServiceForMetadata`, but I ended up re-using it in a few other places after that.

I'll think about this and likely move it, open to suggestions here.

GeorgiSTodorov:
Package-lock conflict as we merged the React 16.8 PR.

GeorgiSTodorov:
I still get '0.0.0' as installed version even after updating and refreshing.
screenshot from 2019-02-21 12-34-48
Also, some tests are failing.
screenshot from 2019-02-21 12-37-25

TattdCodeMonkey:
@GeorgiSTodorov “installed version” is always going to show the client version, which is read from `src/index.html` which doesn’t change when running locally.

🤦‍♂️On the unit test and did a refactor yesterday and didn’t catch that. I’ll fix that ASAP.

Thanks for taking a :eyes:

plugins/index.js
```diff
@@ -7,5 +7,6 @@ module.exports = {
   banner,
   tracking,
   oauth,
-  intercom
+  intercom,
+  "ui-update": require("./ui-update")
```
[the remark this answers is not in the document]
TattdCodeMonkey: I think this might be a vestige of some troubleshooting step. I'll replace with `import`

plugins/ui-update/components/AvailableUpdateRow.tsx
```diff
@@ -0,0 +1,93 @@
...
+    if (this.props.latestVersion !== null) {
+      this.props.onUpdateClick(this.props.latestVersion.version);
+    }
+  }
+
+  renderAction() {
+    const { uiAction } = this.props;
+    let disabled = false;
+    let content = <Trans>Start Update</Trans>;
+    if (uiAction.type === "UIUpdate") {
```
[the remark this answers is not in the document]
pierrebeitz: We could model those solely with types! :)

plugins/ui-update/components/UIDetails.tsx
```diff
@@ -0,0 +1,91 @@
...
+} from "#PLUGINS/ui-update/types/UIAction";
+
+const uiServiceActions$ = new BehaviorSubject<UIAction>(EMPTY_ACTION);
+
+function onClickRollbackUI() {
+  rollbackDCOSUIVersion(uiServiceActions$);
+}
+
+function onClickStartUpdate(version: string) {
+  updateDCOSUIVersion(version, uiServiceActions$);
```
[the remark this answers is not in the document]
TattdCodeMonkey: 🤔 It could, that would basically move the majority of the logic into this file instead of in the util functions. I had put it there to be able to unit test it and keep similar logic together. 

If it returned a stream then the util would basically just be wrapping the logic to make a mutation call. Which is what the query util functions do. I'm fine with that if its better to keep the utils unaware of the action stream.

plugins/ui-update/utils/index.ts
```diff
@@ -0,0 +1,227 @@
...
+      (error: MutationError) => {
+        action$.next({
+          type: actionType,
+          action: UIActions.Error,
+          value: error.message
+        });
+        receivedResult = true;
+      },
+      () => {
+        if (!receivedResult) {
```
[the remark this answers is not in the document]
TattdCodeMonkey: This is an unlikely scenario I was just trying to cover all the bases. But in reality the stream should always either emit a response or error.

plugins/ui-update/utils/index.ts
```diff
@@ -0,0 +1,227 @@
...
+  name: string;
+}
+interface UIUpdateResult {
+  data: { updateDCOSUI: string };
+}
+interface UIRollbackResult {
+  data: { resetDCOSUI: string | null };
+}
+
+function queryCosmosForUIVersions(): Observable<{
```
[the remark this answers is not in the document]
TattdCodeMonkey: Initially i thought I was going to re-use these in another plugin, but since I can use `UIDetails` directly instead it makes less sense to have the logic in the utils file. Since these are more logic then utility.

I'll move these and the mutations back into `UIDetails`.

TattdCodeMonkey:
The integration test failure on this is valid, I broke rollback failure with my refactoring from feedback. I will address this.

plugins/ui-update/components/AvailableUpdateRow.tsx
```diff
@@ -0,0 +1,103 @@
...
+      case UIActions.Started:
+        return <Trans>Updating...</Trans>;
+      case UIActions.Error:
+        return <Trans>Update Failed!</Trans>;
+      default:
+        return <Trans>Start Update</Trans>;
+    }
+  }
+
+  private static actionDisabled(uiAction: UIAction): boolean {
```
nLight: `static` could be a sign that the method doesn't belong to the class.

nLight:
Hangs in this state for me. Is this expected?

screenshot 2019-02-25 at 11 30 48

TattdCodeMonkey:
@nlight does it hang for longer then 45seconds? There is a delay added to the stream currently as a workaround for the race condition.

plugins/ui-update/components/InstalledVersion.tsx
```diff
@@ -0,0 +1,101 @@
...
+}
+
+function renderRollbackContent(props: InstalledVersionProps) {
+  if (!props.uiAction) {
+    return null;
+  }
+  const { uiAction } = props;
+  let disabled = false;
+  let rollbackContent = <Trans>Rollback</Trans>;
+  if (uiAction.type === "UIReset") {
```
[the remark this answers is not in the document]
pierrebeitz: I'm not sure if i understood @nLight's remark correctly, but we could buy ourselves more type safety if we implemented patterns that used [exhaustiveness checking](https://www.typescriptlang.org/docs/handbook/advanced-types.html#exhaustiveness-checking). Therefore we had to model `UIActions` and `UIActionTypes` not as enums but as union types (as i believe exhaustiveness checks don't work with enums. at least it did not ~1.5 years ago).

Subsequently we could use state to conditionally render parts of the UI and only allow certain actions, which get rid of the pattern used in e.g. [renderRollbackAction](https://github.com/dcos/dcos-ui/pull/3598/files#diff-b0b163181531842978f02e234cc5a6c2R24), as we did not need to check whether we are in a state that can handle the current action. But that has nothing to do with this PR in particular and should be discussed in a sync first as to not block here.

I think introducing a exhaustiveness-check-pattern would still be relatively easy to achieve and nice here!

nLight:
@TattdCodeMonkey that could be it. As I left it open and returned it was resolved. I went retesting it and got distracted :)

plugins/ui-update/components/AvailableUpdateRow.tsx
TattdCodeMonkey: I can extract the update & rollback buttons to their own components. 

I prefer to have this logic in the class as you can see how the content/disabled states are determined. Which is why I left them here even as static functions.

plugins/ui-update/components/UIDetails.tsx
```diff
@@ -0,0 +1,217 @@
...
+  name: string;
+}
+interface UIUpdateResult {
+  data: { updateDCOSUI: string };
+}
+interface UIRollbackResult {
+  data: { resetDCOSUI: string | null };
+}
+
+const uiServiceActions$ = new BehaviorSubject<UIAction>(EMPTY_ACTION);
```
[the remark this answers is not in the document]
TattdCodeMonkey: Or an `AsyncSubject` but that would also require some other changes to the `InstalledVersion` component since I'm relying on `EMPTY_ACTION` vs `null` `uiAction` prop for different rendering logic.

TattdCodeMonkey:
@pierrebeitz I see all the tslint errors in my editor and they are all safe to ignore IMO.

plugins/ui-update/components/UIDetails.tsx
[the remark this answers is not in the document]
pierrebeitz: ok! i'll mark that comment as resolved - just wanted to make sure that empty action serves a purpose, as i did not really understand it. :)

pierrebeitz:
@TattdCodeMonkey i'm not talking about `tslint` but actual typescript errors (not warnings). i really think we should not introduce any of those to be able to have a type check across the whole code base soon. there's an effort trying to do just that: https://github.com/dcos/dcos-ui/pull/3623. 

Please think of those errors as invalid typescript.

locale/zh/messages.json
```diff
@@ -329,7 +331,7 @@
   "Error finding node": "查找节点错误",
   "Error finding {item}": "查找 {item} 错误",
   "Exceeded licensed node count on one occasion.": "超过许可节点数 1 次。",
-  "Exceeded licensed node count on {numberBreaches} occasions.": "超过许可节点数 {numberBreaches} 次。",
```
brandonc: Why did this get replaced? The message still exists, but the translation was removed.

plugins/ui-update/components/InstalledVersion.tsx
```diff
@@ -0,0 +1,112 @@
...
+const InstalledVersion = (props: InstalledVersionProps) => {
+  const { uiMetaData } = props;
+  if (!uiMetaData || !uiMetaData.clientBuild) {
+    return null;
+  }
+  const { clientBuild, serverBuild } = uiMetaData;
+
+  let refreshForUpdateContent = null;
+
+  let coercedClientBuild = semver.coerce(clientBuild);
```
brandonc: This and all following variables could be const. You could also extract the button render to a separate render function which could help you get rid of the variable `refreshForUpdateContent`

src/js/data/ui-update/UIMetadata.ts
```diff
@@ -13,3 +13,7 @@ export const UIMetadataSchema = `
     serverBuild: String
   }
 `;
+
+export const DEFAULT_UI_METADATA: UIMetadata = {
```
brandonc: should this file be in the plugin? It's only imported by files in plugins/*

tests/pages/system/settings/UISettings-cy.js
```diff
@@ -0,0 +1,319 @@
+describe("UI Settings", function() {
```
brandonc: ❤️

tests/pages/system/settings/UISettings-cy.js
```diff
@@ -0,0 +1,319 @@
...
+  if (!Cypress.env("FULL_INTEGRATION_TEST")) {
```
brandonc: So this env is present only for ee tests?

plugins/ui-update/components/AvailableUpdateRow.tsx
```diff
@@ -0,0 +1,103 @@
...
+    if (uiAction.type === UIActionType.Update) {
```
brandonc: To me, it feels a little simpler to let the conditions be the return value
```
    return (uiAction.type === UIActionType.Update) || (uiAction.type !== UIActionType.None && uiAction.action !== UIActions.Completed && uiAction.action !== UIActions.Error)
```

plugins/ui-update/components/UIDetails.tsx
```diff
@@ -0,0 +1,217 @@
...
+        <ConfigurationMapSection>
+          <InstalledVersion
+            uiMetaData={uiMetadata}
+            uiAction={uiAction}
+            onRollbackClick={onClickRollbackUI}
+          />
+          <AvailableUpdateRow
+            latestVersion={latestUpdateAvailable}
+            uiAction={uiAction}
+            onUpdateClick={onClickStartUpdate}
```
brandonc: What do  you think about adopting the convention to name callback functions "onXXX" but the handlers that are attached to them as "handleXXX"?

plugins/ui-update/components/UIDetails.tsx
```diff
@@ -0,0 +1,218 @@
...
+        uiServiceActions$.next({
+          type: UIActionType.Reset,
+          action: UIActions.Error,
+          value: error.message
+        });
+      }
+    );
+}
+
+function onClickStartUpdate(version: string, delayMs: number = 45000) {
```
brandonc: I find it a little odd that we export event handlers from the module, but I get why: this is a method you want to test. 

What if you extract this as a query or update command, rename the function to handleClickStartUpdate and just have it invoke the query or command which is tested separately. That could help you break  up the component and simplify the exports

src/js/data/ui-update/UIMetadata.ts
TattdCodeMonkey: I originally had defined it in the plugin, but I felt like it made more sense being here with the interface definition.

tests/pages/system/settings/UISettings-cy.js
TattdCodeMonkey: You can run integration tests against an actual cluster by setting that env variable to true. But the tests I wrote would all fail in that scenario b/c I'm waiting on mocked API calls and updating how the API responds in tests in some cases.

plugins/ui-update/components/UIDetails.tsx
TattdCodeMonkey: I'm only exporting them to be able to test them.

locale/zh/messages.json
TattdCodeMonkey: 🤔 I don't know. I'll see about checking out the messages from master and doing another extract/compile

plugins/ui-update/components/UIDetails.tsx
brandonc: yeah that's my point-- seems like if its complex enough to be tested in isolation it could be its own module

locale/zh/messages.json
TattdCodeMonkey: @brandonc that fixed it.

natmegs:
Couple questions:
- Is it intended that the "Refresh page to load(2.55)" link is still there after you click it and it refreshes the page?
- Is it also intended that the same "Refresh page to load" link stays visible during rollback processing?

TattdCodeMonkey:
@natmegs 
>    * Is it intended that the "Refresh page to load(2.55)" link is still there after you click it and it refreshes the page?
That is due to running locally. That button is displayed when your local version (window.DCOS_UI_VERSION) does not match the version that is available on the server. So it *should* go away when you refresh. But when running from local the build doesn't change when you refresh, unless you manually update the value in `index.html`

>     * Is it also intended that the same "Refresh page to load" link stays visible during rollback processing?

That is a great question and I didn't think about that. I think your right that it should *not* be visible while rolling back or updating. I will change that.

PR #3600, opened 2019-02-15
Testbranch - DO NOT MERGE

pcrews:
## Testing
This is to see how the jenkinsfile changes perform against the main pipeline

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -79,7 +79,7 @@ pipeline {
             withCredentials([
               [
                 $class: "AmazonWebServicesCredentialsBinding",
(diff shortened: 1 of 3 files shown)


Jenkinsfile
```diff
@@ -6,8 +6,7 @@ def master_branches = ["master", ] as String[]
 
 pipeline {
   agent {
-    dockerfile {
-      args  "--shm-size=1g"
+    label 'dcos-ui'
```
juliangieseke: we need to build dockerfile here for dependencies, if you want to change the label - thats defined in jenkins config.

also we want to use dockerfile do be able to eaily change it, docker labels are kinda hardcoded and we had this before… it was a pain…

pierrebeitz:
do we still need this PR?

PR #3602, opened 2019-02-15
[spike] DCOS-48028: [1.12] Metronome JOBs show "Invalid Date".

GeorgiSTodorov:
Check the dates of last successful and last unsuccessful run
and show no tooltip if they are not valid.

Closes https://jira.mesosphere.com/browse/DCOS-48028

## Testing
In `dcos-ui/plugins/jobs/src/js/components/JobsOverviewTable.js`, replace the first line in the `renderLastRunStatusColumn` method with:
```
    let { lastFailureAt, lastSuccessAt, status } = row[prop];
    lastSuccessAt = "foo";
```
Go to jobs tab, create a job (that doesn't fail), run it and wait for it to finish.
From the jobs table, hover over the value for the "Last Run" column for this job.
You should see no tooltip.

## Trade-offs
I am not sure how to reproduce an actual invalid date, so I am mocking one.

## Dependencies
None.

### Before
screenshot from 2019-02-15 13-32-15

### After (no tooltip)
screenshot from 2019-02-15 13-47-18

### DIFF
--- a/plugins/jobs/src/js/components/JobsOverviewTable.js
+++ b/plugins/jobs/src/js/components/JobsOverviewTable.js
@@ -5,6 +5,7 @@ import PropTypes from "prop-types";
 import React from "react";
 import { Table, Tooltip } from "reactjs-components";
 
+import DateUtil from "#SRC/js/utils/DateUtil";
 import Icon from "#SRC/js/components/Icon";
 import ResourceTableUtil from "#SRC/js/utils/ResourceTableUtil";
 import TableUtil from "#SRC/js/utils/TableUtil";
@@ -238,7 +239,10 @@ export default class JobsOverviewTable extends React.Component {
     const nodes = [];
     const statusNode = <span className={statusClasses}>{status}</span>;
 
-    if (lastFailureAt == null && lastSuccessAt == null) {
+    if (
+      !DateUtil.isValidDate(lastFailureAt) ||
+      !DateUtil.isValidDate(lastSuccessAt)
+    ) {
       return statusNode;
     }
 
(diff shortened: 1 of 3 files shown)


plugins/jobs/src/js/components/JobsOverviewTable.js
```diff
@@ -238,7 +239,11 @@ export default class JobsOverviewTable extends React.Component {
     const nodes = [];
     const statusNode = <span className={statusClasses}>{status}</span>;
 
-    if (lastFailureAt == null && lastSuccessAt == null) {
+    if (
+      (lastFailureAt == null && lastSuccessAt == null) ||
+      (lastFailureAt != null && !DateUtil.isValidDate(lastFailureAt)) ||
+      (lastSuccessAt != null && !DateUtil.isValidDate(lastSuccessAt))
```
pierrebeitz: if we'd make `isValidDate` aware of `null`, could we simplify this to the following?

```typescript
if (!DateUtil.isValidDate(lastFailureAt) || !DateUtil.isValidDate(lastSuccessAt)){
  return statusNode;
}
```

GeorgiSTodorov: Good idea, `null` is not a valid date anyway.

plugins/jobs/src/js/components/JobsOverviewTable.js
```diff
@@ -230,15 +231,19 @@ export default class JobsOverviewTable extends React.Component {
   }
 
   renderLastRunStatusColumn(prop, row) {
-    const { lastFailureAt, lastSuccessAt, status } = row[prop];
+    let { lastFailureAt, lastSuccessAt, status } = row[prop];
+    lastSuccessAt = "foo";
```
pierrebeitz: a stray foo!

GeorgiSTodorov: Just noticed, will fix.

PR #3603, opened 2019-02-15
DCOS-48029: [1.13] Metronome JOBs show "Invalid Date"

GeorgiSTodorov:
Check the dates of last successful and last unsuccessful run
and show no console errors if they are not valid.

Closes https://jira.mesosphere.com/browse/DCOS-48029

In 1.13 it doesn't show "Invalid date", because we use format options, but instead it shows console errors.

## Testing
In `dcos-ui/plugins/jobs/src/js/components/JobsOverviewTable.js`, replace the first line in the `renderLastRunStatusColumn` method with:
```
    let { lastFailureAt, lastSuccessAt, status } = row[prop];
    lastSuccessAt = "foo";
```
Go to jobs tab, create a job (that doesn't fail), run it and wait for it to finish.
From the jobs table, open the console.
You should see no date errors.

## Trade-offs
I am not sure how to reproduce an actual invalid date, so I am mocking one.

## Dependencies
None.

### Before
screenshot from 2019-02-15 14-01-07

### After (no console errors)
screenshot from 2019-02-15 14-03-28

### DIFF
--- a/src/js/utils/DateUtil.ts
+++ b/src/js/utils/DateUtil.ts
@@ -145,6 +145,15 @@ const DateUtil = {
 
   getDuration(time: number): string {
     return distanceInWordsStrict(0, time);
+  },
+
+  isValidDate(dateString: string) {
+    if (dateString === null) {
+      return false;
+    }
+
+    const date: Date = new Date(dateString);
+    return date instanceof Date && !isNaN(date.getTime());
   }
 };
 
--- a/plugins/jobs/src/js/components/JobsOverviewTable.js
+++ b/plugins/jobs/src/js/components/JobsOverviewTable.js
@@ -252,7 +252,10 @@ export default class JobsOverviewTable extends React.Component {
     const nodes = [];
     const statusNode = <span className={statusClasses}>{status}</span>;
 
-    if (lastFailureAt == null && lastSuccessAt == null) {
+    if (
+      !DateUtil.isValidDate(lastFailureAt) ||
+      !DateUtil.isValidDate(lastSuccessAt)
+    ) {
       return statusNode;
     }
 
--- a/src/js/utils/__tests__/DateUtil-test.js
+++ b/src/js/utils/__tests__/DateUtil-test.js
@@ -109,4 +109,20 @@ describe("DateUtil", function() {
       expect(DateUtil.strToMs(undefined)).toEqual(null);
     });
   });
+
+  describe("#isValidDate", function() {
+    it("return true for a valid date", function() {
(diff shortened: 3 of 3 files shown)


GeorgiSTodorov:
CI discovered typescript error, I will look into it.

GeorgiSTodorov:
Fixed the typescript error.

src/js/utils/DateUtil.ts
```diff
@@ -135,6 +135,15 @@ const DateUtil = {
     formatKey: moment.DurationInputArg2 = "seconds"
   ): string {
     return moment.duration(time, formatKey).humanize();
+  },
+
+  isValidDate(dateString: string) {
```
DanielMSchmidt: I think this is not really right regarding typescript.
You want to express "I get something thats a string or null", otherwise you would not check for null. It should be `datestring?: string`, imho

GeorgiSTodorov: I don't think `dateString?` is right, I think it should be `dateString: string | null`.

DanielMSchmidt: Yeah, but could it also be undefined? Do we need `dateString?: string | null`

pierrebeitz: Please keep in mind that `datestring?: string` is different from `datestring: string | undefined` in that it allows the call without an argument (`isValidDate()`) which does not make a lot of sense here.

DanielMSchmidt: Yeah right, thanks for adding @pierrebeitz 👍

GeorgiSTodorov: I agree with `Maybe<string>`, but the Maybe PR is not yet merged. So, for now, are we okay with `string | null`?  @pierrebeitz @DanielMSchmidt

pierrebeitz: 👍

nLight:
```
ERROR in /jenkins/workspace/end_dcos-ui-oss-pipeline_PR-3603/src/js/utils/DateUtil.ts

ERROR in /jenkins/workspace/end_dcos-ui-oss-pipeline_PR-3603/src/js/utils/DateUtil.ts(146,43):

TS2345: Argument of type 'Date' is not assignable to parameter of type 'number'.

Child html-webpack-plugin for "../index.html":

```

pierrebeitz:
moment.js has been removed from the main bundle - we should use date-fns here.

GeorgiSTodorov:
@pierrebeitz , good point, something I missed while rebasing.

PR #3604, opened 2019-02-15
Use dcos-ui label for CI

nLight:


Jenkinsfile
```diff
@@ -6,9 +6,7 @@ def master_branches = ["master", ] as String[]
 
 pipeline {
   agent {
-    dockerfile {
-      args  "--shm-size=1g"
-    }
+    label 'dcos-ui'
```
juliangieseke: can you explain why we would want to use the label instead of our dockerfile? 🤔

nLight: to avoid docker in docker and gain some speed

nLight: following recommendations of Patric

PR #3613, opened 2019-02-21
DCOS-44842: add public ip component

Poltergeist:
## Testing

Visit the nodes page this should display the public ip of the public available nodes. 

The polling Stream to the API is shared and should only be shared once. 
Please check the network tab in the dev tools and filter for `/net/v1/nodes` and check that the endpoint is only requested every 2 seconds and not once for each line. best checked in an environment where multiple nodes are available.

## Trade-offs

As a trade-off for the current solution we have that the column is not sortable. Also we are unpacking the Node struct adding the network information and create a new Node struct object. 

## Dependencies

~#3586~

## Screenshots

image


plugins/nodes/src/js/components/NodesTableWrapper.tsx
```diff
@@ -0,0 +1,82 @@
...
+    { privateIP }
+  ).pipe(map(response => response.data));
+};
+// Put in the same props as you would in the NodesTable
+const NodesTableWrapper = componentFromStream<TableWrapperProps>(props$ => {
+  const data$ = (props$ as Observable<TableWrapperProps>).pipe(
+    map((props: TableWrapperProps) => props.data),
+    distinctUntilChanged((x, y) => {
+      if (x.length === y.length) {
+        return JSON.stringify(x) === JSON.stringify(y);
```
jaxesn: This seems pretty strange.  It appears like this will run quite often, even just hovering over the rows, as the table grows this seems like a potential slow down...  Is there way to use some kind of immutable so we could a reference equal?  Would changing the approach as you mentioned in your note avoid this?

GeorgiSTodorov:
It looks like an integration test is failing - "Nodes Page -- Filters nodes table -- shows all nodes -- before each hook".

plugins/nodes/src/js/components/NodesTableWrapper.tsx
Poltergeist: Well the problem here is we are using Node structs on the top level by that making the comparison already difficult.  In the most cases we are avoiding a much more complex computation by unpacking each node struct object adding the network information from the endpoints and then repacking it into node structs. Also I feel adding immutable.js to dcos-ui and changing the nodes table to work with a immutable list feels to me a bit of a scope creep. This is not nice but unfortunately the whole nodes page is not nice. We should plan to do a bigger refactoring of the whole nodes page. However this pattern is not here to stay.

jaxesn: `complex computation by unpacking each node struct object adding the network information from the endpoints and then repacking it into node structs` is that a one time cost?  Can you run a quick test where you load that table up with say 200 nodes and see how it performs, scrolling + hovering and what not?

Poltergeist: I checked against a bigger cluster the performance was not a lot worse then without the change. I however think we should remove it just to make sure also by sharing the connection we are only sending one request. how do you feel about this @juliangieseke who suggested this check in the first place?

Poltergeist:
@GeorgiSTodorov I found the issue the problem was that I imported the `<Trans>` component from `lingui/react` and not from `lingui/macro` which apparently doesn't work. So basically My auto import 
extension screwed me over.

plugins/nodes/src/js/components/NodesTable.tsx
```diff
@@ -159,124 +166,135 @@ export default class NodesTable extends React.Component<
...
-            cellRenderer={gpuRenderer}
-          />
-
-          <Column
-            header={<span title="Spacing" />}
-            cellRenderer={spacingRenderer}
-            width={spacingSizer}
-          />
-        </Table>
+        <NodesTableWrapper data={data.slice()}>
```
brandonc: I'm concerned about the tradeoff you mentioned in the description. This change seems to double the number of async requests to the nodes endpoint, but that seems like a sufficiently high cost to restructure things.

plugins/nodes/src/js/components/NodesTable.tsx
```diff
@@ -29,7 +34,9 @@ import {
   spacingRenderer
 } from "../columns/NodesTableSpacingColumn";
 
-interface NodesTableProps {
+import { default as PublicIPColumn } from "../columns/NodesTablePublicIPColumn";
+import NodesTableWrapper from "./NodesTableWrapper";
+export interface NodesTableProps {
```
brandonc: Why export NodesTableProps?

plugins/nodes/src/js/components/NodesTable.tsx
```diff
@@ -159,124 +166,135 @@ export default class NodesTable extends React.Component<
...
-          />
-
-          <Column
-            header={<span title="Spacing" />}
-            cellRenderer={spacingRenderer}
-            width={spacingSizer}
-          />
-        </Table>
+        <NodesTableWrapper data={data.slice()}>
+          <Table data={data.slice()}>
```
brandonc: You could copy this data once and store it in a local

GeorgiSTodorov:
File conflicts.

plugins/nodes/src/js/components/NodesTable.tsx
Poltergeist: There should be no redundant network requests. The request is shared via rxjs.

plugins/nodes/src/js/components/NodesTable.tsx
Poltergeist: Will remove the export.

plugins/nodes/src/js/components/NodesTable.tsx
Poltergeist: Yes will do

plugins/nodes/src/js/components/NodesTable.tsx
brandonc: I double checked. The polling of the nodes endpoint happen roughly twice that of master branch

Poltergeist: Okay, the problem I see here is that we are not requesting the `/net/v1/nodes` endpoint on the master branch, could it be that you are just filtering for `nodes` and not `/net/v1/nodes` since we also request `/system/health/v1/nodes` in that case the double number of request is a expected outcome.

plugins/nodes/src/js/components/NodesTableWrapper.tsx
```diff
@@ -0,0 +1,76 @@
...
+};
+// Put in the same props as you would in the NodesTable
+const NodesTableWrapper = componentFromStream<TableWrapperProps>(props$ => {
+  const data$ = (props$ as Observable<TableWrapperProps>).pipe(
+    map((props: TableWrapperProps) => props.data)
+  );
+  const dataWithPublicIp$ = data$.pipe(
+    switchMap(data => {
+      const streamArray = data.map(item => {
+        return getGraphQl(item.hostname).pipe(
```
pierrebeitz: it looks like there are too many requests appearing in the network tab. the first requests get canceled immediately though.

plugins/nodes/src/js/data/NodesNetworkClient.ts
```diff
@@ -32,7 +32,7 @@ export function fetchNodesNetwork(): Observable<
 > {
   const [success, error] = partition(
     (response: RequestResponse<NodesNetworkResponse>) => response.code < 300
-  )(request("/net/v1/nodes"));
+  )(request<NodeNetworkResponse[]>("/net/v1/nodes").pipe(share()));
```
pierrebeitz: can we remove the sharing in `NodesResolver` now?

plugins/nodes/src/js/components/NodesTable.tsx
brandonc: No, I'm seeing the normal request to /system/health/v1/nodes and two additional requests that happen nearly simultaneously to /net/v1/nodes

You can see here it's illustrated in the network waterfall and total request count after 80ish seconds

master branch:
screen shot 2019-02-27 at 9 53 41 am

you branch:
screen shot 2019-02-27 at 9 55 42 am

plugins/nodes/src/js/components/NodesTableWrapper.tsx
pierrebeitz: Resolving as Brandon already pointed this out.

src/js/structs/Node.d.ts
```diff
@@ -1,4 +1,4 @@
-import Item from "./Item";
+import Item from './Item';
```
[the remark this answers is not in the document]
Poltergeist: 🦆  & 📔

plugins/nodes/src/js/components/NodesTableWrapper.tsx
```diff
@@ -0,0 +1,76 @@
+import * as React from "react";
+import graphqlObservable from "reactive-graphql";
```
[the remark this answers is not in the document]
Poltergeist: yep

plugins/nodes/src/js/components/NodesTableWrapper.tsx
```diff
@@ -0,0 +1,76 @@
...
+        }
+      }
+    `,
+    schema,
+    { privateIP }
+  ).pipe(map(response => response.data));
+};
+// Put in the same props as you would in the NodesTable
+const NodesTableWrapper = componentFromStream<TableWrapperProps>(props$ => {
+  const data$ = (props$ as Observable<TableWrapperProps>).pipe(
```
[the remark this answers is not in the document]
Poltergeist: We need the cast since props$ is a `Subscribable<TableWrapperProps>` we could omit the generic type on the function call? But I thought it might be a bad idea. If you have a better approach please let me know.

plugins/nodes/src/js/components/NodesTableWrapper.tsx
```diff
@@ -0,0 +1,76 @@
...
+              new Node({
+                ...item.toJSON(),
+                network: publicIp.node.network
+              })
+          )
+        );
+      });
+
+      return combineLatest(...streamArray).pipe(
+        map(([...items]) => {
```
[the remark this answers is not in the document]
Poltergeist: 🤷‍♂️

plugins/nodes/src/js/data/NodesNetworkClient.ts
Poltergeist: If I don't share I get double the shares.

plugins/nodes/src/js/components/NodesTable.tsx
Poltergeist: Okay now I know why and how this happens. So we are requesting only once per update cycle. the unfortunate part is that we force the update 3 times in a cycle. :-( this has to do with how the data flows in the nodes page. The only way I know of to circle around this is to do a distinct until changed and check if the data has changed with `JSON.stringify` which is far to slow unfortunately. If you have another suggestion I am happy to address.

brandonc: I spent some time thinking about this and don't have a great answer. Have you looked at the componentWillReceiveProps method in NodesTableContainer could be short circuited if the mesos stream data is unchanged?

tests/_support/index.js
```diff
@@ -77,6 +77,7 @@ Cypress.Commands.add("configureCluster", function(configuration) {
 
     router
       .route(/service\/marathon\/v2\/apps/, "fx:marathon-1-task/app")
+      .route(/net\/v1\/nodes/, "fx:1-app-for-each-health/nodes")
```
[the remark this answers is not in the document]
Poltergeist: Yes that is true as far as i remember the Mesos Stream does not work correctly on cypress and since Node informations are from there I didn't want to write a test. But I could try to since now the integration test tooling is kinda working.

plugins/nodes/src/js/columns/NodesTablePublicIPColumn.tsx
```diff
@@ -0,0 +1,27 @@
+import { TextCell } from "@dcos/ui-kit";
```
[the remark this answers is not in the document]
Poltergeist: I tryed to add some kind of sorting. Maybe a topic for the next sync.

tests/_support/index.js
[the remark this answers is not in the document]
Poltergeist: Also the nodes we have in our fixtures do have actual hostnames and we can't match them anyways. So the only thing we could check is if the N/A is set.

plugins/nodes/src/js/components/NodesTable.tsx
```diff
@@ -157,9 +164,11 @@ export default class NodesTable extends React.Component<
       return <Loader />;
     }
 
+    const dataCopy = data.slice();
```
GeorgiSTodorov: Why is this needed?

Poltergeist: I'll remove this it was a suggestion from earlier which is not needed anymore.
https://github.com/dcos/dcos-ui/pull/3613#discussion_r259974500

plugins/nodes/src/js/columns/NodesTablePublicIPColumn.tsx
```diff
@@ -0,0 +1,27 @@
+import * as React from "react";
+import { Tooltip } from "reactjs-components";
+import { TextCell } from "@dcos/ui-kit";
+import { Trans } from "@lingui/macro";
+
+import Node from "src/js/structs/Node";
```
DanielMSchmidt:
```suggestion
import Node from "#SRC/js/structs/Node";
```

plugins/nodes/src/js/columns/NodesTablePublicIPColumn.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+const NodesTablePublicIpColumn = (item: Node) => {
```
DanielMSchmidt: Just random nit-picking:

```suggestion
export default (item: Node) => {
```

or 
```suggestion
export default function NodesTablePublicIpColumn(item: Node) {
```

plugins/nodes/src/js/components/NodesTable.tsx
```diff
@@ -1,5 +1,10 @@
 import * as React from "react";
-import { Table, Column, SortableHeaderCell } from "@dcos/ui-kit/dist/packages";
+import {
+  Table,
+  Column,
+  SortableHeaderCell,
+  HeaderCell
+} from "@dcos/ui-kit/dist/packages";
```
DanielMSchmidt: Do we need the `dist/packages` part here?

plugins/nodes/src/js/components/NodesTable.tsx
```diff
@@ -29,6 +34,8 @@ import {
   spacingRenderer
 } from "../columns/NodesTableSpacingColumn";
 
+import { default as PublicIPColumn } from "../columns/NodesTablePublicIPColumn";
```
DanielMSchmidt:
```suggestion
import PublicIPColumn from "../columns/NodesTablePublicIPColumn";
```

plugins/nodes/src/js/pages/nodes/nodes-table/NodesTableContainer.js
```diff
@@ -87,4 +126,18 @@ class NodesTableContainer extends mixin(StoreMixin, QueryParamsMixin) {
...
+  switchMap(() => fetchNodesNetwork()),
+  map(response => response.response),
+  startWith([]),
+  share()
+);
+
+module.exports = componentFromStream(props$ =>
+  combineLatest(props$, nodes$).pipe(
+    map(([props, networks]) => (
+      <NodesTableContainer {...{ ...props, networks }} />
```
DanielMSchmidt: Another nit-pick
```suggestion
      <NodesTableContainer {...props} networks={networks} }} />
```

plugins/nodes/src/js/pages/nodes/nodes-table/NodesTableContainer.js
```diff
@@ -87,4 +126,18 @@ class NodesTableContainer extends mixin(StoreMixin, QueryParamsMixin) {
   }
 }
 
-module.exports = NodesTableContainer;
+const pollingInterval$ = timer(0, Config.getRefreshRate());
+const nodes$ = pollingInterval$.pipe(
+  switchMap(() => fetchNodesNetwork()),
+  map(response => response.response),
+  startWith([]),
+  share()
```
DanielMSchmidt: Is this really shared? I only see one subscribe and no export of this

plugins/nodes/src/js/pages/nodes/nodes-table/NodesTableContainer.js
```diff
@@ -87,4 +126,18 @@ class NodesTableContainer extends mixin(StoreMixin, QueryParamsMixin) {
   }
 }
 
-module.exports = NodesTableContainer;
+const pollingInterval$ = timer(0, Config.getRefreshRate());
+const nodes$ = pollingInterval$.pipe(
```
DanielMSchmidt: Let's use the DL here as discussed offline, so with a simple networks resolver

GeorgiSTodorov:
Failing integration test is just flaky.

plugins/nodes/src/js/pages/nodes/nodes-table/NodesTableContainer.js
```diff
@@ -87,4 +126,18 @@ class NodesTableContainer extends mixin(StoreMixin, QueryParamsMixin) {
   }
 }
 
-module.exports = NodesTableContainer;
+const pollingInterval$ = timer(0, Config.getRefreshRate());
```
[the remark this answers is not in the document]
DanielMSchmidt: You got it dude

GeorgiSTodorov:
@Poltergeist , file conflicts.

plugins/nodes/src/js/columns/NodesTablePublicIPColumn.tsx
Poltergeist: 👍

plugins/nodes/src/js/columns/NodesTablePublicIPColumn.tsx
Poltergeist: We decided that we want to have a single export at the end of the file as a new convention. Which makes these comments outdated.

plugins/nodes/src/js/components/NodesTable.tsx
Poltergeist: If you ask me I would add even more `/table` but I haven't touched the import

plugins/nodes/src/js/pages/nodes/nodes-table/NodesTableContainer.js
Poltergeist: 👍

plugins/nodes/src/js/pages/nodes/nodes-table/NodesTableContainer.js
Poltergeist: Refactored the NodesResolver to be NodesNetworkResolver

plugins/nodes/src/js/pages/nodes/nodes-table/NodesTableContainer.js
```diff
@@ -1,12 +1,18 @@
...
+import { componentFromStream } from "@dcos/data-service";
 
 import CompositeState from "#SRC/js/structs/CompositeState";
 import QueryParamsMixin from "#SRC/js/mixins/QueryParamsMixin";
 import NodesList from "#SRC/js/structs/NodesList";
+import Node from "#SRC/js/structs/Node";
+import Config from "#SRC/js/config/Config";
 
 import NodesTable from "../../../components/NodesTable";
+import { fetchNodesNetwork } from "../../../data/NodesNetworkClient";
```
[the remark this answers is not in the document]
Poltergeist: I always thought we would use relative imports in a plugin or src but use absolute imports for items outside of the package.

plugins/nodes/src/js/pages/nodes/nodes-table/NodesTableContainer.js
```diff
@@ -87,4 +125,24 @@ class NodesTableContainer extends mixin(StoreMixin, QueryParamsMixin) {
   }
 }
 
-module.exports = NodesTableContainer;
+const getGraphQl = () => {
```
DanielMSchmidt: Why do we have this as a function? RxJS observables are lazily invoked, so it could also be a constant, right?

With the current implementation, wouldn't componentFromStream call the inner callback every time it gets new props, which means it would unsubscribe from the old stream and subscribe to the new one returned by `getGraphql()` again?

Poltergeist: 👍

plugins/nodes/src/js/columns/NodesTablePublicIPColumn.tsx
DanielMSchmidt: Ah okay, I might have forgot about this

plugins/nodes/src/js/components/NodesTable.tsx
DanielMSchmidt: Yeah ok, I just did not review the last PR, so 🤷‍♂️

plugins/nodes/src/js/data/NodesNetworkResolver.ts
```diff
@@ -0,0 +1,84 @@
...
+
+        return networks$;
+      }
+    }
+  };
+};
+
+const baseSchema = `
+type Query {
+  networks(privateIPs: [String!]!): [NodeNetwork]
```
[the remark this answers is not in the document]
DanielMSchmidt: @Poltergeist This is still valid

juliangieseke:
CI was green before and system tests are failing because they cant create cluster… -> merging

PR #3614, opened 2019-02-21
DCOS-48944 Use extension kid from npm

TattdCodeMonkey:


src/js/container.js
```diff
@@ -1,4 +1,5 @@
-import { Container } from "inversify";
+import "reflect-metadata"; // Needed for inversify
```
pierrebeitz: Just out of curiosity: why wasn't that needed before?

pierrebeitz:
If i get it correctly no component is using inversify anymore. Can we remove it from `package.json`?

GeorgiSTodorov:
@pierrebeitz , if you think this is a blocking change, add a blocking review, otherwise someone might merge it as it has two approving reviews.

src/js/container.js
nLight: We didn't use `extension-kid` it has been just sitting there. With this PR we do import the library thus engaging `@injectable`-like decorators.

nLight:
@pierrebeitz we don't use `inversify` _directly_. Though it's a peer dep of `@extension-kid/core` https://github.com/dcos-labs/extension-kid/blob/master/package.json#L18

PR #3535, opened 2019-01-17
Revert "fix: update CLI URL in the installation instruction".

ArmandGrillet:
This reverts commit 810b475b3cac7d81659d5892d14549c77b981c68. The original commit was made due to the lack of a DC/OS CLI for DC/OS 1.13, this is now fixed.

## Testing

A cluster using DC/OS 1.13.

## Trade-offs

We will have to update the URL once we start developing DC/OS 1.14.

PR #3615, opened 2019-02-22
DCOS-48976: [1.13] Download cli modal shows wrong version if user has no permission

GeorgiSTodorov:
Change the CLI install modal instuctions to always show latest version.

Closes https://jira.mesosphere.com/browse/DCOS-48976

## Testing
1. Open CLI install modal from the dropdown in the top right corner
2. Verify that the code snippet for OS X and Linux contains "latest" and not "dcos-1.13" (at the end of the second line)

## Trade-offs
I couldn't find an easy way to unit test that the installation instructions contain "latest".

## Dependencies
None.

### Before
screenshot from 2019-02-22 11-55-38

### After
screenshot from 2019-02-22 11-51-12

### DIFF
--- a/src/js/components/modals/CliInstallModal.js
+++ b/src/js/components/modals/CliInstallModal.js
@@ -65,14 +65,9 @@ class CliInstallModal extends React.Component {
     const clusterUrl = `${protocol}://${hostname}${port}`;
     const { selectedOS } = this.state;
 
-    let version = MetadataStore.parsedVersion;
-    // Prepend 'dcos-' to any version other than latest
-    if (version !== "latest") {
-      version = `dcos-${version}`;
-    }
     const downloadUrl = `https://downloads.dcos.io/binaries/cli/${
       osTypes[selectedOS]
-    }/x86-64/${version}/dcos`;
+    }/x86-64/latest/dcos`;
     if (selectedOS === "Windows") {
       return this.getWindowsInstallInstruction(clusterUrl, downloadUrl);
     }


src/js/components/modals/CliInstallModal.js
```diff
@@ -65,14 +65,9 @@ class CliInstallModal extends React.Component {
...
 
-    let version = MetadataStore.parsedVersion;
-    // Prepend 'dcos-' to any version other than latest
-    if (version !== "latest") {
-      version = `dcos-${version}`;
-    }
     const downloadUrl = `https://downloads.dcos.io/binaries/cli/${
       osTypes[selectedOS]
-    }/x86-64/${version}/dcos`;
+    }/x86-64/latest/dcos`;
```
bamarni: @GeorgiSTodorov : I'm just catching-up with this change, can we please hardcode 1.13 here? That was the original intention when we opened https://github.com/dcos/dcos-ui/pull/3535

nLight: @bamarni please help me understand why would we want having `1.13` hardcoded in `master` and when we should actually switch to always show `latest`?

bamarni: At the moment `latest` is the latest stable CLI (which is the one released for DC/OS 1.12), while `dcos-1.13` contains the DC/OS 1.13 CLI (which is currently a dev build, we're planning to push a stable version there at the time DC/OS 1.13 becomes beta).

It is important during the alpha / beta timeframe that customers use the 1.13 CLI, so they can get all the 1.13 specific features. Thus I'd rather have the `dcos-1.13` URL there.

For DC/OS 1.14 we will consider sticking to `latest` instead of `dcos-1.14`, assuming we can figure out how to handle the alpha/beta concern.

GeorgiSTodorov: @bamarni , I used this comment https://jira.mesosphere.com/browse/DCOS-16401?focusedCommentId=220189&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-220189 as guidance while working on this. Is the information in it no longer relevant?

bamarni: Indeed I was myself confused and didn't think about DC/OS beta :D

It is safer to stick to dcos-1.13 for now and we'll figure-out how to handle CLI URLs for beta DC/OS releases in the future, hopefully for 1.14.

nLight: Since we're going to release 1.13 rather soon how critical is this?

bamarni: It is not highly critical but that'd be nice to change it. Is the UI release already frozen for 1.13?

PR #3617, opened 2019-02-22
docs: npm run integration-tests

nLight:
Some documentation on how to run integration tests locally.


GeorgiSTodorov:
What do you think about doing the same for system tests?

nLight:
@GeorgiSTodorov 😅 at some point. How about next week?

PR #3621, opened 2019-02-22
DCOS-49001 extension-kid notification-service

TattdCodeMonkey:
Introduced the @extension-kid/notification-service that will be used to
publish notifications to notification extensions.

## Testing

There are new unit tests for the notification-service. Otherwise this is just introducing the service that can be used by a new extension.

## Trade-offs

The name could be better.
I used `Symbol`s for the Notification type since they are already used heavily with `inversify`, but they could easily be strings, we could also refactor that later if need be.

## Dependencies

N/A

## Screenshots

N/A


packages/@extension-kid/notification-service/index.js
```diff
@@ -0,0 +1,40 @@
+import { ContainerModule } from "inversify";
```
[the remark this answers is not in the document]
pierrebeitz: making this a `.ts`-file would remove the need for almost all type annotations. only `getExtensionModule` would still need its parameter to be typed to `extension: new (args: any[]) => {})` (at least this is what is inferred for inversify's `BindingToSyntax`-type.

packages/@extension-kid/notification-service/__tests__/NotificationService-test.ts
```diff
@@ -0,0 +1,85 @@
...
+const DebugNotification = Symbol("DebugNotification");
+
+const DebugSymbol = Symbol("Debug");
+
+@injectable()
+class DebugNotificationExtension
+  implements NotificationServiceExtensionInterface {
+  id = DebugSymbol;
+
+  supportedNotifications(): symbol[] {
```
pierrebeitz: `symbol[]` is automatically inferred here, so we can omit.

packages/@extension-kid/notification-service/__tests__/NotificationService-test.ts
```diff
@@ -0,0 +1,85 @@
...
+    return [DebugNotification];
+  }
+
+  push(notification: Notification) {
+    /* tslint:disable */
```
pierrebeitz: tslint does not seem to have a problem here, does it?

packages/@extension-kid/notification-service/__tests__/NotificationService-test.ts
```diff
@@ -0,0 +1,85 @@
...
+describe("NotificationService", () => {
+  let container: Container;
+  beforeEach(() => {
+    container = new Container();
+
+    container.load(notificationServiceContainerModuleFactory());
+  });
+
+  it("can get Notification service", () => {
+    const ns: NotificationService = container.get<NotificationService>(
```
pierrebeitz: we only need one of those annotations:

```
const ns: NotificationService = container.get<NotificationService>(
          ^^^^^ EITHER ^^^^^^                 ^^^^^^^ OR ^^^^^^^^
```

the other is inferred automatically

packages/@extension-kid/notification-service/__tests__/NotificationService-test.ts
```diff
@@ -0,0 +1,85 @@
...
+    const logSpy = jest.spyOn(console, "log");
+
+    const ns: NotificationService = container.get<NotificationService>(
+      NotificationServiceType
+    );
+
+    ns.push(new Notification(Symbol("Other_Type"), "This is another test"));
+
+    expect(logSpy).not.toHaveBeenCalled();
+    logSpy.mockRestore();
```
pierrebeitz: I like how digestible the tests are!

packages/@extension-kid/notification-service/index.js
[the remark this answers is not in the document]
nLight: Package candidates in the `packages` directory technically may have `{main: "index.ts"}` as an entry point. Though as we extract them it is required to have `index.js` + `index.d.ts` combination. To avoid this confusion we decided to follow the `js` entry point from the get go.

packages/@extension-kid/notification-service/NotificationService.ts
```diff
@@ -0,0 +1,48 @@
+import { injectable, inject, named } from "inversify";
```
nLight: Would love to see a README in the package with some design decisions that went into the package and some documentation around how to design an extension for the service.

packages/@extension-kid/notification-service/NotificationService.ts
```diff
@@ -0,0 +1,48 @@
...
+    this._extensionProvider = extensionProvider;
+  }
+
+  push(notification: Notification): void {
+    let filteredExtensions: NotificationServiceExtensionInterface[] = [];
+    this._extensionProvider
+      .getAllExtensions()
+      .reduce((extensions, extension) => {
+        if (extension.supportedNotifications().includes(notification.type)) {
+          extensions.push(extension);
```
nLight: curious why you decided to ask an extension instead of fanning out the notifications and let extension decide what to do?

nLight: Could we use `filter` instead of `reduce`? Having `filteredExtensions` looks unnecessary to me.

packages/@extension-kid/notification-service/__tests__/NotificationService-test.ts
nLight: I wonder if we disabled the no-logs rule 🤔

packages/@extension-kid/notification-service/index.js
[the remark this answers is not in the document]
pierrebeitz: > Though as we extract them it is required to have `index.js` + `index.d.ts` combination.

Why exactly? We could tell webpack to grab those `.ts`-files anyway.

If we wanted to do things in a way that every js-user could use our stuff, shouldn't the other files be `.js` as well or is it only about the `main`?

packages/@extension-kid/notification-service/__tests__/NotificationService-test.ts
pierrebeitz: i'm so sorry. some fault in my editor-setup.

packages/@extension-kid/notification-service/NotificationService.ts
```diff
@@ -0,0 +1,48 @@
...
+  constructor(
+    @inject(ExtensionProvider)
+    @named(NotificationServiceExtensionType)
+    extensionProvider: ExtensionProvider<NotificationServiceExtensionInterface>
+  ) {
+    this._extensionProvider = extensionProvider;
+  }
+
+  push(notification: Notification): void {
+    let filteredExtensions: NotificationServiceExtensionInterface[] = [];
```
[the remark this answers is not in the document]
TattdCodeMonkey: your right, but I ended up removing this in favor of `filter` over `reduce` so its not required now.

packages/@extension-kid/notification-service/index.d.ts
```diff
@@ -0,0 +1,18 @@
+import { ContainerModule } from "inversify";
+import {
+  default as NotificationService,
+  NotificationServiceExtensionInterface
+} from "./NotificationService";
+import { Notification } from "./Notification";
+
+export function getExtensionModule<T>(extension: {
+  new (...args: any[]): T;
+}): ContainerModule;
```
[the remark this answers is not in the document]
TattdCodeMonkey: Changed to `throw`

packages/@extension-kid/notification-service/index.js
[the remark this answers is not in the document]
nLight: > Why exactly? We could tell webpack to grab and compile those `.ts`-files anyway.

That's what we do right now, yes, for packages in the `packages` directory. 
Though I'd like us to avoid a special config in webpack for anything that comes from npm. Including the packages we publish and own.

> shouldn't the other files be `.js` as well?

Totally! Those should eventually be compiled. It's just an attempt to balance the effort and confusion. As we extract those packages they basically stop working because the `main` script is `.ts`. While it's natural to setup the missing compilation step for the package, I discovered that it's very easy to forget about the `main` script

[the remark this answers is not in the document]
pierrebeitz: > I'd like us to avoid a special config in webpack for anything that comes from npm
> ...
> I discovered that it's very easy to forget about the `main` script

Thanks, i really appreciate that you took the time to explain!  
Maybe we can have `.ts`-files as source as well as a dist in each package at some point in time, so we don't need to maintain the typings-files. But that's barely something to discuss now.


I'll mark this conversation as resolved.

packages/@extension-kid/notification-service/NotificationService.ts
TattdCodeMonkey: I started a README, I'll expand it more today.

packages/@extension-kid/notification-service/NotificationService.ts
```diff
@@ -0,0 +1,44 @@
...
+
+const NotificationServiceExtensionType = Symbol("NotificationServiceExtension");
+const NotificationServiceType = Symbol("NotificationService");
+interface NotificationServiceExtensionInterface {
+  id: symbol;
+  supportedNotifications(): symbol[];
+  push(notification: Notification): void;
+}
+
+@injectable()
```
natmegs: Thought I was looking at Angular code for a second 🙂

Poltergeist: coffeescript ☕️?

PR #3623, opened 2019-02-25
enable type checks

pierrebeitz:
This PR would reenable our type checks.

* ts-loader is upgraded to make it compatible with webpack4. on master it only works because it is used in `happyPackMode` with `threadLoader`, making communication between ts-loader and webpack use a narrower interface (which swallows our beloved type errors).
* thread-loader is removed, as webpack4 seems to be similar speedy by default.
* `fork-ts-checker-webpack-plugin` is introduced to speed up type checks. It runs in the background and does not block webpack-compilation.
* ~~Lingui is updated by a patch-release to resolve a typing issue.~~ (already in master)

* [This](https://github.com/dcos/dcos-ui/compare/pb/reenable-type-checks?expand=1#diff-096065ef473f5f9d2cecc6f0b655e7b0) is refactored, as type-wise it was not working out.
* Types are added / fixed. ⚠️ 	This PR is not introducing *pretty* types, but just wants to prevent us from introducing more and more errors. Please 🙈 when spotting `any`s here!
* Please test the DX a little (Speed of reloading, convenience, etc...).

* ~~change TS's `module`-setting to commonjs to have an easier time importing from libraries that do not export "real" (in the typescript sense) default modules.~~

## Tradeoffs

1. We activate `skipLibCheck` in `tsconfig.json` because of cypress. This means that we now blindly trust `.d.ts`-files and do not run a type check on them. As we did not run any type checks at all before this is not too bad, but still TS will not find wrong/strange-typings **in our own `.d.ts-files`** now.

2. Currently the fork-checker is configured to work async. DX-wise this means that errors are shown on the console only - which can be good or bad.
    * faster spikes (as you don't need to add types along the way).
    * maybe you're running into a direction that's hard to type afterwards (which may indicate a suboptimal design) without noticing early.


plugins/nodes/src/js/components/NodesGridView.tsx
```diff
@@ -10,6 +10,14 @@ import NodesGridDials from "./NodesGridDials";
 const MAX_SERVICES_TO_SHOW = 32;
 const OTHER_SERVICES_COLOR = 32;
 
+// This interface has just been crafted to reenable type checking.
+// It may very well be that it needs extension.
+interface Service {
```
nLight: Shall we put it into `plugins/services/src/js/structs/Service.d.ts`? That was the way we started introducing types. Maybe there's a better way?

plugins/jobs/src/js/types/JobRunStatusSummary.ts
```diff
@@ -19,77 +19,14 @@ export function JobRunStatusSummaryTypeResolver(
...
-
-  if (summary.lastFailureAt !== null || summary.lastSuccessAt !== null) {
-    if (lastFailureAt > lastSuccessAt) {
-      status = "Failed" as JobRunStatus;
-      time = summary.lastFailureAt;
-    } else {
-      status = "Success" as JobRunStatus;
-      time = summary.lastSuccessAt;
-    }
+  if ([summary.lastFailureAt, summary.lastSuccessAt].every(x => x === null)) {
```
nLight: We could avoid an array and a function call with just `summary.lastFailureAt  === null && summary.lastSuccessAt === null` curious what's the value add?

pierrebeitz: i'm just notoriously bad at parsing long boolean statements in my head and that's one way to directly see "every(thing) is null". 

as that's not a long statement, i'll happily change that to your suggestion!

plugins/nodes/src/js/components/NodesGridView.tsx
pierrebeitz: With my understanding of them it would feel like a misbuse of `.d.ts`-files, as those are a tool to (partially) type libraries that would be hard to translate into pure `.ts`-files. From what i understood is that we want to convert app-code to TS in a boyscout-approach.

To get things done i just put it there, because i did not want to dig into structs just then.

I'd prefer to clean this up in a separate PR. Maybe we just remove that new declaration in favor of an `any[]` for now (although i don't like those).

plugins/jobs/src/js/types/JobRunStatusSummary.ts
```diff
@@ -19,77 +19,14 @@ export function JobRunStatusSummaryTypeResolver(
...
-  return { status, time: time ? DateUtil.strToMs(time) : null };
+  const lastFailureAt = DateUtil.strToMs(summary.lastFailureAt) || 0;
+  const lastSuccessAt = DateUtil.strToMs(summary.lastSuccessAt) || 0;
+  return {
+    status: lastFailureAt > lastSuccessAt ? "Failed" : "Success",
+    time: Math.max(lastFailureAt, lastSuccessAt)
+  };
 }
-
-export const JobRunStatusSummaryFieldResolver = {
```
Poltergeist: Are we sure that we can remove this complete variable?

plugins/services/src/js/components/SDKPlans.tsx
```diff
@@ -22,6 +22,7 @@ import {
 import gql from "graphql-tag";
 import Loader from "#SRC/js/components/Loader";
 import RequestErrorMsg from "#SRC/js/components/RequestErrorMsg";
+// @ts-ignore
```
Poltergeist: Let's add a `// TODO` here, and file a Jira issue for this?

pierrebeitz:
CI breaks as expected when pushing a type error: https://github.com/dcos/dcos-ui/pull/3626

plugins/jobs/src/js/types/JobRunStatusSummary.ts
pierrebeitz: As it can only be imported by name and the name does not appear anywhere else project-wide (or my grep fools me, which hasn't happened in a long time), i'm quite sure.

plugins/services/src/js/components/SDKPlans.tsx
pierrebeitz: will propose to add an ongoing effort to reduce the amounts of `any`s, apparently bad typings and ignore-comments later on.

plugins/jobs/src/js/types/JobRunStatusSummary.ts
```diff
@@ -19,77 +19,14 @@ export function JobRunStatusSummaryTypeResolver(
...
-    }
+  if (summary.lastFailureAt === null && summary.lastSuccessAt === null) {
+    return { status: "N/A", time: null };
   }
 
-  return { status, time: time ? DateUtil.strToMs(time) : null };
+  const lastFailureAt = DateUtil.strToMs(summary.lastFailureAt) || 0;
+  const lastSuccessAt = DateUtil.strToMs(summary.lastSuccessAt) || 0;
+  return {
+    status: lastSuccessAt >= lastFailureAt ? "Success" : "Failed",
```
GeorgiSTodorov: What if they are both `0`?

pierrebeitz: Same as before.

plugins/nodes/src/js/components/NodesGridView.tsx
pierrebeitz: i removed the new interface for now. will type stuff in a spike day effort.

PR #515, opened 2016-06-24
Remove errormessage after json change

Poltergeist:
http://cl.ly/1l1k2b0V3D3M

image

image

Thsi removes the error message when a change in JSON mode is happening.

PR #3625, opened 2019-02-26
feat: update react-transition-group@2.5.0

TattdCodeMonkey:
Updating `react-transition-group` to latest version to match `ui-kit` this is required to use the `Toaster` & `Toast` components. `reactjs-components` has already been updated.

## Testing

Existing functionality should continue to work. Particularly transitions. See sidebar and modals opening/closing etc.

## Dependencies

~[reactjs-components #515](https://github.com/mesosphere/reactjs-components/pull/515) will need to merge and have the version updated in package.json here.~


GeorgiSTodorov:
I just got an e-mail that said `Successfully published reactjs-components@5.0.0`.

TattdCodeMonkey:
I don’t mind at all, thank you!

GeorgiSTodorov:
Retriggering tests to see if it's a legitimate failure.

plugins/services/src/js/components/LogView.js
```diff
@@ -269,17 +269,9 @@ class LogView extends React.Component {
...
-      <CSSTransitionGroup
-        key="log-go-down-button"
-        transitionAppear={true}
-        transitionName="button"
-        transitionAppearTimeout={350}
-        transitionEnterTimeout={350}
-        transitionLeaveTimeout={350}
-        component="div"
-      >
+      <TransitionGroup appear={true}>
```
mperrotti: At some point this may have transitioned in, but there aren't currently any in/out transition styles to I could find for this one 🤷‍♂️ 

I'd recommend we just remove this TransitionGroup

mperrotti: At some point this may have transitioned in, but there aren't currently any in/out transition styles to I could find for this one 🤷‍♂️

I'd recommend we just remove this TransitionGroup

tests/pages/services/ServiceActions-cy.js
```diff
@@ -391,7 +391,8 @@ describe("Service Actions", function() {
       cy.route({
         method: "PUT",
         url: /marathon\/v2\/apps\/\/cassandra-healthy/,
-        response: []
+        response: [],
+        delay: 100
```
mperrotti: Any idea what transition makes these delays necessary?

plugins/services/src/js/components/LogView.js
TattdCodeMonkey: This is for the Go To Bottom button and I think it does transition in if I'm not mistaken.

mperrotti: I checked in this branch and on master, and it just pops in - no transition. If you're seeing a transition, could you post a screen capture?

tests/pages/services/ServiceActions-cy.js
TattdCodeMonkey: It's related to the modal. I added these as a troubleshooting step to try and diagnose the issue. Testing the button is disabled while the API call is made. and all these tests pass locally.

mperrotti: Ah that makes sense. Thanks

plugins/services/src/js/components/LogView.js
TattdCodeMonkey: Maybe I'm mistaken. And I can't find any styles for `button-enter` `button-appear` etc. I'll double check and if it doesn't transition I'll remove this.

TattdCodeMonkey: You are correct there is not transition, removing.

mperrotti: The transition lives on in our imagination <3

GeorgiSTodorov:
I was testing the integration tests and as it was failing locally as well, I decided I might as well fix it.

TattdCodeMonkey:
force pushed after rebase to allow merging.

PR #3626, opened 2019-02-27
check if ci breaks by type error

pierrebeitz:
nothing to see here 🙈


pierrebeitz:
checking if #3623 checks stuff on CI now.

PR #3628, opened 2019-02-27
fix(JSONMultiContainerParser): fully support scheduling.placement

TattdCodeMonkey:
Closes DCOS_OSS-4840

## Testing
See [DCOS_OSS-4840](https://jira.mesosphere.com/browse/DCOS_OSS-4840)

### DIFF
--- a/plugins/services/src/js/reducers/JSONMultiContainerReducers.js
+++ b/plugins/services/src/js/reducers/JSONMultiContainerReducers.js
@@ -29,14 +29,15 @@ module.exports = {
       joinedPath === "scheduling" ? Util.deepCopy(value) : state;
 
     const constraintsState =
-      state != null && state.placement != null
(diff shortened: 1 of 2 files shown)


plugins/services/src/js/reducers/JSONMultiContainerReducers.js
```diff
@@ -29,14 +29,15 @@ module.exports = {
       joinedPath === "scheduling" ? Util.deepCopy(value) : state;
 
     const constraintsState =
-      state != null && state.placement != null
-        ? state.placement.constraints
-        : [];
+      scheduling.placement != null ? scheduling.placement.constraints : [];
```
TattdCodeMonkey: I’m not 100% sure about changing this line. I did it to be consistent with the line below which was required for the fix.

PR #3635, opened 2019-03-02
DCOS-48322: Jobs container form section

natmegs:
This PR adds a container form section to the jobs form. See intended designs [here](https://mesosphere.invisionapp.com/share/Y6Q8EFKBA3G#/screens/346213421).

## Testing

Try testing the form validation and editing via JSON editor. Some validation scenarios:

Container section disabled:
- In General tab, choose Command Only 
- In Container tab, all form fields should be disabled with an inline banner at the top

Sharing common attributes when switching between UCR and docker
- In General tab, choose Container Image
- Enter something in Container Image text input - JSON editor should show a `ucr` object with an `image: { id: "what you entered" }`
- Go to Container tab, switch to Docker - JSON editor should now show a `docker` object with `image: "what you entered"`
- With docker still chosen, check "Force Pull Image On Launch" - JSON editor should add `forcePullImage: true` to `docker` object
- Switch to UCR - JSON editor should show `ucr` object with `image: { ..., forcePull: true }`

Docker parameters
- Choose Docker
- Add parameters with JSON editor open - they should only appear in job JSON if they have a key or value filled out
- Adding parameters should add `parameters: [{ key: "x", value: "y" }]` to `docker` object
- If you submit the form when a docker parameter has either an empty key or empty value, you should get an error

Args
- Choose docker
- Add args with JSON editor open - they should only appear in job JSON if they are not empty
- Adding args should add `args: ["x", "y", ...]` to `run` object
- If you add `args` manually in JSON editor with UCR chosen and then submit, you should get an error
- If you submit the form with both Args and a Command specified, you should get an error

## Trade-offs

Building on patterns established in #3599 

## Dependencies

N/A


plugins/services/src/js/components/SDKPlans.tsx
```diff
@@ -89,7 +89,7 @@ const SDKPlans = componentFromStream(props$ => {
     service: { getName: () => string };
   }>).pipe(
     map((props: { service: { getName: () => string } }) => {
-      const tasks = MesosStateStore.getTasksByService(props.service);
+      const tasks = (MesosStateStore as any).getTasksByService(props.service);
```
natmegs: If I don't add this, `npm build` fails locally

natmegs:
CI check seemingly got skipped when I pushed my last commit. Previously, it was red, and immediately after I pushed it is supposedly good to go. Seems fishy....? @juliangieseke any ideas?

plugins/jobs/src/js/components/form/ContainerFormSection.tsx
```diff
@@ -0,0 +1,186 @@
...
+                  name="grantRuntimePrivileges"
+                  type="checkbox"
+                  value={formData.grantRuntimePrivileges}
+                  disabled={formData.cmdOnly}
+                />
+                <Trans>Grant Runtime Privileges</Trans>
+                <FieldHelp>
+                  <Trans>
+                    By default, containers are "underprivileged" and cannot, for
+                    example, rurn a Docker daemon inside a Docker container.
```
GeorgiSTodorov: I think this should be "run".

plugins/jobs/src/js/components/form/helpers/MetronomeJobValidators.ts
```diff
@@ -262,5 +262,97 @@ export const MetronomeSpecValidators: MetronomeValidators = {
...
+  },
+
+  oneOfUcrOrDocker(formData: JobOutput) {
+    const docker = formData.job.run.docker;
+    const ucr = formData.job.run.ucr;
+    if (docker && ucr) {
+      return [
+        {
+          path: ["job", "run", "docker"],
+          message: i18nMark("Only one of UCR or Docker is allowed.")
```
GeorgiSTodorov: What do you think about "Only UCR or Docker is allowed."?

GeorgiSTodorov:
Looks good, some things I noticed:
- I can enter docker parameters and args consisting of just empty spaces, is this intended?
- I can enter duplicate parameters and args.

plugins/jobs/src/js/components/form/helpers/JobParsers.ts
```diff
@@ -44,16 +80,41 @@ export const jobSpecToFormOutputParser = (jobSpec: JobSpec): FormOutput => {
...
+    jobSpec.job.run.docker.parameters.length > 0
+      ? jobSpec.job.run.docker.parameters
+      : [];
+  const args =
+    Array.isArray(jobSpec.job.run.args) && jobSpec.job.run.args.length > 0
+      ? jobSpec.job.run.args
+      : [];
+  const imageForcePull =
+    container === Container.UCR
+      ? jobSpec.job.run.ucr &&
```
GeorgiSTodorov: We have an util for this - https://github.com/dcos/dcos-ui/blob/master/src/js/utils/Util.js#L54 .

natmegs:
@GeorgiSTodorov I added changes to use `findNestedPropertyInObject` in the `jobSpecToFormOutputParser`, changed the wording in that validation message, and added a type check to the docker parameter reducer to minimize errors if invalid json has been entered. 

> If I change a docker parameter to be a number in the JSON, I get an error, which is good. However, when I delete that parameter and create a new one, the message is still there

I'm not able to reproduce this - which part of the parameter are you changing to a number? I tried changing a parameter key, value and even replacing the parameter object with a number and wasn't able to see this error. 

> The phrase "Wrong type. Expected string, was number." doesn't appear in the message catalog

This would be coming from the JSON editor which is why it's not in the message catalog.

For spaces in params, args - this is the current behaviour in the Services form so I'm leaning towards keeping it. Also I don't know if there are restrictions against duplicate params or args so I would prefer to have the server throw an error if necessary (cc @AmrAbdelrazik)

AmrAbdelrazik:
Re  Params and Args can you check with Vishnu and Ken please

GeorgiSTodorov:
Here is how to reproduce that error:
peek 2019-03-06 10-27

Also in Services > Environment > Labels we can't have a key starting or ending with a whitespace (although environment variables and secrets can).
screenshot from 2019-03-06 10-32-27

natmegs:
@GeorgiSTodorov Thanks for clarifying those repro steps. I see now that the error you are seeing is from the server. So if you retype those values and resubmit it will work. But to avoid the confusion, I've added client-side checks to ensure args and params are strings.

WRT validating starting/ending with spaces - this doesn't seem to affect parameters or args. I created test jobs where params/args started/ended with spaces and there were no server errors.

AmrAbdelrazik:
@natmegs  we should have client side validation for all fields (you should get the info from RAML spec/Json schema and validate with metronome team

juliangieseke:
Jenkins is fixed 🎉

natmegs:
@AmrAbdelrazik We already have client side validation for all the logic constraints and I've pushed up some additional checks to check the type of each field.

GeorgiSTodorov:
I ran the command to update the message catalog and new messages appeared, although I don't see them in this PR.
screenshot from 2019-03-07 11-07-45

Also, a system test is failing because of this error:
screenshot from 2019-03-07 11-18-32

Even though it is docker:
screenshot from 2019-03-07 11-18-39

plugins/jobs/src/js/components/form/helpers/MetronomeJobValidators.ts
```diff
@@ -219,29 +276,165 @@ export const MetronomeSpecValidators: MetronomeValidators = {
...
+    }
+    if (id != undefined && typeof id !== "string") {
+      errors.push({
+        path: ["job", "run", "ucr", "image", "id"],
+        message: stringErrorMessage
+      });
+    }
+    if (
+      kind != undefined &&
+      (kind !== UcrImageKind.Docker || kind !== UcrImageKind.Appc)
```
Poltergeist: This is always going to be true. since kind can't be `Appc` and `Docker` at the same time. Based on the message I guess we want to have `&&` instead of the `||`

natmegs: Ah you are so right! Thank you!

natmegs:
@GeorgiSTodorov Thanks for catching those missing messages - I pushed a change to add the messages and fix the validation issue :)

GeorgiSTodorov:
Some unit tests are failing locally, even though it seems like they have nothing to do with this PR.
screenshot from 2019-03-08 14-06-56

Some messages, for example "Value cannot be empty" are not translated in the form, even though they are in the message catalog (I replaced all empty strings with 123):
screenshot from 2019-03-08 14-10-51
screenshot from 2019-03-08 14-11-02
screenshot from 2019-03-08 14-11-25

plugins/jobs/src/js/components/form/reducers/JsonReducers.ts
```diff
@@ -0,0 +1,74 @@
+import { deepCopy } from "#SRC/js/utils/Util";
```
GeorgiSTodorov: We usually call JSON reducers JSONReducers.
screenshot from 2019-03-08 14-20-51

natmegs:
@GeorgiSTodorov I'm not sure why your unit tests are failing but they are passing on CI and locally for me. I've fixed the error message translation :)

plugins/jobs/src/js/components/form/reducers/JsonReducers.ts
natmegs: Since this is separate from the services reducers I don't think it will cause confusion

GeorgiSTodorov:
Because we merged that icons from ui-kit PR, right now we can only test this after disabling the private plugins. However, when they are disabled, we can't test changing the language for example.

plugins/jobs/src/js/components/form/GeneralFormSection.tsx
```diff
@@ -1,5 +1,5 @@
 import { Trans } from "@lingui/macro";
-import React, { Component } from "react";
+import * as React from "react";
 import { Tooltip } from "reactjs-components";
 
 import Icon from "#SRC/js/components/Icon";
```
mperrotti: When you update with master, you can replace all the `Icon` components used here with `InfoTooltipIcon` ("#SRC/js/components/form/InfoTooltipIcon")

plugins/jobs/src/js/components/form/reducers/ContainerReducers.ts
```diff
@@ -0,0 +1,148 @@
...
+      };
+      stateCopy.job.run.ucr.image = {
+        ...prevUcrImage,
+        ...newUcrImage
+      };
+    }
+    if (stateCopy.job.run.docker) {
+      stateCopy.job.run.docker.image = value;
+    }
+    return stateCopy;
```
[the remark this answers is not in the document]
natmegs: My only reservation here is that there are potentially many object nested properties that would have to be spread separately (labels, restart, etc). Using `deepCopy` we don't have to worry about potentially overriding a nested object property.

plugins/jobs/src/js/components/form/reducers/ContainerReducers.ts
```diff
@@ -0,0 +1,148 @@
...
+
+export const cmdOnlyReducers = {
+  [JobFormActionType.Set]: (value: "true" | "false", state: JobSpec) => {
+    const stateCopy = deepCopy(state);
+    if (value === "true") {
+      stateCopy.cmdOnly = true;
+    }
+    if (value === "false") {
+      stateCopy.cmdOnly = false;
+    }
```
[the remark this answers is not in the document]
natmegs: The problem with throwing an error here is that the user experience will crash instead of give them something useful - for that reason I would rather fail silently. But it really should never be anything other than "true" or "false", because if it's coming through this reducer then it has been edited in a form field, which we have control over.

plugins/jobs/src/js/components/form/reducers/ContainerReducers.ts
```diff
@@ -0,0 +1,148 @@
...
+    if (!Array.isArray(stateCopy.job.run.args) || !stateCopy.job.run.args) {
+      stateCopy.job.run.args = [];
+    }
+    stateCopy.job.run.args.push("");
+    return stateCopy;
+  },
+  [JobFormActionType.RemoveArrayItem]: (value: number, state: JobSpec) => {
+    const stateCopy = deepCopy(state);
+    const args = stateCopy.job.run.args;
+    if (args && Array.isArray(args) && args.length >= value + 1) {
```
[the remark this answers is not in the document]
natmegs: Yes it *should* never happen, but I added this due to the uncertainty of user being able to enter whatever they want in json editor. As I mentioned in an above comment, my gut feeling is to have this (unlikely case) fail silently so the user experience doesn't hang/crash

natmegs:
@mperrotti I replaced the icons, but the disabled ID on edit came up in the original jobs form PR and was determined to be out of scope - see discussion here for more context https://github.com/dcos/dcos-ui/pull/3599#issuecomment-467247473.

@pierrebeitz I'm having a hard time reproducing that issue... See an attempt I made here https://cl.ly/022ac19551da.

pierrebeitz:
@natmegs 	sorry for the incomplete repro. it seems that a container image is needed to break. so 

* specify a container image
* copy the whole config
* make run an array
* paste config
* 💥

GeorgiSTodorov:
@pierrebeitz , I am not sure if I am understanding you correctly. Can you record a gif or a video?

plugins/jobs/src/js/components/form/reducers/ContainerReducers.ts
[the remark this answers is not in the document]
natmegs: Ok I'm convinced :) I've changed this to throw if the value isn't `true` or `false`!

GeorgiSTodorov:
Failing system test in flaky, retriggered.

PR #3637, opened 2019-03-05
DCOS-49106: Nodes table showing infinite loading when filter doesn't find matches

GeorgiSTodorov:
Remove the loader in case no data is found.

Closes https://jira.mesosphere.com/browse/DCOS-49106

## Testing
1. Go to the nodes table in the nodes tab
2. Filter by ip that doesn't exist
3. Ensure that the table doesn't load forever

## Trade-offs
1. We need to decide how to distinguish between no data, because it is loading and not data because no data is found.
2. We need to decide what to show when no data is found.

## Dependencies
None.

### Before
screenshot from 2019-03-05 12-52-58

### After
screenshot from 2019-03-05 12-49-18


mperrotti:
Since we have no way of knowing the difference between loading data, this change removes the loading bubble for all cases.
Maybe we should only hide the loader when there are filters selected? This way we'll at least see a loader when the user first visits the page.

GeorgiSTodorov:
@mperrotti , this is what I did for the services table, but here we don't have a filtered prop or state.

mperrotti:
Ah ok. Then I guess this is the best we can do for Nodes ¯\_(ツ)_/¯

GeorgiSTodorov:
I like @natmegs 's suggestion, but since data expected an array of nodes, I couldn't set it to `null`. This is why I added a `initialLoading` in state. If you think this is excessive, I will remove it.
@natmegs @mperrotti

plugins/nodes/src/js/components/NodesTable.tsx
```diff
@@ -56,7 +57,8 @@ export default class NodesTable extends React.Component<
     this.state = {
       data: [],
```
pierrebeitz: Works great. 

As for the implementation i think that using a separate flag that's strongly coupled to data might be a little brittle. 

I'd prefer to go in the direction of what's suggested in this [article](https://medium.com/javascript-inside/slaying-a-ui-antipattern-in-react-64a3b98242c).

The fully fledged version might be something for later,  could set data to `null` initially and decide on what to render by the difference of `null` and `[]` although that might require a little type-mangling.

What do you think?

GeorgiSTodorov: To me, both doing `Node [] | null` and an additional state property is sub-optimal.

GeorgiSTodorov:
In the end, I decided to go with `null`, following @nLight 's advice. Sorry to everyone who took time reviewing before.

PR #3638, opened 2019-03-05
fix: unset externalplugins config var on ci

juliangieseke:
## Description

CI was failing because it got stale config values from earlier runs…


juliangieseke:
we have to fix root cause instead of trying to mitigate

PR #3640, opened 2019-03-05
DCOS-49578: [1.12] MesosLogContainer / TaskFileViewer never loads logs for failed tasks

GeorgiSTodorov:
Add a check to prevent infinite loading
and a message for missing files.

Closes https://jira.mesosphere.com/browse/DCOS-49578

## Testing
1. Start a service.
2. Open the service and click on an instance.
3. In `plugins/services/src/js/pages/task-details/TaskFileViewer.js`, change line 149 from `const files = this.getLogFiles();` to `const files = []`, to mock missing files.
4. Open the logs tab for that instance.
5. Verify that it doesn't load forever and a message is shown.

## Trade-offs
Something broke while rebasing and it shows that @nLight also commited.

## Dependencies
None.

### Before
screenshot from 2019-03-05 15-09-16

### After
screenshot from 2019-03-05 14-39-36

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -31,6 +31,8 @@ pipeline {
 
     stage("Build") {
       steps {
+        // jenkins seem to have this variable set for no reason, explicitly remiving it…
+        sh "npm config delete externalplugins"
         sh "npm --unsafe-perm install"
         sh "npm run build"
       }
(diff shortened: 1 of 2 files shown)


juliangieseke:
added the same fix we already added to master, should work now. 👍

PR #3641, opened 2019-03-06
DCOS-48333: Jobs form schedule section

natmegs:
This PR adds the schedule tab and form section to the jobs form. See intended designs [here](https://mesosphere.invisionapp.com/share/Y6Q8EFKBA3G#/screens/345675529). 

## ⚠️Dependencies

This PR is based on #3635 and can only be merged after the other one got merged
It makes use of some new Error utils and structural changes.

## Testing

Try testing the form validation and editing via JSON editor. Some specific validation scenarios:

Schedule is blank if form looks blank:
- Go to schedule tab
- Toggle either or both of the checkboxes on and off
- Submit form and ensure there are no schedule errors

Similar to above:
- Go to schedule tab
- enter an ID
- Toggle either or both of the checkboxes on and off
- Delete ID
- Submit form and ensure there are no schedule errors

## Trade-offs

Building on patterns established in #3599 and #3635


GeorgiSTodorov:
Looks good. Some small things:
- Pressing tab from CRON Schedule input does not bring me to Time zone input, I have to press tab twice to go there.

- Schedule ID error message could be improved,
screenshot from 2019-03-06 10-39-55
although it's fine if I enter only empty characters
screenshot from 2019-03-06 10-53-29

When I tried first "1", then "@", I got both messages
screenshot from 2019-03-06 10-54-55

- I think "Schedule ID" and "CRON Schedule" shouldn't be marked as required if the enable checkbox is not selected.

- Clicking the row where the "Enable schedule" checkbox is activates it (same for the concurrent policy checkbox).
peek 2019-03-06 10-45

- Time Zone and Starting Deadline should show the default values. I created a job without setting them and them became "UTC" and "900".
screenshot from 2019-03-06 10-47-46

- I can't edit schedule, because a schedule with this id already exists (is this intended?).
screenshot from 2019-03-06 10-49-47

- "Starting deadline seconds" sounds a little weird to me
screenshot from 2019-03-06 10-51-46

natmegs:
@GeorgiSTodorov Thank you for the thorough testing! I've tried to address your feedback below:
- Tabbing from CRON schedule: focus goes to documentation link. I think this is ok?
- I'm having trouble reproducing the schedule ID errors - here is a screenshot of me entering 1 for schedule ID: https://cl.ly/3230b11085d0. I do not see `1 does not match pattern...`. I have added validation to check that schedule id is a string and maybe this will resolve some of the issues you are seeing.
- Enabled checkbox can be misleading - you can enter a schedule without checking enabled and this is valid. It doesn't change the validation for the rest of the schedule. Id and CRON are only required if any part of a schedule is filled out
- Timezone and deadline: I didn't put them in the job initial JSON because you are not required to send them in the payload and people may not be used to having them specified in a new job JSON. You are right that they have defaults but the users can check these in documentation too.
- Starting deadline: changed to "Starting deadline must be a number". The error for non-integers is valid.
- Editing schedule: what are the repro steps to see that error? I created a job with schedule, then edited it and it succeeded

GeorgiSTodorov:
@natmegs 
> Tabbing from CRON schedule: focus goes to documentation link. I think this is ok?

For me that's counter-intuitive, especially since the link doesn't look focused and is visually on lower level than the cron schedule and time zone inputs, but maybe this is a correct design?
@mperrotti 

> I'm having trouble reproducing the schedule ID errors - here is a screenshot of me entering 1 for schedule ID: https://cl.ly/3230b11085d0. I do not see 1 does not match pattern.... I have added validation to check that schedule id is a string and maybe this will resolve some of the issues you are seeing.

I can no longer reproduce them either. Did you pull @Poltergeist 's commits before you tested? Maybe he fixed it before you commented and after I commented yesterday.

> Clicking rows on checkbox/radio inputs: this is the same behaviour as in the service form

That doesn't make it right. But it's a non-blocking request for me.

> Enabled checkbox can be misleading - you can enter a schedule without checking enabled and this is valid. It doesn't change the validation for the rest of the schedule. Id and CRON are only required if any part of a schedule is filled out

This is somewhat confusing and un-intuitive for me, but if the designers approve, then I am fine with it as well.

> Timezone and deadline: I didn't put them in the job initial JSON because you are not required to send them in the payload and people may not be used to having them specified in a new job JSON. You are right that they have defaults but the users can check these in documentation too.

They can, but will they? But again, a non-blocking request.

> Starting deadline: changed to "Starting deadline must be a number". The error for non-integers is valid.

It's valid, but confusing. If we change it to sting, it says it must be a number, if we change it to a decimal number, it says `Wrong type. Expected integer, was number.`.

> Editing schedule: what are the repro steps to see that error? I created a job with schedule, then edited it and it succeeded

Hm, not sure. Again, maybe @Poltergeist fixed it.
Although, if we try to change the schedule ID, we get an error. Is this intended?

GeorgiSTodorov:
I ran the command to update the message catalog and new messages appear, although I don't see them in this PR.
screenshot from 2019-03-07 11-00-13
screenshot from 2019-03-07 11-00-20
screenshot from 2019-03-07 11-00-29

natmegs:
@GeorgiSTodorov Great catch on the message catalog - I forgot to update it yesterday with the rest of the changes. I have also disabled the schedule ID input if it's being edited.

plugins/jobs/src/js/components/form/helpers/JobFormData.ts
```diff
@@ -46,6 +46,12 @@ export interface FormOutput {
   disk: number;
   dockerParams: DockerParameter[];
   args: string[];
+  scheduleEnabled?: boolean;
```
GeorgiSTodorov: Is this really optional?

plugins/jobs/src/js/components/form/helpers/JobParsers.ts
```diff
@@ -64,6 +69,21 @@ export function jobSpecToOutputParser(jobSpec: JobSpec): JobOutput {
       }
     }
   }
+  if (
```
GeorgiSTodorov: I think we can combine this if and the if below.
```
if ((jobSpecCopy.schedule) {
  if (jobSpecCopy.schedule.startingDeadlineSeconds === "") { ... }
  const filteredSchedule = filterEmptyValues(jobSpecCopy.schedule);
  if (
      !Object.keys(filteredSchedule).length ||
      schedulePropertiesCanBeDiscarded(filteredSchedule)
    ) { ... }
}
```

plugins/jobs/src/js/components/form/helpers/ScheduleUtil.ts
```diff
@@ -0,0 +1,33 @@
...
+ * and it's not clear how to proceed without looking at the JSON editor.
+ *
+ * Instead, when these checkboxes are updated and they are in a toggled off state and are the
+ * only properties in the schedule object, we can remove the schedule object entirely.
+ */
+export const schedulePropertiesCanBeDiscarded = (
+  schedule: Partial<JobSchedule>
+): boolean => {
+  const numProps = Object.keys(schedule).length;
+  if (
```
GeorgiSTodorov: I think this if and the one below can be merged
```
  if (
    (numProps === 1 || numProps === 2) &&
    (schedule.enabled === false &&
      schedule.concurrentPolicy === ConcurrentPolicy.Forbid)
  ) {
    return true;
  }
```

GeorgiSTodorov:
Similar feedback to https://github.com/dcos/dcos-ui/pull/3635#issuecomment-470908252 .
Failing unit tests, messages in the catalog are not actually translated in the form (I replaced all empty string in the catalog with 123).
screenshot from 2019-03-08 14-32-39
screenshot from 2019-03-08 14-30-49

When I am editing a job, the schedule ID field is disabled even if it was empty before. I don't think this is intended.
screenshot from 2019-03-08 14-38-58

plugins/jobs/src/js/components/form/helpers/JobFormData.ts
natmegs: Yes. A schedule can be submitted without the `enabled` property.

plugins/jobs/src/js/components/form/helpers/ScheduleUtil.ts
natmegs: I don't think that would cover every situation - if the schedule only has `{ concurrentPolicy: "FORBID" }` this won't return true.

natmegs:
@GeorgiSTodorov That's what I get for trying to push a quick fix haha. I reverted that disabled id on edit commit. Since it falls into the same category as the regular job id on edit (not ideal but not a regression), it is out of scope for this PR and we will address it separately. I fixed the translation issue but cannot reproduce the unit test failures.

GeorgiSTodorov:
@natmegs , file conflicts.

mperrotti:
Re: link focus
@GeorgiSTodorov - We don't currently have focus styles for links. This is not an issue specific to these changes, but it's something we should fix. I think this has to be fixed in CNVS. I filed an issue: https://jira.mesosphere.com/browse/DCOS-49859

GeorgiSTodorov:
@mperrotti , I mean not just visually showing that a link is focused, but also focusing something that is on a vertically lower level.

natmegs:
@mperrotti Good catch on `concurrencyPolicy` - added a commit to fix that. WRT the schedule ID, that falls into the same category as job id and is out of scope for right now. See this discussion from another PR https://github.com/dcos/dcos-ui/pull/3599#issuecomment-467247473 
@GeorgiSTodorov For me it makes sense for the focus to go to the link since it's part of the form input. Thanks Mike for filing that issue!

natmegs:
@GeorgiSTodorov The 'does not match pattern' error is coming from the server. It seems their guidelines for schedule ID may not be accurate since "1" should be a valid ID. I've pinged metronome team. I will extract any work needed for this to another ticket.

mperrotti:
@GeorgiSTodorov - I think that focus order makes sense. That link appears in hint text that is associated with that input.

plugins/jobs/src/js/components/form/helpers/ScheduleUtil.ts
GeorgiSTodorov: I don't think it will return true with the current code either.

package.json
```diff
@@ -99,6 +99,7 @@
     "@types/jest": "22.2.3",
     "@types/lingui__react": "2.5.0",
     "@types/md5": "2.1.33",
+    "@types/lodash.isequal": "^4.5.4",
```
GeorgiSTodorov: Build is failing because of the `^`.

GeorgiSTodorov: Fixed.

PR #3642, opened 2019-03-06
run config section

pierrebeitz:
## Testing

Create a new Job and fill the inputs in the "Run Configuration"-tab.

Try to insert weird JSON to the form.

Look for cases the API rejects a job that could have been prevented with client-side validation.


locale/zh/messages.json
```diff
@@ -128,6 +134,7 @@
   "Bootstrap Config": "Bootstrap 配置",
   "Bridge": "桥",
   "Built": "已构建",
+  "By default, containers are \"underprivileged\" and cannot, for example, run a Docker daemon inside a Docker container.": "",
```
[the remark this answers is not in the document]
brandonc: I will put this comment on the base PR

GeorgiSTodorov:
Looks good. Some observations:
- If I enter an invalid value, the error doesn't appear under the input like in the other tabs
screenshot from 2019-03-07 11-23-02
- I think "Enter in seconds" should be "Enter a value in seconds" or "Enter a time in seconds".
- Same think @brandonc noticed, I can't submit because of "restartJob" property.
- Message catalog needs to be updated.
- When I enter a name, I get:
screenshot from 2019-03-07 11-27-31

plugins/jobs/src/js/components/JobsForm.tsx
```diff
@@ -85,16 +91,15 @@ class JobModalForm extends React.Component<JobFormProps> {
...
-    const { value, name, type } = event.target;
-    const action = {
-      type:
-        type === "number" ? JobFormActionType.SetNum : JobFormActionType.Set,
-      value,
-      path: name
-    };
-    onChange(action);
+  onInputChange({ target: { value, name, type, dataset } }: any) {
+    const actionType =
```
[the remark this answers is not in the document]
pierrebeitz: I guess those are questions to Natalie as most of this PRs code is her achievement.

As for me: 

* i like long functions and files as long as everything is referentially transparent and/or declarative. So long view functions are fine, but object oriented stuff needs to be short, as my brain does not handle side-effects well. :)
* i try to use `if`s only for early returns. most `if/else`s feel like a violation of the command-query-separation-principle to me and usually they can be avoided by introducing additional guarantees in front or invariants to the containing function.

natmegs:
I think this is looking great so far. A couple high level things I noticed:
- Labels are following a different pattern than the rest of the form elements. The other ones reduce directly to the format expected by the API, whereas the labels are reducing to an array, and then we modify them to an object only for the JobOutput. I think this is okay
- Restart policy design needs a couple modifications now that I compare to the spec. The Yes/No radios actually correspond to policy ("NEVER"/"ON_FAILURE" ) and it doesn't make a lot of sense from a user perspective to disable retry time based on policy (I am not firmly set on this though - willing to discuss)

natmegs:
Looks good - at this point it can only be tested in OSS w/o plugins-private until it is rebased. A few things I noticed in testing:
- There's extra metadata added to the job JSON (`restartJob` and `retryTime`) which is causing the requests to fail. These should probably be removed from the `JobRun` interface as well as from `JsonReducers`
Screen Shot 2019-03-11 at 5 20 31 PM
- I think we should add type validation to our client side validation (MetronomeJobValidators). Right now I could add for ex. a string value for `maxLaunchDelay` and it won't be caught til server
- I think it would also be helpful to add validation to ensure there is a `uri` for every `artifact`?
- If I add `labels` as an array in the JSON editor (incorrect but someone might try it) - they are silently removed when I hit submit. If job succeeds otherwise, user might not realize labels are not there. Could be helpful to throw validation error

GeorgiSTodorov:
@pierrebeitz , file conflicts.

plugins/jobs/src/js/components/form/helpers/JobFormData.ts
```diff
@@ -73,7 +91,10 @@ export interface JobRun {
   user?: string;
   taskKillGracePeriodSeconds?: number;
   restart?: JobRestart;
+  activeDeadlineSeconds?: number;
   volumes?: Array<JobVolume | SecretVolume>;
+  restartJob?: boolean;
+  retryTime?: number;
```
natmegs: I think `restartJob` and `retryTime` should be removed from the interface. They are contained in `restart` and shouldn't be root level properties in this object.

plugins/jobs/src/js/components/form/helpers/MetronomeJobValidators.ts
```diff
@@ -513,19 +393,6 @@ export const MetronomeSpecValidators: MetronomeValidators = {
     return errors;
   },
 
-  argsAreArray(formData: JobOutput) {
```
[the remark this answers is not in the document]
pierrebeitz: Due to a safeguard in the `argsReducer` this can not be set and not an Array, so i was not able to ever see that message in the UI - thus this method does not seem to add any value and has been removed.

pierrebeitz:
@natmegs 	thanks for being patient with me breaking stuff! :)

i changed the validators one more time to take the indexes into account, so the functionality highlighting inputs with an error does not break.
i changed the labels-validation in that the values are only reduced now, when they are an object (formerly the magic of `Object.entries` applied, explaining the weird behaviour you captured in the screenshot).

i would like to keep the misplaced colon as the generation if the headers ("key", "value") is much easier to reason about by the approach taken in this PR. also the other occurrences will most likely be changed to this form in order to fix https://jira.mesosphere.com/browse/DCOS-45215 in a reasonable way.

plugins/jobs/src/js/components/form/helpers/JobFormData.ts
pierrebeitz: have been removed! thanks for the find!

plugins/jobs/src/js/components/form/RunConfigFormSection.tsx
```diff
@@ -0,0 +1,424 @@
...
+    const showErrors = false;
+    const labels = formData.labels || [];
+    const artifacts = formData.artifacts || [];
+
+    try {
+      return (
+        <div className="form-section">
+          {/*
+
+            RUN CONFIGURATION
```
DanielMSchmidt: I like that you introduced these comments for better readability. I sense that you, too, have the feeling that this components render function is just too big. Have you considered breaking it up into multiple smaller ones containing the sections? then we could easily change the order and have named things in there.

plugins/jobs/src/js/components/form/helpers/JobParsers.ts
```diff
@@ -68,7 +68,25 @@ export function jobSpecToOutputParser(jobSpec: JobSpec): JobOutput {
         }
       }
     }
+
+    // RUN CONFIG
+    const { artifacts } = jobSpecCopy.job.run;
+    jobSpecCopy.job.run.artifacts = Array.isArray(artifacts)
+      ? artifacts.filter(_ => _.uri || _.extract || _.executable || _.cache)
```
DanielMSchmidt: Could we change this from `_` to anything else? I don't know if it's something I picked up or a best practice, but I understand `_` as "omited", like in `function(_, secondParamThatIsTheOnlyOneINeed)`

plugins/jobs/src/js/components/form/helpers/MetronomeJobValidators.ts
```diff
@@ -1,64 +1,108 @@
...
+  values: T[]
+) => (errors: FormError[]) =>
+  errors.concat(
+    values.flatMap(
+      (value, index) =>
+        isValid(value) ? [] : [{ message, path: path(index).split(".") }]
+    )
+  );
+
+const isBoolean = validation<boolean | undefined>(
```
DanielMSchmidt: I think all these validators are a great opportunity to leverage [`unknown`](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-0.html#new-unknown-top-type). I mean if I know I put in a boolean already, why check further?

plugins/jobs/src/js/components/form/helpers/MetronomeJobValidators.ts
```diff
@@ -1,64 +1,108 @@
...
+export const MetronomeSpecValidators: MetronomeValidators = {
+  validate(formData: JobOutput): FormError[] {
+    const { run } = formData.job;
+    const parameters = (run.docker && run.docker.parameters) || [];
+
+    // prettier-ignore
+    return pipe(
+      // IS BOOLEAN
+
+      isBoolean(_ => "job.run.docker.forcePullImage", [run.docker && run.docker.forcePullImage]),
```
DanielMSchmidt: That's really nice and declarative

plugins/jobs/src/js/components/form/helpers/MetronomeJobValidators.ts
```diff
@@ -1,64 +1,108 @@
...
+      isString(_ => "job.run.ucr.image.id", [run.ucr && run.ucr.image && run.ucr.image.id]),
+      isString(_ => "job.run.user", [run.user]),
+      isString(i => `job.labels.${i}.key`, Object.keys(formData.job.labels || [])),
+      isString(i => `job.labels.${i}.value`, Object.values(formData.job.labels || [])),
+      isString(i => `job.run.args.${i}`, run.args || []),
+      isString(i => `job.run.artifacts.${i}.uri`, (run.artifacts || []).map(_ => _.uri)),
+      isString(i => `job.run.docker.parameters.${i}.key`, parameters.map(_ => _.key)),
+      isString(i => `job.run.docker.parameters.${i}.value`, parameters.map(_ => _.value))
+
+      // pipe only infers 10 steps, so we need a cast here
```
DanielMSchmidt: Can we use more then one pipe?

plugins/jobs/src/js/components/form/RunConfigFormSection.tsx
pierrebeitz: TL;DR this is intentional and i'd like to keep it that way.

i like long render functions as long as they're dumb. unless we want to reuse some code i don't feel like breaking up stuff here makes the code more readable or maintainable. the only thing i see as a benefit of splitting up would be the ability to jump to a section with our editors' jump-to-definition-functionality faster, once we found the render-function.

plugins/jobs/src/js/components/form/helpers/MetronomeJobValidators.ts
DanielMSchmidt: Oh noeeees, we are still on an old TS version 😢

plugins/jobs/src/js/components/form/helpers/MetronomeJobValidators.ts
DanielMSchmidt: We could, but not casting the type does not give us too much information, so let's forget about it 🤷‍♂️

plugins/jobs/src/js/components/form/helpers/MetronomeJobValidators.ts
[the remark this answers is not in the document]
pierrebeitz: that's the only way i tried to break it (and provoke the validation-message). it neither broke, nor did the message show up. just double checked this by reintroducing this validator.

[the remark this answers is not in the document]
natmegs: Ok cool - looks good as is! Thanks for checking!!

pierrebeitz:
note to self: we could hand the formErrors to the new component to make the inputs glow red when a validation failed.

plugins/jobs/src/js/components/form/RunConfigFormSection.tsx
```diff
@@ -0,0 +1,424 @@
...
+    super(props);
+  }
+
+  render() {
+    const { formData, onAddItem, onRemoveItem } = this.props;
+    const showErrors = false;
+    const labels = formData.labels || [];
+    const artifacts = formData.artifacts || [];
+
+    try {
```
Poltergeist: Why are we using a `try` here? Are we sure about the impacts of missing out on the code optimisation which v8 might not do because of the try catch block or has this been changed in v8?

plugins/jobs/src/js/components/form/RunConfigFormSection.tsx
Poltergeist: Hmm I feel your concerns and I see where you are coming from. Although the render function is huge, I don't have a better Idea. But since this class has only the render function I would feel more comfortable with changing the class to a single function component. Although, this can be addressed at a later point.

pierrebeitz: that's a great idea!

plugins/jobs/src/js/components/form/RunConfigFormSection.tsx
pierrebeitz: because i don't trust the validators and reducers to always bring "good" data here, when someone enters weird stuff via JSON. implementing an approach i'd trust would've been out of scope here.

AFAIK only code within a try-catch is affected. also we have a componentFromStream that will be wrapped in a try-catch due to every observable being wrapper, right? i guess the implications for one render function are not big enough to continue this discussion (unless my understanding is wrong and the whole app is now affected for some reason).

PR #3644, opened 2019-03-06
DCOS-49002 Toast extension

TattdCodeMonkey:
## Testing

create the following file
```typescript
// src/js/ToastTester.ts
import {
  NotificationService,
  NotificationServiceType
} from "@extension-kid/notification-service";
import {
  ToastNotification,
  ToastNotificationOptions
} from "@extension-kid/toast-notifications";
import container from "./container";

const notificationService = container.get<NotificationService>(
  NotificationServiceType
);

// @ts-ignore
window.makeToast = (title: string, options: ToastNotificationOptions = {}) => {
  const notification = new ToastNotification(title, options);

  notificationService.push(notification);
};
```
Add the following two lines to `src/js/index.js`
```javascript
import "./container";
import "./ToastTester";
```
In the browser console you can now call `makeToast`
```
> makeToast("My test Toast Title", { description: "A description" });
```

## Screenshots

image


packages/@extension-kid/toast-notifications/ToastNotification.ts
```diff
@@ -0,0 +1,79 @@
...
+  readonly secondaryActionText: string;
+
+  constructor(
+    title: string,
+    description: string = "",
+    callback: ToastCallback | null = null,
+    appearance: ToastAppearance = ToastAppearance.Default,
+    primaryActionText: string = "",
+    secondaryActionText: string = "",
+    autodismiss: boolean = false
```
pierrebeitz: Our current blueprint-Toast-component does not use it, but we want it to be here in case we need it one day, right?

also there seems to be no test relying on it yet.

packages/@extension-kid/toast-notifications/__tests__/ToastExtension-test.ts
```diff
@@ -0,0 +1,190 @@
...
+  describe("#push", () => {
+    it("updates stream when new notification is pushed", () => {
+      const ext = new ToasterExtension();
+      const notification = new ToastNotification("unit test");
+      const nextCallback = jest.fn();
+      ext.Toast$.subscribe(nextCallback);
+
+      ext.push(notification);
+
+      expect(nextCallback).toHaveBeenCalledTimes(2);
```
pierrebeitz: i had a hard time understanding why this is `2` (as in "i naively did not expect it do be 2 and had to look up the implementation details").

could we `ext.push(notification);` before `ext.Toast$.subscribe(nextCallback);` so we only get one event? maybe this could avoid the need to do `nextCallback.mockReset();` in some of the subsequent tests.

packages/@extension-kid/toast-notifications/__tests__/ToastExtension-test.ts
```diff
@@ -0,0 +1,190 @@
...
+      expect(nextCallback).toHaveBeenCalledTimes(1);
+      expect(nextCallback).toHaveBeenCalledWith([notification02]);
+    });
+
+    it("doesn't emit if notification not found", () => {
+      const ext = new ToasterExtension();
+      const nextCallback = jest.fn();
+      ext.Toast$.subscribe(nextCallback);
+      expect(nextCallback).toHaveBeenCalledWith([]);
+      nextCallback.mockReset();
```
pierrebeitz: > pierre: maybe this could avoid the need to do nextCallback.mockReset(); in some of the subsequent tests.

here i like the `mockReset` though.

packages/@extension-kid/toast-notifications/__tests__/ToastExtension-test.ts
```diff
@@ -0,0 +1,190 @@
...
+
+  describe("#toastPrimaryAction", () => {
+    it("removes notification action executed", () => {
+      const ext = new ToasterExtension();
+      const notification01 = new ToastNotification("unit test one");
+      const nextCallback = jest.fn();
+      ext.Toast$.subscribe(nextCallback);
+      ext.push(notification01);
+      nextCallback.mockReset();
+      expect(ext.Toast$.getValue()).toEqual([notification01]);
```
pierrebeitz: i think expecting this here is redundant (we already tested that above, right?) and may also a little brittle as it relies on impl-details (`getValue`).

packages/@extension-kid/toast-notifications/components/ToastContainer.tsx
```diff
@@ -0,0 +1,72 @@
...
+    return combineLatest<[ToastContainerProps, ToastNotification[]]>([
+      props$,
+      toastExtension.Toast$
+    ]).pipe(
+      map(([props, toasts]) => {
+        return (
+          <div className={props.className}>
+            <Toaster>
+              {toasts.length > 0
+                ? makeToasts(
```
pierrebeitz: i'm not sure just from looking at it, but could we also expose `makeToast` directly, so we could do a `toasts.map(t => makeToast(t, /*...*/))` here that would yield `[]` automatically when no toasts are present instead of having a conditional on `length`? (to me it feels like `makeToasts` is a worse abstraction than `makeToast`)

packages/@extension-kid/toast-notifications/components/ToastContainer.tsx
```diff
@@ -0,0 +1,72 @@
...
+function makeToastContainer(
+  container: Container
+): React.ComponentType<ToastContainerProps> {
+  return componentFromStream<ToastContainerProps>(props$ => {
+    const ns = container.get<NotificationService>(NotificationServiceType);
+    const extension = ns.findExtension(ToastNotification.NotificationType);
+    const toastExtension =
+      extension !== undefined ? (extension as ToastExtension) : undefined;
+
+    if (!toastExtension) {
```
pierrebeitz: if we move that before line 40 we could get rid of the noisy ternary that looks important but actually is not. :)

src/styles/components/toasts/styles.less
```diff
@@ -0,0 +1,14 @@
+& when (@toasts-enabled) {
```
pierrebeitz: wow! that's neat! 😮  👍

packages/@extension-kid/toast-notifications/ToastNotification.ts
```diff
@@ -0,0 +1,79 @@
...
+  ) {
+    super(ToastNotification.NotificationType, `${title}|${description}`);
+    this.title = title;
+    this.description = description || "";
```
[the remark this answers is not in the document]
TattdCodeMonkey: yea, I think this is latent from when I was initially writing this. This will get handled when I refactor to use an option hash.

packages/@extension-kid/toast-notifications/ToastNotification.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  constructor(
+    title: string,
+    description: string = "",
+    callback: ToastCallback | null = null,
+    appearance: ToastAppearance = ToastAppearance.Default,
+    primaryActionText: string = "",
+    secondaryActionText: string = "",
+    autodismiss: boolean = false
+  ) {
```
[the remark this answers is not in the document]
TattdCodeMonkey: Good idea, will refactor.

packages/@extension-kid/toast-notifications/ToastNotification.ts
TattdCodeMonkey: the `Toast` in `ui-kit` supports this and the prop is set when creating the component.

packages/@extension-kid/toast-notifications/__tests__/ToastExtension-test.ts
TattdCodeMonkey: I think I want to refactor the creation of the extension and callback mock into `beforeEach` and then I can separate the two expected calls to two tests.

packages/@extension-kid/toast-notifications/components/ToastContainer.tsx
TattdCodeMonkey: yea `makeToasts` is redundant at this point. I'll refactor it out.

GeorgiSTodorov:
The failing integration test is also failing on master.
I don't understand why we are expecting the modal to exist after we close it - https://github.com/dcos/dcos-ui/blob/master/tests/pages/system/overview/repositories/AddRepositoryFormModal-cy.js#L48 .

TattdCodeMonkey:
Yea those tests are flakey now after #3625 ☹️ I think it’s some kind of timing issue we see on CI because of constrained resources or maybe cypress on Linux.

TattdCodeMonkey:
rebased & force-pushed (again) for latest `master`

PR #3648, opened 2019-03-11
DCOS-45132 Show toasts when UI Updated

TattdCodeMonkey:
Adding stream subscriptions to ui-update plugin that will push Toast
notifications when the UI is updated, an update fails or a rollback fails.
To support this I abstracted the streams from the `UIDetails` component
to the new `streams.ts` file. This allows the streams to be shared
between the component and notifications.

Closes DCOS-45132

## Testing

The UI Updated toast can be triggered by changing the `window.DCOS_UI_VERSION` value in your local `src/index.html` file. Note refreshing locally doesn't change this value so you will get the toast every time the page loads.

## Screenshots

image

image

image


locale/zh/messages.json
```diff
@@ -292,7 +293,7 @@
   "Download": "下载",
   "Download Config": "下载配置",
   "Download Snapshot": "下载快照",
-  "Download and install: <0><1/> Download dcos.exe</0>.": "下载和安装：<0><1/> 下载 dcos.exe</0>.",
+  "Download and install: <0><1/>Download dcos.exe</0>.": "",
```
nLight: Let's please put it back :)

packages/@extension-kid/toast-notifications/ToastNotification.ts
```diff
@@ -21,7 +21,7 @@ type ToastCallback = (
 interface ToastNotificationOptions {
   appearance?: ToastAppearance;
   autodismiss?: boolean;
-  description?: string;
+  description?: string | { id: string; values: Record<string, string> };
```
GeorgiSTodorov: Does it make sense to create an interface for this type of object?

packages/@extension-kid/toast-notifications/ToastNotification.ts
```diff
@@ -30,7 +30,10 @@ interface ToastNotificationOptions {
 class ToastNotification extends Notification {
   static NotificationType = Symbol("Toast");
   readonly title: string;
-  readonly description: string;
+  readonly description:
+    | string
+    | { id: string; values: Record<string, string> }
+    | undefined;
```
GeorgiSTodorov: If it could be undefined, should it be optional?

plugins/ui-update/commands.ts
```diff
@@ -1,22 +1,65 @@
 import { graphqlObservable } from "@dcos/data-service";
 import gql from "graphql-tag";
 import { default as uiServiceSchema } from "#SRC/js/data/ui-update";
-import { delay } from "rxjs/operators";
+import { delay, take } from "rxjs/operators";
```
GeorgiSTodorov: A small thing: we are mixing local and non-local imports.

plugins/ui-update/notifications.ts
```diff
@@ -0,0 +1,189 @@
...
+  ToastAppearance,
+  ToastCallbackType,
+  ToastNotification
+} from "@extension-kid/toast-notifications";
+
+import { getAction$, getUiMetadata$ } from "./streams";
+import { UIActions, UIActionType } from "./types/UIAction";
+import { rollbackUI, updateUI } from "./commands";
+import { getNotificationService } from "./utils";
+import { i18nMark } from "@lingui/react";
```
GeorgiSTodorov: Should be on top.

plugins/ui-update/notifications.ts
```diff
@@ -0,0 +1,189 @@
...
+const activeNotifications: string[] = [];
+
+function clearActiveNotifications() {
+  if (activeNotifications.length) {
+    activeNotifications.splice(0, activeNotifications.length);
+  }
+}
+
+function addActiveNotification(id: string) {
+  activeNotifications.push(id);
```
GeorgiSTodorov: Should we check if a notification with such an id already exists? Or maybe we can make `activeNotificaitons` a set.

plugins/ui-update/notifications.ts
```diff
@@ -0,0 +1,189 @@
...
+}
+
+function setupUIUpdatedNotification() {
+  getUiMetadata$().subscribe(uiMetadata => {
+    const { clientBuild, serverBuild } = uiMetadata;
+    const coercedClientBuild = semver.coerce(clientBuild || "");
+    const coercedServerBuild = semver.coerce(serverBuild || "");
+    if (
+      coercedClientBuild !== null &&
+      coercedServerBuild != null &&
```
GeorgiSTodorov: Is this not `!==` on purpose?

plugins/ui-update/types/UIAction.ts
```diff
@@ -14,11 +14,13 @@ export enum UIActionType {
 export interface UIAction {
   type: UIActionType;
   action: UIActions;
-  value: string;
+  value: { message: string; data?: string };
```
GeorgiSTodorov: Should we have an interface for this kind of object?

GeorgiSTodorov:
For me, nothing happens even after refreshing.
Peek 2019-03-12 13-47

Also, I think the cursor should change to a loading circle when hovering over the disabled button to show that the user should wait.

TattdCodeMonkey:
Update your version in src/index.html to not 0.0.0

I have to write some testing notes this morning.

I’m filtering out the local dev version so we don’t get a toast on every
open

GeorgiSTodorov:
@TattdCodeMonkey , I changed it to 2.3.4 and now when I refresh, it remains 2.3.4.

TattdCodeMonkey:
Yes that is expected, you can’t refresh to change a version when developing
locally since your running through webpack instead of loading the page from
the server.

GeorgiSTodorov:
So is there a way to test it when running locally?

TattdCodeMonkey:
You can test that it displays when the version doesn’t match the server
version and doesn’t display when it does match. And you can test that
clicking reload reloads the page.

But reloading and not getting it again can’t be tested running locally.

packages/@extension-kid/toast-notifications/ToastNotification.ts
TattdCodeMonkey: yes, 👍

plugins/ui-update/notifications.ts
TattdCodeMonkey: nope, fixed

plugins/ui-update/notifications.ts
TattdCodeMonkey: I do check if the array includes the `id` in the `displayNotification` function, which was the only function calling this helper. So I just removed this helper function.

plugins/ui-update/streams.ts
```diff
@@ -0,0 +1,62 @@
...
+} from "rxjs/operators";
+
+import { Package } from "#SRC/js/data/cosmos/Package";
+import { UIMetadata } from "#SRC/js/data/ui-update/UIMetadata";
+
+import { queryCosmosForUIVersions, queryUIServiceForMetadata } from "./queries";
+import { EMPTY_ACTION, UIAction, UIActions } from "./types/UIAction";
+import { versionUpdateAvailable } from "./utils";
+
+const LONG_POLLING_INTERVAL = 300 * 1000;
```
TattdCodeMonkey: Should we pull this from `Config` instead? the current `getRefreshRate` is 2s which is too often but we could create a new value for a longer refresh rate.

packages/@extension-kid/toast-notifications/ToastNotification.ts
```diff
@@ -18,30 +18,40 @@ type ToastCallback = (
...
-  constructor(title: string, options: ToastNotificationOptions = {}) {
-    const description = options.description || "";
-    super(ToastNotification.NotificationType, `${title}|${description}`);
+  constructor(
+    title: string | ToastTranslatableText,
+    options: ToastNotificationOptions = {}
+  ) {
+    super(
+      ToastNotification.NotificationType,
+      `${title}|${options.description || ""}`
```
natmegs: If `title` is an object (`ToastTranslatableText`) I think this will be kinda wonky?

packages/@extension-kid/toast-notifications/components/Toast.tsx
```diff
@@ -1,10 +1,20 @@
...
+import { ToastNotification, ToastTranslatableText } from "../ToastNotification";
 import ToastExtension from "../ToastExtension";
 
+function makeTextContent(
+  value: string | ToastTranslatableText
+): React.ReactElement<HTMLElement> {
+  if (typeof value === "string") {
+    return <Trans id={value} />;
+  }
+  return <Trans id={value.id} values={value.values} />;
```
natmegs: I'm not 100% certain here, but it seems based on the docs that `values` is only a prop on `Trans` from `@lingui/react`? Would it be possible to restrict `value` here to `string | ReactNode` so that it can be either "regular title" or `<Trans>Title formatted in whatever way the user wants so they don't have to know how to format 'values' in this context</Trans>`?

packages/@extension-kid/toast-notifications/ToastNotification.ts
TattdCodeMonkey: Good catch, I just refactor that. I will fix this.

packages/@extension-kid/toast-notifications/components/Toast.tsx
TattdCodeMonkey: Do you mean allowing the value on the ToastNotification class be a ReactNode so we can set it to Trans?

I was trying to avoid that but I’m not completely opposed to it.

natmegs: Yes I guess it would be changing the `title` on ToastNotification as well to `string | ReactNode`. IMO using the `<Trans>` macro is a more familiar pattern than using the bare `<Trans>` @lingui/React component with `values` prop, but it's not a blocker for me

plugins/ui-update/streams.ts
GeorgiSTodorov: I am in favour of adding this value in `Config` and taking it from there.

plugins/ui-update/commands.ts
```diff
@@ -1,22 +1,67 @@
...
         resetDCOSUI
       }
     `,
     uiServiceSchema,
     {}
-  ).pipe(delay(delayMs));
+  )
+    .pipe(
+      take(1),
+      delay(delayMs)
```
DanielMSchmidt: I know, wrong PR, but why do we delay this?

packages/@extension-kid/toast-notifications/ToastNotification.ts
GeorgiSTodorov: Is the PR still in progress then?

plugins/ui-update/components/UIDetails.tsx
```diff
@@ -1,116 +1,39 @@
...
-      },
-      error => {
-        uiServiceActions$.next({
-          type: UIActionType.Reset,
-          action: UIActions.Error,
-          value: error.message
-        });
-      }
-    );
+  rollbackUI();
```
DanielMSchmidt: Can't we directly pass rollbackUI whereever we are passing handleRollbackUI (same for handleUpdateClick)

plugins/ui-update/notifications.ts
```diff
@@ -0,0 +1,185 @@
...
+function setupUIUpdatedNotification() {
+  getUiMetadata$().subscribe(uiMetadata => {
+    const { clientBuild, serverBuild } = uiMetadata;
+    const coercedClientBuild = semver.coerce(clientBuild || "");
+    const coercedServerBuild = semver.coerce(serverBuild || "");
+    if (
+      coercedClientBuild !== null &&
+      coercedServerBuild !== null &&
+      coercedClientBuild.raw !== coercedServerBuild.raw &&
+      coercedClientBuild.raw !== "0.0.0"
```
DanielMSchmidt: Can we move this magic string into a constant?

plugins/ui-update/streams.ts
DanielMSchmidt: +1

packages/@extension-kid/toast-notifications/ToastNotification.ts
TattdCodeMonkey: no, that was just one of the last things I completed and I missed the implication on this call.

plugins/ui-update/commands.ts
TattdCodeMonkey: It's a workaround for the multi-master race condition. We will remove it once the race condition is resolved.

plugins/ui-update/components/UIDetails.tsx
TattdCodeMonkey: no, the click handler would send eventArgs as the first parameter to `rollbackUI`. Which would take them as the `delayMs`.

Once race-condition is fixed, yes we could easily refactor this to be passed directly.

packages/@extension-kid/toast-notifications/components/Toast.tsx
TattdCodeMonkey: I'll try this out today and see what the change would look like.

My two concerns are 1) any file that needs to create and push a notification would then have to import React and be a `tsx` file. Which isn't a deal breaker but its annoying. 2) I'm currently using the content of the title and description to build a hash that is used as the unique id for the toast. This is so that we can ensure previously dismissed toasts aren't re-pushed.

GeorgiSTodorov:
I don't know if this is intended or not, but without the external plugins, we get this for UI settings:
Screenshot from 2019-03-13 14-03-21

TattdCodeMonkey:
@GeorgiSTodorov yes and no. Without external plugins then the config should load the `ui-update` plugin.

GeorgiSTodorov:
Not related to this PR, but I think it would be better if that loading indicator was where "Refresh page" is, because now the User Prefrences move when it finishes loading.
Screenshot from 2019-03-13 14-14-58
@mperrotti , what do you think?

plugins/ui-update/notifications.ts
```diff
@@ -0,0 +1,186 @@
...
+function removeActiveNotification(id: string) {
+  const notificationIndex = activeNotifications.indexOf(id);
+  if (notificationIndex !== -1) {
+    activeNotifications.splice(notificationIndex, 1);
+  }
+}
+
+function displayNotification(notification: ToastNotification) {
+  if (!activeNotifications.includes(notification.id)) {
+    const ns = getNotificationService();
```
nLight: What if we don't have the service on the container? :)

plugins/ui-update/utils/index.ts
```diff
@@ -49,4 +54,8 @@ function versionUpdateAvailable(
   return null;
 }
 
-export { versionUpdateAvailable };
+function getNotificationService(): NotificationService {
+  return container.get<NotificationService>(NotificationServiceType);
```
nLight: I'm not sure that's a good way of using the injection. 
Why don't we create `Notificatior` class and have the service injected into it and then in hooks `plugins/ui-update/hooks.js` we actually use the container `container.get(Notificator); Notificator.setup()`. 
All names are fictional :)

nLight: Ideally we should _not_ call container directly. But if we do we should try doing this at the top most level and let injection do everything else.

TattdCodeMonkey: I can move the get NS service to `hooks.js` and pass the Notification Service in as a parameter?

We'd have to talk about the `Notificatior` stuff I'm not quite grokking it from the comment.

TattdCodeMonkey: @nLight updated to get NotificationService from container in `hooks.js` and pass it as a parameter to `Notifications`. We can refactor to better use injection as an improvement. If thats ok with you.

plugins/ui-update/notifications.ts
TattdCodeMonkey: This is no longer an issue with the latest refactor.

plugins/ui-update/hooks.js
```diff
@@ -1,12 +1,23 @@
...
+import * as Notifications from "./notifications";
 
 module.exports = {
   initialize() {
     MountService.MountService.registerComponent(
       UIDetails,
       "UISettings:UIDetails:Content"
     );
+
+    const notificationService = container.get(NotificationServiceType);
```
nLight: That was almost basically my suggestion :)

```js
@injectable()
class Notifications {
  constructor(@inject(NotificationServiceType) notificationService: NotificationServiceType) {
    this.notificationService = notificationService;
  }
  setup () {
    // use this.notificationService
  }
}
```

`hooks.js`
```js
const notifications = contanier.get(NotificationsType);
notifications.setup(); // Notification Service gets injected
```

plugins/ui-update/notifications.ts
Poltergeist: Nit-pick: can we not abbreviate the `NotificationService` with `ns`.

Also we could also just do `getNotificationService().push(notification)` instead right?

plugins/ui-update/__tests__/notifications-test.ts
```diff
@@ -0,0 +1,243 @@
...
+        mockMetadata$ = m.cold("--j", {
+          j: {
+            clientBuild: "master+v1.0.1",
+            packageVersion: "1.2.0",
+            packageVersionIsDefault: false,
+            serverBuild: "master+v1.2.0"
+          }
+        });
+        //@ts-ignore
+        Notifications.setupUIUpdatedNotification(mockNS);
```
nLight: then here you can create an empty container and bind the mock to it instead of the Notification service

plugins/ui-update/components/UIDetails.tsx
DanielMSchmidt: ah okay :)

mperrotti:
I agree with @GeorgiSTodorov about where the loader appears

This should do it:
```
<ConfigurationMapSection>
  <div className="flex">
    <div className="flex-item flex-item-grow-1">
      <InstalledVersion uiMetaData={DEFAULT_UI_METADATA} />
    </div>
    <ConfigurationMapValue>
      <div className="flex flex-justify-items-end">
        <div className="flex-item">
          <Trans>Checking for new version</Trans>
        </div>
        <div className="flex-item">
          <Loading />
        </div>
      </div>
    </ConfigurationMapValue>
  </div>
</ConfigurationMapSection>
```

GeorgiSTodorov:
In order to try and merge this today, I pushed a commit fixing the message catalog file conflicts (and moving the loader following @mperrotti 's suggestion).

GeorgiSTodorov:
Hm, the chinese catalog broke for some reason.

src/js/config/Config.ts
```diff
@@ -63,6 +65,7 @@ let Config: IConfiguration = {
   rootUrl: "",
   slackChannel: "https://dcos-community.slack.com/messages/general/",
   stateRefresh: 2000,
+  stateLongPoll: 300000,
```
GeorgiSTodorov: Cool.

plugins/ui-update/notifications.ts
```diff
@@ -0,0 +1,192 @@
...
+  getAction$()
+    .pipe(
+      filter(
+        action =>
+          action.action === UIActions.Error &&
+          action.type === UIActionType.Reset
+      )
+    )
+    .subscribe(() => {
+      const notification = new ToastNotification(
```
nLight: we could map the stream to the stream of `new ToastNotification` objects then filter the stream by `notification.id`  and that could go directly to the NotificationService, so that we don't need `displayNotification` with extra filtering logic :)

nLight: I'm coming from seeing streams as flow of data through operators ideally with no extra steps/boilerplate on the end. It's like operating on an array to me `[...actions].filter().map().filter().reduce()`

plugins/ui-update/notifications.ts
```diff
@@ -0,0 +1,192 @@
...
+) {
+  removeActiveNotification(toastNotification.id);
+}
+
+function setupUIUpdatedNotification(notificationService: NotificationService) {
+  getUiMetadata$().subscribe(uiMetadata => {
+    const { clientBuild, serverBuild } = uiMetadata;
+    const coercedClientBuild = semver.coerce(clientBuild || "");
+    const coercedServerBuild = semver.coerce(serverBuild || "");
+    if (
```
nLight: here we have the logic under a conditional with no alternative `else` branch. It looks more like `filter` operator could be leveraged

TattdCodeMonkey:
@nlight that makes sense. I definitely wasn’t thinking in that way but I
see it and like it better. Changes incoming.

PR #3649, opened 2019-03-12
[DCOS-49636] Enable UI prefix label for SDK packages

juliangieseke:
## Testing

* create marathon app having ui labels -> link should be displayed
* install non-sdk package with ui -> link should be displayed
* install sdk package without label -> NO Link should be displayed
* install sdk package with label -> Link from label should be displayed

## Trade-offs

This PR consists of three commits, which I decided to put into one PR:
1. refactoring service util in general to remove duplicate from `src/js/utils` and resolve circular dependency in `plugins/src/js/utils`
2. refactoring structs and adding functionality to enable new label usage (this is one commit b/c I wasn really able to split it up without creating broken code)
3. small ux fix to always display webui icon

## Dependencies

None


plugins/services/src/js/containers/services/ServicesTable.js
```diff
@@ -97,7 +97,7 @@ class ServicesTable extends React.Component {
     if (
       actionItem.id !== EDIT &&
       actionItem.id !== DELETE &&
-      (containsSDKService || isSDKService(service)) &&
+      containsSDKService &&
```
juliangieseke: Assumption: this check is moved into `Framework.js`, because SDK packages _always_ are `Framework`.

plugins/services/src/js/containers/services/ServicesTable.js
```diff
@@ -220,7 +220,6 @@ class ServicesTable extends React.Component {
   hasWebUI(service) {
     return (
       service instanceof Service &&
-      !isSDKService(service) &&
```
juliangieseke: Assumption: this check is moved into `Framework.js`, because SDK packages _always_ are `Framework`.

plugins/services/src/js/structs/Framework.js
```diff
@@ -1,5 +1,5 @@
 import { cleanServiceJSON } from "#SRC/js/utils/CleanJSONUtil";
-import { isSDKService } from "#SRC/js/utils/ServiceUtil";
+import { isSDKService } from "#PLUGINS/services/src/js/utils/ServiceUtil";
```
juliangieseke: There are two `ServiceUtil`, the one in plugins includes `Service` which creates a circular dependency…

plugins/services/src/js/structs/ServiceTree.js
```diff
@@ -48,7 +50,19 @@ module.exports = class ServiceTree extends Tree {
...
+
+      if (ServiceValidatorUtil.isFrameworkResponse(item)) {
+        return new Framework(item);
+      }
+
+      if (ServiceValidatorUtil.isApplicationResponse(item)) {
+        return new Application(item);
+      }
+
+      throw Error("Unknown service response: " + JSON.stringify(item));
```
DanielMSchmidt: Is this error caught anywhere?

plugins/services/src/js/utils/ServiceUtil.js
```diff
@@ -114,6 +114,45 @@ const ServiceUtil = {
       service.getLabels &&
       service.getLabels().DCOS_COMMONS_API_VERSION != null
     );
+  },
+
+  /**
+   * this function extracts webURL logic from all structs to keep it in one place
```
DanielMSchmidt: This comment refers to the refactoring (where the code was before), it would be cooler if you could describe what it does and why

plugins/services/src/js/utils/ServiceUtil.js
```diff
@@ -135,27 +108,51 @@ const ServiceUtil = {
...
+    const {
+      DCOS_SERVICE_NAME,
+      DCOS_SERVICE_PORT_INDEX,
+      DCOS_SERVICE_SCHEME,
+      DCOS_SERVICE_WEB_PATH,
+      DCOS_COMMONS_API_VERSION
+    } = labels;
+    const serviceName = encodeURIComponent(DCOS_SERVICE_NAME);
+
+    // service name label not set or empty
```
pierrebeitz: as i understood you like stylistic discussions, i'd vote for having a couple of intention-revealing booleans to replace the comments and use them for the early return. pseudo-code:


```typescript
    // ...
    const serviceName = encodeURIComponent(DCOS_SERVICE_NAME);
    const hasNewWebUiLabel = serviceName && DCOS_SERVICE_WEB_PATH;
    const hasOldWebUiLabel = serviceName && DCOS_SERVICE_PORT_INDEX && DCOS_SERVICE_SCHEME && !DCOS_COMMONS_API_VERSION;

    if (hasNewWebUiLabel){ // i'd prefer a nested ternary, but eslint does not like that.
      return `${rootUrl}/service/${serviceName}${DCOS_SERVICE_WEB_PATH}` // <-- happy path or new/best/most common way of doing things at the point of most attention.
    } else if (hasOldWebUiLabel) {
      return `${rootUrl}/service/${serviceName}/`;
    } else {
      return "";
    } 
```

or something along those lines...

plugins/services/src/js/structs/__tests__/Framework-test.js
```diff
@@ -283,4 +283,20 @@ describe("Framework", function() {
...
+  describe("getWebURL", () => {
+    // this test mostly makes sure that util isnt removed from `Service` struct - see ServiceUtil tests for more details
+    it("returns url if service is SDK Service providing DCOS_SERVICE_WEB_PATH label", function() {
+      const service = new Framework({
+        labels: {
+          DCOS_SERVICE_NAME: "foo",
+          DCOS_SERVICE_PORT_INDEX: "80",
+          DCOS_SERVICE_SCHEME: "https",
+          DCOS_COMMONS_API_VERSION: "notnull",
+          DCOS_SERVICE_WEB_PATH: "/bar"
```
pierrebeitz: are we sure that there's always a slash in front? can i look that up somewhere?

plugins/services/src/js/structs/__tests__/Application-test.js
```diff
@@ -539,20 +539,32 @@ describe("Application", function() {
...
           DCOS_SERVICE_NAME: "",
           DCOS_SERVICE_PORT_INDEX: "80",
           DCOS_SERVICE_SCHEME: "https"
         }
       });
-      expect(service.getWebURL()).toEqual(null);
+      expect(service.getWebURL()).toEqual("");
+    });
+
+    it("returns empty string if service is SDK Service", function() {
```
pierrebeitz: description and test don't match here :)

tests/pages/services/ServiceTable-cy.js
```diff
@@ -268,6 +268,14 @@ describe("Service Table", function() {
         .contains("stop")
         .should("not.exist");
     });
+
+    it("DCOS_SERVICE_WEB_PATH adds Open Service item", function() {
```
pierrebeitz: i did not grok that on first reading. something like `has an 'Open Service'-DropdownItem when DCOS_SERVICE_WEB_PATH-label is set` would have helped me.

plugins/services/src/js/structs/ServiceTree.js
```diff
@@ -201,10 +215,14 @@ module.exports = class ServiceTree extends Tree {
               return;
             }
 
-            let serviceLabels = service.getLabels();
-
+            let serviceLabels;
             if (service instanceof Service) {
-              serviceLabels = ServiceUtil.convertServiceLabelsToArray(service);
+              serviceLabels = Object.keys(service.getLabels()).map(key => ({
```
pierrebeitz: you basically say those helper functions don't need to be tested anymore? (`convertServiceLabelsToArray `, `createServiceFromResponse `) . 
(i'm totally ok with that. just want to make sure.)

plugins/services/src/js/structs/ServiceTree.js
```diff
@@ -201,10 +215,14 @@ module.exports = class ServiceTree extends Tree {
...
+                key,
+                value: service.getLabels()[key]
```
pierrebeitz: `Object.entries` seems like a good idea here: 

```typescript
serviceLabels = Object.entries(service.getLabels()).map(([key, value])=> ({key, value}));
```

plugins/services/src/js/structs/ServiceTree.js
juliangieseke: well, no - BUT…
i tried refactoring it in a way that preserves the tests, but failed… I can see that some of the functions have been extracted for easier testing and thats a good idea in general - but extracting them into a util and not having that util pure is worse than not having them extracted IMHO… I tried to come up with a test testing this method, but failed in a reasonable timeframe…

not entirely sure if this change is "good", but I found it to be "good enough" 🤷‍♂️

plugins/services/src/js/structs/ServiceTree.js
juliangieseke: copy pasted 🤷‍♂️  I can change, if you want me to change.

plugins/services/src/js/structs/__tests__/Application-test.js
juliangieseke: uh, will fix 👍

plugins/services/src/js/structs/__tests__/Framework-test.js
juliangieseke: well… we _should_… but I think adding a quick catch for that is actually a good idea… will do 👍

plugins/services/src/js/utils/ServiceUtil.js
```diff
@@ -135,27 +108,51 @@ const ServiceUtil = {
...
   },
 
-  convertServiceLabelsToArray(service) {
-    if (!(service instanceof Service)) {
-      return [];
-    }
+  isSDKService(service) {
+    return (
+      service &&
+      service.getLabels &&
```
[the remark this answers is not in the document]
juliangieseke: dont think so… will add that test 👍

plugins/services/src/js/utils/ServiceUtil.js
juliangieseke: fair point 👍

plugins/services/src/js/utils/ServiceUtil.js
juliangieseke: yeah, thanks for that - i just shuffled this code around too much to see where I can further improve it, like your suggestion 👍

plugins/services/src/js/structs/ServiceTree.js
juliangieseke: no idea, just copy&pasted this code, will check 👍

plugins/services/src/js/structs/ServiceTree.js
pierrebeitz: agree

plugins/services/src/js/structs/__tests__/Framework-test.js
pierrebeitz: this would also make the logic in getWebURL a little easier! 🎉

juliangieseke:
addressed feedback & rebased 👍 
cc @pierrebeitz @DanielMSchmidt

plugins/services/src/js/utils/ServiceUtil.js
```diff
@@ -135,27 +108,59 @@ const ServiceUtil = {
...
-  isSDKService(service) {
-    if (!(service instanceof Service)) {
-      return false;
+  /**
+   * this function generates webui url from given labels.
+   *
+   * - DCOS_SERVICE_WEB_PATH is a newly introduced label which
+   *   contains a url part starting with `/`
+   * - DCOS_SERVICE_PORT_INDEX & DCOS_SERVICE_SCHEME are legacy
+   *   labels used to detexct if service has webui
```
pierrebeitz: typo 🙈

plugins/services/src/js/utils/ServiceUtil.js
```diff
@@ -135,27 +108,59 @@ const ServiceUtil = {
...
+      DCOS_COMMONS_API_VERSION
+    } = labels;
+    const serviceName = encodeURIComponent(DCOS_SERVICE_NAME);
+    const baseUrl = `${rootUrl}/service/${serviceName}`;
+
+    if (serviceName && DCOS_SERVICE_WEB_PATH) {
+      const webPath =
+        DCOS_SERVICE_WEB_PATH.charAt(0) === "/"
+          ? DCOS_SERVICE_WEB_PATH
+          : "/" + DCOS_SERVICE_WEB_PATH;
```
pierrebeitz: as i played around with the code before, i have a patch here. i like it a little better. proceed as you like.

```diff
diff --git a/plugins/services/src/js/utils/ServiceUtil.js b/plugins/services/src/js/utils/ServiceUtil.js
index f6515b0d0..c040c758c 100644
--- a/plugins/services/src/js/utils/ServiceUtil.js
+++ b/plugins/services/src/js/utils/ServiceUtil.js
@@ -133,34 +133,23 @@ const ServiceUtil = {
    */
   getWebURL(labels, rootUrl) {
     const {
-      DCOS_SERVICE_NAME,
-      DCOS_SERVICE_PORT_INDEX,
-      DCOS_SERVICE_SCHEME,
-      DCOS_SERVICE_WEB_PATH,
-      DCOS_COMMONS_API_VERSION
+      DCOS_COMMONS_API_VERSION: apiVersion,
+      DCOS_SERVICE_NAME: name,
+      DCOS_SERVICE_PORT_INDEX: portIndex,
+      DCOS_SERVICE_SCHEME: scheme,
+      DCOS_SERVICE_WEB_PATH: webPath = ""
     } = labels;
-    const serviceName = encodeURIComponent(DCOS_SERVICE_NAME);
-    const baseUrl = `${rootUrl}/service/${serviceName}`;
+    const serviceName = encodeURIComponent(name);
+    const hasNewWebUiLabel = serviceName && webPath;
+    const hasOldWebUiLabel = serviceName && portIndex && scheme && !apiVersion;

-    if (serviceName && DCOS_SERVICE_WEB_PATH) {
-      const webPath =
-        DCOS_SERVICE_WEB_PATH.charAt(0) === "/"
-          ? DCOS_SERVICE_WEB_PATH
-          : "/" + DCOS_SERVICE_WEB_PATH;
-
-      return `${baseUrl}${webPath}`;
-    }
-
-    if (
-      serviceName &&
-      !DCOS_COMMONS_API_VERSION &&
-      DCOS_SERVICE_PORT_INDEX &&
-      DCOS_SERVICE_SCHEME
-    ) {
-      return `${baseUrl}/`;
+    if (!hasNewWebUiLabel && !hasOldWebUiLabel) {
+      return "";
     }

-    return "";
+    return `${rootUrl}/service/${serviceName}/${
+      webPath.startsWith("/") ? webPath.substring(1) : webPath
+    }`;
   }
 };
```

juliangieseke: yup, yours is better, mind just commiting to the branch? seems like its on your machine anyway :)

pierrebeitz: Pushed. What do you think on slashing slashes? 

> Also i don't feel confident about manipulating slashes and rather anticipated a "of course there are always slashes, as the API-docs show here". Now that this is not the case i'm not sure whether it's better in or out.

Also someone yesterday forbid me to approve this PR now. :)

juliangieseke: * slashes: its just a string and there is absolutely no enforcement that guarantees it starting with a slash - I'd keep the check here (even if I generally agree with you)
* approve: I reviewed your code - you can approve it again ("in my behalf" when it comes to your commit)

plugins/services/src/js/containers/services/ServicesTable.js
```diff
@@ -155,9 +155,9 @@ class ServicesTable extends React.Component {
       return null;
     }
 
-    return (
+    const webLink = (
```
nLight: I don't see us using the `webLink` variable anywhere I'd assume this is a forgotten change?

plugins/services/src/js/structs/__tests__/Application-test.js
```diff
@@ -498,7 +498,7 @@ describe("Application", function() {
   });
 
   describe("#getWebURL", function() {
```
nLight: I would say if we remove the method from the class we should also move the test into `Service-test.js` otherwise we kinda test the inheritance.

plugins/services/src/js/utils/ServiceUtil.js
```diff
@@ -135,27 +108,48 @@ const ServiceUtil = {
...
-    }
-
-    const labels = service.getLabels();
-    if (labels == null) {
-      return [];
-    }
-
-    return Object.keys(labels).map(key => ({ key, value: labels[key] }));
+  isSDKService(service) {
+    return (
```
nLight: why are we dropping the `if (!(service instanceof Service))` check? 
It feels rather safe to test `getLabels` existence though having the function doesn't mean the output will be expected. Checking on a class would be much a safer bet.

tests/_fixtures/marathon-1-task/groups-sdk-services.json
```diff
@@ -1,56 +1,36 @@
 {
   "id": "/",
-  "dependencies": [
-
-  ],
+  "dependencies": [],
```
nLight: the change in the file can be dropped to shrink down the size of the diff.

tests/_fixtures/marathon-1-task/groups-sdk-services.json
```diff
@@ -66,54 +46,39 @@
...
           "upgradeStrategy": {
             "minimumHealthCapacity": 1,
             "maximumOverCapacity": 1
           },
           "labels": {
             "DCOS_COMMONS_API_VERSION": "v1",
             "DCOS_PACKAGE_NAME": "sdk-sleep",
-            "DCOS_PACKAGE_FRAMEWORK_NAME": "/services/sdk-sleep"
+            "DCOS_PACKAGE_FRAMEWORK_NAME": "/services/sdk-sleep",
+            "DCOS_SERVICE_WEB_PATH": "/test"
```
nLight: Keeping the important part ofcourse. VSCode has an option to save without reformating

juliangieseke:
Going to address change requests and clean up commits tomorrow - if anyone else has time before that, feel free to do it!

tests/_fixtures/marathon-1-task/groups-sdk-services.json
pierrebeitz: those changes have been factored into separate commits, so i'd consider this well done.

also i'd vote for keeping it as it is, as else we'd waste the time of the next person touching the file, wouldn't we? :)

plugins/services/src/js/structs/ServiceTree.js
DanielMSchmidt: Okay, do we want that error to bubble through?

plugins/services/src/js/containers/services/ServicesTable.js
juliangieseke: good catch… it actually was a leftover from a test I did - should have been reverted. will fix it

plugins/services/src/js/structs/ServiceTree.js
juliangieseke: I think in that case its okay, if we cant identify something as Pod/Application/Framework, we have a problem 🤔

plugins/services/src/js/structs/__tests__/Application-test.js
juliangieseke: `Service` struct doesnt has necessary implementation of `getLabels` method, we cant add tests there without extending it. But I removed these and added the single one from `Framework` instead.

plugins/services/src/js/utils/ServiceUtil.js
juliangieseke: it didnt exist in `src/js/ServiceUtil` which was used by `Service` before… if we add it, we're creating circular dependency 😞

tests/_fixtures/marathon-1-task/groups-sdk-services.json
juliangieseke: I'll keep it, but extract the change into its own commit (like it should have been in the first place, my bad)

PR #3650, opened 2019-03-12
Delete extra files from the Universe templates

nLight:
This PR remove extra `marathon` and `config` templates that we don't need in our Universe package. Moreover having `marathon` template would mean the package is not CLI only and can be _installed_ as an application running on dcos.

## Testing

try running steps from the Jenkins file https://github.com/mesosphere/dcos-ui-plugins-private/blob/master/Jenkinsfile#L173

with a change in the last line: `../dcos-commons/tools/build_package.sh 'dcos-ui' ./ -a ./release.tar.gz local`

so `local` instead of `aws`

so that we don't actually publish it to AWS :)


juliangieseke:
Draft? 🤔

nLight:
@juliangieseke I started as a draft and wanted to make sure the Testing part is valid before asking for reviews and got distracted 😅

juliangieseke:
wasnt aware of that feature

PR #3664, opened 2019-03-15
chore: remove obsolete font-files

pierrebeitz:
can't find any reference to them anywhere, so let's remove them!


juliangieseke:
CNVS? 🤔

pierrebeitz:
good idea! it seems to be used for the docs of cnvs (also in 1.1.14, the version we use) but not in any of the less-files that are required by the ones we have:

https://github.com/mesosphere/cnvs/blob/develop/docs/_assets/styles/content/fonts.less#L5

(please double check, as i might be lacking context/history).

also i started dcos-ui and had a look at the network tab. no font-files have been loaded while navigating through random pages and those i'd suspect to load font-files (basically the modals with JSON-editors).

PR #3665, opened 2019-03-15
chore: remove deps that are included via @dcos/tslint-config

pierrebeitz:
those are included via https://github.com/dcos-labs/tslint-config/blob/master/index.js and not used anymore


pierrebeitz:
apparently we need to have those deps in our project without ever referencing them directly. 🤔

PR #3671, opened 2019-03-15
DCOS-50271: Check for labels with duplicate keys

natmegs:
Add validation to jobs form to check that the user hasn't tried to add multiple labels with the same key.

Closes DCOS-50271.

## Testing

- Go to jobs > create
- In the first tab, add an id, command ("test", "sleep 100")
- Go to run config tab
- Add a label with "a" for key and "b" for value
- Add another label with "a" for key and "c" for value. You should see this
Screen Shot 2019-03-15 at 4 19 42 PM

- Click submit, you should see errors in three places (banner in form tab, the label inputs, and in the JSON editor)
Screen Shot 2019-03-15 at 4 20 03 PM
Screen Shot 2019-03-15 at 4 20 12 PM
Screen Shot 2019-03-15 at 4 20 20 PM

## Trade-offs

Need to validate labels separately from the rest of the validation since they are stored in a different format in the job spec (tracked internally) compared to the job output (what's in JSON editor).

## Dependencies

N/A


locale/zh/messages.json
```diff
@@ -918,9 +920,9 @@
   "The CPUs offered": "已提供 CPU",
   "The GPUs offered": "已提供 GPU",
   "The Mesos master process orchestrates agent tasks.": "Mesos 管理节点进程编排代理任务。",
-  "The command that is executed. This value is wrapped by Mesos via `/bin/sh -c job.cmd`. Either `cmd` or `args` must be supplied. It is invalid to supply both `cmd` and `args` in the same job.": "",
   "The UI rollback has failed due to an error.": "",
   "The UI upgrade has failed due to an error.": "",
+  "The command that is executed. This value is wrapped by Mesos via `/bin/sh -c job.cmd`. Either `cmd` or `args` must be supplied. It is invalid to supply both `cmd` and `args` in the same job.": "",
```
natmegs: I have no idea why it would remove this message and then re-insert it a few lines below 🤷‍♀️

juliangieseke: I think lingui has some internal method to match lines with files, like the `line` attribute it had in the other format 🤔

pierrebeitz:
i extracted `allUniq`- and `isUniqIn`-validations and used them in the new validation-function. i also extended all validations to accept a message-override, so one can specify field-specific-messages.

PR #3672, opened 2019-03-16
update renovate config

juliangieseke:
lets try again…

## Trade-offs

`dependencies` that are behind in `major` are pinned to that `major` by only enabling `patch` and `minor` updates
same for all `devDependencies`

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,31 +1,26 @@
 {
-  "enabled": true,
-  "semanticCommits": true,
-  "rebaseStalePrs": true,
-  "labels": [
-    "renovate"
-  ],
+  "extends": ["config:js-app", ":semanticCommits", ":rebaseStalePrs"],
+  "labels": ["renovate"],
   "packageRules": [
     {
       "packageNames": [
-        "@dcos/http-service",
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -1,31 +1,26 @@
...
+        "clipboard",
+        "d3",
+        "flux",
+        "graphql-tools",
+        "inversify",
+        "query-string",
+        "react-ace",
+        "react-bootstrap-typeahead",
+        "react-router",
+        "redux"
```
pierrebeitz: i never used renovate before. do i get correctly that it'll only create PRs for the packages declared in `packageNames`? if so, how is this list compiled?

juliangieseke: no, this config applies special settings to these packages - I added them manually. Its a list of things that are behind in major version (we're using redux v3, but most recent is v5)

Idea is to manually update these packages and remove them from this list.

juliangieseke: everything thats not in this list will be updated to `latest` by renovate.

pierrebeitz: now i also understand the things in the trade-offs-section. thanks!

PR #3704, opened 2019-03-18
DCOS-19315: remove force updates

DanielMSchmidt:
## Testing

- Please click through all pages that might be affected (I don't want to give detailed pages, because if I miss one you miss them too) and see if they still operate like normal. 

## Trade-offs

- Decided to only change this on a level in the component tree that effects a lot of element on rerenders, this should maximize the impact of this PR
- I could not trace back every removed `forceUpdate`, but I still think we should ship them. If they hide an obscure bug, then we need to create better integration tests to catch that in the future. I see no other way, except keeping them and limiting the performance for all users in the future.
- I choose to use the component state in a very trivial way, as this is IMHO also about removing magic

## Screenshots

|  Page | Before  | After  |
|---|---|---|
| Dashboard  |  Bildschirmfoto 2019-03-19 um 16 39 45 |  Bildschirmfoto 2019-03-19 um 16 38 39 |
| Deployments | Bildschirmfoto 2019-03-19 um 16 41 03  | Bildschirmfoto 2019-03-19 um 16 42 18  |
| Nodes Agents  | Bildschirmfoto 2019-03-19 um 16 43 38  |  Bildschirmfoto 2019-03-19 um 16 44 38 |
| Nodes Detail| Bildschirmfoto 2019-03-19 um 16 45 45 | Bildschirmfoto 2019-03-19 um 16 46 43 |
| Service Detail | Bildschirmfoto 2019-03-19 um 16 48 13 | Bildschirmfoto 2019-03-19 um 16 51 40 |

### DIFF
--- a/plugins/nodes/src/js/pages/nodes/NodeDetailPage.js
+++ b/plugins/nodes/src/js/pages/nodes/NodeDetailPage.js
@@ -23,7 +23,7 @@ class NodeDetailPage extends mixin(TabsMixin, StoreMixin) {
 
     this.store_listeners = [
       { name: "summary", events: ["success"], suppressUpdate: true },
-      { name: "state", events: ["success"], suppressUpdate: false },
+      { name: "state", events: ["success"], suppressUpdate: true },
       {
         name: "nodeHealth",
         events: ["nodeSuccess", "nodeError", "unitsSuccess", "unitsError"],
@@ -96,6 +96,13 @@ class NodeDetailPage extends mixin(TabsMixin, StoreMixin) {
     }
   }
 
+  onSummaryStoreSuccess() {
+    this.setState({
+      summaryStatesProcessed: MesosSummaryStore.get("statesProcessed"),
+      summaryStates: MesosSummaryStore.get("states")
+    });
+  }
+
   updateCurrentTab(nextProps) {
     const { routes } = nextProps || this.props;
     const currentTab = RouterUtil.reconstructPathFromRoutes(routes);
@@ -136,10 +143,7 @@ class NodeDetailPage extends mixin(TabsMixin, StoreMixin) {
   }
 
   render() {
-    if (
-      !MesosSummaryStore.get("statesProcessed") ||
-      !this.state.mesosStateLoaded
-    ) {
+    if (!this.state.summaryStatesProcessed || !this.state.mesosStateLoaded) {
       return this.getLoadingScreen();
     }
 
--- a/plugins/nodes/src/js/pages/nodes/NodeDetailPage.js
+++ b/plugins/nodes/src/js/pages/nodes/NodeDetailPage.js
@@ -22,12 +22,12 @@ class NodeDetailPage extends mixin(TabsMixin, StoreMixin) {
     super(...arguments);
 
     this.store_listeners = [
-      { name: "summary", events: ["success"], suppressUpdate: false },
+      { name: "summary", events: ["success"], suppressUpdate: true },
       { name: "state", events: ["success"], suppressUpdate: false },
       {
         name: "nodeHealth",
         events: ["nodeSuccess", "nodeError", "unitsSuccess", "unitsError"],
-        suppressUpdate: false
+        suppressUpdate: true
       }
     ];
 
--- a/plugins/nodes/src/js/pages/nodes/nodes-table/NodesTableContainer.js
+++ b/plugins/nodes/src/js/pages/nodes/nodes-table/NodesTableContainer.js
@@ -30,7 +30,7 @@ class NodesTableContainer extends mixin(StoreMixin, QueryParamsMixin) {
         name: "nodeHealth",
         suppressUpdate: true
       },
-      { name: "state", events: ["success"], suppressUpdate: false }
+      { name: "state", events: ["success"], suppressUpdate: true }
     ];
   }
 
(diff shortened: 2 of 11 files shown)


plugins/nodes/src/js/pages/nodes/NodeDetailPage.js
```diff
@@ -23,12 +23,12 @@ class NodeDetailPage extends mixin(TabsMixin, StoreMixin) {
     super(...arguments);
 
     this.store_listeners = [
-      { name: "summary", events: ["success"], suppressUpdate: false },
-      { name: "state", events: ["success"], suppressUpdate: false },
+      { name: "summary", events: ["success"], suppressUpdate: true },
+      { name: "state", events: ["success"], suppressUpdate: true },
```
nLight: removing those we should make sure we have `on<StoreName>Store<EventName>` functions. I.e. in the component we do have `onStateStoreSuccess` but nothing for summary and nodes

DanielMSchmidt: I mean, nothing breaks like this, so maybe we don't need that anymore

nLight: can we please ensure checking the components don't rely on data from the stores please?

plugins/nodes/src/js/pages/nodes/nodes-table/NodesTableContainer.js
```diff
@@ -30,7 +30,7 @@ class NodesTableContainer extends mixin(StoreMixin, QueryParamsMixin) {
         name: "nodeHealth",
         suppressUpdate: true
       },
-      { name: "state", events: ["success"], suppressUpdate: false }
+      { name: "state", events: ["success"], suppressUpdate: true }
```
DanielMSchmidt: There was a `onStateStoreSuccess` available already

plugins/services/src/js/containers/service-detail/ServiceDetail.js
```diff
@@ -163,7 +163,6 @@ class ServiceDetail extends mixin(TabsMixin) {
       this.hasVolumes()
     ) {
       this.tabs_tabs["/services/overview/:id/volumes"] = "Volumes";
-      this.forceUpdate();
```
DanielMSchmidt: Could not find a root cause for this call

plugins/services/src/js/pages/ServicesPage.js
```diff
@@ -37,7 +37,7 @@ var ServicesPage = createReactClass({
 
   componentWillMount() {
     this.store_listeners = [
-      { name: "notification", events: ["change"], suppressUpdate: false }
```
DanielMSchmidt: No idea which errors [this commit](https://github.com/dcos/dcos-ui/commit/b73c4540f9133b47af0b3ecacab73b3180415452) talks about, I found no data requirement

src/js/components/charts/Chart.js
```diff
@@ -28,7 +28,7 @@ var Chart = createReactClass({
       {
         name: "sidebar",
         events: ["widthChange"],
-        suppressUpdate: false
```
DanielMSchmidt: This was already handled

nLight:
The mixin by default does force updates and in some places we don't have the `suppressUpdate` specified thus having `forceUpdate` i.e. https://github.com/dcos/dcos-ui/blob/danielmschmidt/trying-if-tests-break-on-supress-update/src/js/components/Page.js#L84 Do we want fixing those places as well?

DanielMSchmidt:
Snap, my search did not find those. Will go through them 👍

DanielMSchmidt:
I'll skip small and unprioritized pages though :)

DanielMSchmidt:
@nLight Please take another look, I think I got all big pages and impactful spots.

nLight:
code looks solid haven't tested yet

DanielMSchmidt:
Thank you @mperrotti I got this error covered now 👍

DanielMSchmidt:
@mperrotti @nLight Just rebased, could you take another look 👍

mperrotti:
@DanielMSchmidt - I'm still seeing that error :(
However, I just noticed it's also happening on the soak cluster, so it's not related to these changes. Still something we should address though.

Video of error on soak113s: https://cl.ly/6bc0d5239644

DanielMSchmidt:
@nLight @mperrotti Could you reapprove after the rebase?

src/js/components/Page.js
```diff
@@ -58,7 +57,13 @@ PageHeader.propTypes = {
 const Page = createReactClass({
   displayName: "Page",
 
-  mixins: [InternalStorageMixin, StoreMixin],
+  mixins: [StoreMixin],
```
nLight: I hate `InternalStorageMixin` even more than `StoreMixin`

DanielMSchmidt:
@brandonc @natmegs Hey folks, I invested some time today to debug the root cause and I am unable to even reproduce the issue. Could you maybe check if this is not a preexisting bug and / or take it from here and debug the issue? Neither problem would appear once in like 30 minutes clicking forth and back 🙈

DanielMSchmidt:
Awesome, could you also check @natmegs or may I dismiss your review?

PR #3722, opened 2019-03-20
DCOS-50468: Nodes Table Loader does not behave as expected

GeorgiSTodorov:
Change the initial value for nodes to null
to show loader while data is loading.

Closes https://jira.mesosphere.com/browse/DCOS-50468

## Testing
1. Go to Nodes tab
2. Refresh the page
3. Verify that an empty table is not visible as data is loading

## Trade-offs
None.

## Dependencies
None.

### Before
Peek 2019-03-20 11-42

### After
Peek 2019-03-20 15-41


GeorgiSTodorov:
@natmegs , good catch. I fixed it.
The failing integration test is flaky.

plugins/nodes/src/js/components/NodesTable.tsx
```diff
@@ -155,7 +163,7 @@ export default class NodesTable extends React.Component<
   componentWillReceiveProps(nextProps: NodesTableProps): void {
     this.setState(
       this.updateData(
-        nextProps.hosts.getItems(),
+        nextProps.hosts ? nextProps.hosts.getItems() : null,
```
nLight: do we even need updating data when we don't get `nextProps.hosts`?

nLight: sorting comes from state 🤔

GeorgiSTodorov: I can imagine some situations where we do want to update even when `nextProps.hosts` is null or undefined or empty.

nLight: Ok, I see 1 case at least. When for some reason `props.hosts` turn to undefined we want to `nullify` the `state.data`  thus showing loader.

PR #3727, opened 2019-03-20
DCOS-50195: preserve modal on schedule failure

natmegs:
Ensure the jobs form modal stays open when a job schedule fails when creating the first job in the list.

More background into the root cause of the bug in the [JIRA](https://jira.mesosphere.com/browse/DCOS-50195).

Closes DCOS-50195

## Testing

Testing steps also in video: https://cl.ly/996cf3e97d63 
- Ensure there are no jobs in the jobs page
- Click create a job
- Enter bare minimum for a valid job (ex: ID: "test", Command: "sleep 100")
- Go to schedule tab, enter an _invalid_ schedule (ex: ID: "test", CRON: "1234")
- Submit
- **The form modal should stay open**
- There should be an error visible for `schedule.cron`
- Fix the error (replace with ex: "0 0 4 * *") and submit
- You should see job list with the one job you just created, indicated as "Scheduled"

## Trade-offs

Not necessarily tradeoffs, but there are some things you should be aware of in this PR. It _removes_ the `JobsOverviewEmpty` component and adds the presentational part of that component into `JobsOverviewList`. Previously, the two components rendered their own `JobsFormModal` and had duplicated logic around modal opening/closing. This PR consolidates the view into one component that renders either the empty view or list view depending on the `data` prop. (aka moving control of what gets rendered from `JobsOverview` component to `JobsOverviewList`)

## Dependencies

N/A


plugins/jobs/src/js/components/JobsOverviewList.tsx
```diff
@@ -69,25 +71,56 @@ class JobsOverviewList extends React.Component<
...
+            searchString={this.props.filter || ""}
+          />
+        </FilterBar>
+        <JobsOverviewTable data={data} />
+      </div>
+    );
+  }
+
+  getEmptyContent() {
+    return (
```
DanielMSchmidt: NitPick, feel free to ignore: Why don't we make the getXZYs small functional component that lives inside this file? I have seen this quite some times now, but mostly in older code where I could not ask for the reason 😇

PR #3733, opened 2019-03-21
DCOS-46413: Server side error message for invalid input in package creation does not go away after updating the field

GeorgiSTodorov:
Hide cosmos errors when editing the form.

Closes https://jira.mesosphere.com/browse/DCOS-46413

## Testing
1. Go to Catalog tab.
2. Select a package and enter some error in the configuration that Cosmos would catch, for example in chronos, go to "Chronos" tab and enter "/Chronos" as ID (invalid because of capital letter).
3. Go to "Review & Run" and try to submit.
4. An error should appear. As you go back, the error should disappear. It shouldn't appear even when going back to the review screen before clicking the submit button.

## Trade-offs
My initial wish was to use `onCosmosPackagesStoreServiceDescriptionError` from `EditFrameworkConfiguration.js` to clear errors, but I couldn't pass it as props for some reason.

## Dependencies
None.

### Before
Peek 2019-03-21 15-21

### After
Peek 2019-03-21 15-19


GeorgiSTodorov:
`npm run build` runs fine locally.

PR #3738, opened 2019-03-21
DCOS-49957: code splitting

mperrotti:
Before, all JavaScript loaded had to be download and parsed from 1 bundle before the page could be rendered.
This PR takes out some of our larger dependencies that aren't critical to using DC/OS and loads them asynchronously. Chart-related code is set up to download immediately after the critical JS because the first page after /login is /dashboard, which contains those charts.

## Testing
1. Open the Network tab of dev tools, hit the record button, and reload the page.
First, the main bundle and the vendor bundle should download.
Next, chart-related bundles should load.

2. With the Network tab still open, navigate to the service create form
More JavaScript should be downloaded for the JSON editor

You can also run `npm run analyzebundle` to see the bundle sizes visualized

## Trade-offs
- If a user has a very slow connection, the components that are loaded asynchronously may show a loading state before appearing
- Not really a tradeoff, but the code from `node_modules` is placed in a separate vendor bundle. It doesn't make a difference if it's separated, so I'm happy to change it back to one big bundle if anybody has an issue with this.

## Screenshots
AsyncLoading_CreateService


pierrebeitz:
this is awesome!

i listed the .js-files before and after to see what we pack now. 

### before

```
132B 1.c1f79202fd4e85bb8683.js
 65K 2.c1f79202fd4e85bb8683.js
5.6M index.c1f79202fd4e85bb8683.js
```

### after

```
147K 0.861f6aa80f27361899e6.bundle.js
 65K 13.14df77c90086b636d115.bundle.js
139B 4.94a7941a3add155315d4.bundle.js
5.8K 9.9d30c65a8600d9178e7d.bundle.js
1.4K hosttimeserieschart.55e3cab0ee2d4f946699.bundle.js
 14K hosttimeserieschart~resourcetimeserieschart.fc13aecc8a3db171f2d5.bundle.js
2.9M index.029dfdcdd2a89773d43d.bundle.js
8.7K jsoneditor.1e95067c93c08fdf0551.bundle.js
1.5K resourcetimeserieschart.c397eea906a3ff346760.bundle.js
2.7K runtime.4ffc23fd8971cb8f3dd8.js
5.3K taskschart.7e7e3c33836e2130955c.bundle.js
2.0M vendors~index.55725b0bd0647b1dab8f.bundle.js
 94K vendors~jsoneditor.ae859c9ce898595ddd4d.bundle.js
387K vendors~jsoneditor~yamleditorschemafield.a2717668ffb33b038410.bundle.js
5.7K yamleditorschemafield.94d8a084896deb4e0ba8.bundle.js
```

i wonder whether there's a size we want our modules to have minimally - but that's something for another time as this feels like a huge improvement!

mperrotti:
@natmegs - I investigated route-based splitting. We're going to start with `/organization/`, `/networking/`, and `/secrets/` routes. They're some of our least visited views with the biggest JS footprint.

PR #3743, opened 2019-03-22
DCOS-49183: [SOAK] Plans Tab: URIError: URI malformed

GeorgiSTodorov:
Add a check if service id and scheduler task id
are valid URI and don't display scheduler logs if not.

Closes https://jira.mesosphere.com/browse/DCOS-49183

## Testing
1. In `dcos-ui/plugins/services/src/js/components/SDKPlansTab.tsx` change line 79 to:
```
    let { service, schedulerTaskId } = this.props;
    schedulerTaskId = "\uD800";
```
2. Start a service that support SDK plans, for example kafka.
3. Go to plans tab and see that it works and has no "view scheduler logs" button.

## Trade-offs
I tried to reproduce the error, but couldn't, so mocking it is the next best thing.

## Dependencies
None.

### Before
Screenshot from 2019-03-22 14-02-29

### After
Screenshot from 2019-03-22 14-03-12


plugins/services/src/js/components/SDKPlansTab.tsx
```diff
@@ -77,12 +77,19 @@ class SDKPlansTab extends React.PureComponent<SDKPlansTabProps, {}> {
       return null;
     }
     const { service, schedulerTaskId } = this.props;
+    let serviceIdURI, schedulerTaskIdURI;
+
+    try {
```
juliangieseke: there already is `if(!this.props.schedulerTaskID)` clause above, what about moving if after the deconstruction and just extending that one with ` || !service || !service.id`? 🤔 
```suggestion
const { service, schedulerTaskId } = this.props;

if(!service || !service.id || !schedulerTaskId) {
  return null;
}
```

brandonc: I think there is an issue with not just falsy values getting encoded, but invalid ones as well. encodeURIComponent will raise a URIError if it can't be encoded for some reason

brandonc:
Helpful testing instructions

plugins/services/src/js/components/SDKPlansTab.tsx
GeorgiSTodorov: I agree with @brandonc . I don't know exactly why this occurs, but the problem is not that these are null or undefined (`encodeURIComponent(null)` and `encodeURIComponent(undefined)` will give no error, but rather `"null"` and `"undefined"`), but that they are invalid URI - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Malformed_URI#Encoding .

GeorgiSTodorov:
@pierrebeitz 
1. Do you mean that we should merge only something that will allow us to get more context in sentry?
2. Is the use of `let` a blocking change for you?

pierrebeitz:
@GeorgiSTodorov 	i'm sorry for the noise at all. i first thought that we might be about to render invalid links and thus decided to write a lengthy comment. nothing is blocking here. i will make sure you know if i think something is blocking (e.g. with a 🚫 	as well as a "requesting changes" on github). there's rarely an opportunity that i'd consider code-style or rather -preference blocking.

i still think that opting for a solution that makes us see the error- and corner-cases is favourable to fixing stuff in a fashion that makes us not care about it anymore. we can still choose to ignore those errors in sentry - but at least we know they are existent.

GeorgiSTodorov:
@pierrebeitz , no need to apologize, I value your feedback.
I would say that in this case we are still fixing the error, not just ignoring it (even if we don't know the exact details). How would you go about addressing this issue?

pierrebeitz:
i'd love this to be merged as is to fix the bug. afterwards it would be great to open a follow-up to implement something that makes us aware of something being fishy here if **you** (as the expert on that matter) think that's a reasonable thing to do. 

i would do so, as i don't know why we have to process weird codepoints or values that make `encodeURIComponent` throw and would love to learn more about it.

GeorgiSTodorov:
@pierrebeitz , I will open a follow-up next spike day. Thanks for your help.

GeorgiSTodorov:
@pierrebeitz , throwing an error breaks the page. And I don't think we can use the `const a = b ? c : d;` here, since we have assignment inside a `try{}`.

PR #3746, opened 2019-03-22
Try mockserve

DanielMSchmidt:


pierrebeitz:
do we still need / wanna try that? what would be next steps?

DanielMSchmidt:
Want yes, but time is always a constraint :)

PR #3748, opened 2019-03-22
feat(ServiceStatus): add new framework statuses

Poltergeist:
This adds the sdk packages statuses.

I will also add fixtures for each state.
CLOSES DCOS-49583, DCOS-49584

## Testing

You can test this by setting fixtures to `true` in your `config.dev.ts` file then you should see the service table as in the screenshot.

## Screenshots

image

closes DCOS-50998, DCOS-51166, DCOS-49583, DCOS-49584


plugins/services/src/js/structs/Framework.js
```diff
@@ -66,6 +123,55 @@ module.exports = class Framework extends Application {
     return this._spec;
   }
 
+  getCheckStatus() {
```
natmegs: Is this supposed to be used by consumers of the Framework struct? So far it looks like it is just used internally by `getStatus`, `getServiceStatus`. I wonder if it would be better to move to a util?

plugins/services/src/js/structs/Framework.js
```diff
@@ -66,6 +123,55 @@ module.exports = class Framework extends Application {
...
+    }, false);
+
+    if (checkStatus == null) {
+      return false;
+    }
+
+    return checkStatus;
+  }
+
+  getStatus() {
```
natmegs: I'm finding the naming of this method as well as `getServiceStatus` a bit ambiguous. Hard to tell what the difference is without doing some digging (one returns string display name and the other an object)?

Poltergeist: I can't rename these functions since they are overwriting functions form application. But what I can do is add more documentation as in JS doc blocks.

AmrAbdelrazik:
@poltergeist can you please chrck latest names from the document?

plugins/services/src/js/structs/Framework.js
natmegs: Thinking about this again after we discussed this in standup where the concern was that we would have to pass `this` to a util function - would it make sense to move to a util but pass tasks as param so there's no dependency on `this`? I can see there being confusion about when/where to use `getCheckStatus` vs `getStatus` vs `getServiceStatus` but maybe we could resolve that with some clarifying code comments.

tests/_fixtures/marathon-pods/groups.js
```diff
@@ -1,5 +1,236 @@
+const createFrameworks = () =>
```
pierrebeitz: just to be sure: there's nothing in here that shouldn't be public, right?

plugins/services/src/js/structs/Framework.js
pierrebeitz: i'd second that. factoring a function out of the Framework-class (not necessarily into a new file) and passing tasks would be a good thing.

tests/_fixtures/marathon-pods/groups.js
Poltergeist: I don’t think so but I will all unnecessary information to thin out this change.

src/styles/components/service-table/styles.less
```diff
@@ -34,7 +34,7 @@
       }
 
       &-status {
```
mperrotti: These changes make sense, but there will be conflicts when @GeorgiSTodorov merges the service table replacement PR.

I don't like how much content we're trying to squeeze into this table, but I trust we've tested this with users and this is the data they absolutely need to see in this view.

plugins/services/src/js/structs/Framework.js
```diff
@@ -66,6 +92,70 @@ module.exports = class Framework extends Application {
     return this._spec;
   }
 
+  getCheckStatus() {
+    const tasks = this.get("tasks") || [];
+    const checkStatus = tasks.reduce((acc, cur = {}) => {
```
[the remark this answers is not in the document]
pierrebeitz: here's a proposal: https://github.com/dcos/dcos-ui/commit/9331b155865d73dabe510f92108e1459ff519ed8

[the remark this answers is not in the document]
pierrebeitz: thanks!

pierrebeitz:
@Poltergeist 	could we rebase the removal of the unneeded fixture-information onto the first commit, so this data does not appeart in the git-history at all?

src/styles/components/service-table/styles.less
Poltergeist: In theory we could drop these changes if the service PR has merged but also we can drop all the stylings of the columns made in this file.

GeorgiSTodorov: My PR has two reviews, but CI is broken.

Poltergeist:
let's see to merge #3583 before this one.

natmegs:
I like the latest changes, I will wait to do final review since we anticipate a rebase and some changes after merging the service table PR.

nLight:
Hey @Poltergeist help me understand how the PR resolves `Unify App- and Pod-CheckResponses to be shown in the Services Overview` I don't see anything about apps nor pods here

pierrebeitz:
@nLight 	we found out that the data only needs to be processed and shown for frameworks for now. i renamed the according issue.

nLight:
filcts!

pierrebeitz:
rebased!

plugins/services/src/js/structs/Framework.js
```diff
@@ -1,15 +1,54 @@
...
+const frameworkStatusHelper = (displayName, key, priority) => ({
+  displayName,
+  key,
+  priority
+});
+
+// prettier-ignore
+/* eslint-disable no-multi-spaces */
+const FrameworkStatus = {
+  418: frameworkStatusHelper("Initializing",                    ServiceStatus.DEPLOYING,  1),
```
TattdCodeMonkey: should all of these `displayName` arguments be wrapped in `i18nMark` so we can `Trans` them when they are rendered in the UI?

plugins/services/src/js/structs/Framework.js
```diff
@@ -1,15 +1,54 @@
...
+  200: frameworkStatusHelper("Running",                         ServiceStatus.RUNNING,    1),
+  500: frameworkStatusHelper("Error Creating Service",          ServiceStatus.WARNING,    1),
+  204: frameworkStatusHelper("Deploying (Awaiting Resources)",  ServiceStatus.DEPLOYING,  2),
+  202: frameworkStatusHelper("Deploying",                       ServiceStatus.DEPLOYING,  2),
+  203: frameworkStatusHelper("Degraded (Awaiting Resources)",   ServiceStatus.RECOVERING, 4),
```
TattdCodeMonkey: Should this be `ServiceStatus.Warning` instead ? The design shows a red `CircleClose` icon for this status, not the `Yield` used for the `RECOVERING` status.

plugins/services/src/js/structs/Framework.js
```diff
@@ -1,15 +1,54 @@
...
+  key,
+  priority
+});
+
+// prettier-ignore
+/* eslint-disable no-multi-spaces */
+const FrameworkStatus = {
+  418: frameworkStatusHelper("Initializing",                    ServiceStatus.DEPLOYING,  1),
+  200: frameworkStatusHelper("Running",                         ServiceStatus.RUNNING,    1),
+  500: frameworkStatusHelper("Error Creating Service",          ServiceStatus.WARNING,    1),
```
TattdCodeMonkey: Should we introduce an `ERROR` status instead of using `WARNING` here? the design has groups with `Error` status and `Warning` status and it seems confusing to use `ServiceStatus.WARNING` when we want the status text to be `Error ...` and the icon to be `CircleClose` instead of `Yield`

TattdCodeMonkey: [ServiceStatus](https://github.com/dcos/dcos-ui/blob/ph/feature/DCOS-49583-Accurate-Status/plugins/services/src/js/constants/ServiceStatus.ts#L20-L53) does not have a `WARNING` status defined at all currently.

GeorgiSTodorov:
I agree with @TattdCodeMonkey 's feedback. Otherwise, it looks good.

pierrebeitz:
I converted stuff to typescript and unified the types of FrameworkStatus and ServiceStatus as they seem quite related.

It appears that there never have been icons specified for "N/A", "Delayed" and, "Waiting". I introduced those with the help of Julian. Also the types have been modelled in a way that makes it impossible to forget to add an icon to a new state.

I'm sorry for the noisy PR, but some plumbing had to be done to convert stuff to TS. Please re-review.

plugins/services/src/js/components/ServiceStatusIcon.tsx
```diff
@@ -34,12 +35,30 @@ const getTooltipContent = (service, content) => {
...
-class ServiceStatusIcon extends Component {
-  getDeclinedOffersWarning(service) {
+type ServiceOrPodOrTree = Service | ServiceTree | Pod;
+
+class ServiceStatusIcon extends React.Component<{
+  showTooltip?: boolean;
+  tooltipContent: JSX.Element;
+  service: ServiceOrPodOrTree;
+}> {
+  static propTypes = {
```
TattdCodeMonkey: do propTypes still make sense when using TS? these are just duplicates for whats in the props interface definition above.

plugins/services/src/js/components/__tests__/ServiceStatusProgressBar-test.js
```diff
@@ -1,10 +1,11 @@
 import React from "react";
 import { shallow } from "enzyme";
 
+import ServiceStatusProgressBar from "../ServiceStatusProgressBar";
```
TattdCodeMonkey: nitpick: why move this?

plugins/services/src/js/components/ServiceStatusIcon.tsx
pierrebeitz: as those are runtime-checks that will help us to prevent bugs when using stuff from TS i did not want to remove them. maybe we should gather opinions around that.

src/styles/components/service-table/styles.less
```diff
@@ -44,7 +44,7 @@
     }
 
     .service-status-icon-wrapper {
-      width: 110px;
+      width: 240px;
```
mperrotti: Why don't we just get rid of this width style and add `flex-justify-items-space-between` to the parent flex container?

plugins/services/src/js/components/ServiceStatusIcon.tsx
mperrotti: I'm also in favor of keeping them to help prevent bugs from passing the wrong props from a component that isn't converted to Typescript yet

plugins/services/src/js/constants/FrameworkStatus.ts
```diff
@@ -0,0 +1,25 @@
+import { i18nMark } from "@lingui/react";
+
+import { Status } from "./ServiceStatus";
+import StatusIcon from "#SRC/js/constants/StatusIcon";
+
+// prettier-ignore
+const StatusMap: Record<number, Status> = {
```
mperrotti: Some of the icons we show for framework statuses are inconsistent what we'd show for service statuses:
"Degraded (Awaiting Resources)" shows `StatusIcon.LOADING`
"Deploying (Awaiting Resources)" shows `StatusIcon.LOADING`
"Degraded (Recovering)" shows `StatusIcon.LOADING`
"Degraded" shows `StatusIcon.LOADING`
"Error Creating Service" shows `StatusIcon.STOPPED`
"Service Unavailable" shows `StatusIcon.ERROR`

Is this intentional?

mperrotti:
I'm getting some Typescript errors:
```
ERROR in /Users/mikep/dcos-ui/plugins/services/src/js/components/ServiceStatusIcon.tsx(23,42):
TS2345: Argument of type 'number' is not assignable to parameter of type 'string'.

ERROR in /Users/mikep/dcos-ui/plugins/services/src/js/components/ServiceStatusIcon.tsx(59,9):
TS2531: Object is possibly 'null'.

ERROR in /Users/mikep/dcos-ui/plugins/services/src/js/components/ServiceStatusIcon.tsx(76,20):
TS7006: Parameter 'item' implicitly has an 'any' type.

ERROR in /Users/mikep/dcos-ui/plugins/services/src/js/components/ServiceStatusIcon.tsx(109,46):
TS2345: Argument of type 'number' is not assignable to parameter of type 'string'.

ERROR in /Users/mikep/dcos-ui/plugins/services/src/js/components/ServiceStatusIcon.tsx(133,15):
TS2531: Object is possibly 'null'.
```

pierrebeitz:
@mperrotti i'm not able to reproduce those errors locally. CI also seems to be happy. any ideas?

plugins/services/src/js/constants/FrameworkStatus.ts
pierrebeitz: i don't really understand what you mean.

here it says
```
  203: Status(4, i18nMark("Degraded (Awaiting Resources)"), StatusIcon.ERROR),    // not LOADING
  204: Status(2, i18nMark("Deploying (Awaiting Resources)"), StatusIcon.WARNING), // not LOADING
  205: Status(4, i18nMark("Degraded (Recovering)"), StatusIcon.WARNING),          // not LOADING
  206: Status(3, i18nMark("Degraded"), StatusIcon.WARNING),                       // not LOADING
```

also there's no direct match to the service statuses (that i see). 🤔 

shall we sync tomorrow or do you have an idea how to make me understand what you mean?

mperrotti: Sure, let's chat tomorrow

mperrotti:
@pierrebeitz - That's strange...
We can take a look tomorrow when we're chatting about icons :)

GeorgiSTodorov:
Also, I would prefer to have a feature branch or at least commits saying which jira they are supposed to fix. Right now testing this is hard.

GeorgiSTodorov:
I created an empty group and this is what I got. Doesn't seem intended to me.
Peek 2019-04-05 12-28

pierrebeitz:
@GeorgiSTodorov i can not reproduce the brokenness of your screenshot unfortunately. would you mind debugging a little to find out what we can do to make this work on linux?

GeorgiSTodorov:
@pierrebeitz , you mean the first screenshot?

GeorgiSTodorov:
The icon wrapper class is causing the progress bar to move.
Screenshot from 2019-04-05 17-11-14

GeorgiSTodorov:
The overlapping problem appears fixed now, but these long statuses aren't pretty on smaller windows.
Peek 2019-04-05 17-16

pierrebeitz:
do you have any suggestions on how to make that better? shall we file an issue and leave it to UX?

mperrotti:
@izharbuendia - do you have any thoughts on the layout issue in the Status column @GeorgiSTodorov mentioned [in his last comment](https://github.com/dcos/dcos-ui/pull/3748#issuecomment-480292713)?

src/styles/components/service-table/styles.less
pierrebeitz: love it! will add!

GeorgiSTodorov:
File conflicts.

GeorgiSTodorov:
I am not sure why my comment https://github.com/dcos/dcos-ui/pull/3748#issuecomment-480290148 was marked as resolved, I still have this issue.

GeorgiSTodorov:
Removing the "flex-justify-items-space-between" class fixes it.
Peek 2019-04-08 11-58

pierrebeitz:
as it looks like the expectation is that SDK-packages don't show a progressBar anyways (see newest designs) the commit changing their styling has been removed.

plugins/services/src/js/components/ServiceStatusIcon.tsx
```diff
@@ -104,35 +123,33 @@ class ServiceStatusIcon extends Component {
...
-        <Trans render="span">
-          There are tasks in this queue that DC/OS has failed to deploy for{" "}
-          {DateUtil.getDuration(
-            Date.now() - DateUtil.strToMs(service.getQueue().since)
-          )}
-          .
-        </Trans>
-      );
-    }
+  getUnableToLaunchWarning(service: Service) {
```
GeorgiSTodorov: This can also be a pod, right?

plugins/services/src/js/components/ServiceStatusIcon.tsx
```diff
@@ -104,35 +123,33 @@ class ServiceStatusIcon extends Component {
...
+        There are tasks in this queue that DC/OS has failed to deploy for
+        {duration}.
+      </Trans>
+    );
   }
 
-  isUnableToLaunch(service) {
-    const queue = service.getQueue();
+  isUnableToLaunch(service: ServiceOrPodOrTree) {
+    const queue: any = service.getQueue();
```
GeorgiSTodorov: I think this is an array of objects.

plugins/services/src/js/components/ServiceStatusProgressBar.tsx
```diff
@@ -45,14 +45,9 @@ class ServiceStatusProgressBar extends React.Component<
...
-    const serviceStatus = service.getServiceStatus();
     const runningInstances = service.getRunningInstancesCount();
 
-    if (
-      serviceStatus === ServiceStatus.RUNNING ||
-      serviceStatus === ServiceStatus.STOPPED ||
-      serviceStatus === ServiceStatus.NA
-    ) {
+    if (!ServiceStatus.showProgressBar(service.getServiceStatus())) {
       return null;
```
GeorgiSTodorov: We are changing the logic here. It used to be a positive check and now it is negative. This might be relevant if we add more statuses in the future.

plugins/services/src/js/components/ServiceStatusIcon.tsx
pierrebeitz: A `Pod` is a `Service`. But you're right in that we most likely can get away with declaring `ServiceOrPodOrTree` like so:

```
type ServiceTreeNode = 
  Service | ServiceTree
```

Thanks for the nudge!

plugins/services/src/js/filters/ServiceAttributeIsFilter.js
```diff
@@ -2,16 +2,6 @@ import DSLFilterTypes from "#SRC/js/constants/DSLFilterTypes";
 import DSLFilter from "#SRC/js/structs/DSLFilter";
 import * as ServiceStatus from "../constants/ServiceStatus";
 
-const LABEL = "is";
```
GeorgiSTodorov: Why remove this constant?

plugins/services/src/js/components/ServiceStatusProgressBar.tsx
pierrebeitz: That's intentional. Would you feel more comfortable with explicitly mapping each Status `showProgressBar` to whether is showsProgress or not?

GeorgiSTodorov: Before we didn't show progress bar if it was "running", "stopped" or "na". Now, we show it if its' https://github.com/dcos/dcos-ui/pull/3748/files#diff-850dfaf716778762300d8d5407693b70R154 . We changed from checking what it wasn't to checking what it is.

plugins/services/src/js/filters/ServiceAttributeIsFilter.js
pierrebeitz: good question. i felt like it is really verbose and is not going to change anytime - thus i inlined it to make clearer that it can really be considered hardcoded. but i have no strong feelings here and would change it back if you think that's better.

GeorgiSTodorov: I also don't prefer one or the other way, but I was wondering.

plugins/services/src/js/components/ServiceStatusIcon.tsx
pierrebeitz: i don't have enough experience with the codebase to be capable of typing things easily. i think it's better to keep `any` here until we have a good type for `Queue`, as `any` is kind of screaming for attention while it can look like someone has already made a good decision in typing it (while `object[]` does not seem quite correct here).

so i'd propose to keep it `any` until we have introduced more typing to structs.

GeorgiSTodorov: Okay, but maybe add a todo?

plugins/services/src/js/components/ServiceStatusProgressBar.tsx
pierrebeitz: Yes indeed, that's intentional as we have a fixed set of statuses - with the set of statuses that have a progressBar being far smaller. you now need to opt in to have a statusBar with your status.

Would you feel more comfortable with explicitly mapping Statuses to whether they show a progressBar, e.g.

```
const STOPPED: Status = {
  priority: 2,
  displayName: i18nMark("Stopped"),
  icon: StatusIcon.STOPPED,
  showProgressBar: false
};
```
so that when someone introduces a new Status they are forced to decide on that?

GeorgiSTodorov: Which statuses do actually show a progress bar and which don't?

plugins/services/src/js/components/ServiceStatusIcon.tsx
pierrebeitz: i anticipate us enforcing a `no-any` policy (tslint has a checker for that) in the future, so we need to get rid of all the `any`s left in the code. until then i think we need to find a balance of typing and being able to get work done. i'd consider having a todo at every `any` overkill (as there will be quite some and they can easily be found with the help of `tslint`).

what do you think?

GeorgiSTodorov: I don't mind `any`-s in some cases, because sometimes `any` really is the best we can do, but `queue` is something that we can easily verify and see the type of, so adding the type should be pretty easy to do, even if we don't want this type in this PR. I am not saying to add a todo to every any (just the ones that should be a quick fix) and `no-any` policy doesn't seem like something we can and will do in the near future.

pierrebeitz: i am not sure about the type of queue. to me it looks like it is at least a 

```
type Queue = Array<{
  app: Application,
  count: number,
  delay: {
    timeLeftSeconds: number,
    overdue: boolean
  },
  since: string // iso8601,
  processedOfferSummary: OfferSummary[],
  //...

}>
```

i do not dare to add that, as i am missing too many concepts and knowledge here yet.

GeorgiSTodorov: Yeah, this is why we can add a todo, because it's an `any` that is not impossible to determine.

pierrebeitz: i am a strong believer in that `no-any` is very possible, as i worked with one large TS-codebase without an `any` and know that every elm-codebase is completely typed without any being even a possibility. i've not come across a use case where `any` could not be replaced with proper types.

GeorgiSTodorov: It's possible if the project is started with typescript in mind, but currently we have over 200 `any`-s without counting the private repository. Even if it's possible, it's so much work that I doubt we will ever do it. I think this is a good sync topic, so I added it to our list.

plugins/services/src/js/components/ServiceStatusProgressBar.tsx
pierrebeitz: Resolved in chat. The formerly shown are the ones in `ServiceStatus.showProgressBar`.

plugins/services/src/js/structs/Framework.js
```diff
@@ -1,15 +1,28 @@
...
+import * as ServiceStatus from "../constants/ServiceStatus";
 
 import Application from "./Application";
 import FrameworkSpec from "./FrameworkSpec";
 
+const getHighestPriorityStatus = tasks => {
+  const statuses = tasks
+    .map(t => findNestedPropertyInObject(t, "checkResult.http.statusCode"))
+    .map(ServiceStatus.fromHttpCode)
+    .filter(x => x);
```
Poltergeist: Maybe we should rename this to something with more meaning.

pierrebeitz: like `.filter(isTruthy => isTruthy)` or `.filter(status => status)`?
could you pick your favourite or make a suggestion please, as i hardly see added value here.

Poltergeist: Sorry I mend more describing to what is in the variable my suggestion would be status.

pierrebeitz: done

plugins/services/src/js/components/ServiceStatusIcon.tsx
```diff
@@ -10,15 +10,16 @@ import DateUtil from "#SRC/js/utils/DateUtil";
...
-import * as ServiceStatus from "../constants/ServiceStatus";
+import { Status } from "../constants/ServiceStatus";
 import Pod from "../structs/Pod";
 import Service from "../structs/Service";
 import ServiceTree from "../structs/ServiceTree";
 
 const UNABLE_TO_LAUNCH_TIMEOUT = 1000 * 60 * 30; // 30 minutes
 
-const getTooltipContent = (service, content) => {
+const getTooltipContent = (service: Service, content: JSX.Element) => {
```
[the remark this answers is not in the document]
pierrebeitz: # 👍 

thanks for educating! i will make use of that knowledge in the follow-up #3793.

pierrebeitz:
I think the filters do not make sense like this. i changed the filter bar's behaviour so that it accepts all states, e.g. `is:degraded`, although that's not documented anywhere.
there's an ongoing discussion with UX that i expect to produce a change regarding the filter-behaviour.

PR #3752, opened 2019-03-26
DCOS-50309: remove fixtures from bundle

mperrotti:
## Testing
Run `npm run analyzebundle`
The fixtures should not be included in the main bundle

Change `Config.dev.ts` to use fixtures
Run `npm start`
Click around and see that everything looks as expected with fixtures


GeorgiSTodorov:
Some fixtures do not work for me. For example, on master we have 9001 nodes, here we have 0.

PR #3754, opened 2019-03-26
DCOS-50310: replace momentjs

mperrotti:
## Testing
Running `npm run analyzebundle` should no longer show moment.js in the vendor bundle

Start dcos-ui with the Soak cluster
Click around and make sure relative times look the same as they do in master

Here are some places to look:
Service debug tab
Job run history table (go to Jobs tab and click a job)
Deploy modal

## Trade-offs
English locale is slightly different. Instead of saying `a minute ago`, it says `1 minute ago`. We can extend the default English locale provided by date-fns if we don't like this


plugins/jobs/src/js/components/JobsOverviewTable.js
```diff
@@ -22,6 +22,10 @@ import JobTableHeaderLabels from "../constants/JobTableHeaderLabels";
 
 const METHODS_TO_BIND = ["renderHeadline", "jobSortFunction"];
 
+const JobsCronTooltip = lazy(() =>
```
mperrotti: Lazy-loading this component because it depends on `prettycron`, which depends on `moment`.

I tried using [cronstrue](https://github.com/bradymholt/cRonstrue), but it uses different language than `moment`.

moment: https://cl.ly/69a125629a43
cronstrue: https://cl.ly/3e019537d16f

src/js/utils/DateUtil.ts
```diff
@@ -65,7 +68,25 @@ const DateUtil = {
...
+    //     new Date(ms).toISOString(),
+    //     "YYYY-MM-DD[T]HH:mm:ss.SSS[Z]"
+    //   )}`
+    // );
+    // console.log(
+    //   format(
+    //     new Date(ms).valueOf() + new Date(ms).getTimezoneOffset() * 60000,
+    //     "YYYY-MM-DD[T]HH:mm:ss.SSS[Z]"
+    //   )
+    // );
```
natmegs: 🎨

src/js/utils/DateUtil.ts
```diff
@@ -74,7 +95,11 @@ const DateUtil = {
    * @return {String} time string with the format 'YYYY-MM-DD hh:mm:ss'
    */
   msToLogTime(ms: Date | number): string {
-    return moment.utc(ms).format("YYYY-MM-DD hh:mm:ss");
+    return format(
+      new Date(ms).valueOf() + new Date(ms).getTimezoneOffset() * 60000,
+      "YYYY-MM-DD hh:mm:ss"
+    );
+    // return moment.utc(ms).format("YYYY-MM-DD hh:mm:ss");
```
natmegs: 🎨

GeorgiSTodorov:
It looks good, but as @natmegs correctly pointed out, we should remove the `console.log`s first.
Also, I prefer "1 minute ago" to "a minute ago".

Another thing I noticed is that here we show "half a minute" instead of the exact number of seconds. This might be bad if some people want to know the exact number.
Here:
Screenshot from 2019-03-27 10-19-54
master:
Screenshot from 2019-03-27 10-24-01

plugins/jobs/src/js/components/JobsOverviewTable.js
pierrebeitz: i don't know how crucial it is to change language, but i actually like the cronstrue-format better. also it's only 4kb when gzipped and does not bring a lot of unneeded features (prettycron also brings `later`, which in turn even forces [some webpack-specifics](https://github.com/dcos/dcos-ui/blob/411b5f11f29c5d81a67e1b09dcb6c18a8086f148/webpack/webpack.config.js#L81) on us) . :)

mperrotti: Yea, I like the cronstrue format better too. I just wasn't sure if we were ok with changing that in this ticket. I'd love to hear from others :)

src/js/utils/DateUtil.ts
mperrotti: 😅

plugins/jobs/src/js/data/__tests__/JobModel-test.js
```diff
@@ -669,13 +669,13 @@ describe("JobModel Resolver", () => {
...
                       jobId: 1,
                       status: "foo",
-                      createdAt: "1985-01-03t00:00:00z-1",
+                      createdAt: "1985-01-03t00:00:00z",
                       tasks: []
                     },
                     {
                       jobId: 2,
                       status: "bar",
-                      createdAt: "1990-01-03t00:00:00z-1",
```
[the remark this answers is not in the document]
mperrotti: I'm not sure which big red error blocks you're talking about.
Moment didn't throw an error for UTC format with the `-1` added, but date-fns and the browser both throw an error saying it's an invalid date.

I checked with @DanielMSchmidt to make sure there was no special reason for the `-1` part of the time string, and he couldn't remember.

mperrotti:
@GeorgiSTodorov - good catch on the formatting for < 1 minute.

plugins/jobs/src/js/components/JobsOverviewTable.js
brandonc: Who do we have to convince? product or design? I like cronstrue's format better, plus it has internationalization options.

plugins/services/src/js/containers/service-debug/ServiceDebugContainer.js
```diff
@@ -195,7 +195,8 @@ class ServiceDebugContainer extends React.Component {
     if (frameworkName != null) {
       return (
         <Trans render="span">
-          Rejected offer analysis is not currently supported for {frameworkName}.
+          Rejected offer analysis is not currently supported for {frameworkName}
+          .
```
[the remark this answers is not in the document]
mperrotti: Prettier did this. Do you think it's worthwhile to change it back?

plugins/jobs/src/js/components/JobsOverviewTable.js
mperrotti: @pierrebeitz and @brandonc:
I talked to the designers and Tobi in our Product Design sync today. They're ok with the cronstrue style of formatting.

I filed a ticket to follow-up on this: https://jira.mesosphere.com/browse/DCOS-51104

nLight:
legit integration test failure

juliangieseke:
👀 - we need to merge this to make plugins ci green again

nLight:
@juliangieseke 🤔 it _was_ green without the PR

juliangieseke:
fixed 🤞 

we need to fix our graphql implementation, finding an error is hard when errormessage gets swallowed somewhere…

juliangieseke:
> @juliangieseke 🤔 it _was_ green without the PR
no, plugins CI had a bug and the Run actually tested master instead of the PR 🙈 fix is in my #3767 PR

pierrebeitz:
When using this with fixtures it breaks, because `Dateutil.strToMs` is called with `0` (where it expects a string).

Checked because i anticipated a huge performance gain on the services table (which to date is fully occupied with momentjs trying to parse date-strings). And indeed it seems to be faster by a lot! 🎉

juliangieseke:
extended early exit in `strToMs` 🙈

plugins/services/src/js/containers/service-debug/ServiceDebugContainer.js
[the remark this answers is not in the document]
brandonc: We either need to change it back or run util:extract so that it falls back correctly.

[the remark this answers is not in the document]
brandonc: I'm pretty sure this is associated with this feature https://github.com/prettier/prettier/pull/4717 and not a configurable option. Let's keep the new formatting and I'll add a commit in order to unblock this PR

TattdCodeMonkey:
I ran this against soak and everything LGTM, but I agree with Brandon's request for changes related to `lingui`

brandonc:
https://github.com/dcos/dcos-ui/pull/3754#discussion_r271043098

> I'm pretty sure this is associated with this feature prettier/prettier#4717 and not a configurable option. Let's keep the new formatting and I'll add a commit in order to unblock this PR

pierrebeitz:
@juliangieseke 	@mperrotti 	we should have a (very small) follow up here for https://github.com/dcos/dcos-ui/pull/3754#pullrequestreview-221173174. a `prettier-ignore` will do: https://prettier.io/docs/en/ignore.html#jsx.

juliangieseke:
brandon fixed it: https://github.com/dcos/dcos-ui/pull/3754#discussion_r271043098

mperrotti:
Thank you @juliangieseke and @brandonc for the commits 🙏

pierrebeitz:
@juliangieseke i think that the translation key now [looks a little odd](https://github.com/dcos/dcos-ui/pull/3754/files#diff-3efbfb8074a9a05f9fbe43a2ce435a50L742) containing a space. if i got it correctly prettier mandated this on us which does not feel right, thus the mention of prettier-ignore in the above comment.

if nobody else minds the space, we can ignore me on that matter. :)

PR #3757, opened 2019-03-27
DCOS-50715: add manual disabling of add* calculations

DanielMSchmidt:
## Testing

- Check out the branch
- Add `console.log` calls to the CompositeState add* methods to see in the browser console what happens
- Change export of composite state so you can control it in the browser console:

```diff
- module.exports = new CompositeState({ frameworks: [], slaves: [] });
+ const state = new CompositeState({ frameworks: [], slaves: [] });
+ module.exports = state
+ window.compositeState = state;
```

- Run `window.compositeState.enable()` and `window.compositeState.disable()` in the browser console until you are convinced it works

## Trade-offs

- Choose against EventEmitter, as it seemed harder to implement with an event emitter than with a simple reference counter. 

## Screenshots

ezgif com-optimize


src/js/structs/__tests__/CompositeState-test.js
```diff
@@ -70,7 +70,7 @@ describe("CompositeState", function() {
     });
 
     it("handles null data gracefully", function() {
-      expect(CompositeState.addNodeHealth).not.toThrow();
```
DanielMSchmidt: The previous call overwrote the `this`, making the method fail. I think this is the very same thing as the test wanted to test for, without using a shorthand, so I just added it.

DanielMSchmidt:
@jaxens yes, that's the plan

PR #3758, opened 2019-03-27
DCOS-50727 & DCOS-50728: Add master node region as data source & use it on services table

DanielMSchmidt:
## Testing

- test services page (ideally on multi region clusters) and see if the "(Local)" addition is visible
- see if there are the same amount or less requests to master_info going out

## Screenshots

| *Before*  | *After*  |
|---|---|
| Bildschirmfoto 2019-04-03 um 14 44 00  | Bildschirmfoto 2019-04-03 um 14 41 22   |


plugins/services/src/js/containers/services/ServicesTable.js
```diff
@@ -651,4 +657,28 @@ ServicesTable.propTypes = {
   services: PropTypes.array
 };
 
-module.exports = ServicesTable;
+// TODO: move into it's own file
+function withMasterRegionName(Component) {
+  const master$ = MesosStateStore.getMasterRequestStream();
+  const masterRegionName$ = master$.pipe(
+    map(masterRequestString => {
+      const masterRequest = JSON.parse(masterRequestString);
```
jaxesn: Not important now, would there be a way to share this parsed version of the string?  I imagine most examples that follow this pattern would want to get the object to find the specific key(s) for that component?

plugins/services/src/js/containers/services/ServicesTable.js
```diff
@@ -651,4 +657,28 @@ ServicesTable.propTypes = {
...
+      const nodeRegionName = findNestedPropertyInObject(
+        masterRequest,
+        "get_master.master_info.domain.fault_domain.region.name"
+      );
+
+      return nodeRegionName || "N/A";
```
jaxesn: Maybe its as simple as making this component pure, but should we also go ahead and establish the pattern for making sure this component only re-renders when the region.name changed and no on every stream update?

plugins/services/src/js/containers/services/ServicesTable.js
DanielMSchmidt: Yeah, I'm planning on moving the data parsing up to the nodes store. The only thing I'm a little worried about is that we now have this piece of data selection in the component. I'm thinking about time boxing a data layer effort here to an hour to have this nice encapsulated and still a good performance tradeoff

natmegs:
I played around with this a bit to try to make the `masterRequest$` stream shareable - added a commit to a different branch (https://github.com/dcos/dcos-ui/commit/8b5680ec72e3661eafee9e58ffc0d2e20bd441e3) so you can cherry-pick or take parts of or discard entirely 😄 Basically I moved the actual request observable to a constant, so it's not creating a new request observable each time `getMasterRequestStream` is called. I also removed `wait$` from the `subscribe` method as it was unclear what the purpose was (the entire subscribe setup is a bit convoluted IMO) and it was causing duplicate requests to be make every polling interval. 
Results on services page in network tab when filtering for `get_master`:
Before: https://cl.ly/bc004d020085
After: https://cl.ly/b4650a646660

plugins/services/src/js/containers/services/ServicesTable.js
```diff
@@ -651,4 +655,16 @@ ServicesTable.propTypes = {
   services: PropTypes.array
 };
 
-module.exports = ServicesTable;
+function withMasterRegionName(Component) {
```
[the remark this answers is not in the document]
DanielMSchmidt: Yeah, that was just a POC, I'll find a proper solution in a follow-up ticket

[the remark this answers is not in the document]
DanielMSchmidt: Okay, I decided to do both ticket in one go, but I have no clue on where to extract this to. It has become pretty small, so I wonder if it even makes sense to extract it in this state

mperrotti:
I saw a slight increase in polling time for `get_master` over time, but not a significant increase as @natmegs  described.

DanielMSchmidt:
Thank you for the feedback @natmegs and @mperrotti I will take a look at this strange polling behaviour

src/js/stores/MesosStateStore.js
```diff
@@ -28,12 +24,10 @@ import MesosStateUtil from "../utils/MesosStateUtil";
 import pipe from "../utils/pipe";
 import { linearBackoff } from "../utils/rxjsUtils";
 import { MesosStreamType } from "../core/MesosStream";
+import { MesosMasterRequestType } from "../core/MesosMasterRequest";
```
nLight: Shall we consider refactoring the last place in OverviewDetailTab.js to use request from container and replaying the response instead of polling?

src/js/stores/MesosStateStore.js
```diff
@@ -88,34 +82,25 @@ class MesosStateStore extends GetSetBaseStore {
...
       })
     );
 
     const parsers = pipe(...Object.values(mesosStreamParsers));
-    const dataStream = mesosStream.pipe(
-      merge(getMasterRequest),
+    const data$ = mesos$.pipe(
+      merge(masterRequest$),
       distinctUntilChanged(),
       map(message => parsers(this.getLastMesosState(), JSON.parse(message))),
```
nLight: json being parsed again

src/js/core/MesosMasterRequest.ts
```diff
@@ -0,0 +1,46 @@
...
+
+const { request } = MesosClient;
+const MAX_RETRIES = 5;
+const RETRY_DELAY = 500;
+
+export const MesosMasterRequestType = Symbol("MesosMasterRequest");
+
+export default request({ type: "GET_MASTER" }, "/mesos/api/v1?get_master").pipe(
+  retryWhen(linearBackoff(RETRY_DELAY, MAX_RETRIES)),
+  map((response: string) => JSON.parse(response)),
```
nLight: json being parsed the first time

natmegs:
I didn't realize when I tested this just now that I had fixtures turned on. Without fixtures, I see a different error relating to performing setState on an undefined state, but if I moved all this stuff https://github.com/dcos/dcos-ui/blob/b4edd7dde1dd126a8754552a88a8d17eb4906e71/src/js/pages/system/OverviewDetailTab.js#L86-L123 into `componentWillMount` instead of constructor it fixed it (unfortunately does not fix the integration test though).

DanielMSchmidt:
Thank you very much for investigating @natmegs. I think we got now almost every place that could have gotten wrong 👍

TattdCodeMonkey:
@DanielMSchmidt  Testing the latest against SOAK and I dont see `(Local)` next to the region like I do on Soak UI.

Localhost PR branch:
image
Soak:
image

DanielMSchmidt:
Thank you for the feedback @TattdCodeMonkey - Seems like I forgot to test the main use-case after so much whack a moling :D

nLight:
Legit fails?

src/js/core/MesosMasterRequest.ts
```diff
@@ -0,0 +1,45 @@
...
+import { linearBackoff } from "../utils/rxjsUtils";
+import Util from "#SRC/js/utils/Util";
+
+const { request } = MesosClient;
+const MAX_RETRIES = 5;
+const RETRY_DELAY = 500;
+
+export const MesosMasterRequestType = Symbol("MesosMasterRequest");
+
+export default request({ type: "GET_MASTER" }, "/mesos/api/v1?GET_MASTER").pipe(
```
brandonc: When is this every re-requested? If the master were to change while the UI was open, what would happen?

src/js/stores/MesosStateStore.js
```diff
@@ -327,18 +320,8 @@ if (Config.useFixtures) {
...
-    request = function(options, url) {
-      if (url === "/mesos/api/v1?get_master") {
-        console.log("Stub /mesos/api/v1?get_master");
-
-        return of(JSON.stringify(values[0]));
-      }
-
-      return oldRequest(options, request);
-    };
+    container.rebind(MesosMasterRequestType).toConstantValue(of(values[0]));
```
brandonc: 😎

PR #3760, opened 2019-03-28
DCOS-51008: Bump marathon version

nLight:
We download raml definitions to compile App definition validators for the Service Create Form

Marathon version 1.8.143-7c07b88f9

## Testing

1. Check out
2. npm i
3. Verify service form still works and validates JSON

### DIFF
--- a/scripts/pre-install
+++ b/scripts/pre-install
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 SCRIPT_PATH="$(dirname "$0")/$(dirname "$(readlink "$0")")"
-MARATHON_VERSION="1.7.49-f24b03af3"
+MARATHON_VERSION="1.8.143-7c07b88f9"
 
 # Import utils
 source "${SCRIPT_PATH}/utils/git"
--- a/scripts/utils/marathon
+++ b/scripts/utils/marathon
@@ -39,7 +39,7 @@ function install_marathon_raml() {
 
   # Copy only relevant parts in the RAML directory
   header "Installing raml files"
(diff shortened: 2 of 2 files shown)


scripts/pre-install
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 SCRIPT_PATH="$(dirname "$0")/$(dirname "$(readlink "$0")")"
-MARATHON_VERSION="1.7.49-f24b03af3"
+MARATHON_VERSION="1.8.143-7c07b88f9"
```
brandonc: I see this log when running `npm install`

```
# Installing raml files 
cp: /var/folders/68/gwtbd7vn5jz7clcv3xjv85p00000gn/T/tmp.aSg8Oeml/marathon-docs-1.8.143-7c07b88f9/docs/rest-api/public/api/v2/schema: No such file or directory
! Updated RAML in /Users/brandonc/m/fe/dcos-ui/src/resources/raml/marathon/v2 
```

I would have also assumed `src/resources/raml/marathon/v2` would have been updated after running it.

nLight:
@brandonc 😍 we indeed never used schema and it's been gone from the marathon repo.

scripts/pre-install
nLight: > I would have also assumed src/resources/raml/marathon/v2 would have been updated after running it.

only if they actually have changed anything

nLight: I at least see `.marathon-version` having desired `1.8.143-7c07b88f9` :)

PR #3761, opened 2019-03-28
check types in plugins-private

pierrebeitz:
## Testing

introduce a type error in private plugins, run `npx tsc` and see it failing.

currently you don't even need to introduce an error, as there already are some.

# ⚠️ 	DO NOT MERGE, as this will break EE


GeorgiSTodorov:
The weird thing is that this checks them even after I disabled them with `npm config delete externalplugins`. I don't think that it should.

pierrebeitz:
well i guess there's no elegant way of meeting your expectation, as we'd need a wrapper-script around the execution of TS then. :(

still i think checking the types over there to be sound in this way is still better than not checking them at all, right?

GeorgiSTodorov:
It's good to have checking, but it might be annoying trying to see errors only in dcos-ui and getting private plugins as well.

pierrebeitz:
@mperrotti 	could you check this out some time and have a look at the errors? for most (the js-imports) it should be fine to just `// @ts-ignore` for now, but there's one that could be an issue as a property of a component is missing.

pierrebeitz:
@juliangieseke 	please try to do so. i failed miserably for the reasons mentioned [here](https://github.com/dcos/dcos-ui/pull/3761#issuecomment-477929543).

pierrebeitz:
superseded by https://github.com/dcos/dcos-ui/pull/3851

PR #3764, opened 2019-03-29
DCOS-51098: Simplify Webpack- and Jest-setup

pierrebeitz:
This aims to simplify the configuration of webpack, babel and jest.

Therefore we:

* dry down the webpack config.
* remove our custom jest-preprocessor altogether, just relying on `ts-jest` and `js-jest`. 🎉 
* remove the jison file along with its loader.
* move module aliasing into babel instead of doing it in jest *and* webpack.
* use `glob` instead of `fs` to determine what test-files to run.
* remove some seemingly obsolete test-helpers in the test-env (`requestAnimationFrame`).
* remove some cruft (`.es6`-resolving) and do some cosmetics to old code.

## Testing

* i suggest reviewing commit by commit
* check if the same amount of tests is ran.
* check if the DSL-stuff that's jison-related is still working.
* anything that could break by the changes mentioned above.

⚠️ 	if the overall change seems like a lot, have a look at the separate commits as they should be factored ok-ish. if you don't think so, i'll happily factor this PR into several ones. just let me know if that would help.

Closes DCOS-51098


nLight:
We thought about extracting DSL to a library @wavesoft then we could keep the grammar there while shipping the compiled version. Current change kinda kills the source of truth, isn't it?

pierrebeitz:
@nLight 	yes, it kills the source of truth. i therefore introduced [a comment](https://github.com/dcos/dcos-ui/blob/ef2a226c295cb952a1f7481e1bd14d18354b0528/src/resources/grammar/SearchDSL.js#L4-L11) in case someone wants to modify that stuff.

maybe i should have shared my reasoning for removing right away:

* nobody touched it since 2016 (apart from cosmetic changes). please have a look at the [churn](https://github.com/dcos/dcos-ui/commits/54db46faf5b2d640c06d7897883d02bd5aeed513/src/resources/grammar/SearchDSL.jison). 
* in the HH-office there is the believe that no member of the frontend-team can maintain jison. so it seems basically worthless to have it easily modifiable - if one wants to modify it, there's instructions in the comment linked above.
* it does not seem worth having special-cases in our tooling/config for a single file that hasn't been touched and could have been commited as JS right away, as this increases our flexibility in changing stuff. (e.g. @DanielMSchmidt 	mentioned that he once wanted to try a different bundler and stumbled because of jison back then).

what do you think?

nLight:
@pierrebeitz I agree it doesn't make sense to load it at runtime, though we could keep some tooling to compile `jison`. So that it is accessible when we need it. I don't accept the argument that if nobody is able maintaining it we should just remove it. When we need modifying it people will have to make sense of jison and it would be great having less obstacles at that moment. 

Having said that would you be ok removing only the `jison-loader` part keeping the source with a comment how to compile it?

pierrebeitz:
@nLight 	like [this](https://github.com/dcos/dcos-ui/blob/14e4134190e6101fb2b794f80d7aafddfea5441b/src/resources/grammar/SearchDSL.jison#L1-L10)?

nLight:
@pierrebeitz exactly, thanks for following up! 🙇

GeorgiSTodorov:
@pierrebeitz , file conflicts.

brandonc:
Waiting to update branch until

- [ ] https://github.com/dcos/dcos-ui/pull/3764

is merged

PR #3766, opened 2019-03-29
DCOS-50724: eliminate CompositeState dep from NodeBreadcrumbs

natmegs:
Eliminate dependency on CompositeState by implementing a ~custom hook~ higher order component to pull latest summary from MesosSummaryStore.

This implementation makes use of the mesos state summary we are already polling for (globally) every 2s so that we don't have to make any extra network requests.

Some random notes I used to try to better understand the constraints [here](https://gist.github.com/natmegs/16f9fbd51a9ce6835f9c43a663ba875a).

## Testing

- Go to Nodes page
- Click on a node to go to its detail page
- Ensure breadcrumbs at top of page show Nodes > <node hostname>

## Trade-offs

- The JIRA specifies to use RxJS but this implementation does not. I found that an RxJS solution wasn't super intuitive for this particular case because the MesosSummaryStore uses an event emitter pattern rather than a stream-based request pattern. We could wrap the emitters to emit to a `Subject` and compose a stream from that (which is what I tried initially), but it involved a lot of overhead.
- When we are on the nodes page, not only do we poll for the mesos summary, but we also poll 2 different nodes endpoints every 2s (`/system/health/v1/nodes` and `/net/v1/nodes`). It seems unfortunate that we can't make one request to get the nodes information that we need 

## Dependencies

N/A

## Screenshots

N/A


plugins/nodes/src/js/components/NodeBreadcrumbs.js
```diff
@@ -19,22 +56,16 @@ const NodeBreadcrumbs = ({ nodeID, taskID, taskName, unitID }) => {
       </BreadcrumbTextContent>
     </Breadcrumb>
   ];
-  let node;
-
-  if (nodeID != null && trimmedNodeID.length > 0) {
-    node = CompositeState.getNodesList()
-      .filter({ ids: [nodeID] })
-      .last();
+  const node = useNode(nodeID);
```
brandonc:
```suggestion
  const node = node != null && trimmedNodeID.length > 0 ? useNode(nodeID) : null;
```

If you do this you can avoid the redundant trim and null condition in getNodeFromSummary, as well as whatever other overhead comes with useNode(null)

natmegs: I tried to implement this change but got an error saying I can't use a hook in a conditional 😞

brandonc: of course. I just haven't memorized the rules yet!

plugins/nodes/src/js/components/NodeBreadcrumbs.js
```diff
@@ -19,22 +19,16 @@ const NodeBreadcrumbs = ({ nodeID, taskID, taskName, unitID }) => {
       </BreadcrumbTextContent>
     </Breadcrumb>
   ];
-  let node;
+  const node = useNode(nodeID);
```
nLight: Would it make sense passing the node as a prop to the component instead?

jaxesn: I agree with @nLight. I think ideally you would limit the components who can get data from other places, like stores.  Even down below we do the same for Units.  If we could say something like, `only top level components get data from stores, everything below it get its data from props` we could be much smarter about renders and generally have a better sense of how the data flows since something as nested as Breadcrumb isn't going out to a store to get some random data.

This is more a general comment so not something we can necessarily solve in this PR, but I think it would be worth it to sit down and think about our general arch rules like this so that when writing new code, such as this PR, we can start putting in place better fundamentals.

natmegs: Yes I definitely agree that the ideal case would be top level components retrieving any data necessary and passing down to other components. What we could do in this case, is use the `useNode` hook in components that need to retrieve nodes by ID. In some cases we can already just pass the entire node to `NodeBreadcrumbs` (for example, if we moved `useNode` to `NodeDetailPage`, we can pass the entire node to `NodeBreadcrumbs`). In other cases though, we will need to implement this data fetching at a higher level (for example, `NodesTaskDetailPage` only has the nodeId from the route params, so we would need to fetch the node in this component using `useNode` and pass it to `NodeBreadcrumbs`)

natmegs:
I made a fairly major change in a second commit based on feedback. I moved node fetching out of `NodeBreadcrumbs` and up into the higher level components `NodeDetailPage`, `NodeTaskDetailPage`, and `NodesUnitHealthDetailPage`. I also had to change the node fetcher from a custom hook to a regular HOC (`useNode` -> `withNode`) so it could integrate with regular class components.

natmegs:
@GeorgiSTodorov I cannot reproduce that error (and CI was green 🤔). Maybe if you were running locally on a different branch and then switched branches without restarting the app it could have caused an issue? Anyway I rebased with latest master, let me know if it persists.

GeorgiSTodorov:
I thought the same, but couldn't get it working, I will try again tomorrow. However, I don't have the option to dismiss my review right now.

brandonc:
My dismiss comment is misleading- I meant Georgi might not have permissions to dismiss his own review

plugins/nodes/src/js/components/NodeBreadcrumbs.js
```diff
@@ -19,22 +18,15 @@ const NodeBreadcrumbs = ({ nodeID, taskID, taskName, unitID }) => {
...
+    crumbs.push(
+      <Breadcrumb key="hostname" title={node.hostname}>
+        <BreadcrumbTextContent>
+          <Link to={`/nodes/${encodedNodeID}`}>{node.hostname}</Link>
+        </BreadcrumbTextContent>
+      </Breadcrumb>
+    );
   }
 
   if (taskID != null && taskName != null) {
```
TattdCodeMonkey: should we add `node && ` to this if? if the `node === null` then `encodedNodeID = "null"` based on the code on lines 12 & 13. 

Or we could move this if block inside the `if (node)` block on line 22. Then we could also move the assignment of `trimmedNodeID` & `encodedNodeID` to the top of the same block.

src/js/stores/MesosFetchers.tsx
```diff
@@ -0,0 +1,66 @@
+import * as React from "react";
+import { MESOS_SUMMARY_CHANGE } from "#SRC/js/constants/EventTypes";
+import MesosSummaryStore from "./MesosSummaryStore";
+import Node from "../structs/Node";
+
+const getNodeFromSummary = (nodeId?: string, summary?: any) => {
+  const trimmedNodeID = nodeId
+    ? decodeURIComponent(nodeId).replace(/^\//, "")
```
GeorgiSTodorov: We have sentry bugs with decodeURIComponent breaking in certain cases. Do you think it makes sense to add some validation here?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Malformed_URI#Decoding

src/js/stores/MesosFetchers.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+export function withNode<P extends object>(
+  ComponentWithNode: React.ComponentType<P>
+) {
+  return class extends React.Component<any, WithNodeState> {
+    constructor(props: any) {
+      super(props);
+      this.receiveNewSummary = this.receiveNewSummary.bind(this);
+      this.state = {
+        summary: MesosSummaryStore.getLastSuccessfulSummarySnapshot()
```
GeorgiSTodorov: Is it possible for this to return `null` or `undefined`?

src/js/stores/MesosFetchers.tsx
```diff
@@ -0,0 +1,66 @@
+import * as React from "react";
```
DanielMSchmidt: Can we name the file so it somehow resembles the source of the data (summary store)?

plugins/nodes/src/js/components/NodeBreadcrumbs.js
```diff
@@ -5,12 +5,11 @@ import { ProductIcons } from "@dcos/ui-kit/dist/packages/icons/dist/product-icon
 
 import Breadcrumb from "#SRC/js/components/Breadcrumb";
 import BreadcrumbTextContent from "#SRC/js/components/BreadcrumbTextContent";
-import CompositeState from "#SRC/js/structs/CompositeState";
 import PageHeaderBreadcrumbs from "#SRC/js/components/PageHeaderBreadcrumbs";
 import UnitHealthStore from "#SRC/js/stores/UnitHealthStore";
 
-const NodeBreadcrumbs = ({ nodeID, taskID, taskName, unitID }) => {
-  const trimmedNodeID = decodeURIComponent(nodeID).replace(/^\//, "");
+const NodeBreadcrumbs = ({ node, taskID, taskName, unitID }) => {
```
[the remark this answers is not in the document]
natmegs: I converted to TS but then the NodeBreadcrumbs unit tests all failed pretty dramatically... `macro_1 not found`... sunk a bunch of time into trying to resolve but ended up going back to JS and adding propTypes

plugins/nodes/src/js/components/NodeBreadcrumbs.js
```diff
@@ -1,82 +1,81 @@
...
+    }
   }
 
   return (
     <PageHeaderBreadcrumbs iconID={ProductIcons.Servers} breadcrumbs={crumbs} />
   );
 };
 
+NodeBreadcrumbs.propTypes = {
+  node: PropTypes.object,
```
brandonc: nitpick: I've seen other classes declare `PropTypes.instanceOf(Node)`

src/js/stores/MesosSummaryFetchers.tsx
```diff
@@ -0,0 +1,66 @@
+import * as React from "react";
+import { MESOS_SUMMARY_CHANGE } from "#SRC/js/constants/EventTypes";
+import MesosSummaryStore from "./MesosSummaryStore";
+import Node from "../structs/Node";
+
+const getNodeFromSummary = (nodeId?: string, summary?: any) => {
+  const trimmedNodeID = nodeId
```
brandonc: If you pass in node in stead of nodeId here, you can reuse getEncodedId and getId from there and drop all this trimming stuff

plugins/nodes/src/js/components/NodeBreadcrumbs.js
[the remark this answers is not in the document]
GeorgiSTodorov: `macro_1` is related to lingui. I think there was a way to fix it, but I don't remember what it was.

DanielMSchmidt:
I'm merging this to unblock future efforts, I think these changes are small enough so that we can tackle them when we touch anything else on the nodes page

PR #3767, opened 2019-04-01
chore: minor cleanup after system tests update

juliangieseke:
## Description

* remove unused env vars from Jenkinsfile
* reduce timeout to 3h
* cleanup stages (setup repo, build, setup datadog)
* integrate `test:validate` in `pretest`
* rename bash functions, remove `*_integration` suffix (was added because i didnt expect us being able to use the same funcs for integration & system
* fix `git rebase` line so that it doesnt test master if rebasing fails 🙈 

## Testing

🤷‍♂️


nLight:
wasted 1sp :)

PR #3769, opened 2019-04-02
DCOS-49871 Support service tabs from plugins

TattdCodeMonkey:
Added the ability for a plugin to add a tab to the create service
modals, single and multi-container. This was done using
Hooks.applyFilter so that plugins could add additional navigation items
and TabView items to the CreateServiceModalForm component.

## Testing

Ensure tabs and navigation doesn't break. The Hook filters created here will be used by plugins and nothing else to be tested here.

### DIFF
--- a/plugins/services/src/js/components/modals/CreateServiceModalForm.js
+++ b/plugins/services/src/js/components/modals/CreateServiceModalForm.js
@@ -3,6 +3,7 @@ import { i18nMark, withI18n } from "@lingui/react";
 import classNames from "classnames";
 import isEqual from "lodash.isequal";
 import { MountService } from "foundation-ui";
+import { Hooks } from "PluginSDK";
 import PropTypes from "prop-types";
 import React, { Component, Suspense, lazy } from "react";
 
@@ -473,7 +474,7 @@ class CreateServiceModalForm extends Component {
         ? "Service"
         : "Services";
 
-    const tabList = [
+    let tabList = [
       {
         id: "services",
         label: serviceLabel,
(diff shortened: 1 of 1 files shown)


plugins/services/src/js/components/modals/CreateServiceModalForm.js
```diff
@@ -628,10 +652,16 @@ class CreateServiceModalForm extends Component {
             onRemoveItem={this.handleRemoveItem}
             onAddItem={this.handleAddItem}
           />
-        </TabView>
+        </TabView>,
+        ...pluginTabViews
```
nLight: Could we leverage `<Mount />` component here? And then just mount additional tabs like so https://github.com/mesosphere/dcos-ui-plugins-private/blob/master/secrets/hooks.js#L64 ?

GeorgiSTodorov:
Is this ready for reviews? It's marked as "in progress" in jira.

plugins/services/src/js/components/modals/CreateServiceModalForm.js
TattdCodeMonkey: Given how TabViewList works I don’t think Mount will work here. I was going to try that initially but opted for this instead.

TattdCodeMonkey:
@GeorgiSTodorov this PR is the matching PR for the ticket for plugins private should be up later today.

natmegs:
I am seeing all the tabs duplicated (EE - single and multicontainer) 
Screen Shot 2019-04-03 at 3 25 37 PM

TattdCodeMonkey:
@natmegs 🤔 let me try and reproduce that and adjust accordingly.

TattdCodeMonkey:
@natmegs good catch, I was using `Hooks.applyFilter` incorrectly.

PR #3771, opened 2019-04-03
DCOS-50993: Don't merge MesosStream into CompositeState

nLight:
With this PR we stop merging MesosStream into the CompositeState as we seem not to be using the data.

The only piece of `MesosStateStore` we're using is `master_info` therefore I moved the `addState` from the stream to the `GET_MASTER` request.

I decided to rename `addSummary` to `addState` as the functions were doing the same.  

Closes DCOS-50993

## Testing

Please verify there's no regressions in places described in the ticket
https://jira.mesosphere.com/browse/DCOS-50993

## Trade-offs

1. I assumed the performance gain will be high, but it's marginal. We stop merging `MesosStream` but still merge Summary with _itself_. Otherwise we would lose the nodes health. There's a ticket to address that.
2. I've fixed fixtures just to enable Nodes table but they're incomplete. I.e. you won't see region and zone info in the nodes table running off fixtures.


nLight:
Fixing integration tests...

TattdCodeMonkey:
I re-ran CI as it was failed again, just double-checking if its a flakey issue or not.

I ran this locally against the daily cluster and wasn't not able to find anything broken. Will check back in when CI completes.

TattdCodeMonkey:
Integration test failing again.

tests/_fixtures/v1/get_master.json
```diff
@@ -6,10 +6,10 @@
             "ip": 1694826506,
             "port": 5050,
             "pid": "master@10.0.5.101:5050",
-            "hostname": "10.0.5.101",
+            "hostname": "dcos-0",
```
natmegs: I hope it's ok that I pushed a change to fix an failing integration test assertion to match these changes 🙏

nLight: Thank you!

PR #3772, opened 2019-04-03
DCOS-50840: Debug tab intermittently crashes on soak cluster

GeorgiSTodorov:
Handle the case of having an undefined
node id in the declined offers table and breaking the UI.

Closes https://jira.mesosphere.com/browse/DCOS-50840

## Testing
1. Go to services tab
2. Create a service with very high resource requirements
Example JSON:
```
{
  "id": "/new-service",
  "instances": 1,
  "portDefinitions": [],
  "container": {
    "type": "MESOS",
    "volumes": []
  },
  "cpus": 0.1,
  "mem": 1000000000000,
  "requirePorts": false,
  "networks": [],
  "healthChecks": [],
  "fetch": [],
  "constraints": [],
  "cmd": "sleep 1000000"
}
```
3. Run the service and go to it's debug tab
4. In `dcos-ui/plugins/services/src/js/components/DeclinedOffersTable.js` replace line 209
Currently:
```
          const node = MesosStateStore.getNodeFromHostname(hostname);
```
Replace with:
```
          const node = {id: null};
```
5. Verify that the UI doesn't break
6. Verify that the declined offers table at the bottom isn't shown
7. Test the same for pods

## Trade-offs
~1. The jira says that we should add tests for this scenario, but neither `DeclinedOffersTable`, nor `PodDebugContainer` and `ServiceDebugContainer` have defined test files and for me, creating complete new test files is out of scope for a 2-story point bug. I think we can create the tests as a follow-up.~
~2. Hiding the entire table because we can't render one column might seem over the top, but for me hiding just the Host column doesn't look good and isn't user friendly.~

## Dependencies
None.

### Before
Screenshot from 2019-04-03 15-26-13

### After
(no link)
Screenshot from 2019-04-05 13-29-12


natmegs:
I am OK with this solution, but I wonder if we could avoid hiding the entire DeclinedOffersTable by displaying "N/A" or something like that in the table if the node is null in the host column render method (https://github.com/dcos/dcos-ui/blob/b803824a05b81c2a503b899e8a7ccbeb283ff729/plugins/services/src/js/components/DeclinedOffersTable.js#L209)?

GeorgiSTodorov:
@natmegs , this is pretty close to my initial way of solving it (which was just hiding this column), but I think that showing details for unknown nodes isn't optimal. Although if the designers disagree, I will gladly change it.
Screenshot from 2019-04-04 11-55-17
@mperrotti @izharbuendia 
(I don't know Daniel Schwartz's github profile)

plugins/services/src/js/containers/service-debug/ServiceDebugContainer.js
```diff
@@ -48,11 +50,20 @@ class ServiceDebugContainer extends React.Component {
 
     const queue = service.getQueue();
 
+    const offers = queue.declinedOffers.offers;
+    for (let i = 0; i < offers.length; i++) {
+      const hostname = offers[i].hostname;
+      const node = MesosStateStore.getNodeFromHostname(hostname);
```
juliangieseke: Did you try adding this to `service` struct where it gets created? I feel like that would be a better place then "somewhere down the line" 🤔

nLight: Or actually `DeclinedOffersTable` around line 210, so that we don't render the hostname?

nLight:
We shall _always_ consider tests estimating tickets. Both bugs and features. Please add tests even if it feels like a lot of work.

GeorgiSTodorov:
@nLight , I am not sure I understand the changes you request. If we do the same check in `DeclinedOffersTable`, then if `node.id` is null, we show `Details` heading, but nothing underneath.

nLight:
@GeorgiSTodorov as I can see we use `node` only to render `hostname` am I missing something?

GeorgiSTodorov:
@nLight , yes. What is your suggestion?

nLight:
@GeorgiSTodorov apparently the node went missing from the cluster. We could say something like `N/A` or `Node is not available`

GeorgiSTodorov:
@nLight , that was @natmegs 's suggestion as well. Here is my response: https://github.com/dcos/dcos-ui/pull/3772#issuecomment-479811955 .

nLight:
@GeorgiSTodorov thanks for kindly pointing me to the comment. I'd rather have this version so that we show at least what we can show. Let's wait for @izharbuendia @mperrotti please. Product input could be valuable here @AmrAbdelrazik https://github.com/dcos/dcos-ui/pull/3772#issuecomment-479811955

GeorgiSTodorov:
I also said something similar in the trade-offs section. Let's see what design and product will say, I will accept their decision.

jaxesn:
I'm not sure this decision is so mission critical that we need to run every detail by design.  I want to roll with something here.  @mperrotti please make a call, if you think its big enough to get design to weigh in, please make that happen.  Having them follow along in a random PR, with little to no context, is not going to be super productive.

Unless this PR has a downstream dependency where we really want this merged, please go ahead and spend some time writing tests.  I understand there are no tests for this particular area, but we have to start somewhere.  If you need some help getting it started, since I know starting a fresh test file is always a challenge, please grab someone to pair to get you rolling.

GeorgiSTodorov:
@nLight , "We shall always consider tests estimating tickets." - is everyone on the team aware of this?

mperrotti:
I prefer the solution @natmegs suggested - rendering "N/A" in place of the host text. I agree with @jaxesn - we don't need to run this particular detail by design

plugins/services/src/js/containers/pod-debug/PodDebugContainer.js
```diff
@@ -47,11 +48,20 @@ class PodDebugTabView extends React.Component {
       return null;
     }
 
+    const offers = queue.declinedOffers.offers;
+    for (let i = 0; i < offers.length; i++) {
```
mperrotti: Why do we have to use an old-school `for` loop instead of a `.map()`?

nLight:
@GeorgiSTodorov if we have still have `hostname` let's please render it. As I understand it's only `node` we can't resolve to render a link?

GeorgiSTodorov:
I pushed the new proposed solution, now I am working on tests.

nLight:
@GeorgiSTodorov even a jest snapshot test would do

GeorgiSTodorov:
@nLight , I already started working on an integration tests.

GeorgiSTodorov:
I added tests.

plugins/services/src/js/components/DeclinedOffersTable.js
```diff
@@ -210,12 +210,16 @@ class DeclinedOffersTable extends React.Component {
...
-              </Link>
+              {node && node.id ? (
+                <Link
+                  className="table-cell-link-primary"
+                  to={`/nodes/${node.id}`}
+                >
+                  {hostname}
+                </Link>
+              ) : (
+                hostname
```
mperrotti: Can we wrap this in `<span className="table-cell-link-primary" />`? That way it looks the same as when it has an id and gets linked.

If it was intentional to make it look different, I think we need to make the difference more noticeable than just changing the font weight - maybe lighten the text.

GeorgiSTodorov:
Implemented @mperrotti 's suggestion.

PR #3775, opened 2019-04-03
DCOS_OSS-4954: allow gpus with ucr or command only

natmegs:
~*Hold merge until [blocker ticket](https://jira.mesosphere.com/browse/DCOS_OSS-4934) is resolved (currently integrating)~

Allow users to enter gpus for a job that uses UCR explicitly or no container (command only).

Closes DCOS_OSS-4954

## Testing

*This will error until the blocker ticket is merged

Command only job
- Go to jobs form
- Fill out only job ID and command (do not choose container image)
- **GPU form field should not be disabled**
- Fill out number for GPU
- Submit job (there should be no errors)

Job with explicit UCR container
- Go to jobs form
- Fill out job ID
- Select Container Image
- Fill out container image input
- **GPU form field should not be disabled**
- Fill out number for GPU
- Submit job (there should be no errors)

Job with Docker container
- Go to jobs form
- Fill out job ID
- Select Container Image
- Fill out container image input
- Go to Container tab, select Docker
- Go back to General tab
- **GPU form field should be disabled**

## Trade-offs

N/A

## Dependencies

[RESOLVED] Depends on [blocker ticket](https://jira.mesosphere.com/browse/DCOS_OSS-4934)


GeorgiSTodorov:
Not a blocker, but if I enter no value for GPU, it's 0 next time. I think we should have 0 as the default.

PR #3777, opened 2019-04-04
DCOS-42386: downscale mobile viewport

nLight:
To mitigate Android closing the keyboard because of Modal being redrawn.

When there's not enough space for a modal on the screen it tries adjusting its dimensions causing a repaint. For some reason repaint closes Android keyboard.

So the solution is to scale down the UI so that the modal fits :)

Closes DCOS-42386

## Testing

1. Install Android Studio :)
2. Open Android emulator
3. Navigate to you local UI
4. Verify that you can focus login and password inputs without the keyboard being closed

## Trade-offs

The proper way to fix it would be fixing https://github.com/mesosphere/reactjs-components/blob/master/src/Modal/ModalContents.js#L112 so that id doesn't repaint the screen on Android devices. But that could have side effects on the Desktop version.

## Screenshots

Working version:

Screenshot 2019-04-04 at 13 23 07

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -9,8 +9,8 @@
   <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1">
   <meta name="title" content="Mesosphere DC/OS">
   <meta name="HandheldFriendly" content="True">
-  <meta name="MobileOptimized" content="320">
-  <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+  <meta name="MobileOptimized" content="640">
+  <meta name="viewport" content="width=640, initial-scale=0.5" />
   <link rel="icon" href="./img/components/icons/favicon.ico">
   <!--[if IE]>
       <link rel="icon shortcut" href="./img/components/icons/favicon.ico">


GeorgiSTodorov:
An alternative for testing is https://developers.google.com/web/tools/chrome-devtools/remote-debugging/ .

mperrotti:
I know we don't have very many mobile users, but this makes the app a lot harder to use on a mobile device. Can we explore other solutions?

nLight:
@mperrotti at this point I would rather at least enable people, then we can think of a UX feature/improvement as all the alternatives we've explored so far lead down the rabbit hole.

mperrotti:
Ok, I'm good with that. Is there already another ticket created for finding a different solution?

nLight:
@mperrotti we don't. I'd argue that we should have a capability @AmrAbdelrazik for that. Our product has never actually been 100% mobile friendly. We should revisit what makes sense and what areas we need to tackle. Not having hover-over easter eggs as suggested by Amr would be one of them.

Would you be ok going forward with the PR and then taking over the mobile UX topic?

PR #3778, opened 2019-04-04
Docs for integration and system tests

DanielMSchmidt:
## Testing

- try the commands mentioned in the contributing doc :)


package.json
```diff
@@ -213,11 +213,13 @@
     "release": "semantic-release",
     "start:testing": "NODE_ENV=testing webpack-dev-server --config ./webpack/webpack.dev.js --progress --colors --content-base ./build --host 0.0.0.0 --port $npm_package_config_port",
     "start": "lingui compile && NODE_ENV=development webpack-dev-server --config ./webpack/webpack.dev.js --hot --progress --colors --content-base ./dist --host 0.0.0.0 --port $npm_package_config_port",
-    "test:integration:local": "npx cypress open",
     "test:integration:plugins:setup": "[ -z \"${npm_config_externalplugins}\" ] || rsync -aH ${npm_config_externalplugins}/tests/ ./tests/",
+    "test:integration:local": "PLUGINS_PATH=\"${npm_config_externalplugins}\" ./scripts/run-integration-tests -hnk",
     "test:integration:rerun": "PLUGINS_PATH=\"${npm_config_externalplugins}\" ./scripts/ci/rerun-integration-tests.sh",
     "test:integration": "PLUGINS_PATH=\"${npm_config_externalplugins}\" ./scripts/run-integration-tests",
+    "test:system:plugins:setup": "[ -z \"${npm_config_externalplugins}\" ] || rsync -aH ${npm_config_externalplugins}/system-tests/ ./system-tests/",
```
juliangieseke: I dont think we need this, `test:integration` will do it.

scripts/run-integration-tests
```diff
@@ -42,33 +41,36 @@ CLEANUP_TESTS=${CLEANUP_TESTS:-0}
...
+    echo -e "+-------------------------------+\n"
+    echo -e "Don't forget running npm run build:assets first!\n\n"
+    echo -e "Usage: \n"
+    echo -e "\tnpm run test:integration -- [-s <FILE_NAME>] [-hnk] [-r <RETRIES>]\n"
+    echo -e "Options:\n"
+    printf "  %-17s %-30s \n" "-s <TestFileName>" "Run single Test File. All files if omitted"
+    printf "  %-17s %-30s \n" "-h" "Run Chrome in Headed mode"
+    printf "  %-17s %-30s \n" "-n" "Don't close Chrome after tests are finished"
+    printf "  %-17s %-30s \n" "-r <NUM_RETRIES>" "Number of retries, 3 by default"
+    printf "  %-17s %-30s \n" "-k" "Don't run the build, just assume something runs on the right port"
```
juliangieseke: what about `-d` which does `npm start` in the background? 🤔

scripts/run-system-tests
```diff
@@ -49,24 +49,26 @@ TESTS_FOLDER="system-tests"
...
     echo -e "Don't forget running npm run build:assets first!\n\n"
     echo -e "Usage: \n"
-    echo -e "\tnpm run test:integration -- [-s <FILE_NAME>] [-hn] [-r <RETRIES>]\n"
+    echo -e "\tnpm run test:system -- [-s <FILE_NAME>] [-hnk] [-r <RETRIES>]\n"
     echo -e "Options:\n"
     printf "  %-17s %-30s \n" "-s <TestFileName>" "Run single Test File. All files if omitted"
     printf "  %-17s %-30s \n" "-h" "Run Chrome in Headed mode"
     printf "  %-17s %-30s \n" "-n" "Don't close Chrome after tests are finished"
     printf "  %-17s %-30s \n" "-r <NUM_RETRIES>" "Number of retries, 3 by default"
+    printf "  %-17s %-30s \n" "-k" "Don't run the build, just assume something runs on the right port"
```
juliangieseke: same as above

scripts/run-system-tests
```diff
@@ -49,24 +49,26 @@ TESTS_FOLDER="system-tests"
...
+  k) NO_BUILD="true" ;;
   r) RETRIES=${OPTARG} ;;
   ?)
     echo -e "+-------------------------------+"
     echo -e "| DC/OS Integration test runner |"
     echo -e "+-------------------------------+\n"
     echo -e "Don't forget running npm run build:assets first!\n\n"
     echo -e "Usage: \n"
-    echo -e "\tnpm run test:integration -- [-s <FILE_NAME>] [-hn] [-r <RETRIES>]\n"
+    echo -e "\tnpm run test:system -- [-s <FILE_NAME>] [-hnk] [-r <RETRIES>]\n"
```
juliangieseke: 🙈

GeorgiSTodorov:
Please add an instruction on how to test a single file. I think this is the way most of us will be using these.

package.json
DanielMSchmidt: Not if you run it with the no build option

scripts/run-integration-tests
DanielMSchmidt: I think there could be different usecases, but I would also be fine with that. How would you handle the output?

GeorgiSTodorov:
When I run `npm run test:integration:local`, only one test is being run (UISettings-cy). This is not intentional, right?

DanielMSchmidt:
@GeorgiSTodorov I think it kinda expects a `-- -s` as it's thought for a single file usage. I wonder if we shall extract this into some bash script 🤔

CONTRIBUTING.md
```diff
@@ -390,36 +390,29 @@ it will run, the user browser.
...
 
-1.1 Without Plugins
+#### Without Plugins
 
 Run DC/OS UI in testing mode (you have to close `npm start`).
 
 ```sh
 npm config delete externalplugins
-npm run start:testing
+npm start
```
nLight: why start why not testing?

DanielMSchmidt: I checked them, one is for cypress saving fixture files causing reloading. This is not the case anymore, so I think we can delete that. The other one is for loading the Config.test instead of the Config.dev and I would argue that can go to the trash as well, this might simplify our setup.

GeorgiSTodorov:
Integration tests work.
Why do we need ` Before you start, please make sure you are configured against a new and empty cluster (See `webpack/proxy.dev.js`).`?

DanielMSchmidt:
Because the system tests mutate the cluster (schedule new things) and assume that no one else schedules things (matching for lengths in lists).

If you are configured against soak for example I hope this hint will remind you to abort the script

package.json
```diff
@@ -214,11 +214,13 @@
...
     "start": "lingui compile && NODE_ENV=development webpack-dev-server --config ./webpack/webpack.dev.js --hot --progress --colors --content-base ./dist --host 0.0.0.0 --port $npm_package_config_port",
-    "test:integration:local": "npx cypress open",
     "test:integration:plugins:setup": "[ -z \"${npm_config_externalplugins}\" ] || rsync -aH ${npm_config_externalplugins}/tests/ ./tests/",
+    "test:integration:local": "PLUGINS_PATH=\"${npm_config_externalplugins}\" ./scripts/run-integration-tests -hd",
     "test:integration:rerun": "PLUGINS_PATH=\"${npm_config_externalplugins}\" ./scripts/ci/rerun-integration-tests.sh",
     "test:integration": "PLUGINS_PATH=\"${npm_config_externalplugins}\" ./scripts/run-integration-tests",
+    "test:system:plugins:setup": "[ -z \"${npm_config_externalplugins}\" ] || rsync -aH ${npm_config_externalplugins}/system-tests/ ./system-tests/",
     "test:system": "PLUGINS_PATH=\"${npm_config_externalplugins}\" ./scripts/run-system-tests",
+    "test:system:local": "PLUGINS_PATH=\"${npm_config_externalplugins}\" ./scripts/run-system-tests -hnk",
```
juliangieseke: I'm not really a fan of these `:local` scripts… 🤷‍♂️

scripts/run-system-tests
```diff
@@ -54,12 +54,12 @@ while getopts 's:hknr:?' flag; do
...
   ?)
-    echo -e "+-------------------------------+"
-    echo -e "| DC/OS Integration test runner |"
-    echo -e "+-------------------------------+\n"
+    echo -e "+--------------------------+"
+    echo -e "| DC/OS System test runner |"
+    echo -e "+--------------------------+\n"
     echo -e "Don't forget running npm run build:assets first!\n\n"
     echo -e "Usage: \n"
     echo -e "\tnpm run test:system -- [-s <FILE_NAME>] [-hnk] [-r <RETRIES>]\n"
```
juliangieseke: still outdated `-k` and missing `-d` (same for integration tests)

scripts/utils/tests-functions
```diff
@@ -152,37 +152,46 @@ function setup() {
...
+  if [ -n "${DEVELOPMENT}" ]; then
+    if [ -n "$(lsof -i :"${PROXY_PORT}")" ]; then
+      echo "==> Something serves on ${PROXY_PORT}, we take that as the proxy"
+    else
+      ## Start Development server
+      echo "==> Starting development server using npm start"
+      npm start &
+      PROXY_PID=$!
+      echo "==> Waiting for npm start to finish, sleeping for 20s"
+      sleep 20
```
juliangieseke: :fry: we use curl to ping marathon in system tests, we should do the same approach here IMHO

DanielMSchmidt: Good idea 👍

scripts/run-system-tests
DanielMSchmidt: Will do 👍

DanielMSchmidt:
@juliangieseke @GeorgiSTodorov Could you take a look again?

PR #3779, opened 2019-04-04
DCOS-51221: Delete Repositores button not working for Package Repository

GeorgiSTodorov:
Change data layer command so that deleting repositories works.

Closes https://jira.mesosphere.com/browse/DCOS-51221

~This is still work in progress. I still need to figure out how our integration tests work right now and add one for deleting.~

## Testing
1. Go to Settings > Package Repositories
2. Add a new repository, for example:
```
name: universe 2
url: https://downloads.mesosphere.com/universe/repo/repo-up-to-1.13.json.gz
priority: 0
```
3. Try to delete it
4. Verify that it gets successfully deleted

## Trade-offs
None.

## Dependencies
None.

### Before
Peek 2019-04-04 15-10

### After
Peek 2019-04-04 15-12

### DIFF
--- a/plugins/catalog/src/js/repositories/RepositoriesDelete.js
+++ b/plugins/catalog/src/js/repositories/RepositoriesDelete.js
@@ -42,7 +42,7 @@ const removePackageRepository = gql`
 `;
 
 const removePackageRepositoryGraphql = (name, uri) =>
-  dataLayer.get(removePackageRepository, {
+  dataLayer.query(removePackageRepository, {
     name,
     uri
   });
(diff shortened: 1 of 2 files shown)


GeorgiSTodorov:
Okay, I managed to get integration tests running. However, I think that the cluster we use for tests is not setup to allow deleting packages (and adding as well).

GeorgiSTodorov:
I guess a way to test is to check if the modal has closed.

GeorgiSTodorov:
I added a test that checks if the modal closes when the delete button is clicked. As the modal won't close if the delete action fails (as in this bug), I think this is enough to test this functionality.

PR #3781, opened 2019-04-04
DCOS-51000 - Group aggregate status value

TattdCodeMonkey:
Introduced the `ServiceTreeStatusSummary` response type. 
This summary object can be used to render a status
that can also be translated using the id & value key values. It also
contains the highest priority statuses display name and a count of
statuses grouped by the StatusIcon.name. The highest priority status
count is used for the displayed status values, but all are returned so
they can be used to generate the desired tooltip content as well.

- Closes DCOS-51000
- Closes DCOS-50999

## Testing

1. Navigate to the Service page
2. Create a group
3. Inside your group create 2+ services / pods / groups
4. Navigate back to Services overview
5 Verify status is an aggregate of child service status with the highest priority showing.

## Trade-offs

- I introduced a `name` key to the `StatusIcon` type which required some small refactoring. I used this when generating the aggregate status counts for the `ServiceTreeStatusSummary`.
- I don't love the naming of the keys on `ServiceTreeStatusSummary`, open to better suggestions here. 

## Dependencies

This is based on #3748 and should be merged after it, I have the base set to the branch for now so the diff is reviewable. But that mean CI is broken until #3748 is merged to `master` and this PRs base can be updated.

## Screenshots

image
image


plugins/services/src/js/columns/ServicesTableStatusColumn.tsx
```diff
@@ -19,8 +19,28 @@ const StatusMapping: any = {
 export function statusRenderer(
   service: Service | Pod | ServiceTree
 ): React.ReactNode {
-  const serviceStatusText: string = service.getStatus();
-  const serviceStatusClassSet: string = StatusMapping[serviceStatusText] || "";
+  const serviceStatus = service.getStatus();
+  let serviceStatusClassSet: string = "";
+  let statusContent: JSX.Element | null = null;
+  if (serviceStatus.type === "servicetreestatussummary") {
```
natmegs: I was trying to log what's happening here and it looks like this condition is never met (even with a group in the services table). It always goes to the `else if` fallback, and as a result the entire object is set as the Trans id. This (somehow?) figures out how to be rendered correctly, even though `values` are nested inside of `id`. I'm not certain this would work in a prod build. I tried changing the condition to `if(typeof serviceStatus === "object")` and found that this condition allowed the id and values to be set as expected. Not sure if I'm explaining this right.
Inspecting the status with loop as is (id is an object):
Screen Shot 2019-04-04 at 4 08 04 PM
Inspecting the status with typeof check:
Screen Shot 2019-04-04 at 4 08 53 PM

TattdCodeMonkey: Weird I wonder why I didn’t see that. 

I’m fine with the object check here instead. I actually had that originally but then typescript complained about accessing `.id` etc. So I changed to this to resolve that. I wonder if it would be better to make ServiceTreeStatusSummary a class instead of an interface, then I can do an `instanceOf` check instead.

TattdCodeMonkey: Fixed, weird that it worked at all, but I guess `lingui/Trans` has guards for receiving an object with `id` & `values` as `id`

GeorgiSTodorov:
I created an empty group and this is what I got. Doesn't seem intended to me.
Peek 2019-04-05 12-28

plugins/services/src/js/structs/ServiceTree.js
```diff
@@ -357,8 +386,50 @@ module.exports = class ServiceTree extends Tree {
...
+      statusCounts: statusIconCounts.status,
+      values: {
+        priorityStatusCount: statusIconCounts.status[status.icon.name],
+        totalCount: statusIconCounts.total
+      }
+    };
+    switch (status.icon) {
+      case StatusIcon.SUCCESS:
+        statusSummary.id = i18nMark(
+          "Running ({priorityStatusCount} of {totalCount})"
```
GeorgiSTodorov: It's always `"Something ({priorityStatusCount} of {totalCount})"`, can we add `({priorityStatusCount} of {totalCount})` after the switch?

GeorgiSTodorov: This would also reduce the number of similar messages in the message catalog.

plugins/services/src/js/columns/ServicesTableStatusColumn.tsx
```diff
@@ -1,15 +1,16 @@
 import * as React from "react";
 import { Trans } from "@lingui/macro";
 import { TextCell } from "@dcos/ui-kit";
+import classnames from "classnames";
 
 import StringUtil from "#SRC/js/utils/StringUtil";
-import ServiceStatusIcon from "../components/ServiceStatusIcon";
+import { SortDirection } from "#PLUGINS/services/src/js/types/SortDirection";
+import ServiceStatusProgressBar from "#PLUGINS/services/src/js/components/ServiceStatusProgressBar";
```
TattdCodeMonkey: I don't know why but webpack would intermittently fail to find this module and not compile when this was a relative path vs using an absolute path with `#PLUGINS`, so that is why this is changed.

plugins/services/src/js/structs/ServiceTree.js
TattdCodeMonkey: Maybe? I was unsure which was better for translations, the whole statement or the status and `({priorityStatusCount} of {totalCount})` separately.

plugins/services/src/js/columns/ServicesTableStatusColumn.tsx
GeorgiSTodorov: For me, ` import ServiceStatusProgressBar from "../components/ServiceStatusProgressBar";` works.

GeorgiSTodorov:
Is "Recovering" the same as "Processing"?
Screenshot from 2019-04-08 12-12-31

GeorgiSTodorov:
If we have a nested folder, then the services tab will behave as if the services inside the second folder are in the first folder. This is working as intended for me, but I am wondering if the designers have considered this situation and what do they think about it.
Peek 2019-04-08 12-20

GeorgiSTodorov:
Also, I managed to get this error:
Screenshot from 2019-04-08 12-23-36

pierrebeitz:
@GeorgiSTodorov 
1. yes, "Recovering" is in the category "Processing".
2. do you have any idea how to reproduce the error above? (it's happening while traversing all the tasks statuses and there seems to be a status that does not have an icon associated).

GeorgiSTodorov:
@pierrebeitz , I am not sure how this error appeared, I left my window open for a few minutes and it happened.

plugins/services/src/js/structs/ServiceTree.js
```diff
@@ -166,6 +167,32 @@ module.exports = class ServiceTree extends Tree {
...
+
+    return this.reduceItems(function(counts, item) {
+      if (item instanceof Service) {
+        const itemStatus = item.getServiceStatus();
+        if (itemStatus === null) {
+          return counts;
+        }
+        counts.total++;
+
+        const iconName = itemStatus.icon.name;
```
pierrebeitz: as @GeorgiSTodorov found out: `itemStatus.icon` can be null (e.g. for Services with the status "N/A").
we should add a safeguard here.

GeorgiSTodorov:
> "Recovering" is in the category "Processing"
@pierrebeitz , will this be reflected in the documentation so that users are not confused?

plugins/services/src/js/columns/ServicesTableStatusColumn.tsx
```diff
@@ -3,51 +3,82 @@ import { Trans } from "@lingui/macro";
...
+    | ServiceTreeStatusSummary
+    | null = service.getStatus();
+  let statusContent: JSX.Element | null = null;
+  let tooltipContent: JSX.Element;
+  if (typeof serviceStatus === "object" && serviceStatus !== null) {
+    const statusText = ServiceStatus.toCategoryLabel(serviceStatus.status);
+    statusContent = (
+      <span className="status-bar-text">
+        <Trans id={statusText} />{" "}
+        <Trans id={serviceStatus.countsText} values={serviceStatus.values} />
```
brandonc: I'm not sure this `values` prop is meaningful. What is it?

TattdCodeMonkey: The values for translation of “(x of y)”

TattdCodeMonkey: Now that that’s separate we could do that directly here instead of with an id. I didn’t realize that when refactoring the status out.

In that case values won’t be needed here.

TattdCodeMonkey:
@brandonc I’ll take a look and address issues. I tested against soak last week when implementing this but not with the latest changes and refactoring this morning.

GeorgiSTodorov:
I tested this and didn't notice any visual bugs.

pierrebeitz:
@TattdCodeMonkey i 
* rebased to latest master
* made sure we will earn the benefits of exhaustiveness checking, see commit `make use of exhaustiveness checking`
* fixed a commit-lint
* removed ProgressBars for ServiceTreeNodes
* split apart the rendering for serviceTrees and services, which feels more declarative, see commit `make a difference between Service and ServiceTree`

as always: feel free to throw away or modify stuff at will.

looks good to me, so i will approve once you verified and approved my changes.

plugins/services/src/js/constants/ServiceStatus.ts
```diff
@@ -163,10 +163,7 @@ function showProgressBar(status: Status): boolean {
   return [DEPLOYING, WAITING, DELAYED, RECOVERING, DELETING].includes(status);
 }
 
-function toIcon(status: Status | null): StatusIcon | null {
-  if (status === null) {
```
pierrebeitz: @TattdCodeMonkey i removed that safeguard here, as the types say that we should be good. if you had a crash because of this, please reintroduce it!

brandonc:
@DanielMSchmidt I think this https://github.com/dcos/dcos-ui/pull/3781/files#diff-484e62fd5c4ca1e990e8d33bac21335c addresses your feedback so I'm going to dismiss your request for revisions.

GeorgiSTodorov:
I approved, because I didn't have the option to dismiss my review. I didn't actually test the latest changes.

PR #3784, opened 2019-04-05
CI: speedup CI by creating system test cluster at start

DanielMSchmidt:
## Testing

- [ ] See that it deletes the cluster if system test fail
- [ ] See that it deletes the cluster if system test run successfully
- [ ] See that it deletes the cluster if there is an early failure

## Trade-offs

- We always spawn and destroy a cluster, even if we have an early error. This additional work and cost should be compensated by ~10 minutes less runtime


pierrebeitz:
that's a nice idea! 👍 	
this would help the system-test to finish ~7 minutes (today) earlier. currently those are not the bottleneck as integration tests take longer to finish. closing until that seems helpful again!

PR #3785, opened 2019-04-05
perf: don't wait for first interval to fetch mesos summary

brandonc:
## Overview

While trying to isolate a performance issue, I slowed down the mesos summary interval to 20s but noticed that the page didn't load (showing loading animation only) for 20s. I found that we aren't 'priming the pump' by fetching the mesos summary immediately, as we do other polling endpoints.

## Testing

Hard refresh using master & this branch. This branch should allow rendering to happen sooner.


src/js/stores/MesosSummaryStore.js
```diff
@@ -21,29 +21,35 @@ import SummaryList from "../structs/SummaryList";
...
-        // history when navigating back, for case where history server is down.
-
-        // Use {silent: true} Because we only want to push a summary on the stack without side
-        // effects (like re-rendering etc). The tab is out of focus so we
-        // don't want it to do any work. It only matters that there is
-        // appropriate history when we return focus to the tab.
-        this.processSummaryError({ silent: true });
-      }
-    }, Config.getRefreshRate());
+      fetchSummaryIfActive.call(this);
```
DanielMSchmidt: Is the `.call` this really required? What if we made `fetchSummaryIfActive ` an arrow function?

AmrAbdelrazik:
Sweet!

PR #3786, opened 2019-04-05
[DCOS-51247] fix: gm-scroll-view does not trigger onScroll events

brandonc:
## Overview

reactjs-components Table was expecting `.gm-scroll-view` to raise scroll events in order to update it's virtual list of items. These events were not being raised because this is not the scrollable element in gemini. The new selector is the one added to &lt;GeminiScrollbar&gt; component in our FluidGeminiScrollbar wrapper.

There is another separate issue of the entire &lt;tbody&gt; getting replaced on the DOM when the event is raised, but this fixes the "giant blank screen" issue after scrolling ends and reveals a "janky scroll because the browser is too busy manipulating the DOM" issue.

## Testing

Go to the jobs history page on SOAK to see a long list and start scrolling.

### DIFF
--- a/plugins/nodes/src/js/components/HealthTab.js
+++ b/plugins/nodes/src/js/components/HealthTab.js
@@ -163,7 +163,7 @@ class HealthTab extends React.PureComponent {
             table-borderless-inner-columns table-hover flush-bottom"
           columns={this.getColumns()}
           colGroup={this.getColGroup()}
-          containerSelector=".gm-scroll-view"
+          containerSelector=".gm-scrollbar-container-fluid-view-width"
           data={visibleData}
           itemHeight={TableUtil.getRowHeight()}
           sortBy={{ prop: "health", order: "asc" }}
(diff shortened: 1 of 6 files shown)


brandonc:
Update:

I dove deep into “window” rendering (react-virtualized etc) and determined that our reactjs-components’ `VirtualList` (used by our Table component) has a fatal flaw: It does not render absolutely positioned elements with a diffable `style` attribute and therefore imposes too much DOM activity on React. You can see an ‘overscroll’ behavior in the simplest examples which is simply ridiculous. I sought to replace VirtualList with react-window, but unfortunately ran out of time. Plus, I have a feeling react-window is fundamentally incompatible with &lt;table&gt; positioning anyway, which is imposed by the Table component. If so much work is needed, it might best to replace with the ui-kit table.

juliangieseke:
Hey Brandon, looking at soak with your branch doesnt seem to improve a lot, am I doing something wrong or is it the reason for your follow-up comment? Do you think that replacing the table actually help? My impression was more like there is not "that one" thing that improves performance drastiaclly, but its more like the general amount of load thats on chrome during scrolling 🤔 

https://cl.ly/972399115c0b

brandonc:
@juliangieseke It's still possible to overscroll with this table (even in the documentation example) and I believe the reason to be what I mentioned in the update... But I think things are still improved because the table no longer has to wait for a polling response to receive an update, as opposed to a scroll event.

GeorgiSTodorov:
I also don't see much of a difference.

PR #3787, opened 2019-04-06
fix(TaskDetail): don't hang on loading

natmegs:
Services task details logs used to hang if reloaded on the logs page. This happened because it tried to fetchDirectory once immediately (using the MesosStateStore success event which only fires once), at which point the necessary information (task) wasn't available in the store. Use MesosStateStore events to retry data fetching until necessary info is available.

## Testing

- On soak, go to services page and choose a service with logs (ie data-services > cassandra)
- Go to a task detail page
- Go to logs
- They should load as expected
- Refresh on that page
- Logs should still load

## Trade-offs

It takes a while to load the logs page from a refresh because there is a significant delay in waiting to get a response from the mesos stream. The fetchDirectories request isn't made until the mesos response is available.

## Dependencies

N/A

## Screenshots

N/A

### DIFF
--- a/plugins/services/src/js/pages/task-details/TaskDetail.js
+++ b/plugins/services/src/js/pages/task-details/TaskDetail.js
@@ -14,6 +14,7 @@ import RequestErrorMsg from "#SRC/js/components/RequestErrorMsg";
 import RouterUtil from "#SRC/js/utils/RouterUtil";
 import StoreMixin from "#SRC/js/mixins/StoreMixin";
 import TabsMixin from "#SRC/js/mixins/TabsMixin";
+import { MESOS_STATE_CHANGE } from "#SRC/js/constants/EventTypes";
 
 import TaskDirectoryStore from "../../stores/TaskDirectoryStore";
 import TaskStates from "../../constants/TaskStates";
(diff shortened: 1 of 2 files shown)


mperrotti:
Refreshing the page takes about the same amount of time to show content as before this change.

However, when I navigate to the task detail page organically, `this.state.directory` always comes back `null`, and the screen will start to flash the logs UI and then it will disappear: https://cl.ly/b76111ca52cc (flash of logs UI shown towards the end of the video)

natmegs:
@mperrotti Interesting, when I refresh directly on a logs page on soak without this change, it hangs on loading indefinitely and never loads? I'm still looking into the issue when navigating to logs organically

GeorgiSTodorov:
I am not using soak, but I started cassandra and I can't reproduce the original problem on master.

mperrotti:
@natmegs - I'm still having the same issue I mentioned in April. We can do a screen-share sometime if you want to pair on it or see if I'm doing anything strange

pierrebeitz:
i think this was solved by https://github.com/dcos/dcos-ui/pull/4009.

PR #3789, opened 2019-04-08
DCOS-51370: NodesTable: avoid re-renders

DanielMSchmidt:
## Testing

Open a nodes agents page with react dev tools and [flash on rerender](https://blog.logrocket.com/make-react-fast-again-part-3-highlighting-component-updates-6119e45e6833) enabled and see that there are less then before this change
## Screenshots

| Before                                                                                                                                        | After                                                                                                                                         | After - no render if no data change                                                                                                           |
|-----------------------------------------------------------------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------|
| Bildschirmfoto 2019-04-08 um 15 27 05 | Bildschirmfoto 2019-04-08 um 15 25 40 |                    |


mperrotti:
I can't get the smoke test `"displays one row on the table [00t]"` to fail locally 🤔

package.json
```diff
@@ -50,6 +50,8 @@
     "d3": "3.5.16",
     "date-fns": "1.30.1",
     "dygraphs": "2.1.0",
+    "fast-deep-equal": "2.0.1",
```
brandonc: We already bundle lodash.isequal... Should we just reuse that or maybe migrate other callers?

nLight: @DanielMSchmidt you might want to follow up on this 👼

DanielMSchmidt: Followup: https://github.com/dcos/dcos-ui/pull/3795

PR #3790, opened 2019-04-08
DCOS-49242: [SOAK] TypeError: Cannot read property 'push' of undefined

GeorgiSTodorov:
Display service not found page if the user
tries to edit an unknown servide id.

Closes https://jira.mesosphere.com/browse/DCOS-49242

~This is work in progress, I still have to fix the test.~

## Testing
1. Try to edit a service that doesn't exist.
2. Verify that the UI doesn't break
3. Verify that the "service not found" screen is shown

## Trade-offs
I am not sure what this `props.history` is supposed to do, we don't have it anywhere else in the UI.

## Dependencies
None.

### Before
Peek 2019-04-08 14-25

### After
Peek 2019-04-09 15-10


locale/zh/messages.json
```diff
@@ -946,6 +946,7 @@
   "The secret cannot be empty.": "密钥不能为空。",
   "The service is currently locked by one or more deployments. Press again to force this operation.": "服务当前被一个或多个部署锁定。再次按下以强制执行此操作。",
   "The service with the ID of \"{id}\" could not be found.": "未找到带有 ID \"{id}\" 的服务。",
+  "The service with the ID of \"{serviceID}\" could not be found.": "未找到带有 ID \"{serviceID}\" 的服务。",
```
nLight: 👍

plugins/services/src/js/components/modals/EditServiceModal.js
```diff
@@ -19,9 +24,18 @@ class EditServiceModal extends Component {
 
     // Service not found
     if (!service) {
-      this.props.history.push("/services/404");
```
nLight: I believe it's a forgotten leftover of the react router migration. Please look for `hashHistory.push` as an example.

plugins/services/src/js/components/modals/__tests__/EditServiceModal-test.js
```diff
@@ -28,16 +28,11 @@ describe("EditServiceModal", function() {
         return null;
       }
     };
-    const pushMock = jest.fn();
```
nLight: please refer to `intercomPlugin-test.js` to see how to mock `hashHistory`

plugins/services/src/js/components/modals/EditServiceModal.js
GeorgiSTodorov: It seems that way to me.

nLight: That would be a sep issue. Does it show the 404 page correctly though?

GeorgiSTodorov: It does.

nLight: cool! let's adjust the test and merge it!

GeorgiSTodorov:
@nLight , thank you for the helpful suggestions.

PR #3794, opened 2019-04-09
[DCOS-49688] Remove metronome subscriber from DCOSStore

brandonc:
Metronome data was being kept up to date but not used in the DCOSStore. Without this subscriber, MetronomeStore will no longer poll metronome when visiting unrelated pages.

## Testing

When visiting the dashboard or other non-jobs page, you can see a request to the metronome API for jobs data. With this branch, you won't.


nLight:
😻

PR #3792, opened 2019-04-08
DCOS-51596: Services Table: avoid rerenders

DanielMSchmidt:
## Testing

- Open a services page with react dev tools and [flash on rerender](https://blog.logrocket.com/make-react-fast-again-part-3-highlighting-component-updates-6119e45e6833) enabled and see that there are less then before this change
- Check that loading icons and deploying status bars still display as before

## Screenshots

| Before                                                                                                                                        | After first commit                                                                                                                            | After - no render if no data change                                                                                                           |
|-----------------------------------------------------------------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------|
| Bildschirmfoto 2019-04-08 um 16 07 45 | Bildschirmfoto 2019-04-08 um 16 48 55 | Screen Shot 2019-04-09 at 2 16 10 PM  |

PR #3796, opened 2019-04-09
DCOS-51377: Services Table: avoid rerender (except for status icons)

DanielMSchmidt:
## Testing

- Open a services page with react dev tools and [flash on rerender](https://blog.logrocket.com/make-react-fast-again-part-3-highlighting-component-updates-6119e45e6833) enabled and see that there are less then before this change
- Check that loading icons and deploying status bars still display as before

## Trade-offs

- Left out the status icons. @natmegs did an awesome job in #3792, but we failed to see that @TattdCodeMonkey, @Poltergeist and @pierrebeitz were working on the same column. As this is a lot of work (very entangled struct usage), I bailed out here. I could not rebase the work done in #3792, so I thought I'll just extract what I can to ship this already.

## Screenshots

| Before                                                                                                                                        | After first commit                                                                                                                                         |
|-----------------------------------------------------------------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------|
| Bildschirmfoto 2019-04-08 um 16 07 45 | Bildschirmfoto 2019-04-08 um 16 48 55


nLight:
legit fails

natmegs:
This seems to duplicate #3792 but missing all the service status column stuff?

DanielMSchmidt:
@natmegs Because of the accurate service merges there are quite some conflicts for the status column with every PR we merge on that topic. Let's revisit this soon and do a final rebase once the accurate service work is all done 👍

plugins/services/src/js/containers/services/__tests__/ServicesTableRenderers-test.js
```diff
@@ -1,181 +1,46 @@
...
-        renderer.create(renderRegions(mockService)).toJSON()
+        renderer.create(<ServiceDisk resource="1" />).toJSON()
       ).toMatchSnapshot();
     });
+  });
 
-    it("renders with multiple regions", function() {
-      mockService.getRegions.mockReturnValue(["aws/eu-central-1", "dc-east"]);
-
+  describe("#renderRegions", function() {
```
pierrebeitz: why change to a single region?

plugins/services/src/js/columns/ServicesTableInstancesColumn.tsx
```diff
@@ -11,34 +11,49 @@ import StringUtil from "#SRC/js/utils/StringUtil";
...
+      <NumberCell>
+        <Tooltip content={tooltipContent}>
+          <span>{content}</span>
+        </Tooltip>
+      </NumberCell>
+    );
+  }
+);
+
+export function instancesRenderer(
```
pierrebeitz: i don't like that we now always export the cellRenderer and a memoized component. but that's not to block these awesome changes.

plugins/services/src/js/columns/ServicesTableDiskColumn.tsx
```diff
@@ -8,16 +8,17 @@ import Units from "#SRC/js/utils/Units";
 import { SortDirection } from "plugins/services/src/js/types/SortDirection";
 import ServiceTableUtil from "../utils/ServiceTableUtil";
 
+export const ServiceDisk = React.memo(({ resource }: { resource: string }) => (
+  <NumberCell>
+    <span>{Units.formatResource("disk", resource)}</span>
+  </NumberCell>
+));
+ServiceDisk.displayName = "ServiceDisk";
```
pierrebeitz: some components have been lucky while others have been not? or is there a reason, e.g. `ServiceName` does not have/need that property? (it's totally optional from my POV, but still curious)

plugins/services/src/js/containers/services/__tests__/ServicesTableRenderers-test.js
DanielMSchmidt: Is your concern that there is no test for multiple regions?

plugins/services/src/js/columns/ServicesTableInstancesColumn.tsx
DanielMSchmidt: Can you explain why? From my point of view it looks like the renderer now does a transition from structs to primitives, so the "wiring up / business-logic" part and the component does the rendering. 

The reason why I chose to export the component as well is that it's easier to test that way. Testing that the structs are called in certain ways or respond in certain ways seemed pretty fragile to me, therefore I switched the tests to purely test what is being rendered.

plugins/services/src/js/columns/ServicesTableDiskColumn.tsx
DanielMSchmidt: I'll remove them, I only added them to debug another problem but they weren't of any help there.

plugins/services/src/js/containers/services/__tests__/ServicesTableRenderers-test.js
pierrebeitz: i have no concern here. i just saw that the test checked for multiple regions before and now does not anymore and was curious about why that is.

plugins/services/src/js/columns/ServicesTableInstancesColumn.tsx
pierrebeitz: > Can you explain why?

🤔 i don't like the current factorization of column-modules.
they seem to aim to contain the stuff that's associated with a column but miss e.g. the header-data. i wonder whether we had less code and better comprehensibility if we inlined that stuff into the `ServiceTable`. but maybe that's just me liking big files to get a lot of context fast.

i realise that my concern has not to do with this PR directly, so please resolve.

plugins/services/src/js/containers/services/__tests__/ServicesTableRenderers-test.js
pierrebeitz: after our offline-discussion, i'd agree that these tests are more likely to create churn than save us from mistakes.

DanielMSchmidt:
I split up the commits to be per column for easier rebasing and I also removed the unit tests, because they did not seem to add any value

plugins/services/src/js/columns/ServicesTableInstancesColumn.tsx
juliangieseke: I introduced multiple column files to enable us to easily work on them without too much conflicts. I'm happy inlining them into one file tho. maybe something for spike pr 🤷‍♂️

plugins/services/src/js/columns/ServicesTableNameColumn.tsx
```diff
@@ -14,33 +14,70 @@ import Pod from "../structs/Pod";
...
-        <Link className="table-cell-icon" to={serviceLink}>
-          {getImage(service)}
-        </Link>
-        <span className="table-cell-value table-cell-flex-box">
-          {getServiceLink(service, isFiltered)}
-          {getOpenInNewWindowLink(service)}
-        </span>
-      </div>
-    </TextCell>
+    <ServiceName
```
juliangieseke: 'nit: I think it would have been possible to split this into `ServiceName` and `GroupName` component 🤔 not sure if I'm missing something and I think its not that important tho.

plugins/services/src/js/containers/services/__tests__/ServicesTableRenderers-test.js
```diff
@@ -1,190 +0,0 @@
-import { NumberCell } from "@dcos/ui-kit";
```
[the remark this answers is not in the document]
DanielMSchmidt: Yes, I wrote some of them back then :)

plugins/services/src/js/columns/ServicesTableNameColumn.tsx
DanielMSchmidt: Same as before, I did not touch anything in this direction. I don't see that this is part of the scope of this ticket.

DanielMSchmidt: At least if our goal is still to merge this functionality today. If this goal changed we can also close this PR and use the #3792 one

PR #3803, opened 2019-04-10
Services table: show region for groups as well

DanielMSchmidt:
## Testing

- Check if there are now regions shown for groups and verify that no region is shown that is not shown within the group
- Also check groups in groups behaiour


GeorgiSTodorov:
@DanielMSchmidt , do we still wanna do this? I can take a look at the failing build.

DanielMSchmidt:
Only the rerun script is failing, the real merge run is green. I'll check with @nLight

GeorgiSTodorov:
Do we merge?

DanielMSchmidt:
@nLight Will decide if it makes sense from the product point of view and either merge or close it 👍

nLight:
I'm concerned adding one more `reduce`. I've learned our customers nest groups a lot.
Unless we combine it with resource aggregation somehow I'd reject the PR, sorry.

If we resolve the performance concern we need to work on the UX a bit.

if you'd be up for owning the change I'd suggest involving design and come back to it in September :)

Screenshot 2019-07-18 at 18 10 34

DanielMSchmidt:
Maybe this is not the UX quick win that I hoped for, let's find a more fine-grained and thoroughly designed solution in the future

DanielMSchmidt:
Maybe this is not the UX quick win that I hoped for, let's find a more fine-grained and thoroughly designed solution in the future

PR #3804, opened 2019-04-10
fix(PodContainer): ensure conditions exists

TattdCodeMonkey:
## Testing

view an unhealthy pod. On SOAK try `pull-image-auth` or `pull-image-cert` and verify UI doesnt crash

### DIFF
--- a/plugins/services/src/js/structs/PodContainer.js
+++ b/plugins/services/src/js/structs/PodContainer.js
@@ -71,21 +71,30 @@ module.exports = class PodContainer extends Item {
   }
 
   hasHealthChecks() {
+    const conditions = this.get("conditions");
+
     // According to RAML specs:
     //
     // https://github.com/mesosphere/marathon/blob/feature/pods/docs/docs/rest-api/public/api/v2/types/container-status.raml#L49
     // 'healthy: should only be present if a health check is defined for this endpoint'
     //
-    return this.get("conditions").some(function(cnd) {
-      return cnd.name === "healthy";
-    });
+    return (
+      conditions &&
+      Array.isArray(conditions) &&
+      conditions.some(function(cnd) {
+        return cnd.name === "healthy";
+      })
+    );
   }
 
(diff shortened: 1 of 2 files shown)


jaxesn:
test....

plugins/services/src/js/structs/PodContainer.js
```diff
@@ -71,21 +71,31 @@ module.exports = class PodContainer extends Item {
...
   isHealthy() {
+    const conditions = this.get("conditions");
+
     // If we have at least 1 health check and it has failed, we are assumed to
     // be unhealthy.
-    return !this.get("conditions").some(function(cnd) {
-      return cnd.name === "healthy" && cnd.value === "false";
-    });
+    return (
+      conditions &&
```
DanielMSchmidt: If this is undefined, meaning there are no health checks (I guess), this would return false, wouldn't it?

nLight: not really, empty conditions / undefined means healthy
there's a ancient test that suggests this

nLight:
@TattdCodeMonkey we need at least 1 unit test please.

plugins/services/src/js/structs/PodContainer.js
DanielMSchmidt: yes, and `undefined && ...` returns false, making `isHealth()` false. Is my logic flawed here?

nLight: undefined means `false` and we should return `true` in this case

TattdCodeMonkey:
@nLight  add *2* tests!
@DanielMSchmidt  re-worked logic

plugins/services/src/js/structs/PodContainer.js
TattdCodeMonkey: 👍 I resolved this by wrapping the conditions check in a `this.hasHealthChecks` call and returning `true` if there are no health checks.

PR #3805, opened 2019-04-10
Resizable Nodes and Services table columns

mperrotti:
## Testing
(recommended in a private browsing tab)
1. Navigate to the {Nodes|Services} page
2. Resize some columns
3. Refresh the page

Columns should be resizable, and when you refresh the page, the resized columns should be the same width you resized them to

## Screenshots
ServicesResizeDemo

### DIFF
--- a/plugins/nodes/src/js/components/NodesTable.tsx
+++ b/plugins/nodes/src/js/components/NodesTable.tsx
@@ -11,27 +11,61 @@ import isEqual from "lodash/isEqual";
 import NodesList from "#SRC/js/structs/NodesList";
 import Node from "#SRC/js/structs/Node";
 import Loader from "#SRC/js/components/Loader";
+import TableUtil from "#SRC/js/utils/TableUtil";
+import TableColumnResizeStore from "#SRC/js/stores/TableColumnResizeStore";
 
 import { SortDirection } from "../types/SortDirection";
 
-import { ipSorter, ipRenderer } from "../columns/NodesTableIpColumn";
-import { typeSorter, typeRenderer } from "../columns/NodesTableTypeColumn";
+import { ipSorter, ipRenderer, ipWidth } from "../columns/NodesTableIpColumn";
+import {
+  typeSorter,
+  typeRenderer,
+  typeWidth
+} from "../columns/NodesTableTypeColumn";
 import {
   regionSorter,
-  regionRenderer
+  regionRenderer,
+  regionWidth
 } from "../columns/NodesTableRegionColumn";
-import { zoneSorter, zoneRenderer } from "../columns/NodesTableZoneColumn";
+import {
+  zoneSorter,
+  zoneRenderer,
+  zoneWidth
+} from "../columns/NodesTableZoneColumn";
 import {
   healthSorter,
-  healthRenderer
+  healthRenderer,
+  healthWidth
 } from "../columns/NodesTableHealthColumn";
-import { tasksSorter, tasksRenderer } from "../columns/NodesTableTasksColumn";
-import { cpuSorter, cpuRenderer } from "../columns/NodesTableCPUColumn";
-import { memSorter, memRenderer } from "../columns/NodesTableMemColumn";
-import { diskSorter, diskRenderer } from "../columns/NodesTableDiskColumn";
-import { gpuSorter, gpuRenderer } from "../columns/NodesTableGPUColumn";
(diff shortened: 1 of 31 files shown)


plugins/nodes/src/js/components/NodesTable.tsx
```diff
@@ -297,8 +439,21 @@ export default class NodesTable extends React.Component<
               />
             }
             cellRenderer={memRenderer}
-            growToFill={true}
-            minWidth={110}
+            growToFill={
```
jaxesn: This is all fine for this PR, but I wonder if there could be a future improvement to how we handle these resizeable columns...  Having the Store to store this state makes a lot of sense so having this kind of logic in dcos-ui vs ui-kit is reasonable.  Alot of this PR, esp the lines following, are boilerplate-ish and some of potentially error-proned, like this growToFill "logic"...  Would there be a slick way to have a HOC or some other kind of component/helper in dcos-ui that can almost "generate" this boilerplate to make the ultimate api something like:

<Column minWitdth={110} resizeable={true} columnKey={"mem"} />

GeorgiSTodorov:
Also, unit and snapshot tests are failing.
Screenshot from 2019-04-11 11-54-31

GeorgiSTodorov:
Currently looking into the non-snapshot failure.

GeorgiSTodorov:
Merging master to see if it will fix the issue.

GeorgiSTodorov:
The same test is failing again.
Screenshot from 2019-04-11 13-58-01

plugins/nodes/src/js/components/NodesTable.tsx
nLight: I'd agree with Jackson. We should design the interface a little here.

```jsx
<ConfigurableTable id="table_id" storeBackend={LocalStorageBackend}>
    <ResizableColumn minWitdth={110} columnKey="mem" />
</ConfigurableTable>
```

or even

```jsx
<Configuration id="table_id" storeBackend={LocalStorageBackend}>
  <Table>
    <ResizableColumn minWitdth={110} columnKey="mem" />
  </Table>
</Configuration>
```

Configuration would inject defaults into its child component. We would need figuring out the `onConfigurationChange` interface though.

GeorgiSTodorov:
Taking a look at the failing tests.

GeorgiSTodorov:
I've somewhat narrowed down the problem. It comes from the new ui-kit version. It makes the table have height 0 in integration tests.
I tested this by first running master and seeing tests work and then running master with new ui-kit version 2.1.1 and seeing tests break because of invisible tables.

mperrotti:
This ui-kit PR should fix the issue: https://github.com/dcos-labs/ui-kit/pull/324

Alternatively, we could have added styles to ensure the `div` that is a direct child of `.table-wrapper` fills it's parent's height. The ui-kit fix feels much cleaner though

GeorgiSTodorov:
If there are still failing tests, I will investigate them.

GeorgiSTodorov:
Can't reproduce failing tests locally, retriggered.

GeorgiSTodorov:
Still can't reproduce them locally. Reasons for test failures looking at the artifact video:
Screenshot from 2019-04-18 12-13-20
Screenshot from 2019-04-18 12-14-48

It looks like a flake to me.

mperrotti:
@GeorgiSTodorov - I wasn't able to reproduce locally either. I wonder why these tests aren't flaking on other PRs though 🤔

GeorgiSTodorov:
@mperrotti , are we sure they aren't?

GeorgiSTodorov:
Merging latest master to see if they are still flakes.

GeorgiSTodorov:
Now it fails locally.
Screenshot from 2019-04-25 14-50-36

tests/pages/services/ServiceFormModal-cy.js
```diff
@@ -500,6 +500,7 @@ describe("Service Form Modal", function() {
             .focus()
             .type("{selectall}{backspace}")
             .type("{selectall}{backspace}")
+            .type("{selectall}{backspace}")
```
mperrotti: Locally, this prevents Cypress from typing part of the input at the beginning of the field. This is more of a workaround than an actual fix, but I want to see if it works in CI

nLight:
flicts :)

PR #3812, opened 2019-04-11
DCOS-51670 fix: pre-commit

TattdCodeMonkey:
Jest was always returing `1` when running with pre-commit. This ended up being an issue loading a module from an incorrect path. I had to upgrade jest-cli locally just to see this error. We should work on updating jest too.

Additionally once the module path was fixed this exposed a few TS errors. 

I also updated the pre-commit script to look for tests for both JS & TS* files.

## Testing

make a change and try to commit something, ensure you can commit without `--no-verify`


GeorgiSTodorov:
@TattdCodeMonkey , do we want this merged before or after https://github.com/dcos/dcos-ui/pull/3816 ?

PR #3817, opened 2019-04-11
[spike] 'prefetch' script hints for chart chunks

brandonc:
## Overview

Offer [prefetch](https://developer.mozilla.org/en-US/docs/Web/HTML/Preloading_content#Other_resource_preloading_mechanisms) hints to the browser for unneeded yet useful webpack chunks.

I was doing some research into the critical path for loading the login page and found an easy win: With this webpack-html-plugin extension, you can alter the script loading behavior of various webpack chunks. Previously, all non-dynamic chunk imports (such as the chart chunks) load synchronously in the &lt;head&gt; of the html document in anticipation of the dashboard being loaded.

Collectively, these represent 160K of scripts whose loading can be deprioritized, and they also need not be evaluated, either.

I don't know if this improves loading by a measurable difference, but this change could theoretically help login loading speed, especially on high latency networks.

The second commit (index.html) is a small addition to the new relic (?) script that adds async to that dynamically added script tag. It's hard to spot in the diff. Find "n.async=true"

## Testing

- Ensure dashboard loads as expected

## Trade-offs

None

## Screenshots

Before:

Screen Shot 2019-04-11 at 3 48 24 PM

After:

Screen Shot 2019-04-11 at 3 49 37 PM


GeorgiSTodorov:
This makes dashboard load slower and the charts are blank for a few seconds. This might be a problem. It's especially visible on slow networks (slow 3g in the gifs):
### Here
Peek 2019-04-12 11-32

### Master
Peek 2019-04-12 11-36

nLight:
Brandon, that's awesome! I wonder if we could adding react16 <loading> to those so that they don't look "broken"? :)

brandonc:
@nLight I added Loader to each chart. Here is the effect (slow 3G throttling)

Apr-12-2019 12-59-03

I also removed the NR async commit because it poses an unnecessary risk. It should remain a black box

GeorgiSTodorov:
`pages/services/ServiceDetail` is failing. I will try to reproduce locally.

PR #3820, opened 2019-04-12
DCOS-51269: COPS-4710 affecting : Package-specific icons not showing for some services installed from Universe

GeorgiSTodorov:
Add icons for services that are installed from
Universe, but are not classified as Frameworks.

Closes https://jira.mesosphere.com/browse/DCOS-51269

## Testing
1. Start some universe services that are not classified as frameworks, for example "spark-history" or "kibana".
2. Check that they have their own icons and not the default ones (as icons don't work in production, you can inspect the element to see the link and verify that they are showing a missing icon and not the default one).
3. Check the same thing for services inside folders.
4. Verify that no universe or not-universe service breaks the UI.

## Trade-offs
This is work in progress. I have to update/add tests, but it's otherwise reviewable.

## Dependencies
None.

### Before
Screenshot from 2019-04-12 15-07-49
Screenshot from 2019-04-12 15-08-13

### After
Screenshot from 2019-04-12 14-53-52
Screenshot from 2019-04-12 14-54-14


plugins/services/src/js/stores/__tests__/MarathonStore-test.js
```diff
@@ -138,6 +138,9 @@ describe("MarathonStore", function() {
       var marathonApps = MarathonStore.get("apps");
 
       Object.keys(marathonApps).forEach(function(key) {
+        if (!marathonApps[key].health) {
```
GeorgiSTodorov: This is needed, because otherwise the test breaks because of images in marathonApps.

nLight: Could you please elaborate? What test is breaking? What is the expectation?

plugins/services/src/js/structs/Application.js
```diff
@@ -65,7 +66,14 @@ module.exports = class Application extends Service {
    * @override
    */
   getImages() {
-    return FrameworkUtil.getServiceImages(this.getMetadata().images);
+    if (
```
nLight: Can't we do something like this instead?
```js
const images = this.getMetadata().images || this.get("images");

return FrameworkUtil.getServiceImages(images);
```

plugins/services/src/js/stores/__tests__/MarathonStore-test.js
GeorgiSTodorov: Here https://github.com/dcos/dcos-ui/pull/3820/files#diff-7b6a93d2ba53835d3ed69d3f2e183f45R483 we add images property to Applications and this is a property which didn't exist when this test was written, so the test expected images to have a key and failed.

plugins/services/src/js/structs/Application.js
GeorgiSTodorov: This won't work the same in some cases, for example if the metadata images are the same as the default ones.

nLight: Is there an example when we have default images in the metadata but custom in `images`? If we have default images `getImageSizeFromImagesObject` won't return null, will it?

plugins/services/src/js/stores/__tests__/MarathonStore-test.js
nLight: Do you mean that the old fixture is incomplete?

GeorgiSTodorov: The `forEach` definitely is.

plugins/services/src/js/structs/Application.js
GeorgiSTodorov: If everyone can add to Universe, then such an example is possible.
`getImageSizeFromImagesObject` won't return null, but `getServiceImages` will return the default images.

plugins/services/src/js/stores/__tests__/MarathonStore-test.js
nLight: As I can see in this foreach we just check every marathon app from the store. I fail to understand why would we ever want start skipping some of them?

GeorgiSTodorov: Here we are checking for health checks, so we don't care about marathon apps with no health, so we skip them. If you insist, I can also try changing the fixtures.

nLight: It's not me insisting, I just don't see where we add new fixtures? So in my understanding this test shouldn't fail. Other question would be: how exactly doest it fail?

GeorgiSTodorov: We don't add new fixtures, but we add the possibility for Applications to have images. It fails with `can't find key of undefined` on this line - https://github.com/dcos/dcos-ui/pull/3820/files/e2567b8eedc40c48808e924003772b32ff30ebe6#diff-4da2835a0dd453e39652befc964aa197R150 if I remember correctly.

plugins/services/src/js/stores/MarathonStore.js
```diff
@@ -474,6 +475,14 @@ class MarathonStore extends GetSetBaseStore {
         item["images"] = CosmosPackagesStore.getPackageImages()[
           item.labels.DCOS_PACKAGE_NAME
         ];
+      } else if (
+        ServiceValidatorUtil.isApplicationResponse(item) &&
```
brandonc: It seems to me like you could add this clause to the last else if and lose an entire condition block here:

`if (ServiceValidatorUtil.isFrameworkResponse(item) || ServiceValidatorUtil.isApplicationResponse(item)) {`

GeorgiSTodorov: Yes, I can.

pierrebeitz:
It's great to see this fixed!

Have we found out what the actual issue is? If there are package definitions lacking labels, we should probably tell someone. If it's us (or rather `dcos-ui`) misinterpreting the data, we should plan to refactor this solution sometime to not only have access to the images but rather all of a frameworks functionality, right?

GeorgiSTodorov:
@pierrebeitz , the problem was that some non-framework services from Universe want to have icons as well and until not we didn't allow them to have icons unless they came from the metadata.

plugins/services/src/js/stores/__tests__/MarathonStore-test.js
natmegs: I understand why the test fails if it meets this condition, but I am unclear why this PR's changes would have caused this to fail? Why is `health` affected by the images changes?

GeorgiSTodorov:
@natmegs , the marathonApps used to only have health properties, now they have non-health properties as well (images), so they fail this check since they lack a key.

natmegs:
@GeorgiSTodorov Right, I guess the part I don't understand is why would adding an additional property (`images`) cause a check for `marathonApps[key].health.key` to fail? Are there changes to the `health` property that I missed?

GeorgiSTodorov:
`marathonApps[key].health.key` fails, because not all marathonApps keys have health, as some of them are images.

GeorgiSTodorov:
Tomorrow I will post `marathonApps` before and after to hopefully clear this confusion.

GeorgiSTodorov:
### Before
Screenshot from 2019-04-18 10-41-24

### After
Screenshot from 2019-04-18 10-38-57

We have no name here because the name is not defined here - https://github.com/dcos/dcos-ui/pull/3820/files#diff-7b6a93d2ba53835d3ed69d3f2e183f45R505 . Our first instinct might be to add a name, but the test will still fail, because it doesn't have health.

### After with name
Screenshot from 2019-04-18 10-37-55

An alternative fix would be to check if name exists here - https://github.com/dcos/dcos-ui/pull/3820/files#diff-7b6a93d2ba53835d3ed69d3f2e183f45R505 , but in practice all applications have names and I don't want to change the code only for some test to pass.

brandonc:
@natmegs I think the new addition to processMarathonGroups is causing this data to show up in the test.

I think we should add health data to the application fixtures, or if they aren't supposed to have health data, just modify the test to use a filter() before the forEach() to make it look more deliberate.

GeorgiSTodorov:
@brandonc , I think adding health data to the fixtures is also an option. I will try and I will do it instead, if it's not too complicated.

plugins/services/src/js/stores/MarathonStore.js
```diff
@@ -495,6 +501,12 @@ class MarathonStore extends GetSetBaseStore {
         };
       }
 
+      if (item instanceof Application) {
+        map[item.getName().toLowerCase()] = {
+          images: item.getImages()
```
juliangieseke: Framework `if` above also has `health` and `snapshot` keys. adding these here fixes the broken test.

GeorgiSTodorov:
@juliangieseke said that he wants to do a follow-up PR for these tests.

juliangieseke:
no follow up PR. found a legit bug in this one.

GeorgiSTodorov:
@juliangieseke , what is the bug?

plugins/services/src/js/stores/MarathonStore.js
GeorgiSTodorov: Are we sure Applications have health and snapshots?

juliangieseke: I think the correct question should be "why is there this `if framework` clause in the first place? why do we need it?" can you figure that out please? Also `getFrameworkName` returns the predefined framework name, not the services ID… so shouldnt that block be broken if thee are multiple frameworks of the same kind installed? 🤔

GeorgiSTodorov: About the first question: I think we didn't want non-frameworks to have images.

GeorgiSTodorov: About the second question: it is actually the service id.
```
  processMarathonGroups(data) {
    const groups = new ServiceTree(data);
    const apps = groups.reduceItems(function(map, item) {
      if (item instanceof Framework) {
        console.log(item.getFrameworkName());
        map[item.getFrameworkName().toLowerCase()] = {
          health: item.getHealth(),
          images: item.getImages(),
          snapshot: item.get()
        };
      }
...
```
Screenshot from 2019-04-23 13-09-42

GeorgiSTodorov:
I pushed @juliangieseke 's fix with the assumption that Applications always have health and snapshot.

plugins/services/src/js/stores/MarathonStore.js
```diff
@@ -470,7 +471,12 @@ class MarathonStore extends GetSetBaseStore {
     data.items.forEach(item => {
       if (item.items && Array.isArray(item.items)) {
         this.injectGroupsWithPackageImages(item);
-      } else if (ServiceValidatorUtil.isFrameworkResponse(item)) {
+      } else if (
+        (ServiceValidatorUtil.isFrameworkResponse(item) ||
+          ServiceValidatorUtil.isApplicationResponse(item)) &&
```
nLight: I wonder if we should add `Pod` as well. So basically every item that has `DCOS_PACKAGE_NAME`

plugins/services/src/js/stores/MarathonStore.js
```diff
@@ -495,6 +501,14 @@ class MarathonStore extends GetSetBaseStore {
         };
       }
 
+      if (item instanceof Application) {
+        map[item.getName().toLowerCase()] = {
+          health: item.getHealth(),
```
nLight: do we really need adding `health` and `snapshot` along with images?

plugins/services/src/js/stores/MarathonStore.js
```diff
@@ -495,6 +501,14 @@ class MarathonStore extends GetSetBaseStore {
         };
       }
 
+      if (item instanceof Application) {
+        map[item.getName().toLowerCase()] = {
```
nLight: This will also affect line 517 https://github.com/dcos/dcos-ui/pull/3820/files#diff-7b6a93d2ba53835d3ed69d3f2e183f45R531

We used to collect only frameworks there now all the apps will also land into the resulting structure with unknown consequences.

plugins/services/src/js/stores/MarathonStore.js
GeorgiSTodorov: The alternative is changing the marathon store test.

plugins/services/src/js/stores/MarathonStore.js
GeorgiSTodorov: I don't know if Pods can have images. If everything can, we can just check for name and images.
But I am not sure if this change will be in scope here. The bug is that some catalog services don't have icons.

plugins/services/src/js/stores/MarathonStore.js
GeorgiSTodorov: That was suggested by @juliangieseke , and my assumption is that if frameworks can have health and snapshot, then pseudo-frameworks can have them as well.

plugins/services/src/js/stores/MarathonStore.js
GeorgiSTodorov: The way I understand it, our framework check is not ideal, since we only check a label. So everything with the label is a framework.

nLight: exactly right, we consider everything that has `DCOS_PACKAGE_FRAMEWORK_NAME` to be a framework, though for icons we check for another label, namely `DCOS_PACKAGE_NAME` not every catalog package is a Mesos Framework

plugins/services/src/js/stores/MarathonStore.js
nLight: Those are catalog packages, not frameworks

plugins/services/src/js/stores/MarathonStore.js
nLight: Ok, creeping scope is a fair observation and the bug is about service coming from the catalog/universe. I believe one can create a catalog package that will be essentially a Pod.

plugins/services/src/js/stores/MarathonStore.js
GeorgiSTodorov: Aren't catalog packages usually frameworks?

plugins/services/src/js/stores/MarathonStore.js
GeorgiSTodorov: This is why I want to allow Applications that have images to show them.

GeorgiSTodorov:
@nLight , your suggestions and help are always welcome. What exactly are your concerns?

plugins/services/src/js/stores/MarathonStore.js
nLight: nope, a catalog package it's just a configurable Marathon application definition with some artefacts like tar files and docker images. It may or may not be a framework. 

Framework in DCOS and Mesos is an application that schedules mesos tasks when needed. In this regard Marathon is a Framework, Spark and Kafka are as well, but `ngnix` isn't for example.

http://mesos.apache.org/documentation/latest/frameworks/

GeorgiSTodorov: In theory, yes, but in practice, aren't the majority of catalog packages frameworks?

nLight: I would argue that in practice the majority of packages are _not_ frameworks.

GeorgiSTodorov: Then we have one more reason to allow non-frameworks to have images.

plugins/services/src/js/stores/MarathonStore.js
nLight: Are we sure that there will be no side effects augmenting this data structure?

plugins/services/src/js/stores/MarathonStore.js
nLight: images yes, my question was regarding `health` and `snapshot` fields

GeorgiSTodorov: Adding health and snapshot was @juliangieseke 's suggestion and I added them with the assumption that Applications also have them.

plugins/services/src/js/stores/MarathonStore.js
GeorgiSTodorov: I am not absolutely sure, but I can't think of any. If you do, please share so I can handle these.

GeorgiSTodorov:
I will look into the failing tests next week to see if they are legitimate.

DanielMSchmidt:
Had to restart the current test run, there was a jenkins internal error

GeorgiSTodorov:
Thanks @natmegs , @Poltergeist , @DanielMSchmidt .

PR #3822, opened 2019-04-12
DCOS-52769 - Refactor ServicesTable

pierrebeitz:
The `ServiceTableUtil` has been used before to find out how to sort columns. It was used by the column-files that each exposed a sorter. The ServiceTable then switched over the columnsName to obtain one of those sorters that then looked up by column-name in `ServiceTableUtil`. If this sounds confusing you may like this PR.

The tests have been rewritten one to one for more clarity. Previously fixtures have been set up in the beginning whose properties would have to be remembered to fully understand what is about to be tested. Now the behaviour under test should be comprehensible by looking around in the nearby scope.

## Review

the most important things to understand are 
* the sorting has been inlined directly into `sortForColumn` in `ServicesTable.js` (no more code in the `*Column`-files, no more code in `ServiceTableUtil`).
* a version-module has been factored with what was then left in `ServiceTableUtil`
* afterwards the tests have been rewritten 1 to 1

## Testing

* Play with the `ServicesTable` and see whether everything still works. The behaviour when sorting `instances` and `region` columns has been changed/fix. Before one could not sort by regions at all and instances were broken for pods, see 
`fix(ServiceTableUtil): sort by instances`.

## Next steps

This is rather a memo for myself.

### Sort Direction 

I feel like it should be discussed whether the sortDirection-indicators are used in the usual way - i feel like they are used in the opposite way.

current state: 
```
▲ C  |  ▼ A
  B  |    B
  A  |    C
```

i feel like the tip of the arrow is lighter and thus lighter things should show up (this feels especially weird with numeric columns like mem/disk/cpu).

### `ServiceTree` vs (table entry for) `Group`

on a future spike day i want to have a look on whether the Column-files are still necessary. also i now feel certain enough in the domain to introduce a distinction between `ServiceTree` and `Group`, which are used in a very confusing way sometimes causing a lot of null checks.

### See whether the column files are still necessary


GeorgiSTodorov:
> I feel like it should be discussed whether the sortDirection-indicators are used in the usual way - i feel like they are used in the opposite way.

I think the way they are currently is the intended way. @mperrotti might have more info on this.

GeorgiSTodorov:
Also, it is a good idea to create a jira for this and add it to the spikes sprint so it has more visibility.

plugins/services/src/js/containers/services/ServicesTable.js
```diff
@@ -77,6 +61,75 @@ const METHODS_TO_BIND = [
...
+    return "";
+  }
+
+  return pipe(
+    Version.fromService,
+    Version.toDisplayVersion
+  )(serviceTreeNode);
+};
+
+function sortForColumn(name) {
```
GeorgiSTodorov: I think it would be better to call this `columnName` to avoid confusion?

pierrebeitz: i'll gladly change that in a follow up

mperrotti:
@pierrebeitz - you're right about the arrow directions. The design documentation (see the "Sorting" section under the "Readme" tab in Storybook) says:
>Upward chevron is for ascending the items in the column. Ascending means smallest to largest, increasing in size.
>Downward chevron is for descending the items in the column. Descending means largest to smallest, decreasing in size.

@TattdCodeMonkey - I get a sort triggered when I click the sorting indicator. Could you send me a video of the behavior you're seeing?

TattdCodeMonkey:
@mperrotti 🤔 can't reproduce now, assuming it was a PEBKAC.

PR #3825, opened 2019-04-15
DCOS-51824: In Service Create/Edit form, expand the advanced settings if they contain errors

GeorgiSTodorov:
Expand the advanced settings if they contain error
so that the user can see the errors.

Closes https://jira.mesosphere.com/browse/DCOS-51824

## Testing
1. In Services tab, edit or create a single-container service.
2. In the first tab, click "More Settings" and add some error there, for example set Disk to "-1".
3. Close "More Settings".
4. Try to run the service.
5. Verify that "More Settings" was expanded.
6. Fix the error and close "More Settings".
7. Run the service (with another error, for example missing id) and verify that "More Settings" wasn't opened.

## Trade-offs
While working on this, I discovered a bug that also happens on master: https://jira.mesosphere.com/browse/DCOS-52162 .

## Dependencies
None.

### Before
Peek 2019-04-15 16-34

### After
Peek 2019-04-15 16-44


GeorgiSTodorov:
Failing tests are passing locally.

plugins/services/src/js/components/modals/CreateServiceModalForm.js
```diff
@@ -140,6 +140,10 @@ class CreateServiceModalForm extends Component {
   }
 
   componentWillReceiveProps(nextProps) {
+    if (nextProps.expandAdvancedSettings) {
+      this.forceUpdate();
```
GeorgiSTodorov: This is needed, otherwise the form won't get re-rendered and the advanced settings can't expand.

nLight: Please find another solution, we have just finished removing those. Not to start adding them again

src/js/components/form/AdvancedSection.js
```diff
@@ -18,6 +18,12 @@ class AdvancedSection extends React.Component {
     });
   }
 
+  componentWillReceiveProps(nextProps) {
```
mperrotti: `componentWillReceiveProps` is being deprecated. Could we use [getDerivedStateFromProps](https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops) instead?

GeorgiSTodorov: Thanks for the feedback, I will take a look.

GeorgiSTodorov: I couldn't get it working with `getDerivedStateFromProps`. The problem is that the advanced section does not receive the new props, unless something is changed, so we can't close the settings until we click something.

Peek 2019-04-22 14-51

`getDerivedStateFromProps` is also not recommended - https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html .

GeorgiSTodorov: I am not sure it's possible to do it without refactoring every component in the chain - `CreateServiceModalForm`, `GeneralServiceFormSection` and `AdvancedSection`.

GeorgiSTodorov: I think I figured it out.

plugins/services/src/js/components/forms/GeneralServiceFormSection.js
```diff
@@ -92,7 +92,10 @@ class GeneralServiceFormSection extends Component {
     const initialIsExpanded = this.shouldShowAdvancedOptions();
 
     return (
-      <AdvancedSection initialIsExpanded={initialIsExpanded}>
+      <AdvancedSection
+        initialIsExpanded={initialIsExpanded}
+        shouldExpand={expandAdvancedSettings}
```
nLight: Why can't we reuse `initialIsExpanded` and need separate `shouldExpand`?

GeorgiSTodorov: Good question. I think it will be confusing to use `initialIsExpanded` for non-initial expansion.

plugins/services/src/js/components/modals/CreateServiceModal.js
```diff
@@ -491,6 +497,32 @@ class CreateServiceModal extends Component {
     this.setState({ isPending: true });
   }
 
+  setShouldExpand() {
```
nLight: this is a very broad name for `Advanced section of the General Config Tab`. Having `shouldExpand` in this context implies that the method controls whether the whole `CreateServiceModal` should expand.

plugins/services/src/js/components/forms/GeneralServiceFormSection.js
nLight: Having two props that do the same thing is less confusing?

plugins/services/src/js/components/modals/CreateServiceModal.js
GeorgiSTodorov: Good point, I will change the name of this function.

plugins/services/src/js/components/forms/GeneralServiceFormSection.js
GeorgiSTodorov: They both expand it, but in different scenarios.

nLight: How those scenarios are different? Can we converge to one prop that controls whether the section is expanded or not?

GeorgiSTodorov: I will try to do it with one prop. Thank you for the helpful feedback.

GeorgiSTodorov:
Pushed @nLight 's suggestions, currently working on @mperrotti 's request.

GeorgiSTodorov:
I spend a lot of time trying to fix a bug thinking that I introduced it, but it also happens on master.

src/js/components/form/AdvancedSection.js
mperrotti: @GeorgiSTodorov - If it's a lot of work to avoid using `componentWillReceiveProps` here, I'll still approve because it's not deprecated until React 17.

I'll check back tomorrow to see if you pushed up a change :)

GeorgiSTodorov: I will try some more, but it's definitely tricky. If I don't do it today, I will keep it in mind for my next spike day.

plugins/services/src/js/components/forms/GeneralServiceFormSection.js
GeorgiSTodorov: I think if we want to use `getDerivedStateFromProps` like @mperrotti suggested, we will need two separate props. Otherwise, as soon as we change anything in the form, `initialIsExpanded` becomes true and we can't close the advanced settings.

nLight: We can't close them until we fix the error, right?

GeorgiSTodorov: You mean that we shouldn't we able to?

GeorgiSTodorov:
@natmegs , I put it in `CreateServiceModal`, because I want it to run on clicking the review button.

plugins/services/src/js/components/modals/CreateServiceModal.js
```diff
@@ -491,6 +497,32 @@ class CreateServiceModal extends Component {
     this.setState({ isPending: true });
   }
 
+  getExpandAdvancedSettings() {
+    // If we have errors inside the advanced options, expand the options to allow the user to see the errors.
+    const errors = this.anyCriticalFormErrors();
+
+    if (errors.length > 0) {
```
nLight: `anyCriticalFormErrors` used to return boolean, right? If we change what it returns we should also change the name and JSDoc

plugins/services/src/js/components/modals/CreateServiceModal.js
```diff
@@ -491,6 +497,32 @@ class CreateServiceModal extends Component {
     this.setState({ isPending: true });
   }
 
+  getExpandAdvancedSettings() {
```
nLight: Why should the Modal know about the GeneralSection behaviour? We're passing down both `errors` and `serviceSpec` to `<CreateServiceModalForm />` would it make sense to move this logic all the way into GeneralServiceFormSection?

plugins/services/src/js/components/forms/GeneralServiceFormSection.js
nLight: once we move the expand on error logic here we not going to need the second prop.

plugins/services/src/js/components/modals/CreateServiceModal.js
GeorgiSTodorov: I made it this way to run this code on clicking the review and run button.

plugins/services/src/js/components/modals/CreateServiceModal.js
GeorgiSTodorov: I changed the name, I can't find the JSDoc for this file.

GeorgiSTodorov: Done.

natmegs:
@GeorgiSTodorov that check seems to only depend on the form errors and spec? I think this check could be moved to the general section which uses these props anyway.

GeorgiSTodorov:
Correct me if I am wrong, but I think that this expanding needs to happen as soon as the review and run button is clicked and for that we cannot have it in the general form section.

plugins/services/src/js/components/modals/CreateServiceModal.js
```diff
@@ -491,21 +497,45 @@ class CreateServiceModal extends Component {
...
+            error.path.includes("container") &&
+            error.path.includes("docker") &&
+            error.path.includes("image"))
+      );
+
+      return shouldExpand;
+    }
+
+    return false;
+  }
```
pierrebeitz: what do you think of 
1. omitting the `error.length > 0`-check (as the result from the following code will be the same and we don't expect performance issues with a one-time-check.
2. using the complete path instead of checking with an `includes`, as that might break in unexpected ways more easily.

i played around a bit and arrived at the following. it does not check for `container.type === "DOCKER"`, as i thought that this is a dependency of the error occurring at all. that might be wrong though.

```typescript
  getExpandAdvancedSettings() {
    // If we have errors inside the advanced options, expand the options to allow the user to see the errors.
    return this.criticalFormErrors().some(
      ({ path }) =>
        isEqual(path, ["gpus"]) ||
        isEqual(path, ["disk"]) ||
        isEqual(path, ["container", "docker", "image"])
    );
  }
```

plugins/services/src/js/components/modals/CreateServiceModalForm.js
```diff
@@ -201,6 +201,16 @@ class CreateServiceModalForm extends Component {
...
     }
 
+    if (nextProps.expandAdvancedSettings) {
+      nextProps.resetExpandAdvancedSettings();
+    }
+
+    if (
+      nextProps.expandAdvancedSettings !== this.props.expandAdvancedSettings
+    ) {
+      return true;
```
pierrebeitz: can we get rid of `shouldComponentUpdate` here?
i don't see this being a performance issue, thus being a good spot to reduce complexity.

plugins/services/src/js/components/modals/CreateServiceModalForm.js
```diff
@@ -201,6 +201,16 @@ class CreateServiceModalForm extends Component {
       return true;
     }
 
+    if (nextProps.expandAdvancedSettings) {
+      nextProps.resetExpandAdvancedSettings();
```
pierrebeitz: to me kicking off a side effect in `shouldComponentUpdate` (which has the only purpose of figuring out whether the component should be rerendered or not) does not feel good.

plugins/services/src/js/components/modals/CreateServiceModalForm.js
GeorgiSTodorov: There might be a way to remove it, but it was not added in this PR, so I would say removing it is out of scope here.

GeorgiSTodorov:
@pierrebeitz , thank you for the review. You are right about the design of `AdvancedSection` being peculiar.

plugins/services/src/js/components/modals/CreateServiceModal.js
GeorgiSTodorov: I will try this.

plugins/services/src/js/components/modals/CreateServiceModalForm.js
GeorgiSTodorov: Where do you prefer to have it in this case?

plugins/services/src/js/components/modals/CreateServiceModal.js
GeorgiSTodorov: We do need the check for Docker, you are right about the rest. Thanks.

GeorgiSTodorov:
@pierrebeitz 

> Also it feels a little weird that we manipulate a property (expandAdvancedSettings) to open the section when it has errors, but close it via a function (resetExpandAdvancedSettings). Have you considered using functions for both?

`expandAdvancedSettings` is basically the function `getExpandAdvancedSettings`. Do you mean that we should pass the function?

plugins/services/src/js/components/modals/CreateServiceModalForm.js
GeorgiSTodorov: I moved it to `componentDidUpdate`.

PR #3826, opened 2019-04-15
DCOS-51384 name unknownvolumes parser

TattdCodeMonkey:
Gave UnknownVolumesParser function a unique name so that it could be
replaced by a plugin.


brandonc:
Can you elaborate on how changing the export can allow it to be replaced by a plugin? Sorry I'm unclear

TattdCodeMonkey:
@brandonc the parsers are passed through the plugin `Hooks.applyFilter` as a list of functions. If it has a unique name I can remove it from the list and add a replacement.

GeorgiSTodorov:
The commit message is missing a jira number.

TattdCodeMonkey:
This is a subtask and doesn’t close a JIRA in its own.

PR #3828, opened 2019-04-16
DCOS-45215 - Fix Form Colons

pierrebeitz:
factored out separate header-sections that are independent of the lines to
follow after. this way we have a definitive border to measure against: the top
of a form row. note that the bottom of a row is not suitable as it can vary
depending on validation states (see jobs run config with duplicate keys) (@mperrotti).

this has been done in preparation of the new "environment" tab in the jobs form
that looks very broken if this issue in not addressed.

i think it is worth exploring whether we can factor this repeating pattern into a component at some point in the future (e.g. account for it when the next colon-form-section is planned).

## Testing

Make sure to check out pb/DCOS-45215-fix-form-colon in ee as well. 

Have a look at form-sections with the colon. you should be able to grep the code for 'form-colon' to see (all?) the partials containing it.

Please note that secrets (in service/create) are EE and fixed in a separate EE-PR. They will still look broken. Every other occurrence should be fixed.

## Screenshots

**Before** (this is a rather old screenshot taken from JIRA)

before

**After**

OS Enterprise 2019-04-16 18-11-48

## Tradeoffs

`position: relative;` is introduced on `.form-row`, which i don't really like, as it feels rather invasive. a separate approach would be to go even further and move the colons within the markup and have flexbox align the colons - that's not too much work, so if you prefer that solution, please leave an according comment.

Closes DCOS-45215


pierrebeitz:
@natmegs good catch! put up a PR for EE.

plugins/jobs/src/js/components/form/ParametersSection.tsx
```diff
@@ -117,10 +93,36 @@ class ParametersSection extends React.Component<
...
+            <FormGroup className="column-6 short-bottom">
+              <FieldLabel>
+                <FormGroupHeading>
+                  <FormGroupHeadingContent primary={true}>
+                    <Trans>Parameter Value</Trans>
+                  </FormGroupHeadingContent>
+                </FormGroupHeading>
+              </FieldLabel>
+            </FormGroup>
+            <div style={{ visibility: "hidden", height: "0" }}>
```
mperrotti: I think we can find a way to get rid of this

mperrotti:
The "Parameter Value" label is misaligned with the field
Screen Shot 2019-04-17 at 9 00 15 PM

mperrotti:
>  note that the bottom of a row is not suitable as it can vary
depending on validation states

It's too bad we couldn't position the colon based off the bottom. Good catch though!

---

> i think it is worth exploring whether we can factor this repeating pattern into a component at some point in the future (e.g. account for it when the next colon-form-section is planned).

I agree we should turn this pattern into a component

---

Let me know if there's anything you want to pair on with this ticket. I love solving tricky layout issues :)

GeorgiSTodorov:
I can reproduce @mperrotti 's observation.

pierrebeitz:
i used the wrong button as a spacer in this particular form. 🤦‍♂️ 

please re-review. i rebase onto master. the current change is:

```diff
@ plugins/jobs/src/js/components/form/ParametersSection.tsx:121 @ class ParametersSection extends React.Component<
              </FieldLabel>
            </FormGroup>
            <div style={{ visibility: "hidden", height: "0" }}>
-              <AddButton />
+              <DeleteRowButton />
            </div>
          </FormRow>
        ) : null}
```

pierrebeitz:
@GeorgiSTodorov can not reproduce. are you sure, you have the latest CSS? looks like this change did not get compiled: https://github.com/dcos/dcos-ui/pull/3828/files#diff-88ae5e2666ed29dcfb6f6e1faf5a5c6c.

GeorgiSTodorov:
It looks like you pushed while I was testing.
Screenshot from 2019-04-23 13-43-13

pierrebeitz:
just double-checked: it was not broken in f517b42. so the issue is most likely about about how the change was tested. let's please slack/pair to figure out what information i need to add to the testing instructions if/when you have a similar issue! 💜

mperrotti:
I was having the same issue Georgi was. The browser was getting the old styles even though `src/styles/components/form-elements/styles.less` showed the correct styles.

I was able to get it working after making edits to the `.less` file, rebuilding, and checking out the related branch from plugins-private

GeorgiSTodorov:
CI is having issues again.

PR #3830, opened 2019-04-16
[DCOS-51776] chore: add secretsDefaultStore to app config

brandonc:
## Overview

This information was moved to the secrets plugin but will be needed by multiple plugins (Ex: service accounts can create associated secrets).

We already include another secrets config key here but both will remain unused in the app. All secrets seem to be store in the default secret store.


brandonc:
I've got a better idea.

PR #3832, opened 2019-04-17
DCOS-5080: notify users if connection drops

mperrotti:
Note to reviewers: the design team changed their mind - they wanted to show a page banner instead of a toast. See comments on the Jira ticket.

## Testing
I recommend testing with Chrome so you can toggle the connection in the browser instead of turning your WiFi on and off

- Open the dev tools and go to the Network tab
- Toggle the "Offline" checkbox in the top of the Network tab

A warning banner should appear when "Offline" is checked, and disappear when it's not checked

## Trade-offs
If you're in a state where we show the `LicenseBanner` and you go offline, both banners will be displayed.
This could have been avoided by using extension-kid to create a Banner service (like the Toast service), but that would be well out of the scope of a 2-point ticket.

## Screenshots
OfflineBanner

### DIFF
--- a/locale/en/messages.json
+++ b/locale/en/messages.json
@@ -1113,6 +1113,7 @@
   "You have several protocol options. <0>More Information</0>.": "",
   "You must specify a command, an argument or a container with an image.": "",
   "You need at least one package repository with some packages to be able to install packages. For more <0>information on repositories</0>.": "",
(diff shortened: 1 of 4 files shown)


mperrotti:
Wow @brandonc - thanks! I'll give it a try

PR #3833, opened 2019-04-17
🤫 feat: parameterized root marathon

nLight:
## Outstanding issues:

- [ ] Images are broken (i.e. Mesosphere Logo)
- [ ] Update service tries calling the relative path -> Config.rootUrl most probably should be set to "/" or we rather should disable the update plugin
- [ ] Disable Catalog
- [ ] Some kind of indication that UI is running on MOM
---

Introducing 2 global variables to parameterize root Marathon.
With the change it will be possible directing the UI to a MoM (Marathon on Marathon) so that operators could operate MoMs easily.

## Testing

1. Launch `marathon` from the Catalog usually it goes under name `marathon-user`
2. Set `window.MARATHON_NAME = "marathon-user";`
3. Set `window.MARATHON_API_PREFIX = "/service/marathon-user/v2";`
4. Check the UI, it should be _empty_ unless something is running on the MoM already
5. Launch an app and a pod
6. See screenshots

## Trade-offs

Oh well, global variables

## Dependencies

None

## Screenshots
Left screenshot: docs-ui pointing to the root Marathon
Right screenshot: dcos-ui pointing to mom
Screenshot 2019-04-17 at 10 43 12


src/js/config/Config.ts
```diff
@@ -56,7 +56,7 @@ let Config: IConfiguration = {
   historyLength: 31,
   historyServer: "",
   fullProductName: "Mesosphere DC/OS",
-  marathonAPIPrefix: "/service/marathon/v2",
+  marathonAPIPrefix: global.MARATHON_API_PREFIX,
```
nLight: @pierrebeitz maybe you could help me silencing TS lint. It complains `MARATHON_API_PREFIX` is not defined on global #no_pressure

pierrebeitz: having discussed this exact issue a lot in the past, i'd propose to add a special file to augment builtins. this way there's a single place to look up the "imperfections" rather than relying on `(global as any).SOME_BLORGH` or extending those modules on a per-file basis.

it could look something like the following (although i did not test this).
```
// e.g. src/typings/global.d.ts
declare namespace NodeJS {
  interface Global {
    MARATHON_API_PREFIX: string;
  }
}
```

GeorgiSTodorov:
The UI breaks for me.
Screenshot from 2019-04-19 11-28-51
Screenshot from 2019-04-19 11-29-39

GeorgiSTodorov:
I get this on the first step.
Screenshot from 2019-04-25 11-40-37

src/js/config/Config.ts
nLight: Thanks! Had to adjust tslint config for that to work 😬

PR #3837, opened 2019-04-18
DCOS-46178: update dcos-cli in system tests

nLight:
CLI team dropped `.dcos.toml` config file support a while back and we've been using 1.11 cli for our tests

Closes DCOS-46178

## Testing

System tests should pass


nLight:
@juliangieseke looks like we can finally upgrade. Tests pass but flaky 😩

PR #3840, opened 2019-04-18
DCOS-12400: highlight tabs with errors

mperrotti:
## Testing
1. Go to the create service form
2. Enter invalid data

Error badges should appear in the tabs, and the sum of the numbers in the badges should equal the sum of the number of errors in the error alert banner.

## Screenshots
Screen Shot 2019-04-18 at 6 13 56 PM


plugins/services/src/js/components/forms/ContainerServiceFormSection.js
```diff
@@ -61,7 +61,8 @@ class ContainerServiceFormSection extends Component {
           target="_blank"
         >
           More information
-        </a>.
+        </a>
+        .
```
GeorgiSTodorov: I think these will also change the translations.

GeorgiSTodorov:
For the service
```
{
  "portDefinitions": [],
  "container": {
    "type": "MESOS",
    "volumes": []
  },
  "requirePorts": false,
  "networks": [],
  "healthChecks": [],
  "fetch": [],
  "constraints": [
    [
      "@zone",
      "GROUP_BY",
      "0"
    ]
  ],
  "cpus": -1,
  "mem": -1,
  "instances": -1,
  "disk": -1,
  "gpus": -1
}
```
the error count is wrong - it shows 5 errors, even though there are 7.

DanielMSchmidt:
I got another one: 	When I add a value of an environment variable, but not the key: 
Screenshot 2019-04-23 at 15 30 39

mperrotti:
@DanielMSchmidt - we don't show an error in the error banner either, so I chose not to include that error in the badges either.

When a user submits a form with environment variables that have values but not keys, we just don't include those environment variables in the configuration.

mperrotti:
CI was green this morning. Now it's just getting stuck and timing out during the unit test portion of the 
"Build" step

GeorgiSTodorov:
I think this is a service side error, do we support badges for them?
Peek 2019-04-24 14-44

plugins/services/src/js/utils/CreateServiceModalFormUtil.js
```diff
@@ -38,6 +40,72 @@ const CreateServiceModalFormUtil = {
...
+   * @param {Array} pathTranslationRules - The path translation rules
+   * @param {object} i18n - lingui internationalization instance object
+   * @returns {Object} An object where where keys map to the service form tabs,
+   * and the values are an array of objects containing error data
+   */
+  getTopLevelTabErrors(errors, tabPathRegexes, pathTranslationRules, i18n) {
+    const errorPaths = errors.map(error => error.path.join("."));
+
+    return errorPaths.reduce((errorsByTabObj, pathString) => {
+      Object.keys(tabPathRegexes).forEach(tabId => {
```
GeorgiSTodorov: Is it possible for this to not have keys?

GeorgiSTodorov: I saw it's a constants file. Why do we not use it directly here?

mperrotti: I was originally using it directly, but I changed it to a parameter when I was writing tests. If we do something like this on other forms, we can move this to another util file and pass a different object of the same shape.

If you or any other reviewers feel strongly that it should be used directly, I think it's ok to change.

mperrotti:
> I think this is a service side error, do we support badges for them?

Yes. Any error that has a message in the `ErrorsAlert` should support badges. I fixed this one.

---

> If we have multiple containers and they have the same error, only the first container gets a badge, but I am not sure if changing this is possible.

It's possible if we either:
- change the error shown in the `ErrorsAlert` to read something like `container.0.name: must be defined`
- manually define which errors are grouped into one error message in the `ErrorsAlert`
- allow the number in each error badge be inconsistent with the number or error messages shown in the `ErrorsAlert`

None of these seem like great options, so I'm not sure what to do here. I'd love feedback from other reviewers on this 🙏

GeorgiSTodorov:
One more error that doesn't get a badge: If we enter "1" as command.
Screenshot from 2019-04-25 11-22-49

mperrotti:
>One more error that doesn't get a badge: If we enter "1" as command.

Entering "1" as a command doesn't come back as an error in `this.props.errors`

---

>And one more: for labels starting with whitespace.
The badge on "Environment" is actually for the label being a whitespace char. We just don't highlight the "Label" field as having an error in that case.

I can't get the first two errors in your screenshot to appear. Maybe you can show me how you got to that state?

GeorgiSTodorov:
@mperrotti , I got the two errors by entering "1" as command and create an environment variable with value, but no key (or maybe whitespace key).

mperrotti:
@GeorgiSTodorov - I tried that, but I still can't get it to appear. Maybe we can pair and you can see if I'm missing a step?

mperrotti:
I haven't been able to get that one too show up. All you're doing is
putting in a value and no key (or whitespace key), right?

natmegs:
I didn't initially realize how wonky the repro steps are. Basically if I create a job with an environment variable, edit it, and remove the key from the environment variable, I get this server error. Repro video here https://cl.ly/d17030dafcf8. Since it's such an edge case I would be ok with it being a non-blocker

mperrotti:
Thanks for the video! It's an easy fix :)

GeorgiSTodorov:
The environment variable error has a badge now, but not the command error.
Peek 2019-05-02 12-21

mperrotti:
@GeorgiSTodorov - I'm only seeing that error if I enter a single character (any character - letter, number, symbol, whitespace, etc), and the error comes back without a `path`.

This is confusing because a single character is not an empty string. `appConfig` state is still updated with that single character as a value for `cmd`, and the JSON editor shows a value for `cmd` as well.

This happens on `master` as well, so I'd consider this to be a bug separate from this ticket.

PR #3842, opened 2019-04-19
do not merge: testing integration test fails in CI

mperrotti:


mperrotti:
Able to reproduce test failures locally now. No need for this PR anymore

PR #3843, opened 2019-04-23
ci: fix race condition due to entry point

DanielMSchmidt:
Errors like this can occur:
https://jenkins.mesosphere.com/service/jenkins/blue/organizations/jenkins/frontend%2Fdcos-ui-ee-pipeline/detail/PR-972/7

## Testing

We already have this on EE, but never added it to OSS as well.


juliangieseke:
hm, changing OSS Jenkinsfile does not affect EE pipeline, also we have removed entrypoints from our dockerfiles… the mentioned error does not seem to be related to this PR 🤔

DanielMSchmidt:
Yeah, it is not supposed to do, we have this flag already in EE, that's all :)

I got the info to add this by devprod, but I wonder why we need to overwrite the entrypoint as well.

pierrebeitz:
from the description and commit i neither understand the problem nor solution. 🙈

GeorgiSTodorov:
CI was red, merging master to see if it makes a difference.

DanielMSchmidt:
Sorry @pierrebeitz, I meant to write a little more, my bad.

I was (blindly) following [this recommendation](https://mesosphere.slack.com/archives/C5U03P5T6/p1556022199118600?thread_ts=1556012906.118000&cid=C5U03P5T6), but judging from the referred date this might have been a problem on the old dockerfile. I'll check with them again to see if this can really be the fix. Closing until then.

PR #3844, opened 2019-04-23
fix(jest): add --forceExit to resolve issue with jest hanging

TattdCodeMonkey:
Adding `--forceExit` flag when we call `jest` in `npm test` to try and ensure jest exits when all tests are completed. We are seeing some issues in CI where jest will failed to exit. This flag will hopefully resolve this issue.

## Testing

```
$ npm run test
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -223,7 +223,7 @@
     "test:system": "PLUGINS_PATH=\"${npm_config_externalplugins}\" ./scripts/run-system-tests",
     "test:validate": "PLUGINS_PATH=\"${npm_config_externalplugins}\" ./scripts/validate-tests",
     "test:watch": "npm test -- --watch",
-    "test": "lingui compile && NODE_PATH=node_modules jest --config=jest.config.js --no-cache",
+    "test": "lingui compile && NODE_PATH=node_modules jest --config=jest.config.js --no-cache --forceExit --detectOpenHandles",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -223,7 +223,7 @@
     "test:system": "PLUGINS_PATH=\"${npm_config_externalplugins}\" ./scripts/run-system-tests",
     "test:validate": "PLUGINS_PATH=\"${npm_config_externalplugins}\" ./scripts/validate-tests",
     "test:watch": "npm test -- --watch",
-    "test": "lingui compile && NODE_PATH=node_modules jest --config=jest.config.js --no-cache",
+    "test": "lingui compile && NODE_PATH=node_modules jest --config=jest.config.js --no-cache --forceExit --detectOpenHandles",
```
brandonc: `--detectOpenHandles` seems like more of a debug option. Does it reveal anything about jest not exiting?

mperrotti:
Works as expected, and all tests pass when I check out `rnorris/upgrade-jest` from plugins-private

package.json
TattdCodeMonkey: Using `--forceExit` without it jest will print a message suggesting you use it, so I added it. Assuming it will give some info if jest has to forceExit.

TattdCodeMonkey:
@natmegs it should be seamless locally, I've only seen jest not exit on CI which is what this will hopefully resolve.

PR #3848, opened 2019-04-25
DCOS-48818: add file input support

DanielMSchmidt:


src/js/components/form/FieldFile.js
```diff
@@ -0,0 +1,25 @@
+import React from "react";
```
TattdCodeMonkey: could this be a .tsx?

DanielMSchmidt:
Merged this with an older green run in place and aborted the current CI run.

PR #3850, opened 2019-04-26
Add types for marathon

DanielMSchmidt:
## Testing

- It's dead code for now, no testing needed. If one knows a good place to use these types initially, I'm all in for it 👍 

## Trade-offs

- These types are auto-generated from the marathon raml definitions using https://github.com/DanielMSchmidt/dcos-api-schemas


DanielMSchmidt:
We can use them everywhere we get a response back from Marathon. These parts are not moved to typescript yet, so this is an enabler to move them over

GeorgiSTodorov:
@DanielMSchmidt , sounds good.

DanielMSchmidt:
@pierrebeitz I will show the way they were generated, but I think we need to think about them more as a one time generation

PR #3851, opened 2019-04-29
Improve typescript integration with externalplugins

brandonc:
## Overview

This adds package support for what seems to be the best method of integrating our dev tooling with external plugins: symlinking ../dcos-ui-plugins-private

Here are the changes:

1. adds `npm run util:env:link-externalplugins` and `npm run util:env:unlink-externalplugins` which simply creates and removes the symlink.

2. the symlink is now used to extract lingui messages (a temporary one was created before), and without it, an error message is displayed.  I think this is a good simplification.

3. the symlink is used as a typescript "includes" so that modules can be resolves and other tooling works with our own settings

4. vscode workspace settings have been added to allow the symlink to be ignored by git, but not ignored by the editor. I had to add new entries to search exclude in order to reach some parity with the default setting.

## Testing

- `npm run util:env:link-externalplugins` before taking the next step...
- `npm run lint:ts` to see typescript errors locally. This does not affect CI yet.
- `npm run util:lingui:extract-with-plugins` to ensure lingui extraction still works as expected. This command should also link external plugins as well
- Try out your editor integrations and add appropriate workspace config if necessary

## Trade-offs

- Not a new tradeoff, but open source contributions containing new translation messages are basically not allowed since we require you to clone plugins-private before extraction. This is not ideal.

- If your editor depends on .gitignore to hide files, it may require reconfiguration.

#### Typescript errors!

Screen Shot 2019-04-29 at 11 58 52 AM

#### Editor integration!

Screen Shot 2019-04-29 at 11 56 29 AM

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,10 +18,9 @@ tmp/
 jest/test-results
 webpack/proxy.dev.js
 src/resources/raml
-.vscode/
 .chrome/
 .envrc
 .env
 locale/_build/
 locale/*/messages.js
-.dcos-ui-plugins-private
+plugins-ee
--- a/.linguirc
+++ b/.linguirc
@@ -7,7 +7,7 @@
   "srcPathDirs": [
       "./src",
       "./plugins",
-      "./.dcos-ui-plugins-private"
+      "./plugins-ee"
   ],
   "srcPathIgnorePatterns": [
       "node_modules/",
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,7 @@
+{
+  "search.exclude": {
+    "**/_build": true,
+    "cypress": true
+  },
+  "search.useIgnoreFiles": false
+}
--- a/package.json
+++ b/package.json
@@ -222,8 +222,10 @@
     "test:validate": "PLUGINS_PATH=\"${npm_config_externalplugins}\" ./scripts/validate-tests",
     "test:watch": "npm test -- --watch",
     "test": "lingui compile && NODE_PATH=node_modules jest --config=jest.config.js --no-cache",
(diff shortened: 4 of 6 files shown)


pierrebeitz:
from what i get from the EE-Jenkinsfile this would not enable type checking on CI, as we're not calling `npm run util:env:link-externalplugins` along the way. We'll either need a separate PR on EE for that or should stick to the naming conventions used in EE.

brandonc:
@pierrebeitz yes; I wanted to take that action as a second step so that we didn't block everything by breaking CI until all the type fixes were in

brandonc:
@natmegs hmm- they do show up for me. I took a screenshot of my console after running lint:ts. You ran it after running `npm run util:env:link-externalplugins`, right?

natmegs:
Yes I see the errors in the console, and I see the TS errors in the editor, but I don't see the tslint errors from the console in my editor. For ex, in the screenshot the console shows linter errors in `SecretsSelect` but I don't see them in the editor when I open that file. Not a blocker though
Screen Shot 2019-04-29 at 2 29 18 PM

brandonc:
@natmegs I made a mistake. vscode is showing the output of tsc (`npx typescript --build tsconfig.json`)

I'll try to debug tslint warnings

brandonc:
@natmegs to get tslint warnings in vscode, I had to install tslint and typescript globally. Not sure if that affects other tooling yet so maybe try that and re-test?

Screen Shot 2019-04-29 at 3 43 25 PM

brandonc:
@natmegs my guess is that tslint extension can't "find" the node_modules version when it walks up a directory and encounters a symlink.

natmegs:
I'm unable to see the errors locally even with typescript and tslint installed globally, but I suspect this is an issue with my local set up. Being able to see the errors in the console is sufficient for me for now!

pierrebeitz:
a hint in the readme would make for a good follow up!

PR #3853, opened 2019-04-30
chore: more accurate type interfaces

brandonc:
These changes are needed in order to fix typescript issues in the plugins-private companion PR


brandonc:
Companion PR: https://github.com/mesosphere/dcos-ui-plugins-private/pull/991

PR #3854, opened 2019-05-02
DCOS-48339: volumes form section

natmegs:
Add the volumes section to the jobs form.

## Testing

- Open jobs form, go to Volumes tab
- Add volumes
- Check JSON editor to make sure they appear as expected
- Try to submit with invalid volumes (missing fields, wrong field type, etc)

## Trade-offs

N/A

## Dependencies

DCOS-48339 is not complete without EE changes.

## Screenshots

Screen Shot 2019-05-01 at 5 01 21 PM


GeorgiSTodorov:
- When I change the mode in the JSON, the form still shows "Read and Write" mode.
Screenshot from 2019-05-02 12-47-34
Peek 2019-05-02 12-50

- Errors persist.

- Like in the Placement PR, if I introduce a JSON error and later fix it, I can no longer submit the form.
Peek 2019-05-02 13-04

- Error for mode doesn't disappear.
Peek 2019-05-02 13-02

natmegs:
@GeorgiSTodorov thanks for the prompt review!! I changed the mode input so that it will show `Select ...` if one of the RO/RW options are not provided. Any errors that are persisting are actually from the server - the reason for that is we cannot revalidate a rule that our client side validation didn't catch. To mitigate the issue, I've added more validation to check that `mode` is one of RO/RW. 

The invalid JSON issue actually affects the whole form... we do not receive an error change event from the JSON editor in this particular situation. It is definitely something that should be addressed but I would say it is out of scope of this particular PR.

plugins/jobs/src/js/components/form/VolumesFormSection.tsx
```diff
@@ -0,0 +1,220 @@
...
+          (volume as JobVolume).mode === VolumeMode.RO ||
+          (volume as JobVolume).mode === VolumeMode.RW
+        )
+      ) {
+        // If mode is changed to something not in the list (for ex: in the JSON editor),
+        // show the default "Select ..." option
+        mode = "";
+      }
+
+      if (index === 0) {
```
brandonc: This could be > 0 if the first volumes were SecretVolumes, right?

plugins/jobs/src/js/components/form/VolumesFormSection.tsx
```diff
@@ -0,0 +1,220 @@
...
+  getVolumesLines() {
+    const {
+      formData: { volumes = [] },
+      onRemoveItem,
+      errors,
+      showErrors
+    } = this.props;
+
+    return volumes.map((volume: JobVolume | SecretVolume, index: number) => {
+      if (isFBS(volume)) {
```
brandonc: Does this indicate SecretVolume? I think it must since everything gets cast as JobVolumn below. I wonder if it would be better to filter out all the SecretVolumns before the map and then avoid all the casting later? Your index === 0 thing would also not be a bug (if I understand this correctly)

```
volumes.filter((volume: JobVolume | SecretVolume) => isFBS(volume)
       .map((volume: JobVolume) => { ... }
```

plugins/jobs/src/js/components/form/VolumesFormSection.tsx
```diff
@@ -0,0 +1,220 @@
...
+        errors
+      );
+      const hostPathErrors = getFieldError(
+        `job.run.volumes.${index}.hostPath`,
+        errors
+      );
+      const modeErrors = getFieldError(`job.run.volumes.${index}.mode`, errors);
+
+      return (
+        <FormRow key={index}>
```
brandonc: index as key is probably insufficient since it's not permanent and unique. Is there another ID you can use?

plugins/jobs/src/js/components/form/VolumesFormSection.tsx
natmegs: If I filter the list the error indices may not match up properly in the UI - but I changed the return type on `isFBS` so that the type casting is not necessary anymore!

TattdCodeMonkey:
@natmegs When I try to type into `Container Path` it loses focus after every key press.

natmegs:
@TattdCodeMonkey oof good catch! Made a bad choice for a `key` on the form row - it should be fixed now.

natmegs:
@Poltergeist I appreciate the thoughtful review! For 1) and 2), I think they are definitely worth addressing but I think these are out of scope for this PR. These all involve shared components that are used throughout the jobs and services forms (and likely other places too). 
For 3), my concern here is it would make validation difficult. For example, if someone adds a volume row but doesn't fill it out, this should be ignored. But if there is a default mode value, how do we know if it's truly an empty row or if the user chose it intentionally and we should be giving them a validation message to fill in host & container paths?

PR #3858, opened 2019-05-02
[DCOS-53297] fix(ui-update): parse build version in data-layer

TattdCodeMonkey:
Updated the ui-update data-layer to parse the build version from string
provided instead of relying on semver.coerce. We currently use the
template of `<branch>+<build_version>(+<build_hash>)?` when branch is
`master` coercing the value gives use the build version, but if the
build is made from a release branch, `1.13` then `1.13` is returned from
semver.coerce instead of our desired build verions after the first `+`.

I added a `parseVersion` utility to the ui-update data-layer to handle
getting the build version from the string assuming the template is
followed. Unit tests for the utility and data-layer were added as well.

- Closes DCOS-53297
- Closes DCOS-53295

## Testing

- running locally update `src/index.html` line 34 from `window.DCOS_UI_VERSION = "0.0.0-dev+semantic-release";` to `window.DCOS_UI_VERSION = "1.13+v2.78.2";`
- navigate to UI-Setting page
- Verify `Installed Version` is `2.78.2` and not `1.13`

## Screenshots

Before:
image
After:
image


src/js/data/ui-update/utils.ts
```diff
@@ -0,0 +1,18 @@
+const LOCAL_DEV_VERSION = "0.0.0-dev+semantic-release";
+
+export function parseVersion(version: string): string {
+  if (version === LOCAL_DEV_VERSION) {
+    return "0.0.0";
+  }
+  if (!version) {
```
GeorgiSTodorov: Shouldn't we make this check first?

src/js/data/ui-update/utils.ts
```diff
@@ -0,0 +1,18 @@
+const LOCAL_DEV_VERSION = "0.0.0-dev+semantic-release";
```
pierrebeitz: do you think it would make sense to merge this with `plugins/services/src/js/utils/Version.ts` some time?

PR #3860, opened 2019-05-02
DCOS-53305: Nodes table progress bars are being colored inconsistently

mperrotti:
## Testing

1. In `NodesTableContainer.js`, pass a large NodesList to the `NodesTable` hosts prop. I've created a gist with the NodesList from the mixed workload test: https://gist.github.com/mperrotti/15d2e299ddfd093fca401c8f4813cb83
2. Go to the Nodes table and ensure that all of the progress bar colors are the same for each column

## Trade-offs
Adds `fast-memoize` dependency.

This is a TINY library, but if anybody has an issue with this, there's an alternative fix: explicitly pass a `className` prop to `<ProgressBar>` in each of the nodes table columns, and use the value of the classname prop instead of whatever comes back from `memoizedGetProgressBarData`

## Screenshots
Before:
Screenshot 2019-05-02 at 13 12 31

After:
Screen Shot 2019-05-02 at 4 53 29 PM


pierrebeitz:
sorry for adding noise here, but i noticed that we're about to remove the last "lodash"-imports and just could not resist the urge to remove them. 

i also found a third way of fixing the problem, which seems to have occurred due to lodash using only the first parameter of the function to memoize to determine the caching-key. by providing a custom "resolver"-function (`[].join`) we now get around that.

as always: feel free to just throw away my commit if you feel like that saves trouble. i would try to get it in during a spike day then.

finally some stats to convince you that this might be worth having:

**before**


before_lodash

**after**


after_lodash

mperrotti:
Thanks @pierrebeitz - that's a much better solution!

PR #3863, opened 2019-05-03
[DCOS-53141] Jobs File Based Secrets Dependencies

brandonc:
## Overview

Some type additions and clarifying renames that will be needed for Jobs File Based Secrets.

Should not change any behavior.

## Important

Hold merge until companion PR is ready

### DIFF
--- a/plugins/jobs/src/js/components/form/helpers/JobFormData.ts
+++ b/plugins/jobs/src/js/components/form/helpers/JobFormData.ts
@@ -64,6 +64,13 @@ export interface JobSpec {
 export type EnvModel = Array<[string, string]>;
 export type SecretModel = Array<[string, string, string]>;
 
+export interface JobSecretExposure {
+  exposureType: "" | "env" | "file";
+  exposureValue: string;
+  key: string;
+  secretPath: string;
+}
+
 export interface FormOutput {
   jobId: string;
   description?: string;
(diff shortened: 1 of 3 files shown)


plugins/jobs/src/js/components/form/helpers/JobFormData.ts
```diff
@@ -57,13 +57,20 @@ export enum Container {
 export interface JobSpec {
   cmdOnly: boolean;
   container?: Container | null;
-  job: JobFormData;
+  job: JobSpecData;
   schedule?: JobSchedule;
 }
 
 export type EnvModel = Array<[string, string]>;
 export type SecretModel = Array<[string, string, string]>;
```
natmegs: Can we delete this now?

brandonc: Removed it in a follow up commit

plugins/jobs/src/js/components/form/helpers/JobFormData.ts
```diff
@@ -26,7 +26,7 @@ export interface JobRun<Env, Secrets> {
   secrets?: Secrets;
 }
 
-export type JobFormData = Job<ArrayLabels, EnvModel, SecretModel>;
+export type JobSpecData = Job<ArrayLabels, EnvModel, JobSecretExposure>;
```
natmegs: I think we will need `JobSecretExposure[]` here

brandonc: 😭  of course

PR #3864, opened 2019-05-06
DCOS-52749 allow to route packages without a version

pierrebeitz:
this aims to solve the problem what let to
https://jira.mesosphere.com/browse/DCOS-52749 without sticking to the solution
proposed by that very issue.

instead of adding a redirect, we make use of the cosmos-api always responding
with data for `latest` if no version was given. as we previously assumed that a
version had to be specified always, that assumption has been factored out, only
sending the `version` query parameter if it is actually set. this evades the need for a redirect altogether and looks like a cleaner solution to me.

also prettier was updated lately and no one ran it over the whole codebase yet, so
a dot has to be moved to comply to our linting...

## Testing

go to http://localhost:4200/#/catalog/packages/kubernetes and see that nothing breaks, also try to launch that app and try every functionality that might rely on the version-information in that form. 

also of course try to think of other problems that are introduced by that approach. :)


src/js/pages/catalog/PackageDetailTab.js
```diff
@@ -289,7 +289,8 @@ class PackageDetailTab extends mixin(StoreMixin) {
               target="_blank"
             >
               documentation
-            </a>.
+            </a>
+            .
```
pierrebeitz: this change must be in here as else CI fails :(

src/js/pages/catalog/PackageDetailTab.js
```diff
@@ -157,7 +157,7 @@ class PackageDetailTab extends mixin(StoreMixin) {
     router.push(
       `/catalog/packages/${encodeURIComponent(
         params.packageName
-      )}/deploy?version=${location.query.version}`
+      )}/deploy?${qs.stringify({ version: location.query.version })}`
```
pierrebeitz: `qs` omits properties that are `undefined`.

src/js/events/CosmosPackagesActions.js
```diff
@@ -183,7 +183,10 @@ const CosmosPackagesActions = {
       },
       method: "POST",
       url: `${Config.rootUrl}${Config.cosmosAPIPrefix}/describe`,
-      data: JSON.stringify({ packageName, packageVersion }),
+      data: JSON.stringify({
+        packageName,
+        ...(packageVersion && { packageVersion })
```
pierrebeitz: this only adds the `packageVersion`-property if it is truthy, which is what we are looking for here.

i like this syntax and think it's concise, if you feel like it is confusing, we could add everything and filter unset stuff out, e.g. by introducing a dependency on our `Util`:

```
import Util from "../../utils/Util"

// ....
  data: Json.stringify(Util.filterEmptyValues({ packageName, packageVersion }))
// ...
```

brandonc: Syntax is not confusing after the initial read

PR #3865, opened 2019-05-06
DCOS-52747 Disable package installation until "manager" requirement is satisfied

pierrebeitz:
https://jira.mesosphere.com/browse/DCOS-52747

⚠️ depends on https://github.com/dcos/dcos-ui/pull/3864, you might want to get that merged first.

some packages (currently only kubernetes-cluster) have a `manager` that must be
installed before they can be used. currently the API restricts the installation
of kubernetes-cluster if kubernetes is not installed yet. you have to go through
the whole installation-process in the UI to afterwards be notified that stuff
did not work. as a remedy a check has been implemented that renders an infobox
with a nice warning as well as the reviewAndRun-button disabled - so now we
can't waste a lot of work in the first place when we don't know about
dependencies upfront.

## Review

The functionality has been implemented in two steps (commit). The first commit (i mean the one that starts with `refactor`) refactors `geReviewAndRunButtons -> renderReviewAndRunButton` in the `PackageDetailTab` as i can not deal with `let`s somehow... it also does some dummy-changes that cause huge whitespace-changes.

the second commit should contain all the logic related to dependency-checking. it should enable to get an overview of the approach taken.

## Testing

Try to install [kubernetes-cluster](http://localhost:4200/#/catalog/packages/kubernetes-cluster) on a cluster that does not have a kubernetes service running to see the warning. install kubernetes and confirm that the message disappears (it should disappear in realtime as soon as the kubernetes-task has the state "TASK_RUNNING").

To make the warning show up programatically you can make `hasUnresolvedDependency` return `true`.


src/js/pages/catalog/PackageDetailTab.js
```diff
@@ -117,6 +125,31 @@ class PackageDetailTab extends mixin(StoreMixin) {
...
+      this.onMesosStateChange
+    );
+  }
+
+  onMesosStateChange() {
+    this.setState({
+      runningPackageNames: {
+        state: "success",
+        data: (MesosStateStore.getLastMesosState().tasks || [])
+          .filter(_ => _.state === "TASK_RUNNING")
```
brandonc: Not blocking feedback but just thought I would share that this confused me for a sec. Usually i take "_" params to be unused, and "_" operands to be lodash

src/js/pages/catalog/PackageDetailTab.js
```diff
@@ -562,4 +634,19 @@ PackageDetailTab.contextTypes = {
...
 
+const dependency = cosmosPackage => (cosmosPackage.manager || {}).packageName;
+
+const unresolvedDependencyMessage = dependency => (
+  <div>
+    <Icon shape={SystemIcons.CircleInformation} size={iconSizeXs} />{" "}
+    <Trans>
+      This service cannot run without the "
+      <Link to={`/catalog/packages/${dependency}`}>
+        {StringUtil.capitalize(dependency)}
```
brandonc: Worth noting that this is not a foolproof method of capitalization -- Thinking of "Dcos" here

I think it's better to leave it unchanged since I mentioned this to design already and the package names are in lowercase in the catalog.

src/js/pages/catalog/PackageDetailTab.js
pierrebeitz: good point. when wrangling data i seemingly have developed a weird style of trying to hide stuff that's not really needed in my head.

we can definitely make that `task => task.state === "TASK_RUNNING"`.

src/js/pages/catalog/PackageDetailTab.js
brandonc: Also, I can't seem to get the Tooltip to dismiss. I thought these were supposed to be magically dismissed by ui-kit.

.eslintrc
```diff
@@ -55,13 +55,13 @@
...
     "comma-dangle": ["error", "never"],
     "compat/compat": "error",
+    "destructuring/in-methods-params": "off",
     "destructuring/in-params": "off",
     "destructuring/no-rename": "off",
     "eqeqeq": ["error", "smart"],
     "jsx-max-len/jsx-max-len": ["error", {"lineMaxLength": 80, "tabWidth": 2, "maxAttributesPerLine": 1}],
     "keyword-spacing": ["error"],
     "multiline-ternary": "off", // allow both multiline and single line
-    "newline-before-return": "error",
```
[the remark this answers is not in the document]
pierrebeitz: oh, i should have put this into an own commit.

it's because that lint bugs me every single day and we allow a return without a newline in front of it in typescript. will put that up for discussion tomorrow.

src/js/pages/catalog/PackageDetailTab.js
```diff
@@ -275,33 +308,68 @@ class PackageDetailTab extends mixin(StoreMixin) {
...
+          </a>
+          .
+        </Trans>
+      </div>
+    );
+  }
+
+  renderInstallButton({ tooltipContent, disabled }) {
+    return (
+      <Tooltip
```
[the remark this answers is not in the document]
pierrebeitz: we did so before, so i saw no evil here. but you're right in that we don't need to! 

i'll see whether i can come up with a solution that's not too bad regarding implementation cost.

src/js/pages/catalog/PackageDetailTab.js
mperrotti: @brandonc - this Tooltip is from `reactjs-components`.

I can't get it to dismiss either, but it's dismissing just fine everywhere else.

src/js/pages/catalog/PackageDetailTab.js
```diff
@@ -514,45 +568,63 @@ class PackageDetailTab extends mixin(StoreMixin) {
...
-                        : cosmosPackage.getIcons()["icon-large"]
-                    }
+
+        {!isLoading ? (
+          <div>
+            <div className="container">
+              {this.hasUnresolvedDependency(cosmosPackage) && (
+                <div className="infoBoxWrapper">
+                  <InfoBoxInline
+                    appearance="warning"
```
mperrotti: The designers wanted this to be the neutral (grey) InfoBox. You can get this by not passing an `appearance`.

src/js/pages/catalog/PackageDetailTab.js
pierrebeitz: interesting: the tooltip does not trigger `onMouseLeave` from a disabled input. 

i changed the implementation to add a `disabled`-class instead of using the attribute. this way it seems to work just fine.

this also fixes the tooltip behaviour for the other cases the button is disabled!

pierrebeitz:
> @TattdCodeMonkey I did notice that we're checking that the manager package exists, but not the status or it. So if its deploying you will get an error when trying to run the package with the dependency.

We're currently looking for a `TASK_RUNNING`, which was the best i could come up with. There seems to be no accurate service status for the kubernetes package yet, so this currently is the best we can do, right? would it be ok for you, if i opened an issue that reminds us of implementing a more sophisticated check once a package with a manager implements those statuses?

i'll happily implement more/different checks though if you have something in mind!

brandonc:
Finding it disquieting that the cosmos system test failed here. Could be relevant! 👀 I'll watch it

brandonc:
Frustrating because they pass locally

brandonc:
@pierrebeitz Cypress can't seem to complete mesos stream proxy requests so all the universe system tests fail with this change. So in order to preserve the universe system tests, I've added a commit that modifies the PackageDetailsTab to consult the DCOSStore (marathon groups endpoint) to check for an _installed_ kubernetes package.

Unfortunately, this is a worse solution because "kubernetes" could appear to be satisfied even if it has failed to start. This can result in the failure to install kubernetes-cluster. I don't think this is a huge deal because in the end we give the user an appropriate error message after kubernetes-cluster fails to install.

Another alternative is to disable universe system tests but I opted to not do this.

Another thought I had was that there's not enough information to figure out if "manager dependencies" are truly satisfied. Would it require that _all_ or _any_ tasks of that package be running?  For kubernetes, this is not an issue because there's only one task.

GeorgiSTodorov:
The button looks disabled, but I can still click it. Is this intentional?
Peek 2019-05-08 13-00

GeorgiSTodorov:
@brandonc , I approve of your current solution.

src/js/pages/catalog/PackageDetailTab.js
```diff
@@ -117,6 +132,23 @@ class PackageDetailTab extends mixin(StoreMixin) {
     }
   }
 
+  componentWillUnmount() {
+    DCOSStore.removeChangeListener(DCOS_CHANGE, this.onStoreChange);
+  }
+
+  componentWillMount() {
```
GeorgiSTodorov: Not blocking, but `componentWillMount` will be removed in React 17.

pierrebeitz:
> The button looks disabled, but I can still click it. Is this intentional?

no, not at all! should be fixed now!

> Why do we have "kubernetes" in quotation marks?

apparently i worked with a wrong design doc in the beginning. removed them!

> I ran "kubernetes", tried to install "kubernetes-cluster", but got this error. Maybe we need to show the infobox in case "kubernetes" isn't configured correctly?

that seems to be an error from the server. any idea on how to find out whether kubernetes is configured correctly upfront?

> Are there designs for the infobox in this case? I think gray is sort of easily ignored color, I think yellow or red are better.

@mperrotti talked to design. they want it to be gray. here's what i consider the design doc: https://mesosphere.invisionapp.com/share/2TRTTM5SZQX#/screens/361260526_k8cluster-Details-Page-Disabled

GeorgiSTodorov:
@pierrebeitz , thank you for addressing my feedback and for the quick response.

GeorgiSTodorov:
I have one concern though. If our logic for checking if the dependency is somewhat incorrect, we prevent the user from installing a certain package. I think we should just display a warning. We have a server error in case the user insists in trying to install the package.

GeorgiSTodorov:
Also, the tooltip message isn't in the message catalog and cannot be translated.

pierrebeitz:
> I have one concern though. If our logic for checking if the dependency is somewhat incorrect, we prevent the user from installing a certain package. I think we should just display a warning. We have a server error in case the user insists in trying to install the package.

i really admire that idea! we need to talk to design about this.

> Also, the tooltip message isn't in the message catalog and cannot be translated.

@GeorgiSTodorov updated the fixup-commit

brandonc:
@GeorgiSTodorov @TattdCodeMonkey the ServiceTree().getLabels method flattens out those running labels so we should be able to detect. The only thing we can't detect is whether or not the underling task requirement is satisfied. But I think that is a limitation of the API.

brandonc:
@pierrebeitz I think this is ready to go

brandonc:
I'm merging this because the "installed" requirement seems like the correct one to focus on since there are many scenarios we can't detect and this is meant as a helpful shortcut for someone who definitely does not have kubernetes installed before installing kubernetes-cluster

PR #3866, opened 2019-05-07
DCOS-52843: make json editor fit v0 spec

natmegs:
Make the JSON editor contents of the job form appear to be in metronome V0 spec format. Send payload to the API in V1 format.

**Note**: Please read/consider tradeoffs

Closes DCOS-52843

## Testing

Try interacting with the jobs form and all of its sections. Try adding jobs, editing jobs, etc. Try editing using form only, JSON editor only, a combination of both.

This is a fairly major change so the potential for bugs is high.

## Trade-offs

I'm hesitant about these changes so here are some points to consider:
- This is not a typical "bug fix." This changes the behaviour of the jobs form in a non-trivial way
- These changes are designed to allow users to continue using V0 formatted jobs despite the fact that we've moved on to V1 format behind the scenes. This could raise the questions: what benefit are we getting from using V1 format now? Are we potentially building in technical debt in exchange for backwards compatibility? 
- Server errors are returned with paths corresponding to v1 spec

## Dependencies

mesosphere/dcos-ui-plugins-private#996

## Screenshots

Note changes to the JSON structure in the editor.

Before:
Screen Shot 2019-05-06 at 5 07 19 PM

After:
Screen Shot 2019-05-06 at 4 59 52 PM

### DIFF
--- a/plugins/jobs/src/js/components/JobsForm.tsx
+++ b/plugins/jobs/src/js/components/JobsForm.tsx
@@ -85,11 +85,11 @@ class JobsForm extends React.Component<JobsFormProps> {
 
   getJSONEditorData(jobSpec: JobSpec): JobOutput {
     const jobJSON = jobSpecToOutputParser(jobSpec);
-    if (jobJSON.hasOwnProperty("schedule") && jobJSON.schedule === undefined) {
+    if (jobJSON.hasOwnProperty("schedule") && jobJSON.schedules === undefined) {
       // jobSpecToOutputParser returns object with `schedule: undefined` if there is no schedule present,
       // but this triggers an update of the JSONEditor and leads to issues where the state of the JSON in the
       // editor is replaced with old values.
-      delete jobJSON.schedule;
+      delete jobJSON.schedules;
     }
     return jobJSON;
   }
--- a/plugins/jobs/src/js/components/JobsFormModal.tsx
+++ b/plugins/jobs/src/js/components/JobsFormModal.tsx
@@ -30,7 +30,8 @@ import {
   JobSpec,
   JobOutput,
   Action,
-  Container
+  Container,
+  JobAPIOutput
 } from "./form/helpers/JobFormData";
 import { JobResponse } from "src/js/events/MetronomeClient";
 import JobForm from "./JobsForm";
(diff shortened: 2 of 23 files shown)


GeorgiSTodorov:
- (It's the same on master) For duplicate docker parameters, we don't show errors for the inputs, we just outline them in red.
Снимка от 2019-05-16 15-34-11
- Unlike master, we don't show errors in the form for schedule errors (it does show errors for starting deadline).
Снимка от 2019-05-16 15-37-23

natmegs:
@GeorgiSTodorov Thank you for the detailed review as always!

`For duplicate docker parameters, we don't show errors for the inputs, we just outline them in red.`

This is intentional - showing a message under every affected input looked noisy.

`Unlike master, we don't show errors in the form for schedule errors (it does show errors for starting deadline).`

That is because these come from the server and the paths match the v1 spec instead of the v0 spec that we display now using this change. An important tradeoff for others to consider as well!

`Same for Volumes.`

I'm having a hard time reproducing this. Maybe I fixed this while working on this feedback without realizing :) This is what I see:
Screen Shot 2019-05-16 at 5 28 49 PM

`Double dot if we manually change concurrency policy in the JSON.`

This is a side effect of adding a dot in the ErrorAlert component... the job errors are returned with dots so appending a dot in the error component causes the double dot.

`Concurrency policy errors sometimes persists even when we have no concurrency policy.`

This is a server error. Server errors persist until you resubmit.

`No form errors for run config (expect for labels) and double dot in the message.`

Run config section needs to have errors added to form UI (out of scope for this ticket). Double dot due to previously mentioned issue with ErrorsAlert/server error messages.

`Missing form error.`

Not able to reproduce this. The only way I can see this state in the form is if I put an empty space for variable name, and in that case I see this:
Screen Shot 2019-05-16 at 5 27 45 PM

`The error path does not start with job anymore.`

This is a necessary change so that errors are shown correctly in the JSON editor.

GeorgiSTodorov:
@natmegs , thank you for the fast response.
> This is a server error. Server errors persist until you resubmit.
Sometimes it persisted even after resubmitting.

> I'm having a hard time reproducing this.
> Not able to reproduce this.
Maybe you fixed this while working on something else.

juliangieseke:
regarding the tradeoffs:
afaik v0 is still supported, which means we want to support it. esp. when we're talking about backporting 1.13 to 1.12 eventually. once v0 is decommissioned, we can get rid of these changes in a breaking change release of the UI. (e.g. v2 -> v3).

brandonc:
feel free to merge once you resolve the conflicts and re-test

PR #3870, opened 2019-05-08
DCOS-12637: add metadata to health check endpoint dropdown

mperrotti:
Not all of the possible states were covered in the design Lee attached to the ticket, so here's a spreadsheet showing metadata content for all the states I was able to find: [Health check port metadata](https://docs.google.com/spreadsheets/d/1myz9yy9Xx1H4Vybjl7zC9j7DtZosrZCXRyjevhuFGcU/edit?usp=sharing)

## Testing

Single Container
1. In the create service form, go to the Network tab
2. Add a few endpoints - some with names and some without
3. Go to the Health Checks tab and add a health check
4. Select "HTTP" protocol

The "Select Endpoints" select dropdown should have additional metadata in parentheses in addition to what it was showing before this change.
If an endpoint has a Host Port auto-assigned, it should show the environemnt variable text "$PORT{index}"

The results should be consistent no matter which "Network Type" you select from the dropdown

Multi-container
1. In the create service form, go to the Network tab
2. Add a few endpoints to a container - some with names and some without
3. Go to the Health Checks tab and add a health check to that container
4. Select "HTTP" protocol

The "Select Endpoints" select dropdown should have additional metadata in parentheses in addition to what it was showing before this change.
If an endpoint has a Host Port auto-assigned, it should show the environment variable text "$ENDPOINT_{endpoint name}"

The results should be consistent no matter which "Network Type" you select from the dropdown

## Screenshots
Before:
Screen Shot 2019-05-07 at 8 13 47 PM

After:
Screen Shot 2019-05-07 at 8 12 50 PM


GeorgiSTodorov:
"Services -- Applications -- renders proper review screen and JSON for an app with HTTP health check" is failing. I will try to fix it.

mperrotti:
@natmegs - I agree that's a little weird. What's weirder though is if I manually assign the port, then change it back to an auto-assigned port, `hostPort` is still set to whatever I manually assigned it even after I "Review & Run". So technically, I think it's not being auto-assigned if you've manually entered a port.

It's an easy fix to show `$PORT{index}` if "Assign Host Ports Automatically" is checked, but is it more accurate to show the `hostPort` we have in the JSON? 🤔 

Would love to hear from other reviewers too

PR #3871, opened 2019-05-08
Proposal: chore: remove vscode settings file

Poltergeist:
I personally have some specific vs code settings to my dcos-ui workspace. Since we introduced the  `settings.json` to dcos-ui I have to run `git stash` and `git stash pop` regularly. Which means this change has really influenced how I work. Therefore I propose to remove the file again and put a note into the README with the appropriate settings. Then everybody can add these settings if they would like to, and therefore keep their existing settings as well.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,4 @@ src/resources/raml
 locale/_build/
 locale/*/messages.js
 plugins-ee
+.vscode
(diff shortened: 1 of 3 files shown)


GeorgiSTodorov:
@Poltergeist , I would suggesting adding this as a discussion topic to the sync, I don't think it will get enough visibility otherwise.

PR #3873, opened 2019-05-08
fix: change message according to design

pierrebeitz:
changing the infobox-message in case the package you're about to install has a unfulfilled manager dependency.

i'd consider this mergeable once the tests are green.


GeorgiSTodorov:
@pierrebeitz , I fixed a failing snapshot, I hope you don't mind.

PR #3878, opened 2019-05-10
[Spike] DCOS-53721: update jobs config page

natmegs:
Update configuration tab of job details page to reflect new form options.

Included in this PR:
- Updated General section
- Container section (including docker parameters)
- Schedule section
- Labels section
- Environment Variables section

Not implemented, to be done in separate tasks:
- Run config section
- Secrets
- Volumes
- Placement constraints

## Testing

Create jobs with different configurations, for example:
- With/without a schedule
- With UCR, Docker, and no container
- With/without environment variables
- With/without labels
Ensure that the configuration tab reflects the job configuration.

## Trade-offs

The `job` property being passed to the JobsConfiguration component is the result of a graphql query that returns a flattened but incomplete job config as well as a stringified complete job json configuration. I decided that changing the graphql query to include all necessary properties, as well as develop types for the return of the query, was out of scope for this work. For this reason, the complete parsed job json is used.

## Dependencies

N/A

## Screenshots

With labels and env vars:
Screen Shot 2019-05-09 at 5 04 16 PM

With docker container + parameters:
Screen Shot 2019-05-09 at 5 04 45 PM

With schedule:
Screen Shot 2019-05-09 at 5 04 29 PM


natmegs:
@GeorgiSTodorov Thanks for the reminder about the translations! We poll jobs every 2s so the slight lag comes from waiting to get the updated job details.

GeorgiSTodorov:
@natmegs , file conflicts.

natmegs:
This should be postponed until #3866 has been decided or merged.

juliangieseke:
#3866 got merged 🎉

mperrotti:
If a section has no data in the services configuration tab, we leave it empty. Is there a reason we need to show an empty state for the jobs configuration?

Screen Shot 2019-06-10 at 1 52 15 PM

natmegs:
@mperrotti The table is used for the key/value sections like env vars and labels (anything that isn't the standard single value with a label). This is used in the services config components as well for example here https://github.com/dcos/dcos-ui/blob/master/plugins/services/src/js/service-configuration/ServiceLabelsConfigSection.js#L100

natmegs:
@mperrotti good catch on the empty env var section - the rest of the sections don't get displayed if they are empty but I forgot to filter out the secrets in the env object before checking whether it's empty. Should be fixed now

PR #3882, opened 2019-05-10
[spike] DCOS-7157: Filter tasks by Failed/Killed status

GeorgiSTodorov:
Add the option to filter tasks that have the status
failed or killed. This allows for easier debugging.

Closes https://jira.mesosphere.com/browse/DCOS-7157

## Testing
1. Start a service with failed and killed tasks.
To have failed tasks, the command of the service could be something like `sleep 10 && exit 1`.
2. Open that service.
3. Click the button that opens the filter dropdown menu.
4. Verify that it contains checkboxes for failed and killed statuses.
5. Verify that these filters work as expected.
6. Verify that the old filters still work as expected.
7. Verify that the added tests make sense.

## Trade-offs
I am not sure if it's possible for pods to have a failed status. I will leave the filter as it is, though.

## Dependencies
None.

### Before
Screenshot from 2019-05-10 15-17-43

### After
Peek 2019-06-21 11-27

### DIFF
--- a/tests/pages/services/TasksTable-cy.js
+++ b/tests/pages/services/TasksTable-cy.js
@@ -232,6 +232,14 @@ describe("Tasks Table", function() {
       cy.get("table tr td.task-table-column-name").should("to.have.length", 4);
     });
 
+    it("Filters tasks that are in failed state", function() {
+      cy.get(".form-control.filter-input-text")
+        .type("{selectall}{backspace}")
+        .type("{selectall}{backspace}");
+      cy.get(".form-control.filter-input-text").type("is:failed");
+      cy.get("table tr td.task-table-column-name").should("to.have.length", 0);
+    });
+
     it("Filters tasks by region", function() {
       cy.get(".form-control.filter-input-text")
         .type("{selectall}{backspace}")
--- a/plugins/services/src/js/components/dsl/TaskStatusDSLSection.js
+++ b/plugins/services/src/js/components/dsl/TaskStatusDSLSection.js
@@ -13,7 +13,9 @@ import FormGroup from "#SRC/js/components/form/FormGroup";
 
 const EXPRESSION_PARTS = {
   is_active: DSLExpressionPart.attribute("is", "active"),
-  is_completed: DSLExpressionPart.attribute("is", "completed")
+  is_completed: DSLExpressionPart.attribute("is", "completed"),
+  is_failed: DSLExpressionPart.attribute("is", "failed"),
+  is_killed: DSLExpressionPart.attribute("is", "killed")
 };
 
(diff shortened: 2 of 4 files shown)


TattdCodeMonkey:
Seeing the following when testing this against daily cluster:
image
Choosing `Killed` shows 0 tasks, but choosing `Failed` shows tasks with status value `Killed`.

Also I'm not sure if its related but the size of the filter box changes when filters are selected:

image

GeorgiSTodorov:
@TattdCodeMonkey 
> Also I'm not sure if its related but the size of the filter box changes when filters are selected:

This is not related to this PR and something that also puzzles me.

>  Choosing Killed shows 0 tasks, but choosing Failed shows tasks with status value Killed.

I will investigate.

GeorgiSTodorov:
Apparently `Failed` shows both failed and killed and `Killed` shows nothing. I am working on a fix.

GeorgiSTodorov:
It should work as expected now. Nice catch, @TattdCodeMonkey .

tests/pages/services/TasksTable-cy.js
```diff
@@ -232,6 +232,14 @@ describe("Tasks Table", function() {
       cy.get("table tr td.task-table-column-name").should("to.have.length", 4);
     });
 
+    it("Filters tasks that are in failed state", function() {
+      cy.get(".form-control.filter-input-text")
+        .type("{selectall}{backspace}")
+        .type("{selectall}{backspace}");
+      cy.get(".form-control.filter-input-text").type("is:failed");
```
nLight: Test is checking for `is:failed` but in the code it is `is:failure`.
Let's please change it to `is:failed` in the code as well

GeorgiSTodorov: Good catch, I will update it.

GeorgiSTodorov: I think the `is:failure` is correct, so we should update the test.

nLight: We use verbiage: 
`is:active`
`is:killed`
`is:completed`
should be
`is:failed` following the logic

GeorgiSTodorov: `failure` already existed on master - https://github.com/dcos/dcos-ui/blob/master/plugins/services/src/js/constants/TaskStates.js#L45 , this is why I used it.

nLight:
1. There was a mixup between `failing` as an artificial state in the UI and `Failed` to be task status. I decided to use consistent verbiage `is:killed` `is:failed` `is:completed` etc.
2. I made `killed` to be separate aggregated state from `failure` so that filter would show them separately.

PR #3886, opened 2019-05-16
DCOS-53692: supporting changes

natmegs:
Supporting changes for mesosphere/dcos-ui-plugins-private#1007. Updates ui-kit version (to be able to use Typeahead) and adjusts a form input style so that it does not need to be applied directly to an input element. See EE PR for more details.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -72,9 +72,9 @@
       }
     },
     "@babel/types": {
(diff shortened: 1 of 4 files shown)


GeorgiSTodorov:
The build is failing because of failing snapshot tests.

PR #3887, opened 2019-05-16
add crypess:open to make better use of cypress during development

pierrebeitz:
if you open cypress and change a test file, it actually reruns that file. when
paired with `.only` on a single test case, it looks like writing integration
tests will be waaaay more convenient for me, as the feedback cycle actually
feels quite convenient.

without the `fixtures` config in place a `npx cypress open` would fail for me on most tests.

this also enables snapshots in development mode!

snaps

### DIFF
--- a/cypress.json
+++ b/cypress.json
@@ -10,6 +10,5 @@
   "supportFile": "tests/_support",
   "waitForAnimations": true,
   "videoUploadOnPasses": false,
-  "numTestsKeptInMemory": 0,
   "chromeWebSecurity": false
 }
--- a/package.json
+++ b/package.json
@@ -192,13 +192,14 @@
     "webpack-merge": "4.1.1"
   },
   "scripts": {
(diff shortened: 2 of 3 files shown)


pierrebeitz:
sorry for pushing again, but i discovered one more really cool thing!

snapshots!


snaps

cypress.json
```diff
@@ -10,6 +10,5 @@
   "supportFile": "tests/_support",
   "waitForAnimations": true,
   "videoUploadOnPasses": false,
-  "numTestsKeptInMemory": 0,
```
GeorgiSTodorov: What' s the advantage of having this in tests-functions and not here?

pierrebeitz: that if you `npm run cypress:open` you actually get snapshots!

i'll spike a little on jenkins and CI and want to adjust configuration in a way that makes cypress "dev-first" (without sacrificing on the security it gives us on CI of course).

i think if it's easier to write integration tests, we'll have more of those, which should give us even more confidence in our stuff.

PR #3891, opened 2019-05-16
refactor: let's always fakeAnalytics

pierrebeitz:
along with removing that flag the _support_/index.js has be refactored for
clarity.

### DIFF
--- a/tests/_support/index.js
+++ b/tests/_support/index.js
@@ -735,39 +735,24 @@ Cypress.Commands.add("clusterCleanup", function(fn) {
 });
 
 Cypress.Commands.add("visitUrl", function(options) {
-  var callback = function() {};
-
-  if (options.logIn && options.remoteLogIn) {
-    callback = function(win) {
-      // {"uid":"ui-bot","description":"UI Automated Test Bot","is_remote":true}
+  cy.visit(Cypress.env("CLUSTER_URL") + "/#" + options.url, {
+    onBeforeLoad(win) {
       win.document.cookie =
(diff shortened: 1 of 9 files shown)


tests/_support/index.js
```diff
@@ -735,39 +735,24 @@ Cypress.Commands.add("clusterCleanup", function(fn) {
...
+            "dcos-acs-info-cookie=eyJ1aWQiOiJ1aS1ib3QiLCJkZXNjcmlwdGlvbiI6IlVJIEF1dG9tYXRlZCBUZXN0IEJvdCIsImlzX3JlbW90ZSI6dHJ1ZX0K"
+          : // {"uid":"ui-bot","description":"UI Automated Test Bot"}
+            "dcos-acs-info-cookie=eyJ1aWQiOiJ1aS1ib3QiLCJkZXNjcmlwdGlvbiI6IlVJIEF1dG9tYXRlZCBUZXN0IEJvdCJ9Cg==";
 
-  if (options.identify && options.fakeAnalytics) {
-    var identifyCallback = callback;
-    callback = function(win) {
-      identifyCallback(win);
+      // mock a global analytics object
       win.analytics = {
```
GeorgiSTodorov: Where and how do we use that global object?

pierrebeitz: i believe that the usage is limited to the tracking plugin

pierrebeitz: thanks for asking this! 

in a next iteration we might only add that stuff to window, if the tracking plugin is enabled. i would prefer to not add this to the scope of this PR though.

PR #3893, opened 2019-05-17
chore: upgrade lingui to 2.8.0

brandonc:
## Overview

I [added a feature to lingui](https://github.com/lingui/js-lingui/pull/501) that allows a config file to be specified which really helps out with our externalplugins configuration. Therefore we no longer need to swap out config files when extracting messages.

The practical result is simplified lingui-extract and lingui-compile scripts

## Testing

check out companion plugins-private PR https://github.com/mesosphere/dcos-ui-plugins-private/pull/1008

- `npm run util:lingui:extract`
- `npm run util:lingui:compile`
(With and without plugins-ee symlink)

## Dependencies

I had to rename the .linguirc config file in the plugins-private repo (https://github.com/mesosphere/dcos-ui-plugins-private/pull/1008)


GeorgiSTodorov:
Build is failing with `npm ERR! cipm can only install packages when your package.json and package-lock.json or npm-shrinkwrap.json are in sync. Please update your lock file with `npm install` before continuing.`.

PR #3895, opened 2019-05-17
DCOS-53996: Replace Secrets select with Typeahead

mperrotti:
## Testing
Before testing, make sure there are a few secrets added

In the service or job form:
1. Go to the "Secrets" tab
2. Click "+ Add a Secret"
3. Focus the Secret field

Selecting an item from the Typeahead dropdown should set the value of the input and update the JSON.
The list in the typeahead dropdown should filter as you type 

## Dependencies
Same branch name in dcos-ui-plugins-private. PR: https://github.com/mesosphere/dcos-ui-plugins-private/pull/1010

## Screenshots
TypeaheadDemo_jobs


GeorgiSTodorov:
I think this solves https://jira.mesosphere.com/browse/DCOS-53996 , right? If yes, we should change the commit message.

package.json
```diff
@@ -48,6 +48,8 @@
     "create-react-class": "15.6.3",
     "d3": "3.5.16",
     "date-fns": "1.30.1",
+    "downshift": "3.2.10",
+    "react-draggable": "3.2.1",
```
GeorgiSTodorov: Why do we need these two?

mperrotti: Oops, we don't. This is a leftover from when I was testing with a local build of ui-kit

mperrotti:
@GeorgiSTodorov - yes, you're right, it should be say "Closes DCOS-53996"

GeorgiSTodorov:
@mperrotti , build is failing because of the commit message:
```
���   input: 

feat(SecretInput): use the Typeahead component instead for secret selection


Bumping ui-kit to a version that contains the Typeahead component, and a change to FieldInput to trigger an onChange when a value is set using the Typeahead component.


Closes DCOS-53996


���   header must not be longer than 72 characters [header-max-length]

���   found 1 problems, 0 warnings
```

mperrotti:
@DanielMSchmidt - I wouldn't consider that broken. I think it's better to horizontally scroll to see the rest of a long string without spaces than to truncate it or use `word-break` to break it to a new line.

This is just my opinion though - I'd be happy to make changes if any other reviewers agree with you.

DanielMSchmidt:
@mperrotti Sorry, I realize this is tough to see. The columns in the background are just as long as the text, this looks visually broken on hover

mperrotti:
@DanielMSchmidt - I see what you mean now after loading it up on my laptop. Simple ui-kit fix :)

DanielMSchmidt:
Awesome 👍

mperrotti:
Fix is in this PR: https://github.com/dcos-labs/ui-kit/pull/336

DanielMSchmidt:
Awesome, I approved it, let's ship it 💪

GeorgiSTodorov:
I updated ui-kit's version to solve @DanielMSchmidt 's problem.

GeorgiSTodorov:
I think this is safe to merge.

PR #3904, opened 2019-05-23
DCOS-43588: UI doesn't reflect executor's resources for Pods on the Services page

GeorgiSTodorov:
Include executor resources when showing
pods in the services table.

Closes https://jira.mesosphere.com/browse/DCOS-43588

This PR is work in progress and is still missing tests.

## Testing
1. Go to services tab.
2. Create a pod.
3. In the services table, verify that the resource columns also show executor resources. Compare them with the resources when clicking on the pod. In my example:
Снимка от 2019-05-23 19-10-18

## Trade-offs
None.

## Dependencies
None.

### Before
Снимка от 2019-05-23 19-05-01

### After
Снимка от 2019-05-23 19-03-07

### DIFF
--- a/plugins/services/src/js/structs/Service.js
+++ b/plugins/services/src/js/structs/Service.js
@@ -116,12 +116,24 @@ module.exports = class Service extends Item {
       gpus = 0,
       disk = 0
     } = this.getSpec().getResources();
+    let executorCpus = 0;
+    let executorMem = 0;
+    let executorGpus = 0;
+    let executorDisk = 0;
+
+    if (this.getSpec().get("executorResources")) {
+      const executor = this.getSpec().get("executorResources");
+      executorCpus = executor.cpus ? executor.cpus : 0;
+      executorMem = executor.mem ? executor.mem : 0;
+      executorGpus = executor.gpus ? executor.gpus : 0;
+      executorDisk = executor.disk ? executor.disk : 0;
+    }
 
     return {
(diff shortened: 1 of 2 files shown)


plugins/services/src/js/structs/Service.js
```diff
@@ -116,12 +116,24 @@ module.exports = class Service extends Item {
...
+      executorGpus = executor.gpus ? executor.gpus : 0;
+      executorDisk = executor.disk ? executor.disk : 0;
+    }
 
     return {
-      cpus: cpus * instances,
-      mem: mem * instances,
-      gpus: gpus * instances,
-      disk: disk * instances
+      cpus: (cpus + executorCpus) * instances,
```
brandonc: Isn't the correct calculation `(cpus * instances) + executorCpus` ? I can't imagine why we'd want to multiply executorCpus by instances

TattdCodeMonkey: According to the comments on the JIRA ticket  it is "per container"

brandonc: We may need more clarification because instances ≠ containers --  There may be one executor per container but not per instance container. Or maybe its per instance in which case this would be  correct!!

GeorgiSTodorov: @brandonc , if you create a pod with two instances and open it, you will see that each instance has executor resources.

PR #3916, opened 2019-05-30
[1.13] DCOS-54491: Framework configuration form tab bug on Firefox <=63 on Windows/Linux

mperrotti:
## Testing
1. Open Firefox version 63 or older on a Windows or Linux machine (if you need an invite to our BrowserStack account, ping me)
2. Go to the Catalog, pick a framework, and click "Review & Run" to open the framework configuration form

The tabs on the left should all be clickable

## Trade-offs
I wasn't able to find what exactly changed between 1.12 and 1.13 that surfaced this bug, so this is just a quick fix

## Screenshots
Before:
FrameworkConfigFFTabFix_before

After:
FrameworkConfigFFTabFix_after

### DIFF
--- a/src/styles/components/menu-tabbed/styles.less
+++ b/src/styles/components/menu-tabbed/styles.less
@@ -11,11 +11,13 @@
   }
 
   .menu-tabbed-container-fixed {
+    // force the menu to it's own composite layer to fix a bug in
+    // Firefox 63 and below on Windows and Linux
+    transform: translateZ(0);
 
     .menu-tabbed-vertical {
       height: 100%;
       position: fixed;
-      top: 135px;
     }
   }
 


src/styles/components/menu-tabbed/styles.less
```diff
@@ -11,11 +11,13 @@
...
 
   .menu-tabbed-container-fixed {
+    // force the menu to it's own composite layer to fix a bug in
+    // Firefox 63 and below on Windows and Linux
+    transform: translateZ(0);
 
     .menu-tabbed-vertical {
       height: 100%;
       position: fixed;
-      top: 135px;
```
DanielMSchmidt: What is this change helping with?

mperrotti: This number wasn't accurate, so the tab border was positioned incorrectly. It's much worse in those older FF versions.

Screen Shot 2019-06-04 at 2 49 09 PM

PR #3920, opened 2019-06-04
DCOS-43777: provide default volume type for DSS volumes

orlandohohmeier:
DSS (profile volumes) will as of today only successfully _launch_ if the type is set to `mount` thus we need to provide the resp. volume type property.

* Adjust `Volumes` and `MultiContainerVolumes` reducers to provide the correct default volume type
* Add validation rules to catch volume type issues during initial creation e.g., changing a local persist volume with volume type "path" or "root" to a profile volume

Closes DCOS-43777

## Testing

* Create a pod/app with a profile volume and check whether a default volume type with the correct value (`mount`) is provided.
* Create a pod/app with a local persistent volume and make sure that no default is provided
* Create a pod/app with a local persistent volume, define a volume type e.g. "path", and then change the volume type to DSS

## Trade-offs

I've introduced validation rules instead of overwriting the volume type when changing the volume type to keep the user input. 

## Notes

We will also change the Marathon default volume type for profile volumes – for details please see: https://jira.mesosphere.com/browse/DCOS-54690


orlandohohmeier:
I'll back-port the fix to release/1.13 once I get a thumbs up.

GeorgiSTodorov:
@orlandohohmeier , are you still planning on creating a 1.13 backport?

orlandohohmeier:
@GeorgiSTodorov done. :) https://github.com/dcos/dcos-ui/pull/3922

PR #3921, opened 2019-06-10
DCOS-41120: don't show tooltip on services table region cell if it's "N/A"

mperrotti:
## Testing
1. Navigate to the Services table
2. If there aren't already any rows with "N/A" in the region column, add a group from the ellipsis button in the page header
3. Hover the "N/A" text in the region column

No tooltip should be shown

## Screenshots
Before:
RegionTooltip-before

After:
RegionTooltip-after


GeorgiSTodorov:
I think having a small integration test for this would be good.

mperrotti:
Thanks @GeorgiSTodorov - I'll add one today

PR #3923, opened 2019-06-11
fix: improve error message for invalid env variable name

TattdCodeMonkey:
[DCOS-54437](https://jira.mesosphere.com/browse/DCOS-54437) was written as an error exposing a secret as an env variable for a pod, but when testing I found the provided env variable `1234` was not valid for Marathon. Updated ServiceErrorMessages to treat `evironment.[key]` error paths as an invalid environment variable name and not "must be defined"

## Testing

Define a pod or service with an invalid env variable name:
```json
{
  "environment": {
    "123": "TEST"
  },
  "id": "/",
  "containers": [
    {
      "name": "container-1",
      "resources": {
        "cpus": 0.1,
        "mem": 128
      },
      "exec": {
        "command": {
          "shell": "sleep 1000"
        }
      }
    }
  ],
  "scaling": {
    "instances": 1,
    "kind": "fixed"
  },
  "networks": [
    {
      "mode": "host"
    }
  ],
  "volumes": [],
  "fetch": [],
  "scheduling": {
    "placement": {
      "constraints": []
    }
  }
}
```
2. click review & run, and run service

3. Verify Error message is "Environment variables must be valid." and not "Environment variables must be defined."
## Trade-offs

I added a new entry with a different regex and made the existing error match `environment.[key].+` to retain any existing behavior, but we may just want to replace the existing error message of `Must be defined` with `Must be valid` I took a "safer" approach since I wasn't sure use case where we would get other `evironment.[key].[value]` error paths.

## Dependencies

n/a

## Screenshots

image

### DIFF
--- a/plugins/services/src/js/constants/ServiceErrorMessages.js
+++ b/plugins/services/src/js/constants/ServiceErrorMessages.js
@@ -92,7 +92,12 @@ const ServiceErrorMessages = [
     )
   },
   {
-    path: /^environment\./,
+    path: /^environment\.[^.]+$/,
+    type: "GENERIC",
+    message: i18nMark("Must be valid")
+  },
+  {
+    path: /^environment\..*\./,
     type: "GENERIC",
     message: i18nMark("Must be defined")
   }
--- a/locale/en/messages.json
+++ b/locale/en/messages.json
@@ -486,6 +486,7 @@
   "Must be specified when using the Docker Engine runtime": "",
   "Must be specified when using the Docker Engine runtime. You can change runtimes under \"Advanced Settings\"": "",
   "Must be unique": "",
+  "Must be valid": "",
   "Must contain a property that matches `||pattern||`": "",
   "Must contain at least one item": "",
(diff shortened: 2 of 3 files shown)


GeorgiSTodorov:
Adding hold merge as there is a design feedback in the jira.

TattdCodeMonkey:
Removed hold merge, design signed off on error message as written here.

plugins/services/src/js/constants/ServiceErrorMessages.js
```diff
@@ -92,7 +92,12 @@ const ServiceErrorMessages = [
     )
   },
   {
-    path: /^environment\./,
+    path: /^environment\.[^.]+$/,
+    type: "GENERIC",
+    message: i18nMark("Must be valid")
```
GeorgiSTodorov: Not specific to this PR, but I just noticed that these are not getting translated (maybe because we change them to lower case?).
Screenshot from 2019-06-13 11-18-44
Screenshot from 2019-06-13 11-19-27

PR #3924, opened 2019-06-12
feat: allow for link-tags in headings

pierrebeitz:
# ⚠️ 	do not merge

this commit has been created while trying to address an internal request
regarding links in form-headings.

the implementation certainly has implications regarding the security of our
website.
everyone with access to the schema definitions can now do evil things to a lot
of users.

⚠️ 	i just realized that in the other occurrences using `dangerouslySetInnerHTML` we pipe our stuff through a markdown-parser first. If we consider going this route that had to be implemented here as well.

here's a screenshot: 
OS Enterprise 2019-06-12 14-38-54

### DIFF
--- a/src/js/components/SchemaField.js
+++ b/src/js/components/SchemaField.js
@@ -244,7 +244,7 @@ class SchemaField extends Component {
         {requiredSymbol}
         <FormGroupHeadingContent primary={false}>
           <Tooltip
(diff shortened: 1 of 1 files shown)


pierrebeitz:
that's specifically why it's a draft and said to be evil. 👍

PR #3925, opened 2019-06-12
DCOS-48532: No text button in Marathon configuration

GeorgiSTodorov:
Add a text for a textless button.

Closes https://jira.mesosphere.com/browse/DCOS-48532

## Testing
1. Go to catalog.
2. Open `marathon`.
3. Click "Review & Run".
4. Open the "Marathon" tab and scroll to the bottom.
5. Verify that the button is not empty.

## Trade-offs
Somewhat of a band-aid solution.

## Dependencies
None.

### Before
Снимка от 2019-06-12 15-59-26

### After
Снимка от 2019-06-12 15-38-21


src/styles/components/framework-configuration/styles.less
```diff
@@ -29,3 +29,7 @@
     z-index: 0;
   }
 }
+
+.glyphicon-plus:before {
```
mperrotti: This element is a `<i>`, so the browser styles it as italic.

Could you add `font-style: normal` to these styles?

mperrotti:
I see this original component is coming from `react-jsonschema-form`. Is there a way we can replace it with our `AddButton`?

GeorgiSTodorov:
@mperrotti , where do you see the button from `react-jsonschema-form`?

mperrotti:
@GeorgiSTodorov - by inspecting through the React inspector

GeorgiSTodorov:
@mperrotti , I hadn't considered that. Still, I am not sure if that wouldn't be an over-kill for this bug. I initially tried to use our icons from the ui-kit, but I couldn't get the imports in the `less` file to work, so I decided to use the current solution as good enough.

src/styles/components/framework-configuration/styles.less
GeorgiSTodorov: The plus is not _italic_ in the screenshot, is it italic for you?

mperrotti:
Yup, I agree - current solution is fine. Just change the `font-style` from italic :)

src/styles/components/framework-configuration/styles.less
mperrotti: @GeorgiSTodorov - yes, it was _italic_ for me in Chrome

PR #3928, opened 2019-06-13
DCOS-43896: Progress indicator doesn't go away without a page reload

GeorgiSTodorov:
Update the breadcrumbs for the services when
instances or status change.

Closes https://jira.mesosphere.com/browse/DCOS-43896

## Testing
1. Create a service with multiple instances.
Example JSON:
```
{
  "id": "/new-service-1",
  "instances": 10,
  "portDefinitions": [],
  "container": {
    "type": "MESOS",
    "volumes": []
  },
  "cpus": 0.1,
  "mem": 128,
  "requirePorts": false,
  "networks": [],
  "healthChecks": [],
  "fetch": [],
  "constraints": [],
  "cmd": "sleep 1000;"
}
```
2. Start it and quickly open the service details for that service.
2.5. (Optional) Open services overview in another tab.
3. As soon as the service is running, it should have an updated breadcrumb in the service detail page.

## Trade-offs
1. I tried adding an integration test, but I couldn't figure out how to mock the actual scale so that we can see the changed status. Even in the test where we test the actual scale action, we don't really scale. If someone knows how this could work, that would be great. This is what I have so far.
In `ServiceDetail-cy` in the `Navigate to service page` context.
```
      it("updates breadcrumbs on status change", function() {
        cy.route({
          method: "PUT",
          url: /marathon\/v2\/apps\/\/net/,
          response: []
        });
        cy.visitUrl({
          url: "/services/detail/%2Fnet"
        });

        cy.get(".button-narrow").click();
        cy.get(".dropdown-menu-items")
          .contains("Scale")
          .click();
        cy.get(".form-group input[name='instances']").type("00");
        cy.get(".button")
          .contains("Scale Service")
          .click();
        cy.get(".status-bar").should("exist");
      });
```
2. While working on this, I noticed that pods don't have a status in the breadcrumbs for some reason.

## Dependencies
None.

### Before
Peek 2019-06-13 14-40
Peek 2019-06-13 14-41

### After
Peek 2019-06-13 14-30
Peek 2019-06-13 14-37


src/js/components/Breadcrumb.js
```diff
@@ -1,13 +1,17 @@
...
+import isEqual from "lodash.isequal";
 
 class Breadcrumb extends React.Component {
   shouldComponentUpdate(nextProps) {
     return (
       this.props.isCaret !== nextProps.isCaret ||
       this.props.isIcon !== nextProps.isIcon ||
-      this.props.title !== nextProps.title
+      this.props.title !== nextProps.title ||
+      this.props.instancesCount !== nextProps.instancesCount ||
```
mperrotti: Would it be possible to do these `shouldComponentUpdate` checks in `ServiceBreadcrumb` instead?

TattdCodeMonkey: These props are specific to `ServiceBreadcrumbs` and we should not be referencing them here.

GeorgiSTodorov: We do them there, but we also have to do them here.

GeorgiSTodorov: If we don't check them here, we don't trigger a refresh.

Poltergeist: But then we are messing up something else. We should not need to pass foreign properties to a component for re-rendering although the component does not use the prop in any way. I will look into the issue later / tomorrow.

GeorgiSTodorov: If that's the problem, then it's a problem for `ServiceBreadcrumbs` as well. There we have the service id, so we also don't need them.

GeorgiSTodorov: But if we don't update `Breadcrumb`, then the children also don't get updated.

Poltergeist: So I played around a little with `shouldComponentUpdate` I unfortunately didn't found a good solution to the problem. Since this is a bug in order to fix it now I propose that we create a follow up issue where we want to refactor this and clean the whole component up and progress the PR as it is?

TattdCodeMonkey: I would rather add a generic optional prop to `Breadcrumbs` and set it from `ServiceBreadcrumbs` based on the values we care about changing. That way we aren't adding `ServiceBreadcrumbs` specific props too the generic `Breadcrumbs` component, and if we run into a similar issue we can use the same solution. I don't have a good idea on the name of said prop though, something generic?

brandonc: Why not just not implement shouldComponentUpdate? It's an optimization, after all.

GeorgiSTodorov: @brandonc , which `shouldComponentUpdate` do you mean? The one in `Breadcrumb` is not added by this PR.

brandonc: Right, I meant getting rid of it. It shouldn't be necessary to update the component

GeorgiSTodorov: As far as I know, we don't mix bug fixing and refactoring.

nLight: Yep, we don't :) let's fix the bug and follow-up with a refactoring PR

brandonc: Maybe I'm not being understood... This fixes the bug but the feedback has to do with code in the wrong place. Instead of refactoring the fix to satisfy Rodney's concern, you could delete the method altogether and sidestep the issue.

GeorgiSTodorov: @brandonc , you are saying to remove my new code in `Breadcrumb.js` or to remove the entire `shouldComponentUpdate` in `Breadcrumb.js`? If we did the first, it wouldn't work, if we did the second, it would be a refactoring.

TattdCodeMonkey: Removing `shouldComponentUpdate` in `Breadcrumb` would address the root cause of the bug, not be a refactor IMO.

GeorgiSTodorov: Are you sure this wouldn't break something else?

brandonc: If I were a lawyer, I might say "to a reasonable degree of analytical certainty" lol

plugins/services/src/js/components/ServiceBreadcrumbs.js
```diff
@@ -230,7 +249,13 @@ class ServiceBreadcrumbs extends React.Component {
         }
 
         return (
-          <Breadcrumb key={index} title={ids.slice(0, index + 1).join("/")}>
+          <Breadcrumb
+            key={index}
+            title={ids.slice(0, index + 1).join("/")}
+            instancesCount={instancesCount}
```
brandonc: I don't think these service-specific props need to be given anymore.

GeorgiSTodorov: Good catch.

PR #3929, opened 2019-06-14
DCOS-46214: Sorting in Active Deployments modal is broken

GeorgiSTodorov:
Fix the sorting for time.

Closes DCOS-46214

## Testing
1. Go to services tab.
2. Add a few services that can't start (for example very high cpu or memory requirements) few minutes within each other.
3. Click the "active deployments" link in the top right corner of the Services tab.
4. Verify that sorting by start time works correctly.

## Trade-offs
The caret is reversed for this column because we are sorting by time it was created but we are showing time passed. I felt like fixing this is out of scope here.

## Dependencies
None.

### Before
Peek 2019-06-14 14-19

### After
Peek 2019-06-14 16-00

### DIFF
--- a/plugins/services/src/js/components/DeploymentsModal.js
+++ b/plugins/services/src/js/components/DeploymentsModal.js
@@ -25,6 +25,7 @@ import ServiceUtil from "#PLUGINS/services/src/js/utils/ServiceUtil";
 import ProgressBar from "#SRC/js/components/ProgressBar";
 import StringUtil from "#SRC/js/utils/StringUtil";
 import TimeAgo from "#SRC/js/components/TimeAgo";
+import TableUtil from "#SRC/js/utils/TableUtil";
 
 import defaultServiceImage from "../../img/icon-service-default-small@2x.png";
 import MarathonActions from "../events/MarathonActions";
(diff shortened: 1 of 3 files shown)


plugins/services/src/js/components/DeploymentsModal.js
```diff
@@ -37,7 +38,7 @@ const columnClasses = {
 };
 const columnHeadings = ResourceTableUtil.renderHeading({
   id: i18nMark("Affected Services"),
-  startTime: i18nMark("Started"),
+  version: i18nMark("Started"),
```
mperrotti: Sorry, I think I'm missing something here - why does this key need to be changed to "version"?

natmegs: Same question from me. My concern is that it's not an accurate description of the column and could be confusing in the future. I tried changing this and line 195 to "startTime" and it seems to be ok?

TattdCodeMonkey: So debugging this it appears that `startTime` on `Deployment` is a `Date` object where are `version` is an ISO-8601 timestamp. So doing a naive comparison with `TableUtil.getSortFunction` (see line 181) on the `version` prop fixes this sorting issue, where as we would have to write a sort function to do comparisons on the Date object.

GeorgiSTodorov: @mperrotti , @natmegs , startTime of the object was undefined. Maybe it was a property that we used to have.

PR #3932, opened 2019-06-14
DCOS-55176 - fix: ensure getStatus exists before calling

TattdCodeMonkey:
We have seen errors in Sentry where the item being rendered into the
deployments table does not have a getStatus function. This doesn't make
a lot of sense since the item _should_ be an Application, Service or
ServiceTree. And all of those structs have a getStatus function. But
adding a guard against an unhandled exception here regardless.

- Closes DCOS-55176

### DIFF
--- a/plugins/services/src/js/components/__tests__/DeploymentsModal-test.js
+++ b/plugins/services/src/js/components/__tests__/DeploymentsModal-test.js
@@ -1,0 +1,31 @@
+import Item from "#SRC/js/structs/Item";
+/* eslint-disable no-unused-vars */
+const React = require("react");
+/* eslint-enable no-unused-vars */
+
+const DeploymentsModal = require("../DeploymentsModal").WrappedComponent;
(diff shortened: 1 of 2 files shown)


plugins/services/src/js/components/__tests__/DeploymentsModal-test.js
```diff
@@ -0,0 +1,31 @@
+import Item from "#SRC/js/structs/Item";
+/* eslint-disable no-unused-vars */
+const React = require("react");
+/* eslint-enable no-unused-vars */
+
+const DeploymentsModal = require("../DeploymentsModal").WrappedComponent;
+const Application = require("../../structs/Application");
+
+describe("DeploymentsModal", function() {
+  describe("#renderStatus", function() {
```
brandonc: This feels like you're testing some private implementation detail of this component... In general I feel it would be nice to either extract renderStatus and test that, or test the rendering of the entire component

PR #3933, opened 2019-06-14
DCOS-55177 - fix: ensure getStatus exists before calling

TattdCodeMonkey:
We have seen errors in Sentry where the item being rendered into the
deployments table does not have a getStatus function. This doesn't make
a lot of sense since the item _should_ be an Application, Service or
ServiceTree. And all of those structs have a getStatus function. But
adding a guard against an unhandled exception here regardless.

- Closes DCOS-55177

### DIFF
--- a/plugins/services/src/js/components/DeploymentsModal.js
+++ b/plugins/services/src/js/components/DeploymentsModal.js
@@ -490,7 +490,7 @@ class DeploymentsModal extends mixin(StoreMixin) {
       {});
     }
 
-    let statusText = !item.isStale ? item.getStatus() : null;
+    let statusText = !item.isStale && item.getStatus ? item.getStatus() : null;
     const itemId = item.isStale ? item.serviceID : item.id;
 
     if (currentActions[itemId] != null) {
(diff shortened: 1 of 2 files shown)


plugins/services/src/js/components/DeploymentsModal.js
```diff
@@ -483,7 +483,7 @@ class DeploymentsModal extends mixin(StoreMixin) {
       {});
     }
 
-    let statusText = !item.isStale ? item.getStatus() : null;
+    let statusText = !item.isStale && item.getStatus ? item.getStatus() : null;
```
natmegs: Do you think it would be overkill to check `typeof item.getStatus === "function"`?

TattdCodeMonkey: I also considered doing `item.isStale === false` b/c if the struct has `isStale` it *should* also have `getStatus()`, but it felt less explicit and could be confusing looking at that code sometime down the road.

TattdCodeMonkey:
rebased & force pushed

PR #3934, opened 2019-06-14
DCOS-8665: active state for action buttons

mperrotti:
## Testing
Just smoke test and let me know if you can think of any buttons I may have missed.

## Trade-offs
I put the ellipsis in the text, but in the ui-kit button components, the ellipsis is added in CSS with the `:after` pseudo element. This means when/if we replace these buttons with ui-kit buttons, we'll have to delete the "..." in the test, or we'll get a double ellipsis.

## Dependencies
dcos-ui-plugins-private PR: https://github.com/mesosphere/dcos-ui-plugins-private/pull/1038

## Screenshots
Too many spots to screenshot everything


plugins/services/src/js/components/modals/KillPodInstanceModal.js
```diff
@@ -121,6 +126,10 @@ class KillPodInstanceModal extends React.PureComponent {
       buttonText = i18n._(t`Force`) + " " + buttonText;
     }
 
+    if (isPending) {
+      buttonText = ACTION_DISPLAY_NAMES_CONTINUOUS[action];
```
brandonc: It looks like buttonText never gets translated.

GeorgiSTodorov:
I second what @brandonc said, some new messages are not in the message catalog, for example "Submitting", "Restarting", "Stopping", "Deploying...". 

Also, "stopping" for jobs doesn't appear for me.
Peek 2019-06-17 08-58

Other than that, it looks good.

locale/en/messages.json
```diff
@@ -786,7 +795,7 @@
   "There are no plans currently available.": "",
   "There are tasks in this queue that DC/OS has failed to deploy for {duration}.": "",
   "There is an error with your configuration": "",
-  "This action <0>CANNOT</0> be undone. This will permanently delete the <1>{serviceName}</1> {0}.": "",
+  "This action <0>CANNOT</0> be undone. This will permanently delete the <1>{serviceName}</1> {0} .": "",
```
brandonc: Are these prettier artifacts? 😭

locale/zh/messages.json
```diff
@@ -786,7 +795,7 @@
   "There are no plans currently available.": "",
   "There are tasks in this queue that DC/OS has failed to deploy for {duration}.": "",
   "There is an error with your configuration": "您的配置有错误",
-  "This action <0>CANNOT</0> be undone. This will permanently delete the <1>{serviceName}</1> {0}.": "此操作 <0>无法</0> 撤销。这将永久删除 <1>{serviceName}</1> {0}.",
```
brandonc: ...Because they are killing translations

plugins/catalog/src/js/repositories/components/AddRepositoryFormModal.js
```diff
@@ -79,14 +79,18 @@ class AddRepositoryFormModal extends React.Component {
...
     return [
       {
         text: i18nMark("Cancel"),
         className: "button button-primary-link flush-left",
         isClose: true
       },
       {
-        text: i18nMark("Add Repository"),
+        text: pendingRequest
+          ? i18nMark("Adding...")
```
brandonc: This didn't actually work for me. Where does the pendingRequest prop come from?

plugins/catalog/src/js/repositories/components/AddRepositoryFormModal.js
```diff
@@ -79,14 +79,18 @@ class AddRepositoryFormModal extends React.Component {
...
+          : i18nMark("Add Repository"),
         className: "button button-primary",
         isSubmit: true
```
brandonc: Can buttons using this declarative system be disabled?

plugins/catalog/src/js/repositories/components/RepositoriesDeleteConfirm.js
```diff
@@ -37,7 +37,9 @@ const RepositoriesDeleteConfirm = ({
       <Trans render="span">Delete Repository</Trans>
     </ModalHeading>
   );
-  const rightButtonText = i18nMark("Delete Repository");
+  const rightButtonText = pendingRequest
+    ? i18nMark("Deleting...")
+    : i18nMark("Delete Repository");
```
brandonc: I didn't detect a change with this button, either.

locale/en/messages.json
mperrotti: Yes, these are Prettier artifacts. I'll try and figure something out so we're not breaking translations

plugins/catalog/src/js/repositories/components/AddRepositoryFormModal.js
mperrotti: Good catch. It should be coming from the result of `addRepository$`, but we're only ever setting it to false, and we don't pass it as a prop to the `AddRepositoryFormModal` component

plugins/catalog/src/js/repositories/components/RepositoriesDeleteConfirm.js
mperrotti: I can't get it to stay on screen long enough to show a change

plugins/catalog/src/js/repositories/components/AddRepositoryFormModal.js
mperrotti: Yes, the confirm modal takes a `disabled` prop, which disables the buttons

GeorgiSTodorov:
Why do we delete so many JobStopRun files?

mperrotti:
@GeorgiSTodorov - they weren't being used

GeorgiSTodorov:
I tried a few different tabs, but it didn't work for me in any of them.

plugins/catalog/src/js/repositories/components/RepositoriesDeleteConfirm.js
GeorgiSTodorov: I console.log-ed pendingRequest here and it is false even when deleting.
Peek 2019-06-21 09-48

plugins/services/src/js/components/DeploymentsModal.js
```diff
@@ -429,6 +429,10 @@ class DeploymentsModal extends mixin(StoreMixin) {
       </ModalHeading>
     );
 
+    const rollbackActionText = awaitingRevertDeploymentResponse
```
GeorgiSTodorov: `awaitingRevertDeploymentResponse` seems to be always `undefined`.

GeorgiSTodorov:
Tested extensively and saw two cases where it doesn't work as expected.

mperrotti:
@GeorgiSTodorov - what are the two cases?

GeorgiSTodorov:
@mperrotti , I mentioned them in individual comments.
https://github.com/dcos/dcos-ui/pull/3934#discussion_r296115140
https://github.com/dcos/dcos-ui/pull/3934#pullrequestreview-252665918

plugins/catalog/src/js/repositories/components/RepositoriesDeleteConfirm.js
GeorgiSTodorov: Now it works.

plugins/services/src/js/components/DeploymentsModal.js
GeorgiSTodorov: Now it works.

PR #3936, opened 2019-06-18
DCOS-49181: Version column in service table shows misleading information

GeorgiSTodorov:
Show full version for frameworks.

Closes https://jira.mesosphere.com/browse/DCOS-49181

## Testing
1. Go to the services tab.
2. Start some services.
3. Go to catalog.
4. Start some frameworks.
5. Go to services table.
6. Verify that for services we show no version and for frameworks we show the full version (when hovering the cell, the tooltip should show the same version).
7. Verify that the added test makes sense.

## Trade-offs
None.

## Dependencies
None.

### Before
Screenshot from 2019-06-18 14-37-02

### After
Screenshot from 2019-06-18 13-16-47

### DIFF
--- a/plugins/services/src/js/utils/Version.ts
+++ b/plugins/services/src/js/utils/Version.ts
@@ -37,7 +37,4 @@ function compare(a: Version, b: Version) {
 
 const fromService = (service: Service | Pod) => service.getVersion();
 
-const toDisplayVersion = (v: Version) =>
-  FrameworkUtil.extractBaseTechVersion(v);
-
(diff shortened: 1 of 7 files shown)


plugins/services/src/js/utils/Version.ts
```diff
@@ -37,7 +37,4 @@ function compare(a: Version, b: Version) {
 
 const fromService = (service: Service | Pod) => service.getVersion();
```
GeorgiSTodorov: I am not sure why we need this.

PR #3941, opened 2019-06-21
[Spike] Refactor ServicesPage: update with es6 class

natmegs:
Move ServicesPage from createReactClass to es6 class.

## Testing

Go to services page, tab around and make sure nothing breaks.

## Trade-offs

I 90% converted this to TS but got hung up on the mixin types 😞 

## Dependencies

N/A

### DIFF
--- a/plugins/services/src/js/pages/ServicesPage.js
+++ b/plugins/services/src/js/pages/ServicesPage.js
@@ -1,9 +1,9 @@
 import { i18nMark } from "@lingui/react";
 import React from "react";
-import createReactClass from "create-react-class";
 import { routerShape } from "react-router";
 import { StoreMixin } from "mesosphere-shared-reactjs";
 import { Icon } from "@dcos/ui-kit";
+import mixin from "reactjs-mixin";
 import { ProductIcons } from "@dcos/ui-kit/dist/packages/icons/dist/product-icons-enum";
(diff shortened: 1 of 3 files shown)


nLight:
> I 90% converted this to TS but got hung up on the mixin types 😞

Can't we ignore/`any` those missing types? I feel like otherwise it's to heavy to do in one take

nLight:
@pierrebeitz to rescue? 🙏

pierrebeitz:
PR is already on the radar as i think TS's static properties will also make this PR easier to review.

pierrebeitz:
@natmegs 	added a commit that addresses what i think might have been the problem.

as always feel free to use, alter or throw away my commits.

natmegs:
Hey @pierrebeitz thanks for the help! I had actually tried something super similar, but the problem that I see is that the typing of the mixin as returning `any` makes it so that we are not getting the benefit of types at all inside of the component (it doesn't know what types the props or state should be. For ex if you take out the state initializer it should yell at you but it doesn't). Maybe this is ok for a first pass though?

pierrebeitz:
to properly infer all that information we'd have to make up for our magic mixins - i.e. the `TabsMixin` with dynamically generated `render${currentTab}TabView`-functions. i'm not sure whether it is possible to create sound and complete typings for that mixin-mechanism.

i strongly believe that the codebase would benefit from removing the mixins (e.g. by making everything they do explicit) and thus think there's not much value in trying to come up with those really complex types.

PR #3943, opened 2019-06-24
DCOS-41310, DCOS-41842, DCOS-42506

GeorgiSTodorov:
Update the node count whenever we filter by framework.

Closes https://jira.mesosphere.com/browse/DCOS-41310 .
Closes https://jira.mesosphere.com/browse/DCOS-41842 .
Closes https://jira.mesosphere.com/browse/DCOS-42506 .

## Testing
DCOS-41310:
1. Go to nodes page.
2. Use the "filter by framework" dropdown to filter.
3. Verify that every time it shows the correct number of nodes above the filter (also when using "Clear" to clear).

DCOS-41842:
1. Go to nodes page.
2. Type something in the filter.
3. Click "Clear".
4. Verify that the filter is empty.
5. Select something from the filter dropdown (for example "healthy" or "unhealthy").
6. Click "Clear".
7. Verify that the filter is empty.
8. Verify the same filter is still applied after refreshing the page.
9. Verify that the added test makes sense.

DCOS-42506:
1. Go to nodes page.
2. Switch to graph view.
3. Apply a filter using the dropdown.
4. Switch to grid view.
5. Verify that the filter is kept.
6. Verify that the added test makes sense.

## Trade-offs
None.

## Dependencies
None.

### Before
Peek 2019-06-25 14-01

### After
Peek 2019-06-25 14-01-2

### Before
Peek 2019-06-25 13-53

Peek 2019-06-25 13-50

### After
Peek 2019-06-25 13-53-2

Peek 2019-07-17 20-14

### Before
Peek 2019-06-25 13-47

### After
Peek 2019-07-17 20-14-2


GeorgiSTodorov:
@DanielMSchmidt , for the case of switching between the different modes the filter is being reset on master as well. I only reset the dropdown selection so that it's clear that we are not filtering.

GeorgiSTodorov:
Same for visiting the site with filters. Right now if you apply a filter, for example "unhealthy", copy the URL and open it in another tab, the filter will show "unhealthy", but the table will not be actually filtered.

DanielMSchmidt:
I am aware and the bug ticket points out that this is an inconsistency. Instead of fixing it towards the "clear the textfield" side we should resolve the inconsistency by applying the filters that are set via query params

GeorgiSTodorov:
>  Instead of fixing it towards the "clear the textfield" side we should resolve the inconsistency by applying the filters that are set via query params

@DanielMSchmidt , why do you think that we should?

DanielMSchmidt:
I think we should take what the URL gives us as query parameters and put the into the DSL form and apply the filter. And I think if there is something missing as an URL query parameter we should add it. In the end I would like to be able to deep link to a specific search here. By taking the URL as a single source of truth we can actually remove quite some complexity I would imagine.

But that's only my opinion, @nLight what do you think is the right mix  of customer first / ship it here?

GeorgiSTodorov:
@DanielMSchmidt , I don't know if the filters ever worked as you described. If they didn't, it's not a bug, but a feature and out of scope here.

DanielMSchmidt:
If they never did they were build with this bug. Because there is no explanation why you should change the URL to reflect your filtering if you don't want to enable anyone to deep link to this filter

GeorgiSTodorov:
Even though the jira was created by Fabs, the bug was reported by me. And the bug I saw was that we didn't clear the filtering. If the URL thing is a bug, it is a different bug for me.

GeorgiSTodorov:
I talked with @nLight and he says we should fix the filter when visiting a URL with a filter.

pierrebeitz:
to me it feels like we have to unravel a lot of logic to start the data-flow from the routing and also make the selected-framework-filter a part of the DSL-filter (as suggested in DCOS-41310). that's going to be a bigger refactoring but i think it will be well worth it. 

in the meantime the current behaviour is quite a bit better than on master although some things i'd suspect to work (select-framework-via-url, updating the url updates the filters) and that would fit the context of these issues could still be addressed.

while digging into the code [some cleanups/changes emerged](https://github.com/dcos/dcos-ui/tree/pb/DCOS-41310-nodes-dsl-filter-additions), which you might want to have a look at. if you think it does not make sense to incorporate those, i'd save them for a spike day.

GeorgiSTodorov:
@pierrebeitz , I definitely agree with the need for a refactoring, but I don't think that a bug fix PR is the right place for this.

GeorgiSTodorov:
I will take care of the file conflict tonight or tomorrow.

GeorgiSTodorov:
@natmegs , thank you for the review.
> For me it shows the correct number of nodes, but it doesn't show the actual node in the table (this is using fixtures to have more options to choose from)

This shouldn't happen, I will take a look.

> The filters do not refresh on reset for me as they seem to be persisted in the url (which is a good thing I think? Just wanted to check)

Yes, this is expected, I hadn't updated the testing instructions.

> (^ Switching between graph and table). Filter is not reset when I switch from graph -> table

Same here, it's expected, I didn't update the testing instructions.

mperrotti:
Everything behaves as expected, and the changes look good to me.

I noticed one potential issue, but I think it may just be an issue with fixtures:
I'm not sure if the node count is correct. When I use fixtures and filter by "Draining" status, it says "300 out of 5000". Then if I add the "Unhealthy" filter, it still says "300 out of 5000". See screen recording: https://cl.ly/f07e6ff9099f

GeorgiSTodorov:
@mperrotti , I also think that this is related to fixture and is not a real issue. For me, only if we observe something similar happening on a real cluster (soak for example), then we should be alarmed.

PR #3947, opened 2019-06-24
DCOS-55627: show icons for healthy and unhealthy statuses

mperrotti:
## Testing
Either browse around tasks in a soak cluster to find services with health checks, or create services on the daily cluster with health checks that will fail and health checks that will pass.

Confirm that the "Health" column shows icons instead of just colored dots to indicate status when the service is healthy or unhealthy

## Screenshots
Before:
Screen Shot 2019-06-24 at 3 43 38 PM

After:
Screen Shot 2019-06-24 at 3 39 54 PM


GeorgiSTodorov:
If you want to have a successful health check, add a command `exit 0` in the health check tab. If you want unsuccessful health check, add a command `exit 1`.

PR #3949, opened 2019-06-25
DCOS-55092 show past jobRun tasks

pierrebeitz:
## Testing

Verify that we now can access the logs for past JobRuns. Those should have an associated task now that can be clicked to view the runs details along with its logs.

OS Enterprise 2019-06-25 12-16-04

For tasks already cleaned up by Mesos, we now show this (without the "Sorry"): 

OS Enterprise 2019-06-27 15-12-32

Design has been informed of those states and might come up with follow ups to polish messages/design.

## Trade-offs

I had some trouble to find the correct place to make those changes and threw away some dead code along the way. I'm not sure whether we want that to be reviewed right now (although i think it's not a whole lot of work to confirm that the stuff thrown away is not used anymore) or whether we just want to merge the added functionality right away and leave all the dead code elimination for some time later.

## Review

commit by commit... the second to last one contains the interesting parts.


plugins/jobs/src/js/types/Job.ts
```diff
@@ -87,138 +86,80 @@ type Job {
...
-  },
-  mem(job: MetronomeGenericJobResponse): number {
-    return job.run.mem;
-  },
-  cpus(job: MetronomeGenericJobResponse): number {
-    return job.run.cpus;
-  },
-  activeRuns(job: MetronomeGenericJobResponse): JobRunConnection | null {
-    return isMetronomeJobDetailResponse(job)
+    activeRuns: isMetronomeJobDetailResponse(job)
```
DanielMSchmidt: We could have used real field resolvers now, we would write in the job Model (like here: https://github.com/dcos/dcos-ui/blob/82b6efa19fefbfcc8a91201ad9de9ab47109023f/plugins/jobs/src/js/data/JobModel.ts#L154)

```ts
return {
  Job: {
    activeRuns(job) {
      return isMetronomeJobDetailResponse(job) ? job.activeRuns : null
    }
  }
}
```

But I think this is out of scope for this PR. Maybe we can use a spike day to make this pretty

Also Kudos for creating separate commits, this makes the review way easier

src/js/components/modals/JobStopRunModal.js
```diff
@@ -35,8 +36,15 @@ class JobStopRunModal extends mixin(StoreMixin) {
...
+
     if (selectedItems.length === 1) {
-      MetronomeStore.stopJobRun(jobID, selectedItems[0]);
+      const jobRunID = selectedItems[0];
+
+      if (jobID == null || jobRunID == null) {
+        return;
+      }
+
+      MetronomeClient.stopJobRun(jobID, jobRunID).subscribe();
```
DanielMSchmidt: Do we need a take(1) or something to make sure it finishes and no subscription is tangling in the air?

GeorgiSTodorov:
@mperrotti is clearing up pretty much the same files in https://github.com/dcos/dcos-ui/pull/3934/files .

pierrebeitz:
@GeorgiSTodorov 	wow, great eye for detail!  
as there won't be conflicts as long as both PRs just delete those files, i think we don't need to sync on that. 🙇

GeorgiSTodorov:
@pierrebeitz , there will be conflicts. For example, you are changing `JobStopRunModal` and Mike is deleting it.

src/js/components/modals/JobStopRunModal.js
pierrebeitz: i had a similar concern although i believe i only moved code on a structural level.  
the observable from http-service's `request` `complete`s once the request is finished. so we don't need to make doubly sure here.
if there's some edge case that would still requires us to jump through hoops here i think that should be fixed on a library-level to not convolute the semantics of a "request" (as i understand them 😊)

pierrebeitz:
@GeorgiSTodorov 	thank you so much for preventing conflict-resolving-frustration!! 🙇 
I'll sync with Mike as soon as possible.

src/js/components/modals/JobStopRunModal.js
DanielMSchmidt: 👍

plugins/jobs/src/js/types/Job.ts
pierrebeitz: @DanielMSchmidt can i just change that stuff to the syntax you're outlining and everything will work automagically or do i need to think throughout the process.

Also thanks for requesting a system test! I'll add one.

DanielMSchmidt: If you move to this syntax it should work automatically. parent is the object returned by the parent field resolver, you also have access to args and context.

pierrebeitz:
i synced up with mike and we are pretty sure that our changes do not conflict, so i removed the `hold merge`-label again

tests/pages/jobs/JobDetails-cy.js
```diff
@@ -79,6 +79,29 @@ describe("Job Details", function() {
         expect($children.length).to.equal(20);
       });
     });
+
+    it("expands a row for already finished job runs", function() {
+      cy.get(".page table tbody tr .is-expandable").as("tableRows");
+      console.log(cy.get("@tableRows").eq(2));
```
GeorgiSTodorov: Forgotten console.log.

GeorgiSTodorov:
It works and I like it, but I find it weird that when we expand, we show the status twice, and sometimes the two statuses are different. I think this might be confusing.
Снимка от 2019-06-27 10-28-36

pierrebeitz:
@GeorgiSTodorov 	we currently also show the status twice for active runs ("Running"). 
The logic to show the statuses has already been in place and looked reasonable to me. We have a distinction between `RunStatus` and `TaskStatus` - that's why they can be different ("Completed" vs. "Finished").

I'd propose to merge this as is - as the value of being able to access the logs of past runs outweighs the mentioned oddity IMO - while i throw this at design, so they can come back at us if they also think that this is confusing, ok?

pierrebeitz:
i just noticed that we can't look at tasks that are older than 6 hours on SOAK. maybe mesos wipes information for old tasks?

pierrebeitz:
I added a notice to a least let users know why they can't see detailed information for old tasks. Also see the screenshot in the updated description.

GeorgiSTodorov:
Should we run the new message by the designers?

pierrebeitz:
As mentioned in the updated description i already informed them. Do you think we should wait for feedback?

My approach would be to merge this and wait until they have something up their sleeves (which might be the wrong approach) as i'm pretty sure that message won't offend anyone.

GeorgiSTodorov:
I am a little skeptic if the "sorry" is a good idea, but it's not like it will break anything.

GeorgiSTodorov:
IP Addresses seems to be always empty. Is this known?
Снимка от 2019-06-27 16-39-45

PR #3950, opened 2019-06-25
feat: link to support portal

nLight:
I figured it doesn't make sense to run this change through the planning pipeline

## Screenshots

Screenshot 2019-06-25 at 13 22 20

@izharbuendia would you please check the wording? Is it ok to have `Support` sub-header next to `Support Portal` link? Thanks


locale/en/messages.json
```diff
@@ -572,7 +572,6 @@
   "Please try again later.": "",
   "Please try again.": "",
   "Pod": "",
-  "Pods": "",
```
nLight: Looks like a forgotten piece. After lingui upgrade we need to `npm i`

PR #3952, opened 2019-06-25
DCOS-50792: TypeError: Cannot read property 'jsonValue' of null

GeorgiSTodorov:
Don't crash the UI if getNewJsonState
inside handleChange is null

Closes https://jira.mesosphere.com/browse/DCOS-50792

## Testing
1. In `/src/js/components/JSONEditor.js` in `getNewJsonState` change line 313 from `if (this.jsonText === jsonText) {` to `if (this.jsonText !== jsonText) {`.
2. Open the JSON editor in the jobs form.
3. Change something in the editor.
4. Verify that no console error is shown.
5. Verify that the added test makes sense.

## Trade-offs
None.

## Dependencies
None.

### Before
Screenshot from 2019-06-25 16-29-12

### DIFF
--- a/src/js/components/JSONEditor.js
+++ b/src/js/components/JSONEditor.js
@@ -226,7 +226,12 @@ class JSONEditor extends React.Component {
     const lastError = this.jsonError;
 
     // Calculate what the next state is going to be
-    const { jsonValue, jsonMeta, jsonError } = this.getNewJsonState(jsonText);
+    const getNewJsonState = this.getNewJsonState(jsonText);
+    if (getNewJsonState === null) {
+      return; // No change.
+    }
+
+    const { jsonValue, jsonMeta, jsonError } = getNewJsonState;
 
     // Update the `isTyping` flag
     this.isTyping = true;
--- a/src/js/components/__tests__/JSONEditor-test.js
+++ b/src/js/components/__tests__/JSONEditor-test.js
@@ -170,6 +170,17 @@ describe("JSONEditor", function() {
 
       expect(onChangeHandler).toBeCalledWith(JSON.parse(validJSONText));
     });
+
+    it("does not crash when getNewJsonState is null", function() {
+      const instance = ReactDOM.render(
+        <JSONEditor value={123} />,
+        thisContainer
+      );
+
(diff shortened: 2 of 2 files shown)


src/js/components/JSONEditor.js
```diff
@@ -226,7 +226,16 @@ class JSONEditor extends React.Component {
     const lastError = this.jsonError;
 
     // Calculate what the next state is going to be
-    const { jsonValue, jsonMeta, jsonError } = this.getNewJsonState(jsonText);
```
pierrebeitz: would providing a default value in `handleChange` do as well?

`  handleChange(jsonText = "") {` or 

```
  handleChange(jsonText) {
    if (typeof jsonText != "string") {
      jsonText = ""
    }
    ...
```

GeorgiSTodorov: This might work. Do you think the current solution is lacking in some way?

pierrebeitz: yes, i do think it is lacking in clarity. 

the reason is that `getNewJsonState` is quite convoluted. it's either returning the object you resembled or a `null`.

i'd now expect consumers to distinct between the object and `null`.

so we could write 
```
const newJsonState = this.getNewJsonState(jsonText);
if (newJsonState === null) {
  return;
}
const { jsonValue, jsonMeta, jsonError } = newJsonState;
...
```

so we would need to introduce 4 lines to properly and explicitly handle a `null` (which is mentioned in the docs of `getNewJsonState`).

instead we now provide an empty default object to destruct it and return in case each value has not changed between lines 230 and 236. while this is relatively simple to see here it still requires a lot more mental overhead than necessary to understand the intention of the author here.

------------

IF we ever want to fix the root of that complication we'd need to narrow down the interface of `getNewJsonState` from `getNewJsonState(jsonText: string | null): ThatObject | null` to `getNewJsonState(jsonText: string): ThatObject`. 

To do that we should handle `null`s outside of that function and that's why i asked the original question.

GeorgiSTodorov: Okay, I agree with you. I will change it.

PR #3953, opened 2019-06-25
DCOS-45226: fix(jobsform): fix jobs form to preserve docker params

Poltergeist:
Docker params are now preserved although no docker image has been provided.

CLOSES DCOS-39298

## Testing

1. Add new jobs and enter docker parameters. Open edit page of the job. The docker parameter fields are blank.
2. Open add job or edit job page. Enter data in the docker parameters fields. Switch to json mode. The entered parameter are not visible. Switch back to forms view. The docker parameters fields are blank.

### DIFF
--- a/src/js/utils/JobUtil.js
+++ b/src/js/utils/JobUtil.js
@@ -46,7 +46,7 @@ const JobUtil = {
         id: null
       },
       labels = {},
-      docker,
+      docker = {},
       docker_parameters,
       schedule
     } = formModel;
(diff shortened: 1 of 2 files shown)


src/js/utils/JobUtil.js
```diff
@@ -72,16 +72,22 @@ const JobUtil = {
...
     });
 
-    if (docker && docker.image) {
-      Object.assign(spec.run, { docker });
-      if (docker_parameters != null && docker_parameters.items != null) {
-        spec.run.docker.parameters = docker_parameters.items;
-      } else {
-        spec.run.docker.parameters = [];
-      }
+    if (docker_parameters != null && docker_parameters.items != null) {
```
DanielMSchmidt: Why not `!===` ?

src/js/utils/JobUtil.js
```diff
@@ -72,16 +72,22 @@ const JobUtil = {
...
+    } else {
+      spec.run.docker.parameters = [];
+    }
+
+    if (
+      spec.run.docker &&
+      !spec.run.docker.image &&
+      spec.run.docker.parameters.length === 0
+    ) {
+      delete spec.run.docker;
```
DanielMSchmidt: Why do we want to delete it?

Poltergeist: If we don't delete it if the user added parameters there is no way they are able to produce a valid spec without touching the JSON.

src/js/utils/JobUtil.js
Poltergeist: I always using not the strict comparison for null since it will catch more cases. For example on an object these two cases yield different examples and I want to cover the second case.
```JS
({}).a === null; // false
({}).a == null; // true
```

GeorgiSTodorov:
@Poltergeist , it's failing now.

PR #3955, opened 2019-06-26
DCOS-50259: Possible exception in the Services Form Constraints reducer

GeorgiSTodorov:
Check if a variable exists before trying to
acccess its property.

Closes https://jira.mesosphere.com/browse/DCOS-50259

## Testing
Verify that the added checks make sense for the given sentry error.
Verify that the added test makes sense.

## Trade-offs
None.

## Dependencies
None.

## Screenshots
None.


GeorgiSTodorov:
@pierrebeitz ,  For me in this case it's not worth it to invest time into investigating an edge case that was only in 1.12 and stopped appearing three months ago. But if you have an idea how we can learn more about this, I would be glad to pair.

pierrebeitz:
@GeorgiSTodorov 	ok, i can see your point. why are we fixing something in master, if we have only seen this error for `1.12`? i'd personally would go with rather not silently swallowing an error and just closing the issue then.

GeorgiSTodorov:
@nLight suggested that if something can still happen on master, we could take care of it.

pierrebeitz:
it happens when you e.g. add an empty object in constraints via the JSON-editor: `constraints: [{}]`.

we should definitely think of using decoders for those inputs at some point in time. i'm convinced that it's not worth the effort to make for a proper error message.

the bad news is that the form still crashes - even with this fix.

GeorgiSTodorov:
I will take a look.

GeorgiSTodorov:
@pierrebeitz , I tried adding such an object in the jobs JSON form and it didn't crash on master. Can you please give me more detailed reproduction steps?
Peek 2019-07-02 14-37

pierrebeitz:
here's it: 
screencast 2019-07-02 13-44-33

fortunately i retested on your branch as well and it seems to work. sorry for the hassle!

GeorgiSTodorov:
Okay, you were using the services form.

GeorgiSTodorov:
@pierrebeitz , thank you for figuring out the real world scenario where this happens and testing it.

PR #3958, opened 2019-06-27
DCOS-49826: TypeError: Cannot read property 'includes' of null

GeorgiSTodorov:
Check if a property is null before
using the include method.

Closes DCOS-49826

## Testing
Verify that the added check makes sense given the sentry error.

## Trade-offs
Sentry bug has disappeared, so we can't really know the scenario in which this happened.

## Dependencies
None.

## Screenshots
None.


GeorgiSTodorov:
The related store has already been deleted.

PR #3959, opened 2019-06-27
DCOS_OSS-867: Unspecific GUI Error Message

GeorgiSTodorov:
Display an error even when the response is empty.

Closes https://jira.mesosphere.com/browse/DCOS_OSS-867

## Testing
1. In Chrome, enable request blocking and block `localhost:4200/package/search`.
2. Go to Catalog tab.
3. Verify that the error message is not empty.
4. Verify that the added test makes sense.

## Trade-offs
None.

## Dependencies
None.

### Before
Снимка от 2019-06-27 12-52-18

### After
Снимка от 2019-06-27 16-44-29


GeorgiSTodorov:
@izharbuendia , what do you think about the message?

GeorgiSTodorov:
I combined two existing messages.

src/js/components/CosmosErrorMessage.js
```diff
@@ -43,7 +43,17 @@ class CosmosErrorMessage extends React.Component {
       );
     }
 
-    return <Trans id={message} render="span" />;
+    if (message) {
+      return <Trans id={message} render="span" />;
+    }
+
+    i18nMark("Looks Like Something is Wrong. Please try again.");
```
TattdCodeMonkey: We don't need this with the Trans component on line 53, your "marking" this string twice essentially.

PR #3960, opened 2019-06-27
DCOS-55012 Maintenance Status Column

pierrebeitz:
soon DCOS will tell us whether a node is currently under maintenance or
deactivated. To prepare, we're extending the fixtures, add a check of whether
this feature seems to be present and add an according status column in the
nodes table.

OS 2019-07-01 11-11-33

# Review

Enable fixtures to see the feature in action!

Please review commit by commit, as we clean up and refactor before the actual feature is implemented.

As always with my PRs: if you feel that some commits should not be in here, please tell me and i'll take them out timely if feasible.

# Tradeoffs

I somehow decided to go with a "feature detection" approach, so currently we try to find a node that has either "drain_info" or "deactivated" to determine whether we have a DCOS that supports maintenance. We could also want to go with a feature flag... i'm not sure.

Also i used an inline style (please don't hit me!) as there seems to be no component that encapsulates an icon with a text within a cell, or is there?


GeorgiSTodorov:
Failing test is flaky, retriggering.

GeorgiSTodorov:
When I open the nodes table:
Screenshot from 2019-06-28 11-17-58

I am not sure how the tests run fine.

pierrebeitz:
i retrieved `publicIp`, while it should have been `publicIP`. that could very well lead to the issue you're observing. please try again now. :bowing_man:

GeorgiSTodorov:
It works now.

plugins/nodes/src/js/components/NodesTable.tsx
```diff
@@ -242,18 +216,10 @@ export default class NodesTable extends React.Component<
           />
         }
         cellRenderer={ipRenderer}
-        minWidth={
-          TableUtil.isColWidthCustom(columnWidthsStorageKey, "host")
-            ? undefined
-            : 120
-        }
+        minWidth={hasCustomWidth("host") ? undefined : 120}
```
GeorgiSTodorov: I am having trouble understanding how this works. Why is minWidth undefined?

plugins/nodes/src/js/components/NodesTable.tsx
```diff
@@ -82,25 +96,26 @@ export default class NodesTable extends React.Component<
...
+        return sorter(compareByRegion);
       case "zone":
-        return zoneSorter;
+        return sorter(compareByZone);
       case "health":
-        return healthSorter;
+        return sorter(UnitHealthUtil.getHealthSortFunction);
       case "tasks":
-        return tasksSorter;
+        return sorter((a, b) => getTasks(a) - getTasks(b));
```
GeorgiSTodorov: Non-blocking, but I think by removing some code, we might be making this harder to understand.

plugins/nodes/src/js/components/NodesTable.tsx
pierrebeitz: please note that i did not change semanticts here but only aliased    `TableUtil.isColWidthCustom(columnWidthsStorageKey, column)`.

i think ui-kit wants minWidth to not be present when the user chose a custom width which could be smaller than minWidth. you could try to ask Mike or Philipp about this.

pierrebeitz:
i forgot to squash a fixup-commit... pushing again 🤦‍♂

plugins/nodes/src/js/pages/nodes/nodes-table/NodesTableContainer.js
```diff
@@ -122,9 +122,17 @@ class NodesTableContainer extends mixin(StoreMixin, QueryParamsMixin) {
     const { nodeHealthResponse, filteredNodes, masterRegion } = this.state;
     const { networks = [] } = this.props;
 
+    // Detecting whether DCOS already supports maintenance mode.
+    // We might want to remove this flag at some point in the future.
+    const hasMaintenanceMode = CompositeState.getNodesList().list.some(
```
brandonc: I'm wondering if we should centralize these types of conditions and use a context and context consumer to get these flags from anywhere in the hierarchy 🤔

src/js/structs/Node.js
```diff
@@ -1,3 +1,5 @@
+import * as Status from "#PLUGINS/nodes/src/js/types/Status";
```
brandonc: Minor issue but this feels like a backwards dependency. Our node struct is in the main bundle and referenced by other main modules as well as placement and services plugins-- should the [Node]Status type also be there?

I don't know the correct answer to this.

plugins/nodes/src/js/components/NodesTable.tsx
```diff
@@ -108,7 +108,9 @@ export default class NodesTable extends React.Component<
       case "zone":
         return compareByZone;
       case "health":
-        return UnitHealthUtil.getHealthSortFunction;
```
brandonc: is anything lost by bypassing this function?

src/js/structs/CompositeState.js
```diff
@@ -74,6 +74,23 @@ class CompositeState {
...
+    if (this._isDisabled()) {
+      storedState = newData;
+
+      return;
+    }
+    if (newData == null) {
+      return;
+    }
+
+    this.data = { ...this.data, ...newData };
```
brandonc: The only difference with addState seems to be that this is preserving all keys. Which keys are missing and can they be added to the PRESERVE_KEYS array?

plugins/nodes/src/js/columns/NodesTableStatusColumn.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+import { Status } from "../types/Status";
+
+import Node from "#SRC/js/structs/Node";
+
+const StatusCell = React.memo(({ status }: { status: Status }) => (
+  <Cell>
+    <span>
+      <Icon {...status.icon} size={iconSizeXs} />
+      <span style={{ marginLeft: "7px" }}>{status.displayName}</span>
```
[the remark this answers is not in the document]
pierrebeitz: couldn't find one: https://github.com/dcos-labs/ui-kit/search?q=7px&unscoped_q=7px

@mperrotti?

src/js/structs/Node.js
```diff
@@ -22,6 +24,26 @@ class Node extends Item {
     return this.get("framework_ids");
   }
 
+  getStatus() {
+    const drainInfo = this.get("drain_info");
+    const deactivated = this.get("deactivated");
+    const active =
```
brandonc: In addition, active is a specific field. I wonder how it changes when drain_info is set.

brandonc: Might be worth a unit test in order to keep it clear

plugins/nodes/src/js/columns/NodesTableStatusColumn.tsx
```diff
@@ -0,0 +1,25 @@
...
+  </Cell>
+));
+
+export function statusRenderer(data: Node): React.ReactNode {
+  const status = data.getStatus();
+
+  return <StatusCell status={status} />;
+}
+
+export const compareByStatus = (a: Node, b: Node): number =>
```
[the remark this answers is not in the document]
pierrebeitz: for no good reason. i now inlined all but `compareByIp`, which made for another small cleanup: https://github.com/dcos/dcos-ui/commit/214d8076754734345c6665f401044fbacdf93c83. thanks!

plugins/nodes/src/js/pages/nodes/nodes-table/NodesTableContainer.js
pierrebeitz: i'm not sure. i'd rather have a static feature flag or a feature-flag-module that allows for runtime-toggling. 🤔 

i think there's nothing too bad here. having to look through the list naively admittedly feels bad although it should terminate on the first entry for DCOS'es having that feature.

good enough for now?

plugins/nodes/src/js/components/NodesTable.tsx
pierrebeitz: no. we know exactly by what we want to have our data sorted. and we reach for exactly that function.

also note that `getHealthSortFunction` does not do its job on master.

src/js/structs/CompositeState.js
pierrebeitz: i updated the commit and hopefully made things a lot more explicit. the convolution here was that someone wanted to be able to set all data via `addState`. we now handle `master_info` explicitly.

src/js/structs/Node.js
pierrebeitz: i also felt bad when i introduced this but then forgot about ironing it out.

we now pass a complete node to the status-module, so that it can figure out the status based on the information it gets from the node.

src/js/structs/Node.js
pierrebeitz: i added a unit test reflecting the information i got from Joseph.

GeorgiSTodorov:
When sorting by status, we sort Drained, Draining, Deactivated, Active, Unknown. What is the reason for this?

pierrebeitz:
> Edit: I saw that we are sorting by priority, but are we sure that these priorities are up to date?

That's the order i got from Brandon. I'll make a request to design while proposing to not let this be blocking as it can be easily changed.

plugins/nodes/src/js/types/Status.ts
```diff
@@ -1,20 +1,45 @@
...
+  priority,
+  icon,
+  displayName: i18nMark(name)
+});
+
+const active = build(2, StatusIcon.SUCCESS, "Active");
+const deactivated = build(3, StatusIcon.STOPPED, "Deactivated");
+const drained = build(5, StatusIcon.STOPPED, "Drained");
+const draining = build(4, StatusIcon.LOADING, "Draining");
+const unknown = build(1, StatusIcon.ERROR, "Unknown");
```
GeorgiSTodorov: Shouldn't 1 be the highest priority? Also, a small thing, but I think it would help with readability if these were sorted.

GeorgiSTodorov:
Good catch, I forgot about translations.

PR #3963, opened 2019-06-28
DCOS-55631 Quota Overview

TattdCodeMonkey:
Closes https://jira.mesosphere.com/browse/DCOS-55631 .

## Testing

1. Login to DC/OS UI
2. Update your `Config.dev.ts`
```
uiConfiguration: {
...
  features: {
    quota: true
  }
},
  useFixtures: true,
  ...
```
3. Go to services and open the Quota tab.
4. Verify that the table is shown correctly and follows the design - exact columns, percentages, tooltip when hovered, "no limit" state, etc.
5. Verify that the table is sortable.

## Trade-offs
- Actions are not in scope here, so actions column won't be rendered now.

- The table has a weird transition when sorting.
Peek 2019-07-02 11-12

## Dependencies
None.

## Screenshots
Screenshot from 2019-07-08 17-51-58


plugins/services/src/js/columns/QuotaOverviewCPUConsumedColumn.tsx
```diff
@@ -0,0 +1,64 @@
...
+import { Cell, Tooltip } from "@dcos/ui-kit";
+
+import { ServiceGroup } from "#PLUGINS/services/src/js/types/ServiceGroup";
+import ProgressBar from "#SRC/js/components/ProgressBar";
+import * as ResourcesUtil from "#SRC/js/utils/ResourcesUtil";
+import { findNestedPropertyInObject } from "#SRC/js/utils/Util";
+
+const className = `color-${ResourcesUtil.getResourceColor("cpus")}`;
+
+interface Quota {
```
GeorgiSTodorov: Do we have a place for this?

mperrotti:
@GeorgiSTodorov - I can't reproduce that progress bar bug on the Quota table or on the Nodes table.

There's one difference between Nodes and Quota sorting though: I see a subtle animation on the filled part the progress bar in the Quota table when sorting

plugins/services/src/js/columns/QuotaOverviewGPUConsumedColumn.tsx
```diff
@@ -0,0 +1,64 @@
...
+            data={ProgressBar.getDataFromValue(
+              getGPUConsumedPercent(gpuQuota),
+              className
+            )}
+            total={100}
+          />
+          <div className="table-content-spacing-left">
+            <Tooltip
+              id="quota-gpu-tooltip"
+              trigger={<span>{getGPUConsumedPercent(gpuQuota)}%</span>}
```
mperrotti: We don't need to include the "%" in translation, right?

GeorgiSTodorov:
@mperrotti , what do you mean with progress bar bug? The subtle animation should be the progress bar bug.

plugins/services/src/js/columns/QuotaOverviewGPUConsumedColumn.tsx
GeorgiSTodorov: Good point. I don't know how the percentage sign is displayed in other languages. For example, in Chinese they might say %number or something different.

GeorgiSTodorov: And this is not even a translation, it's just a tooltip.

plugins/services/src/js/filters/__tests__/GenericObject-test.ts
```diff
@@ -0,0 +1,50 @@
...
+      expect(filterByObject({ foo: 100 }, { foo: 0 })).toEqual(false);
+      expect(filterByObject({ foo: 100 }, { foo: 100 })).toEqual(true);
+    });
+    it("can filter by boolean", () => {
+      expect(filterByObject({ foo: true }, { foo: false })).toEqual(false);
+      expect(filterByObject({ foo: true }, { foo: true })).toEqual(true);
+    });
+    it("can filter for undefined", () => {
+      expect(filterByObject({ foo: true }, { foo: undefined })).toEqual(false);
+      expect(filterByObject({ foo: true }, { bar: undefined })).toEqual(true);
```
pierrebeitz: i admire the complexity of `filterByObject` but have no idea what it does.

i imagine jest's `objectContaining`?

in particular i don't understand this very test. why would i want this to be true, if my `foo` does not contain a `bar`? i would want the key to be present, even if its value is undefined.

Currently we only ever search for `{ quota: { enforced: true } }`, right?

GeorgiSTodorov:
Hm, the integration test still doesn't work.

plugins/services/src/js/filters/__tests__/GenericObject-test.ts
pierrebeitz: if it wasn't for that case we could do something like: 

```typescript
type Rec = Record<string, unknown>;

export function filterByObject(value: Rec, filterObj: Rec): boolean {
  const keys = Object.keys(value || {});

  return Object.keys(filterObj || {}).every(k =>
    keys.includes(k) ? matchValue(value[k], filterObj[k]) : false
  );
}

const matchValue = (a: unknown, b: unknown) => {
  if (isRecord(a) && isRecord(b)) {
    return filterByObject(a, b);
  }
  return a === b;
};
const isRecord = (a: unknown): a is Rec => typeof a === "object" && a !== null;
```

this makes all but the `undefined` test pass.

TattdCodeMonkey: I added `filterByObject` so we could easily send a filter object to the graphQL query. We currently only need it to get groups with quota enforced for the overview page, but that is temporary for MVP, we would want to show all groups in that view in the future.

And I didn’t want to have a single Boolean on a group query because that would of been too specific. So I tried to write filterByObject as something more generic that can be expanded upon. But also not to invest too much time now since we only needed to filter by a single sub property value.

Maybe I overdid it there but I was trying to balance current feature requirements with the spirit of graphql and future needs.

tests/_fixtures/config/no-plugins.json
```diff
@@ -1,7 +1,10 @@
 {
   "uiConfiguration": {
+    "features": {
+      "maintenance": true
```
GeorgiSTodorov: I think the integration test is failing, because this has to include Quota as well.

pierrebeitz: oh boy -.-
copy-paste...

GeorgiSTodorov: It should be fine now, I rebased.

plugins/services/src/js/filters/__tests__/GenericObject-test.ts
pierrebeitz: i added a refactor-commit changing the behaviour in the mentioned case because i think that if a key with an `undefined` value has significance in

```
filterByObject({ foo: true }, { foo: undefined })
```

it should also have significance in

```
(filterByObject({ foo: true }, { bar: undefined })
```

If you don't like it, feel free to throw away that commit again.

plugins/services/src/js/data/groups/__tests__/index-test.ts
```diff
@@ -121,6 +121,144 @@ describe("Services Data Layer - Groups", () => {
         m.expect(dl.query(query, {}).pipe(take(1))).toBeObservable(expected$);
       })
     );
+
+    it(
+      "handles query for filtering enforce quota",
+      marbles(m => {
+        const marathonServiceTree = makeServiceTree(marathonGroups);
+        const roles$ = m.cold("(a|)", {
```
pierrebeitz: no action items here, just curious:
why do we mock this with a marble? wouldn't `of` be more straightforward?

plugins/services/src/js/data/groups/__tests__/index-test.ts
```diff
@@ -121,6 +121,144 @@ describe("Services Data Layer - Groups", () => {
...
+          a: {
+            code: 200,
+            message: "OK",
```
pierrebeitz: no action items here, just curious:
what's the significance of `message: "OK"`?

GeorgiSTodorov:
I couldn't figure out how to add fixtures that test the table. I tried adding `.route(/mesos\/roles/, "fx:quota-management/roles");` and similar routes in `tests/_support/index.js`.

GeorgiSTodorov:
Work in progress as I am working on adding more integration tests.

GeorgiSTodorov:
I rebased and pushed a commit that hopefully fixes the broken cypress selectors.

plugins/services/src/js/filters/__tests__/GenericObject-test.ts
```diff
@@ -16,7 +16,7 @@ describe("GenericObject", () => {
     });
     it("can filter for undefined", () => {
       expect(filterByObject({ foo: true }, { foo: undefined })).toEqual(false);
-      expect(filterByObject({ foo: true }, { bar: undefined })).toEqual(true);
+      expect(filterByObject({ foo: true }, { bar: undefined })).toEqual(false);
```
TattdCodeMonkey: you should just remove this line if this doesn't work anymore. This was testing that you could filter by a key *not* being present. But that feature is not required. So we should just remove this.

pierrebeitz: i removed "can filter for undefined". please don't hesitate to make changes to my changes! :)

GeorgiSTodorov:
I have improved the integration tests that we had. If CI is green, this should be reviewable.

pierrebeitz:
@GeorgiSTodorov 	you reintroduced the test i just removed. i guess that was not on purpose?

GeorgiSTodorov:
@pierrebeitz , no.

TattdCodeMonkey:
I did a rebase and had to cherry-pick some commits over to this PR: https://github.com/dcos/dcos-ui/pull/3977

PR #3964, opened 2019-06-28
DCOS-55890 fix lingui templates

TattdCodeMonkey:
Replaced usages of `i18nMark(``)`  with `i18nMark("")`, `i18nMark` does not support string templates and these strings we're not being extracted. 

Translated strings for `ToastNotification` manually in the `ServicesContainer`, the ToastNotification extension for the notification service does not handle any translation, so any text strings sent to it must be translated beforehand. I updated the `ServicesContainer` to get the lingui `i18n` object from the `container` and translate the strings manually.

## Testing

1. Verify strings are extracted now.
2. Test toast notications for delayed start?

## Trade-offs

I add the `i18nTranslate` helper function to the `ServicesContainer` so that tests wouldn't need i18n to exist in the container.


plugins/services/src/js/containers/services/ServicesContainer.js
```diff
@@ -91,6 +92,15 @@ const SERVICE_FILTERS = new DSLFilterList([
 
 const notificationService = container.get(NotificationServiceType);
 
+function i18nTranslate(id, values) {
```
pierrebeitz: does that mean that every containerModule should implement such a function and all marks would have to be wrapped in `i18nTranslate(i18nMark("meh"))` to not couple tests to having a i18n-containerModule loaded?

that does not feel elegant to me. maybe we can tackle that on a spike day.

PR #3967, opened 2019-07-02
[DCOS-55014] add drain status to node details

brandonc:
## Testing

There is a bug that prevents the mesos state store from loading when useFixtures is true (this does not seem to affect integration tests?) Let me know if you are affected by the same issue. If so, just bypass the state store loading by changing NodeDetailPage line 41 from `mesosStateLoaded: false` to `mesosStateLoaded: true`

Visit the node details page (details tab) and see the drain status and new section there

## Dependencies

This PR is currently based on #3960 

## Screenshots

Screen Shot 2019-07-01 at 3 01 25 PM

### DIFF
--- a/plugins/nodes/src/js/pages/nodes/NodeDetailTab.js
+++ b/plugins/nodes/src/js/pages/nodes/NodeDetailTab.js
@@ -2,6 +2,8 @@ import { Trans, DateFormat } from "@lingui/macro";
 import React, { PureComponent } from "react";
 import PropTypes from "prop-types";
 import { request } from "@dcos/mesos-client";
+import { Icon } from "@dcos/ui-kit";
+import { iconSizeXs } from "@dcos/ui-kit/dist/packages/design-tokens/build/js/designTokens";
 
 import { MESOS_STATE_CHANGE } from "#SRC/js/constants/EventTypes";
 import MesosStateStore from "#SRC/js/stores/MesosStateStore";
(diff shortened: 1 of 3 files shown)


GeorgiSTodorov:
Build is failing because of the commit message.

plugins/nodes/src/js/pages/nodes/NodeDetailTab.js
```diff
@@ -57,12 +59,24 @@ class NodeDetailTab extends PureComponent {
...
   }
 
+  hasMaintenanceData() {
+    const { node } = this.props;
+
+    // Detecting whether DCOS already supports maintenance mode.
+    // We might want to remove this flag at some point in the future.
+    return (
+      typeof node.get("drain_info") !== "undefined" ||
+      typeof node.get("deactivated") !== "undefined"
```
GeorgiSTodorov: Why are we checking the type and not just the value?

pierrebeitz: that's a really good question. will change in #3960 as well!

GeorgiSTodorov:
> Let me know if you are affected by the same issue.

I am.

Nitpick: I think it looks better with a slightly smaller margin-left:
Original 7px margin.
Screenshot from 2019-07-02 10-16-31

4px margin.
Screenshot from 2019-07-02 10-16-51

GeorgiSTodorov:
It's failing again. I think it's because it's not pointing to master.

pierrebeitz:
> Nitpick: I think it looks better with a slightly smaller margin-left: 

i used the same paddings as are used in the services table in case someone wants to clean that up at some point. `7px` is pretty nice to grep for.

brandonc:
Going ahead and merging on the basis of previous approvals

PR #3968, opened 2019-07-02
DCOS-55633: Group quota detail page

GeorgiSTodorov:
Quota detail page with cards.

Closes https://jira.mesosphere.com/browse/DCOS-55633

## Testing
0. Start localhost and login.
```
uiConfiguration: {
...
  features: {
    quota: true
  }
},
  useFixtures: true,
  ...
```
2. Go to services page.
3. Open quota tab.
4. Click on a group.
5. Verify that the detailed view for the groups is shown and it looks like the design from the jira.
5.5 (Optional) Test it on smaller window to make sure it looks okay.

## Trade-offs
None.

## Dependencies
https://github.com/dcos/dcos-ui/pull/3977

## Screenshots
Screenshot from 2019-07-04 14-18-26


nLight:
I pulled the PR and see that Service overview action menu is available on the Quota tab but hidden on the Services tab

Screenshot 2019-07-03 at 11 33 19

GeorgiSTodorov:
Work in progress as I am working on adding more integration tests.

plugins/services/src/js/containers/services/ServicesQuotaView.tsx
```diff
@@ -46,13 +46,14 @@ class ServicesQuotaView extends React.Component<ServicesQuotaViewProps, {}> {
       this.context.router.push("/services/overview/create");
     };
     const tabs = this.getTabs();
```
TattdCodeMonkey: We also need to update the tabs when in detail view so that `Services` takes you to the service detail for the group and not back to overview.

plugins/services/src/js/components/ServicesQuotaOverview.tsx
```diff
@@ -27,17 +43,26 @@ const ServicesQuotaOverview = componentFromStream(() => {
...
+      if (id !== EMPTY_ID) {
+        const group = groups.filter(
+          (group: { id: string }) => group.id === id
+        )[0];
+        return <ServicesQuotaOverviewDetail group={group} />;
+      }
+
+      return <ServicesQuotaOverviewTable groups={groups} />;
+    }),
+    catchError(() => of(<div>Error getting groups with Quota</div>)),
```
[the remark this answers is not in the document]
GeorgiSTodorov: I don't think it should be different, but we can ask @izharbuendia .
I thought this was a temporary error message.

plugins/services/src/js/containers/services/ServicesQuotaView.tsx
GeorgiSTodorov: Do we want this for this PR? I don't remember seeing this in the design.

plugins/services/src/js/containers/services/ServicesQuotaView.tsx
```diff
@@ -46,13 +46,14 @@ class ServicesQuotaView extends React.Component<ServicesQuotaViewProps, {}> {
...
-
-    const content = <ServicesQuotaOverview />;
+    const id: string = serviceTree.getId();
+    //@ts-ignore
+    const content = <ServicesQuotaOverview id={id} />;
```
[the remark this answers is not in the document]
GeorgiSTodorov: I don't have a strong opinion.

plugins/services/src/js/containers/services/ServicesQuotaView.tsx
TattdCodeMonkey: I'm not sure if its in the design, but I think that matches the expected UX IMO.

GeorgiSTodorov: Let's ask @izharbuendia .

plugins/services/src/js/containers/services/ServicesQuotaView.tsx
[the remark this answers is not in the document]
pierrebeitz: i tend towards @TattdCodeMonkey's suggestion.

plugins/services/src/js/components/ServicesQuotaOverviewDetail.tsx
```diff
@@ -0,0 +1,132 @@
...
+    this.getCard = this.getCard.bind(this);
+  }
+
+  getCard(resource: string): React.ReactNode {
+    const { group } = this.props;
+
+    const title =
+      resource === "cpus" || resource === "gpus"
+        ? resource.slice(0, -1).toUpperCase()
+        : resource[0].toUpperCase() + resource.slice(1);
```
pierrebeitz: that looks complicated. could we achieve the same result with a dumb lookup?

what would you think of introducing a resource type and a function like the following?
```typescript
const titleFor = (resource: Resource):string => {
  switch (resource) {
    case "cpus": return "CPUs"
    case "gpus": return "GPUs"
    case "disk": return "Disk"
    case "mem": return "Mem"
    // no default here!
  }
}
```

that function would automatically force developers to provide a title for a new resource, if they add one.

plugins/services/src/js/components/ServicesQuotaOverview.tsx
```diff
@@ -9,12 +14,23 @@ import Loader from "#SRC/js/components/Loader";
 import container from "#SRC/js/container";
 
 import ServicesQuotaOverviewTable from "./ServicesQuotaOverviewTable";
+import ServicesQuotaOverviewDetail from "./ServicesQuotaOverviewDetail";
 import EmptyServicesQuotaOverview from "./EmptyServicesQuotaOverview";
 
-const ServicesQuotaOverview = componentFromStream(() => {
+const EMPTY_ID = "/";
```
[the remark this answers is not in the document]
GeorgiSTodorov: Sounds good.

GeorgiSTodorov:
@pierrebeitz , I will address some of your points.

plugins/services/src/js/columns/QuotaOverviewNameColumn.tsx
```diff
@@ -6,17 +6,25 @@ import {
 } from "@dcos/ui-kit/dist/packages/design-tokens/build/js/designTokens";
 import { SystemIcons } from "@dcos/ui-kit/dist/packages/icons/dist/system-icons-enum";
 import * as React from "react";
+import { Link } from "react-router";
 
 export function nameRenderer(group: ServiceGroup) {
+  const serviceLink = `/services/quota/${encodeURIComponent(
+    group.id.toString()
```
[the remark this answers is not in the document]
GeorgiSTodorov: I was using the service table name column as an example. I will change it.

tests/pages/services/QuotaOverview-cy.js
```diff
@@ -39,118 +39,204 @@ describe("Quota Tab", function() {
...
-        .contains(name);
-    }
-
-    function clickHeading(name) {
-      cy.get(".TopRightGrid_ScrollWrapper")
-        .contains(name)
-        .click();
-    }
+    context("Quota table", function() {
+      function getFirstRowName(name) {
```
TattdCodeMonkey: A good spike day task would be to implement these as Cypress commands in `tests/_support` similar to some of the existing form commands we use for service modal tests etc. Not in scope for this ticket though.

GeorgiSTodorov: I like the idea.

plugins/services/src/js/components/ServicesQuotaOverviewDetail.tsx
```diff
@@ -0,0 +1,173 @@
...
+      return "GPUs";
+    case "disk":
+      return "Disk";
+    case "memory":
+      return "Memory";
+  }
+  return "";
+};
+
+function getCard(group: ServiceGroup, resource: string): React.ReactNode {
```
TattdCodeMonkey: I feel like this should be a component outside of this file.

locale/zh/messages.json
```diff
@@ -247,8 +247,6 @@
   "Disk (MiB)": "磁盘 (Mib)",
   "Disk Allocation": "磁盘分配",
   "Disk Consumed": "",
-  "Disk Space (MiB)": "磁盘空间 (MiB)",
```
TattdCodeMonkey: 🤔 we should ensure we want to lose these

GeorgiSTodorov: I checked, they no longer appear.

plugins/services/src/js/containers/services/ServicesQuotaView.tsx
```diff
@@ -46,13 +46,14 @@ class ServicesQuotaView extends React.Component<ServicesQuotaViewProps, {}> {
       this.context.router.push("/services/overview/create");
     };
     const tabs = this.getTabs();
-
-    const content = <ServicesQuotaOverview />;
+    const id: string = serviceTree.getId();
+    //@ts-ignore
```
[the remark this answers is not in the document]
TattdCodeMonkey: FYI I resolved this with a commit when I was reviewing.

plugins/services/src/js/components/ServicesQuotaOverviewDetail.tsx
TattdCodeMonkey: But I'm ok with refactoring this later.

mperrotti:
I tried to review this and the other Quota PR, but I'm getting the "Access Denied" page. @GeorgiSTodorov - could we do a Zoom in the morning and take a look at my `Config.dev.ts`?

plugins/services/src/js/components/ServicesQuotaOverviewDetail.tsx
mperrotti: I agree with both of Rodney's comments

TattdCodeMonkey:
You need to login then change useFixtures to true

GeorgiSTodorov:
@mperrotti , I cannot reproduce that problem neither on Firefox, nor on Chrome. This is how the cards look for me.
Снимка от 2019-07-11 08-51-10

GeorgiSTodorov:
@TattdCodeMonkey , I don't see the breadcrumbs for Quota.

package.json
```diff
@@ -172,7 +172,7 @@
     "semantic-release-jira": "1.2.3",
     "speed-measure-webpack-plugin": "1.2.2",
     "style-loader": "0.20.1",
-    "stylelint": "7.2.0",
+    "stylelint": "^7.13.0",
```
GeorgiSTodorov: Build is failing because of the `^`.

src/styles/components/charts/styles.less
```diff
@@ -23,7 +23,7 @@
     .grid-graph .tick,
     .axis line {
       fill: none;
-      shape-rendering: crispedges;
+      shape-rendering: crispEdges;
```
GeorgiSTodorov: I don't know why this is related, but this was breaking the build.

TattdCodeMonkey: update to `stylelint` most likely. we need to go to > 7.8.0 to get support for `fr` units to use CSS grid.

mperrotti:
@GeorgiSTodorov - that's probably because macOS has a different default font than Ubuntu. The default Ubuntu font may not have an 100 weight.

Can another non-Ubuntu user let me know if they're seeing the problem too?

GeorgiSTodorov:
@TattdCodeMonkey , can you confirm weather for you it looks like it does for Mike? Thanks.

TattdCodeMonkey:
Yes, I'm also seeing 100 font-weight too.

GeorgiSTodorov:
Okay, I will change it tomorrow, either in this PR or in another one.

PR #3970, opened 2019-07-03
DCOS-55588 Node Status Filter

pierrebeitz:
## Testing

* Add a feature flag in `Config.dev.ts`

```
export default {
  uiConfigurationFixture: {
    uiConfiguration: {
      features: {
        maintenance: true,
      },
      //...
}
```
* Enable fixtures
* Go to the NodesTable and filter by Status!

## Reviewing

Commit-by-commit should be way easier than having a look at the complete changes.

Again we have some small cleanups in here. As always: If you feel uncomfortable with any of those we can remove them from this PR.  

## Screenshots

OS 2019-07-03 23-05-59


plugins/nodes/src/js/filters/NodesRegionFilter.ts
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * This filter handles the `region:XXXX` for nodes
+ */
+const filter: DSLFilter = {
+  /*
+   * Handle all `region:XXXX` attribute filters that we can handle.
+   */
+  filterCanHandle: (filterType, filterArguments) =>
+    filterType === DSLFilterTypes.ATTRIB && filterArguments.label === "region",
```
pierrebeitz: while reviewing this myself, i noticed that i may have broken the ability to apply multiple `is:`-filters when regions are involved. will investigate tomorrow.

src/js/structs/DSLFilter.ts
```diff
@@ -0,0 +1,29 @@
...
+  filterApply(
+    resultset: List<Node>,
+    type: FilterType,
+    args: Arguments
+  ): List<Node>;
+}
+
+// we need this as long as you find something when grep'ing for `extends DSLFilter`.
+// it's a backwards-compatibility-measure on our way of converting filters to TS.
+export default class DSLFilter {}
```
natmegs: 2 default exports in this file?

plugins/nodes/src/js/filters/NodesRegionFilter.ts
natmegs: My brain is having a hard time understanding how we are able to drop the `regions.includes`... checks in these functions?

GeorgiSTodorov:
Filtering works for me. Why do we have no filter for the "Unknown" type?

GeorgiSTodorov:
The failing integration test in passing locally, retriggering.

plugins/nodes/src/js/components/NodesTable.tsx
```diff
@@ -188,7 +188,7 @@ export default class NodesTable extends React.Component<
         width={hasCustomWidth("host") ? customWidthFor("host") : undefined}
       />,
 
-      withMaintenanceStatus ? (
+      (Config.features || {}).maintenance ? (
```
GeorgiSTodorov: What does the `|| {}` do? Wouldn't `{}.mainenence` give an error?

pierrebeitz: no, it would give you `undefined`. That's a safeguard for the case that a config has no `features`-prop, which i assume might be common.

plugins/nodes/src/js/filters/NodesRegionFilter.ts
pierrebeitz: My brain is having a hard time understanding why the `regions.includes` was present. 😄

When the `DSLUtil` checks which filters to apply it asks every filters `filterCanHandle`. Our filter now says: "i can handle things with a region-label". It formerly said: "i can handle things with a region-label **and** a region-value that is within all the region-values that occur in the nodes table".

We now got all the filters that claim that they can handle the current `DSLExpression` and call `filterApply` on them. We now do a "show if the region matches `node.getRegionName().toLowerCase()`". We formerly did a "show if the region matches `node.getRegionName().toLowerCase() ` AND that region is an element of all the regions in the `NodesTable`" (which it must be anyway if a node matches on that).

I hope that makes things clearer. If not i think i can try again with more code.

src/js/structs/DSLFilter.ts
pierrebeitz: yes, one for the new type and one for an actual implementation.

GeorgiSTodorov:
I figured out the failing integration test. Feature flags don't work in integration tests as far as I know, so it can't find the status heading. It fails locally as well if we disable the feature flag.

pierrebeitz:
@natmegs 
> When I try to filter, nothing matches for any of the statuses. Is this expected given our current fixtures?

thanks a lot for finding this. we now hide those filters if the maintenance-flag is not set. please add the feature-flag to your `Config.dev.ts` to see the filter and fixtures in action!

pierrebeitz:
needed to update to make feature-flagged stuff run in CI. also i had to fix a couple eslints that were introduced here 😬

src/js/structs/DSLFilter.ts
natmegs: Oh ok, thanks for the explanation. I wonder would this work to make it clear which is actually exported?
```
type Bla = string 
export const Bla = {} 
```

plugins/nodes/src/js/filters/NodesRegionFilter.ts
natmegs: Makes sense - thanks!! 🙌

src/js/structs/DSLFilter.ts
pierrebeitz: This would export the constant but not the type. I'm explicitly seeking to export a type with the same name, so type-signatures read good und imports are not as noisy.

PR #3973, opened 2019-07-05
[DCOS-43441] Remove misleading info

nLight:
Background: the terms themselves misleading i.e. `container/bridge` is not in the list. Or `USER networking` is being called `Virtual network: <name>`. 

I believe it's better to say nothing specific as the info differs cluster to cluster, than to say misleading things.

Closes DCOS-43441

## Testing

Check both single and multi container.

## Trade-offs

Not to loose complete translation over the change I trimmed it to `Refer to the <0>ports documentation</0> for more information.` part

## Screenshots

Screenshot 2019-07-09 at 14 43 13


nLight:
@GeorgiSTodorov I decided to trim the translation not to loose it as depicted in the tradeoffs section :)

GeorgiSTodorov:
Yeah, I saw that, but will the translators know that the string is only half-translated?

nLight:
Removed the redundant `Choose Network Type` as per @izharbuendia suggestion. @GeorgiSTodorov I think now we can safely copy the translation :)

PR #3976, opened 2019-07-08
chore: remove pluginSDKStruct

pierrebeitz:
a byproduct of a review.

pluginSDKStruct is copying the keys of the given config object to a new object-instance (so it's basically copying that object). also we seem to only instantiate it once which makes me even more confident that we don't break anything with this change.


nLight:
@pierrebeitz could it be that we mutate the data somewhere and now we will potentially have side effects?

pierrebeitz:
@nLight 	i thought the same and came to the conclusion that the side effects on an instance would be the same as on a POJO.

PR #3977, opened 2019-07-08
DCOS-55631 Quota Overview

TattdCodeMonkey:
Closes https://jira.mesosphere.com/browse/DCOS-55631 .

## Testing

1. Login to DC/OS UI
2. Update your `Config.dev.ts`
```
uiConfiguration: {
...
  features: {
    quota: true
  }
},
  useFixtures: true,
  ...
```
3. Go to services and open the Quota tab.
4. Verify that the table is shown correctly and follows the design - exact columns, percentages, tooltip when hovered, "no limit" state, etc.
5. Verify that the table is sortable.

## Trade-offs
- Actions are not in scope here, so actions column won't be rendered now.

- The table has a weird transition when sorting.
Peek 2019-07-02 11-12

## Dependencies
None.

## Screenshots
Screenshot from 2019-07-08 17-51-58


plugins/services/src/js/containers/services/ServiceTreeView.js
```diff
@@ -64,6 +65,32 @@ class ServiceTreeView extends React.Component {
     return null;
   }
 
+  getTabs() {
+    const { quota } = ConfigStore.get("config").uiConfiguration.features || {};
+    // Workaround for Cypress.
```
pierrebeitz: a smell. but well... 🙈

plugins/services/src/js/containers/services/ServicesContainer.js
```diff
@@ -688,6 +701,10 @@ class ServicesContainer extends React.Component {
       return this.getServiceTree(item);
     }
 
+    if (currentRoutePath.startsWith("/services/quota")) {
```
pierrebeitz: the render-method could definitely benefit from a tell-don't-ask-refactoring. 🙈

plugins/services/src/js/data/groups/resolvers.ts
```diff
@@ -25,12 +26,28 @@ const isGroupArgs = (
...
   return {
     id: serviceTree.getId(),
+    name: serviceTree.getName(),
     quota: {
       enforced: serviceTree.getEnforceRole() === true
     }
   };
 }
 
+function processServiceGroupFilter(
```
pierrebeitz: i personally don't like the mentality of that function. it claims to work for unknown filter-args, so the signature does not tell you what it is that the function wants. when looking at the implementation it looks like it definitely wants a `string`. when looking further it looks like that string must be `JSON.parse`able or else you'll just get your groups back.

this of course is not a big thing on its own but i think we should try to not introduce code that's dealing with so much uncertainty regarding input and output as stuff tends to get hard to trace down in case a debugging is necessary.

🙈

src/styles/content/tables/styles.less
```diff
@@ -439,6 +439,7 @@
 }
 
 .table-content-spacing-left {
+  display: inline-block;
```
pierrebeitz: this is a workaround for Tooltip rendering a nested `div`-structure. i'm not saying we should address this now, but `.table-content-spacing-left` now definitely does more than its name is suggesting.

we could have added a class that keeps the Tooltip in check to keep things separated.

tests/pages/services/QuotaOverview-cy.js
```diff
@@ -0,0 +1,156 @@
...
+    });
+
+    function getFirstRowName(name) {
+      cy.get(".ReactVirtualized__Grid")
+        .eq(2)
+        .children()
+        .first()
+        .children()
+        .first()
+        .contains(name);
```
pierrebeitz: this selector feels very brittle to me. what do you think of using e.g. a `data-`-property next time to evade the need for such selectors?

tests/pages/services/QuotaOverview-cy.js
```diff
@@ -0,0 +1,156 @@
...
+      clickHeading("Disk Consumed");
+      getFirstRowName("10_apps");
+      getSecondRowName("10000_apps");
+
+      clickHeading("Disk Consumed");
+      getFirstRowName("10000_apps");
+      getSecondRowName("10_apps");
+    });
+
+    it("Sorts the table by GPU consumed", function() {
```
pierrebeitz: i love how those test cases read!

plugins/services/src/js/containers/services/ServiceTreeView.js
GeorgiSTodorov: :man_shrugging:

tests/pages/services/QuotaOverview-cy.js
GeorgiSTodorov: What do you mean with a `data` property?

pierrebeitz: a `data-column=name` on the node we are looking for would make the selector less brittle. we could maybe do something like 

`cy.get(".ReactVirtualized__Grid [data-column=name]").first()` instead of encoding assumptions about the structure of that grid.

GeorgiSTodorov: I agree that the selector is sub-optimal, but our table structure is a bit peculiar. I still don't understand how would `cy.get(".ReactVirtualized__Grid [data-column=name]").first()` work. I don't see a `data-column` property anyway, do you mean that we should add it?

pierrebeitz: yes, that was what i meant to suggest.

plugins/services/src/js/containers/services/ServicesContainer.js
TattdCodeMonkey: I agree, but if you look above to line 694 this pattern was already used. I was trying to match the existing code.

pierrebeitz: i'm not asking to change that here and now as the PR is pretty old/big already. 

as we're good boyscouts and clean stuff up no matter who made the mess in the first place i still wanted to point that out. :)

plugins/services/src/js/data/groups/resolvers.ts
TattdCodeMonkey: We could have a larger conversation around structuring graphql schema and arguments for the data layer. The basic trade-off I was trying to meet here was generic vs specific schema. If we went with a more specific schema for quota, then having more specific arguments would make sense. I opt-ed for a more general schema that can be grown, which then IMO lends itself to more generic arguments. As these seems to be what makes graphQL useful IMO. To me having a bunch of specific queries doesn't add much value over just having stores with functions.

pierrebeitz: my comment is not about graphQL at all but only about the second argument that can be unknown but is only handled if it's a `JSON.parse`-able string.

if that's related and i don't see that because i'm missing some part of the bigger picture here, please ignore my comment.

TattdCodeMonkey: That argument is coming from directly from GraphQL arguments which are `unknown`, we could do the type check in the resolver but I thought it better to just pass it down. 🤷‍♂

pierrebeitz: ah! thanks for pointing that out! 🤦‍♂

i'll have another look then and am perfectly fine with this till then.

PR #3980, opened 2019-07-09
COPS-4412: highlight unmatched role

nLight:
The PR fixes the issue when having an unmatched role with matching resources it would still highlight resources. This is misleading.

Closes https://jira.mesosphere.com/browse/COPS-4412

## Testing

Launch an app with a placement constraint on the hostname and changing cpu/mem check following scenarios:

1. Assuming rather free cluster request minimum CPU 0.001 and MEM 32 - checking only role is highlighted
2. Request CPU: 2.5 that can't be fulfilled checking the code highlights 2.5 and doesn't show the role. Assuming insufficient resources is more important

```
{
  "id": "/offer",
  "backoffFactor": 1.15,
  "backoffSeconds": 1,
  "cmd": "sleep infinity;",
  "constraints": [
    [
      "hostname",
      "IS",
      "dima"
    ]
  ],
  "container": {
    "type": "MESOS",
    "volumes": []
  },
  "cpus": 2.1,
  "disk": 0,
  "instances": 1,
  "maxLaunchDelaySeconds": 300,
  "mem": 128,
  "gpus": 0,
  "networks": [
    {
      "mode": "host"
    }
  ],
  "portDefinitions": [],
  "requirePorts": false,
  "upgradeStrategy": {
    "maximumOverCapacity": 1,
    "minimumHealthCapacity": 1
  },
  "unreachableStrategy": {
    "inactiveAfterSeconds": 0,
    "expungeAfterSeconds": 0
  },
  "healthChecks": [],
  "fetch": []
}
```

## Trade-offs

`areResourcesUnmatched ` was incorrectly named all the way. You can see from the code it returns true when received resource is greater or equals requested amount. Meaning resource offer actually _matches_ the request.

#### Resource is matching, role doesn't - still highlighting the resource

Screenshot 2019-07-09 at 14 16 04

#### Requested resource exceeds available amount - not showing role
Screenshot 2019-07-09 at 14 13 14

#### Requested resource suffice available amount - highlight unmatched role
Screenshot 2019-07-09 at 14 13 06


locale/en/messages.json
```diff
@@ -939,6 +940,7 @@
   "your service's requirements ({requestedValue}).": "",
   "your service's role ({requestedValue}).": "",
   "{0, plural, one {# instance running out of {instancesCount}} other {# instances running out of {instancesCount}}}": "",
+  "{0}": "",
```
GeorgiSTodorov: In case someone is wondering: This is also happening on master, I don't know where it comes from.

PR #3972, opened 2019-07-05
DCOS-55638: ServiceTreeView - No limit banner & badges

GeorgiSTodorov:
Closes https://jira.mesosphere.com/browse/DCOS-55638 .

## Testing
0. Start localhost and login.
1. Update your `Config.dev.ts`
```
uiConfiguration: {
...
  features: {
    quota: true
  }
},
  useFixtures: true,
  ...
```
2. On services page, click "2_apps" group.
3. Verify that the infobox and the badge are shown.
4. Verify that the infobox and the badge are not shown for other groups.
5. In `tests/_fixtures/marathon-pods/groups.js` change `enforceRole` on line 281 to false and verify that he badge and the infobox are no longer shown.
6. Verify that the added test makes sense.

## Trade-offs
None.

## Dependencies
None.

## Screenshots
Снимка от 2019-07-17 11-25-56

PR #3981, opened 2019-07-11
DCOS-56348: Quota Overview - Limit Column

GeorgiSTodorov:
Add a limit column for Quota.

Closes https://jira.mesosphere.com/browse/DCOS-56348

## Testing
0. Start localhost and login.
```
uiConfiguration: {
...
  features: {
    quota: true
  }
},
  useFixtures: true,
  ...
```
2. Go to services page.
3. Open quota tab.
4. Verify that the new "Quota Limit" column is there and it looks and behaves as expected.

## Trade-offs
Some of our solutions will probably need to be changed.

## Dependencies
None.

## Screenshots
Screenshot from 2019-07-11 16-51-54


plugins/services/src/js/components/ServicesQuotaOverviewTable.tsx
```diff
@@ -108,6 +109,15 @@ class ServicesQuotaOverviewTable extends React.Component<
             }
             cellRenderer={nameRenderer}
           />
+          <Column
+            key="limit"
+            header={
+              <HeaderCell>
```
GeorgiSTodorov: Non-sortable (for now).

plugins/services/src/js/types/ServiceGroup.ts
```diff
@@ -19,6 +19,8 @@ export type ServiceGroup = {
   id: string;
   name: string;
   quota?: null | ServiceGroupQuota;
+  rolesLength: number;
```
TattdCodeMonkey: I think these should both be a part of `ServiceGroupQuota` and the names could be better.

plugins/services/src/js/structs/ServiceTree.js
```diff
@@ -370,4 +370,13 @@ module.exports = class ServiceTree extends Tree {
       return serviceTreeLabels;
     }, []);
   }
+
+  getGroupRolesLength() {
```
TattdCodeMonkey: I think we should have 1 function that returns the could of both group and non-group roles so that we only have to walk the tree a single time to get those counts.

TattdCodeMonkey: We should also add unit tests for that function and make sure we test for deeply nested groups, apps, services and Pods.

plugins/services/src/js/types/ServiceGroup.ts
GeorgiSTodorov: > the names could be better.

I agree, but I didn't want them to be too long like `groupAppWithRolesLength`.

plugins/services/src/js/structs/ServiceTree.js
GeorgiSTodorov: We can do both of these things.

plugins/services/src/js/data/groups/fetchers.ts
```diff
@@ -14,8 +14,15 @@ export function fetchServiceGroups(): Observable<ServiceTree[]> {
   if (!(serviceTree instanceof ServiceTree)) {
     return of([]);
   }
-  const groupsList = serviceTree.getGroups();
-  const groups = groupsList.getItems() as ServiceTree[];
+  const groups = serviceTree.getItems().reduce(
```
pierrebeitz: nit: we could get rid of the typecast if we wrote it like that: 

```typescript
  const groups = serviceTree
    .getItems()
    .reduce((groups: ServiceTree[], item) => {
      if (item instanceof ServiceTree) {
        groups.push(item);
      }
      return groups;
    }, []);

```

plugins/services/src/js/types/ServiceGroup.ts
```diff
@@ -7,8 +7,21 @@ export type QuotaResources = {
...
+export type ServiceGroupQuotaLimit =
+  | "Enforced"
+  | "Not Enforced"
+  | "Partially Enforced"
+  | "N/A";
+
 export type ServiceGroupQuota = {
   enforced: boolean;
+  limitStatus: ServiceGroupQuotaLimit;
+  serviceRoles?: ServiceGroupQuotaRoles;
```
pierrebeitz: could we provide a default value for `ServiceGroup.quota`, so we could change the type like so?

```diff
export type ServiceGroup = {
  id: string;
  name: string;
-  quota?: null | ServiceGroupQuota;
+  quota: ServiceGroupQuota;
};
```

It looks like we could remove some uncertainty in the code using this type as a follow up.

plugins/services/src/js/columns/QuotaOverviewLimitColumn.tsx
```diff
@@ -0,0 +1,78 @@
...
+      ? group.quota.serviceRoles.count
+      : 0;
+  const groupRoles =
+    group.quota && group.quota.serviceRoles
+      ? group.quota.serviceRoles.groupRoleCount
+      : 0;
+  const limitNumber = roles - groupRoles;
+
+  let icon = null;
+  if (limit === "Enforced") {
```
pierrebeitz: if we changed to a `switch-case` here, we would be notified by typescript whenever the `ServiceGroupQuotaLimit`-type is extended (or a case is removed). The compiler would force us to handle (or remove) that case here, which would be great regarding maintainability! :)

plugins/services/src/js/types/ServiceGroup.ts
TattdCodeMonkey: I don't think so, since this type is the return value from graphql queries we want quota to be optional so that you can make queries where you dont' get "quota" and don't make the additional API calls. Granted we're only using this for quota currently, but I think for graphql query types we should stick with flexibility IMO. 

We might be able to remove the null value though. I will look into that when I work on refactoring DL stuff for DCOS-56352

GeorgiSTodorov:
This is work in progress as we have to refactor it to use the `ServiceTree` function https://github.com/dcos/dcos-ui/pull/3972/files#diff-2f23d4286259700270abc15fd4492804R374 instead of graphql.

plugins/services/src/js/columns/QuotaOverviewLimitColumn.tsx
```diff
@@ -22,45 +22,62 @@ export function limitRenderer(group: ServiceGroup) {
...
+            id="quota-limit-tooltip"
+            trigger={
+              <Icon
+                color={greyLight}
+                shape={SystemIcons.CircleMinus}
+                size={iconSizeXs}
+              />
+            }
+          >
+            <Plural
```
GeorgiSTodorov: @TattdCodeMonkey , Plural is broken in Chinese, this is why I didn't use it.
Example (notice the "s" in English):
Screenshot from 2019-07-15 09-07-40

plugins/services/src/js/data/groups/resolvers.ts
```diff
@@ -23,12 +29,43 @@ const isGroupArgs = (
   return (args as ServiceGroupQueryArgs).id !== undefined;
 };
 
+function getQuotaLimit(
+  roles: ServiceGroupQuotaRoles | undefined
+): ServiceGroupQuotaLimit {
+  if (roles === undefined) {
```
GeorgiSTodorov: @TattdCodeMonkey , shouldn't be also check for null?

plugins/services/src/js/columns/QuotaOverviewLimitColumn.tsx
pierrebeitz: @GeorgiSTodorov 	where is that screenshot from? using a simple `Plural` i don't seem to be able to reproduce the `s` inbetween.

GeorgiSTodorov: @pierrebeitz , it's from the nodes table after you filter the nodes.

pierrebeitz: On that instance we use a lingui-`Trans` and pluralize with our own implementation via `StringUtil.pluralize(name, totalLength)` which is hardcoded to english.

Lingui looks fine to me.

GeorgiSTodorov: Maybe I got confused and thought that it was broken there because of lingui.

GeorgiSTodorov: @pierrebeitz , another example of a broken translation in a place where we are using lingui Plural.
Screenshot from 2019-07-15 12-35-22

https://github.com/dcos/dcos-ui/blob/78f3b127ae4ce48a53f89f924346c55030e67b4e/plugins/services/src/js/columns/ServicesTableStatusColumn.tsx#L54

https://github.com/dcos/dcos-ui/blob/54a3114436bc127200165c956284d598fd4a7565/locale/zh/messages.json#L968

plugins/services/src/js/data/groups/resolvers.ts
TattdCodeMonkey: yes we should

PR #3982, opened 2019-07-15
DCOS-56382 Fix Maintenance Status Change in NodesTable

pierrebeitz:
### Setup

make sure you're using the maintenance feature in `Config.dev.ts`:

```
  useFixtures: false,
  useUIConfigFixtures: true,
  uiConfigurationFixture: {
    uiConfiguration: {
      features: {
        maintenance: true,
      },
```

Is a bit tricky. Either follow the instructions in https://jira.mesosphere.com/browse/DCOS-56382. Create a cluster via ccm with channel `testing/pull/6158` and use the curl-commands given in the JIRA to initiate the draining/reactivation of a Node (make sure you add a cookie-header and adapt the agent-ID).

OR 

apply this patch, if you think it makes sense:
```
curl https://gist.githubusercontent.com/pierrebeitz/94dc57b43f3702f9464856cf0271c2b6/raw/727f75cc6ef28b328bd2e7f2b0415ddb101af363/gistfile1.txt | git apply
```

It makes use of the randomized `drain_info` that now changes for each node with every stubbed request. Note that you need to `useFixtures: true` for this method.

### To confirm

The data (i.e. the maintenance status) in the table would not change before. Now it does! 🎉

## Screenshots

OS Enterprise 2019-07-15 12-45-08

### DIFF
--- a/plugins/nodes/src/js/components/NodesTable.tsx
+++ b/plugins/nodes/src/js/components/NodesTable.tsx
@@ -37,7 +37,6 @@ import PublicIPColumn from "../columns/NodesTablePublicIPColumn";
 interface NodesTableProps {
   withPublicIP: boolean;
   hosts: NodesList;
-  nodeHealthResponse: boolean;
   masterRegion: string;
 }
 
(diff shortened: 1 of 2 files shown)


pierrebeitz:
i would love to have one more review from the former performance-task-force.

IIRC that would be @natmegs, @brandonc, @DanielMSchmidt or @nLight.

PR #3986, opened 2019-07-15
DCOS-56352 refactor: groups data-layer

TattdCodeMonkey:
Removed the filter argument from the groups query in favor of filters on
the result stream for groups that have quota.

I introduced `QuotaUtil` and moved the new has QuotaLimit functions,
existing getQuotaLimit function and extracted code from the quota
resolver to the populateResourcesFromRole function. I did this to groups
quota related functions and easily unit test all of these functions.

I created `groupHasQuotaLimit` and `quotaHasLimit` as separate functions
so that the same logic could be used by the Overview component and the
resolver even though they were operating on related but different types.

- Closes DCOS-56352

## Testing

0. Start localhost and login.
```
uiConfiguration: {
...
  features: {
    quota: true
  }
},
  useFixtures: true,
  ...
```
2. Go to services page and open the quota tab.
3. Verify only group with quota limits shown

## Trade-offs

I'm not crazy about all of the util function names, but I tried to make them relevant in context. 

`populateResourcesFromRole` mutates the given `ServiceGroupQuota` object when returning it. I *could* of made a copy but it seemed like a waste. But I could easily be talked into changing that.

## Dependencies

N/A

## Screenshots

N/A


GeorgiSTodorov:
@TattdCodeMonkey , what do you mean with "Verify only group with quota limits shown"? I see 4 groups.

PR #3987, opened 2019-07-16
DCOS-56482/COPS-5100: fix: update status priorities

juliangieseke:
⚠️ needs backport to 1.13 after merged (DCOS-56483)

old setup was misleading since stopped is user initiated, having a group dislay 'Running (2 of 3)' is okay if third one is stopped.

Closes DCOS-56482
Closes COPS-5100

## Testing

have stopped and running services in a group - it should say "Running (1 of 2)" now.

## Screenshots
Before:
Screenshot 2019-07-16 at 08 48 55

After:
Screenshot 2019-07-16 at 08 48 58


GeorgiSTodorov:
Do we want a test for this? If yes, I can add one.

GeorgiSTodorov:
Also, have we asked the designers? I am hesitant to do something like this, just because one customer requested it. It might be that the old way is better for most of the customers.

juliangieseke:
re tests: good point, there should be a failing test after this change 🤔if not, would be awesome if you could do it 🙇 

re change: yeah, we discussed it with dima and designers, you can check COPS issue for reference 👍

GeorgiSTodorov:
If it's okay with the designers, I am okay with it, too. I will add a test.

GeorgiSTodorov:
@juliangieseke , I added an integration test for this specific scenario. I think that adding integration tests for other similar scenarios (for example warning having higher priority) is a good idea, but not in the scope of this bug fix.

PR #3991, opened 2019-07-17
DCOS-56638 Quota UI - Feature Flag Removal

TattdCodeMonkey:
## Testing

0. Ensure fixtures are off.
1. Create a top-level group
2. Using the Mesos API set quota for a role with the same name as your group
```
$ ## Set Quota
curl -X "POST" "<CLUSTER_URL>/mesos/api/v1?SET_QUOTA" \
     -H 'Authorization: token=<token>' \
     -H 'Accept: application/json' \
     -H 'Content-Type: application/json' \
     -d $'{
  "type": "SET_QUOTA",
  "set_quota": {
    "quota_request": {
      "force": true,
      "guarantee": [
        {
          "scalar": {
            "value": 1
          },
          "name": "cpus",
          "role": "*",
          "type": "SCALAR"
        },
        {
          "scalar": {
            "value": 512
          },
          "name": "mem",
          "role": "*",
          "type": "SCALAR"
        }
      ],
      "role": "<group_name>"
    }
  }
}'
```
3. Navigate to Quota tab Services -> Quota tab
4. Verify the group you created is displayed in the list.

## Screenshots

image
image

### DIFF
--- a/plugins/services/src/js/containers/services/ServiceTreeView.js
+++ b/plugins/services/src/js/containers/services/ServiceTreeView.js
@@ -12,7 +12,6 @@ import DSLExpression from "#SRC/js/structs/DSLExpression";
 import DSLFilterField from "#SRC/js/components/DSLFilterField";
 
 import Page from "#SRC/js/components/Page";
-import ConfigStore from "#SRC/js/stores/ConfigStore";
 
 import DeploymentStatusIndicator from "../../components/DeploymentStatusIndicator";
 import EmptyServiceTree from "./EmptyServiceTree";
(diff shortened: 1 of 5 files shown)


plugins/services/src/js/containers/services/ServiceTreeView.js
```diff
@@ -110,32 +102,29 @@ class ServiceTreeView extends React.Component {
...
+    if (createModalOpen) {
       return [];
     }
     const { serviceTree } = this.props;
-    if (
-      serviceTree.isRoot() ||
-      typeof serviceTree.getEnforceRole() === "boolean"
-    ) {
+    const id = serviceTree.getId();
+    if (id.split("/").length <= 2) {
```
TattdCodeMonkey: This is just an easy way to check if the group id is `/` or `/<name>` ie root or top-level. 

I have some additional work in progress that will remove this check in favor of a `hasQuota` check for top-level groups.

plugins/services/src/js/utils/QuotaUtil.ts
```diff
@@ -58,10 +58,15 @@ export function populateResourcesFromRole(
       quota.gpus.limit = getValue(groupQuota.limit.gpus);
     }
     if (groupQuota.consumed) {
-      quota.cpus.consumed = getValue(groupQuota.consumed.cpus);
-      quota.memory.consumed = getValue(groupQuota.consumed.mem);
-      quota.disk.consumed = getValue(groupQuota.consumed.disk);
-      quota.gpus.consumed = getValue(groupQuota.consumed.gpus);
+      quota.cpus.consumed = getValue(groupQuota.consumed.cpus) || 0;
```
TattdCodeMonkey: In testing consumed is empty until we have tasks with the role, but I still wanted to show 0 so I added this code to always default consumed values to 0 if we have a role.

plugins/services/src/js/containers/services/ServiceTreeView.js
GeorgiSTodorov: We have a `isRoot` function for `ServiceTree`.

PR #3992, opened 2019-07-18
DCOS-56347: refactor add service button

mperrotti:
## Testing
1. Go to the Services page
2. Click the "+" in the page header

A dropdown should appear with "Run a Service" and "Create Group", and there should be no overflow button on the right where "Create Group" used to be

## Trade-offs
We're waiting to implement the exact design ([InVision mockup](https://mesosphere.invisionapp.com/d/main#/console/17664445/366217481/preview)) until the design team can have some time to consider a dropdown add button style for all of DC/OS.

## Screenshots
Screen Shot 2019-07-17 at 8 24 57 PM


pierrebeitz:
amended a small change to `refactor: removes service page header overflow menu` so that cypress would still find the button to open the service modal.

PR #3994, opened 2019-07-18
DCOS-56452: DCOS UI to recognize DELAYED state of SDK services

GeorgiSTodorov:
Add delayed service status with 208 HTTP code.

Closes https://jira.mesosphere.com/browse/DCOS-56452

## Testing
1. In `Config.dev.ts` make `useFixtures` to be `true`.
2. Find the service `208-hello-world` and verify that it has a "Delayed" status.

## Trade-offs
I left the status name as it is, but maybe we want to change it to "Delayed (Failed to launch tasks)"?

## Dependencies
None.

## Screenshots
Screenshot from 2019-07-18 16-35-02


nLight:
`Delayed` is ok

PR #3997, opened 2019-07-18
[spike] DCOS-56743 SDK Plans to container DL

TattdCodeMonkey:
## Testing

SDK Plans should still work


nLight:
@TattdCodeMonkey I'm not sure how the fix fixes the issue 🤔 but looks like it does.

TattdCodeMonkey:
@nlight we were waiting until the service returned a running status which meant you could never check deployment plan.

This basically undos the bug fix Georgi did for 404s until the scheduler exists.

PR #3999, opened 2019-07-22
Simplify CLI installation instructions on UNIX

bamarni:
This removes the need to have `sudo` 2 times, by running `chmod`
before moving the file.

It also removes the creation of `/usr/local/bin` directory, it should be
present in any UNIX installation. Having this line is not really useful,
if the directory is missing it's most likely not in the $PATH either,
so the subsequent `dcos cluster setup` command would still fail anyways.

### DIFF
--- a/src/js/components/modals/CliInstallModal.js
+++ b/src/js/components/modals/CliInstallModal.js
@@ -75,10 +75,9 @@ class CliInstallModal extends React.Component {
     }
 
     const instructions = [
-      `[ -d /usr/local/bin ] || sudo mkdir -p /usr/local/bin`,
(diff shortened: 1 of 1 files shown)


GeorgiSTodorov:
The command works, I will fix the commit message so that the build would pass.

GeorgiSTodorov:
@bamarni , do we want this in older versions?

bamarni:
@GeorgiSTodorov : no need to backport, not a bugfix it just simplifies these instructions.

GeorgiSTodorov:
@bamarni , okay, cool.

PR #4000, opened 2019-07-22
DCOS-55845 fullscreen group modal

TattdCodeMonkey:
Introduced a feature flagged full-screen modal to be used when creating
a top-level group. Current implementation only has group create with
name using data-layer. This is the first step in supporting creating a
group with quota.

## Testing

In `Config.dev` set `useUIConfigFixtures: true` and 
```
  uiConfigurationFixture: {
    uiConfiguration: {
      features: {     
        quota: true
...
```

1. Go to `/services/overview`
2. Click `Create Group`
3. *Verify full screen modal is displayed*
4. Set a name for you group
5. Click Save
6. Verify Group is created.

## Screenshots

image
image
image


plugins/services/src/js/components/modals/ServiceRootGroupModal.tsx
```diff
@@ -0,0 +1,361 @@
...
+                    <Trans>You do not have permission to create a group.</Trans>
+                  )
+                },
+                isPending: false
+              });
+              return;
+            default:
+              this.setState({
+                errors: {
+                  misc: <span>Unable to create group: {e.message}</span>
```
DanielMSchmidt:
```suggestion
                  misc: <Trans values={{ msg: e.message }} id="Unable to create group: {e.message} />
```

plugins/services/src/js/components/modals/ServiceRootGroupModal.tsx
```diff
@@ -0,0 +1,361 @@
...
+              </FormGroupHeading>
+            </FieldLabel>
+            <div>{`/${data.name}`}</div>
+          </FormGroup>
+        </FormRow>
+      </form>
+    );
+  }
+
+  getMiscErrors() {
```
mperrotti: Is there a reason we couldn't use the `ErrorsAlert` component directly?

TattdCodeMonkey: nope, I just didn't even think about using `ErrorsAlert` I can swap that out.

TattdCodeMonkey: So `ErrorsAlert` is specific to errors from a service and expect the errors to be in a specific format, and parses them through ErrorUtils. I wanted similar markup but didn't want to be confined to the error data structure used for the create service form.

plugins/services/src/js/components/modals/ServiceRootGroupModal.tsx
```diff
@@ -0,0 +1,367 @@
...
+            {
+              className: "button-primary-link button-flush-horizontal",
+              clickHandler: this.handleClose,
+              label: "Cancel"
+            }
+          ]}
+          type="secondary"
+        />
+        <FullScreenModalHeaderTitle>
+          <Trans>New Group</Trans>
```
GeorgiSTodorov: We don't have "New Group" in the message catalog.

GeorgiSTodorov:
Pressing enter in the input refreshes the page instead of submitting.

plugins/services/src/js/components/modals/ServiceRootGroupModal.tsx
GeorgiSTodorov: What are misc errors supposed to be?

PR #4005, opened 2019-07-24
DCOS-56963 - fix(Quota): ensure quota values are properly truncated for display

TattdCodeMonkey:
## Screenshots

Before:
image
image

After:
image
image

### DIFF
--- a/plugins/services/src/js/components/ServicesQuotaOverviewDetail.tsx
+++ b/plugins/services/src/js/components/ServicesQuotaOverviewDetail.tsx
@@ -21,6 +21,7 @@ import ProgressBar from "#SRC/js/components/ProgressBar";
 import * as ResourcesUtil from "#SRC/js/utils/ResourcesUtil";
 
 import { ServiceGroup, QuotaResources } from "../types/ServiceGroup";
+import * as QuotaUtil from "../utils/QuotaUtil";
 
 export interface ServicesQuotaOverviewDetailProps {
   id: string;
@@ -42,7 +43,10 @@ function getQuotaPercentage(
     return 0;
   }
 
-  return (resourceQuota.consumed / resourceQuota.limit) * 100;
+  return QuotaUtil.formatQuotaPercentageForDisplay(
+    resourceQuota.consumed,
+    resourceQuota.limit
+  );
 }
 
 function getQuotaConsumedOfLimit(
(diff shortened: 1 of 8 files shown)


plugins/services/src/js/utils/QuotaUtil.ts
```diff
@@ -119,3 +119,15 @@ export function serviceTreeHasQuota(
     Object.keys(role.quota.limit).length > 0
   );
 }
+
+export function formatQuotaValueForDisplay(value: number): number {
+  //@ts-ignore
+  return +(Math.round(value + "e+2") + "e-2");
```
mperrotti: This is so cool!

TattdCodeMonkey:
force pushed to get `closed DCOS-56963` in a commit message.

plugins/services/src/js/utils/QuotaUtil.ts
TattdCodeMonkey: I can't take credit I lifted it from the [Dashboard component](https://github.com/dcos/dcos-ui/blob/master/src/js/components/charts/ResourceTimeSeriesChart.js#L70-L72)

pierrebeitz: interestingly it's not mentioned in [the spec](https://tc39.es/ecma262/#sec-math.round) but works across browsers. 🤔

would you object to adding a `parseFloat()` to get rid of the `ts-ignore`?

```
+(Math.round(parseFloat(value + "e+2")) + "e-2");
```

nLight: We might want extracting the code into a Util or ui-kit component. So that our charts are always accurately rounded.

TattdCodeMonkey: `ui-kit` components could make sense since we do similar things in the other products.

TattdCodeMonkey: @pierrebeitz instead of `parseFloat` I just replaced the string concatenation with math.
`+(Math.round(value + "e+2") + "e-2")` -> `Math.round(value * 100) / 100` which is essentially the same thing right?

nLight: I _think_ those are't the same... at least if we trust internet.

pierrebeitz: There are subtle differences in that we might have rounding errors with the new solution. I *think* `+parseFloat(value).toFixed(2)` would be equivalent, more intention revealing and typesafe. :)

The `+` in front is for getting rid of trailing zeros.

TattdCodeMonkey: How many ways can you round a number in JS ;)

TattdCodeMonkey: Updated to `+value.toFixed(2)`

plugins/services/src/js/components/ServicesQuotaOverviewDetail.tsx
```diff
@@ -57,19 +61,15 @@ function getQuotaConsumedOfLimit(
...
+  );
+  const limit = QuotaUtil.formatQuotaValueForDisplay(resourceQuota.limit);
 
   if (resource === "cpus" || resource === "gpus") {
-    return (
-      <Trans>
-        {resourceQuota.consumed} of {resourceQuota.limit} Cores
-      </Trans>
-    );
+    return <Trans id="{0} of {1} Cores" values={{ 0: consumed, 1: limit }} />;
```
GeorgiSTodorov: Interesting.

GeorgiSTodorov:
Rebased.

PR #4006, opened 2019-07-26
DCOS-55894 Support quota limits on group create

TattdCodeMonkey:
Support setting quota limits when creating a top-level group.
I also refactored the `ServiceRootGroupModal` to a directory to allow
modularizing it. The file was getting very large. And since we will want
to expand this modal even more in the future this seemed like a worth
while refactor now.

~*Note* this PR is based on #4000 and will need to be rebased once that PR merges.~

## Testing

In `Config.dev` set `useUIConfigFixtures: true` and 
```
  uiConfigurationFixture: {
    uiConfiguration: {
      features: {     
        quota: true
...
```

1. Go to `/services/overview`
2. Click `Create Group`
3. Verify that it works with and without quota.

## Dependencies

~#4000~

## Screenshots

image
image
image


GeorgiSTodorov:
Unlike the service creation modal, here updating the field doesn't clear the error. Is this intended?
Peek 2019-07-29 12-42

plugins/services/src/js/components/modals/ServiceRootGroupModal/index.tsx
```diff
@@ -0,0 +1,431 @@
...
+          //TODO: If error creating quota switch to Edit Mode for group, must
+          // Be done after edit mode supported.
+          switch (e.message) {
+            case "Conflict - Group":
+              this.setState({
+                errors: {
+                  id: <Trans>Name already exists. Try a different name.</Trans>,
+                  form: [
+                    <Trans key="groupIdConflict">
+                      A group with the same name already exists. Try a different
```
GeorgiSTodorov: If I try a group with a duplicate name, I don't get this message, but rather:
Screenshot from 2019-07-29 12-51-11

plugins/services/src/js/components/modals/ServiceRootGroupModal/index.tsx
```diff
@@ -0,0 +1,431 @@
...
+              Quota
+            </Trans>
+          </FormGroupHeading>
+        </h2>
+        <Trans render="p">
+          Define the maximum amount of resources that can be used by services in
+          this group.
+        </Trans>
+        <FormRow>
+          <FormGroup
```
GeorgiSTodorov: @TattdCodeMonkey , what do you think about creating these form groups with a function? They are pretty similar.

plugins/services/src/js/components/modals/ServiceRootGroupModal/utils.tsx
```diff
@@ -0,0 +1,100 @@
...
+    if (!ValidatorUtil.isNumber(value)) {
+      setQuotaError(field, <Trans>Must be a number</Trans>);
+      addFormError(<Trans>{QuotaFieldLabels[field]} must be a number</Trans>);
+    } else {
+      const numberValue = parseFloat(value);
+      if (numberValue < 0) {
+        setQuotaError(field, <Trans>Must be bigger than or equal to 0</Trans>);
+        addFormError(
+          <Trans>
+            {QuotaFieldLabels[field]} must be bigger than or equal to 0
```
GeorgiSTodorov: Is this for all fields? When creating a service, everything, except CPU must be simply bigger than 0. CPU "Must be bigger than or equal to 0.001".

plugins/services/src/js/data/MarathonClient.ts
```diff
@@ -0,0 +1,53 @@
...
+
+export function createGroup(
+  id: string,
+  enforceRole: boolean
+): Observable<CreateGroupResponse> {
+  return request(buildMarathonURI("/groups"), {
+    method: "POST",
+    body: JSON.stringify({ id, enforceRole })
+  }).pipe(
+    map(reqResp => {
```
GeorgiSTodorov: This appears to be called for success, but not for some errors (duplicate group name).

GeorgiSTodorov:
The two scrolling issues should be fixed now. The last change looks big, but it's just adding three wrappers to the modal content.

GeorgiSTodorov:
Rebased.

plugins/services/src/js/components/modals/ServiceRootGroupModal/index.tsx
TattdCodeMonkey: I updated from `Conflict` to `Conflict - Groups` I thought I got the switch statement in the component too. If not please fix.

plugins/services/src/js/components/modals/ServiceRootGroupModal/index.tsx
TattdCodeMonkey: I'd be fine with that. We will likely eventually have to abstract this whole section into its own component.

plugins/services/src/js/components/modals/ServiceRootGroupModal/utils.tsx
TattdCodeMonkey: I used this for all but we can add a custom one for CPU, we would also want to update the validation in that case b/c we're just checking `>=` 0 currently

plugins/services/src/js/data/MarathonClient.ts
TattdCodeMonkey: yes that is expected, an error will exit the stream and if there is an error creating the group we don't want to create quota.

plugins/services/src/js/components/modals/ServiceRootGroupModal/index.tsx
```diff
@@ -0,0 +1,479 @@
...
+          value = value === "true";
+        }
+        break;
+      default:
+        value = target.value;
+        break;
+    }
+
+    const { data } = this.state;
+    this.setState({ data: set(data, fieldName, value) });
```
brandonc: For fun I took a stab at writing a version of set that does not require a dependency (lodash.set is 5KB minified)

```typescript
function set(object: any, path: string, value: any): any {
  if (typeof object !== "object") {
    return object;
  }
  const delimiterIndex = path.indexOf(".");
  if (delimiterIndex >= 0) {
    const thisPart = path.substring(0, delimiterIndex);
    object[thisPart] = set(
      object[thisPart] || {},
      path.substring(delimiterIndex + 1),
      value
    );
  } else {
    object[path] = value;
  }
  return object;
}

let obj = set({}, "quota.cpus", 1);
set(obj, "quota.superman", 1);
set(obj, "quota.cpus.whatHappens", 0);

console.log(obj);
// { quota: { cpus: 1, superman: 1 } }
```

plugins/services/src/js/components/modals/ServiceRootGroupModal/index.tsx
```diff
@@ -0,0 +1,479 @@
...
+  }
+
+  handleClose() {
+    // Start the animation of the modal by setting isOpen to false
+    this.setState(
+      { isOpen: false, isPending: false, data: emptyGroupFormData() },
+      () => {
+        // Once state is set, start a timer for the length of the animation and
+        // navigate away once the animation is over.
+        setTimeout(this.context.router.goBack, 300);
```
brandonc: This seems unorthodox and brittle with respect to animation timing. I can't think of something better that does not involve always rendering this modal.

plugins/services/src/js/data/MesosClient.ts
```diff
@@ -0,0 +1,75 @@
...
+    quota_configs: [
+      {
+        role: roleFromGroupId(groupId),
+        limits: limitsFromQuotaFormData(quotaData)
+      }
+    ]
+  };
+}
+
+export function updateQuota(
```
brandonc: I also added some mesos request wrappers when building maintenance features, but I added them as some simple actions-- do you think this should eventually define the entire API client? The name hints at that, but it seems overly ambitious. Maybe this should just be the QuotaActions instead?

plugins/services/src/js/components/modals/ServiceRootGroupModal/index.tsx
```diff
@@ -0,0 +1,479 @@
...
+                >
+                  <FieldLabel>
+                    <FormGroupHeading>
+                      <Trans render={<FormGroupHeadingContent />}>CPUs</Trans>
+                    </FormGroupHeading>
+                  </FieldLabel>
+                  <FieldAutofocus>
+                    <FieldInput
+                      name="quota.cpus"
+                      type="text"
```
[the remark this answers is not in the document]
TattdCodeMonkey: I'd be fine with placeholder text, @izharbuendia  ?

We don't use number because these fields are optional and empty is valid here.

plugins/services/src/js/components/modals/ServiceRootGroupModal/index.tsx
TattdCodeMonkey: I initially had a function written for this and opt-ed for `lodash.set` instead.

plugins/services/src/js/components/modals/ServiceRootGroupModal/index.tsx
TattdCodeMonkey: This came directly from the `CreateServiceModal`

plugins/services/src/js/data/MesosClient.ts
TattdCodeMonkey: My idea was this could be abstracted out to a package and be a more robust client.

For plans I wrote a very specific `SDKPlansClient` and that works, but I assumed as we added more Mesos calls to data-layer having a central client would be useful, same with `MarathonClient`. But that certainly could be too ambitious.

TattdCodeMonkey: Maybe I should of put these in `packages/` to begin with.

plugins/services/src/js/components/modals/ServiceRootGroupModal/index.tsx
TattdCodeMonkey: I switched to lodash b/c it seemed like a common enough use case that it had to of been solved before, I found the lodash function and opted for it since it's less code for use to maintain/test etc. But that is a total preference thing.

GeorgiSTodorov:
@natmegs 

> Is it in scope to have a more descriptive error message for when you're trying to enter a duplicate group name? (and maybe an error message on the form field?)

This should be the case: https://github.com/dcos/dcos-ui/pull/4006/files#diff-310a0c75dfad1e9b5b80eb45a187560dR153 . I will take a look why it doesn't work.

The message

> In other modals like services & jobs we tend to show errors on submit, and then live validate the fields after that so that errors clear as fields are corrected. Is it intentional do not live validate these fields after submit here?

This is known and we intend to fix in the future, but not in this PR.

GeorgiSTodorov:
I will take a look at the failing tests.

src/styles/components/modal/full-screen/styles.less
```diff
@@ -40,6 +42,11 @@
     .modal-body-offset {
       margin: -@modal-full-screen-body-padding-vertical -@modal-full-screen-body-padding-horizontal;
     }
+
+    .gm-scrollbar-container {
```
GeorgiSTodorov: @mperrotti removed this in https://github.com/dcos/dcos-ui/pull/4010 , but without it, our modal has 0 height on integration tests.

GeorgiSTodorov:
Let's see now.

PR #4008, opened 2019-07-29
[DCOS-55018] [DCOS-55019] Deactivate/Reactivate nodes via node overview table

brandonc:
Closes DCOS-55018
Closes DCOS-55019

## Testing

1. add this key to the "uiConfiguration" key in Config.dev.ts
```
features: {
  maintenance: true
},
```
2. enable useUIConfigFixtures
3. Visit the nodes page and find the new actions dropdown on the far right (for active nodes)
4. Try deactivating & reactivating nodes

## Screenshots

Screen Shot 2019-07-29 at 4 31 16 PM


plugins/nodes/src/js/components/modals/DeactivateNodeConfirm.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+  const [inProgress, setInProgress] = useState<boolean>(false);
+  const [networkError, setNetworkError] = useState<React.ReactElement | null>(
+    null
+  );
+
+  let [prevOpenState, setPrevOpenState] = useState<boolean | null>(null);
+
+  // Reset form when closed
+  if (open !== prevOpenState) {
```
natmegs: I wonder if there could be a handler to extend `props.onClose` to perform these state changes, to avoid having to keep track of `prevOpenState`?

plugins/nodes/src/js/pages/nodes/nodes-table/NodesTableContainer.js
```diff
@@ -142,7 +190,12 @@ class NodesTableContainer extends mixin(StoreMixin, QueryParamsMixin) {
         <DrainNodeModal
           open={selectedNodeToDrain != null}
           node={selectedNodeToDrain}
-          onClose={this.handleDrainClose}
+          onClose={this.handleCloseModal}
+        />
+        <DeactivateNodeConfirm
+          open={selectedNodeToDeactivate != null}
```
GeorgiSTodorov: Why "!="?

GeorgiSTodorov:
- Is it possible to add a state like "deactivating" or "activating", because right now it's not clear that something is happening.
Peek 2019-07-30 09-18

- This is somewhat out of scope, but "Deactivated" isn't getting translated (I inserted a fake translation). Same with "Drained".
Screenshot from 2019-07-30 09-26-31
Screenshot from 2019-07-30 09-38-47

- I still think the tooltip should be on the icon and not on the entire cell.
Screenshot from 2019-07-29 10-46-38

- I still think that we can just do `header=""` in the nodes table.
https://github.com/dcos/dcos-ui/blob/9cf9b2ac74e1956dd97cd593fbab45ecef765e80/plugins/nodes/src/js/components/NodesTable.tsx#L374

- It's not longer possible to enter negative values for "Max Grace Period", but we can still enter letters and symbols.
Peek 2019-07-30 09-35

plugins/nodes/src/js/components/modals/DeactivateNodeConfirm.tsx
brandonc: Good idea

plugins/nodes/src/js/pages/nodes/nodes-table/NodesTableContainer.js
brandonc: I was lazy and didnt look up the behavior of null strict comparison

brandonc:
@GeorgiSTodorov I addressed your feedback with two exceptions:

1. I didn't reposition the tooltip on "Connection to node lost" because it appears to want to be absolutely positioned in the center of the trigger. I could also just make the trigger itself be the icon, in which case it would be centered on it. It's also notable that this column wasn't modified for this feature.

2. I added some code to re-fetch the mesos summary after each action, but I'm not sure it increased the responsiveness of the column. I'm very hesitant to start merging in optimistic client state into the nodes collection. Let me know if you think the situation has improved.

Thanks for the excellent review!

plugins/nodes/src/js/pages/nodes/nodes-table/NodesTableContainer.js
```diff
@@ -141,7 +142,9 @@ class NodesTableContainer extends mixin(StoreMixin, QueryParamsMixin) {
       this.setState({ selectedNodeToDeactivate: node });
     } else if (action === "reactivate") {
       NodeMaintenanceActions.reactivateNode(node, {
-        onSuccess: () => {},
+        onSuccess: () => {
+          MesosSummaryActions.fetchSummary();
```
natmegs: What happens with this newly fetched summary?

GeorgiSTodorov:
@brandonc , thank you for addressing my feedback so well so quickly. The only thing that worries me now is that we allow letters and symbols for "Max Grace Period". But I checked and if we enter a non-numeric value, we just don't send it in the network request and as far as I can tell, the value is non-required anyway. Feel free to do a follow-up fixing this. Or maybe we can fix it in another task.

GeorgiSTodorov:
I looked into this issue and decided to create a follow-up to fix it: https://github.com/dcos/dcos-ui/pull/4011 .

plugins/nodes/src/js/pages/nodes/nodes-table/NodesTableContainer.js
brandonc: My understanding was that subscribers (nodes table) would be updated.

PR #4009, opened 2019-07-30
DCOS-55889 show tooltip if task is not existent

pierrebeitz:
Mesos garbage collects tasks. We don't want to tasks that are not present anymore in Mesos. We want to show a tooltip instead.

## Review

commit-by-commit

## Testing

You most likely want a DCOS with a Job that has both kinds of tasks. It looks like a pretty good one is this job on SOAK: http://localhost:4200/#/jobs/detail/spark-hdfs-kerberos-terasort.

Make sure that we don't link to old tasks and that we link to tasks when they are present. Also make sure there are no weird states when reloading the page.

## Screenshots

Monosnap 2019-07-30 17-19-34
OS Enterprise 2019-07-30 17-20-25

Closes DCOS-55889


plugins/services/src/js/components/SDKPlans.tsx
```diff
@@ -127,7 +127,11 @@ const SDKPlans = componentFromStream(props$ => {
         concat(throwError(errors))
       )
     ),
-    catchError(() => of(<ErrorScreen />)),
+    catchError(e => {
+      console.warn(e);
```
GeorgiSTodorov: Build is failing because of this line.

plugins/jobs/src/js/pages/JobRunHistoryTable.js
```diff
@@ -246,17 +256,39 @@ class JobRunHistoryTable extends React.Component {
...
+          <CollapsingString endLength={15} string={taskID} />
+        </Link>
+      );
+      const hint = (
+        <Tooltip
+          preferredDirections={["top-left", "bottom-left"]}
+          trigger={<span>{taskID}</span>}
+        >
+          <Trans>
+            The data related to this tasks has already been cleaned up.
```
GeorgiSTodorov: These* tasks.

plugins/jobs/src/js/pages/JobTaskDetailPage.js
```diff
@@ -30,6 +59,16 @@ class JobTaskDetailPage extends React.Component {
     ];
 
     const task = MesosStateStore.getTaskFromTaskID(taskID);
+
+    if (!task) {
+      return (
+        <Trans>
+          Either the data related to that task has already been cleaned up or
```
GeorgiSTodorov: I think "this" instead of "that" sounds better.

plugins/jobs/src/js/pages/JobTaskDetailPage.js
```diff
@@ -30,6 +59,16 @@ class JobTaskDetailPage extends React.Component {
...
+          the given task-id does not exist.
```
GeorgiSTodorov: I think that "task-id" should be written without a dash.

plugins/jobs/src/js/pages/JobRunHistoryTable.js
```diff
@@ -246,17 +248,34 @@ class JobRunHistoryTable extends React.Component {
...
+          className="table-cell-link-secondary"
+          to={`/jobs/detail/${this.props.job.id}/tasks/${taskID}`}
+          title={taskID}
+        >
+          <CollapsingString endLength={15} string={taskID} />
+        </Link>
+      );
+      const hint = (
+        <Tooltip
+          preferredDirections={["top-left", "bottom-left"]}
```
GeorgiSTodorov: What do you think about only having "bottom-left" here so that we don't hide the job id?
### Currently
Screenshot from 2019-07-31 09-53-22

### My suggestion
Screenshot from 2019-07-31 09-54-04

GeorgiSTodorov:
I left some comments. Other than that, it looks good.

plugins/jobs/src/js/pages/JobRunHistoryTable.js
```diff
@@ -246,17 +256,39 @@ class JobRunHistoryTable extends React.Component {
...
-      const id = this.props.job.id;
+
+      if (!this.state.mesosStateStoreLoaded) {
+        return <Trans>Loading...</Trans>;
+      }
+
+      // It may be that the data only *seems* to be present because we don't
+      // remove tasks from the MesosStateStore when they're garbage collected in
+      // the background.
+      const dataSeemsStillPresent = !!MesosStateStore.getTaskFromTaskID(taskID);
```
GeorgiSTodorov: What do you think about something like:
```
const content = !!MesosStateStore.getTaskFromTaskID(taskID) ?
  <Link
  ...
  :
  <Tooltip
  ...
```
since we never need both at the same time.

plugins/services/src/js/components/SDKPlans.tsx
pierrebeitz: removed the according commit for now to come up with proper logging-facilities.

plugins/jobs/src/js/pages/JobRunHistoryTable.js
pierrebeitz: fixed!

plugins/jobs/src/js/pages/JobTaskDetailPage.js
pierrebeitz: totally agree!

plugins/jobs/src/js/pages/JobTaskDetailPage.js
pierrebeitz: i made it `Task ID` now as it looks like a proper noun to me

plugins/jobs/src/js/pages/JobRunHistoryTable.js
pierrebeitz: really good catch!

plugins/jobs/src/js/pages/JobRunHistoryTable.js
pierrebeitz: i thought the same at some point in time but must have forgotten about it again...
it's changed!

pierrebeitz:
@GeorgiSTodorov 	thanks for the awesome review! i think i addressed all of the points. 

regarding `Edit: I ran the extract command and there were changes.`: i see them as well but did not commit them as i thought it might be strange to see those changes popping up in this PR. maybe we should fix this directly on master.

GeorgiSTodorov:
@pierrebeitz , thank you for addressing my feedback so well and so fast. I will take a look again when I am done with another task.

> regarding Edit: I ran the extract command and there were changes.: i see them as well but did not commit them as i thought it might be strange to see those changes popping up in this PR. maybe we should fix this directly on master.

We had some issues with the extraction of these strings, and maybe you started working on your PR before these issues were fixed.

GeorgiSTodorov:
@pierrebeitz , do you want a second review or should we merge when it is green?

pierrebeitz:
i think we can merge once green.

nLight:
@izharbuendia please PDUX validate

PR #4010, opened 2019-07-30
DCOS-42072: make JSON editors resizable

mperrotti:
## Testing
1. Go to either the create service form, create job form, or framework configuration form
2. Toggle the JSON editor open
3. Drag the handle on the left of the editor to resize the editor

## Trade-offs
The component I made (`SplitPanel`) to handle this only supports a side panel on the right. I can easily support it being on the left or right, but we don't have any cases where the side panel appears on the left.

## Screenshots
JSONResizeDemo

### DIFF
--- a/plugins/jobs/src/js/components/JobsForm.tsx
+++ b/plugins/jobs/src/js/components/JobsForm.tsx
@@ -1,13 +1,16 @@
 import { Trans } from "@lingui/macro";
 import { i18nMark, withI18n } from "@lingui/react";
 import { Hooks } from "#SRC/js/plugin-bridge/PluginSDK";
-import classNames from "classnames";
 import * as React from "react";
 import { MountService } from "foundation-ui";
 
 import ErrorsAlert from "#SRC/js/components/ErrorsAlert";
 import FluidGeminiScrollbar from "#SRC/js/components/FluidGeminiScrollbar";
 import JSONEditorLoading from "#SRC/js/components/JSONEditorLoading";
+import SplitPanel, {
+  PrimaryPanel,
+  SidePanel
+} from "#SRC/js/components/SplitPanel";
 import TabButton from "#SRC/js/components/TabButton";
 import TabButtonList from "#SRC/js/components/TabButtonList";
 import Tabs from "#SRC/js/components/Tabs";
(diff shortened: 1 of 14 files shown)


GeorgiSTodorov:
Some integration tests are failing. I will take a look and try to fix them.

GeorgiSTodorov:
It's possible to have the JSON editor larger than the screen if I make it very large and then resize the window. This is not that dramatic, since it's fixed by going back, but still.
Peek 2019-07-31 10-14

GeorgiSTodorov:
The tests are failing since there is a big white block at the bottom of the create service modal.
Screenshot from 2019-07-31 10-31-23
Peek 2019-07-31 10-31

mperrotti:
@GeorgiSTodorov 

Integration test failures:
- I was able to fix these, but I could only reproduce the issue in the Cypress UI's browser

Resize part appears in place immediately:
- Fixed

Possible to have the JSON editor larger than the screen if I make it very large and then resize the window
- Now the JSON editor will not be able to get smaller than the viewport width when the user resizes

Scrollbar issues
- I fixed the issue where scrollbars don't appear when the editor is resized really small, but scrollbars appearing too early seems to be an issue with `react-ace`, and it happens in master as well

plugins/jobs/src/js/components/JobsForm.tsx
```diff
@@ -73,6 +80,10 @@ class JobsForm extends React.Component<JobsFormProps> {
   constructor(props: Readonly<JobsFormProps>) {
     super(props);
 
+    this.state = {
+      editorWidth: undefined
```
brandonc: Oh, interesting... Is this an equivalent TS interface?

```
interface JobsFormState {
  editorWidth?: number;
}

interface JobsFormState {
  editorWidth: number | undefined;
}
```

I would have assumed this is a typescript error because the editorWidth key exists but undefined is not a number.

mperrotti: Yea, I thought `key?: number` was the same as `key: number | undefined`. If you think `editorWidth: number | undefined;` is clearer, I'll gladly change it :)

mperrotti:
@brandonc - that's a good idea. I called it "PrimaryPanel" because that's the panel that contains the main content, and if we ever change this to make the left or right panel resizable, you could put the children in either order:
```
<SplitPanel>
  <PrimaryPanel />
  <SidePanel />
</SplitPanel>

<SplitPanel>
  <SidePanel />
  <PrimaryPanel />
</SplitPanel>
```

mperrotti:
@natmegs - yes, that's intentional. Can be easily changed though :)

GeorgiSTodorov:
@mperrotti , thank you for addressing my feedback.

GeorgiSTodorov:
@mperrotti , I still have two of the problems: scrollbars not appearing (only for services, for jobs and frameworks it's fine) and the JSON editor taking the whole screen when resizing. I consider these points non-blocking, we can fix them if a follow-up PR if you want.

PR #4012, opened 2019-07-31
DCOS-55895: Support editing Quota limits for an existing top-level Group

GeorgiSTodorov:
Closes https://jira.mesosphere.com/browse/DCOS-55895

## Testing
0. In `Config.dev:
```
  uiConfigurationFixture: {
    uiConfiguration: {
      features: {     
        quota: true
...
```
1. Go to services page.
2. Create a group if there are none.
3. Verify that an edit action appears in the action dropdown for top-level groups.
4. Verify that an edit action appears in the top right dropdown after clicking on a top-level-group.
5. Verify that the edit modal works (you can go to the quota tab to confirm if the limits actually changed).
6. Verify that the edit action appears only for top-level groups.
7. Verify that the added tests make sense.

## Trade-offs
None.

## Dependencies
None.

## Screenshots
Screenshot from 2019-08-05 14-44-36


plugins/services/src/js/components/modals/ServiceRootGroupModal/index.tsx
```diff
@@ -90,14 +122,17 @@ class ServiceRootGroupModal extends React.Component<
...
+    isEdit: false,
+    group: null,
+    id: ""
   };
 
   constructor() {
     // @ts-ignore
     super(...arguments);
 
     this.state = this.getInitialState();
```
TattdCodeMonkey: we can pass the props to `getInitialState` to initialize in a loading state.

plugins/services/src/js/components/modals/ServiceRootGroupModal/index.tsx
```diff
@@ -90,14 +122,17 @@ class ServiceRootGroupModal extends React.Component<
...
+    this.getGroupFormData();
```
TattdCodeMonkey: Should this be in `componentWillMount` or `componentDidMount`?

plugins/services/src/js/components/modals/ServiceRootGroupModal/index.tsx
```diff
@@ -196,8 +243,36 @@ class ServiceRootGroupModal extends React.Component<
...
+          next: groupData => {
+            const data = groupFormDataFromGraphql(groupData.data.group);
+            this.setState({
+              data,
+              originalData: JSON.parse(JSON.stringify(data))
+            });
+          }
+        });
+    }
+    this.setState({
```
TattdCodeMonkey: should this be in an else? in edit mode we wouldn't want data to be set until the query has finished?

plugins/services/src/js/components/modals/ServiceRootGroupModal/index.tsx
```diff
@@ -196,8 +243,36 @@ class ServiceRootGroupModal extends React.Component<
...
+      data: emptyGroupFormData()
+    });
+  }
+
   getModalContent() {
-    const { data, errors } = this.state;
+    const { id } = this.props;
+    const { errors, data } = this.state;
+    // If id exists, then we must be editing.
+    const isEdit = !!id;
```
TattdCodeMonkey: we should move `isEdit` to state and just load it from there instead of continually checking. and remove it from props if we aren't using that value.

plugins/services/src/js/components/modals/ServiceRootGroupModal/utils.tsx
```diff
@@ -28,6 +30,26 @@ export function emptyGroupFormData(): GroupFormData {
...
+  const cpus = findNestedPropertyInObject(quota, "cpus.limit");
+  const mem = findNestedPropertyInObject(quota, "memory.limit");
+  const disk = findNestedPropertyInObject(quota, "disk.limit");
+  const gpus = findNestedPropertyInObject(quota, "gpus.limit");
+  return {
+    id: data.id || "",
+    enforceRole: enforceRole !== undefined ? enforceRole : true,
+    quota: {
+      force: false,
+      cpus: cpus !== undefined ? cpus : "",
```
TattdCodeMonkey: should we `.toString()` these values since they come out of the GQL as floats but we treat them as strings in the form?

plugins/services/src/js/data/MarathonClient.ts
```diff
@@ -51,3 +51,37 @@ export function createGroup(
...
   );
 }
+
+export function editGroup(
+  id: string,
+  enforceRole: boolean
+): Observable<CreateGroupResponse> {
+  return request(buildMarathonURI("/groups/" + id), {
+    method: "PUT",
+    body: JSON.stringify({ id, enforceRole })
```
TattdCodeMonkey: we don't want to send the `id` here, only `enforceRole` is editable.

plugins/services/src/js/data/MarathonClient.ts
```diff
@@ -51,3 +51,37 @@ export function createGroup(
...
+  }).pipe(
+    map(reqResp => {
+      const { code } = reqResp;
+      if (code > 300) {
+        const response = (reqResp.response || {
+          message: "Unknown"
+        }) as CreateGroupErrorResponse;
+        let message: string;
+        switch (reqResp.code) {
+          case 409:
```
TattdCodeMonkey: I think we can remove 409, since it shouldn't be possible to get a 409 for a `PUT`

plugins/services/src/js/data/groups/resolvers.ts
```diff
@@ -144,6 +144,27 @@ export function resolvers({ pollingInterval }: ResolverArgs): IResolvers {
             return updateQuota(args.data.id, args.data.quota);
           })
         );
+      },
+      editGroup(
+        _parent = {},
+        args: Record<string, unknown> = {}
+      ): Observable<string> {
+        if (!isGroupCreateArgs(args)) {
```
TattdCodeMonkey: we will likely want different args for `GroupEdit` from `GroupCreate` we could make more fields optional and then make less API calls based on the fields present.

plugins/services/src/js/columns/ServicesTableActionsColumn.tsx
```diff
@@ -129,7 +130,7 @@ export function actionsRendererFactory(
       });
     }
 
-    if (!isGroup) {
+    if (!isGroup || isRoot) {
       actions.push({
         id: EDIT,
         html: <Trans render="span" id={ServiceActionLabels.edit} />
```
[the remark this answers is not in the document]
GeorgiSTodorov: What is the difference?

plugins/services/src/js/components/modals/ServiceRootGroupModal/index.tsx
GeorgiSTodorov: I am not sure, but I think we don't even need it.

plugins/services/src/js/components/modals/ServiceRootGroupModal/index.tsx
```diff
@@ -113,7 +150,9 @@ class ServiceRootGroupModal extends React.Component<
       isPending: false,
       expandAdvancedSettings: false,
       data: emptyGroupFormData(),
-      errors: {}
+      originalData: null,
+      errors: {},
+      isEdit: false
```
TattdCodeMonkey: we can set `isEdit` from props here, just remove the `_` from `_props` argument and `!!props.id`

plugins/services/src/js/components/modals/ServiceRootGroupModal/index.tsx
```diff
@@ -113,7 +150,9 @@ class ServiceRootGroupModal extends React.Component<
       isPending: false,
       expandAdvancedSettings: false,
       data: emptyGroupFormData(),
```
TattdCodeMonkey: if `isEdit` we want this to be null until query in `getGroupFormData` completes, correct?

plugins/services/src/js/columns/ServicesTableActionsColumn.tsx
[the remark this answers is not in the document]
TattdCodeMonkey: A different action should send to a different route, and not use `EditServiceModal`.

plugins/services/src/js/components/modals/ServiceRootGroupModal/index.tsx
```diff
@@ -352,16 +422,23 @@ class ServiceRootGroupModal extends React.Component<
...
         group
       </Trans>
     );
 
+    if (!data) {
+      return;
+    }
+
+    const isDisabled =
+      !!this.props.id && originalData && originalData.enforceRole;
```
TattdCodeMonkey: we can change `!!this.props.id` to `this.state.isEdit`

GeorgiSTodorov:
@TattdCodeMonkey 

> A different action should send to a different route, and not use EditServiceModal.

Why? A group is still a type of service for us and I think the url "/services/detail/%2Fgroup-name/edit/" isn't wrong or misleading.

plugins/services/src/js/components/modals/ServiceRootGroupModal/index.tsx
```diff
@@ -196,14 +244,44 @@ class ServiceRootGroupModal extends React.Component<
...
+        .subscribe({
+          next: groupData => {
+            const data = groupFormDataFromGraphql(groupData.data.group);
+            this.setState({
+              data,
+              originalData: JSON.parse(JSON.stringify(data))
+            });
+          },
+          error: () => {
+            this.setState({ error: true });
```
TattdCodeMonkey: can this be `loadingError` or something a bit more descriptive.

GeorgiSTodorov: We can use this for other types of errors as well in the future.

PR #4013, opened 2019-07-31
simplify lingui-extraction-script

pierrebeitz:
by making the following change in plugins-ee, we should be able to evade the
need for .babelrc-handling.

```diff
- { "extends": "../dcos-ui/.babelrc" }
+ { "extends": "../.babelrc" }
```

This change makes use of the fact that we now work with a symlink at plugins-ee.

To enable the aforementioned change, we need to change a couple details in our build-phase. Please review commit-by-commit to best be able to think about implications i might have missed.

## Testing

Make sure you checked out the [according ee-branch](https://github.com/mesosphere/dcos-ui-plugins-private/pull/1048).

Test that the extraction still works properly. OSS and EE each have their own localization-files. 

# ⚠️ 	The EE-PR should merge at the same time or before this!

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -565,7 +565,7 @@ Add this part to your `jsconfig.json` in the root of this project
     "paths": {
       "#SRC/*": ["dcos-ui/src/*"],
       "#LOCALE/*": ["dcos-ui/locale/*"],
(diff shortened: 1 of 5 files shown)


GeorgiSTodorov:
It works, but I am hesitant to approve and merge this, because, as @brandonc said, this is likely to break something on CI. @pierrebeitz , if you are available to test and fix CI afterwards, then I am fine with merging this.

PR #4004, opened 2019-07-24
[DCOS-55016]: node action dropdown, drain active nodes

brandonc:
Closes DCOS-55016

## Testing

1. add this key to the "uiConfiguration" key in Config.dev.ts
```
features: {
  maintenance: true
},
```
2. enable useUIConfigFixtures
3. Visit the nodes page and find the new actions dropdown on the far right (for active nodes)
4. Ensure actions don't appear when feature is disabled
5. Try draining a node (this deactivates the node as well)
6. In order to reactivate the node, issue the following curl command, but replace the Cookie header VALUE and NODE_ID tokens with your own values
```
curl -v -XPOST 'http://localhost:4200/mesos/api/v1' -H 'Content-Type: application/json' -H 'Accept: application/json' --data-binary '{"type":"REACTIVATE_AGENT", "reactivate_agent": {"agent_id": {"value": "NODE_ID"}}}' -H 'Cookie: dcos-acs-auth-cookie=VALUE; dcos-acs-info-cookie=VALUE' --compressed
```

## Notes

I'm told the drained node will eventually be garbage collected by mesos but I'm not sure how long it takes.

## Screenshots

Screen Shot 2019-07-24 at 2 35 36 PM

Screen Shot 2019-07-24 at 2 37 08 PM

PR #4014, opened 2019-07-31
[DCOS-55020] Drain/Deactivate actions on node details

brandonc:
## Testing

1. add this key to the "uiConfiguration" key in Config.dev.ts
```
features: {
  maintenance: true
},
```
2. enable useUIConfigFixtures
3. Visit the node details page
4. Choose the details tab for more status clarity
5. Try deactivating & reactivating nodes
6. Ensure feature is not present unless feature flag is set

Should exhibit the same behaviors as nodes overview table. #4008 & #4004 

## Notes

Suggest reviewing commit-wise

### DIFF
--- a/plugins/nodes/src/js/pages/nodes/NodeDetailPage.js
+++ b/plugins/nodes/src/js/pages/nodes/NodeDetailPage.js
@@ -14,6 +14,9 @@ import Page from "#SRC/js/components/Page";
 import TabsMixin from "#SRC/js/mixins/TabsMixin";
 import RouterUtil from "#SRC/js/utils/RouterUtil";
 import { generateDefaultNetworkErrorHandler } from "#SRC/js/utils/DefaultErrorUtil";
+import { findNestedPropertyInObject } from "#SRC/js/utils/Util";
+// @ts-ignore
+import ConfigStore from "#SRC/js/stores/ConfigStore";
 
 import { Status, actionAllowed, StatusAction } from "../../types/Status";
 import DrainNodeModal from "../../components/modals/DrainNodeModal";
@@ -139,6 +142,15 @@ class NodeDetailPage extends mixin(TabsMixin, StoreMixin) {
   }
 
   getActions() {
+    const hasMaintenance = findNestedPropertyInObject(
+      ConfigStore.get("config"),
+      "uiConfiguration.features.maintenance"
+    );
+
+    if (!hasMaintenance) {
+      return [];
+    }
+
     const actions = [];
     const { node } = this.props;
     const status = Status.fromNode(node);
--- a/plugins/nodes/src/js/pages/nodes/NodeDetailPage.js
+++ b/plugins/nodes/src/js/pages/nodes/NodeDetailPage.js
@@ -1,4 +1,4 @@
-import { i18nMark } from "@lingui/react";
+import { i18nMark, withI18n } from "@lingui/react";
 import { Trans } from "@lingui/macro";
 import mixin from "reactjs-mixin";
 /* eslint-disable no-unused-vars */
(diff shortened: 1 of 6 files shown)


plugins/nodes/src/js/pages/nodes/NodeDetailPage.js
```diff
@@ -128,6 +141,67 @@ class NodeDetailPage extends mixin(TabsMixin, StoreMixin) {
...
+
+    if (!hasMaintenance) {
+      return [];
+    }
+
+    const actions = [];
+    const { node } = this.props;
+    const status = Status.fromNode(node);
+
+    if (actionAllowed(StatusAction.DRAIN, status)) {
```
natmegs: Is there any possibility of using `NodeActionsDropdown` here? I see that the action shape is slightly different but thought I'd check since they both seem to use reactjs-components `Dropdown` under the hood

GeorgiSTodorov:
Looks good, but I think we should add some integration tests. I can help you if you want.

plugins/nodes/src/js/pages/nodes/NodeDetailPage.js
brandonc: Hi-- This was my original intent but it turns out that PageHeader has it's own actions rendering logic and API. I found it was better to duplicate the actions here instead of rewrite page header actions to look more like reactjs-dropdown

src/js/utils/DefaultErrorUtil.tsx
```diff
@@ -0,0 +1,38 @@
+import { i18nMark } from "@lingui/react";
+
+// @ts-ignore
```
TattdCodeMonkey: I don't think this `// @ts-ignore` is needed

src/js/utils/DefaultErrorUtil.tsx
```diff
@@ -0,0 +1,38 @@
...
+  ToastAppearance
+} from "@extension-kid/toast-notifications";
+
+import container from "#SRC/js/container";
+import {
+  NotificationService,
+  NotificationServiceType
+} from "@extension-kid/notification-service";
+
+const notificationService = container.get(
```
TattdCodeMonkey: you can also type this like: `const notificationService = container.get<NotificationService>(NotificationServiceType);`

src/js/utils/DefaultErrorUtil.tsx
```diff
@@ -0,0 +1,38 @@
...
+  NotificationServiceType
+) as NotificationService;
+
+const generateDefaultNetworkErrorHandler = (i18n: any) => {
```
TattdCodeMonkey: since you already have `container` imported you could get the i18n object from it.

```javascript
import { I18n } from "@lingui/core";
import { TYPES } from "#SRC/js/types/containerTypes";
const i18n = container.get<I18n>(TYPES.I18n);
```

brandonc:
@GeorgiSTodorov done, thanks!

src/js/utils/DefaultErrorUtil.tsx
```diff
@@ -0,0 +1,38 @@
...
+  return ({ message, code }: { message: string; code: number }) => {
+    notificationService.push(
+      new ToastNotification(i18n._(i18nMark("Cannot Reactivate Node")), {
+        appearance: ToastAppearance.Danger,
+        autodismiss: true,
+        description:
+          code === 0
+            ? i18n._(i18nMark("Network is offline"))
+            : i18n._(
```
TattdCodeMonkey: What do you think about:

```
i18n._(
  i18nMark(
    "Unable to complete request. Please try again. The error returned was: {message}"
  ),
  { message: ` ${code} ${message}` }
)
```

brandonc: Nice, thanks

PR #4015, opened 2019-07-31
DCOS_OSS-5279 setup notification when logged in

TattdCodeMonkey:
Updated the ui-update plugin to wait until a user is logged in to setup
notifications. Also added code to unsubscribe notification streams when
a user is logged out or redirected to login

- Closes DCOS_OSS-5279

## Testing

Open the console at the login page and verify there are no 401 errors


GeorgiSTodorov:
I think this is only happening at the open source login page.

PR #4017, opened 2019-08-01
DCOS-20150: feat(jsoneditor): save state of the json editor globally

Poltergeist:
This introduces the feature to save the JSON editor state so if you activated the json editor once
it will always automatically get opened on that machine.

CLOSES DCOS-20150

## Testing

Open the Service create form and activate the JSON editor, then Reload the page and ensure the JSON editor is open after opening the form again. This setting should be system wide so all JSON editors should be active upon opening the corresponding forms.

Also after deactivating the JSON editor this should also be deactivated system wide.

## Trade-offs

N/A except that the JSON editor will be active everywhere.

## Dependencies

- N/A -

## Screenshots

- N/A - No Visual change should have happened.


GeorgiSTodorov:
@Poltergeist , I added an integration test, I hope you don't mind.
Regarding the functionality itself, I like it, but I think it would be better if we don't just set "JSONEditor" to false in the local storage, but do something like {isAutoExpanded: value} like we do for the sidebar. The current wording makes it seem like we disable the JSON editor.
Screenshot from 2019-08-02 10-32-48

Poltergeist:
@GeorgiSTodorov Do you suggest to rename `JSONEditor` to `JSONEditorExpanded` for example? I was refering to `deactive` and `active` because we use this terminology everywhere in the code base. But I can get behind the `expanded` as well.

GeorgiSTodorov:
@Poltergeist , I suggested an object in case we want to save other JSON settings in the future like the width.

src/js/utils/LocalStorageUtil.ts
```diff
@@ -0,0 +1,17 @@
+const LocalStorageUtil = {
+  set(_key: string, _value: unknown) {
+    return (global as any).localStorage.setItem.apply(
+      (global as any).localStorage,
+      arguments
```
pierrebeitz: why not use `_key` and `_value` here?

i see no point in using `apply`. i know it was this way before, but maybe we can change that due to the transition to typescript?

src/js/stores/UserSettingsStore.ts
```diff
@@ -36,6 +37,27 @@ class UserSettingsStore extends EventEmitter {
   get storeID() {
     return "userSettings";
   }
+
+  get JSONEditorExpandedSetting(): boolean {
+    const settings = (this.getKey(JSON_EDITOR_SETTING) || {}) as {
```
pierrebeitz: maybe we can do without a type-cast?

that would most likely also make sure that our code can not possibly break here (which it currently could, if `this.getKey(JSON_EDITOR_SETTING)` returned something truthy that's not an object.

src/js/utils/LocalStorageUtil.ts
```diff
@@ -0,0 +1,17 @@
+const LocalStorageUtil = {
+  set(_key: string, _value: unknown) {
+    return (global as any).localStorage.setItem.apply(
```
pierrebeitz: can we do without the type cast?

maybe you like one of these options:
* extend global to know about the builtin `WindowLocalStorage`-type.
* use `window` instead of `global`. the type-hassle should vanish.

src/js/stores/__tests__/UserSettingsStore-test.js
```diff
@@ -47,4 +47,64 @@ describe("UserSettingsStore", function() {
...
+      LocalStorageUtil.set = jasmine.createSpy();
+    });
+
+    afterEach(function() {
+      LocalStorageUtil.set = thisPrevSet;
+    });
+
+    it("sets the JSONEditor key to true", function() {
+      UserSettingsStore.setJSONEditorExpandedSetting(true);
+      expect(LocalStorageUtil.set).toHaveBeenCalledWith(
```
pierrebeitz: what we're testing is that `UserSettingsStore` calls `LocalStorageUtil`. that does not make me very confident that our code works. 

we could make those tests more meaningful if we only tested against the public interface of `UserSettingsStore`, right?

like "set to `a`, get value, is it `a`? set to `b`, get value, is it `b`?".

src/js/stores/UserSettingsStore.ts
```diff
@@ -36,6 +37,27 @@ class UserSettingsStore extends EventEmitter {
...
+      expanded?: boolean;
+    };
+
+    if (settings.expanded && typeof settings.expanded === "boolean") {
```
pierrebeitz: if we have a `typeof` for `boolean` here, we should also guard against `settings` not being an object.

src/js/stores/UserSettingsStore.ts
```diff
@@ -2,7 +2,8 @@ import EventEmitter from "events";
 
 import LocalStorageUtil from "../utils/LocalStorageUtil";
 
-const LOCAL_STORAGE_KEY = "dcosUserSettings";
+const LOCAL_STORAGE_KEY: string = "dcosUserSettings";
+const JSON_EDITOR_SETTING: string = "JSONEditor";
```
pierrebeitz: we have other stuff that stores data in the `UserSettingsStore`. that other stuff knows the key to store stuff by and does not introduce an extra method within `UserSettingsStore`. do we have a reason to introduce a new pattern here?

src/js/utils/LocalStorageUtil.ts
Poltergeist: I was wondering about that as well. But I was more focused on the conversion and not refactoring. I am happy to refactor it.

src/js/stores/UserSettingsStore.ts
Poltergeist: Since I have converted the store to typescript I was considering using the existing functions with a return type of `unknown` but I considered that this would result into an awful lot of unnecessary code ensuring that something is of a certain type. And frankly speaking I was trying to set a good example by creating an abstraction with a clear return value and handling the type validation within those functions. Also I did not need to pass the store key to the components who would need the setting. 

So yes I had a reason to introduce the new pattern. That reason was a better abstraction and type validation which otherwise would have been unnecessary code duplication within this PR.

src/js/stores/__tests__/UserSettingsStore-test.js
pierrebeitz: we discussed that offline. i still think that tests that are tied to implementation details are not a good practice. i might be wrong though.

GeorgiSTodorov:
The build is failing for some reason, I am retriggering.

src/js/utils/LocalStorageUtil.ts
Poltergeist: done

src/js/stores/UserSettingsStore.ts
Poltergeist: fixed

src/js/utils/LocalStorageUtil.ts
Poltergeist: done

src/js/stores/UserSettingsStore.ts
Poltergeist: done

src/js/stores/UserSettingsStore.ts
```diff
@@ -0,0 +1,61 @@
...
+import LocalStorageUtil from "../utils/LocalStorageUtil";
+
+const LOCAL_STORAGE_KEY: string = "dcosUserSettings";
+const JSON_EDITOR_SETTING: string = "JSONEditor";
+const isObject = (data: unknown): data is Record<string, unknown> =>
+  typeof data === "object" && data != null;
+
+function getLocalStorageObject() {
+  const localStorageObject = LocalStorageUtil.get(LOCAL_STORAGE_KEY);
+  if (localStorageObject != null) {
```
pierrebeitz: why do we need to introduce this check here?

src/js/stores/UserSettingsStore.ts
pierrebeitz: now that makes sense. thanks for explaining. we could create a follow-up issue to convert the "old style" to the new as that does not look like a lot of work.

src/js/utils/LocalStorageUtil.ts
pierrebeitz: much better! thanks for following up! 😍

src/js/stores/__tests__/UserSettingsStore-test.js
Poltergeist: Let's try to discuss this in a bigger group. I feel you and we definitely want to get a better test quality. I'll do a follow up to this after the discussion to  unblock the PR right now, if that is okay for you.

src/js/stores/UserSettingsStore.ts
Poltergeist: null would occur in situations where a given key is not set in localStorage. The return value of the `localStorage.getItem` is `string | null`. So without the check I was given a TS problem.

PR #4022, opened 2019-08-06
DCOS-57015: Jobs - Enable/Disable Schedule sets concurrencyPolicy to ALLOW

GeorgiSTodorov:
Persist concurency policy when enabling and
disabling schedule using the dropdown.

Closes https://jira.mesosphere.com/browse/DCOS-57015

## Testing
0. Go to Jobs tab.
1. Create a job with a schedule (doesn't matter if it is enabled or not). It just has to have "Concurrency Policy" set  to false.
Example JSON:
```
{
  "id": "new-job-1",
  "run": {
    "cpus": 1,
    "mem": 128,
    "disk": 0,
    "gpus": 0,
    "cmd": "sleep 1000",
    "artifacts": [],
    "maxLaunchDelay": 3600,
    "restart": {
      "policy": "NEVER"
    },
    "ucr": {
      "image": {
        "id": "nginx",
        "kind": "docker",
        "forcePull": false
      },
      "privileged": false
    }
  },
  "schedules": [
    {
      "id": "123",
      "cron": "1 2 3 4 5",
      "timezone": "UTC",
      "startingDeadlineSeconds": 900,
      "concurrencyPolicy": "FORBID",
      "enabled": false,
      "nextRunAt": "2020-04-03T02:01:00.000+0000"
    }
  ]
}
```
2. Go to that job's detail screen.
3. Use the actions dropdown on the top right to enable or disable the schedule.
4. Open the job and verify that Concurrency Policy did not change to true.

## Trade-offs
None.

## Dependencies
None.

### Before
Peek 2019-08-06 11-18

### After
Peek 2019-08-06 10-59


plugins/jobs/src/js/components/form/helpers/DefaultFormData.ts
```diff
@@ -10,7 +10,8 @@ import {
 export const JobDataPlaceholders = {
   maxLaunchDelay: 3600,
   timezone: "UTC",
-  startingDeadlineSeconds: 900
+  startingDeadlineSeconds: 900,
+  concurrencyPolicy: "FORBID"
```
natmegs: Let's remove this or make the placeholder "ALLOW", as that is what the metronome API defaults to

plugins/jobs/src/js/types/JobSchedule.ts
```diff
@@ -1,19 +1,35 @@
...
+export type ConcurrentPolicyType = "ALLOW" | "FORBID";
+
+export const ConcurrentPolicySchema = `
+  enum ConcurrentPolicy {
+    ALLOW
+    FORBID
+  }
+  `;
+
+export type ConcurrencyPolicyType = "ALLOW" | "FORBID";
```
natmegs: Can we remove this and line 13 in favour of the type from `JobFormData`?

plugins/jobs/src/js/components/form/helpers/DefaultFormData.ts
GeorgiSTodorov: Okay, I made it "forbid" since the checkbox is unchecked by default.

plugins/jobs/src/js/types/JobSchedule.ts
```diff
@@ -1,19 +1,35 @@
...
   enabled: boolean;
   id: string;
   startingDeadlineSeconds: number;
   timezone: string;
+  concurrencyPolicy: ConcurrentPolicy;
 }
 
+export type ConcurrentPolicyType = "ALLOW" | "FORBID";
+
+export const ConcurrentPolicySchema = `
```
pierrebeitz: is there a difference between Concurren**t**Policy`and Concurren**cy**Policy? If not i'd  prefer to stick to one of those terms for better grep-/discoverability.

GeorgiSTodorov: ConcurrentPolicy is the name of the type and concurrencyPolicy is the name of the attribute.

pierrebeitz: i just wanted to make sure that the difference in spelling was intentional. 👍

PR #4025, opened 2019-08-06
don't let useUIConfigFixtures interfere with cypress

pierrebeitz:
we already stub the plugins-setting of ui-config.json in our tests.
if we load a developers usual fixture here, the followin happens:

* the (ee-)plugins specified in that fixture will be loaded
* those plugins might make requests
* those requests might not be stubbed
* unstubbed requests return a bad status code
* the router points us to #/access-denied
* the test is broken :(

i'm usually not a fan of changing production-code so the tests don't break, but
this change seems to be worth it, as devs don't need to figure out what's going
on when integration-tests fail.

IF WE WANTED TO SOLVE THIS PROPERLY
we would need to change the way cypress and `RequestUtil.stubRequest` interact,
because it currently hides stubbed requests from cypress completely, directly
firing (flux- or redux-)actions.
coming up with an elegant solution will require more holistic changes. maybe
something for a spike day!

## Testing

verify that nothing breaks and that you can run cypress regardless of whether `useFixtures` or `useUIConfigFixtures` are true.

## Trade-offs

we touch production code to make tests run better... 😒

### DIFF
--- a/src/js/config/Config.ts
+++ b/src/js/config/Config.ts
@@ -92,4 +92,11 @@ if (Config.environment === "development") {
   Config.useFixtures = false;
 }
 
+// we want to stub requests sometimes.
+// as tests bring their own, we never want to stub if cypress is at work.
+if (window.Cypress) {
+  Config.useFixtures = false;
+  Config.useUIConfigFixtures = false;
+}
+
 export default Config;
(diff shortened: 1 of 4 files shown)


GeorgiSTodorov:
The second test of `JobsOverview-cy` broke for me with `useFixtures: true` and `useUIConfigFixtures: false`.
Screenshot from 2019-08-06 15-09-23

pierrebeitz:
thanks for finding that issue @GeorgiSTodorov!

that led to a change i like much better! :)

GeorgiSTodorov:
@pierrebeitz , do you want a second person to take a look?

PR #4030, opened 2019-08-07
remove internalstoragemixin

pierrebeitz:
# ⚠️ 	must be merged with https://github.com/mesosphere/dcos-ui-plugins-private/pull/1054

have a look at the internalStorageMixin (i removed some irrelevant details):

```javascript
var InternalStorageMixin = {
  internalStorage_data: {},

  internalStorage_get() {
    return this.internalStorage_data;
  },

  internalStorage_update(diff) {
    Object.assign(this.internalStorage_data, diff);
  },

  internalStorage_set(data) {
    this.internalStorage_data = data;
  }
};
```

it only manages some internal property called `internalStorage_data`. most
likely to have a dedicated place that's NOT a components state (which would
rerender on update).

unfortunately it has a couple drawbacks:
* it is misused in several places, as a `forceRender()` is called after updates.
* it makes it more difficult to comprehend things as it introduces mental
  overhead when looking at usages.
* it's hard to express mixins in typescript

so i thought it might be easy enough to just remove it.

# Review

the idea is that for every property formerly stored in the internalStorage, we
introduce an instance member, like so

```diff
-    this.internalStorage_set({
-      clipPathID: `clip-${Util.uniqueID()}`,
-      maskID: `mask-${Util.uniqueID()}`
-    });
+    this.clipPathID = `clip-${Util.uniqueID()}`;
+    this.maskID = `mask-${Util.uniqueID()}`;
```

also note that `reactjs-mixin` automatically makes a class a `React.Component`,
so at the two places we could completely remove all mixins, that class now
extends `React.Component`
that's all there is about it. the rest of changes is very mechanical.

If there were REALLY straightforward refactorings i went for it, but there
should be very little noise aside from just removing the mixin.

i suggest reviewing commit-by-commit

# Testing

integration-tests should have our backs. i would rather suggest thinking about the change for a couple minutes and see if there is any flaw in the conversion than doing extensive smoke testing (i hope i looked at every page that was touched).


GeorgiSTodorov:
I was initially concerned about using instance variables in react, but after a quick research, I think it's fine.
https://www.reddit.com/r/reactjs/comments/54gl72/why_use_state_vs_class_properties/
https://stackoverflow.com/questions/25207703/instance-v-state-variables-in-react-js

@pierrebeitz , what do you think about storing these variables inside an object like
`this.internalStorage.mesosHosts = getMesosHosts(this.state);`
because right now they are all over the place.
`this` in `NodesAgents`:
Screenshot from 2019-08-08 10-29-08

pierrebeitz:
fixed the bug: https://github.com/mesosphere/dcos-ui-plugins-private/pull/1054/commits/960a3f6a048a1d219bfe01e83bd010faa2d3ea29 and double checked that this exact error won't appear somewhere else.

> I was initially concerned about using instance variables in react, but after a quick research, I think it's fine.

that's what we've been doing before. its name was `internalStorage_data`. so we should be fine using this method.

> what do you think about storing these variables inside an object like
this.internalStorage.mesosHosts = getMesosHosts(this.state);

i am not sure what the motivation of this idea is. we changed the location of the data from being at `this.internalStorage_data.mesosHosts` to only `this.mesosHosts`. it's still contained in there. would you prefer something that has `internalStorage` in the name or did you think we splatted all the properties into the `this` (which should not be the case)?

GeorgiSTodorov:
> i am not sure what the motivation of this idea is. we changed the location of the data from being at this.internalStorage_data.mesosHosts to only this.mesosHosts. it's still contained in there. would you prefer something that has internalStorage in the name or did you think we splatted all the properties into the this (which should not be the case)? the other things you see on this are from other mixins, react, ...

It's not a big deal for me, but I think it would be more clear what we are doing with these if we put them all inside a single object. Do you disagree?

pierrebeitz:
> It's not a big deal for me, but I think it would be more clear what we are doing with these if we put them all inside a single object. Do you disagree?

they already are in a single object with the name `mesosHosts`, or am i missing a point here?

GeorgiSTodorov:
I am saying is that all the properties that we are saving to this can be inside a single object. For example, in `TimeSeriesChart.js`, we can save both `clipPathID and maskID` in a single object internalStore = {clipPathId: a, maskID: b}. My original exmaple was bad, I didn't realize we only had `mesosHosts` in `NodesAgents`.

pierrebeitz:
# 💡!

Thanks for clarification.

i would tend not to do so, as IIRC the maximum of properties is two for all touched components. if you think it would help, i would not mind if you changed the places where we have more than one internal property though.

GeorgiSTodorov:
I think it's fine. I thought we had a lot more properties for some reason.

PR #4034, opened 2019-08-07
[DCOS-57146] Services quota table

brandonc:
Adds a table in the quota overview to see quota limit information for services and subgroups.

## Testing

Visit a quota tab of a quota group. You can add services and groups to the group

## Screenshots

Screen Shot 2019-08-07 at 3 40 31 PM

## Notes

- I don't know how to create a service that has tasks with mixed roles


plugins/services/src/js/columns/QuotaOverviewLimitColumn.tsx
```diff
@@ -97,3 +91,47 @@ export function limitRenderer(group: ServiceGroup) {
...
+    getQuotaLimit({
+      count: servicesCount,
+      groupRoleCount
+    }),
+    servicesCount - groupRoleCount
+  ];
+}
+
+export function limitRenderer(item: ServiceGroup | ServiceTree | Service) {
+  var limit: string = "N/A";
```
GeorgiSTodorov: Can we call this something else like limitString?

plugins/services/src/js/columns/QuotaOverviewLimitColumn.tsx
```diff
@@ -97,3 +91,47 @@ export function limitRenderer(group: ServiceGroup) {
...
+  var limitNumber: number = 0;
+
+  if (itemIsServiceGroup(item)) {
+    if (!item.quota || !item.quota.serviceRoles) {
+      return <React.Fragment />;
```
GeorgiSTodorov: Why do we have to return a React.Fragment here?

GeorgiSTodorov:
> I don't know how to create a service that has tasks with mixed roles.

For testing, we can use fixtures I guess.

When I have a group inside a group, I can click the nested group in the table.
Peek 2019-08-08 13-24

TattdCodeMonkey:
> When I have a group inside a group, I can click the nested group in the table and it shows the cards for the first group.

That is the behavior I would expect, the quota is on the top-level group not the sub group. We should allow navigating into sub groups so we can see the status of the services inside that group, but display the quota for the entire top-most parent group.

We can confer with @izharbuendia if that was his expectation as well.

> In services, we have one + button that has both services and groups. In quota, we have two buttons. It should be only one button.

We should fix this in a different PR, file a bug?

> After I create a service in the quota tab, it takes me back to the services tab. Is this expected?

That is also what I would expect.

plugins/services/src/js/columns/QuotaOverviewLimitColumn.tsx
brandonc: If I recall, the caller interface demanded a React element

GeorgiSTodorov:
Rebased.

GeorgiSTodorov:
I no longer see a Quota tab when on the services overview table. It's only visible after clicking on a group.
Peek 2019-08-09 11-51

plugins/services/src/js/containers/services/ServiceTreeView.js
```diff
@@ -123,7 +123,7 @@ class ServiceTreeView extends React.Component {
     }
     const { serviceTree } = this.props;
     const id = serviceTree.getId();
-    if (serviceTree.isRoot() || hasQuota) {
```
GeorgiSTodorov: Why was this removed?

plugins/services/src/js/structs/Service.js
```diff
@@ -141,6 +141,23 @@ module.exports = class Service extends Item {
     return this.get("role") || "";
   }
 
+  getQuotaRoleStats(roleName) {
+    return (this.get("tasks") || []).reduce(
```
TattdCodeMonkey: Have we tested this with a `Pod`? I want to make sure `tasks` is available the same way or if we need to override there like we had to do for `getRole`

plugins/services/src/js/structs/Service.d.ts
```diff
@@ -1,11 +1,18 @@
 import Item from "#SRC/js/structs/Item";
 import { Status } from "#PLUGINS/services/src/js/constants/ServiceStatus";
 
+interface QuotaRolesStats {
```
TattdCodeMonkey: Should we move this to another file instead of duplicating here and in `ServiceTree.d.ts` & `ServiceGroup`?

plugins/services/src/js/columns/QuotaOverviewLimitColumn.tsx
```diff
@@ -108,17 +108,17 @@ export function getLimitInfoForService(item: Service | ServiceTree): LimitInfo {
 
   return {
     limitText: getQuotaLimit({
-      count: stats.servicesCount,
+      count: stats.count,
```
GeorgiSTodorov: Why are we renaming this? I think that servicesCount is a lot more descriptive.

GeorgiSTodorov:
I rebased and added the things that were missing. Feel free dropping any of my commits that are not helpful.

plugins/services/src/js/columns/QuotaOverviewLimitColumn.tsx
TattdCodeMonkey: `servicesCount` isn't accurate since it could be a Pod, Service or Task that we are counting with this variable. I wondering if there is still something we can prefix `count` with to be more descriptive, or being a part of the `ServiceQuotaRolesStats` is enough?

plugins/services/src/js/components/ServicesQuotaOverviewTable.tsx
```diff
@@ -1,83 +1,63 @@
...
-      return compatatorFor("memory");
-    case "disk":
-      return compatatorFor("disk");
-    case "gpus":
-      return compatatorFor("gpus");
+      return (a, b) => a.getName().localeCompare(b.getName());
     default:
-      return () => 0;
+      return (a, b) => {
+        const aText = getLimitInfoForService(a).limitText;
```
TattdCodeMonkey: I wonder if we should be caching either the limit or the QuotaStats on the `Service` || `ServiceTree`. Counting everything every render & sort seems like a waste, but maybe that is also a pre-mature optimization.

plugins/services/src/js/components/ServicesQuotaOverviewTable.tsx
```diff
@@ -1,83 +1,63 @@
...
   ServicesQuotaOverviewTableProps,
-  ServicesQuotaOverViewTableState
+  ServicesQuotaOverviewTableState
 > {
   constructor(props: Readonly<ServicesQuotaOverviewTableProps>) {
     super(props);
 
     this.state = {
-      groups: [],
+      items: props.serviceTree.getItems(),
```
TattdCodeMonkey: Should we do a default sort on these items here?

TattdCodeMonkey:
Integration test failures.

plugins/services/src/js/components/ServicesQuotaOverviewTable.tsx
brandonc: I don't think so. This is very cheap to derive and that is The React Way

plugins/services/src/js/components/ServicesQuotaOverviewTable.tsx
brandonc: Great catch!

plugins/services/src/js/structs/Service.js
brandonc: Pods are supported now- insofar as they are enforced if they have a role that matches the group

GeorgiSTodorov:
Rebased.

GeorgiSTodorov:
I solved the issue of the invisible root table.

TattdCodeMonkey:
This needs another rebase :/

GeorgiSTodorov:
Rebased.

plugins/services/src/js/columns/QuotaOverviewLimitColumn.tsx
```diff
@@ -89,11 +89,70 @@ export function limitRenderer(group: ServiceGroup) {
...
+          id={limitInfo.limitText}
           render="span"
           className="table-cell-value table-cell-flex-box"
         />
       </div>
     </TextCell>
   );
 }
+
+function itemIsServiceGroup(item: any): item is ServiceGroup {
```
pierrebeitz: that pattern can lead to runtime errors as `itemIsServiceGroup` gladly takes any input, but calls `.quota` on it, which will fail for e.g. all primitive types.

plugins/services/src/js/columns/QuotaOverviewLimitColumn.tsx
```diff
@@ -89,11 +89,70 @@ export function limitRenderer(group: ServiceGroup) {
...
+  return item.quota && item.quota.limitStatus;
+}
+
+export function getLimitInfoForPod(item: Pod): LimitInfo {
```
pierrebeitz: is this exported intentionally? it looks like `getLimitInfoForService` is good enough as public API.

plugins/services/src/js/structs/Service.d.ts
```diff
@@ -1,11 +1,14 @@
 import Item from "#SRC/js/structs/Item";
 import { Status } from "#PLUGINS/services/src/js/constants/ServiceStatus";
+import { ServiceQuotaRolesStats } from "../types/ServiceQuotaRolesStats";
 
 export default class Service extends Item {
   getId(): string;
   getMesosId(): string;
   getName(): string;
   getRole(): string;
+  getRootGroupName(): string;
```
pierrebeitz: this is true, because every serviceTree or service will always contain a `/`, right?

plugins/services/src/js/containers/services/ServicesQuotaView.tsx
```diff
@@ -82,7 +82,7 @@ class ServicesQuotaView extends React.Component<ServicesQuotaViewProps, {}> {
     }
 
     return (
-      <Page dontScroll={serviceTree.isRoot()} flushBottom={true}>
+      <Page dontScroll={false} flushBottom={true}>
```
pierrebeitz: if `dontScroll` is always `false`, can't we delete the prop?

PR #4038, opened 2019-08-08
DCOS-57467 Remove Quota Feature Flag Checks

TattdCodeMonkey:
- Updated group modal to open Advanced Settings when editing an existing group that is not currently set to "Use Group Roles" (`enforceRole: true`)

## Testing

Quota  features should be available regardless of `uiFixtures` being on or off.


mperrotti:
It looks like system tests are failing for the same reason release branches were red earlier this week. Hopefully it goes back to green again soon 🤞

GeorgiSTodorov:
I removed the `handleAdvancedSectionClick` method, since the `AdvancedSection` component has a method for handling expanding and `handleAdvancedSectionClick` wasn't getting called anyway. So right now `expandAdvancedSettings` is used only for the initial expand.

PR #4048, opened 2019-08-09
DCOS-39095: declare that Services Table shows allocated resources

mperrotti:
See the comments on [DCOS-39095](https://jira.mesosphere.com/browse/DCOS-39095) for a little more context

## Testing
Go to the services table and confirm the labels are updated

## Trade-offs
Those column's headers will almost always be truncated

## Dependencies
https://github.com/dcos-labs/ui-kit/pull/389


GeorgiSTodorov:
@mperrotti , do we merge this and then update the ui-kit in another PR or should we wait for the ui-kit to be updated first?

pierrebeitz:
added a commit updating ui-kit

pierrebeitz:
while i see the new title-attributes for normal cells, header cells don't seem to ship with those :/

i guess we need to adjust ui-kit once more.

here are the DOM-nodes i looked at: 
OS Enterprise 2019-08-12 10-14-23

GeorgiSTodorov:
I currently get no tooltip for any of the service table cells.

GeorgiSTodorov:
@mperrotti , which cells do we want tooltips for? Just the resource headings?

GeorgiSTodorov:
Now the snapshots are failing.

mperrotti:
The first solution didn't work with the `<Trans>` component because it uses a higher-order component, so the children are not fully resolved.

Here's a new ui-kit PR that sets the `title` attribute using what's rendered to the DOM: https://github.com/dcos-labs/ui-kit/pull/390

nLight:
@mperrotti snapshot tests are failing

pierrebeitz:
updated ui-kit to the version containing the necessary fix along with the snapshots.

pierrebeitz:
It works as intended. Unfortunately i'm not sure whether we want to ship it like this, because it now feels odd sometimes:

OS 2019-08-14 15-29-59

I guess it's better than before as one now can access more of the invisible text (but seemingly not all as Georgi suggests with his firefox-screenshot above) but it does not feel quite insanely great to me.

I'd propose to merge and follow up on that.

mperrotti:
@pierrebeitz 
Duplicate tooltips:
I agree the duplicate tooltips can be a bit strange, but less so if the React tooltip content is different from the "title" tooltip content. I mentioned this tradeoff to the design team when we were discussing how to resolve the original ticket, and we decided to move forward. We're planning to iterate on design patterns for making truncated text fully visible.

Firefox screenshots with missing titles:
I don't see these screenshots, and I don't think this is still an issue with the latest ui-kit update

`undefined` title in snapshots:
This also happened in ui-kit snapshots. I think it's because we're using React DOM to get the rendered content, and the DOM doesn't get fully rendered in some snapshot tests.

GeorgiSTodorov:
Some values in the instances table don't get tooltips. Is this known and intended?
Peek 2019-08-15 10-30

mperrotti:
@GeorgiSTodorov - the instance table isn't using the ui-kit Table yet

GeorgiSTodorov:
@mperrotti , okay.

PR #4057, opened 2019-08-13
DCOS-57043: Support DELAYED plan status

GeorgiSTodorov:
Add delayed status to plan status.

Closes https://jira.mesosphere.com/browse/DCOS-57043

## Testing
I am not sure how to test this.

## Trade-offs
None.

## Dependencies
None.

## Screenshots
None.


plugins/services/src/js/data/ServicePlansClient.ts
```diff
@@ -10,7 +10,8 @@ export type ServicePlanStatus =
   | "STARTING"
   | "STARTED"
   | "COMPLETE"
-  | "IN_PROGRESS";
+  | "IN_PROGRESS"
+  | "DELAYED";
```
pierrebeitz: i know i'm being annoying, but why do we have a need to duplicate this type (in `ServicePlanStatus.ts`)? can we dedupe it?

TattdCodeMonkey: that is a good catch, not sure why we don't just import from `ServicePlanStatus`

PR #4058, opened 2019-08-13
Rename React methods

GeorgiSTodorov:
## Testing
1. Verify that all instances of "componentWillMount" and "componentWillReceiveProps" have been renamed to "UNSAFE_componentWillMount" and "UNSAFE_componentWillReceiveProps".
2. (Optional) Smoke test the UI to see that everything works
3. (Optional) Check to see that we have less console warnings.

## Trade-offs
I didn't rename `super.componentWillMount`. I don't think that we should.

## Dependencies
None.

## Screenshots
None.

### DIFF
--- a/packages/foundation-ui/src/mount/Mount.js
+++ b/packages/foundation-ui/src/mount/Mount.js
@@ -35,11 +35,11 @@ class Mount extends React.Component {
     });
   }
 
-  componentWillMount() {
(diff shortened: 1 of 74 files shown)


pierrebeitz:
why shouldn't we rename `super.componentWillMount`?

GeorgiSTodorov:
@pierrebeitz , we should, I have a new PR up.

PR #4062, opened 2019-08-14
Rename React methods

GeorgiSTodorov:
Everything should still work the same way as before.

EDIT: please review commit-by-commit as there is a noise change in between that has been introduced to shut up the `StoreMixin` (and make it fully functional again).

You'll want to merge https://github.com/mesosphere/dcos-ui-plugins-private/pull/1057 along.

EDIT 2:
sorry for delaying this but it turned out that seemingly we can get rid of ALL `super.UNSAFE_*`s and the `SaveStateMixin`! 🎉


plugins/nodes/src/js/pages/nodes/NodeDetailPage.js
```diff
@@ -83,7 +83,7 @@ class NodeDetailPage extends mixin(TabsMixin, StoreMixin) {
     this.updateCurrentTab(nextProps);
   }
 
-  componentWillUpdate() {
+  UNSAFE_componentWillUpdate() {
     super.componentWillUpdate(...arguments);
```
pierrebeitz: shouldn't we change those calls as well?

GeorgiSTodorov: We should, this PR is still work in progress.

GeorgiSTodorov:
I am looking into the failing unit tests.

GeorgiSTodorov:
Rebased.

GeorgiSTodorov:
For some reason, having `super.UNSAFE_` breaks some pages, for example Networking > Networks and Package Detail. @pierrebeitz , do you know why this might break? Also, it leads to some weird unit test failures.

GeorgiSTodorov:
Thanks, @pierrebeitz .

PR #4064, opened 2019-08-14
DCOS-50331: TypeError: Cannot convert undefined or null to object

GeorgiSTodorov:
Add a non-object check and a null check
to the findVip function.

Closes https://jira.mesosphere.com/browse/DCOS-50331

## Testing
1. Go to services tab.
2. Open the create service modal.
3. Open the single container service form.
4. Go to the Networking tab.
5. Click "Add Service Endpoint".
6. Click "Enable Load Balanced Service Address" checkbox.
7. Manually change portDefinitions.labels to null in the JSON Editor.
8. Verify that no console error is shown.
9. Verify the same thing for pods.
10. Verify that the added test makes sense.

## Trade-offs
None.

## Dependencies
None.

### Before
Peek 2019-08-14 14-56

### After
Peek 2019-08-14 14-53

### DIFF
--- a/plugins/services/src/js/utils/VipLabelUtil.js
+++ b/plugins/services/src/js/utils/VipLabelUtil.js
@@ -29,6 +29,9 @@ const VipLabelUtil = {
   },
 
   findVip(labels = {}) {
+    if (typeof labels !== "object" || labels === null) {
+      return undefined;
+    }
     return Object.entries(labels).find(([key, _]) => key.match(/^(vip|VIP)/));
   },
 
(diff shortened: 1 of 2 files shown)


plugins/services/src/js/utils/VipLabelUtil.js
```diff
@@ -29,6 +29,9 @@ const VipLabelUtil = {
   },
 
   findVip(labels = {}) {
+    if (typeof labels !== "object" || labels === null) {
+      return undefined;
```
GeorgiSTodorov: This is undefined, since this is the default for find. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

pierrebeitz:
oh how i wish we had decoders in dcos-ui...

shall we agree to adopt those before implementing the next form?

GeorgiSTodorov:
> shall we agree to adopt those before implementing the next form?

I agree.

PR #4068, opened 2019-08-15
DCOS-49571: Services Table > Resume Service modal has an incomplete error description

GeorgiSTodorov:
Add a readable error for a negative instance count
in service actions modals.

Closes https://jira.mesosphere.com/browse/DCOS-49571

## Testing
1. Go to services tab.
2. Start a service if there are none.
3. Open the scale modal for that service.
4. Enter a negative number and submit.
5. Verify that the new readable error message is shown.
6. Stop the service.
7. Open the resume modal.
8. Enter a negative number and submit.
9. Verify that the new readable error message is shown.
10. Verify that the added tests make sense.

## Trade-offs
1. I wondered if it's worth it adding an util, but I think we might expand it in the future.
2. I wondered about the function name. Suggestions are appreciated.
3. I tried to add an integration test for this, but cypress refused to type a minus.

## Dependencies
None.

### Before
Screenshot from 2019-08-15 11-27-10

### After
Screenshot from 2019-08-15 11-09-46


plugins/services/src/js/components/modals/ServiceResumeModal.js
```diff
@@ -100,7 +101,9 @@ class ServiceResumeModal extends React.PureComponent {
     }
 
     return (
-      <h4 className="text-align-center text-danger flush-top">{errorMsg}</h4>
+      <Trans render="h4" className="text-align-center text-danger flush-top">
```
TattdCodeMonkey: This won't work. If you want to translate `"Must be bigger than or equal to 0"` but then fallback to the error message if it doesn't `== " error.min"` then you'll need to mark and translated `Must be bigger than or equal to 0` in `getActionModalReadableError` and not use `Trans` here.

plugins/services/src/js/utils/ServiceActionModalsUtil.ts
```diff
@@ -0,0 +1,6 @@
+export function getActionModalReadableError(errorMsg: any): any {
```
TattdCodeMonkey: You can import `i18n` from `container` and update this to use it inside the `if` to return the translated text and otherwise default to the untranslated errorMsg.

GeorgiSTodorov:
Thanks, @TattdCodeMonkey .

plugins/services/src/js/utils/ServiceActionModalsUtil.ts
```diff
@@ -0,0 +1,9 @@
+import { t } from "@lingui/macro";
+
+export function getActionModalReadableError(i18n: any, errorMsg: any): any {
+  if (errorMsg === " error.min") {
+    return i18n._(t`Must be bigger than or equal to 0`);
```
pierrebeitz: let's use trans here, so we don't have to pass the `i18n`-prop down all the way to this helper.

GeorgiSTodorov:
Thank you, @pierrebeitz !

plugins/services/src/js/utils/ServiceActionModalsUtil.ts
```diff
@@ -0,0 +1,6 @@
+export function getActionModalReadableError(errorMsg: any): any {
+  if (errorMsg === " error.min") {
+    return "Must be bigger than or equal to 0";
```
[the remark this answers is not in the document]
GeorgiSTodorov: So the `<Trans>` in the modal won't work for extracting this? Good catch, thank you.

PR #4069, opened 2019-08-15
DCOS-57614: table-multiselect doesn't work as expected

GeorgiSTodorov:
Allow multiple job run to be selected
in the job run history table

Closes https://jira.mesosphere.com/browse/DCOS-57614

## Testing
1. Go to jobs tab.
2. Create a job that wouldn't finish immediately.
Example JSON:
```
{
  "id": "new-job-1",
  "description": "",
  "run": {
    "cpus": 1,
    "disk": 0,
    "mem": 128,
    "ucr": {
      "image": {
        "id": "nginx",
        "kind": "docker"
      }
    },
    "cmd": "sleep 1000"
  }
}
```
3. Open it.
4. Run it at least once.
5. In the runs history table, verify that the actions dropdown exists and it contains the stop actions. It should only exist for running instances.
6. Verify that clicking stop works.
7. Verify that overall the table works the same way as before.
8. Verify that the added test makes sense.

## Trade-offs
~https://github.com/dcos/dcos-ui/pull/4069#discussion_r314611962~

## Dependencies
None.

### Before
Peek 2019-08-15 12-46-2

### After
Peek 2019-08-23 15-43


GeorgiSTodorov:
Unrelated system tests are failing. Maybe we broke master?

pierrebeitz:
does not seem to work as intended: 

screencast 2019-08-15 15-59-35

GeorgiSTodorov:
The failing system tests are unrelated to this PR and are also failing on other PRs.

GeorgiSTodorov:
@pierrebeitz , the bug you mentioned is actually a known one: https://jira.mesosphere.com/browse/DCOS-8763 . I guess this means we need to change the table type.

GeorgiSTodorov:
Alternatively, we can just run through the selected items.

src/js/components/modals/JobStopRunModal.js
```diff
@@ -35,17 +35,18 @@ class JobStopRunModal extends mixin(StoreMixin) {
 
   handleButtonConfirm() {
     const { selectedItems, jobID } = this.props;
-    // TODO DCOS-8763 introduce support for multiple job run IDs
 
-    if (selectedItems.length === 1) {
-      const jobRunID = selectedItems[0];
+    let i = 0;
+    selectedItems.forEach(function() {
```
GeorgiSTodorov: Now we have to send a network request for every run, which can mean thousands of network requests. I  think this downside is non-trivial. @nLight @pierrebeitz 

Alternatively, we can just change the table type to not confuse the users into thinking that multi-selection is possible.
Also, should we backport this?

pierrebeitz: two folks in the office mentioned that this has been bugging them for quite some time.

GeorgiSTodorov: > two folks in the office mentioned that this has been bugging them for quite some time.

What about the customers?

> i think it makes sense to get rid of the illusion that multi-selection would be possible.

I think so too. I am interested in what @nLight has to say.

tests/_fixtures/metronome/job.json
```diff
@@ -130,7 +130,7 @@
         "createdAt": "1990-01-02T12:10:59.571+0000",
         "finishedAt": "1990-01-02T12:11:19.762+0000",
         "id": "20160705121059J7cPJ",
-        "tasks": ["completedTask.42"]
+        "tasks": ["completedTask.42", "failedTask.42"]
```
GeorgiSTodorov: Without this, the test was failing. I am not sure why.

plugins/jobs/src/js/pages/JobRunHistoryTable.js
```diff
@@ -6,6 +6,7 @@ import PropTypes from "prop-types";
 import React from "react";
 import { Icon, Tooltip } from "@dcos/ui-kit";
 import { SystemIcons } from "@dcos/ui-kit/dist/packages/icons/dist/system-icons-enum";
+import { Dropdown } from "reactjs-components";
```
pierrebeitz: why did we chose not to use `ui-kit`s Dropdown?

plugins/jobs/src/js/pages/JobRunHistoryTable.js
```diff
@@ -365,65 +346,78 @@ class JobRunHistoryTable extends React.Component {
...
-
-      return filteredItems;
-    },
-    {});
-
-    if (hasCheckedTasks) {
-      rightAlignLastNChildren = 1;
+    const disabledIDs = this.getDisabledItemsMap(job);
+    if (disabledIDs[selectedID]) {
+      this.handleItemSelect(null);
```
pierrebeitz: what does this do? (i don't mean codewise - i mean "what user story is covered by this"?)

GeorgiSTodorov: The same we did before with https://github.com/dcos/dcos-ui/pull/4069/files/99262bc4d88b14e945ae5149c516d17bfc532604#diff-40f057127c4e7bc0ed23cfdca70e59b3L381 , but for a single ID - we remove it from selectedItem.

plugins/jobs/src/js/pages/JobRunHistoryTable.js
GeorgiSTodorov: We don't use it anywhere else, I used the service table actions as an example.

plugins/jobs/src/js/pages/JobRunHistoryTable.js
pierrebeitz: why would that be necessary now that we only have a single id? is `render` the right place to do this?

plugins/jobs/src/js/pages/JobRunHistoryTable.js
pierrebeitz: let's please make use of `ui-kit` as much as we can!

plugins/jobs/src/js/pages/JobRunHistoryTable.js
GeorgiSTodorov: It's necessary, since the function that should do this https://github.com/dcos/dcos-ui/pull/4069/files#diff-816c1e88efffe7eed722c8faa498cd04R49 doesn't work. It used to be in render and it is still in render. It need a refactoring, but not in this PR.

plugins/jobs/src/js/pages/JobRunHistoryTable.js
GeorgiSTodorov: If we want to use the ui-kit dropdown, we should use it everywhere for consistency, since it looks and behaves different.

PR #4070, opened 2019-08-15
Max-width to breadcrumb text

GeorgiSTodorov:
### Before
Screenshot from 2019-08-15 16-07-43

### After
Screenshot from 2019-08-15 16-05-05


nLight:
Do we have a tooltip when the line is too long

GeorgiSTodorov:
@nLight , no , but I think that's a good idea. @mperrotti , what do you think?

mperrotti:
Sure, I think a tooltip would be good

GeorgiSTodorov:
@mperrotti @nLight 
This is what I have right now.
It looks alright for service with large, but not super large names.
Peek 2019-08-16 11-46

For super large names, the tooltip isn't fully visible, but I don't think these cases are common. And it's nevertheless much better than before.
Peek 2019-08-16 11-46-2

I intentionally made this only in `ServiceBreadcrumbs`. If I add a max width and a tooltip for all breadcrumbs, we have to pass all children to the tooltip, including the link, and I think that this looks ugly.
Peek 2019-08-16 11-56

pierrebeitz:
merging as this is clearly better than before.

before long names could prevent you from even seeing the edit-button without any way to access it after all as overflow is just hidden.

pierrebeitz:
i would love to have a width-option for `Tooltip`, like the `Legacy.Tooltip` has.

what do you think @GeorgiSTodorov 	, @mperrotti ?

GeorgiSTodorov:
@pierrebeitz , it's @mperrotti 's call, but I would like to see what you suggested and also some other imrpovements for the ui-kit tooltip like it not being div, since it is currently harder to style.

PR #4072, opened 2019-08-16
DCOS-57646: refactor(endpointstab): remove service port from endpoints tab

Poltergeist:
This removes the service port from the endpoints tab since the service port was an implementation
detail which is not giving any useful information to the user therefore we decided to remove this
information.

CLOSES DCOS-57646

## Testing

Create a service with a network config,
visit the endpoints tab and ensure it does not include service port information.

## Trade-offs

We will not show the service port in the endpoints tab anymore, but the service port is an implementation detail which is not yielding any information to the user.
## Dependencies

N/A

## Screenshots

N/A


plugins/services/src/js/containers/service-connection/ServiceConnectionEndpointList.js
```diff
@@ -134,12 +134,6 @@ class ServiceConnectionEndpointList extends React.Component {
             {this.getHostPortValue(portDefinition, service)}
           </ConfigurationMapValue>
         </ConfigurationMapRow>
-        <ConfigurationMapRow key="service-port">
-          <Trans render={<ConfigurationMapLabel />}>Service Port</Trans>
-          <ConfigurationMapValue>
-            {this.getServicePortValue(portDefinition)}
```
pierrebeitz: i'd love if we removed `getServicePortValue` as this was the only place we used it.

PR #4073, opened 2019-08-16
DCOS-57645: feat(endpointstab): add weburl to endpoints tab

Poltergeist:
This adds the weburl from a service to the endpoints tab as a click to copy function

CLOSES DCOS-57645

## Testing

Start a  package with a frontend (marathon) for example.
visit the endpoints tab on the detail page and check if the url is present there.

## Trade-offs

N/A

## Dependencies

N/A

## Screenshots

image

### DIFF
--- a/plugins/services/src/js/containers/service-connection/ServiceConnectionEndpointList.js
+++ b/plugins/services/src/js/containers/service-connection/ServiceConnectionEndpointList.js
@@ -147,13 +147,7 @@ class ServiceConnectionEndpointList extends React.Component {
     });
   }
 
-  getEndpoints(service) {
-    const webUrl = service.getWebURL();
-
-    if (webUrl === "") {
-      return null;
-    }
-
+  getEndpoints(webUrl) {
     return (
       <ConfigurationMapSection>
         <ConfigurationMapHeading>
@@ -199,12 +193,13 @@ class ServiceConnectionEndpointList extends React.Component {
         />
       );
     }
+    const webUrl = service.getWebURL();
 
     return (
       <div className="container">
         <ConfigurationMap>
           {this.getPortDefinitions(endpoints, service)}
-          {this.getEndpoints(service)}
+          {webUrl ? this.getEndpoints(webUrl) : null}
         </ConfigurationMap>
       </div>
     );
--- a/plugins/services/src/js/containers/service-connection/ServiceConnectionEndpointList.js
+++ b/plugins/services/src/js/containers/service-connection/ServiceConnectionEndpointList.js
@@ -147,6 +147,30 @@ class ServiceConnectionEndpointList extends React.Component {
     });
   }
 
+  getEndpoints(service) {
+    const webUrl = service.getWebURL();
+
+    if (webUrl === "") {
+      return null;
+    }
+
+    return (
+      <ConfigurationMapSection>
+        <ConfigurationMapHeading>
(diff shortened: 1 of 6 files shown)


plugins/services/src/js/containers/service-connection/ServiceConnectionEndpointList.js
```diff
@@ -214,4 +244,12 @@ ServiceConnectionEndpointList.contextTypes = {
   router: routerShape
 };
 
+function hasWebUI(service) {
+  return (
+    service instanceof Service &&
+    service.getWebURL() != null &&
```
GeorgiSTodorov: Why do we do "!=" here?

plugins/services/src/js/containers/service-connection/ServiceConnectionEndpointList.js
```diff
@@ -165,6 +166,34 @@ class ServiceConnectionEndpointList extends React.Component {
     });
   }
 
+  getEndpoints(service) {
+    const webUrl =
+      service instanceof Pod ||
```
GeorgiSTodorov: Do pods always have web urls?

GeorgiSTodorov: This page isn't even shown for pods as far as I can tell, for them we use `plugins/services/src/js/containers/service-connection/ServicePodConnectionEndpointList.js`.

plugins/services/src/js/containers/service-connection/ServiceConnectionEndpointList.js
```diff
@@ -214,4 +244,12 @@ ServiceConnectionEndpointList.contextTypes = {
   router: routerShape
 };
 
+function hasWebUI(service) {
```
GeorgiSTodorov: What is the point of this check? Can't we just make `getWebURL` return null or "'?

plugins/services/src/js/containers/service-connection/ServiceConnectionEndpointList.js
Poltergeist: Good catch.

GeorgiSTodorov:
An integration test appears to be failing.

GeorgiSTodorov:
For some reason, the web url wasn't showing in the integration test. I removed this check for now and also rebased.

pierrebeitz:
i rebased the fixups directly in master...

GeorgiSTodorov:
Why do we want no fixups on master?

pierrebeitz:
fixup-commits are not meant to be separate commits but rather should be "squashed" into the commits they are considered an amendment for because they modify stuff to be correct in the first place. 

all by themselves they're just unnecessary noise without a proper commit-messsage.

GeorgiSTodorov:
This seems like a pretty big downside to me. I am not sure if it's worth it using them.

pierrebeitz:
i used them to "suggest changes" here as discussed in the sync.

you would run `git rebase -i master` locally to squash them into the places they're meant to be if you like the changes and them push and merge. maybe it's worth discussing this in the sync again.

GeorgiSTodorov:
Yes, I would be in favour of discussing this. If it's a small change, I think we can just say what we changed (also, it would be much more clear in the PR since there would be comments) and if it's a big change then it shouldn't be a fixup. If we have to rebase and push again, we would have to dismiss reviews also.

PR #4074, opened 2019-08-20
DCOS-57500: [UI] Add filter hiding community packages to catalog page

GeorgiSTodorov:
Add a filter to the pacakges page that
allows the user to filter for certified, community
and all packages with certified as default.

Closes https://jira.mesosphere.com/browse/DCOS-57500

## Testing
1. Go to Catalog tab.
2. Verify that the new filters are shown.
3. Verify that clicking them works as expected.
4. Verify that after using the input filter, they work as expected (changing the badge nubmers, etc.).

## Trade-offs
See my comments.

## Dependencies
None.

## Screenshots
Screenshot from 2019-08-20 14-03-50
Screenshot from 2019-08-20 14-05-35


src/styles/views/packages/styles.less
```diff
@@ -9,6 +9,25 @@
     }
   }
 
+  .filter-bar-item {
+
+    .button-outline {
+
+      span {
+        margin: @base-spacing-unit * 1/6;
+        padding: 1px 6px;
```
GeorgiSTodorov: I know @mperrotti said not to change the default styles, but without this the badges look ugly, since they are getting this rule
```
button > :last-child, .button > :last-child {
    padding-right: 0;
```
Screenshot from 2019-08-20 15-23-31

src/js/pages/catalog/PackagesTab.js
```diff
@@ -197,10 +207,28 @@ class PackagesTab extends mixin(StoreMixin) {
     return <Trans id={copy} render={<Badge appearance={appearance} />} />;
   }
 
-  getCertifiedPackagesGrid(packages) {
-    if (this.state.searchString || packages.getItems().length === 0) {
+  getCertifiedPackagesGrid(packages, communityPackages) {
+    const isSearchActive = this.state.searchString !== "";
+    // If we are showing all packages, and we are filtering with a string
```
GeorgiSTodorov: This might be a little confusing at first, but I still think it's clear what we are trying to do.

src/js/pages/catalog/PackagesTab.js
```diff
@@ -284,26 +350,47 @@ class PackagesTab extends mixin(StoreMixin) {
...
+                {this.getFilterButtons(
+                  selectedPackages.list.length,
+                  communityPackages.list.length
+                )}
+              </FilterBar>
             </div>
-            {this.getCertifiedPackagesGrid(selectedPackages)}
-            {this.getCommunityPackagesGrid(communityPackages)}
+            {state.searchFilter !== COMMUNITY
+              ? this.getCertifiedPackagesGrid(
```
GeorgiSTodorov: I debated having only one function for packages grid, but in the end I decided against this.

src/js/pages/catalog/PackagesTab.js
```diff
@@ -29,6 +30,9 @@ import StringUtil from "../../utils/StringUtil";
 import CatalogPackageOption from "./CatalogPackageOption";
 import MetadataStore from "../../stores/MetadataStore";
 
+const CERTIFIED = "certified";
```
GeorgiSTodorov: I don't think it's worth having a constants file just for these two.

GeorgiSTodorov:
Snapshots are failing in CI, but passing locally.

GeorgiSTodorov:
I (hopefully) fixed a failing integration test.

pierrebeitz:
i'm on it. looks like you updated the snapshots with an old version of ui-kit!

pierrebeitz:
thanks for doing the heavy lifting. 
* i changed the behaviour of "view all" to also switch the filter to "all". 
* i made some structural changes to the code.
* i removed some functionality that first confused me. digging through the git-history revealed that by now we should not need to "rank" packages, because they are now rendered in their own sections.

src/js/pages/catalog/PackagesTab.js
pierrebeitz: i pulled the functionality to show the "view all"-message out, so we don't complect rendering the `CertifiedPackagesGrid` with showing that message. they don't need to be coupled IMHO.

src/js/pages/catalog/PackagesTab.js
```diff
@@ -218,102 +239,130 @@ class PackagesTab extends mixin(StoreMixin) {
...
+      });
+
+      const name = StringUtil.capitalize(filter);
+      return (
+        <button
+          key={filter}
+          className={classSet}
+          onClick={this.getSearchFilterChangeHandler(filter)}
+        >
+          <Trans render="b" id={name}>
```
TattdCodeMonkey: if your using the `id` prop then you dont need to put content inside the `Trans` component. This should be written `<Trans render="b" id={name} />` using `{name}` twice is redundant.

GeorgiSTodorov: Thanks, I will change it.

GeorgiSTodorov:
@pierrebeitz , thanks for the help.

PR #4075, opened 2019-08-20
DCOS-49193: fix blurry dialog modals on lo-res screens

mperrotti:
## Testing
This only happens on low-resolution screens, so test this on an external monitor if you use a Macbook Pro or some other high-end machine.

1. Open up a dialog modal (for example: the "Repositories" modal under Settings > Package Repositories)
2. Confirm the text and text input borders aren't blurry
3. Click around to other dialog modals and confirm they don't look blurry either

## Trade-offs
The fix is a little hacky to avoid having to do a CNVS upgrade

## Screenshots
**Before**
BlurryModal_before

**After**
BlurryModal_after

### DIFF
--- a/src/styles/components/modal/styles.less
+++ b/src/styles/components/modal/styles.less
@@ -1,5 +1,23 @@
 & when (@modal-enabled) {
 
+  // using flexbox for centering fixes the blurry subpixel 
+  // rendering of dialog modals on low resolution screens
+  .modal-wrapper {
+    align-items: center;
+    bottom: 0;
+    display: flex;
+    justify-content: center;
+    left: 0;
+    right: 0;
+    top: 0;
+  }
+
+  // part of the flexbox centering fix
+  .modal-small {
+    position: unset !important;
+    transform: none !important;
+  }
+
   .modal {
     // We need the extra specificity here, so we have to nest.
     .modal-header,


pierrebeitz:
this fix does not seem to help much on my external monitor: 
OS Enterprise 2019-08-21 11-02-08

i noticed that the effect gets way worse if you scale the page down a bit (10% is enough to consistently make it REALLY blurry on non-retina-displays)

GeorgiSTodorov:
@pierrebeitz , it does look better for me. Do you want to request changes?

mperrotti:
Thanks for catching that, Pierre.
The flexbox solution may be the most reliable path to take. I'll update the PR

GeorgiSTodorov:
An integration test might be legitimately failing. I will take a look on Monday.

src/styles/components/modal/styles.less
```diff
@@ -1,6 +1,22 @@
...
+    align-items: center;
+    bottom: 0;
+    display: flex;
+    justify-content: center;
+    left: 0;
+    right: 0;
+    top: 0;
+  }
+
   .modal {
```
GeorgiSTodorov: @mperrotti , the failing integration test is fixed if we change this to `modal-small` and I think this won't affect the actual big that we are fixing, since it only happens for small modals anyway.

pierrebeitz: since my screencast-app fails after an update earlier this day i'll try to describe what's happening:

the arrow-right on an image gallery within the packagesDetail is failing. you can have a look at that here: https://soak114s.testing.mesosphe.re/#/catalog/packages/instana-agent?version=1.0.5.

when looking for packages that have screenshots i could reproduce the error from CI on soak. so it looks like we may "just" have made things worse.

PR #4077, opened 2019-08-22
DCOS-57551: [UI] show a warning if a package has not been updated in a while

GeorgiSTodorov:
Add a last updated label for packages
that have not been updated soon.

Closes https://jira.mesosphere.com/browse/DCOS-57551

## Testing
1. Go to catalog tab.
2. Select a package that has been updated recently, for example cassandra and click it.
3. Verify that "Last Updated" label is shown on the package detail tab. The framework configuration tab (after clicking "Review & Run") should look the same.
4. Select a package that hasn't been updated for over a year, for example grafana and click it.
5. Verify that the "Last Updated" label is shown on the package detail tab. Next to it a warning icon with a tooltip should exist. The framework configuration tab (after clicking "Review & Run") should also have a warning icon with a tooltip.
6. Select a package with no last updated property, for example dcos-ui and click it.
7. Verify that the package detail page and the edit configuration page don't break.
8. Verify that the added tests makes sense.

## Trade-offs
See my comments.

## Dependencies
None.

## Screenshots
OS Enterprise 2019-08-28 17-38-49
Screenshot from 2019-08-28 14-20-12


GeorgiSTodorov:
Hold merge as we are waiting for designer feedback.

plugins/services/src/js/utils/FrameworkUtil.js
```diff
@@ -1,7 +1,14 @@
+import { Trans } from "@lingui/macro";
+import React from "react";
+
 import Util from "#SRC/js/utils/Util";
+import DateUtil from "#SRC/js/utils/DateUtil";
 
 import ServiceImages from "../constants/ServiceImages";
 
+const YEAR = 31557600000;
```
GeorgiSTodorov: This is possibly temporary, we are waiting for designer feedback.

plugins/services/src/js/utils/FrameworkUtil.js
```diff
@@ -76,6 +83,52 @@ const FrameworkUtil = {
...
+   * @returns {null | React.ReactNode} null or a label saying when was the last update of this package
+   */
+  getLastUpdated(cosmosPackage) {
+    // This looks like magic, but we are only decoding the package configuration into a string,
+    // and then finding the PACKAGE_BUILD_TIME_STR property.
+    const v2AppMustacheTemplate = Util.findNestedPropertyInObject(
+      cosmosPackage,
+      "marathon.v2AppMustacheTemplate"
+    );
+    if (!v2AppMustacheTemplate) {
```
GeorgiSTodorov: Some packages don't even have marathon property.

plugins/services/src/js/utils/FrameworkUtil.js
```diff
@@ -76,6 +83,52 @@ const FrameworkUtil = {
...
+      return null;
+    }
+
+    const decodedBuildTime = atob(v2AppMustacheTemplate).match(
+      /PACKAGE_BUILD_TIME_STR": "(.*)"/
```
GeorgiSTodorov: And some packages don't have "PACKAGE_BUILD_TIME_STR".

plugins/services/src/js/utils/FrameworkUtil.js
```diff
@@ -76,6 +83,52 @@ const FrameworkUtil = {
...
+
+    const packageUpdatedTime = decodedBuildTime[1];
+    const now = Date.now();
+
+    const difference = now - DateUtil.strToMs(packageUpdatedTime);
+    if (difference < SIX_MONTHS) {
+      return null;
+    }
+
+    const className = difference > YEAR ? "danger" : "warning";
```
GeorgiSTodorov: I decided to show a yellow label for packages older than six months, but younger than a year and a red label for packages older that a year. Again, the designers might change this.

plugins/services/src/js/utils/FrameworkUtil.js
```diff
@@ -76,6 +83,52 @@ const FrameworkUtil = {
...
+
+    // Format to only show the date.
+    const formattedDate = DateUtil.msToUTCDay(packageUpdatedTime);
+
+    // TODO: decide if we want to format the date (for example day-month-year).
```
GeorgiSTodorov: Right not it's year-month-day.

src/js/components/FrameworkConfiguration.js
```diff
@@ -319,6 +324,7 @@ class FrameworkConfiguration extends Component {
             {StringUtil.capitalize(packageDetails.getName()) +
               " " +
               packageDetails.getVersion()}
+            {lastUpdated ? <div>{lastUpdated}</div> : null}
```
GeorgiSTodorov: Again, it's a question for the designers if we want to have this here.

src/js/pages/catalog/PackageDetailTab.js
```diff
@@ -271,6 +272,14 @@ class PackageDetailTab extends mixin(StoreMixin) {
       );
     }
 
+    if (!label) {
```
GeorgiSTodorov: We don't want this label, since it's unstyled.

GeorgiSTodorov:
The tests failing in CI are passing locally for me.

GeorgiSTodorov:
I tried to follow the designer's feedback from the epic as close as possible.

src/js/components/FrameworkConfiguration.js
```diff
@@ -319,6 +324,7 @@ class FrameworkConfiguration extends Component {
             {StringUtil.capitalize(packageDetails.getName()) +
               " " +
               packageDetails.getVersion()}
```
GeorgiSTodorov: I am not sure that this is what @danielone meant when he said "Next to service icon, have the date be part of the selected version number so "Version: 2.6.03.11. (2019-03-12)"
Screenshot from 2019-08-23 14-34-40

tests/pages/catalog/packages/package/PackageTab-cy.js
```diff
@@ -110,6 +110,55 @@ describe("Package Detail Tab", function() {
...
+      });
+
+      it("Shows a yellow warning if the package has not been updated for over a six months", function() {
+        cy.configureCluster({
+          mesos: "1-task-healthy",
+          universePackages: "old"
+        }).visitUrl({ url: "/catalog/packages/marathon?version=1.4.6" });
+
+        cy.get(".last-updated-package-warning").contains(
+          "Last updated: 2019-01-24"
```
GeorgiSTodorov: I think this will eventually fail, since we use the current date when checking. However, I cannot think of a solution now, we will have to change the config after the 24th of January 2020.

plugins/services/src/js/utils/FrameworkUtil.js
```diff
@@ -76,6 +84,62 @@ const FrameworkUtil = {
...
+   * Return a warning or an error label
+   * if the pacakge has cosmosPackage been updated soon.
+   * @param {object} cosmosPackage The cosmos package to check.
+   * @param {Boolean} withLabel Indicates whether we want the "Last Updated: " label or not.
+   * @returns {null | String | React.ReactNode} Null or string or an HTML saying when was the last update of this package
+   */
+  getLastUpdated(cosmosPackage, withLabel) {
+    // This looks like magic, but we are only decoding the package configuration into a string,
+    // and then finding the PACKAGE_BUILD_TIME_STR property.
+    const v2AppMustacheTemplate = Util.findNestedPropertyInObject(
```
GeorgiSTodorov: I wonder if we wanna show nothing for packages that don't have this property or should we do "Last Updated: N/A".

mperrotti:
@danielone and @izharbuendia: I think the yellow for updates > 6 months is way too low contrast on the white background

plugins/services/src/js/utils/FrameworkUtil.js
```diff
@@ -76,6 +84,62 @@ const FrameworkUtil = {
...
+    } else if (difference > SIX_MONTHS) {
+      className = "-warning";
+    }
+
+    // Format to only show the date.
+    const formattedDate = DateUtil.msToUTCDay(packageUpdatedTime);
+
+    if (!withLabel) {
+      return (
+        <span className={"last-updated-package" + className}>
```
mperrotti: You could have avoided writing styles by using the [Text component](https://dcos-labs.github.io/ui-kit/?path=/story/style-utilities-typography-text--color)
```
let dateColor;
 if (difference > YEAR) {
      dateColor = red;
 else if (difference > SIX_MONTHS) {
      dateColor = yellow;
}
<Text color={dateColor} dataCy="last-updated-package">
```

This is definitely not a blocker, just sharing info :)

GeorgiSTodorov:
@mperrotti 
Using the read and the yellow was a decision made be me and @pierrebeitz . However, I used the standard warning yellow that we already had. What would you recommend?

src/styles/components/framework-configuration/styles.less
```diff
@@ -37,3 +37,11 @@
 .glyphicon-plus {
   font-style: normal;
 }
+
+.package-last-updated-danger {
```
pierrebeitz: anything speaking against just reusing `.text-danger` here?

plugins/services/src/js/utils/FrameworkUtil.js
pierrebeitz: will clarify with izhar

src/styles/components/framework-configuration/styles.less
GeorgiSTodorov: Since the design wasn't finished, I wanted to make it easy to add new styles.

pierrebeitz: that makes a lot of sense! thanks for sharing! 💡

GeorgiSTodorov: Maybe that's overengineering,  and won't be needed.

GeorgiSTodorov:
CI is telling us to extract again.

pierrebeitz:
new designs...

download (1)
download (2)

GeorgiSTodorov:
@pierrebeitz , I can take care of these.

src/styles/components/framework-configuration/styles.less
```diff
@@ -37,3 +37,12 @@
 .glyphicon-plus {
   font-style: normal;
 }
+
+.outdated-package-warning-оutline {
+  border: @base-spacing-unit * 1/12 solid @warning-color;
+  border-bottom-left-radius: 5% 20%;
```
GeorgiSTodorov: This will fail the build, since it expects us to use the shorthand property. But the shorthand property `5% 20%` does not yield the same result. And if we use `5% / 20%`, it's interpreted as `4%` probably because we use .less.

plugins/services/src/js/utils/FrameworkUtil.js
```diff
@@ -1,7 +1,17 @@
+import { Trans } from "@lingui/macro";
+
+import React from "react";
+import { Icon, Legacy } from "@dcos/ui-kit";
```
GeorgiSTodorov: I tried to use the ui-kit tooltip, but not only did it require special styling in the package detail tab, in the framework configuration tab it didn't appear at all.

pierrebeitz: argh! ui-kit's tooltip has `z-index: 2`, which sticks behind `.modal-wrapper`s `z-index: 1000`. cc @mperrotti

src/styles/components/framework-configuration/styles.less
pierrebeitz: @mperrotti 	we might want to bring this pattern to `ui-kit`.

it looks like `outline` is not an `appearance` but rather an independent prop that changes the look of a badge different from the usual appearances.

63838781-4740c480-c97e-11e9-9f40-dec13423270b

pierrebeitz:
@GeorgiSTodorov 	i applied the changes i did before again to clean things up a bit. i went without CSS-adaptions for now, waiting for ui-kit to either adapt or reject the pattern introduced by design.

GeorgiSTodorov:
I thought my way was, although longer, more clear on what we are doing and where we are getting the information.

GeorgiSTodorov:
Also, I don't think that everything should be styled in ui-kit.

src/styles/components/framework-configuration/styles.less
GeorgiSTodorov: It's from Izhar.

pierrebeitz:
> I thought my way was, although longer, more clear on what we are doing and where we are getting the information.

i thought you're ok with the way in here, as i pushed that before. it will be replaced with a proper date on the `UniversePackage`-struct, so i don't think we should keep that noise. this way all black magic is in a handful lines and the next pair of eyes won't flee right away.

> Also, I don't think that everything should be styled in ui-kit.

hmm... you're right. now that i think of it i never really saw outlined badges (but only buttons).

GeorgiSTodorov:
@pierrebeitz 
> i thought you're ok with the way in here, as i pushed that before. it will be replaced with a proper date on the UniversePackage-struct, so i don't think we should keep that noise. this way all black magic is in a handful lines and the next pair of eyes won't flee right away.

I am okay with it, I just preferred my way, but it's not a big deal.

PR #4084, opened 2019-08-26
speed up eslint for less painful rebases

pierrebeitz:
there are a couple things to note here:

1. we don't run eslint twice anymore on post-rewrite because we now have
 plugins-ee symlinked. they are linted automatically this way.

2. we update `.eslintignore` for no good reason. i have did not measure the
 impact of changing those.

3. we now use pretty-quick to make sure everything is formatted the way prettier
 likes it. why? because this is where eslint spends its time:

  ```
  Rule                              | Time (ms) | Relative
  :---------------------------------|----------:|--------:
  prettier/prettier                 | 26058.548 |    43.4%
  import/no-duplicates              |  9835.163 |    16.4%
  import/no-named-as-default        |  9109.186 |    15.2%
  import/no-named-as-default-member |  6969.773 |    11.6%
  indent                            |   698.182 |     1.2%
  react/no-multi-comp               |   501.391 |     0.8%
  react/no-deprecated               |   391.670 |     0.7%
  react/require-render-return       |   391.670 |     0.7%
  keyword-spacing                   |   335.591 |     0.6%
  semi-spacing                      |   280.738 |     0.5%
  ```

4. we disable some import-rules. why? have a look at the timing-table afterwards:

  ```
  Rule                        | Time (ms) | Relative
  :---------------------------|----------:|--------:
  indent                      |   633.055 |     8.6%
  react/no-multi-comp         |   469.786 |     6.4%
  react/require-render-return |   382.254 |     5.2%
  react/no-deprecated         |   327.625 |     4.5%
  keyword-spacing             |   283.417 |     3.9%
  comma-style                 |   255.195 |     3.5%
  semi-spacing                |   246.565 |     3.4%
  compat/compat               |   212.421 |     2.9%
  no-unused-vars              |   187.942 |     2.6%
  react/jsx-curly-spacing     |   181.527 |     2.5%
  ```

  i think it's good to not have multiple imports and thus
  enabled the builtin but much faster 'no-duplicate-imports' instead. 

  the docs of `import/no-named-as-default-member` state that the rule is not
  needed once one has upgraded to babel 6, so the only thing we don't have
  anymore is `import/no-named-as-default`. here are its docs:

  ```
  // foo.js
  export default 'foo';
  export const bar = 'baz';

  // ...this would be valid:
  import foo from './foo.js';

  // ...and this would be reported:
  import bar from './foo.js';
  ```

  i think the rule might actually be helpful sometimes and especially for
  juniors. if you think it might be a good idea to put it back in, please do so!


GeorgiSTodorov:
@pierrebeitz , file conflict, as we merged the webpack PR.

pierrebeitz:
conflicts resolved

PR #4088, opened 2019-08-27
Fix message catalog

GeorgiSTodorov:
## Testing
1. Try to start a community package from the Catalog.
2. Verify that the confirmation looks the same as before.
3. The stop job run modal content should be still translated in Chinese.

## Trade-offs
None.

## Dependencies
None.

## Screenshots
None.


pierrebeitz:
just re-extracted on master and fixed the lingui-check.

PR #4122, opened 2019-08-30
fix: don't error on pageNotFound

pierrebeitz:
problem: navigate `http://localhost:4200/#/dashboard123` and see the app crashing.

fix: not registering things multiple times on the MountService. it looks like this was the only place we chose to dynamically register a component and relied on the header only being instantiated once.

TODO: follow up with an integration test, making sure PageNotFound is correctly rendered when a page is not found...

### DIFF
--- a/src/js/components/ClusterHeader.js
+++ b/src/js/components/ClusterHeader.js
@@ -16,6 +16,11 @@ import ClusterDropdown from "./ClusterDropdown";
 
 const METHODS_TO_BIND = ["handleItemSelect", "handleTextCopy"];
 
+MountService.MountService.registerComponent(
+  ClusterDropdown,
+  "Header:ClusterDropdown"
+);
+
 export default class ClusterHeader extends mixin(StoreMixin) {
   constructor() {
     super(...arguments);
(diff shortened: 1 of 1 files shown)


pierrebeitz:
yep, totally!

GeorgiSTodorov:
@pierrebeitz , do you want to add it or should I?

GeorgiSTodorov:
This doesn't work for the login page. Is this expected?
Screenshot from 2019-09-02 10-44-53

GeorgiSTodorov:
Cypress fails whenever I try to visit a non-existing url.

pierrebeitz:
oh boy. the issue seems complicated now. 

open questions: 
* does this also happen on master? 
* do we need to test every react-router-`path` for whether it breaks or not?
* doesn't `react-router` have a feature to just render a not-found? we seem to have a "catchall" in `routes/index.js`.

GeorgiSTodorov:
It does also happen on master. I don't know the answer to the two other questions. My own question would be if we are okay with only fixing it for some pages.

pierrebeitz:
i'd vow to merge this as i'm really sure it's a simplification over the previous state and less error prone AND file a jira to resolve the bad behaviour we're seeing here.

PR #4127, opened 2019-09-03
DCOS-57932: Allow unknown values in container

DanielMSchmidt:
## Testing

1. Create a single container service with

```json
{
  "id": "/app1",
  "instances": 1,
  "container": {
    "type": "MESOS",
    "linuxInfo": {"ipcInfo": {"mode": "PRIVATE", "shmSize": 128}}
  },
  "cpus": 0.1,
  "mem": 32,
  "cmd": "df -m /dev/shm && sleep 1000"
}
```

2. Change the form and see that it does not exist

## Trade-offs

I was unsure about testing, so I just went with a unit test for this. Do you think we need more?

### DIFF
--- a/plugins/services/src/js/reducers/serviceForm/Container.js
+++ b/plugins/services/src/js/reducers/serviceForm/Container.js
@@ -251,7 +251,7 @@ module.exports = {
     );
 
     this.internalState = newState;
-    if (type === SET && path.join(".") === "container") {
+    if (type === SET && path.join(".") === "container" && value) {
       newState = value;
     }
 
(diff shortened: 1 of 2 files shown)


plugins/services/src/js/reducers/serviceForm/Container.js
```diff
@@ -245,20 +245,23 @@ module.exports = {
...
     }
 
-    const newState = Object.assign(
+    let newState = Object.assign(
       {},
       containerJSONReducer.apply(this, [this.internalState, ...args])
     );
 
     this.internalState = newState;
+    if (type === SET && path.join(".") === "container") {
```
nLight: That should do. When switching to multi-container we're emit `SET` transaction with `value == undefined` to delete the `container` field.  This results in the exception.

```suggestion
    if (type === SET && path.join(".") === "container" && value) {
```

PR #4128, opened 2019-09-05
DCOS-58366: Nodes: Filter field - some special symbols are breaking the field

GeorgiSTodorov:
Make sure the page is not broken when entering certain symbols.

Closes https://jira.mesosphere.com/browse/DCOS-58366

## Testing
1. Go to nodes page.
2. Type "%" in the filter.
3. Verify that the page does not break.
4. Verify that the added test makes sense.

## Trade-offs
Like Slavena noticed, the field is red for some symbols. I will investigate if this is intentional.

## Dependencies
None.

### Before
Peek 2019-09-05 13-20

### After
Peek 2019-09-05 13-22


plugins/nodes/src/js/pages/nodes-overview/HostsPageContent.js
```diff
@@ -95,7 +95,7 @@ class HostsPageContent extends React.PureComponent {
     const { regions, zones } = this.state.defaultFilterData;
 
     const query = decodeURIComponent(
-      props.location.query["filterExpression"] || ""
+      encodeURIComponent(props.location.query["filterExpression"] || "")
```
nLight: Do we need to `decodeURIComponent`? Right now it looks like we're encoding and decoding right away.

GeorgiSTodorov: I think we need it, because without it we will type one thing, but we could be actually filtering for another. For example, if we enter "%", we will be filtering for "%25".

nLight: I mean we're doing `decodeURIComponent(encodeURIComponent(x))` which should be just `x`

GeorgiSTodorov: Yes, but if we don't encode it, it breaks for some symbols, like "%". And if we don't decode it, we could type "%" and it would actually be searching for "%25".

nLight: I replaced the whole thing with `const query = props.location.query["filterExpression"] || "";` and nothing broke. Am I missing a case here?

GeorgiSTodorov: Hm, I think you are right, good catch. It seems we introduced the decode a long time ago and don't need it anymore.

PR #4129, opened 2019-09-06
[GA BLOCKER] DCOS-58445: Replace the "plus icon" dropdown with two separate dropdowns

nLight:
Replacing `+` icon with suggested dropdowns for Service/Quota screens

## Testing

Please check empty states when group is empty
also top level group, 1 level, 2 levels deep

## Trade-offs

1. I had to leverage supplemental content and completely drop `addButton` prop on the header component
2. Overrode CNVS `button:hover` style in a nasty way 😅 

### After

Screenshot 2019-09-10 at 15 59 17

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,3 @@
+**/*.less
+locale/en/messages.json
+locale/zh/messages.json
--- a/package-lock.json
+++ b/package-lock.json
@@ -735,7 +735,7 @@
     "@dcos/copychars": {
       "version": "0.1.2",
       "resolved": "https://registry.npmjs.org/@dcos/copychars/-/copychars-0.1.2.tgz",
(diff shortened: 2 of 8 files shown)


GeorgiSTodorov:
I also think that it would look better without the hover effects. I also agree with them being too close. Can we introduce a single dropdown that says actions?
Снимка от 2019-09-06 20-40-47

nLight:
If we say `Actions` we can't really use the `+` icon.

/cc @izharbuendia

GeorgiSTodorov:
"Actions" was just an example. The idea was to have a dropdown that includes both the create and the edit actions.

GeorgiSTodorov:
Tests are passing now, but we still have the hovers. @mperrotti

nLight:
Updated the PR, I discovered that we have "native" actions menu so I re-used it. The hover color change for the `New` button still there I will just override it.

PR #4130, opened 2019-09-10
[GA BLOCKER] DCOS-57644: feat(services): add mesos dns to endpoints tab

Poltergeist:
This adds the mesos dns to the endpoints tab for services.

CLOSES DCOS-57644

## Testing

Create some services (SDK, single container service and multi container service). Make sure that the DNS is shown in the endpoints tab. The DNS should be constructed like this: `%name%.%schedular%.mesos`

## Trade-offs

We are only showing the logical created mesos DNS which in Pod's will only link to one of the containers. We are not showing all the mesos DNS's which can be resolved like this [here](https://docs.d2iq.com/mesosphere/dcos/1.13/networking/DNS/mesos-dns/service-naming/#discovering-the-dns-names-for-a-service)

## Dependencies

N/A

## Screenshots

image


GeorgiSTodorov:
When I run `npm run util:lingui:check` locally, it looks like it's failing because of this empty line at the bottom of the file.
Screenshot from 2019-09-11 13-32-16

PR #4132, opened 2019-09-27
Human readable quota values

nLight:
Currently we always show values in megabytes which is not very convenient to read

## Testing

1. Create a quota group with Disk / Memory > 1000 (you can overcommit the cluster asking for bigger quota that it has)
2. Launch a service that consumes the quota in the group
3. check if two places are looking good

## Screenshots

**Before:**
Screenshot 2019-09-27 at 14 30 14
Screenshot 2019-09-27 at 14 29 58

**After:**
Screenshot 2019-09-27 at 14 27 38
Screenshot 2019-09-27 at 14 26 59


GeorgiSTodorov:
@nLight , shall we create a bump or are there other changes we wish to include?

nLight:
@GeorgiSTodorov I think that would be it for 1.14 GA please create a bump, thank you! 🙇

PR #4133, opened 2019-10-10
Update CLI URL in UI installation instructions

bamarni:
Starting with DC/OS 2.0 we're confident that the CLI is now cluster-version agnostic.

As such we can stick to the `latest` tagged version.

https://jira.mesosphere.com/browse/DCOS-59799

### DIFF
--- a/src/js/components/modals/CliInstallModal.js
+++ b/src/js/components/modals/CliInstallModal.js
@@ -69,8 +69,9 @@ class CliInstallModal extends React.Component {
 
     // Starting from DC/OS >=2.0, users can stick to the "latest" CLI.
     // However on DC/OS 1.x we must use the correct CLI for the given version of DC/OS.
-    let dcosVersion = MetadataStore.parsedVersion
-    cliVersion = dcosVersion.substring(0, 2) == '1.' ? `dcos-${dcosVersion}` : "latest";
+    const dcosVersion = MetadataStore.parsedVersion;
+    const cliVersion =
+      dcosVersion.substring(0, 2) === "1." ? `dcos-${dcosVersion}` : "latest";
 
     const downloadUrl = `https://downloads.dcos.io/cli/releases/binaries/dcos/${osTypes[selectedOS]}/x86-64/${cliVersion}/dcos`;
     if (selectedOS === "Windows") {
--- a/src/js/components/modals/CliInstallModal.js
+++ b/src/js/components/modals/CliInstallModal.js
@@ -67,7 +67,12 @@ class CliInstallModal extends React.Component {
     const clusterUrl = `${protocol}://${hostname}${port}`;
     const { selectedOS } = this.state;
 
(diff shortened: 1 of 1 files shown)


pierrebeitz:
@bamarni updated the commit-message to comply with our linting-rules. looks good otherwise 👍

makkes:
@bamarni @pierrebeitz To which version of DC/OS is this change going to be rolled out?

pierrebeitz:
☝️ @nLight ?

nLight:
@bamarni we should make sure it's backward compatible as we're going to be releasing the same artefact all the way down to 1.12.x

makkes:
@nLight, so is this change backported to all currently supported versions of DC/OS? I saw you have release branches for 1.12 and 1.13 as well?!

nLight:
@makkes those shall be abandoned. Our master shall be backwards compatible.

bamarni:
I see, how can we apply it only to >=2.0 then? Would it be by parsing the DC/OS version at runtime?

nLight:
@bamarni we could use the trick we had before please check https://github.com/dcos/dcos-ui/commit/810b475b3cac7d81659d5892d14549c77b981c68#diff-64630c1b266f4f5896bc16b1ae18bb17

bamarni:
@nLight : thanks for the tip, I've updated it however I was not able to test it locally. I couldn't run the UI locally with docker compose (following https://github.com/dcos/dcos-ui#development-with-docker). On `docker-compose up -d`, the `toolchain_1` container just exits with 0 in my case.

src/js/components/modals/CliInstallModal.js
```diff
@@ -67,7 +67,12 @@ class CliInstallModal extends React.Component {
     const clusterUrl = `${protocol}://${hostname}${port}`;
     const { selectedOS } = this.state;
 
-    const downloadUrl = `https://downloads.dcos.io/binaries/cli/${osTypes[selectedOS]}/x86-64/dcos-1.13/dcos`;
+    // Starting from DC/OS >=2.0, users can stick to the "latest" CLI.
+    // However on DC/OS 1.x we must use the correct CLI for the given version of DC/OS.
+    let dcosVersion = MetadataStore.parsedVersion
+    cliVersion = dcosVersion.startsWith('1.') ? `dcos-${dcosVersion}` : "latest";
```
makkes: I don't know whether the UI provides a polyfill for `startsWith`, otherwise this will not work in IE: https://caniuse.com/#search=startswith Oldie but goldie.

As a workaround you can use `substring`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substring

bamarni: Updated

makkes:
@bamarni, linter problem: all of the PR's commit messages must adhere to the `type: subject` format. 😩

PR #4140, opened 2019-10-30
DCOS-60342, DCOS-60344: Add refresh rate configuration

Poltergeist:
## Testing

Go to ui Settings, 
Reload the page.
Open network tab.
Open services page.

## Trade-offs

- You have to hard reload the page to use the new refresh rate on streams

## Dependencies

n/a

## Screenshots

n/a

CLOSES DCOS-60342
CLOSES DCOS-60344


src/js/pages/settings/UISettingsPage.js
```diff
@@ -83,8 +101,60 @@ class UISettingsPage extends React.Component {
                 }
               />
             </ConfigurationMapSection>
+            <ConfigurationMapSection>
+              <ConfigurationRow
+                keyValue="refreshRate"
+                title={<Trans>Refresh Rate</Trans>}
+                value={`${Config.getRefreshRate() / 1000} ${i18nMark(
+                  t`seconds`
```
GeorgiSTodorov: Maybe in some languages, we don't say seconds after the number.

GeorgiSTodorov:
It looks like this to me:

src/js/pages/settings/UISettingsPage.js
```diff
@@ -83,8 +101,60 @@ class UISettingsPage extends React.Component {
...
+          definition={[
+            {
+              fieldType: "select",
+              label: i18nMark(t`Select your Refresh Rate`),
+              showLabel: true,
+              options: [2, 5, 10, 15, 30, 60].map(refreshRate => ({
+                html: `${refreshRate} ${i18nMark(t`seconds`)}`,
+                id: `${refreshRate}seconds`
+              })),
+              value: `${Config.getRefreshRate() / 1000}seconds`,
```
nLight: we gonna need a space here

Poltergeist:
Investigating ... Thanks and sorry for not testing the i18n part ...

src/js/pages/settings/UISettingsPage.js
nLight: that's ok

Poltergeist: Is there an example somewhere for the units being before or being after the value?

GeorgiSTodorov: @Poltergeist , in English there are after the value, in [Shona](https://en.wikipedia.org/wiki/Shona_language) (one example I found, there could be other), they are before - https://translate.google.com/?hl=en#view=home&op=translate&sl=en&tl=sn&text=five%20seconds%0Aten%20seconds .

nLight: that's not important. Let's keep it like it is

nLight:
@Poltergeist It should take effect right after selecting the new value. If we need to refresh the page to achieve it please refresh the page forcibly.
Right now I'm selecting the new value and it keeps calling the API every two seconds even when I navigate to other pages. Only reloading the browser helps

src/js/pages/settings/UISettingsPage.js
Poltergeist: I fixed the issue. anyways. I was looking for a code example to make this possible. I found one.

locale/en/messages.json
```diff
@@ -1019,6 +1022,7 @@
   "{0} must be bigger than or equal to 0": "",
   "{0} of {1}": "",
   "{0} of {1} Cores": "",
+  "{0} seconds": "",
```
GeorgiSTodorov: Why do we have once this and once https://github.com/dcos/dcos-ui/pull/4140/files#diff-3efbfb8074a9a05f9fbe43a2ce435a50R1034 ?

src/js/pages/settings/UISettingsPage.js
```diff
@@ -83,8 +100,58 @@ class UISettingsPage extends React.Component {
...
+              className: "button button-primary",
+              isSubmit: true
+            }
+          ]}
+          definition={[
+            {
+              fieldType: "select",
+              label: i18n._(t`Select your Refresh Rate`),
+              showLabel: true,
+              options: [2, 5, 10, 15, 30, 60].map(refreshRate => ({
```
GeorgiSTodorov: Can we add a const for these so it's clear where they come from?

locale/en/messages.json
nLight: I believe because in once case we use named variable and in another case an expression? https://github.com/dcos/dcos-ui/pull/4140/files#diff-9f9e1d6769448e9c866a1ab4b7b664ebR139 @Poltergeist we might want to converge to one version

pierrebeitz:
the services-tab seems to have a minimum wait time of the chosen value.

pierrebeitz:
It looks like we have a race condition which makes for a waiting time of a little over the chosen value in 80% of my service tree visits.

pierrebeitz:
i added an ugly fix that rerenders the whole ServicesContainer in case it is in loading-state and DCOSStore has new data. this is necessary as the serviceTree is not passed into our component but retrieved during `render`...

src/js/pages/settings/UISettingsPage.js
```diff
@@ -83,8 +100,58 @@ class UISettingsPage extends React.Component {
                 }
               />
             </ConfigurationMapSection>
+            <ConfigurationMapSection>
+              <ConfigurationRow
```
mperrotti: Non-blocking suggestion: can we move this into the same `ConfigurationMapSection` as the Language setting? I think it looks a little nicer that way.

Screen Shot 2019-11-06 at 2 17 28 PM

PR #4141, opened 2019-11-04
Backport 1.13 refresh rate feature / fix.

Poltergeist:
This is a Backport of #4140 

## Testing

Go to ui Settings, 
Reload the page.
Open network tab.
Open services page.

## Trade-offs

~- You have to hard reload the page to use the new refresh rate on streams~

## Dependencies

n/a

## Screenshots

n/a


GeorgiSTodorov:
It looks like tests are actually failing.

src/js/stores/UserSettingsStore.ts
```diff
@@ -2,19 +2,23 @@ import EventEmitter from "events";
 
 import LocalStorageUtil from "../utils/LocalStorageUtil";
 
-const LOCAL_STORAGE_KEY = "dcosUserSettings";
+const LOCAL_STORAGE_KEY: string = "dcosUserSettings";
+const REFRESH_RATE_SETTING: string = "RefreshRate";
```
GeorgiSTodorov: I think these two don't need types.

src/js/stores/UserSettingsStore.ts
```diff
@@ -2,19 +2,23 @@ import EventEmitter from "events";
...
+    } catch (e) {
+      return null;
+    }
   }
+  return null;
 }
 
 class UserSettingsStore extends EventEmitter {
-  getKey(key) {
+  getKey(key: string): unknown {
```
GeorgiSTodorov: Is the return type really `unknown`?

Poltergeist: Yes since it is a catchall function it can basically return anything.

GeorgiSTodorov:
It works, I will approve once the tests are fixed (the easiest way is by changing the exports).

PR #4142, opened 2019-11-07
backport: fix: don't let users wait `refreshRate` before rendering services

Poltergeist:
## Testing

Go to ui Settings, 
Reload the page.
Open network tab.
Open services page.

If you select the 60 sec. setting make sure that the service page or any other page loads in a reasonable time and not only 60 sec. later.


nLight:
Backport, CI is green

PR #4165, opened 2019-11-21
chore: remove fake page-load-delay

pierrebeitz:
it could not be seen on most installations anyway anymore as the delay has
gradually been tweaked down to one second.


pierrebeitz:
FYI

PR #4181, opened 2019-11-29
chore: remove unused dependency

pierrebeitz:


pierrebeitz:
it's in use in `scripts/build-validate`

PR #4188, opened 2019-11-29
[WIP] protobuf types

pierrebeitz:
# ⚠️ DRAFT-PR FOR DISCUSSION!

# DO NOT MERGE!

we're adding the static types here that protobuf generates for the given protobuf-files.
unfortunately we rely on `toObject` currently to convert enums to readable strings, which leaves us with `Record<string, any>`s... no gain here. Of course we could make use of some interfaces from within that file, but i think we're better of converting the structure we get from Mesos to something we actually want to work with (as we currently kind of do with `parsers`). I imagine something like the [Decoders](https://github.com/mesosphere/kommander/blob/72eb950d7951411e35dd39851909d66115c1fe0c/packages/decode/src/index.ts) we use in Kommander to to this in a typesafe manner.

WDYT?

--------------

To be clear: i still think it might be a good idea to add the protos to the repo and add something like the `npm run gen:protobuf`-task introduced here.


pierrebeitz:
will come up with something that dynamically builds those files later on.

PR #4211, opened 2019-12-11
chore: use chrome on CI

pierrebeitz:
locally chrome performs about 30%-40% faster than electron. let's give it a try!


pierrebeitz:
no... it's not as easy. times seem to be about the same and chrome introduces trouble regarding our DSL-filters (which happen to be real bugs in current chrome versions :( )

PR #4253, opened 2019-12-19
TYPESCRIPT!

pierrebeitz:
This converts most code to TypeScript.

please make sure to review commit-by-commit and to read the commit messages.


jest/setupEnv.js
```diff
@@ -1,15 +1,29 @@
...
+// the second problem is that ts-jest claims to be multistage (tsc -> babel), but it seems that it somehow does not respect the "target", so
+// the output of tsc will mess with the imports so that "@lingui/babel-preset-react" does not pick them up to transform them.
+// This means that we'll get a "can't find macro_1" when running tests without this workaround for now.
+// It might be worthwhile to dig into ts-jest as to why it's transpiling ts-files differently.
+global.macro_1 = {
+  DateFormat,
+  Plural: ({ one }) => one,
+  Trans,
+  t: (...args) => args.join("")
+};
```
pierrebeitz: this is a workaround until we got rid of all lingui-macros.

lingui explicitly supports usage without macros. that way we should be able to build jest- and webpack-build-pipelines that solely depend on typescript.

package.json
```diff
@@ -142,6 +143,7 @@
     "eslint-plugin-react-hooks": "1.7.0",
     "eslint-plugin-test-names": "1.0.1",
     "esprima": "4.0.1",
+    "execa": "4.0.0",
```
pierrebeitz: somehow the stdout of the "normal" exec from `child_process` was giving inconsistent and incomplete results. i did not investigate as to why, but fell back on this lib for now.

package.json
```diff
@@ -226,6 +228,8 @@
     "test:validate": "PLUGINS_PATH=\"${npm_config_externalplugins}\" ./scripts/validate-tests",
     "test:watch": "npm test -- --watch",
     "test": "./scripts/lingui-compile && jest --config=jest.config.js --no-cache",
+    "typecheck": "jest --config=jest.config.js typecheck --no-cache",
+    "updateTSSnapshots": "jest typecheck -u; jest tslint -u",
```
pierrebeitz: keep that in mind. you'll use it quite a lot in the next couple weeks!

src/js/__tests__/helper.ts
```diff
@@ -15,9 +15,20 @@ export const getOutput = (cmd: string, params: string[]) =>
...
+            // inlined mid-term anyway
+            !line.includes("plugins-ee") &&
+            // Those are included if plugins-ee is around.
+            // Ignoring those here is a quickfix, to evade trouble between
+            // machines and CI.
+            !line.includes("CheckboxTable") &&
+            !line.includes("ServiceAccountForm") &&
+            !line.includes("TextInput") &&
+            !line.includes("Typeahead") &&
+            !line.includes("ServiceAccountFormModal")
```
pierrebeitz: That means that this stuff should rather be in EE as only EE uses it and it does not need to be bundled in OSS. Not "fixing" this as we might integrate ee-plugins anyways.

PR #4254, opened 2019-12-19
feat(refreshratee): add refreshrate to the header

Poltergeist:
## Testing

Change you refresh rate setting to be not the default value of `2 seconds`. Ensure it is only showing if that applies.

## Screenshots

image

### DIFF
--- a/src/js/components/HeaderBar.tsx
+++ b/src/js/components/HeaderBar.tsx
@@ -1,21 +1,45 @@
 import * as React from "react";
 import { Link } from "react-router";
 import { MountService } from "foundation-ui";
+import { Tooltip } from "@dcos/ui-kit";
+
+import { Trans } from "@lingui/macro";
 
 import ClusterHeader from "./ClusterHeader";
 import SidebarToggle from "./SidebarToggle";
+import Config from "../config/Config";
 
-export default () => (
-  <div className="header-bar">
-    <div className="header-bar-left-align-wrapper">
-      <SidebarToggle />
-      <Link to="/dashboard" className="header-bar-logo-wrapper">
-        <span className="header-bar-logo" />
-      </Link>
+export default () => {
+  const refreshRate = Config.getRefreshRate() / 1000;
+  const refreshRateHint =
+    Config.getRefreshRate() !== Config.defaultRefreshRate ? (
+      <span>
+        <Tooltip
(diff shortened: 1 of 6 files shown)


src/js/components/HeaderBar.tsx
```diff
@@ -18,6 +21,19 @@ export default class HeaderBar extends React.Component {
           </Link>
         </div>
         <div className="header-bar-right-align-wrapper">
+          {refreshRate > 2 ? (
```
pierrebeitz: if we want to show that if we're deviating from the default, why don't we express that here?

something like `currentRefreshRate != defaultRefreshRate`?

src/js/components/HeaderBar.tsx
```diff
@@ -18,6 +21,19 @@ export default class HeaderBar extends React.Component {
...
+            <span>
+              <Tooltip
+                id="refreshRate"
+                trigger={`Refresh Rate: ${refreshRate} sec`}
+                maxWidth={150}
+              >
+                Data will only be requested every {refreshRate} seconds per
+                endpoint
```
pierrebeitz: let's remove the "per endpoint" here as users most likely won't care and it also might not be true if we throttle repainting on some pages.

pierrebeitz:
btw: what's the PR-title and commit message about?

Poltergeist:
I messed up the commit I will create a new commit soon sorry :-( and therefore the title.

src/js/components/HeaderBar.tsx
```diff
@@ -1,14 +1,17 @@
 import * as React from "react";
 import { Link } from "react-router";
 import { MountService } from "foundation-ui";
+import { Tooltip } from "@dcos/ui-kit";
 
 import ClusterHeader from "./ClusterHeader";
 import SidebarToggle from "./SidebarToggle";
+import Config from "../config/config";
```
pierrebeitz: this path does not work on all unixes. your mac seems to handle paths case insensitive. you'll want to use `"../config/Config"`.

pierrebeitz:
@Poltergeist added a fixup-commit (so please rebase before merging in case you like it).

pierrebeitz:
ping @Poltergeist

Poltergeist:
pong 🎾

PR #4314, opened 2020-01-02
chore: remove all direct usages of react-transition-group

pierrebeitz:


pierrebeitz:
wow... we need to get rid of `mesosphere-reactjs` before we can drop the dep.

PR #4342, opened 2020-01-11
refactor VirtualNetworksStore

pierrebeitz:
this is a rather big and unwieldy commit, sorry!

we're
* removing references to unused data (things like `vtep_mac_oui`, `vtep_subnet`)
* removing `OverlayList` and the convolutions around it in favor of a brand new
`Overlay`-type that has the idea of implementing a "decoder" that builds
something of type `Overlay` from an API-Response.
* pulling out methods from classes that can actually be functions. this narrows
interfaces and is a step on the way to function components.
* making use of optional chaining if it boils down code.
* making sure the overlay-type flows to everywhere an Overlay is used in a
component.
* address some easy-to-fix type-issues along the way.

it actually started to prepare for integrating a calico-network-type in the
service-form. as we're currently in the process of massively refactoring the
code of dcos-ui in order to become faster on future features (and keep our
sanity) it evolved to the start of a POC on how to get rid of our current
store-architecture. that will be handled in the next commit.


pierrebeitz:
@Poltergeist (maybe also @orlandohohmeier) i think it makes sense to do a pair-review on this to share some understanding regarding some patterns i find either useful or more maintainable.

PR #4346, opened 2020-01-13
DCOS-62535 support calico networking in service forms

pierrebeitz:
👉 https://jira.mesosphere.com/browse/DCOS-62535

adds a select-option for calico for single- and multicontainer-forms.

OS Enterprise 2020-01-13 15-30-03

### DIFF
--- a/plugins/services/src/js/components/forms/MultiContainerNetworkingFormSection.tsx
+++ b/plugins/services/src/js/components/forms/MultiContainerNetworkingFormSection.tsx
@@ -546,25 +546,13 @@ class MultiContainerNetworkingFormSection extends mixin(StoreMixin) {
   }
 
   getTypeSelections() {
-    const networkType = findNestedPropertyInObject(
-      this.props.data,
-      "networks.0.mode"
-    );
-    const networkName = findNestedPropertyInObject(
-      this.props.data,
-      "networks.0.name"
-    );
-
-    let network = networkType;
-    if (networkName) {
-      network = `${networkType}.${networkName}`;
-    }
+    const { mode, name } = this.props.data?.networks?.[0] || {};
+    const selectedValue = name ? `${mode}.${name}` : mode;
 
     return (
-      <FieldSelect name="networks.0" value={network}>
-        <Trans key="host" render={<option value={HOST} key="host" />}>
-          Host
-        </Trans>
+      <FieldSelect name="networks.0" value={selectedValue}>
+        <Trans key="host" id="Host" render={<option value={HOST} />} />
+        <option value={`${CONTAINER}.calico`}>Virtual Network: Calico</option>
         {getVirtualNetworks()}
       </FieldSelect>
     );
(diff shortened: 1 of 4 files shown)


plugins/services/src/js/components/forms/MultiContainerNetworkingFormSection.tsx
```diff
@@ -546,25 +546,13 @@ class MultiContainerNetworkingFormSection extends mixin(StoreMixin) {
...
+    const selectedValue = name ? `${mode}.${name}` : mode;
 
     return (
-      <FieldSelect name="networks.0" value={network}>
-        <Trans key="host" render={<option value={HOST} key="host" />}>
-          Host
-        </Trans>
+      <FieldSelect name="networks.0" value={selectedValue}>
+        <Trans key="host" id="Host" render={<option value={HOST} />} />
+        <option value={`${CONTAINER}.calico`}>Calico</option>
```
mainred: I am afraid we may treat `calico` as one of the virtual networks as, the same as `dcos`

plugins/services/src/js/components/forms/NetworkingFormSection.tsx
```diff
@@ -554,6 +554,7 @@ class NetworkingFormSection extends mixin(StoreMixin) {
       <FieldSelect name="networks.0.network" value={selectedValue}>
         <Trans key="host" id="Host" render={<option value={HOST} />} />
         <Trans key="bridge" id="Bridge" render={<option value={BRIDGE} />} />
+        <option value={`${CONTAINER}.calico`}>Calico</option>
```
mainred: ditto

plugins/services/src/js/components/forms/MultiContainerNetworkingFormSection.tsx
pierrebeitz: why are you afraid of this? what are the consequences?

the thing i made sure of is that the form-mechanics are able to add (and identify) the json you mentioned in the JIRA - so one can choose it via a select-input. what's missing?

mainred: Let me elaborate on this.
Throughout the existing DC/OS documentation, a virtual network is a kind of network implemented by CNI or CNM, 

_NOTE: The network tab currently displays information about containers that are associated with virtual networks managed by DC/OS Overlay. It does not have information about containers running on virtual networks managed by any other CNI/CNM provider`._
from https://docs.d2iq.com/mesosphere/dcos/1.13/networking/SDN/dcos-overlay/#limitations

_Using CNI and CNM, DC/OS is able to plumb containers onto any virtual network defined by an SDN provider that supports the CNI or CNM standard_
from https://docs.d2iq.com/mesosphere/dcos/1.13/networking/#container-mode-networking

Here, `calico` is a network based on CNI, so it should be treated as a virtual network, and show up in the select-input as `Virtual Network: calico`.

CC @orlandohohmeier

pierrebeitz: thanks for the response. that was exactly what i needed to understand the issue.

i think you're right and will change the wording here. 👍

PR #4353, opened 2020-01-15
refactor virtualNetworksStore

pierrebeitz:
to be able to get rid of stores this serves as a first exercise. unfortuntately
this store was a bit special as it was only ever used to request data from a
server (the polling in the background turned out to be there only to retry a
couple of times in case the server does not respond right away), so that we did
not need to introduce rxjs-polling-magic here.

nonetheless a couple patterns arose, the data-flow is much more clear and some
type errors have been fixed.

I think we could develop a migration path away from stores that might go
something like this:

1. get rid of the AppDispatcher
  It's used to handle "Server Actions" but has been extended to also
  handle sidebar actions and a special case for the language modal - both of
  which should be easily replaceable by more straightforward solutions.

  After implementing those, we should be able to just remove everything related
  to ServerActions as our action-identifiers are unique throughout the
  application and this type has only been used to safeguard stores from
  consuming messages that are not meant for them but other components.

2. extend all actions with a onSuccess / onError-handle that immediately invokes
  the callback. request-consumers can use those right away (it looks like
  `suppressUpdate: true` is an indicator for those).

3. pass the data along the emitted events.
  we currently have a pattern to emit `WHATEVER_SUCCESS` only to invoke a
  `on${STORE}store${ACTION}handler` that in the best case does alter some state
  and in the worst case is empty. if it's empty the storeMixin still might
  invoke a `forceUpdate` on the component in order to grab some data or kick of
  some request out of its `render` method... yes, you read that correct.

  by passing the data along we can then wade through all the handlers which now
  get some data. often times that data will be the data we actually want to
  render some new stuff or update some state explicitly. this will very likely
  make some flags you can pass to `store_listeners` obsolete because we now
  don't need to rely on the `forceUpdate` anymore (apart from the code being
  much more read- and maintainable because of the explicit and more
  react-idiomatic dataflow). we should be able to get rid of quite some stores
  at this point in time.

4. implement polling stores as rxjs-streams.
  maybe incorporate
  gist.github.com/marinho/3637210b13c0f298e1692a0b7b104e64.
  maybe it's also worth to do this very early in order to be able to use the
  solution in all the stores that are about to be replaced.

TIL i learned about a pattern i like to reduce the constructor-noise:

1. setting state without a constructor:

```
class Bla extends React.Component {
  // look ma! so clean!
  state = { initial: "value" }

  render() {
    <div>Works: {this.state.initial}</div>
  }
}
```


plugins/services/src/js/components/forms/MultiContainerNetworkingFormSection.tsx
```diff
@@ -26,40 +24,61 @@ import FormRow from "#SRC/js/components/form/FormRow";
...
-    this.forceUpdate();
+  state: {
+    virtualNetworks: Overlay[];
+  } = {
+    virtualNetworks: []
   };
 
+  componentDidMount() {
+    VirtualNetworksActions.fetch(virtualNetworks => {
+      this.setState({ virtualNetworks });
```
Poltergeist: Couldn't we do the `getVirtualNetworks` part here as well? I don't know if this would add performance but it would decrease the confusion that `this.state.virtualNeworks` is only used in the `getVirtualNetworks` function. Or am I missing something?

pierrebeitz: that's a great remark.

`setState` explicitly tells us that the component will rerender with a new internal state. I prefer to only put data into state (and not the `Trans`-components that would be returned by `getVirtualNetworks`) because the type of the state will be more intention-revealing/documenting.

we'll likely have under 10000 options (it should actually be under 10 most of the time, right?) in that select - so i think performance is not something to be concerned about.

local state is used for a single rendering-purpose in a lot of components in dcos-ui. do you think this is confusing and/or should be considered a bad practice?

pierrebeitz:
most of this stuff is merged by now. postponing the virtualNetwork-Stuff as it's not as important as e.g. the ServicesTable.

PR #4394, opened 2020-01-26
chore: update JIRA API URL

juliangieseke:
JIRA Was moved to new URL, old links seem to be valid, but some housekeeping cant be bad :)


pierrebeitz:
nice, thanks!

PR #4399, opened 2020-01-28
DCOS-62251 add reservations to service table

Poltergeist:
This introduces the virtual bursting information to the services table.

### DIFF
--- a/plugins/services/src/js/columns/ServicesTableMemColumn.tsx
+++ b/plugins/services/src/js/columns/ServicesTableMemColumn.tsx
@@ -1,5 +1,6 @@
 import * as React from "react";
-import { NumberCell } from "@dcos/ui-kit";
+import { Trans } from "@lingui/macro";
+import { NumberCell, Tooltip } from "@dcos/ui-kit";
 
 import Pod from "../structs/Pod";
 import Service from "../structs/Service";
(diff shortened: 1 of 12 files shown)


plugins/services/src/js/columns/ServicesTableMemColumn.tsx
```diff
@@ -1,19 +1,60 @@
...
+);
 
 export function memRenderer(
   service: Service | Pod | ServiceTree
 ): React.ReactNode {
-  return <ServiceMem resource={service.getResources()[`mem`]} />;
+  return (
+    <ServiceMem
+      id={service.getId()}
+      resource={service.getResources()[`mem`]}
```
pierrebeitz: can we simplify this?

``service.getResources()[`mem`] `` => `service.getResources().mem`?

plugins/services/src/js/columns/ServicesTableMemColumn.tsx
```diff
@@ -1,19 +1,60 @@
...
+export const ServiceMem = React.memo(
+  ({
+    resource,
+    limit
+  }: {
+    resource: string;
+    limit: number | string | null;
+    id: string;
+  }) => {
+    if (limit != null && limit !== 0) {
```
pierrebeitz: what's the exact thing we need to do here?

do we want to handle empty strings? do we only bail out in case of an exact `null`?

plugins/services/src/js/columns/ServicesTableMemColumn.tsx
Poltergeist: ✅

PR #4414, opened 2020-01-29
chore: remove cruft around our test-setup

pierrebeitz:
the major motivation behind this was to understand our system- and integration-test-setup better in order to be able to run the system-tests for ee and oss in parallel at some point in time.

:tada: the major benefit of these changes is that we're now able to easily run system-tests locally (try `npm run test:system:local` - it's not perfect but now way easier to get going!)

1. a `cypress.system-tests.json` is introduced (we had something similar before in plugins-ee, but that was never used) and used to account for the different configuration of system-tests instead of passing a config directly via `cypress --config`.

2. a `load-auth-env-vars`-script is added. you can source that in order to obtain all the env vars you need to talk to a cluster. the exact mechanism may be replaced by using `dcos` itself at some point in time.

3. the `test:*:local`-tasks have been replaced by usages of `cypress open` which makes it quite a bit more handy to run system-tests locally.

4. debugging the scripts used by CI should be simpler now as some complexity and indirection has been removed. it now automatically detects whether a proxy is running and uses that instead of starting an `http-server` on dist - no need to specify a flag anymore. i think of replacing all of the machinery with a script that has a CI-mode and asks questions if ran locally.

# later todos

* figure out how to make the dcos-cli command aware of separate clusters being orchestrated from the same machine. `dcos cluster setup` won't work as it takes ages to run and makes us vulnerable to race conditions.


pierrebeitz:
@orlandohohmeier, @Poltergeist: i'm not really requesting a thorough review here but rather want you to stray though the changes in case you're familiar with that code.
if you're not, please have a look at the `scripts`-section in `package.json` to internalize that we now can run system-tests locally quite easy.

PR #4433, opened 2020-02-04
chore: run unit tests in parallel to system/integration tests

pierrebeitz:
parallelity level 9000!


pierrebeitz:
unfortunately cypress and jest don't want to run at the same time... :(

a memory-problem?

PR #4568, opened 2020-04-01
use an rxjs stream for /dcos-metadata/dcos-version.json

pierrebeitz:
we want to flag features by the underlying version of DCOS. to understand the
flow of data (and to decoupled a lot of things) we switch to using an
rxjs-stream for the route that tells us details about the cluster we're running
on.


src/js/stores/dcos-version.ts
```diff
@@ -0,0 +1,55 @@
...
+            description: `An error occurred while fetching ${Config.rootUrl}/dcos-metadata/dcos-version.json. Here's the internal error: ${e}`,
+          })
+        );
+      }),
+      // retry every 10 seconds
+      delay(10000)
+    )
+  ),
+  shareReplay<ClusterData>(1)
+);
```
pierrebeitz, line 55: @orlandohohmeier review requested mainly for this part. 

we're trying forever with a delay of 10 seconds, show a toast in case something failed and `shareReplay` the result amongst subscribers.

Now that i see the stuff above i realize i forgot to i18n'ize. Will do that soon. 😊

orlandohohmeier: 🤔 What do you expect users do with the information?

pierrebeitz: File better COPS issues...

pierrebeitz: would merging be ok with you or do we want to do something different in the case of an error. till now every page would have remained in a loading state  in case that endpoint was not available. so showing an error that might be confusing or too detailed seems still better than not doing anything at all, right?

orlandohohmeier: As discussed offline, let's adjust this to apply capped exponential back-off plus a a message after about 5s to not block interaction any longer then needed while waiting for the response.

PR #4631, opened 2020-04-21
fix: show proper error message if secret store is sealed

pierrebeitz:
this fixes a bug when trying to view secrets on a DC/OS with a sealed store.

the alert has been present before, but now you can actually see it:

OS Enterprise 2020-04-22 14-04-18

to reproduce you can spin up a new instance and run the commands from here to seal the store: https://docs.d2iq.com/mesosphere/dcos/1.13/security/ent/secrets/seal-store/

afterwards visit the page to see the above screen.

---------------

the actual change is quite small and i can isolate it into a single commit if you think that's worthwhile. having a look at the surroundings i cleaned up a bit of stuff - mainly removing the structs as we did not need them and they computed conditions in a suboptimal way. 

----------------

if you think that's worthwhile, we can pair on reviewing this, as quite some stuff happened to the codebase since you last worked with it, i assume.


plugins/secrets/pages/SecretsPage.tsx
```diff
@@ -220,6 +243,16 @@ class SecretsPage extends mixin(StoreMixin) {
       selectedAction,
     } = this.state;
 
+    // the list-request will fail, thus this needs to be before the SecretsError-component
+    if (SecretStore.getStores().some((s) => s.sealed)) {
+      return (
+        <Page>
```
pierrebeitz, line 249: moving this section up a couple lines is the fix. everything else is trimming down complexity (mainly removing structs in favor of static typing).

PR #4660, opened 2020-05-14
fix: calculation of resource limits for pods/service table

pierrebeitz:
due to the former implementation calculating resource limits in a couple places
there were inconsistencies in display.
we now calculate and display the allocated resource along with its aggregated
limit (as we formerly ignored the number of instances in one place and did not
properly multiply it in another case...).

also i could not help it and added a couple signatures to reduce the number of
type errors.

closes D2IQ-66986


DominikDary:
@pierrebeitz do you have the permissions to merge this?

pierrebeitz:
sure!

PR #4662, opened 2020-05-15
chore: try to fix ci

pierrebeitz:
progress: 
- seems to make things better... the actual error is gone and tests now seem to be running somewhat.
- dcos-cli tries to download `https://downloads.dcos.io/cli/testing/plugins/dcos-core-cli/linux/x86-64/dcos-core-cli-2.2-patch.x.zip` which gives access denied. there's a 2.1 version though 🤔 
- tried to convince CLI's CI to place according releases to the places dcos-cli is looking for them.


pierrebeitz:
fixed CI by fixing CLI's CI! 🎉

PR #4670, opened 2020-05-18
fix: add no-cache to dcos-version request

pierrebeitz:
we used to add a `?ts=UNIX_TIMESTAMP` to every request that's routed through
request-util. by changing implementation to use rxjs we removed that and now had
an instance of a caching-issue while updating a soak-cluster.

to fix the issue, we add a no-cache in this place (which is the only place for
now) and will extract a new request-utility to handle cache-invalidation for all
requests.

### DIFF
--- a/src/js/stores/dcos-version.ts
+++ b/src/js/stores/dcos-version.ts
@@ -22,9 +22,9 @@ type DCOSData = {
   hasQuotaSupport: boolean;
 };
 
-export default fromFetch(
-  `${Config.rootUrl}/dcos-metadata/dcos-version.json`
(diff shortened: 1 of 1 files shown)


fatz:
wouldn't this be something we rather want to do on admin router site? So always apply no-cache to potentially dynamic resources?

pierrebeitz:
you're totally right. 🤔 

actually i could not reproduce the issues you faced on soak when trying to manipulate the `dcos-version.json` directly on a master, as nginx already sends E-Tags. i'll try again next week (after PTO).

PR #4706, opened 2020-06-03
D2IQ-69020 update docs-links

pierrebeitz:
https://jira.d2iq.com/browse/D2IQ-69020

### DIFF
--- a/plugins/services/src/js/containers/pod-debug/PodDebugContainer.tsx
+++ b/plugins/services/src/js/containers/pod-debug/PodDebugContainer.tsx
@@ -223,7 +223,7 @@ class PodDebugTabView extends React.Component {
       primaryAction = (
         <Trans render="span">
           <a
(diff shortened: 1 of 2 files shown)


plugins/services/src/js/containers/pod-debug/PodDebugContainer.tsx
```diff
@@ -223,7 +223,7 @@ class PodDebugTabView extends React.Component {
       primaryAction = (
         <Trans render="span">
           <a
-            href={MetadataStore.buildDocsURI("/services#service-status")}
+            href={MetadataStore.buildDocsURI("/gui/services#service-status")}
```
janisz, line 226: Can you explain why we need this change?

pierrebeitz: because the docs-team seems to have rearranged the structure without setting up redirects.

PR #4711, opened 2020-06-03
fix: actually hide cpu limits removed via the forms

pierrebeitz:
if you deleted a cpu limit via the service form it would actually be removed
from the JSON but the value cached in the form state would still re-appear
again.

i consider this an ugly, yet the only feasible (due to time constraints)
solution as we don't have a proper representation of the form within the "state"
(which is called a "Batch" holding "Transactions"). thus we now
figure out that while reducing a new state, we're at a "Transaction" that represents unchecking the "unlimited" checkbox (that's what
`value === false` represents). only in this case we'll now reintroduce the value
that's been stored in the state - which has formerly been put in whenever the parsing function did not succeed - like when you want to delete a value from the form.

what i actually want to say here is that our forms are accidentally complex -
and i mean as complex as having to follow the dataflow through 8 files upwards
(while evaluating some conditionals in order to figure out what file to go
through next) to find this place which is only responsible to calculate the
value to be shown in the form field from the aforementioned "Batch".

closes COPS-6202

----------

here's a video of our new-won freedom to delete the cpu limit in the form: 

screencast 2020-06-03 20-29-21

much wow!


pierrebeitz:
note that CI is actually passing (but with a flake)

PR #4740, opened 2020-06-11
COPS-6139 don't make up job-groups

pierrebeitz:
https://jira.d2iq.com/browse/COPS-6139

before we render jobs, we expect our graphql-server to only give us jobs that
are considered to be within the group we're currently looking at. we're grouping
by spliting job-ids by "." - everything except for the last segment is a group,
while the last segment is the jobs "name" (in the context of this code).

unfortunately our graphql-server (yes, we have a graphql-server bundled with our
application. i don't think it's making code more maintainable btw) filtered jobs
like so:

imagine there're the jobs `foo.bar.baz.blorgh` and `foo.barbar.blorgh` and we're
currently looking at the `foo.bar` group. we're now filtering out all jobs that
don't start with that `path` (`["foo", "bar"]`). we did that with a
`startsWith`, so `foo.barbar.blorgh` was be considered to be in that group. now
the rendering takes over and displays all the "third-level" groups and jobs. in
this case we'd see `baz` (being the third-level of `foo.bar.baz.blorgh`) AND
`blorgh` (being the third-level of `foo.barbar.blorgh`), while we actually don't
want that group to show up.

in code:
```
foo.bar
foo.bar.baz.blorgh <- starts with foo.bar
foo.barbar.blorgh  <- starts with foo.bar (the error)

```

this fix seeks to compare differently:

```
foo.bar.           <- notice the dot at the end
foo.bar.baz.blorgh <- starts with foo.bar.
foo.barbar.blorgh  <- does not start with foo.bar. :woohoo:

```

--------------------

please let's follow up with removing that graphQL-server and -client at some point in time. they are mainly used in the context of jobs and are highly confusing as they add a lot of accidental complexity (that's very hard to debug).

### DIFF
--- a/plugins/jobs/src/js/types/JobConnection.ts
+++ b/plugins/jobs/src/js/types/JobConnection.ts
@@ -38,7 +38,7 @@ export function JobConnectionTypeResolver(
 
   const namespacedResponse: MetronomeJobResponse[] =
     path.length > 0
-      ? response.filter(({ id }) => id.startsWith(path.join(".")))
+      ? response.filter(({ id }) => id.startsWith(path.join(".") + "."))
       : response;
 
   const filteredResponse: MetronomeJobResponse[] =
--- a/tests/_fixtures/metronome/jobs.json
+++ b/tests/_fixtures/metronome/jobs.json
@@ -334,5 +334,108 @@
         ]
       }
     ]
+  },
+  {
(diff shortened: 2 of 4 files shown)


pierrebeitz:
@janisz sure. added one and made sure that it fails without the fix.  
thanks!

PR #4764, opened 2020-07-09
chore: enable vertical bursting system test

pierrebeitz:


system-tests/services/test-apps-cy.js
```diff
@@ -1061,35 +1063,41 @@ describe("Services", () => {
...
-        cy.root().contains(serviceName).click();
-        cy.root().contains("Edit").click();
-
-        cy.root().getFormGroupInputFor("CPUs").should("have.value", "1");
-        cy.root()
-          .getFormGroupInputFor("Memory (MiB)")
-          .should("have.value", "42");
-      });
+    describe("Vertical Bursting", () => {
+      if (compareVersions(Cypress.env("CLUSTER_VERSION"), "2.1.0") > 0) {
```
janisz, line 1067: How about inverting the condition so text below will have the same indentation?

PR #4783, opened 2020-07-24
Remove dead code

pierrebeitz:
review commit-by-commit. this adds to 0b94de1165c7d5ae17c553b1005bd0230de9444a in that the download now also works on 1.13 clusters.

### DIFF
--- a/src/js/components/modals/CliInstallModal.tsx
+++ b/src/js/components/modals/CliInstallModal.tsx
@@ -1,10 +1,8 @@
 import { Trans } from "@lingui/macro";
 import { i18nMark } from "@lingui/react";
 
-import browserInfo from "browser-info";
 import classNames from "classnames";
 import { Modal } from "reactjs-components";
-import PropTypes from "prop-types";
 import * as React from "react";
 import { Icon } from "@dcos/ui-kit";
 import { SystemIcons } from "@dcos/ui-kit/dist/packages/icons/dist/system-icons-enum";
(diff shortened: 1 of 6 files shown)


src/js/components/modals/CliInstallModal.tsx
```diff
@@ -67,13 +50,7 @@ class CliInstallModal extends React.Component {
...
     const clusterUrl = `${protocol}://${hostname}${port}`;
     const { selectedOS } = this.state;
-
-    // Starting from DC/OS >=2.0, users can stick to the "latest" CLI.
-    // However on DC/OS 1.x we must use the correct CLI for the given version of DC/OS.
-    const dcosVersion = MetadataStore.parsedVersion;
-    const cliVersion =
-      dcosVersion.substring(0, 2) === "1." ? `dcos-${dcosVersion}` : "latest";
-    const downloadUrl = `https://downloads.dcos.io/binaries/cli/${osTypes[selectedOS]}/x86-64/${cliVersion}/dcos`;
+    const downloadUrl = `https://downloads.dcos.io/binaries/cli/${osTypes[selectedOS]}/x86-64/latest/dcos`;
```
janisz, line 53: Can we do that?
```
// However on DC/OS 1.x we must use the correct CLI for the given version of DC/OS.
```

pierrebeitz: most likely we can't! thanks for catching!

PR #4831, opened 2020-08-17
chore: upgrade ui-kit

pierrebeitz:
we can make good use of the new table to overhaul the jobs-form


package.json
```diff
@@ -152,6 +152,7 @@
     "purify-css": "1.2.5",
     "raml-validator-loader": "0.1.11",
     "react-addons-test-utils": "15.6.2",
+    "react-popper": "2.2.3",
```
janisz, line 155: Why do we need new dependency?

plugins/nodes/src/js/components/NodesTable.tsx
```diff
@@ -1,10 +1,10 @@
 import * as React from "react";
 import {
-  Table,
+  Table_Deprecated,
```
janisz, line 3: I understand table was deprecated in 4.x release
Will you use non deprecated Table in the future?

package.json
pierrebeitz: it looks like we need to declare this explicitly here as else npm resolves that to an older version that does not use hooks yet and breaks.

plugins/nodes/src/js/components/NodesTable.tsx
pierrebeitz: yes! next PR will have one.

package.json
pierrebeitz: we just learned that we had a dependency that relied on popper.js 1.x, which is why i initially had to add this explicitly.

now that dependency is gone. 🎉

PR #4833, opened 2020-08-17
D2IQ-57657 COPS-4665 add cpus, mem, disk to JobsOverviewTable

pierrebeitz:
Add CPUs, Mem and Disk columns to the Jobs Table.

Here's a screenshot:

OS Enterprise 2020-08-17 18-14-48

EDIT: added GPUs as well as they were requested in COPS-4665.


pierrebeitz:
@janisz could you have a look?

system-tests/jobs/test-jobs-cy.js
```diff
@@ -1,185 +1,47 @@
...
-  it("runs, stops and deletes a job", () => {
-    // first create a simple job
-    const jobName = "delete";
-    const fullJobName = `${Cypress.env("TEST_UUID")}.${jobName}`;
-    const cmdline = "while true; do echo 'test' ; sleep 100 ; done";
-
-    // Visit jobs
-    cy.visitUrl("jobs/overview");
-    cy.get(".button.button-primary-link.button-narrow").click();
+    // ////////////////////////////////////////////////////////////////////////
```
janisz, line 7: Is this the proper way of organizing tests? Why not use `it` or `context`?

janisz:
I updated branch to check if tests pass

system-tests/jobs/test-jobs-cy.js
pierrebeitz: those are system-tests against a real cluster.

team formerly wrote those tests like unit tests, often times repeating the process or creating and deleting a service lots of times just to test e.g. whether different form fields persist their value.

having many assertions in a single run, speeds up our tests massively and also proved to reduce flakiness for integration- and system-tests.

while one could argue that we could put `it`s inbetween, i think that an `it` should be self-contained, so that one can move it freely in the according file.

pierrebeitz:
checks also passed the last time: https://jenkins.mesosphere.com/service/jenkins/blue/organizations/jenkins/frontend%2Fdcos-ui-oss-pipeline/detail/PR-4833/5/

janisz:
@pierrebeitz Can we get somebody more experienced to look at this? LGTM but I'd like to get more :eyes: on this

pierrebeitz:
@DanielMSchmidt ?

DanielMSchmidt (approved):
Looks good to me, but I haven't checked it out and tested it :)

PR #4864, opened 2020-08-29
chore: separate junit-files per env

pierrebeitz:
we're currently merging all system-test-results into a big mess that makes our
reporting within jenkins inaccurate.
this is rather a POC that i intent to make cleaner later.


pierrebeitz:
note to self: integration test reporting would be broken as is, as we're directly feeding the mochaReporter-option to cypress. but let's first see if the approach can work at all.

PR #4875, opened 2020-08-31
feat: display dependencies on job list and detail view

pierrebeitz:
this paves the way to display dependencies on the job-list and job-details view.
we currently display those and while i'm almost happy with the links in the
config-map (a.k.a. the detail view), i think we'll need to make another pass on
the list-view.

choosing to go with another icon here seemed reasonable for two reasons:
* ease of implementation: we only show this icon when a job brings dependencies,
which it only can do on a dcos >= 2.2. thus we don't need to introduce
feature-flagging here.
* we don't waste table-space in case a customer does not work with dependencies.

i'm totally open to changing this to a separate column that is only shown on
clusters >= 2.2 OR if at least one job is configured to have a dependency.

## Reviewing

in order to play around with the changes in here, you'll want to start via `npm start` without specifying a `CLUSTER_URL`. that way our new mock-server kicks in and will provide some job-fixtures that have dependencies specified as we don't have a dcos-build with the according metronome changes we need to test this end-to-end yet.

## Screenshots

Here's the "detail view".

Adding a "Dependencies" section that shows all dependencies as a bordered list. Each entry links to the according job.
detail

The list view. I'm not yet happy with this but requested help from UX. While i like the idea of "just" adding a an icon in the name-column, i think we need a different icon (@danielone please!). also i think we could change the tooltip [to a popover](https://dcos-labs.github.io/ui-kit/?path=/story/overlays-popover--default) and also display the entries as a list of links. possibly with the according state (e.g. running, pending).

list

### DIFF
--- a/plugins/jobs/src/js/JobDetailPageContainer.tsx
+++ b/plugins/jobs/src/js/JobDetailPageContainer.tsx
@@ -43,6 +43,9 @@ const getGraphQL = (id) =>
           path
           command
           schedules
+          dependencies {
+            id
+          }
           cpus
           description
           mem
--- a/plugins/jobs/src/js/components/JobsFormConfig.tsx
+++ b/plugins/jobs/src/js/components/JobsFormConfig.tsx
@@ -13,6 +13,7 @@ import ParametersConfigSection from "./config/ParametersConfigSection";
 import ScheduleConfigSection from "./config/ScheduleConfigSection";
 import LabelsConfigSection from "./config/LabelsConfigSection";
 import EnvVarConfigSection from "./config/EnvVarConfigSection";
+import DependenciesConfigSection from "./config/DependenciesConfigSection";
 
 const PRIORITIES_PAD_NUMBER = 100;
 const DEFAULT_DISPLAY_COMPONENTS = [
@@ -25,6 +26,7 @@ const DEFAULT_DISPLAY_COMPONENTS = [
       ScheduleConfigSection,
       LabelsConfigSection,
       EnvVarConfigSection,
+      DependenciesConfigSection,
     ],
   },
 ];
--- a/plugins/jobs/src/js/components/JobsOverviewTable.tsx
+++ b/plugins/jobs/src/js/components/JobsOverviewTable.tsx
@@ -3,21 +3,28 @@ import classNames from "classnames";
 import { Link } from "react-router";
 import * as React from "react";
 import { Tooltip } from "reactjs-components";
-import { designTokens as dt, Icon, Table } from "@dcos/ui-kit";
+import {
+  Tooltip as UIKitTooltip,
+  designTokens as dt,
+  Icon,
+  Table,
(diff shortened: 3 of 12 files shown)


danielone:
If job depdencies will be common, then worth a column

vinodkone:
@danielone "My understanding is the # of dependencies can be quite large". I think the number of direct dependencies for a given job wouldn't be that high (1-5). I can't imagine a user to be able to reason about dependencies if they are a higher number.

DominikDary:
@danielone can you please review this pr so that we can add this to the beta1?

janisz:
I tried to test it but It looks like metronome is not ready yet

Run on 

```
$ curl -k 'https://54.89.204.197/service/metronome/v1/jobs' -X POST --data-raw '{"id":"test","run":{ "dependencies": [{ "id": "dep" }], "cpus":1,"mem":128,"disk":0,"gpus":0,"cmd":"sleep 100","maxLaunchDelay":3600,"restart":{"policy":"NEVER"}}}'

{"message":"Object is not valid","details":[{"path":"/run","errors":["Additional properties are not allowed, but found properties 'dependencies'."]}]}% 
```

janisz:
My bad I fixed request but It's still looks not good

```
curl -s "$(dcos config show core.dcos_url)/service/metronome/v1/jobs"  -k --header "Authorization: token=$(dcos config show core.dcos_acs_token)" | jq
[
  {
    "id": "test",
    "dependencies": [
      "dep"
    ],
    "labels": {},
    "run": {
      "cpus": 1,
      "mem": 128,
      "disk": 0,
      "gpus": 0,
      "cmd": "sleep 100",
      "env": {},
      "placement": {
        "constraints": []
      },
      "artifacts": [],
      "maxLaunchDelay": 3600,
      "volumes": [],
      "restart": {
        "policy": "NEVER"
      },
      "secrets": {},
      "networks": []
    },
    "schedules": []
  },
  {
    "id": "dep",
    "dependencies": [],
    "labels": {},
    "run": {
      "cpus": 1,
      "mem": 128,
      "disk": 0,
      "gpus": 0,
      "cmd": "sleep 100",
      "env": {},
      "placement": {
        "constraints": []
      },
      "artifacts": [],
      "maxLaunchDelay": 3600,
      "volumes": [],
      "restart": {
        "policy": "NEVER"
      },
      "secrets": {},
      "networks": []
    },
    "schedules": []
  }
]
```
Screenshot from 2020-09-09 15-38-46
Screenshot from 2020-09-09 15-39-12
No tool-tip and no dependencies listed? Maybe my configuration is wrong?

danielone:
> @danielone can you please review this pr so that we can add this to the beta1?

@DominikDary I did review the PR, please see my comments above, unless there is a new iteration? If not, there are some issues to address for when @pierrebeitz is available to partner to do so.

janisz:
Tested with latest master 
image

